PR #1984, opened 2019-04-18
Automate the next release version bump [MAILPOET-1869]

wxa:
Now the `jira:release-version` command will automatically determine the next release version according to the "Required version increment" field of Jira tickets.

You can also check what the next version will be with `next:release-version`.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -594,10 +594,17 @@ class RoboFile extends \Robo\Tasks {
       ->run();
   }
 
-  public function jiraReleaseVersion($version = null, $opts = []) {
-    if ($version) {
-      $this->validateVersion($version);
+  public function nextReleaseVersion($version = null) {
+    if (!$version) {
+      $version = $this->getReleaseVersionController()
+        ->determineNextVersion();
     }
+    $this->validateVersion($version);
+    return $version;
+  }
+
+  public function jiraReleaseVersion($version = null, $opts = []) {
+    $version = $this->nextReleaseVersion($version);
     try {
       list($version, $output) = $this->getReleaseVersionController()
         ->assignVersionToCompletedTickets($version);
@@ -656,7 +663,7 @@ class RoboFile extends \Robo\Tasks {
   }
 
   protected function validateVersion($version) {
-    if (!preg_match('/\d+\.\d+\.\d+/', $version)) {
+    if (!\MailPoetTasks\Release\VersionHelper::validateVersion($version)) {
       $this->yell('Incorrect version format', 40, 'red');
       exit(1);
     }
--- a/tasks/release/ReleaseVersionController.php
+++ b/tasks/release/ReleaseVersionController.php
@@ -27,7 +27,7 @@ class ReleaseVersionController {
 
     $output = [];
     $output[] = "Setting version $version to completed tickets in $this->project...";
-    $issues = $this->getDoneIssuesWithoutVersion();
+    $issues = $this->getUnreleasedDoneIssues();
     $result = array_map(function ($issue) use ($version) {
       return $this->setIssueFixVersion($issue['key'], $version);
     }, $issues);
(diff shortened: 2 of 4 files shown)


RoboFile.php
```diff
@@ -594,10 +594,17 @@ public function writeReleaseVersion($version) {
...
-  public function jiraReleaseVersion($version = null, $opts = []) {
-    if ($version) {
-      $this->validateVersion($version);
+  public function nextReleaseVersion($version = null) {
+    if (!$version) {
+      $version = $this->getReleaseVersionController()
+        ->determineNextVersion();
     }
+    $this->validateVersion($version);
+    return $version;
```
JanJakes: This is also a command, right? Wouldn't returning a version cause it be "failed" (return code != 0)?

tasks/release/ReleaseVersionController.php
```diff
@@ -36,15 +36,50 @@ function assignVersionToCompletedTickets($version = null) {
...
+
+    $parsed_last_version[$version_to_increment]++;
+    $next_version = VersionHelper::buildVersion($parsed_last_version);
+    return $next_version;
+  }
+
+  private function checkProjectVersionIncrement($project) {
+    $this->jira->setProject($project);
+    $issues = $this->getUnreleasedDoneIssues();
+    $this->jira->setProject(JiraController::PROJECT_MAILPOET); // restore project
```
JanJakes: This doesn't restore the original value but always sets project "MAILPOET" instead I think.

What I'd suggest here is to get rid of the `setProject` setter on `JiraController` and pass it as a parameter to the exact method that needs it. It seems to be `getUnreleasedDoneIssues` below but actually that one seems to have project already in the JQL query 🤔

tasks/release/ReleaseVersionController.php
```diff
@@ -36,15 +36,50 @@ function assignVersionToCompletedTickets($version = null) {
...
+    if ($this->project === JiraController::PROJECT_MAILPOET) {
+      $free_increment = $this->checkProjectVersionIncrement(JiraController::PROJECT_MAILPOET);
+      $premium_increment = $this->checkProjectVersionIncrement(JiraController::PROJECT_PREMIUM);
+
+      if (in_array(VersionHelper::MINOR, [$free_increment, $premium_increment])) {
+        $version_to_increment = VersionHelper::MINOR;
+      }
+    }
+
+    $parsed_last_version[$version_to_increment]++;
```
JanJakes: If we increase minor version, then patch should be reset to `0`. I think that's missing here. Maybe a method in the `VersionHelper` could do it?

tasks/release/ReleaseVersionController.php
```diff
@@ -36,15 +36,50 @@ function assignVersionToCompletedTickets($version = null) {
...
+    $this->jira->setProject($project);
+    $issues = $this->getUnreleasedDoneIssues();
+    $this->jira->setProject(JiraController::PROJECT_MAILPOET); // restore project
+
+    $version_to_increment = VersionHelper::PATCH;
+    $field_id = JiraController::VERSION_INCREMENT_FIELD_ID;
+
+    foreach ($issues as $issue) {
+      if (!empty($issue['fields'][$field_id]['value'])
+        && $issue['fields'][$field_id]['value'] === VersionHelper::MINOR
```
JanJakes: Maybe a nicer coding style would be:
```
if (
  !empty(...)
  && $issue[...]
)
```
Not important though.

tasks/release/ReleaseVersionController.php
```diff
@@ -36,15 +36,50 @@ function assignVersionToCompletedTickets($version = null) {
...
+        $version_to_increment = VersionHelper::MINOR;
+        break;
+      }
+    }
+
+    return $version_to_increment;
+  }
+
+  private function getUnreleasedDoneIssues() {
+    $jql = "project = $this->project AND status = Done AND (fixVersion = EMPTY OR fixVersion IN unreleasedVersions()) AND updated >= -52w";
```
JanJakes: Two questions here:
1. Why do we need the `updated >= -52w`? Some perf optimization?
2. Do we want to pick all unreleased tickets or do we actually first want to assign the new release version (another step in our release checklist) and then here check only against that version? It's just a question, I don't know which approach is better.

tasks/release/ReleaseVersionController.php
```diff
@@ -36,15 +36,50 @@ function assignVersionToCompletedTickets($version = null) {
...
+      ) {
+        $version_to_increment = VersionHelper::MINOR;
+        break;
+      }
+    }
+
+    return $version_to_increment;
+  }
+
+  private function getUnreleasedDoneIssues() {
```
JanJakes: So yeah, this method could take `$project` parameter and then there would be no need for the project setter I guess.

tasks/release/ReleaseVersionController.php
wxa: I think the current one better adheres to PSR-2: 
> There MUST NOT be a space after the opening parenthesis

https://www.php-fig.org/psr/psr-2/#5-control-structures

tasks/release/ReleaseVersionController.php
wxa: 1. I took this from the Jira interface's queries, and it seems to be necessary, otherwise it picks up some odd ancient tickets.
2. See item no. 3 from specs, we want to pick all unreleased tickets.

RoboFile.php
wxa: This command is not intended for CI use, more for local debugging. Anyway, its returned value is converted to an exit code of 0. Must be exactly an integer between 0-255 to serve as an error code.

tasks/release/ReleaseVersionController.php
wxa: Good catch, I've added a helper method.

tasks/release/ReleaseVersionController.php
wxa: Nice find. Fixed as you suggested.

tasks/release/ReleaseVersionController.php
wxa: Fixed.

tasks/release/ReleaseVersionController.php
JanJakes: I think that's just for single-line conditions. They don't really specify multiline, however all other structures (function calls, etc.) specify using the style I suggested. Not important so let's leave it as is.

PR #1991, opened 2019-04-23
Update vulnerable JS libraries [MAILPOET-1994]

NeosinneR:


NeosinneR:
@amine-mp you're right, totally forgot about that. I'm on it.

PR #1994, opened 2019-04-23
Exclude non-public posts hidden by 3rd-party plugins from ALC [MAILPOET-1964]

NeosinneR:


NeosinneR:
@michelleshull I can't replicate this. It correctly shows number of posts (see screenshots below)
uam-private-admin
uam-private-email
uam-public-admin
uam-public-email

One thing that comes to mind, in UAM settings, did you set Posts to be completely hidden? Otherwise, the title is visible and therefore shown and counted in email.
uam-settings

PR #1995, opened 2019-04-23
Display subscribers with correct status [MAILPOET-1944]

pavel-mailpoet:


lib/API/JSON/v1/Subscribers.php
```diff
@@ -95,24 +95,42 @@ function listing($data = array()) {
...
+
+    return $this->successResponse($result, array(
       'count' => $listing_data['count'],
       'filters' => $listing_data['filters'],
       'groups' => $listing_data['groups']
     ));
   }
 
+  private function fixSubscriberStatus(array $subscriber, $segment_id) {
+    $segment_status = Subscriber::STATUS_SUBSCRIBED;
```
NeosinneR: Why is this set to `subscribed`? It seems unnecessary.

pavel-mailpoet: The line is not necessary, but it is here to make sure the variable is always defined and always the same type. But maybe it is confusing and I'll refactor the function a bit.

pavel-mailpoet:
> In the spec, there is mentioned acceptance test, but there is integration. Is that something you agreed on?

Yes, we spoke about it with Michelle and we agreed this is an edge case that can live with an integration test.

pavel-mailpoet:
@michelleshull I am trying to reproduce the problem with WordPress user you described. But I cannot reproduce it. Deleting WordPress users is in the code area that my branch didn't even touch. So it seems unrelated. I rebased master so if there was something in master this could fix it.

michelleshull:
Send it back to me, I'll test again on some different sites.

michelleshull:
I can't reproduce now, either. I think it was related to that user being marked as spam by Akismet, but it's not related to this branch.

PR #1996, opened 2019-04-23
Release commands wrapup [MAILPOET-1977]

JanJakes:
DO NOT TEST THE COMMANDS!

They would trigger full release and testing it is hard. It was tested by author and we’ll handle any potential problems during the upcoming releases.

To see a draft of the release checklist after this is merged have a look at https://mailpoet.atlassian.net/wiki/spaces/MAILPOET/pages/685801481/DRAFT+Releasing+Free+plugin.


.env.sample
```diff
@@ -25,5 +25,10 @@ WP_TEST_MAILER_SMTP_PASSWORD=""
 WP_SVN_USERNAME=""
 WP_SVN_PASSWORD=""
 WP_TRANSIFEX_API_TOKEN=""
-WP_JIRA_USER="" // Jira username/email
-WP_JIRA_TOKEN="" // Jira token (https://confluence.atlassian.com/cloud/api-tokens-938839638.html) or password
+WP_JIRA_USER="" // JIRA username/email
```
NeosinneR: I would add a comment as in the premium plugin
```# Optional: for plugin deployment```

File: .env.sample [hunk 2]
```diff
@@ -25,5 +25,10 @@ WP_TEST_MAILER_SMTP_PASSWORD=""
...
+WP_JIRA_TOKEN="" // JIRA token from https://id.atlassian.com/manage/api-tokens
+WP_CIRCLECI_USERNAME="" // CircleCI organization or user
+WP_CIRCLECI_TOKEN="" // CircleCI token from https://circleci.com/gh/{user|org}/{project}/edit#api
+WP_GITHUB_USERNAME="" // GitHub username (not email)
+WP_GITHUB_TOKEN="" // GitHub token from https://github.com/settings/tokens
+WP_SLACK_WEBHOOK_URL="" // Webhook URL to post release info on Slack
```
NeosinneR: Do we have this URL documented somewhere or do devs have to go to Slack's settings and copy it from there? If it's documented, might be useful to point to it here

.env.sample
JanJakes: It is a few lines above, actually.

PR #1997, opened 2019-04-24
Random strings [MAILPOET-1902]

amine-mp:
[MAILPOET-1902]

[MAILPOET-1902]: https://mailpoet.atlassian.net/browse/MAILPOET-1902

### DIFF
--- a/lib/Subscribers/SubscriberPersonalDataEraser.php
+++ b/lib/Subscribers/SubscriberPersonalDataEraser.php
@@ -16,7 +16,6 @@ class SubscriberPersonalDataEraser {
         'done' => true,
       );
     }
-
     $subscriber = Subscriber::findOne(trim($email));
     $item_removed = false;
     $items_retained = true;
@@ -44,7 +43,7 @@ class SubscriberPersonalDataEraser {
   }
 
   private function anonymizeSubscriberData($subscriber) {
-    $subscriber->email = sprintf('deleted-%s@site.invalid', uniqid('', true));
+    $subscriber->email = sprintf('deleted-%s@site.invalid', bin2hex(random_bytes(12))); // phpcs:ignore
     $subscriber->first_name = 'Anonymous';
     $subscriber->last_name = 'Anonymous';
     $subscriber->status = Subscriber::STATUS_UNSUBSCRIBED;
--- a/lib/Util/Security.php
+++ b/lib/Util/Security.php
@@ -13,9 +13,8 @@ class Security {
   }
 
   static function generateRandomString($length = 5) {
-    // non-cryptographically strong random generator
     return substr(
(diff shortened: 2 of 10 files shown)


lib/Util/Security.php
```diff
@@ -13,9 +13,8 @@ static function generateToken($action = 'mailpoet_token') {
   }
 
   static function generateRandomString($length = 5) {
-    // non-cryptographically strong random generator
     return substr(
-      md5(uniqid((string)mt_rand(), true)),
+      bin2hex(random_bytes(16)), // phpcs:ignore
       0,
       min(max(5, (int)$length), 32)
```
pavel-mailpoet: I think this was here before because we used `md5` and we couldn't generate a string longer than 32. But now using `random_bytes` we can generate anything. So we could maybe fix this line as well. A couple of lines below we call this function with 64. We couldn't generate such string before but now we can.

lib/Subscribers/SubscriberPersonalDataEraser.php
```diff
@@ -44,7 +43,7 @@ private function eraseCustomFields($subscriber_id) {
   }
 
   private function anonymizeSubscriberData($subscriber) {
-    $subscriber->email = sprintf('deleted-%s@site.invalid', uniqid('', true));
+    $subscriber->email = sprintf('deleted-%s@site.invalid', bin2hex(random_bytes(12))); // phpcs:ignore
```
pavel-mailpoet: How does this work on PHP 5.6? I don't see any polyfill added or the function `random_bytes` being defined...

lib/Util/Security.php
```diff
@@ -13,9 +13,8 @@ static function generateToken($action = 'mailpoet_token') {
   }
 
   static function generateRandomString($length = 5) {
-    // non-cryptographically strong random generator
     return substr(
-      md5(uniqid((string)mt_rand(), true)),
+      bin2hex(random_bytes(16)), // phpcs:ignore
```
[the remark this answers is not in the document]
amine-mp: `bin2hex(random_bytes(16))` produces a 32 characters string. I can't pass `$length / 2` because `$length` is not always pair.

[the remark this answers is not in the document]
pavel-mailpoet: But couldn't you generate more than is needed passing $length and then cut the string using substr to the desired length?

lib/Util/Security.php
amine-mp: Right. I will fix that.

lib/Subscribers/SubscriberPersonalDataEraser.php
amine-mp: WP already includes a polyfill for `random_bytes` here https://github.com/WordPress/WordPress/tree/master/wp-includes/random_compat

PR #2003, opened 2019-04-25
Move tests with encapsulated WP functions from integration to unit [MAILPOET-2009]

wxa:


lib/Subscription/Pages.php
```diff
@@ -41,6 +44,7 @@ function __construct($action = false, $data = array(), $init_shortcodes = false,
       $this->new_subscriber_notification_sender = new NewSubscriberNotificationMailer();
     }
     $this->settings = new SettingsController();
+    $this->url_helper = new UrlHelper(new WPFunctions());
```
amine-mp: Ideally we would use DI to inject all dependencies, is there a reason why you didn't use it here?

wxa: Refactoring this class was more work than I planned, it also drags along refactoring of `Config\Shortcodes`, `Router\EndpointsSubscription` etc. I had to stop somewhere so there is something to do on the next hacking day 😄

amine-mp: Makes sense. Thanks for the clarification.

PR #2004, opened 2019-04-25
Fix flaky acceptance test [MAILPOET-2018]

costasovo:


tests/acceptance/WooCommerceListImportPageCest.php
```diff
@@ -6,11 +6,12 @@
 require_once __DIR__ . '/../DataFactories/WooCommerceProduct.php';
```
NeosinneR: This is no longer needed

PR #2011, opened 2019-04-30
Use site-specific form field obfuscation hash [MAILPOET-1999]

JanJakes:


tests/unit/Form/Util/FieldNameObfuscatorTest.php
```diff
@@ -2,7 +2,18 @@
...
 
+use Codeception\Stub;
+use MailPoet\WP\Functions as WPFunctions;
+
 class FieldNameObfuscatorTest extends \MailPoetUnitTest {
+  function _before() {
+    parent::_before();
+    WPFunctions::set(
+      Stub::make(WPFunctions::class, [
+        'homeUrl' => 'http://example.com',
```
wxa: Please reset the stubbed function after tests.

JanJakes: 🤔 We don't do that in any other unit test for WPFunctions. The reason is that each test creates its own mock and never reuses the global instance. So there is probably no harm but you're right it is more correct to reset it, I can add it.

wxa: In theory you're right until someone forgets to stub out their WP functions, and you may get a flaky test that depends on some other test. The person who has to dig into this won't be happy.

PR #2014, opened 2019-04-30
Add backward compatibility for wpdb parse_db_host [MAILPOET-2023]

costasovo:


lib/WP/Functions.php
```diff
@@ -534,7 +534,23 @@ function wpautop($pee, $br = true) {
    */
   function parseDbHost($host) {
     global $wpdb;
-    return $wpdb->parse_db_host($host);
+    if (method_exists($wpdb, 'parse_db_host')) {
+      return $wpdb->parse_db_host($host);
+    } else {
+      // Backward compatibility for WP 4.7 and 4.8
+      $db_host = DB_HOST;
```
wxa: Why do you use the `DB_HOST` constant and not the `$host` function argument?

costasovo: Ohh. Good catch! It was a copy paste from old version. I did not realized that. Will fix it.

PR #2016, opened 2019-05-01
Improve batch custom field update query [MAILPOET-2028]

costasovo:


lib/Models/SubscriberCustomField.php
```diff
@@ -63,16 +63,17 @@ static function createMultiple($values) {
...
 
   static function updateMultiple($values) {
+    $subscriber_ids = array_map(function($value) {
+      return $value[1];
+    }, $values);
+
+    $params = Helpers::flattenArray($values);
+    $query = sprintf("UPDATE `" . self::$_table . "` SET value = (CASE " .
+      str_repeat('WHEN custom_field_id = ? AND subscriber_id = ? THEN ? ', count($values)) .
+      "ELSE value END) WHERE subscriber_id IN (%s)", implode(',', array_unique($subscriber_ids)));
```
JanJakes: I know it kinda follows the previous formatting but it's wrong - multiline function call needs newlines before/after `(`/`)`, concatenation should probably start on next line, not on the previous one. It's definitely more readable like this:
```
$query = sprintf(
  "UPDATE `" . self::$_table . "` SET value = (CASE "
    . str_repeat('WHEN custom_field_id = ? AND subscriber_id = ? THEN ? ', count($values))
    . "ELSE value END) WHERE subscriber_id IN (%s)",
  implode(',', array_unique($subscriber_ids))
);
```

But I'd also avoid mixing concatenation of variables with `sprintf`'s `%s` and use `'` instead of `"` when possible:
```
$query = sprintf(
  'UPDATE %s SET value = (CASE %s ELSE value END) WHERE subscriber_id IN (%s)',
  self::$_table,
  str_repeat('WHEN custom_field_id = ? AND subscriber_id = ? THEN ? ', count($values))
  implode(',', array_unique($subscriber_ids))
);
```

Now when it's readable I can decode what was changed. Looks definitely like a good thing to add the `WHERE` part but in general I don't think it is good to do a multi-update with CASE... WHEN. Not only it's very difficult to read I'd also be doubtful about performance. Here it is probably not so important (I'm guessing, I have no context) so let's keep it this way if it solves the problem we're solving but otherwise I think multiple updates in a transaction or a [single update with inner join](https://stackoverflow.com/questions/5008147/update-multiple-rows-using-one-query) seem much better to me.

lib/Models/SubscriberCustomField.php
```diff
@@ -63,16 +63,17 @@ static function createMultiple($values) {
   }
 
   static function updateMultiple($values) {
+    $subscriber_ids = array_map(function($value) {
+      return $value[1];
+    }, $values);
+
+    $params = Helpers::flattenArray($values);
```
JanJakes: Why was this extracted to a variable and moved here when it is not used anywhere except its original place?

lib/Models/SubscriberCustomField.php
```diff
@@ -63,16 +63,17 @@ static function createMultiple($values) {
   }
 
   static function updateMultiple($values) {
+    $subscriber_ids = array_map(function($value) {
```
JanJakes: I'd add `array` typehint but actually... you should use `array_column`. I'd also add the `array_unique` from below right here.

PR #2017, opened 2019-05-01
Send confirmation email also within subscribeToLists [MAILPOET-1992]

JanJakes:


lib/API/MP/v1/API.php
```diff
@@ -217,17 +232,9 @@ function addSubscriber(array $subscriber, $segments = array(), $options = array(
...
-      // send confirmation email
-      if ($send_confirmation_email && $new_subscriber->status === Subscriber::STATUS_UNCONFIRMED) {
-        $result = $this->_sendConfirmationEmail($new_subscriber);
-        if (!$result && $new_subscriber->getErrors()) {
-          throw new \Exception(
-            WPFunctions::get()->__(sprintf('Subscriber added, but confirmation email failed to send: %s', strtolower(implode(', ', $new_subscriber->getErrors()))), 'mailpoet')
-          );
-        }
-      }
+      $this->subscribeToLists($new_subscriber->id, $segments, [
```
amine-mp: Here we fetch the subscriber from the DB then we pass his id to `subscribeToLists` which fetches him again from DB. Is there a way to not do the same DB query twice?

JanJakes: Good point. I'm afraid there is no easy way to fix that now. Considering:
1. It was like this already before, I didn't change that behavior (I didn't add `subscribeToLists`, only added params) .
2. Some ORM's would cache the subscriber per request (Doctrine).

Do you think it needs a fix? I cannot think of a one.

amine-mp: A fix is possible but it requires changing the API interface, and since we will switch to a better ORM with caching feature in the future I think this is not a big deal.

PR #2034, opened 2019-05-09
Skip deleted segments in SubscribersFinder [MAILPOET-2026]

wxa:
A newsletter can have deleted segments, i.e. orphaned relations to segments (see here: https://github.com/mailpoet/mailpoet/blob/master/lib/Models/Newsletter.php#L488). Thus we should not clean up these relations but skip them during processing.


wxa:
@costasovo A deleted segment existence is handled in code by displaying 'Deleted list' in the newsletter segments column and is verified with tests. I think that's handy for sent emails: you'll see that they were sent to a deleted list, not in a void.

costasovo:
@wxa I see. Makes sense. Thanks for explanation.

PR #2035, opened 2019-05-09
Add newsletter sending status page [MAILPOET-1155]

amine-mp:
[MAILPOET-1155]

## Note about tests

Instead of writing acceptance test for every type of newsletters. I have written generic integration tests and then tested only one newsletter type on the acceptance test. The sending status page should work the same for other types. 

[MAILPOET-1155]: https://mailpoet.atlassian.net/browse/MAILPOET-1155

### DIFF
--- a/assets/js/src/newsletters/sending_status.jsx
+++ b/assets/js/src/newsletters/sending_status.jsx
@@ -145,7 +145,7 @@ const ListingItem = ({
   }
   return (
     <>
-      <td className={rowClasses}>
+      <td data-automation-id={`name_${taskId}_${subscriberId}`} className={rowClasses}>
         <strong>
           <a
             className="row-title"
@@ -158,10 +158,10 @@ const ListingItem = ({
           { `${firstName} ${lastName}` }
         </p>
       </td>
-      <td className="column" data-colname={MailPoet.I18n.t('sendingStatus')}>
+      <td className="column" data-automation-id={`status_${taskId}_${subscriberId}`} data-colname={MailPoet.I18n.t('sendingStatus')}>
         { status }
       </td>
-      <td className="column" data-colname={MailPoet.I18n.t('failureReason')}>
+      <td className="column" data-automation-id={`error_${taskId}_${subscriberId}`} data-colname={MailPoet.I18n.t('failureReason')}>
         { error }
       </td>
     </>
--- a/assets/js/src/newsletters/sending_status.jsx
+++ b/assets/js/src/newsletters/sending_status.jsx
@@ -1,7 +1,7 @@
+import React from 'react';
 import MailPoet from 'mailpoet';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import React, { Fragment } from 'react';
 import { Link } from 'react-router-dom';
 import Listing from 'listing/listing.jsx';
 import { CronMixin, MailerMixin } from 'newsletters/listings/mixins.jsx';
@@ -50,7 +50,7 @@ const SendingStatus = (props) => {
   }, [newsletterId]);
 
   return (
-    <Fragment>
+    <>
       <h1>{MailPoet.I18n.t('sendingStatusTitle')}</h1>
       <StatsLink
         newsletterId={newsletterId}
@@ -73,7 +73,7 @@ const SendingStatus = (props) => {
           CronMixin.checkCronStatus(state);
         }}
       />
-    </Fragment>
+    </>
   );
 };
 SendingStatus.propTypes = {
@@ -144,7 +144,7 @@ const ListingItem = ({
     }
   }
   return (
-    <Fragment>
+    <>
       <td className={rowClasses}>
         <strong>
           <a
@@ -164,7 +164,7 @@ const ListingItem = ({
       <td className="column" data-colname={MailPoet.I18n.t('failureReason')}>
         { error }
       </td>
-    </Fragment>
+    </>
   );
 };
 ListingItem.propTypes = {
--- a/assets/js/src/newsletters/sending_status.jsx
+++ b/assets/js/src/newsletters/sending_status.jsx
@@ -1,13 +1,19 @@
 import Hooks from 'wp-js-hooks';
 import MailPoet from 'mailpoet';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 import React, {Fragment} from 'react';
 import { Link } from 'react-router-dom';
+import Listing from 'listing/listing.jsx';
+import { CronMixin, MailerMixin } from 'newsletters/listings/mixins.jsx';
 
 const SendingStatus = props => {
   const newsletterId = props.match.params.id;
-  const [error, setError] = React.useState(null);
-  const [newsletterSubject, setNewsletterSubject] = React.useState(null);
+  const [state, setState] = React.useState({
+    error: null,
+    newsletterSubject: null,
+  });
+
   React.useEffect(() => {
     MailPoet.Ajax.post({
       api_version: window.mailpoet_api_version,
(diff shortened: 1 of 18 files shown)


pavel-mailpoet:
@amine-mp I started testing your branch and I noticed straight away that when I open the new `sending-status` page it makes 2 ajax requests to the same endpoint `sending_task_subscribers` listings. I think that should be fixed. We should make only 1 request.

I am not sure if that is not my computer but those requests are quite slow. I have 12 subscribers and it takes more than 20 seconds to load the listings.

assets/js/src/newsletters/sending_status.jsx
```diff
@@ -0,0 +1,184 @@
...
+      endpoint: 'newsletters',
+      action: 'get',
+      data: {
+        id: newsletterId,
+      },
+    })
+      .done(res => setNewsletterSubject(res.data.subject))
+      .fail((res) => {
+        setNewsletterSubject('');
+        MailPoet.Notice.error(
```
pavel-mailpoet: You should use `showApiErrorNotice` instead of `error` it has some checks if everything is working.

assets/js/src/newsletters/sending_status.jsx
```diff
@@ -0,0 +1,184 @@
...
+        afterGetItems={(state) => {
+          MailerMixin.checkMailerStatus(state);
+          CronMixin.checkCronStatus(state);
+        }}
+      />
+    </>
+  );
+};
+SendingStatus.propTypes = {
+  location: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
```
pavel-mailpoet: Do we always have to disable eslint on this line? you know what you need, don't you? Can't you use `shape` instead of `object` here?

assets/js/src/newsletters/sending_status.jsx
```diff
@@ -0,0 +1,184 @@
...
+}) => {
+  const resend = () => {
+    MailPoet.Ajax.post({
+      api_version: window.mailpoet_api_version,
+      endpoint: 'sending_task_subscribers',
+      action: 'resend',
+      data: { taskId, subscriberId },
+    })
+      .done(() => window.mailpoet_listing.forceUpdate())
+      .fail(res => MailPoet.Notice.error(
```
pavel-mailpoet: Same as above, we should use `showApiErrorNotice`

lib/API/JSON/v1/SendingTaskSubscribers.php
```diff
@@ -0,0 +1,102 @@
...
+    $newsletter_id = !empty($data['params']['id']) ? (int)$data['params']['id'] : false;
+    $tasks_ids = SendingQueueModel::select('task_id')
+      ->where('newsletter_id', $newsletter_id)
+      ->findArray();
+    if (empty($tasks_ids)) {
+      return $this->errorResponse([
+        APIError::NOT_FOUND => __('This newsletter is not being sent to any subcriber yet.', 'mailpoet'),
+      ]);
+    }
+    $data['params']['task_ids'] = array_map(function($item) {
```
pavel-mailpoet: We could use `array_column` here, but since it is such a broken function, I don't know if `array_map` isn't better here...

tests/DataFactories/Newsletter.php
```diff
@@ -213,6 +219,16 @@ public function withSendingQueue(array $options = []) {
     return $this;
   }
 
+  public function beingSentToSubscriber($subscriber, array $data = []) {
```
pavel-mailpoet: I am not sure about the name. All other methods in the data factories start with `with`. I think this shouldn't be an exception...

tests/acceptance/SendingStatusCest.php
```diff
@@ -0,0 +1,66 @@
...
+      ])
+      ->create();
+    // When I visit the newsletters page
+    $I->login();
+    $I->amOnMailPoetPage('Emails');
+    $I->waitForText($newsletter->subject);
+    // I click on the "Sent to 3 of 3" link
+    $I->click('[data-automation-id="sending_status_' . $newsletter->id . '"]');
+    // Check I am on the sending status page
+    $I->seeInCurrentUrl('?page=mailpoet-newsletters#/sending-status/' . $newsletter->id);
```
pavel-mailpoet: Why do we need this check?

pavel-mailpoet:
And another thing that needs to be updated is the text of the notices, see Jira comments

tests/acceptance/SendingStatusCest.php
amine-mp: Just to make sure the link is working.

tests/DataFactories/Newsletter.php
amine-mp: I will rename it to `withSubscriber`.

tests/acceptance/SendingStatusCest.php
pavel-mailpoet: Wouldn't that be obvious from the next check(s)?

lib/API/JSON/v1/SendingTaskSubscribers.php
amine-mp: I think `array_column` makes more sense in that case. At least it's less verbose.

tests/acceptance/SendingStatusCest.php
amine-mp: Yeah, you're right. I will remove that line.

lib/API/JSON/v1/SendingTaskSubscribers.php
```diff
@@ -47,7 +47,7 @@ function listing($data = []) {
       ->findArray();
     if (empty($tasks_ids)) {
       return $this->errorResponse([
-        APIError::NOT_FOUND => __('This newsletter is not being sent to any subcriber yet.', 'mailpoet'),
+        APIError::NOT_FOUND => __( 'This email has not been sent yet.', 'mailpoet'),
```
pavel-mailpoet: Why did you add a space here? `__(  '`

assets/js/src/newsletters/sending_status.jsx
```diff
@@ -0,0 +1,185 @@
...
+  }, [newsletterId]);
+
+  return (
+    <>
+      <h1>{MailPoet.I18n.t('sendingStatusTitle')}</h1>
+      <StatsLink
+        newsletterId={newsletterId}
+        newsletterSubject={newsletterSubject}
+      />
+      {!isLoading && (
```
pavel-mailpoet: I understand what you did here, but the User Experience is quite bad now. User will come to a blank page, sees nothing and after a while, everything just jumps to the page. Before this change, there was a table showing them we're loading the data.

lib/API/JSON/v1/SendingTaskSubscribers.php
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  function listing($data = []) {
+    $newsletter_id = !empty($data['params']['id']) ? (int)$data['params']['id'] : false;
+    $tasks_ids = SendingQueueModel::select('task_id')
+      ->where('newsletter_id', $newsletter_id)
+      ->findArray();
+    if (empty($tasks_ids)) {
+      return $this->errorResponse([
+        APIError::NOT_FOUND => __( 'This email has not been sent yet.', 'mailpoet'),
```
pavel-mailpoet: I think space shouldn't be here. `__( 'This`

lib/API/JSON/v1/SendingTaskSubscribers.php
```diff
@@ -0,0 +1,100 @@
...
+      ]);
+    }
+    $data['params']['task_ids'] = array_column($tasks_ids, 'task_id');
+    $listing_data = $this->listing_handler->get('\MailPoet\Models\ScheduledTaskSubscriber', $data);
```
pavel-mailpoet: Shall we use some default order? I would think that failed subscribers should be in the beginning so that it is obvious what failed.

lib/API/JSON/v1/SendingTaskSubscribers.php
amine-mp: This was a typo.

lib/API/JSON/v1/SendingTaskSubscribers.php
amine-mp: This may be a good idea but it's not part of the specs.

pavel-mailpoet: That is true, it is not part of the specs. But shouldn't we try to make the plugin the best for our users? Shouldn't we develop our features so that users would like to use them? We should go above and beyond to give our users a plugin they will love. Not just what is part of the specs.

assets/js/src/newsletters/sending_status.jsx
amine-mp: I fixed that using `React.memo`

PR #2039, opened 2019-05-13
Authorized emails validation and error notice [MAILPOET-1987]

costasovo:


lib/Util/Notices/UnauthorizedEmailNotice.php
```diff
@@ -0,0 +1,112 @@
...
+        endpoint: 'mailer', action: 'resumeSending'
+        }).done(function() {
+          jQuery('.mailpoet-js-error-unauthorized-emails-notice').slideUp();
+          MailPoet.Notice.success('" . WPFunctions::get()->__('Sending has been resumed.') . "');}
+        ).fail(function(response) {
+          if (response.errors.length > 0) {
+            MailPoet.Notice.error(response.errors.map(function(error) { return error.message }), { scroll: true });
+          }
+        });
+      });
```
JanJakes: Uff, I'm not familiar with reasons for this and how other messages are rendered but does this really have to be in PHP? Is there anything preventing us from doing this the "standard" way?

tests/integration/Cron/Workers/AuthorizedSendingEmailsCheckTest.php
```diff
@@ -23,4 +23,17 @@ function testItRunsCheckOnBridge() {
     $worker = new AuthorizedSendingEmailsCheck($bridge_mock);
     $worker->processTaskStrategy(ScheduledTask::createOrUpdate([]));
   }
+
+  function testItDoesScheduleAutomatically() {
```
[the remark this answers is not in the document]
costasovo: Oh. It should be does not. Good catch!

lib/Util/Notices/UnauthorizedEmailNotice.php
costasovo: I don't like it either :(  I used a system for displaying permanent notices but it can only render HTML. I checked other places and found that we do similar thing here: 
https://github.com/mailpoet/mailpoet/blob/296ad974d8e7433b345df2766309609ab7239d5d/lib/Mailer/Methods/ErrorMappers/MailPoetMapper.php#L92 so I decided to do it same way.

I was thinking about 2 alternative solutions:
- rendering only HTML and adding a js code for resume sending [here](https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/dismissible-notice.jsx) or to some new js file. (preferred current solution since code is together)
- creating an extra react component only for this notice (felt like non system solution and would have to take care e.g. that it disappears when settings are saved etc.)

Do you have some suggestion?

JanJakes: I see. It's not in the scope of this feature to refactor the notice system to something nicer but inlining JS like this is really bad. It not only avoids linters but it can also be forgotten when someone will refactor `MailPoet.anything...` API in JS. Don't you think it would be better to extract JS at least? I think it's better to have in two places than to call our JS classes like `MailPoet` from PHP.

costasovo: Good point. I'll extract it.

lib/Cron/Workers/SimpleWorker.php
```diff
@@ -42,7 +43,7 @@ function process() {
     $scheduled_tasks = self::getScheduledTasks();
     $running_tasks = self::getRunningTasks();
 
-    if (!$scheduled_tasks && !$running_tasks) {
+    if (!$scheduled_tasks && !$running_tasks && static::AUTOMATIC_SCHEDULING) {
```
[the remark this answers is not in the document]
costasovo: Actually we need to know that there are no running or scheduled tasks so we need to fetch them. But there is another issue. The check has to be nested inside so that we end process same way even if automatic scheduling is disabled.

costasovo:
I fixed coding issues and added 2 new commits: refactor JS code out of php + removed old similar message from admin pages.

michelleshull:
This looks great, but it needs a rebase before I can merge, @costasovo

costasovo:
@michelleshull rebased

costasovo:
I've added a two commits:
- fixes for coding style errors found by new coding style check
- show messages with one email address in case both default sender and sing-confirmation are same

lib/Util/Notices/UnauthorizedEmailNotice.php
```diff
@@ -73,9 +77,13 @@ private function getSettingsButtons($validation_error) {
...
-    if (!empty($validation_error['invalid_sender_address']) && !empty($validation_error['invalid_confirmation_address'])) {
-      $authorize_link = $this->wp->x('Authorize %email and %email2', 'Link for user to authorize their email address');
-      $authorize_link = str_replace('%email', EscapeHelper::escapeHtmlText($validation_error['invalid_sender_address']), $authorize_link);
+    if (
+      !empty($validation_error['invalid_sender_address'])
+      && !empty($validation_error['invalid_confirmation_address'])
+      && $validation_error['invalid_sender_address'] !== $validation_error['invalid_confirmation_address']
+    ) {
+      $authorize_link = $this->wp->x('Authorize %email1 and %email2', 'Link for user to authorize their email address');
+      $authorize_link = str_replace('%email1', EscapeHelper::escapeHtmlText($validation_error['invalid_sender_address']), $authorize_link);
```
JanJakes: Looks good. However (and I notice that just now), considering `UnauthorizedEmailNotice.php` is a completely new class I don't think we should use "hacky" `str_replace('%email1', ...)`. PHP has `sprintf` exactly for this. Would you mind adding a commit to fix that?

costasovo: I used str_replace to keep nicer placeholders for translators.

costasovo:
@JanJakes I added one commit with translations function calls fix

PR #2041, opened 2019-05-13
Update Beamer announcement date [MAILPOET-2071]

amine-mp:


michelleshull:
We don't need this ticket at this time, closing and marked as "won't do" in Jira.

PR #2045, opened 2019-05-14
Fix inactive subscribers issue with MP2 migration [MAILPOET-2040]

costasovo:


tests/integration/Subscribers/InactiveSubscribersControllerTest.php
```diff
@@ -188,6 +208,24 @@ function testItDoesNotActivateOldSubscribersWithUnopenedEmail() {
     expect($subscriber->status)->equals(Subscriber::STATUS_INACTIVE);
   }
 
+  function testItActivatesSubscribersWhenMP2MigrationHappenedWithinInterval() {
```
pavel-mailpoet: This test is very hard to understand. We should probably stop using numbers and use constants or variables instead.

PR #2047, opened 2019-05-14
Track WC revenues with cookies: Ask permission [MAILPOET-1856]

pavel-mailpoet:


views/settings/woocommerce.html
```diff
@@ -80,4 +80,32 @@
...
+      <td>
+        <p>
+          <input
+            type="checkbox"
+            value="1"
+            id="settings[mailpoet_accept_cookie_revenue_tracking]"
+            name="woocommerce[accept_cookie_revenue_tracking][enabled]"
+            <% if(settings.woocommerce.accept_cookie_revenue_tracking) %>checked="checked"<% endif %>
+          >
+          <input
```
costasovo: This hidden field is here just so that saving of setting doesn't reset information that screen was displayed. Am I correct? Just double checking if I understand correctly.

tests/acceptance/NewsletterStatisticsCest.php
```diff
@@ -20,10 +20,18 @@
...
+    $this->settings = $settings;
+  }
+
   function _before(\AcceptanceTester $I) {
     $I->activateWooCommerce();
-    (new Settings())->withWooCommerceListImportPageDisplayed(true);
-    (new Settings())->withDisplayRevenuesEnabled();
+    $this->settings->withWooCommerceListImportPageDisplayed(true);
+    $this->settings->withDisplayRevenuesEnabled();
+    $this->settings->withCookieRevenueTrackingDisabled();
```
costasovo: I think that we will need to add this setting to more tests (probably all which work with WooCommerce), but it could be probably done once we remove the feature switch or turn the feature on by default.

assets/css/src/components/_welcomeWizard.scss
```diff
@@ -32,6 +32,12 @@
   }
 }
 
+.mailpoet_welcome_wizard_step_revenue_tracking {
+  input[type="submit"] {
```
costasovo: Style lint is already in master and I think this rule should use single quotes https://stylelint.io/user-guide/rules/string-quotes/ so please rebase on master.

assets/js/src/wizard/wizard.jsx
```diff
@@ -17,10 +17,10 @@ if (container) {
   ReactDOM.render((
     <HashRouter>
       <div>
-        <Route exact path="/" render={() => <Redirect to={basePath} />} />
         <Route path="/steps/:step" component={WelcomeWizardStepsController} />
         <Route path="/import" component={WooCommerceImportController} />
         <Route path="/revenue-tracking-permission" component={RevenueTrackingPermissionController} />
+        <Route render={() => <Redirect to={basePath} />} />
```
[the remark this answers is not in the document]
pavel-mailpoet: I didn't just move it. I also removed the matching part. It didn't work previously when redirecting from newsletters and `#standard` stayed in the URL and none of the routes worked for me.

views/settings/woocommerce.html
pavel-mailpoet: Correct.

assets/js/src/wizard/revenue_tracking_permission.jsx
```diff
@@ -1,12 +1,46 @@
-import React from 'react';
+import React, { useState } from 'react';
+import MailPoet from 'mailpoet';
+import RevenueTrackingPermissionStep from './steps/revenue_tracking_permission_step.jsx';
 
 function RevenueTrackingPermission() {
+  const [loading, setLoading] = useState(false);
+
+  const handleApiError = (response) => {
```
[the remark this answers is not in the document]
pavel-mailpoet: I am not sure how to do that. First of all, I need to call `setLoading(false)` so I still need to have some code that is called. But then can I assume I can show the notice everywhere on every ajax call?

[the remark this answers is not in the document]
costasovo: I was thinking about something like that:

```js
const handleApiError = (response) => {
    setLoading(false);
    MailPoet.Notice.showApiErrorNotice(response);
}
```

michelleshull:
@pavel-mailpoet This needs a rebase, I think to pick up the feature flag PR changes I just merged.

PR #2050, opened 2019-05-15
Use our own wp-js-hooks fork [MAILPOET-2082]

pavel-mailpoet:


pavel-mailpoet:
@michelleshull those hooks are used for free-premium integration. So when testing focus on both free an premium plugin if they work ok together. Hopefully, there is no change in the code.

PR #2052, opened 2019-05-16
Fix form messages translation

amine-mp:


lib/Models/Form.php
```diff
@@ -151,3 +151,6 @@ static function updateSuccessMessages() {
   }
 
 }
+
+Form::$message_when_confirmation_enabled = __('Check your inbox or spam folder to confirm your subscription.', 'mailpoet');
+Form::$message_when_confirmation_disabled = __('You’ve been successfully subscribed to our newsletter!', 'mailpoet');
```
pavel-mailpoet: I don't think it is a good idea to leave this code here. It means the code will be run each time we load this file. And that may have some unwanted consequences. I don't think it is a big harm if we keep the messages duplicated in the code here.

PR #2055, opened 2019-05-16
Show notice in Beamer when MailPoet update is available [MAILPOET-2038]

NeosinneR:


assets/css/src/components/_featureAnnouncement.scss
```diff
@@ -23,3 +23,34 @@
   top: -4px;
   width: 10px;
 }
+
+.mailpoet_update_notice {
```
costasovo: Just a suggestion: What about name `.mailpoet_in_beamer_update_notice` I think that this name could help explain why such a big z-index is needed.

assets/js/src/announcements/feature_announcement.jsx
```diff
@@ -60,6 +62,16 @@ class FeatureAnnouncement extends React.Component {
       action: 'set',
       data,
     });
+    if (window.mailpoet_update_available && !document.getElementById('mailpoet_update_notice')) {
```
costasovo: This would be nicer and more readable as an extra function. 
```
window.Beamer.show();
this.showPluginUpdateNotice();
```

lib/Util/Helpers.php
```diff
@@ -95,4 +98,8 @@ static function recursiveTrim($value) {
     return $value;
   }
 
+  static function isMailPoetUpdateAvailable() {
```
costasovo: I think it is not good idea adding this here. It has a dependency on WordPress which we want to inject and adding it this way will make it hard to refactor.

I would suggest:
- moving `MAILPOET_PLUGIN_PATH ` into Env::plugin_path (note that the property already exists but is not filled in initialization or used anywhere else)
- performing the check in-place in Menu class which already has WPFunctions injected. `$data['is_mailpoet_update_available'] = array_key_exists(Env::$plugin_path,  $this->wp->getPluginUpdates())`

PR #2062, opened 2019-05-21
Clear schedule date when duplicating newsletter [MAILPOET-2080]

NeosinneR:


lib/Models/Newsletter.php
```diff
@@ -411,8 +411,16 @@ function duplicate($data = []) {
       $options = NewsletterOption::where('newsletter_id', $this->id)
         ->findMany();
 
+      $ignored_option_field_ids = [
+        1, // isScheduled
```
costasovo: Please use option names instead of ids. Ids are autogenerated values so names should be safer.

PR #2066, opened 2019-05-21
Run WooCommerce Customers list sync on settings change [MAILPOET-2079]

NeosinneR:


lib/API/JSON/v1/Settings.php
```diff
@@ -51,6 +57,16 @@ function set($settings = []) {
...
         $this->onInactiveSubscribersIntervalChange();
       }
+
+      // Sync WooCommerce Customers list on settings change
+      $new_subscribe_old_woocommerce_customers = 0;
+      if (isset($settings['mailpoet_subscribe_old_woocommerce_customers']['enabled'])) {
+        $new_subscribe_old_woocommerce_customers = (int)$settings['mailpoet_subscribe_old_woocommerce_customers']['enabled'];
+      }
+      if ($original_subscribe_old_woocommerce_customers !== $new_subscribe_old_woocommerce_customers) {
+        $this->woocommerce_segment->synchronizeCustomers();
```
pavel-mailpoet: We should schedule this task and let the cron run it. This can take a long time to run.

lib/API/JSON/v1/Settings.php
```diff
@@ -51,6 +57,16 @@ function set($settings = []) {
       ) {
         $this->onInactiveSubscribersIntervalChange();
       }
+
+      // Sync WooCommerce Customers list on settings change
+      $new_subscribe_old_woocommerce_customers = 0;
```
pavel-mailpoet: This needs to be refactored. We shouldn't just keep adding stuff to this API function. We don't even write tests for it. 

I would suggest to:
1.  Get all settings from the settings controller at the beginning.
2. Run the save.
3. Get all settings from the settings controller and create a diff.
4. Create another class(es) that would have a method `onSettingsChange` and run the code there.

NeosinneR: I refactored it but left the code in the same file, because it seems logical for it to be there. What do you think?

lib/API/JSON/v1/Settings.php
```diff
@@ -61,6 +60,40 @@ function set($settings = []) {
     }
   }
 
+  private function onSettingsChange($old_settings, $new_settings) {
+    // Recalculate inactive subscribers
+    $old_inactivation_interval = $old_settings['deactivate_subscriber_after_inactive_days'];
+    $new_inactivation_interval = $new_settings['deactivate_subscriber_after_inactive_days'];
```
pavel-mailpoet: Do we also need `isset` here?

NeosinneR: I don't think so, this setting should always be set. The other one depends on WooCommerce being active and also when it's disabled, the value is not in the database at all.

PR #2073, opened 2019-05-23
Validate from address before sending newsletter [MAILPOET-1985]

NeosinneR:


assets/js/src/newsletters/send.jsx
```diff
@@ -34,11 +34,13 @@ const NewsletterSend = createReactClass({ // eslint-disable-line react/prefer-es
       item: {},
       loading: true,
       thumbnailPromise: null,
+      authorizedEmailAddresses: [],
     };
   },
 
   componentDidMount: function componentDidMount() {
     this.loadItem(this.props.match.params.id);
+    this.loadAuthorizedEmailAddresses();
```
costasovo: It makes sense to do this API call only if sending method is set to MailPoet so I think it would be better to call it only for MailPoet sending method to save unnecessary API call.

Also this is not chained or combined with loadItem, which turns off loading state once it is done. There is a possibility that user will be able to fill the form but the authorized email validation will fail because emails will not be loaded. The probability is rather low but it the api call does another api call in background (Bridge API) so it is still possible.

MailPoet.Ajax.Post returns jQuery deferred objects which could be combined using [jQuery.when](https://api.jquery.com/jquery.when/)

assets/js/src/newsletters/send.jsx
```diff
@@ -74,6 +74,38 @@ const NewsletterSend = createReactClass({ // eslint-disable-line react/prefer-es
     return jQuery('#mailpoet_newsletter').parsley().isValid();
   },
 
+  isValidFromAddress: function isValidFromAddress() {
+    const fromAddress = jQuery('#field_sender_address').val();
```
costasovo: You should be able to read fromAddress from state (`this.state.item. sender_address `) which is I think a nicer way of doing it.

assets/js/src/newsletters/send.jsx
```diff
@@ -74,6 +74,38 @@ const NewsletterSend = createReactClass({ // eslint-disable-line react/prefer-es
...
+    if (window.mailpoet_mta_method !== 'MailPoet') {
+      return true;
+    }
+    return this.state.authorizedEmailAddresses.indexOf(fromAddress) !== -1;
+  },
+
+  showInvalidFromAddressError: function showInvalidFromAddressError() {
+    const errorMessage = MailPoet.I18n.t('newsletterInvalidFromAddress')
+      .replace('%$1s', jQuery('#field_sender_address').val())
+      .replace('[i]', '<i>')
```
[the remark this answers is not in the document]
NeosinneR: Didn't know about using simple tags like `<i>` in strings 👍 

About the `ReactStringReplace` - I was thinking about using it, but ended up with `.replace` because it looks more readable (especially now with removed replace for `[i]`).

```
const errorMessage = ReactStringReplace(
  MailPoet.I18n.t('newsletterInvalidFromAddress')
    .replace('%$1s', this.state.item.sender_address),
  /\[link\](.*?)\[\/link\]/g,
  (match) => (
    <a href="https://account.mailpoet.com/authorization" target="_blank" rel="noopener noreferrer">{match}</a>
  )
);
```

vs.

```
const errorMessage = MailPoet.I18n.t('newsletterInvalidFromAddress')
  .replace('%$1s', this.state.item.sender_address)
  .replace('[link]', '<a href="https://account.mailpoet.com/authorization" target="_blank" rel="noopener noreferrer">')
  .replace('[/link]', '</a>');
```

What do you think?

NeosinneR:
@michelleshull I can't replicate this. Also, it doesn't seem that this feature would break something on listing page, there were no changes in those parts of the code.

michelleshull:
I had a mismatch between Premium and free versions, which is what broke the stats links for me, so mystery solved there. Looks great, passing on to product.

PR #2078, opened 2019-05-23
Automate release branch preparation [MAILPOET-2004]

pavel-mailpoet:


RoboFile.php
```diff
@@ -557,18 +557,42 @@ public function releaseCheckIssues($version = null) {
...
       ->run();
+    // make sure release branch doesn't exist on github
+    $release_branch_status = $this->taskGitStack()
+      ->printOutput(false)
+      ->exec('git ls-remote --heads git@github.com:mailpoet/mailpoet.git release')
+      ->run();
+    if (strlen(trim($release_branch_status->getMessage())) > 0) {
+      throw new \Exception('Delete old release branch before running release.');
+    }
+    $git_status = $this->taskGitStack()
```
JanJakes: This checks local "release" branch and deletes it if it exists? Comment would be useful, Git's CLI commands are incredibly unintuitive.

RoboFile.php
```diff
@@ -553,6 +556,16 @@ public function releaseCheckIssues($version = null) {
     }
   }
 
+  public function releasePrepareGit() {
+    $git_status = $this->taskGitStack()
+      ->printOutput(false)
+      ->exec('git status --porcelain')
+      ->run();
+    if (strlen(trim($git_status->getMessage())) > 0) {
+      throw new \Exception('Please make sure your working directory is clean before running release.');
```
JanJakes: In some other places we use `yell()` and `exit(1)` but I guess this will have the same result?

RoboFile.php
```diff
@@ -595,6 +598,18 @@ public function releasePrepareGit() {
       ->run();
   }
 
+  public function releaseCreatePullRequest($version) {
+    $this->taskGitStack()
+      ->stopOnFail()
+      ->add('-A')
+      ->commit('Release ' . $version)
+      ->tag($version)
```
JanJakes: Are you creating a `tag` here already or what does this do? The release is tagged automatically when created in GitHub releases. Here, I think we don't want to tag yet since the pull request may not be approved. Not sure what this does exactly though.

RoboFile.php
```diff
@@ -613,6 +613,9 @@ public function releaseCreatePullRequest($version) {
   public function releasePublish($version = null) {
     $version = $this->releaseVersionGetNext($version);
     return $this->collectionBuilder()
+      ->addCode(function ()  use ($version) {
```
JanJakes: Extra space in there. Unfortunately this seems to be one of the files we don't sniff...

tasks/release/GitHubController.php
```diff
@@ -32,6 +35,12 @@ public function __construct($username, $token, $project) {
       ],
       'base_uri' => "https://api.github.com/repos/mailpoet/$github_path/",
     ]);
+    $this->http_client_no_base = new Client([
```
JanJakes: Actually not needed since Guzzle will ignore base in a request where you pass full absolute URL.

tasks/release/GitHubController.php
```diff
@@ -59,6 +68,74 @@ private function assignPullRequest($pull_request_number) {
     ]);
   }
 
+  public function checkReleasePullRequestPassed($version) {
```
JanJakes: Wow, this is some detailed checking 👍

RoboFile.php
pavel-mailpoet: Really? I always created the tag locally when releasing...  This does create a new git tag.

JanJakes: The `release:publish-github` handles that. Actually, GitHub handles that - same as in UI:

Screenshot 2019-05-28 at 11 33 21

PR #2079, opened 2019-05-27
Update readme.txt for the new free plan [MAILPOET-2052]

pavel-mailpoet:


michelleshull:
Sending back to remove feature flag so we can release.

PR #2083, opened 2019-05-27
Change the course form [MAILPOET-2043]

pavel-mailpoet:


assets/js/src/wizard/steps/email_course_step.jsx
```diff
@@ -12,10 +12,12 @@ const WelcomeWizardEmailCourseStep = props => (
       scrolling="no"
       frameBorder="0"
       title="Apply to course"
-      src="https://newsletters.mailpoet.com?mailpoet_form_iframe=13"
+      src="https://www.mailpoet.com?mailpoet_form_iframe=13"
       className="mailpoet_form_iframe"
+      onLoad="if (window['MailPoet']) MailPoet.Iframe.autoSize(this);"
```
NeosinneR: 1. There is no `MailPoet.Iframe` object - it would throw an error
2. This `onLoad` is processed by React and not HTML and so expecting different format.

PR #2090, opened 2019-05-30
Create custom fields api [MAILPOET-2041]

pavel-mailpoet:


lib/API/MP/v1/API.php
```diff
@@ -82,6 +88,23 @@ function getSubscriberFields() {
...
   }
 
+  function addSubscriberField(array $data = []) {
+    try {
+      $custom_field = CustomField::createOrUpdate($this->custom_fields_data_sanitizer->sanitize($data));
+      $errors = $custom_field->getErrors();
+      if (!empty($errors)) {
+        throw new \Exception('Failed to save a new subscriber field');
+      }
+      $custom_field = CustomField::findOne($custom_field->id);
```
JanJakes: Is it really necessary to fetch it again? I think it shouldn't be - when there were no errors in `createOrUpdate` then you should be able to rely on the field being saved and `$custom_field` containing it.

lib/CustomFields/ApiDataSanitizer.php
```diff
@@ -0,0 +1,164 @@
...
+    $this->checkParamsType($data);
+    return [
+      'name' => $data['name'],
+      'type' => strtolower($data['type']),
+      'params' => $this->sanitizeParams($data),
+    ];
+  }
+
+  private function checkMandatoryStringParameter(array $data, $parameter_name) {
+    if (!isset($data[$parameter_name]) || empty($data[$parameter_name])) {
```
JanJakes: Only the `empty(...)` part would be enough here: https://3v4l.org/TihcG

lib/CustomFields/ApiDataSanitizer.php
```diff
@@ -0,0 +1,164 @@
...
+      throw new InvalidArgumentException(sprintf(__('Mandatory argument "%s" is missing', 'mailpoet'), $parameter_name));
+    }
+    if (!is_string($data[$parameter_name])) {
+      throw new InvalidArgumentException(sprintf(__('Mandatory argument "%s" has to be string', 'mailpoet'), $parameter_name));
```
JanJakes: Do we need to translate this? Is it just for consistency with other endpoints?

lib/CustomFields/ApiDataSanitizer.php
```diff
@@ -0,0 +1,164 @@
...
+      if (in_array($validate, ['number', 'alphanum', 'phone'], true)) {
+        return ['validate' => $validate];
+      }
+      throw new InvalidArgumentException(__('Validate parameter is not valid', 'mailpoet'));
+    }
+    return [];
+  }
+
+  private function getExtraParamsForCheckbox($params) {
+    if (!isset($params['values']) || empty($params['values']) || count($params['values']) > 1) {
```
JanJakes: Only `empty` is sufficient here.

lib/CustomFields/ApiDataSanitizer.php
```diff
@@ -0,0 +1,164 @@
...
+        throw new InvalidArgumentException(__('Invalid value for date_type', 'mailpoet'));
+    }
+    return [
+      'date_type' => $date_type,
+      'date_format' => $date_format,
+    ];
+  }
+
+  private function getExtraParamsForSelect($params) {
+    if (!isset($params['values']) || empty($params['values'])) {
```
JanJakes: Only `empty` is sufficient here.

lib/CustomFields/ApiDataSanitizer.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+    $values = [];
+    foreach ($params['values'] as $value) {
+      $values[] = $this->sanitizeValue($value);
+    }
+    return ['values' => $values];
+  }
+
+  private function sanitizeValue($value) {
+    if (!isset($value['value']) || empty($value['value'])) {
```
JanJakes: Only `empty` is sufficient here.

lib/CustomFields/ApiDataSanitizer.php
JanJakes: But maybe here (and in other cases where `empty` is used) it should be considered that even some non-empty value are considered empty by this function (like string `'0'`) - https://www.php.net/empty. It's really an edge case thought so probably not much to worry about.

doc/Readme.md
```diff
@@ -22,6 +22,7 @@ Class `\MailPoet\API\API` becomes available once MailPoet plugin is loaded by Wo
 ### Available API Methods
 * [Add List (addList)](api_methods/AddList.md)
 * [Add Subscriber (addSubscriber)](api_methods/AddSubscriber.md)
+* [Add Subscriber Field(addSubscriberField)](api_methods/AddSubscriberField.md)
```
JanJakes: Missing space between `Field` and `(`.

doc/api_methods/AddSubscriberField.md
```diff
@@ -0,0 +1,65 @@
...
+| Property | Type | Description |
+| --- | --- | --- |
+| values | array | Contains a list of options. Each element must contain a string `value` and can contain `is_checked` |
+
+#### `$params` for date type
+
+| Property | Type | Description |
+| --- | --- | --- |
+| date_type | string | Possible values are: Values: `year_month_day`, `year_month`, `month`, `day` |
+| date_format | string | Values: for year_month_day: 'MM/DD/YYYY', 'DD/MM/YYYY', 'YYYY/MM/DD', for year_month: 'YYYY/MM', 'MM/YY', for year: 'YYYY', for month: 'MM' |
```
JanJakes: This uses `'` instead of backticks `.

doc/api_methods/AddSubscriberField.md
```diff
@@ -0,0 +1,65 @@
...
+## Response
+
+| Property | Type | Limits | Description |
+| --- | --- | --- | --- |
+| id | string | 11 chars |Field Id |
+| name | string | 90 chars | Human readable name. Intended to be used, as an example, as a label for form input. |
+| type | string | - | Type of the field. Possible values are: `text`, `date`, `textarea`, `radio`, `checkbox`, `select` |
+| params | array | - | Contains various information, see examples below. |
+| created_at | string\|null | - | UTC time of creation in 'Y-m-d H:i:s' format |
+| updated_at | string | - | UTC time of last update in 'Y-m-d H:i:s' format |
```
JanJakes: This uses `'` instead of backticks `.

doc/api_methods/GetSubscriberFields.md
```diff
@@ -7,6 +7,7 @@
 Each subscriber has a set of default properties (`email`, `first_name`, `last_name`).
 MailPoet enables the extension of these properties by adding custom properties.
 This method returns list of all properties available for a subscriber (default + custom).
+See also [addSubscriberField function](./AddSubscriberField.md)
```
JanJakes: Personally, I'd ended this with a dot.

lib/API/MP/v1/API.php
pavel-mailpoet: It is consistent with other methods in the API. 

Also without this, we wouldn't return all the fields. `created_at` and `updated_at` would not be populated.

lib/CustomFields/ApiDataSanitizer.php
pavel-mailpoet: Correct, all the exceptions in the mp api are translated so I did the same.

PR #2095, opened 2019-06-03
Inactive subscribers notice [MAILPOET-2088]

JanJakes:


lib/Util/Notices/InactiveSubscribersNotice.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Models\Subscriber;
+use MailPoet\Settings\SettingsController;
+use MailPoet\Util\Helpers;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class InactiveSubscribersNotice {
```
pavel-mailpoet: No tests? Why do you think we don't need them for this class?

JanJakes: The ticket says no tests needed but thinking about it... I'd add some integration tests at least.

PR #2098, opened 2019-06-04
Add codes to MP API exceptions [MAILPOET-2107]

pavel-mailpoet:


doc/api_methods/AddList.md
```diff
@@ -17,3 +17,18 @@ An associative array which contains list data.
 | --- | --- | --- | --- |
 | name (required) | string | 90 chars | A name of the list. |
 | description (optional) | string\|null| 250 chars | A description of the list. |
+
+## Error handling
```
wxa: This text is repeated in all docs, maybe we should create a separate document about error handling and move it there? Not sure whether to leave error codes where they are or to move them as well. If the codes of different methods don't collide then they shouldn't be confusing if organized on a single page, should they?

doc/api_methods/AddSubscriberField.md
```diff
@@ -63,3 +63,25 @@ The common properties for all types:
...
+Code of the exception is populated to distinguish between different errors.
+
+An exception of base class `\Exception` can be thrown when something unexpected happens.
+
+Codes description:
+
+| Code | Description |
+| --- | ---  |
+| 1 | The subscriber couldn’t be created in the database |
+| 1001 | Missing a mandatory field in the `$data` argument |
```
wxa: Why do these validation errors have a different numeration scheme? Shouldn't then all error codes be uniform (e.g. 1XXX - general errors, 2XXX - validation errors etc.)?

lib/API/MP/v1/API.php
```diff
@@ -93,15 +93,15 @@ function addSubscriberField(array $data = []) {
       $custom_field = CustomField::createOrUpdate($this->custom_fields_data_sanitizer->sanitize($data));
       $errors = $custom_field->getErrors();
       if (!empty($errors)) {
-        throw new \Exception('Failed to save a new subscriber field');
+        throw new APIException('Failed to save a new subscriber field ' . join(', ', $errors), 1);
```
wxa: For better readability I think we should use constants for error codes instead of bare numbers. It would be also beneficial for consuming code to do easier comparations against them.

lib/CustomFields/ApiDataSanitizer.php
```diff
@@ -20,16 +20,16 @@ function sanitize(array $data = []) {
 
   private function checkMandatoryStringParameter(array $data, $parameter_name) {
     if (empty($data[$parameter_name])) {
-      throw new InvalidArgumentException(sprintf(__('Mandatory argument "%s" is missing', 'mailpoet'), $parameter_name));
+      throw new InvalidArgumentException(sprintf(__('Mandatory argument "%s" is missing', 'mailpoet'), $parameter_name), 1001);
```
wxa: The docs don't mention that an `InvalidArgumentException` can be thrown, or was this implied by the "exception of base class `\Exception`" line?

doc/api_methods/AddList.md
pavel-mailpoet: Hmm, I was thinking about that and I decided to copy and paste it everywhere. The reason being is the convenience for the reader. So when they are working on a feature they can see everything in a single page without needing to go back and forth between different pages. 

> the codes of different methods don't collide

They sometimes do when they throw a similar exception, for example UnsubscribeFromLists and SubscribeToLists share codes.

So if we wanted to create a single page with all the codes we would have to separate the code by the function anyway, so I thought it is better to leave it here.

doc/api_methods/AddSubscriberField.md
pavel-mailpoet: No big reason really. It is just because I was adding those errors in a different file so I used a different scheme. Just to prevent future collisions.

I was thinking about your suggestion, but it is quite hard to separate the errors into different classes to be any useful so I didn't do that.

lib/API/MP/v1/API.php
pavel-mailpoet: Good idea, I'll do that...

lib/CustomFields/ApiDataSanitizer.php
pavel-mailpoet: This error is caught in the main MP API class and converted to `APIException` so we don't need to mention it.

lib/API/MP/v1/APIException.php
```diff
@@ -4,4 +4,19 @@
...
+  const SUBSCRIBE_TO_WC_LIST_NOT_ALLOWED = 7;
+  const SUBSCRIBE_TO_LIST_NOT_ALLOWED = 8;
+  const CONFIRMATION_FAILED_TO_SEND = 10;
+  const REQUIRED_EMAIL_ADDRESS = 11;
+  const SUBSCRIBER_EXIST = 12;
+  const FAILED_TO_SAVE_SUBSCRIBER = 13;
+  const REQUIRED_LIST_NAME = 14;
+  const LIST_EXIST = 15;
+  const FAILED_TO_SAVE_LIST = 16;
+  const WELCOME_FAILED_TO_SEND = 17;
```
wxa: Looks good, I would just correct some spelling:
* `*_EXIST` => `*_EXISTS`;
* `REQUIRED_*` => `*_REQUIRED`;
* `SUBSCRIBE_TO_*` => `SUBSCRIBING_TO_*`.

pavel-mailpoet: Thanks, updated.

PR #2101, opened 2019-06-05
Fix registration of rejected users [MAILPOET-2114]

pavel-mailpoet:
There are multiple hooks for user registration.
https://usersinsights.com/wp/wp-content/uploads/2018/07/user-registration-hooks-2.png
The `register_post` hook is executed first and
after that, there is `registration_errors`.
The captcha verification is done in `registration_errors` hook.
But because we were using `register_post` we were saved
the subscriber before the post request was verified.

### DIFF
--- a/lib/Config/Hooks.php
+++ b/lib/Config/Hooks.php
@@ -132,8 +132,12 @@ class Hooks {
           'register_form',
           [$this->subscription_registration, 'extendForm']
         );
-        $this->wp->addAction(
-          'register_post',
+        // we need to process new users while they are registered.
+        // We used `register_post` before but that is too soon
+        //   because if registration fails during `registration_errors` we will keep the user as subscriber.
+        // So we are hooking to `registration_error` with a low priority.
+        $this->wp->addFilter(
+          'registration_errors',
           [$this->subscription_registration, 'onRegister'],
           60,
           3
--- a/lib/Subscription/Registration.php
+++ b/lib/Subscription/Registration.php
@@ -54,9 +54,9 @@ class Registration {
   }
 
   function onRegister(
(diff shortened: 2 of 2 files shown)


lib/Config/Hooks.php
```diff
@@ -132,8 +132,8 @@ function setupSubscriptionEvents() {
           'register_form',
           [$this->subscription_registration, 'extendForm']
         );
-        $this->wp->addAction(
-          'register_post',
+        $this->wp->addFilter(
+          'registration_errors',
```
JanJakes: I understand the problem but is using `registration_errors` a good fix? Isn't this hook used rather for error processing or validation? Looking at https://usersinsights.com/wp/wp-content/uploads/2018/07/user-registration-hooks-2.png I have a question - could we use `user_register` or `register_new_user`? I.e. we would register after the user is really saved in WordPress DB?

pavel-mailpoet: I don't think so. Maybe? I am not sure. 

We are already using `user_register` to catch any user creation and create the user in the `\MailPoet\Segments\WP::synchronizeUser` function. That hook can be created from anywhere where a user can be created

This is to only run the code in the registration form and get the data from it.

JanJakes: Hm, I don't see the whole context, I just thought hooking on `registration_errors` with our `onRegister` action looks weird. If you think it's better to keep it in this hook, we can do it but it may look confusing to someone one day.

pavel-mailpoet: I see what you mean. I don't think `user_register` is a better place and I don't know what would be better than `registration _errors`. I'll leave a comment.

PR #2103, opened 2019-06-06
Siteurl -> homeurl

utolosa002:
If the site is not equal to home, It doesn't render well images.

### DIFF
--- a/lib/Form/Util/Export.php
+++ b/lib/Form/Util/Export.php
@@ -20,7 +20,7 @@ class Export {
         // generate url to load iframe's content
         $iframe_url = WPFunctions::get()->addQueryArg([
           'mailpoet_form_iframe' => $form['id'],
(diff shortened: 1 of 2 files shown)


flakas:
Hey @utolosa002, thanks for the PR.
We are still evaluating the impact of this change, as in different setups it may have different implications that aren't immediately obvious.

Thanks for your patience and for your help.

michelleshull:
Hi there, @utolosa002 - I'm Michelle, the QA. I've been doing some testing on your fix, and while I haven't found any immediate conflicts, I think we can address this better programmatically from the plugin, so we can account for edge cases that I haven't thought of or seen yet. I've opened a ticket to have our devs take a look. Thanks so much for your time and effort on this, you've helped us see a gap we can close in an even more thorough way. Please consider contributing again, we greatly appreciate your efforts. :)

PR #2109, opened 2019-06-18
 Add command to fix code style on a single file [MAILPOET-2128]

amine-mp:


RoboFile.php
```diff
@@ -351,6 +351,53 @@ function qaCodeSniffer($severity='errors') {
...
+      return $this->collectionBuilder()
+        ->taskExec(
+          'npx eslint -c .eslintrc.es5.json ' .
+            '--max-warnings 0 ' .
+            '--fix ' .
+            $filePath
+        )
+        ->run();
+    }
+    if (substr($filePath, -8) === '.spec.js') {
```
JanJakes: This will be first run through the previous IF since it also ends with `.js`. I suggest using `elseif` and changing the order.

RoboFile.php
```diff
@@ -351,6 +351,53 @@ function qaCodeSniffer($severity='errors') {
...
 
+  function qaFixFile($filePath) {
+    if (substr($filePath, -4) === '.php') {
+      // fix PHPCS rules
+      return $this->collectionBuilder()
+        ->taskExec(
+          './tasks/code_sniffer/vendor/bin/phpcbf ' .
+            '--standard=./tasks/code_sniffer/MailPoet ' .
+            '--runtime-set testVersion 5.6-7.3 ' .
+            $filePath . ' -n'
```
JanJakes: I don't see any `fix` argument here. Or is that the `-n`?

amine-mp: Here I am using `phpcbf` instead of `phpcs` to fix the errors. The `-n` flag is to not print warnings.

RoboFile.php
amine-mp: I need to change the order but no need to use `elseif` because I am doing `return`.

PR #2110, opened 2019-06-18
Sending error notice fix [MAILPOET-2115]

JanJakes:


amine-mp:
The code is verbose for me, and there should be a better solution. But since it will require a lot of refactoring I am just approving this for now.

PR #2111, opened 2019-06-18
Update screenshots of templates [MAILPOET-2102]

JanJakes:


JanJakes:
@NeosinneR I noticed but thought it doesn't matter since this goes to the plugin repo's CDN, not to the plugin. I just checked and it seems they don't resize it though 🤔 Do you think we should optimize it? And if so, do you know a good service for that?

NeosinneR:
I'm using https://compressor.io/compress (online) for JPGs and https://pngmini.com/ (app) for PNGs. I tried running it through it, down to 143KB, I think better than nothing.
screenshot-1

JanJakes:
@NeosinneR Repushed.

PR #2112, opened 2019-06-19
Undo/redo [MAILPOET-1976]

NeosinneR:


assets/js/src/newsletter_editor/components/history.js
```diff
@@ -19,6 +25,12 @@ Module.HistoryView = Marionette.View.extend({
     App.getChannel().on('afterEditorSave', this.setCurrentState, this);
   },
 
+  onAttach: function onRender() {
+    this.elements.redo = jQuery('#mailpoet-history-arrow-redo');
+    this.elements.undo = jQuery('#mailpoet-history-arrow-undo');
```
JanJakes: Just an idea... do we really need jQuery here? I think it's easy without.

assets/js/src/newsletter_editor/components/history.js
```diff
@@ -29,6 +41,26 @@ Module.HistoryView = Marionette.View.extend({
...
+
+  updateArrowsUI: function updateArrowsUI() {
+    this.elements.undo.addClass('mailpoet_history_arrow_inactive');
+    this.elements.redo.addClass('mailpoet_history_arrow_inactive');
+    if (this.canUndo()) {
+      this.elements.undo.removeClass('mailpoet_history_arrow_inactive');
+    };
+    if (this.canRedo()) {
+      this.elements.redo.removeClass('mailpoet_history_arrow_inactive');
+    };
```
JanJakes: Even without jQuery this could just be:
```
this.elements.undo.classList.toggle('mailpoet_history_arrow_inactive', !this.canUndo());
this.elements.redo.classList.toggle('mailpoet_history_arrow_inactive', !this.canRedo());
```

views/newsletter/editor.html
```diff
@@ -409,6 +409,10 @@ <h1 style="display:none"><%= __('Newsletter Editor') %></h1>
     'template': __('Template'),
     'designer': __('Designer'),
     'send': __('Send'),
+    'canUndo': __('Undo'),
+    'canNotUndo': __('No past history available.'),
+    'canRedo': __('Redo'),
+    'canNotRedo': __('No future history available.'),
```
JanJakes: The two longer phrases may need some context that they belong to "undo/redo" feature. Also it should have domain parameter with `mailpoet` (https://developer.wordpress.org/reference/functions/__/) although this file seems to have none.

assets/js/src/newsletter_editor/components/history.js
```diff
@@ -44,7 +44,7 @@ Module.HistoryView = Marionette.View.extend({
     if (this.model.currentStateIndex > 0) {
       this.model.states.splice(0, this.model.currentStateIndex);
     }
-    this.model.states.unshift(json.body);
+    this.model.states.unshift(JSON.stringify(json.body));
```
JanJakes: Does this really need to be stringified (and later parsed I suppose)?

assets/js/src/newsletter_editor/components/history.js
```diff
@@ -62,12 +62,18 @@ Module.HistoryView = Marionette.View.extend({
     if (!this.canUndo()) {
       return;
     }
+    this.model.currentStateIndex += 1;
```
JanJakes: Could it occur by some mistake you'd get out of the array bounds with this index? And for "redo" below 0? Should we check/sanitize it here (i.e. with min/max functions)?

assets/js/src/newsletter_editor/components/history.js
```diff
@@ -29,6 +41,26 @@ Module.HistoryView = Marionette.View.extend({
...
     this.model.states.length = Math.min(this.model.states.length, this.MAX_HISTORY_STATES);
+    this.updateArrowsUI();
+  },
+
+  canUndo: function canUndo() {
+    return this.model.currentStateIndex < (this.model.states.length - 1)
+  },
+
+  canRedo: function canRedo() {
+    return this.model.currentStateIndex !== 0;
```
JanJakes: For consistency with `canUndo` using `> 0` would be a bit nicer.

assets/js/src/newsletter_editor/components/content.js
```diff
@@ -63,6 +63,21 @@ Module.findModels = function findModels(predicate) {
   return _.filter(blocks, predicate);
 };
 
+Module.renderContent = function renderContent(content) {
+  if (App._contentContainerView) {
+    App._contentContainer.destroy();
+  }
+  if (App._contentContainerView) {
```
JanJakes: Both IFs check for `_contentContainerView` but one destroys `_contentContainer` and other one `_contentContainerView`. Is that intentional?

assets/js/src/newsletter_editor/components/history.js
```diff
@@ -28,7 +29,14 @@ Module.HistoryView = Marionette.View.extend({
   },
 
   initialize: function initialize() {
+    var that = this;
     App.getChannel().on('afterEditorSave', this.addState, this);
+    Mousetrap.bind(['ctrl+z', 'command+z'], function keyboardUndo() {
+      that.undo();
+    });
```
JanJakes: Can this be written just like `Mousetrap.bind(['ctrl+z', 'command+z'], this.undo)`?

tests/acceptance/EditorHistoryCest.php
```diff
@@ -0,0 +1,44 @@
...
+
+    // Keyboard undo
+    $I->pressKey('body', [\WebDriverKeys::CONTROL, 'z']);
+    $I->waitForElementNotVisible(self::BUTTON_SELECTOR);
+
+    // Keyboard redo
+    $I->pressKey('body', [\WebDriverKeys::SHIFT, \WebDriverKeys::CONTROL, 'z']);
+    $I->waitForElementVisible(self::BUTTON_SELECTOR);
+  }
+
```
JanJakes: If it's not difficult it could also assert that undo/redo buttons are/aren't clickable.

assets/js/src/newsletter_editor/components/content.js
NeosinneR: That's a typo, fixed

assets/js/src/newsletter_editor/components/history.js
NeosinneR: It can't :( Inside `undo`/`redo` methods, they are using `this`, which would be undefined if written as you wrote.

assets/css/src/components/newsletterEditor/components/_history.scss
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.mailpoet_history_arrow {
+  cursor: pointer;
+  margin-right: 20px;
+
+  svg {
+    display: inline-block;
+    height: 26px;
+    stroke: #0074a2;
```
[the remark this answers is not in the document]
NeosinneR: Completely forgot about variables, fixed

assets/js/src/newsletter_editor/components/history.js
NeosinneR: In later commits, I added a comparison of the current state with the one being saved (so you won't create same states just by clicking on Save button), and for that I need both states stringified.

PR #2116, opened 2019-06-20
Prevent importing role addresses as subscribers [MAILPOET-2101]

pavel-mailpoet:


assets/js/src/subscribers/importExport/import.jsx
```diff
@@ -157,17 +157,26 @@ jQuery(document).ready(() => {
...
       jQuery('#subscribers_data_parse_results').html(
         subscribersDataParseResultsTemplate(importResults)
       );
     }
 
     jQuery('.mailpoet_subscribers_data_parse_results_details_show')
-      .click((event) => {
+      .click(function detailsClick() {
         const details = jQuery('.mailpoet_subscribers_data_parse_results_details');
         jQuery(details).toggle();
```
NeosinneR: Not part of this issue, but `jQuery(details).toggle();` can be also written as `details.toggle();` - the `jQuery` is unnecessary.

PR #2121, opened 2019-06-20
Move Symfony polyfills to vendor-prefixed [MAILPOET-2011]

pavel-mailpoet:


prefixer/scoper.inc.php
```diff
@@ -35,7 +35,16 @@
   'patchers' => [
     function (string $filePath, string $prefix, string $contents): string {
       // Change the contents here.
-
+      if (
+        strpos($filePath, 'vendor/symfony/polyfill-mbstring/bootstrap.php')
+        || strpos($filePath, 'vendor/symfony/polyfill-php72/bootstrap.php')
+      ) {
+        return preg_replace(
```
JanJakes: It's not very important but I'd suggest using just `str_replace`.

PR #2126, opened 2019-06-24
Beamer API get last announcement [MAILPOET-2037]

amine-mp:
[MAILPOET-2037]

[MAILPOET-2037]: https://mailpoet.atlassian.net/browse/MAILPOET-2037

### DIFF
--- a/lib/Config/Menu.php
+++ b/lib/Config/Menu.php
@@ -706,10 +706,11 @@ class Menu {
 
     $user_id = $data['current_wp_user']['ID'];
 
+    $last_announcement_date = $this->settings->get('last_announcement_date');
     $last_announcement_seen = $this->user_flags->get('last_announcement_seen');
     $data['feature_announcement_has_news'] = (
       empty($last_announcement_seen) ||
-      $last_announcement_seen < strtotime(self::LAST_ANNOUNCEMENT_DATE)
+      $last_announcement_seen < $last_announcement_date
     );
     $data['last_announcement_seen'] = $last_announcement_seen;
 
--- a/lib/Config/Menu.php
+++ b/lib/Config/Menu.php
@@ -38,7 +38,6 @@ if (!defined('ABSPATH')) exit;
 
 class Menu {
   const MAIN_PAGE_SLUG = 'mailpoet-newsletters';
-  const LAST_ANNOUNCEMENT_DATE = '2019-06-10 10:00:00';
 
   /** @var WooCommerceHelper */
   private $woocommerce_helper;
--- a/lib/Config/Populator.php
+++ b/lib/Config/Populator.php
@@ -5,6 +5,7 @@ use Carbon\Carbon;
 use MailPoet\Config\PopulatorData\DefaultForm;
 use MailPoet\Cron\CronTrigger;
 use MailPoet\Cron\Workers\AuthorizedSendingEmailsCheck;
+use MailPoet\Cron\Workers\Beamer;
 use MailPoet\Cron\Workers\InactiveSubscribers;
 use MailPoet\Mailer\MailerLog;
 use MailPoet\Models\NewsletterTemplate;
@@ -139,6 +140,7 @@ class Populator {
     $this->updateMetaFields();
     $this->scheduleInitialInactiveSubscribersCheck();
     $this->scheduleAuthorizedSendingEmailsCheck();
+    $this->initLastAnnouncementDate();
     // Will be uncommented on task [MAILPOET-1998]
     // $this->updateFormsSuccessMessages();
   }
@@ -547,6 +549,13 @@ class Populator {
       AuthorizedSendingEmailsCheck::TASK_TYPE,
       Carbon::createFromTimestamp(WPFunctions::get()->currentTime('timestamp'))
     );
+  }
+
+  private function initLastAnnouncementDate() {
+    if (!$this->settings->get('last_announcement_date')) {
+      $beamer = new Beamer($this->settings);
+      $beamer->setLastAnnouncementDate();
+    }
   }
 
   private function scheduleTask($type, $datetime) {
--- a/lib/Config/Populator.php
+++ b/lib/Config/Populator.php
@@ -553,7 +553,7 @@ class Populator {
 
   private function initLastAnnouncementDate() {
     if (!$this->settings->get('last_announcement_date')) {
-      $beamer = new Beamer($this->settings);
+      $beamer = new Beamer($this->settings, WPFunctions::get());
       $beamer->setLastAnnouncementDate();
     }
   }
--- a/lib/Config/Populator.php
+++ b/lib/Config/Populator.php
@@ -37,10 +37,13 @@ class Populator {
   private $default_segment;
   /** @var SettingsController */
   private $settings;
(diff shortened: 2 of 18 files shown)


lib/Cron/Daemon.php
```diff
@@ -51,8 +51,9 @@ private function getWorkers() {
     yield $this->workers_factory->createPremiumKeyCheckWorker($this->timer);
     yield $this->workers_factory->createBounceWorker($this->timer);
     yield $this->workers_factory->createExportFilesCleanupWorker($this->timer);
+    yield $this->workers_factory->createBeamerkWorker($this->timer);
     yield $this->workers_factory->createInactiveSubscribersWorker($this->timer);
     yield $this->workers_factory->createWooCommerceSyncWorker($this->timer);
-    yield $this->workers_factory->createAuthorizedSendingEmailsCheckWorker($this->timer);
+    yield $this->workers_factory-> createAuthorizedSendingEmailsCheckWorker($this->timer);
```
pavel-mailpoet: Why did you modify this line?

lib/Config/Populator.php
```diff
@@ -549,6 +551,13 @@ private function scheduleAuthorizedSendingEmailsCheck() {
     );
   }
 
+  private function initLastAnnouncementDate() {
```
pavel-mailpoet: I think we should rather schedule the task instead of getting the date directly here. It can take some time to load the Beamer API and our plugin activation/upgrade is already slow. We should keep it as fast as possible.

lib/Config/Populator.php
```diff
@@ -549,6 +551,13 @@ private function scheduleAuthorizedSendingEmailsCheck() {
...
+    if (!$this->settings->get('last_announcement_date')) {
+      $beamer = new Beamer($this->settings, WPFunctions::get());
```
pavel-mailpoet: Just thinking if we should pass the instance through DI here... What do you think?

lib/Cron/Workers/Beamer.php
```diff
@@ -0,0 +1,51 @@
...
+
+  function setLastAnnouncementDate() {
+    $response = $this->wp->wpRemoteGet(self::API_URL . '/posts?published=true&maxResults=1', [
+      'headers' => [
+        'Beamer-Api-Key' => self::API_KEY,
+      ],
+    ]);
+    $posts = $this->wp->wpRemoteRetrieveBody($response);
+    if (empty($posts)) return false;
+    $posts = json_decode($posts);
```
pavel-mailpoet: Should we check if `$posts` contains the correct data before trying to use them? If the network request didn't finish properly we could get invalid JSON or JSON with unexpected structure.

lib/Config/Menu.php
```diff
@@ -706,10 +706,11 @@ function newsletters() {
 
     $user_id = $data['current_wp_user']['ID'];
 
+    $last_announcement_date = $this->settings->get('last_announcement_date');
     $last_announcement_seen = $this->user_flags->get('last_announcement_seen');
     $data['feature_announcement_has_news'] = (
       empty($last_announcement_seen) ||
-      $last_announcement_seen < strtotime(self::LAST_ANNOUNCEMENT_DATE)
```
pavel-mailpoet: You should also remove the `LAST_ANNOUNCEMENT_DATE` constant definition on top of this class.

tests/integration/Cron/Workers/BeamerTest.php
```diff
@@ -0,0 +1,26 @@
+<?php
+namespace MailPoet\Test\Cron\Workers;
+
+use Carbon\Carbon;
+use Codeception\Stub;
+use MailPoet\Cron\Workers\Beamer;
+use MailPoet\Settings\SettingsController;
+use MailPoet\WP\Functions as WPFunctions;
+
+class BeamerTest extends \MailPoetTest {
```
pavel-mailpoet: First of all, I thought our coding standards require an empty line between the class header and the first method definition. The same on the end. I am surprised no check caught it. Could you please add those lines?

tests/integration/Cron/Workers/BeamerTest.php
```diff
@@ -0,0 +1,26 @@
...
+    $settings->set('last_announcement_date', Carbon::createFromTimeString($oldDate)->getTimestamp());
+    $wp = Stub::make(new WPFunctions, [
+      'wpRemoteGet' => null,
+      'wpRemoteRetrieveBody' => json_encode([
+        ['date' => $newDate],
+      ]),
+    ]);
+    $beamer = new Beamer($settings, $wp);
+    $beamer->setLastAnnouncementDate();
+    expect($settings->get('last_announcement_date'))->equals( Carbon::createFromTimeString($newDate)->getTimestamp());
```
pavel-mailpoet: Another code styling issue here. There shouldn't be space after `(`

tests/integration/Cron/Workers/BeamerTest.php
```diff
@@ -0,0 +1,26 @@
...
+namespace MailPoet\Test\Cron\Workers;
+
+use Carbon\Carbon;
+use Codeception\Stub;
+use MailPoet\Cron\Workers\Beamer;
+use MailPoet\Settings\SettingsController;
+use MailPoet\WP\Functions as WPFunctions;
+
+class BeamerTest extends \MailPoetTest {
+  function testItSetsLastAnnouncementDate() {
```
pavel-mailpoet: Should there be more tests? You only test the happy path, but should we add a test for:
- 404 response from the Beamer API
- 200 response with an empty body
- incomplete JSON that fails to decode
- valid JSON but an unexpected structure

lib/Cron/Daemon.php
amine-mp: That's my VSCode adding spaces when I paste something :(

lib/Config/Populator.php
amine-mp: I can make `$wp` a property and initialize it on the constructor, but in order to properly inject it with DI we need to refactor all classes using Populator which may easily bring a lot of refactoring.
Do you think `$wp` as property is good enough?

pavel-mailpoet: > in order to properly inject it with DI we need to refactor all classes using Populator

I don't think it is that big problem. AFAIK Populator is only used in Activator and Activator is already configured in DI container. So this is not going to be a big refactor. 

If we don't keep adding more and more classes to the DI we will never reach the goal to use DI everywhere.

lib/DI/ContainerConfigurator.php
```diff
@@ -57,6 +57,7 @@ function configure(ContainerBuilder $container) {
     // Config
     $container->autowire(\MailPoet\Config\AccessControl::class)->setPublic(true);
     $container->autowire(\MailPoet\Config\Activator::class)->setPublic(true);
+    $container->autowire(\MailPoet\Config\Populator::class)->setPublic(true);
```
pavel-mailpoet: Does this have to be public? I don't think so...

amine-mp: I agree, it does not have to be public.

michelleshull:
Awesome!!!! One less thing to think about on release day.

PR #2131, opened 2019-06-26
Remove deprecated react method [MAILPOET-2133]

pavel-mailpoet:


pavel-mailpoet:
@michelleshull I am afraid this is not going to be an easy one for testing. What has changed:

- all forms (😞 sorry)
- search in listings
- congratulate page
- newsletter date time
- listings
- the final step in newsletter sending

I had to update how those components receive their data. So either when you are editing or creating new stuff.

assets/js/src/listing/search.jsx
```diff
@@ -7,13 +7,17 @@ class ListingSearch extends React.Component {
...
+      componentMounted: false,
     };
     this.handleSearch = this.handleSearch.bind(this);
     this.onChange = this.onChange.bind(this);
   }
 
-  componentWillReceiveProps(nextProps) {
-    this.setState({ search: nextProps.search });
+  static getDerivedStateFromProps(prevProps, prevState) {
+    if (!prevState.componentMounted && prevProps.search) {
```
amine-mp: This condition passes only the first time that `props.search` has a value. If this prop changes after that, the state will not!
I think using `componentDidMount` and comparing previous and next props would be better to reproduce the same behavior as before.

assets/js/src/newsletters/send/date_time.jsx
```diff
@@ -13,8 +13,15 @@ class DateTime extends React.Component {
...
   }
 
-  componentWillReceiveProps(nextProps) {
-    this.setState(this.buildStateFromProps(nextProps));
+  componentDidUpdate(prevProps) {
+    if (
+      this.props.value !== prevProps.value
+      || this.props.defaultDateTime !== prevProps.defaultDateTime
+    ) {
+      setImmediate(() => {
```
amine-mp: Is this `setImmediate` here just to stop ESLint from complaining about using `setState`? If so then why to use a comment to make it ignore this statement?

pavel-mailpoet: That is not the only reason. The other reason is to leave react a little bit of space to finish rendering before doing the second render after the state is changed. So really just to make it more asynchronous.

assets/js/src/listing/search.jsx
pavel-mailpoet: I don't understand what do you mean... I don't see how using `componentDidMount` could help in your scenario. Isn't that only executed once at the beginning of the lifecycle? So if this prop changes after that, the state will not anyway? What am I missing?

amine-mp: Sorry, I meant `componentDidUpdate`!

PR #2132, opened 2019-06-26
Stats Page available without Premium plugin [MAILPOET-2104]

NeosinneR:


assets/js/src/newsletters/campaign_stats/page.jsx
```diff
@@ -73,8 +73,8 @@ class CampaignStatsPage extends React.Component {
 
     MailPoet.Ajax.post({
       api_version: window.mailpoet_api_version,
-      endpoint: 'stats',
-      action: 'get',
+      endpoint: window.mailpoet_premium_active ? 'stats' : 'newsletters',
+      action: window.mailpoet_premium_active ? 'get' : 'getWithStats',
```
JanJakes: What's the difference between those two? Do they return different data?

assets/js/src/newsletters/campaign_stats/newsletter_stats.jsx
```diff
@@ -47,9 +46,7 @@ const NewsletterGeneralStats = ({ newsletter }) => {
             headline={headlineClicked}
           />
         </div>
-        <RevenuesStats
-          revenue={newsletter.statistics.revenue}
-        />
+        {Hooks.applyFilters('mailpoet_newsletters_revenues_stats', null, newsletter.statistics.revenue)}
```
JanJakes: What's the second (`null`) parameter for?

lib/Config/Menu.php
```diff
@@ -703,6 +703,8 @@ function newsletters() {
     $data['premium_plugin_active'] = License::getLicense();
     $data['is_woocommerce_active'] = $this->woocommerce_helper->isWooCommerceActive();
     $data['is_mailpoet_update_available'] = array_key_exists(Env::$plugin_path, $this->wp->getPluginUpdates());
+    $data['subscribers_count'] = Subscriber::getTotalSubscribers();
```
JanJakes: This will count subscribers on every single newsletters page request, right? I'm afraid about performance - some sites have hundreds of thousands of subscribers and that's not and easy COUNT query. At the moment we use this only for some specific pages (settings, premium). Could we count it only for the stats page here?

assets/js/src/newsletters/campaign_stats/page.jsx
NeosinneR: `stats` API is available only in the premium plugin, so in free I had to call `newsletters` API, which returns basic stats like opens, clicks and unsubscribes

assets/js/src/newsletters/campaign_stats/newsletter_stats.jsx
NeosinneR: When no `Hooks.addFilters()` is called (because in free there is none) that `null` parameter is what's being rendered. And since the free plugin has no revenue stats, nothing is rendered and only premium content is visible with a hook in the premium plugin.

lib/Config/Menu.php
NeosinneR: I thought it can be pretty fast since that query looks rather simple (https://github.com/mailpoet/mailpoet/blob/master/lib/Models/Subscriber.php#L594). But I can change it to load only on the stats page.

JanJakes: In most cases it is but COUNT in SQL databases is unfortunately not read from any "global" per-table meta value because such value would cause a bottleneck in concurrent environment. Multiple concurrent transactions can see different states of DB data and COUNT query really has to go through whole primary key index to count the rows every time (https://stackoverflow.com/questions/10976328/mysql-count-performance-on-very-big-tables). For hundreds of thousands to millions of rows this can take even a second or more, depending on the power of HW.

NeosinneR: Fixed by loading `subscribers_count` only for non-premium users, which should be fast, because of the 2000 limit.

PR #2135, opened 2019-06-27
Remove react create class [MAILPOET-2151]

pavel-mailpoet:


pavel-mailpoet:
@michelleshull that is weird. How did you test this? I tried it locally and also I downloaded the `.zip` file from Circle.CI and uploaded to test. and it worked properly... Could it be some caching issue in your browser?


Also, acceptance tests passed, don't we cover this scenario already?

michelleshull:
It's on a test site where you have credentials, 3furiespublishing.com, you're pavel@mailpoet there. I had previously tested a few other tickets, this ticket is definitely where I lost the Woo functionality.

pavel-mailpoet:
It seems like you have a very old premium build because `StatisticsMixin` has been removed by Amine a couple of weeks ago. But at least you discovered there is another place with React.createClass in the premium which I missed. I am creating premium PR right now.

PR #2143, opened 2019-07-08
Add a built-in captcha [MAILPOET-2015]

wxa:
**Note to reviewer:**
*Please first check this PR: https://github.com/mailpoet/wordpress-images/pull/9 After it is approved and the Docker image is uploaded, I will re-enable the itegration test for the captcha image generation method.*

**How the built-in captcha works:**
Upon submitting a subscription form, if all validations pass, the form data is stored in the session and the user is redirected to the captcha page. When he submits the captcha, the form data is taken from the session and merged with the captcha input value, then the captcha check is performed and the user is subscribed.

**Why does captcha check come after the form validation?**
Our subscription logic has many checks and validations which, if they fail, show errors back in the form. Since we have the captcha on a separate page, we have to perform all validation beforehand. This way in case of errors we don't have to ping-pong between the captcha and the subscription form pages where the data is already gone or it will increase the complexity of the whole interaction.

**Notes to QA:**
* Please check that all types of subscription forms (widget, shortcode, PHP, iframe) work correctly in all three modes: built-in captcha, recaptcha, no captcha. All possible errors should be handled correctly in the UI;
* Make sure that MP API and other subscription methods 3rd parties may depend on still work with captcha enabled.

### DIFF
--- a/assets/js/src/public.js
+++ b/assets/js/src/public.js
@@ -23,12 +23,18 @@ jQuery(function ($) { // eslint-disable-line func-names
   }
 
   function updateCaptcha(e) {
-    var captcha = $('img.mailpoet_captcha');
-    var captchaSrc = captcha.attr('src');
-    var hashPos = captchaSrc.indexOf('#');
-    var newSrc = hashPos > 0 ? captchaSrc.substring(0, hashPos) : captchaSrc;
+    var captcha;
+    var captchaSrc;
+    var hashPos;
+    var newSrc;
+    captcha = $('img.mailpoet_captcha');
+    if (!captcha.length) return false;
+    captchaSrc = captcha.attr('src');
+    hashPos = captchaSrc.indexOf('#');
+    newSrc = hashPos > 0 ? captchaSrc.substring(0, hashPos) : captchaSrc;
     captcha.attr('src', newSrc + '#' + new Date().getTime());
     if (e) e.preventDefault();
+    return true;
   }
 
   $(function () { // eslint-disable-line func-names
--- a/assets/js/src/public.js
+++ b/assets/js/src/public.js
@@ -60,11 +60,22 @@ jQuery(function ($) { // eslint-disable-line func-names
           data: formData.data,
         })
           .fail(function handleFailedPost(response) {
-            form.find('.mailpoet_validate_error').html(
-              response.errors.map(function buildErrorMessage(error) {
-                return error.message;
-              }).join('<br />')
-            ).show();
+            if (
+              response.meta !== undefined
+                && response.meta.redirect_url !== undefined
+            ) {
+              // go to page
+              window.top.location.href = response.meta.redirect_url;
+            } else {
+              if (response.meta && response.meta.refresh_captcha) {
+                $('.mailpoet_captcha_update').click();
+              }
+              form.find('.mailpoet_validate_error').html(
+                response.errors.map(function buildErrorMessage(error) {
+                  return error.message;
+                }).join('<br />')
+              ).show();
+            }
           })
           .done(function handleRecaptcha(response) {
             if (window.grecaptcha && formData.recaptcha) {
@@ -83,6 +94,10 @@ jQuery(function ($) { // eslint-disable-line func-names
             } else {
               // display success message
               form.find('.mailpoet_validate_success').show();
+              // hide elements marked with a class
+              form.find('.mailpoet_form_hide_on_success').each(function hideOnSuccess() {
+                $(this).hide();
+              });
             }
 
             // reset form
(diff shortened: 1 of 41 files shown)


assets/js/src/public.js
```diff
@@ -60,11 +60,22 @@ jQuery(function ($) { // eslint-disable-line func-names
...
-            ).show();
+            if (
+              response.meta !== undefined
+                && response.meta.redirect_url !== undefined
+            ) {
+              // go to page
+              window.top.location.href = response.meta.redirect_url;
+            } else {
+              if (response.meta && response.meta.refresh_captcha) {
+                $('.mailpoet_captcha_update').click();
```
costasovo: What about creating `updateCaptcha` function and call it directly here and also within `$('.mailpoet_captcha_update').click()`. In my opinion it would be more robust and readable than triggering it using dom event.

lib/API/JSON/v1/Subscribers.php
```diff
@@ -201,6 +192,64 @@ function subscribe($data = []) {
       ]);
     }
 
+    $is_builtin_captcha_required = false;
```
costasovo: It seems we have over 100 lines of code for captcha validation. Maybe we could move it into a private method which would return a bad request in case of invalid captcha or maybe even refactor this to some separate class.
I think it would make `subscribe` method in API class much more readable.

wxa:
@costasovo I have made changes for all your comments.

lib/API/JSON/v1/Subscribers.php
```diff
@@ -204,18 +258,18 @@ function subscribe($data = []) {
...
-    if (!empty($captcha['type']) && $captcha['type'] === Captcha::TYPE_RECAPTCHA && empty($data['recaptcha'])) {
+    if (!empty($captcha_settings['type']) && $captcha_settings['type'] === Captcha::TYPE_RECAPTCHA && empty($data['recaptcha'])) {
       return $this->badRequest([
         APIError::BAD_REQUEST => WPFunctions::get()->__('Please check the CAPTCHA.', 'mailpoet'),
       ]);
     }
 
-    if (!empty($captcha['type'])) {
-      if ($captcha['type'] === Captcha::TYPE_RECAPTCHA) {
+    if (!empty($captcha_settings['type'])) {
```
costasovo: Using fast termination here would make this method a bit more readable by removing one level of nesting in conditions.
```
if (empty($captcha_settings['type'])) {
  return true;
}
```
You can also put the condition at the beginning of the `validateCaptcha` method so that it doesn't have to be repeated in each condition.

lib/Subscription/CaptchaRenderer.php
```diff
@@ -0,0 +1,85 @@
...
+use MailPoet\Models\Form as FormModel;
+use MailPoet\Util\Url as UrlHelper;
+use MailPoet\Form\Renderer as FormRenderer;
+use MailPoet\WP\Functions as WPFunctions;
+
+class CaptchaRenderer {
+  /** @var UrlHelper */
+  private $url_helper;
+
+  function __construct() {
```
costasovo: Since this is a new class and it is used only on one place I think it would be nicer to inject it's dependencies.
```
function __construct(UrlHelper $url_helper, WPFunctions $wp) {
  ...
}
It is constructed in Pages.php which is not refactored to DI yet so this class could be constructed in Pages.php. 
Also within methods we could call `$this->wp->`... instead `WPFunctions::get()->`

File: lib/Subscription/Captcha.php [hunk 5]
```diff
@@ -0,0 +1,105 @@
...
+        return true;
+      }
+    }
+
+    // Check limits per IP address
+    $subscription_captcha_window = $this->wp->applyFilters('mailpoet_subscription_captcha_window', MONTH_IN_SECONDS);
+
+    $subscriber_ip = Helpers::getIP();
+
+    if (!empty($subscriber_ip)) {
```
costasovo: We can get rid of one level nesting conditions by immediately returning `false` `if (empty($subscriber_ip))`

lib/Router/Endpoints/Subscription.php
```diff
@@ -21,6 +25,17 @@ class Subscription {
     'global' => AccessControl::NO_ACCESS_RESTRICTION,
   ];
 
+  function captcha($data) {
+    $subscription = $this->initSubscriptionPage(UserSubscription\Pages::ACTION_CAPTCHA, $data);
```
costasovo: I find this unused variable a bit confusing. Could you please change it just to `$this->initSubscriptionPage(UserSubscription\Pages::ACTION_CAPTCHA, $data);`

wxa:
Fixed all remarks.

wxa:
@costasovo I've made changes after @qfrery's review/

PR #2144, opened 2019-07-08
Convert import subscribers into react [MAILPOET-1809]

pavel-mailpoet:


assets/js/src/subscribers/importExport/import/step_data_manipulation/match_table.jsx
```diff
@@ -1,118 +1,127 @@
...
+  subscribers: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.string)).isRequired,
+};
+
 function MatchTable({
   subscribersCount,
   subscribers,
   header,
 }) {
-  let selectedColumns = [];
+  const selectedColumns = [];
```
amine-mp: Maybe I am missing something, but the `selectedColumns` variable seems to not be used anywhere! We assign values to it but never use it.

pavel-mailpoet: I think so. I was about to do this using react state and props and pass this data the modern way. But in the end, I had to re-use some old jquery code I didn't need to use this variable but forgot to delete it.

PR #2149, opened 2019-07-08
Mark spammy subscribers as unconfirmed [MAILPOET-2150]

pavel-mailpoet:


tests/integration/Segments/WPTest.php
```diff
@@ -282,6 +282,23 @@ function testItMarksSpammySubscribersAsUnconfirmed() {
     expect($db_subscriber->status)->equals(Subscriber::STATUS_UNCONFIRMED);
   }
 
+  function testItMarksSpammySubscribersWithUserStatus2AsUnconfirmed() {
```
wxa: The test will fail if there is no `user_status` column (e.g. in future WP versions). I suggest to skip it in that case.

PR #2153, opened 2019-07-16
Refactor Menu [MAILPOET-2200]

costasovo:
I've added a new namespace `MailPoet\AdminPages` and refactored all methods used for fetching data for templates to extra classes and injected container so that a class for a particular page is instantiated when it is needed for rendering.

## QA Notes
I would be good to go thru all MailPoet admin pages and check that all renders correctly.
Here are all pages which are handled by Menu class:

mailpoet/lib/AdminPages/Pages/ExperimentalFeatures.php
mailpoet/lib/AdminPages/Pages/FormEditor.php
mailpoet/lib/AdminPages/Pages/Forms.php
mailpoet/lib/AdminPages/Pages/Help.php
mailpoet/lib/AdminPages/Pages/MP2Migration.php
mailpoet/lib/AdminPages/Pages/NewsletterEditor.php
mailpoet/lib/AdminPages/Pages/Newsletters.php
mailpoet/lib/AdminPages/Pages/Premium.php
mailpoet/lib/AdminPages/Pages/RevenueTrackingPermission.php
mailpoet/lib/AdminPages/Pages/Segments.php
mailpoet/lib/AdminPages/Pages/Settings.php
mailpoet/lib/AdminPages/Pages/Subscribers.php
mailpoet/lib/AdminPages/Pages/SubscribersAPIKeyInvalid.php
mailpoet/lib/AdminPages/Pages/SubscribersExport.php
mailpoet/lib/AdminPages/Pages/SubscribersImport.php
mailpoet/lib/AdminPages/Pages/SubscribersLimitExceeded.php
mailpoet/lib/AdminPages/Pages/Update.php
mailpoet/lib/AdminPages/Pages/WelcomeWizard.php
mailpoet/lib/AdminPages/Pages/WooCommerceListImport.php

### DIFF
--- a/lib/Config/Menu.php
+++ b/lib/Config/Menu.php
@@ -3,6 +3,7 @@
 namespace MailPoet\Config;
 
 use Carbon\Carbon;
+use MailPoet\AdminPages\PageRenderer;
 use MailPoet\Cron\CronHelper;
 use MailPoet\Cron\CronTrigger;
 use MailPoet\Features\FeaturesController;
@@ -30,7 +31,6 @@ use MailPoet\Util\License\Features\Subscribers as SubscribersFeature;
 use MailPoet\Util\License\License;
 use MailPoet\WooCommerce\Helper as WooCommerceHelper;
 use MailPoet\WP\DateTime;
-use MailPoet\WP\Notice as WPNotice;
 use MailPoet\WP\Readme;
 use MailPoet\WP\Functions as WPFunctions;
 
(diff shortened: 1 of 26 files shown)


lib/Config/Menu.php
```diff
@@ -439,15 +446,7 @@ function migration() {
...
-    if ((bool)(defined('DOING_AJAX') && DOING_AJAX)) return;
-    $data = [
-      'is_mp2_migration_complete' => (bool)$this->settings->get(MP2Migrator::MIGRATION_COMPLETE_SETTING_KEY),
-      'is_woocommerce_active' => $this->woocommerce_helper->isWooCommerceActive(),
-      'finish_wizard_url' => $this->wp->adminUrl('admin.php?page=' . self::MAIN_PAGE_SLUG),
-      'sender' => $this->settings->get('sender'),
-      'admin_email' => get_option('admin_email'),
-    ];
-    $this->page_renderer->displayPage('welcome_wizard.html', $data);
+    $this->container->get(WelcomeWizard::class)->render();
```
JanJakes: You're fetching this service from DI container but the service is not public. That generates warnings. Probably not a big issue but still... Thinking about it, I think we should just mark all services as public (automatically maybe) because we'll just always need this (also like fetching services in integration tests, etc.).

JanJakes: Now I see another commit fixed that.

costasovo:
Thank you @michelleshull! This was a very good catch.

costasovo:
@JanJakes I've added one commit with a fix for the reported issue.

PR #2158, opened 2019-07-17
Refactor Shortcodes, Pages and Blacklist to DI [MAILPOET-2208] 

wxa:
The `Subscription\Pages` constructor via the `getSubscriber` method called WP functions which weren't loaded yet during DI initialization, so I have moved this class initialization to an `init` method.

### DIFF
--- a/lib/Config/Shortcodes.php
+++ b/lib/Config/Shortcodes.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Config;
 
+use MailPoet\Form\Widget;
 use MailPoet\Models\Newsletter;
 use MailPoet\Models\Subscriber;
 use MailPoet\Models\SubscriberSegment;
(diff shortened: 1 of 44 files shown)


lib/Config/Shortcodes.php
```diff
@@ -39,18 +45,18 @@ function init() {
...
   }
 
   function formWidget($params = []) {
     // IMPORTANT: fixes conflict with MagicMember
-    WPFunctions::get()->removeShortcode('user_list');
+    $this->wp->removeShortcode('user_list');
 
     if (isset($params['id']) && (int)$params['id'] > 0) {
-      $form_widget = new \MailPoet\Form\Widget();
+      $form_widget = new Widget();
```
pavel-mailpoet: Why didn't you pass Widget through DI as well?

lib/DI/ContainerConfigurator.php
```diff
@@ -64,6 +64,8 @@ function configure(ContainerBuilder $container) {
     $container->autowire(\MailPoet\Config\Menu::class)->setPublic(true);
     $container->autowire(\MailPoet\Config\RendererFactory::class)->setPublic(true);
     $container->autowire(\MailPoet\Config\ServicesChecker::class);
+    $container->autowire(\MailPoet\Config\Shortcodes::class)
+      ->setShared(false);
```
pavel-mailpoet: This `setShared` is not something we are used to. Don't you think we should add a comment here for others to understand what it does and why is it important to keep it here?

lib/Config/Shortcodes.php
wxa: This class extends `\WP_Widget` and is consumed by WP via [register_widget](https://codex.wordpress.org/Function_Reference/register_widget) function. WP does not pass any parameters to a widget constructor and, moreso, our DI container throws a non-existent `WP_Widget` class error when dumping container cache during build with this class added. Therefore, even if refactoring it to DI is still possible, it would be a longer and more quirky refactoring session.

lib/DI/ContainerConfigurator.php
wxa: Sure, I'll add an explanation.

PR #2168, opened 2019-07-23
Block risky lists import [MAILPOET-2129]

pavel-mailpoet:


pavel-mailpoet:
@wxa thanks for the update. Kim just removed the request so we can move it towards QA.

PR #2169, opened 2019-07-24
Sender task scheduler [MAILPOET-2146]

pavel-mailpoet:


pavel-mailpoet:
@michelleshull I had to refactor Mailer class in this pull request. So when testing, please focus on sending.

PR #2171, opened 2019-07-24
Referral id detection  [MAILPOET-2203]

costasovo:
There is a small [related PR](https://github.com/mailpoet/mailpoet-premium/pull/275) in premium.

### DIFF
--- a/tests/unit/Referrals/ReferralDetectorTest.php
+++ b/tests/unit/Referrals/ReferralDetectorTest.php
@@ -1,0 +1,69 @@
+<?php
+namespace MailPoet\Referrals;
+
+use MailPoet\Settings\SettingsController;
+use MailPoet\WP\Functions as WPFunctions;
+use PHPUnit_Framework_MockObject_MockObject as MockObject;
+
+class ReferralDetectorTest extends \MailPoetUnitTest {
+
+  /** @var MockObject */
+  private $settings_mock;
+
+  /** @var MockObject */
+  private $wp_mock;
+
+  function _before() {
+    $this->settings_mock = $this->createMock(SettingsController::class);
+    $this->wp_mock = $this->createMock(WPFunctions::class);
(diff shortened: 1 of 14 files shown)


tests/unit/Referrals/ReferralDetectorTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+  function _before() {
+    $this->settings_mock = $this->createMock(SettingsController::class);
+    $this->wp_mock = $this->createMock(WPFunctions::class);
+    if (!defined(ReferralDetector::REFERRAL_CONSTANT_NAME)) {
+      define(ReferralDetector::REFERRAL_CONSTANT_NAME, 'constant_referral_id');
+    }
+  }
+
+  function testItPrefersSettingsValueOverAll() {
```
JanJakes: The method name sounds weird to me. Probably `testItPrefersSettingsValue` is enough but "others" or "any other" would be correct, same as "prefer to" instead of "prefer over".

tests/unit/Referrals/ReferralDetectorTest.php
```diff
@@ -0,0 +1,69 @@
...
+      ->method('get')
+      ->willReturn('settings_referral_id');
+    $this->wp_mock
+      ->expects($this->never())
+      ->method('getOption');
+    $referral_detector = new ReferralDetector($this->wp_mock, $this->settings_mock);
+    expect($referral_detector->detect())->equals('settings_referral_id');
+  }
+
+  function testItPrefersOptionValueOverConstantAndStoreValueToSettings() {
```
JanJakes: Again the name is wrong. Should be "and storeS" and "prefer to" is more common than "prefer over".

costasovo:
@JanJakes Rebased and fixed.

costasovo:
Rebased. I think we can merge it after release.

PR #2191, opened 2019-07-31
Validate from address on form submit [MAILPOET-2135]

amine-mp:
[MAILPOET-2135]

[MAILPOET-2135]: https://mailpoet.atlassian.net/browse/MAILPOET-2135

### DIFF
--- a/assets/js/src/newsletters/send.jsx
+++ b/assets/js/src/newsletters/send.jsx
@@ -35,17 +35,14 @@ class NewsletterSend extends React.Component {
       item: {},
       loading: true,
       thumbnailPromise: null,
-      authorizedEmailAddresses: [],
     };
   }
 
   componentDidMount() {
-    jQuery.when(
-      this.loadItem(this.props.match.params.id),
-      this.loadAuthorizedEmailAddresses()
-    ).always(() => {
-      this.setState({ loading: false });
-    });
+    this.loadItem(this.props.match.params.id)
+      .always(() => {
+        this.setState({ loading: false });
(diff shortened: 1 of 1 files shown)


assets/js/src/newsletters/send.jsx
```diff
@@ -81,13 +78,12 @@ class NewsletterSend extends React.Component {
 
   isValid = () => jQuery('#mailpoet_newsletter').parsley().isValid();
 
-  isValidFromAddress = () => {
+  isValidFromAddress = async () => {
```
costasovo: Could you please keep a condition for MailPoet sending method here?
```
if (window.mailpoet_mta_method !== 'MailPoet') {	    
      return true;	
}
```
It is handled in `loadAuthorizedEmailAddresses`  but current code in function `isValidFromAddress ` was a bit confusing for me because a condition `if (addresses === false) return true;` looks like we mark from address as valid when user has no authorized addresses.

assets/js/src/newsletters/send.jsx
```diff
@@ -81,13 +78,14 @@ class NewsletterSend extends React.Component {
...
+    const addresses = await this.loadAuthorizedEmailAddresses();
     const fromAddress = this.state.item.sender_address;
     if (window.mailpoet_mta_method !== 'MailPoet') {
```
costasovo: Can you please put this condition as a very first thing in the function so that we avoid doing a useless API request in `loadAuthorizedEmailAddresses` ?

PR #2192, opened 2019-07-31
Add index.php only if no Index.php [MAILPOET-2234]

amine-mp:
[MAILPOET-2234]

[MAILPOET-2234]: https://mailpoet.atlassian.net/browse/MAILPOET-2234

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -141,7 +141,12 @@ cp uninstall.php $plugin_name
 
 # Add index files if they don't exist to all folders
 echo '[BUILD] Adding index files to all project folders'
-find $plugin_name -type d -exec touch {}/index.php \;
+find $plugin_name -type d -print0 | while read -d $'\0' dir
+do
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -141,7 +141,12 @@ cp uninstall.php $plugin_name
...
 echo '[BUILD] Adding index files to all project folders'
-find $plugin_name -type d -exec touch {}/index.php \;
+find $plugin_name -type d -print0 | while read -d $'\0' dir
+do
+  if [ ! -f "$dir/Index.php" ]; then
+    touch "$dir/index.php"
+  fi
+done
 # Remove second index.php from Doctrine directory
 rm -f $plugin_name/vendor-prefixed/doctrine/orm/lib/Doctrine/ORM/Mapping/index.php
```
wxa: Do we still need this line after the fix?

amine-mp: No we don't, I will remove it.

PR #2198, opened 2019-08-01
Fix setting wait_timeout for database [MAILPOET-2243]

costasovo:


lib/Doctrine/ConnectionFactory.php
```diff
@@ -43,7 +43,7 @@ private function getDriverOptions($timezone_offset, $charset, $collation) {
     $driver_options = [
       "@@session.time_zone = '$timezone_offset'",
       '@@session.sql_mode = REPLACE(@@sql_mode, "ONLY_FULL_GROUP_BY", "")',
-      "@@session.wait_timeout = GREATEST(@@wait_timeout, $this->min_wait_timeout)",
+      "@@session.wait_timeout = GREATEST(CONVERT(@@wait_timeout, SIGNED), $this->min_wait_timeout)",
```
JanJakes: Please add a comment why `CONVERT` and `SIGNED` is needed, ideally with links to MySQL/MariaDB issues 🙏

PR #2200, opened 2019-08-01
Remove Doctrine unneeded annotations directory [MAILPOET-2236]

amine-mp:
[MAILPOET-2236]

[MAILPOET-2236]: https://mailpoet.atlassian.net/browse/MAILPOET-2236


JanJakes:
@codemonkey-jack @amine-mp Hmm, this would have to be run before `./composer.phar dump-autoload` (or the autoload after).

amine-mp:
I tried removing `vendor-prefixed/doctrine/annotations` in local then run `composer dump-autoload` and it worked fine.

PR #2205, opened 2019-08-02
Start session only on MailPoet pages [MAILPOET-2246]

amine-mp:
[MAILPOET-2246]

[MAILPOET-2246]: https://mailpoet.atlassian.net/browse/MAILPOET-2246


costasovo:
Closing this PR since it will be addressed in another [PR](https://github.com/mailpoet/mailpoet/pull/2206).

PR #2206, opened 2019-08-05
Session workaround for captcha [MAILPOET-2248][MAILPOET-2246]

costasovo:
[MAILPOET-2248](https://mailpoet.atlassian.net/browse/MAILPOET-2248), [MAILPOET-2246](https://mailpoet.atlassian.net/browse/MAILPOET-2246)

## Idea

1. Replace PHP Session with minimalist [MailPoet session](https://github.com/mailpoet/mailpoet/blob/captcha-session/lib/Config/Session.php) which is only responsible for handling session id. It has no storage for session data but it uses a cookie to store and read session id, so that we could relate requests made by one user.
2. Create [a service which uses the session id and stores captcha related data (captcha hash + form data)](https://github.com/mailpoet/mailpoet/blob/captcha-session/lib/Subscription/CaptchaSession.php) using this id so that they can be used in different requests. This service has own expiration time (lower than session's expiration) and uses [WP Transient API](https://codex.wordpress.org/Transients_API) to store captcha related data.
3. Use this session instead of PHP session in all places...

## Reasoning
This is a minimalist solution needed for captcha. I want to avoid storing anything on server until it is really needed. PHP Session stores some data in filesystem when initialized. I separated session id from storage so that we could store anything once it is really needed. Later we can add another session services using different storages or create a common session storage service.


tests/integration/API/JSON/v1/SubscribersTest.php
```diff
@@ -517,7 +545,7 @@ function testItCanSubscribeWithBuiltInCaptchaWhenEnabled() {
     $subscriber->count_confirmations = 1;
     $subscriber->save();
     $captcha_value = 'ihg5w';
-    $_SESSION[Captcha::SESSION_KEY] = $captcha_value;
+    $this->captcha_session->setCaptchaHash('ihg5w');
```
wxa: The string is duplicated, let's use the variable above.

lib/Subscription/CaptchaSession.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace MailPoet\Subscription;
+
+use MailPoet\Config\Session;
+use MailPoet\WP\Functions as WPFunctions;
+
+class CaptchaSession {
+  const EXPIRATION = 300; // 5 minutes
```
wxa: 5 minute lifetime is too little, 30 minutes would be better.

costasovo:
Thanks for fixing those remarks @wxa!

PR #2207, opened 2019-08-05
Release 3.34.5

amine-mp:


codemonkey-jack:
The release for the week ending August 9th didn't happen, so we don't need to merge this. After discussing with Amine, we agreed to close this PR and we'll have new ones next week.

PR #2213, opened 2019-08-12
Add referral id to URLs [MAILPOET-2182]

costasovo:
## QA Notes
I updated @babel/core-js from version 2 to version 3 so that we can use modern js features. Please do some basic walkthrough in supported browsers. 

### Urls
Here is a list of places where I've added the parameter to a link:
#### Premium page 
 - http://mailpoet.localhost/wp-admin/admin.php?page=mailpoet-premium
#### Free plan annoucement 
- delete free_plan_announcement_seen setting
- http://mailpoet.localhost/wp-admin/admin.php?page=mailpoet-newsletters

#### Sending error notices with PHPMailer
- set sending method to PHPMailer but must be misconfigured on host (or just break mailer https://github.com/mailpoet/mailpoet/blob/master/lib/Mailer/Methods/PHPMail.php#L36 to always return error)
- on email editor send preview and link in error notie should have ref
- send an email (with misconfigured PHPmailer) and wait for error notice on newsletter page

#### Newsletter stats premium banner
- deactivate premium and open newsletter stats
- note there is different button when there is more than 1000 subscribers

#### Settings -> send with
- http://mailpoet.localhost/wp-admin/admin.php?page=mailpoet-settings#mta
- if MSS active and no/invalid api key is set

#### Invalid key page when MSS active and API key is invalid
 - http://mailpoet.localhost/wp-admin/admin.php?page=mailpoet-newsletters


views/layout.html
```diff
@@ -60,6 +60,7 @@
   // RFC 5322 standard; http://emailregex.com/ combined with https://google.github.io/closure-library/api/goog.format.EmailAddress.html#isValid
   var mailpoet_email_regex = /(?=^[+a-zA-Z0-9_.!#$%&'*\/=?^`{|}~-]+@([a-zA-Z0-9-]+\.)+[a-zA-Z0-9]{2,63}$)(?=^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,})))/;
   var mailpoet_feature_flags = <%= json_encode(feature_flags) %>;
+  var mailpoet_referral_id = <%= json_encode(referral_id) %>;
```
JanJakes: Does this need `json_encode`? I don't know if we use it for scalar values.

assets/js/src/announcements/free_plan_announcement.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import MailPoet from 'mailpoet';
+import addReferralId from 'referral_url_decorator.jsx';
```
JanJakes: As discussed, this could be on the `MailPoet` object. Also, the MailPoet URLs could be created by a simple factory, maybe.

assets/js/src/mailpoet_url_factory.js
```diff
@@ -0,0 +1,27 @@
...
+    return url;
+  }
+  const refUrl = new URL(url);
+  refUrl.searchParams.set('ref', referralId);
+  return refUrl.toString();
+};
+
+const MailPoetUrlFactory = referralId => ({
+  getFreePlanUrl: (utmCampaign = null, utmMedium = null) => {
+    const url = new URL(addReferralId('https://www.mailpoet.com/free-plan/?utm_source=plugin', referralId));
```
JanJakes: This doesn't look very nice - you create a URL string, parse it to URL object (in `addReferralId`), modify some parameters, output it to string and then parse it again.

This should work differently - i.e. adding a ("private") method `getUrl` that would return URL object with base URL and extending on that (by adding path and params). Above that, such method could already register the referral ID parameter.

assets/js/src/newsletter_editor/components/sidebar.js
```diff
@@ -354,7 +353,7 @@ Module.SidebarPreviewView = Marionette.View.extend({
                   <p>${MailPoet.I18n.t('newsletterPreviewErrorUseSendingService')}</p>
                   <p>
                     <a
-                      href=${addReferralId('https://www.mailpoet.com/free-plan/?utm_source=plugin&utm_campaign=sending-error')}
+                      href=${MailPoet.MailPoetUrlFactory.getFreePlanUrl('plugin', 'sending-error')}
```
JanJakes: You have the params wrong here (switched).

assets/js/src/mailpoet_url_factory.js
```diff
@@ -0,0 +1,27 @@
...
+    if (utmCampaign) {
+      url.searchParams.set('utm_campaign', utmCampaign);
+    }
+    if (utmMedium) {
```
JanJakes: I also don't like too much having two optional positional parameters (you already used them wrong, see below). What about allowing to pass `utmParams` object and have it checked by PropTypes? Or maybe even a new method (`getFreePlanUrlWithUtmParams`)? Or some kind of method to add them to any URL object?

PR #2216, opened 2019-08-13
Release 3.35.0

pavel-mailpoet:


JanJakes:
@codemonkey-jack This is probably OK, it is typographically correct and it was implemented that way in the changelog generator (it doesn't change `;`, `.`, `!` or `?` because all of these are correct in a list, it only adds a semicolon to lines where punctuation is missing).

As for the last line, when there's a `;` the script replaces it with `.`.

PR #2219, opened 2019-08-13
Hide WC Customer list on non WC websites [MAILPOET-2163]

NeosinneR:


lib/Models/Segment.php
```diff
@@ -155,6 +155,24 @@ static function getWooCommerceSegment() {
     return $wc_segment;
   }
 
+  static function shouldShowWooCommerceSegment() {
+    $is_woocommerce_active = class_exists('WooCommerce');
+    $woocommerce_users_count = Subscriber::where('is_woocommerce_user', 1)->count();
```
JanJakes: For big segments it would be more performant to use sth like "WHERE is_woocommerce_user = 1 LIMIT 1", it also seems `is_woocommerce_user` doesn't have index but this should help.

Btw. I also checked the option of looking to the segment directly and sth like `SubscriberSegment::where('segment_id', $wc_segment_id)->limit(1)` should work if you know the segment ID.

lib/Models/Segment.php
```diff
@@ -159,9 +159,11 @@ static function getWooCommerceSegment() {
   static function shouldShowWooCommerceSegment() {
     $woocommerce_helper = new WCHelper();
     $is_woocommerce_active = $woocommerce_helper->isWooCommerceActive();
-    $woocommerce_users_count = Subscriber::where('is_woocommerce_user', 1)->count();
+    $woocommerce_user_exists = Subscriber::where('is_woocommerce_user', 1)
+      ->limit(1)
+      ->findOne();
```
JanJakes: I just realized that if you have no WC users this will always result in table full scan due to the fact there's no index on `is_woocommerce_user`(and the fact it will never find one). I guess that's risky for large customers. Should we try sth like below? That should use indexes.

```
Segment::tableAlias('segment')
  ->where('segment.type', Segment::TYPE_WC_USERS)
  ->join(
    MP_SUBSCRIBER_SEGMENT_TABLE,
    'segment_subscribers.segment_id = segment.id',
    'segment_subscribers'
  )
  ->limit(1)
  ->findOne();
```

NeosinneR: 👍 used that, it's working correctly

PR #2222, opened 2019-08-14
Speed up newsletters list loading [MAILPOET-2213]

costasovo:


lib/Config/Migrator.php
```diff
@@ -428,7 +428,7 @@ function statisticsOpens() {
       'queue_id int(11) unsigned NOT NULL,',
       'created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,',
       'PRIMARY KEY  (id),',
-      'KEY newsletter_id (newsletter_id),',
+      'KEY newsletter_id_subscriber_id (newsletter_id, subscriber_id),',
```
JanJakes: Shouldn't you create the same key in `statistics_unsubscribes` as well?

costasovo: Good catch! 👍

PR #2229, opened 2019-08-15
Update 'View in browser' header link [MAILPOET-2196] 

NeosinneR:


assets/js/src/newsletter_editor/blocks/header.js
```diff
@@ -13,7 +13,7 @@ Module.HeaderBlockModel = base.BlockModel.extend({
   defaults: function defaults() {
     return this._getDefaults({
       type: 'header',
-      text: 'Display problems? <a href="[link:newsletter_view_in_browser_url]">View it in your browser</a>',
+      text: '<a href="[link:newsletter_view_in_browser_url]">View it in your browser</a>',
```
JanJakes: Some templates have `this`, some `it`. Should we unify also that? The specs use `this`.

PR #2233, opened 2019-08-19
Add check for logged PHP errors after acceptance tests [MAILPOET-2267]

pavel-mailpoet:


amine-mp:
I am sure how to go about testing this. Should we introduce an exception and check if CircleCI would catch it? How to do that without causing unit/integration tests to fail?

PR #2236, opened 2019-08-19
Update the default list name [MAILPOET-2198]

pavel-mailpoet:


pavel-mailpoet:
> I'm still seeing My First List on new installs and after using MP's reinstall setting, not seeing the new list name.

@michelleshull did you see the Jira specification? It says `Put this behind feature flag to wait for translation` Which I did. This new name is behind a feature flag. So you cannot see the new name. See my comment in the Jira. This ticket only gets the new name to production for translators to pick it up.

michelleshull:
I did see it, but the comments indicated we were not doing the feature flag. Should have checked the code. Will test again.

pavel-mailpoet:
There is not much to be tested. I added the new string to the code so that it can be picked by translators. But you won't see any change in the plugin.

PR #2243, opened 2019-08-21
Fix image resizing when aligned right [MAILPOET-2184]

NeosinneR:


NeosinneR:
@michelleshull are you sure? While adding a post or product, you can't resize the image, only after it's added. But then, it's just a regular image widget.

michelleshull:
Yep, just double-checked. I took a video of the Image, Post, and Product widgets with right-aligned images so you can see it in action: https://monosnap.com/file/73wbl9U41M5m6L5IR1KxAajHBPYkBC

NeosinneR:
Aaah, I see. That setting in posts/products is only for alignment within the post/product, not alignment for an actual image widget that is inserted. After adding post/product, if you click on the image, it's centered, that's why the handler is at right bottom (and resizing should normally work).

PR #2244, opened 2019-08-21
Refactor Feature Flags to Doctrine [MAILPOET-2219]

pavel-mailpoet:


lib/Features/FeatureFlagsRepository.php
```diff
@@ -15,4 +15,33 @@ class FeatureFlagsRepository extends Repository {
...
+    }
+    if (isset($data['value'])) {
+      $feature_flag->setValue($data['value']);
+    }
+
+    try {
+      $this->flush();
+    } catch (\Exception $e) {
+      throw new \RuntimeException("Error when saving feature " . $data['name']);
+    }
```
JanJakes: Not sure if we need to rethrow this exception. Maybe the try/catch is not necessary. Failing to save to DB would definitely be a severe error that we don't need to handle I guess. Or does this try to prevent exposing some data, etc.? My point is - we'll probably use `flush()` in many other places and we won't wrap them there I guess.

lib/Features/FeatureFlagsRepository.php
```diff
@@ -15,4 +15,33 @@ class FeatureFlagsRepository extends Repository {
...
+      throw new \InvalidArgumentException('Missing name');
+    }
+    $feature_flag = $this->findOneBy([
+      'name' => $data['name'],
+    ]);
+    if (!$feature_flag) {
+      $feature_flag = new FeatureFlagEntity($data['name']);
+      $this->persist($feature_flag);
+    }
+    if (isset($data['value'])) {
```
JanJakes: If someone passes `['name' => 'xxx', 'value' => null]`, this will not save the `null`. I'd suggest using `array_key_exists`.

lib/Features/FeatureFlagsRepository.php
pavel-mailpoet: I tried to leave the functionality as it was before. The previous code was throwing this exception if saving failed, so I left it there.

PR #2245, opened 2019-08-21
Prevent showing WC Customers UI to new WC website [MAILPOET-2206]

pavel-mailpoet:


lib/Config/Changelog.php
```diff
@@ -107,6 +107,7 @@ private function checkWooCommerceListImportPage() {
       && !$this->settings->get('woocommerce_import_screen_displayed')
       && $this->wooCommerceHelper->isWooCommerceActive()
       && $this->wooCommerceHelper->getOrdersCount() >= 1
+      && $this->wooCommerceHelper->getOrdersCountCreatedBefore($this->settings->get('installed_at')) === 0
```
costasovo: I think the condition should be `$this->wooCommerceHelper->getOrdersCountCreatedBefore($this->settings->get('installed_at')) > 0`  because we don't want to show it to users who have 0 orders older than MP installation date.

I think we can also improve this check a bit by using `max($this->settings->get('installed_at'), $date_wc_list_feature_was_released)` so that users who have WooCommerce + MailPoet installed for long time would get the screen. 

Also we can remove `$this->wooCommerceHelper->getOrdersCount() >= 1`

PR #2252, opened 2019-08-22
New loading animation [MAILPOET-2005]

NeosinneR:
I completely skipped using JS for animating the bird as we'd need to include GreenSock's JS lib. The animation is done in CSS (I tried to do it the same, but in CSS it won't be that precise). Easy way to show loader is run `MailPoet.Modal.loading(true)` in console.


NeosinneR:
Discussed on Slack in #product, we are not doing this anymore. Sticking with old animation.

PR #2254, opened 2019-08-22
 Open KB links in Beacon popover [MAILPOET-2179]

NeosinneR:


amine-mp:
A link was missing, check Jira comments.

NeosinneR:
@amine-mp rebased because of conflict, added missing beacon ID as a separate commit

PR #2261, opened 2019-08-26
i18n: avoid using HTML tags in translation strings

ramiy:
The HTML tags `<h2>`, `<h3>` and `<p>` should be removed from the translation strings.

**Why?**

1. We did the same in WordPress core.

2. If the HTML tag is inside the translation string, the translator can accidentally change it from `<h2>` to let's say `<h1>`, or even worse, to forget to properly close the tag and break the code.

3. In RTL languages it's hard to translate strings with HTML code inside. The string is in RTL but the HTML tags are LTR. Very hard to translate this way.

**The Solution**

The best practice is to avoid using HTML tags in translation strings if possible.

### DIFF
--- a/lib/Config/PopulatorData/Templates/NewsletterBlank121Column.php
+++ b/lib/Config/PopulatorData/Templates/NewsletterBlank121Column.php
@@ -163,7 +163,7 @@ class NewsletterBlank121Column {
                 "blocks" => [
                   [
                     "type" => "text",
-                    "text" => WPFunctions::get()->'<h2>'.__('This template has...', 'mailpoet').'</h2>',
+                    "text" => '<h2>' . WPFunctions::get()->__('This template has...', 'mailpoet') . '</h2>',
                   ],
                   [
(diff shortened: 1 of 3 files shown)


lib/Config/PopulatorData/Templates/NewsletterBlank121Column.php
```diff
@@ -163,7 +163,7 @@ private function getBody() {
                 "blocks" => [
                   [
                     "type" => "text",
-                    "text" => WPFunctions::get()->__("<h2>This template has...</h2>", 'mailpoet'),
+                    "text" => WPFunctions::get()->'<h2>'.__('This template has...', 'mailpoet').'</h2>',
```
pavel-mailpoet: Hi ramiy, 
thank you for your work, but have you try to execute the code? It is not a valid code.  Also, it doesn't conform to our coding standards. The `.` operator should be separated by a single space on each side.

PR #2266, opened 2019-08-27
i18n: merge two similar translation strings

ramiy:
This PR merges two similar translation strings, the same way we do it in WordPress Core - using `sprintf()` function and `%s` placeholder.

**Old Strings:**

* `Cookie expiry: 14 days.`
* `Cookie expiry: 3,650 days.`

**New String:**

* `Cookie expiry: % days.`

**Note:**

* The `%s` placeholder makes sure the translator won't accidentally change the value while translating the string.

### DIFF
--- a/lib/Config/PrivacyPolicy.php
+++ b/lib/Config/PrivacyPolicy.php
@@ -48,14 +48,14 @@ class PrivacyPolicy {
         '<p>' .
         WPFunctions::get()->__('Cookie name: mailpoet_revenue_tracking', 'mailpoet') .
         '<br>' .
-        WPFunctions::get()->__('Cookie expiry: 14 days.', 'mailpoet') .
+        WPFunctions::get()->sprintf( __('Cookie expiry: % days.', 'mailpoet'), '14' ) .
         '<br>' .
         WPFunctions::get()->__('Cookie description: The purpose of this cookie is to track which newsletter sent from your website has acquired a click-through and a subsequent purchase in your WooCommerce store.', 'mailpoet') .
         '</p> ' .
         '<p>' .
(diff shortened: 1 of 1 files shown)


lib/Config/PrivacyPolicy.php
```diff
@@ -48,14 +48,14 @@ function getPrivacyPolicyContent() {
...
-        WPFunctions::get()->__('Cookie expiry: 14 days.', 'mailpoet') .
+        sprintf(WPFunctions::get()->__('Cookie expiry: %s days.', 'mailpoet'), '14' ) .
         '<br>' .
         WPFunctions::get()->__('Cookie description: The purpose of this cookie is to track which newsletter sent from your website has acquired a click-through and a subsequent purchase in your WooCommerce store.', 'mailpoet') .
         '</p> ' .
         '<p>' .
         WPFunctions::get()->__('Cookie name: mailpoet_abandoned_cart_tracking', 'mailpoet') .
         '<br>' .
-        WPFunctions::get()->__('Cookie expiry: 3,650 days.', 'mailpoet') .
+        sprintf(WPFunctions::get()->__('Cookie expiry: %s days.', 'mailpoet'), '3,650' ) .
```
amine-mp: Can you please use `number_format_i18n` to make the number follow the locale as well? As not all locales use `,` inside numbers. This line should become something like:
```
sprintf(WPFunctions::get()->__('Cookie expiry: %s days.', 'mailpoet'), WPFunctions::get()->numberFormatI18n(3650)) .
```
Note that we are using `WPFunctions::get()` to call WordPress functions to be able to easily mock those during tests.

PR #2267, opened 2019-08-27
i18n: merge two similar translation strings

ramiy:
This PR merges two similar translation strings, the same way we do it in WordPress Core - using `sprintf()` function and `%s` placeholder.

**Old Strings:**

* `Cookie name: mailpoet_revenue_tracking`
* `Cookie name: mailpoet_abandoned_cart_tracking`

**New String:**

* `Cookie name: %s`

**Note:**

* The `%s` placeholder makes sure the translator won't accidentally change the cookie name while translating the string.

### DIFF
--- a/lib/Config/PrivacyPolicy.php
+++ b/lib/Config/PrivacyPolicy.php
@@ -46,14 +46,14 @@ class PrivacyPolicy {
         WPFunctions::get()->__('MailPoet creates and stores two cookies if you are using WooCommerce and MailPoet together. Those cookies are:', 'mailpoet') .
         '</p>' .
         '<p>' .
-        WPFunctions::get()->__('Cookie name: mailpoet_revenue_tracking', 'mailpoet') .
+        WPFunctions::get()->sprintf( __('Cookie name: %s', 'mailpoet'), 'mailpoet_revenue_tracking' ) .
         '<br>' .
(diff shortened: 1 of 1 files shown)


lib/Config/PrivacyPolicy.php
```diff
@@ -46,14 +46,14 @@ function getPrivacyPolicyContent() {
         WPFunctions::get()->__('MailPoet creates and stores two cookies if you are using WooCommerce and MailPoet together. Those cookies are:', 'mailpoet') .
         '</p>' .
         '<p>' .
-        WPFunctions::get()->__('Cookie name: mailpoet_revenue_tracking', 'mailpoet') .
+        WPFunctions::get()->sprintf( __('Cookie name: %s', 'mailpoet'), 'mailpoet_revenue_tracking' ) .
```
pavel-mailpoet: This code doesn't work. There is no `sprintf` function on the `WPFunctions` object. This has to be the other way around: `sprintf(WPFunctions::get()->__('Cookie name: %s', 'mailpoet'), 'mailpoet_revenue_tracking')`.

Another problem is the coding standards. We don't use a space in the parentheses. This would break our build.

ramiy: Fixed.

PR #2268, opened 2019-08-27
Convert WordPress media blocks to links

fergbrain:
Replaces embed/self-hosted WordPress video and audio blocks with links to the source instead of stripping them out. If a thumbnail is present for a video file, it will use that as the link image.


wxa:
Thank you for your effort! This needs test coverage and non-Gutenberg support. If you'll be able to provide those, please reopen this PR.

fergbrain:
Is there a way I could hook into this as a filter instead?

PR #2273, opened 2019-08-28
Clean tests [MAILPOET-2321]

pavel-mailpoet:
@michelleshull Most of the changes are in tests. Changes in the code are two. I refactored the logic of displaying PHP notice a little. And scheduling of all welcome newsletters has been updated.


tests/integration/Cron/Workers/SendingQueue/SendingQueueTest.php
```diff
@@ -621,10 +620,13 @@ function testItDoesNotSendToInactiveSubscribers() {
...
-    $sending_task = Mock::double(SendingTask::create(), [
-      'updateProcessedSubscribers' => false,
-    ]);
-    $sending_task->id = 100;
+    $sending_task = $this->createMock(SendingTask::class);
+    $sending_task
+      ->method('updateProcessedSubscribers')
+      ->will($this->returnValue(false));
+    $sending_task
+      ->method('__get')
```
costasovo: I would be better to add `->with('id')` so that this mocks only `$sending_task->id` calls.

lib/Newsletter/Scheduler/Welcome.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace MailPoet\Newsletter\Scheduler;
+
+use MailPoet\Models\Newsletter;
+use MailPoet\Models\SendingQueue;
+use MailPoet\Tasks\Sending as SendingTask;
+
+class Welcome {
```
costasovo: I think we could follow the naming pattern we started using with entities.
`MailPoet\Entities\NewsletterEntity`
So this would be:
`MailPoet\Newsletter\Scheduler\WelcomeScheduler` or maybe `MailPoet\Newsletter\Scheduler\WelcomeEmailScheduler`

PR #2282, opened 2019-09-02
 Display captcha success/error messages without JS [MAILPOET-2323] 

wxa:


lib/Subscription/CaptchaRenderer.php
```diff
@@ -84,11 +101,20 @@ public function getCaptchaPageContent() {
...
-    $form_html .= '<div class="mailpoet_message">';
-    $form_html .= '<p class="mailpoet_validate_success" style="display:none;">' . $form_model['settings']['success_message'] . '</p>';
-    $form_html .= '<p class="mailpoet_validate_error" style="display:none;"></p>';
-    $form_html .= '</div>';
+    $form_html .= $this->renderFormMessages($form_model, $show_success_message, $show_error_message);
     $form_html .= '</form>';
     return $form_html;
   }
+
+  private function renderFormMessages(
```
JanJakes: This method and its usage are a bit strange:
- Two boolean arguments deciding what to render while we actually never render both of the messages.
- Hiding message by `display: none` instead not rendering it at all.

Since Alex is out I'm taking over to fix this and assign to someone else for review.

wxa:
@JanJakes 
 >  Two boolean arguments deciding what to render while we actually never render both of the messages.
>    Hiding message by display: none instead not rendering it at all.

We need these messages for the JS-enabled flow as well, so they should be both present, even if empty.

JanJakes:
@wxa Thanks, I tested and discovered that already. I'll leave it as-is although it's quite a confusing code.

PR #2283, opened 2019-09-02
Remove bulk option to resend confirmation emails [MAILPOET-2315]

NeosinneR:


pavel-mailpoet:
@michelleshull When testing this, could you please also check that Thrive Leads works? This pull request is removing a function that they used to use. I am not sure if they are still using it or not.

PR #2297, opened 2019-09-06
Subscribe new user to WP list only [MAILPOET-2210]

amine-mp:
[MAILPOET-2210]

[MAILPOET-2210]: https://mailpoet.atlassian.net/browse/MAILPOET-2210


amine-mp:
I forgot to add the acceptance test. It's added now.

PR #2300, opened 2019-09-09
Send confirmation emails using Default Sender address [MAILPOET-2328]

amine-mp:
[MAILPOET-2328]


lib/Subscribers/ConfirmationEmailMailer.php
```diff
@@ -92,18 +91,12 @@ function sendConfirmationEmail(Subscriber $subscriber) {
     ];
 
     // set from
-    $from = (
-      !empty($signup_confirmation['from'])
-      && !empty($signup_confirmation['from']['address'])
-    ) ? $signup_confirmation['from']
-      : false;
+    $from = $this->settings->get('sender', false);
+    if (empty($from['address'])) $from = false;
```
pavel-mailpoet: We can delete those 2 lines and also both `$reply_to` lines. Mailer will take care of everything.

PR #2309, opened 2019-09-10
Hide Beacon on wizard page [MAILPOET-2344]

amine-mp:
[MAILPOET-2344]


views/welcome_wizard.html
```diff
@@ -54,3 +55,12 @@
...
 <% endblock %>
+
+<% block after_javascript %>
+  <script type="text/javascript">
+    console.log(window.Beacon)
+    if (window.Beacon) {
+      window.Beacon('destroy');
+    }
+  </script>
+<% endblock %>
```
JanJakes: There is a `console.log()` here. But also - what is this whole block actually for?

amine-mp: oops! this block was only for debugging, I removed it.

PR #2321, opened 2019-09-13
Send meta to mss [MAILPOET-2333]

amine-mp:
[MAILPOET-2333]

I added the following email types that were not mentioned in the specs:

**The sending method test email** 
```
email_type = 'sending_test'
subscriber_status = 'unknown'
subscriber_source = 'administrator'
```

**Preview email**
```
email_type = 'preview'
subscriber_status = 'unknown'
subscriber_source = 'administrator'
```

I have also used `administrator` as `subscriber_source` for stats notifications, new subscriber notifications, and WordPress transactional emails.

[MAILPOET-2333]: https://mailpoet.atlassian.net/browse/MAILPOET-2333

### DIFF
--- a/lib/API/JSON/v1/Mailer.php
+++ b/lib/API/JSON/v1/Mailer.php
@@ -6,6 +6,7 @@ use MailPoet\API\JSON\Endpoint as APIEndpoint;
 use MailPoet\API\JSON\Error as APIError;
 use MailPoet\Config\AccessControl;
 use MailPoet\Mailer\MailerLog;
+use MailPoet\Mailer\MetaInfo;
 use MailPoet\Services\AuthorizedEmailsController;
 use MailPoet\Services\Bridge;
 use MailPoet\Settings\SettingsController;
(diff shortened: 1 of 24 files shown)


lib/API/JSON/v1/Mailer.php
```diff
@@ -40,7 +45,10 @@ function send($data = []) {
         (isset($data['sender'])) ? $data['sender'] : false,
         (isset($data['reply_to'])) ? $data['reply_to'] : false
       );
-      $result = $mailer->send($data['newsletter'], $data['subscriber']);
+      $extra_params = [
+        'meta' => $this->mailerMetaInfo->getSendingTestMetaInfo(),
```
JanJakes: This looks like a generic sending API class. I don't think we can hardcode here "sending test" meta info. Can we?

lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -67,6 +72,11 @@ function process() {
       }
       // clone the original object to be used for processing
       $_newsletter = (object)$newsletter->asArray();
+      $options = $newsletter->options()->findMany();
+      if (!empty($options)) {
+        $options = array_column($options, 'value', 'name');
```
JanJakes: Please use https://github.com/mailpoet/mailpoet/blob/master/lib/Util/ArrayColumn.php.

lib/API/JSON/v1/Mailer.php
amine-mp: I checked the code and found that we only call it to send the test email on from the settings page. I know that this may change in the future, but then what do you suggest as metadata for this email?

JanJakes: @amine-mp Yeah, it is kind of dangerous to suppose this is called for tests only. Maybe we should extend the API with some parameters? Or maybe we can ask what others think on Slack?

lib/Subscribers/LinkTokens.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
JanJakes: Why is this file here and how it is related to this commit?

PR #2327, opened 2019-09-17
Add backward compatibility for link tokens [MAILPOET-2364]

costasovo:
[MAILPOET-2364](https://mailpoet.atlassian.net/browse/MAILPOET-2364)

### DIFF
--- a/lib/Models/Subscriber.php
+++ b/lib/Models/Subscriber.php
@@ -38,7 +38,9 @@ class Subscriber extends Model {
   const STATUS_BOUNCED = 'bounced';
   const STATUS_INACTIVE = 'inactive';
 
+  const DEPRECATED_LINK_TOKEN_LENGTH = 6;
   const LINK_TOKEN_LENGTH = 32;
+
   /** @var string|bool */
   public $token;
 
@@ -121,7 +123,10 @@ class Subscriber extends Model {
     return self::where('wp_user_id', $wp_user->ID)->findOne();
   }
 
-  static function generateToken($email = null, $length = 32) {
+  /**
+   * @deprecated Only for backward compatibility for old tokens
+   */
+  static function generateToken($email = null, $length = self::DEPRECATED_LINK_TOKEN_LENGTH) {
     if ($email !== null) {
       $auth_key = '';
       if (defined('AUTH_KEY')) {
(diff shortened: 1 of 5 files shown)


lib/Models/Subscriber.php
```diff
@@ -38,6 +38,9 @@ class Subscriber extends Model {
   const STATUS_BOUNCED = 'bounced';
   const STATUS_INACTIVE = 'inactive';
 
+  const DEPRECATED_LINK_TOKEN_LENGTH = 6;
```
pavel-mailpoet: Shall we add a comment describing why is this deprecated? Why is it deprecated anyway? I thought that deprecated is something that can be removed eventually. Can this be removed?

lib/Models/Subscriber.php
```diff
@@ -120,7 +123,10 @@ static function getCurrentWPUser() {
     return self::where('wp_user_id', $wp_user->ID)->findOne();
   }
 
-  static function generateToken($email = null, $length = 32) {
+  /**
+   * @deprecated Only for backward compatibility for old tokens
```
pavel-mailpoet: I don't think this can be deprecated. We have to leave this here forever for compatibility reasons, don't we?

lib/Subscribers/ImportExport/Import/Import.php
```diff
@@ -297,14 +298,20 @@ private function setSubscriptionStatusToSubscribed($subscribers_data) {
     return $subscribers_data;
   }
 
-  function setSource($subscribers_data) {
+  private function setSourceAndLinkToken($subscribers_data) {
```
pavel-mailpoet: Don't you think it would be clearer and easier to understand if there were two methods? One for setting the source and second for the token?

costasovo: You are right two methods will be nicer.

lib/Models/Subscriber.php
costasovo: Deprecated might not be the best word. I just want to make sure that no one will use this function with some new feature. I think we can go with comment.

lib/Models/Subscriber.php
costasovo: What name do you suggest? What about OLD_LINK_TOKEN_LENGTH ?

pavel-mailpoet: Maybe `OBSOLETE_TOKENS_LENGTH`?

PR #2330, opened 2019-09-18
Tests skipping [MAILPOET-2358]

pavel-mailpoet:
[MAILPOET-2358]

[MAILPOET-2358]: https://mailpoet.atlassian.net/browse/MAILPOET-2358


pavel-mailpoet:
> is there a possibility to check if any were skipped after the test suite

I don't think so. I tried to find a way, but haven't found one.

PR #2331, opened 2019-09-18
Don't allow deleting list which is assigned to automatic email [MAILPOET-2352]

NeosinneR:
[MAILPOET-2352]

[MAILPOET-2352]: https://mailpoet.atlassian.net/browse/MAILPOET-2352

### DIFF
--- a/views/segments.html
+++ b/views/segments.html
@@ -59,6 +59,7 @@
     'restore': __('Restore'),
     'deletePermanently': __('Delete permanently'),
     'save': __('Save'),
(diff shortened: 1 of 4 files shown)


views/segments.html
```diff
@@ -59,6 +59,7 @@
     'restore': __('Restore'),
     'deletePermanently': __('Delete permanently'),
     'save': __('Save'),
+    'trashDisallowed': _x("List cannot be deleted because it's used for %$1s email", 'Alert shown when trying to delete segment, which is assigned to any automatic emails.'),
```
pavel-mailpoet: Could you use `it’s` instead of `it's` here?

codemonkey-jack:
This fails QA. 

When you try and delete a list, the error message shows. However, it just keeps repeating the newsletter title. 

image

PR #2332, opened 2019-09-18
Update template sections [MAILPOET-2345]

pavel-mailpoet:
[MAILPOET-2345]

[MAILPOET-2345]: https://mailpoet.atlassian.net/browse/MAILPOET-2345

### DIFF
--- a/assets/js/src/newsletters/templates.jsx
+++ b/assets/js/src/newsletters/templates.jsx
@@ -36,10 +36,6 @@ if (window.mailpoet_woocommerce_active) {
 templatesCategories.push(
   ...[
     {
-      name: 'all',
-      label: MailPoet.I18n.t('allTemplates'),
-    },
-    {
       name: 'recent',
       label: MailPoet.I18n.t('recentlySent'),
     },
--- a/assets/js/src/newsletters/templates.jsx
+++ b/assets/js/src/newsletters/templates.jsx
@@ -24,6 +24,10 @@ const templatesCategories = [
     name: 'notification',
     label: MailPoet.I18n.t('tabNotificationTitle'),
   },
+  {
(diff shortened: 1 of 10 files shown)


assets/js/src/newsletters/templates.jsx
```diff
@@ -35,10 +35,6 @@ if (window.mailpoet_woocommerce_active) {
 
 templatesCategories.push(
   ...[
-    {
-      name: 'all',
-      label: MailPoet.I18n.t('allTemplates'),
```
amine-mp: We should also remove `allTemplates` from translations on `newsletters.html` view.

lib/Config/Populator.php
```diff
@@ -554,20 +552,6 @@ private function createSourceForSubscribers() {
     );
   }
 
-  private function updateNewsletterCategories() {
-    global $wpdb;
-    // perform once for versions below or equal to 3.14.0
-    if (version_compare($this->settings->get('db_version', '3.14.1'), '3.14.0', '>')) {
-      return false;
-    }
-    $query = "UPDATE `%s` SET categories = REPLACE(REPLACE(categories, ',\"blank\"', ''), ',\"sample\"', ',\"all\"')";
```
amine-mp: This query removed `blank` and replaced `sample` with `all`. Now that we create the `blank` category, what happens to templates having the `sample` category? Wouldn't it make sense to replace `sample` by `blank`?

pavel-mailpoet: > Wouldn't it make sense to replace `sample` by `blank`?

Definitely not. That would show all the templates in the blank section. That is not the point. We don't want to show all templates there, only the simple text ones.

PR #2334, opened 2019-09-18
Create and store WC transactional email template [MAILPOET-2280]

amine-mp:
[MAILPOET-2280]

[MAILPOET-2280]: https://mailpoet.atlassian.net/browse/MAILPOET-2280

### DIFF
--- a/lib/Entities/NewsletterEntity.php
+++ b/lib/Entities/NewsletterEntity.php
@@ -23,6 +23,7 @@ class NewsletterEntity {
   const TYPE_WELCOME = 'welcome';
   const TYPE_NOTIFICATION = 'notification';
   const TYPE_NOTIFICATION_HISTORY = 'notification_history';
+  const TYPE_WC_TRANSACTIONAL_EMAIL = 'wc_transactional';
 
   // standard newsletters
   const STATUS_DRAFT = 'draft';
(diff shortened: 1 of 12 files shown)


lib/WooCommerce/TransactionalEmails.php
```diff
@@ -0,0 +1,833 @@
...
+
+  function __construct(WPFunctions $wp, SettingsController $settings) {
+    $this->wp = $wp;
+    $this->settings = $settings;
+  }
+
+  function init() {
+    $saved_email_id = (bool)$this->settings->get(self::SETTING_EMAIL_ID, false);
+    if (!$saved_email_id) {
+      $email = Newsletter::createOrUpdate([
```
pavel-mailpoet: Shouldn't this be a newsletter template? That is what the specification says.

Also, why did you decide to keep the file here instead of having it somewhere in the `Config` where the rest of the templates live?

lib/Entities/NewsletterEntity.php
```diff
@@ -23,6 +23,7 @@ class NewsletterEntity {
   const TYPE_WELCOME = 'welcome';
   const TYPE_NOTIFICATION = 'notification';
   const TYPE_NOTIFICATION_HISTORY = 'notification_history';
+  const TYPE_WC_TRANSACTIONAL_EMAIL = 'wc_transactional';
```
pavel-mailpoet: Correct me if I am wrong, but I thought we decided to use the full `WooCommerce` rather than using the `WC` abbreviation.

codemonkey-jack:
@amine-mp This fails QA. I enabled the feature flag and have WooCommerce installed.
No email template was created and no options are saved to the database on my test site.

PR #2340, opened 2019-09-19
Fix features controller usage before plugin init [MAILPOET-2373]

costasovo:


lib/Features/FeaturesController.php
```diff
@@ -33,7 +35,13 @@ function isSupported($feature) {
     if (!$this->exists($feature)) {
       throw new \RuntimeException("Unknown feature '$feature'");
     }
-    $this->ensureFlagsLoaded();
+    // In case the controller is called before plugin initialization with plugin version prior 3.26.1
+    // which doesn't have feature_flags table we return default value
```
JanJakes: We can make the comment more universal:
```
// ensure controller works even if used before migrator, return default value in such case
```

PR #2342, opened 2019-09-20
readme.txt - add Hebrew language to the translations list

ramiy:


codemonkey-jack:
@ramiy I'm going to close this without merging. Why? Because we currently have some rendering/display issues with the UI with RTL languages and we don't officially support them, which is why we haven't listed the language anywhere.

ramiy:
@codemonkey-jack I can help we RTL and all the required CSS changes.

ramiy:
@codemonkey-jack

I should note that you do display other RTL languages like "Arabic" and "Persian".

PR #2343, opened 2019-09-23
Fix user being unsubscribed from non default lists [MAILPOET-2292]

NeosinneR:
[MAILPOET-2292]

[MAILPOET-2292]: https://mailpoet.atlassian.net/browse/MAILPOET-2292


lib/API/JSON/v1/Subscribers.php
```diff
@@ -356,6 +358,28 @@ function save($data = []) {
     );
   }
 
+  private function getNonDefaultSubscribedSegments(array $data) {
+    $subscribed_segment_ids = [];
+    if (isset($data['id']) && (int)$data['id'] > 0) {
```
costasovo: It would be nicer and more readable to use fast termination here.

```
if (!isset($data['id']) || (int)$data['id'] <= 0) { return ... }
```

PR #2345, opened 2019-09-23
Allow modifying cron limits [MAILPOET-2367]

pavel-mailpoet:
[MAILPOET-2367]

[MAILPOET-2367]: https://mailpoet.atlassian.net/browse/MAILPOET-2367


wxa:
Could you replace `DAEMON_EXECUTION_LIMIT` with a function call in tests for consistency?

PR #2359, opened 2019-09-25
Prevent multiple instances of inactive subscribers sync worker [MAILPOET-2385]

wxa:
Now if we want to prevent a simple worker from running in parallel in case of slowness, we need it to extend `SingleInstanceSimpleWorker` and update its code to reset the `in_progress` flag in case of exceptions or cron timeouts.


lib/Cron/Workers/InactiveSubscribers.php
```diff
@@ -28,26 +28,31 @@ function __construct(
...
+      // Handle activation/deactivation within interval
+      while ($this->inactive_subscribers_controller->markInactiveSubscribers($days_to_inactive, self::BATCH_SIZE) === self::BATCH_SIZE) {
+        CronHelper::enforceExecutionLimit($this->timer);
+      };
+      while ($this->inactive_subscribers_controller->markActiveSubscribers($days_to_inactive, self::BATCH_SIZE) === self::BATCH_SIZE) {
+        CronHelper::enforceExecutionLimit($this->timer);
+      };
       self::schedule();
-      return true;
+    } catch (\Exception $e) {
```
costasovo: I think it would be more readable to move the wrapping into try/catch block one level up into `processTask`. I see exactly same try/catch with same error handling in `WooCommerceSync` worker.
If we move it in process task and wrap `processTaskStrategy` we could easily see what happens in case of exception directly from the processTask method which is the one which is handling the "administration" stuff of task processing.

wxa:
@costasovo Very good comments. I've made changes to address all of them. 

I was unsure about implementing the feature in `SimpleWorker` as it might become too big, but your opinion tipped the scale. I also hesitated to make progress flag-related methods private: I think it somewhat breaks encapsulation by having tests to access the inner structure of the `meta` array. E.g. if we want to make `in_progress` a separate column in the future, we'll have to rework tests a bit more. But it's a model, not a class property and it's not widely used so it's not a hard point.

lib/Cron/Workers/SimpleWorker.php
```diff
@@ -97,12 +101,34 @@ function processTask(ScheduledTask $task) {
     // abort if execution limit is reached
     CronHelper::enforceExecutionLimit($this->timer);
 
-    if ($this->processTaskStrategy($task)) {
+    if (!static::SUPPORT_MULTIPLE_INSTANCES) {
+      if ($this->isInProgress($task)) {
```
costasovo: I think these two conditions (isInProgress and rescheduleOutdated) have to switch order. I had it wrong in my previous comment, sry :(
With the current order, it would not reschedule a task which is in progress.

lib/Cron/Workers/SimpleWorker.php
```diff
@@ -97,12 +101,34 @@ function processTask(ScheduledTask $task) {
...
+      if ($this->rescheduleOutdated($task)) {
+        return false;
+      }
+      $this->startProgress($task);
+    }
+
+    try {
+      $completed = $this->processTaskStrategy($task);
+    } catch (\Exception $e) {
+      if (!static::SUPPORT_MULTIPLE_INSTANCES) {
```
costasovo: Just an idea. I'm thinking about moving $this->startProgress($task) and $this->endProgress($task) out of the `if (!static::SUPPORT_MULTIPLE_INSTANCES)` block so that we start storing this for all workers. I think it might be useful information for all workers. We would be able to see in DB which task is now actually being processed and could be useful during debugging.

lib/Cron/Workers/SimpleWorker.php
wxa: Good catch, this is a bug indeed, fixed.

lib/Cron/Workers/SimpleWorker.php
wxa: Sounds reasonable. Made a change, we now track progress for all workers.

wxa: Regarding this change, while it may be handy to set this for all workers, at the time some workers may use the `meta` field in their own way which may conflict with an array having an `in_progress` key. The failed test shows that. I can fix the test, but there is still risk of introducing bugs in other workers. The most clean way to avoid such issues would be to make `in_progress` a separate column. 

We decided to roll back this change and refactor later.

PR #2360, opened 2019-09-25
Templates speed-up [MAILPOET-2378]

NeosinneR:
[MAILPOET-2378]

[MAILPOET-2378]: https://mailpoet.atlassian.net/browse/MAILPOET-2378

### DIFF
--- a/assets/js/src/ajax.js
+++ b/assets/js/src/ajax.js
@@ -29,6 +29,9 @@ MailPoet.Ajax = {
   post: function post(options) {
     return this.request('post', options);
   },
+  get: function get(options) {
+    return this.request('get', options);
+  },
   init: function init(options) {
     // merge options
     this.options = jQuery.extend({}, this.defaults, options);
@@ -70,7 +73,7 @@ MailPoet.Ajax = {
     }
 
     // ajax request
-    deferred = jQuery.post(
+    deferred = jQuery[method](
       this.options.url,
       params,
       null,
--- a/assets/js/src/ajax.js
+++ b/assets/js/src/ajax.js
@@ -56,7 +56,7 @@ MailPoet.Ajax = {
       data: this.options.data || {},
     };
   },
-  constructGETUrl: function constructGETUrl(options) {
+  constructGetUrl: function constructGetUrl(options) {
     this.init(options);
     return this.options.url + '?' + jQuery.param(this.getParams());
   },
--- a/assets/js/src/ajax.js
+++ b/assets/js/src/ajax.js
@@ -56,6 +56,10 @@ MailPoet.Ajax = {
       data: this.options.data || {},
     };
   },
+  constructGETUrl: function constructGETUrl(options) {
+    this.init(options);
+    return this.options.url + '?' + jQuery.param(this.getParams());
+  },
   request: function request(method, options) {
     var params;
     var deferred;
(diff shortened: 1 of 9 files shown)


lib/API/JSON/API.php
```diff
@@ -79,7 +79,11 @@ function init() {
 
   function setupAjax() {
     $this->wp->doAction('mailpoet_api_setup', [$this]);
-    $this->setRequestData($_POST);
+    if (isset($_POST['api_version'])) {
+      $this->setRequestData($_POST);
+    } else {
+      $this->setRequestData($_GET);
+    }
```
JanJakes: Allowing GET for all API endpoints is probably not a good idea. GET can transport a limited amount of data, GET should have no side-effects (well, in REST), GET can be cached, and it could cause security issues (pointing an admin to URL that would perform something).

What we can do is to go a bit forward REST here maybe. Converting this one single endpoint to `GET`, completely (allowing only GET for this one). Another option is to add a GET wrapper - a new route just for this. There is also https://developer.wordpress.org/rest-api/ but that may be a bigger topic, not sure.

assets/js/src/ajax.js
```diff
@@ -56,6 +56,10 @@ MailPoet.Ajax = {
       data: this.options.data || {},
     };
   },
+  constructGETUrl: function constructGETUrl(options) {
```
JanJakes: Not important but nicer version of camel case is `constructGetUrl` - more readable, unambiguous.

lib/API/JSON/API.php
```diff
@@ -244,4 +238,43 @@ function createErrorResponse($error_type, $error_message, $response_status) {
...
+      return [];
+    }
+    if (
+      $type === 'get'
+      && (!isset(self::$get_endpoints[$endpoint]) || !in_array($method, self::$get_endpoints[$endpoint]))
+      ||
+      $type === 'post'
+      && (isset(self::$get_endpoints[$endpoint]) && in_array($method, self::$get_endpoints[$endpoint]))
+    ) {
+      return [];
```
JanJakes: Wrong method should end at least with _some_ error. We should not pretend that all is OK. Normally, it would be a bad-method or a 404 error but in our API [we seem to convert everything to bad-request](https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/API.php#L188) so I guess that's OK.

Moreover, such check should probably be in [processRoute](https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/API.php#L156), not here when getting data.

lib/API/JSON/API.php
```diff
@@ -36,6 +36,11 @@ class API {
   /** @var SettingsController */
   private $settings;
 
+  /** @var array */
+  private static $get_endpoints = [
+    'newsletterTemplates' => ['getAll'],
+  ];
```
JanJakes: I'd recommend not doing this here in the parent API class. I'd suggest to keep it in endpoint class themselves:

1. Make it `protected static $get_methods = []`, leave it empty here.
2. In the inherited class itself make it `protected static $get_methods = ['getAll']`.
3. Use late static binding when accessing the array (`static::$get_methods` instead of `self::$get_methods`).

lib/API/JSON/API.php
NeosinneR: Discarded the whole commit and refactored it as suggested.

PR #2361, opened 2019-09-25
Disable buttons after first click [MAILPOET-2273]

amine-mp:
[MAILPOET-2273]

[MAILPOET-2273]: https://mailpoet.atlassian.net/browse/MAILPOET-2273

### DIFF
--- a/assets/js/src/newsletters/types.jsx
+++ b/assets/js/src/newsletters/types.jsx
@@ -13,6 +13,13 @@ class NewsletterTypes extends React.Component {
     }).isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      isCreating: false,
+    };
+  }
+
   setupNewsletter = (type) => {
     if (type !== undefined) {
       this.props.history.push(`/new/${type}`);
@@ -23,7 +30,7 @@ class NewsletterTypes extends React.Component {
     }
   };
 
-  getAutomaticEmails = () => {
+  getAutomaticEmails() {
     if (!window.mailpoet_automatic_emails) return [];
 
     return _.map(window.mailpoet_automatic_emails, (automaticEmail) => {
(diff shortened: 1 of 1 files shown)


assets/js/src/newsletters/types.jsx
```diff
@@ -23,7 +30,7 @@ class NewsletterTypes extends React.Component {
     }
   };
 
-  getAutomaticEmails = () => {
+  getAutomaticEmails() {
```
JanJakes: Why did you change the type of the function? I see you used `.bind(this)` below so it's OK I guess but is there any reason? Now some functions in this class are arrow functions, some are not.

assets/js/src/newsletters/types.jsx
```diff
@@ -73,98 +81,96 @@ class NewsletterTypes extends React.Component {
...
+                event.preventDefault();
+                createWelcomeNewsletter();
+              }
+            }}
+            role="button"
+            tabIndex={0}
+          >
+            {MailPoet.I18n.t('setUp')}
+          </a>
+        ),
```
JanJakes: I see you unwrapped the closure and then added the `disabled` attribute. It's all in one commit so I don't know why. Then I have to ask - did you unwrap it just because it's nicer or because it's needed?

assets/js/src/newsletters/types.jsx
amine-mp: Normally all methods should be functions in order to access `this` properly. But ESLint complains about using a `function` and not using `this` inside it. So we ended up converting those to `arrow functions`. Now I changed this to be a `function` because I needed to use `this` inside.

assets/js/src/newsletters/types.jsx
amine-mp: Because it's needed. if I keep the closure/anonymous function, then I can't access the class instance using `this`.

assets/js/src/newsletters/types.jsx
JanJakes: I see. However:
1. Arrow functions do allow you to access the correct `this` without the need to bind it.
2. It may be indeed better to use bound functions in some cases (see https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1) but in such case, the `bind()` belongs to the constructor, not as a side-effect of `render` method (https://github.com/mailpoet/mailpoet/pull/2361/files#diff-e7956c90afbfb37c0f8a5bca3e5ebfa3R95).

amine-mp:
You are right @JanJakes, there is no need to change methods from arrow functions because I removed closures on action render, which were causing the actual issue.

PR #2363, opened 2019-09-27
Disable WooCommerce plugin's email settings [MAILPOET-2283]

NeosinneR:
There is a bit of a difference from specs - I'm not using `woocommerce_email_settings_after` hook as suggested, because it's a hook for specific email settings. Instead, I'm using `woocommerce_settings_start` (which is called on every settings page) and then filtering in MailPoet which tab is shown. 

It's done without JS so the positioning of buttons is fixed and not calculated. It may break if they change the settings page layout, but that might break even if JS would be included. 

[MAILPOET-2283]

[MAILPOET-2283]: https://mailpoet.atlassian.net/browse/MAILPOET-2283

### DIFF
--- a/lib/WooCommerce/Settings.php
+++ b/lib/WooCommerce/Settings.php
@@ -2,29 +2,29 @@
 
 namespace MailPoet\WooCommerce;
 
+use MailPoet\Config\Renderer;
 use MailPoet\Features\FeaturesController;
 use MailPoet\Settings\SettingsController;
-use MailPoet\WP\Functions as WPFunctions;
 
 class Settings {
 
   /** @var FeaturesController */
   private $features_controller;
 
+  /** @var Renderer */
+  private $renderer;
+
   /** @var SettingsController */
   private $settings;
 
-  /** @var WPFunctions */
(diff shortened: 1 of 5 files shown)


lib/WooCommerce/Settings.php
```diff
@@ -2,8 +2,58 @@
...
-    // TODO
+    if ($_GET['tab'] !== 'email') {
+      return;
+    }
+    if (!(bool)$this->settings->get('woocommerce.use_mailpoet_editor')) {
+      return;
+    }
+    $woocommerce_template_id = $this->settings->get('woocommerce.transactional_email_id');
+    ?>
+
```
amine-mp: Is it possible to put the CSS and HTML code inside a Twig view and render it instead of mixing everything here?

NeosinneR: Great suggestion, moved to Twig view. Also fixed checking undefined array items.

amine-mp:
Good job on using Twig to render the HTML, I am just curious if it would make sense to have an acceptance test for this feature. What do you think?

PR #2364, opened 2019-09-27
Merge React contexts into a global one [MAILPOET-2388]

amine-mp:
[MAILPOET-2388]

[MAILPOET-2388]: https://mailpoet.atlassian.net/browse/MAILPOET-2388

### DIFF
--- a/assets/js/src/context/index.jsx
+++ b/assets/js/src/context/index.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
-import getFeaturesContext from './getFeaturesContext';
-import getConstantsContext from './getConstantsContext';
+import getFeaturesContext from './getFeaturesContext.jsx';
+import getConstantsContext from './getConstantsContext.jsx';
 
 /**
  * Builds the value of the global context.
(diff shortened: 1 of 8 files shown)


assets/js/src/context/index.jsx
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Builds the value of the global context.
+ * This starts with `use` because it needs to call
+ * some React hooks to build some parts of the context.
+ */
+export function useGlobalContextValue(data) {
+  const features = getFeaturesContext(data);
+  const constants = getConstantsContext(data);
+  return { features, constants };
```
pavel-mailpoet: I am thinking about the grouping you did here and I am not sure if it is a good idea or not. It seems to me that we will have a growing `constants` context and not much anything else. I am not sure if that is good or not. Just an idea, not sure if better or worse:
```
context: {
  users: {
    isNewUser: true,
  },
  segments: {
    all: [...],
  },
  features: {},
},
```

amine-mp: I am not sure which structure is better for now, because I don't know all constants we put into `window`. My idea was to just create the context now with minimal data. Then after providing it to all components, we can then take a look at all constants we pass into window and define a good structure for the global context. What do you think about that?

PR #2368, opened 2019-09-30
Remove intro [MAILPOET-2402]

pavel-mailpoet:
[MAILPOET-2402]

[MAILPOET-2402]: https://mailpoet.atlassian.net/browse/MAILPOET-2402


lib/Config/Changelog.php
```diff
@@ -84,9 +84,6 @@ private function checkMp2Migration($version) {
 
   private function setupNewInstallation() {
     // ensure there was no MP2 migration (migration resets $version so it must be checked)
```
JanJakes: The commend needs to be removed as well. Otherwise, the result is confusing.

PR #2369, opened 2019-09-30
Ensure that unsubscribe link is always saved [MAILPOET-2245]

costasovo:
[MAILPOET-2245](https://mailpoet.atlassian.net/browse/MAILPOET-2245)

### DIFF
--- a/lib/Models/NewsletterLink.php
+++ b/lib/Models/NewsletterLink.php
@@ -10,6 +10,7 @@ namespace MailPoet\Models;
  */
 class NewsletterLink extends Model {
   public static $_table = MP_NEWSLETTER_LINKS_TABLE;
+  const UNSUBSCRIBE_LINK_SHORT_CODE = '[link:subscription_unsubscribe_url]';
 
   /**
    * @param Newsletter $newsletter
--- a/lib/Newsletter/Links/Links.php
+++ b/lib/Newsletter/Links/Links.php
@@ -12,6 +12,8 @@ use MailPoet\Util\Helpers;
 use MailPoet\Util\Security;
 use MailPoet\Util\pQuery\pQuery as DomParser;
 
+use function MailPoet\Util\array_column;
+
 class Links {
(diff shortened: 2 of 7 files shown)


lib/Models/NewsletterLink.php
```diff
@@ -10,6 +10,7 @@
  */
 class NewsletterLink extends Model {
   public static $_table = MP_NEWSLETTER_LINKS_TABLE;
+  const UNSUBSCRIBE_LINK_SHORT_CODE = '[link:subscription_unsubscribe_url]';
```
pavel-mailpoet: If you introduced this constant should we also replace strings to start using it?

tests/integration/Newsletter/Links/LinksTest.php
```diff
@@ -8,6 +8,7 @@
 use MailPoet\Newsletter\Links\Links;
 use MailPoet\Newsletter\Shortcodes\Categories\Link;
 use MailPoet\Router\Router;
+use function MailPoetVendor\Symfony\Component\DependencyInjection\Loader\Configurator\expr;
```
pavel-mailpoet: Is this used? I can see where.

lib/Newsletter/Links/Links.php
```diff
@@ -170,6 +133,20 @@ static function save(array $links, $newsletter_id, $queue_id) {
     }
   }
 
+  static function ensureUnsubscribeLink(array $processed_links) {
+    if (in_array(
+      NewsletterLink::UNSUBSCRIBE_LINK_SHORT_CODE,
+      \MailPoet\Util\array_column($processed_links, 'link'))
```
pavel-mailpoet: Why did you use fully qualified namespace here? AFAIK we use `use` everywhere.

tests/integration/Newsletter/Links/LinksTest.php
costasovo: Good catch! This was added by PHP Storm by accident...

PR #2370, opened 2019-09-30
Code sniffer rules [MAILPOET-2409]

JanJakes:


costasovo:
@codemonkey-jack These are just coding styles changes.

PR #2373, opened 2019-10-01
Show TinyMCE toolbar above WP menu [MAILPOET-2404]

NeosinneR:
I didn't find any option in docs to set offset for the fixed TinyMCE toolbar. So the "solution" here is, that it's shown above the WordPress menu.

[MAILPOET-2404]

[MAILPOET-2404]: https://mailpoet.atlassian.net/browse/MAILPOET-2404

### DIFF
--- a/assets/css/src/components/newsletterEditor/_libraryOverrides.scss
+++ b/assets/css/src/components/newsletterEditor/_libraryOverrides.scss
@@ -29,7 +29,9 @@
     border: 1px solid $content-border-color;
     border-radius: 3px;
     box-shadow: 0 0 3px 1px rgba(0, 0, 0, .05);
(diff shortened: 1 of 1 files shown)


codemonkey-jack:
@NeosinneR This fails QA due to the negative impact the fix has on the user experience. 

This includes:

- difficult to select existing items that the toolbar covers on scroll up and scroll down. 
- On scroll down toolbar is far above the content you want to edit. 

I've upload a screencast to the Jira ticket, as GitHub doesn't allow MP4 uploads. 

I'd recommend checking with Quentin in #product on Slack about what he wants to do. My recommendation would be to do nothing as the original ticket didn't seem to have that large a impact and we hadn't received complaints about it before.

PR #2375, opened 2019-10-01
Swiftmailer [MAILPOET-2363]

JanJakes:


lib/Config/RequirementsChecker.php
```diff
@@ -17,9 +17,6 @@ class RequirementsChecker {
   public $vendor_classes = [
     '\ORM',
     '\Model',
-    '\Swift_Mailer',
-    '\Swift_SmtpTransport',
-    '\Swift_Message',
```
pavel-mailpoet: Why did you remove those lines from here? The commit message says "Move Swiftmailer to vendor-prefixed" I don't see how it is relevant.

JanJakes: @pavel-mailpoet It is, classes such as `\Swift_Mailer` won't exist anymore, it will be `MailPoetVendor\Swift_Mailer` but that one we don't need to check for conflicts.

PR #2378, opened 2019-10-03
Create WC Content widget [MAILPOET-2279]

amine-mp:
[MAILPOET-2279]

[MAILPOET-2279]: https://mailpoet.atlassian.net/browse/MAILPOET-2279

### DIFF
--- a/assets/js/src/newsletter_editor/blocks/woocommerceContent.js
+++ b/assets/js/src/newsletter_editor/blocks/woocommerceContent.js
@@ -1,10 +1,9 @@
-/* eslint-disable func-names */
 import App from 'newsletter_editor/App';
 import BaseBlock from 'newsletter_editor/blocks/base';
 
 const BlockModel = BaseBlock.BlockModel.extend({
   stale: ['styles'],
-  defaults: function () {
+  defaults() {
     return this._getDefaults({
       type: 'woocommerceContent',
       styles: {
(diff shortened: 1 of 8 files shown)


assets/js/src/newsletter_editor/blocks/woocommerceContent.js
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable func-names */
```
pavel-mailpoet: WHY? I thought we want to remove those and not add more 😞

amine-mp: I removed that comment and fixed the ESLint warnings.

codemonkey-jack:
@amine-mp I tried on a new install and on an existing install and in both cases the email didn't show the new content block (screenshot attached).

MailPoet - Newsletter ‹ qa7 — WordPress 2019-10-07 10-01-33

amine-mp:
Hey @codemonkey-jack 
Sorry for not clarifying this before. if the `woocomerce.transactional_email_id` exists then the plugin will not create the new transactional email containing the new widget. That's why you need to remove the `woocommerce` setting before reactivating the plugin:
```sql
DELETE FROM wp_mailpoet_settings WHERE name = 'woocommerce';
```
Please try and let me know.

codemonkey-jack:
@amine-mp Thanks! Ok, so after testing this does fail QA. There are two issues. 

1. In the text: `Just to let you know — we've received your order #0001, and it is now being processed:` the dash shows as the HTML entity and not a dash. 

2. The specs state: 

> Get color from WooCommerce > Settings > Emails > Base color  and use it to set the color of the titles

But if this color is changed the email isn't updated. It's only done for the initial generation. You need to hook in and update the color when the settings change. 

MailPoet - Newsletter ‹ qa7 — WordPress 2019-10-08 10-08-36
MailPoet - Newsletter ‹ qa7 — WordPress 2019-10-08 10-11-36
WooCommerce settings ‹ qa7 — WordPress 2019-10-08 10-08-20

PR #2385, opened 2019-10-07
Block access to WC Email customizer if disabled [MAILPOET-2408]

NeosinneR:
[MAILPOET-2408]

[MAILPOET-2408]: https://mailpoet.atlassian.net/browse/MAILPOET-2408

### DIFF
--- a/lib/AdminPages/Pages/NewsletterEditor.php
+++ b/lib/AdminPages/Pages/NewsletterEditor.php
@@ -54,7 +54,7 @@ class NewsletterEditor {
       )
     ) {
       header('Location: admin.php?page=mailpoet-settings&enable-customizer-notice#woocommerce', true, 302);
-      return;
+      exit;
     }
 
     $subscriber = Subscriber::getCurrentWPUser();
--- a/lib/AdminPages/Pages/NewsletterEditor.php
+++ b/lib/AdminPages/Pages/NewsletterEditor.php
@@ -9,6 +9,7 @@ use MailPoet\Newsletter\Shortcodes\ShortcodesHelper;
 use MailPoet\Services\Bridge;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Settings\UserFlagsController;
+use MailPoet\WooCommerce\Helper as WooCommerceHelper;
 use MailPoet\WP\Functions as WPFunctions;
 
 class NewsletterEditor {
@@ -21,6 +22,9 @@ class NewsletterEditor {
   /** @var UserFlagsController */
   private $user_flags;
 
+  /** @var WooCommerceHelper */
+  private $woocommerce_helper;
+
   /** @var WPFunctions */
   private $wp;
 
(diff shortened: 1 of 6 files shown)


tests/acceptance/WooCommerceSettingsTabCest.php
```diff
@@ -62,4 +67,20 @@ function checkWooCommercePluginSettingsAreDisabled(\AcceptanceTester $I) {
     $I->amOnPage("/wp-admin/admin.php?page=wc-settings&tab=email&section=wc_email_new_order");
     $I->dontSeeElementInDOM(self::CUSTOMIZE_SELECTOR);
   }
+
+  function showNoticeWhenWooCommerceCustomizerIsDisabled(\AcceptanceTester $I) {
+    $this->features->withFeatureEnabled(FeaturesController::WC_TRANSACTIONAL_EMAILS_CUSTOMIZER);
+    $this->transactional_emails->init(); // TODO: remove when WC_TRANSACTIONAL_EMAILS_CUSTOMIZER flag is removed
```
JanJakes: Instead of this "hack", could you please try to `$I->deactivatePlugin('mailpoet');` and then reactivate it to trigger the populator? It will be slow but you can add the comment about removing it when the feature flag is removed.

This way of using `_inject` on some internal project class is something that's not really "allowed" and it's only due to old static models that it actually works in this case. Let's not add such things to our tests.

assets/js/src/newsletter_editor/initializer.jsx
```diff
@@ -29,7 +29,7 @@ const initializeEditor = (config) => {
     !parseInt(window.mailpoet_woocommerce_settings.use_mailpoet_editor, 10)
     && getUrlParam('id') === window.mailpoet_woocommerce_settings.transactional_email_id
   ) {
-    window.location.href = 'admin.php?page=mailpoet-settings#woocommerce';
+    window.location.href = 'admin.php?page=mailpoet-settings#woocommerce/enable-notice';
```
JanJakes: The URL sounds like "enable WooCommerce" notice. I'd suggest something like `#woocommerce/enable-customizer-notice`.

views/newsletter/editor.html
```diff
@@ -1416,6 +1416,7 @@ <h1 style="display:none"><%= __('Newsletter Editor') %></h1>
       installedAt: '<%= settings["installed_at"] %>',
       mtaMethod: '<%= settings["mta"]["method"] %>',
     };
+    var mailpoet_woocommerce_settings = <%= json_encode(settings['woocommerce']) %>;
```
JanJakes: In `newsletters.html` it's passed like this https://github.com/mailpoet/mailpoet/blob/master/views/newsletters.html#L35. Not important, just saying, we may want to unify it.

assets/js/src/newsletter_editor/initializer.jsx
```diff
@@ -25,6 +25,14 @@ const initializeEditor = (config) => {
 
   if (!editorContainer || !window.EditorApplication) return;
 
+  if (
+    !parseInt(window.mailpoet_woocommerce_settings.use_mailpoet_editor, 10)
+    && getUrlParam('id') === window.mailpoet_woocommerce_settings.transactional_email_id
+  ) {
+    window.location.href = 'admin.php?page=mailpoet-settings#woocommerce';
+    return;
+  }
```
JanJakes: I was wondering why this is not done in PHP since 'id' is a real URL param. Here: https://github.com/mailpoet/mailpoet/blob/master/lib/AdminPages/Pages/NewsletterEditor.php#L39

But I guess it's not too important? I mean, in PHP it's safe and it will occur sooner. You could also use `WPNotice` (although that won't survive redirect I guess but you can still use some `?param=...` and do `WPNotice` in PHP). Does this need to be secure? Is it worth it to move to PHP? It would also be a bit simpler - no new routes and stuff.

NeosinneR:
@JanJakes rewritten to use PHP for redirect and notice, which addresses all the issues

lib/AdminPages/Pages/NewsletterEditor.php
```diff
@@ -21,22 +22,41 @@ class NewsletterEditor {
...
   }
 
   function render() {
+    $newsletter_id = (isset($_GET['id']) ? (int)$_GET['id'] : 0);
+    $woocommerce_template_id = (int)$this->settings->get('woocommerce.transactional_email_id', null);
+    if (
+      $woocommerce_template_id
+      && $newsletter_id === $woocommerce_template_id
+      && (
+        !$this->woocommerce_helper->isWooCommerceActive()
```
JanJakes: Seeing this line I think when WC is not active then `page=mailpoet-settings#woocommerce` won't even exist. I guess the spec doesn't say what to do in such case, does it?

NeosinneR: Added redirect in JS when the user tries to access the WooCommerce tab without WooCommerce being active. I specifically didn't use `this.tabs('basics)` because that would only change the tab without changing the URL.

PR #2391, opened 2019-10-08
Store 'in_progress' flag for workers in a separate column [MAILPOET-2443]

wxa:


lib/Cron/Workers/SimpleWorker.php
```diff
@@ -150,21 +150,20 @@ function reschedule(ScheduledTask $task, $timeout) {
   }
 
   private function isInProgress(ScheduledTask $task) {
-    $meta = $task->getMeta();
-    if (!empty($meta['in_progress'])) {
+    if (!empty($task->in_progress)) {
```
costasovo: This could be refactored just to `return !empty($task->in_progress)`, but it doesn't have to be addressed.

PR #2395, opened 2019-10-09
Logger refactor [MAILPOET-2444]

costasovo:


tests/integration/Logging/LoggerTest.php
```diff
@@ -9,66 +9,70 @@ class LoggerTest extends \MailPoetTest {
   /** @var SettingsController */
   private $settings;
 
+  /** @var LoggerFactory */
+  private $logger_factory;
```
pavel-mailpoet: This is a unit test now, isn't it? We can move it to the unit tests suite, can't we?

costasovo: Good point. Settings are now injected so we can mock it and move the test to unit tests. I'll add a commit.

PR #2403, opened 2019-10-10
Remove WC templates from saved templates tab [MAILPOET-2467]

amine-mp:
[MAILPOET-2467]

[MAILPOET-2467]: https://mailpoet.atlassian.net/browse/MAILPOET-2467

### DIFF
--- a/assets/js/src/newsletters/templates.jsx
+++ b/assets/js/src/newsletters/templates.jsx
@@ -101,6 +101,9 @@ class NewsletterTemplates extends React.Component {
     const categoriesNames = templatesCategories.map((category) => category.name);
     let categories;
 
+    if (categoriesNames.indexOf('woocommerce') === -1) {
+      categoriesNames.push('woocommerce');
+    }
     try {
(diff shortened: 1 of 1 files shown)


assets/js/src/newsletters/templates.jsx
```diff
@@ -101,6 +101,9 @@ class NewsletterTemplates extends React.Component {
     const categoriesNames = templatesCategories.map((category) => category.name);
     let categories;
 
+    if (categoriesNames.indexOf('woocommerce') === -1) {
```
pavel-mailpoet: From the specification, It seems that we should show those templates if WooCommerce is active. This looks like you hide those templates always, don't you?

amine-mp: No, I don't. If WooCommerce is active, we do add `woocommerce` category on line 34, which makes this condition false.
In other terms, we will always have the `woocommerce` category but we will show the WC tab only when WC is active. Makes sense?

PR #2413, opened 2019-10-18
Drop down list of WC transactional emails [MAILPOET-2282]

amine-mp:


lib/WooCommerce/TransactionalEmails.php
```diff
@@ -34,21 +54,27 @@ public function init() {
     }
   }
 
-  public function getEmailHeading() {
-    $default_heading = __('Thank you for your order', 'woocommerce');
-    $settings = $this->wp->getOption('woocommerce_customer_processing_order_settings');
-    if (!$settings) {
-      return $default_heading;
+  public function getEmailHeadings() {
```
pavel-mailpoet: Do you think this should have a unit test?

views/newsletter/templates/blocks/woocommerceContent/new_account.hbs
```diff
@@ -0,0 +1,14 @@
...
+    .mailpoet_editor_view_{{ viewCid }} .mailpoet_content h1,
+    .mailpoet_editor_view_{{ viewCid }} .mailpoet_content h2,
+    .mailpoet_editor_view_{{ viewCid }} .mailpoet_content h3 {
+        color: {{ styles.titleColor }};
+    }
+</style>
+<div class="mailpoet_content mailpoet_woocommerce_content" data-automation-id="woocommerce_content">
+<p style="margin:0 0 16px"><%= __('Hi %s,', 'woocommerce') | format('Elon') %></p>
+<p style="margin:0 0 16px"><%= __('Thanks for creating an account on %1$s. Your username is %2$s. You can access your account area to view orders, change your password, and more at: %3$s', 'woocommerce') | format('{{siteName}}', '<strong>elon.musk</strong>', '<a href="http://{{siteAddress}}" style="color:#fe5301;font-weight:normal;text-decoration:underline" target="_blank">{{siteAddress}}</a>') | raw %></p>
+<p style="margin:0 0 16px"><%= __('We look forward to seeing you soon.', 'woocommerce') %>.</p>
```
pavel-mailpoet: I think there is a redundant `.` at the end.

tests/integration/WooCommerce/TransactionalEmailsTest.php
```diff
@@ -55,6 +55,30 @@ function testInitUsesImageFromWCSettings() {
     expect($email->body)->contains('my-awesome-image-url');
   }
 
+  function testGetEmailHeadings() {
```
pavel-mailpoet: I have two questions: 
- Why is this an integration test and not a unit test? 
- Why did you include both changes into a single commit, those are not related changes, are they?

amine-mp: - This is an integration test because it needs to call some WP functions internally (`wp_specialchars_decode` and `wp_parse_url`).
- You are right, my bad. I should have separated the typo fix into its own commit.

pavel-mailpoet: > This is an integration test because it needs to call some WP functions internally (wp_specialchars_decode and wp_parse_url).
 
Didn't we create the `WPFunctions` class for this exact reason? To be able to mock WP calls ?

codemonkey-jack:
@amine-mp Can you fix the conflicts, please? I can then test. :)

Thanks!

amine-mp:
@codemonkey-jack I fixed the conflict.

codemonkey-jack:
@amine-mp Thanks! This is now with Quentin in product QA.

PR #2418, opened 2019-10-23
Black friday banner [MAILPOET-2469]

pavel-mailpoet:


lib/Util/Notices/PermanentNotices.php
```diff
@@ -61,10 +61,8 @@ public function init() {
     $this->inactive_subscribers_notice->init(
       Menu::isOnMailPoetAdminPage($exclude = ['mailpoet-welcome-wizard'])
     );
-    $this->discounts_announcement->init(
-      empty($_GET['page'])
-      && $this->wp->isAdmin()
-      && strpos($_SERVER['SCRIPT_NAME'], 'wp-admin/index.php') !== false
+    $this->black_friday_notice->init(
+      Menu::isOnMailPoetAdminPage($exclude = ['mailpoet-welcome-wizard'])
```
costasovo: I think we might want to exclude also pages
`mailpoet-woocommerce-list-import` and `mailpoet-revenue-tracking-permission` possibly also for older notices. These two are displayed to users with WooCommerce right after the wizard.

lib/Util/Notices/BlackFridayNotice.php
```diff
@@ -0,0 +1,43 @@
...
+use MailPoet\WP\Notice as WPNotice;
+
+class BlackFridayNotice {
+
+  const OPTION_NAME = 'dismissed-black-friday-notice';
+  const DISMISS_NOTICE_TIMEOUT_SECONDS = 2592000; // 30 days
+
+  function init($should_display) {
+    $should_display = $should_display
+      && (time() <= strtotime('2019-11-30 23:59:59'))
```
codemonkey-jack: @pavel-mailpoet Should we not use the current date/time of the WordPress install itself and check against that using the core function? https://codex.wordpress.org/Function_Reference/current_time

It'd ensure there can be no bugs with various date/time formats. What do you think?

pavel-mailpoet: @codemonkey-jack The time format is not an issue here, because both `time` and `strtotime` return the same format. See here:

- https://www.php.net/manual/en/function.time.php
- https://www.php.net/manual/en/function.strtotime.php

PR #2419, opened 2019-10-23
Refactor emoji encoding/decoding out of models [MAILPOET-2460]

wxa:
Since emoji handling is now decoupled from models, we need to put it anywhere newsletter source/rendered body is saved to the database or retrieved for rendering. Hope I didn't miss such cases.


codemonkey-jack:
@wxa All good! Just needs conflicts resolving and then I can merge. :)

wxa:
@codemonkey-jack Conflicts resolved.

PR #2420, opened 2019-10-24
Fix saving options for WooCommerce product/category emails [MAILPOET-2499]

JanJakes:
[MAILPOET-2499]

[MAILPOET-2499]: https://mailpoet.atlassian.net/browse/MAILPOET-2499

### DIFF
--- a/assets/js/src/newsletters/automatic_emails/listings.jsx
+++ b/assets/js/src/newsletters/automatic_emails/listings.jsx
@@ -220,10 +220,15 @@ class Listings extends React.Component {
 
   renderSettings = (newsletter) => {
     const event = automaticEmails[newsletter.options.group].events[newsletter.options.event];
-    const meta = newsletter.options.meta || {};
-    const metaOptionValues = (meta.option) ? _.pluck(meta.option, 'name') : [];
+    let meta;
+    try {
+      meta = JSON.parse(newsletter.options.meta || null);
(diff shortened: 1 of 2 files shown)


lib/API/JSON/v1/Newsletters.php
```diff
@@ -537,7 +537,7 @@ function listing($data = []) {
         $queue
       );
 
-      $data[] = $this->wp->applyFilters('mailpoet_api_newsletters_listing_item', $newsletter->asArray());
```
pavel-mailpoet: I don't think we should remove this filter. Someone can be still using it, don't you think?

JanJakes: @pavel-mailpoet Hmm, you mean third party plugins? Because according to Git history this filter was added only to perform that side-effect. Honestly, I'd like to get rid of all filters that can modify our API data or anything. It is a very ugly and dangerous way of extending stuff. I understand when a filter changes some parameters but whole API data, I don't know.

Back to this filter - there's no way I could ensure nobody's using it. Do you think we have to keep it?

pavel-mailpoet: It could be a 3rd party plugin, or it could be a customer who can have a piece of code that is using it. My understanding is we cannot remove any filters because we don't know who is using them.

PR #2425, opened 2019-10-28
Remove new premium page feature flag [MAILPOET-2493]

pavel-mailpoet:
[MAILPOET-2493]

[MAILPOET-2493]: https://mailpoet.atlassian.net/browse/MAILPOET-2493


assets/css/src/components/_premiumPage.scss
```diff
@@ -10,6 +10,7 @@
   }
 
   .button.mailpoet-button-bigger {
+    line-height: normal;
```
wxa: To me the buttons' text looks better vertically aligned without this line.

pavel-mailpoet: I will take a look. I am using the new WP 5.3 with the new theme Twenty Twenty. It was broken in this theme, that is why I had to add the line.  But with this line it is broken in other themes.

wxa:
The buttons are still ugly, no difference. It's the screenshot from WP 5.2.4. @qfrery Is that good for you?

image

P.S.: By the way, why do you say the look depends on WP version? The old Premium page had a copy of WP CSS rules for its own use to not be dependent. Otherwise any future WP update can wreak havoc to it.

pavel-mailpoet:
@wxa it is dependent. If I switch to master and use the new theme Twenty Twenty from WP 5.3 the buttons look like this: https://d.pr/i/ZQv95k

Note that it is the old premium page. So the problem is not related to this PR.

wxa:
> So the problem is not related to this PR.

Alright, the feature flag is removed. But the button look should be fixed, preferably for all WP versions. I'll make a new ticket.

codemonkey-jack:
> > So the problem is not related to this PR.
> 
> Alright, the feature flag is removed. But the button look should be fixed, preferably for all WP versions. I'll make a new ticket.

@wxa I'm going to merge this as you're creating a new ticket for the button issue. Once created, please share a link here so we have a trace to it from this original PR. 

Thanks!

wxa:
@codemonkey-jack The button ticket: https://mailpoet.atlassian.net/browse/MAILPOET-2505

PR #2427, opened 2019-10-29
Dynamic segments with missing woocommerce database [MAILPOET-2486]

pavel-mailpoet:
[MAILPOET-2486]

[MAILPOET-2486]: https://mailpoet.atlassian.net/browse/MAILPOET-2486

### DIFF
--- a/lib/DynamicSegments/RequirementsChecker.php
+++ b/lib/DynamicSegments/RequirementsChecker.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace MailPoet\DynamicSegments;
+
+use MailPoet\DynamicSegments\Filters\Filter;
+use MailPoet\Models\DynamicSegment;
+use MailPoet\WooCommerce\Helper;
+
+class RequirementsChecker {
+
+  /** @var Helper */
+  private $woocommerce_helper;
+
+  function __construct(Helper $woocommerce_helper = null) {
+    if (!$woocommerce_helper) {
+      $woocommerce_helper = new Helper();
+    }
+    $this->woocommerce_helper = $woocommerce_helper;
+  }
+
+  function shouldSkipSegment(DynamicSegment $segment) {
+    foreach ($segment->getFilters() as $filter) {
+      if ($this->shouldSkipFilter($filter)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private function shouldSkipFilter(Filter $filter) {
+    if ($this->woocommerce_helper->isWooCommerceActive()) {
+      return false;
+    }
+    $class = get_class($filter);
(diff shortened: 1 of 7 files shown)


tests/unit/DynamicSegments/RequirementsCheckerTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+  /** @var Helper|\PHPUnit_Framework_MockObject_MockObject */
+  private $woocommerce_helper;
+
+  /** @var RequirementsChecker */
+  private $requirements_checker;
+
+  function _before() {
+    parent::_before();
+    if (!defined('MP_SEGMENTS_TABLE')) define('MP_SEGMENTS_TABLE', '');
```
JanJakes: Why is this needed? Isn't it a forgotten hotfix/workaround for something?

tests/unit/DynamicSegments/RequirementsCheckerTest.php
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  function testShouldntBlockSegmentIfWooCommerceIsActive() {
+    $this->woocommerce_helper->method('isWooCommerceActive')->willReturn(true);
+    $segment = DynamicSegment::create();
+    $segment->setFilters([new WooCommerceCategory(1)]);
+    expect($this->requirements_checker->shouldSkipSegment($segment))->false();
+  }
+
+  function testShouldntBlockWooCommerceSegmentIfWooCommerceIsInactive() {
```
JanJakes: The name - `testShouldnt` or `testShould`? I think the latter in this case, right?

lib/DynamicSegments/RequirementsChecker.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+    return false;
+  }
+
+  private function shouldSkipFilter(Filter $filter) {
+    if ($this->woocommerce_helper->isWooCommerceActive()) {
+      return false;
+    }
+    $class = get_class($filter);
+    return strpos($class, 'WooCommerce') !== false;
```
JanJakes: This is a bit risky in case of some refactors, etc. Could we make use of https://github.com/mailpoet/mailpoet/blob/master/lib/DynamicSegments/Filters/WooCommerceCategory.php#L10? I know constants are not enforceable by interface but we can either add a `getSegmentType` method to the interface or at least check if the constant exists.

tests/unit/DynamicSegments/RequirementsCheckerTest.php
pavel-mailpoet: This is needed. If this line is not here and the Segments model has loaded PHP complaints this constant is undefined.

PR #2428, opened 2019-10-29
Add missing translation domains [MAILPOET-2494]

pavel-mailpoet:
[MAILPOET-2494]

[MAILPOET-2494]: https://mailpoet.atlassian.net/browse/MAILPOET-2494


wxa:
@codemonkey-jack The code is ok, please check that translations are displayed now.

codemonkey-jack:
@pavel-mailpoet This fails QA due to the untranslated notice still displaying. Turns out we also have this file: `lib/Util/Notices/UnauthorizedEmailNotice.php` which is also missing the text domain from the strings. 

Please add the `mailpoet` text domain in the above file and then this will work. 

Thanks!

PR #2430, opened 2019-10-29
Remove AspectMock [MAILPOET-2464]

wxa:


pavel-mailpoet:
@wxa From the Jira specification:

> simplify PHPStan RoboTask

I don't see that. Am I missing something?

wxa:
You're right, added PHPStan changes.

PR #2433, opened 2019-10-31
Refactor CronHelper to a service [MAILPOET-2459]

wxa:


lib/Cron/CronHelper.php
```diff
@@ -19,20 +18,29 @@ class CronHelper {
...
-  static function getDaemonExecutionLimit() {
-    $wp = Functions::get();
-    $limit = $wp->applyFilters('mailpoet_cron_get_execution_limit', self::DAEMON_EXECUTION_LIMIT);
+  /** @var SettingsController */
+  private $settings;
+
+  /** @var WPFunctions */
+  private $wp;
+
+  public function __construct(SettingsController $settings, WPFunctions $wp) {
```
JanJakes: Just one question - can some of the methods in this class be made `private`? Or do we need all of them `public`?

lib/Cron/Supervisor.php
```diff
@@ -6,34 +6,41 @@ class Supervisor {
...
-  function __construct() {
-    $this->token = CronHelper::createToken();
+  /** @var CronHelper */
+  private $cron_helper;
+
+  function __construct(CronHelper $cron_helper) {
+    $this->cron_helper = $cron_helper;
+  }
+
+  function init() {
```
JanJakes: This can make the service "invalid" when injected somewhere if `init` wasn't called. I have an idea:
- Instead of `$this->daemon` use `$this->getDaemon()` inside this class - it already kind of behaves as a lazy property, maybe `$this->daemon` is not needed at all. Or... it can be lazy.
- Make token lazy as well (with private getter method), if needed (= if you think it shouldn't be generated in constructor).

tests/integration/Cron/SupervisorTest.php
```diff
@@ -21,51 +22,53 @@ function _before() {
     $this->settings->set('cron_trigger', [
       'method' => 'none',
     ]);
+    $this->cron_helper = ContainerWrapper::getInstance()->get(CronHelper::class);
+    $this->supervisor = ContainerWrapper::getInstance()->get(Supervisor::class);
```
JanJakes: Just so you know, you can also use `$this->di_container->get(...)`. No need to change it now...

lib/Cron/CronHelper.php
wxa: Currently they all need to be public: all but two are referenced in other classes, and the two remaining (`queryCronUrl`, `getSiteUrl`) are used in tests.

lib/Cron/Supervisor.php
wxa: I've introduced `init` because there was a problem with `Triggers` tests sharing the state of this class. I'll just use `setShared(false)` in the DI container and restore the original constructor.

codemonkey-jack:
@wxa Could you resolve the conflicts, please? I'll take a look after they are resolved.

Thanks!

wxa:
@codemonkey-jack Fixed the conflict - please check that the Linux Cron is working fine.

JanJakes:
@wxa This is still (or again?) showing conflicts here.

wxa:
@codemonkey-jack It had conflicts again with another merged branch. Rebased and fixed.

PR #2436, opened 2019-11-04
Form Editor #1 [MAILPOET-2450]

costasovo:
[MAILPOET-2450]

[MAILPOET-2450]: https://mailpoet.atlassian.net/browse/MAILPOET-2450


assets/js/src/form_editor/components/notices.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { NoticeList } from '@wordpress/components';
+import { useDispatch, useSelect } from '@wordpress/data';
+
+export default () => {
+  const dismissibleNotices = useSelect(
+    (select) => select('mailpoet-form-editor').getNotices(true),
```
pavel-mailpoet: I don't think using boolean argument here is a good idea. From here I don't know what it does. Why don't you create two functions `getDismissibleNotices` and `getAllNotices`?

assets/js/src/form_editor/store/actions.jsx
```diff
@@ -28,3 +28,20 @@ export function saveFormDone(result) {
     result,
   };
 }
+
+export function addNotice(content, status, isDismissible = false, id = null) {
```
pavel-mailpoet: The same comment goes here, why would you use a boolean argument instead of two different actions?

assets/js/src/form_editor/store/actions.jsx
```diff
@@ -45,3 +36,13 @@ export function removeNotice(id) {
     id,
   };
 }
+
+export function* saveForm() {
+  yield {
+    type: 'SAVE_FORM_STARTED',
+  };
+  yield removeNotice('save-form');
```
pavel-mailpoet: I am not sure if it is a good idea to have this here. Why does `saveForm` need to know anything about notices and if they should or should not be removed? Should something else react on `SAVE_FORM_STARTED` action and hide notice if needed?

assets/js/src/form_editor/store/actions.jsx
pavel-mailpoet: I am not sure why is the `id` argument optional.

views/forms.html
```diff
@@ -61,6 +61,7 @@
     'numberOfItemsMultiple': __('%$1d items'),
 
     'formName': __('Name'),
+    'noName': _x('no name', 'fallback for forms without a name in a form list'),
```
pavel-mailpoet: Did you consult this with Quentin? This seems quite odd to me.

assets/js/src/form_editor/components/form_title.jsx
```diff
@@ -0,0 +1,34 @@
...
+        <div>
+          <label htmlFor="post-title" className="screen-reader-text">
+            Add form name
+          </label>
+          <textarea
+            id="form-title"
+            className="editor-post-title__input"
+            placeholder="Add form name"
+            rows="1"
+            style={{
```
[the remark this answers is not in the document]
costasovo: I copypasted those from the component from Gutenberg. I'll check if they are really needed because our component is simpler than the one used in Gutenberg.

assets/js/src/form_editor/store/index.jsx
```diff
@@ -0,0 +1,14 @@
...
+
+const config = {
+  reducer,
+  actions,
+  selectors,
+  controls: {},
+  resolvers: {},
+};
+
+export default () => (registerStore('mailpoet-form-editor', config));
```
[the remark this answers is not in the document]
costasovo: Function [select](https://www.npmjs.com/package/@wordpress/data#select) expects one parameter the store name so I would not add the suffix `-store`.
I was thinking about using constant but it means importing it everywhere 🤔. I'll take a look.

assets/js/src/form_editor/store/actions.jsx
```diff
@@ -1,14 +1,14 @@
...
-  },
-  changeFormName(name) {
-    return {
-      type: 'CHANGE_FORM_NAME',
-      name,
-    };
-  },
-};
+
+export function toggleSidebar(toggleTo) {
```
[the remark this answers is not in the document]
costasovo: For actions) I would keep then in one file or maybe later divide them to some groups. That is how they [are organized within Gutenberg repo](https://github.com/WordPress/gutenberg/blob/master/packages/edit-post/src/store/actions.js). Action function (in most cases) just returns a simple object. I find more convenient to have it in one place.

For reducers) I think we will eventually separate them and use [combineReducers](https://www.npmjs.com/package/@wordpress/data#combineReducers). I just thought
that for now, it is not necessary since we have just a few simple reducers.

assets/js/src/form_editor/store/actions.jsx
costasovo: Good point. It would be probably better to move this to a reducer.

assets/js/src/form_editor/store/actions.jsx
costasovo: It is just for convenience if you don't pass any it will be generated, but this is not needed for now. This notification system will evolve as we add new notifications so I'll make it non-optional for now.

views/forms.html
costasovo: I've done it based on [the specs](https://mailpoet.atlassian.net/browse/MAILPOET-2450) see  B2.
It is similar to Gutenberg. When you save a post without a title it shows (no title) in the list. I think it makes sense.

PR #2448, opened 2019-11-07
Use MailPoet template for WC emails [MAILPOET-2286]

amine-mp:
[MAILPOET-2286]

[MAILPOET-2286]: https://mailpoet.atlassian.net/browse/MAILPOET-2286


tests/integration/WooCommerce/TransactionalEmailsTest.php
```diff
@@ -49,13 +52,76 @@ function testInitUsesImageFromWCSettings() {
...
+    $transactional_emails->useTemplateForWoocommerceEmails();
+    expect($added_actions)->count(1);
+    expect($added_actions['woocommerce_init'])->isCallable();
+    $added_actions['woocommerce_init']();
+    expect($removed_actions)->count(2);
+    expect($added_actions)->count(4);
+    expect($added_actions['woocommerce_email_header'])->isCallable();
+    ob_start();
+    $added_actions['woocommerce_email_header']('heading text');
+    expect(ob_get_clean())->equals('HTML before content with heading text');
```
pavel-mailpoet: Working with output buffering in tests is a kind of code smell. It seems like the class has more responsibilities than it should have. I think that calls for separating the code into two different classes. One for the logic that only returns the values and the other one that uses the first one, doesn't have any logic and only outputs it.

tests/integration/_bootstrap.php
```diff
@@ -216,4 +216,21 @@ function __unset($name) {
   }
 }
 
+if (!function_exists('WC')) {
+  class WC_Mailer { // phpcs:ignore
```
pavel-mailpoet: This code seems a bit surprising to me. We've been working with woocommerce for a long time and we didn't need this. Why do we need it now?

lib/WooCommerce/TransactionalEmails/Renderer.php
```diff
@@ -0,0 +1,65 @@
...
+
+  public function getHTMLAfterContent() {
+    if (empty($this->html_after_content)) {
+      throw new \Exception("You should call 'render' before 'getHTMLAfterContent'");
+    }
+    return '</div>' . $this->html_after_content;
+  }
+
+  public function prefixCss($css) {
+    $parser = new csstidy();
```
pavel-mailpoet: Shouldn't this be instantiated using DI?

lib/WooCommerce/TransactionalEmails/Renderer.php
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public function render(Newsletter $newsletter) {
+    $renderer = new NewsletterRenderer($newsletter, true);
+    $html = explode(Preprocessor::WC_CONTENT_PLACEHOLDER, $renderer->render('html'));
+    $this->html_before_content = $html[0];
+    $this->html_after_content = $html[1];
+  }
+
+  public function getHTMLBeforeContent($heading_text) {
```
pavel-mailpoet: I must admit this class looks quite odd to me. All methods are public. It is a one-shot class... You can instantiate it once, call render (which doesn't return anything) and after that using getters pull pieces of rendered code. And the `prefixCss` is here for an unknown reason... 

Could this be done as a reusable service?

lib/WooCommerce/TransactionalEmails.php
```diff
@@ -61,13 +66,46 @@ public function getEmailHeadings() {
...
+      $this->wp->addAction('woocommerce_email_footer', function() {
+        echo $this->renderer->getHTMLAfterContent();
+      });
+      $this->wp->addAction('woocommerce_email_styles', [$this->renderer, 'prefixCss']);
+    });
+  }
+
+  private function createNewsletter() {
+    $wc_email_settings = $this->getWCEmailSettings();
+    return Newsletter::createOrUpdate([
```
pavel-mailpoet: Could we use Doctrine for this? We are transitioning towards it but if we keep using the old models we will never finish.

amine-mp: This PR is already big, I would make it even bigger by refactoring `Newsletter::createOrUpdate` into Doctrine. What about creating a separate task for that?

pavel-mailpoet: I don't like this approach. This new task would take a while before it starts. This will go through QA and product QA. 

If we keep doing this we will never completely switch to Doctrine and we've already seen problems with the old ORM. We cannot just use the old models and keep creating tasks for refactoring.

This PR is not THAT big. It seems big because there was a big piece of code moved.

lib/WooCommerce/TransactionalEmails.php
```diff
@@ -94,12 +100,12 @@ public function useTemplateForWoocommerceEmails() {
...
-    return Newsletter::createOrUpdate([
-      'type' => Newsletter::TYPE_WC_TRANSACTIONAL_EMAIL,
-      'subject' => 'WooCommerce Transactional Email',
-      'preheader' => '',
-      'body' => json_encode($this->template->create($wc_email_settings)),
-    ]);
+    $newsletter = new NewsletterEntity;
+    $newsletter->setType(NewsletterEntity::TYPE_WC_TRANSACTIONAL_EMAIL);
+    $newsletter->setSubject('WooCommerce Transactional Email');
+    $newsletter->setBody(json_encode($this->template->create($wc_email_settings)));
```
pavel-mailpoet: You don't have to run `json_encode` here, the type on the entity is JSON. I wonder how does that even work.

tests/integration/WooCommerce/TransactionalEmailsTest.php
```diff
@@ -52,7 +60,13 @@ function testInitUsesImageFromWCSettings() {
...
-    $transactional_emails = new TransactionalEmails($wp, $this->settings, new Template, new Renderer);
+    $transactional_emails = new TransactionalEmails(
+      $wp,
+      $this->settings,
+      ContainerWrapper::getInstance()->get(Template::class),
+      ContainerWrapper::getInstance()->get(Renderer::class),
+      ContainerWrapper::getInstance()->get(NewslettersRepository::class)
+    );
     $transactional_emails->init();
     $email = Newsletter::where('type', Newsletter::TYPE_WC_TRANSACTIONAL_EMAIL)->findOne();
```
pavel-mailpoet: Why did you leave the old model here?

PR #2451, opened 2019-11-12
WP mailer fallback [MAILPOET-2515]

JanJakes:


lib/Mailer/WordPress/WordPressMailer.php
```diff
@@ -31,20 +36,30 @@ function send() {
     // Prepare everything (including the message) for sending.
     $this->preSend();
 
+    $extra_params = [
+      'meta' => $this->mailerMetaInfo->getWordPressTransactionalMetaInfo(),
+    ];
+    $email = $this->getEmail();
+    $address = $this->formatAddress($this->getToAddresses());
+
+    $send_with_mailer = function ($mailer) use ($email, $address, $extra_params) {
```
costasovo: Why did you use an anonymous function instead of a private class method? I think that the private method would be more readable + easy to click through the name to the definition in IDE. Not a big deal I just wonder...

JanJakes: @costasovo Because `use` binds three parameters in a "fixed" way while only one remains "passable" which I thought was nice. Do you think it is a bad idea?

costasovo: Yeah, the private method would require 4 parameters. I don't think is a bad idea. I would personally go with a private method because of navigation in IDE but as I said it is not a big deal. I was just curious.

JanJakes: If you mean jump-to-definition, that works in PHPStorm. If the method is used just below then I guess it's OK. I can also change it to a private method, I have no preference.

PR #2457, opened 2019-11-20
Shortcodes inside links are URL encoded when adding GA params [MAILPOET-2540]

amine-mp:
[MAILPOET-2540]

[MAILPOET-2540]: https://mailpoet.atlassian.net/browse/MAILPOET-2540

### DIFF
--- a/lib/Statistics/GATracking.php
+++ b/lib/Statistics/GATracking.php
@@ -44,11 +44,7 @@ class GATracking {
 
   private function addParams($extracted_links, $ga_campaign, $internal_host = null) {
     $processed_links = [];
-    $params = [
-      'utm_source' => 'mailpoet',
-      'utm_medium' => 'email',
-      'utm_campaign' => urlencode($ga_campaign),
-    ];
+    $params = 'utm_source=mailpoet&utm_medium=email&utm_campaign=' . urlencode($ga_campaign);
     $internal_host = $internal_host ?: parse_url(home_url(), PHP_URL_HOST);
     $internal_host = $this->secondLevelDomainNames->get($internal_host);
     foreach ($extracted_links as $extracted_link) {
@@ -58,7 +54,9 @@ class GATracking {
         // Process only internal links (i.e. pointing to current site)
         continue;
       }
-      $processed_link = WPFunctions::get()->addQueryArg($params, $extracted_link['link']);
+      list($path, $search, $hash) = $this->splitLink($extracted_link['link']);
+      $search = empty($search) ? $params : $search . '&' . $params;
+      $processed_link = $path . '?' . $search . ($hash ? '#' . $hash : '');
       $link = $extracted_link['link'];
       $processed_links[$link] = [
         'type' => $extracted_link['type'],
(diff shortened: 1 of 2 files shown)


lib/Statistics/GATracking.php
```diff
@@ -59,6 +59,7 @@ private function addParams($extracted_links, $ga_campaign, $internal_host = null
         continue;
       }
       $processed_link = WPFunctions::get()->addQueryArg($params, $extracted_link['link']);
+      $processed_link = str_replace(['%5B', '%3A', '%5D'], ['[', ':', ']'], $processed_link);
```
JanJakes: Replacing all `%3A` with a `:` will break many valid links (same for `[` and `]`). This is a very dangerous change to apply on the whole link.

I'd suggest looking at https://github.com/mailpoet/mailpoet/pull/2457/files#diff-75a68118206e3c038e3933dbd423c024R50 - you can check if `$ga_campaign` contains a shortcode (using a regex) and if yes, then you can urlencode only the part that is not the shortcode. Something like this: https://3v4l.org/0ZEroD (Although I'm not sure if `$shortcode_regex` is good like this or should be something stricter.)

amine-mp: it's the `$extracted_link['link']` that contains a shortcode not `$ga_campaign`. When `addQueryArg` is executed, it url_encodes the link, I am trying to undo the encoding on shortcodes because they will be replaced later with their values.
Can you please explain why is it dangerous to do it that way?

JanJakes: @amine-mp As I said:

> Replacing all `%3A` with a `:` will break many valid links (same for `[` and `]`).

You will simply break all valid links where users themselves have `:`, `[`, and `]`. We really can't do that. There are many valid links containing these symbols and `[` and `]` are even used for array parameters.

> I am trying to undo the encoding on shortcodes because they will be replaced later with their values.

I didn't look deep into `add_query_arg` but if it has such a side-effect that it encodes some of the previous stuff in the link then we should not _undo_ it but avoid it. The thing that it modifies the extracted link is _the_ bug but instead of fixing it you're trying to undo it after that. There are many ways of just not having the bug in the first place. Something like:

```
processed_link .= (strpos($extracted_link['link'], '?') === false ? '?' : '&') . http_build_query($params);
```
You may say what to do if the link already has UTM params, etc. Well, I don't know, it's an edge-case I guess. I just think we shouldn't convert the original params in any way. That would be a side-effect of this service even if it didn't break anything.

lib/Statistics/GATracking.php
```diff
@@ -58,7 +54,9 @@ private function addParams($extracted_links, $ga_campaign, $internal_host = null
         // Process only internal links (i.e. pointing to current site)
         continue;
       }
-      $processed_link = WPFunctions::get()->addQueryArg($params, $extracted_link['link']);
+      list($path, $search, $hash) = $this->splitLink($extracted_link['link']);
+      $search = empty($search) ? $params : $search . '&' . $params;
+      $processed_link = $path . '?' . $search . '#' . $hash;
```
JanJakes: I'm afraid this will add `#` at the end of every link without `$hash`. Could you handle that?

lib/Statistics/GATracking.php
```diff
@@ -68,4 +66,13 @@ private function addParams($extracted_links, $ga_campaign, $internal_host = null
     }
     return $processed_links;
   }
+
+  private function splitLink($link) {
+    $parts = explode('#', $link);
+    $hash = implode('#', array_slice($parts, 1));
```
JanJakes: Btw. there is also `explode('#', $link, 2)` which will split a string to max. 2 parts by the first occurrence of `#`. I guess no need to change it now but it's a handy parameter to know about.

amine-mp: Cool, I didn't know about it.

PR #2458, opened 2019-11-20
Cron refactor [MAILPOET-2538]

JanJakes:
[MAILPOET-2538]

[MAILPOET-2538]: https://mailpoet.atlassian.net/browse/MAILPOET-2538


lib/Cron/CronTrigger.php
```diff
@@ -22,17 +21,34 @@ class CronTrigger {
   const DEFAULT_METHOD = 'WordPress';
   const SETTING_NAME = 'cron_trigger';
 
-  function __construct(SettingsController $settings) {
+  /** @var MailPoet */
+  private $mailoet_trigger;
```
costasovo: Typo in the property and variable (in the constructor).

PR #2459, opened 2019-11-20
Trigger Cron workers only when needed [MAILPOET-2530]

amine-mp:
[MAILPOET-2530]

[MAILPOET-2530]: https://mailpoet.atlassian.net/browse/MAILPOET-2530

### DIFF
--- a/lib/Cron/Workers/SimpleWorker.php
+++ b/lib/Cron/Workers/SimpleWorker.php
@@ -47,15 +47,18 @@ abstract class SimpleWorker {
 
     // abort if execution limit is reached
     $this->cron_helper->enforceExecutionLimit($timer);
+    $scheduled_tasks = $this->getDueTasks();
+    $running_tasks = $this->getRunningTasks();
 
     if (!$this->checkProcessingRequirements()) {
+      foreach (array_merge($scheduled_tasks, $running_tasks) as $task) {
+        $task->delete();
+      }
       return false;
     }
 
     $this->init();
 
-    $scheduled_tasks = $this->getDueTasks();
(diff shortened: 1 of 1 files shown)


lib/Cron/Workers/SimpleWorker.php
```diff
@@ -42,6 +42,11 @@ function init() {
 
   function process() {
     if (!$this->checkProcessingRequirements()) {
+      ScheduledTask::where('type', static::TASK_TYPE)
+        ->whereLte('scheduled_at', Carbon::createFromTimestamp($this->wp->currentTime('timestamp')))
+        ->whereNull('deleted_at')
+        ->whereNotEqual('status', ScheduledTask::STATUS_COMPLETED)
+        ->deleteMany();
```
JanJakes: I just came across this and I have a question - could we reuse `$scheduled_tasks` and `$running_tasks` that are fetched just a few lines below and call `->delete()` on them? Because those two arrays are the exact tasks that cause the problems and because this is just a logic duplicity, moreover not a nice one since it's a SQL query, and not a nice one since the query is not targetting _exactly_ the same tasks.

We could fetch `$scheduled_tasks` and `$running_tasks` a bit earlier or save the `checkProcessingRequirements` to a variable and act later or both, to some extent.

amine-mp: I thought about that. But Is it possible to call `delete()` on the result of `findMany()`? or do we have to loop over them and delete one by one?

JanJakes: We can just loop over them, I think. In most cases there will probably be just one anyway.

PR #2460, opened 2019-11-24
Fine tune WC email customizer [MAILPOET-2520]

wxa:


wxa:
@amine-mp Updated the branch according to the changed specs, the only departure is a slightly different notice when WC is disabled, see my latest comment in the Jira task.

lib/AdminPages/Pages/NewsletterEditor.php
```diff
@@ -54,29 +54,38 @@ function render() {
...
       $woocommerce_template_id
       && $newsletter_id === $woocommerce_template_id
-      && (
-        !$this->woocommerce_helper->isWooCommerceActive()
-        || !(bool)$this->settings->get('woocommerce.use_mailpoet_editor', false)
-      )
+      && !$this->woocommerce_helper->isWooCommerceActive()
     ) {
-      header('Location: admin.php?page=mailpoet-settings&enable-customizer-notice#woocommerce', true, 302);
+      $location = 'admin.php?page=mailpoet-settings&enable-customizer-notice#woocommerce';
```
amine-mp: From the specs
> If the user tries to access the template editor URL, redirect him to `?page=mailpoet-newsletters` and ...

But you redirect to settings page here.

wxa: I guess it's a typo in specs, we changed them to redirect to the message which is in the Settings (see last 3 comments in the Jira task).

amine-mp: I see, that makes sense.

PR #2461, opened 2019-11-25
Form editor - Form settings [MAILPOET-2455]

costasovo:
[MAILPOET-2455]

[MAILPOET-2455]: https://mailpoet.atlassian.net/browse/MAILPOET-2455


assets/js/src/form_editor/components/form_settings/custom_css_panel.jsx
```diff
@@ -1,24 +1,33 @@
...
 import MailPoet from 'mailpoet';
 import { useSelect, useDispatch } from '@wordpress/data';
+import CodeMirror from './codemirror_wrap.jsx';
 
 export default () => {
   const styles = useSelect(
     (select) => select('mailpoet-form-editor').getFormStyles(),
     []
   );
+  const options = useRef({
```
pavel-mailpoet: I am not sure why do we need to use `useRef` here. Could you please explain?

assets/js/src/form_editor/components/form_settings/codemirror_wrap.jsx
```diff
@@ -0,0 +1,42 @@
...
+      autoComplete="off"
+    />
+  );
+};
+
+CodemirrorWrap.propTypes = {
+  value: PropTypes.string.isRequired,
+  onChange: PropTypes.func.isRequired,
+  options: PropTypes.shape({
+    lineNumbers: PropTypes.bool.isRequired,
```
pavel-mailpoet: Do all of these properties need to be required? Could we make them optional and use default values?

assets/js/src/form_editor/store/selectors.jsx
```diff
@@ -14,6 +14,15 @@ export default {
   getFormExports(state) {
     return state.formExports;
   },
+  getFormSettings(state) {
+    return state.formData.settings;
+  },
+  getSegments(state) {
+    return state.segments;
+  },
+  getPages(state) {
```
pavel-mailpoet: Could we rename those two properties? I was confused about what do they mean. I initially thought that it something that a user selected. What about the name `getAllAvailablePages` and `getAllAvailableSegments`?

views/form/editor.html
```diff
@@ -30,7 +32,12 @@
 'formSaved': __('Form saved.'),
 'addFormWidgetHint': __('Add this form to your sidebar or footer on the [link]Widgets page[/link].'),
 'addFormShortcodeHint': __('Copy and paste this [link]shortcode[/link] on to a post or page.'),
-'addFormPhpIframeHint': __('%sPHP%s and %siFrame%s versions are also available.')
+'addFormPhpIframeHint': __('%sPHP%s and %siFrame%s versions are also available.'),
+'settingsListLabel': __('This form adds the subscribers to these lists:'),
+'settingsAfterSubmit': __('After submit...'),
```
pavel-mailpoet: I think the interpunction should not be part of the translations. That way we couldn't be messed up by translators. And you should use … instead of ...

assets/js/src/form_editor/store/controls.jsx
```diff
@@ -6,9 +6,14 @@ export default {
     if (select('mailpoet-form-editor').getIsFormSaving()) {
       return;
     }
-    dispatch('mailpoet-form-editor').saveFormStarted();
-    const formData = select('mailpoet-form-editor').getFormData();
     dispatch('mailpoet-form-editor').removeNotice('save-form');
+    const formData = select('mailpoet-form-editor').getFormData();
+    if (!formData.settings.segments || formData.settings.segments.length === 0) {
```
pavel-mailpoet: I am not sure about this one. The responsibility of this control is not to validate, and so I think this should be moved to some other place. I am not sure if or where, but I think that validation should be done elsewhere.

assets/js/src/form_editor/components/form_settings/basic_settings_panel.jsx
```diff
@@ -59,10 +64,17 @@ export default () => {
...
     ? segments.filter((seg) => (settings.segments.includes(seg.id.toString())))
     : [];
+  const shouldDisplayMissingListError = missingListError && !selectedSegments.length;
   return (
     <Panel>
       <PanelBody title={MailPoet.I18n.t('formSettings')}>
-        <BaseControl label={MailPoet.I18n.t('settingsListLabel')}>
+        <BaseControl
+          label={MailPoet.I18n.t('settingsListLabel')}
+          className={shouldDisplayMissingListError ? 'mailpoet-form-missing-lists' : null}
```
pavel-mailpoet: `classnames`?

costasovo:
@pavel-mailpoet All should be fixed.

codemonkey-jack:
@costasovo This fails QA. 

In the specs it's written;

>When the user clicks on the Save button,

>If the list selected is empty, don’t save the form and display the following error message: Please select a list.

>Also, expand the sidebar section Settings, add a red note Please select a list. above the list selector and display the list selector input border in red.

The error handling is there, but if `Settings` is collapsed it doesn't expand to show the message. Based on the specs I believe it should expand if collapsed. Unless I'm mis-reading the specs.

assets/js/src/form_editor/store/reducers/toggleSidebarPanel.jsx
```diff
@@ -0,0 +1,20 @@
+import { remove } from 'lodash';
+
+export default (state, action) => {
+  const openedPanels = [...state.sidebar.openedPanels];
+  const isOpenedCurrent = openedPanels.includes(action.id);
+  const isOpenedFinal = action.isOpened !== undefined ? action.isOpened : !isOpenedCurrent;
```
pavel-mailpoet: Could we split this function into more functions, please?

codemonkey-jack:
LGTM! Now with Quentin for product QA.

PR #2463, opened 2019-11-26
Display a new limit warning for free users [MAILPOET-2394]

amine-mp:
[MAILPOET-2394]

[MAILPOET-2394]: https://mailpoet.atlassian.net/browse/MAILPOET-2394


lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+   * @ORM\Column(type="bigint")
+   * @var int|null
+   */
+  private $wp_user_id;
+
+  /**
+   * @ORM\Column(type="boolean")
+   * @var bool|null
+   */
+  private $is_woocommerce_user;
```
costasovo: This is not nullable in DB and has default value 0 so I guess we could set this property as `false` by default and remove null from the annotation.

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+
+  /**
+   * @ORM\Column(type="string")
+   * @var string
+   */
+  private $first_name;
```
costasovo: We have an empty string as a default value for `first_name` and `last_name` in DB. It is not nice but we should add it as default to the entity property as well.

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+  private $email;
+
+  /**
+   * @ORM\Column(type="string")
+   * @var string
+   */
+  private $status = self::STATUS_UNCONFIRMED;
+
+  /**
+   * @ORM\Column(type="string")
```
costasovo: It is nullable in DB so please add a nullable annotation for doctrine `@ORM\Column(type="string", nullble=true)`

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+   * @var string
+   */
+  private $subscribed_ip;
+
+  /**
+   * @ORM\Column(type="string")
```
costasovo: It is nullable in DB so please add a nullable annotation for doctrine @ORM\Column(type="string", nullble=true)

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+   * @var string|null
+   */
+  private $confirmed_ip;
+
+  /**
+   * @ORM\Column(type="datetimetz")
```
costasovo: It is nullable in DB so please add a nullable annotation for doctrine @ORM\Column(type="datetimetz", nullble=true)

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+   * @var DateTimeInterface|null
+   */
+  private $confirmed_at;
+
+  /**
+   * @ORM\Column(type="datetimetz")
```
costasovo: It is nullable in DB so please add a nullable annotation for doctrine @Orm\Column(type="datetimetz", nullble=true)

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+   * @var DateTimeInterface|null
+   */
+  private $last_subscribed_at;
+
+  /**
+   * @ORM\Column(type="text")
```
costasovo: It is nullable in DB so please add a nullable annotation for doctrine @Orm\Column(type="text", nullble=true)

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+   * @ORM\Column(type="string")
+   * @var string|null
+   */
+  private $source = 'unknown';
+
+  /**
+   * @ORM\Column(type="int")
+   * @var int|null
+   */
+  private $count_confirmations;
```
costasovo: This has default value 0 in DB so we should set it as default to property and remove `null` from the annotation. When changing annotation update also getter and setter.

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+   * @ORM\Column(type="text")
+   * @var string|null
+   */
+  private $unconfirmed_data;
+
+  /**
+   * @ORM\Column(type="string")
+   * @var string|null
+   */
+  private $source = 'unknown';
```
costasovo: It is nullable in DB so please add a nullable annotation for doctrine @Orm\Column(type="string", nullble=true)

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+
+  /**
+   * @ORM\Column(type="int")
+   * @var int|null
+   */
+  private $count_confirmations;
+
+  /**
+   * @ORM\Column(type="string")
```
costasovo: It is nullable in DB so please add a nullable annotation for doctrine @Orm\Column(type="string", nullble=true)

lib/Entities/SubscriberEntity.php
```diff
@@ -0,0 +1,350 @@
...
+   * @var string|null
+   */
+  private $unsubscribe_token;
+
+  /**
+   * @ORM\Column(type="string")
```
costasovo: It is nullable in DB so please add a nullable annotation for doctrine @Orm\Column(type="string", nullble=true)

lib/Util/License/Features/Subscribers.php
```diff
@@ -3,18 +3,37 @@
...
+
+  /** @var int */
+  private $installation_time;
+
+  /** @var int */
+  private $subscribers_count;
+
+  function __construct(SettingsController $settings) {
+    $has_mss_key = !empty($settings->get(Bridge::API_KEY_SETTING_NAME));
+    $has_premium_key = !empty($settings->get(Bridge::PREMIUM_KEY_SETTING_NAME));
```
costasovo: Please move this `$settings->get` and model calls from the constructor. We should use the constructor for passing dependencies. This code would access the database and this could be dangerous because DI container may instantiate this class before we safely check DB connection in Initializer.
You could just set $settings controller as private property and add `getLicence`. Settings are cached internally so accessing those is cheap. You could also cach `subscribers_count` later.
I also noticed that you added a repository `SubscribersRepository` with method `getTotalSubscribers`
Could you use this repository instead static call on old model?

amine-mp:
Thanks for taking time to write all the comments. I rebased the branch and fixed the typos. 4 commits were added

codemonkey-jack:
@amine-mp This fails QA. There's no warning displaying (admin notice) and it's letting me send.

I have 2085 confirmed subscribers. And MailPoet premium is not installed.

codemonkey-jack:
Now with Quentin in product QA.

PR #2464, opened 2019-11-26
Acceptance test on WP 5.3/PHP 7.3 [MAILPOET-2544]

costasovo:
[MAILPOET-2544]

[MAILPOET-2544]: https://mailpoet.atlassian.net/browse/MAILPOET-2544

### DIFF
--- a/tests/_support/AcceptanceTester.php
+++ b/tests/_support/AcceptanceTester.php
@@ -342,7 +342,7 @@ class AcceptanceTester extends \Codeception\Actor { // phpcs:ignore PSR1.Classes
    */
   public function selectPaymentMethod($method = 'cod') {
     $I = $this;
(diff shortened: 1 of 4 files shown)


tests/_support/AcceptanceTester.php
```diff
@@ -342,7 +342,7 @@ public function optOutOfSubscription() {
    */
   public function selectPaymentMethod($method = 'cod') {
     $I = $this;
-    $I->scrollTo('#payment_method_' . $method);
+    $I->scrollTo('#payment_method_' . $method, 0, -40);
```
pavel-mailpoet: This magic number needs to have some comment. Why do we need it? Why exactly 40?

costasovo:
I added comments.

PR #2466, opened 2019-11-27
Promote agency plan on Premium page [MAILPOET-2534]

JanJakes:
[MAILPOET-2534]

[MAILPOET-2534]: https://mailpoet.atlassian.net/browse/MAILPOET-2534

### DIFF
--- a/views/premium.html
+++ b/views/premium.html
@@ -172,6 +172,32 @@
   </div>
   <hr>
   <div class="mailpoet-premium-page-text">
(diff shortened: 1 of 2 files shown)


views/premium.html
```diff
@@ -184,6 +184,32 @@ <h1><%= _x('100% Satisfaction Guarantee', 'This text resides in the Premium page
     </p>
   </div>
   <hr>
+  <div class="mailpoet-premium-page-text">
```
amine-mp: According to the screenshot on specs, this should go before the "Try MailPoet Premium for 30 days" section.

JanJakes: Fixed.

PR #2475, opened 2019-12-04
Add First and Last Name blocks in the Form Editor [MAILPOET-2452]

pavel-mailpoet:
[MAILPOET-2452]

[MAILPOET-2452]: https://mailpoet.atlassian.net/browse/MAILPOET-2452

### DIFF
--- a/tests/acceptance/FormEditorAddNamesCest.php
+++ b/tests/acceptance/FormEditorAddNamesCest.php
@@ -37,6 +37,11 @@ class FormEditorAddNamesCest {
     $I->click('[data-automation-id="form_save_button"]');
     $I->waitForText('Form saved', 10, '.automation-dismissible-notices');
     $I->seeNoJSErrors();
+    // Reload page and check data were saved
+    $I->reloadPage();
+    $I->waitForElement('[data-automation-id="form_title_input"]');
+    $I->seeElement('[data-automation-id="editor_first_name_input"]');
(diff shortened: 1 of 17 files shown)


tests/acceptance/FormEditorAddNamesCest.php
```diff
@@ -0,0 +1,42 @@
...
+    $I->click('.editor-inserter__results .components-panel__body-toggle'); // toggle fields
+    $I->click('.editor-block-list-item-mailpoet-form-first-name-input'); // add first name block to the editor
+    $I->click('.block-list-appender button');// CLICK the big button that adds new blocks
+    $I->waitForElement('.editor-inserter__results .components-panel__body-toggle');
+    $I->click('.editor-inserter__results div:nth-child(2) .components-panel__body-toggle'); // toggle fields, get the second field, first one is now "Most Used"
+    $I->click('.editor-block-list-item-mailpoet-form-last-name-input'); // add last name block to the editor
+
+    $I->click('[data-automation-id="form_save_button"]');
+    $I->waitForText('Form saved', 10, '.automation-dismissible-notices');
+    $I->seeNoJSErrors();
```
costasovo: We should check that the last name was really added into the block editor area and after saving we could do a page reload and after reload check that it is still present to verify that it was really saved.

costasovo:
I noticed we have .only in on JS test. It is my fault 🤦‍♂ . Could you please add eslint check for that.

PR #2476, opened 2019-12-04
Transactional emails setting [MAILPOET-2536]

NeosinneR:
[MAILPOET-2536]

[MAILPOET-2536]: https://mailpoet.atlassian.net/browse/MAILPOET-2536

### DIFF
--- a/views/settings/advanced.html
+++ b/views/settings/advanced.html
@@ -165,7 +165,7 @@
     <tr>
       <th scope="row">
         <label>
-          <%= _x("Send all site's emails with...", 'Transational emails settings title') %>
+          <%= _x('Send all site’s emails with…', 'Transational emails settings title') %>
         </label>
         <p class="description">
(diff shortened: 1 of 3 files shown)


views/settings/advanced.html
```diff
@@ -161,6 +161,49 @@
...
+        <p>
+          <label>
+            <input
+              type="radio"
+              name="send_transactional_emails"
+              value="1"
+              <% if(settings.send_transactional_emails) %>
+              checked="checked"
+              <% endif %>
+            /><%= _x("MailPoet's active sending method (recommended)", 'Transactional emails settings option') %>
```
costasovo: In the `MailPoet's` you could use an apostrophe (`MailPoet’s`) instead of a single quote and then you can use single quotes for the whole string.  Similar in `Send all site’s emails with...` where you can change three dots ... to ellipsis …
I think that using apostrophe and ellipsis is typographically correct.

PR #2480, opened 2019-12-09
Form editor list selection [MAILPOET-2461]

pavel-mailpoet:
[MAILPOET-2461]

[MAILPOET-2461]: https://mailpoet.atlassian.net/browse/MAILPOET-2461


assets/js/src/form_editor/blocks/segment_select/settings.jsx
```diff
@@ -27,12 +27,30 @@ const Preview = ({ segments, updateSegment }) => {
...
-      key={segment.id}
-    />
+    <div className="mailpoet-form-segments-settings-list" key={segment.id}>
+      <CheckboxControl
+        label={segment.name}
+        checked={!!segment.isChecked}
+        onChange={partial(onCheck, segment.id)}
+        key={`check-${segment.id}`}
+      />
+      <span
```
costasovo: Maybe we could use [IconButton component](https://developer.wordpress.org/block-editor/components/icon-button/) provided by wordpress/components. Did you try it?

package.json
```diff
@@ -50,6 +50,7 @@
     "prop-types": "^15.6.2",
     "react": "^16.10.2",
     "react-autosize-textarea": "^7.0.0",
+    "react-beautiful-dnd": "^12.2.0",
```
costasovo: I looked at package-lock.json and this library looks pretty big. It requires e.g. whole redux. I googled a bit and found [another library](https://www.npmjs.com/package/react-dnd). It looks updated and has over 12k stars so it has similar popularity. Here is a demo for sortable list: https://react-dnd.github.io/react-dnd/examples/sortable/simple
It also provides custom hooks so it may work nicely with our codebase. What do you think?

PR #2481, opened 2019-12-09
WC email customizer changes [MAILPOET-2521]

wxa:


codemonkey-jack:
Now with Quentin in Product QA.

PR #2493, opened 2019-12-12
WC email customizer: sync WC settings [MAILPOET-2569]

wxa:


codemonkey-jack:
Looks good, now with Quentin in product QA.

PR #2497, opened 2019-12-16
Form editor add existing custom fields [MAILPOET-2453]

pavel-mailpoet:
[MAILPOET-2453]

[MAILPOET-2453]: https://mailpoet.atlassian.net/browse/MAILPOET-2453


assets/js/src/form_editor/blocks/custom_text/custom_field_settings.jsx
```diff
@@ -0,0 +1,93 @@
...
+  customFieldId,
+  updateAttributes,
+}) => {
+  const [localMandatory, setLocalMandatory] = useState(mandatory);
+  const [localValidate, setLocalValidate] = useState(validate);
+  const { saveCustomField } = useDispatch('mailpoet-form-editor');
+  const isSaving = useSelect(
+    (sel) => sel('mailpoet-form-editor').getIsCustomFieldSaving(),
+    []
+  );
```
costasovo: Maybe we should pass those two using props so that this component is reusable for user import etc.

lib/API/JSON/v1/CustomFields.php
```diff
@@ -45,7 +45,9 @@ function save($data = []) {
     }
     $custom_field = CustomField::findOne($custom_field->id);
```
costasovo: Maybe we could use the newly created response builder here. We are already re-fetching the model from DB...

lib/Entities/CustomFieldEntity.php
```diff
@@ -57,4 +58,25 @@ public function getType() {
...
+   * @param string $type
+   */
+  public function setType($type) {
+    $this->type = $type;
+  }
+
+  /**
+   * @param array $params
+   */
+  public function setParams($params) {
```
costasovo: We could add a typehint here.

assets/js/src/form_editor/store/blocks_to_form_body.jsx
```diff
@@ -2,15 +2,29 @@
...
     ...mappedCommonProperties,
     id: block.attributes.customFieldId.toString(),
     name: customField.name,
-    params: {
-      ...mappedCommonProperties.params,
-      validate: block.attributes.validate,
-    },
+    type: typesMap[block.name],
   };
+  if (Object.prototype.hasOwnProperty.call(block.attributes, 'validate')) {
```
costasovo: Maybe we could use `https://lodash.com/docs/4.17.15#has` instead of `Object.prototype.hasOwnProperty` here

PR #2499, opened 2019-12-16
Manage subscription page unsubscribes from hidden lists [MAILPOET-2537]

amine-mp:
[MAILPOET-2537]

[MAILPOET-2537]: https://mailpoet.atlassian.net/browse/MAILPOET-2537

### DIFF
--- a/lib/Subscription/Manage.php
+++ b/lib/Subscription/Manage.php
@@ -56,7 +56,9 @@ class Manage {
 
   private function addHiddenSegments(Subscriber $subscriber, array $data) {
     $shownSegments = $this->settings->get('subscription.segments');
-    if (empty($shownSegments)) return $data;
+    if (empty($shownSegments)) {
+      return $data;
+    }
     if (empty($data['segments'])) $data['segments'] = [];
     $subscriber->withSubscriptions();
     foreach ($subscriber->subscriptions as $subscription) {
--- a/lib/Subscription/Manage.php
+++ b/lib/Subscription/Manage.php
@@ -5,6 +5,7 @@ namespace MailPoet\Subscription;
 use MailPoet\Form\Util\FieldNameObfuscator;
 use MailPoet\Models\CustomField;
 use MailPoet\Models\Subscriber;
+use MailPoet\Settings\SettingsController;
 use MailPoet\Subscribers\LinkTokens;
 use MailPoet\Util\Url as UrlHelper;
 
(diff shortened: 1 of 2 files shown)


lib/Subscription/Manage.php
```diff
@@ -48,6 +54,19 @@ function onSave() {
     $this->url_helper->redirectBack();
   }
 
+  private function addHiddenSegments(Subscriber $subscriber, array $data) {
+    $shown_segments = $this->settings->get('subscription.segments');
+    if (empty($shown_segments)) return $data;
+    if (empty($data['segments'])) $data['segments'] = [];
```
JanJakes: What does this line do? The `$data` variable must be an array therefore `empty($data)` already is `[]`.

lib/Subscription/Manage.php
```diff
@@ -48,6 +54,19 @@ function onSave() {
     $this->url_helper->redirectBack();
   }
 
+  private function addHiddenSegments(Subscriber $subscriber, array $data) {
+    $shown_segments = $this->settings->get('subscription.segments');
+    if (empty($shown_segments)) return $data;
```
JanJakes: We don't have a sniff for this yet but the coding style should be:
```php
if (empty($shown_segments)) {
  return $data;
}
```

lib/Subscription/Manage.php
```diff
@@ -48,6 +54,19 @@ function onSave() {
...
+    if (empty($data['segments'])) $data['segments'] = [];
+    $subscriber->withSubscriptions();
+    foreach ($subscriber->subscriptions as $subscription) {
+      if ($subscription['status'] === Subscriber::STATUS_SUBSCRIBED && !in_array($subscription['segment_id'], $shown_segments)) {
+        $data['segments'][] = $subscription['segment_id'];
+      }
+    }
```
JanJakes: This is quite a hard to read logic and there's no comment to explain it. But I don't like it anyway - it seems we take data from user's form and then add some stuff to it from the DB (and only specific subscription status). Why don't we do naturally, the other way round? I think we should first fetch all subscription data from the DB, then apply user changes on it, then save it. That would be a more straightforward and safe way to do it.

lib/Subscription/Manage.php
amine-mp: I am checking and initializing `$data['segments']` not `$data`

PR #2504, opened 2019-12-17
Better empty listings [MAILPOET-2383]

NeosinneR:
[MAILPOET-2383]

[MAILPOET-2383]: https://mailpoet.atlassian.net/browse/MAILPOET-2383

### DIFF
--- a/assets/js/src/newsletters/types.jsx
+++ b/assets/js/src/newsletters/types.jsx
@@ -245,7 +245,7 @@ class NewsletterTypes extends React.Component {
     return (
       <div>
         <link rel="prefetch" href={window.mailpoet_editor_javascript_url} as="script" />
-        <h1>{MailPoet.I18n.t('pickCampaignType')}</h1>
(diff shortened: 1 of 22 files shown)


assets/js/src/newsletters/types.jsx
```diff
@@ -245,9 +250,11 @@ class NewsletterTypes extends React.Component {
     return (
       <div>
         <link rel="prefetch" href={window.mailpoet_editor_javascript_url} as="script" />
-        <h1>{MailPoet.I18n.t(window.mailpoet_newsletters_count === 0 ? 'createFirstEmailTitle' : 'pickCampaignType')}</h1>
 
-        <Breadcrumb step="type" />
+        {this.props.showHeader
+          && <h1>{MailPoet.I18n.t(window.mailpoet_newsletters_count === 0 ? 'createFirstEmailTitle' : 'pickCampaignType')}</h1>
```
wxa: The header is never shown because only the last part of the logical expression is returned.

NeosinneR: Good catch, fixed and also rebased since there were conflicts

PR #2505, opened 2019-12-17
Cron split [MAILPOET-2539]

JanJakes:
[MAILPOET-2539]

[MAILPOET-2539]: https://mailpoet.atlassian.net/browse/MAILPOET-2539

### DIFF
--- a/tests/integration/Cron/DaemonHttpRunnerTest.php
+++ b/tests/integration/Cron/DaemonHttpRunnerTest.php
@@ -2,9 +2,12 @@
 
 namespace MailPoet\Test\Cron;
 
+use Codeception\Stub;
 use Codeception\Stub\Expected;
 use MailPoet\Cron\CronHelper;
 use MailPoet\Cron\CronTrigger;
+use MailPoet\Cron\CronWorkerInterface;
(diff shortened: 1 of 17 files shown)


tests/integration/Cron/DaemonHttpRunnerTest.php
```diff
@@ -68,20 +71,19 @@ function testItStoresErrorMessageAndContinuesExecutionWhenWorkersThrowException(
       'token' => 123,
     ];
 
-    $workers_factory_mock = $this->createWorkersFactoryMock([
-      'createScheduleWorker' => $this->makeEmpty(SimpleWorker::class, [
-        'process' => function () {
+    $run = 0;
+    $cron_worker_runner_mock = $this->make(CronWorkerRunner::class, [
```
pavel-mailpoet: Is there a better way to create this mock? If we add logic to our tests should we also test it?

JanJakes: With Codeception stubs it is not possible to pass callbacks on consecutive calls, only values (https://codeception.com/docs/reference/Stub#public-static-consecutive). Maybe I can mock it with PHPUnit, it won't be much simpler than now I guess.

PR #2516, opened 2019-12-19
Upgrade phpstan [MAILPOET-2588]

JanJakes:
[MAILPOET-2588]

[MAILPOET-2588]: https://mailpoet.atlassian.net/browse/MAILPOET-2588


amine-mp:
I am merging this PR because it's purely technical, there is nothing for QA to test, and because we need it to continue with other PRs.

PR #2528, opened 2020-01-07
Merge styles into single file [MAILPOET-2482]

NeosinneR:
[MAILPOET-2482]

[MAILPOET-2482]: https://mailpoet.atlassian.net/browse/MAILPOET-2482


assets/css/src/components/_importExport.scss
```diff
@@ -24,7 +24,10 @@ span {
 
   &.select2-search {
     &.select2-search--dropdown {
-      display: none !important;
+      #mailpoet_subscribers_import &,
```
costasovo: The whole dropdown is appended to the document body so the prefixing doesn't help here. We will have to probably sneak some additional class into the select2 component itself for this case.

See the second step of import on localhost vs test.mailpoet.com
Screen Shot 2020-01-08 at 9 22 54 AM

NeosinneR: Nice finding 👍 I changed the selector to body class (which is generated by WP based on the current page). The search field is now hidden.

And looking at those screenshots, you are missing group names (`Actions` and `System fields`) - I can't replicate this, I see them correctly. But this shouldn't have anything to do with CSS, you must be missing them in HTML already.

costasovo: I can see Actions and System Fields now. So maybe it was related somehow.

PR #2534, opened 2020-01-09
Fix subscribers limit warning [MAILPOET-2624]

amine-mp:
[MAILPOET-2624]

[MAILPOET-2624]: https://mailpoet.atlassian.net/browse/MAILPOET-2624

### DIFF
--- a/lib/Subscribers/SubscribersRepository.php
+++ b/lib/Subscribers/SubscribersRepository.php
@@ -25,7 +25,7 @@ class SubscribersRepository extends Repository {
       ->createQueryBuilder()
       ->select('count(n.id)')
       ->from(SubscriberEntity::class, 'n')
-      ->where('n.status IN (:statuses)')
+      ->where('n.deleted_at IS NULL AND n.status IN (:statuses)')
       ->setParameter('statuses', [
         SubscriberEntity::STATUS_SUBSCRIBED,
         SubscriberEntity::STATUS_UNCONFIRMED,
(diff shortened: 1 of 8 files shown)


lib/Subscribers/SubscribersRepository.php
```diff
@@ -25,7 +25,7 @@ public function getTotalSubscribers() {
       ->createQueryBuilder()
       ->select('count(n.id)')
       ->from(SubscriberEntity::class, 'n')
-      ->where('n.status IN (:statuses)')
+      ->where('n.deleted_at IS NULL AND n.status IN (:statuses)')
```
costasovo: You can use ->andWhere('n.deleted_at IS NULL') but this is ok as well.

lib/Util/License/Features/Subscribers.php
```diff
@@ -32,22 +34,50 @@ public function check() {
   }
 
   public function hasAPIKey() {
```
costasovo: I would consider renaming this to `hasValidApiKey`. I think this could be easily misinterpreted later by another developer because one could expect that `hasApiKey` covers also the situation when a user has an invalid API key in the database.  I see we pass the value to JS also as `has_api_key` so I think it would be safer to rename it as well.
What do you think?

amine-mp: I agree, I will do that.

codemonkey-jack:
I'm not sure about something in the specs versus the finished version so I've asked Quentin to clarify.

PR #2537, opened 2020-01-13
Camel case [MAILPOET-1796]

JanJakes:
[MAILPOET-1796]

[MAILPOET-1796]: https://mailpoet.atlassian.net/browse/MAILPOET-1796


tests/integration/Router/Endpoints/ViewInBrowserTest.php
```diff
@@ -106,7 +106,7 @@ public function testItFailsValidationWhenSubscriberTokenDoesNotMatch() {
 
   public function testItFailsValidationWhenNewsletterIdIsProvidedButSubscriberDoesNotExist() {
     $data = (object)$this->browserPreviewData;
```
costasovo: Maybe in this case we could just drop conversion to object and use array as we do in API so that we don't have to add so many `// phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps` but it is probably out of scope of this task.

PR #2538, opened 2020-01-13
Fix stats redirect [MAILPOET-2554]

pavel-mailpoet:
When an unauthenticated user comes to any page they are redirected
to the login screen. After they authenticate they are redirected
to the original page. But the # part of the URL is ignored.
So we have to use regular query params from the email and
redirect the user to the canonical URL later.

[MAILPOET-2554]

[MAILPOET-2554]: https://mailpoet.atlassian.net/browse/MAILPOET-2554

### DIFF
--- a/lib/AdminPages/Pages/Newsletters.php
+++ b/lib/AdminPages/Pages/Newsletters.php
@@ -75,6 +75,15 @@ class Newsletters {
   public function render() {
     global $wp_roles; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
 
+    if (isset($_GET['stats'])) {
+      $this->wp->wpSafeRedirect(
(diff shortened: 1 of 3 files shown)


lib/AdminPages/Pages/Newsletters.php
```diff
@@ -75,6 +75,14 @@ public function __construct(
   public function render() {
     global $wp_roles;
 
+    if (isset($_GET['stats'])) {
+      $this->wp->wpSafeRedirect(
+        $this->wp->getSiteUrl(null,
+          '/wp-admin/admin.php?page=mailpoet-newsletters#/stats/' . $_GET['stats']
+        )
+      );
+    }
```
amine-mp: Will the PHP process end here? should we add an `exit` or something?

PR #2558, opened 2020-01-21
Refactor Statistics to Doctrine [MAILPOET-2503]

amine-mp:
[MAILPOET-2503]

[MAILPOET-2503]: https://mailpoet.atlassian.net/browse/MAILPOET-2503

### DIFF
--- a/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php
+++ b/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php
@@ -59,18 +59,18 @@ class NewslettersResponseBuilder {
         $data['options'] = $this->buildOptions($newsletter);
       }
       if ($relation === self::RELATION_TOTAL_SENT) {
-        $data['totalSent'] = $this->newslettersStatsRepository->getTotalSentCount($newsletter);
+        $data['total_sent'] = $this->newslettersStatsRepository->getTotalSentCount($newsletter);
       }
       if ($relation === self::RELATION_CHILDREN_COUNT) {
-        $data['childrenCount'] = $this->newslettersStatsRepository->getChildrenCount($newsletter);
+        $data['children_count'] = $this->newslettersStatsRepository->getChildrenCount($newsletter);
       }
       if ($relation === self::RELATION_SCHEDULED) {
-        $data['totalScheduled'] = (int)SendingQueue::findTaskByNewsletterId($newsletter->getId())
+        $data['total_scheduled'] = (int)SendingQueue::findTaskByNewsletterId($newsletter->getId())
           ->where('tasks.status', SendingQueue::STATUS_SCHEDULED)
           ->count();
       }
       if ($relation === self::RELATION_STATISTICS) {
-        $data['statistics'] = $this->newslettersStatsRepository->getStatistics($newsletter);
+        $data['statistics'] = $this->newslettersStatsRepository->getStatistics($newsletter)->asArray();
       }
     }
     return $data;
--- a/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php
+++ b/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php
@@ -5,12 +5,29 @@ namespace MailPoet\API\JSON\ResponseBuilders;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SendingQueueEntity;
+use MailPoet\Models\SendingQueue;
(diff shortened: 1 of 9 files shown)


lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php
```diff
@@ -27,12 +44,36 @@ public function build(NewsletterEntity $newsletter) {
...
+        $data['segments'] = $this->buildSegments($newsletter);
+      }
+      if ($relation === self::RELATION_OPTIONS) {
+        $data['options'] = $this->buildOptions($newsletter);
+      }
+      if ($relation === self::RELATION_TOTAL_SENT) {
+        $data['totalSent'] = $this->newslettersStatsRepository->getTotalSentCount($newsletter);
+      }
+      if ($relation === self::RELATION_CHILDREN_COUNT) {
+        $data['childrenCount'] = count($newsletter->getChildren());
```
costasovo: This will require to fetch and hydrate all children entities. What about creating a getChildrenCount() method which would only query DB for the count on newslettersRepository?

lib/Entities/StatisticsWooCommercePurchaseEntity.php
```diff
@@ -0,0 +1,64 @@
...
+   * @var SendingQueueEntity|null
+   */
+  private $queue;
+
+  /**
+   * @ORM\ManyToOne(targetEntity="MailPoet\Entities\SubscriberEntity")
+   * @ORM\JoinColumn(name="subscriber_id", referencedColumnName="id")
+   * @var SubscriberEntity|null
+   */
+  private $subscriber;
```
costasovo: I noticed that when we delete a subscriber we don't delete his purchases. I'm not sure what happens to this entity when the association refers to non-existing entity.

costasovo: It is probably ok for now since we don't work with the entity directly. It seems that when we delete subscriber (possible even other entities) we don't clean all related records and we leave some foreign keys. This may not pay off in the future. It makes sense to keep some anonymous statistic data after we delete a subscriber. Maybe we should make those relations nullable in DB and cleanup foreign keys when we delete a subscriber. So this is probably a more general problem not related directly to this PR.

lib/Newsletter/Statistics/NewsletterStatistics.php
```diff
@@ -16,11 +16,15 @@ class NewsletterStatistics {
   /** @var int */
   private $totalSentCount;
 
-  public function __construct($clickCount, $openCount, $unsubscribeCount, $totalSentCount) {
+  /** @var NewsletterWooCommerceRevenue|null */
```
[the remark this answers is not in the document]
amine-mp: It's used here https://github.com/mailpoet/mailpoet/pull/2558/files/c7ff32e1966256aa17a958ff3c2a23caf7e8cb49..72e4f11db8106de596694489b78ec2b72a1d68d7#diff-7915794b23c9896cc65d74576b61cfcaR62

lib/Entities/StatisticsWooCommercePurchaseEntity.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+class StatisticsWooCommercePurchaseEntity {
+  use AutoincrementedIdTrait;
+  use CreatedAtTrait;
+  use UpdatedAtTrait;
+
+  /**
+   * @ORM\ManyToOne(targetEntity="MailPoet\Entities\NewsletterEntity")
+   * @ORM\JoinColumn(name="newsletter_id", referencedColumnName="id")
+   * @var NewsletterEntity|null
```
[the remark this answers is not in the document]
amine-mp: Ok, I added the constructor.

amine-mp:
The goal of this task was to get rid of Paris models: `StatisticsOpens`, `StatisticsClicks` and `StatisticsUnsubscribe`, and replace them with Doctrine entities. After some refactoring, I noticed that we need to refactor `API\JSON\Newsletters::listing` to use Doctrine and that would take more time so we decided to do it in a separated task https://mailpoet.atlassian.net/browse/MAILPOET-2655

amine-mp:
I fixed the bug in the stats (is was a camelCase vs underscores issue) and I added a test.
For premium plugin, it will require rewriting `CampaignStats::getClickedLinks`, do you think we need to do that as part of this PR?

costasovo:
@amine-mp I think it would be better to do the premium plugin part in a separate task so that we don't block this one. Could you please create the ticket?

amine-mp:
I created the task https://mailpoet.atlassian.net/browse/PREMIUM-142

PR #2559, opened 2020-01-21
Form editor launch [MAILPOET-2549]

pavel-mailpoet:
[MAILPOET-2549]

[MAILPOET-2549]: https://mailpoet.atlassian.net/browse/MAILPOET-2549


pavel-mailpoet:
@codemonkey-jack This is the task that covers the whole new form editor. Also, it probably needs cross-browser testing.

PR #2561, opened 2020-01-22
Send a confirmation email when a user is added to a website [MAILPOET-2567]

pavel-mailpoet:
[MAILPOET-2567]

[MAILPOET-2567]: https://mailpoet.atlassian.net/browse/MAILPOET-2567


wxa:
@pavel-mailpoet When I add a user manually through WP admin (Users > Add New), a confirmation email is sent twice.

PR #2564, opened 2020-01-22
Fix integration tests on PHP 7.4 [MAILPOET-2577]

wxa:


.circleci/config.yml
```diff
@@ -313,7 +320,7 @@ jobs:
           path: /tmp/fake-mailer/
           destination: fake-mailer
   php7_integration:
-    executor: wordpress_mysql80_php73
```
NeosinneR: Since this executor is no longer used, please remove it also from the list of executors.

wxa: Fixed.

PR #2571, opened 2020-01-27
Update Gutenberg packages [MAILPOET-2610]

costasovo:
[MAILPOET-2610]

[MAILPOET-2610]: https://mailpoet.atlassian.net/browse/MAILPOET-2610


assets/css/src/components/formEditor/components/_custom_field.scss
```diff
@@ -21,3 +21,18 @@
     text-align: left;
   }
 }
+
+// Fix for placeholder component alignment bug introduced in @wordpress/components@9.0.0
```
pavel-mailpoet: Is there an issue in GitHub we could link from here? We need to know when can we remove this code otherwise this will be here forever.

costasovo: There is no issue. It seems that in new version they simply changed all placeholders so that they render aligned to left. It seems that this is rather a permanent change so maybe I should update the comment.

costasovo: I thought that this was a bug because all placeholders appeared ugly to me, but it seems that it is an intended change. So we can either redesign the create custom field block to be left-aligned or just keep it as is using css change.

PR #2579, opened 2020-01-29
Form editor: Display below pages and below posts [MAILPOET-2639]

pavel-mailpoet:
[MAILPOET-2639]

[MAILPOET-2639]: https://mailpoet.atlassian.net/browse/MAILPOET-2639


assets/js/src/form_editor/store/selectors.jsx
```diff
@@ -17,6 +17,12 @@ export default {
   getFormSettings(state) {
     return state.formData.settings;
   },
+  placeFormBellowAllPages(state) {
```
costasovo: So far we have been using `getSomething` naming convention for selectors. Why did you decide to change it? Shouldn't we change it for all?

tests/javascript/form_editor/store/map_form_data_before_saving.spec.js
```diff
@@ -0,0 +1,62 @@
...
+const data = {
+  id: '1',
+  name: 'My First Form',
+  settings: {
+    segments: ['3'],
+    on_success: 'message',
+    success_message: 'Check your inbox or spam folder to confirm your subscription.',
+    success_page: '5',
+    segments_selected_by: 'admin',
+    placeFormBellowAllPages: true,
```
costasovo: We use underscore or naming properties within the request data. I think we should not mix it.

assets/js/src/form_editor/store/map_form_data_before_saving.jsx
```diff
@@ -0,0 +1,10 @@
+export default function mapFormDataBeforeSaving(data) {
+  return {
+    ...data,
+    settings: {
+      ...data.settings,
+      placeFormBellowAllPages: data.settings.placeFormBellowAllPages === true ? '1' : '',
```
costasovo: We use underscore for request data.

lib/Form/DisplayFormInWPContent.php
```diff
@@ -0,0 +1,56 @@
...
+  public function __construct(WPFunctions $wp, FormsRepository $formsRepository) {
+    $this->wp = $wp;
+    $this->formsRepository = $formsRepository;
+  }
+
+  // TODO remove transient in the api on form save
+
+  public function display(string $content): string {
+    $result = $content;
+    if (!$this->wp->isSingle()) return $result;
```
costasovo: I can see we have a test for this but for a simple comment saying that this code ensures that we display the form only on a page which is related to single post would be nice here.

lib/Form/DisplayFormInWPContent.php
```diff
@@ -0,0 +1,56 @@
...
+    if (!is_array($settings)) return false;
+    if (!isset($settings['placeFormBellowAllPosts'])) return false;
+    if (
+      ($settings['placeFormBellowAllPosts'] === '1')
+      && !$this->wp->isPage()
+    ) return true;
+    if (
+      ($settings['placeFormBellowAllPages'] === '1')
+      && $this->wp->isPage()
+    ) return true;
```
costasovo: I think this check is not correct. We have two switches: below post and below pages. With this implementation both do the same they will display if is_page returns true. We need somehow distinguish if the page is related to post or if it is a page. I can see that in wp_post table there is a post_type column so maybe we can use that information somehow. I'm not really sure what is the best way of doing it. Maybe we should also consider what to do with custom post types e.g. WooCommerce product. 
If we distinguish types it would probably require to have more caching keys.

assets/js/src/form_editor/store/selectors.jsx
pavel-mailpoet: I didn't decide to change it. I didn't know about the convention. It makes sense for most of the fields. I thought if it returns boolean it makes sense to omit the `get` prefix. You think all of those functions should be prefixed with `get`?

costasovo: We probably don't have to prefix all. I think that it would be good so that these methods were named as getters. Name `placeFormBellowAllPages` doesn't sound to me as a getter name but it might be just my problem. You decide.

lib/Form/DisplayFormInWPContent.php
costasovo: Hey my bad. I missed the negation in condition `!$this->wp->isPage()` for posts 🤦‍♂

lib/Form/DisplayFormInWPContent.php
```diff
@@ -39,10 +39,18 @@ public function display(string $content): string {
 
   private function shouldDisplay():bool {
     if (!$this->wp->isSingle()) return false;
-    if ($this->wp->getTransient(DisplayFormInWPContent::NO_FORM_TRANSIENT_KEY)) return false;
+    $cache = $this->wp->getTransient(DisplayFormInWPContent::NO_FORM_TRANSIENT_KEY);
+    if (isset($cache[$this->wp->getPostType()])) return false;
     return true;
   }
 
+  private function saveNoPosts() {
```
costasovo: Maybe better name would be `saveNoForms`. What do you think?

codemonkey-jack:
@pavel-mailpoet This fails QA as it's behind a feature flag and based on the spec it shouldn't be.

codemonkey-jack:
This fails QA. The "Below all posts" option works, but when I set the form to display below all pages it never displays. For example: http://camel.wpsatchel.com/sample-page/

PR #2580, opened 2020-01-29
Remove old form editor [MAILPOET-2625]

pavel-mailpoet:
[MAILPOET-2625]

[MAILPOET-2625]: https://mailpoet.atlassian.net/browse/MAILPOET-2625


assets/css/src/components/formEditor/components/_sidebar.scss
```diff
@@ -85,9 +85,3 @@ $gutenberg-control-border-color-focus: #007cba;
     margin: 8px 0;
   }
 }
-
-// Styles for old editor link
-.mailpoet_old_form_editor_link,
```
costasovo: Hmm, actually we can't delete the whole definition here. It would cause this: https://d.pr/i/Qdf284
What about renaming the class to something like `mailpoet_sidebar_note`?

PR #2583, opened 2020-01-30
Remove WooCommerce step from setup wizard [MAILPOET-2619]

JanJakes:
[MAILPOET-2619]

[MAILPOET-2619]: https://mailpoet.atlassian.net/browse/MAILPOET-2619


assets/js/src/wizard/steps/woo_commerce_step.jsx
```diff
@@ -1,43 +0,0 @@
...
-        ${(props.loading ? 'mailpoet_welcome_wizard_step_controls_loading' : '')}`
-      }
-    >
-      <button
-        type="button"
-        className="button button-primary"
-        onClick={props.next}
-        disabled={props.loading}
-      >
-        {MailPoet.I18n.t('gotIt')}
```
NeosinneR: This string should also be removed, it's not used anywhere else. https://github.com/mailpoet/mailpoet/blob/master/views/welcome_wizard.html#L60

assets/js/src/wizard/welcome_wizard_controller.jsx
```diff
@@ -135,21 +134,6 @@ const WelcomeWizardStepsController = (props) => {
...
-      { stepName === 'WelcomeWizardWooCommerceStep'
-        ? (
-          <WelcomeWizardStepLayout
-            step={step}
-            stepsCount={stepsCount}
-            illustrationUrl={window.wizard_woocommerce_illustration_url}
-          >
-            <WelcomeWizardWooCommerceStep
-              next={() => redirect(step)}
-              screenshot_src={window.wizard_woocommerce_box_url}
```
NeosinneR: These two URL definitions should also be removed https://github.com/mailpoet/mailpoet/blob/master/views/welcome_wizard.html#L9-L10  (and also `is_woocommerce_active` 3 lines below)

PR #2590, opened 2020-02-04
Add Gutenberg Block for MailPoet subscription form [MAILPOET-1798]

pavel-mailpoet:
[MAILPOET-1798]

[MAILPOET-1798]: https://mailpoet.atlassian.net/browse/MAILPOET-1798


assets/js/src/post_editor_block/form_block.jsx
```diff
@@ -0,0 +1,19 @@
...
+const { registerBlockType } = wp.blocks;
+
+registerBlockType('mailpoet/form-block', {
+  title: 'Example: Basic (esnext)',
+  icon: 'universal-access-alt',
+  category: 'widgets',
+  example: {},
+  edit() {
+    return (
+      <div className="mailpoet-block-div">Hello World, step 1 (from the editor).</div>
```
JanJakes: Is the `-div` suffix something we use (or Gutenberg) in other places? I'm wondering whether that suffix is necessary (or whether it should be something rather semantical such as `-wrapper`).

lib/Form/FormsRepository.php
```diff
@@ -18,4 +18,14 @@ protected function getEntityClassName() {
...
   }
 
+  public function findAllNotDeleted() {
+    return $this->entityManager
+      ->createQueryBuilder()
+      ->select('f')
+      ->from(FormEntity::class, 'f')
+      ->where('f.deletedAt IS NULL')
+      ->getQuery()
+      ->getResult();
```
JanJakes: We could add an "orderBy" here to return the forms in some defined order ("updated_at" maybe? or "name"?).

assets/js/src/post_editor_block/edit.jsx
```diff
@@ -1,18 +1,26 @@
...
+function Edit({ attributes, setAttributes }) {
   function displayFormsSelect() {
     if (!Array.isArray(allForms)) return null;
     if (allForms.length === 0) return null;
     return (
-      <select>
+      <select
+        onChange={(event) => {
+          setAttributes({
+            selectedForm: parseInt(event.target.value, 10),
```
JanJakes: This is just a small naming detail but maybe simply `formId` would be nicer considering the rendered code (`<!-- wp:mailpoet/form-block {"selectedForm":1} /-->`).

lib/Entities/FormEntity.php
```diff
@@ -102,4 +102,17 @@ public function setStyles($styles) {
...
+      'name' => $this->getName(),
+      'body' => $this->getBody(),
+      'settings' => $this->getSettings(),
+      'styles' => $this->getStyles(),
+      'created_at' => $this->getCreatedAt(),
+      'updated_at' => $this->getUpdatedAt(),
+      'deleted_at' => $this->getDeletedAt(),
+    ];
+  }
+
```
JanJakes: We're not doing this in other places yet and I'm just thinking whether it's a good idea. Self-serialization of entities is always a problem since different use cases require different data but I guess there's no harm in this case.

But maybe instead of implementing `\JsonSerializable` you could just and special method such as `toArray()`? It has the advantage of being easily searchable if one day we want to know in how many places this is used.

lib/Config/PostEditorBlock.php
```diff
@@ -0,0 +1,103 @@
...
+    );
+
+    $this->wp->wpEnqueueStyle(
+      'mailpoetblock-form-block-css',
+      Env::$assetsUrl . '/dist/css/' . $this->renderer->getCssAsset('post-editor-block.css'),
+      ['wp-edit-blocks'],
+      Env::$version
+    );
+
+    register_block_type('mailpoet/form-block', [
```
JanJakes: We should add all this functions to `WP\Functions` and use it from there.

assets/js/src/post_editor_block/form_block.jsx
pavel-mailpoet: No, this was a temporary code form a tutorial.

codemonkey-jack:
@pavel-mailpoet Looks like checks failed so I don't have the zip to test with. And there are changes requested too it seems. Sending back to you.

codemonkey-jack:
@pavel-mailpoet And now on a refresh of the repo it works, ignore the previous comment.

codemonkey-jack:
The forms display correctly for me, the only thing is they don't bring over the output button style into the Gutenberg preview, but I guess this is ok? Over to you for Product Review @qfrery.

Also when we take them to create a form after clicking "create a new form" it'd be nice to open a URL like: http://camel.wpsatchel.com/wp-admin/admin.php?page=mailpoet-form-editor&id=3 in the actual form editor, rather than the form page and then ask them to create a form again.

codemonkey-jack:
Ok so some emails were delayed to my gmail and there some problems so I have sent this back. 

1. I received the welcome email when subscribing via the Gutenberg block before confirming my subscription. 
2. I received 3 confirmation emails before. 
3. As the admin I received the new subscriber notification before the subscriber had confirmed their email.

pavel-mailpoet:
@codemonkey-jack I fixed the "Create a new form". But I cannot replicate the problem you got with those emails locally. I use a regular widget so it would be surprising if the widget was broken only on this page. Do you have a link where I can test it?

assets/js/src/post_editor_block/subscription_form/edit.jsx
```diff
@@ -46,7 +46,7 @@ function Edit({ attributes, setAttributes }) {
     return (
       <div className="mailpoet-block-create-new-content">
         <a
-          href="admin.php?page=mailpoet-forms"
+          href="admin.php?page=mailpoet-form-editor&operation=create"
```
JanJakes: I think we use `action` (not `operation`) on other pages, don't we?

codemonkey-jack:
Now with @qfrery in product QA. 👍

PR #2593, opened 2020-02-06
Preview links [MAILPOET-2430]

JanJakes:
[MAILPOET-2430]

[MAILPOET-2430]: https://mailpoet.atlassian.net/browse/MAILPOET-2430


lib/API/JSON/v1/Newsletters.php
```diff
@@ -435,19 +424,20 @@ public function showPreview($data = []) {
       $newsletter->body = $this->emoji->encodeForUTF8Column(MP_NEWSLETTERS_TABLE, 'body', $newsletter->body);
       $newsletter->save();
       $subscriber = Subscriber::getCurrentWPUser();
-      $previewUrl = NewsletterUrl::getViewInBrowserUrl(
-        NewsletterUrl::TYPE_LISTING_EDITOR,
-        $newsletter,
-        $subscriber
-      );
+      $previewUrl = NewsletterUrl::getViewInBrowserUrl($newsletter, $subscriber);
+      $publicPreviewUrl = NewsletterUrl::getViewInBrowserUrl($newsletter);
```
amine-mp: I think we don't need two URLs here. `$previewURL` is used for the iframe when previewing the email on the editor.
in which case, the user is already logged-in so `$publicPreviewURL` would use the current user and display the same content.

JanJakes: Good point. I added a commit that uses a single link _without_ subscriber - that way it will be "anonymous" when logged out and with a current user when logged in.

No I'm actually wondering about all other places where a subscriber is included. I mean - we do need subscriber info in the links inside sent emails ("view this in browser", etc.), which means shortcodes, but what about other use cases? I found:

1. Newsletters API: [here](https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/v1/Newsletters.php#L126) and [here](https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/v1/Newsletters.php#L157) and [here](https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/v1/Newsletters.php#L286).
2. Export: [here](https://github.com/mailpoet/mailpoet/blob/438e26b42ff1d0fdf82ea4bfa52b6cc7361468c5/lib/Subscribers/ImportExport/PersonalDataExporters/NewslettersExporter.php#L70)

I think for 1. (the API) we don't need to pass `$subscriber` at all - it's our API used only when you're logged in anyway and it does seem to pass current WP user. For 2., the export, I'm not sure.

Do you think I should further simplify those two occurrences? @amine-mp

amine-mp: I agree on 1. There is no need to include the subscriber.
For the export, I think we should keep the subscriber in the URL, because we are using them when exporting data of a specific subscriber so it makes sense to have his specific URLs.
So I suggest simplifying the ones on API/Newsletters.

JanJakes: Added a commit and rebased on master.

PR #2596, opened 2020-02-11
Form Renderer Refactor [MAILPOET-2665]

costasovo:
[MAILPOET-2665]

[MAILPOET-2665]: https://mailpoet.atlassian.net/browse/MAILPOET-2665

### DIFF
--- a/lib/Form/Block/Base.php
+++ b/lib/Form/Block/Base.php
@@ -19,7 +19,7 @@ class Base {
     $this->wp = $wp;
   }
 
-  public function getInputValidation($block, $extraRules = []) {
+  public function getInputValidation(array $block, array $extraRules = []): string {
     $rules = [];
 
     if ($block['id'] === 'email') {
@@ -78,7 +78,7 @@ class Base {
     return join(' ', $validation);
   }
 
-  public function renderLabel($block) {
+  public function renderLabel(array $block): string {
     $html = '';
     if (
       isset($block['params']['hide_label'])
@@ -106,7 +106,7 @@ class Base {
     return $html;
   }
 
-  public function renderInputPlaceholder($block) {
+  public function renderInputPlaceholder(array $block): string {
     $html = '';
     // if the label is displayed as a placeholder,
     if (
@@ -126,7 +126,7 @@ class Base {
   }
 
   // return field name depending on block data
-  public function getFieldName($block = []) {
+  public function getFieldName(array $block = []): string {
     if ((int)$block['id'] > 0) {
       return 'cf_' . $block['id'];
     } elseif (isset($block['params']['obfuscate']) && !$block['params']['obfuscate']) {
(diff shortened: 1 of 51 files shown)


lib/Form/Util/DateConverter.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+namespace MailPoet\Form\Util;
```
pavel-mailpoet: I don't understand why is this in the `Form` namespace? It seems like it has nothing to do with forms. Does it?

tests/unit/Form/Util/FieldNameObfuscatorTest.php
```diff
@@ -2,46 +2,43 @@
 
 namespace MailPoet\Form\Util;
 
-use Codeception\Stub;
 use MailPoet\WP\Functions as WPFunctions;
+use PHPUnit\Framework\MockObject\MockObject;
 
 class FieldNameObfuscatorTest extends \MailPoetUnitTest {
+
+  /** @var MockObject | WPFunctions */
```
pavel-mailpoet: I think this should be an intersection type using `&` instead of `|` https://medium.com/@ondrejmirtes/union-types-vs-intersection-types-fd44a8eacbb

It is also used in other tests.

tests/unit/Form/HtmlParser.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace MailPoet\Test\Form;
+
+class HtmlParser extends \MailPoetUnitTest {
```
pavel-mailpoet: Could you please explain why is this extending `MailPoetUnitTest`?

lib/Form/Renderer.php
```diff
@@ -26,6 +26,12 @@ public function __construct(
     $this->blocksRenderer = $blocksRenderer;
   }
 
+  public function render(array $form = []): string {
```
pavel-mailpoet: I am not sure if we need this method here. If it is used only in one place we can use those other methods as well in `lib/Form/DisplayFormInWPContent.php`. We could maybe in the future split this class further because `renderStyles` and `renderHtml` have little in common and could be separated.

lib/Form/Block/Base.php
```diff
@@ -6,7 +6,7 @@
 use MailPoet\Models\ModelValidator;
 use MailPoet\WP\Functions as WPFunctions;
 
-abstract class Base {
+class Base {
```
pavel-mailpoet: I am not sure about the class name. I mean it was called `Base` because it was extended in all the other block renderers. Now it is not extended any more so we should probably use a better name. 

I am doubtful about the need for this class. Why are all of those methods in a single class?

tests/unit/Form/Block/BaseTest.php
```diff
@@ -0,0 +1,117 @@
...
+use MailPoet\Form\Block\Base;
+use MailPoet\Form\Util\FieldNameObfuscator;
+use MailPoet\WP\Functions as WPFunctions;
+use PHPUnit\Framework\MockObject\MockObject;
+
+class BaseTest extends \MailPoetUnitTest {
+  /** @var Base */
+  private $baseRenderer;
+
+  /** @var MockObject|WPFunctions */
```
pavel-mailpoet: I noticed you separated the `|` operator by space from both sides in other tests. You don't do that here, is there a reason for that?

PR #2597, opened 2020-02-11
Import missing functions from polyfill-mbstring [MAILPOET-2695]

wxa:
Please test with the `mbstring` extension disabled.


wxa:
@JanJakes That will be done in another ticket which may require thorough testing: https://mailpoet.atlassian.net/browse/MAILPOET-2580 
The current one is just a quick fix.

PR #2605, opened 2020-02-12
Allow null content in MailPoet\Form\DisplayFormInWPContent [MAILPOET-2699]

costasovo:
[MAILPOET-2699]

[MAILPOET-2699]: https://mailpoet.atlassian.net/browse/MAILPOET-2699

### DIFF
--- a/lib/Form/DisplayFormInWPContent.php
+++ b/lib/Form/DisplayFormInWPContent.php
@@ -20,8 +20,14 @@ class DisplayFormInWPContent {
     $this->formsRepository = $formsRepository;
   }
 
-  public function display(string $content): string {
-    if(!$this->shouldDisplay()) return $content;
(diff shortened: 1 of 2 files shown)


lib/Form/DisplayFormInWPContent.php
```diff
@@ -20,8 +20,11 @@ public function __construct(WPFunctions $wp, FormsRepository $formsRepository) {
     $this->formsRepository = $formsRepository;
   }
 
-  public function display(string $content): string {
-    if(!$this->shouldDisplay()) return $content;
+  /**
+   * @return string|null
+   */
+  public function display(string $content = null) {
```
pavel-mailpoet: We should be safe and remove the type-hint all together here and check if the value is a string.

PR #2614, opened 2020-02-17
Remove MailPoet.Notice from React components [MAILPOET-2391]

amine-mp:
[MAILPOET-2391]

[MAILPOET-2391]: https://mailpoet.atlassian.net/browse/MAILPOET-2391

### DIFF
--- a/assets/js/src/experimental_features/experimental_features.jsx
+++ b/assets/js/src/experimental_features/experimental_features.jsx
@@ -7,28 +7,25 @@ import Notices from 'notices/notices.jsx';
 const ExperimentalFeatures = () => {
   const [flags, setFlags] = useState(null);
   const contextValue = useGlobalContextValue(window);
-  const [mounted, setMounted] = useState(false);
+  const showError = contextValue.notices.error;
 
   useEffect(() => {
-    if (!mounted) {
-      MailPoet.Ajax.post({
-        api_version: window.mailpoet_api_version,
-        endpoint: 'featureFlags',
-        action: 'getAll',
-      }).done((response) => {
-        const flagsMap = response.data.reduce((obj, item) => ({ ...obj, [item.name]: item }), {});
-        setFlags(flagsMap);
-      }).fail((response) => {
-        if (response.errors.length > 0) {
-          contextValue.notices.error(
-            <>{response.errors.map((error) => <p>{error.message}</p>)}</>,
-            { scroll: true }
-          );
-        }
-      });
-      setMounted(true);
-    }
-  }, [contextValue.notices, mounted]);
+    MailPoet.Ajax.post({
+      api_version: window.mailpoet_api_version,
+      endpoint: 'featureFlags',
+      action: 'getAll',
+    }).done((response) => {
+      const flagsMap = response.data.reduce((obj, item) => ({ ...obj, [item.name]: item }), {});
+      setFlags(flagsMap);
+    }).fail((response) => {
+      if (response.errors.length > 0) {
+        showError(
+          <>{response.errors.map((error) => <p>{error.message}</p>)}</>,
+          { scroll: true }
+        );
+      }
+    });
+  }, [showError]);
 
   function handleChange(event) {
     const name = event.target.name;
@@ -49,7 +46,7 @@ const ExperimentalFeatures = () => {
       contextValue.notices.success(<p>{message}</p>);
     }).fail((response) => {
       if (response.errors.length > 0) {
-        contextValue.notices.error(
+        showError(
           response.errors.map((error) => <p key={error.message}>{error.message}</p>),
           { scroll: true }
         );
--- a/assets/js/src/experimental_features/experimental_features.jsx
+++ b/assets/js/src/experimental_features/experimental_features.jsx
@@ -7,24 +7,28 @@ import Notices from 'notices/notices.jsx';
 const ExperimentalFeatures = () => {
   const [flags, setFlags] = useState(null);
   const contextValue = useGlobalContextValue(window);
+  const [mounted, setMounted] = useState(false);
 
   useEffect(() => {
-    MailPoet.Ajax.post({
-      api_version: window.mailpoet_api_version,
-      endpoint: 'featureFlags',
-      action: 'getAll',
-    }).done((response) => {
-      const flagsMap = response.data.reduce((obj, item) => ({ ...obj, [item.name]: item }), {});
-      setFlags(flagsMap);
-    }).fail((response) => {
-      if (response.errors.length > 0) {
-        MailPoet.Notice.error(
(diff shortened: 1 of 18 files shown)


assets/js/src/experimental_features/experimental_features.jsx
```diff
@@ -7,24 +7,28 @@ import Notices from 'notices/notices.jsx';
...
+        if (response.errors.length > 0) {
+          contextValue.notices.error(
+            <>{response.errors.map((error) => <p>{error.message}</p>)}</>,
+            { scroll: true }
+          );
+        }
+      });
+      setMounted(true);
+    }
+  }, [contextValue.notices, mounted]);
```
JanJakes: Adding `mounted` here is kind of a hack. I tested this and was trying to understand why you do it this way (and @costasovo helped me to understand the problem) but I don't think this is a good solution. Now, this "mounted" workaround is used here, in a single place. However, looking at `useGlobalContextValue(...)`, there's a potential risk of doing this hack in many other components in the future.

Looking at how `useGlobalContextValue(...)` is implemented, it seems it always creates a new object and returns it. That's why this ends up in a loop (and adding a "mounted" state is a workaround to prevent that loop). My question is - why do we always return a new object?

Going deeper, the `useGlobalContextValue(...)` creates four other subcontexts, and those too are always new objects. Then `useEffect` with not only `contextValue` will loop but also a `useEffect` with `contextValue.notices` only. Moreover, this could have some memory-usage consequences - can't we end up creating many of these context objects over and over again for every single usage?

One more question concerning the `useGlobalContextValue(...)` - why do we have to pass `window` from every single component where we'll use it? Do you think it's a good idea that many components along our codebase will touch `window` object? I personally don't like it (in fact, I think `window` should be prohibited by eslint, with maybe one or a few exceptions).

Could you maybe explain those concerns a bit (and the design of this global context hook) in regard to the points raised above? If there's more to discuss it may worth a Slack chat with others as well. I think we should handle this first so we don't end up polluting our code with more occurrences of thins such as the "mounted".

It may as well worth explaining why we have some of the other contexts (for instance, segments/users context). I don't think it was really discussed and I'm not sure their existence is justifiable, those are static values that won't change and may be passed by props in many cases. How do we decide which data should be passed by props and which should have a context wrapper? Context may be overkill for some use cases, not to mention it's not a very intuitive construct. We should use it carefully.

JanJakes:
Btw. the linked [ticket](https://mailpoet.atlassian.net/browse/MAILPOET-2391) seems to be only about planning how to start using the new notices. This seems to be implementation. Did you change your mind and implement it all? Or is it just a part of it?

amine-mp:
Hey @JanJakes, thanks for the detailed comment. It revealed some issues I was not aware of.

First for the Jira task, yes it was only about planing but I had some time so I started the implementation and I added a comment to the ticket explaining what remains to be done. Here I only refactored 10 files, there are 56 React components left.

Now let's answer your concerns, I summarized them as follows:
1- `mounted` is a hack! isn't there a cleaner way?
2- Will we be creating many context objects for no reason!
3- Why do we pass `window` to `useGlobalContextValue`, aren't global variables bad?
4- Why are there constants in the context? couldn't they be passed by props or accessed directly from `window`?
5- How do we decide which data should be passed by props and which should have a context wrapper?

Let me explain the design I have in mind. Each admin page of our plugin has its own React application or entry point.
I wanted to create a context to be used in all these entry points so we are able to use it within any of our components.
The context should contain two types of data (this answers **question 5**):
- The common data that is shared between many components at different levels (like Notices in our case).
- Global external data that comes from the `window` object.

> Why do you want to put all `mailpoet_*` data on `window` into the context?

To be able to mock these data during tests. Yes, we don't write tests for React components yet, but preparing our code to be testable is a good thing. Ideally, I want our entry points to look like:
```jsx
ReactDOM.render(
  <Wrapper data={window}>
    <App/>
  </Wrapper>
)
```
The `Wrapper` would make the context from the given data and provide it to the `App`.
This would be the only place where we access `window` (**questions 3 and 4**).
That way, we can do this during tests
```jsx
render(
  <Wrapper data={mockedWindowData}>
    <AnyComponent />
  </Wrapper>
)
// assert behaviour of the component given the window data ...
```

> But can't we just use the `window` object during tests? I mean we are using `jsdom` so we already have a mocked window object right?

Yes we do, but what about running multiple tests in parallel, they would target the same `window` and cause conflicts.

> But with current code, we are creating a new context object every time something changes in the context. This will cause memory and performance problems!

Right. We should use `React.useMemo` and `React.useCallback` to memoize values. I added [this commit](https://github.com/mailpoet/mailpoet/pull/2614/commits/cd78f9dc72cb37882e25ecc6f54a175486e38f88) to do that (**question 2**).

> Does memoizing values so the issue that is now solved with the `mount` hack?

No, memoizing only solved half the issue. We had the following
```jsx
React.useEffect(() => {
  //...
  contextValue.notices.error(...)
}, [contextValue.notices])
```
Whenever `contextValue.notices.error` is called, the value of `contextValue.notices` changes and the effect is run again causing an infinite loop.

> Here the effect only uses `notices.error` so it should only depend on it and not on the whole `notices` object, right?

Right, but when I use `[contextValue.notices.error]` as dependencies array, ESLint doesn't accept it. So my solution was to do
```jsx
const showError = contextValue.notices.error;
React.useEffect(() => {
  //...
  showError(...)
}, [showError])
```
Now the effect will run only when `error` changes its reference. But looking at `useNotices` definition:
```js
const [items, setItems] = React.useState([]);
const [nextId, setNextId] = React.useState(1);

const getNextId = React.useCallback(() => {
  setNextId((x) => x + 1);
  return nextId;
}, [nextId]);

const add = React.useCallback((item) => {
  setItems((xs) => [...xs, { ...item, id: item.id || getNextId() }]);
}, [getNextId]);
//...
const error = React.useCallback(
  (content, props = {}) => add({ ...props, type: 'error', children: content }),
  [add]
);
//...
```
Here `error` changes only when `add` changes, which does only when `getNextId` changes, which does when `nextId` changes, which does when `add` is called, so when `error` is called.

**Result:** `error` changes every time we call it. So calling it from inside a `useEffect` will cause an infinite loop!

> So this whole React context thing is a mistake?!

Nope. The mistake is in my implementation of `useNotices`. I fixed it in [this commit](https://github.com/mailpoet/mailpoet/pull/2614/commits/6e3d1e9e3ab1fe80b51b48fdd1fa430fd561e98e)

**To sum up**
- I want to avoid accessing `window` from our components by providing data in the context, this will help us to test components easier.
- When using `useEffect` make sure you depend only on what you are really using.
- Use `useMemo` and `useCallback` to memoize values and avoid recreating the same objects/functions.
- When exposing a function from a hook. This function should not change its reference every time it's called.

PR #2619, opened 2020-02-20
Fix signup confirmation email for WP users trigger [MAILPOET-2700]

JanJakes:
[MAILPOET-2700]

[MAILPOET-2700]: https://mailpoet.atlassian.net/browse/MAILPOET-2700

### DIFF
--- a/lib/AdminPages/Pages/Settings.php
+++ b/lib/AdminPages/Pages/Settings.php
@@ -61,7 +61,6 @@ class Settings {
 
   public function render() {
     $settings = $this->settings->getAll();
-    $flags = $this->getFlags();
 
     $premiumKeyValid = $this->servicesChecker->isPremiumKeyValid(false);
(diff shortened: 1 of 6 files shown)


wxa:
This implements what is said in the ticket, but the ticket itself could break another customer case (https://mailpoet.atlassian.net/browse/MAILPOET-2565):

> A customer had user registration disabled in his WP settings, but had a WooCommerce registration form on the frontend which still added WP users

This customer wanted to send confirmation emails to those WP users, but couldn't due to inability to activate the "Subscribe in registration form" setting:

> MP3 settings said "Registration is disabled on this site" so you couldn't enable subscription in registration form.

Maybe we should add a possible solution here: 
> keep the "Subscribe in registration form" form control in settings even if registration is disabled

JanJakes:
@wxa Added a commit.

wxa:
Updated the setting description according to the updated specs.

PR #2624, opened 2020-02-26
Add possibility to use TypeScript [MAILPOET-2658]

amine-mp:
[MAILPOET-2658]

Please check [this post on Slack](https://mailpoet.slack.com/files/T02AL75TW/FUC1MFL2F?origin_team=T02AL75TW) for details about this PR.

[MAILPOET-2658]: https://mailpoet.atlassian.net/browse/MAILPOET-2658


assets/js/src/notices/notice.tsx
```diff
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { FC, ReactNode } from 'react';
 import ReactDOM from 'react-dom';
-import PropTypes from 'prop-types';
 import MailPoet from 'mailpoet';
 
-const Notice = ({
+type Props = {
+  type: 'success' | 'info' | 'warning' | 'error';
+  children: ReactNode;
```
JanJakes: This was string/element/array-of-elements originally. Is ReactNode sufficient? Thinking about it, maybe it is - ReactNode can contain any children, right? Just double checking.

assets/js/src/notices/api_errors_notice.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { FC } from 'react';
+import Notice from 'notices/notice';
+
+type Props = {
+  errors: Array<{ message: string }>
+}
+
+const APIErrorsNotice: FC<Props> = ({ errors }) => {
```
JanJakes: Is it necessary to import and use `FC` in all functional components now? Or can we do something like this (https://github.com/milesj/babel-plugin-typescript-to-proptypes):
```
const APIErrorsNotice: (props: Props) => {
  const errors = props.errors;
```

Or is there a difference? Just trying to figure out what will be the best practices now.

.eslintrc.ts.json
```diff
@@ -0,0 +1,137 @@
...
+  },
+  "plugins": [
+    "react-hooks",
+    "no-only-tests",
+    "@typescript-eslint"
+  ],
+  "settings": {
+    "import/resolver": "webpack"
+  },
+  "rules": {
```
JanJakes: These rules are something we want to turn on, eventually?

assets/js/src/notices/notice.tsx
```diff
@@ -1,9 +1,19 @@
...
+  scroll?: boolean;
+  closable?: boolean;
+  renderInPlace?: boolean;
+  onDisplay?: () => void;
+  onClose?: () => void;
+  timeout?: number | false;
+};
+
+const Notice: FC<Props> = ({
```
JanJakes: Same question if we need to use `FC` here or if there's other syntax. Also... is it really nicer to unwrap the properties to such a large number of variables? Kind of thinking if `props.onClose`... isn't more convenient.

assets/js/src/notices/notice.tsx
amine-mp: Yes!

assets/js/src/notices/api_errors_notice.tsx
amine-mp: The two ways are equivalent, the only difference is that using `FC<Props>` will implicitly add an optional `children` property of type `ReactNode`.

.eslintrc.ts.json
amine-mp: Yes, we want to decide which ones to turn on and do it later.

assets/js/src/notices/notice.tsx
amine-mp: The props were spread like that before this change. But yeah I would prefer to use `props.*` in this case. Do you think I should make the change in this PR?

PR #2625, opened 2020-02-27
Form global styles [MAILPOET-2600]

pavel-mailpoet:
[MAILPOET-2600]

[MAILPOET-2600]: https://mailpoet.atlassian.net/browse/MAILPOET-2600


costasovo:
@pavel-mailpoet Approved but needs rebase.

PR #2632, opened 2020-03-05
Bootstrap the new settings page [MAILPOET-2676]

amine-mp:
[MAILPOET-2676]

The new settings URL is `?page=mailpoet-new-settings`.

## WP Redux and Typescript

I created a selector `getSetting` that takes the path to a setting and return its value. The usage looked like that:
```jsx
const email = useSelect(
  (sel) => sel(STORE_NAME).getSetting(['sender', 'address']),
  []
);
```
But there were some issues with that:
- I could give any array of strings to `getSetting` without any error.
- The type of `email` was `any` which is not helping in detecting errors.
- I could even write `sel(STORE_NAME).anyThing()` and Typescript won't be able to help me!

So first I made the `useSelector` hook and the code above became:
```jsx
const email = useSelector('getSetting')(['sender', 'address']);
```
The `useSelector` takes the name of a selector (and not any string) and returns a selector function with correct types (based on the types definitions on `store/selectors.ts`). I had to use some Typescript magic to return the same selector function type without the `state` parameter. Now calling `useSelector('somethingElse')` will be a Typescript error. I did the same for actions with the `useAction` hook.
But now I could still give any wrong setting path and the return type was still `any`. That's why I created another hook `useSetting` that calls `useSelector('getSetting')` internally but gives proper type checking. The code became:
```jsx
const email = useSetting('sender', 'address');
```
Now `email` has type `string` and calling `useSetting('some','missing', 'thing')` will give type error.
With the same logic, I made the hook `useSettingSetter` which returns a setter function for a specific setting.

[MAILPOET-2676]: https://mailpoet.atlassian.net/browse/MAILPOET-2676


assets/js/src/settings/store/actions.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function* saveSettings(data: Settings) {
+  yield { type: 'SAVE_STARTED' };
+  const error = yield { type: 'SEND_DATA_TO_API', data };
+  if (error) {
+    return { type: 'SAVE_FAILED', error };
+  }
+  yield { type: 'TRACK_SETTINGS_SAVED', data };
+  return { type: 'SAVE_DONE' };
+}
```
costasovo: Just a note: In the form editor, we decided to keep actions without any logic and we [put the flow into control](https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/form_editor/store/controls.jsx#L36). The idea behind this decision is to have the logic in one place. It is not a best practice or something we need to enforce.
We also don't pass all the data we want to save to the action but we [select it from the store within the code of the control.](https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/form_editor/store/controls.jsx#L45) That way you don't have to distribute those data in every component which could possibly trigger saving.

amine-mp:
Thanks, I added 3 commits to address your remarks.

PR #2636, opened 2020-03-06
Shortcodes cannot spread to multiple lines [MAILPOET-2660]

pavel-mailpoet:
Wordpress as well doesn't use those modifiers

[MAILPOET-2660]

[MAILPOET-2660]: https://mailpoet.atlassian.net/browse/MAILPOET-2660


pavel-mailpoet:
There is another related bug: https://mailpoet.atlassian.net/browse/MAILPOET-2659

PR #2637, opened 2020-03-06
Add KB about the new form [MAILPOET-2713]

pavel-mailpoet:
[MAILPOET-2713]

[MAILPOET-2713]: https://mailpoet.atlassian.net/browse/MAILPOET-2713

### DIFF
--- a/views/form/editor.html
+++ b/views/form/editor.html
@@ -7,6 +7,13 @@
 <% endblock %>
 
 <% block container %>
+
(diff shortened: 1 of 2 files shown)


pavel-mailpoet:
Good find @veljkho . We didn't specify what articles to display on that page so the beacon showed only the top articles. I updated it and now it is displaying only the article about the new editor.

PR #2638, opened 2020-03-06
Remove Feedback section from form editor [MAILPOET-2735]

pavel-mailpoet:
[MAILPOET-2735]

[MAILPOET-2735]: https://mailpoet.atlassian.net/browse/MAILPOET-2735


qfrery:
@costasovo I agree, please do!

PR #2639, opened 2020-03-09
Disable columns blocks in form editor [MAILPOET-2609]

costasovo:
[MAILPOET-2609]

[MAILPOET-2609]: https://mailpoet.atlassian.net/browse/MAILPOET-2609


costasovo:
@pavel-mailpoet I added a commit with a fix for default margin CSS. A theme may have a CSS reset which may cancel our CSS so I've slightly increased specificity of the rule.

PR #2642, opened 2020-03-09
Remove absolute positioning [MAILPOET-2744]

pavel-mailpoet:
[MAILPOET-2744]


amine-mp:
After the change, the "Open in new tab" and Desktop/Mobile radio buttons are no longer in the same line. is that fine?
preview

pavel-mailpoet:
I think it is. It was originally like that before it got broken: https://mailpoet.atlassian.net/browse/MAILPOET-2430

PR #2649, opened 2020-03-12
Use name as fallback label [MAILPOET-2747]

pavel-mailpoet:
[MAILPOET-2747]

[MAILPOET-2747]: https://mailpoet.atlassian.net/browse/MAILPOET-2747


pavel-mailpoet:
As stated in JIRA, our current custom field implementation is still WIP and not ideal. I used the name as a fallback and when we have a place for custom fields editing this will improve.

PR #2653, opened 2020-03-16
Newsletter templates doctrine [MAILPOET-2647]

JanJakes:
[MAILPOET-2647]

[MAILPOET-2647]: https://mailpoet.atlassian.net/browse/MAILPOET-2647


lib/Entities/NewsletterTemplateEntity.php
```diff
@@ -0,0 +1,137 @@
...
+   * @var NewsletterEntity|null
+   */
+  private $newsletter;
+
+  /**
+   * @ORM\Column(type="string")
+   * @Assert\NotBlank()
+   * @var string
+   */
+  private $name;
```
pavel-mailpoet: I think the type of this property is not correct. Also, getter and setter have an incorrect type. `$name` can be null if I write:
```
  $t = new NewsletterTemplateEntity();
  $t->getName();
```

We should prevent creating the entity without the name or update the type.

This also applies to other properties.

JanJakes: In DB this column is not nullable so I'd rather not make it nullable. What remains is adding it to a constructor which causes a bit of an issue in the incoming data processing - I either have to add an extra check for `$data['name']` and throw an exception or something. However, the same is done by the `Assert\NotBlank()` annotation so I solved it by using `$data['name'] ?? ''` to avoid duplicating the validation logic.

PR #2657, opened 2020-03-17
Implement the Sign-up Confirmation settings tab [MAILPOET-2678]

amine-mp:
[MAILPOET-2678]

- This PR is based on https://github.com/mailpoet/mailpoet/pull/2648 so that one should be merged first.

[MAILPOET-2678]: https://mailpoet.atlassian.net/browse/MAILPOET-2678


assets/js/src/settings/pages/signup_confirmation/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import EnableSignupConfirmation from './enable_signup_confirmation';
+
+export default function SignupConfirmation() {
```
pavel-mailpoet, line 8: As far as I know, we don't use the name `index` for files.  We always use the folder name. See `form` or `listing` for example.

assets/js/src/settings/pages/signup_confirmation/email_content.tsx
```diff
@@ -1,4 +1,6 @@
-import React from 'react';
+import React, { Fragment } from 'react';
```
pavel-mailpoet, line 1: Could you please help me understand why do we need `Fragment` here?

assets/js/src/settings/pages/signup_confirmation/email_content.tsx
```diff
@@ -1,4 +1,6 @@
...
+import ReactStringReplace from 'react-string-replace';
```
pavel-mailpoet, line 2: I don't see this being used, am I missing something?

assets/js/src/settings/pages/signup_confirmation/index.tsx
amine-mp: I use `index.tsx` so that I can use the folder name in the import
```ts
import SignupConfirmation from 'settings/page/signup_confirmation';
```
This is a convention used widely by many libraries.

pavel-mailpoet: I understand that @amine-mp but that convention has not been used by MailPoet.

assets/js/src/settings/pages/signup_confirmation/email_content.tsx
amine-mp: You are right, I will remove it.

assets/js/src/settings/pages/signup_confirmation/email_content.tsx
amine-mp: This is a leftover from the refactoring, I will remove it.

assets/js/src/settings/pages/signup_confirmation/email_content.tsx
pavel-mailpoet: I wonder why didn't eslint complain. I thought we have it configured like that.

PR #2659, opened 2020-03-17
Update UsageExamples.md [MAILPOET-2820]

creativecatapps:
Changed variable name of $subscriber to $getsubscriber, so the subscriber array isn't overwritten.


creativecatapps:
I've updated the file - never done a pull request before, so unsure whether I need to do anything else. It doesn't look like it but please let me know.

costasovo:
Thank you for fixing this!

PR #2660, opened 2020-03-17
Unapproved key notices [MAILPOET-2729]

JanJakes:
[MAILPOET-2729]

[MAILPOET-2729]: https://mailpoet.atlassian.net/browse/MAILPOET-2729

### DIFF
--- a/lib/AdminPages/Pages/Newsletters.php
+++ b/lib/AdminPages/Pages/Newsletters.php
@@ -6,6 +6,7 @@ use MailPoet\AdminPages\PageRenderer;
 use MailPoet\Config\Env;
 use MailPoet\Config\Installer;
 use MailPoet\Config\Menu;
+use MailPoet\Config\ServicesChecker;
 use MailPoet\Features\FeaturesController;
 use MailPoet\Listing\PageLimit;
 use MailPoet\Models\Newsletter;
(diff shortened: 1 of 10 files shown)


lib/AdminPages/Pages/Newsletters.php
```diff
@@ -102,6 +108,11 @@ public function render() {
     $data['roles'] = $wp_roles->get_names(); // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
     $data['roles']['mailpoet_all'] = $this->wp->__('In any WordPress role', 'mailpoet');
 
+    $mssActive = Bridge::isMPSendingServiceEnabled();
+    $mssKeyValid = $this->servicesChecker->isMailPoetAPIKeyValid();
+    $mssKeyPendingApproval = $this->settings->get('mta.mailpoet_api_key_state.data.is_approved') === false;
+    $data['mss_key_pending_approval'] = $mssActive && $mssKeyValid && $mssKeyPendingApproval;
```
costasovo: I can see this exact same logic in two places. What about putting it into service checker service method?

JanJakes: Good idea, done.

PR #2663, opened 2020-03-19
Add paragraph block to the form editor [MAILPOET-2614]

pavel-mailpoet:
[MAILPOET-2614]

[MAILPOET-2614]: https://mailpoet.atlassian.net/browse/MAILPOET-2614

### DIFF
--- a/assets/js/src/form_editor/store/form_body_to_blocks.jsx
+++ b/assets/js/src/form_editor/store/form_body_to_blocks.jsx
@@ -110,6 +110,22 @@ const mapColorSlug = (colorDefinitions, colorValue) => {
   return result ? result.slug : undefined;
 };
 
+/**
+ * @param {Array.<{name: string, slug: string, size: number}>} fontSizeDefinitions
+ * @param {string} fontSizeValue
+ */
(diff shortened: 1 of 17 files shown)


assets/js/src/form_editor/store/form_body_to_blocks.jsx
```diff
@@ -67,6 +67,22 @@ const mapColorSlug = (colorDefinitions, colorValue) => {
...
 
+/**
+ * @param {Array.<{name: string, slug: string, size: number}>} fontSizeDefinitions
+ * @param {string} fontSizeValue
+ */
+const mapFontSizeSlug = (fontSizeDefinitions, fontSizeValue) => {
+  let value = 0;
+  if (fontSizeValue) {
+    value = parseInt(fontSizeValue, 10);
+    if (Number.isNaN(value)) {
```
costasovo: Just wondering why do we need to handle this?

pavel-mailpoet: Just to make sure that we don't break something if a non-numeric value comes from the server.

pavel-mailpoet:
@veljkho Good find. We don't support HTML editing. I removed it from the menu.

PR #2667, opened 2020-03-20
Implement the Advanced settings tab [MAILPOET-2680]

amine-mp:
[MAILPOET-2680]

[MAILPOET-2680]: https://mailpoet.atlassian.net/browse/MAILPOET-2680


amine-mp:
Good points @costasovo 
For point 1. I removed the old settings template as we don't use it anymore.
For point 2. After discussing with Tautvidas, I added the bounce address again, It seems that we need it after all.

costasovo:
@amine-mp Approved! There is a conflict so I'm assigning it bac to you.

amine-mp:
@veljkho I changed CSS to add more spacing between fields and little spacing for checkboxes.
Let me know if anything else should be changed :)

PR #2676, opened 2020-03-25
Pop up form [MAILPOET-2741]

pavel-mailpoet:
[MAILPOET-2741]

[MAILPOET-2741]: https://mailpoet.atlassian.net/browse/MAILPOET-2741

### DIFF
--- a/assets/css/src/components-form-editor/_sidebar.scss
+++ b/assets/css/src/components-form-editor/_sidebar.scss
@@ -131,23 +131,18 @@
 .form-placement-settings-overlay {
   background-color: rgba(0, 0, 0, .3);
   border-radius: 2px;
-  height: $form-placement-option-height - (2 * $form-placement-option-border);
+  height: 100%;
   left: 0;
   pointer-events: none;
   position: absolute;
   top: 0;
-  width: $form-placement-option-width - (2 * $form-placement-option-border);
+  width: 100%;
   z-index: 3;
 }
 
 .form-placement-option-active {
   border: solid $form-placement-option-active-border $gutenberg-control-active-color;
   box-shadow: 0 4px 8px 0 rgba(156, 166, 204, .5);
-
-  .form-placement-settings-overlay {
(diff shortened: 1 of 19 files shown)


assets/css/src/components-public/_public.scss
```diff
@@ -157,3 +157,36 @@ $form-columns-space-between: 20px;
     @include maxWidthColumn;
   }
 }
+
+div.mailpoet_form_popup {
+  background-color: white;
+  border-radius: 10px;
+  box-shadow: 0 4px 35px 0 rgba(195, 65, 2, .2);
+  display: none;
+  height: 240px;
```
NeosinneR: It can't have fixed height, because there can be many inputs. 

Screenshot 2020-03-25 19 03 56

assets/css/src/components-form-editor/_sidebar.scss
```diff
@@ -105,6 +105,11 @@
   border: 1px solid #eee;
 }
 
+.form-placement-option-list {
+  display: flex;
+  flex-direction: row;
+}
+
 .form-placement-option {
```
NeosinneR: This shouldn't have any left/right margin but rather add `justify-content: space-between` and `flex-wrap: wrap;` to `.form-placement-option-list`

assets/js/src/public.jsx
```diff
@@ -2,50 +2,91 @@ import MailPoet from 'mailpoet';
...
+function setCookie(name, value, days) {
+  let expires = '';
+  if (days) {
+    const date = new Date();
+    date.setTime(date.getTime() + (days * 24 * 60 * 60 * 1000));
+    expires = `; expires=${date.toUTCString()}`;
+  }
+  document.cookie = `${name}=${value}${expires}; path=/`;
+}
+function getCookie(name) {
```
NeosinneR: I know this is from SO, but maybe at least some reasonable variable names inside this function so it's readable for devs? It'll be minified when compiled.

PR #2679, opened 2020-03-26
Normalize settings values [MAILPOET-2823]

amine-mp:
[MAILPOET-2823]

[MAILPOET-2823]: https://mailpoet.atlassian.net/browse/MAILPOET-2823


assets/js/src/settings/store/normalize_settings.ts
```diff
@@ -0,0 +1,219 @@
+import _ from 'lodash';
+import { t } from 'common/functions';
+import { Settings } from './types';
+
+export default function normalizeSettings(data: any): Settings {
+  const text = asString('');
+  const disabledCheckbox = asBoolean('1', '0', '0');
+  const enabledCheckbox = asBoolean('1', '0', '1');
+  const disabledRadio = asBoolean('1', '', '');
+  const enabledRadio = asBoolean('1', '', '1');
```
JanJakes: The aim of this task is to "normalize" data values. This, however, creates many different options how to save boolean. Maybe I don't understand the reasons but I don't think this is a good idea. Instead, I think in our `Settings` type boolean should always be boolean. Not that false is `0` for some fields, `''` for others, and never actually `false`. Don't you think?

assets/js/src/settings/store/normalize_settings.ts
```diff
@@ -0,0 +1,219 @@
...
+  const emptyArray = asStringArray([]);
+  const smtpServer = asEnum(['server', 'manual', 'AmazonSES', 'SendGrid'], 'server');
```
JanJakes: For enums, shouldn't we use native TS enums? I mean, for instance, using a named type `SettingsSmtpServerEnum` here. Or is that bad since those are internally represented as numbers? 🤔

assets/js/src/settings/store/normalize_settings.ts
amine-mp: I totally agree with you that we should have a unique way to represent things, but right now we don't. Changing that now will change what we send to the API and may break the plugin behaviour somewhere. I am trying to keep the save values (even if they are wrong) to not have to refactor the PHP code (which will not be easy).

assets/js/src/settings/store/normalize_settings.ts
amine-mp: We do use TS enums in `settings/store/types.ts`, but here we need a function that takes any value and forces it to be in the enum.

amine-mp:
Good point @JanJakes, I agree that this is a generic problem and there may be libraries for that already. But given the time constraints, I choose to implement our own simple solution because it's faster. We can revisit that later and decide whether to keep this home-made solution or switch to a generic library.

assets/js/src/settings/store/normalize_settings.ts
JanJakes: Those are union types in fact, but that's fine, it seems the TS "Enum" would indeed be a number, internally.

assets/js/src/settings/store/normalize_settings.ts
```diff
@@ -0,0 +1,219 @@
...
+}
+function asObject<T extends Schema>(schema: T) {
+  return (value: any): SchemaResult<T> => {
+    const object = Object.keys(schema).reduce((result, field) => ({
+      [field]: schema[field](value ? value[field] : undefined),
+      ...result,
+    }), {});
+    return object as SchemaResult<T>;
+  };
+}
```
JanJakes: Wow, this is elegant. A bit hard to grasp with the generics and return types but nice!

amine-mp: The power of Typescript :D

veljkho:
Having difficulties to verify

- Saving Basics will not save WooCommerce screen
This is happening when I go to WC screen, and let's say tick the first box "Use MailPoet to customize WC.." and untick all the rest checkboxes on the page, then save it. Then just go to Basics screen and just press Save button. Then go back to WC screen, observe that checkboxes aren't ticked as they should. If you can't reproduce this at the first, please try to tick/untick differently in WC screen, and save Basics then.

I must say that also saving WC screen will always auto untick 2 boxes in Basics screen under "Stats notifications", for some reason. So that's another issue by the same logic.

amine-mp:
Very good findings @veljkho 

## First issue

> Changing something on the WC tab (or any old tab) and switching to Basics tab then saving does not save the changes on WC tab.

This happens because right now the Basics and WC tabs are on two different pages. When you go from one tab to the other, the whole page is reloaded and all unsaved changes are lost. Just like if you changed something on settings then went to the Emails or Forms page before saving.

This is a temporary problem that should be fixed once we migrate all settings tabs. Also I don't think users are doing changes and switching tabs before clicking the save button.

## Second issue

> Saving WC tab (or any old tab) changes the Basics settings. This happened due to the old Basics  fields that were still in the view and were sent to the API. I removed them from the HTML view and the issue should be now fixed.

Good job on finding these issues, please let me know if you find any other bug :)

PR #2680, opened 2020-03-26
Authorize from modal [MAILPOET-2804]

JanJakes:
[MAILPOET-2804]

[MAILPOET-2804]: https://mailpoet.atlassian.net/browse/MAILPOET-2804

### DIFF
--- a/assets/css/src/components-plugin/_set-from-address-modal.scss
+++ b/assets/css/src/components-plugin/_set-from-address-modal.scss
@@ -1,14 +1,14 @@
-#set-from-address-modal {
+.set-from-address-modal {
   p:first-child {
     margin-top: 0;
   }
 
-  #mailpoet_set_from_address_modal_address {
+  input[type=text] {
     display: block;
     width: 230px;
   }
 
-  #mailpoet_set_from_address_modal_save {
+  input[type=submit] {
     display: block;
     margin-top: 30px;
   }
--- a/assets/css/src/components-plugin/_set-from-address-modal.scss
+++ b/assets/css/src/components-plugin/_set-from-address-modal.scss
@@ -1,0 +1,15 @@
+#set-from-address-modal {
+  p:first-child {
+    margin-top: 0;
+  }
+
+  #mailpoet_set_from_address_modal_address {
+    display: block;
+    width: 230px;
+  }
+
+  #mailpoet_set_from_address_modal_save {
+    display: block;
+    margin-top: 30px;
+  }
+}
--- a/assets/js/src/common/set_from_address_modal.tsx
+++ b/assets/js/src/common/set_from_address_modal.tsx
@@ -3,6 +3,7 @@ import ReactStringReplace from 'react-string-replace';
 import jQuery from 'jquery';
 import MailPoet from 'mailpoet';
 import Modal from 'common/modal/modal.jsx';
+import { GlobalContext } from 'context';
 
 const mailPoetApiVersion = (window as any).mailpoet_api_version as string;
 
@@ -30,9 +31,22 @@ const getErrorMessage = (error: any|null): string => {
   return error.message || MailPoet.I18n.t('setFromAddressEmailUnknownError');
 };
 
-const getSuccessMessage = (): string => MailPoet.I18n.t('setFromAddressEmailSuccess').replace(
-  /\[link\](.*?)\[\/link\]/g,
-  '<a href="?page=mailpoet-settings#basics" rel="noopener noreferrer">$1</a>'
+const getSuccessMessage = (): JSX.Element => (
+  <p>
+    { ReactStringReplace(
+      MailPoet.I18n.t('setFromAddressEmailSuccess'),
+      /\[link\](.*?)\[\/link\]/g,
+      (match) => (
+        <a
+          key="setFromAddressModalBasicsTabLink"
+          href="?page=mailpoet-settings#basics"
+          rel="noopener noreferrer"
+        >
+          {match}
+        </a>
+      )
+    )}
+  </p>
 );
 
 const removeUnauthorizedEmailNotices = () => {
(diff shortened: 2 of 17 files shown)


assets/css/src/components-plugin/_set-from-address-modal.scss
```diff
@@ -0,0 +1,15 @@
+#set-from-address-modal {
```
pavel-mailpoet: You should use classes in css.

assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import ReactDOMServer from 'react-dom/server';
+import ReactStringReplace from 'react-string-replace';
+import MailPoet from 'mailpoet';
+
+const showSetFromAddressModal = async () => {
```
pavel-mailpoet: Can I ask why did you mark this function as `async`? There is nothing wrong with it, but it doesn't return anything (if I didn't miss anything) so I don't see the reason for it.

assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -0,0 +1,47 @@
...
+  MailPoet.Modal.popup({
+    title: MailPoet.I18n.t('setFromAddressModalTitle'),
+    template: ReactDOMServer.renderToString(
```
pavel-mailpoet: I am not sure about the consequences of using `ReactDOMServer` here. I have never used it before so I am just asking about it. AFAIK its intended purpose is to render React on the server. I am not sure about the usage in the browser. Do you think it is fine?

lib/Mailer/Mailer.php
```diff
@@ -74,7 +74,7 @@ private function buildMailer() {
           $this->sender,
           $this->replyTo,
           new MailPoetMapper(),
-          new AuthorizedEmailsController($this->settings, new Bridge)
+          ContainerWrapper::getInstance()->get(AuthorizedEmailsController::class)
```
pavel-mailpoet: Why didn't you inject it directly?

assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -41,6 +44,22 @@ const showSetFromAddressModal = async () => {
         />
       </div>
     ),
+    onInit: () => {
+      const saveButton = document.getElementById('mailpoet_set_from_address_modal_save') as HTMLInputElement;
+      const addressInput = document.getElementById('mailpoet_set_from_address_modal_address') as HTMLInputElement;
+      const addressValidator = jQuery(addressInput).parsley();
+
+      saveButton.addEventListener('click', async () => {
```
pavel-mailpoet: This looks a bit odd to me. Again, I am not rejecting here, just asking for clarification. 
- why is this function async? it doesn't return anything, does it?
- you are combining react, jquery and native JS listeners in a single file, that doesn't seem right to me. There is technically nothing wrong with that. It just doesn't seem correct to me. Care to explain the reasoning?

assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -59,6 +90,20 @@ const showSetFromAddressModal = async () => {
         if (!address) {
           return;
         }
+        try {
+          await handleSave(address);
+          MailPoet.Modal.close();
+          MailPoet.Notice.success(getSuccessMessage());
```
pavel-mailpoet: Why are you using this old notice? I thought they are deprecated since Amine created the new mechanism.

assets/js/src/common/set_from_address_modal.tsx
JanJakes: For the `async` here - the function uses `await` (below). I may have added the `await` in a later commit due to some rebases which is probably confusing, but in the end it needs `async`.

To explain React, jQuery, & JS (plus also old vs. new notices):
1. This is a modal which is a non-React component.
2. To get a body rendered by React in there I need to render the to string and pass it to the modal. This is done my `ReactDOMServer`. It's not new in our codebase, same logic is used in [confirm alert](https://github.com/mailpoet/mailpoet/blob/a8b31d1a745688acfff41ee8b23976803c02990f/assets/js/src/common/confirm_alert.jsx) or [NPS poll](https://github.com/mailpoet/mailpoet/blob/a8b31d1a745688acfff41ee8b23976803c02990f/assets/js/src/nps_poll.jsx) modals.
3. Since the result is a string, there are no hooks, no events, no "dynamic" React functionality. That's why any behavior must be done using native JS `addEventListener`.
4. I need form validation here and we use `Parsley`. Parsley is a library that extends jQuery which is why I need it to be able to call the `.parsley()` at all.
5. I also need notices but since I'm not in the React tree I don't have a way to use React context with the new notices. (Now, about this last point I'm not 100% sure, maybe there's a way how to interact with React notices from the modal but I guess it's not really important - because what we really need here is a rewrite of Modal component to React.)

assets/js/src/common/set_from_address_modal.tsx
JanJakes: Explained in https://github.com/mailpoet/mailpoet/pull/2680/files#r400026892. Simply, I don't know if there's a way to interact with the React-rendered tree from non-react code.

lib/Mailer/Mailer.php
JanJakes: There are a couple of places that use `new Mailer()`. I'd have to refactor all of those and it wouldn't be a small refactor.

assets/js/src/common/set_from_address_modal.tsx
JanJakes: We already use it like this, explained in https://github.com/mailpoet/mailpoet/pull/2680/files#r400026892.

assets/js/src/common/set_from_address_modal.tsx
pavel-mailpoet: Do I understand it correctly that the reason for this combination of technologies is the modal? Why didn't you use this one then? Is there something wrong with it? https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/common/modal/modal.jsx

JanJakes:
@veljkho Good finds!

Added two commits with fixes and rebased on master.

PR #2685, opened 2020-03-30
Implement the Key Activation settings tab [MAILPOET-2681]

amine-mp:
[MAILPOET-2681]

[MAILPOET-2681]: https://mailpoet.atlassian.net/browse/MAILPOET-2681


assets/js/src/settings/pages/key_activation/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const activatePremiumPlugin = useAction('activatePremiumPlugin');
+  const [senderAddress] = useSetting('sender', 'address');
+  const [unauthorizedAddresses] = useSetting('authorized_emails_addresses_check');
+
+  const verifyKey = async (event) => {
+    if (!state.key) {
+      notices.error(<p>{t('premiumTabNoKeyNotice')}</p>, { scroll: true });
+      return;
+    }
+    const isTrusted = event.isTrusted;
```
JanJakes, line 27: This had a [better name](https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/old_settings/premium_tab/premium_tab.jsx#L193) before, now we're losing the information on why it was used.

But - is it still relevant? It was clicked on by another settings tab. Will it be still clickable now? See https://github.com/mailpoet/mailpoet/blob/42fc6c0e3e7840da22b45727aeef213e4159f6db/views/old_settings.html#L121.

In any case we must not forget to remove this when "send with" tab is ready!

assets/js/src/settings/store/actions.ts
```diff
@@ -25,3 +30,138 @@ export function* saveSettings() {
   yield { type: 'TRACK_SETTINGS_SAVED' };
   return { type: 'SAVE_DONE' };
 }
+
+export function* verifyMssKey(key: string, activateMssIfKeyValid: boolean) {
+  const { success, error, res } = yield {
```
JanJakes, line 35: I have a question here since Redux is new to me. When generators should be used? Is it when any action that may change the state is performed? Is `yield` the way to go through the reducers? Thanks.

And if it is so, why do we also use the API call this way?

assets/js/src/settings/store/actions.ts
```diff
@@ -25,3 +30,138 @@ export function* saveSettings() {
...
+    const call = yield {
+      type: 'CALL_API',
+      endpoint: 'settings',
+      action: 'set',
+      data: {
+        mta_group: 'mailpoet',
+        mta: { method: 'MailPoet', mailpoet_api_key: key },
+        signup_confirmation: { enabled: '1' },
+      },
+    };
```
JanJakes, line 60: Are we doing a bit of duplicated work here? Shouldn't this be managed only by "setSetting" and save?

assets/js/src/settings/store/actions.ts
```diff
@@ -25,3 +30,138 @@ export function* saveSettings() {
...
+  let pluginActive = res.meta.premium_plugin_active;
+
+  if (!pluginInstalled) {
+    const actions = installPremiumPlugin();
+    let action = actions.next();
+    while (!action.done) {
+      yield action.value;
+      action = actions.next();
+    }
+    pluginInstalled = action.value;
```
JanJakes, line 110: Why does this need to be iterated? Could we just yield it as with "activatePremiumPlugin" below?

assets/js/src/settings/store/make_default_state.ts
```diff
@@ -1,19 +1,59 @@
...
+  const segments = window.mailpoet_segments;
+  const save = { inProgress: false, error: null };
+  const data = normalizeSettings(window.mailpoet_settings);
+  const flags = {
+    error: false,
+    woocommerce: !!window.mailpoet_woocommerce_active,
+    newUser: !!window.mailpoet_is_new_user,
+    mssKeyValid: window.mailpoet_mss_key_valid,
+    premiumKeyValid: window.mailpoet_premium_key_valid,
+    premiumPluginInstalled: window.mailpoet_premium_plugin_installed,
```
JanJakes, line 16: These 3 new flags seem to be used only in this file. What about avoiding adding them to the state? I mean - instead of adding them to flags we can use them just like local variables here (or touch "window" directly).

views/old_settings.html
```diff
@@ -114,12 +109,6 @@ <h2 class="nav-tab-wrapper" id="mailpoet_settings_tabs">
           if ($('.mailpoet_mta_setup_save').is(':visible')) {
             $('.mailpoet_mta_setup_save').trigger('click');
           }
-          var mailpoet_premium_key = $('#mailpoet_premium_key').val();
-          // sync mss key with premium key
-          $('#mailpoet_api_key').val(mailpoet_premium_key);
-          if (mailpoet_premium_key.length > 0) {
-            $('#mailpoet_premium_key_verify').trigger('click', false);
-          }
```
JanJakes, line 122: Interesting. Will we use only a single key in the settings store? It actually seems like we used it like that already before... just in DB there seem to be 2 keys. So I guess this is OK since [we indeed save it for MSS as well](https://github.com/mailpoet/mailpoet/pull/2685/files#diff-2909d3626e5a2ce4276dfd5d29be6628R57).

assets/js/src/settings/pages/key_activation/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+      premiumStatus: null,
+      premiumInstallationStatus: null,
+    });
+    MailPoet.Modal.loading(true);
+    const mssStatus: MssStatus = (await verifyMssKey(state.key, isTrusted)) as any;
+    await verifyPremiumKey(state.key);
+    MailPoet.Modal.loading(false);
+    if (isTrusted) {
+      const authorizedAddressNeeded = !senderAddress || unauthorizedAddresses;
+      if (mssStatus === 'valid_mss_active' && authorizedAddressNeeded) {
```
JanJakes, line 39: One more thing - I noticed that when you have an unauthorized email, then click "Verify", then in the modal add an authorized email and then click "Verify" again, it will display the modal again. But it shouldn't. Also, I think once I had to click twice to actually display the modal (not sure how to replicate now).

Is this caused by using `senderAddress` and `unauthorizedAddresses` which could have changed for the next render cycle but not within the run here in the callback itself? I think that may be the issue - a "wrong" usage of "state" vars.

assets/js/src/settings/pages/key_activation/messages/premium_messages.tsx
```diff
@@ -0,0 +1,77 @@
...
+  <div className="mailpoet_error">
+    {message || MailPoet.I18n.t('premiumTabPremiumKeyNotValidMessage')}
+  </div>
+);
+
+type Props = {
+  keyMessage?: string
+  activationCallback: () => any
+  installationCallback: () => any
+  installationStatus: PremiumInstallationStatus
```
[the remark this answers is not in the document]
amine-mp: commas are optional in `type` declarations. Do we need to enforce them? do you get warnings on the file `settings/store/types.ts`?

assets/js/src/settings/pages/key_activation/messages/premium_messages.tsx
```diff
@@ -0,0 +1,77 @@
...
+  <div className="mailpoet_error">
+    {MailPoet.I18n.t('premiumTabPremiumNotActiveMessage')}
+    {' '}
+    <button type="button" className="button-link" onClick={callback}>
+      {MailPoet.I18n.t('premiumTabPremiumActivateMessage')}
+    </button>
+  </div>
+);
+
+type NotValidMessageProps = { message?: string }
```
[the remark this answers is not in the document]
amine-mp: I think any function that returns JSX should be declared as a React component. I have even seen an ESLint rule proposal to enforce that https://github.com/yannickcr/eslint-plugin-react/issues/578
That's why I am using props here.
```jsx
const NotValidMessage = ({ message }: { message?: string }) => (
```
Looks ugly that's why I am declaring a type alias then using it.

assets/js/src/settings/index.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import jQuery from 'jquery';
+import 'parsleyjs';
```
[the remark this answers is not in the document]
amine-mp: This is because the Modal uses `parsley` to validate the input. Please note that this is the webpack entry point for the settings page, it doesn't use the common entry point that already includes jQuery and parsley.

assets/js/src/settings/pages/key_activation/index.tsx
amine-mp: This variable was already there and is passed to the `verifyMssKey` action. I just reused it for the modal. I will rename it to `isUserTriggered`

assets/js/src/settings/store/actions.ts
amine-mp: An [action creator](https://developer.wordpress.org/block-editor/packages/packages-data/#actions) is defined as a generator function when it needs to get some asynchronous data in order to decide which actions to dispatch. The way to get asynchronous data is by `yield`ing a [control](https://developer.wordpress.org/block-editor/packages/packages-data/#controls), so here the `CALL_API` action is a controlled action that calls the API and returns the response, a boolean indicating if the request was successful, and the error if any.

assets/js/src/settings/store/actions.ts
amine-mp: Good point. Unfortunately, I can't do that because the `saveSettings` action will do additional things that we don't want here (track mixpanel event), and also we will not know if it succeeded or failed (an action creator should not access the state directly unless there is no other choice).

assets/js/src/settings/store/actions.ts
amine-mp: You are right, my bad.

assets/js/src/settings/store/make_default_state.ts
amine-mp: You are right, I missed those.

assets/js/src/settings/pages/key_activation/index.tsx
amine-mp: This happens because the modal is shown based on data on the store, but when the new authorized address is sent to the API, the store is not updated, so it shows the modal again. I fixed that by passing a `setAuthorizedAddress` callback to the modal so that it can save the new address on the store.

amine-mp:
Very good remarks @JanJakes, I added 11 commits to address them separately.

assets/js/src/settings/pages/key_activation/messages/premium_messages.tsx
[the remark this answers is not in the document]
JanJakes: I think it comes from JS eslint file that considers it an object. It's about IDE configuration, not important now.

PR #2691, opened 2020-03-31
Woocommerce double opt in [MAILPOET-2799]

pavel-mailpoet:
[MAILPOET-2799]

[MAILPOET-2799]: https://mailpoet.atlassian.net/browse/MAILPOET-2799

### DIFF
--- a/lib/Segments/WooCommerce.php
+++ b/lib/Segments/WooCommerce.php
@@ -80,8 +80,14 @@ class WooCommerce {
     $wcSegment = Segment::getWooCommerceSegment();
 
     if ($wcOrder === false or $wcSegment === false) return;
+    $signupConfirmation = $this->settings->get('signup_confirmation');
+    $status = Subscriber::STATUS_UNCONFIRMED;
+    if ((bool)$signupConfirmation['enabled'] === false) {
+      $status = Subscriber::STATUS_SUBSCRIBED;
+    }
 
-    $insertedEmails = $this->insertSubscribersFromOrders($orderId);
+    $insertedEmails = $this->insertSubscribersFromOrders($orderId, $status);
+
     if (empty($insertedEmails[0]['email'])) {
       return false;
     }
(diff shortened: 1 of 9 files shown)


lib/Segments/WooCommerce.php
```diff
@@ -80,8 +80,14 @@ public function synchronizeGuestCustomer($orderId, $currentFilter = null) {
     $wcSegment = Segment::getWooCommerceSegment();
 
     if ($wcOrder === false or $wcSegment === false) return;
+    $signupConfirmation = $this->settings->get('signup_confirmation');
+    $status = Subscriber::STATUS_UNCONFIRMED;
+    if ((bool)$signupConfirmation['enabled'] === false) {
```
amine-mp: The specs don't mention that we should only use `unconfirmed` status if the confirmation email setting in on. My understanding is to always use `unconfirmed` status.

lib/WooCommerce/Subscription.php
```diff
@@ -111,6 +116,29 @@ public function subscribeOnCheckout($orderId, $data) {
...
+    $subscriber->save();
+  }
+
+  private function updateNotSubscribedSubscriber(Subscriber $subscriber) {
+    $signupConfirmation = $this->settings->get('signup_confirmation');
+    if ((bool)$signupConfirmation['enabled'] === false) {
+      // if double opt-in is disabled we act as if the subscriber was already subscribed
+      $this->updateAlreadySubscribedSubscriber($subscriber);
+      return;
+    }
```
amine-mp: It would make more sense to do this test before calling this method. Doing it on the `subscribeOnCheckout` method and calling `updateAlreadySubscribedSubscriber` directly from there seems more reasonable.

lib/Segments/WooCommerce.php
pavel-mailpoet: @flakas I think you should decide this one because I am not sure.

When the sign-up confirmation is disabled, what should happen? Do we still keep the status `unconfirmed`?

flakas: @amine-mp @pavel-mailpoet Good catch, I have not taken into account that "Signup confirmation" can be turned off.
As much as I would like to always force double opt-in, since we do permit the option to disable "Signup confirmation", for consistency's sake I believe we should follow the "Signup confirmation" setting.

So I think the behavior that Pavel implemented is good:
a) If "Signup confirmation" is enabled, we set the subscriber as unconfirmed, and in most cases we send them a confirmation email (in a different code part).
b) If "Signup confirmation" is disabled, we sync the customer as "subscribed" if the checkbox is checked.

PR #2693, opened 2020-04-01
Remove paused sending [MAILPOET-2643]

pavel-mailpoet:
[MAILPOET-2643]

[MAILPOET-2643]: https://mailpoet.atlassian.net/browse/MAILPOET-2643


lib/Config/ServicesChecker.php
```diff
@@ -33,12 +33,15 @@ public function isMailPoetAPIKeyValid($displayErrorNotice = true, $forceCheck =
...
-          ['target' => '_blank']
-        );
-        WPNotice::displayError($error);
+        $error = '<h3>' . __('All sending is currently paused!', 'mailpoet') . '</h3>';
+        $error .= '<p>' . __('Your key to send with MailPoet is invalid.', 'mailpoet') . '</p>';
+        $error .= '<p><a '
+          . ' href="https://account.mailpoet.com?s=' . (Subscriber::getTotalSubscribers() + 1) . '"'
+          . ' class="button button-secondary" '
+          . ' target="_blank"'
+          . '>' . __('Visit MailPoet.com to purchase a key') . '</a></p>';
```
costasovo, line 42: I think this function call is missing the second argument 'mailpoet'.

pavel-mailpoet: I thought phpstan can catch such errors.

costasovo: The second argument is optional. I guess we have to pass the correct domain https://mailpoet.slack.com/archives/G0A1L2KHV/p1571673372003200

lib/Config/Menu.php
```diff
@@ -554,8 +554,15 @@ public static function errorPageCallback() {
...
     if (self::isOnMailPoetAdminPage()) {
+      $showNotices = isset($_REQUEST['page'])
+        && (
+          stripos($_REQUEST['page'], self::MAIN_PAGE_SLUG) !== false
+          || stripos($_REQUEST['page'], 'mailpoet-segments') !== false
+          || stripos($_REQUEST['page'], 'mailpoet-subscribers') !== false
+      );
       $checker = $checker ?: $this->servicesChecker;
-      $this->mpApiKeyValid = $checker->isMailPoetAPIKeyValid();
+      $checker = $checker ?: $this->servicesChecker;
```
costasovo: This line of code is duplicated. Probably accidentally copy-pasted?

lib/AdminPages/Pages/Segments.php
```diff
@@ -32,6 +49,15 @@ public function render() {
     $data['has_valid_api_key'] = $this->subscribersFeature->hasValidApiKey();
     $data['subscriber_count'] = Subscriber::getTotalSubscribers();
 
+    $data['mss_key_invalid'] = false;
+
+    if ($this->servicesChecker->isMailPoetAPIKeyValid() === false) {
+      $mssKey = $this->settings->get(Bridge::API_KEY_STATE_SETTING_NAME);
+      if (isset($mssKey['state']) && ($mssKey['state'] === Bridge::KEY_INVALID)) {
+        $data['mss_key_invalid'] = true;
+      }
```
costasovo, line 58: I see these lines repeated in 2 places: here and in SubscribersPage. What about adding a method to servicesChecker ?

lib/Config/Menu.php
costasovo: I see that this was deleted in a later commit.

PR #2695, opened 2020-04-02
Unauthorized notices [MAILPOET-2805]

JanJakes:
[MAILPOET-2805]

[MAILPOET-2805]: https://mailpoet.atlassian.net/browse/MAILPOET-2805

### DIFF
--- a/assets/js/src/common/set_from_address_modal.tsx
+++ b/assets/js/src/common/set_from_address_modal.tsx
@@ -58,6 +58,10 @@ const removeUnauthorizedEmailNotices = () => {
   if (unauthorizedEmailInNewsletterNotice) {
     unauthorizedEmailInNewsletterNotice.remove();
   }
+  const unauthorizedEmailInNewsletterDynamicNotice = document.querySelector('[data-id="mailpoet_authorization_error"]');
+  if (unauthorizedEmailInNewsletterDynamicNotice) {
+    unauthorizedEmailInNewsletterDynamicNotice.remove();
+  }
 };
 
 type Props = {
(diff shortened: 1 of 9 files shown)


JanJakes:
Thanks for the inputs, @veljkho. I don't think these are actual issues. Let me explain:

1. This ticket is not meant to change the conditions when the notice appears. It changes it's content only. It doesn't appear only after clicking on "Verify" - when some sending is active and the emails get refused by MSS, it will appear. It's as you say (_nothing will trigger warning until someone purchase something or sign up for newsletter_) and that's correct.

2. This is not wrong, this is the desired behavior. It you look at the ticket, it says that it should use exactly the same modal as in settings, which it does. Also, it would be very confusing for users to have to fill in the exactly the same modal window twice - I think they wouldn't understand why they're doing it twice.

veljkho:
Hi @JanJakes,
Thanks for clarification. Regarding (2.) problem, we can agree to leave it as is, so will update all From fields.

But for the (1.) one, I think there are two problems actually..
1. First one is about not showing unauthorized notice when there are unauthorized emails set in automatic emails.. when I want to trigger them by purchasing some item, the notice wouldn't appear
**+** I receive all the time automatic mailpoet emails but just for the first time purchasing (all wc emails), and further will receive only transactional wc emails and not automatic emails.
**Note**: To be more complex, this will lead to another new bug which I'm going to share on Slack about sending newsletter with any Sender email address.

2. Second one is related to the green notice after fixing.. the green notice is also present after fixing automatic emails, but it should'nt, because there is no "b) The green admin notice.." spec in Jira for the "A. For the automatic emails, change the following".

Please let me know if you'd like to try all of these on my environment on stage.qawp.net, and if you need any steps to reproduce help on Slack.

PR #2705, opened 2020-04-07
Avoid entity not found errors [MAILPOET-2818]

costasovo:
[MAILPOET-2818]

[MAILPOET-2818]: https://mailpoet.atlassian.net/browse/MAILPOET-2818

### DIFF
--- a/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php
+++ b/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php
@@ -181,7 +181,9 @@ class NewslettersResponseBuilder {
 
   private function buildQueue(SendingQueueEntity $queue) {
     $task = $queue->getTask();
-
+    if ($task === null) {
+      return null;
+    }
     // the following crazy mix of '$queue' and '$task' comes from 'array_merge($task, $queue)'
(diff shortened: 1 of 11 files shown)


lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php
```diff
@@ -198,7 +198,7 @@ private function buildQueue(SendingQueueEntity $queue) {
       'deleted_at' => ($deletedAt = $queue->getDeletedAt()) ? $deletedAt->format(self::DATE_FORMAT) : null,
       'meta' => $queue->getMeta(),
       'task_id' => (string)$task->getId(), // (string) for BC
-      'newsletter_id' => (string)$queue->getNewsletter()->getId(), // (string) for BC
+      'newsletter_id' => ($queue->getNewsletter() instanceof NewsletterEntity) ? (string)$queue->getNewsletter()->getId() : null, // (string) for BC
```
JanJakes, line 201: How is it possible PHPStan doesn't scream? You shouldn't call the function twice - PHPStan doesn't know if there are any side-effects.

Correct way should be:
```
($newsletter = $queue->getNewsletter()) ? (string)$newsletter->getId() : null;
```

PR #2706, opened 2020-04-07
Settings send with [MAILPOET-2679]

amine-mp:
[MAILPOET-2679]

[MAILPOET-2679]: https://mailpoet.atlassian.net/browse/MAILPOET-2679


assets/js/src/settings/pages/send_with/send_with.tsx
```diff
@@ -0,0 +1,146 @@
...
+                })}
+              >
+                <span className="dashicons dashicons-format-video" />
+                {t('seeVideo')}
+              </a>
+            </p>
+          </div>
+        </div>
+        <div className={classnames('mailpoet_status', {
+          mailpoet_invalid_key: !isMssActive || !isMssKeyValid,
```
costasovo, line 77: I think this should be `isMssActive && !isMssKeyValid`. The div  should have `mailpoet_invalid_key` class only when MSS is activated and the key is invalid. 

I also noticed a difference in CSS.
This is how `mailpoet_invalid_key` renders in old settings: https://d.pr/i/1lI9EF
This is new settings: https://d.pr/i/FQyw63

amine-mp: I agree that this condition looks wrong, but it's the equivalent of what we had before
https://github.com/mailpoet/mailpoet/blob/3.46.8/views/settings/mta.html#L94
```
<div class="mailpoet_status <% if(settings.mta_group != 'mailpoet' or not mss_key_valid) %> mailpoet_invalid_key <% endif %>">
```
`settings.mta_group != 'mailpoet'` is `!isMssActive` and `not mss_key_valid` is `!isMssKeyValid`.

For the difference in CSS, I think you are mixing the case when MSS is active vs when another method is active.
When MSS is active and the key is invalid, it should have the red border. When another method is active and the key is invalid, the border is grey.

costasovo: Ahh. I see now. It works as expected. Sorry. The condition really confused me.

costasovo:
@amine Approved. Needs a rebase though.

PR #2707, opened 2020-04-08
Pending approval warning [MAILPOET-2806]

JanJakes:
[MAILPOET-2806]

[MAILPOET-2806]: https://mailpoet.atlassian.net/browse/MAILPOET-2806

### DIFF
--- a/lib/API/JSON/v1/Services.php
+++ b/lib/API/JSON/v1/Services.php
@@ -102,9 +102,10 @@ class Services extends APIEndpoint {
     }
 
     // pause sending when key is pending approval, resume when not pending anymore
-    if ($this->servicesChecker->isMailPoetAPIKeyPendingApproval()) {
+    $isPendingApproval = $this->servicesChecker->isMailPoetAPIKeyPendingApproval();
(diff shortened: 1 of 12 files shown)


lib/API/JSON/v1/Services.php
```diff
@@ -85,6 +101,13 @@ public function checkMSSKey($data = []) {
       ]);
     }
 
+    // pause sending when key is pending approval, resume when not pending anymore
+    if ($this->servicesChecker->isMailPoetAPIKeyPendingApproval()) {
```
amine-mp, line 105: I think we can avoid pausing sending multiple times by only pausing if the key was not pending approval before. What do you think?

JanJakes:
@veljkho Thanks for the issues.
a. This is on purpose, looks better IMO. Don't you think?
b. This is expected - clicking on "Resume sending" will still try to resume sending. The only thing is, that later, MSS will refuse to send the emails in case of `is_approved = false`. But since there's no `is_approved = false` on the MSS side yet, it works now. Does that make sense?

PR #2710, opened 2020-04-09
Add request feedback link to the footer [MAILPOET-2802]

pavel-mailpoet:
[MAILPOET-2802]

[MAILPOET-2802]: https://mailpoet.atlassian.net/browse/MAILPOET-2802

### DIFF
--- a/lib/Config/Hooks.php
+++ b/lib/Config/Hooks.php
@@ -104,6 +104,7 @@ class Hooks {
     $this->setupPostNotifications();
     $this->setupWooCommerceSettings();
     $this->dynamicSegmentHooks->init();
+    $this->setupFooter();
   }
 
   public function initEarlyHooks() {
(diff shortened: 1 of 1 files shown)


lib/Config/Hooks.php
```diff
@@ -365,4 +366,19 @@ public function setupPostNotifications() {
...
+      [$this, 'setFooter'],
+      1, 1
+    );
+  }
+
+  public function setFooter($text) {
+    if (!Menu::isOnMailPoetAdminPage()) {
+      return $text;
+    }
+    return '<a href="https://feedback.mailpoet.com/" target="_blank">Give feedback</a>';
```
JanJakes, line 382: I guess we should add also `noopener noreferrer`, right?

When at it, do you think adding the check for `Menu::isOnMailPoetAdminPage()` directly to `setupFooter` is safe or could there be any problem? Just thinking we wouldn't need to register the hook on non-MP pages at all.

pavel-mailpoet: you are right, I'll make the changes. 

- I didn't add  `noopener noreferrer` because the target is our domain so it is not really necessary here
- the code is inspired by woocommerce that does it this way as well, but I can move the check up, no problem.

PR #2711, opened 2020-04-09
Fix props warnings on templates page [MAILPOET-2816]

costasovo:
[MAILPOET-2816]

[MAILPOET-2816]: https://mailpoet.atlassian.net/browse/MAILPOET-2816


amine-mp:
Good job on fixing this component, but we should also to fix the component that's calling it to use the right types/values. Here are some examples:
- `readonly` needs to be a boolean here and the `id` is missing https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/newsletters/templates.jsx#L83
- No need to parse the `id` as integer if it's already an integer https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/newsletters/templates.jsx#L132

costasovo:
@amine-mp Good catch! I added one more commit. I didn't find more places than the examples you mentioned.

PR #2713, opened 2020-04-09
Fixed bar form [MAILPOET-2740]

pavel-mailpoet:
[MAILPOET-2740]

[MAILPOET-2740]: https://mailpoet.atlassian.net/browse/MAILPOET-2740


assets/css/src/components-public/_public.scss
```diff
@@ -89,6 +89,12 @@ $form-columns-space-between: 20px;
   .mailpoet_form_columns .mailpoet_form_column .mailpoet_paragraph:last-child {
     margin-bottom: 0;
   }
+
+  @include breakpoint-max-width($form-break-small-width) {
```
costasovo, line 93: I think it would be better instead of adding this new rule to modify the one above to set 0 bottom margin only for displays bigger than mobile. Because now we have 3 rules for the bottom margin:
1) adds margin
2) removes margin
3) add margin again
and every rule has a higher specificity. So I guess 2 rules would be nicer for future modifications.

assets/js/src/public.jsx
```diff
@@ -70,25 +70,34 @@ jQuery(($) => {
...
-    $('.mailpoet_popup_close_icon').click((event) => {
+    $('.mailpoet_form_close_icon').click((event) => {
       const closeIcon = $(event.target);
       const formDiv = closeIcon.parent();
-      closePopupForm(formDiv);
+      closeForm(formDiv);
     });
 
     $('div.mailpoet_form_fixed_bar').each((index, element) => {
+      const cookieValue = getCookie('popup_form_dismissed');
```
costasovo: Shouldn't we distinguish between `popup_form_dismissed` and `fixed_bar_form_dismissed`?  I still think that adding form id into the cookie would be the best option.

PR #2723, opened 2020-04-14
Slide in form [MAILPOET-2742]

pavel-mailpoet:
[MAILPOET-2742]

[MAILPOET-2742]: https://mailpoet.atlassian.net/browse/MAILPOET-2742


lib/Form/DisplayFormInWPContent.php
```diff
@@ -103,15 +100,28 @@ private function saveNoForms() {
...
+   * @return array<string, FormEntity>
    */
   private function getForms(): array {
     $forms = $this->formsRepository->findBy(['deletedAt' => null]);
-    return array_filter($forms, function($form) {
-      return (
-        $this->shouldDisplayForm($form)
-      );
-    });
+    $forms = $this->filterOneFormInEachDisplayType($forms);
```
costasovo, line 107: I think you need to add some order here to ensure that the latest updated will be selected.

PR #2725, opened 2020-04-14
Congratulatory email [MAILPOET-2831]

JanJakes:
[MAILPOET-2831]

[MAILPOET-2831]: https://mailpoet.atlassian.net/browse/MAILPOET-2831

### DIFF
--- a/views/emails/congratulatoryMssEmail.html
+++ b/views/emails/congratulatoryMssEmail.html
@@ -1,0 +1,155 @@
+<html lang="en" style="margin:0;padding:0">
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+  <meta name="format-detection" content="telephone=no" />
+  <title>Subject</title>
+  <style type="text/css"> @media screen and (max-width: 480px) {
+    .mailpoet_button {width:100% !important;}
+  }
+  @media screen and (max-width: 599px) {
+    .mailpoet_header {
+      padding: 10px 20px;
+    }
+    .mailpoet_button {
+      width: 100% !important;
+      padding: 5px 0 !important;
+      box-sizing:border-box !important;
+    }
+    div, .mailpoet_cols-two, .mailpoet_cols-three {
+      max-width: 100% !important;
+    }
+  }
+  </style>
+
+</head>
+<body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" style="margin:0;padding:0;background-color:#f0f0f0">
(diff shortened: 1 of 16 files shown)


views/emails/congratulatoryMssEmail.html
```diff
@@ -0,0 +1,155 @@
...
+    }
+    .mailpoet_button {
+      width: 100% !important;
+      padding: 5px 0 !important;
+      box-sizing:border-box !important;
+    }
+    div, .mailpoet_cols-two, .mailpoet_cols-three {
+      max-width: 100% !important;
+    }
+  }
```
costasovo, line 23: Some of those classes are not used in the content: `.mailpoet_button`, `.mailpoet_cols-three`.  
It seems to be the default styles we use for all emails. I see the same in https://github.com/mailpoet/mailpoet/blob/master/views/emails/statsNotificationLayout.html#L8
So I guess we could somehow include those or maybe even create a common layout for all fancy HTML email. What do you think?

views/emails/congratulatoryMssEmail.html
```diff
@@ -0,0 +1,155 @@
+<html lang="en" style="margin:0;padding:0">
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+  <meta name="format-detection" content="telephone=no" />
+  <title>Subject</title>
```
costasovo, line 7: I guess the title should be different.

JanJakes:
@veljkho Thanks for the detailed review 👍 Looking into those and I'll see what we can/should fix now and what deserves a ticket.

JanJakes:
@veljkho 

> I experienced new not related issue in this PR: when you click to activate Other sending method, MSS is still in green as activated, and when you reload the page it will show you Other as activated (this is just a visual bug)

I managed to replicate this about twice but then couldn't anymore. Is it possible this happens randomly (sometimes yes, sometimes no)? I'd recommend checking it in https://github.com/mailpoet/mailpoet/pull/2714, it's likely it's fixed in there since those subpages should already be rewritten in that PR.

> Another like not related to changes bug in this PR: when you have everything set up, go in Basics tab and add some unauthorized email as From address, then just click Key Activation tab, it will automatically show you a popup to change From address (we may leave this as is since it is not that bad to show this to user with unauthorized email).

Good catch! Adding a fix.

> This one may be related: When I paste the key which extended plan limit and is not anymore useful for sending with MSS until upgraded, I can see the message and I received MSS cogratulatory email (but this may be okay, since this is just a test email and edgecase).

You're saying "I can see the message", what message do you mean? A success message or an error message? In general, if the key is valid for MSS and MSS is being activated, we should send the email, the limit is not an issue here.

PR #2732, opened 2020-04-15
Prepare Storybook.js [MAILPOET-2769]

NeosinneR:
[MAILPOET-2769]

[MAILPOET-2769]: https://mailpoet.atlassian.net/browse/MAILPOET-2769


.storybook/main.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  stories: ['../assets/js/src/**/_stories/*.js'],
```
costasovo: Shouldn't we rather use typescript for stories? It is a new feature so I guess we could even force TS for this. We will probably write components in TS as well.
This will probably require some additional configuration https://storybook.js.org/docs/configurations/typescript-config/ but think we need it.

NeosinneR:
@costasovo changed stories to `.tsx` and added the addon you suggested. However, Storybook throws errors when TS is configured to also process `.js` files, so I had to disable it (at least temporary) in `.tsconfig`. I couldn't get the Storyboard to use specific TS config, so that's why this temporary fix (the last commit).

NeosinneR:
@costasovo I used a previous version of `@storybook/preset-typescript`, which was still using `ts-loader` and special `tsconfig` could be used. Now watching and building is working correctly.

PR #2733, opened 2020-04-16
Merge Lists and Segments pages [MAILPOET-2480]

pavel-mailpoet:
[MAILPOET-2480]

[MAILPOET-2480]: https://mailpoet.atlassian.net/browse/MAILPOET-2480


assets/js/src/segments/dynamic_segments_list.jsx
```diff
@@ -116,25 +117,23 @@ function renderItem(item, actions) {
...
-          data-automation-id="new-segment"
-        >
-          {MailPoet.I18n.t('new')}
-        </Link>
+        <Link className="page-title-action" to="/new">{MailPoet.I18n.t('new')}</Link>
+        <Link className="page-title-action" to="/new-segment" data-automation-id="new-segment">{MailPoet.I18n.t('newSegment')}</Link>
       </h1>
 
+      <Tabs />
+
```
costasovo, line 128: This code fragment is very similar to the header in segments/list.jsx. I noticed that header for default lists contains also SubscribersLimitNotice and InvalidMssKeyNotice components. Shouldn't we display these notices also for dynamic segments listing? Actually maybe we could create a component <SegmentsListingHeader title={xxx} /> containing title, notices, new buttons and tabs and then reuse this one for both tabs. What do you think?

pavel-mailpoet: The specification of the `InvalidMssKeyNotice` only specified to show it on List page and not on Segments page. I had to re-do it when I displayed it on Segments page https://mailpoet.atlassian.net/browse/MAILPOET-2643

The other notice wasn't on the Segments page as well, so I am not sure why would we add it here...

Do you think they need to be displayed on both tabs?

costasovo: Ahh. Thanks for explaining. I didn't know we didn't want to show those on segments page. It felt like now when the pages are merged we could keep it for both and remove some code duplicity. But if we need to hide it for segments I guess it is fine as is.

PR #2734, opened 2020-04-16
Update swiftmailer/swiftmailer dependencies [MAILPOET-2863]

costasovo:
[MAILPOET-2863]

[MAILPOET-2863]: https://mailpoet.atlassian.net/browse/MAILPOET-2863


pavel-mailpoet:
@veljkho this PR is updating the library that is used for sending SMTP emails so when testing please check it is still working.

PR #2739, opened 2020-04-21
Form styling 2 [MAILPOET-2809]

pavel-mailpoet:
[MAILPOET-2809]

[MAILPOET-2809]: https://mailpoet.atlassian.net/browse/MAILPOET-2809


assets/js/src/form_editor/blocks/submit/edit.jsx
```diff
@@ -31,7 +31,7 @@ const SubmitEdit = ({ attributes, setAttributes }) => {
 
   const styles = !attributes.styles.inheritFromTheme ? {
     fontWeight: attributes.styles.bold ? 'bold' : 'inherit',
-    borderRadius: attributes.styles.borderRadius ? `${attributes.styles.borderRadius}px` : 0,
```
costasovo, line 34: Shouldn't we make it consistent the other way around? I mean we should probably allow a user to set 0 border-radius.

assets/js/src/form_editor/blocks/submit/edit.jsx
```diff
@@ -37,7 +37,7 @@ const SubmitEdit = ({ attributes, setAttributes }) => {
 
   const styles = !attributes.styles.inheritFromTheme ? {
     fontWeight: attributes.styles.bold ? 'bold' : 'inherit',
-    borderRadius: attributes.styles.borderRadius !== undefined ? `${attributes.styles.borderRadius}px` : 1,
+    borderRadius: attributes.styles.borderRadius !== undefined ? `${attributes.styles.borderRadius}px` : 0,
```
costasovo, line 40: With this change, you reintroduced the inconsistency (0 is rendered but 1 is displayed in settings) in settings and rendering. We need to update [the value in submit settings component](https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/form_editor/blocks/submit/styles_settings.tsx#L94) to use zero as well.

assets/js/src/form_editor/store/map_form_data_after_loading.jsx
```diff
@@ -25,6 +25,10 @@ export default function mapFormDataAfterLoading(data) {
       placeSlideInFormOnAllPosts: data.settings.place_slide_in_form_on_all_posts === '1',
       slideInFormDelay: asNum(data.settings.slide_in_form_delay),
       slideInFormPosition: data.settings.slide_in_form_position,
+      borderRadius: asNum(data.settings.borderRadius),
```
costasovo, line 28: Maybe I missed that in the first review. Do we save those properties on API as underscore or camelcase? I think it should be as underscore so that it is consistent with other properties.

assets/js/src/form_editor/blocks/convert_alignment_to_margin.ts
```diff
@@ -0,0 +1,12 @@
+
+function convert(alignment: 'left' | 'center' | 'right'): string {
```
costasovo, line 2: Just an idea. What about using [typescript string enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) for the alignment?

pavel-mailpoet:
@veljkho I had to change the way the properties are stored in the database so if you have an older form created in this branch you need to delete it and create a new one for testing.

PR #2744, opened 2020-04-23
Display number of subscribers in plan [MAILPOET-2832]

pavel-mailpoet:
[MAILPOET-2832]

[MAILPOET-2832]: https://mailpoet.atlassian.net/browse/MAILPOET-2832

### DIFF
--- a/assets/js/src/common/subscribers_in_plan.tsx
+++ b/assets/js/src/common/subscribers_in_plan.tsx
@@ -7,7 +7,7 @@ type Props = {
   hasValidApiKey: boolean,
 };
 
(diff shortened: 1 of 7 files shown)


NeosinneR:
@costasovo @pavel-mailpoet now I see, how it could be understood differently :-( 

The number should show an actual number of all subscribed, unconfirmed and inactive subscribers, not the maximum number according to a plan. We are adding this number because `All (<count>)` is showing also unsubscribed and bounced, which are not counted towards the plan limit. 

With this number, the user can without a doubt see, how many subscribers are counted towards the limit and if he needs a larger plan. 

And as for when it's visible - when MSS is active. In which case the user has to have a plan.

PR #2752, opened 2020-04-29
Form Editor: extend Spacer / Divider [MAILPOET-2835]

pavel-mailpoet:
[MAILPOET-2835]

[MAILPOET-2835]: https://mailpoet.atlassian.net/browse/MAILPOET-2835


lib/Form/Block/Divider.php
```diff
@@ -3,8 +3,41 @@
...
+    $dividerHeight = $block['params']['divider_height'] ?? self::DEFAULT_ATTRIBUTES['dividerHeight'];
+    $color = $block['params']['color'] ?? self::DEFAULT_ATTRIBUTES['color'];
+    $height = $block['params']['height'] ?? self::DEFAULT_ATTRIBUTES['height'];
+
+    $dividerStyles = [
+      "border-top-style: $style",
+      "border-top-width: {$dividerHeight}px",
+      "border-top-color: $color",
+      "height: {$height}px",
+      "width: $width%",
```
costasovo, line 38: This renders the divider with some height differently than in form editor.
In the form editor, the divider is vertically centred https://d.pr/i/7cj6hO 
On the front end, the divider is displayed on top https://d.pr/i/FkVR7c

I see that this behaviour was not specified in spec. I would personally expect the behaviour we have in form editor - vertically centred.

PR #2754, opened 2020-04-29
Redesign - typography [MAILPOET-2770]

NeosinneR:
[MAILPOET-2770]

[MAILPOET-2770]: https://mailpoet.atlassian.net/browse/MAILPOET-2770


assets/css/src/components-plugin/_breadcrumb.scss
```diff
@@ -9,10 +9,10 @@
...
 
 .mailpoet_breadcrumb a {
-  color: #444;
+  color: $color-text;
   text-decoration: none;
 }
 
 .mailpoet_breadcrumb a:hover {
-  color: darken(#444, 50%);
+  color: darken($color-text, 50%);
```
pavel-mailpoet, line 17: I wonder if we should keep all the colours in the settings and use variables everywhere else. If we use functions like `darken` only in settings we will have ALL colours defined in one place. If we spread them everywhere we will have no idea how many colours we actually use.

assets/css/src/generic/_typography.scss
```diff
@@ -15,3 +15,25 @@ a {
...
+
+h1,
+h2,
+h3,
+h4,
+h5 {
+  color: $color-text;
+  font-weight: bold;
+  line-height: 1.25;
+  margin: 0 0 20px 0;
```
pavel-mailpoet, line 24: Our colours definitions are in the settings file. Should we do the same with dimensions? I like how gutenberg does that: https://github.com/WordPress/gutenberg/blob/master/packages/base-styles/_variables.scss

assets/css/src/generic/_typography.scss
```diff
@@ -15,3 +15,25 @@ a {
...
+}
+
+h4,
+h5 {
+  margin: 0 0 10px 0;
+}
+
+h1 { font-size: 32px; }
```
pavel-mailpoet, line 32: Don't we need to scale down the heading sizes on smaller devices?

assets/js/src/storybook/_stories/header.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+export default {
+  title: 'Typography',
+};
+
+export const Headers = () => (
+  <>
+    <h1>&lt;h1&gt; header</h1>
```
pavel-mailpoet, line 9: I thought the idea was to create react components for the elements so that we would use: `<Heading level={1}>Header</Heading>` or something like that. Again Gutenberg does that https://developer.wordpress.org/block-editor/components/text/

assets/css/src/generic/_typography.scss
NeosinneR: No, the sizes stay the same

PR #2755, opened 2020-04-29
Add confirm unsubscribe page [MAILPOET-2736]

costasovo:
[MAILPOET-2736]

**Note for QA:**
Old unsubscribe links from newsletters sent prior upgrading to this version should unsubscribe a subscriber just by opening the link.

[MAILPOET-2736]: https://mailpoet.atlassian.net/browse/MAILPOET-2736

### DIFF
--- a/lib/Newsletter/Shortcodes/Categories/Link.php
+++ b/lib/Newsletter/Shortcodes/Categories/Link.php
@@ -24,6 +24,14 @@ class Link {
       case 'subscription_unsubscribe_url':
         return self::processUrl(
           $shortcodeDetails['action'],
+          $subscriptionUrlFactory->getConfirmUnsubscribeUrl($wpUserPreview ? null : $subscriber),
+          $queue,
+          $wpUserPreview
+        );
+
+      case 'subscription_instant_unsubscribe_url':
+        return self::processUrl(
+          $shortcodeDetails['action'],
           $subscriptionUrlFactory->getUnsubscribeUrl($wpUserPreview ? null : $subscriber),
           $queue,
           $wpUserPreview
@@ -76,12 +84,11 @@ class Link {
     $subscriptionUrlFactory = SubscriptionUrlFactory::getInstance();
     switch ($shortcodeAction) {
       case 'subscription_unsubscribe_url':
-        $settings = SettingsController::getInstance();
-        // track unsubscribe event
-        if ((boolean)$settings->get('tracking.enabled') && !$wpUserPreview) {
-          $unsubscribeEvent = new Unsubscribes();
-          $unsubscribeEvent->track($newsletter->id, $subscriber->id, $queue->id);
-        }
+        self::trackUnsubscribe($newsletter, $subscriber, $queue, $wpUserPreview);
+        $url = $subscriptionUrlFactory->getConfirmUnsubscribeUrl($subscriber);
+        break;
+      case 'subscription_instant_unsubscribe_url':
+        self::trackUnsubscribe($newsletter, $subscriber, $queue, $wpUserPreview);
         $url = $subscriptionUrlFactory->getUnsubscribeUrl($subscriber);
         break;
       case 'subscription_manage_url':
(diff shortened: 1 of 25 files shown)


tests/acceptance/ManageSubscriptionLinkCest.php
```diff
@@ -62,12 +62,42 @@ public function manageSubscriptionLink(\AcceptanceTester $i) {
...
+
+  private function verifyUnsubscribeLinks(\AcceptanceTester $i) {
+    $this->sendEmail($i);
     $formStatusElement = '[data-automation-id="form_status"]';
+    $i->wantTo('Verify that "Unsubscribe List" header link works and subscriber status is set to unsubscribed instantly');
+    $i->amOnMailboxAppPage();
+    $i->click(Locator::contains('span.subject', $this->newsletterTitle));
+    $i->click('#show-headers');
+    $i->waitForText('List-Unsubscribe');
+    $link = $i->grabTextFrom('.headers tbody tr:nth-child(4) td');
```
JanJakes, line 85: This looks quite flaky. You could select the `td` that is just after `td` with text `List-Unsubscribe` (with XPath it should be possible).

Another option - grab full message source (on the source tab) and parse it out using a regex.

tests/acceptance/ManageSubscriptionLinkCest.php
```diff
@@ -62,12 +62,42 @@ public function manageSubscriptionLink(\AcceptanceTester $i) {
     $i->seeNoJSErrors();
   }
 
-  public function unsubscribeLink(\AcceptanceTester $i) {
-    $i->wantTo('Verify that "unsubscribe" link works and subscriber status is set to unsubscribed');
-    $this->sendEmail($i);
+  public function unsubscribeLinksWithLinkTracking(\AcceptanceTester $i) {
```
JanJakes, line 65: This makes me think - do we still track unsubscribe as a "click" and do we still want to?

lib/Subscription/Pages.php
```diff
@@ -307,6 +320,12 @@ private function getUnsubscribeTitle() {
     }
   }
 
+  private function getConfirmUnsubscribeTitle() {
+    if ($this->isPreview() || $this->subscriber !== false) {
+      return $this->wp->__("Confirm you want to unsubscribe.", 'mailpoet');
```
JanJakes, line 325: There should be no `.` at the end of the title.

lib/Newsletter/Shortcodes/Categories/Link.php
```diff
@@ -76,12 +84,11 @@ public static function processShortcodeAction(
     $subscriptionUrlFactory = SubscriptionUrlFactory::getInstance();
     switch ($shortcodeAction) {
       case 'subscription_unsubscribe_url':
-        $settings = SettingsController::getInstance();
-        // track unsubscribe event
-        if ((boolean)$settings->get('tracking.enabled') && !$wpUserPreview) {
-          $unsubscribeEvent = new Unsubscribes();
-          $unsubscribeEvent->track($newsletter->id, $subscriber->id, $queue->id);
-        }
+        self::trackUnsubscribe($newsletter, $subscriber, $queue, $wpUserPreview);
```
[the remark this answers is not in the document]
costasovo: Very good catch. It is a mistake.

tests/acceptance/ManageSubscriptionLinkCest.php
costasovo: Very good question. We track it. I haven't thought about removing the tracking. It makes sense to remove it or maybe at least filter these clicks from stats because unsubscribes are not kind of clicks user want to get. Not sure if removing it has some other consequences.

costasovo:
@JanJakes Fixed. I had to refactor unsubscribe tracking since it was done during shortcode processing.

>  do we still track unsubscribe as a "click"
We still track it. After this refactor it should be easier to remove this click tracking because unsubscribes tracking is no longer dependent on data passed in click tracking URL.

PR #2764, opened 2020-05-04
Redesign - buttons [MAILPOET-2771]

NeosinneR:
[MAILPOET-2771]

[MAILPOET-2771]: https://mailpoet.atlassian.net/browse/MAILPOET-2771


assets/js/src/common/button/button.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import classnames from 'classnames';
+
+type Props = {
+  children: React.ReactNode,
+  size?: 'small' | 'large',
+  variant?: 'light' | 'dark' | 'link' | 'link-dark',
+  isLoading?: boolean,
```
pavel-mailpoet, line 8: I think this is not a great name. The button is not loading anything. The button has a spinner, so maybe a better name would be `withSpinner`

The same applies to the class names.

assets/js/src/common/button/button.tsx
```diff
@@ -0,0 +1,42 @@
...
+  size,
+  variant,
+  isLoading,
+  isFullWidth,
+  onClick,
+  href,
+}: Props) => (
+  <a
+    href={href}
+    onClick={onClick}
```
pavel-mailpoet, line 29: I think the button should allow disabling the action. For example if the button has spinner this action shouldn't be triggered. Maybe in some other cases, it would be useful to disable `onClick`

NeosinneR: Now, that I'm thinking about it, there should also be `isDisabled` parameter and actions should also be disabled in this state.

But, `href` and `onClick` are actually disabled when there is a spinner - via CSS `pointer-events: none;`. Do you think this is enough (I'd add a comment to those parameters) or should I add a new parameter (e.g. `disableActions`) and only disable them when this is true?

pavel-mailpoet: > Do you think this is enough

I am not sure. I cannot foresee all the consequences. Certainly, comments are useful for developers.

`isDisabled` is also good.

NeosinneR: Added `isDisabled` parameter with a comment that it also disables `href` and `onClick` (the same for `withSpinner`). In the future, if we need more control when a button shouldn't be actionable, we can add a new parameter, but I think this is good enough for now.

PR #2774, opened 2020-05-07
Change default form spacing [MAILPOET-2869]

wxa:


lib/Form/Util/Styles.php
```diff
@@ -10,6 +10,9 @@ class Styles {
 .mailpoet_form {
   padding: 20px;
 }
+.mailpoet_form.mailpoet_form_slide_in {
+  max-width: 350px;
```
pavel-mailpoet, line 14: You are using custom CSS for those changes. Rosta and I have been adding our changes to `public.scss` and I am not sure which is better. I feel like the `public.scss` is slightly better because we can later change it or add settings to control those values. If a user saves a form with this custom CSS it would be more difficult for us to override those values.

lib/Form/Util/Styles.php
```diff
@@ -74,6 +74,10 @@ class Styles {
 .mailpoet_message {
 }
 
+.mailpoet_form_image {
+  margin-bottom: 15px;
```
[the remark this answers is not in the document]
wxa: Fixed, now it should be only applied to the last image in a stacked column.

lib/Form/Util/Styles.php
wxa: Sure I can move this to `public.scss`, which I did, increasing the range between `min-width` and `max-width` a bit. But I still have some doubts: this probably needs to be somewhat customizable, the only way is to set the `width` custom CSS property which will still make it harder to override. Then, it may still be overridden by a theme. Therefore it's still not a perfect solution, but I have no better idea.

costasovo: I'm currently working on form width settings https://mailpoet.atlassian.net/browse/MAILPOET-2811
In this feature, each form will have defined width and user will be able to set their own.

lib/Form/Util/Styles.php
```diff
@@ -8,13 +8,16 @@ class Styles {
...
 .mailpoet_form {
-  padding: 10px;
+  padding: 20px;
 }
 
 /* columns */
 .mailpoet_column_with_background {
   padding: 10px;
 }
+.mailpoet_form_columns.mailpoet_paragraph {
```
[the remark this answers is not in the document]
wxa: I can't apply the fixed bar fix as-is, because the class structure of a slide-in form in the Jira task (spec A, URL is in the screenshot) is different: `.mailpoet_form_slide_in` is applied to both `div` and `form`, `.mailpoet_paragraph` is applied to `.mailpoet_form_columns`, there is no `.mailpoet_paragraph` inside `.mailpoet_form_column` or in `p` tags. So I have to modify this a bit. You're welcome to check out.

PR #2777, opened 2020-05-11
Optimize produtcs loading on segments page [MAILPOET-2904]

costasovo:
[MAILPOET-2904]

[MAILPOET-2904]: https://mailpoet.atlassian.net/browse/MAILPOET-2904

### DIFF
--- a/lib/AdminPages/Pages/Segments.php
+++ b/lib/AdminPages/Pages/Segments.php
@@ -83,8 +83,10 @@ class Segments {
   }
 
   private function getProducts() {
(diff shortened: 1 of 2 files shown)


lib/AdminPages/Pages/Segments.php
```diff
@@ -83,8 +83,10 @@ public function render() {
   }
 
   private function getProducts() {
-    $args = ['post_type' => 'product', 'orderby' => 'title', 'order' => 'ASC', 'numberposts' => -1];
-    $products = $this->wp->getPosts($args);
+    $products = $this->wp->getResultsFromWpDb(
```
pavel-mailpoet, line 86: The order here is missing

PR #2778, opened 2020-05-11
Refactor statistics API to doctrine [PREMIUM-142]

pavel-mailpoet:


lib/Entities/StatisticsWooCommercePurchaseEntity.php
```diff
@@ -104,4 +104,11 @@ public function getClick() {
     $this->safelyLoadToOneAssociation('click');
     return $this->click;
   }
+
+  /**
+   * @return int
+   */
+  public function getOrderId() {
```
JanJakes, line 111: This could have a PHP 7 return typehint.

lib/Entities/NewsletterLinkEntity.php
```diff
@@ -22,20 +22,20 @@ class NewsletterLinkEntity {
...
    * @ORM\JoinColumn(name="newsletter_id", referencedColumnName="id")
-   * @var NewsletterEntity|null
+   * @var NewsletterEntity
    */
   private $newsletter;
 
   /**
    * @ORM\ManyToOne(targetEntity="MailPoet\Entities\SendingQueueEntity")
    * @ORM\JoinColumn(name="queue_id", referencedColumnName="id")
-   * @var SendingQueueEntity|null
```
JanJakes, line 32: Those two still can be `null` due to possible data inconsistencies (`safelyLoadToOneAssociation`). Not sure if PHPStan complains if you keep the `|null`. I guess without foreign keys we'll always have this problem 😞 

https://github.com/mailpoet/mailpoet/blob/master/lib/Doctrine/EntityTraits/SafeToOneAssociationLoadTrait.php#L27

lib/Entities/StatisticsClickEntity.php
```diff
@@ -73,4 +73,39 @@ public function getLink() {
...
+  public function setSubscriberId($subscriberId) {
+    $this->subscriberId = $subscriberId;
+  }
+
+  /**
+   * @param NewsletterLinkEntity|null $link
+   */
+  public function setLink($link) {
+    $this->link = $link;
+  }
```
JanJakes, line 103: All of these are in fact non-nullable in the DB so I think it would make sense to use constructor (in all of the statistic entities). (They are nullable on PHP side but only due to `safelyLoadToOneAssociation`.)

PR #2780, opened 2020-05-11
Form: columns in popup form [MAILPOET-2850]

costasovo:
[MAILPOET-2850]

[MAILPOET-2850]: https://mailpoet.atlassian.net/browse/MAILPOET-2850


pavel-mailpoet:
@veljkho I am a little bit afraid of some future regressions in this area. We are planning some other tasks that will affect the popup modals. We will change paddings, borders and others. I am afraid we will bring this bug back. I don't think we can write an automated test. Do you have any idea to prevent this breaking in the future?

PR #2788, opened 2020-05-14
Update the close button [MAILPOET-2915]

pavel-mailpoet:


assets/js/src/public.jsx
```diff
@@ -67,15 +67,26 @@ jQuery(($) => {
...
+  const closeForm = (formDiv) => {
+    formDiv.removeClass('active');
+    formDiv.prev('.mailpoet_form_popup_overlay').removeClass('active');
+    Cookies.set('popup_form_dismissed', '1', { expires: 365, path: '/' });
+  };
+
+  $(document).keyup((e) => {
+    if (e.key === 'Escape') {
+      $('div.mailpoet_form').each((index, element) => {
+        if ($(element).children('.mailpoet_form_close_icon').length !== 0) {
```
costasovo, line 79: Shoudn't we rather add some data attribute saying that form is closable? I'm afraid we can change the icon class and this will brake. What do you think?

PR #2792, opened 2020-05-14
Improve form editor background style [MAILPOET-2914]

pavel-mailpoet:
[MAILPOET-2914]

[MAILPOET-2914]: https://mailpoet.atlassian.net/browse/MAILPOET-2914

### DIFF
--- a/assets/js/src/form_editor/components/form_styling_background.jsx
+++ b/assets/js/src/form_editor/components/form_styling_background.jsx
@@ -58,6 +58,8 @@ const FormStylingBackground = ({ children }) => {
         borderStyle,
         textAlign,
         padding,
+        width: 700,
+        margin: '0 auto',
       }}
     >
       {children}


assets/js/src/form_editor/components/form_styling_background.jsx
```diff
@@ -58,6 +58,9 @@ const FormStylingBackground = ({ children }) => {
         borderStyle,
         textAlign,
         padding,
+        width: 700,
+        margin: '0 auto',
+        paddingBottom: 30,
```
costasovo, line 63: I think we should drop the paddingBottom here. It overrides padding set by a user in styles settings and when user changes it it gets overridden again. See https://d.pr/v/6Pz3i8. I would respect the padding set in form styles settings.

pavel-mailpoet: Well, the padding is in the specification. It shouldn't be overridden by user's settings though.

costasovo: Maybe we could ask PM to reconsider the bottom padding, since it is "fighting" with the one user specifies.

PR #2794, opened 2020-05-14
Doctrine save newsletter [MAILPOET-2900]

JanJakes:
[MAILPOET-2900]

[MAILPOET-2900]: https://mailpoet.atlassian.net/browse/MAILPOET-2900

@veljkho For QA we need to look at saving emails in the editor and sending all their options (scheduling, etc.) in the second step (Activate/Schedule/Send). Also, email editing after sending should work and previews/view-in-browser link should still render correctly. All should work as before.


lib/Newsletter/NewsletterSaveController.php
```diff
@@ -253,4 +244,34 @@ private function updateSegments(NewsletterEntity $newsletter, array $segments) {
 
     $this->entityManager->flush();
   }
+
+  private function updateOptions(NewsletterEntity $newsletter, array $options) {
```
costasovo, line 210: We don't handle orphaned newsletter options fields here. I guess it is because the list of options is defined by $optionFields so the removal of orphaned newsletter options should be handled when someone removes an option field. Am I correct?

lib/Newsletter/NewsletterSaveController.php
```diff
@@ -291,4 +276,31 @@ private function rescheduleIfNeeded(NewsletterEntity $newsletter, Newsletter $ne
...
     }
   }
+
+  private function updateQueue(NewsletterEntity $newsletter, Newsletter $newsletterModel, array $options) {
+    $queue = $newsletter->getLatestQueue();
+    if (!$queue) {
+      return;
+    }
+
+    if (in_array($newsletter->getType(), [NewsletterEntity::TYPE_NOTIFICATION, NewsletterEntity::TYPE_NOTIFICATION_HISTORY], true)) {
```
costasovo, line 269: I know it is a reimplementation of current state but this condition makes me think. Do we need to do this for other automatic emails like welcome emails or WooCommerce emails? This will regenerate the body and subject of the last sent email. Maybe we even don't want to do it for welcome emails and WooCommerce emails.

lib/Newsletter/NewsletterSaveController.php
```diff
@@ -119,61 +111,114 @@ public function save(array $data = []): array {
...
+      $newsletter->setSenderAddress($data['sender_address'] ?? '');
+    }
+
+    if (array_key_exists('reply_to_name', $data)) {
+      $newsletter->setReplyToName($data['reply_to_name'] ?? '');
+    }
+
+    if (array_key_exists('reply_to_address', $data)) {
+      $newsletter->setReplyToAddress($data['reply_to_address'] ?? '');
+    }
```
costasovo, line 174: In the old [model's createUpdate is a functionality](https://github.com/mailpoet/mailpoet/blob/master/lib/Models/Newsletter.php#L990-L1019) that sets these properties based on values from settings instead of an empty string as a fallback. I'm not sure if removing this may cause some issues. Maybe we should keep it. What do you think?

tasks/phpstan/phpstan-baseline.neon
```diff
@@ -243,3 +243,8 @@ parameters:
 			message: "#^Parameter \\#1 \\$newsletter of method MailPoet\\\\Entities\\\\NewsletterTemplateEntity\\:\\:setNewsletter\\(\\) expects MailPoet\\\\Entities\\\\NewsletterEntity\\|null, object given\\.$#"
 			count: 1
 			path: ../../lib/NewsletterTemplates/NewsletterTemplatesRepository.php
+
+		-
+			message: "#^Parameter \\#2 \\$segment of class MailPoet\\\\Entities\\\\NewsletterSegmentEntity constructor expects MailPoet\\\\Entities\\\\SegmentEntity, object given\\.$#"
```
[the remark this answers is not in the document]
JanJakes: Yes. The Doctrine extension for PHPStan is not working in our case and even though I have a hacking branch with prefixing it, I'm not sure if we can safely use it because we're using a pretty old Doctrine version and I guess the extension has stubs for current versions. We can also create our own stubs I guess (this is now writable using generics).

lib/Newsletter/NewsletterSaveController.php
JanJakes: I thought so too but it was misbehaving... and the reason is the code you point to is only for `onCreate`. Note that the `save` endpoint is update-only now (we never use it for create, there's a different endpoint for that).

lib/Newsletter/NewsletterSaveController.php
```diff
@@ -221,4 +222,35 @@ private function updateNewsletter(NewsletterEntity $newsletter, array $data) {
...
+        $newsletterSegment = new NewsletterSegmentEntity($newsletter, $segment);
+        $this->entityManager->persist($newsletterSegment);
+      }
+      $newsletter->getNewsletterSegments()->add($newsletterSegment);
+    }
+
+    // remove orphaned newsletter segments
+    foreach (array_diff($oldNewsletterSegments, $newsletter->getNewsletterSegments()->toArray()) as $newsletterSegment) {
+      $this->newsletterSegmentRepository->remove($newsletterSegment);
+    }
```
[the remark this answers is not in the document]
JanJakes: I guess I wanted to make it more explicit. But maybe it's actually safer to add it? The relation is `NewsletterEntity <- NewsletterSegmentEntity -> SegmentEntity` so I guess I can add it to:

1. `NewsletterEntity` (remove `NewsletterSegmentEntity` when removed from `NewsletterEntity`).
2. `SgmentEntity` (remove `NewsletterSegmentEntity` when removed from `SegmentEntity`).

Right?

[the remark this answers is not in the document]
JanJakes: Adding a commit & test. Only from the `NewsletterEntity` side for now (not sure about performance implications if we do this the other way too).

lib/Newsletter/NewsletterSaveController.php
JanJakes: We don't delete any options here but I'm adding Doctrine's orphan removal for options too.

lib/Newsletter/NewsletterSaveController.php
JanJakes: @costasovo In other words, we should regenerate the body and subject only for `standard` newsletters, right? (All types are `automatic`, `standard`, `welcome`, `notification`, `notification_history`, `wc_transactional`.)

costasovo: Hey. I would say so.

JanJakes:
@veljkho Good catch! Fixed.

JanJakes:
@costasovo Added one more fix (last commit).

PR #2796, opened 2020-05-18
Forms: add input colour [MAILPOET-2908]

pavel-mailpoet:
[MAILPOET-2908]

[MAILPOET-2908]: https://mailpoet.atlassian.net/browse/MAILPOET-2908

### DIFF
--- a/lib/Form/Block/Text.php
+++ b/lib/Form/Block/Text.php
@@ -41,14 +41,27 @@ class Text {
     if (in_array($block['id'], ['email', 'last_name', 'first_name'], true)) {
       $automationId = 'data-automation-id="form_' . $block['id'] . '" ';
     }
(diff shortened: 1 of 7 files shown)


lib/Form/Block/Text.php
```diff
@@ -41,14 +41,27 @@ public function render(array $block, array $formSettings): string {
...
+      && $block['params']['label_within']
+      && isset($block['styles']['font_color'])
+    ) {
+      $html .= '<style>'
+        . 'input[name="data[' . $name . ']"]::placeholder{'
+        . 'color:' . $block['styles']['font_color'] . ';'
+        . 'opacity: 1;'
+        . '}'
+        . '</style>';
+    }
```
costasovo, line 58: This code repeats itself twice. I think we could add it to `BlockStylesRenderer` as a method. What do you think?

PR #2807, opened 2020-05-21
Prevent deleting the last radio item [MAILPOET-2889]

pavel-mailpoet:


pavel-mailpoet:
@veljkho That is something that Rosta mentioned as well. But that was fixed in a different pull request: https://github.com/mailpoet/mailpoet/pull/2809

PR #2809, opened 2020-05-25
Check params before accessing values [MAILPOET-2941]

amine-mp:
[MAILPOET-2941]

[MAILPOET-2941]: https://mailpoet.atlassian.net/browse/MAILPOET-2941

### DIFF
--- a/assets/js/src/form_editor/store/store.jsx
+++ b/assets/js/src/form_editor/store/store.jsx
@@ -12,10 +12,14 @@ import validateForm from './form_validator.jsx';
 import { formBodyToBlocksFactory } from './form_body_to_blocks.jsx';
 import mapFormDataAfterLoading from './map_form_data_after_loading.jsx';
 
+const customFields = window.mailpoet_custom_fields.map(
+  (field) => ({ ...field, params: field.params || {} })
+);
+
 const formBodyToBlocks = formBodyToBlocksFactory(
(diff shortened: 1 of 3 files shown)


assets/js/src/form_editor/store/store.jsx
```diff
@@ -12,10 +12,14 @@ import validateForm from './form_validator.jsx';
 import { formBodyToBlocksFactory } from './form_body_to_blocks.jsx';
 import mapFormDataAfterLoading from './map_form_data_after_loading.jsx';
 
+const customFields = window.mailpoet_custom_fields.map(
```
costasovo, line 15: I'm thinking that maybe we could fix this in PHP when we generate the response https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/ResponseBuilders/CustomFieldsResponseBuilder.php#L25
or maybe even in the entity in getter [getParams](https://github.com/mailpoet/mailpoet/blob/master/lib/Entities/CustomFieldEntity.php#L57).
The getter is annotated that it should return an array. So probably we could return an empty array or at least fix the annotation. What do you think?

amine-mp: I thought about returning an empty array but this will make `params` in JS an empty array and not an object!
I will fix the annotation

PR #2812, opened 2020-05-26
Refactor renderer [MAILPOET-2899]

pavel-mailpoet:


lib/Services/Bridge.php
```diff
@@ -53,6 +53,16 @@ public static function isMPSendingServiceEnabled() {
     }
   }
 
+  public function isMailpoetSendingServiceEnabled() {
```
JanJakes, line 56: Is it on purpose that we keep also the `isMPSendingServiceEnabled` static method?

lib/Entities/NewsletterPostEntity.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+class NewsletterPostEntity {
+  use AutoincrementedIdTrait;
+  use CreatedAtTrait;
+  use UpdatedAtTrait;
+  use SafeToOneAssociationLoadTrait;
+
+  /**
+   * @ORM\ManyToOne(targetEntity="MailPoet\Entities\NewsletterEntity", inversedBy="newsletterSegments")
+   * @var NewsletterEntity
```
JanJakes, line 23: I guess you need to annotate this `@var NewsletterEntity|null`. After https://github.com/mailpoet/mailpoet/pull/2814 was merged this would fail PHPStan, I guess.

(By default phpstan-doctrine thinks the association is nullable. In DB it is not, which would be fixable by adding `@JoinColumn(nullable=false)` but since we use the `SafeToOneAssociationTrait` we're actually making it nullable indeed.)

lib/Services/Bridge.php
pavel-mailpoet: The original static method is still used in many places. I should probably mark it as `@deprecated`.

PR #2815, opened 2020-05-27
Update form responsiveness [MAILPOET-2924]

pavel-mailpoet:
[MAILPOET-2924]

[MAILPOET-2924]: https://mailpoet.atlassian.net/browse/MAILPOET-2924

### DIFF
--- a/lib/Form/Util/Styles.php
+++ b/lib/Form/Util/Styles.php
@@ -165,6 +165,7 @@ EOL;
         top: auto;
         transform: none;
         width: 100%;
+        min-width: 100%;
       }} ";
     }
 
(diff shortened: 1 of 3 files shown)


lib/Form/Util/Styles.php
```diff
@@ -158,7 +158,19 @@ public function renderFormSettingsStyles(array $form, string $selector, string $
     $formWrapperStyles = $selector . '{' . join(';', $styles) . ';}';
 
     // Media styles
-    $media = "@media (max-width: 500px) {{$selector} {background-image: none;}}";
+    $media = "@media (max-width: 500px) {{$selector} {
+      animation: none;
```
costasovo, line 162: The new rules should apply only to slide-in, fixed bar and pop up. With this implementation, it will apply to all forms including widgets and below posts. We need to apply additional selectors for form types. You can use $displayType variable and FormEtity display type constants to implement it.

PR #2824, opened 2020-05-28
Reduce size of package [MAILPOET-2692]

costasovo:
[MAILPOET-2692]

[MAILPOET-2692]: https://mailpoet.atlassian.net/browse/MAILPOET-2692

### DIFF
--- a/prefixer/composer.json
+++ b/prefixer/composer.json
@@ -30,7 +30,8 @@
       "php fix-twig.php",
       "php fix-doctrine.php",
       "php fix-swiftmailer.php",
-      "php fix-validator.php"
+      "php fix-validator.php",
+      "php fix-monolog.php"
     ]
   },
   "config": {
--- a/tasks/strip-whitespaces.php
+++ b/tasks/strip-whitespaces.php
@@ -1,0 +1,18 @@
+<?php
+
+// Strip comments and whitespaces from php files in folder
+
+// absolutize path (if not absolute)
+$directory = rtrim($argv[1] . '/');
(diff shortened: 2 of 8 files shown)


tasks/strip-whitespaces.php
```diff
@@ -0,0 +1,18 @@
...
+  $directory = getcwd() . DIRECTORY_SEPARATOR . $directory;
+}
+
+$iterator = new RecursiveDirectoryIterator($directory, RecursiveDirectoryIterator::SKIP_DOTS);
+$files = new RecursiveIteratorIterator($iterator, RecursiveIteratorIterator::SELF_FIRST);
+foreach ($files as $file) {
+  if (substr($file, -3) !== 'php') {
+    continue;
+  }
+  file_put_contents($file, php_strip_whitespace ( $file));
```
JanJakes, line 17: Weird spacing after `php_strip_whitespace`.

prefixer/composer.json
```diff
@@ -30,7 +30,8 @@
       "php fix-twig.php",
       "php fix-doctrine.php",
       "php fix-swiftmailer.php",
-      "php fix-validator.php"
+      "php fix-validator.php",
```
JanJakes, line 33: We could start to think of a more sophisticated "cleaner" than those copy-pasted scripts 🤔

costasovo: Hmmm, maybe we could. There is a copy-pasted error handling code snippet. On the other hand, what we have is a simple solution and one can see easily from composer.json that there are packages and some scripts which modify them.

PR #2828, opened 2020-06-01
Show subscribers counted towards plan and plugin limits [MAILPOET-2917]

wxa:


views/subscribers/subscribers.html
```diff
@@ -63,8 +66,14 @@
     'numberOfItemsSingular': __('1 item'),
     'numberOfItemsMultiple': __('%$1d items'),
 
-    'subscribersInPlan': __('%$1d subscribers in your plan'),
-    'subscribersInPlanTooltip': __('This is the total of subscribed, unconfirmed and inactive subscribers. The count excludes unsubscribed and bounced (invalid) email addresses.'),
+    'subscribersInPlanCount': _x('%$1d / %$2d', 'count / total subscribers'),
+    'subscribersInPlan': _x('%s subscribers in your plan', 'number of subscribers in a sending plan'),
+    'subscribersInPlanTooltip': __('This is the total of subscribed, unconfirmed and inactive subscribers we count when you are sending with MailPoet Sending Service. The count excludes unsubscribed and bounced (invalid) email addresses.'),
+
+    'mailpoetSubscribersCount': _x('%$1d / %$2d', 'count / total subscribers'),
```
pavel-mailpoet, line 73: Those two strings are the same: `subscribersInPlanCount` and `mailpoetSubscribersCount`. Can we reuse them? Why do we even need this to go through translation? Do we expect this to be different in some language?

wxa: > Can we reuse them?
This way it's easier to decouple them if a need arises. Other than that no big reasons for not reusing them, so I'll do it.
> Why do we even need this to go through translation? Do we expect this to be different in some language?
I was thinking of RTL languages.

PR #2831, opened 2020-06-02
Add offer clearout step to subscribers import [MAILPOET-2919]

costasovo:
[MAILPOET-2919]

[MAILPOET-2919]: https://mailpoet.atlassian.net/browse/MAILPOET-2919


assets/js/src/subscribers/importExport/import/step_offer_clearout.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import Button from 'common/button/button';
+import MailPoet from 'mailpoet';
+
+type Props = {
+  history: History,
+}
+
+export default ({ history }: Props) => (
+  <div className="mailpoet-offer-clearout-step__container">
```
pavel-mailpoet: Did we start using BEM? I am not aware.

assets/js/src/context/useSegmentsContext.jsx
```diff
@@ -3,5 +3,7 @@ import React from 'react';
 export default function useSegmentsContext(data) {
   return React.useMemo(() => ({
     all: data.mailpoetSegments,
-  }), [data]);
+    // eslint-disable-next-line no-param-reassign
```
pavel-mailpoet, line 6: 😞 is it necessary to disable eslint here?

PR #2832, opened 2020-06-03
Form: font selection [MAILPOET-2849]

pavel-mailpoet:
[MAILPOET-2849]

[MAILPOET-2849]: https://mailpoet.atlassian.net/browse/MAILPOET-2849


assets/js/src/form_editor/blocks/custom_textarea/edit.jsx
```diff
@@ -159,10 +159,14 @@ const CustomTextAreaEdit = ({
 
   const textarea = useRef(null);
   const getTextArea = (placeholder) => {
-    let style = '';
+    let style = `#${id}::placeholder {`;
```
costasovo, line 162: Just an idea. Because of placeholder, we render <style> block with every component (input, textarea) and we use the component id in CSS selector. So I'm thinking that we could use <[FormStyles](https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/form_editor/components/form_styles.jsx)> component (currently used only for Custom CSS rendering) and render these rules inside this component using block class names in a selector. This could be possibly used for all form global styles. What do you think?

assets/js/src/form_editor/components/font_family_settings.tsx
```diff
@@ -155,14 +95,19 @@ const FontFamilySettings = ({
...
-  />
-);
+export const CustomFontsStyleSheetLink = () => {
+  const customFonts = useSelect(
+    (select) => select('mailpoet-form-editor').getAllCustomFonts(),
+    []
+  );
+  const customFontsUrl = customFonts
+    .map((fontName) => fontName.replace(' ', '+'))
+    .map((fontName) => fontName.concat(':400,400i,700,700'))
```
costasovo, line 105: I see that on front end we have `:400,400i,700,700i` so maybe the `i` is missing here.

assets/js/src/form_editor/blocks/custom_textarea/edit.jsx
pavel-mailpoet: Good idea.

PR #2833, opened 2020-06-03
Form editor: fix inconsistencies in form styles rendering [MAILPOET-2938]

costasovo:
[MAILPOET-2938]

[MAILPOET-2938]: https://mailpoet.atlassian.net/browse/MAILPOET-2938


wxa:
Note: initially I've got an error when clicking the preview button:
> Target container is not a DOM element.

The error happened in the `Modal` component and the preview wasn't opening. But after that I could not reproduce it, so I'm passing the ticket further.

PR #2834, opened 2020-06-03
Fix fetching posts for ALC [MAILPOET-2960]

costasovo:
[MAILPOET-2960]

[MAILPOET-2960]: https://mailpoet.atlassian.net/browse/MAILPOET-2960

### DIFF
--- a/lib/Entities/NewsletterPostEntity.php
+++ b/lib/Entities/NewsletterPostEntity.php
@@ -19,13 +19,13 @@ class NewsletterPostEntity {
   use SafeToOneAssociationLoadTrait;
 
   /**
-   * @ORM\ManyToOne(targetEntity="MailPoet\Entities\NewsletterEntity", inversedBy="newsletterSegments")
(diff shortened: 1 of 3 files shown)


lib/Entities/NewsletterPostEntity.php
```diff
@@ -19,13 +19,13 @@ class NewsletterPostEntity {
...
   /**
-   * @ORM\ManyToOne(targetEntity="MailPoet\Entities\NewsletterEntity", inversedBy="newsletterSegments")
+   * @ORM\ManyToOne(targetEntity="MailPoet\Entities\NewsletterEntity")
    * @var NewsletterEntity|null
    */
   private $newsletter;
 
   /**
-   * @ORM\Column(type="int")
+   * @ORM\Column(type="integer")
```
JanJakes, line 28: How is it possible Doctrine didn't scream and PHPStan didn't catch those? 🙀

PR #2842, opened 2020-06-04
Update import subscribers test [MAILPOET-2966]

veljkho:
I'd like to commit an update of the existing test case SubscriberManageImportExportCest.php to include scenario of creating a new subscribers list plus searching and selecting while importing subscribers.

Jira -> [MAILPOET-2966](https://mailpoet.atlassian.net/browse/MAILPOET-2966)

### DIFF
--- a/tests/acceptance/SubscriberManageImportExportCest.php
+++ b/tests/acceptance/SubscriberManageImportExportCest.php
@@ -58,7 +58,7 @@ class SubscriberManageImportExportCest {
     // Test reimporting the same list
     $i->click('Import again');
     $this->uploadCsvFile($i);
-    $this->chooseListAndConfirm($i);
+    $this->createNewListAndConfirm($i);
     $i->see('9 existing subscribers were updated and added to');
 
     //confirm subscribers from import list were added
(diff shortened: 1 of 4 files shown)


tests/acceptance/SubscriberManageImportExportCest.php
```diff
@@ -101,6 +101,21 @@ private function chooseListAndConfirm(\AcceptanceTester $i) {
...
 
+  private function createNewListAndConfirm(\AcceptanceTester $i) {
+    $newListName = 'Simple List';
+    // first create a new list
+    $i->click('a.mailpoet_create_segment');
+    $i->fillField('input#new_segment_name', $newListName);
+    $i->fillField('textarea#new_segment_description', 'This is just a simple list.');
+    $i->click('input#new_segment_process');
+    // trigger dropdown to display selections and search for recently created list
+    $i->waitForElementVisible('input.select2-search__field');
```
costasovo, line 112: We have a helper function for selecting in select2. I think you can use `$i->selectOptionInSelect2($newListName);`
See https://github.com/mailpoet/mailpoet/blob/master/tests/_support/AcceptanceTester.php#L122

tests/acceptance/SubscriberManageImportExportCest.php
```diff
@@ -101,6 +101,21 @@ private function chooseListAndConfirm(\AcceptanceTester $i) {
...
+    $i->fillField('input.select2-search__field', $newListName);
+    $i->click(['xpath' => '//*[@id="select2-mailpoet_segments_select-results"]/li[1]']);
+    $i->click('.mailpoet_data_manipulation_step [data-automation-id="import-next-step"]');
```
costasovo, line 115: You don't need to combine those `[data-automation-id="import-next-step"]` should be enough. We add those data-automation-id attributes to have stable selectors for acceptance tests.

tests/acceptance/SubscriberManageImportExportCest.php
```diff
@@ -101,6 +101,21 @@ private function chooseListAndConfirm(\AcceptanceTester $i) {
     $i->waitForText('Import again');
   }
 
+  private function createNewListAndConfirm(\AcceptanceTester $i) {
+    $newListName = 'Simple List';
+    // first create a new list
+    $i->click('a.mailpoet_create_segment');
```
costasovo, line 107: Can you use $i->click(`Create a new list`) here? It is closer to user's context.

veljkho: Sure, I just changed to your solution. Thanks.

tests/acceptance/SubscriberManageImportExportCest.php
veljkho: I agree, I just copy/paste that code from existing, so I now changed both mine and another inside. Thanks.

PR #2843, opened 2020-06-04
Redesign datepicker [MAILPOET-2779]

wxa:
Notes:
1. Our newsletter send page datepicker is actually two different components: datepicker and time picker (a simple select box). I've replaced both, but did not design their composite appearance, i.e. they don't fit into the new grid width as one component. I've only shrunk the time picker so they could fit on one line as before.
2. MP2Migrator still uses a jQuery UI progressbar, but the corresponding styles are encapsulated in `mp2-migrator.scss`, so I've deleted the huge jQuery UI CSS file.


assets/css/src/components/_datepicker.scss
```diff
@@ -0,0 +1,95 @@
...
+    box-shadow: 0 2px 8px 0 $color-tertiary-light;
+    font-family: $font-family;
+    font-size: $font-size;
+    height: 400px;
+    width: $grid-column;
+  }
+
+  .react-datepicker-wrapper,
+  .react-datepicker-wrapper input {
+    width: 100%;
```
NeosinneR, line 20: `box-sizing: border-box;` should be added, so the width is correctly calculated (and not add the padding resulting in more than 100%)

assets/css/src/components/_datepicker.scss
```diff
@@ -0,0 +1,95 @@
+@import '../../../../node_modules/react-datepicker/src/stylesheets/datepicker';
+
+.mailpoet-datepicker {
+  display: inline-block;
+  vertical-align: middle;
+
+  .react-datepicker {
```
NeosinneR, line 7: Add `left: -1px;` so it's aligned perfectly with an input (otherwise it's positioned from the border resulting in 1px shift).

wxa:
@NeosinneR Thank you for reviewing and the input on styles! 

I've added two more CSS changes because for me the calendar still did not look nice: 
* `  .react-datepicker__month-container { float: none; }` because the calendar was sticking to the left edge of the panel due to `float: left;` in the original stylesheet;
* `padding: ($grid-gap / 2) 0;` in `.react-datepicker__current-month` because the current month name was vertically below center.

NeosinneR:
@wxa good call 👍 updated it a bit more, now should be all good

PR #2858, opened 2020-06-10
Forms: styles for headings and paragraphs [MAILPOET-2965]

pavel-mailpoet:
[MAILPOET-2965]

[MAILPOET-2965]: https://mailpoet.atlassian.net/browse/MAILPOET-2965


assets/js/src/form_editor/rich_text/font_selection_format.tsx
```diff
@@ -0,0 +1,52 @@
...
+import FontFamilySettings from '../components/font_family_settings';
+
+const name = 'mailpoet-form/font-selection';
+const title = 'Font Selection';
+
+const settings = {
+  name,
+  title,
+  tagName: 'span',
+  className: 'has-font',
```
costasovo, line 15: This is probably too common class name. Let's rather use `mailpoet-has-font`. I think we should also add a fallback font family on frontend using this class in public.js

assets/js/src/form_editor/rich_text/init.ts
```diff
@@ -0,0 +1,7 @@
+import '@wordpress/format-library';
```
costasovo, line 1: We could add a comment here to explain why we do this import here.

PR #2862, opened 2020-06-11
Load Gutenberg JS translations for form editor [MAILPOET-2942]

costasovo:
[MAILPOET-2942]

[MAILPOET-2942]: https://mailpoet.atlassian.net/browse/MAILPOET-2942

### DIFF
--- a/lib/AdminPages/Pages/FormEditor.php
+++ b/lib/AdminPages/Pages/FormEditor.php
@@ -4,6 +4,7 @@ namespace MailPoet\AdminPages\Pages;
 
 use MailPoet\AdminPages\PageRenderer;
 use MailPoet\API\JSON\ResponseBuilders\CustomFieldsResponseBuilder;
+use MailPoet\Config\Localizer;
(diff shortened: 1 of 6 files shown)


lib/AdminPages/Pages/FormEditor.php
```diff
@@ -117,4 +118,43 @@ private function getPreviewPageUrl() {
...
+  /**
+   * JS Translations are distributed and loaded per script. We can't use wp_set_script_translations
+   * because translation filename is determined based on script filename and path.
+   * This function loads JSON files with Gutenberg script's translations distributed within WordPress.
+   * Implemented based on load_script_textdomain function
+   * @see https://developer.wordpress.org/reference/functions/load_script_textdomain/
+   * @return string[]
+   */
+  private function getGutenbergScriptsTranslations() {
+    $locale = $this->wp->determineLocale();
```
pavel-mailpoet, line 131: I think we should use `\MailPoet\Config\Localizer::locale` here.

PR #2864, opened 2020-06-15
QA build refactor [MAILPOET-2949]

pavel-mailpoet:
[MAILPOET-2949]

[MAILPOET-2949]: https://mailpoet.atlassian.net/browse/MAILPOET-2949


.circleci/config.yml
```diff
@@ -136,6 +136,22 @@ jobs:
     - run:
         name: "Static analysis"
         command: ./do qa:phpstan
+  qa_js:
+    executor: wpcli_php_oldest
```
JanJakes, line 140: Why "oldest"?

RoboFile.php
```diff
@@ -348,11 +348,23 @@ public function doctrineGenerateProxies() {
...
+  public function qaPhp() {
     $collection = $this->collectionBuilder();
     $collection->addCode([$this, 'qaLint']);
     $collection->addCode(function() {
       return $this->qaCodeSniffer('all');
     });
+    return $collection->run();
+  }
+
+  public function qaJavascript() {
```
JanJakes, line 366: This includes also CSS... should we split it from JS?

.circleci/config.yml
pavel-mailpoet: no reason, I copied that from the `js_tests` job

PR #2873, opened 2020-06-16
Convert null shortcode params to false (implementations rely on false) [MAILPOET-2972]

JanJakes:
[MAILPOET-2972]

[MAILPOET-2972]: https://mailpoet.atlassian.net/browse/MAILPOET-2972


JanJakes:
@costasovo Unfortunately, that will be a much bigger task, I guess. At the moment `null` and `false` can mean different things, i.e. here: https://github.com/mailpoet/mailpoet/blob/master/lib/Newsletter/Shortcodes/Categories/Subscriber.php#L18

There's even a test for that (empty value just returns the shortcode while `false` renders the default). This is just a quick fix based on a support ticket.

PR #2877, opened 2020-06-18
Keep records when a subscriber unsubscribes [MAILPOET-2737]

pavel-mailpoet:
[MAILPOET-2737]

[MAILPOET-2737]: https://mailpoet.atlassian.net/browse/MAILPOET-2737

### DIFF
--- a/lib/API/JSON/v1/Subscribers.php
+++ b/lib/API/JSON/v1/Subscribers.php
@@ -385,7 +385,6 @@ class Subscribers extends APIEndpoint {
         isset($data['status'])
         && ($data['status'] === SubscriberEntity::STATUS_UNSUBSCRIBED)
         && ($oldSubscriber instanceof Subscriber)
-        && ($oldSubscriber->status === SubscriberEntity::STATUS_SUBSCRIBED)
       ) {
         $currentUser = $this->wp->wpGetCurrentUser();
         $this->unsubscribesTracker->track(
--- a/lib/API/JSON/v1/Subscribers.php
+++ b/lib/API/JSON/v1/Subscribers.php
@@ -6,6 +6,7 @@ use MailPoet\API\JSON\Endpoint as APIEndpoint;
 use MailPoet\API\JSON\Error as APIError;
 use MailPoet\API\JSON\Response as APIResponse;
 use MailPoet\Config\AccessControl;
+use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\StatisticsUnsubscribeEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Form\Util\FieldNameObfuscator;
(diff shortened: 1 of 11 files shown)


lib/Statistics/Track/Unsubscribes.php
```diff
@@ -23,7 +23,9 @@ public function __construct(
     $this->statisticsUnsubscribesRepository = $statisticsUnsubscribesRepository;
   }
 
-  public function track(int $subscriberId, int $queueId = null, string $source) {
+  public function track(int $subscriberId, int $queueId = null, string $source, string $meta = null) {
```
wxa, line 26: An optional argument (`queueId`) is followed by a mandatory argument (`source`). 

Ideally, this should not be allowed by our coding standard as we seem to move closer to [PSR-2](https://www.php-fig.org/psr/psr-2/#44-method-arguments):
> Method arguments with default values MUST go at the end of the argument list.

lib/Config/Migrator.php
```diff
@@ -564,4 +568,19 @@ private function sqlify($model, $attributes) {
 
     return implode("\n", $sql);
   }
+
+  private function updateNullInUnsubscribeStats() {
+    global $wpdb;
+    // perform once for versions below or equal to 3.47.5
```
wxa, line 575: Please increment the version, this will probably be out next week.

lib/API/JSON/v1/Subscribers.php
```diff
@@ -346,6 +354,23 @@ public function save($data = []) {
     }
     $data['segments'] = array_merge($data['segments'], $this->getNonDefaultSubscribedSegments($data));
     $newSegments = $this->findNewSegments($data);
+    if (isset($data['id']) && (int)$data['id'] > 0) {
+      $oldSubscriber = Subscriber::findOne((int)$data['id']);
+      if (
+        isset($data['status'])
+        && ($data['status'] === SubscriberEntity::STATUS_UNSUBSCRIBED)
+        && ($oldSubscriber instanceof Subscriber)
+        && ($oldSubscriber->status === SubscriberEntity::STATUS_SUBSCRIBED)
```
wxa, line 388: Should we track if unsubscription by admin happens from any other status?

lib/API/JSON/v1/Subscribers.php
```diff
@@ -385,7 +385,6 @@ public function save($data = []) {
         isset($data['status'])
         && ($data['status'] === SubscriberEntity::STATUS_UNSUBSCRIBED)
         && ($oldSubscriber instanceof Subscriber)
-        && ($oldSubscriber->status === SubscriberEntity::STATUS_SUBSCRIBED)
```
wxa: @pavel-mailpoet  This should probably be `&& ($oldSubscriber->status !== SubscriberEntity::STATUS_UNSUBSCRIBED)`, otherwise it will write to unsubscribe log every time an unsubscribed user is edited.

pavel-mailpoet: @wxa great catch. That would be a big problem.

PR #2884, opened 2020-06-23
Templates selection 1 [MAILPOET-2987]

pavel-mailpoet:
[MAILPOET-2987]

[MAILPOET-2987]: https://mailpoet.atlassian.net/browse/MAILPOET-2987


lib/AdminPages/Pages/FormEditor.php
```diff
@@ -110,6 +112,147 @@ public function render() {
...
+            .mailpoet_checkbox {}
+            .mailpoet_submit {}
+            .mailpoet_divider {}
+            .mailpoet_message {}
+            .mailpoet_form_loading {  width: 30px;  text-align: center;  line-height: normal;}
+            .mailpoet_form_loading > span {  width: 5px;  height: 5px;  background-color: #5b5b5b;}
+        ',
+      ],
+    ];
+    if (empty($templates) || !$this->flagsController->isSupported(FeaturesController::TEMPLATES_SELECTION)) {
```
JanJakes, line 238: `$templates` seems to be never empty, at least at this point.

assets/css/src/components-form-editor/_template-selection.scss
```diff
@@ -0,0 +1,3 @@
+.template-selection {
+  margin: 30px;
```
JanJakes, line 2: It's probably better to use `$grid-gap` or `$grid-gap-large` but I guess styling will come with another ticket, right?

assets/js/src/form_editor/template_selection.tsx
```diff
@@ -0,0 +1,30 @@
...
+const App = () => (
+  <GlobalContext.Provider value={useGlobalContextValue(window)}>
+    <>
+      <Notices />
+      <Selection />
+    </>
+  </GlobalContext.Provider>
+);
+
+window.addEventListener('DOMContentLoaded', () => {
```
JanJakes, line 17: What if DOM is loaded before you register this event? Can it happen? (jQuery's ready function is very complex to catch such scenarios)

assets/js/src/forms/list.jsx
```diff
@@ -127,25 +127,12 @@ const itemActions = [
...
-          response.errors.map((error) => error.message),
-          { scroll: true }
-        );
-      }
+    MailPoet.trackEvent('Forms > Add New', {
+      'MailPoet Free version': window.mailpoet_version,
     });
+    setTimeout(() => {
+      window.location = window.mailpoet_form_edit_url;
+    }, 200); // leave some time for the event to track
```
JanJakes: I see the timeout is not a new addition but why don't we just wait for the promise? And is it needed (since we don't care how the request ends)?

lib/AdminPages/Pages/FormEditor.php
```diff
@@ -70,12 +191,12 @@ public function __construct(
...
   public function render() {
-    if (!isset($_GET['id']) && !isset($_GET['action'])) {
+     if (!isset($_GET['id']) && !isset($_GET['action'])) {
       $this->renderTemplateSelection();
       return;
     }
-    if ($_GET['action'] === 'create') {
-      $this->createEmptyForm();
+    if (isset($_GET['action']) && $_GET['action'] === 'create') {
+      $this->createForm($_GET['template-id']);
```
JanJakes: What if `template-id` is not present? This is changing in multiple commits so I'm a bit confused now.

lib/AdminPages/Pages/FormEditor.php
JanJakes: Hmm, looking at it now (in a later commit) the flags controller check should probably be somewhere before calling `renderTemplateSelection` and the IF removed from here completely.

lib/Config/Menu.php
```diff
@@ -454,6 +474,10 @@ public function formEditor() {
     $this->container->get(FormEditor::class)->render();
   }
 
+  public function formEditorTemplateSelection() {
+    $this->container->get(FormEditor::class)->renderTemplateSelection();
```
JanJakes, line 478: I think the page should reside in https://github.com/mailpoet/mailpoet/tree/master/lib/AdminPages/Pages.

assets/js/src/forms/list.jsx
pavel-mailpoet: Because there is no promise there in the `MailPoet.trackEvent`. We are using mixpanel library and it doesn't return a promise: https://developer.mixpanel.com/docs/javascript-full-api-reference#mixpaneltrack so I don't see a better way here.

lib/AdminPages/Pages/FormEditor.php
pavel-mailpoet: This code was removed. Sorry.

PR #2893, opened 2020-06-24
Update editing subscriber scenario [MAILPOET-3013]

veljkho:
[MAILPOET-3013](https://mailpoet.atlassian.net/browse/MAILPOET-3013)


tests/acceptance/SubscriberManagementCest.php
```diff
@@ -4,6 +4,7 @@
 
 use MailPoet\Test\DataFactories\Segment;
 use MailPoet\Test\DataFactories\Subscriber;
+use Codeception\Util\Locator;
```
amine-mp, line 7: This dependency is added but not used, please remove it

veljkho:
Hi @amine-mp , Can we just merge this beside failed test? Seems to me that this is some WC test which fails in last days.. do we have another fix in another ticket? I believe yes.

PR #2894, opened 2020-06-25
Fix acceptamce wc order [MAILPOET-2992]

costasovo:
[MAILPOET-2992]

[MAILPOET-2992]: https://mailpoet.atlassian.net/browse/MAILPOET-2992

### DIFF
--- a/tests/_support/AcceptanceTester.php
+++ b/tests/_support/AcceptanceTester.php
@@ -376,7 +376,7 @@ class AcceptanceTester extends \Codeception\Actor {
     }
   }
 
-    /**
+  /**
    * Select a payment method (cheque, cod, ppec_paypal)
    */
   public function selectPaymentMethod($method = 'cod') {
--- a/tests/_support/AcceptanceTester.php
+++ b/tests/_support/AcceptanceTester.php
@@ -357,8 +357,11 @@ class AcceptanceTester extends \Codeception\Actor {
    */
   public function optInForSubscription() {
     $i = $this;
-    $i->waitForElementClickable('#mailpoet_woocommerce_checkout_optin');
-    $i->checkOption('#mailpoet_woocommerce_checkout_optin');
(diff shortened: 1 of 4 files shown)


tests/_support/AcceptanceTester.php
```diff
@@ -357,20 +357,26 @@ public function optInForRegistration() {
...
-    $i->uncheckOption('#mailpoet_woocommerce_checkout_optin');
+    $checked = $i->executeJS('return document.getElementById("mailpoet_woocommerce_checkout_optin").checked');
+    if ($checked) {
+      $i->waitForElementClickable('[data-automation-id="woo-commerce-subscription-opt-in"]');
+      $i->click('[data-automation-id="woo-commerce-subscription-opt-in"]');
+    }
   }
 
-  /**
+    /**
```
JanJakes, line 379: Typo

tests/_support/AcceptanceTester.php
```diff
@@ -357,20 +357,26 @@ public function optInForRegistration() {
...
    */
   public function optOutOfSubscription() {
     $i = $this;
-    $i->waitForElementClickable('#mailpoet_woocommerce_checkout_optin');
-    $i->uncheckOption('#mailpoet_woocommerce_checkout_optin');
+    $checked = $i->executeJS('return document.getElementById("mailpoet_woocommerce_checkout_optin").checked');
+    if ($checked) {
+      $i->waitForElementClickable('[data-automation-id="woo-commerce-subscription-opt-in"]');
+      $i->click('[data-automation-id="woo-commerce-subscription-opt-in"]');
+    }
```
JanJakes, line 376: Maybe we should put this into a private method? As you prefer...

costasovo: I'll leave it as is. I think in this case it is slightly better for readability.

PR #2895, opened 2020-06-25
Wizard redesign [MAILPOET-2784]

NeosinneR:
[MAILPOET-2784]

[MAILPOET-2784]: https://mailpoet.atlassian.net/browse/MAILPOET-2784


assets/js/src/common/button/button.tsx
```diff
@@ -12,6 +12,7 @@ type Props = {
   iconEnd?: JSX.Element,
   onClick?: () => void,
   href?: string,
+  type?: string,
```
JanJakes, line 15: This should be `'a' | 'button'` I guess. Also, a default should be in the component header below (but all defaults seem to be missing, actually).

JanJakes: Hmm, looking further, it seems to be unused...

assets/js/src/common/typography/list/list.tsx
```diff
@@ -0,0 +1,20 @@
...
+  children: React.ReactNode,
+  isOrdered?: boolean,
+};
+
+const List = ({
+  children,
+  isOrdered,
+}: Props) => {
+  const Element = isOrdered ? 'ol' : 'ul';
+  return React.createElement(
```
JanJakes, line 13: I've seen this in a couple of new components so just a note - we should probably pick one style and use it. I'd vote for `<Element...>` since it's the syntax we normally use (and this is a `.tsx` file).

assets/css/src/components-plugin/_welcome-wizard.scss
```diff
@@ -1,3 +1,15 @@
+.admin_page_mailpoet-welcome-wizard {
```
JanJakes, line 1: Do you target both variants (with/without Premium plugin active)?

assets/css/src/components-plugin/_welcome-wizard.scss
```diff
@@ -1,3 +1,15 @@
...
+  background: #fff;
+
+  #wpcontent {
+    padding: 0;
+  }
+
+  .wrap {
+    margin: 0;
+  }
+}
```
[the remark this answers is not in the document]
NeosinneR: FIxed, hidden this notice from the wizard page. Other notices were already hidden.

assets/js/src/common/button/button.tsx
NeosinneR: Good catch, I automatically assumed (knowing how many components work), that all HTML attributes are automatically used as part of `...attributes` prop - but that's not used in `<Button>` component. Added. 

And this `type` is used for [HTML button type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), not to differentiate what element to use - added possible values, so it's clear.

assets/js/src/common/typography/list/list.tsx
NeosinneR: Unified, also updated `<Heading>` component

PR #2902, opened 2020-06-29
Redesign listing header/footer [MAILPOET-2778]

wxa:
I did not add the `Listing` component to Storybook because it requires communication with the server to work, but our Storybook is static.

**Note to QA:**
 
Please ensure that all listings work and look good in both desktop and mobile versions with different screen dimensions and different contents, in the Free and Premium plugins. Please check correct functioning of all listing components: 
* groups;
* search;
* filters;
* pagination;
* selection with checkboxes;
* "Select all" action;
* bulk actions including subscribers' lists manipulation with modals;
* custom actions like "Empty trash" or "Create segment";
* sorting;
* special states like loading/no items;
* etc.


assets/css/src/components-plugin/_listing.scss
```diff
@@ -97,6 +97,7 @@
 }
 
 .mailpoet-listing-pages-num {
+  font-weight: normal;
```
wxa, line 100: It's `font-weight: 600` in the design, normal is 400. Why are you using it?

NeosinneR: You're right, I guess I mixed it up with `font-stretch`. I removed that commit, so it's back to were it was.

wxa:
> However, this commit breaks newsletter editor, which is why many acceptance tests are failing. 

Thank you for pointing it out. I've replaced `lodash` with `underscore` in this file, the newsletter editor seems to be working now.

PR #2906, opened 2020-06-29
Redesign Tooltip component [MAILPOET-2998]

amine-mp:
[MAILPOET-2998]

[MAILPOET-2998]: https://mailpoet.atlassian.net/browse/MAILPOET-2998

### DIFF
--- a/assets/js/src/common/tooltip/tooltip.tsx
+++ b/assets/js/src/common/tooltip/tooltip.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import classNames from 'classnames';
+import ReactTooltip, { TooltipProps } from 'react-tooltip';
+
+const Tooltip = ({
+  effect, textColor, backgroundColor, border, borderColor, className, children, ...props
+}: TooltipProps) => (
+  <ReactTooltip
+    effect={effect || 'solid'}
+    textColor={textColor || '#071c6d'}
+    backgroundColor={backgroundColor || '#fafbfe'}
+    borderColor={borderColor || '#e5e9f8'}
+    className={classNames('mailpoet-tooltip', className)}
+    border={border === undefined ? true : border}
+    {...props}
+  >
+    {children}
+  </ReactTooltip>
+);
+
+export default Tooltip;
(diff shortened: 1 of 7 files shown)


assets/js/src/common/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import classNames from 'classnames';
+import ReactTooltip, { TooltipProps } from 'react-tooltip';
+
+const Tooltip = (props: TooltipProps) => (
```
NeosinneR, line 5: Wouldn't it be better to use
```
const Tooltip = ({
    backgroundColor, 
    border, 
    borderColor, 
    effect, 
    textColor, 
    ...props
}: TooltipProps) => (
  <ReactTooltip
      effect={effect || 'solid'}
      textColor={textColor || '#071c6d'}
      backgroundColor={backgroundColor || '#fafbfe'}
      borderColor={borderColor || '#e5e9f8'}
      className={classNames('mailpoet-tooltip', className)}
      border={border === undefined ? true : border}
      {...props}
  >
);
```
? It's less messy and this way if a new prop is added to React Tooltip, we don't have to update our component.

amine-mp: I wanted to do that, but I had the following ESLint error:
```
 Prop spreading is forbidden  react/jsx-props-no-spreading
```
Should we ignore this rule for this specific file?

pavel-mailpoet: The rule has a good explanation: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md

> Enforces that there is no spreading for any JSX attribute. This enhances readability of code by being more explicit about what props are received by the component. It is also good for maintainability by avoiding passing unintentional extra props and allowing react to emit warnings when invalid HTML props are passed to HTML elements.

Maybe we don't have to pass all of them? just those we are using.

NeosinneR: Yes, we already do so for new components. Maybe it's time to rethink this rule. I'll ask on Slack. https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/common/form/checkbox/group.tsx#L51

JanJakes:
@amine-mp The tooltips don't look as they should according to the design in Zeplin.
Also, some tests are red (I guess just flakiness).

PR #2909, opened 2020-06-29
Update transactional emails label [MAILPOET-2944]

amine-mp:
[MAILPOET-2944]

[MAILPOET-2944]: https://mailpoet.atlassian.net/browse/MAILPOET-2944


costasovo:
@amine-mp I notified JM and he updated the spec [MAILPOET-2944]. So we can just show the warning all the time.

amine-mp:
@costasovo it's already done :)

PR #2912, opened 2020-06-30
Form close button styles [MAILPOET-2963]

pavel-mailpoet:
[MAILPOET-2963]

[MAILPOET-2963]: https://mailpoet.atlassian.net/browse/MAILPOET-2963


pavel-mailpoet:
@veljkho 
> Missing 6th style of white cross
I used the icons that were attached to the JIRA ticket. Also, the Zeplin design page has only 5 icons: https://app.zeplin.io/project/5ccadeea21d5db688697b384/screen/5ef3503c296fcc9f51a1f204

> According to the specs, white cross on black square and circle, the cross is just transparent, it is not white. ...
Again, I used the icon that was in the JIRA ticket, if we want to use a different icon I need the icon file.

assets/js/src/form_editor/components/close_button_settings.tsx
```diff
@@ -47,6 +47,7 @@ const CloseButtonsSettings = ({
     (sel) => sel('mailpoet-form-editor').getCloseIconsUrl(),
     []
   );
+  const current = value ?? 'round_white';
```
costasovo, line 50: Please move the default value to data mapper "after loading from server".

PR #2913, opened 2020-06-30
Settings save on verify, fix resume sending [MAILPOET-2991] [MAILPOET-3006]

JanJakes:
This fixes both [MAILPOET-2991] and [MAILPOET-3006].

[MAILPOET-2991]: https://mailpoet.atlassian.net/browse/MAILPOET-2991
[MAILPOET-3006]: https://mailpoet.atlassian.net/browse/MAILPOET-3006

### DIFF
--- a/lib/Config/ServicesChecker.php
+++ b/lib/Config/ServicesChecker.php
@@ -122,7 +122,8 @@ class ServicesChecker {
   public function isMailPoetAPIKeyPendingApproval(): bool {
     $mssActive = Bridge::isMPSendingServiceEnabled();
     $mssKeyValid = $this->isMailPoetAPIKeyValid();
-    $mssKeyPendingApproval = $this->settings->get('mta.mailpoet_api_key_state.data.is_approved') === false;
+    $isApproved = $this->settings->get('mta.mailpoet_api_key_state.data.is_approved');
+    $mssKeyPendingApproval = $isApproved === false || $isApproved === 'false'; // API unfortunately saves this as a string
     return $mssActive && $mssKeyValid && $mssKeyPendingApproval;
   }
 }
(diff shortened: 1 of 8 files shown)


lib/Config/ServicesChecker.php
```diff
@@ -122,7 +122,8 @@ public function isPremiumKeyValid($displayErrorNotice = true) {
   public function isMailPoetAPIKeyPendingApproval(): bool {
     $mssActive = Bridge::isMPSendingServiceEnabled();
     $mssKeyValid = $this->isMailPoetAPIKeyValid();
-    $mssKeyPendingApproval = $this->settings->get('mta.mailpoet_api_key_state.data.is_approved') === false;
+    $isApproved = $this->settings->get('mta.mailpoet_api_key_state.data.is_approved');
```
costasovo, line 125: I guess this fix has to be in this place since some users already might have it saved like that, correct?

lib/Cron/Workers/KeyCheck/SendingServiceKeyCheck.php
```diff
@@ -40,10 +41,17 @@ public function getNextRunDate() {
...
+
     $mssKey = $this->settings->get(Mailer::MAILER_CONFIG_SETTING_NAME)['mailpoet_api_key'];
     $result = $this->bridge->checkMSSKey($mssKey);
     $this->bridge->storeMSSKeyAndState($mssKey, $result);
     $this->bridge->updateSubscriberCount($result);
+
+    $isPendingApproval = $this->servicesChecker->isMailPoetAPIKeyPendingApproval();
+    if ($wasPendingApproval && !$isPendingApproval) {
+      MailerLog::resumeSending();
+    }
```
costasovo, line 54: I think it would be nice to add a test for this functionality. We already have an [integration test for this service](https://github.com/mailpoet/mailpoet/blob/master/tests/integration/Cron/Workers/KeyCheck/SendingServiceKeyCheckTest.php) so I guess we could add a test case for this.

lib/Config/ServicesChecker.php
JanJakes: Yes, but also - the problem still exists. It will be fixed only by converting the API to JSON.

lib/Cron/Workers/KeyCheck/SendingServiceKeyCheck.php
JanJakes: 👍  I'll add a test.

PR #2915, opened 2020-07-01
Form templates backend [MAILPOET-2985]

costasovo:
[MAILPOET-2985]

[MAILPOET-2985]: https://mailpoet.atlassian.net/browse/MAILPOET-2985


lib/Form/Templates/InitialForm.php
```diff
@@ -0,0 +1,123 @@
+<?php
+
+namespace MailPoet\Form\Templates;
```
pavel-mailpoet, line 3: Should we move the actual templates into a separate directory? When we have 50 templates it will be difficult to find the other classes in this directory.

PR #2916, opened 2020-07-02
Form editor: enable gradients [MAILPOET-3005]

costasovo:
[MAILPOET-3005]

[MAILPOET-3005]: https://mailpoet.atlassian.net/browse/MAILPOET-3005


assets/js/src/form_editor/store/mapping/from_blocks/styles_mapper.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (has(styles, 'padding') && styles.padding !== undefined) {
+    mappedStyles.padding = styles.padding;
+  }
+  if (has(styles, 'fontFamily') && styles.fontFamily) {
+    mappedStyles.font_family = styles.fontFamily;
+  }
+  return mappedStyles;
+};
+
+export const mapColorSlugToValue = (
```
pavel-mailpoet, line 48: Should this function be here? It has nothing to do with mapping server data to blocks and we are using it in `form_editor/blocks/html/edit.jsx` and that is not related to store or server or anything.

assets/js/src/form_editor/store/mapping/to_blocks/styles_mapper.ts
```diff
@@ -27,22 +27,22 @@ export const mapInputBlockStyles = (styles) => {
     mappedStyles.backgroundColor = styles.background_color;
   }
   if (has(styles, 'border_size') && styles.border_size !== undefined) {
-    mappedStyles.borderSize = parseInt(styles.border_size, 10);
+    mappedStyles.borderSize = +styles.border_size;
```
pavel-mailpoet, line 40: Why this change? I find it less readable and less understandable. I don't understand why is it in this commit, what does it have in common with the rest of the changes in here? Also what happens if `border_size` is an empty string? Can it be?
https://github.com/airbnb/javascript#coercion--numbers

lib/Form/Block/Columns.php
```diff
@@ -12,9 +12,12 @@ private function getStyles(array $params): string {
     if (isset($params['text_color'])) {
       $styles[] = "color:{$params['text_color']};";
     }
-    if (isset($params['background_color'])) {
+    if (isset($params['background_color']) && !isset($params['gradient'])) {
```
pavel-mailpoet, line 15: Hmmm, I think this needs an explanation, I don't understand this line of code, can you add a code comment, please?

assets/js/src/form_editor/blocks/blocks.jsx
```diff
@@ -1,3 +1,4 @@
+import '@wordpress/block-editor/build-module/hooks/padding.js';
```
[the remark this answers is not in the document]
costasovo: Good find. It is a leftover from some testing. 👍

assets/js/src/form_editor/store/mapping/from_blocks/styles_mapper.ts
costasovo: Well, this maps some string constants used in form editor to CSS value so I would say it is related to mapping styles data.

PR #2917, opened 2020-07-02
Fix templates translation [MAILPOET-2616]

amine-mp:
[MAILPOET-2616]

[MAILPOET-2616]: https://mailpoet.atlassian.net/browse/MAILPOET-2616


amine-mp:
> I noticed that in some cases you changed Unsubscribe | Manage subscription to Unsubscribe. | Manage your subscription. Is that intentional?

Yes, it is intentional. The translated strings has `Manage your subscription`, but some templates had `your` missing so I added it in other to use the same translation for all templates.

For code style, my editor formats the file according to our style rules once I save it. So only the files that I changed and saved manually were formatted. As for others I did a "search and replace in all" that why they stayed unformatted.

Our QA command does not complain about these files because we are ignoring them here https://github.com/mailpoet/mailpoet/blob/master/RoboFile.php#L400

PR #2919, opened 2020-07-06
Add scenario various status newsletters [MAILPOET-3042]

veljkho:
[MAILPOET-3042](https://mailpoet.atlassian.net/browse/MAILPOET-3042)
Added new scenario in existing test case SettingsArchivePageCest and also updated Newsletter data factory.

### DIFF
--- a/tests/acceptance/SettingsArchivePageCest.php
+++ b/tests/acceptance/SettingsArchivePageCest.php
@@ -37,4 +37,42 @@ class SettingsArchivePageCest {
     $i->waitForText($pageTitle2);
     $i->waitForText('SentNewsletter');
   }
+
+  public function createArchivePageWithVariousStatusNewsletters(\AcceptanceTester $i) {
+    $i->wantTo('Create page with MP archive shortcode, showing only sent newsletters but having various in database');
+    $segmentFactory = new Segment();
+    $segment3 = $segmentFactory->withName('SentNewsletters')->create();
+    $newsletterFactory = new Newsletter();
+    $newsletterFactory->withSubject('SentNewsletter')->withSentStatus()->withSendingQueue()->withSegments([$segment3])->create();
+    $newsletterFactory->withSubject('DraftNewsletter')->withDraftStatus()->withScheduledQueue()->withSegments([$segment3])->create();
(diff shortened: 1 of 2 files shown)


tests/acceptance/SettingsArchivePageCest.php
```diff
@@ -37,4 +37,42 @@ public function createArchivePageWithSentNewsletters(\AcceptanceTester $i) {
...
+    $i->wantTo('Create page with MP archive shortcode, showing only sent newsletters but having various in database');
+    $segmentFactory = new Segment();
+    $segment3 = $segmentFactory->withName('SentNewsletters')->create();
+    $newsletterFactory = new Newsletter();
+    $newsletterFactory->withSubject('SentNewsletter')->withSentStatus()->withSendingQueue()->withSegments([$segment3])->create();
+    $newsletterFactory->withSubject('DraftNewsletter')->withDraftStatus()->withScheduledQueue()->withSegments([$segment3])->create();
+    $newsletterFactory->withSubject('ScheduledNewsletter')->withScheduledStatus()->withScheduledQueue()->withSegments([$segment3])->create();
+    $pageTitle3 = 'SentNewsletterArchive';
+    $pageContent3 = "[mailpoet_archive segments=\"$segment3->id\"]";
+    $i->cli(['post', 'create', '--post_type=page', '--post_status=publish', "--post_title=$pageTitle3", "--post_content=$pageContent3"]);
```
costasovo, line 51: We have a helper for creating a post. See https://github.com/mailpoet/mailpoet/blob/master/tests/_support/AcceptanceTester.php#L420
I think you could extend it so that it supports also pages by adding a parameter postType with default value 'post'.
This helper return URL so you can then just go directly to the page without going to admin and clicking view. It is faster and more reliable. I remember we experienced some flakiness with posts listing.
Here is some usage: 
https://github.com/mailpoet/mailpoet/blob/3276d32b6090f4dd46b6deddda426d2071e89a1d/tests/acceptance/DisplayFormBellowPostCest.php#L45-L46

veljkho: Thanks, I added that helper across all three scenarios in this test case. Please dont merge this PR, just assign back to me.

PR #2927, opened 2020-07-08
Form mobile view adjustments [MAILPOET-3024]

costasovo:
[MAILPOET-3024]

[MAILPOET-3024]: https://mailpoet.atlassian.net/browse/MAILPOET-3024

## Note for QA
Please test how heading margins in the mobile version look on different forms.

### DIFF
--- a/assets/css/src/components-public/_public.scss
+++ b/assets/css/src/components-public/_public.scss
@@ -408,3 +408,9 @@ div.mailpoet_form_slide_in.active {
 .mailpoet-has-background-color {
   padding: 20px 30px;
 }
+
+.mailpoet-heading {
+  @include breakpoint-max-width($form-break-small-width - 1) {
+    font-size: 18px !important;
+  }
+}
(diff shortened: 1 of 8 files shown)


assets/css/src/components-public/_public.scss
```diff
@@ -7,6 +12,26 @@
 .mailpoet_list_label,
 .mailpoet_date_label {
   display: block;
+
+  @include breakpoint-max-width($form-break-small-width - 1) {
+    font-size: 16px !important;
```
pavel-mailpoet, line 17: Shouldn't we also use a variable for those sizes? Both `16px` and `1.4`? They are used in at least two places right now.

assets/css/src/components-public/_public.scss
```diff
@@ -432,5 +432,8 @@ div.mailpoet_form_slide_in.active {
 .mailpoet-heading {
   @include breakpoint-max-width($form-break-small-width - 1) {
     font-size: 18px !important;
```
pavel-mailpoet, line 434: Same here, shouldn't this be a variable?

PR #2928, opened 2020-07-08
Acceptance tests fixes [MAILPOET-3044]

costasovo:
[MAILPOET-3044]

[MAILPOET-3044]: https://mailpoet.atlassian.net/browse/MAILPOET-3044

### DIFF
--- a/assets/js/src/listing/listing.jsx
+++ b/assets/js/src/listing/listing.jsx
@@ -381,6 +381,7 @@ class Listing extends React.Component {
       action: 'bulkAction',
       data,
     }).done(() => {
+      // Reload items after a bulk action except for empty trash action which redirects to All tab.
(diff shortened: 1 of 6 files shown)


assets/js/src/listing/listing.jsx
```diff
@@ -381,7 +381,10 @@ class Listing extends React.Component {
       action: 'bulkAction',
       data,
     }).done(() => {
-      this.getItems();
+      const isEmptyTrashAction = selectedIds === 'all' && params.group === 'trash' && params.action === 'delete';
+      if (!isEmptyTrashAction) {
```
pavel-mailpoet, line 385: I think we need an explanation here. if an engineer comes here after some time they wouldn't understand why do we have this here.

PR #2934, opened 2020-07-14
Return array instead of null from api [MAILPOET-2518]

pavel-mailpoet:
[MAILPOET-2518]

[MAILPOET-2518]: https://mailpoet.atlassian.net/browse/MAILPOET-2518

### DIFF
--- a/lib/API/JSON/Response.php
+++ b/lib/API/JSON/Response.php
@@ -30,14 +30,13 @@ abstract class Response {
     if (!empty($this->meta)) {
       $response['meta'] = $this->meta;
     }
-    if ($data !== null) {
-      $response = array_merge($response, $data);
+    if ($data === null) {
+      $data = [];
     }
+    $response = array_merge($response, $data);
 
-    if (!empty($response)) {
-      @header('Content-Type: application/json; charset=' . get_option('blog_charset'));
-      echo WPFunctions::get()->wpJsonEncode($response);
-    }
+    @header('Content-Type: application/json; charset=' . get_option('blog_charset'));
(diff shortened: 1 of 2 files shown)


lib/API/JSON/Response.php
```diff
@@ -30,9 +30,10 @@ public function send() {
...
       $response['meta'] = $this->meta;
     }
-    if ($data !== null) {
-      $response = array_merge($response, $data);
+    if ($data === null) {
+      $data = [];
     }
+    $response = array_merge($response, $data);
 
     if (!empty($response)) {
```
wxa, line 38: This response is still an empty body because `[]` is considered empty here.

wxa: How about allowing an empty array: `if (!empty($response) || is_array($response))`.

pavel-mailpoet: https://circleci.com/gh/mailpoet/mailpoet/91305?utm_campaign=chatroom-integration&utm_medium=referral&utm_source=slack

I tried that but PHPStan is complaining that the condition is always true. And it is because `$response` is always an array. I think we could remove this condition altogether and send the response always. What do you think?

wxa: Yes, that's an option.

wxa:
@veljkho Please check that in the upcoming WP 5.5 (install it using the WordPress Beta Tester plugin) sending a test email in "Send with... > Other" and force syncing WP users list work without errors. Before this fix there was an error:
> An error has happened while performing a request, the server has responded with response code 200

PR #2952, opened 2020-07-30
Update manage subscription [MAILPOET-1635]

costasovo:
[MAILPOET-1635]

[MAILPOET-1635]: https://mailpoet.atlassian.net/browse/MAILPOET-1635

### DIFF
--- a/views/subscription/manage_subscription.html
+++ b/views/subscription/manage_subscription.html
@@ -8,7 +8,7 @@
   <p class="mailpoet_paragraph">
     <label> <%= __('Email', 'mailpoet') %>*<br /><strong><%= email|escape %></strong></label>
     <br />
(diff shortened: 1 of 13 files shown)


views/subscription/manage_subscription.html
```diff
@@ -8,7 +8,7 @@
   <p class="mailpoet_paragraph">
     <label> <%= __('Email', 'mailpoet') %>*<br /><strong><%= email|escape %></strong></label>
     <br />
-    <span style="font-size:85%;"><%= editEmailInfo|raw %></span>
+    <span class="mailpoet-change-email-info" style="font-size:85%;"><%= editEmailInfo|raw %></span>
```
pavel-mailpoet, line 11: Why did you leave the font size here an in the css file as well?

PR #2957, opened 2020-07-31
Redesign newsletter listing stats badges [MAILPOET-3000]

wxa:
[MAILPOET-3000](https://mailpoet.atlassian.net/browse/MAILPOET-3000)

To make styling easier I've replaced the component in the actual listing while developing and then decided to leave it there. This is a bit broader than the task scope.

I also kept the old badge components/styles, they are still used on the stats page.


NeosinneR:
@veljkho @wxa I changed the font size to 11px on purpose, because 10px is very small and sometimes not readable (I was even thinking about 12px, but that seemed big). The vertical centering is a bit problematic now because Chrome and Firefox are acting differently with the current font. I specifically added top padding, because it looks a little bit better when it's off to the bottom then to the top. 

Also, we'll be changing the font to Proxima Nova Soft in a few weeks (as part of listings redesign) so the line-height property of the font will probably change again and we'll have to go through this again. I'd keep this as is and added a note to https://mailpoet.atlassian.net/browse/MAILPOET-3073 to see how it looks with a new font. What do you think?

veljkho:
Then I agree also @NeosinneR , changing the font in the near future can also lead to this same issue reappearing.

Issue #2960, opened 2020-08-03
Don't change status when creating WP user if already subscribed [MAILPOET-3016]


NeosinneR:
Currently, when a user subscribes to a newsletter and confirms, he is in our database as `subscribed`. But when a new WP user account is created with the same email, the subscriber status changes back from `subscribed` to `unconfirmed`. 

Video: https://take.ms/cEXBY

### Specs

1. When the new WordPress user is registered but is already as a subscriber in MailPoet, don’t change the global status - keep it to whatever it was before.

PR #2961, opened 2020-08-03
Fix subscriber status resetting on WP user creation [MAILPOET-3016]

lysyjan:
Fixes:#2960
[MAILPOET-3016]

[MAILPOET-3016]: https://mailpoet.atlassian.net/browse/MAILPOET-3016

### DIFF
--- a/lib/Segments/WP.php
+++ b/lib/Segments/WP.php
@@ -43,6 +43,11 @@ class WP {
     $wpSegment = Segment::getWPSegment();
     if (!$wpSegment) return;
 
+    // find subscriber by email when is false
+    if (!$subscriber) {
+      $subscriber = Subscriber::where('email', $wpUser->user_email)->findOne(); // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
+    }
+
     $scheduleWelcomeNewsletter = false;
     if (in_array($currentFilter, ['profile_update', 'user_register'])) {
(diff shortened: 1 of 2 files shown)


lib/Segments/WP.php
```diff
@@ -61,7 +68,7 @@ private static function updateSubscriber($currentFilter, $wpUser, $subscriber =
       'email' => $wpUser->user_email, // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
       'first_name' => $firstName,
       'last_name' => $lastName,
-      'status' => $signupConfirmationEnabled ? Subscriber::STATUS_UNCONFIRMED : Subscriber::STATUS_SUBSCRIBED,
+      'status' => $subscriberOldStatus ?: ($signupConfirmationEnabled ? Subscriber::STATUS_UNCONFIRMED : Subscriber::STATUS_SUBSCRIBED),
```
costasovo, line 71: We already have logic for keeping the original status. See https://github.com/mailpoet/mailpoet/pull/2961/commits/843c5e6abf811dab31d92670f87f2f4deb32fd43#diff-dd3a1bae08a4e17e5d2bb8b31568fde4R75-R79

I think that if we just fetch the subscriber by his email it should be enough to get behavior we want.

lysyjan: Good point 👍 I simplified the code.

PR #2970, opened 2020-08-05
Fix search in sending status [MAILPOET-3050]

pavel-mailpoet:
[MAILPOET-3050]

[MAILPOET-3050]: https://mailpoet.atlassian.net/browse/MAILPOET-3050

### DIFF
--- a/lib/Models/ScheduledTaskSubscriber.php
+++ b/lib/Models/ScheduledTaskSubscriber.php
@@ -76,7 +76,7 @@ class ScheduledTaskSubscriber extends Model {
 
   public static function listingQuery($data) {
     $group = isset($data['group']) ? $data['group'] : 'all';
-    return self::join(Subscriber::$_table, ["subscriber_id", "=", "subscribers.id"], "subscribers")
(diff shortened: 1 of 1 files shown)


lib/Models/ScheduledTaskSubscriber.php
```diff
@@ -86,6 +86,19 @@ public static function listingQuery($data) {
...
+    if (isset($data['search'])) {
+      $search = trim($data['search']);
+      $search = str_replace(['\\', '%', '_'], ['\\\\', '\\%', '\\_'], $search); // escape for 'LIKE'
+      if (strlen($search) === 0) {
+        return $query;
+      }
+
+      return $query->whereRaw(
+        '(`subscribers.email` LIKE ? OR `subscribers.first_name` LIKE ? OR `subscribers.last_name` LIKE ?)',
+        ['%' . $search . '%', '%' . $search . '%', '%' . $search . '%']
```
amine-mp, line 98: I would prefer to do `$search = '%' . $search . '%'` before then just use `$search` here, I think it will be cleaner that way.

Issue #2968, opened 2020-08-05
Changing subscribers status on import [MAILPOET-2957]


NeosinneR:
Right now, there is no way to bulk change subscribers statuses. The workaround is to change the statuses directly in the database. However, we already trust users when setting up MailPoet for the first time with their WooCoommerce customers and with combination with Clearout integration, we want to give them more options when migrating from other services. 

### Use cases

1. A customer wants to migrate to MailPoet and he does have a list of active subscribers or on the other hand a list of bounced/inactive subscribers and wants to create a list in MailPoet to never send an email to those customers (a suppression list) - he’ll use the new Set new subscribers' status to option
2. Regular synchronization of subscribers with another CRM. A customer keeps a list of subscribers/customers in another service and only uses MailPoet to send emails, but wants to respect unsubscribed customers. He’ll use both new options, setting status for both new and existing customers.

### Specs

<img width="825" alt="Screenshot 2020-06-16 16 30 27" src="https://user-images.githubusercontent.com/470616/89379100-76ff5d00-d6f5-11ea-9937-f2bd4988e5bd.png">

**1. Add new option**

- Position: below lists select
- Title: Set new subscribers' status to
- Options: `Subscribed` (default), `Inactive`. `Unsubscribed`

**2. Add new option** 

- Position: below previous 
- Title: Update existing subscribers' status to
- Options: `Don’t update` (default), `Subscribed`, `Inactive`, `Unsubscribed`

**3. On submit:**

- Always change the status for a selected list(s) as `Subscribed`.
- Change the global subscriber status to the selected value (even for inactive users) - respect different values for new and existing customers

**4. Change `Next step` button to `Import` on this step**

PR #2972, opened 2020-08-05
Add possibility set status for new or existing users in import [MAILPOET-2957]

lysyjan:
Fixes:#2968
[MAILPOET-2957]

[MAILPOET-2957]: https://mailpoet.atlassian.net/browse/MAILPOET-2957

### DIFF
--- a/views/subscribers/importExport/import.html
+++ b/views/subscribers/importExport/import.html
@@ -133,5 +133,11 @@
 'offerClearoutText3': __('We recommend cleaning your list with Clearout because the first 500 emails are free, it’s fast and it detects up to 50% of invalid addresses. It’s our support team’s favorite.'),
 'tryClearout': _x('Try clearout.io for free', 'CTA button label'),
 'clearoutGotIt': _x('Got it, I’ll proceed to import', 'Text in a link'),
+'subscribed': __('Subscribed'),
+'unsubscribed': __('Unsubscribed'),
+'inactive': __('Inactive'),
+'dontUpdate': __('Don\’t update'),
+'newSubscribersStatus': __('Set new subscribers\' status to'),
+'existingSubscribersStatus': __('Update existing subscribers\' status to'),
 }) %>
 <% endblock %>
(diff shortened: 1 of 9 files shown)


views/subscribers/importExport/import.html
```diff
@@ -133,5 +133,11 @@ <h1 class="title">
 'offerClearoutText3': __('We recommend cleaning your list with Clearout because the first 500 emails are free, it’s fast and it detects up to 50% of invalid addresses. It’s our support team’s favorite.'),
 'tryClearout': _x('Try clearout.io for free', 'CTA button label'),
 'clearoutGotIt': _x('Got it, I’ll proceed to import', 'Text in a link'),
+'subscribed': __('Subscribed'),
+'unsubscribed': __('Unsubscribed'),
+'inactive': __('Inactive'),
+'dontUpdate': __('Don\’t update'),
```
pavel-mailpoet, line 139: You don't need the backslash here. It is better to not include it here because when searching in the source code for "Don't update" you find nothing because of the backslash.

views/subscribers/importExport/import.html
```diff
@@ -133,5 +133,11 @@ <h1 class="title">
...
+'newSubscribersStatus': __('Set new subscribers\' status to'),
```
pavel-mailpoet, line 140: The same note about the `\'` here. We should use a typographical apostrophe because it looks better on the page and we don't need to escape it.

views/subscribers/importExport/import.html
```diff
@@ -133,5 +133,11 @@ <h1 class="title">
...
+'existingSubscribersStatus': __('Update existing subscribers\' status to'),
```
pavel-mailpoet, line 141: The same note about the `\'` here. We should use a typographical apostrophe because it looks better on the page and we don't need to escape it.

veljkho:
Sorry @lysyjan 
it was due to my cache, after clearing cache problem solved.
Will test it soon.

PR #2975, opened 2020-08-11
Add form templates - part 1 [MAILPOET-2810]

costasovo:
[MAILPOET-2810]

[MAILPOET-2810]: https://mailpoet.atlassian.net/browse/MAILPOET-2810


costasovo:
@veljkho OK I'll turn Display on all pages/posts off. Turning those off will cause that all templates will be opened in form editor as widget type form (default preview will be a widget and also a width rendered in the editor will be taken from widget). This is already fixed in an upcoming ticket that will add a master toggle and toggles for all display types. Just so that you know.

PR #2976, opened 2020-08-11
On off switches in form editor [MAILPOET-3075]

pavel-mailpoet:
[MAILPOET-3075]

[MAILPOET-3075]: https://mailpoet.atlassian.net/browse/MAILPOET-3075

### DIFF
--- a/assets/js/src/form_editor/store/map_form_data_after_loading.jsx
+++ b/assets/js/src/form_editor/store/map_form_data_after_loading.jsx
@@ -8,18 +8,22 @@ export default function mapFormDataAfterLoading(data) {
       ...data.settings,
       placeFormBellowAllPages: data.settings.place_form_bellow_all_pages === '1',
       placeFormBellowAllPosts: data.settings.place_form_bellow_all_posts === '1',
+      placementBellowAllPostsEnabled: data.settings.form_placement_bellow_posts_enabled === '1',
       placePopupFormOnAllPages: data.settings.place_popup_form_on_all_pages === '1',
       placePopupFormOnAllPosts: data.settings.place_popup_form_on_all_posts === '1',
+      placementPopupEnabled: data.settings.form_placement_popup_enabled === '1',
       popupFormDelay: data.settings.popup_form_delay !== undefined
         ? asNum(data.settings.popup_form_delay) : defaults.popupForm.formDelay,
       placeFixedBarFormOnAllPages: data.settings.place_fixed_bar_form_on_all_pages === '1',
       placeFixedBarFormOnAllPosts: data.settings.place_fixed_bar_form_on_all_posts === '1',
+      placementFixedBarEnabled: data.settings.form_placement_fixed_bar_enabled === '1',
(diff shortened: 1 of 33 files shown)


lib/Config/Populator.php
```diff
@@ -688,6 +702,50 @@ private function updateSentUnsubscribeLinksToInstantUnsubscribeLinks() {
     ));
   }
 
+  private function addPlacementStatusToForms() {
+    if (version_compare($this->settings->get('db_version', '3.48.1'), '3.48.1', '>')) {
```
costasovo, line 706: I think this should be `$this->settings->get('db_version', '3.48.2')` so that it doesn't run for new installs and also next release will be minor so maybe even '3.49.0'

assets/js/src/form_editor/store/reducer.jsx
```diff
@@ -30,12 +30,18 @@ import {
...
+  disableForm,
+  enableForm,
+} from './reducers/toggle_form';
 
 const createCustomFieldStarted = createCustomFieldStartedFactory(MailPoet);
 const saveFormStarted = saveFormStartedFactory(MailPoet);
 
 export default (defaultState) => (state = defaultState, action) => {
   switch (action.type) {
+    case 'ENABLE_FORM': return enableForm(state);
```
costasovo, line 43: Just wondering, why did you decide to go with 2 reducers.action types instead one using value from action?

lib/Form/DisplayFormInWPContent.php
```diff
@@ -182,14 +186,21 @@ private function shouldDisplayFormType(FormEntity $form, string $formType): bool
     if (!is_array($settings)) return false;
 
     $keys = self::SETUP[$formType];
+
+    $switchKey = $keys['switch'];
+    if (isset($settings[$switchKey]) && ($settings[$switchKey] !== '1')) {
```
costasovo, line 191: I think it might be better to use `!isset($settings[$switchKey]) || ($settings[$switchKey] !== '1'` so that forms without the property are disabled as well.

assets/js/src/form_editor/store/map_form_data_after_loading.jsx
```diff
@@ -8,18 +8,22 @@ export default function mapFormDataAfterLoading(data) {
       ...data.settings,
       placeFormBellowAllPages: data.settings.place_form_bellow_all_pages === '1',
       placeFormBellowAllPosts: data.settings.place_form_bellow_all_posts === '1',
+      placementBellowAllPostsEnabled: data.settings.form_placement_bellow_posts_enabled === '1',
```
costasovo, line 11: It would be nice to add these new props mappings into a test so that someone doesn't remove them in some refactoring. The test should also test that in case these are not set they are mapped as false.

assets/js/src/form_editor/store/reducer.jsx
pavel-mailpoet: No reason really. I didn't want to use a boolean attribute and so I made 2 reducer functions. And I made only a single action so that it is easier to use in the component.

PR #2983, opened 2020-08-12
Update Gutenberg [MAILPOET-3049]

pavel-mailpoet:
[MAILPOET-3049]

[MAILPOET-3049]: https://mailpoet.atlassian.net/browse/MAILPOET-3049


tests/_support/AcceptanceTester.php
```diff
@@ -487,4 +487,12 @@ public function checkTemplateIsPresent(int $templateIndex, string $templateCateg
     $i->waitForElement($template);
     return $template;
   }
+
+  public function clearFormField(string $selector) {
```
costasovo, line 491: With this implementation, it will work only for inputs that have up to 3 characters. I think you could use https://codeception.com/docs/modules/WebDriver#grabattributefrom method and based on value length press backspace as many times as needed.

PR #2984, opened 2020-08-12
Move form title to sidebar [MAILPOET-3086]

pavel-mailpoet:
[MAILPOET-3086]

[MAILPOET-3086]: https://mailpoet.atlassian.net/browse/MAILPOET-3086

### DIFF
--- a/assets/js/src/form_editor/components/form_title.jsx
+++ b/assets/js/src/form_editor/components/form_title.jsx
@@ -15,25 +15,21 @@ export default () => {
   const { changeFormName } = useDispatch('mailpoet-form-editor');
 
   return (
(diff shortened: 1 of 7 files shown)


assets/js/src/form_editor/components/form_title.jsx
```diff
@@ -2,15 +2,14 @@ import React, { useState } from 'react';
 import { useDispatch, useSelect } from '@wordpress/data';
 import classnames from 'classnames';
 import MailPoet from 'mailpoet';
-import Textarea from 'react-autosize-textarea';
```
costasovo, line 5: This was probably the only place where 'react-autosize-textarea' was used. I guess we can remove the package now.

PR #2985, opened 2020-08-13
Premium banner component [MAILPOET-3083]

costasovo:
[MAILPOET-3083]

[MAILPOET-3083]: https://mailpoet.atlassian.net/browse/MAILPOET-3083

### DIFF
--- a/assets/css/src/components/_badge.scss
+++ b/assets/css/src/components/_badge.scss
@@ -1,0 +1,10 @@
+.mailpoet-badge {
+  background-color: $color-tertiary;
+  border-left: 12px solid $color-secondary;
+  border-radius: 2px;
+  box-shadow: inset 12px 0 0 $color-secondary-middle;
+  color: $color-white;
+  display: inline-block;
+  line-height: $line-height;
+  padding: 1px 8px 1px 20px;
+}
--- a/assets/css/src/components/_badge.scss
+++ b/assets/css/src/components/_badge.scss
@@ -1,10 +1,14 @@
+$badge-fancy-border-width: $grid-gap-medium / 2;
+
 .mailpoet-badge {
   background-color: $color-tertiary;
-  border-left: 12px solid $color-secondary;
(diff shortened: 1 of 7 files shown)


assets/css/src/components/_badge.scss
```diff
@@ -0,0 +1,10 @@
+.mailpoet-badge {
```
pavel-mailpoet, line 1: Should all those sizes also come from variables? Why do we have the `grid` settings if we just keep using some random numbers in our CSS?

NeosinneR: I wouldn't do that. Not every single space is tied to the grid. We would end up creating grid variables for every pixel size just for the sake of using a variable and might end up with a much less readable code. In this case, `<Badge>` being a very specific visual component, I don't see a problem with it having its own spacing according to the design.

NeosinneR: Just to clarify, some of them could be replaced with already defined variables, but it wouldn't make sense to replace all. And if I should choose between `padding: 1px 8px 1px 20px;` and `padding: 1px $grid-gap-half 1px ($grid-gap-half + 12px);` I'd choose the first.

pavel-mailpoet: Definitely not `($grid-gap-half + 12px)` that is no difference. I was thinking about adjusting this component so that it fits into the grid. 

> <Badge> being a very specific visual component

What is the rule then? 

I thought we are trying to do the redesign and we are using the grid for some reasons (consistency, visual cleanness, ...). If we keep having exceptions it undermines the purpose. The page will still look like a mess. But don't answer me I don't care that much about the design to argue about it. Just leave it as it is.

costasovo: The double border of the badge has a width of $grid-gap-medium thanks to that I was able to somehow define the spacing using grid variables.

costasovo:
@pavel-mailpoet I've added one new commit.

PR #2995, opened 2020-08-18
Forms: new position for shortcode [MAILPOET-3094]

pavel-mailpoet:
[MAILPOET-3094]

[MAILPOET-3094]: https://mailpoet.atlassian.net/browse/MAILPOET-3094

### DIFF
--- a/assets/css/src/components-form-editor/_form-placement.scss
+++ b/assets/css/src/components-form-editor/_form-placement.scss
@@ -29,6 +29,7 @@
   z-index: 2;
 
   p {
+    line-height: 1.2;
     margin-bottom: 0;
   }
 }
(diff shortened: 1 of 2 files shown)


veljkho (changes requested):
Sorry, maybe I haven't been clear enough when I said to update existing css code line-height.
The problem is still present, we should somehow lower the line-height of the paragraph text "Shortcode & other". Currently, it inherit css line-height from some WP based css style.. here is the [screenshot](https://i.imgur.com/nG20bxY.png). If possible to somehow override that rule with just 1.2 value, would be great.

flakas (approved):

veljkho (approved):

PR #2996, opened 2020-08-19
Send date as a string instead of an object [MAILPOET-3109]

pavel-mailpoet:
[MAILPOET-3109]

[MAILPOET-3109]: https://mailpoet.atlassian.net/browse/MAILPOET-3109


flakas (approved):
The code looks fine, so I'll approve the PR.
We may need to reapprove once WooCommerce releases their fixed WC version, which should also fix the failing builds.

veljkho (approved):

PR #3000, opened 2020-08-25
Add new scenario of not-opting to receive email

veljkho:
Adding new scenario in existing test case.
[MAILPOET-3114](https://mailpoet.atlassian.net/browse/MAILPOET-3114)

### DIFF
--- a/tests/acceptance/SendCategoryPurchaseEmailCest.php
+++ b/tests/acceptance/SendCategoryPurchaseEmailCest.php
@@ -69,15 +69,14 @@ class SendCategoryPurchaseEmailCest {
     $userEmail = Security::generateRandomString() . '-user@email.example';
     $i->orderProduct($product, $userEmail, true, false);
 
-    $this->verifyNoScheduledSending($i);
+    $this->verifyNoScheduledSending($i, $emailSubject);
     
     $i->amOnMailboxAppPage();
     $i->dontSee($emailSubject);
     $i->dontSee($userEmail);
   }
 
-  private function verifyNoScheduledSending(\AcceptanceTester $i) {
-    $emailSubject = 'Product In Category Purchase Test';
+  private function verifyNoScheduledSending(\AcceptanceTester $i, $emailSubject) {
     $i->login();
     $i->amOnMailpoetPage('Help');
     $i->click('System Status');
--- a/tests/acceptance/SendCategoryPurchaseEmailCest.php
+++ b/tests/acceptance/SendCategoryPurchaseEmailCest.php
@@ -69,8 +69,18 @@ class SendCategoryPurchaseEmailCest {
     $userEmail = Security::generateRandomString() . '-user@email.example';
     $i->orderProduct($product, $userEmail, true, false);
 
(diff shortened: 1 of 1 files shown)


tests/acceptance/SendCategoryPurchaseEmailCest.php
```diff
@@ -46,4 +46,31 @@ public function sendCategoryPurchaseEmail(\AcceptanceTester $i) {
...
+      ->withSubject($emailSubject)
+      ->withAutomaticTypeWooCommerceProductInCategoryPurchased([$product])
+      ->withActiveStatus()
+      ->create();
+
+    $userEmail = Security::generateRandomString() . '-user@email.example';
+    $i->orderProduct($product, $userEmail, true, false);
+
+    $i->amOnMailboxAppPage();
+    $i->dontSee($emailSubject);
```
costasovo, line 73: Hmm, this is tricky. It takes some time to deliver the email (see [code for checking it was delivered](https://github.com/mailpoet/mailpoet/blob/master/tests/_support/AcceptanceTester.php#L466)) so if we check immediately after the order was placed the test will pass even though the email was actually sent because we were checking too early.
I guess we will have to wait some time before checking the mailbox, but this will make the test slow.
Maybe we could use [Codeception DB module](https://codeception.com/docs/modules/Db) and create a helper that would check scheduled tasks table and it would wait until there are no scheduled or running sending tasks.
Somethin like `$i->waitUntilAllSendingIsDone()` and we could check the mailbox after that.

costasovo (changes requested):

tests/acceptance/SendCategoryPurchaseEmailCest.php
veljkho: Hi @costasovo,
I understand, so I added one private function to check if proper email is scheduled for sending in our Help section.
Tested also negative case. So, I hope this method will be fine at least for now, as quick solution to this test.

tests/acceptance/SendCategoryPurchaseEmailCest.php
```diff
@@ -46,4 +46,41 @@ public function sendCategoryPurchaseEmail(\AcceptanceTester $i) {
...
+    $i->orderProduct($product, $userEmail, true, false);
+
+    $this->verifyNoScheduledSending($i);
+    
+    $i->amOnMailboxAppPage();
+    $i->dontSee($emailSubject);
+    $i->dontSee($userEmail);
+  }
+
+  private function verifyNoScheduledSending(\AcceptanceTester $i) {
```
costasovo, line 79: I guess you can pass $emailSubject as a second argument to this function and you won't have to redeclare it inside the function's body. It will make it more reusable.

costasovo (changes requested):

tests/acceptance/SendCategoryPurchaseEmailCest.php
veljkho: Yeah, that's good point. I updated it.

costasovo (approved):

PR #3001, opened 2020-08-26
Disable full group by [MAILPOET-3115]

costasovo:


wxa (approved):
Works on MySQL. @veljkho Could you please test this on a host with MariaDB if possible?

veljkho (approved):

PR #3003, opened 2020-08-26
Add 2 new scenarios to newsletter preview testcases

veljkho:
[MAILPOET-3118](https://mailpoet.atlassian.net/browse/MAILPOET-3118)

### DIFF
--- a/tests/acceptance/PreviewPostNotificationNewsletterCest.php
+++ b/tests/acceptance/PreviewPostNotificationNewsletterCest.php
@@ -2,9 +2,19 @@
 
 namespace MailPoet\Test\Acceptance;
 
+use Codeception\Util\Locator;
 use MailPoet\Test\DataFactories\Newsletter;
+use MailPoet\Test\DataFactories\Settings;
 
 class PreviewPostNotificationNewsletterCest {
+
+  /** @var Settings */
+  private $settings;
+
+  public function _before() {
+    $this->settings = new Settings();
+  }
+
   public function previewPostNotificationNewsletter(\AcceptanceTester $i) {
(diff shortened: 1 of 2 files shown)


tests/acceptance/PreviewPostNotificationNewsletterCest.php
```diff
@@ -18,4 +18,33 @@ public function previewStandardNewsletter(\AcceptanceTester $i) {
...
+    $i->waitForText('Send preview');
+    $i->click('Send preview');
+    $i->waitForText('Your test email has been sent!');
+    $i->click('.notice-dismiss');
+    $i->clearField('#mailpoet_preview_to_email');
+    $i->click('Send preview');
+    $i->waitForText('Enter an email address to send the preview newsletter to.');
+    $i->fillField('#mailpoet_preview_to_email', 'test2@test.com');
+    $i->click('Send preview');
+    $i->waitForText('Your test email has been sent!');
```
wxa, line 43: Why do we need to send the preview email twice? Looks like the first one may be eliminated if I'm not missing something. 

It also might make sense to check that a preview email was received using the `checkEmailWasReceived` method, see examples in other test cases.

veljkho: Regarding twice sending check - we had issue in the past that only sending to one email was possible, now with this check we should cover it.
Thanks for your suggestion, I have added `checkEmailWasReceived` at the end of both tests, didn't want to include it in the middle of testing, I think at the end is just fine. Also, I added comments in my tests, so now it is more clear for everyone.

wxa (approved):

PR #3006, opened 2020-08-28
Expand display options - refactor data structure [MAILPOET-3072]

pavel-mailpoet:
[MAILPOET-3072]

@veljkho This is just a prepare step, nothing is changing during this ticket. It is just a refactor. Please test all the form placement options to see if nothing breaks. It should also migrate old forms to the new structure so please test some existing and some new forms. You need to adjust the version number in your database again.

[MAILPOET-3072]: https://mailpoet.atlassian.net/browse/MAILPOET-3072


lib/Config/Populator.php
```diff
@@ -738,6 +740,68 @@ private function addPlacementStatusToForms() {
...
+  private function migrateFormPlacement() {
+    if (version_compare($this->settings->get('db_version', '3.49.1'), '3.49.0', '>')) {
+      return;
+    }
+    $forms = $this->formsRepository->findAll();
+    foreach ($forms as $form) {
+      $settings = $form->getSettings();
+      $settings['form_placement'] = [
+        FormEntity::DISPLAY_TYPE_POPUP => [
+          'enabled' => $settings['form_placement_popup_enabled'],
```
costasovo, line 753: I see we don't unset these outdated properties in migration. They will be removed on form save. Am I correct?

costasovo (changes requested):
Looks Good! I just think that with this change we also need to update form templates we already have.
We probably need just to remove these lines from all templates: https://github.com/mailpoet/mailpoet/blob/master/lib/Form/Templates/Templates/Template10BelowPages.php#L179-L191

and should also remove placement styles if they are set to default values like [these](https://github.com/mailpoet/mailpoet/blob/master/lib/Form/Templates/Templates/Template10BelowPages.php#L199-L228). If there are some non-default values we need to rewrite those to new data structure.

lib/Config/Populator.php
```diff
@@ -798,6 +798,23 @@ private function migrateFormPlacement() {
...
+      if (isset($settings['place_popup_form_on_all_posts'])) unset($settings['place_popup_form_on_all_posts']);
+      if (isset($settings['popup_form_delay'])) unset($settings['popup_form_delay']);
+      if (isset($settings['place_fixed_bar_form_on_all_pages'])) unset($settings['place_fixed_bar_form_on_all_pages']);
+      if (isset($settings['place_fixed_bar_form_on_all_posts'])) unset($settings['place_fixed_bar_form_on_all_posts']);
+      if (isset($settings['fixed_bar_form_delay'])) unset($settings['fixed_bar_form_delay']);
+      if (isset($settings['fixed_bar_form_position'])) unset($settings['fixed_bar_form_position']);
+      if (isset($settings['place_slide_in_form_on_all_pages'])) unset($settings['place_slide_in_form_on_all_pages']);
+      if (isset($settings['place_slide_in_form_on_all_posts'])) unset($settings['place_slide_in_form_on_all_posts']);
+      if (isset($settings['slide_in_form_delay'])) unset($settings['slide_in_form_delay']);
+      if (isset($settings['slide_in_form_position'])) unset($settings['slide_in_form_position']);
```
costasovo, line 817: I think we need to add unset also for outdated placement styles properties:
below_post_styles
slide_in_styles
fixed_bar_styles
popup_styles
other_styles

costasovo (approved):

veljkho (changes requested):
Some problems occured
**1.a**: When I try to change form's width values in the Preview screen, nothing has changed in the preview actually (it's frozen)
**1.b**: The same happens if I try to move form from left-righ or top-bottom (slide-in, fixedbar)
**2.a**: Also, if I publish the form with fixed bar - on the front end it will be centered and not on the top as it should
**2.b**: And, if I try to publish slide-in (right), it will be on the front end something like in the middle of the screen - [screenshot for both 2 a & b](https://i.imgur.com/wfNpSHD.png)
**3**. In my local environment, blank page for the whole form editor is present if I click on Style expandable block in the settings (but this is not present on my online environment, at prod.qawp.net)

All issues except 3rd one can be reproduced at prod.qawp.net (admin / arabica99!)
I changed db_version in database to 3.44.0 and performed deactivation/activation.

costasovo (approved):

veljkho (changes requested):
Unfortunately,
I have found another issue, but it probably existed already in previous testing.
In Forms list page, all forms are Type: Others (widget), no matter what type you've set inside. This is for both new and old forms. Good is that issue is not present inside form editor or on the front end (so issue in listing only).

costasovo (approved):

veljkho (approved):

PR #3007, opened 2020-08-28
Newsletter preview move [MAILPOET-2981]

pavel-mailpoet:
[MAILPOET-2981]

[MAILPOET-2981]: https://mailpoet.atlassian.net/browse/MAILPOET-2981


tests/acceptance/PreviewPostNotificationNewsletterCest.php
```diff
@@ -41,26 +41,28 @@ public function previewPostNotificationNewsletterInEditor(\AcceptanceTester $i)
...
+    $i->waitForElement('[data-automation-id="switch_send_to_email"]');
+    $i->click('[data-automation-id="switch_send_to_email"]');
     $i->clearField('#mailpoet_preview_to_email');
     $i->click('Send preview');
     $i->waitForText('Enter an email address to send the preview newsletter to.');
     //set different email and test sending
     $i->fillField('#mailpoet_preview_to_email', 'test2@test.com');
     $i->click('Send preview');
     $i->waitForText('Your test email has been sent!');
-    //check the preview email in browser
```
wxa, line 58: Why was this test section removed?

tests/javascript_newsletter_editor/newsletter_editor/components/sidebar.spec.js
```diff
@@ -180,15 +180,4 @@ describe('Sidebar', function () {
       });
     });
   });
-
-  describe('preview view', function () {
```
wxa, line 184: Why don't we move this test to `save.spec.js` instead of deleting it?

wxa (changes requested):
There are a few issues with the "Send to email" tab:

1. No loading animation when clicking "Send preview", makes it kind of frustrating to guess whether form submission worked or not.
2. No disabling of the "Send preview" button, I can click it a dozen times, each request is heavy on the server.
3. Problems with error handling: when I simulate a 404/500 error or switch off the internet connection/shut down my local server, no errors are displayed & the modal is not closed as said in specs. Just nothing happens.

tests/acceptance/PreviewPostNotificationNewsletterCest.php
pavel-mailpoet: Because it doesn't make sense any more. The button `View in browser` has been removed and the preview modal works - we opened it a couple of lines above.

wxa (commented):
It works sufficiently well now, but what happens after preview sending is not according to specs. Spec 3.c says: 
> If there is an error when sending a preview and a notice is shown, close modal window so the notice is visible.

Spec 3.d.ii:
> After successful sending, show another sentence below (with 1 empty line between them): Didn't receive the test email? Read our [link]quick guide[/link] to sending issues.

It actually does the opposite: closes the modal and shows a notice on success, displays the "Didn't receive ..." message on failure without closing the modal. This doesn't seem critical to me, so maybe specs could be altered instead, let's ask @NeosinneR .

NeosinneR:
@wxa @pavel-mailpoet the error message showing in the modal is fine (maybe even better, not to close the modal), but it should probably say something different than `Didn't receive...` because there is nothing to receive - I guess there is specific error why it wasn't sent. 

In case of success message, the `Didn't receive...` message should be shown. Because if MailPoet is misconfigured and the user expects an email, we need to let him know why he didn't receive it. In this case, both "in modal" and "close modal and show notice" are fine.

pavel-mailpoet:
ok. I misread the specs and implemented it the other way round. I leave the modal opened and display either success message or an error. I took the error message from the previous implementation.

wxa:
Thank you for the fixes. We're almost there, just a couple visual improvements needed:
1. Please add a `mailpoet_success` class to the "Your test email has been sent!" message to make it green and more visible.
2. A long error message in a small browser window is clipped, can we stretch the modal or add scrolling in this case? See the screenshot:
изображение

pavel-mailpoet:
@wxa I cannot replicate the problem with a small browser window. Scrolling works for me normally: https://d.pr/v/oyA6oM

wxa:
@pavel 
>  I cannot replicate the problem with a small browser window

It happens if the browser window itself is resized down or if the browser console is open, not in the mobile screen emulation mode.

wxa (dismissed):

veljkho (changes requested):
Everything's good, except:
- After entering different email in **Send to email** field.. and pressing Send button: I see my own email address in the field while sending is in the progress. The good thing is that it has been sent to proper address, so this is just visual issue.
- Suggestion: Maybe is better to see "quick guide" opened inside Beacon and not as external link.

wxa (approved):

veljkho (approved):

PR #3012, opened 2020-09-01
Add Categories component [MAILPOET-3082]

wxa:
[MAILPOET-3082](https://mailpoet.atlassian.net/browse/MAILPOET-3082)


assets/js/src/common/categories/categories.tsx
```diff
@@ -0,0 +1,71 @@
...
+import PropTypes from 'prop-types';
+
+type Category = {
+  name: string,
+  label: string,
+  count?: number | string,
+};
+
+type Props = {
+  onSelect: (category: any) => any,
```
amine-mp, line 12: I see you are calling the `onSelect` with the `category.name` so its argument is `name: string`, right?

amine-mp (changes requested):
Good job, I just have some remarks.

wxa:
@amine-mp Nice remarks, all were addressed.

assets/js/src/common/categories/categories_item.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
```
amine-mp, line 4: This can be written as
```
type Props = Category & {
  onSelect: (name: string) => any,
  active?: boolean
}
```
This way, if we change the `Category` type, this will update automatically.

wxa: Fixed.

amine-mp (approved):

veljkho (approved):

PR #3014, opened 2020-09-01
Redesign send with [MAILPOET-1705]

amine-mp:
[MAILPOET-1705]

[MAILPOET-1705]: https://mailpoet.atlassian.net/browse/MAILPOET-1705


wxa (changes requested):
Specs 5.a.i.2 and 5.a.ii.2 say that when a key is not valid for MSS:
> Choosing the radio input will switch to Key activation tab

This does not happen: nothing changes when I click the MSS radio input with an invalid key.

wxa (dismissed):

veljkho (changes requested):
Everything's good except one thing:
- I see all the time "Solve all of your sending problems" message but even after I fixed everything. Seems the message is stuck always.
Hint: Maybe it was caused by above message for inappropriate version? [Screenshot](https://i.imgur.com/iYgMCYv.png)
Are they related? If yes, than this issue might not be valid.

NeosinneR:
@veljkho I'm not sure I understand it correctly, but `Solve all of your sending problems` is a description for MSS option, it should be always visible

NeosinneR:
@veljkho I stand corrected, I haven't realized, that currently, when MSS is active, there is `You're now sending with MailPoet!` instead. We should keep the same functionality even after the redesign. I'll update specs

wxa (approved):

veljkho (approved):

PR #3021, opened 2020-09-02
Fix too much space at the end of form [MAILPOET-3090]

pavel-mailpoet:
[MAILPOET-3090]

[MAILPOET-3090]: https://mailpoet.atlassian.net/browse/MAILPOET-3090

### DIFF
--- a/lib/Form/Util/Styles.php
+++ b/lib/Form/Util/Styles.php
@@ -113,7 +113,9 @@ class Styles {
 
     $messagesStyles = $this->renderMessagesStyles($formSettings, $selector);
 
-    $additionalStyles = $selector . ' .mailpoet_paragraph.last {margin-bottom: 0}';
+    $additionalStyles = $selector . ' .mailpoet_paragraph.last {margin-bottom: 0} ';
+    $media .= " @media (min-width: 500px) {{$selector} .last .mailpoet_paragraph:last-child {margin-bottom: 0}} ";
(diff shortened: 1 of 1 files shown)


lib/Form/Templates/FormTemplate.php
```diff
@@ -39,6 +39,10 @@ public function getStyles(): string {
   margin-bottom: 20px;
 }
 
+.last .mailpoet_paragraph {
```
costasovo, line 42: I think it would be nicer to have it in one place here https://github.com/mailpoet/mailpoet/blob/master/lib/Form/Util/Styles.php#L116

costasovo (dismissed):

veljkho (changes requested):
We have new issue now. If we place inside columns block, blocks like email, subscribe, first name or image.. they will be without space at bottom of each block, looking merged together.. on both mobile and web view.
[Screenshot](https://i.imgur.com/TEAdRKO.png)

costasovo (approved):

veljkho (changes requested):
Issue partially verified.
We have still problem on mobile, but in this case:
- Column block 50-50
- Place email block in left place, subscribe button in right place (inside that column block)
- Preview it in mobile and observe the same issue with missing spacing bottom

All other verified

costasovo (approved):

veljkho (approved):

PR #3024, opened 2020-09-03
Forms: add "exit intent" [MAILPOET-3057]

pavel-mailpoet:
[MAILPOET-3057]

[MAILPOET-3057]: https://mailpoet.atlassian.net/browse/MAILPOET-3057


assets/js/src/public.jsx
```diff
@@ -82,13 +91,16 @@ jQuery(($) => {
...
-      if (showOverlay) {
-        formDiv.prev('.mailpoet_form_popup_overlay').addClass('active');
-      }
+      doDisplayForm(formDiv, showOverlay);
     }, delay * 1000);
+
+    const exitIntentEnabled = form.data('exit-intent-enabled');
+    if (exitIntentEnabled) {
+      $(document).on('mouseleave', () => {
+        doDisplayForm(formDiv, showOverlay);
```
costasovo, line 100: I think we need to clear the timeout for displaying popup here otherwise the popup will be displayed again after the timeout.

assets/js/src/public.jsx
```diff
@@ -82,13 +91,16 @@ jQuery(($) => {
...
+        $(document).off('mouseleave');
```
costasovo, line 101: Based on [docs](https://api.jquery.com/off/) this will unbind all `mouseleave` handlers from document. I think we may potentially break some other plugin. I guess we need to pass a handler as second argument here or maybe more elegant solution would be using [event namespaces](https://api.jquery.com/on/#event-names) so that we just rename `mouseleave` to `mouseleave.mailpoet` according to the docs this should work

costasovo (changes requested):

costasovo (approved):

veljkho (approved):

PR #3025, opened 2020-09-03
Update tinymce [MAILPOET-3098]

pavel-mailpoet:
[MAILPOET-3098]

[MAILPOET-3098]: https://mailpoet.atlassian.net/browse/MAILPOET-3098


assets/js/src/newsletter_editor/behaviors/tinymce_icons.js
```diff
@@ -0,0 +1,7 @@
+
```
wxa, line 1: Should this file be ignored if it is created after `npm install`?

wxa (approved):

veljkho (approved):

PR #3027, opened 2020-09-04
Form templates - Batch 3 [MAILPOET-3076]

costasovo:
### Note for QA

When testing this locally you need to turn off CDN (`define('MAILPOET_USE_CDN', false);`) to see form assets.


lib/Form/Templates/Templates/Template7FixedBar.php
```diff
@@ -0,0 +1,290 @@
...
+                  'background_color' => '',
+                  'anchor' => '',
+                  'class_name' => '',
+                ],
+              ],
+              [
+                'type' => 'paragraph',
+                'id' => 'paragraph',
+                'params' => [
+                  'content' => _x('<span style="font-family: Cairo" data-font="Cairo" class="mailpoet-has-font">Let us do the hard work for you. Sign up to receive our latest deals directly in your inbox. We\'ll never send you spam - promise. Find out more in our <a href="https://www.mailpoet.com/privacy-notice/">Privacy Policy</a>.</span>', 'Text in a web form. Keep HTML tags!', 'mailpoet'),
```
pavel-mailpoet, line 60: Two things here.
1) Should we clear the link here? I think the idea is for the user to add their own privacy policy.
2) Could we not use this syntax? `We\'ll` It is hard to search for it when you need to do a full-text search in the source code.

pavel-mailpoet (changes requested):

pavel-mailpoet (changes requested):
See my comment

pavel-mailpoet (approved):

costasovo:
@pavel-mailpoet I added 2 more commits with some fixes in the template data.

pavel-mailpoet (approved):

veljkho (approved):

PR #3031, opened 2020-09-08
Redesign email type selection page [MAILPOET-2786]

wxa:


NeosinneR (approved):
@veljkho there are small visual differences (not centered, `Woo` icon is full-width, close button sometimes overlaying the box) but because this page shares a bunch of styles with [[MAILPOET-2788]](https://mailpoet.atlassian.net/browse/MAILPOET-2788) (which I'm working on) I'll fix those issue in my ticket to prevent conflicts. It requires bigger refactor for both types and templates pages.

veljkho (approved):

PR #3036, opened 2020-09-08
Detect whitespace in output that can break tracking images and captcha [MAILPOET-2983]

wxa:
[MAILPOET-2983](https://mailpoet.atlassian.net/browse/MAILPOET-2983)

I formatted the captcha message as a separate sentence for easier translation.

**For QA:**

To test this, try adding `?>` followed by some whitespace at the end of theme's `functions.php` or in some 3rd-party plugin files.

### DIFF
--- a/lib/Util/Notices/HeadersAlreadySentNotice.php
+++ b/lib/Util/Notices/HeadersAlreadySentNotice.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace MailPoet\Util\Notices;
+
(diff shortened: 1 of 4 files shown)


NeosinneR (dismissed):

wxa:
@NeosinneR Added one more check for a case when output buffering is on, `headers_sent` won't work in this case.

NeosinneR (approved):

veljkho (approved):

PR #3037, opened 2020-09-08
Expand display options [MAILPOET-3120]

pavel-mailpoet:
[MAILPOET-3120]

[MAILPOET-3120]: https://mailpoet.atlassian.net/browse/MAILPOET-3120

### DIFF
--- a/assets/js/src/form_editor/store/map_form_data_after_loading.jsx
+++ b/assets/js/src/form_editor/store/map_form_data_after_loading.jsx
@@ -135,7 +135,6 @@ export default function mapFormDataAfterLoading(data) {
       backgroundImageUrl: data.settings.background_image_url,
       backgroundImageDisplay: data.settings.background_image_display,
       closeButton: data.settings.close_button ?? defaults.formStyles.closeButton,
-
     },
   };
 
--- a/assets/js/src/form_editor/store/map_form_data_after_loading.jsx
+++ b/assets/js/src/form_editor/store/map_form_data_after_loading.jsx
@@ -20,9 +20,15 @@ export default function mapFormDataAfterLoading(data) {
           styles: { ...defaults.popupForm.styles, ...data.settings.form_placement?.popup?.styles },
           posts: {
             all: data.settings.form_placement?.popup?.posts?.all === '1',
+            selected: Array.isArray(data.settings.form_placement?.popup?.posts?.selected)
+              ? data.settings.form_placement?.popup?.posts?.selected
+              : [],
           },
           pages: {
             all: data.settings.form_placement?.popup?.pages?.all === '1',
+            selected: Array.isArray(data.settings.form_placement?.popup?.pages?.selected)
+              ? data.settings.form_placement?.popup?.pages?.selected
+              : [],
           },
         },
         fixedBar: {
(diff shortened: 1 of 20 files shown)


assets/js/src/form_editor/store/map_form_data_after_loading.jsx
```diff
@@ -17,9 +17,15 @@ export default function mapFormDataAfterLoading(data) {
           styles: { ...defaults.popupForm.styles, ...data.settings.form_placement?.popup?.styles },
           posts: {
             all: data.settings.form_placement?.popup?.posts?.all === '1',
+            selected: Array.isArray(data.settings.form_placement?.popup?.posts?.selected)
```
costasovo, line 22: Shouldn't we add this property to mapper test? I think it could be handy if we decide to refactor the mapper in the future.

tests/acceptance/FormEditorPlaceFormOnSpecifiedPageCest.php
```diff
@@ -0,0 +1,42 @@
...
+    $i->checkOption('Enable');
+    $i->waitForText('Display on all posts');
+    $i->selectOptionInSelect2($pageTitle, '[data-automation-id="form-placement-select-page"] input.select2-search__field');
+
+    // Save form
+    $i->saveFormInEditor();
+
+    // check the form
+    $i->amOnUrl($pageUrl);
+    $i->wantTo('Place form on a specific page');
```
costasovo, line 39: Shouldn't this be at the beginning of the test?

costasovo (approved):

veljkho (changes requested):
Some issues found:
- Field titles are not in proper position, please refer to this [screenshot](https://i.imgur.com/xwggziO.png)
- Unable to search for WooCommerce product's categories and tags
- Unable to search for posts in Select field. Seems like only possible to search for posts that are present in dropdown menu when you click on the field (5 results already shown if you click).
Note: Maybe issue here is that only last 5 posts are present in Select's dropdown if you click it, because it is like that, only last 5 posts present... but for Page's select field I see all pages are present actually..
- Posts are usually with long titles. When I select them [this is how it looks](https://i.imgur.com/xsUMPpr.png). All fields are extended in the sidebar. But I believe this is how it works, and we dont need to do nothing about it. Maybe if we could make it truncated... shorted, or something, but fine as is if no any ideas to improve it.

lib/Form/DisplayFormInWPContent.php
```diff
@@ -187,10 +187,10 @@ private function shouldDisplayFormType(FormEntity $form, string $formType): bool
       return false;
     }
 
-    if ($this->wp->isSingular('post')) {
+    if ($this->wp->isSingular('post') || $this->wp->isSingular('product')) {
```
costasovo, line 190: With this change when a user selects `display on all posts` the form will be displayed also on all product pages.
I'm not sure if that an intended outcome. We put products into the selection of posts so I guess it makes sense, on the other hand, I'm not sure if users understand that + it will change the behaviour of already created forms.

Alternatively, we could display forms on product pages only if they are listed in select.
cc @veljkho

costasovo (approved):
Codewise looks good 👍 
I had just one comment which is more UX related. Leaving it to QA.

costasovo (approved):

veljkho (changes requested):
There are some issues:
1. Radio button for "Display on all pages" is turned OFF automatically if user do the following:
- Select some pages for form to be displayed on and save it
- Then just click radio button for "Display on all pages" and save it
- Observe issue while saving, radio button auto turned OFF in the meantime while saving
2. In Preview screen, labels like "Display on posts with these cats, tags" are too big (probably using improper component, since issue is not present in form editor).
3. In both Preview screen and Form Editor, label "Display with a delay of" is bolded. Should be normal text.

All other verified and functioning properly. [Screenshot](https://i.imgur.com/3XwrpNO.png) for 2. and 3.

costasovo (approved):

veljkho (approved):

PR #3041, opened 2020-09-09
Redesign - Emails - Congratulation page [MAILPOET-2833]

amine-mp:
[MAILPOET-2833]

[MAILPOET-2833]: https://mailpoet.atlassian.net/browse/MAILPOET-2833


wxa (approved):
Looks good. I've changed loader default size to 32px (was 34px) to match specs.

NeosinneR:
@wxa added a few commits, when you have time, please re-review this PR, thanks :)

wxa:
@NeosinneR One UX thing is annoying me: after clicking "Close" on the congrats page there is no visible reaction until the settings-saving AJAX request returns, there may be a lag of several seconds, so I'm tempted to click "Close" several times until I see that a page is loading. What about adding a standard loading overlay/animation?

NeosinneR:
@wxa added spinner straight on the button, which also disables clicking on it multiple times

wxa (approved):

veljkho (approved):

PR #3042, opened 2020-09-10
Add form animations [MAILPOET-3088]

pavel-mailpoet:
[MAILPOET-3088]

[MAILPOET-3088]: https://mailpoet.atlassian.net/browse/MAILPOET-3088

### DIFF
--- a/assets/js/src/public.jsx
+++ b/assets/js/src/public.jsx
@@ -89,6 +89,7 @@ jQuery(($) => {
     if (animation !== '') {
       formDiv.addClass(`mailpoet_form_animation_${animation}`);
       formDiv.prev('.mailpoet_form_popup_overlay').addClass(`mailpoet_form_overlay_animation_${animation}`);
+      formDiv.prev('.mailpoet_form_popup_overlay').addClass(`mailpoet_form_overlay_animation`);
     }
     const background = form.data('background-color');
     formDiv.css('background-color', background || 'white');
--- a/assets/js/src/public.jsx
+++ b/assets/js/src/public.jsx
@@ -89,7 +89,7 @@ jQuery(($) => {
     if (animation !== '') {
       formDiv.addClass(`mailpoet_form_animation_${animation}`);
       formDiv.prev('.mailpoet_form_popup_overlay').addClass(`mailpoet_form_overlay_animation_${animation}`);
-      formDiv.prev('.mailpoet_form_popup_overlay').addClass(`mailpoet_form_overlay_animation`);
+      formDiv.prev('.mailpoet_form_popup_overlay').addClass('mailpoet_form_overlay_animation');
     }
     const background = form.data('background-color');
     formDiv.css('background-color', background || 'white');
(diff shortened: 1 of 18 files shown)


assets/js/src/public.jsx
```diff
@@ -85,6 +85,10 @@ jQuery(($) => {
     const form = formDiv.find('form');
     const position = form.data('position');
     formDiv.addClass(`mailpoet_form_position_${position}`);
+    const animation = form.data('animation');
+    if (animation !== '') {
+      formDiv.addClass(`mailpoet_form_animation_${animation}`);
+    }
```
costasovo: Do we have to do this using data attributes?
Why we just don't set the class in Twig template file here: https://github.com/mailpoet/mailpoet/blob/master/views/form/front_end_form.html#L13 ?
I find it as simpler and nicer solution.

The same applies to `position` two lines above.
Or is there some reason why this has to be done in JS?

assets/js/src/public.jsx
```diff
@@ -85,6 +85,10 @@ jQuery(($) => {
...
     const background = form.data('background-color');
     formDiv.css('background-color', background || 'white');
```
costasovo, line 95: Not related to this PR but I think we can remove these lines since we already set this in CSS we generate here: https://github.com/mailpoet/mailpoet/blob/master/lib/Form/Util/Styles.php#L65-L78
We can probably also remove the data attribute `data-background-color="<%= backgroundColor %>"`

assets/js/src/public.jsx
```diff
@@ -88,6 +88,7 @@ jQuery(($) => {
     const animation = form.data('animation');
     if (animation !== '') {
       formDiv.addClass(`mailpoet_form_animation_${animation}`);
+      formDiv.prev('.mailpoet_form_popup_overlay').addClass(`mailpoet_form_overlay_animation_${animation}`);
```
costasovo, line 91: I guess we could set this class in twig view https://github.com/mailpoet/mailpoet/blob/master/views/form/front_end_form.html#L10

costasovo (changes requested):

lib/Form/DisplayFormInWPContent.php
```diff
@@ -156,7 +156,6 @@ private function getContentBellow(FormEntity $form, string $displayType): string
     $templateData['delay'] = $formSettings['form_placement'][$displayType]['delay'] ?? 0;
     $templateData['position'] = $formSettings['form_placement'][$displayType]['position'] ?? '';
     $templateData['animation'] = $formSettings['form_placement'][$displayType]['animation'] ?? '';
-    $templateData['backgroundColor'] = $formSettings['backgroundColor'] ?? '';
```
costasovo, line 158: I think we need to remove this from [the preview page](https://github.com/mailpoet/mailpoet/blob/master/lib/Form/PreviewPage.php#L100).

costasovo (changes requested):

costasovo (dismissed):

veljkho (changes requested):
Everything's good, I just missed in the specs to include also Slide Left.

costasovo (approved):

veljkho (approved):

PR #3047, opened 2020-09-15
Fix Normalizer prefixing [MAILPOET-3148]

wxa:
Please double-check that all necessary cases are prefixed now.

*For QA:*

If possible, please test this on a host where Intl extension functions are missing to make sure errors do not occur. Test with WooCommerce enabled.

### DIFF
--- a/prefixer/fix-symfony-polyfill.php
+++ b/prefixer/fix-symfony-polyfill.php
@@ -8,9 +8,16 @@ set_error_handler(function ($severity, $message, $file, $line) {
 $file = __DIR__ . '/../vendor-prefixed/symfony/polyfill-intl-idn/Idn.php';
 $data = file_get_contents($file);
 $data = str_replace('\\Normalizer::', '\\MailPoetVendor\\Normalizer::', $data);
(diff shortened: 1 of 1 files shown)


prefixer/fix-symfony-polyfill.php
```diff
@@ -8,9 +8,16 @@
 $file = __DIR__ . '/../vendor-prefixed/symfony/polyfill-intl-idn/Idn.php';
 $data = file_get_contents($file);
 $data = str_replace('\\Normalizer::', '\\MailPoetVendor\\Normalizer::', $data);
+$data = str_replace('use Normalizer;', 'use MailPoetVendor\\Normalizer;', $data);
```
costasovo, line 11: I think that this one is nor necessary since all usages of Normalizer within this file use fully qualified class name. But it makes no harm keeping it.

costasovo (approved):
Tested on a container with PHP 7.0 and Intl disabled. The plugin worked correctly and I was able to enable/disable it.

veljkho (approved):

PR #3049, opened 2020-09-17
Fixed three flaky tests

veljkho:
[MAILPOET-3153](https://mailpoet.atlassian.net/browse/MAILPOET-3153)

In this task I will fix flaky tests that keep failing in nightly build.

Flaky tests:

- EditorProductsCest:filterProducts (keep failing because warning message for PHP update is present)
- SettingsPageBasicsCest:allowSubscribeInComments (failing because require additional scroll)
- ManageListsCest:editTrashRestoreAndDeleteExistingList (failing because we redesigned fields recently)

### DIFF
--- a/tests/acceptance/SettingsPageBasicsCest.php
+++ b/tests/acceptance/SettingsPageBasicsCest.php
@@ -71,10 +71,12 @@ class SettingsPageBasicsCest {
     $i->click('Post Comment');
     $i->waitForText($comment, 10, '.comment-content');
     //check if user is really subscribed to a list
-    $i->amOnMailpoetPage('Subscribers');
+    $i->amOnMailpoetPage('Lists');
+    $i->waitForText('Lists');
+    $i->clickItemRowActionByItemName('Newsletter mailing list', 'View Subscribers');
     $i->waitForText('Subscribers');
+    $i->click('[data-automation-id="filters_unconfirmed"]');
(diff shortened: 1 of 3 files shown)


tests/acceptance/SettingsPageBasicsCest.php
```diff
@@ -61,17 +61,17 @@ public function allowSubscribeInComments(\AcceptanceTester $i) {
...
     $i->click('#comment');
     $i->fillField('#comment', $comment);
     $i->waitForText($optinMessage);
     $i->click('#mailpoet_subscribe_on_comment');
     $i->click('Post Comment');
     $i->waitForText($comment, 10, '.comment-content');
     //check if user is really subscribed to a list
-    $i->amOnMailpoetPage('Lists');
-    $i->clickItemRowActionByItemName('Newsletter mailing list', 'View Subscribers');
+    $i->amOnMailpoetPage('Subscribers');
```
pavel-mailpoet, line 74: This makes the test weaker. If the subscriber ends up in a wrong list the test still passes. How does it help with the flakiness?

veljkho: Yes, you're totally right! Why I changed that to be more weak, only God knows. Now I strengthed it. Thanks!

pavel-mailpoet (approved):

PR #3052, opened 2020-09-22
Add hooks for Twig cache path and disabling cache (for WP Engine) [MAILPOET-3157]

wxa:
[MAILPOET-3157]

[MAILPOET-3157]: https://mailpoet.atlassian.net/browse/MAILPOET-3157


flakas (changes requested):
Thanks for the PR.
Could you please add tests to ensure these hooks can be used, and that we will maintain these hooks?

flakas (approved):
Thanks!

veljkho (approved):

PR #3054, opened 2020-09-23
Refactor dynamic segments filtering to Doctrine [MAILPOET-3077]

costasovo:
[MAILPOET-3077]

[MAILPOET-3077]: https://mailpoet.atlassian.net/browse/MAILPOET-3077

## Note for QA
Please check all functionality related to dynamic segments thoroughly (all types). These changes are mostly related to fetching subscribers from both dynamic segments (Segments in UI) and static segments (Lists in UI) so it affects subscribers listings and fetching subscribers when sending emails.

It is probably a good idea to merge [a PR which fixes editation of WooCommerce dynamic segments](https://github.com/mailpoet/mailpoet/pull/3058) first and rebase this PR before checking.


lib/Entities/SegmentEntity.php
```diff
@@ -105,4 +105,8 @@ public function setDescription($description) {
   public function getDynamicFilters() {
     return $this->dynamicFilters;
   }
+
+  public function isDynamic(): bool {
```
pavel-mailpoet, line 109: Why do you think it is better to use this function instead of directly comparing the type?

lib/Entities/SegmentEntity.php
```diff
@@ -109,4 +109,8 @@ public function getDynamicFilters() {
   public function isDynamic(): bool {
     return $this->getType() === self::TYPE_DYNAMIC;
   }
+
+  public function isStatic(): bool {
```
pavel-mailpoet, line 113: I am not sure why do we need this.

lib/Subscribers/SubscriberListingRepository.php
```diff
@@ -183,4 +225,66 @@ public function getFilters(ListingDefinition $definition): array {
     }
     return ['segment' => $segmentList];
   }
+
+  private function sanitizeSearch(string $search): string {
```
pavel-mailpoet, line 229: I think this should be a public function somewhere.

lib/AdminPages/Pages/Newsletters.php
```diff
@@ -69,7 +73,8 @@ public function __construct(
     FeaturesController $featuresController,
     SubscribersFeature $subscribersFeature,
     ServicesChecker $servicesChecker,
-    NewsletterTemplatesRepository $newsletterTemplatesRepository
+    NewsletterTemplatesRepository $newsletterTemplatesRepository,
+    AddToNewslettersSegments $dynamicSegmentsLoader
```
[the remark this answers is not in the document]
costasovo: You are right, the loader is not a very good name. I'll rename it.

lib/Entities/SegmentEntity.php
```diff
@@ -43,6 +44,19 @@ class SegmentEntity {
    */
   private $description;
 
+  /**
+   * @ORM\OneToMany(targetEntity="MailPoet\Entities\DynamicSegmentFilterEntity", mappedBy="segment")
+   * @var DynamicSegmentFilterEntity[]|ArrayCollection
```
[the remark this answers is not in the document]
costasovo: Dynamic segments and segments share the same entity for now. We could use [single table inheritance](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/inheritance-mapping.html#single-table-inheritance) to create a different entities for static and dynamic segments. I think it is a bit overkill since we have only one additional property. I also noticed that in some places in the old code we kind of expect that there might be some unknown segment type and this may cause issues with @DiscriminatorMap since it must contain complete list.
I also think it is not a big deal if Segment which is not dynamic has an empty list of filters.

lib/Entities/SegmentEntity.php
costasovo: This condition is used in [SubscriberFinder](https://github.com/mailpoet/mailpoet/blob/master/lib/Segments/SubscribersFinder.php#L53) and I added the same one to a different place so I put it in this method so that it is defined on one place.

lib/Entities/SegmentEntity.php
costasovo: It is not used anywhere. Will remove it.

costasovo:
@pavel-mailpoet I added all changes as new commits. This is the first one https://github.com/mailpoet/mailpoet/pull/3054/commits/650818586e8255a0489c780df22ac3c35435804f

pavel-mailpoet (approved):
Great job @costasovo. It was a lot of work, but it is a big step in the right direction.

costasovo:
@pavel-mailpoet I added one more commit with a fix for an issue I found when testing locally.

pavel-mailpoet (dismissed):

veljkho (approved):

PR #3057, opened 2020-09-23
Improved scenario to pass multisite testing

veljkho:
Details at [MAILPOET-3171](https://mailpoet.atlassian.net/browse/MAILPOET-3171)

### DIFF
--- a/tests/acceptance/CreateNewWordPressUserCest.php
+++ b/tests/acceptance/CreateNewWordPressUserCest.php
@@ -29,12 +29,13 @@ class CreateNewWordPressUserCest {
     $i->click('[data-automation-id="settings-submit-button"]'); //save settings
 
     // create a wp user via registration
+    // Note: Xpath used to avoid flakyness and to pass multisite testing where we have different registration page designs
     $i->logOut();
     $i->amOnUrl(\AcceptanceTester::WP_URL . '/wp-login.php?action=register');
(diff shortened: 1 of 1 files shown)


tests/acceptance/CreateNewWordPressUserCest.php
```diff
@@ -29,12 +29,13 @@ public function sendConfirmationEmail(\AcceptanceTester $i) {
     $i->click('[data-automation-id="settings-submit-button"]'); //save settings
 
     // create a wp user via registration
+    // Note: Xpath used to avoid flakyness and to pass multisite testing where we have different registration page designs
     $i->logOut();
     $i->amOnUrl(\AcceptanceTester::WP_URL . '/wp-login.php?action=register');
-    $i->fillField('#user_login', 'newuser');
-    $i->fillField('#user_email', 'newuser@test.com');
+    $i->fillField(['xpath' => "//input[@type='text'][contains(@name, 'user')]"], 'newuser');
```
wxa, line 35: I would go for `contains(@name, 'login')` because both login and email fields have `user` in their names, only field types differ which are more likely to change.

veljkho: Unfortunately, that wouldn't work because on multisite test there's user_name and on WP default is user_login. So I used user to bind them.

wxa (approved):

PR #3066, opened 2020-09-29
Forms::get() API refactored to Doctrine [MAILPOET-3035]

lysyjan:
[MAILPOET-3035]

[MAILPOET-3035]: https://mailpoet.atlassian.net/browse/MAILPOET-3035


wxa (approved):
Looks good!

Note: looks like the `Forms::get()` JSON API method is unused as of now, but it will be used later in other Forms API calls.

veljkho (approved):

PR #3067, opened 2020-09-30
Update supported WordPress version

lysyjan:
[MAILPOET-3176]

[MAILPOET-3176]: https://mailpoet.atlassian.net/browse/MAILPOET-3176

### DIFF
--- a/mailpoet.php
+++ b/mailpoet.php
@@ -7,8 +7,8 @@
  * Description: Create and send newsletters, post notifications and welcome emails from your WordPress.
  * Author: MailPoet
  * Author URI: http://www.mailpoet.com
- * Requires at least: 5.2
- * Tested up to: 5.4
+ * Requires at least: 5.3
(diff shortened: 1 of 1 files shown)


mailpoet.php
```diff
@@ -31,7 +31,7 @@ function mailpoet_deactivate_plugin() {
 }
 
 // Check for minimum supported WP version
-if (version_compare(get_bloginfo('version'), '4.6', '<')) {
+if (version_compare(get_bloginfo('version'), '5.3', '<')) {
```
costasovo, line 34: I think we should be more conservative here. We still have a bunch of users with 5.2 < who updates. WP itself shows some warnings.
The plugin should still work 5.0+ so I would go with 5.0 as a first step

costasovo (changes requested):

costasovo (approved):

veljkho (approved):

PR #3072, opened 2020-09-30
Select form template [MAILPOET-2988]

costasovo:
[MAILPOET-2988]

[MAILPOET-2988]: https://mailpoet.atlassian.net/browse/MAILPOET-2988


assets/js/src/form_editor/templates/selection.tsx
```diff
@@ -1,47 +1,72 @@
...
 import { Button } from '@wordpress/components';
-
-import { TemplateType } from './store/types';
+import Categories from 'common/categories/categories';
+import Background from 'common/background/background';
+import { TemplateData } from './store/types';
 
 export default () => {
-  const templates: Array<TemplateType> = useSelect(
+  const [selectedCategory, setSelectedCategory] = useState('popup');
```
pavel-mailpoet, line 13: Why do you keep the selected category in the local state rather than passing it to the store?

assets/css/src/components-form-editor/_template-selection.scss
```diff
@@ -1,3 +1,22 @@
-.template-selection {
-  margin: $grid-gap-large;
+.mailpoet-templates {
+  display: grid;
+  grid-gap: 16px;
```
pavel-mailpoet: Why do you use hardcoded values rather than already predefined values in this file? Also, some values are dependant on others so we should calculate them rather than hardcoding them.

assets/js/src/form_editor/templates/components/template_box.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { TemplateType } from 'form_editor/templates/store/types';
+
+type Props = {
+  template: TemplateType,
+  onSelect: (id: string) => any,
+}
+const TemplateBox = ({ template, onSelect }: Props) => (
+  <div className="mailpoet-template mailpoet-form-template" data-automation-id="select_template_box">
+    <div className="mailpoet-template-thumbnail">
+      <img src={template.thumbnail} alt={template.name} />
```
pavel-mailpoet: Should we specify the dimension here so that the browser can calculate the space before loading the image? Also, we could add an argument `loading="lazy"` for more optimal loading...

assets/js/src/form_editor/templates/selection.tsx
```diff
@@ -46,6 +46,11 @@ export default () => {
   const { selectTemplate } = useDispatch('mailpoet-form-editor-templates');
   return (
     <>
+      {categories.map((category) => (
+        templates[category.name].map((template) => (
+          <link key={`thumbnail_prefetch_${template.id}`} rel="preload" href={template.thumbnail} as="image" />
```
pavel-mailpoet, line 58: I am not sure if this is the best way to preload the images. Chrome is yelling at us: https://d.pr/i/yj6kZX

pavel-mailpoet (changes requested):

assets/js/src/form_editor/templates/selection.tsx
costasovo: I'll reduce the preload so that we load only 4 thumbnails per category. That will reduce the number of requests (warnings) and the user should still see images instantly.

pavel-mailpoet (dismissed):

veljkho (changes requested):
Everything's good, except:
- Sort order of template filter is not accordin to specs (expected: popup, slidein, fixedbar, belowp, widget)
- Some form families have 0s display delay option as default (affected: Newsletter Signup, Fitness Tips, Keep in Touch)
- Widget thumbnails are not correct (wrong color) I will provide you with better ones

costasovo:
@veljkho To see up to date thumbnail images you have to use `define('MAILPOET_USE_CDN', false);`

pavel-mailpoet (approved):

veljkho (approved):

PR #3080, opened 2020-10-01
Form templates part 4 [MAILPOET-3190]

costasovo:
[MAILPOET-3190]

[MAILPOET-3190]: https://mailpoet.atlassian.net/browse/MAILPOET-3190


pavel-mailpoet (approved):

veljkho (changes requested):
Some corrections needed:
- Priority List (Popup) - Display on all posts/pages radio buttons turned ON by default
- I forgot to add form editor's custom css `.wp-block-column:not(:first-child),` to add padding inside editor only for the following forms: **Test 12 #Popup** & **Test 12 #SlideIn** only. So I added now and please copy/paste from mailpoet.info site.

lib/Form/Block/Heading.php
```diff
@@ -82,6 +82,8 @@ private function renderStyle(array $block): string {
     }
     if (!empty($block['params']['line_height'])) {
       $styles[] = 'line-height: ' . $block['params']['line_height'];
+    } elseif (!empty($block['params']['font_size'])) {
+      $styles[] = 'line-height: 1.4;';
```
pavel-mailpoet, line 86: This overrides a value from custom css. I think we should use a different approach.

pavel-mailpoet (changes requested):

pavel-mailpoet (dismissed):

pavel-mailpoet (dismissed):

pavel-mailpoet (dismissed):

veljkho (approved):

PR #3082, opened 2020-10-05
Add WooCommerce abandoned cart content block [MAILPOET-2979]

wxa:
[MAILPOET-2979](https://mailpoet.atlassian.net/browse/MAILPOET-2979)


pavel-mailpoet (commented):
@wxa I have nothing major to say about your pull request. Just a couple of stuff to consider:

- `abandonedCartContentTransformedProducts`  and `automatedLatestContentTransformedPosts` could maybe go to a separate class outside of block renderer
- because sending and sending preview we have to use a lot of optional boolean arguments. I don't think the code `$this->renderer->render($newsletter, false, false, $sendingTask);` is very readable. I think it would be better to have a method `renderAsPreview` and `render` to make stuff more obvious when reading in the future. You can leave the boolean arguments inside a class. They are ok if they are in a single class
- I am not sure about all the `meta`. I don't have a better solution but such a vague name adds a complexity. When a new developer will come to this code in the future they will need a lot of time to debug what all the stuff does.

wxa:
> `abandonedCartContentTransformedProducts` and `automatedLatestContentTransformedPosts` could maybe go to a separate class outside of block renderer

Done, now they're two separate classes.

> I don't think the code $this->renderer->render($newsletter, false, false, $sendingTask); is very readable. I think it would be better to have a method renderAsPreview and render to make stuff more obvious 

Done, we have two methods now, arguments were also reordered.

>  I am not sure about all the meta. I don't have a better solution but such a vague name adds a complexity. When a new developer will come to this code in the future they will need a lot of time to debug what all the stuff does.

I don't think cart data qualifies for a separate field in the model, hence `meta` it is. Have no better idea either.

pavel-mailpoet (dismissed):

veljkho (changes requested):
This is not working properly:
- When I add 1st item to the cart... the email is properly scheduled for sending and with that item in the content. When I then add any additional item to the cart... the email is rescheduled for sending but with just 1st item in the email content. Should be fetched with all cart items.

All other verified.

wxa:
> the email is rescheduled for sending but with just 1st item in the email content

Fixed. It was updating scheduled task's `meta` instead of sending queue's `meta` when rescheduling.

pavel-mailpoet (approved):

veljkho (approved):

PR #3087, opened 2020-10-06
Redesign - Email stats 1/3 - Header [MAILPOET-3102]

pavel-mailpoet:
[MAILPOET-3102]

[MAILPOET-3102]: https://mailpoet.atlassian.net/browse/MAILPOET-3102


NeosinneR:
@pavel-mailpoet just a quick note, could you please use `.mailpoet-` prefix for CSS classes? For consistency with other components and to prevent potential conflict with other plugins (if they are whitelisted) or even WordPress itself in future versions.

wxa:
Issues I found:
1. No mobile logo on small screen;
2. When hovering mouse over MailPoet logo or Beamer logo the cursor does not indicate they're clickable, please add `cursor: pointer` to CSS;
3. After opening the Beamer sidebar (if there's news, i.e. a red dot) on the stats page an orange bell icon appears that obstructs Help Scout widget:
image
4. I don't see screen options being restored after being hidden. Would be nice to use the opposite function when a stats page is unmounted:
https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/newsletters/listings/heading_display.jsx#L6

wxa (changes requested):
See my previous comment.

wxa:
@pavel-mailpoet Now the MailPoet logo looks clickable, but it's not clear what would happen if you click on it. Do you think it makes sense to add a `title` attribute to the logo, something like "Back to section root"?

wxa (approved):
Looks nice now, thanks!

veljkho (approved):

PR #3091, opened 2020-10-07
Better explain import options [MAILPOET-3181]

wxa:
[MAILPOET-3181]

[MAILPOET-3181]: https://mailpoet.atlassian.net/browse/MAILPOET-3181

### DIFF
--- a/views/subscribers/importExport/import.html
+++ b/views/subscribers/importExport/import.html
@@ -113,7 +113,7 @@
 'pickListsDescription': __('Pick the list that you want to import these subscribers to.'),
 'select': _x('Select', ' Verb'),
 'createANewList': __('Create a new list'),
-'updateExistingSubscribers': __("Update existing subscribers' information"),
+'updateExistingSubscribers': __("Update existing subscribers’ information, like first name, last name, etc."),
 'updateExistingSubscribersYes': __('Yes'),
 'updateExistingSubscribersNo': __('No'),
 'addSubscribersToSegment': __(' To add subscribers to a mailing segment, [link]create a list[/link].'),
--- a/views/subscribers/importExport/import.html
+++ b/views/subscribers/importExport/import.html
@@ -113,7 +113,7 @@
 'pickListsDescription': __('Pick the list that you want to import these subscribers to.'),
 'select': _x('Select', ' Verb'),
 'createANewList': __('Create a new list'),
-'updateExistingSubscribers': __("Update existing subscribers’ information, like first name, last name, etc."),
+'updateExistingSubscribers': __('Update existing subscribers’ information, like first name, last name, etc.'),
 'updateExistingSubscribersYes': __('Yes'),
 'updateExistingSubscribersNo': __('No'),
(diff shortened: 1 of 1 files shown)


views/subscribers/importExport/import.html
```diff
@@ -113,7 +113,7 @@ <h1 class="title">
 'pickListsDescription': __('Pick the list that you want to import these subscribers to.'),
 'select': _x('Select', ' Verb'),
 'createANewList': __('Create a new list'),
-'updateExistingSubscribers': __("Update existing subscribers' information"),
+'updateExistingSubscribers': __("Update existing subscribers' information, like first name, last name, etc."),
```
pavel-mailpoet, line 116: @wxa When updating this translation, I think we should also change the regular apostrophe to the typographic one.  We do that for the new translations everywhere. It has many benefits as stated previously. (mainly because it looks better)

Also, why do you keep assigning Jan Lysy? I thought the new developers always start doing code reviews after a couple of weeks. Do I misremember that?

wxa: >  change the regular apostrophe to the typographic one

Fixed. 
> We do that for the new translations everywhere

Could you please point me to a discussion or a document where this was set as a rule? I have only found it mentioned in a few comments to PRs I did not participate in.

> why do you keep assigning Jan Lysy? I thought the new developers always start doing code reviews after a couple of weeks. Do I misremember that?

Well, it's his third week already, so all good 😃 But seriously, I have asked him beforehand if he can review PRs and he said yes. And I did not mean to overwhelm him, the PRs are tiny, more to get him familiar with the process than anything else. Is something wrong with that?

pavel-mailpoet: > Could you please point me to a discussion or a document where this was set as a rule?

That is not a rule, but that has been discussed a couple of times during various pull requests. Also, we discussed it with Quentin where we agreed that devs will do it but that has been a very long time ago and I don't remember if we did it on slack or Jira or maybe even in person.

>  Is something wrong with that?

no, there is nothing wrong with that, I am just surprised, that's it.

views/subscribers/importExport/import.html
```diff
@@ -113,7 +113,7 @@ <h1 class="title">
 'pickListsDescription': __('Pick the list that you want to import these subscribers to.'),
 'select': _x('Select', ' Verb'),
 'createANewList': __('Create a new list'),
-'updateExistingSubscribers': __("Update existing subscribers' information"),
+'updateExistingSubscribers': __("Update existing subscribers’ information, like first name, last name, etc."),
```
lysyjan, line 116:
```suggestion
'updateExistingSubscribers': __('Update existing subscribers’ information, like first name, last name, etc.'),
```
After Pavel's suggestion, you can replace double-quotes with the apostrophes. What do you mean?

wxa: Done.

lysyjan (approved):

veljkho (approved):

PR #3104, opened 2020-10-13
Stop loading 3rd party libraries [MAILPOET-3128]

pavel-mailpoet:
[MAILPOET-3128]

[MAILPOET-3128]: https://mailpoet.atlassian.net/browse/MAILPOET-3128


wxa (approved):

veljkho (changes requested):
After selecting Yes on step 3, on step 5 now I cant select Yes or No:
- https://i.imgur.com/BwPHz1v.png
but able to continue further if click the button.

wxa:
There's also a CSS inclusion from https://use.typekit.net/aps3nxt.css which sometimes generates a network request in console, can it be considered a 3rd party library?

pavel-mailpoet:
@wxa I am aware of it. But it wasn't part of this specification so it can be a part of a different task. Frankly, I am not sure how to block that since it is loaded from a CSS file.

wxa:
@pavel-mailpoet 
> I am not sure how to block that since it is loaded from a CSS file.
One way is to bundle this import in a separate CSS file which is conditionally included on a page. The thing is, our new design uses this font everywhere so it will look very different without it. You're right, let's take care of it later if there's a need.

wxa (approved):

veljkho (approved):

PR #3106, opened 2020-10-16
Permit trashing of WP User subscribers [MAILPOET-3138]

lysyjan:
[MAILPOET-3138]

[MAILPOET-3138]: https://mailpoet.atlassian.net/browse/MAILPOET-3138

### DIFF
--- a/lib/Subscribers/SubscribersRepository.php
+++ b/lib/Subscribers/SubscribersRepository.php
@@ -100,7 +100,8 @@ class SubscribersRepository extends Repository {
       return 0;
     }
 
-    $this->entityManager->transactional(function (EntityManager $entityManager) use ($ids) {
+    $count = 0;
+    $this->entityManager->transactional(function (EntityManager $entityManager) use ($ids, &$count) {
       // Delete subscriber segments
       $this->bulkRemoveFromAllSegments($ids);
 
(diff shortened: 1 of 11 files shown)


lib/Subscribers/SubscribersRepository.php
```diff
@@ -112,13 +113,14 @@ public function bulkDelete(array $ids): int {
       ", ['ids' => $ids], ['ids' => Connection::PARAM_INT_ARRAY]);
 
       $queryBuilder = $entityManager->createQueryBuilder();
-      $queryBuilder->delete(SubscriberEntity::class, 's')
+      $count = $queryBuilder->delete(SubscriberEntity::class, 's')
         ->where('s.id IN (:ids)')
+        ->andWhere('s.wpUserId IS NULL')
```
costasovo, line 118: I think this is something that should be covered with a test. We have [a test for this repository](https://github.com/mailpoet/mailpoet/blob/master/tests/integration/Subscribers/SubscribersRepositoryTest.php). What do you think about extending it with this use-case?

costasovo (commented):
Good job! I would just suggest extending subscribers repository test for new use-cases.

lib/Subscribers/SubscribersRepository.php
lysyjan: Good idea, I'll add two or three tests for this use case.

costasovo (approved):

veljkho (approved):

PR #3108, opened 2020-10-19
Stats summary redesign [MAILPOET-2791]

pavel-mailpoet:
[MAILPOET-2791]

[MAILPOET-2791]: https://mailpoet.atlassian.net/browse/MAILPOET-2791


costasovo (changes requested):
Looks good. I have just two small remarks (in comments) and I think we should improve spacing in the top part of the stats page. 

I took a screenshot https://d.pr/i/4uEAV8
On the screenshot, the design is on top. It seems that the top margin of the newsletter subject heading is bigger and also there is too little space vertically around newsletter segments (displayed with the light blue background).

costasovo (dismissed):

veljkho (changes requested):
I see some issues:
- When I don't have Premium activated, I can't see the [order amount](https://i.imgur.com/B5vGxpj.png) in the stats - is this a valid issue?
- [These titles are not enough bold](https://i.imgur.com/9qKqmvJ.png) on my side, require a higher amount of font-weight. I tested it with 600 and 700, both worked good.
- Can we add some padding between [To: & List Name + at the top](https://i.imgur.com/30dbDK6.png)
- Looks like the [whole content here is moved up a bit](https://i.imgur.com/nbqalT7.png), we need to move down from the top line. In Rosta's screenshot in the above comment, the issue was not present.

pavel-mailpoet:
> When I don't have Premium activated, I can't see the order amount in the stats - is this a valid issue?

That has always been like that. The premium plugin adds the order amount.

costasovo (dismissed):

veljkho (changes requested):
Everything's good, except this one [new issue](https://i.imgur.com/8lRaHRM.png).

costasovo (approved):

veljkho (approved):

PR #3111, opened 2020-10-20
Change purchased product trigger [MAILPOET-2335] [MAILPOET-2334]

pavel-mailpoet:
[MAILPOET-2335] [MAILPOET-2334]

[MAILPOET-2335]: https://mailpoet.atlassian.net/browse/MAILPOET-2335
[MAILPOET-2334]: https://mailpoet.atlassian.net/browse/MAILPOET-2334

### DIFF
--- a/lib/Models/Newsletter.php
+++ b/lib/Models/Newsletter.php
@@ -369,17 +369,39 @@ class Newsletter extends Model {
   }
 
   public function wasScheduledForSubscriber($subscriberId) {
+    $query = SendingQueue::selectExpr('COUNT(*)', 'count');
+    $query = $this->getAllQueuesForSubscscriberQuery($query, $subscriberId);
     /** @var \stdClass */
-    $queue = SendingQueue::rawQuery(
-      "SELECT COUNT(*) as count
-      FROM `" . SendingQueue::$_table . "`
-      JOIN `" . ScheduledTask::$_table . "` ON " . SendingQueue::$_table . ".task_id = " . ScheduledTask::$_table . ".id
-      JOIN `" . ScheduledTaskSubscriber::$_table . "` ON " . ScheduledTask::$_table . ".id = " . ScheduledTaskSubscriber::$_table . ".task_id
-      WHERE " . ScheduledTaskSubscriber::$_table . ".subscriber_id = " . $subscriberId . "
-      AND " . SendingQueue::$_table . ".newsletter_id = " . $this->id
-    )->findOne();
+    $queue = $query->findOne();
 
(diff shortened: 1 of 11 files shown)


lib/Models/Newsletter.php
```diff
@@ -369,19 +369,41 @@ public function withOptions() {
...
-      JOIN `" . ScheduledTaskSubscriber::$_table . "` ON " . ScheduledTask::$_table . ".id = " . ScheduledTaskSubscriber::$_table . ".task_id
-      WHERE " . ScheduledTaskSubscriber::$_table . ".subscriber_id = " . $subscriberId . "
-      AND " . SendingQueue::$_table . ".newsletter_id = " . $this->id
-    )->findOne();
+    $queue = $query->findOne();
 
     return ((int)$queue->count) > 0;
   }
 
+  private function getAllQueuesForSubscscriberQuery($orm, $subscriberId) {
```
costasovo, line 380: Here we add more usages of ORM we want to get rid of. Could you create a Doctrine repository instead? Automatic event class are not yet refactored to use DI container but you can use ContainerWrapper::getInstance()->ger() in the event class constructor as we do in other places that are not refactored yet.

tests/integration/AutomaticEmails/WooCommerce/Events/PurchasedInCategoryTest.php
```diff
@@ -156,6 +160,51 @@ private function getOrderMock($categories = ['123']) {
...
+
+    $subscriberSegment = SubscriberSegment::create();
+    $subscriberSegment->hydrate([
+      'subscriber_id' => $subscriber->id,
+      'segment_id' => Segment::getWooCommerceSegment()->id,
+      'status' => Subscriber::STATUS_SUBSCRIBED,
+    ]);
+    $subscriberSegment->save();
+    return $subscriber;
+  }
```
costasovo, line 206: This could be written using Doctrine as well. I think you need only ids in the test.

costasovo (commented):
Looks good. I would just replace newly added Idiorm queries by Doctrine so that we save ourselves work in the future. What do you think?

tests/integration/AutomaticEmails/WooCommerce/Events/PurchasedInCategoryTest.php
pavel-mailpoet: Well, yes, but I didn't write this code, I only moved it around.

costasovo (approved):

veljkho (approved):

PR #3112, opened 2020-10-20
Update carbon [MAILPOET-3232]

costasovo:
[MAILPOET-3232]

[MAILPOET-3232]: https://mailpoet.atlassian.net/browse/MAILPOET-3232

### DIFF
--- a/prefixer/fix-carbon.php
+++ b/prefixer/fix-carbon.php
@@ -41,3 +41,28 @@ foreach ($replacements as $singleFile) {
   }
   file_put_contents($singleFile['file'], $data);
 }
+
+// Remove all locales except default english
+// We don't use Carbon translate capabilities so we keep only default english locale to reduce size of the library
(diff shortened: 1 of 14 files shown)


prefixer/fix-carbon.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+// throw exception if anything fails
```
pavel-mailpoet, line 3: I think this is a temporary fix, isn't it? Could you please leave a comment on what needs to happen for this file to be removed?

costasovo: Only ternary operator fixes are temporary (until new phpScoper with updated parser is released). Cleanups (added in later commit) are not temporary. I'll add a comment to the ternary operators fix and create a Jira for updating PHPScoper.

pavel-mailpoet (approved):

veljkho (approved):

PR #3114, opened 2020-10-21
Update symfony/dependency-injection [MAILPOET-3239]

costasovo:
[MAILPOET-3239]

[MAILPOET-3239]: https://mailpoet.atlassian.net/browse/MAILPOET-3239

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -55,7 +55,7 @@ mkdir vendor-prefixed
 ./tools/vendor/composer.phar install --no-dev --prefer-dist --optimize-autoloader --no-scripts
 
 echo '[BUILD] Fetching prefixed production libraries'
-./tools/vendor/composer.phar install --prefer-dist --working-dir=./prefixer/
+./tools/vendor/composer.phar install --no-dev --prefer-dist --working-dir=./prefixer/
 
 # Remove Doctrinne Annotations (no need since generated metadata are packed)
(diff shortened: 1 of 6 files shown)


build.sh
```diff
@@ -55,7 +55,7 @@ mkdir vendor-prefixed
 ./tools/vendor/composer.phar install --no-dev --prefer-dist --optimize-autoloader --no-scripts
 
 echo '[BUILD] Fetching prefixed production libraries'
-./tools/vendor/composer.phar install --prefer-dist --working-dir=./prefixer/
+./tools/vendor/composer.phar install --no-dev --prefer-dist --working-dir=./prefixer/
```
pavel-mailpoet, line 58: Could you explain this change to me? This is different than what we do in the ci and locally, right? So we will release a different version than we test. Could we break the production build by this?

costasovo: With this flag composer skips required-dev, dependencies. We do the same for non-prefixed dependencies. Take a look  a couple of lines above. We don't have any dev dependencies in prefixed composer.json so actually, this change will not affect anything. It only allows us adding prefixed dev dependencies, that won't be added to build.
I added this change because I needed to add dev dependency symfony/config, but I removed it later and used a workaround. I think we can keep this change it makes no harm.

pavel-mailpoet (approved):

veljkho (approved):

PR #3126, opened 2020-10-23
Fix flaky scenario addHeader [MAILPOET-3216]

veljkho:
[MAILPOET-3216](https://mailpoet.atlassian.net/browse/MAILPOET-3216)

### DIFF
--- a/tests/acceptance/EditorHeaderBlockCest.php
+++ b/tests/acceptance/EditorHeaderBlockCest.php
@@ -18,6 +18,7 @@ class EditorHeaderBlockCest {
     $i->dragAndDrop('#automation_editor_block_header', '#mce_0');
     //Open settings by clicking on block
     $i->moveMouseOver($headerInEditor, 3, 2);
+    $i->waitForElement($headerInEditor);
     $i->waitForText('View this in your browser.');
     $i->click($headerSettingsIcon);
     $i->waitForElementVisible($headerSettingsAssertion);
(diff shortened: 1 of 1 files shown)


tests/acceptance/EditorHeaderBlockCest.php
```diff
@@ -18,6 +18,7 @@ public function addHeader(\AcceptanceTester $i) {
     $i->dragAndDrop('#automation_editor_block_header', '#mce_0');
     //Open settings by clicking on block
     $i->moveMouseOver($headerInEditor, 3, 2);
+    $i->waitForElement($headerInEditor);
```
wxa, line 21: Are you sure you want to use `waitForElement`? If you want to wait for a fixed amount of time, `wait` is better suited because `waitForElement` will only wait if there is no `headerInEditor` element on the page, but it's already there, you just interacted with it a line above. So it may have no effect.

veljkho: I used `waitForElement` because it looks like it's enough to fix the flakiness and I never use `wait`. But, in case `wait(1)` would be a better solution, I agree to implement it. But just, in this case, looks like `waitForElement` is just enough.

wxa: Okay, if you think this will be enough to combat the flakiness, let's go for it. But keep in mind the alternative way in case it won't fly. I've just added a comment to clarify that waiting for the element after it's already there is intentional.

wxa (approved):

PR #3127, opened 2020-10-23
Fix flaky scenario addSpacer [MAILPOET-3250]

veljkho:
[MAILPOET-3250](https://mailpoet.atlassian.net/browse/MAILPOET-3250)


costasovo (commented):
Can you please explain what was flaky and how these changes help? There isn't a link to the failing build in JIRA. It is a bit confusing. Why we have to hover on footer block?

veljkho:
> Can you please explain what was flaky and how these changes help? There isn't a link to the failing build in JIRA. It is a bit confusing. Why we have to hover on footer block?

The problem was that the automation had problems to see spacerResizeHandle element when hovering over the spacer block in the editor (after adding it). So, with my solution to just hover the footer element and then hover the spacer, automation will then always be able to find the spacerResizeHandle element successfully.

costasovo (approved):

PR #3134, opened 2020-10-27
Add a description for a form of type Bellow pages [MAILPOET-3159]

lysyjan:
[MAILPOET-3159]

[MAILPOET-3159]: https://mailpoet.atlassian.net/browse/MAILPOET-3159

### DIFF
--- a/assets/js/src/form_editor/components/form_settings/form_placement_options/settings_panels/below_posts_settings.tsx
+++ b/assets/js/src/form_editor/components/form_settings/form_placement_options/settings_panels/below_posts_settings.tsx
@@ -17,6 +17,7 @@ const BelowPostsSettings = () => {
 
   return (
     <>
+      <p>{MailPoet.I18n.t('placeFormBellowPagesDescription')}</p>
       <ToggleControl
         label={MailPoet.I18n.t('enable')}
         checked={isActive}


views/form/editor.html
```diff
@@ -93,6 +93,7 @@
   'exitIntentDescription': __('Show the form immediately if the visitor attempts to leave the site.'),
   'exitIntentSwitch': __('Exit-intent display'),
   'placeFormBellowPages': _x('Below pages', 'This is a text on a widget that leads to settings for form placement'),
+  'placeFormBellowPagesPagesDescription': __('Display your form at the bottom of posts or pages.'),
   'placeFormBellowPagesDescription': __('This form placement allows you to add this form at the end of all the pages or posts, below the content.'),
```
costasovo, line 97: I can see that we already have a translation for bellow pages description. Just take a look one line below. We probably accidentally removed it from the react component during refactoring.
I guess we could reuse the old translation key, which seems a bit nicer than the new one with double "pages". Maybe we could also reuse the older wording as is. The older text has been probably already translated. You can ask @veljkho about it.

lysyjan: You are right. I'm feeling really blind at this moment. I'm sorry.

costasovo: We have about a hundred keys there so it is easy to miss one and you didn't know we used to have some description. So you don't need to be sorry.

lysyjan: We decided to use the old description.

costasovo (approved):

veljkho (approved):

PR #3135, opened 2020-10-27
Remove default form [MAILPOET-3238]

costasovo:
[MAILPOET-3238]

[MAILPOET-3238]: https://mailpoet.atlassian.net/browse/MAILPOET-3238


assets/js/src/forms/heading.jsx
```diff
@@ -5,35 +5,33 @@ import MailPoet from 'mailpoet';
...
-        >
-          {plusIcon}
-          <span>{MailPoet.I18n.t('new')}</span>
-        </Link>
-      </TopBarWithBeamer>
-      <h1 className="mailpoet-newsletter-listing-heading-empty title">{' '}</h1>
-    </>
-  );
+export const goToSelectTemplate = () => {
+  setTimeout(() => {
```
pavel-mailpoet, line 9: Do we need the timeout here? I don't see any events here mentioned in the comment, or should we add an event?

costasovo: Hmm, good question. I just moved the code. I can't see any event even before the change. I'll try to find out what happened to it.

pavel-mailpoet: I **think** we used to have a Mixpanel event for tracking a form creation. That may have happened here.

costasovo: Hmm, the event is gone :( I'll add it to template selection function.

pavel-mailpoet (dismissed):

veljkho (approved):

PR #3140, opened 2020-10-28
Add create blank form option to form template selection [MAILPOET-3207]

costasovo:
[MAILPOET-3207]

[MAILPOET-3207]: https://mailpoet.atlassian.net/browse/MAILPOET-3207

### DIFF
--- a/assets/css/src/components-form-editor/_template-selection.scss
+++ b/assets/css/src/components-form-editor/_template-selection.scss
@@ -34,7 +34,7 @@ $templates-one-column-breakpoint: 2 * ($mailpoet-form-template-thumbnail-width +
 /**
  The header uses grid to position heading in center (second column) and a new form button on right (third column)
  */
-.template-selection-header {
+.mailpoet-template-selection-header {
   @include formTemplatesGrid;
   background: $color-input-background;
   border-bottom: 1px solid $color-tertiary-light;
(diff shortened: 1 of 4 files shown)


assets/css/src/components-form-editor/_template-selection.scss
```diff
@@ -27,11 +27,19 @@ $mailpoet-form-template-thumbnail-height: 316px;
 }
 
 .template-selection-header {
-  align-items: center;
   background: $color-input-background;
   border-bottom: 1px solid $color-tertiary-light;
-  display: flex;
-  flex-wrap: wrap;
-  justify-content: center;
+  display: grid;
```
pavel-mailpoet, line 32: could you please explain why did you change this to a grid? I can see only a single line in there...

costasovo: I added the button so there are 2 items in the top bar. I needed heading to be positioned in centre and button on right. I found out that it is quite tricky to do with flexbox and I found using the grid as the most elegant solution.

pavel-mailpoet: I reckon that grid "trick" deserves a comment in the code, what do you think?

pavel-mailpoet (dismissed):

veljkho (changes requested):
Is it possible to [align the button](https://i.imgur.com/XPalwKo.png) with the template box grid width? Please try, since like this it's too on the right side.

NeosinneR (approved):

veljkho (approved):

PR #3141, opened 2020-10-28
Add cart contents block to email templates [MAILPOET-3182]

pavel-mailpoet:
[MAILPOET-3182]

[MAILPOET-3182]: https://mailpoet.atlassian.net/browse/MAILPOET-3182


lib/Config/PopulatorData/Templates/Fitness.php
```diff
@@ -25,63 +25,66 @@ public function get() {
   }
 
   private function getThumbnail() {
-    return $this->template_image_url . '/thumbnail.20190411-1500.jpg';
+    return $this->template_image_url . '/fitness-thumbnail.20201028.jpg';
```
wxa, line 28: We already have the template name as a folder in `$this->template_image_url`, so I think adding a name to the thumbnail file is redundant.

wxa (changes requested):
Please rename the templates according to specs 1b, 2b.

wxa (approved):
Both templates had the name "Abandoned Cart - Kids", but I fixed that. Approved.

veljkho (approved):

PR #3142, opened 2020-10-29
Improvement useability on small devices [MAILPOET-2794]

lysyjan:
[MAILPOET-2794]

[MAILPOET-2794]: https://mailpoet.atlassian.net/browse/MAILPOET-2794

### DIFF
--- a/assets/js/src/form_editor/components/editor.jsx
+++ b/assets/js/src/form_editor/components/editor.jsx
@@ -107,9 +107,11 @@ export default () => {
                 settings={editorSettings}
                 useSubRegistry={false}
               >
-                <div className="interface-interface-skeleton__left-sidebar">
-                  {isInserterOpen && (<Inserter setIsInserterOpened={setIsInserterOpen} />)}
-                </div>
+                {(isInserterOpen) && (
+                  <div className="interface-interface-skeleton__left-sidebar">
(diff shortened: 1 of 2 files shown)


assets/js/src/form_editor/components/editor.jsx
```diff
@@ -107,9 +109,11 @@ export default () => {
                 settings={editorSettings}
                 useSubRegistry={false}
               >
-                <div className="interface-interface-skeleton__left-sidebar">
-                  {isInserterOpen && (<Inserter setIsInserterOpened={setIsInserterOpen} />)}
-                </div>
+                {(!isMobileViewport || isInserterOpen) && (
```
wxa, line 112: It seems like the `!isMobileViewport` condition does not have any effect. If I understand properly, it should affect a default state: closed on mobile, open on desktop. Yet it doesn't work like that in my tests (maybe I'm not reproducing it correctly, e.g. with a clean install). It behaves as if this condition wasn't there, only the sidebar status determines if it will be open or closed on both screen types. But it's not very important since you can close the sidebar anyway. Can we remove this condition then?

lysyjan: You are right. I have changed it.

wxa (approved):

veljkho (approved):

PR #3143, opened 2020-10-29
Fix filtering subscribers without a segment [MAILPOET-3197]

lysyjan:
[MAILPOET-3197]

[MAILPOET-3197]: https://mailpoet.atlassian.net/browse/MAILPOET-3197

### DIFF
--- a/lib/Subscribers/SubscriberListingRepository.php
+++ b/lib/Subscribers/SubscriberListingRepository.php
@@ -16,6 +16,8 @@ use MailPoetVendor\Doctrine\ORM\Query\Expr\Join;
 use MailPoetVendor\Doctrine\ORM\QueryBuilder;
 
 class SubscriberListingRepository extends ListingRepository {
+  public const FILTER_WITHOUT_LIST = 'without-list';
+
   const DEFAULT_SORT_BY = 'createdAt';
 
   private static $supportedStatuses = [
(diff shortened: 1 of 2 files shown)


lib/Subscribers/SubscriberListingRepository.php
```diff
@@ -88,6 +90,11 @@ protected function applyListingDefinition(QueryBuilder $queryBuilder, ListingDef
     if (!$filters || !isset($filters['segment'])) {
       return;
     }
+    if ($filters['segment'] === self::FILTER_WITHOUT_LIST) {
+      $queryBuilder->leftJoin('s.subscriberSegments', 'ss')
+        ->andWhere('ss.segment IS NULL');
```
costasovo, line 95: I found [the original implementation in the old model](https://github.com/mailpoet/mailpoet/blob/master/lib/Models/Subscriber.php#L542-L553) and it seems that subscribers without list should also contain subscribers who unsubscribed from their only list or subscribers whose only list is in the trash.

We still show subscribers who were added to a list and then unsubscribed from it as unsubscribed in listing for a list when the user filter a specific list.
So maybe we could exclude those who are on some list as unsubscribed from `without-list` so that it is more consistent count them as they have some list. Maybe discuss it with @NeosinneR.

If we do some changes remember also [to alter query for fetching counts into the filter data](https://github.com/mailpoet/mailpoet/blob/master/lib/Subscribers/SubscriberListingRepository.php#L190-L197).

costasovo (changes requested):

costasovo (approved):

veljkho (approved):

PR #3145, opened 2020-11-02
Improve success message design [MAILPOET-3195]

lysyjan:
[MAILPOET-3195]

[MAILPOET-3195]: https://mailpoet.atlassian.net/browse/MAILPOET-3195

### DIFF
--- a/assets/js/src/public.jsx
+++ b/assets/js/src/public.jsx
@@ -81,6 +81,19 @@ jQuery(($) => {
     }
   }
 
+  function displaySuccessMessage(form) {
+    // hide all form elements instead of .mailpoet_message
+    form.children().not('.mailpoet_message').css('visibility', 'hidden');
+    // add class that form was successfully send
+    form.toggleClass('mailpoet_form_successfully_send');
+    // display success message
+    form.find('.mailpoet_validate_success').show();
+    // hide elements marked with a class
+    form.find('.mailpoet_form_hide_on_success').each(function hideOnSuccess() {
+      $(this).hide();
+    });
+  }
+
   function showForm(formDiv, showOverlay = false) {
     const form = formDiv.find('form');
     let delay = form.data('delay');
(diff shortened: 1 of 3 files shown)


assets/js/src/public.jsx
```diff
@@ -99,6 +99,19 @@ jQuery(($) => {
     });
   }
 
+  function hideSucessMessage(form) {
+    // show success message
```
costasovo, line 103: Shouldn't the comment be "hide success message" ?

assets/js/src/public.jsx
```diff
@@ -81,6 +81,24 @@ jQuery(($) => {
     }
   }
 
+  function displaySuccessMessage(form) {
+    // hide all form elements
+    form.children().each((index, element) => {
```
costasovo, line 86: You can use jquery's [.not selector](https://api.jquery.com/not/) to exclude '.mailpoet_message' and get rid of the forEach. I think it would be a bit nicer to read.

costasovo (changes requested):
Looks Good 👍 I have just 2 small remarks.

costasovo (dismissed):

veljkho (changes requested):
I just can't subscribe. Getting error:
`Something went wrong with your subscription. Please contact the website owner.`
The user is present in the pool with the status "Unconfirmed". Didn't receive a confirmation email.

UPDATE: this was due to banned key license. Testing is in progress.

veljkho (changes requested):
There is some issue:
- X button is not clickable when on the success message screen, this is happening only for the Popups and for some templates like Halloween and Black Friday. An interesting fact that it is not present on the mobile...

pavel-mailpoet (dismissed):

veljkho (changes requested):
We should have some padding for the general error message:
- https://i.imgur.com/YtFtoWM.png

pavel-mailpoet (approved):

veljkho (approved):

PR #3147, opened 2020-11-04
Enable full screen mode [MAILPOET-3205]

lysyjan:
[MAILPOET-3205]

[MAILPOET-3205]: https://mailpoet.atlassian.net/browse/MAILPOET-3205

### DIFF
--- a/assets/js/src/form_editor/components/editor.jsx
+++ b/assets/js/src/form_editor/components/editor.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import '@wordpress/core-data';
 import { select, useSelect, useDispatch } from '@wordpress/data';
 import {
@@ -27,6 +27,7 @@ import FormStyles from './form_styles.jsx';
 import Preview from './preview/preview';
 import FormStylingBackground from './form_styling_background.jsx';
 import { CustomFontsStyleSheetLink } from './font_family_settings';
+import Fullscreen from './fullscreen';
 
 /**
  * This component renders the form editor app.
@@ -41,10 +42,6 @@ export default () => {
   const [isInserterOpen, setIsInserterOpen] = useState(false);
   const sidebarOpened = useSelect(
     (sel) => sel('mailpoet-form-editor').getSidebarOpened(),
-    []
-  );
-  const isFullscreen = useSelect(
-    (sel) => sel('mailpoet-form-editor').isFullscreenEnabled(),
     []
   );
   const canUserUpload = useSelect(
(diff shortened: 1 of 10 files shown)


assets/js/src/form_editor/components/header.jsx
```diff
@@ -1,13 +1,20 @@
+/* eslint-disable no-undef */
```
costasovo, line 1: Why do we need to add this? Can we get rid of it?

assets/js/src/form_editor/components/header.jsx
```diff
@@ -53,6 +61,22 @@ const Header = ({ isInserterOpened, setIsInserterOpened }) => {
           isPressed={sidebarOpened}
           className="mailpoet-editor-header-button"
         />
+        <DropdownMenu
+          icon={moreVertical}
+          label="More tools & options"
```
costasovo, line 74: This should be translated. In case this text is taken from WP post editor you can use `__()`  from @wordpress/i18n since we [load translations for post editor that are distributed in WP](https://github.com/mailpoet/mailpoet/blob/master/lib/AdminPages/Pages/FormEditor.php#L301-L338).

assets/js/src/form_editor/components/header.jsx
```diff
@@ -53,6 +61,22 @@ const Header = ({ isInserterOpened, setIsInserterOpened }) => {
...
+        >
+          {() => (
+            <MenuGroup label="View" className="mailpoet-dropdown-menu-group">
+              <FeatureToggle
+                shortcut="Ctrl+Shift+Alt+F"
+                label="Fullscreen mode"
+                info="Work without distraction"
```
costasovo, line 81: Translations...

assets/js/src/form_editor/components/editor.jsx
```diff
@@ -43,6 +43,10 @@ export default () => {
     (sel) => sel('mailpoet-form-editor').getSidebarOpened(),
     []
   );
+  const isFullscreen = useSelect(
```
costasovo, line 46: Now I see that this is used only for changing body class. So I think we could create a component FullscreenMode that would render null but it would contain this selector and effect. We could then just place the component <FullscreenMode /> in editor component.  Editor component is already quite big so this could make it a bit more readable. What do you think?

costasovo (changes requested):
Good job! I added just a couple of notes.

assets/js/src/form_editor/components/header.jsx
lysyjan: 🤦🏻 I definitely forgot on translation. I'm sorry.

assets/js/src/form_editor/components/editor.jsx
lysyjan: Right, it will be better.

assets/js/src/form_editor/components/header.jsx
lysyjan: It added probably editor automatically. I'll get rid.

lysyjan:
@costasovo It should be better now.

costasovo (approved):

veljkho (approved):

PR #3148, opened 2020-11-05
Update scenario to cover form new button [MAILPOET-3272]

veljkho:
[MAILPOET-3272](https://mailpoet.atlassian.net/browse/MAILPOET-3272)

### DIFF
--- a/tests/acceptance/FormsListingCest.php
+++ b/tests/acceptance/FormsListingCest.php
@@ -8,7 +8,6 @@ class FormsListingCest {
   public function formsListing(\AcceptanceTester $i) {
     $i->wantTo('Open forms listings page');
     $formName = 'Test Form';
-    $newFormButton = '.mailpoet-button';
     $form = new Form();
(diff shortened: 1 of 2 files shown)


tests/acceptance/FormsListingCest.php
```diff
@@ -6,16 +6,20 @@
 
 class FormsListingCest {
   public function formsListing(\AcceptanceTester $i) {
+    $i->wantTo('Open forms listings page');
+    $formName = 'Test Form';
+    $newFormButton = '.mailpoet-button';
```
wxa, line 11: This selector is too broad, if more buttons appear on the page, the test will be broken. One of the buttons already has `data-automation-id="create_new_form"`, we can add the same id to another button, or, to not mix them up, use the `create_new_form_2` ID and select both with `[data-automation-id^="create_new_form"]` (`^=` means "begins with").

wxa (approved):

PR #3165, opened 2020-11-12
Improve WC customizer [MAILPOET-2584]

wxa:
Item D (KB article) is already done in https://mailpoet.atlassian.net/browse/MAILPOET-2561


lysyjan (approved):

veljkho (changes requested):
The point _B-Dark blue border is missing_ is [still present ](https://i.imgur.com/aNTpPd7.png)as an issue.

veljkho (approved):

PR #3171, opened 2020-11-16
Remove disabled rules in eslint typescript config [MAILPOET-3201]

pavel-mailpoet:
[MAILPOET-3201]

[MAILPOET-3201]: https://mailpoet.atlassian.net/browse/MAILPOET-3201


lysyjan (approved):
I think that it looks very good.

veljkho (approved):

PR #3178, opened 2020-11-17
Add undo&redo to the Form editor [MAILPOET-3085]

lysyjan:
[MAILPOET-3085]

[MAILPOET-3085]: https://mailpoet.atlassian.net/browse/MAILPOET-3085

### DIFF
--- a/assets/js/src/form_editor/components/editor.jsx
+++ b/assets/js/src/form_editor/components/editor.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import '@wordpress/core-data';
-import { select, useSelect, useDispatch } from '@wordpress/data';
+import { useSelect, useDispatch } from '@wordpress/data';
 import {
   DropZoneProvider,
   Popover,
@@ -44,6 +44,11 @@ export default () => {
     (sel) => sel('mailpoet-form-editor').getSidebarOpened(),
     []
   );
+  const formBlocks = useSelect(
+    (sel) => sel('mailpoet-form-editor').getFormBlocks(),
+    []
+  );
+
   const canUserUpload = useSelect(
     (sel) => sel('core').canUser('create', 'media'),
     []
@@ -102,7 +107,7 @@ export default () => {
             </div>
             <div className="interface-interface-skeleton__body">
               <BlockEditorProvider
-                value={select('mailpoet-form-editor').getFormBlocks()}
+                value={formBlocks}
                 onInput={blocksChangedInBlockEditor}
                 onChange={blocksChangedInBlockEditor}
                 settings={editorSettings}
--- a/assets/js/src/form_editor/components/form_settings/selection.jsx
+++ b/assets/js/src/form_editor/components/form_settings/selection.jsx
@@ -26,6 +26,9 @@ class Selection extends React.Component {
         .trigger('change');
     }
 
+    // After change data in store this component didn't call trigger for change
+    // It happened only when component allowed multipleValues
+    // Following lines are modified lines on top for multipeValues
     if (
       (this.props.item !== undefined && prevProps.item !== undefined)
       && this.allowMultipleValues()
--- a/assets/js/src/form_editor/components/form_settings/selection.jsx
+++ b/assets/js/src/form_editor/components/form_settings/selection.jsx
@@ -26,6 +26,17 @@ class Selection extends React.Component {
         .trigger('change');
     }
 
+    if (
+      (this.props.item !== undefined && prevProps.item !== undefined)
+      && this.allowMultipleValues()
+      && _.isArray(this.props.item[this.props.field.name])
+      && !_.isEqual(this.props.item[this.props.field.name], prevProps.item[this.props.field.name])
+    ) {
+      jQuery(`#${this.selectRef.current.id}`)
+        .val(this.getSelectedValues())
+        .trigger('change');
+    }
+
     if (this.isSelect2Initialized()
       && (this.getFieldId(this.props) !== this.getFieldId(prevProps))
       && this.props.field.resetSelect2OnUpdate !== undefined
--- a/assets/js/src/form_editor/components/history_redo.tsx
+++ b/assets/js/src/form_editor/components/history_redo.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import { __ } from '@wordpress/i18n';
+import { Button } from '@wordpress/components';
+import { useSelect, useDispatch } from '@wordpress/data';
+import { redo as redoIcon } from '@wordpress/icons';
+
+function HistoryRedo(props) {
+  const hasRedo = useSelect(
+    (select) => select('mailpoet-form-editor').hasEditorRedo(),
+    []
+  );
+  const { historyMove } = useDispatch('mailpoet-form-editor');
+  const redoAction = () => {
+    historyMove('redo');
+  };
+  return (
+    <Button
+      {...props}
+      icon={redoIcon}
+      label={__('Redo')}
+      aria-disabled={!hasRedo}
+      onClick={hasRedo ? redoAction : undefined}
+      className="editor-history__redo"
+    />
+  );
+}
+
+export default HistoryRedo;
--- a/assets/js/src/form_editor/components/history_redo.tsx
+++ b/assets/js/src/form_editor/components/history_redo.tsx
@@ -3,6 +3,8 @@ import { __ } from '@wordpress/i18n';
 import { Button } from '@wordpress/components';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { redo as redoIcon } from '@wordpress/icons';
+import { displayShortcut } from '@wordpress/keycodes';
+import { useShortcut } from '@wordpress/keyboard-shortcuts';
 
 function HistoryRedo(props) {
   const hasRedo = useSelect(
@@ -10,14 +12,38 @@ function HistoryRedo(props) {
     []
   );
   const { historyRedo } = useDispatch('mailpoet-form-editor');
+  const { registerShortcut } = useDispatch('core/keyboard-shortcuts');
+
   const redoAction = () => {
     historyRedo();
   };
+
+  useShortcut(
+    // Shortcut name
+    'mailpoet-form-editor/redo',
+    // Shortcut callback
+    (event) => {
+      redoAction();
+      event.preventDefault();
+    },
+  );
+
+  registerShortcut({
+    name: 'mailpoet-form-editor/redo',
+    category: 'block',
+    description: __('Redo your last undo.'),
+    keyCombination: {
+      modifier: 'primaryShift',
+      character: 'z',
+    },
+  });
+
   return (
     <Button
       {...props}
       icon={redoIcon}
       label={__('Redo')}
+      shortcut={displayShortcut.primaryShift('z')}
       aria-disabled={!hasRedo}
       onClick={hasRedo ? redoAction : undefined}
       className="editor-history__redo"
--- a/assets/js/src/form_editor/components/history_redo.tsx
+++ b/assets/js/src/form_editor/components/history_redo.tsx
@@ -9,9 +9,9 @@ function HistoryRedo(props) {
     (select) => select('mailpoet-form-editor').hasEditorRedo(),
     []
   );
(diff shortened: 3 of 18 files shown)


lysyjan:
Hey, please look at this PR. I'm open to all discussions.
I had troubles with [Selection.jsx](https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/form_editor/components/form_settings/selection.jsx) and I didn't find another way.
Some issues were with displaying when I used undo&redo.

assets/js/src/form_editor/store/actions.ts
```diff
@@ -33,15 +33,21 @@ export function* toggleFullscreen(toggleTo) {
   };
 }
 
-export function changeFormBlocks(blocks) {
-  return {
+export function* changeFormBlocks(blocks) {
+  yield {
+    type: 'CREATE_HISTORY_RECORD',
```
pavel-mailpoet, line 38: Why do we need a new action? This is called when other actions are also triggered. I don't see why is this necessary. We can have this in the reducer, I think that would be nicer and keep the actions cleaner

assets/js/src/form_editor/store/actions.ts
```diff
@@ -295,3 +307,23 @@ export function* applyStylesToAllTextInputs(styles) {
     styles,
   };
 }
+
+export function createHistoryRecord() {
```
pavel-mailpoet, line 311: I don't see this being used, is it?

assets/js/src/form_editor/store/actions.ts
```diff
@@ -295,3 +307,23 @@ export function* applyStylesToAllTextInputs(styles) {
...
+  return {
+    type: 'CREATE_HISTORY_RECORD',
+  };
+}
+
+export function* historyMove(action) {
+  const editorHistoryOffset = select('mailpoet-form-editor').getEditorHistoryOffset();
+  if (editorHistoryOffset === 0) {
```
pavel-mailpoet, line 319: I am not sure I understand. Why do we want to create a new action here? I mean we are at the beginning of the history, right? So I think nothing should happen when the user goes back. So I don't understand why is here the `CREATE_HISTORY_RECORD` triggered.

assets/js/src/form_editor/components/history_redo.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { redo as redoIcon } from '@wordpress/icons';
+
+function HistoryRedo(props) {
+  const hasRedo = useSelect(
+    (select) => select('mailpoet-form-editor').hasEditorRedo(),
+    []
+  );
+  const { historyMove } = useDispatch('mailpoet-form-editor');
+  const redoAction = () => {
+    historyMove('redo');
```
pavel-mailpoet, line 14: You can convince me otherwise, but I think it is nicer to have two different functions for undo and redo rather than using a string argument. We wouldn't have to check the value of the argument and it reads a bit better I think.

assets/js/src/form_editor/store/selectors.jsx
```diff
@@ -187,4 +187,18 @@ export default {
       return result;
     }, null);
   },
+  getEditorHistoryOffset(state) {
+    return state.editorHistoryOffset;
+  },
+  hasEditorUndo(state) {
```
pavel-mailpoet, line 193: Do you think we should write a test for this function?

assets/js/src/form_editor/store/selectors.jsx
```diff
@@ -187,4 +187,18 @@ export default {
...
+    let offset = state.editorHistoryOffset;
+    if (state.editorHistory.length > 1) {
```
pavel-mailpoet, line 195: Do you think this would deserve a comment? It took me awhile to understand why do we need to add 1.

assets/js/src/form_editor/components/editor.jsx
```diff
@@ -44,6 +44,11 @@ export default () => {
     (sel) => sel('mailpoet-form-editor').getSidebarOpened(),
     []
   );
+  const formBlocks = useSelect(
+    (sel) => sel('mailpoet-form-editor').getFormBlocks(),
```
pavel-mailpoet, line 48: Nothing wrong here, I just wonder. Why did you have to make this change? What is wrong with `select('mailpoet-form-editor').getFormBlocks()`?

assets/js/src/form_editor/components/form_settings/selection.jsx
```diff
@@ -26,6 +26,17 @@ class Selection extends React.Component {
         .trigger('change');
     }
 
+    if (
```
pavel-mailpoet, line 29: I think we need a comment or something to understand what is going on. The commit message is not very helpful here and if we come back in two years we will wonder why is this code here.

assets/js/src/form_editor/store/reducers/history_record.ts
```diff
@@ -0,0 +1,83 @@
+import { isEqual } from 'lodash';
+
+export const createHistoryRecord = (state) => {
+  let editorHistory: any[] = state.editorHistory;
```
pavel-mailpoet, line 4: It is great that you are using types here. But please don't use `any`, in my branch I updated TS to disallow this. The build would fail.
You know the structure, right? So it should be easier to create types for this.

assets/js/src/form_editor/store/reducers/history_record.ts
```diff
@@ -0,0 +1,83 @@
+import { isEqual } from 'lodash';
+
+export const createHistoryRecord = (state) => {
```
pavel-mailpoet, line 3: I think this whole file deserves a test, what do you think?

assets/js/src/form_editor/store/reducers/history_record.ts
```diff
@@ -0,0 +1,83 @@
...
+  let editorHistory: any[] = state.editorHistory;
+  let editorHistoryOffset = state.editorHistoryOffset;
+  const lastHistoryRecord = editorHistory[editorHistory.length - 1];
+
+  let newHistoryRecord;
+  const time = Date.now();
+
+  if (
+    lastHistoryRecord !== undefined
+    && (time - lastHistoryRecord.time) < 1000
```
pavel-mailpoet, line 13: Would it be better if we used a constants instead of using a magical numbers 1000 or 100 here?

assets/js/src/form_editor/store/reducers/history_record.ts
```diff
@@ -0,0 +1,83 @@
...
+    && editorHistory.length > 1
+  ) {
```
pavel-mailpoet, line 15: I think this function need either split into a couple of smaller ones and/or some comments why do we need all of the logic here.

pavel-mailpoet (commented):
Hey, good job doing this. It looks great. I just have a couple of nitpicks. I am open to discussion about all of them. I think we need a couple of tests though, unit tests and maybe even an acceptance test for the new functionality.

assets/js/src/form_editor/store/actions.ts
lysyjan: This is an excellent question. I will try to explain it by example. When you make one change then you will have one history record. Right? And when you return back then you will have restored the same record that you have in history. And my question is. What do you restore when you will want to move forward? Without creating a record when you move from zero you haven't the last state. Do you agree?

assets/js/src/form_editor/store/reducers/history_record.ts
lysyjan: Yes, you are right. I'm not used to write constants in TS. I'll be better for next time.

assets/js/src/form_editor/store/actions.ts
pavel-mailpoet: I am not sure I follow your logic, but I reckon that still needs an explanation in the code then:

> When you make one change then you will have one history record. Right?
right, I follow that logic
> And when you return back then you will have restored the same record that you have in history
I think I follow, but you mean do undo and redo, correct?
>  What do you restore when you will want to move forward?
I don't know, why do you need to restore anything, you are at the end of the queue, you are not? Why then restore something if you are at the end.

lysyjan: > When you make one change then you will have one history record. Right?
You make some change in form. For example, move up with a button.
> And when you return back then you will have restored the same record that you have in history
I mean **undo** 🔙.
> What do you restore when you will want to move forward?
When you would have only one record in history, and after **undo** you want to do **redo**, you will have nothing to restore. This is the reason why I create new history record before **undo**.

assets/js/src/form_editor/components/history_redo.tsx
lysyjan: The functions would be almost the same instead of two lines. I don't guess that it's the right way.
Maybe something like a private function and two public functions?

pavel-mailpoet: Yeah, one private with two public functions sound like a great plan.

assets/js/src/form_editor/components/editor.jsx
lysyjan: Form Block didn't refresh when I clicked on undo without useSelect.

assets/js/src/form_editor/store/actions.ts
lysyjan: Right, it's redundant.

assets/js/src/form_editor/store/reducers/history_record.ts
lysyjan: Right, it's a nice place for type.

assets/js/src/form_editor/store/reducers/history_record.ts
lysyjan: 👍🏻 Absolutely yes.

assets/js/src/form_editor/store/actions.ts
lysyjan: It isn't necessary. I preferred this solution. I think that this functionality will be in reducers a little bit hidden. But if you want to move it I have no problem with it.

costasovo: I think putting it in reducer would be nicer. We have one big reducer for all actions, but we can create another reducer only for undo/redo it would listen to all actions that need creating history. We can use [combineReducers](https://developer.wordpress.org/block-editor/packages/packages-data/#combineReducers) to combine the reducer we have now and this undo-redo. I think this is the recommended approach.

tests/acceptance/FormEditorUndoAndRedoCest.php
```diff
@@ -0,0 +1,42 @@
...
+    $formName = 'My fancy form';
+    $form = new Form();
+    $form->withName($formName)->withSegments([$segment])->create();
+    $i->wantTo('Add first and last name to the editor');
+    $i->login();
+    $i->amOnMailPoetPage('Forms');
+    $i->waitForText($formName);
+    $i->clickItemRowActionByItemName($formName, 'Edit');
+    $i->waitForElement('[data-automation-id="form_title_input"]');
+    // Add First name block
```
pavel-mailpoet, line 22: Instead of using comments in the Acceptance tests, you should use `$i->wantTo()`. It has the same meaning and can be displayed in the output if using verbose output.

pavel-mailpoet (dismissed):

veljkho (changes requested):
Everything's good except keyboard shortcuts. It just won't work with the keys.

assets/js/src/form_editor/components/history_redo.tsx
```diff
@@ -3,21 +3,47 @@ import { __ } from '@wordpress/i18n';
...
+    (event) => {
+      redoAction();
+      event.preventDefault();
+    },
+  );
+
+  registerShortcut({
+    name: 'mailpoet-form-editor/redo',
+    category: 'block',
+    description: __('Redo your last undo.'),
```
pavel-mailpoet, line 34: This is using the gutenberg default translation function. Is that intentional?

lysyjan: Yes, both translations are from Gutenberg.

pavel-mailpoet (approved):

veljkho (approved):

PR #3183, opened 2020-11-18
Add new scenarios to form editor's block custom field [MAILPOET-3310]

veljkho:
[MAILPOET-3310](https://mailpoet.atlassian.net/browse/MAILPOET-3310)

### DIFF
--- a/assets/js/src/form_editor/blocks/custom_checkbox/custom_field_settings.jsx
+++ b/assets/js/src/form_editor/blocks/custom_checkbox/custom_field_settings.jsx
@@ -53,6 +53,7 @@ const CustomFieldSettings = ({
         />
         <input
           value={localCheckboxLabel}
+          data-automation-id="custom_field_value_settings_value"
           type="text"
           onChange={(event) => setLocalCheckboxLabel(event.target.value)}
         />
--- a/assets/js/src/form_editor/blocks/custom_checkbox/custom_field_settings.jsx
+++ b/assets/js/src/form_editor/blocks/custom_checkbox/custom_field_settings.jsx
@@ -53,7 +53,7 @@ const CustomFieldSettings = ({
         />
         <input
           value={localCheckboxLabel}
-          data-automation-id="custom_field_value_settings_value"
+          data-automation-id="settings_custom_checkbox_value"
           type="text"
           onChange={(event) => setLocalCheckboxLabel(event.target.value)}
         />
--- a/assets/js/src/form_editor/blocks/custom_date/custom_field_settings.jsx
+++ b/assets/js/src/form_editor/blocks/custom_date/custom_field_settings.jsx
@@ -51,6 +51,7 @@ const CustomFieldSettings = ({
     return (
       <SelectControl
         label={MailPoet.I18n.t('customFieldDateFormat')}
+        data-automation-id="settings_custom_date_order"
         value={localDateFormat}
         onChange={(value) => setLocalDateFormat(value)}
         options={dateFormats.map((format) => ({
@@ -75,6 +76,7 @@ const CustomFieldSettings = ({
       />
       <SelectControl
         label={MailPoet.I18n.t('customFieldDateType')}
+        data-automation-id="settings_custom_date_type"
         value={localDateType}
         onChange={(value) => {
           setLocalDateType(value);
--- a/assets/js/src/form_editor/blocks/custom_date/custom_field_settings.jsx
+++ b/assets/js/src/form_editor/blocks/custom_date/custom_field_settings.jsx
@@ -51,7 +51,7 @@ const CustomFieldSettings = ({
     return (
       <SelectControl
         label={MailPoet.I18n.t('customFieldDateFormat')}
-        data-automation-id="settings_custom_date_order"
+        data-automation-id="settings_custom_date_format"
         value={localDateFormat}
         onChange={(value) => setLocalDateFormat(value)}
         options={dateFormats.map((format) => ({
--- a/tests/acceptance/FormEditorCreateCustomFieldCest.php
+++ b/tests/acceptance/FormEditorCreateCustomFieldCest.php
@@ -7,18 +7,17 @@ use MailPoet\Test\DataFactories\Segment;
 
 class FormEditorCreateCustomFieldCest {
   public function createCustomSelect(\AcceptanceTester $i) {
+    $i->wantTo('Create custom field: select');
     $segmentFactory = new Segment();
     $segmentName = 'Fancy List';
     $segment = $segmentFactory->withName($segmentName)->create();
     $formName = 'My fancy form';
     $form = new Form();
     $form->withName($formName)->withSegments([$segment])->create();
-    $i->wantTo('Add first and last name to the editor');
-    $i->login();
-    $i->amOnMailPoetPage('Forms');
-    $i->waitForText($formName);
-    $i->clickItemRowActionByItemName($formName, 'Edit');
-    $i->waitForElement('[data-automation-id="form_title_input"]');
+    
+    // Go and edit the form
+    $this->editTheForm($i);
+
     // Insert create custom field block
     $i->addFromBlockInEditor('Create Custom Field');
 
@@ -30,11 +29,8 @@ class FormEditorCreateCustomFieldCest {
     $i->fillField('[data-automation-id="custom_field_value_settings_value"]', 'First option'); // Configure first option
     $i->click('[data-automation-id="custom_field_values_add_item"]'); // Add second option
 
-    // Save custom field
-    $i->click('[data-automation-id="create_custom_field_submit"]');
-    $i->waitForText('Custom field saved', 10, '.automation-dismissible-notices');
-    $i->seeNoJSErrors();
-    $i->click('.automation-dismissible-notices .components-notice__dismiss'); // Hide notice
+    // Save the custom field
+    $this->saveCustomFieldBlock($i);
 
     // Check field was added correctly
     $this->checkCustomSelectInForm($i);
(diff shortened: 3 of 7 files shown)


assets/js/src/form_editor/blocks/custom_checkbox/custom_field_settings.jsx
```diff
@@ -53,6 +53,7 @@ const CustomFieldSettings = ({
         />
         <input
           value={localCheckboxLabel}
+          data-automation-id="custom_field_value_settings_value"
```
lysyjan, line 56: The word _value_ is twice in this ID. Do you think that we could change it?

assets/js/src/form_editor/blocks/custom_date/custom_field_settings.jsx
```diff
@@ -51,6 +51,7 @@ const CustomFieldSettings = ({
     return (
       <SelectControl
         label={MailPoet.I18n.t('customFieldDateFormat')}
+        data-automation-id="settings_custom_date_order"
```
lysyjan, line 54: Label and value have suffix _format_. Why did you use suffix _order_ here?

tests/acceptance/FormEditorCreateCustomFieldCest.php
```diff
@@ -47,11 +43,301 @@ public function createCustomSelect(\AcceptanceTester $i) {
     $this->checkCustomSelectInForm($i);
   }
 
+  public function createCustomTextInput(\AcceptanceTester $i) {
+    $i->wantTo('Create custom field: text input');
+    $segmentFactory = new Segment();
```
lysyjan, line 48: You created a function `editTheForm` which contains recurring code. This is nice. But when I'm looking at the start of new methods then there are 6 still repeating lines. Maybe I'm wrong a this lines have a reason why they aren't in some function. What do you think?

tests/acceptance/FormEditorCreateCustomFieldCest.php
```diff
@@ -47,11 +43,301 @@ public function createCustomSelect(\AcceptanceTester $i) {
...
+
+    // Save the custom field
+    $this->saveCustomFieldBlock($i);
+
+    // Save the form
+    $i->saveFormInEditor();
+
+    // Reload page and check data were saved
+    $i->reloadPage();
+    $this->checkCustomCheckboxInForm($i);
```
lysyjan, line 220: You created a method to check the custom checkbox. If you add a parameter with the checkbox's value, you can reuse this method at the end after the second reload. You would call this method twice. First time with the second parameter's value `Option 1` and the second time with the second's parameter `New option`. What do you mean about this change? If you agree, then this solution could be used in other methods too.

assets/js/src/form_editor/blocks/custom_checkbox/custom_field_settings.jsx
veljkho: Yes, you're right, my fault.

assets/js/src/form_editor/blocks/custom_date/custom_field_settings.jsx
veljkho: Yes, you're right. My fault.

assets/js/src/form_editor/blocks/custom_checkbox/custom_field_settings.jsx
veljkho: Yes, I changed it to be more concise.

tests/acceptance/FormEditorCreateCustomFieldCest.php
veljkho: Yes, this is a good idea. I Will update accordingly. Thanks!

tests/acceptance/FormEditorCreateCustomFieldCest.php
veljkho: Well, never seen this in our acceptance suite, but you're right. Let's try with this now, since it is very frequently used within the tests.

tests/acceptance/FormEditorCreateCustomFieldCest.php
```diff
@@ -6,19 +6,32 @@
 use MailPoet\Test\DataFactories\Segment;
 
 class FormEditorCreateCustomFieldCest {
-  public function createCustomSelect(\AcceptanceTester $i) {
+  private function prepareTheForm(\AcceptanceTester $i) {
```
costasovo, line 9: I can see that this function and also editTheForm is used at the beginning of every test in this suite so I guess you can just put the code into _before hook function that is triggered before every test.

tests/acceptance/FormEditorCreateCustomFieldCest.php
```diff
@@ -30,11 +43,8 @@ public function createCustomSelect(\AcceptanceTester $i) {
     $i->fillField('[data-automation-id="custom_field_value_settings_value"]', 'First option'); // Configure first option
     $i->click('[data-automation-id="custom_field_values_add_item"]'); // Add second option
 
-    // Save custom field
-    $i->click('[data-automation-id="create_custom_field_submit"]');
-    $i->waitForText('Custom field saved', 10, '.automation-dismissible-notices');
-    $i->seeNoJSErrors();
-    $i->click('.automation-dismissible-notices .components-notice__dismiss'); // Hide notice
+    // Save the custom field
+    $this->saveCustomFieldBlock($i);
```
costasovo, line 47: I think this comment and maybe a couple of others are redundant since they actually just repeat what is already said in the name of the following method.
Also [recently we found out](https://mailpoet.slack.com/archives/CAUJT7D1Q/p1599647422007800) that there is a nicer way of adding comments into acceptance tests using `$i->wantTo('Do something');` so that the comment is also printed out on the output.
I think it would be nicer to remove redundant comments and replace others with `$i->wantTo('Do something');` type of comments. What do you think?

costasovo (commented):
Nice extension of the test! Overall looks good I have just a couple of notes/suggestions.

tests/acceptance/FormEditorCreateCustomFieldCest.php
```diff
@@ -6,19 +6,32 @@
...
     $segmentFactory = new Segment();
     $segmentName = 'Fancy List';
     $segment = $segmentFactory->withName($segmentName)->create();
     $formName = 'My fancy form';
     $form = new Form();
     $form->withName($formName)->withSegments([$segment])->create();
-    $i->wantTo('Add first and last name to the editor');
+  }
+
+  private function editTheForm($i) {
```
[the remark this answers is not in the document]
veljkho: Okay, that sounds like a better name.

tests/acceptance/FormEditorCreateCustomFieldCest.php
veljkho: Yes, will make such a method.

tests/acceptance/FormEditorCreateCustomFieldCest.php
veljkho: Well, I don't think that's a good idea, because commenting is used for dividing the code between the notice.
With `$i->wantTo('Do something');`, the code will be all the same, hard to see commenting which is important when previewing the code, but on the other hand we would have the output. Not sure if we really need additional output, was able to see where failed and why.

pavel-mailpoet: I agree with Rosta here. I think it is better to use the syntax `$i->wantTo('Do something');` you still see it in the code as comments and you can divide the code using empty lines if you want to.

costasovo: @veljkho Here is a sample of the change from the current PR:

Before:
```php
$i->wantTo('Create custom field: text input');
// Configure, check and save the custom field block
$i->waitForElement('[data-automation-id="create_custom_field_form"]');
$i->selectOption('[data-automation-id="create_custom_field_type_select"]', 'Text Input');
$i->fillField('[data-automation-id="create_custom_field_name_input"]', 'My custom text input');
$i->selectOption('[data-automation-id="settings_custom_text_input_validation_type"]', 'Numbers only');
$this->saveCustomFieldBlock($i);

// Save, reload and check data were saved
$i->saveFormInEditor();
$i->reloadPage();
$this->checkCustomTextInputInForm($i);


// Change text input validation
$i->selectOption('[data-automation-id="settings_custom_text_input_validation_type"]', 'Alphanumerical');
$i->click('Update custom field');
$i->waitForText('Custom field saved.');

// Update label and save the form
$i->fillField('[data-automation-id="settings_custom_text_label_input"]', 'My updated custom text input');
$i->saveFormInEditor();
```

After:
```php
$i->wantTo('Create custom field: text input');

$i->wantTo('Configure, check and save the custom field block');
$i->waitForElement('[data-automation-id="create_custom_field_form"]');
$i->selectOption('[data-automation-id="create_custom_field_type_select"]', 'Text Input');
$i->fillField('[data-automation-id="create_custom_field_name_input"]', 'My custom text input');
$i->selectOption('[data-automation-id="settings_custom_text_input_validation_type"]', 'Numbers only');
$this->saveCustomFieldBlock($i);

$i->wantTo('Configure, check and save the custom field block');
$i->saveFormInEditor();
$i->reloadPage();
$this->checkCustomTextInputInForm($i);

$i->wantTo('Change input validation');
$i->selectOption('[data-automation-id="settings_custom_text_input_validation_type"]', 'Alphanumerical');
$i->click('Update custom field');
$i->waitForText('Custom field saved.');

$i->wantTo('Update input label');
$i->fillField('[data-automation-id="settings_custom_text_label_input"]', 'My updated custom text input');

$i->saveFormInEditor();
```
I see your point that comments are better highlighted in a code editor. On the other hand, if you separate those blocks with empty lines I think it is not bad either plus you have a benefit of human-readable text on the output. I think seeing `'Change input validation'` in error output helps to find out what part of a test failed without having to see full source code. We lived with simple comments all the time so I guess it is not that important. I'll approve the PR and let you decide if you want to add the change.

costasovo (approved):

PR #3190, opened 2020-11-24
Form editor tutorial [MAILPOET-3206]

pavel-mailpoet:
[MAILPOET-3206]

[MAILPOET-3206]: https://mailpoet.atlassian.net/browse/MAILPOET-3206

### DIFF
--- a/assets/js/src/form_editor/components/tutorial.tsx
+++ b/assets/js/src/form_editor/components/tutorial.tsx
@@ -26,7 +26,7 @@ export const Tutorial = () => {
       isDismissible
       onRequestClose={onClose}
     >
-      <div className="mailpoet_drag_and_drop_tutorial">
+      <div className="mailpoet_drag_and_drop_tutorial" data-automation-id="form-editor-tutorial">
         {/* eslint-disable-next-line jsx-a11y/media-has-caption */}
         <video
           style={{ height: '640px' }}
--- a/assets/js/src/form_editor/components/tutorial.tsx
+++ b/assets/js/src/form_editor/components/tutorial.tsx
@@ -1,6 +1,9 @@
 import React from 'react';
-import Modal from 'common/modal/modal';
 import { useSelect, useDispatch } from '@wordpress/data';
+
+import Modal from 'common/modal/modal';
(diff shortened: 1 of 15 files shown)


assets/js/src/form_editor/store/controls.jsx
```diff
@@ -167,6 +167,15 @@ export default {
     dispatch('core/block-editor').resetBlocks(updatedBlocks);
   },
 
+  TUTORIAL_DISMISS() {
```
costasovo, line 170: We have a common control `CALL_API` so there is no need to create a new control for that. `CALL_API` also returns info if an action was a success or failed so you can react on that in your actions generator function.

Here is an [example usage in the form editor](https://github.com/mailpoet/mailpoet/blob/master/assets/js/src/form_editor/store/actions.ts#L207-L220).

I think it would be nicer to reuse this control so that we don't repeat ajax calls code and error handling etc.

assets/js/src/form_editor/components/tutorial.tsx
```diff
@@ -17,6 +18,13 @@ export const Tutorial = () => {
     return null;
   }
 
+  const installedAtDate = new Date(MailPoet.installedAt);
+  const sevenDaysAgo = new Date();
+  sevenDaysAgo.setDate(sevenDaysAgo.getDate() - 7);
+  if (installedAtDate < sevenDaysAgo) {
+    return null; // only show to new users
+  }
+
```
costasovo, line 27: I couldn't find this in spec. I'm not sure if that is a good idea because many users may decide to use form editor or even the plugin much later than in 7 days after installing it.
If we don't want to show it to users who have some experience with the form editor maybe we could check if they already have some older form.
What do you think?

costasovo (changes requested):
Good job! Did you think about adding an acceptance test checking that tutorial is shown? It is not critical functionality, but it has some conditions for displaying so it might be useful to check we didn't break it.

costasovo (dismissed):

veljkho (changes requested):
Missing title for the popup video.

costasovo (approved):

veljkho (approved):

PR #3191, opened 2020-11-24
Add "US East (Ohio)" region to Amazon SES options

derek-j-pitman:
New AWS accounts in US East are automatically assigned to the region "us-east-2", which corresponds to "US East (Ohio)".

This PR adds the corresponding strings where there are AWS regions listed in the code. Editing "US East (N. Virginia)" in place is not sufficient.


pavel-mailpoet (approved):

pavel-mailpoet:
Thank you @derek-j-pitman for your contribution, we will test it and merge it.

veljkho (approved):

PR #3194, opened 2020-11-25
Remove premium autoinstall [MAILPOET-3315]

pavel-mailpoet:
[MAILPOET-3315]

[MAILPOET-3315]: https://mailpoet.atlassian.net/browse/MAILPOET-3315


assets/js/src/settings/pages/key_activation/messages/premium_messages.tsx
```diff
@@ -3,6 +3,7 @@ import MailPoet from 'mailpoet';
 import { useSelector } from 'settings/store/hooks/index';
 import { PremiumInstallationStatus, PremiumStatus } from 'settings/store/types';
 import PremiumInstallationMessages from './premium_installation_messages';
+import Button from '../../../../common/button/button';
```
costasovo, line 5: Can you please path from the root instead of a relative path?

views/settings.html
```diff
@@ -178,10 +179,8 @@
     'premiumTabPremiumActiveMessage': __('MailPoet Premium is active', 'mailpoet'),
     'premiumTabPremiumInstallingMessage': __('MailPoet Premium plugin is being installed', 'mailpoet'),
     'premiumTabPremiumActivatingMessage': __('MailPoet Premium plugin is being activated', 'mailpoet'),
```
costasovo, line 181: I think we no longer need these 2 strings.

views/settings.html
```diff
@@ -178,10 +179,8 @@
...
-    'premiumTabPremiumNotInstalledMessage': __('MailPoet Premium is not installed.', 'mailpoet'),
-    'premiumTabPremiumInstallMessage': __('Install MailPoet Premium plugin', 'mailpoet'),
-    'premiumTabPremiumNotActiveMessage': __('MailPoet Premium is not active.', 'mailpoet'),
-    'premiumTabPremiumActivateMessage': __('Activate MailPoet Premium plugin', 'mailpoet'),
+    'premiumTabPremiumNotActiveMessage': __('MailPoet Premium is not installed.', 'mailpoet'),
+    'premiumTabPremiumActivateMessage': __('Download MailPoet Premium plugin', 'mailpoet'),
```
costasovo, line 183: Can we use text stubs that fit a bit better? I find these two a bit confusing.
For example `premiumTabPremiumNotInstalledMessage` `premiumTabPremiumDownloadMessage `

costasovo (changes requested):
Looks good! I have just 3 small remarks.

costasovo (approved):

veljkho (changes requested):
Sorry, there was missing the 3rd case in the specs of the Jira ticket. I now updated the 3rd case when Premium is installed but not activated, we should show a different message without a button.

veljkho (approved):

PR #3199, opened 2020-11-26
Update scenario to include heading and paragraph [MAILPOET-3328]

veljkho:
[MAILPOET-3328](https://mailpoet.atlassian.net/browse/MAILPOET-3328)

Check the Jira ticket for more info , if needed.

### DIFF
--- a/tests/acceptance/FormEditorTextInputStylesCest.php
+++ b/tests/acceptance/FormEditorTextInputStylesCest.php
@@ -41,13 +41,13 @@ class FormEditorTextInputStylesCest {
 
     $i->wantTo('Add heading block and write some title');
     $i->addFromBlockInEditor('Heading');
-    $i->fillField('[data-title="Heading"]', 'Lorem Ipsum');
-    $i->see('Lorem Ipsum');
+    $i->fillField('[data-title="Heading"]', 'Heading Lorem');
+    $i->see('Heading Lorem');
 
     $i->wantTo('Add paragraph block and write some text');
     $i->addFromBlockInEditor('Paragraph');
-    $i->fillField('[data-title="Paragraph"]', 'Lorem ipsum dolor sit amet');
-    $i->see('Lorem ipsum dolor sit amet');
+    $i->fillField('[data-title="Paragraph"]', 'Paragraph ipsum dolor');
+    $i->see('Paragraph ipsum dolor');
 
     $i->wantTo('Check email block has styles too and save the form');
     $i->assertAttributeContains('[data-automation-id="editor_first_name_input"]', 'style', 'border-width: 10px;');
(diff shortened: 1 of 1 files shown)


tests/acceptance/FormEditorTextInputStylesCest.php
```diff
@@ -35,12 +35,24 @@ public function changeTextInputStyles(\AcceptanceTester $i) {
     // Check element has styles
     $i->assertAttributeContains('[data-automation-id="editor_first_name_input"]', 'style', 'border-width: 10px;');
     $i->assertAttributeContains('[data-automation-id="editor_first_name_label"]', 'style', 'font-weight: bold;');
+
     // Apply to all
     $i->click('[data-automation-id="styles_apply_to_all"]');
-    // Check email block has styles too
+
+    // Add heading block and write some title
```
pavel-mailpoet, line 42: Could you please use `$i->wantTo` instead of comments in the test?

tests/acceptance/FormEditorTextInputStylesCest.php
```diff
@@ -35,12 +35,24 @@ public function changeTextInputStyles(\AcceptanceTester $i) {
...
+    $i->addFromBlockInEditor('Heading');
+    $i->fillField('[data-title="Heading"]', 'Lorem Ipsum');
+    $i->see('Lorem Ipsum');
+
+    // Add paragraph block and write some text
+    $i->addFromBlockInEditor('Paragraph');
+    $i->fillField('[data-title="Paragraph"]', 'Lorem ipsum dolor sit amet');
+    $i->see('Lorem ipsum dolor sit amet');
```
pavel-mailpoet, line 50: Should we also check that those new elements are present in the editor after the reload below? And also check that these elements are present in the rendered page?

pavel-mailpoet (changes requested):

tests/acceptance/FormEditorTextInputStylesCest.php
```diff
@@ -32,23 +32,37 @@ public function changeTextInputStyles(\AcceptanceTester $i) {
...
-    // Reload page and check data were saved
+
+    $i->wantTo('Reload page and check data were saved');
     $i->reloadPage();
     $i->waitForElement('[data-automation-id="form_title_input"]');
     $i->assertAttributeContains('[data-automation-id="editor_first_name_input"]', 'style', 'border-width: 10px;');
     $i->assertAttributeContains('[data-automation-id="editor_first_name_label"]', 'style', 'font-weight: bold;');
     $i->assertAttributeContains('[data-automation-id="editor_first_name_input"]', 'style', 'border-width: 10px;');
+    $i->see('Lorem Ipsum');
+    $i->see('Lorem ipsum dolor sit amet');
```
pavel-mailpoet, line 63: I am not sure those checks will work properly. You are using two texts that only differ in a single letter. I can see two problems with this:

1. Is the method `see` case sensitive? I tried to find it in the documentation, but I couldn't find it mentioned anywhere. 
2. Those texts are quite vague so if we see that the test failed because `'Lorem Ipsum'` was not found we would need to read the whole text to understand what is missing. I think it would be better to add the name of the block in the string. Or even better use the automation id.

pavel-mailpoet (changes requested):

veljkho:
Yes, you're right. `see` is not case sensitive.

pavel-mailpoet (approved):

PR #3202, opened 2020-11-30
Use curl for download Woocommerce plugin [MAILPOET-3321]

lysyjan:
[MAILPOET-3321]

[MAILPOET-3321]: https://mailpoet.atlassian.net/browse/MAILPOET-3321

### DIFF
--- a/tests/docker/codeception/docker-entrypoint.sh
+++ b/tests/docker/codeception/docker-entrypoint.sh
@@ -45,8 +45,36 @@ else
   wp core multisite-install $WP_CORE_INSTALL_PARAMS
 fi
 
+# Load Composer dependencies
+# Set SKIP_DEPS environment flag to not download them. E.g. you have downloaded them yourself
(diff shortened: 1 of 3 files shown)


tests/docker/codeception/docker-entrypoint.sh
```diff
@@ -46,7 +46,18 @@ else
 fi
 
 # install WooCommerce (activate & deactivate it to populate DB for backup)
-wp plugin install woocommerce
+if [[ ! -d "/wp-core/wp-content/plugins/woocommerce" ]]; then
```
pavel-mailpoet, line 49: Does this happen every single time when running acceptance tests? That means at least 5 times per build. Shall we do this functionality in the `build` phase and then share it in all acceptance test jobs?

lysyjan: I will try it now.

pavel-mailpoet (approved):

veljkho (approved):

PR #3203, opened 2020-11-30
Update woocommerce types search [MAILPOET-2553]

pavel-mailpoet:
[MAILPOET-2553]

[MAILPOET-2553]: https://mailpoet.atlassian.net/browse/MAILPOET-2553

### DIFF
--- a/lib/AdminPages/Pages/Newsletters.php
+++ b/lib/AdminPages/Pages/Newsletters.php
@@ -168,6 +168,27 @@ class Newsletters {
     $data['display_detailed_stats'] = Installer::getPremiumStatus()['premium_plugin_initialized'];
     $data['newsletters_templates_recently_sent_count'] = $this->newsletterTemplatesRepository->getRecentlySentCount();
 
+    $data['product_categories'] = $this->wp->getCategories(['taxonomy' => 'product_cat']);
+
+    usort($data['product_categories'], function ($a, $b) {
+      return strcmp($a->catName, $b->catName);
+    });
+
+    $data['products'] = $this->getProducts();
+
     $this->pageRenderer->displayPage('newsletters.html', $data);
+  }
+
+  private function getProducts() {
+    $products = $this->wp->getResultsFromWpDb(
+      "SELECT `ID`, `post_title` FROM {$this->wp->getWPTableName('posts')} WHERE `post_type` = %s ORDER BY `post_title` ASC;",
+      'product'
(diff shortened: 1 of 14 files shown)


lib/AdminPages/Pages/Newsletters.php
```diff
@@ -168,6 +168,27 @@ public function render() {
...
+      "SELECT `ID`, `post_title` FROM {$this->wp->getWPTableName('posts')} WHERE `post_type` = %s ORDER BY `post_title` ASC;",
+      'product'
+    );
+    return array_map(function ($product) {
+      return [
+        'title' => $product->post_title, // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
+        'ID' => $product->ID,
+      ];
+    }, $products);
+  }
```
costasovo, line 193: We have a bunch of methods in [form editor page controller](https://github.com/mailpoet/mailpoet/blob/master/lib/AdminPages/Pages/FormEditor.php#L358-L415) that do basically the same (maybe without ordering). I think it would be nice to put those into some reusable service and reuse it here and also unify formatted output. With this, we could also avoid additional mapping in views/newsletters.html. I think would be much nicer since now the fetching code is duplicated here and on the forms page and we have 2 slightly different output formats for products and categories. What do you think?

views/newsletters.html
```diff
@@ -29,6 +29,16 @@
       var mailpoet_premium_active = <%= json_encode(premium_plugin_active) %>;
       var mailpoet_transactional_emails_opt_in_notice_dismissed = '<%= transactional_emails_opt_in_notice_dismissed %>';
       var mailpoet_mss_key_invalid = <%= mss_key_invalid ? 'true' : 'false' %>;
+      var mailpoet_product_categories = <%= json_encode(product_categories)  %>;
+      mailpoet_product_categories = mailpoet_product_categories.map(function(category) {
```
costasovo, line 33: I would be nicer to do that in PHP so that all mappings happen in one place otherwise it is harder to follow where data comes from.

costasovo (changes requested):

pavel-mailpoet:
@veljkho as part of this ticket, there were some refactors. I changed the data loading for the segments autocomplete, and also for the form editor autocomplete.

costasovo (approved):

veljkho (approved):

PR #3208, opened 2020-12-02
Add enabling and disabling WP Users List [MAILPOET-3142]

lysyjan:
[MAILPOET-3142]

[MAILPOET-3142]: https://mailpoet.atlassian.net/browse/MAILPOET-3142

### DIFF
--- a/assets/js/src/segments/list.jsx
+++ b/assets/js/src/segments/list.jsx
@@ -110,6 +110,11 @@ const bulkActions = [
   },
 ];
 
+const isItemDeletable = (segment) => {
+  const isDeletable = !isSpecialSegment(segment);
+  return isDeletable;
+};
+
 const itemActions = [
   {
     name: 'edit',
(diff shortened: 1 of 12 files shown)


assets/js/src/segments/list.jsx
```diff
@@ -233,6 +238,23 @@ const itemActions = [
       return !isSpecialSegment(segment) && segment.automated_emails_subjects.length > 0;
     },
   },
+  {
+    name: 'delete_wp_users',
+    className: 'mailpoet-hide-on-mobile',
+    label: MailPoet.I18n.t('trashAndDisable'),
+    onClick: function onClick(segment) {
+      MailPoet.Notice.error(
```
pavel-mailpoet, line 246: Is the `error` really the best type of notice here? And also instead of the outdated `MailPoet.Notice`, we should use `assets/js/src/notices/notice.tsx`.

pavel-mailpoet (changes requested):
Looks good @lysyjan. I have only one small remark about the notice. But also, should we add an acceptance test for enabling/disabling the List? What do you think?

assets/js/src/segments/list.jsx
lysyjan: I used the same what is on line 229. I didn't know that it's outdated.

pavel-mailpoet (approved):

veljkho (changes requested):
There is one thing left from the specs:
- 1.a.iii. - Issue is that when I have simple newsletter scheduled for sending and using `WordPress Users` list, I was able to trash the list `WordPress Users` list, but should have been warned that it's attached to the newsletter.
PS. It is working for Woo automatic emails.

pavel-mailpoet (approved):

veljkho (approved):

PR #3209, opened 2020-12-02
PHP8 support fixes [MAILPOET-3296]

costasovo:
[MAILPOET-3296]

[MAILPOET-3296]: https://mailpoet.atlassian.net/browse/MAILPOET-3296


lib/Doctrine/Driver/PDOConnection.php
```diff
@@ -39,15 +40,21 @@ public function exec($statement) {
...
   }
 
   public function prepare($prepareString, $driverOptions = []) {
     try {
-      return $this->createStatement(
-        $this->connection->prepare($prepareString, $driverOptions)
-      );
+      $preparedStatement = $this->connection->prepare($prepareString, $driverOptions);
+      if ($preparedStatement === false) {
+        throw new RuntimeException('Unable to prepare PDOStatement for ' . $prepareString);
```
pavel-mailpoet, line 55: Is this safe enough? Can we throw the prepared string? What if we leak any private data here? Can we?

pavel-mailpoet (dismissed):

veljkho (approved):

PR #3212, opened 2020-12-03
Use Doctrine in Segments::save() [MAILPOET-3167]

lysyjan:
[MAILPOET-3167]

[MAILPOET-3167]: https://mailpoet.atlassian.net/browse/MAILPOET-3167

### DIFF
--- a/lib/Segments/SegmentSaveController.php
+++ b/lib/Segments/SegmentSaveController.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace MailPoet\Segments;
+
+use InvalidArgumentException;
+use MailPoet\Entities\SegmentEntity;
+use MailPoet\NotFoundException;
+
+class SegmentSaveController {
+  /** @var SegmentsRepository */
+  private $segmentsRepository;
+
+  public function __construct(
+    SegmentsRepository $segmentsRepository
+  ) {
+    $this->segmentsRepository = $segmentsRepository;
+  }
+
+  public function save(array $data = []): SegmentEntity {
+    $this->checkSegmenUniqueName($data['name'] ?? '');
+
+    $segment = $this->createOrUpdateSegment($data);
+    return $segment;
+  }
+
+  private function checkSegmenUniqueName(string $name): void {
+    $segment = $this->segmentsRepository->findOneBy(['name' => $name]);
+    if ($segment) {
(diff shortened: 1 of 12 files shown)


lib/Segments/SegmentSaveController.php
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  private function checkSegmenUniqueName(string $name): void {
+    $segment = $this->segmentsRepository->findOneBy(['name' => $name]);
+    if ($segment) {
+      throw new InvalidArgumentException("Segment with name: '{$name}' already exists.");
+    }
+  }
+
+  private function createOrUpdateSegment(array $data): SegmentEntity {
```
pavel-mailpoet, line 33: Can (should) we have this method available for other code as well? I am thinking about moving it to the segments repository. This can be useful universally, can't it? We can use it in tests and everywhere where we want to create a segment.

pavel-mailpoet (commented):
This is great, thanks. I have just a small question.

lib/Segments/SegmentSaveController.php
lysyjan: I understand, but I'm not sure that this type of method should be part of the repository. I guess that this looks like a method for some class of type factory. What thinks @costasovo?

costasovo: The method is dependent on a specific data format we send in API. So with this implementation is coupled with API. What about changing the method interface to  `createOrUpdateSegment(string $name, string $description, ?int $id)`?

pavel-mailpoet: > The method is dependent on a specific data format we send in API.

I don't see that. It is just an array with the data. It has the same signature as the old `createOrUpdate` method from old models.

lysyjan: Change the method interface makes sense. I would do it.

costasovo: I don't like the old models signature that much. I can imagine the usage of an array for data coming from API or some other input, but for example in tests or some other internal usage, I find it better to just call a method with typed params instead of crafting some specific array.

pavel-mailpoet (approved):

veljkho (approved):

PR #3220, opened 2020-12-17
Fix update segment when the name is the same [MAILPOET-3343]

lysyjan:
[MAILPOET-3343]

[MAILPOET-3343]: https://mailpoet.atlassian.net/browse/MAILPOET-3343

### DIFF
--- a/lib/Segments/SegmentsRepository.php
+++ b/lib/Segments/SegmentsRepository.php
@@ -33,6 +33,23 @@ class SegmentsRepository extends Repository {
     return $countMap;
   }
 
+  public function isNameUnique(string $name, ?int $id): bool {
+    $qb = $this->doctrineRepository->createQueryBuilder('s')
+      ->select('s')
+      ->where('s.name = :name')
+      ->setParameter('name', $name);
+
+    if ($id !== null) {
+      $qb->andWhere('s.id != :id')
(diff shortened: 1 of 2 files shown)


lib/Segments/SegmentsRepository.php
```diff
@@ -33,6 +33,23 @@ public function getCountsPerType(): array {
     return $countMap;
   }
 
+  public function isNameUnique(string $name, ?int $id): bool {
```
costasovo, line 36: Does this method have to be public?

costasovo (commented):
Looks good! Just a one question.

lib/Segments/SegmentsRepository.php
lysyjan: How can I call this method from the Controller when it is private/protected?

costasovo: Oh Sorry! My bad! I read the diff as it was one file and thought that it is also in the controller.

costasovo (approved):

lib/Segments/SegmentsRepository.php
lysyjan: Don't worry. Thank you, that you have looked at this PR during support rotation

veljkho (approved):

PR #3225, opened 2020-12-28
Fix saving clicked links in dynamic segments [MAILPOET-3292]

wxa:
Please make sure saving and editing clicked links in dynamic segments works properly.


costasovo (approved):
Works as expected! 👍

veljkho (approved):

PR #3229, opened 2020-12-29
Fix list selection in form when placed in column block [MAILPOET-3297]

costasovo:
[MAILPOET-3297]

[MAILPOET-3297]: https://mailpoet.atlassian.net/browse/MAILPOET-3297


wxa (approved):
Nice job!

veljkho (approved):

PR #3239, opened 2021-01-04
Remove closing button on listings pages [MAILPOET-3313]

pavel-mailpoet:
[MAILPOET-3313]

[MAILPOET-3313]: https://mailpoet.atlassian.net/browse/MAILPOET-3313


costasovo (changes requested):
I think you have found another close button that we may want to hide 👍 
We still need to hide the close button for the case when the newsletters DB table is empty see https://d.pr/v/vpLTuq.

costasovo (approved):

veljkho (approved):

PR #3244, opened 2021-01-06
Add new scenario covering import WP users as subscribers [MAILPOET-3355]

veljkho:
[MAILPOET-3355](https://mailpoet.atlassian.net/browse/MAILPOET-3355)


costasovo (approved):
@veljkho Looks good! For some reason, this PR is marked as a draft so I'm not merging it.

PR #3249, opened 2021-01-07
Source reCAPTCHA API only when needed [MAILPOET-3363]

lunar-debian:
When using a subscription form as an iframe, our CSP catched up that the reCAPTCHA API was included despite MailPoet being configured to use its internal CAPTCHA. The following change will only output the relevant <script> tag when MailPoet uses reCAPTCHA.


wxa (approved):
Thank you!

veljkho (approved):

PR #3250, opened 2021-01-07
Fix getting subscribers without a list [MAILPOET-3285]

lysyjan:
[MAILPOET-3285]

[MAILPOET-3285]: https://mailpoet.atlassian.net/browse/MAILPOET-3285

### DIFF
--- a/lib/Subscribers/SubscriberListingRepository.php
+++ b/lib/Subscribers/SubscriberListingRepository.php
@@ -130,10 +130,14 @@ class SubscriberListingRepository extends ListingRepository {
       return;
     }
     if ($filters['segment'] === self::FILTER_WITHOUT_LIST) {
-      $queryBuilder->leftJoin('s.subscriberSegments', 'ssg', Join::WITH, (string)$queryBuilder->expr()->eq('ssg.status', ':statusSubscribed'))
-        ->leftJoin('ssg.segment', 'sg', Join::WITH, (string)$queryBuilder->expr()->isNull('sg.deletedAt'))
+        $queryBuilder->leftJoin('s.subscriberSegments', 'ssg')
+        ->leftJoin('ssg.segment', 'sg')
+        ->leftJoin(SubscriberEntity::class, 's2', Join::WITH, (string)$queryBuilder->expr()->eq('s.id', 's2.id'))
+        ->leftJoin('s2.subscriberSegments', 'ssg2', Join::WITH, 'ssg2.status = :statusSubscribed AND sg.id <> ssg2.segment')
+        ->leftJoin('ssg2.segment', 'sg2', Join::WITH, (string)$queryBuilder->expr()->isNull('sg2.deletedAt'))
         ->andWhere('s.deletedAt IS NULL')
-        ->andWhere('sg.id IS NULL')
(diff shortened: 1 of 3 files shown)


lib/Subscribers/SubscriberListingRepository.php
```diff
@@ -130,10 +130,17 @@ protected function applyFilters(QueryBuilder $queryBuilder, array $filters) {
...
+      $qbSubselect->select('s2.id')
+        ->from(SubscriberEntity::class, 's2')
+        ->leftJoin('s2.subscriberSegments', 'ssg2', Join::WITH, (string)$qbSubselect->expr()->eq('ssg2.status', ':statusSubscribed'))
+        ->join('ssg2.segment', 'sg2', Join::WITH, (string)$qbSubselect->expr()->isNull('sg2.deletedAt'));
+
+        $queryBuilder->leftJoin('s.subscriberSegments', 'ssg', Join::WITH, (string)$queryBuilder->expr()->eq('ssg.status', ':statusSubscribed'))
         ->leftJoin('ssg.segment', 'sg', Join::WITH, (string)$queryBuilder->expr()->isNull('sg.deletedAt'))
         ->andWhere('s.deletedAt IS NULL')
         ->andWhere('sg.id IS NULL')
+        ->andWhere($queryBuilder->expr()->notIn('s.id', $qbSubselect->getDQL()))
```
costasovo, line 143: I think this might cause performance issues. As far as I know, subqueries in WHERE IN are executed for each row.
I see some other conditions so planner should probably use those to limit rows in result first, but still, it might be an issue for users who have more users with no list. And also it might be safer not to rely too much on MySQL planner.

I think we could either execute the subselect and use the result to avoid it or I think there is some trick using the subquery within JOIN. I think that in this case we could probably add left join and join the subquery and in where clause we could then use something like `s2.id IS NULL` 🤔

lib/Subscribers/SubscriberListingRepository.php
```diff
@@ -234,13 +241,19 @@ public function getFilters(ListingDefinition $definition): array {
...
+      ->join('ssg2.segment', 'sg2', Join::WITH, (string)$queryBuilderNoSegmentSubselect->expr()->isNull('sg2.deletedAt'));
     $queryBuilderNoSegment = clone $queryBuilder;
     $subscribersWithoutSegment = $queryBuilderNoSegment
-      ->select('COUNT(s) AS subscribersCount')
+      ->select('COUNT(DISTINCT s) AS subscribersCount')
       ->leftJoin('s.subscriberSegments', 'ssg', Join::WITH, (string)$queryBuilderNoSegment->expr()->eq('ssg.status', ':statusSubscribed'))
       ->leftJoin('ssg.segment', 'sg', Join::WITH, (string)$queryBuilderNoSegment->expr()->isNull('sg.deletedAt'))
       ->andWhere('s.deletedAt IS NULL')
       ->andWhere('sg.id IS NULL')
+      ->andWhere($queryBuilder->expr()->notIn('s.id', $queryBuilderNoSegmentSubselect->getDQL()))
```
costasovo, line 256: Similar to the above comment.

costasovo (commented):
I think there might be a performance issue. See comments.

costasovo (approved):

veljkho (approved):

PR #3252, opened 2021-01-08
Refactor dynamic segments listings to doctrine [MAILPOET-3361]

pavel-mailpoet:
[MAILPOET-3361]

[MAILPOET-3361]: https://mailpoet.atlassian.net/browse/MAILPOET-3361

### DIFF
--- a/lib/Segments/DynamicSegments/DynamicSegmentsListingRepository.php
+++ b/lib/Segments/DynamicSegments/DynamicSegmentsListingRepository.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace MailPoet\Segments\DynamicSegments;
+
+use MailPoet\Entities\SegmentEntity;
+use MailPoet\Segments\SegmentListingRepository;
+
+class DynamicSegmentsListingRepository extends SegmentListingRepository {
+  protected $types = [SegmentEntity::TYPE_DYNAMIC];
+}
--- a/lib/Segments/DynamicSegments/DynamicSegmentsListingRepository.php
+++ b/lib/Segments/DynamicSegments/DynamicSegmentsListingRepository.php
@@ -4,7 +4,12 @@ namespace MailPoet\Segments\DynamicSegments;
 
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Segments\SegmentListingRepository;
+use MailPoetVendor\Doctrine\ORM\QueryBuilder;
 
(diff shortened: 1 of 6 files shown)


lib/Segments/SegmentListingRepository.php
```diff
@@ -50,7 +52,7 @@ protected function applyFilters(QueryBuilder $queryBuilder, array $filters) {
   }
 
   protected function applyParameters(QueryBuilder $queryBuilder, array $parameters) {
-    $types = [SegmentEntity::TYPE_DEFAULT, SegmentEntity::TYPE_WP_USERS];
+    $types = $this->types;
     if ($this->wooCommerce->shouldShowWooCommerceSegment()) {
       $types[] = SegmentEntity::TYPE_WC_USERS;
     }
```
costasovo, line 58: The listing repository is used for both static and dynamic segments and WooCommerce segment is added dynamically based on WooCommerce plugin status. This causes that when WooCommerce plugin is active we display WooCommerce segment in both static and dynamic segment lists

lib/Segments/DynamicSegments/DynamicSegmentsListingRepository.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace MailPoet\Segments\DynamicSegments;
+
+use MailPoet\Entities\SegmentEntity;
+use MailPoet\Segments\SegmentListingRepository;
+
+class DynamicSegmentsListingRepository extends SegmentListingRepository {
+  protected $types = [SegmentEntity::TYPE_DYNAMIC];
```
costasovo, line 9: I think that instead of inheritance we could pass this as an argument or via a constructor or a setter.

costasovo (changes requested):
I found one issue that needs to be fixed and have one comment. Otherwise looks good.

costasovo (approved):

costasovo (changes requested):
Oh! I've just checked the related [Jira issue](https://mailpoet.atlassian.net/browse/MAILPOET-3164) and it seems that you did a bit different stuff that is in the spec. The name of the Jira card is probably misleading. This task is about removing the rest of the dynamic segments old model usages.
It is used in various places for fetching a list of segments please see the spec. all usages are listed there.

costasovo (approved):

veljkho (approved):

PR #3256, opened 2021-01-09
Missing i18n namespace in checkout optin label [MAILPOET-3360]

barryceelen:
MAILPOET-3360


costasovo:
Nice catch @barryceelen! Thank you!

costasovo (approved):

veljkho (approved):

PR #3268, opened 2021-01-14
Minor updates to the instructions on how to setup a MailPoet development environment

rodrigoprimo:
This PR makes three minor adjustments to the portion of the README.md file that details how to setup a MailPoet development environment:

- Removes docker and docker compose as requirements as they are used to run the acceptance tests but not MailPoet itself
- Adds composer as a MailPoet requirement
- Adds a note making it explicit that the instructions assume that there is already a working WP dev environment configured on the machine.

Since I'm new to MailPoet there is a high chance that I'm missing something and that the changes that I'm suggesting don't make sense for some reason. If that is the case, just let me know :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,6 @@ MailPoet done the right way.
 - PHP 7.0
 - NodeJS
 - WordPress
-- Docker & Docker Compose
 
 ## Installation
 ```bash
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ MailPoet done the right way.
 - WordPress
 
 ## Installation
+
+The instructions below assume you already have a working WordPress development environment.
+
 ```bash
 # go to WP plugins directory
 $ cd path_to_wp_directory/wp-content/plugins
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ MailPoet done the right way.
 # Setup
 
 ## Requirements
-- PHP 7.0
+- PHP >= 7.3 (only for the development environment, to run the plugin PHP >= 7.1 is required)
 - NodeJS
 - WordPress
 


README.md
```diff
@@ -13,11 +13,14 @@ MailPoet done the right way.
 
 ## Requirements
 - PHP 7.0
+- Composer
```
costasovo, line 16: We install `composer.phar` during [initialization of the `./do` command](https://github.com/mailpoet/mailpoet/blob/master/do#L16-L20). Composer is then available in `./tools/vendor/composer.phar`  and used in the `./do install` command so theoretically, the globally installed composer is not a must-have requirement. On the other hand, it might be more convenient for devs to use a global composer.
I would suggest marking it as an optional requirement. What do you think?
We should probably also mention we require composer 2.0 or newer.

Thanks to your PR I noticed that also PHP version in requirements is outdated. The plugin needs PHP 7.1 but the latest version of PHPScoper we use requires PHP7.3 to run so that means a developer needs to have at least PHP 7.3 installed.

Could you please update also the PHP version requirement?

costasovo (changes requested):
Thank you for the updates! I left a note about the composer requirement.

README.md
rodrigoprimo: Thanks for the review, @costasovo!

> We install composer.phar during initialization of the ./do command. 

I wasn't aware that composer is installed during the initialization of the `/do` command. Thanks for the explanation. For now, I simply removed the commit that was adding composer to the list of requirements as I don't have a strong opinion on whether we should include it as optional or not.

> Composer is then available in ./tools/vendor/composer.phar and used in the ./do install command so theoretically, the globally installed composer is not a must-have requirement.

I noticed that there is a [composer.phar in the root directory](https://github.com/mailpoet/mailpoet/blob/master/composer.phar) that was added to the repository. Is there a reason why we need composer installed in two places? Maybe the composer.phar in the root directory was added by mistake as the message of the [commit](https://github.com/mailpoet/mailpoet/commit/a751a2a459db433d3f81c6b72659a929105bfa16#diff-67ba8bffe9b4b295cfd4c5cd23d23edd02bf81febb63694af6c216db6f4741c0) and the [description of the related issue](https://mailpoet.atlassian.net/browse/MAILPOET-3150) doesn't mention anything about it?

> Could you please update also the PHP version requirement?

Sure, just added a commit to include it.

costasovo: > I noticed that there is a composer.phar in the root directory that was added to the repository. Is there a reason why we need composer installed in two places? Maybe the composer.phar in the root directory was added by mistake as the message of the commit and the description of the related issue doesn't mention anything about it?

Good catch! That one must have been added by a mistake. composer.phar was never part of the repository. Actually, it [used to be listed in .gitignore and we removed it from the ignored files when we moved it to the `tools` directory](https://github.com/mailpoet/mailpoet/commit/c0d77e456458b9a6fc3baf5438796e40d5acf4ab#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947).

Could you please add a commit with a deletion of the file?

rodrigoprimo: @costasovo, sure, but instead of adding a commit to this PR, I opted to create a separate PR: https://github.com/mailpoet/mailpoet/pull/3269

costasovo (approved):
Thanks again for updating the docs!

PR #3271, opened 2021-01-15
Fix a syntax issue in .env.sample

rodrigoprimo:
The .env.sample file contained a few comments using `//`. The problem is that `//` is not supported by vlucas/phpdotenv and this was resulting in the following fatal error when following the dev environment setup instructions and running `./do install`:

```
ERROR: Uncaught Dotenv\Exception\InvalidFileException: Failed to parse dotenv file due to unexpected whitespace. Failed at ["" // URL of your site (used for multisite env and equals to DOMAIN_CURRENT_SITE from wp-config.php)]. in /srv/www/mailpoet/public_html/wp-content/plugins/mailpoet/vendor/vlucas/phpdotenv/src/Parser.php:148
```

To fix this problem, this commit replaces all instances of `//` in .env.sample with the `#` (see
https://github.com/vlucas/phpdotenv#comments).

Quick test show that `//` doesn't work and `#` does:

```
>>> Dotenv\Dotenv::parse("FOO=Bar // some comment")
Dotenv\Exception\InvalidFileException with message 'Failed to parse dotenv file. Encountered unexpected whitespace at [Bar // some comment].'
>>> Dotenv\Dotenv::parse("FOO=Bar # some comment")
=> [
     "FOO" => "Bar",
   ]
```

Related slack discussion: https://a8c.slack.com/archives/C01H71ZLBGQ/p1610740191014600

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -6,12 +6,12 @@ WP_TEST_MAILER_ENABLE_SENDING="false"
 # Optional: for multisite acceptance tests
 WP_ROOT_MULTISITE="/var/www/wordpress"
 WP_TEST_MULTISITE_SLUG=""
-HTTP_HOST="" // URL of your site (used for multisite env and equals to DOMAIN_CURRENT_SITE from wp-config.php)
+HTTP_HOST="" # URL of your site (used for multisite env and equals to DOMAIN_CURRENT_SITE from wp-config.php)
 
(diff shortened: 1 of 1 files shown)


costasovo (approved):
Thank you for fixing it! 👍

PR #3272, opened 2021-01-18
Add check that newsletter body is not empty [MAILPOET-3295]

lysyjan:
[MAILPOET-3295]

[MAILPOET-3295]: https://mailpoet.atlassian.net/browse/MAILPOET-3295

### DIFF
--- a/assets/js/src/newsletter_editor/components/config.jsx
+++ b/assets/js/src/newsletter_editor/components/config.jsx
@@ -2,7 +2,7 @@ import App from 'newsletter_editor/App';
 import SuperModel from 'backbone.supermodel';
 import _ from 'underscore'; // eslint-disable-line func-names
 
-var Module = {};
+const Module = {};
 
 Module.ConfigModel = SuperModel.extend({
   defaults: {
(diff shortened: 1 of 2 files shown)


assets/js/src/newsletter_editor/components/config.js
```diff
@@ -32,7 +32,7 @@ App.on('before:start', function (BeforeStartApp, options) { // eslint-disable-li
 
   config.blockDefaults = _.extend(
     config.blockDefaults,
-    options.newsletter.body.blockDefaults || {}
+    (options.newsletter.body ? options.newsletter.body.blockDefaults : null) || {}
```
costasovo, line 35: I think it would be nicer if we used the same fallback value for both cases. I guess we can go with an empty object.
We could also simplify it by using optional chaining syntax `options.newsletter.body?.blockDefaults || {}`

costasovo (commented):
Good job finding the issue! I would suggest a small change so that we use the same fallback value. See my comment.

assets/js/src/newsletter_editor/components/config.js
lysyjan: I had a solution with optional chaining syntax at the begging, but eslint didn't allow it. I am going to try it again and I will investigate why.

costasovo: Oh! I read the condition incorrectly 🤦 . It is ok as is. My bad! Sorry. If you want to add the optional chaining please do so and switch back to me or just switch it back to me.

costasovo (approved):
👍

veljkho (approved):

PR #3273, opened 2021-01-18
Override #wpbody color to default for form-editor [MAILPOET-3305]

lysyjan:
[MAILPOET-3305]

[MAILPOET-3305]: https://mailpoet.atlassian.net/browse/MAILPOET-3305

### DIFF
--- a/assets/css/src/generic/_typography.scss
+++ b/assets/css/src/generic/_typography.scss
@@ -5,6 +5,10 @@
   line-height: $line-height;
 }
 
+.admin_page_mailpoet-form-editor #wpbody {
+  color: inherit;
+}
+
 a {
   text-decoration: underline;
 


assets/css/src/generic/_typography.scss
```diff
@@ -1,5 +1,4 @@
 #wpbody {
-  color: $color-text;
```
costasovo, line 2: I think we can't simply remove it from here as this would affect some other MailPoet plugin pages (e.g. https://d.pr/i/IBZOe0 or some texts in "help" page).
Maybe there is some common wrapper used for all MailPoet pages and we could move this definition there and somehow exclude problematic form editor page.

costasovo (changes requested):

costasovo (approved):
Thank you! 👍

veljkho (approved):

PR #3277, opened 2021-01-20
Remove patching for ternary operators warnigns [MAILPOET-3240]

lysyjan:
[MAILPOET-3240]

PHPScoper was updated with PR for composer v2.

[MAILPOET-3240]: https://mailpoet.atlassian.net/browse/MAILPOET-3240

### DIFF
--- a/prefixer/fix-carbon.php
+++ b/prefixer/fix-carbon.php
@@ -6,20 +6,6 @@ set_error_handler(function ($severity, $message, $file, $line) {
 });
 
 $replacements = [
-  // Fix for PHP 7.4 deprecation warnings for nested ternary operators without brackets
-  // This is caused by a bug in PHP Parser used within PHP Scoper. We can remove it after we update to PHP Scoper
-  // that contains updated PHP Parser.
-  [
-    'file' => '../vendor-prefixed/nesbot/carbon/src/Carbon/CarbonInterval.php',
-    'find' => [
-      '$relativeToNow ? $isFuture ? \'from_now\' : \'ago\' : ($isFuture ? \'after\' : \'before\')',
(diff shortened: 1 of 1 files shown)


prefixer/fix-carbon.php
```diff
@@ -1,70 +0,0 @@
...
-// cleanup file types by extension
-exec('find ' . __DIR__ . "/../vendor-prefixed/nesbot/carbon -type f -name '*.xml' -delete");
-exec('find ' . __DIR__ . "/../vendor-prefixed/nesbot/carbon -type f -name '*.neon' -delete");
-
-// cleanup
-exec('rm -r ' . __DIR__ . '/../vendor-prefixed/nesbot/Carbon/bin');
-exec('rm -r ' . __DIR__ . '/../vendor-prefixed/nesbot/carbon/src/Carbon/Cli');
-exec('rm -r ' . __DIR__ . '/../vendor-prefixed/nesbot/carbon/src/Carbon/PHPStan');
-exec('rm -r ' . __DIR__ . '/../vendor-prefixed/nesbot/carbon/src/Carbon/Laravel');
-exec('rm -r ' . __DIR__ . '/../vendor-prefixed/nesbot/carbon/src/Carbon/Doctrine');
```
costasovo, line 70: I think we shouldn't remove the whole file. It contains more stuff than just the fix for ternary operators. I guess we should keep these cleanups so that the plugin package size doesn't grow. It looks like it should be safe to remove replacement fixes.

costasovo (changes requested):

costasovo (approved):

veljkho (approved):

PR #3281, opened 2021-01-21
Fix flakey integration test [MAILPOET-3336]

pavel-mailpoet:
[MAILPOET-3336]

[MAILPOET-3336]: https://mailpoet.atlassian.net/browse/MAILPOET-3336


costasovo (approved):
Thanks!

PR #3282, opened 2021-01-21
Use email heading from Woo for the title in transactional emails [MAILPOET-3329]

lysyjan:
[MAILPOET-3329]

[MAILPOET-3329]: https://mailpoet.atlassian.net/browse/MAILPOET-3329

### DIFF
--- a/lib/WooCommerce/TransactionalEmailHooks.php
+++ b/lib/WooCommerce/TransactionalEmailHooks.php
@@ -37,7 +37,7 @@ class TransactionalEmailHooks {
       $this->wp->removeAction('woocommerce_email_header', $emailHeaderCallback);
       $this->wp->removeAction('woocommerce_email_footer', $emailFooterCallback);
       $this->wp->addAction('woocommerce_email_header', function($emailHeading) {
-        $this->renderer->render($this->getNewsletter());
+        $this->renderer->render($this->getNewsletter(), $emailHeading);
         echo $this->renderer->getHTMLBeforeContent($emailHeading);
       });
       $this->wp->addAction('woocommerce_email_footer', function() {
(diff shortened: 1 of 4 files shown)


lib/WooCommerce/TransactionalEmailHooks.php
```diff
@@ -37,7 +37,7 @@ public function useTemplateForWoocommerceEmails() {
       $this->wp->removeAction('woocommerce_email_header', $emailHeaderCallback);
       $this->wp->removeAction('woocommerce_email_footer', $emailFooterCallback);
       $this->wp->addAction('woocommerce_email_header', function($emailHeading) {
-        $this->renderer->render($this->getNewsletter());
+        $this->renderer->render($this->getNewsletter(), $emailHeading);
```
wxa, line 40: `$emailHeading` is not the email subject, but the [heading](https://github.com/woocommerce/woocommerce/blob/master/templates/emails/email-header.php#L48) used in an email body. Not exactly what the task requires, but still does the job. What do you think?

lysyjan: I have discussed this topic with Pavel because I wanted to use the subject, but we didn't find a possible solution. In my opinion, the current solution is good. I would update a specification in the ticket.

wxa (approved):

lib/WooCommerce/TransactionalEmailHooks.php
wxa: I've updated the ticket spec.

wxa: > I have discussed this topic with Pavel
If you were aware of that issue beforehand, you could mention this in the PR to avoid the back-and-forth.

lysyjan: Thank you. This is my fault. I'm sorry.

veljkho (approved):

PR #3283, opened 2021-01-21
Upgrade required npm version to 7.4 [MAILPOET-3280]

lysyjan:
[MAILPOET-3280]

[MAILPOET-3280]: https://mailpoet.atlassian.net/browse/MAILPOET-3280


costasovo (changes requested):
After upgrading to node 15 my local builds started failing on node-sass error:

`Error: Node Sass does not yet support your current environment: OS X 64-bit with Unsupported runtime (88)`

It seems that our current version of node-sass doesn't support Node 15 (https://github.com/sass/node-sass/releases/tag/v4.14.1)

We need to update to a newer node-sass version.

costasovo (approved):
Thank you! 👍

veljkho (approved):

PR #3284, opened 2021-01-21
Use timestamp for DateTime compare in Segments WPTest [MAILPOET-3367]

lysyjan:
[MAILPOET-3367]

[MAILPOET-3367]: https://mailpoet.atlassian.net/browse/MAILPOET-3367

### DIFF
--- a/tests/integration/Segments/WPTest.php
+++ b/tests/integration/Segments/WPTest.php
@@ -271,11 +271,15 @@ class WPTest extends \MailPoetTest {
     $subscribersCount = $this->getSubscribersCount();
     expect($subscribersCount)->equals(2);
     $subscriber1 = Subscriber::where("wp_user_id", $id)->findOne();
+    $deletedAt1 = Carbon::createFromFormat('Y-m-d H:i:s', $subscriber1->deletedAt);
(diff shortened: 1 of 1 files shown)


wxa (approved):

tests/integration/Segments/WPTest.php
```diff
@@ -271,11 +271,15 @@ public function testItSynchronizesNewUsersToDisabledWPSegmentAsUnconfirmedAndTra
     $subscribersCount = $this->getSubscribersCount();
     expect($subscribersCount)->equals(2);
     $subscriber1 = Subscriber::where("wp_user_id", $id)->findOne();
+    $deletedAt1 = Carbon::createFromFormat('Y-m-d H:i:s', $subscriber1->deletedAt);
+    assert($deletedAt1 instanceof Carbon);
```
wxa, line 275: Just a note for myself: I asked @lysyjan why does he use native PHP `assert` instead of Codeception's `expect`. He said it's required by PHPStorm: it throws an error when a method can return something other than a specific object. Interesting that PHPStan does not complain about it.

veljkho (approved):

PR #3286, opened 2021-01-25
Refactor Segments::duplicate() to Doctrine [MAILPOET-3170]

lysyjan:
[MAILPOET-3170]

[MAILPOET-3170]: https://mailpoet.atlassian.net/browse/MAILPOET-3170

### DIFF
--- a/lib/Segments/SegmentSaveController.php
+++ b/lib/Segments/SegmentSaveController.php
@@ -5,26 +5,20 @@ namespace MailPoet\Segments;
 use InvalidArgumentException;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SubscriberSegmentEntity;
-use MailPoet\Subscribers\SubscriberSegmentRepository;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class SegmentSaveController {
   /** @var SegmentsRepository */
   private $segmentsRepository;
 
-  /** @var SubscriberSegmentRepository */
-  private $subscriberSegmentRepository;
-
   /** @var EntityManager */
   private $entityManager;
 
   public function __construct(
     SegmentsRepository $segmentsRepository,
-    SubscriberSegmentRepository $subscriberSegmentRepository,
     EntityManager $entityManager
   ) {
     $this->segmentsRepository = $segmentsRepository;
-    $this->subscriberSegmentRepository = $subscriberSegmentRepository;
     $this->entityManager = $entityManager;
   }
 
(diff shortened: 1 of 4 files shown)


lib/Segments/SegmentSaveController.php
```diff
@@ -25,6 +33,33 @@ public function save(array $data = []): SegmentEntity {
...
+      $subscriber = $subscriberSegment->getSubscriber();
+      if (!$subscriber) {
+        continue;
+      }
+      $subscriberDuplicate = new SubscriberSegmentEntity(
+        $duplicate,
+        $subscriber,
+        SubscriberEntity::STATUS_SUBSCRIBED
+      );
+      $this->subscriberSegmentRepository->persist($subscriberDuplicate);
```
pavel-mailpoet, line 55: I am not sure if it is possible or practical, but should we close all of the queries here in a transaction?

lib/Segments/SegmentSaveController.php
```diff
@@ -25,6 +33,33 @@ public function save(array $data = []): SegmentEntity {
...
+    $duplicate->setName(sprintf(__('Copy of %s', 'mailpoet'), $segmentEntity->getName()));
+
+    $this->checkSegmenUniqueName($duplicate->getName(), $duplicate->getId());
+
+    $this->segmentsRepository->persist($duplicate);
+    $this->segmentsRepository->flush();
+    
+    $subscriberSegments = $this->subscriberSegmentRepository->findBy(['segment' => $segmentEntity]);
+    foreach ($subscriberSegments as $subscriberSegment) {
+      $subscriber = $subscriberSegment->getSubscriber();
```
pavel-mailpoet, line 46: Again I am not sure if this is possible in Doctrine, but this makes a separate query for each subscriber, right? Can we prefetch all the subscribers first so that we save the database? I am not sure if that is a good idea.

lib/Segments/SegmentSaveController.php
lysyjan: This is a perfect idea. I thought about it, but I forgot then.

lib/Segments/SegmentSaveController.php
lysyjan: Good point. We can do it. It should be possible by query instead of working with entities. When we use entities, there will be one query per entity, but it will be only one if we use a query.

pavel-mailpoet: Can't we use some eager loading or something? I am not sure how that works though.

pavel-mailpoet (approved):

veljkho (approved):

PR #3296, opened 2021-02-01
Remove PHP security check, it doesn't work any more [MAILPOET-3327]

pavel-mailpoet:
[MAILPOET-3327]

[MAILPOET-3327]: https://mailpoet.atlassian.net/browse/MAILPOET-3327


lysyjan (approved):
Thank you! 👍🏻

veljkho (approved):

PR #3307, opened 2021-02-04
Stop using onload callback after loading reCaptcha v2 [MAILPOET-3390]

lysyjan:
[MAILPOET-3390]

[MAILPOET-3390]: https://mailpoet.atlassian.net/browse/MAILPOET-3390

### DIFF
--- a/assets/js/src/public.jsx
+++ b/assets/js/src/public.jsx
@@ -6,19 +6,27 @@ import 'parsleyjs';
 const exitIntentEvent = 'mouseleave.mailpoet.form-exit-intent';
 
 jQuery(($) => {
-  window.reCaptchaCallback = function reCaptchaCallback() {
-    $('.mailpoet_recaptcha').each((index, element) => {
-      const recaptcha = $(element);
-      const sitekey = recaptcha.attr('data-sitekey');
-      const container = recaptcha.find('> .mailpoet_recaptcha_container').get(0);
-      const field = recaptcha.find('> .mailpoet_recaptcha_field');
-      let widgetId;
-      if (sitekey) {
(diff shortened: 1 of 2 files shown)


assets/js/src/public.jsx
```diff
@@ -6,19 +6,27 @@ import 'parsleyjs';
...
-      const sitekey = recaptcha.attr('data-sitekey');
-      const container = recaptcha.find('> .mailpoet_recaptcha_container').get(0);
-      const field = recaptcha.find('> .mailpoet_recaptcha_field');
-      let widgetId;
-      if (sitekey) {
-        widgetId = window.grecaptcha.render(container, { sitekey, size: 'compact' });
-        field.val(widgetId);
+  function renderCaptcha(element, iteration) {
+    if (!window.recaptcha || !window.grecaptcha.ready) {
+      if (iteration < 20) {
```
pavel-mailpoet, line 11: Could you please explain to me why is 20 here? Why not some bigger or smaller number? why is 20 the right amount? does it need a comment?

pavel-mailpoet (commented):
Sorry @lysyjan but I need an explanation for this pull request. I don't understand what is going on here and why do we need this change and why it is fixing the original issue. Could you please leave a comment explaining it?

lysyjan:
I apologize. I could write some description.
The problem is that sometimes callback reCaptchaCallback isn't loaded yet and in the console is the message: "reCAPTCHA couldn't find user-provided function: reCaptchaCallback".  I haven't found a solution for this except to manage to wait on load reCaptcha on the own side.
The numbers for waiting are a little bit random, but I tested if captcha loads with the slowest network throttling in Firefox.

pavel-mailpoet (approved):

veljkho (approved):

PR #3319, opened 2021-02-11
Fix form rendering issues [MAILPOET-3415]

costasovo:
[MAILPOET-3415]

[MAILPOET-3415]: https://mailpoet.atlassian.net/browse/MAILPOET-3415


lib/API/JSON/v1/Forms.php
```diff
@@ -188,6 +194,7 @@ public function saveEditor($data = []) {
     $formId = (isset($data['id']) ? (int)$data['id'] : 0);
     $name = (isset($data['name']) ? $data['name'] : WPFunctions::get()->__('New form', 'mailpoet'));
     $body = (isset($data['body']) ? $data['body'] : []);
+    $body = $this->dataSanitiser->sanitiseBody($body);
```
pavel-mailpoet, line 199: Although I appreciate your British spelling we should probably use `sanitize`

tests/integration/Form/ApiDataSanitiserTest.php
```diff
@@ -42,7 +42,7 @@ public function testItSanitizesBody() {
     $nestedHeading = $result[1]['body'][0];
     expect($paragraph['params']['content'])->equals('alert(1);Paragraph');
     expect($paragraph['params']['align'])->equals('left');
-    expect($nestedHeading['params']['content'])->equals('alert(1);Heading');
+    expect($nestedHeading['params']['content'])->equals('alert(2);Heading');
```
pavel-mailpoet, line 45: I guess this commit "fixup! Add service for fixing form API data" should be squashed, right?

assets/js/src/form_editor/blocks/blocks.jsx
```diff
@@ -46,6 +46,7 @@ export const initBlocks = () => {
   registerBlockType(firstName.name, firstName.settings);
   registerBlockType(lastName.name, lastName.settings);
   registerBlockType(segmentSelect.name, segmentSelect.settings);
+  html.settings.supports.inserter = select('mailpoet-form-editor').isUserAdministrator();
```
pavel-mailpoet, line 49: Do you think this is the best place for this line? In the middle of the block registering? It feels to me like an outliner here. Would it be better maybe in the block definition?

pavel-mailpoet (changes requested):

assets/js/src/form_editor/blocks/blocks.jsx
costasovo: It can't be done in the block definition since we need the store to be initialized. I'll move the above registering and add a comment.

pavel-mailpoet (approved):

veljkho (approved):

PR #3326, opened 2021-02-16
Export subscription consent evidence [MAILPOET-3393]

pavel-mailpoet:
[MAILPOET-3393]

[MAILPOET-3393]: https://mailpoet.atlassian.net/browse/MAILPOET-3393

### DIFF
--- a/assets/js/src/subscribers/importExport/export.ts
+++ b/assets/js/src/subscribers/importExport/export.ts
@@ -3,25 +3,44 @@ import jQuery from 'jquery';
 import MailPoet from 'mailpoet';
 import Handlebars from 'handlebars';
 
-jQuery(document).ready(function documentReady() {
-  var segmentsContainerElement;
-  var subscriberFieldsContainerElement;
-  var nextStepButton;
-  var renderSegmentsAndFields;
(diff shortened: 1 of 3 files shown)


lysyjan (dismissed):

veljkho (changes requested):
There are some issues in this PR:
- Unable to `Select all ...` `Deselect all ...`. Unclickable issue
- `Subscription time` and `Confirmation time` are always with literally the same values. Shouldn't they be different a bit? The time when subscribed and confirmed, that's how I understood.

lysyjan (approved):

veljkho (changes requested):
A new issue has arrived. Subscription time is not preselected.
The rest is verified.

Another potential issue is not clear to me. The values for `subscription time` & `confirmation time` [are always the same](https://d.pr/i/pWYoM3) no matter if user subscribes again to another list, or unsubscribe and then re-subscribe (to any list). The system counts it only once and remembers the first time user subscribed/confirmed. For all lists.

lysyjan (approved):

veljkho (approved):

PR #3278, opened 2021-01-20
Refactor segment lists fetching to Doctrine [MAILPOET-3164]

costasovo:
[MAILPOET-3164]

### Note for QA:
This change affects all select inputs where a user picks a segment or multiple segments on following pages:

- /wp-admin/admin.php?page=mailpoet-export
- /wp-admin/admin.php?page=mailpoet-import
- /wp-admin/admin.php?page=mailpoet-form-editor
- /wp-admin/admin.php?page=mailpoet-subscribers
- /wp-admin/admin.php?page=mailpoet-settigns
- /wp-admin/admin.php?page=mailpoet-newsletters (send page and automatic newsletters configuration)

[MAILPOET-3164]: https://mailpoet.atlassian.net/browse/MAILPOET-3164

PR #3328, opened 2021-02-16
Fix searching segments in subscribers import [MAILPOET-3423]

costasovo:
[MAILPOET-3423]

The bug was introduced [in this commit](https://github.com/mailpoet/mailpoet/pull/3278/commits/31e0be8feabe3e21beb2e36c545d92b3b3e06c10) that caused that `segment.text` property was not set. We don't use `segment.text` property in our code, but it is needed inside select2 for search.

[MAILPOET-3423]: https://mailpoet.atlassian.net/browse/MAILPOET-3423


wxa (approved):

wxa:
I'll merge it since @veljkho will be AFK, and this is a straightforward change, I've confirmed it works.

PR #3329, opened 2021-02-16
Drop MailPoet tables when a site is deleted in a multisite install [MAILPOET-3265]

rodrigoprimo:
This PR makes sure that MailPoet tables are dropped when a site is deleted in a multisite install. It uses the filter `wpmu_drop_tables` to add the list of MailPoet tables to the list of tables that WP drops whenever a site is deleted in a multisite install.

This will only work if MailPoet is network active. If it is not, MailPoet code is not executed inside the WP network admin panel, and thus our filter is not added to `wpmu_drop_tables`, and MP tables are not deleted.

**How to test this PR**
- On a multisite install, network activate MailPoet and create a few sites.
- Open the dashboard of the new sites so that MP tables are created.
- Delete one of the new sites and check that MP tables for the correct site were deleted.

Important: if there is a bug in this code, we have the potential to delete tables from other sites in the same multisite install or to not delete tables that should be deleted. I kept this in mind while creating this PR, but it is important that the reviewer considers this as well and thinks of edge cases that I might have missed to mitigate the chance of destroying data that shouldn't be destroyed.

[MAILPOET-3265]

[MAILPOET-3265]: https://mailpoet.atlassian.net/browse/MAILPOET-3265

### DIFF
--- a/lib/Config/Initializer.php
+++ b/lib/Config/Initializer.php
@@ -168,6 +168,12 @@ class Initializer {
       new DeferredAdminNotices,
       'printAndClean',
     ]);
+
+    WPFunctions::get()->addFilter('wpmu_drop_tables', [
+      $this,
+      'multisiteDropTables',
+    ]);
+
     $this->hooks->initEarlyHooks();
   }
 
@@ -343,6 +349,13 @@ class Initializer {
     $automaticEmails = new AutomaticEmails();
     $automaticEmails->init();
     $automaticEmails->getAutomaticEmails();
+  }
+
+  public function multisiteDropTables($tables) {
+    global $wpdb;
+    $tablePrefix = $wpdb->prefix . Env::$pluginPrefix;
+    $mailpoetTables = $wpdb->get_col("SHOW TABLES LIKE '$tablePrefix%'");
+    return array_merge($tables, $mailpoetTables);
   }
 
   private function setupWoocommerceTransactionalEmails() {


pavel-mailpoet:
@rodrigoprimo Just a note about the commits:
https://d.pr/i/ezBb2i
1) The first commit was WIP so why keeping it? You reverted some of the changes in the next commit so why didn't you delete the first commit in the first place? 
2) There is missing JIRA id in the second commit: [MAILPOET-3265]. It turned out a very useful practice for us. When debugging a problem it provides a context when trying to find why was some code added. Without it, we need to try to find the pull request and that takes the time.
3) It is better to amend the previous commit instead of creating a new one with the code quality fix. Again when someone in the future tries to understand the code they need to dig deeper and ignore this commit.

pavel-mailpoet (dismissed):

rodrigoprimo:
@pavel-mailpoet, thanks for the review and the comments regarding the commits. I share your concern regarding keeping the commit history clean.

As a reviewer, I find it helpful to know the history of the changes in a particular PR, and that is why I prefer to have all the commits in the PR. What I assumed would happen in this case is that, once the PR is approved, the commits would be squashed before being merged. This should address the issue of making it easier for us in the future when we are trying to understand the code and history of the changes. IMO, this workflow also has the benefit of making it easier to check changes that might be required after the first review as, for example, in this case here: https://github.com/mailpoet/mailpoet/pull/3321

Only now I saw that "Squash and merge" is disabled in this repository, and thus the workflow that I had in mind is not possible. Could you tell me a bit about the reasons for this decision to disable this option? Is the workflow to always combine commits in PRs documented somewhere that I'm forgetting about or missed? I'm asking as this might help future developers that will join the team.

Thanks again

pavel-mailpoet:
@rodrigoprimo 

First of all, I don't think the last commit when the code QA is fixed is helpful to the reviewer so that can be amended. 

According to "Squash and merge" we never considered that. We only used the option "Create a merge commit" and then we switched to "Rebase and merge" to keep the history cleaner. I am not sure about squashing all the commits. That would create very big commits. I am not sure if that would be useful. 

> As a reviewer, I find it helpful to know the history of the changes in a particular PR
I am not sure about that. I don't see why. As a reviewer, I am not interested in the approaches that were tried. I am only interested in the final way.

rodrigoprimo:
I went ahead and squashed the three commits. I'm happy to adapt to the workflow that you folks use.

> First of all, I don't think the last commit when the code QA is fixed is helpful to the reviewer so that can be amended.

Yeah, I agree. I left it initially as I thought it could be squashed when merging the PR (I missed that this option is disabled in this repository).

> According to "Squash and merge" we never considered that. We only used the option "Create a merge commit" and then we switched to "Rebase and merge" to keep the history cleaner. I am not sure about squashing all the commits. That would create very big commits. I am not sure if that would be useful.

I can see this being useful for a PR like this one or https://github.com/mailpoet/mailpoet/pull/3321. But for larger PRs, it might not be ideal (and not used) as it could result in huge commits like you said. It is not clear to me what I should do going forward in a case like #3321, where the reviewer found an issue and asked for a change. I don't think I should squash the changes in the original commits, as this would make it harder for the reviewer to check the changes that were implemented. Any suggestion? What do you folks normally do in a situation like this?

> I am not sure about that. I don't see why. As a reviewer, I am not interested in the approaches that were tried. I am only interested in the final way.

We disagree here, but that is ok.

pavel-mailpoet:
> I went ahead and squashed the three commits

Thank you

> It is not clear to me what I should do going forward in a case like #3321, where the reviewer found an issue and asked for a change.

I understand the confusion, sorry about that. So this is how we've been doing it:
- in the first version of PR we can amend and change any commits as we want so we try to make them nice
- what to do if a reviewer asks for changes? - it depends on the change and on the pull request. If it is a small change in a small pull request we may change the original commit and tell that to the person doing code review. Or sometimes the reviewer directly asks to amend the commit. If it is multiple changes in a big pull request we tend to create new commits. But not always. So it is different for each pull request, if we are in doubts the people involved just discuss and choose the best approach.

pavel-mailpoet (approved):

rodrigoprimo:
> what to do if a reviewer asks for changes? - it depends on the change and on the pull request. If it is a small change in a small pull request we may change the original commit and tell that to the person doing code review. Or sometimes the reviewer directly asks to amend the commit. If it is multiple changes in a big pull request we tend to create new commits. But not always. So it is different for each pull request, if we are in doubts the people involved just discuss and choose the best approach.

Sounds good to me. Thanks for the clarification.

veljkho (approved):

PR #3341, opened 2021-02-18
Fix missing close button for success notice [MAILPOET-3435]

lysyjan:
[MAILPOET-3435]

[MAILPOET-3435]: https://mailpoet.atlassian.net/browse/MAILPOET-3435


wxa (approved):

wxa (approved):
One remark: the fix is actually for all notice types, not only for the success one.

veljkho (approved):

PR #3344, opened 2021-02-19
Update SECURITY.md to point security researchers to HackerOne [MAILPOET-3410]

rodrigoprimo:
This PR updates the SECURITY.md file to point security researchers to HackerOne. I copied the SECURITY.md file from WooCommerce and made some minor changes to reflect MailPoet specificities.

[MAILPOET-3410]

[MAILPOET-3410]: https://mailpoet.atlassian.net/browse/MAILPOET-3410

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -1,3 +1,37 @@
-## Security contact information
+# Security Policy
 
-To report a security vulnerability, please use email address [security@mailpoet.com](mailto:security@mailpoet.com).
+Full details of the Automattic Security Policy can be found on [automattic.com/security](https://automattic.com/security/).
+
+## Supported Versions
+
+Generally, *only the latest version of MailPoet has continued support*. If a critical vulnerability is found in the current version of MailPoet, we may opt to backport any patches to previous versions.
+
+## Reporting a Vulnerability
+
+[MailPoet](https://wordpress.org/plugins/mailpoet) is an open-source plugin for WordPress. Our HackerOne program covers the plugin software, as well as a variety of related projects and infrastructure.
+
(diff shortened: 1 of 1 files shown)


.github/SECURITY.md
```diff
@@ -1,3 +1,37 @@
-## Security contact information
+# Security Policy
 
-To report a security vulnerability, please use email address [security@mailpoet.com](mailto:security@mailpoet.com).
+Full details of the Automattic Security Policy can be found on [automattic.com/security](https://automattic.com/security/).
+
+## Supported Versions
+
+Generally, *only the latest version of MailPoet has continued support*. If a critical vulnerability is found in the current version of MailPoet, we may opt to backport any patches to previous versions.
```
wxa, line 7: Should we specify that the latest *major* version is supported (MailPoet 3 in this case - no need to mention the specific version)? Otherwise the statement about backporting doesn't make much sense: for the version 3 we only release updates with fixes, backporting may happen for the version 2.

rodrigoprimo: This part I just copied from the WooCommerce version of this file, and it is the same for Jetpack. I don't know if we need to change it, but I don't feel strongly about it.

I slightly prefer the original version of the phrase because my understanding is that it is more specific in saying that only the latest version is supported. Assuming that 3.60.1 is the latest version. If we say that only the latest major version is supported, do we mean that we also support 3.59.0 and other versions in the 3.x branch as well? IMO, saying __only the latest version of MailPoet has continued support__ without mentioning __major__ is better to make it clear that only 3.60.1 is supported by default but that, depending on the circumstances, we might opt to backport patches to previous versions (and that could be 2.X.Y or 3.59.X, even though, the latter is unlikely). What do you think?

wxa: Your reasoning sounds rational to me, let's keep it like it is.

wxa (approved):

PR #3345, opened 2021-02-19
Fix link to buy the agency license of MailPoet Premium [MAILPOET-3384]

rodrigoprimo:
[MAILPOET-3384]

[MAILPOET-3384]: https://mailpoet.atlassian.net/browse/MAILPOET-3384

### DIFF
--- a/views/premium.html
+++ b/views/premium.html
@@ -246,10 +246,10 @@
       </ul>
       <a
         target="_blank"
-        href="<%= add_referral_id("https://account.mailpoet.com/?s=" ~ subscriber_count ~ "&email=" ~ current_wp_user.user_email ~ '&utm_source=plugin&utm_medium=premium&utm_campaign=purchase') | escape('html_attr') %>"
+        href="https://www.mailpoet.com/premium/agency"
         class="mailpoet-button"
       >
-        <%= _x('Buy Now (currently 33% discount)', 'This text resides in the Premium page: /wp-admin/admin.php?page=mailpoet-premium') %>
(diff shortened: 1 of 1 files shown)


lysyjan (dismissed):

veljkho (changes requested):
Issue with the improper link is fixed. 
But the issue with incorrect button label is still present. We should remove `(currently 33% discount)` from the button as well.

rodrigoprimo:
@veljkho, please correct me if I'm mistaken, but I didn't look into changing the label of the button because what I understood from @NeosinneR's comment on the Jira issue is that he wanted us to fix just the link.

NeosinneR:
@rodrigoprimo a slight misunderstanding here. The label should also be changed (to remove the 33% discount). The Jira ticket description previously contained two issues (one of which was removed from the description) so my comment became obsolete.

rodrigoprimo:
@NeosinneR, thanks for the clarification. I wrongly assumed that the problem that should be ignored was updating the label as I missed that the original description of the issue was edited.

I just updated the label of the button as well. Since this is a fairly small change, I opted to amend the original commit and force push it to the remote branch to keep the git history clean.

In a case like this one where QA asks for changes in a PR, should I request a new review from @veljkho directly, or should I first request a new review from the developer that reviewed the PR?

NeosinneR:
@rodrigoprimo if it's a small change like this, you can assign @veljkho right away

veljkho (approved):

PR #3350, opened 2021-02-23
Use Doctrine for export subscribers [MAILPOET-3376]

lysyjan:
[MAILPOET-3376]

[MAILPOET-3376]: https://mailpoet.atlassian.net/browse/MAILPOET-3376

### Note for QA
This is a refactor of a backend functionality for subscribers export. It should work the same as before the change + it fixes [MAILPOET-3454]. Please test export for various list types (default list, WooCommerce, WordPress, dynamic segments).

[MAILPOET-3454]: https://mailpoet.atlassian.net/browse/MAILPOET-3454


costasovo (commented):
Nice job! I played with it a bit at it seems to work as expected.

I would suggest just adding a small comment into the code (see my comment) and I think we could also remove some files that are no longer needed after this refactor. I found these that I think we can delete:

- https://github.com/mailpoet/mailpoet/blob/master/lib/Subscribers/ImportExport/Export/DefaultSubscribersGetter.php
- https://github.com/mailpoet/mailpoet/blob/master/lib/Subscribers/ImportExport/Export/DynamicSubscribersGetter.php
- https://github.com/mailpoet/mailpoet/blob/master/lib/Subscribers/ImportExport/Export/SubscribersGetter.php

and probably in Subcriber Model we could mark Subscriber::filterWithCustomFieldsForExport as deprecated ([add PHP doc and trigger an error](https://github.com/mailpoet/mailpoet/blob/1d7d56af52fc5e0cc70bc271e3385f0ea883a1a6/lib/Models/Segment.php#L85-L90))

There might be some more...

costasovo (commented):
👍

costasovo (approved):

veljkho (changes requested):
I noticed that exporting the "Not in a List" list does not count all subscribers without a list. The problem is that when a user **manually removes all the lists via Manage Subscription**, then when you export the list "Not in a List" you will not see that user on the list. Alternatively, if you simply as admin **manually remove all the lists from some subscriber** and save it, then you can reproduce issue as well.

Otherwise, it would work if you select subscribers on the Subscribers page and just press Remove from all lists, they would be on the list when exporting.

Maybe we had this problem before refactoring, unable to test due to bug on production (fixed in this branch).. but just let me know if I need to open a new Jira for this issue.

lysyjan:
I have created a new [issue](https://mailpoet.atlassian.net/browse/MAILPOET-3462) for the trouble with _Not in a List_.
This problem was here before refactoring.

veljkho (approved):

PR #3351, opened 2021-02-23
Update documentation of some Robo commands [MAILPOET-3444]

rodrigoprimo:
This PR updates the documentation of some Robo commands that were using `:` instead of `-` to separete words for commands with more than two words resulting in command not found errors. Simply replacing the aforementioned character fixes the commands.

For two release related commands, it was also necessary to add the `release` prefix.

[MAILPOET-3444]

[MAILPOET-3444]: https://mailpoet.atlassian.net/browse/MAILPOET-3444


lysyjan (approved):
Good catch! 👍🏻

PR #3354, opened 2021-02-25
Fix join parameters in SubscribersRepository [MAILPOET-3445]

lysyjan:
[MAILPOET-3445]

[MAILPOET-3445]: https://mailpoet.atlassian.net/browse/MAILPOET-3445


wxa (approved):
Nice job! The fix works for one of the affected customers, so I assume it's good.

PR #3357, opened 2021-02-26
Truncate first and last names before adding to wp_mailpoet_subscribers [MAILPOET-3246]

rodrigoprimo:
MailPoet syncs users from wp_users to wp_mailpoet_subscribers. The problem is that WP stores first and last names in a longtext field, and MP uses a varchar(255) field. This was causing a fatal error when synchronizing names over 255 characters. This commit fixes this problem by using SUBSTRING() to make sure that the 255 characters limit is enforced when adding values to the columns first_name and last_name of the wp_mailpoet_subscribers table. This should get rid of the fatal error, and it shouldn't be a problem to most users as it is unlikely that a real user has a first or last name that is longer than 255 characters.

@NeosinneR, you mentioned in the issue description performance concerns with SUBSTRING(). As far as I can tell, this should not impact the query in any meaningful way but I didn't check closely. Do you have any specific concerns?

[MAILPOET-3246]

[MAILPOET-3246]: https://mailpoet.atlassian.net/browse/MAILPOET-3246

### DIFF
--- a/lib/Segments/WP.php
+++ b/lib/Segments/WP.php
@@ -229,7 +229,7 @@ class WP {
     Subscriber::rawExecute(sprintf('
       UPDATE %1$s
         JOIN %2$s as wpum ON %1$s.wp_user_id = wpum.user_id AND wpum.meta_key = "first_name"
-      SET %1$s.first_name = wpum.meta_value
+      SET %1$s.first_name = SUBSTRING(wpum.meta_value, 1, 255)
         WHERE %1$s.first_name = ""
         AND %1$s.wp_user_id IS NOT NULL
         AND wpum.meta_value IS NOT NULL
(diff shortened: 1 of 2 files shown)


NeosinneR:
@rodrigoprimo I didn't write these specs, so I'll let @wxa answer :-)

rodrigoprimo:
Ah, sorry. I saw Alex asking you a question in a comment and assumed that you had written the specs. My bad.

costasovo (approved):
Thanks! 👍

veljkho (approved):

wxa:
@rodrigoprimo I don't have specific concerns about this, just general thinking that the more functions called in an already heavy query the slower it is.

PR #3359, opened 2021-03-03
Fix orphan sending tasks [MAILPOET-3379]

wxa:
I have decided to just log errors when saving a sending task because I think handling those when processing is enough (see below).

When fetching tasks, I think it's okay to delete them like we always did with tasks that [lack a newsletter](https://github.com/mailpoet/mailpoet/blob/master/lib/Cron/Workers/SendingQueue/SendingQueue.php#L96-L100).

### DIFF
--- a/lib/Tasks/Sending.php
+++ b/lib/Tasks/Sending.php
@@ -164,6 +164,14 @@ class Sending {
   public function save() {
     $this->task->save();
     $this->queue->save();
+    $errors = $this->getErrors();
+    if ($errors) {
+      $loggerFactory = LoggerFactory::getInstance();
+      $loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->addError(
+        'error saving sending task',
+        ['task_id' => $this->task->id, 'queue_id' => $this->queue->id, 'errors' => $errors]
+      );
+    }
     return $this;
   }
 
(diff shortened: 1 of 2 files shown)


lib/Tasks/Sending.php
```diff
@@ -97,11 +98,22 @@ public static function createManyFromTasks($tasks) {
...
     return $result;
   }
 
+  public static function handleInvalidTask(ScheduledTask $task) {
+    $loggerFactory = LoggerFactory::getInstance();
+    $loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->addError(
+      'invalid sending task found, deleting',
+      ['task_id' => $task->id]
+    );
+    $task->delete();
```
costasovo, line 114: I have a concern about deleting the task. When we delete it we also delete the task subscribers' data. The original idea was to add a new state `invalid`. If someone reports an issue with a stuck newsletter we would be able to recreate the sending queue and we may have more data for debugging. On the other hand, we can't easily match a sending task with a newsletter so maybe it is ok to delete it. Maybe keeping more data is a safer option. I don't have a strong opinion here. Just wanted to point this out. What do you think?

wxa: I chose to delete tasks without sending queue only because we always did the same with tasks without newsletters. If you think keeping them as `invalid` would help in debugging, let's do that. I've made the change.

costasovo (approved):
Thanks!

veljkho (approved):

PR #3360, opened 2021-03-03
Logs display page [MAILPOET-3135]

pavel-mailpoet:
[MAILPOET-3135]

[MAILPOET-3135]: https://mailpoet.atlassian.net/browse/MAILPOET-3135


lib/Config/Menu.php
```diff
@@ -371,6 +372,16 @@ public function setup() {
...
       [$this, 'experimentalFeatures']
     );
+
+    // display loggs page
+    $this->wp->addSubmenuPage(
+      true,
+      $this->setPageTitle('Logs'),
+      '',
+      AccessControl::PERMISSION_ACCESS_PLUGIN_ADMIN,
+      'logs',
```
costasovo, line 382: The url slug in spec is `mailpoet-logs`. I think it would be better to use the one from spec since `http://example.com/wp-admin/admin.php?page=logs` may cause some conflicts.

lib/Logging/LogRepository.php
```diff
@@ -12,4 +12,27 @@ class LogRepository extends Repository {
...
+        ->where('l.createdAt > :dateFrom')
+        ->setParameter('dateFrom', $dateFrom->format('Y-m-d H:i:s'));
+    }
+    if ($dateTo instanceof \DateTimeInterface) {
+      $query
+        ->andWhere('l.createdAt < :dateTo')
+        ->setParameter('dateTo', $dateTo->format('Y-m-d H:i:s'));
+    }
+
+    return $query->getQuery()->getResult();
```
costasovo, line 46: We don't have any other limit than dates here. I think that there might be cases that there will be many logs (e.g. some blocked cron that generate a bunch of logs every 30 seconds). It was not in the specification but we could add limit and offset inputs + a number of rows displayed to the filter area so that we are able somehow "paginate". What do you think? cc @NeosinneR  My concern is that in some cases we won't be able to render the default recent 7 days.

costasovo (commented):
Looks good! I've 2 suggestions to consider.

lib/Logging/LogRepository.php
pavel-mailpoet: The specification says: 
> No pagination, show all available results

NeosinneR: I wanted to skip pagination to keep the ticket as simple as possible. What do you think would be the reason to fail to render the page? The table is very simple, so I think even a few megabytes of logs should be renderable (although it might take some time on slower connections).

costasovo: The table is rendered by React and I'm not sure about performance with e.g. 10K of rows. On the other hand, this is not for users so in some probably rare cases we can live with a slow page and we can always add some paginations later or use some "[virtual table](https://bvaughn.github.io/react-virtualized/#/components/Table)" later.

pavel-mailpoet: I am not sure if react will have issues, but the whole DB query, processing the log, and passing it to the template may time out.

costasovo: Is it difficult to add those offset and limit inputs (with defaults 0, 100) and apply them? It is not super UX for pagination but could be enough for digging in logs and could protect us from potential issues.

pavel-mailpoet: I guess not that much, I would just pass them with the other existing inputs.

costasovo (approved):

veljkho (changes requested):
Unfortunately, I couldn't see any single log when filtering any dates, but I could see logs in the database table mailpoet_log. Probably something's wrong with displaying them on the page.

As a side note, I will mention that if I set -1 and let's say 5 for the "Offset/Limit" filter, I can experience 500 error if I try to filter it.. so the minus value in the Offset field triggers the issue. Not sure if that issue is worth fixing..should we block users to experience that.. but just mentioning it.

veljkho (changes requested):
After the latest investigation, here is the problem description:
- Logs are present, but if I filter I don't see them. It is because we are showing the logs not including the boundary dates. And because all the logs were created on March 11th I dont see them.

lib/Logging/LogRepository.php
```diff
@@ -34,13 +34,13 @@ public function getLogs(
 
     if ($dateFrom instanceof \DateTimeInterface) {
       $query
-        ->andWhere('l.createdAt > :dateFrom')
-        ->setParameter('dateFrom', $dateFrom->format('Y-m-d H:i:s'));
+        ->andWhere('l.createdAt >= :dateFrom')
+        ->setParameter('dateFrom', $dateFrom->format('Y-m-d H:i:s 00:00:00'));
```
costasovo, line 38: I think this is a typo `Y-m-d H:i:s 00:00:00` should probably be `Y-m-d 00:00:00`.

costasovo (changes requested):

costasovo (approved):

veljkho (approved):

PR #3363, opened 2021-03-04
Fix count returned from shortcode mailpoet_subscribers_count [MAILPOET-3424]

lysyjan:
[MAILPOET-3424]

[MAILPOET-3424]: https://mailpoet.atlassian.net/browse/MAILPOET-3424

### DIFF
--- a/lib/Config/Shortcodes.php
+++ b/lib/Config/Shortcodes.php
@@ -2,11 +2,12 @@
 
 namespace MailPoet\Config;
 
+use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Form\Widget;
 use MailPoet\Models\Newsletter;
 use MailPoet\Models\Subscriber;
-use MailPoet\Models\SubscriberSegment;
 use MailPoet\Newsletter\Url as NewsletterUrl;
+use MailPoet\Segments\SegmentSubscribersRepository;
 use MailPoet\Subscription\Pages;
 use MailPoet\WP\Functions as WPFunctions;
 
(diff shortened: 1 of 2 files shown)


lib/Config/Shortcodes.php
```diff
@@ -74,12 +83,11 @@ public function getSubscribersCount($params) {
...
     } else {
-      return $this->wp->numberFormatI18n(
-        SubscriberSegment::whereIn('segment_id', $segmentIds)
-          ->select('subscriber_id')->distinct()
-          ->filter('subscribed')
-          ->findResultSet()->count()
-      );
+      $count = 0;
+      foreach ($segmentIds as $segmentId) {
+        $count += $this->segmentSubscribersRepository->getSubscribersCount($segmentId, SubscriberEntity::STATUS_SUBSCRIBED);
```
costasovo, line 88: I'm afraid that this will not give correct results. When a subscriber is subscribed to multiple segments it will count them multiple times. On the other hand, the original code doesn't handle dynamic segments. So maybe this could be an acceptable trade-off. I guess this is something we could discuss with @NeosinneR .

NeosinneR: We should not count subscribers multiple times, because that could result in more subscribers than there really is. So if a subscriber is subscribed to multiple lists, he should be counted only once. I think it's fine not counting dynamic segments, as there is a high chance, that if someone is subscribed to a dynamic segment, they are also subscribed to the regular list.

costasovo (approved):

veljkho (approved):

PR #3365, opened 2021-03-05
Dynamic Segments API to Doctrine [MAILPOET-3177]

costasovo:
[MAILPOET-3177]

[MAILPOET-3177]: https://mailpoet.atlassian.net/browse/MAILPOET-3177


lib/Segments/DynamicSegments/FilterDataMapper.php
```diff
@@ -0,0 +1,76 @@
...
+    if (!isset($data['segmentType'])) {
+      throw new InvalidFilterException('Segment type is not set', InvalidFilterException::MISSING_TYPE);
+    }
+    return $data['segmentType'];
+  }
+
+  /**
+   * @throws InvalidFilterException
+   */
+  private function createEmail(array $data): DynamicSegmentFilterData {
```
lysyjan, line 41: You verify the name of the action in method `createWooCommerce`, but this method can pass another than the supported action. I would prefer to check that the action value is valid. What do you think?
One more thing: If I remember right so then is `link_id` required for click and not click actions.

lysyjan (commented):
A lot of work! Great job.
I have only a little note.

lib/Segments/DynamicSegments/FilterDataMapper.php
costasovo: It makes sense to validation for all those cases you mentioned 👍 I copied the validation from the old mapper.

costasovo: Ha! Actually, it seems that link_id is not required. Without the specific link, the segments check whether users clicked on some or didn't click at any link. So I think it will be fine to add just supported actions validation.

costasovo: I've added validation of supported actions for EmailAction filter. As for the wooCommerce filter, there is an [else branch that throws an error in case the action is not known](https://github.com/mailpoet/mailpoet/pull/3365/commits/a457c725656ecfd4efebac3f3a8e6c02eb72ce07#diff-4b71db0f185b4e8e279f557aa4723d13d273b56640c750f2cf44c074bc63eb33R71-R73).

lysyjan: I'm sorry. I thought that the link is required for those actions.

lysyjan (dismissed):

veljkho (changes requested):
Hey, I see critical error while accessing the following pages in the plugin:
- Emails
- Subscribers
- Settings

And this error for accessing Segments tab in Lists:
`Argument 2 passed to MailPoet\Segments\DynamicSegments\FilterHandler::apply() must be an instance of MailPoet\Entities\DynamicSegmentFilterData, array given, called in /var/www/html/wp-content/plugins/mailpoet/lib/Segments/SegmentSubscribersRepository.php on line 139`

costasovo:
@lysyjan I've found some issues after rebasing on master and addressed them in a [new commit](https://github.com/mailpoet/mailpoet/pull/3365/commits/5661467b7d20530b5b030f068149b8dbb69567f5).
Note that the fix is not related to the error that @veljkho reported. I think that the error was caused by problems in the synchronization of files in docker volumes after switching branches.

lysyjan (approved):

veljkho (approved):

PR #3368, opened 2021-03-06
Preserve new key when saving settings [MAILPOET-3447]

rodrigoprimo:
This PR fixes a bug in the code where a user could override their key if, after validating a new key, they went to a different settings tab and clicked "Save settings". When reloading the page, they would see the old key instead of the new one.

This was happening because the function verifyPremiumKey() was updating the key in the backend but was not updating the settings object in the frontend. So when the user switched to another settings tab and clicked on the "Save settings" button, the function saveSettings() was saving to the backend a settings object containing the old key instead of the new one.

To fix this problem, this PR updates the settings object with the new premium key whenever a new key is validated.

Please note that I'm fairly new to modern JS and React development so there is a chance that the way that I'm fixing this is not the best way or that I'm missing something.

[MAILPOET-3447]

[MAILPOET-3447]: https://mailpoet.atlassian.net/browse/MAILPOET-3447

### DIFF
--- a/assets/js/src/settings/store/actions/mss_and_premium.ts
+++ b/assets/js/src/settings/store/actions/mss_and_premium.ts
@@ -6,7 +6,7 @@ import { STORE_NAME } from 'settings/store';
 import {
   Action, KeyActivationState, MssStatus, PremiumStatus,
 } from 'settings/store/types';
-import { setSettings } from './settings';
+import { setSettings, setSetting } from './settings';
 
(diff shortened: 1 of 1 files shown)


costasovo (approved):
Good find! As far as I remember there used to be an automatic reload on the page with key validation. It seems that we removed it when we finished refactoring settings and that's probably how the issue surfaced.
I think that this is a proper fix for the issue. 👍

veljkho (approved):

PR #3373, opened 2021-03-09
Duplicate the name property as the text for Select2 search [MAILPOET-3458]

lysyjan:
[MAILPOET-3458]

I duplicate a property `name` as `text` because Select2 needs properties `id` and `text`.
Maybe it would be a better solution to refactor our code for using the property `text` instead of `name`. When I have tried it then didn't work correctly import. I don't want to spend more time with it but if you think that it's worth it. I don't think because we want to remove Select2 in the future.

[MAILPOET-3458]: https://mailpoet.atlassian.net/browse/MAILPOET-3458

### DIFF
--- a/assets/js/src/subscribers/importExport/export.ts
+++ b/assets/js/src/subscribers/importExport/export.ts
@@ -14,9 +14,11 @@ interface ExportWindow extends Window {
   }>;
   subscriberFieldsSelect2: Array<{
     name: string;
+    text: string; // Required select2 property
     children: Array<{
       id: string;
       name: string;
+      text: string; // Required select2 property
       type: string | null;
       custom: boolean;
       params: object;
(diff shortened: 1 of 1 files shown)


lib/Subscribers/ImportExport/ImportExportFactory.php
```diff
@@ -100,20 +102,24 @@ public function formatFieldsForSelect2(
         [
           'id' => 'ignore',
           'name' => __('Ignore field...', 'mailpoet'),
+          'text' => __('Ignore field...', 'mailpoet'),
```
pavel-mailpoet, line 105: Do we have to duplicate the fields on the PHP level? Could we only make the duplication on the javascript level and copy the values in there? So that we don't have to keep the duplication here and keep sending them?

lysyjan: The duplication on the JavaScript level sounds good. I will try it.

pavel-mailpoet (approved):

veljkho (approved):

PR #3374, opened 2021-03-09
Fix Newsletter's Create button is without loading effect when clicked [MAILPOET-3455]

pavel-mailpoet:
[MAILPOET-3455]

[MAILPOET-3455]: https://mailpoet.atlassian.net/browse/MAILPOET-3455


lysyjan (approved):
Good job!

veljkho (approved):

PR #3379, opened 2021-03-10
Refactor Subscriber::getCurrentWPUser to Doctrine [MAILPOET-3268]

lysyjan:
[MAILPOET-3268]

[MAILPOET-3268]: https://mailpoet.atlassian.net/browse/MAILPOET-3268

### DIFF
--- a/lib/AdminPages/Pages/NewsletterEditor.php
+++ b/lib/AdminPages/Pages/NewsletterEditor.php
@@ -119,7 +119,7 @@ class NewsletterEditor {
     return [
       'id' => $subscriber->getId(),
       'wp_user_id' => $subscriber->getWpUserId(),
-      'is_woocommerce_user' => (string)$subscriber->isWPUser(), // BC compatibility
+      'is_woocommerce_user' => (string)$subscriber->getIsWoocommerceUser(), // BC compatibility
       'first_name' => $subscriber->getFirstName(),
       'last_name' => $subscriber->getLastName(),
       'email' => $subscriber->getEmail(),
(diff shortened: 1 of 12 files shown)


lib/AdminPages/Pages/NewsletterEditor.php
```diff
@@ -106,4 +114,28 @@ public function render() {
 
     $this->pageRenderer->displayPage('newsletter/editor.html', $data);
   }
+
+  private function formatSubscriber(SubscriberEntity $subscriber): array {
+    return [
+      'id' => $subscriber->getId(),
+      'wp_user_id' => $subscriber->getWpUserId(),
+      'is_woocommerce_user' => (string)$subscriber->isWPUser(), // BC compatibility
```
wxa, line 122: Why didn't you use the `getIsWoocommerceUser()` method? And why is this marked as BC compatibility?

tests/integration/WooCommerce/SubscriptionTest.php
```diff
@@ -49,8 +50,9 @@ public function _before() {
         },
       ]
     );
+    $subscribersRepository = $this->diContainer->get(SubscribersRepository::class);
```
wxa, line 53: In this file there are a couple usages of `Subscriber::getCurrentWPUser()`, is it possible to refactor them? Then this method will be unused and we can deprecate it (in case it's used by 3rd parties) and later remove.

lib/AdminPages/Pages/NewsletterEditor.php
lysyjan: The usage of the wrong getter is a mistake. Thanks 👍🏻 
And the comment is as a note because the original type from the getter is boolean, and we want to type it as a string because it was used in the past.

tests/integration/WooCommerce/SubscriptionTest.php
lysyjan: I don't know how I could forget. Thanks

wxa (approved):
Thank you! I have added an error message when calling the deprecated function.

veljkho (approved):

PR #3386, opened 2021-03-12
Show number of subscribers when editing and creating segments [MAILPOET-3219]

lysyjan:
[MAILPOET-3219]

[MAILPOET-3219]: https://mailpoet.atlassian.net/browse/MAILPOET-3219

### DIFF
--- a/assets/js/src/ajax.js
+++ b/assets/js/src/ajax.js
@@ -2,17 +2,21 @@ import MailPoet from 'mailpoet';
 import jQuery from 'jquery';
 import _ from 'underscore';
 
+function buildErrorResponse(message) {
+  return {
+    errors: [
+      {
+        message,
+      },
+    ],
+  };
+}
+
 function requestFailed(errorMessage, xhr) {
   if (xhr.responseJSON) {
     return xhr.responseJSON;
   }
-  return {
-    errors: [
-      {
-        message: errorMessage.replace('%d', xhr.status),
-      },
-    ],
-  };
+  return buildErrorResponse(errorMessage.replace('%d', xhr.status));
 }
 
 // Renew MailPoet nonce via heartbeats to keep auth
@@ -55,8 +59,8 @@ MailPoet.Ajax = {
     }
 
     // set default timeout
-    if (this.options.token === null) {
-      this.options.timeout = 0;
+    if (this.options.timeout === null) {
+      this.options.timeout = 2000;
     }
   },
   getParams: function getParams() {
@@ -90,14 +94,20 @@ MailPoet.Ajax = {
 
     // ajax request
     const deferred = jQuery.Deferred();
+    const timeout = Math.ceil(this.options.timeout / 1000); // convert milliseconds to seconds
     jQuery[method]({
       url: this.options.url,
       data: params,
       success: null,
       dataType: 'json',
       timeout: this.options.timeout,
-    }).then(deferred.resolve, (failedXhr) => {
-      const errorData = requestFailed(MailPoet.I18n.t('ajaxFailedErrorMessage'), failedXhr);
+    }).then(deferred.resolve, (failedXhr, textStatus) => {
+      let errorData;
+      if (textStatus === 'timeout') {
+        errorData = buildErrorResponse(MailPoet.I18n.t('ajaxTimeoutErrorMessage').replace('%d', timeout.toString()));
+      } else {
+        errorData = requestFailed(MailPoet.I18n.t('ajaxFailedErrorMessage'), failedXhr);
+      }
       deferred.reject(errorData);
     });
 
--- a/assets/js/src/ajax.js
+++ b/assets/js/src/ajax.js
@@ -60,7 +60,7 @@ MailPoet.Ajax = {
 
     // set default timeout
     if (this.options.timeout === null) {
-      this.options.timeout = 2000;
+      this.options.timeout = 0;
     }
   },
   getParams: function getParams() {
--- a/assets/js/src/ajax.js
+++ b/assets/js/src/ajax.js
@@ -53,6 +53,11 @@ MailPoet.Ajax = {
     if (this.options.token === null) {
       this.options.token = window.mailpoet_token;
     }
+
+    // set default timeout
+    if (this.options.token === null) {
+      this.options.timeout = 0;
+    }
   },
   getParams: function getParams() {
     return {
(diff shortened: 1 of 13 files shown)


assets/js/src/segments/subscribers_counter.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import MailPoet from 'mailpoet';
+
+type SubscriberCount = {
```
pavel-mailpoet, line 4: In the calculator, you were using a different syntax for the type. You used an interface and number | undefined. I think we should start unifying it. So instead of using {} here, I would recommend using `interface`. and instead of  using number | undefined I would use `?` syntax

assets/js/src/segments/subscribers_calculator.ts
```diff
@@ -0,0 +1,103 @@
...
+  wooCommerceActions: string[] = ['purchasedCategory', 'purchasedProduct'];
+
+  // Names of keys from interface FormItem
+  allowedItemKeys: string[] = ['wordpressRole', 'segmentType', 'action', 'newsletter_id', 'category_id', 'product_id'];
+
+  lastFormItem: FormItem;
+
+  lastResult: Result;
+
+  validateInputData(formItem: FormItem): boolean {
```
pavel-mailpoet, line 36: I am not sure about having this here. The idea, in the beginning, was to keep filter specific logic in the `dynamic_segments_filters` directory. The reason is to keep it in one place and when we are going to add new segments we will have only a couple of files to modify. If we keep adding logic like this, adding new segments will be more complicated. 

The same applies to the actions above.

assets/js/src/segments/subscribers_calculator.ts
```diff
@@ -0,0 +1,103 @@
+import MailPoet from 'mailpoet';
+import _ from 'lodash';
+
+export interface FormItem {
```
pavel-mailpoet, line 4: Is it possible to have a better type? The item shouldn't have category_id and `wordpressRole` at the same time, right?

Also, can we rename the items so that we don't combine underscore and camelcase?

lib/Segments/SegmentSubscribersRepository.php
```diff
@@ -85,6 +86,18 @@ public function getSubscribersCountBySegmentIds(array $segmentIds, string $statu
     return (int)$result;
   }
 
+  public function getDynamicSubscribersCount(DynamicSegmentFilterData $data): int {
```
pavel-mailpoet, line 89: We already have a function for calculating subscribers in dynamic segments. When we display segments we almost always display a count of subscribers next to them. Can't we reuse that? Why do we need a new one? 

This function calculates all subscribers and doesn't filter out deleted, doesn't check status. Is that intentional?

pavel-mailpoet (changes requested):

assets/js/src/segments/dynamic_segments_form.jsx
```diff
@@ -102,12 +115,48 @@ class DynamicSegmentForm extends React.Component {
     }
   }
 
+  getCount() {
+    const { item } = this.state;
+    switch (item.segmentType) {
```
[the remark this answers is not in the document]
lysyjan: Of course, there can be `if`. My first idea was a little bit different than the final solution.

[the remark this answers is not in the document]
pavel-mailpoet: I mean, we don't need anything here, right? The calculator does check the value anyway.

assets/js/src/segments/subscribers_calculator.ts
lysyjan: I understand, but how would you solve that we don't want to display the message about calculating and don't call API without all data?

lysyjan: I get it. There is a place for some refactoring maybe?

pavel-mailpoet: I would still have the validation, but move it in the filters. Each filter could validate its own data and return false/true if the data is valid. And having the main validation function that would call dynamically the filter validation based on the type.

assets/js/src/segments/subscribers_calculator.ts
```diff
@@ -0,0 +1,103 @@
...
+  errors: string[];
+}
+
+interface ApiVersionWindow extends Window {
+  mailpoet_api_version: string;
+}
+
+declare let window: ApiVersionWindow;
+
+class SubscribersCalculator {
```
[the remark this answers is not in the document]
lysyjan: I used class because I wanted to store a previous state. I don't want to call API when the name of the description is changed. I would be glad when you help me with a better solution.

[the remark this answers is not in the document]
pavel-mailpoet: you can do that even with a function:
```
let previous = undefined;

function SubscribersCalculator() {
  if (previous) return previous;
  ... 
}
```

assets/js/src/segments/subscribers_calculator.ts
lysyjan: > The item shouldn't have category_id and wordpressRole at the same time, right?

How can I do it? Could you help me, please?

> Also, can we rename the items so that we don't combine underscore and camelcase?

I used current input names from the form.

pavel-mailpoet: > I used current input names from the form.

I know and I think we should rename those input names to make them consistent.

pavel-mailpoet: Could we use something like this?
```
interface FormItem { 
  segmentType: string; 
} 

interface WordpressRoleFormItem extends FormItem { 
   wordpressRole?: string;
} 

interface WooCommerceFormItem extends FormItem { 
 ...
}

type AnyFormItem = WordpressRoleFormItem | WooCommerceFormItem

let lastFormItem: AnyFormItem;
```

lib/Segments/SegmentSubscribersRepository.php
lysyjan: I have overlooked this question. I can't reuse the same method because the dynamic segment isn't created or updated in the DB. I know only the filter parameters. I will try to change it a little bit.

assets/js/src/ajax.js
```diff
@@ -55,8 +59,8 @@ MailPoet.Ajax = {
     }
 
     // set default timeout
-    if (this.options.token === null) {
-      this.options.timeout = 0;
+    if (this.options.timeout === null) {
+      this.options.timeout = 2000;
```
pavel-mailpoet, line 63: Is this change intentional? Do you really want to limit all requests to 2 seconds?

lysyjan: This is definitely my mistake. I tested an error message and I didn't notice this change after that. Thank you

pavel-mailpoet (approved):

veljkho (changes requested):
Everything's good except clicked links in a segment.

For example:

I have 2 subscribers who clicked on different links in an email.
Then I create a `Segment` -> `Select` action `clicked` and I see **2 subscribers** (which is correct). Then I choose `links` and always see **[2 subscribers](https://d.pr/i/DNxjda)**? Is that expected? asking because I expect to see **1 subscriber**, because only 1 has clicked on that specific link chosen in the dropdown. When I save it I see **[1 subscriber](https://d.pr/i/bopecb)** on Segments page, so its a bit confusing for users to see different values in editor and who is really counted.

pavel-mailpoet (approved):

lysyjan:
@veljkho Nice catch 👍🏻

veljkho (dismissed):

veljkho (approved):

PR #3390, opened 2021-03-16
Connecting segments [MAILPOET-3212]

pavel-mailpoet:
[MAILPOET-3212]

[MAILPOET-3212]: https://mailpoet.atlassian.net/browse/MAILPOET-3212

### DIFF
--- a/lib/Segments/SegmentSubscribersRepository.php
+++ b/lib/Segments/SegmentSubscribersRepository.php
@@ -189,9 +189,8 @@ class SegmentSubscribersRepository {
     // For BC compatibility fetching an empty result
     if (count($filters) === 0) {
       return $queryBuilder->andWhere('0 = 1');
-    }
-    foreach ($filters as $filter) {
-      $queryBuilder = $this->filterHandler->apply($queryBuilder, $filter);
+    } elseif ($segment instanceof SegmentEntity) {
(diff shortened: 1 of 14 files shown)


lib/Segments/SegmentSubscribersRepository.php
```diff
@@ -189,9 +189,8 @@ private function filterSubscribersInDynamicSegment(
     // For BC compatibility fetching an empty result
     if (count($filters) === 0) {
       return $queryBuilder->andWhere('0 = 1');
-    }
-    foreach ($filters as $filter) {
-      $queryBuilder = $this->filterHandler->apply($queryBuilder, $filter);
+    } elseif ($segment instanceof SegmentEntity) {
```
costasovo, line 192: This change will cause that this method will not work correctly when $segmentEntity is not passed an it was expected to work. It is used here https://github.com/mailpoet/mailpoet/blob/7e0c718683e98d3c26fb14ef0995a6a23d868ba4/lib/Segments/SegmentSubscribersRepository.php#L80

And it results in broken functionality for showing subscriber count during creation of a dynamic segment.

costasovo (changes requested):
Good job! I found one issue that needs to be addressed.

costasovo (approved):
Good job!

veljkho (approved):

PR #3391, opened 2021-03-17
Disable scheduled tasks when pausing email [MAILPOET-3481]

pavel-mailpoet:
[MAILPOET-3481]

[MAILPOET-3481]: https://mailpoet.atlassian.net/browse/MAILPOET-3481

### DIFF
--- a/lib/Config/Populator.php
+++ b/lib/Config/Populator.php
@@ -10,6 +10,9 @@ use MailPoet\Cron\Workers\StatsNotifications\Worker;
 use MailPoet\Cron\Workers\SubscriberLinkTokens;
 use MailPoet\Cron\Workers\UnsubscribeTokens;
 use MailPoet\Entities\FormEntity;
+use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\UserFlagEntity;
 use MailPoet\Form\FormsRepository;
 use MailPoet\Mailer\MailerLog;
@@ -32,6 +35,7 @@ use MailPoet\Subscription\Captcha;
 use MailPoet\Util\Helpers;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class Populator {
   public $prefix;
@@ -50,6 +54,8 @@ class Populator {
   private $formsRepository;
   /** @var WP */
   private $wpSegment;
+  /** @var EntityManager */
+  private $entityManager;
 
   public function __construct(
     SettingsController $settings,
@@ -57,6 +63,7 @@ class Populator {
     Captcha $captcha,
     ReferralDetector $referralDetector,
     FormsRepository $formsRepository,
+    EntityManager $entityManager,
     WP $wpSegment
   ) {
     $this->settings = $settings;
@@ -146,6 +153,7 @@ class Populator {
       'FarmersMarket',
     ];
     $this->formsRepository = $formsRepository;
+    $this->entityManager = $entityManager;
   }
 
   public function up() {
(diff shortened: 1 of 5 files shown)


lib/Entities/NewsletterEntity.php
```diff
@@ -257,6 +258,25 @@ public function getStatus() {
...
+    $typesWithActivation = [self::TYPE_NOTIFICATION, self::TYPE_WELCOME, self::TYPE_AUTOMATIC];
+
+    if (($status === self::STATUS_DRAFT) && in_array($this->type, $typesWithActivation)) {
+      $newTaskStatus = ScheduledTaskEntity::STATUS_PAUSED;
+    }
+    if (($status === self::STATUS_ACTIVE) && in_array($this->type, $typesWithActivation)) {
+      $newTaskStatus = ScheduledTaskEntity::STATUS_SCHEDULED;
+    }
+
+    if (!$newTaskStatus) return;
```
costasovo, line 274: I think it would be nicer to fetch the unfinished queues after this check so that we don't fetch them when they are not needed.

lib/Entities/NewsletterEntity.php
```diff
@@ -257,6 +258,25 @@ public function getStatus() {
...
+    foreach ($queues as $queue) {
+      /** @var SendingQueueEntity $queue */
+      $task = $queue->getTask();
+      if ($task === null) continue;
+      $task->setStatus($newTaskStatus);
```
costasovo, line 283: I'm not sure whether we should always set the new status here. In the old models, we checked whether the task status was `paused` ([see here](https://github.com/mailpoet/mailpoet/blob/67824f9eb17abf4fa663c7964dedbdde301ba8b9/lib/Models/ScheduledTask.php#L75)) respectively `scheduled` ([see here](https://github.com/mailpoet/mailpoet/blob/67824f9eb17abf4fa663c7964dedbdde301ba8b9/lib/Models/ScheduledTask.php#L58)). I think it would be safer to stick with the previous behavior.

lib/Config/Populator.php
```diff
@@ -689,6 +690,22 @@ private function updateSentUnsubscribeLinksToInstantUnsubscribeLinks() {
...
+    global $wpdb;
+    $wpdb->query(
+      'UPDATE `' . ScheduledTask::$_table . '` t ' .
+      'JOIN `' . SendingQueue::$_table . '` q ON t.`id` = q.`task_id` ' .
+      'JOIN `' . Newsletter::$_table . '` n ON n.`id` = q.`newsletter_id` ' .
+      'SET t.`status` = "' . ScheduledTask::STATUS_PAUSED . '" ' .
+      'WHERE ' .
+      ' t.`status` = "' . ScheduledTask::STATUS_SCHEDULED . '" ' .
+      ' AND n.`status` = "' . Newsletter::STATUS_DRAFT . '" '
+    );
```
costasovo, line 706: It would be nicer to obtain table names using Doctrine and also use constants from entities instead of the old models so that we don't add more old models usage.
Here is how to get [table name from entity manager](https://github.com/mailpoet/mailpoet/blob/080a8221356019c513a80326c3789ab6d91ff56d/lib/Newsletter/NewslettersRepository.php#L221).

costasovo: If we inject the entityManager we could also use doctrine query instead of wpdb

costasovo (changes requested):
I've added a few comments. The ones in Populator are more like nice to have.

lib/Entities/NewsletterEntity.php
```diff
@@ -257,6 +259,35 @@ public function getStatus() {
...
+      /** @var SendingQueueEntity $queue */
+      $task = $queue->getTask();
+      if ($task === null) continue;
+
+      $scheduled = new Carbon($task->getScheduledAt());
+      if ($scheduled < (new Carbon())->subDays(30)) return;
+
+      if (($status === self::STATUS_DRAFT) && ($task->getStatus() !== ScheduledTaskEntity::STATUS_SCHEDULED)) return;
+      if (($status === self::STATUS_ACTIVE) && ($task->getStatus() !== ScheduledTaskEntity::STATUS_PAUSED)) return;
+
```
costasovo, line 288: I've just noticed that in these 3 conditions we use `return` but there should be `continue` or we may skip some tasks.

costasovo (changes requested):

lib/Entities/NewsletterEntity.php
```diff
@@ -257,6 +259,35 @@ public function getStatus() {
...
+    $queues = $this->getUnfinishedQueues();
+
+    foreach ($queues as $queue) {
+      /** @var SendingQueueEntity $queue */
+      $task = $queue->getTask();
+      if ($task === null) continue;
+
+      $scheduled = new Carbon($task->getScheduledAt());
+      if ($scheduled < (new Carbon())->subDays(30)) return;
```
costasovo, line 284: Here is one more `return` that should be `continue`.

costasovo (changes requested):

costasovo (approved):

veljkho (approved):

PR #3397, opened 2021-03-18
Layout components updates [MAILPOET-3503]

NeosinneR:
[MAILPOET-3503]

[MAILPOET-3503]: https://mailpoet.atlassian.net/browse/MAILPOET-3503


assets/js/src/common/grid/centered_row.tsx
```diff
@@ -0,0 +1,15 @@
...
+  className?: string;
+};
+
+const CenteredRow = ({ children, className }: Props) => (
+  <div className={classnames(className, 'mailpoet-centered-row')}>
+    {children}
+  </div>
+);
+
+export default CenteredRow;
```
lysyjan, line 15: We want to avoid using export default, and we should use named exports everywhere.

.eslintrc.ts.json
```diff
@@ -113,6 +113,7 @@
         "assets/js/src/common/form/yesno/yesno.tsx",
         "assets/js/src/common/functions/change_handlers.ts",
         "assets/js/src/common/grid/_stories/grid.tsx",
+        "assets/js/src/common/grid/centered_row.tsx",
```
lysyjan, line 116: We shouldn't add more ignored files to this list. You created a new component that shouldn't be added to this long list.

lysyjan (changes requested):

lysyjan (approved):
Thanks

veljkho (approved):

PR #3401, opened 2021-03-20
Rewrite Forms::listing() API to use Doctrine [MAILPOET-3036]

rodrigoprimo:
This PR replaces the usage of Paris with Doctrine inside MailPoet\API\JSON\v1\Forms::listing(). It also introduces a new class MailPoet\Form\Listing\FormListingRepository that is used by listing() to prepare the query that is executed by Doctrine and a new MailPoet\API\JSON\ResponseBuilders\FormsResponseBuilder::buildForListing() method to prepare the response that is returned by listing(). A few tests were adjusted and new tests were added for the new class. Parts of FormsTests were refactored in a separate commit to use Doctrine entities instead of Paris objects.

[MAILPOET-3036]

### DIFF
--- a/lib/Form/Listing/FormListingRepository.php
+++ b/lib/Form/Listing/FormListingRepository.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace MailPoet\Form\Listing;
+
+use MailPoet\Entities\FormEntity;
+use MailPoet\Listing\ListingDefinition;
+use MailPoet\Listing\ListingRepository;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoetVendor\Doctrine\ORM\QueryBuilder;
+
+class FormListingRepository extends ListingRepository {
+  public function getGroups(ListingDefinition $definition): array {
+    $queryBuilder = clone $this->queryBuilder;
+    $this->applyFromClause($queryBuilder);
+    $this->applyParameters($queryBuilder, $definition->getParameters());
+
+    // total count
+    $countQueryBuilder = clone $queryBuilder;
+    $countQueryBuilder->select('COUNT(f) AS formCount');
+    $countQueryBuilder->andWhere('f.deletedAt IS NULL');
+    $totalCount = (int)$countQueryBuilder->getQuery()->getSingleScalarResult();
+
+    // trashed count
+    $trashedCountQueryBuilder = clone $queryBuilder;
+    $trashedCountQueryBuilder->select('COUNT(f) AS formCount');
+    $trashedCountQueryBuilder->andWhere('f.deletedAt IS NOT NULL');
+    $trashedCount = (int)$trashedCountQueryBuilder->getQuery()->getSingleScalarResult();
+
+    return [
+      [
+        'name' => 'all',
+        'label' => WPFunctions::get()->__('All', 'mailpoet'),
+        'count' => $totalCount,
+      ],
+      [
+        'name' => 'trash',
+        'label' => WPFunctions::get()->__('Trash', 'mailpoet'),
+        'count' => $trashedCount,
+      ],
+    ];
+  }
+
+  protected function applySelectClause(QueryBuilder $queryBuilder) {
+    $queryBuilder->select("PARTIAL f.{id,name,status,settings,createdAt,updatedAt,deletedAt}");
+  }
+
+  protected function applyFromClause(QueryBuilder $queryBuilder) {
+    $queryBuilder->from(FormEntity::class, 'f');
+  }
+
+  protected function applyGroup(QueryBuilder $queryBuilder, string $group) {
+    // include/exclude deleted
+    if ($group === 'trash') {
+      $queryBuilder->andWhere('f.deletedAt IS NOT NULL');
+    } else {
+      $queryBuilder->andWhere('f.deletedAt IS NULL');
+    }
+  }
+
+  protected function applySorting(QueryBuilder $queryBuilder, string $sortBy, string $sortOrder) {
+    $queryBuilder->addOrderBy("f.$sortBy", $sortOrder);
+  }
+
+  protected function applySearch(QueryBuilder $queryBuilder, string $search) {
+    // TODO: Implement applySearch() method.
+  }
+
+  protected function applyFilters(QueryBuilder $queryBuilder, array $filters) {
+    // TODO: Implement applyFilters() method.
+  }
+
+  protected function applyParameters(QueryBuilder $queryBuilder, array $parameters) {
+    // TODO: Implement applyParameters() method.
+  }
+}
--- a/tests/integration/API/JSON/ResponseBuilders/FormsResponseBuilderTest.php
+++ b/tests/integration/API/JSON/ResponseBuilders/FormsResponseBuilderTest.php
@@ -4,12 +4,62 @@ namespace MailPoet\API\JSON\ResponseBuilders;
 
 use MailPoet\DI\ContainerWrapper;
 use MailPoet\Entities\FormEntity;
+use MailPoet\Models\StatisticsForms;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class FormsResponseBuilderTest extends \MailPoetTest {
+  /** @var ContainerWrapper */
+  protected $container;
+
+  /** @var string */
+  protected $formName;
+
+  /** @var array */
+  protected $formBody;
+
+  /** @var array */
+  protected $formSettings;
+
+  public function _before() {
+    parent::_before();
+
+    $this->container = ContainerWrapper::getInstance();
+    $this->entityManager = $this->container->get(EntityManager::class);
+  }
+
   public function testItBuildsForm() {
-    $name = 'Form Builder Test';
-    $body = [
+    $form = $this->createForm('Form 1');
+
+    $responseBuilder = new FormsResponseBuilder();
+    $response = $responseBuilder->build($form);
+
+    expect($response['name'])->equals($this->formName);
+    expect($response['status'])->equals(FormEntity::STATUS_ENABLED);
+    expect($response['body']['name'])->equals($this->formBody['name']);
+    expect($response['body']['params']['label'])->equals($this->formBody['params']['label']);
+    expect($response['settings']['success_message'])->equals($this->formSettings['success_message']);
+  }
+
+  public function testItBuildsFormsForListing() {
+    $form1 = $this->createForm('Form 1');
+    $form2 = $this->createForm('Form 2');
+
+    $responseBuilder = new FormsResponseBuilder();
+    $response = $responseBuilder->buildForListing([$form1, $form2]);
+
+    expect($response)->count(2);
+    expect($response[0]['signups'])->equals(0);
+    expect($response[0]['segments'])->equals($this->formSettings['segments']);
+  }
+
+  public function _after() {
+    $this->truncateEntity(FormEntity::class);
+    StatisticsForms::deleteMany();
+  }
+
+  private function createForm($name) {
+    $this->formName = $name;
+    $this->formBody = [
       'type' => 'text',
       'name' => 'First name',
       'id' => 'first_name',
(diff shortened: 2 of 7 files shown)


lib/Form/Listing/FormListingRepository.php
```diff
@@ -0,0 +1,75 @@
...
+    } else {
+      $queryBuilder->andWhere('f.deletedAt IS NULL');
+    }
+  }
+
+  protected function applySorting(QueryBuilder $queryBuilder, string $sortBy, string $sortOrder) {
+    $queryBuilder->addOrderBy("f.$sortBy", $sortOrder);
+  }
+
+  protected function applySearch(QueryBuilder $queryBuilder, string $search) {
```
rodrigoprimo, line 64: I'm unsure what is the best thing to do with this method and the other two below it. It seems to me that the page that lists the forms does not support search, filters, and parameters, but those three methods must exist as we are extending ListingRepository.

costasovo: Ahh looks like we kind of expected that all listings will have some filters, search, etc 🤔 I guess we could alter the ListingRepository and mark those methods not abstract or keep them here just updating those Todo comments to some comments saying that we don't support the functionality for the forms listing.

lib/Form/Listing/FormListingRepository.php
```diff
@@ -0,0 +1,75 @@
...
+    // trashed count
+    $trashedCountQueryBuilder = clone $queryBuilder;
+    $trashedCountQueryBuilder->select('COUNT(f) AS formCount');
+    $trashedCountQueryBuilder->andWhere('f.deletedAt IS NOT NULL');
+    $trashedCount = (int)$trashedCountQueryBuilder->getQuery()->getSingleScalarResult();
+
+    return [
+      [
+        'name' => 'all',
+        'label' => WPFunctions::get()->__('All', 'mailpoet'),
```
costasovo, line 32: We don't have to use WPFunctions for translation functions like __, _x, etc. It is an exception. But there is no need to fix that. Just so that you know.

tests/integration/API/JSON/v1/FormsTest.php
```diff
@@ -186,40 +192,41 @@ public function testItCanExtractListsFromNestedListSelectionBlock() {
   }
 
   public function testItCanRestoreAForm() {
-    $this->form1->trash();
+    $this->form1->setDeletedAt(new \DateTime());
+    $this->formsRepository->flush();
 
-    $trashedForm = Form::findOne($this->form1->id);
+    $trashedForm = Form::findOne($this->form1->getId());
```
costasovo, line 198: Could you please also replace this usage of the old model? In Doctrine you can refresh an entity from DB (`$entityManager->refresh($entity)`).

In this case you could just call `$this->entityManager->refresh($this->form1)` and then make asserts on `$this->form1`.

tests/integration/API/JSON/v1/FormsTest.php
```diff
@@ -186,40 +192,41 @@ public function testItCanExtractListsFromNestedListSelectionBlock() {
...
     assert($trashedForm instanceof Form);
     expect($trashedForm->deletedAt)->notNull();
 
-    $response = $this->endpoint->restore(['id' => $this->form1->id]);
+    $response = $this->endpoint->restore(['id' => $this->form1->getId()]);
     expect($response->status)->equals(APIResponse::STATUS_OK);
     expect($response->data)->equals(
-      $this->reloadForm((int)$this->form1->id)->asArray()
+      $this->reloadForm((int)$this->form1->getId())->asArray()
     );
```
costasovo, line 206: `$this->reloadForm` internally uses the old model class. It is something that we want to get rid of. Looking at the whole expect, it makes not much sense to me. We compare that the response, that is created by a [specialized response builder class](https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/ResponseBuilders/FormsResponseBuilder.php) is equal to an old model instance converted to an array. The response builder has its own integration test so I guess we could either drop the response format assert from this test or if we want to check that API returns the response built by the response builder we could add one test case for this into this test. Does that make sense to you? What do you think?

tests/integration/API/JSON/v1/FormsTest.php
```diff
@@ -186,40 +192,41 @@ public function testItCanExtractListsFromNestedListSelectionBlock() {
...
     expect($response->data)->isEmpty();
     expect($response->status)->equals(APIResponse::STATUS_OK);
     expect($response->meta['count'])->equals(1);
   }
 
   public function testItCanDuplicateAForm() {
-    $response = $this->endpoint->duplicate(['id' => $this->form1->id]);
+    $response = $this->endpoint->duplicate(['id' => $this->form1->getId()]);
     expect($response->status)->equals(APIResponse::STATUS_OK);
     $form = Form::where('name', 'Copy of Form 1')->findOne();
```
costasovo, line 231: Could you please replace also this old model usage? You can use `$this->formsRepository`
`$form = $this->formsRepository->findOneBy(['name' => 'Copy of Form 1']);`

tests/integration/API/JSON/v1/FormsTest.php
```diff
@@ -19,16 +21,20 @@ class FormsTest extends \MailPoetTest {
...
     $this->endpoint = ContainerWrapper::getInstance()->get(Forms::class);
+    $this->formsRepository = ContainerWrapper::getInstance()->get(FormsRepository::class);
     $this->wp = WPFunctions::get();
-    $this->form1 = Form::createOrUpdate(['name' => 'Form 1']);
-    $this->form2 = Form::createOrUpdate(['name' => 'Form 2']);
-    $this->form3 = Form::createOrUpdate(['name' => 'Form 3']);
+    $this->form1 = $this->createForm('Form 1');
+    $this->form2 = $this->createForm('Form 2');
+    $this->form3 = $this->createForm('Form 3');
     Segment::createOrUpdate(['name' => 'Segment 1']);
```
costasovo, line 38: Could you please do the same small factory method for Segment the same way you did for the form here? It would be nice to clean the test from the usage of old models.

tests/integration/API/JSON/ResponseBuilders/FormsResponseBuilderTest.php
```diff
@@ -4,12 +4,62 @@
...
+    $response = $responseBuilder->buildForListing([$form1, $form2]);
+
+    expect($response)->count(2);
+    expect($response[0]['signups'])->equals(0);
+    expect($response[0]['segments'])->equals($this->formSettings['segments']);
+  }
+
+  public function _after() {
+    $this->truncateEntity(FormEntity::class);
+    StatisticsForms::deleteMany();
```
costasovo, line 57: Looks like we don't have a Doctrine entity for StatisticsForms yet. I'll create a Jira for that.

costasovo (commented):
Thanks! Good job! 👍 As you started with refactoring the Forms API test, I would just ask you to refactor a few more bits there so that we can get rid of more old model usage.

lib/Form/Listing/FormListingRepository.php
rodrigoprimo: Thanks for the information. I pushed a commit fixing this anyway for the sake of consistency and to help me remember about this next time.

Would it make sense to remove \MailPoet\WP\Functions::__() and other i18n functions and replace all the usages?

tests/integration/API/JSON/v1/FormsTest.php
rodrigoprimo: I will replace the usage of the old model as you suggested. I wasn't sure how much time I should spend on doing this for this test class.

On a related note, what are your thoughts on using \PHPUnit\Framework\Assert::assertInstanceOf() instead of assert() for assertions like this one (https://github.com/mailpoet/mailpoet/pull/3401/files#diff-e72a2e2b6300aa942cd56ac3d1cb382dee3064052c53d2aba74b20fc2aaf1f27R199)? I personally prefer more specialized assertions instead of the generic assert(). One of the reasons for that is that we get specific error messages that can help when debugging problems in the test.

tests/integration/API/JSON/v1/FormsTest.php
rodrigoprimo: Sure, done

tests/integration/API/JSON/ResponseBuilders/FormsResponseBuilderTest.php
rodrigoprimo: Thanks

tests/integration/API/JSON/v1/FormsTest.php
rodrigoprimo: > $this->reloadForm internally uses the old model class. It is something that we want to get rid of.

Some of the \MailPoet\API\JSON\v1\Forms methods that are being tested in this class still use the old form model. I believe it will be easier to change \MailPoet\Test\API\JSON\v1\FormsTest::reloadForm() once we update all the methods inside the class being tested so outside of the scope of this PR. Sounds good?

> We compare that the response, that is created by a specialized response builder class is equal to an old model instance converted to an array. The response builder has its own integration test so I guess we could either drop the response format assert from this test or if we want to check that API returns the response built by the response builder we could add one test case for this into this test. Does that make sense to you? What do you think?

I tend to agree with your assessment of this test, but I'm inclined to suggest that we address this in a separate PR as it seems outside of the scope of this one and it will probably be easier to address it once all methods of the class being tested are migrated to the new model. Once that happens, we might be able to remove reloadForm(). What do you think?

rodrigoprimo:
Thanks for your detailed review, @costasovo. I believe I addressed all the points that you raised. Could you please check this PR again?

lib/Form/Listing/FormListingRepository.php
costasovo: I think it is something we should do so that it is used consistently. It is probably not a high priority though. I remove it when I see in the code I'm touching.

tests/integration/API/JSON/v1/FormsTest.php
costasovo: We use assert because of PHPStan. I think it doesn't understand assertInstanceOf.

veljkho (approved):

tests/integration/API/JSON/v1/FormsTest.php
rodrigoprimo: I was not able to reproduce this but maybe I'm missing something. Anyway, generally speaking, whenever possible I believe we should try to use more specific assertions. So if it indeed PHPStan complains about the PHPUnit assertions maybe we should look into updating our PHPStan configuration file to fix this?

Here is what I did to test it:

```shell
$ git diff tests/integration/API/JSON/v1/FormsTest.php
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: tests/integration/API/JSON/v1/FormsTest.php
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ tests/integration/API/JSON/v1/FormsTest.php:199 @ class FormsTest extends \MailPoetTest {
    $this->formsRepository->flush();
    $this->entityManager->refresh($this->form1);

-   assert($this->form1 instanceof FormEntity);
+   $this->assertInstanceOf(FormEntity::class, $this->form1);
    expect($this->form1)->notNull();

    $response = $this->endpoint->restore(['id' => $this->form1->getId()]);

$ php -d memory_limit=-1 /home/rodrigo/devel/mailpoet-dev-env/mailpoet/tasks/phpstan/vendor/bin/phpstan  analyse --configuration=tasks/phpstan/phpstan-tests.neon tests/integration/API/JSON/v1/FormsTest.php
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                                                        
 [OK] No errors                                                                                                         
```                                                                                                                        

Should I do something different to see that PHPStan error when we use assertInstanceOf()?

costasovo: @rodrigoprimo Thanks for checking! Sorry, my bad! I think that its `expect($something)->isInstanceOf('Something');` what was causing the issue. Looks like `$this->assertInstanceOf()` works correctly with PHPStan so I guess we should use it. Good find!

rodrigoprimo: Thanks for confirming!

PR #3402, opened 2021-03-22
Add support for arguments in custom shortcodes [MAILPOET-3465]

wxa:
[MAILPOET-3465]

We already had limited support for arguments: it was using a different syntax than WordPress (`[category:action | argument:argument_value]`), supported only one argument per shortcode and that argument was not passed to custom shortcode functions.

I have added support for WP-style shortcodes with multiple arguments (`[category:action arg1="value1" arg2="value2"]`), they are now passed to custom functions. The support for MailPoet-style shortcodes is retained for backwards compatibility.

One thing to be aware of is that the shortcode string passed to a custom function can be full (with arguments) or stripped of arguments depending on the `tracking.enabled` setting value. So if arguments are used in a custom function, the check for a proper shortcode inside it should verify only the beginning of a string, e.g. `if (strpos($shortcode, '[some:shortcode') !== 0) return $shortcode;` because the latter part of a shortcode can be changing. We should document this in the KB along with the new `$arguments` argument after the change is released.

[MAILPOET-3465]: https://mailpoet.atlassian.net/browse/MAILPOET-3465

### DIFF
--- a/lib/Newsletter/Shortcodes/Shortcodes.php
+++ b/lib/Newsletter/Shortcodes/Shortcodes.php
@@ -80,19 +80,60 @@ class Shortcodes {
       false;
   }
 
+  /**
+   * Parse a MailPoet-style shortcode.
+   * The syntax is [category:action | argument:argument_value], it can have a single argument.
+   */
   public function match($shortcode) {
     preg_match(
       '/\[(?P<category>\w+)?:(?P<action>\w+)(?:.*?\|.*?(?P<argument>\w+):(?P<argument_value>.*?))?\]/',
       $shortcode,
       $match
     );
+    // If argument exists, copy it to the arguments array
+    if (!empty($match['argument'])) {
+      $match['arguments'] = [$match['argument'] => isset($match['argument_value']) ? $match['argument_value'] : ''];
+    }
     return $match;
+  }
+
+  /**
+   * Parse a WordPress-style shortcode.
+   * The syntax is [category:action arg1="value1" arg2="value2"], it can have multiple arguments.
+   */
+  public function matchWPShortcode($shortcode) {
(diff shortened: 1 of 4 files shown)


lib/Newsletter/Shortcodes/Shortcodes.php
```diff
@@ -80,19 +80,60 @@ public function extract($content, $categories = false) {
...
+    }
     return $match;
   }
 
+  /**
+   * Parse a WordPress-style shortcode.
+   * The syntax is [category:action arg1="value1" arg2="value2"], it can have multiple arguments.
+   */
+  public function matchWPShortcode($shortcode) {
+    $atts = WPFunctions::get()->shortcodeParseAtts(trim($shortcode, '[]/'));
```
pavel-mailpoet, line 105: We are already using DI in this class, you don't need to use `WPFunctions::get()`. You can inject the functions class directly.

pavel-mailpoet (changes requested):

lib/Newsletter/Shortcodes/Shortcodes.php
wxa: Fixed.

pavel-mailpoet (approved):

veljkho (changes requested):
Im getting issue when accessing Emails page:
`Too few arguments to function MailPoet\Newsletter\Shortcodes\Shortcodes::__construct(), 4 passed in /var/www/html/wp-content/plugins/mailpoet/generated/FreeCachedContainer.php on line 1775 and exactly 5 expected`

I do not see this when testing on online server, just locally. Can you see the same? Is everything good in `FreeCachedContainer.php` on line 1775? I'm not sure if I could merge even if its not working only locally.

wxa:
@veljkho Could you please clear your local `/generated/` folder and try again?

veljkho (approved):

PR #3406, opened 2021-03-23
Fix flaky counter in test ReinstallFromScratch [MAILPOET-3513]

veljkho:
[MAILPOET-3513](https://mailpoet.atlassian.net/browse/MAILPOET-3513)


costasovo (approved):
Looks good! Did you by a chance find out why `count_users` returned an incorrect value?

veljkho:
> Looks good! Did you by a chance find out why `count_users` returned an incorrect value?

From what I have seen, that counter counted only previously added users (6) and not + admin (7th).

PR #3408, opened 2021-03-24
Refactor Subscribers::save to Doctrine [MAILPOET-3031]

lysyjan:
[MAILPOET-3031]

[MAILPOET-3031]: https://mailpoet.atlassian.net/browse/MAILPOET-3031

### DIFF
--- a/lib/Subscribers/SubscriberSaveController.php
+++ b/lib/Subscribers/SubscriberSaveController.php
@@ -1,0 +1,218 @@
+<?php
+
+namespace MailPoet\Subscribers;
+
+use MailPoet\CustomFields\CustomFieldsRepository;
+use MailPoet\Entities\StatisticsUnsubscribeEntity;
+use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Entities\SubscriberSegmentEntity;
+use MailPoet\Newsletter\Scheduler\WelcomeScheduler;
+use MailPoet\Segments\SegmentsRepository;
+use MailPoet\Settings\SettingsController;
+use MailPoet\Statistics\Track\Unsubscribes;
+use MailPoet\Util\Security;
+use MailPoet\WP\Functions as WPFunctions;
+
+class SubscriberSaveController {
+  /** @var CustomFieldsRepository */
+  private $customFieldsRepository;
+
+  /** @var Security */
+  private $security;
+
+  /** @var SettingsController */
+  private $settings;
+
+  /** @var SegmentsRepository */
+  private $segmentsRepository;
+
+  /** @var SubscriberCustomFieldRepository */
+  private $subscriberCustomFieldRepository;
+
+  /** @var SubscribersRepository */
+  private $subscribersRepository;
+
+  /** @var SubscriberSegmentRepository */
+  private $subscriberSegmentRepository;
+
+  /** @var Unsubscribes */
+  private $unsubscribesTracker;
+
+  /** @var WelcomeScheduler */
+  private $welcomeScheduler;
+
+  /** @var WPFunctions */
+  private $wp;
+
+  public function __construct(
+    CustomFieldsRepository $customFieldsRepository,
+    Security $security,
+    SettingsController $settings,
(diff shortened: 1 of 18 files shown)


lib/Util/Security.php
```diff
@@ -70,7 +71,7 @@ static public function generateUnsubscribeToken($model) {
 
   public function generateUnsubscribeTokenByEntity($entity): string {
     $repository = null;
-    if ($entity instanceof NewsletterEntity) {
+    if ($entity instanceof NewsletterEntity || $entity instanceof SubscriberEntity) {
```
pavel-mailpoet, line 74: Hmm, I don't think this is going to work. If the instance is `SubscriberEntity` we still use the newsletters repository and line 82 down bellow would never work.

lib/Subscribers/SubscriberSegmentRepository.php
```diff
@@ -12,4 +15,77 @@ class SubscriberSegmentRepository extends Repository {
   protected function getEntityClassName() {
     return SubscriberSegmentEntity::class;
   }
+
+  public function getNonDefaultSubscribedSegments(int $subscriberId): array {
+    $qb = $this->entityManager->createQueryBuilder();
+    return $qb->select('ss')
+      ->from(SubscriberSegmentEntity::class, 'ss')
+      ->join('ss.segment', 'seg', Join::WITH, $qb->expr()->eq('seg.type', ':typeDefault'))
```
pavel-mailpoet, line 23: Sorry, I probably don't understand this, but the name of this method suggests it should return all segments except the default one. But here you are only joining the default segment. What am I missing?

lib/Subscribers/SubscriberSaveController.php
```diff
@@ -0,0 +1,218 @@
...
+    $this->subscribersRepository = $subscribersRepository;
+    $this->subscriberCustomFieldRepository = $subscriberCustomFieldRepository;
+    $this->unsubscribesTracker = $unsubscribesTracker;
+    $this->welcomeScheduler = $welcomeScheduler;
+    $this->wp = $wp;
+  }
+
+  public function save(array $data): SubscriberEntity {
+    if (is_array($data) && !empty($data)) {
+      $data = WPFunctions::get()->stripslashesDeep($data);
```
pavel-mailpoet, line 73: I wonder why are you using the static call here when WPFunctions are injected into this class.

pavel-mailpoet (changes requested):
Good job, I have just a couple of small remarks.

lib/Util/Security.php
lysyjan: Yes, you are right. This is my stupid mistake. Thank you for the catch.

lib/Subscribers/SubscriberSaveController.php
lysyjan: @pavel-mailpoet This method is used in [the old model](https://github.com/mailpoet/mailpoet/blob/master/lib/Models/Subscriber.php#L343-L347)
I expected that there is a good reason. If you think that this shouldn't be here, I can remove it.

lib/Subscribers/SubscriberSegmentRepository.php
lysyjan: You are right. I wrote the incorrect condition in the join. I will fix it.

lib/Models/Subscriber.php
```diff
@@ -340,6 +340,9 @@ public function customFields() {
     )->select_expr(MP_SUBSCRIBER_CUSTOM_FIELD_TABLE . '.value');
   }
 
+  /**
+   * @deprecated Will be removed after 2021/09/30. Use MailPoet\Subscribers\SubscriberSaveController
```
pavel-mailpoet, line 344: I think we should also add `trigger_error('Calling ' . $methodName . ' is deprecated and will be removed.', E_USER_DEPRECATED);` so that everyone still using this will get a notice.

pavel-mailpoet (approved):

veljkho (approved):

PR #3409, opened 2021-03-24
Add dynamic segments support to count shortcode [MAILPOET-3425]

lysyjan:
[MAILPOET-3425]

[MAILPOET-3425]: https://mailpoet.atlassian.net/browse/MAILPOET-3425

### DIFF
--- a/lib/Segments/SegmentSubscribersRepository.php
+++ b/lib/Segments/SegmentSubscribersRepository.php
@@ -55,24 +55,37 @@ class SegmentSubscribersRepository {
   }
 
   public function getSubscribersCountBySegmentIds(array $segmentIds, string $status = null): int {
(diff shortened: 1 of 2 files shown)


costasovo (commented):
Good job! 👍 The getSubscribersCountBySegmentIds method is now a bit more complex. I think it would be nice to extend [the test for this class](https://github.com/mailpoet/mailpoet/blob/40c9f047be76fc8867ad936f956c94f7cd49ce5a/tests/integration/Segments/SegmentSubscribersRepositoryTest.php) and add also a test case (or cases?) for this method. What do you think?

lysyjan:
👍🏻  That sounds good. I will add some tests.

costasovo (approved):
👍

veljkho (approved):

PR #3411, opened 2021-03-25
Simplify choosing segment condition [MAILPOET-3476]

pavel-mailpoet:
[MAILPOET-3476]

@veljkho the whole segments form is built from scratch. So when testing please test the whole creating/updating dynamic segments.

[MAILPOET-3476]: https://mailpoet.atlassian.net/browse/MAILPOET-3476


costasovo (changes requested):
Thanks for refactoring the whole form 👍 and 👏 for fixing the issue with the underscore. Good job!

I added a comment with some suggestions related to the usage of the global window within some components.

I also tested the change locally and found an issue with segments saved with the old version of the form. Looks like the change caused that some data we send on API changed from integers to strings and dynamic segments saved with the old version of the form don't load properly. I took a screenshot. I hope it explains the issue...
It is probably related to all ids we save within the form.

Screen Shot 2021-03-26 at 09 41 45

assets/js/src/segments/dynamic/form.tsx
```diff
@@ -85,6 +86,18 @@ const DynamicSegmentForm: React.FunctionComponent = () => {
...
+    }): AnyFormItem {
+      const converted = JSON.parse(JSON.stringify(data));
+      // for compatibility with older data
+      if (has('link_id', data)) converted.link_id = data.link_id.toString();
+      if (has('newsletter_id', data)) converted.newsletter_id = data.newsletter_id.toString();
+      if (has('product_id', data)) converted.product_id = data.product_id.toString();
+      if (has('category_id', data)) converted.category_id = data.category_id.toString();
+      return converted;
+    }
+
```
costasovo, line 100: This doesn't fix that some data that used to be stored as integers would be stored as strings. I'm afraid that this may cause some unexpected issues. 
For example, I wanted to test the email link clicked segment but for some reason, I'm not able to create or edit such a segment, because I can't select and save the link. I recorded video https://d.pr/v/qL20X9. I don't get any error in the console but the input just disappears and appears again without the selected link.

Maybe we could fix it the other way around by altering components for selecting emails, links, products, and categories to accept integers. What do you think?

costasovo (changes requested):
I found an issue when I was trying to test the email segment.

assets/js/src/segments/dynamic/form.tsx
pavel-mailpoet: > Maybe we could fix it the other way around by altering components for selecting emails, links, products, and categories to accept integers. What do you think?

Yes, that would be better, but I am afraid that is not possible because we use a library that doesn't support that.

costasovo (approved):
👍

veljkho (changes requested):
Everything's good so far except:
- 4th point is still the old name "WordPress Users Roles"
- I don't see calculations on Emails and WooCommerce, but it is present for WordPress User Roles only

veljkho (changes requested):
Everything's good so far except:
- 4th point is still the old name "WordPress Users Roles"
- I don't see calculations on Emails and WooCommerce, but it is present for WordPress User Roles only

Edit: Sorry for duplicated comments, it's GitHub's issue.

costasovo (approved):

veljkho (changes requested):
There are 2 more issues that I found, and I believe everything else is working well.

1st issue: Empty specific link field
- I found out that if you select clicked specific link, and save it.. then go back to edit that segment, you'll not see that specific link selected anymore - empty field.

2nd issue: Calculation of subscribers who clicked specific link doesn't change if you change links
- If you set clicked > specific link, you'll see proper number of subscribers who clicked.. but if you then change link you'll still see the old number of clicked subscribers. Looks like it won't re-calculate for specific links if you changing them.

costasovo (approved):

PR #3412, opened 2021-03-25
Check existence panels when creating debugger [MAILPOET-3420]

pavel-mailpoet:
[MAILPOET-3420]

[MAILPOET-3420]: https://mailpoet.atlassian.net/browse/MAILPOET-3420

### DIFF
--- a/lib/Doctrine/EntityManagerFactory.php
+++ b/lib/Doctrine/EntityManagerFactory.php
@@ -5,8 +5,6 @@ namespace MailPoet\Doctrine;
 use MailPoet\Doctrine\EventListeners\EmojiEncodingListener;
 use MailPoet\Doctrine\EventListeners\TimestampListener;
 use MailPoet\Doctrine\EventListeners\ValidationListener;
-use MailPoet\Tracy\ApiPanel\ApiPanel;
-use MailPoet\Tracy\DIPanel\DIPanel;
 use MailPoet\Tracy\DoctrinePanel\DoctrinePanel;
 use MailPoetVendor\Doctrine\DBAL\Connection;
 use MailPoetVendor\Doctrine\ORM\Configuration;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/EntityManagerFactory.php
```diff
@@ -43,10 +45,15 @@ public function __construct(
...
 
-  public function createEntityManager() {
+  public function createEntityManager(): EntityManager {
     $entityManager = EntityManager::create($this->connection, $this->configuration);
     $this->setupListeners($entityManager);
-    if (class_exists(Debugger::class)) {
+    if (
+      class_exists(Debugger::class)
+      && class_exists(DoctrinePanel::class)
+      && class_exists(DIPanel::class)
```
wxa, line 54: I'm not sure if we need to check `DIPanel` and `ApiPanel` existence here because they're not referenced in the file or inside `DoctrinePanel`. But there are other files where these panels are explicitly instantiated, they will likely also break so we need safety checks in them too:
 https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/API.php#L194-L197
 https://github.com/mailpoet/mailpoet/blob/master/lib/AdminPages/PageRenderer.php#L67-L69

wxa (changes requested):
There are also a couple interfaces used in the plugin: `Tracy/ILogger` (used [here](https://github.com/mailpoet/mailpoet/blob/master/lib/API/JSON/API.php#L18)) and `Tracy/IBarPanel` (used in our custom panels, e.g. [here](https://github.com/mailpoet/mailpoet/blob/master/lib/Tracy/ApiPanel/ApiPanel.php#L8)). Do you think it makes sense to check their existence where necessary or, since they're a part of Tracy, they should exist if `Debugger` exists, and broken Tracy installations aren't our problem?

wxa (approved):

veljkho (approved):

PR #3416, opened 2021-03-25
Clean up old dynamic segments code [MAILPOET-3484]

wxa:
[MAILPOET-3484]

I have removed the unused `Segments\BulkAction` class which was the only thing still referencing the `MailPoet\DynamicSegments` namespace. That said, cleaning up all obsolete bulk action code from all models etc. is worth a separate task which I have created: https://mailpoet.atlassian.net/browse/MAILPOET-3520

[MAILPOET-3484]: https://mailpoet.atlassian.net/browse/MAILPOET-3484


lib/Models/DynamicSegment.php
```diff
@@ -94,4 +107,16 @@ public static function bulkDelete($orm) {
 
     return ['count' => $count];
   }
+
+  /**
+   * @deprecated This is here for displaying the deprecation warning for properties.
+   */
+  public function __get($key) {
+    self::deprecationError('property "' . $key . '"');
+    return parent::__get($key);
```
costasovo, line 116: A very good idea to add this! I'm thinking about adding something similar also for `__callStatic` so that we cover also methods like `Model::findOne` etc. Do you think it is worth adding?

costasovo (commented):
Wow! 👍 Good job finding those other unused classes! I have one comment/suggestion to consider.

lib/Models/DynamicSegment.php
wxa: Sure, I have added it.

costasovo (approved):

veljkho (approved):

PR #3424, opened 2021-03-29
Pause emails with deleted/trashed lists [MAILPOET-3418]

costasovo:
[MAILPOET-3418]

[MAILPOET-3418]: https://mailpoet.atlassian.net/browse/MAILPOET-3418

### DIFF
--- a/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -32,6 +32,7 @@ class SendingQueue {
   public $batchSize;
   const BATCH_SIZE = 20;
   const TASK_BATCH_SIZE = 5;
+  const EMAIL_WITH_INVALID_LIST_OPTION = 'mailpoet_email_with_invalid_list';
 
   /** @var StatsNotificationsScheduler */
   public $statsNotificationsScheduler;
@@ -57,6 +58,9 @@ class SendingQueue {
   /** @var SegmentsRepository */
   private $segmentsRepository;
 
+  /** @var WPFunctions */
+  private $wp;
+
   public function __construct(
     SendingErrorHandler $errorHandler,
     StatsNotificationsScheduler $statsNotificationsScheduler,
@@ -65,6 +69,7 @@ class SendingQueue {
     CronHelper $cronHelper,
     SubscribersFinder $subscriberFinder,
     SegmentsRepository $segmentsRepository,
+    WPFunctions $wp,
     $mailerTask = false,
     $newsletterTask = false
   ) {
@@ -75,7 +80,7 @@ class SendingQueue {
     $this->newsletterTask = ($newsletterTask) ? $newsletterTask : new NewsletterTask();
     $this->segmentsRepository = $segmentsRepository;
     $this->mailerMetaInfo = new MetaInfo;
-    $wp = new WPFunctions;
+    $this->wp = $wp;
     $this->batchSize = $wp->applyFilters('mailpoet_cron_worker_sending_queue_batch_size', self::BATCH_SIZE);
     $this->loggerFactory = $loggerFactory;
     $this->newslettersRepository = $newslettersRepository;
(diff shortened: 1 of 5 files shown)


lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -111,6 +123,18 @@ public function process($timer = false) {
       $this->mailerTask->configureMailer($newsletter);
       // get newsletter segments
       $newsletterSegmentsIds = $this->newsletterTask->getNewsletterSegments($newsletter);
+      // Pause task in case some of related segments was deleted or trashed
+      if ($newsletterSegmentsIds && !$this->checkDeletedSegments($newsletterSegmentsIds)) {
+        $this->loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->addInfo(
+          'pause task in sending queue due deleted or trashed segment',
+          ['task_id' => $queue->taskId]
+        );
+        $queue->status = ScheduledTaskEntity::STATUS_PAUSED;
```
pavel-mailpoet, line 132: This pauses the sending queue. Shall we also change the status of the scheduled task?

lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -279,6 +308,29 @@ public function sendNewsletters(
     );
   }
 
+  /**
+   * Checks whether some of segments was deleted or trashed
+   * @param int[] $segmentIds
+   * @return bool
+   */
+  private function checkDeletedSegments(array $segmentIds) {
```
pavel-mailpoet, line 316: Is there a reason not to specify the returned type? `private function checkDeletedSegments(array $segmentIds): boolean`

lib/Util/Notices/EmailWithInvalidListNotice.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Cron\Workers\SendingQueue\SendingQueue;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class EmailWithInvalidListNotice {
+  const OPTION_NAME = SendingQueue::EMAIL_WITH_INVALID_LIST_OPTION;
```
pavel-mailpoet, line 10: Nothing wrong here, I am just curious. Why do you copy the const from Sending Queue instead of using in the code the const from the queue?

lib/Util/Notices/EmailWithInvalidListNotice.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Cron\Workers\SendingQueue\SendingQueue;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class EmailWithInvalidListNotice {
```
pavel-mailpoet, line 9: I am not sure, but we've been calling `List` as `Segment` in the code. Should we be consistent with that?

lib/Cron/Workers/SendingQueue/SendingQueue.php
costasovo: Actually this pauses the scheduled task. The $queue variable in this case refers to [MailPoet\Tasks\Sending](https://github.com/mailpoet/mailpoet/blob/master/lib/Tasks/Sending.php) that combines those 2 models.

lib/Util/Notices/EmailWithInvalidListNotice.php
costasovo: I noticed that all Notice classes uses OPTION_NAME and actually it is used in [the notices controller for dismissing](https://github.com/mailpoet/mailpoet/blob/master/lib/Util/Notices/PermanentNotices.php#L87-L107). So I wanted to keep it similar.

lib/Util/Notices/EmailWithInvalidListNotice.php
costasovo: Good point 👍

pavel-mailpoet (approved):

veljkho (approved):

PR #3425, opened 2021-03-30
Move calling newsletters redirect to the menu [MAILPOET-3536]

lysyjan:
[MAILPOET-3536]

[MAILPOET-3536]: https://mailpoet.atlassian.net/browse/MAILPOET-3536

### DIFF
--- a/lib/AdminPages/Pages/Newsletters.php
+++ b/lib/AdminPages/Pages/Newsletters.php
@@ -103,15 +103,6 @@ class Newsletters {
   public function render() {
     global $wp_roles; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
 
-    if (isset($_GET['stats'])) {
-      $this->wp->wpSafeRedirect(
-        $this->wp->getSiteUrl(null,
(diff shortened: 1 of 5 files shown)


lib/AdminPages/Pages/Newsletters.php
```diff
@@ -100,9 +100,7 @@ public function __construct(
     $this->segmentsListRepository = $segmentsListRepository;
   }
 
-  public function render() {
-    global $wp_roles; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
-
+  public function redirect(): void {
```
costasovo, line 103: I find the name a bit confusing since it doesn't redirect every time. So maybe we could rename it to something like `checkStatsRedirect` and return URL to redirect or `maybeStatsRedirect` or maybe even move the whole code to Menu.php since it works kind of as a router.

lysyjan: Good point. I will change it. I thought about routing as a part of Menu.php, but I liked that the redirect logic stays in the same place. So, I would move the whole routing to the Menu.

costasovo (approved):

veljkho (approved):

PR #3432, opened 2021-04-02
Update Storybook.js library [MAILPOET-3546]

NeosinneR:
[MAILPOET-3546]

[MAILPOET-3546]: https://mailpoet.atlassian.net/browse/MAILPOET-3546


rodrigoprimo (approved):
This looks good to me. To test it, I simply generated a storybook using `./do storybook:build` first on the master branch and then on this branch. After building the storybook on both occasions, I checked http://mp3.localhost/wp-content/plugins/mailpoet/storybook-static/index.html and the results seemed ok to me. That being said, I'm not familiar with Storybook and how we use it. So please let me know if I should have tested this in a different way.

veljkho (approved):

PR #3439, opened 2021-04-06
Update PHPStan to 0.12.83 [MAILPOET-3491]

rodrigoprimo:
This PR updates PHPStan to version 0.12.83 to fix a problem that we had where it was not possible to run PHPStan with PHP > 8.0.0. For more information, see this thread: https://github.com/phpstan/phpstan/discussions/4762.

Please note that I'm running Composer 2, and thus when I used to update PHPStan, it also updated the value of the field `plugin-api-version` in `composer.lock`. I'm not sure if this is a problem or not and if we have a reason to want to keep using Composer 1.

[MAILPOET-3491]

[MAILPOET-3491]: https://mailpoet.atlassian.net/browse/MAILPOET-3491

### DIFF
--- a/lib/Entities/NewsletterEntity.php
+++ b/lib/Entities/NewsletterEntity.php
@@ -280,7 +280,7 @@ class NewsletterEntity {
       $task = $queue->getTask();
       if ($task === null) continue;
 
-      $scheduled = new Carbon($task->getScheduledAt());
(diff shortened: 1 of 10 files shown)


costasovo (approved):
Good job! 👍

veljkho (approved):

PR #3443, opened 2021-04-08
Fix woo category segment edit [MAILPOET-3555]

costasovo:
[MAILPOET-3555]

[MAILPOET-3555]: https://mailpoet.atlassian.net/browse/MAILPOET-3555


lysyjan (approved):
Good job! Thanks

veljkho (approved):

PR #3448, opened 2021-04-13
Implement new WooCommerce dynamic segment based on the number of orders [MAILPOET-3228]

rodrigoprimo:
Creating this PR as a draft for now for some early feedback before I finalize it. Please note that I'm not familiar with React so there is a high chance that I'm missing something on that part of the code.

See inline comments for specific questions that I have.

Besides that, I'm planning to write some acceptance tests for the changes introduced here as this is the kind of test that we are using for the other WooCommerce related segments. I plan to work on this while I wait for initial feedback on this PR. Another thing remaining for me to do is wrap some strings with i18n functions. I started a conversation about this on Slack as some of those strings require interpolation with React components.

[MAILPOET-3228]

[MAILPOET-3228]: https://mailpoet.atlassian.net/browse/MAILPOET-3228


assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -70,21 +85,67 @@ export const WooCommerceFields: React.FunctionComponent<Props> = ({ onChange, it
...
+              onChange,
+              assign(item),
+            ])({ category_id: option.value })}
+          />
+        </div>
+      </div>
+    );
+  } else if (item.action === WooCommerceActionTypes.NUMBER_OF_ORDERS) {
+    if (!item.number_of_orders_type) {
+      item.number_of_orders_type = '=';
```
rodrigoprimo, line 106: This assignment produces this error `106:7  error  Assignment to property of function parameter 'item'  no-param-reassign` (https://app.circleci.com/pipelines/github/mailpoet/mailpoet/6280/workflows/f6561af2-88bd-4f95-9026-4259b6e11605/jobs/122376), but I don't know what would be another way to make sure that item.number_of_orders_type is initialized with a default value. I tried setting `defaultValue` inside the Select component and that works for the rendered HTML but it doesn't set the value of item.number_of_orders_type.

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -70,21 +85,67 @@ export const WooCommerceFields: React.FunctionComponent<Props> = ({ onChange, it
...
-  }
+  } else if (item.action === WooCommerceActionTypes.PURCHASED_CATEGORY) {
+    optionFields = (
+      <div className="mailpoet-form-field">
+        <div className="mailpoet-form-input mailpoet-form-select">
+          <Select
+            placeholder={MailPoet.I18n.t('selectWooPurchasedCategory')}
+            options={categoryOptions}
+            value={find(['value', item.category_id], categoryOptions)}
+            onChange={(option: SelectOption): void => compose([
```
rodrigoprimo, line 96: I copied this `onChange` from the other action types but I don't fully understand what is happening here.

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -70,21 +85,67 @@ export const WooCommerceFields: React.FunctionComponent<Props> = ({ onChange, it
...
-            assign(item),
-          ])({ category_id: option.value })}
-        />
+    optionFields = (
+      <div>
+        <Grid.CenteredRow className="mailpoet-form-field">
+          <Select
+            options={numberOfOrdersTypeOptions}
+            value={find(['value', item.number_of_orders_type], numberOfOrdersTypeOptions)}
+            onChange={(option: SelectOption): void => compose([
```
rodrigoprimo, line 115: I copied this onChange from the other action types but I don't fully understand what is happening here.

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -70,21 +85,67 @@ export const WooCommerceFields: React.FunctionComponent<Props> = ({ onChange, it
...
+              onChange,
+              assign(item),
+            ])({ number_of_orders_type: option.value })}
+          />
+          <Input
+            type="number"
+            min={0}
+            value={item.number_of_orders_count || ''}
+            placeholder="count"
+            onChange={(event): void => compose([
```
rodrigoprimo, line 125: Similar to the comment above, I don't fully understand what is happening inside this `onChange`. I made some modifications from the one used above in the Select component and it seems to work. But I don't know if there is a better way to do this.

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
pavel-mailpoet: This doesn't work. You cannot change the `item` properties and expect them to propagate everywhere. The item is defined up in the components tree here: https://github.com/mailpoet/mailpoet/blob/99eff6129ccd96bc11738cb5e6f6b5cf444bd07c/assets/js/src/segments/dynamic/form.tsx#L68

To change the property we need to call the function `setItem` on `DynamicSegmentForm`. That function is passed in this place as `onChange` so you need to call it to update `item`. If you did it in the code here like this: 
`onChange({...})` that would always override the value using the default one. So we need to wrap it in the `useEffect` hook like this:
```
useEffect(() => {
    if (item.number_of_orders_type === undefined) {
      onChange(assign(item, { number_of_orders_type: '=' }));
    }
  }, [onChange, item, item.operator]);
```

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
pavel-mailpoet: > I copied this onChange from the other action types
I understand. This code is a result of a couple of iterations and could be probably simplified now.
Let me explain: 

- `compose` gets a list of functions and it calls them from the end and passes the result of the previous call to the next call.
- `assign` takes two objects and assigns the properties from the second one into the first one. 
- both functions are curried so they are called only after all parameters are passed.

The equivalent code is: `onChange(assign(item, { number_of_orders_type: option.value }))`. That can be simplified even more using only JS vanilla.

lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
```diff
@@ -0,0 +1,59 @@
...
+
+    if ($type === '=') {
+      $queryBuilder->having("COUNT(posts.ID) = :count");
+    } elseif ($type === '>') {
+      $queryBuilder->having("COUNT(posts.ID) > :count");
+    } elseif ($type === '<') {
+      $queryBuilder->having("COUNT(posts.ID) < :count");
+    }
+
+    $queryBuilder->setParameter('count', $count);
```
pavel-mailpoet, line 55: Be careful when using these parameters in dynamic segments. More dynamic segments will be combined into one big query. Users will be able to create a dynamic filter:
`I want all subscribers who made more than 10 orders in the last 100 days OR 5 orders in the last 10 days.`

That creates a big query and the parameter `count` would be twice in the query and that wouldn't work as intended.  We need to make the parameters unique. Look here how it is done: https://github.com/mailpoet/mailpoet/blob/5b45bdac1e0ba38e99d20aad7f6883c8863ba892/lib/Segments/DynamicSegments/Filters/EmailAction.php#L78

tests/unit/Segments/DynamicSegments/FilterDataMapperTest.php
```diff
@@ -166,4 +167,35 @@ public function testItChecksWooCommerceProductId() {
       'action' => WooCommerceProduct::ACTION_PRODUCT,
     ]);
   }
+
+  public function testItMapsWooCommerceNumberOfOrders() {
```
pavel-mailpoet, line 171: You wrote that you are going to create some acceptance tests. I would recommend writing unit tests and integration tests and test your logic in those. They are faster and acceptance tests tend to be more flakey and take more time. Have you heard about [testing pyramid](https://3fxtqy18kygf3on3bu39kh93-wpengine.netdna-ssl.com/wp-content/uploads/2020/01/test-automation-pyramid.jpg)?

I would write just a single acceptance test and more integration tests.

pavel-mailpoet (changes requested):
Just a word about commits:
- we tend to split our work into multiple commits. Especially such a big change that you made here. It makes the life of code reviewer easier
- we don't merge master in our local branches, we rebase it so the merge commit is not shown in the pull request

lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
```diff
@@ -0,0 +1,59 @@
...
+    $queryBuilder->innerJoin(
+      $subscribersTable,
+      $wpdb->postmeta,
+      'postmeta',
+      "postmeta.meta_key = '_customer_user' AND $subscribersTable.wp_user_id=postmeta.meta_value"
+    )->leftJoin(
+      'postmeta',
+      $wpdb->posts,
+      'posts',
+      'posts.ID = postmeta.post_id AND posts.post_date >= :date AND postmeta.post_id NOT IN ( SELECT id FROM ' . $wpdb->posts . ' as p WHERE p.post_status IN ("wc-cancelled", "wc-failed"))'
```
pavel-mailpoet, line 40: I am not sure about this subselect. I am not very good at MySQL, but I was told to avoid subselects as much as possible because they tend to be slow. The reason, I was told, is that MySQL needs to evaluate the subselect for each row. That can be slow But I am just asking here, did you test this? or do you know much about the final query's performance?

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
rodrigoprimo: Thanks for the explanation!

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -10,13 +10,17 @@ import {
   WooCommerceFormItem,
 } from '../types';
 import { SegmentFormData } from '../segment_form_data';
+import { Grid } from '../../../common/grid';
```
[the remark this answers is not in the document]
rodrigoprimo: :+1:

lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
rodrigoprimo: Thanks for pointing this out. I just fixed it.

tests/unit/Segments/DynamicSegments/FilterDataMapperTest.php
rodrigoprimo: Yeah, I'm familiar with the testing pyramid. I agree with your general recommendation of writing more unit tests and integration tests than acceptance tests because of the reasons you laid out. As I mentioned in the PR description, I was planning to write only acceptance tests for this segment because, unless I'm missing something, the other WooCommerce segments contain mostly just acceptance tests, so I was planning to follow this pattern.

I will look into adding unit and integration tests for this new segment.

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
rodrigoprimo: Thanks for the detailed response. I still need to familiarize myself with all React components. I had also missed where `item` was defined. I pushed a commit implementing your suggestion.

lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
rodrigoprimo: I'm not a MySQL expert as well, but I think that this query in general is not great. That being said, I'm not sure if there is much we can do about it at the moment in this particular PR. We are already using exactly the same subquery in the other two WooCommerce segments:

https://github.com/mailpoet/mailpoet/blob/5b45bdac1e0ba38e99d20aad7f6883c8863ba892/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php#L63
https://github.com/mailpoet/mailpoet/blob/5b45bdac1e0ba38e99d20aad7f6883c8863ba892/lib/Segments/DynamicSegments/Filters/WooCommerceProduct.php#L39

That is why I'm using it here, and I did not test it. I'm sure this query will be slow for large databases, but I don't think the subquery is the worst part. Probably the worst part is the inner join with wp_postmeta, which uses meta_value since this field is not indexed:

```
mysql> explain SELECT DISTINCT wp_mailpoet_subscribers.id as inner_subscriber_id FROM wp_mailpoet_subscribers INNER JOIN wp_postmeta postmeta ON postmeta.meta_key = '_customer_user' AND wp_mailpoet_subscribers.wp_user_id=postmeta.meta_value LEFT JOIN wp_posts posts ON posts.ID = postmeta.post_id AND posts.post_date >= '2021-04-04 18:24:20' AND postmeta.post_id NOT IN ( SELECT id FROM wp_posts as p WHERE p.post_status I
N ("wc-cancelled", "wc-failed")) GROUP BY inner_subscriber_id HAVING COUNT(posts.ID) > 1;
+----+--------------------+-------------------------+------------+-----------------+-------------------------------------------------------------------------------------------------------+------------+---------+-------------------------------+------+----------+----------------------------------------------+
| id | select_type        | table                   | partitions | type            | possible_keys                                                                                         | key        | key_len | ref                           | rows | filtered | Extra                                        |
+----+--------------------+-------------------------+------------+-----------------+-------------------------------------------------------------------------------------------------------+------------+---------+-------------------------------+------+----------+----------------------------------------------+
|  1 | PRIMARY            | postmeta                | NULL       | ref             | meta_key                                                                                              | meta_key   | 767     | const                         |   11 |   100.00 | Using where; Using temporary; Using filesort |
|  1 | PRIMARY            | wp_mailpoet_subscribers | NULL       | ref             | PRIMARY,email,unsubscribe_token,wp_user_id,updated_at,status_deleted_at,last_subscribed_at,link_token | wp_user_id | 9       | wordpress.postmeta.meta_value |    1 |   100.00 | Using where; Using index                     |
|  1 | PRIMARY            | posts                   | NULL       | eq_ref          | PRIMARY                                                                                               | PRIMARY    | 8       | wordpress.postmeta.post_id    |    1 |   100.00 | Using where                                  |
|  2 | DEPENDENT SUBQUERY | p                       | NULL       | unique_subquery | PRIMARY                                                                                               | PRIMARY    | 8       | func                          |    1 |    20.00 | Using where                                  |
+----+--------------------+-------------------------+------------+-----------------+-------------------------------------------------------------------------------------------------------+------------+---------+-------------------------------+------+----------+----------------------------------------------+
4 rows in set, 1 warning (0.01 sec)
```

If we are concerned about the performance of our database queries, and I think we should be, I suggest we create a task to investigate what are our options for all WooCommerce segments as I don't think there is an easy fix here. What do you think?

rodrigoprimo:
Thanks for your review, @pavel-mailpoet. I added a few commits based on your feedback and left some follow-up questions for you. I still need to add automated tests and i18n. Once that is done, I will ask for a final review.

> we tend to split our work into multiple commits. Especially such a big change that you made here. It makes the life of code reviewer easier

I think I disagree here. I tend to try to split commits into functional units as much as possible. In my opinion, splitting my initial commit would mean creating commits that independently are not functional. This might make reviewing the PR a bit harder, but I don't think by much. The main benefit that I see is that it makes investigating the changes here in the future much easier if, at some point, we need to debug something related to this code. Once we find a commit, we don't have to look for other commits to put together how the number of orders segment works. Happy to discuss this further.

> we don't merge master in our local branches, we rebase it so the merge commit is not shown in the pull request

I don't have a strong opinion here on whether I prefer rebasing or merging. I don't see a problem with the merge commit in the PR. I feel like I had done a merge already, and someone told me that I should rebase instead, but I completely forgot it when creating this PR. I will try to remember this for my next PRs. If we feel strongly for some reason that merges shouldn't be used in this case, maybe we can use the pre-merge git hook to display a confirmation message reminding developers that rebasing is preferable to prevent this from happening in the future. I think it is very likely that I will forget about it again :-)

lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
pavel-mailpoet: > If we are concerned about the performance of our database queries, and I think we should be, I suggest we create a task to investigate what are our options for all WooCommerce segments as I don't think there is an easy fix here. What do you think?

You're right. I didn't realise this is like that in other segments as well. So leave it for now, but yeah, we should revisit that later. I think it will be part of [these tasks](https://wordpress.com/read/feeds/113756048/posts/3267798944), right?

rodrigoprimo: I think it is related (here is the link to the task https://mailpoet.atlassian.net/browse/MAILPOET-3557), but I'm not sure it is the same thing. The task that I created is focused on describing how we are currently interacting with WooCommerce data and doesn't focus on the performance of the queries. Maybe we could create a related task to investigate our options to improve the performance of the queries used in the WooCommerce segments?

pavel-mailpoet: Sure we can do that. 

I was thinking that eventually, we want to switch from the SQL queries to Woocommerce functions calls. That would solve that. But I am not sure if that will happen or when, so creating a ticket is an option.

rodrigoprimo:
@pavel-mailpoet, I believe I implemented all your suggestions. Could you please review this PR again?

Regarding i18n, the string that needs to be translated in this PR is a bit more complex than the one that you had in the example that you shared on Slack. I opted to split the string into multiple substrings which is not ideal for translation but I did this because I felt like it would take too much time to implement something that would work in this case and that wouldn't feel too hacky. My suggestion is that we look into creating a generic solution for translating strings that we want to interpolate with React components but I believe this is outside of the scope of this PR (cc @NeosinneR).

tests/acceptance/WooCommerceDynamicSegmentsCest.php
```diff
@@ -101,6 +109,23 @@ public function addCustomerOnlyToCategorySegment(\AcceptanceTester $i) {
     $i->see('0', $productSegmentRow . " [data-colname='Number of subscribers']");
   }
 
+  public function checkThatCustomersAreAddedToNumberOfOrdersSegment(\AcceptanceTester $i) {
```
pavel-mailpoet, line 112: I can see you are checking here that the segment filters subscribers correctly. That is great and we definitely need that. But why did you decide to create the test using a DB call? Shall we create the segment in the UI? Just to check that the creating works?

tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
```diff
@@ -0,0 +1,115 @@
...
+    $this->orders[] = $this->createOrder(['user_id' => $userId3]);
+  }
+
+  public function testItGetsCustomersThatPlacedTwoOrdersInTheLastDay() {
+    $segmentFilter = $this->getSegmentFilter('=', 2, 1);
+    $queryBuilder = $this->numberOfOrders->apply($this->getQueryBuilder(), $segmentFilter);
+    $result = $queryBuilder->execute()->fetchAll();
+    expect(count($result))->equals(1);
+    $subscriber1 = $this->entityManager->find(SubscriberEntity::class, $result[0]['inner_subscriber_id']);
+    assert($subscriber1 instanceof SubscriberEntity);
```
pavel-mailpoet, line 37: Rosta told me that you told him that it is better to use `expect($subscriber1)->isInstanceOf(SubscriberEntity::class);` rather than `assert`. So why are you using `assert` in this test, is there a benefit here?

costasovo: I think `expect($subscriber1)->isInstanceOf(SubscriberEntity::class);` still doesn't work with PHPStan. `$this->asertInstanceOf` was the one with nicer error mesages.

lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
rodrigoprimo: Ticket created here: https://mailpoet.atlassian.net/browse/MAILPOET-3571

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -126,27 +126,27 @@ export const WooCommerceFields: React.FunctionComponent<Props> = ({ onChange, it
...
             value={item.number_of_orders_count || ''}
-            placeholder="count"
+            placeholder={MailPoet.I18n.t('wooNumberOfOrdersCount')}
             onChange={(event): void => compose([
               onChange,
               assign(item),
             ])({ number_of_orders_count: event.target.value })}
           />
-          <div>orders</div>
+          <div>{MailPoet.I18n.t('wooNumberOfOrdersOrders')}</div>
```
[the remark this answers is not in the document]
rodrigoprimo: Maybe you are not missing anything, and I'm the one that is missing something :-)

But I give what you suggested a try and could get to a half working solution where, to proceed, I would have to rely on the value of English words in the sentence and not the placeholders to add `<div>`s around them. That is obviously not an option, as those strings can change depending on the language. That is when I stopped and felt like it wasn't worth investing time to make this work in this particular PR and that we would probably need to come up with a more generic solution for strings that we want to interpolate with React components.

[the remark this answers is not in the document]
pavel-mailpoet: I think this function with the `split` above should work as intended, please if there is something wrong, let me know. I am not sure how to make a generic solution from this.
```
        (match) => {
             if (match === '{type}') {
               return (
                 <Select />
               );
             }
            if (match === '{count}') {
               return (
                 <Input />
               );
             }
            if (match === '{days}') {
               return (
                 <Input />
               );
             }
             if ((typeof match === 'string') && match.trim().length > 1) {
               return (
                 <div key={match}>
                   {match}
                 </div>
               );
             }
             return null;
           }
```

tests/acceptance/WooCommerceDynamicSegmentsCest.php
rodrigoprimo: I just followed the pattern that we are already using for the acceptance tests for the other WooCommerce related segments. I don't know what was the reasoning behind only testing that subscribing works correctly in the acceptance tests and not including a test to make sure that creating the segment works. I will look into adding a test covering the creation of the segment, and if it is straightforward, I will add a new commit. I just suggest that we don't hold this PR because of the lack of this test as tomorrow is my last day before my sabbatical, so today is basically the last day that I have to finish this while someone else from the team is around to approve it.

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -3,14 +3,14 @@ import MailPoet from 'mailpoet';
...
+import { Grid } from 'common/grid';
 import {
   OnFilterChange,
   SegmentTypes,
   SelectOption,
   WooCommerceFormItem,
 } from '../types';
 import { SegmentFormData } from '../segment_form_data';
-import { Grid } from '../../../common/grid';
 import Input from '../../../common/form/input/input';
```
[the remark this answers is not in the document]
rodrigoprimo: My bad. I didn't check if there were other instances where I had added an import with multiple `../`. This is now fixed.

tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
rodrigoprimo: Yeah, the error message of `$this->assertInstanceOf()` is more detailed and precise than the error message of `assert()`. I used `assert()` here because I based those tests on the UserRoleTest tests, and it didn't occur to me to replace `assert()` with `$this->assertInstanceOf()`. I pushed a commit that replaces all the generic assertions with more specific ones, not just for the case that you mentioned here.

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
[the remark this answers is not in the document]
rodrigoprimo: Thanks for sharing the example, @pavel-mailpoet. I got to this point but that is when I stopped. Unless I'm missing something, the example that you shared only works well for a single row. But in this case, we are splitting a sentence into two rows and that is when it gets tricky. We could split the string into two and then use the code above for each of the rows, but I don't know if that would be much better as the whole point is to let translators translate a single string and change the order of the input elements to match what makes more sense in their language. Do you see any solution to this that I'm seeing? If not, maybe we can proceed the way it is as @NeosinneR mentioned it in the ticket description as an option?

Also, @NeosinneR if we want to make this fully translatable maybe we should consider a different format for displaying the input fields? Maybe instead of interpolating them in the string, we create a row for each input field and use an individual string to describe each input field? Something like the example below but with a better wording:

```
Select the type of the comparison: <Select />
Select the number of orders: <Input />
Select the number of days: <Input />
```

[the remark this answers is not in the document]
pavel-mailpoet: Oh, that is what I was missing. The sentence is split into 2 rows. Yeah, I don't think we can code that.

tests/acceptance/WooCommerceDynamicSegmentsCest.php
rodrigoprimo: Just a quick update that I found that Rosta added new acceptance tests for creating and editing the other WooCommerce segments earlier this week (https://github.com/mailpoet/mailpoet/pull/3443/files#diff-dea08370546d62b9522564eef934ea96d12212880401ee358d09445b316b08d2R316), so I will use those to create a similar test for this segment.

rodrigoprimo:
Thanks again for your detailed review, @pavel-mailpoet. I believe I addressed all the points that you mentioned. Could you please check it again?

pavel-mailpoet (dismissed):

veljkho (approved):

veljkho (approved):

PR #3454, opened 2021-04-16
Update customer names on checkout using model [MAILPOET-3565]

costasovo:
[MAILPOET-3565]

[MAILPOET-3565]: https://mailpoet.atlassian.net/browse/MAILPOET-3565

### DIFF
--- a/lib/Segments/WooCommerce.php
+++ b/lib/Segments/WooCommerce.php
@@ -92,11 +92,14 @@ class WooCommerce {
           $data['source'] = Source::WOOCOMMERCE_USER;
         }
         $data['id'] = $subscriber->id();
-
+        if ($wpUser->first_name) { // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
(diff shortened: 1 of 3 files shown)


lib/Segments/WooCommerce.php
```diff
@@ -247,7 +259,7 @@ private function updateFirstNames($subscriberId = null) {
     global $wpdb;
     $collate = '';
     if ($this->needsCollationChange()) {
-      $collate = ' COLLATE ' . $this->mailpoetEmailCollation;
+        $collate = ' COLLATE ' . $this->mailpoetEmailCollation;
```
pavel-mailpoet, line 262: I think this change is not intended or is it?

pavel-mailpoet (changes requested):

lib/Segments/WooCommerce.php
costasovo: I've fixed it with a fixup commit.

pavel-mailpoet (approved):

veljkho (approved):

PR #3458, opened 2021-04-19
Mention Consent and Suppression lists in Subscriber Import [MAILPOET-3526]

pavel-mailpoet:
[MAILPOET-3526]

[MAILPOET-3526]: https://mailpoet.atlassian.net/browse/MAILPOET-3526


lysyjan (dismissed):

veljkho (changes requested):
Can we move those KBs inside a Beacon? To be opened via Beacon, I thought this was by default for all KBs.

Otherwise, looks good and no issues found. 👍

NeosinneR:
@pavel-mailpoet I updated the specs with Beacon IDs

lysyjan (approved):

veljkho (approved):

PR #3459, opened 2021-04-19
Add opt-in checkbox to WooCommerce registration form [MAILPOET-3483]

lysyjan:
[MAILPOET-3483]

[MAILPOET-3483]: https://mailpoet.atlassian.net/browse/MAILPOET-3483

### DIFF
--- a/lib/Config/Hooks.php
+++ b/lib/Config/Hooks.php
@@ -161,11 +161,11 @@ class Hooks {
       }
       $this->wp->addAction(
         'woocommerce_register_form',
-        [$this->subscriptionRegistration, 'extendForm']
+        [$this->hooksWooCommerce, 'extendForm']
       );
       $this->wp->addFilter(
         'woocommerce_registration_errors',
-        [$this->subscriptionRegistration, 'onRegister'],
+        [$this->hooksWooCommerce, 'onRegister'],
         60,
         3
       );
--- a/lib/Config/Hooks.php
+++ b/lib/Config/Hooks.php
@@ -159,6 +159,16 @@ class Hooks {
           3
         );
       }
+      $this->wp->addAction(
+        'woocommerce_register_form',
+        [$this->subscriptionRegistration, 'extendForm']
+      );
+      $this->wp->addFilter(
+        'woocommerce_registration_errors',
+        [$this->subscriptionRegistration, 'onRegister'],
+        60,
+        3
+      );
     }
 
     // Manage subscription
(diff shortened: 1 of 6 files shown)


pavel-mailpoet (dismissed):

veljkho (changes requested):
Wrong checkbox present, the one from the checkout... should be the one from the Settings > Basic .. `Yes, add me to your mailing list`

lib/Config/Hooks.php
```diff
@@ -159,6 +159,16 @@ public function setupSubscriptionEvents() {
           3
         );
       }
+      $this->wp->addAction(
+        'woocommerce_register_form',
+        [$this->subscriptionRegistration, 'extendForm']
```
pavel-mailpoet, line 164: I think this should go through WooCommerce hooks class so that it catches any errors, shouldn't it?

lysyjan:
For QA:
After discussing with Tautvidas, I changed the default state for subscribers when a checkbox in the registration form is unchecked. This change solves an unwanted confirmation email. At this moment, users added via `wp_insert_user` have status unsubscribed.

Let me know if you have any questions.

lib/Segments/WP.php
```diff
@@ -71,15 +71,13 @@ private function createOrUpdateSubscriber($currentFilter, $wpUser, $subscriber =
...
-    $signupConfirmationEnabled = SettingsController::getInstance()->get('signup_confirmation.enabled');
-    $status = $signupConfirmationEnabled ? Subscriber::STATUS_UNCONFIRMED : Subscriber::STATUS_SUBSCRIBED;
     // subscriber data
     $data = [
       'wp_user_id' => $wpUser->ID,
       'email' => $wpUser->user_email, // phpcs:ignore Squiz.NamingConventions.ValidVariableName.NotCamelCaps
       'first_name' => $firstName,
       'last_name' => $lastName,
-      'status' => $status,
+      'status' => SubscriberEntity::STATUS_UNSUBSCRIBED,
```
flakas, line 80: @lysyjan it looks like the WC user signup is handled.
But because of this change, if a WP user signs up via the WordPress user registration form, and they check the "I want to subscribe" checkbox - do they get handled as if they want to subscribe? (added as unconfirmed, sent a confirmation email, scheduled welcome emails)

Based on the code below it doesn't appear to be the case, but maybe the consent check is handled elsewhere?

lysyjan: @flakas Everything works as we expect. I have tested it again. I have tried the welcome email, the confirmation email, and I have checked the subscriber status.

The checkbox from the registration form is managed in the [Registration.php](https://github.com/mailpoet/mailpoet/blob/master/lib/Subscription/Registration.php#L58-L74).

pavel-mailpoet (approved):

veljkho (approved):

PR #3465, opened 2021-04-20
Rebranding buttons [MAILPOET-3545]

NeosinneR:
[MAILPOET-3545]

[MAILPOET-3545]: https://mailpoet.atlassian.net/browse/MAILPOET-3545


NeosinneR:
emails
export
help
import-1
import-2
key-verification
segment
send-with
storybook-1
storybook-2
subscribers

NeosinneR:
@j111q I did some changes (fewer primary buttons on a single page) and here are updated screenshots. Let me know if you have more suggestions. 

Screenshot 2021-05-02 at 16 56 55
Screenshot 2021-05-02 at 16 58 12
Screenshot 2021-05-02 at 16 58 28
Screenshot 2021-05-02 at 16 58 36
Screenshot 2021-05-02 at 16 58 49
Screenshot 2021-05-02 at 16 59 13
Screenshot 2021-05-02 at 17 00 52
Screenshot 2021-05-02 at 17 01 05
Screenshot 2021-05-02 at 17 01 16
Screenshot 2021-05-02 at 17 01 25
Screenshot 2021-05-02 at 17 02 12
story1
story2

j111q:
Sorry for the late review here @NeosinneR!

Looking good overall! 👍 A few notes:

**Let's switch always using a `Secondary` button when there's button(s) in the header.**

As you mentioned, those buttons are all equal in importance, but when they're all Primary buttons, they do look very heavy and visually dense. So if we switch them all to secondary buttons, it will work better. Specifically for the screenshots you gave above, I'm referring to the Subscribers, Forms, and Lists pages - all the buttons in the header next to the Mailpoet logo:

image
image
image


---

**One more thing** - In case it wasn't clear before - the 'blue' should change according to the theme colour of wp-admin [as chosen by the user like](https://www.siteground.com/tutorials/wordpress/color-scheme/) in this screen:

image

Let me know if you have any other questions or thoughts around these^ comments! Looking great otherwise :) 👍

j111q (approved):

NeosinneR:
@j111q I finally got to fixing those two issues (secondary buttons in header and themed buttons) and here are screenshots :) I'll let a developer review the PR so it can be merged 

Screenshot 2021-06-02 at 12 06 11

Screenshot 2021-06-02 at 11 16 18

NeosinneR:
@costasovo, there is also a small change in the premium plugin. One shortcoming of this is that in Storybook.js, all button variants look the same because I couldn't import WordPress styles in Storybook.js (they are generated on the fly, and developers can have a different domain where their local WordPress is running).

j111q:
Awesome thank you! Looks great! (though I personally hope there are very few users who will choose a purple and green colour palette 😆 )

assets/css/src/generic/_buttons.scss
```diff
@@ -1,29 +1,38 @@
...
+  }
+
   &:focus {
-    background: $color-secondary-hover;
+    background: $color-primary-hover;
+    border-color: #fff;
+    box-shadow: 0 0 0 1.5px $color-primary-hover;
     color: #fff;
+    outline: none;
     text-decoration: none;
```
costasovo: Do we need to keep all those styles? After adding WordPress button classes (`button-secondary` etc.) most of these rules are overwritten by CSS from WP. Does it makes sense to keep them?

Screen Shot on 2021-06-03 at 11-21-48

costasovo (changes requested):
Good job! 👍 

I added one question (see comment) + I found one incorrectly styled button on the subscriber's stats page.
Screen Shot on 2021-06-03 at 11-23-12

NeosinneR:
@costasovo good catch! Button fixed. 

For the redundant styles - I kept them there on purpose as a fallback. It's possible that they aren't really needed, and `.button-*` classes will always be present in WordPress. But if for any reason they are not, or they are changed in the future, MailPoet buttons would still look usable. 

But I don't have a strong opinion on this, and I'm open to removing them, as it's probably highly unlikely any of that will happen. What do you think?

costasovo:
> For the redundant styles - I kept them there on purpose as a fallback. It's possible that they aren't really needed, and .button-* classes will always be present in WordPress. But if for any reason they are not, or they are changed in the future, MailPoet buttons would still look usable.

> But I don't have a strong opinion on this, and I'm open to removing them, as it's probably highly unlikely any of that will happen. What do you think?

I don't have a strong opinion either 😄  I lean slightly towards removing them. I can think of 3 reasons:

1. We want every button to have WordPress's `.button-*` so that they respect admin theme colors. With the fallback styles, a button may look OKish even without the class so a developer may forget to add it.
2. We may by accident increase specificity and those fallback styles may override WP's styles.
3. The existence of those styles makes a developer think about why we have them.

NeosinneR:
Fair enough. I removed the styles and also fixed an issue with small buttons. And added a generated CSS, so the Storybook.js buttons look the same as in the plugin.

costasovo (dismissed):
Thanks!

veljkho (changes requested):
1. For some reason there are 2 same buttons on Post Notification listing page, [View History](https://d.pr/i/MgSorg) buttons. This should be caused by rebranding buttons.

2. Can we [align Create Segment button](https://d.pr/i/MkrjCO) with filter dropdown on the left?

3. Can we make the `Empty Trash` [button a bit bigger so it fits nicely](https://d.pr/i/QwiNEK) with the left dropdown size? Actually height is not good.

NeosinneR:
@veljkho 
1. Fixed
2. Fixed
3. The button is bigger now, but it's still 2px smaller than the inputs on the left. This will be fixed once the rest of the form controls is also restyled, but for now, there will be this 2px difference when the input and button are next to each other.

veljkho (approved):

PR #3466, opened 2021-04-21
Add a new segment: total spent [MAILPOET-3227]

lysyjan:
[MAILPOET-3227]

[MAILPOET-3227]: https://mailpoet.atlassian.net/browse/MAILPOET-3227


assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -43,6 +45,9 @@ export function validateWooCommerce(formItems: WooCommerceFormItem): boolean {
   if (formItems.action === 'numberOfOrders' && (!formItems.number_of_orders_count || !formItems.number_of_orders_days || !formItems.number_of_orders_type)) {
     return false;
   }
+  if (formItems.action === 'totalSpent' && (!formItems.total_spent_amount || !formItems.total_spent_days || !formItems.total_spent_type)) {
```
pavel-mailpoet, line 48: Should we use the enum form above here? Instead of using the strings `'totalSpent'`?

views/segments.html
```diff
@@ -173,6 +173,14 @@
     'wooNumberOfOrdersInTheLast': _x('in the last', 'Appears together with wooNumberOfOrdersDays when creating a new WooCommerce segment based on the number of orders.'),
     'wooPurchasedCategory': __('purchased in this category'),
     'wooPurchasedProduct': __('purchased this product'),
+    'wooTotalSpent': __('total spent'),
+    'wooTotalSpentMoreThan': __('more than'),
```
pavel-mailpoet, line 178: Do we need this here? we already have __('more than') a couple of lines above. Also, other translations are here already that could be re-used.

views/segments.html
```diff
@@ -173,6 +173,14 @@
...
+    'wooTotalSpentLessThan': __('less than'),
+    'wooTotalSpentAmount': __('amount'),
+    'wooTotalSpentDaysPlaceholder': __('days'),
+    'wooTotalSpentDays': _x('days', 'Appears together with wooTotalSpentInTheLast when creating a new WooCommerce segment based on the total spent money.'),
+    'wooTotalSpentCurrency': __('$'),
```
pavel-mailpoet, line 183: I am not sure what can translators do about this. Do we need to translate this? And if so how? At least there should be a comment, but I am not sure about it.

lib/WooCommerce/Helper.php
```diff
@@ -35,6 +35,10 @@ public function getWoocommerceCurrency() {
     return get_woocommerce_currency();
   }
 
+  public function getWoocommerceCurrencySymbol() {
+    return get_woocommerce_currency_symbol();
```
pavel-mailpoet, line 39: What happens when the store uses multi currencies?

lib/Segments/DynamicSegments/Filters/WooCommerceTotalSpent.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace MailPoet\Segments\DynamicSegments\Filters;
+
+use MailPoet\Entities\DynamicSegmentFilterEntity;
+use MailPoet\Entities\SubscriberEntity;
+use MailPoetVendor\Carbon\Carbon;
+use MailPoetVendor\Doctrine\DBAL\Query\QueryBuilder;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
+
+class WooCommerceTotalSpent implements Filter {
```
pavel-mailpoet, line 11: Should we create a test for this class? Similar to `WooCommerceNumberOfOrdersTest`? I am not sure about that so that is why I am asking you about your opinion.

tests/acceptance/WooCommerceDynamicSegmentsCest.php
```diff
@@ -58,6 +62,10 @@ public function _before(\AcceptanceTester $i) {
       ->withName(self::NUMBER_OF_ORDERS_SEGMENT)
       ->withWooCommerceNumberOfOrdersFilter()
       ->create();
+    $this->totalSpentSegment = $segmentFactory
```
pavel-mailpoet, line 65: Why did you decide to only test if the created segment works? Would it be a better test if we created the segment using the UI?

pavel-mailpoet (commented):
Good job, I have a couple of questions.

views/segments.html
lysyjan: I thought about this, and I wouldn't say I like that the key will contain a name from the other type. So, should I use the key with the original name, for example, `wooNumberOfOrdersMoreThan`?

views/segments.html
lysyjan: This is the rest of the first code. I will remove it. Thanks

views/segments.html
pavel-mailpoet: I'd rename the key. Maybe just `moreThan`?

lysyjan: Yeah, that sounds fair. I wasn't sure if I can rename some existing keys.

lib/Segments/DynamicSegments/Filters/WooCommerceTotalSpent.php
lysyjan: Yes, it's possible. I will have to add some metadata about order value. Good idea.

tests/acceptance/WooCommerceDynamicSegmentsCest.php
lysyjan: I knew about this test only. I will add a test for creation too.

lib/WooCommerce/Helper.php
lysyjan: Good question. I will find out how multiple currencies work and let you know.

NeosinneR: I don't think WooCommerce supports multiple currencies by default, you need a plugin for that (and there is plenty of them), so I'm not sure if we want to support each plugin specifically.

tests/acceptance/WooCommerceDynamicSegmentsCest.php
pavel-mailpoet: Thanks. Also, to save some time during the test execution I'd recommend doing everything in a single test and not adding another test. 
Check Rodrigo's segment he did a similar test.

lib/WooCommerce/Helper.php
pavel-mailpoet: True. There is one on our marketplace: https://woocommerce.com/products/multi-currency/# 

We don't need to support multiple currencies. But at least we should know what happens and handle that gracefully.

assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
lysyjan: I used the same pattern as is on the lines above. I will change it.

lib/WooCommerce/Helper.php
lysyjan: I have tested the multi-currency plugin, and everything works well because the plugin recalculates prices for products on the page shop with the rate to the main currency. It means that this function returns the main currency, which is configured in WooCommerce.

pavel-mailpoet (approved):

veljkho (approved):

PR #3467, opened 2021-04-21
Visualise subscribers engagement [MAILPOET-3525]

pavel-mailpoet:
[MAILPOET-3525]

[MAILPOET-3525]: https://mailpoet.atlassian.net/browse/MAILPOET-3525


lib/API/JSON/v1/Settings.php
```diff
@@ -81,6 +105,17 @@ public function set($settings = []) {
     }
   }
 
+  public function recalculateSubscribersScore() {
+    $this->statisticsOpensRepository->resetSubscribersScoreCalculation();
+    $task = new ScheduledTaskEntity();
+    $task->setType(SubscribersEngagementScore::TASK_TYPE);
+    $task->setScheduledAt(Carbon::createFromTimestamp($this->wp->currentTime('timestamp')));
+    $task->setStatus(ScheduledTaskEntity::STATUS_SCHEDULED);
+    $this->entityManager->persist($task);
```
lysyjan, line 114: Is there a reason why you used the entityManager instead of the specific repository? I thought that we should use a repository when we can. I'm only curious.

lib/API/JSON/v1/Settings.php
```diff
@@ -30,22 +35,41 @@ class Settings extends APIEndpoint {
...
   private $servicesChecker;
 
+  /** @var WPFunctions */
+  private $wp;
+
   public $permissions = [
     'global' => AccessControl::PERMISSION_MANAGE_SETTINGS,
   ];
 
+  /**
```
lysyjan, line 45: I noticed that you mixed one-line annotations and multi-line annotations. It's a detail only.

lysyjan (commented):
Nice, in my opinion, it looks good. I have only a few small questions/notes.

lib/Statistics/StatisticsOpensRepository.php
```diff
@@ -0,0 +1,51 @@
...
+      ->from(ScheduledTaskSubscriberEntity::class, 'scheduledTaskSubscriber')
+      ->where('scheduledTaskSubscriber.subscriber = :subscriber')
+      ->setParameter('subscriber', $subscriber)
+      ->join('scheduledTaskSubscriber.task', 'task')
+      ->andWhere('task.type = :sending')
+      ->setParameter('sending', Sending::TASK_TYPE)
+      ->getQuery()
+      ->getSingleScalarResult();
+    if ($newslettersSentCount < 3) {
+      $this->entityManager->flush();
```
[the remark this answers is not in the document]
pavel-mailpoet: There is a change on line 20.

assets/js/src/common/tag/tag.tsx
```diff
@@ -3,7 +3,7 @@ import classnames from 'classnames';
 
 type Props = {
   children?: React.ReactNode;
-  variant?: 'average' | 'good' | 'excellent' | 'list';
+  variant?: 'average' | 'good' | 'excellent' | 'list' | 'unknown';
```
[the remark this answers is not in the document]
pavel-mailpoet: Because `badge` uses `tag` internally. If I didn't add this here Typescript would yell at me that the `unknown` value is not allowed.

lib/API/JSON/v1/Settings.php
pavel-mailpoet: I didn't know you can do that. I thought you need to call `persist` on the entity manager. Can I use any repository or a specific one?

lysyjan: > Can I use any repository or a specific one?
Yes, you can call persist and flush on the specific repository. I think that there is a check on the specific entity. You don't have to change it now. I was only curious.

lysyjan (approved):

veljkho (approved):

PR #3469, opened 2021-04-21
WooCommerce versions in test [MAILPOET-3482]

costasovo:
This PR contains a few more changes.

- I've refactored download the WooCommerce plugin so that it uses the same mechanism as WooCommere Subscriptions
- Changed directory where we download the 3rd party plugins
- Updated version of the image we use on Circle CI machine executor for acceptance tests. The image version refers to AWS Ami that is available. We were using some old default Ubuntu 14 with a quite old docker version.
- Fixed a couple of flaky acceptance test I run into during testing

[MAILPOET-3482]

[MAILPOET-3482]: https://mailpoet.atlassian.net/browse/MAILPOET-3482

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,6 +172,7 @@ jobs:
       - run:
           name: Download additional WP Plugins for tests
           command: |
+            ./do download:woo-commerce-zip 5.2.2
             ./do download:woo-commerce-subscriptions-zip 3.0.14
       - run:
           name: Dump tests ENV variables for acceptance tests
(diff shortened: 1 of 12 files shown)


pavel-mailpoet (commented):
Should we add documentation in the readme or somewhere?

costasovo:
> Should we add documentation in the readme or somewhere?

@pavel-mailpoet Agree. It is good to document things. I guess we could document the new commands for downloading plugins. Is there anything else that might need documenting?

pavel-mailpoet (approved):

veljkho (approved):

PR #3475, opened 2021-04-27
Subscribed date dynamic segment [MAILPOET-3222]

pavel-mailpoet:
[MAILPOET-3222]

[MAILPOET-3222]: https://mailpoet.atlassian.net/browse/MAILPOET-3222

### DIFF
--- a/lib/Segments/DynamicSegments/FilterDataMapper.php
+++ b/lib/Segments/DynamicSegments/FilterDataMapper.php
@@ -49,7 +49,7 @@ class FilterDataMapper {
         'segmentType' => DynamicSegmentFilterData::TYPE_USER_ROLE,
         'action' => $data['action'],
         'value' => $data['value'],
-        'operator' => $data['operator'] ?? 'before',
+        'operator' => $data['operator'] ?? SubscriberSubscribedDate::BEFORE,
       ]);
     }
     if (empty($data['wordpressRole'])) throw new InvalidFilterException('Missing role', InvalidFilterException::MISSING_ROLE);
--- a/lib/Segments/DynamicSegments/FilterDataMapper.php
+++ b/lib/Segments/DynamicSegments/FilterDataMapper.php
@@ -6,6 +6,7 @@ use MailPoet\Entities\DynamicSegmentFilterData;
 use MailPoet\Segments\DynamicSegments\Exceptions\InvalidFilterException;
 use MailPoet\Segments\DynamicSegments\Filters\EmailAction;
 use MailPoet\Segments\DynamicSegments\Filters\EmailOpensAbsoluteCountAction;
+use MailPoet\Segments\DynamicSegments\Filters\SubscriberSubscribedDate;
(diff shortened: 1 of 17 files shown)


lib/Segments/DynamicSegments/FilterDataMapper.php
```diff
@@ -42,6 +39,26 @@ private function getSegmentType(array $data): string {
...
+        'value' => $data['value'],
+        'operator' => $data['operator'] ?? 'before',
+      ]);
+    }
+    if (empty($data['wordpressRole'])) throw new InvalidFilterException('Missing role', InvalidFilterException::MISSING_ROLE);
+    return new DynamicSegmentFilterData([
+      'segmentType' => DynamicSegmentFilterData::TYPE_USER_ROLE,
+      'wordpressRole' => $data['wordpressRole'],
+    ]);
+  }
```
costasovo, line 60: Since the logic for mapping use role type segment became more complex I guess it would be nice to add a few test cases in related unit test https://github.com/mailpoet/mailpoet/blob/master/tests/unit/Segments/DynamicSegments/FilterDataMapperTest.php

lib/Segments/DynamicSegments/Filters/SubscriberSubscribedDate.php
```diff
@@ -21,13 +21,13 @@ public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $fi
     $operator = $filterData->getParam('operator');
 
     if ($operator === self::BEFORE) {
-      $queryBuilder->andWhere('last_subscribed_at < :date'. $filter->getId());
+      $queryBuilder->andWhere('last_subscribed_at < :date' . $filter->getId());
```
costasovo, line 24: Maybe using a variable for the parameter name would make the code more readable.

```
$dateParameter = 'date' . $filter->getId();
$queryBuilder->andWhere("last_subscribed_at < :$dateParameter");
```

costasovo (changes requested):
Good job! I added 2 small notices + I noticed that when switching between before/in last days (one requires date second number of days) there is a bad request sent to the plugins API (https://d.pr/i/fvvVnz)

costasovo (approved):

veljkho (changes requested):
Not verified

(1) Point 2.b says no default value.. but [it has default value of 1](https://d.pr/i/f1zTXe). This is important because otherwise we will hide 'days' placeholder and users wouldn't know what value is supposed to be.

(2) Point 2.c [is missing](https://d.pr/i/cFtKIG).

(3) Issue with not today's date [when changing from last days to specific date](https://d.pr/i/CDNHIV), but must follow these steps to reproduce:
1. First create a segment with subscribers `in the last 10 days`
2. Then edit it and change to `before` or `after`, and observe issue

All other points are verified and working as expected.

costasovo (approved):

veljkho (approved):

PR #3483, opened 2021-04-29
Fix subscribers score calculation [MAILPOET-3585]

lysyjan:
[MAILPOET-3585]

[MAILPOET-3585]: https://mailpoet.atlassian.net/browse/MAILPOET-3585

### DIFF
--- a/lib/Cron/Workers/SubscribersEngagementScore.php
+++ b/lib/Cron/Workers/SubscribersEngagementScore.php
@@ -28,21 +28,31 @@ class SubscribersEngagementScore extends SimpleWorker {
   }
 
   public function processTaskStrategy(ScheduledTask $task, $timer) {
-    $subscribers = $this->subscribersRepository->findBy(
-      ['engagementScoreUpdatedAt' => null],
-      [],
-      SubscribersEngagementScore::BATCH_SIZE
-    );
+    $subscribers = $this->subscribersRepository->findByUpdatedScoreNotInLastMonth(SubscribersEngagementScore::BATCH_SIZE);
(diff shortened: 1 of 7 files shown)


lib/Cron/Workers/SubscribersEngagementScore.php
```diff
@@ -28,21 +28,31 @@ public function __construct(
...
     if ($subscribers) {
+      $this->scheduleImmediately();
+    } else {
       $this->schedule();
     }
     return true;
   }
 
-  public function getNextRunDate() {
+  public function scheduleImmediately(): void {
```
pavel-mailpoet, line 43: Should we make this method `protected` and add it to the parent class?

lib/Cron/Workers/SubscribersEngagementScore.php
```diff
@@ -28,21 +28,31 @@ public function __construct(
...
+
+  public function getNextRunDateImmediately(): Carbon {
     return Carbon::createFromTimestamp($this->wp->currentTime('timestamp'));
   }
+
+  public function getNextRunDate() {
+    // random day of the next week
+    $date = Carbon::createFromTimestamp($this->wp->currentTime('timestamp'));
+    $date->addDay();
+    $date->setTime(mt_rand(0, 23), mt_rand(0, 59));
```
pavel-mailpoet, line 55: do we need to use `mt_rand` here? I think it is way slower than `rand` and we don't need the extra randomness it provides, do we?

pavel-mailpoet (approved):

veljkho (approved):

PR #3490, opened 2021-04-30
Add throttling for a sending batch size [MAILPOET-3588]

lysyjan:
[MAILPOET-3588]

[MAILPOET-3588]: https://mailpoet.atlassian.net/browse/MAILPOET-3588

### DIFF
--- a/lib/Cron/Workers/SendingQueue/SendingThrottlingHandler.php
+++ b/lib/Cron/Workers/SendingQueue/SendingThrottlingHandler.php
@@ -47,7 +47,9 @@ class SendingThrottlingHandler {
     $batchSize = $this->getBatchSize();
     if ($batchSize > 1) {
       $batchSize = (int)ceil($this->getBatchSize() / 2);
+      $throttlingSettings = $this->loadSettings();
       $throttlingSettings['batch_size'] = $batchSize;
+      unset($throttlingSettings['success_count']);
       $this->logger->error("MailPoet throttling: decrease batch_size to: {$batchSize}");
       $this->saveSettings($throttlingSettings);
     }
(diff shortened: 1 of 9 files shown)


lib/Cron/Workers/SendingQueue/SendingThrottlingHandler.php
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  public function processSuccess(): void {
+    $throttlingSettings = $this->loadSettings();
+    if (!isset($throttlingSettings['batch_size'])) {
+      return;
+    }
+    $throttlingSettings['success_in_row'] = isset($throttlingSettings['success_in_row']) ? ++$throttlingSettings['success_in_row'] : 1;
+    $this->logger->info("MailPoet throttling: increase success_in_row to: {$throttlingSettings['success_in_row']}");
+    if ($throttlingSettings['success_in_row'] === self::SUCCESS_THRESHOLD_TO_INCREASE) {
```
costasovo, line 65: I would rather use `>=`. Right now === seems to be safe but we may introduce some changes in the future combined with the fact cron may run by accident in parallel and if we for some reason get over SUCCESS_THRESHOLD_TO_INCREASE we would never return to a bigger batch.

tests/integration/Cron/Workers/SendingQueue/SendingThrottlingHandlerTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace MailPoet\Test\Cron\Workers;
+
+use MailPoet\Cron\Workers\SendingQueue\SendingThrottlingHandler;
+use MailPoet\Settings\SettingsController;
+
+class SendingThrottlingHandlerTest extends \MailPoetTest {
```
costasovo, line 8: I think we should add test cases for increasing batch size after certain amount of successes and maybe also that the lowest possible batch size is 1.

costasovo (changes requested):
Good job! I have just a few remarks.

costasovo (approved):
👍

veljkho (approved):

PR #3491, opened 2021-05-03
Fix double escaped button in newsletter [MAILPOET-3538]

costasovo:
I've fixed this by modifying API data sanitiser for newsletter so that it fixes text property only for specific blocks, that are using TinyMCE for editing.

[MAILPOET-3538]

[MAILPOET-3538]: https://mailpoet.atlassian.net/browse/MAILPOET-3538


lysyjan (approved):
Good job!

veljkho (approved):

PR #3501, opened 2021-05-07
Add list score [MAILPOET-3533]

lysyjan:
[MAILPOET-3533]

I have added as a part of this PR some small UI changes discussed with Tautvidas.

[MAILPOET-3533]: https://mailpoet.atlassian.net/browse/MAILPOET-3533

### DIFF
--- a/assets/js/src/segments/list.jsx
+++ b/assets/js/src/segments/list.jsx
@@ -24,7 +24,7 @@ const columns = [
   },
   {
     name: 'average_subscriber_score',
-    label: MailPoet.I18n.t('averageScore'),
+    label: MailPoet.I18n.t('listScore'),
     display: mailpoetTrackingEnabled,
   },
   {
--- a/assets/js/src/segments/list.jsx
+++ b/assets/js/src/segments/list.jsx
@@ -270,7 +270,7 @@ class SegmentList extends React.Component {
         <td data-colname={MailPoet.I18n.t('description')}>
           <abbr>{ segment.description }</abbr>
         </td>
-        { (mailpoetTrackingEnabled === true) ? (
+        { (mailpoetTrackingEnabled) ? (
           <td className="column mailpoet-listing-stats-column" data-colname={MailPoet.I18n.t('averageScore')}>
             <div className="mailpoet-listing-stats">
               <ListingsEngagementScore
--- a/assets/js/src/segments/list.jsx
+++ b/assets/js/src/segments/list.jsx
@@ -5,10 +5,12 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 
 import Listing from 'listing/listing.jsx';
+import { ListingsEngagementScore } from '../subscribers/listings_engagement_score';
 
 const isWPUsersSegment = (segment) => segment.type === 'wp_users';
 const isWooCommerceCustomersSegment = (segment) => segment.type === 'woocommerce_users';
 const isSpecialSegment = (segmt) => isWPUsersSegment(segmt) || isWooCommerceCustomersSegment(segmt);
+const mailpoetTrackingEnabled = (!!(window.mailpoet_tracking_enabled));
 
 const columns = [
   {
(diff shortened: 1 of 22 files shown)


assets/js/src/segments/list.jsx
```diff
@@ -263,6 +270,16 @@ class SegmentList extends React.Component {
         <td data-colname={MailPoet.I18n.t('description')}>
           <abbr>{ segment.description }</abbr>
         </td>
+        { (mailpoetTrackingEnabled === true) ? (
```
costasovo, line 273: Just a detail but it caught my eye. Why can't use just `if (mailpoetTrackingEnabled)` as we did on the line 28 (`display: mailpoetTrackingEnabled`) ?

lib/Statistics/StatisticsOpensRepository.php
```diff
@@ -58,4 +59,54 @@ public function resetSubscribersScoreCalculation() {
...
+      ->andWhere('subscriber.engagementScore IS NOT NULL')
+      ->setParameter('segment', $segment)
+      ->setParameter('subscribed', SubscriberEntity::STATUS_SUBSCRIBED)
+      ->getQuery()
+      ->getSingleScalarResult();
+    $segment->setAverageEngagementScore((float)$avgScore);
+    $this->entityManager->flush();
+  }
+
+  public function resetNewslettersScoreCalculation(): void {
```
costasovo, line 106: I think the name should be `resetSegmentsScoreCalculation` as it resets data for SegmentEntity.

lib/Statistics/StatisticsOpensRepository.php
```diff
@@ -58,4 +59,54 @@ public function resetSubscribersScoreCalculation() {
...
+    $this->entityManager->createQueryBuilder()->update(SegmentEntity::class, 's')
+      ->set('s.averageEngagementScoreUpdatedAt', ':verified')
```
costasovo, line 108: I'm not sure why the parameter name is `:verified`. It would be nicer to name it `:updatedAt` or similar so that it doesn't make other devs think about it.

tests/acceptance.suite.yml
```diff
@@ -27,7 +27,7 @@ modules:
         url: 'http://test.local'
         browser: chrome
         port: 4444
-        window_size: '1280x980'
+        window_size: '1366x980'
```
costasovo, line 30: Just curious. Why we need to to this?

costasovo (changes requested):
Nice job! I added a couple of small remarks.

lib/API/JSON/v1/Settings.php
```diff
@@ -104,10 +110,17 @@ public function set($settings = []) {
...
+      'type' => SubscribersEngagementScore::TASK_TYPE,
+      'status' => ScheduledTaskEntity::STATUS_SCHEDULED,
+    ]);
+    if ($task) {
+      $task->setScheduledAt(Carbon::createFromTimestamp($this->wp->currentTime('timestamp')));
+    } else {
+      $task = new ScheduledTaskEntity();
+      $task->setType(SubscribersEngagementScore::TASK_TYPE);
+      $task->setStatus(ScheduledTaskEntity::STATUS_SCHEDULED);
+    }
```
costasovo, line 123: It looks like that a newly created task has not scheduledAt set. I guess it could be done the other way around.
```php
if (!$task) {
  $task = new ScheduledTaskEntity();
  $task->setType(SubscribersEngagementScore::TASK_TYPE);
  $task->setStatus(ScheduledTaskEntity::STATUS_SCHEDULED);
} 
$task->setScheduledAt(Carbon::createFromTimestamp($this->wp->currentTime('timestamp')));
```

tests/acceptance.suite.yml
costasovo: Aha I see. It makes sense. It looks like that after the addition the table became quite big. I don't know how many users use 1280px wide screens, but as I see from [the screenshot](https://d.pr/i/bH9mp7) it may at least look a bit broken for such users.
Maybe we could compress the information about subscribers counts into one column to save some space, but that is out of scope of this task. I guess @veljkho could check how the table works for users with smaller screens.

costasovo (changes requested):

lib/API/JSON/v1/Settings.php
lysyjan: Right, it was my original idea, but I don't know why I didn't make it. Thank you

costasovo (approved):

veljkho (dismissed):

veljkho (approved):

PR #3510, opened 2021-05-13
Use SubscribersActions for subscribe via manage subscriptions [MAILPOET-2496]

lysyjan:
[MAILPOET-2496]

[MAILPOET-2496]: https://mailpoet.atlassian.net/browse/MAILPOET-2496

### DIFF
--- a/lib/Subscription/Manage.php
+++ b/lib/Subscription/Manage.php
@@ -4,13 +4,18 @@ namespace MailPoet\Subscription;
 
 use MailPoet\Entities\StatisticsUnsubscribeEntity;
 use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Entities\SubscriberSegmentEntity;
 use MailPoet\Form\Util\FieldNameObfuscator;
 use MailPoet\Models\CustomField;
+use MailPoet\Models\Segment;
 use MailPoet\Models\Subscriber;
 use MailPoet\Models\SubscriberSegment;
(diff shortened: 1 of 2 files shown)


lib/Subscription/Manage.php
```diff
@@ -111,12 +117,8 @@ private function updateSubscriptions(Subscriber $subscriber, array $subscriberDa
...
       $segmentsIds = array_intersect($segmentsIds, $allowedSegments);
     }
-    foreach ($segmentsIds as $segmentId) {
-      SubscriberSegment::createOrUpdate([
-        'subscriber_id' => $subscriber->id,
-        'segment_id' => $segmentId,
-        'status' => Subscriber::STATUS_SUBSCRIBED,
-      ]);
+    if ($segmentsIds) {
+      $this->subscriberActions->subscribe($subscriberData, $segmentsIds);
```
costasovo, line 121: I think we can't use `$this->subscriberActions->subscribe` in this case. If we keep it as is it would trigger subscription confirmation email, new subscriber notification email and also welcome email scheduling each time the subscriber changes something in the manage subscription form.
The specification mentions that we should schedule the welcome email when subscriber subscribes on manage subscription page. I think that this means that we should do these actions only when a subscriber subscribes to a new list.

costasovo (changes requested):

costasovo (approved):
👍

veljkho (approved):

PR #3512, opened 2021-05-14
Segment MailPoet custom field [MAILPOET-3220]

pavel-mailpoet:
[MAILPOET-3220]

[MAILPOET-3220]: https://mailpoet.atlassian.net/browse/MAILPOET-3220

### DIFF
--- a/assets/js/src/segments/dynamic/dynamic_segments_filters/custom_fields/date.tsx
+++ b/assets/js/src/segments/dynamic/dynamic_segments_filters/custom_fields/date.tsx
@@ -173,6 +173,47 @@ const DateFullDate = ({ onChange, item }: ComponentProps) => {
   );
 };
 
+const DateMonthYear = ({ onChange, item }: ComponentProps) => {
+  useEffect(() => {
+    if ((item.value === undefined) || item.value === '') {
+      onChange(
+        assign(item, {
+          value: `${format(new Date(), 'yyyy-MM-dd')} 00:00:00`,
+          operator: 'equals',
+        })
+      );
+    }
+  }, [onChange, item]);
+
+  return (
+    <Grid.CenteredRow>
+      <Select
+        key="select-operator"
+        value={item.operator}
+        onChange={compose([
+          onChange,
+          assign(item),
+          set('operator', __, {}),
+          get('value'),
+          get('target'),
+        ])}
+      >
+        <option value="equals">{MailPoet.I18n.t('equals')}</option>
+        <option value="before">{MailPoet.I18n.t('before')}</option>
+        <option value="after">{MailPoet.I18n.t('after')}</option>
+      </Select>
+      <Datepicker
+        onChange={(value): void => onChange(
+          assign(item, { value: convertDateToString(value) })
+        )}
+        selected={item.value ? parseDate(item.value) : undefined}
+        dateFormat="MM/yyyy"
+        showMonthYearPicker
+      />
+    </Grid.CenteredRow>
+  );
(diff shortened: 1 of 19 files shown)


assets/js/src/segments/dynamic/dynamic_segments_filters/custom_fields/text.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <>
+      <div className="mailpoet-gap" />
+      <Grid.CenteredRow>
+        <Select
+          key="select"
+          value={item.operator}
+          onChange={compose([
+            onChange,
+            assign(item),
+            set('operator', __, {}),
```
costasovo, line 53: Just to make sure I understand this. __ will be replaced by the value returned from previous function, correct? I tried to google it but the documentation for the library is either hard to find or non-existing.

assets/js/src/segments/dynamic/dynamic_segments_filters/custom_fields/date.tsx
```diff
@@ -0,0 +1,228 @@
...
+          set('operator', __, {}),
+          get('value'),
+          get('target'),
+        ])}
+      >
+        <option value="equals">{MailPoet.I18n.t('equals')}</option>
+        <option value="before">{MailPoet.I18n.t('before')}</option>
+        <option value="after">{MailPoet.I18n.t('after')}</option>
+      </Select>
+      <Datepicker
```
costasovo, line 165: We render the same Datepicker for both `year_mont_day` and `year_month` date types. The date picker library can be configured as a month picker (see [demo](https://reactdatepicker.com/#example-month-picker)). I guess it would improve UX if we use the month picker for the year_moth date type.

costasovo (changes requested):
Good job! I have just a couple of suggestions.

assets/js/src/segments/dynamic/dynamic_segments_filters/custom_fields/date.tsx
pavel-mailpoet: @NeosinneR I guess this is a question for you.

assets/js/src/segments/dynamic/dynamic_segments_filters/custom_fields/text.tsx
pavel-mailpoet: Correct. I know it from this [library](https://ramdajs.com/docs/#__). Here is something from [lodash](https://github.com/lodash/lodash/wiki/FP-Guide#placeholders)

assets/js/src/segments/dynamic/dynamic_segments_filters/custom_fields/date.tsx
NeosinneR: Good idea, I was not aware of such option. Let's use that instead!

costasovo (dismissed):

veljkho (changes requested):
There are some issues found:
1. Unable to select `Unchecked` [for MP CF checkbox](https://d.pr/i/YplZTo)
2. If I previously set `MailPoet custom field` > `Any CF` and then switch to subscribed date, I'll see [before without calendar](https://d.pr/i/4UgrdK)
3. What's the purpose of `Months MP CF`? I can't see any calculation nor save it for [Date custom field with only months](https://d.pr/i/vB0Nsw).

Everything else should be fine.

costasovo (approved):

veljkho (approved):

PR #3520, opened 2021-05-19
Use WP connection db port if missing in config [MAILPOET-3580]

pavel-mailpoet:
[MAILPOET-3580]

[MAILPOET-3580]: https://mailpoet.atlassian.net/browse/MAILPOET-3580

### DIFF
--- a/lib/Doctrine/ConnectionFactory.php
+++ b/lib/Doctrine/ConnectionFactory.php
@@ -29,6 +29,7 @@ class ConnectionFactory {
   ];
 
   public function createConnection() {
+    global $wpdb;
     $platformClass = self::PLATFORM_CLASS;
     $connectionParams = [
       'wrapperClass' => SerializableConnection::class,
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ConnectionFactory.php
```diff
@@ -49,7 +50,11 @@ public function createConnection() {
       $connectionParams['unix_socket'] = Env::$dbSocket;
     } else {
       $connectionParams['host'] = Env::$dbIsIpv6 ? ('[' . Env::$dbHost . ']') : Env::$dbHost;
-      $connectionParams['port'] = Env::$dbPort;
+      if (!empty(Env::$dbPort)) {
+        $connectionParams['port'] = Env::$dbPort;
+      } else {
+        $connectionParams['port'] = $wpdb->get_var('SELECT @@port');
```
lysyjan, line 56: You removed two checks on the port in the integration tests. It would be nice to have a test for this new option. Do you think that it is possible to add it? Maybe I'm wrong. It is only an idea.

pavel-mailpoet: I added a test when using `Env::$dbPort` but I am not sure how to test this line.

lysyjan (approved):

veljkho (approved):

PR #3523, opened 2021-05-20
Use WP transient for lock running migrations [MAILPOET-3537]

lysyjan:
[MAILPOET-3537]

[MAILPOET-3537]: https://mailpoet.atlassian.net/browse/MAILPOET-3537

### DIFF
--- a/lib/Config/Activator.php
+++ b/lib/Config/Activator.php
@@ -2,9 +2,13 @@
 
 namespace MailPoet\Config;
 
+use MailPoet\InvalidStateException;
 use MailPoet\Settings\SettingsController;
+use MailPoet\WP\Functions as WPFunctions;
 
 class Activator {
+  public const TRANSIENT_ACTIVATE_KEY = 'mailpoet_activator_activate';
+  private const TRANSIENT_EXPIRATION = 120;
 
   /** @var SettingsController */
   private $settings;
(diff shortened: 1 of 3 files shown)


lib/Config/Activator.php
```diff
@@ -2,22 +2,41 @@
 
 namespace MailPoet\Config;
 
+use MailPoet\InvalidStateException;
 use MailPoet\Settings\SettingsController;
+use MailPoet\WP\Functions as WPFunctions;
 
 class Activator {
+  public const TRANSIENT_ACTIVATE_KEY = 'mailpoet_activator_activate';
+  private const TRANSIENT_EXPIRATION = 120;
```
pavel-mailpoet, line 11: Hey, could you leave a comment here on what units are we using? Is it 120 seconds or minutes?

pavel-mailpoet (approved):

lib/Config/Initializer.php
```diff
@@ -346,6 +349,13 @@ public function handleFailedInitialization($exception) {
     return WPNotice::displayError($exception);
   }
 
+  private function handleRunningInitialization(InvalidStateException $exception) {
```
pavel-mailpoet, line 352: Hey, this is probably not a great name, here should probably be `handleRunningMigration`, right?

lysyjan: Thanks, this name will be better.

pavel-mailpoet (approved):

veljkho (approved):

PR #3525, opened 2021-05-21
Fix performance trouble in segment score calculation [MAILPOET-3633]

lysyjan:
[MAILPOET-3633]

The solution has three steps:

1. I removed the unnecessary problematic query because the check is solved in the subscriber's score calculation, and the aggregation function doesn't compute with null values.
2. I dropped the segment's score calculation from the tracking.
3. The segment score is outdated after a day.

[MAILPOET-3633]: https://mailpoet.atlassian.net/browse/MAILPOET-3633


lysyjan:
Circle CI has a capacity problem. I will try to run it again on Monday.

pavel-mailpoet (approved):

veljkho (approved):

PR #3530, opened 2021-05-25
Add curl error logging for MSS messages api calls [MAILPOET-3621]

costasovo:
[MAILPOET-3621]

[MAILPOET-3621]: https://mailpoet.atlassian.net/browse/MAILPOET-3621

### DIFF
--- a/lib/Services/Bridge/API.php
+++ b/lib/Services/Bridge/API.php
@@ -4,6 +4,7 @@ namespace MailPoet\Services\Bridge;
 
 use MailPoet\Logging\LoggerFactory;
 use MailPoet\WP\Functions as WPFunctions;
+use WP_Error;
 
 class API {
   const SENDING_STATUS_OK = 'ok';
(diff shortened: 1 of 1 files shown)


lib/Services/Bridge/API.php
```diff
@@ -175,4 +186,13 @@ private function request($url, $body, $method = 'POST') {
     ];
     return $this->wp->wpRemotePost($url, $params);
   }
+
+  private function logCurlError(WP_Error $error) {
+    $logData = [
+      'curl_errno' => $this->curlHandle ? curl_errno($this->curlHandle) : 'n/a',
+      'curl_error' => $this->curlHandle ? curl_error($this->curlHandle) : $error->get_error_message(),
+      'curl_getinfo' => $this->curlHandle ? curl_getinfo($this->curlHandle) : 'n/a',
```
lysyjan, line 194: I noticed that in `logCurlInformation`, we use the key `curl_info`. I would prefer to use the same key, but it is nothing important. What do you think?

lysyjan (commented):
Looks good. I have only one note.

lysyjan (approved):
Thank you 👍🏻

veljkho (approved):

PR #3531, opened 2021-05-25
Gracefully handle unsupported Email Editor blocks [MAILPOET-3437]

pavel-mailpoet:
[MAILPOET-3437]

[MAILPOET-3437]: https://mailpoet.atlassian.net/browse/MAILPOET-3437

### DIFF
--- a/assets/css/src/components-editor/content-blocks/_unknown_block_fallback.scss
+++ b/assets/css/src/components-editor/content-blocks/_unknown_block_fallback.scss
@@ -1,0 +1,3 @@
+.mailpoet_unknown_block_fallback_heading {
+  text-align: center;
+}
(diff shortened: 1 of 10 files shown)


costasovo (changes requested):
The block works works as expected in the editor. I ran into an error when I tried form preview (https://d.pr/i/QXOM3l)

costasovo (approved):

veljkho (changes requested):
The point 3. of the specs is not present.
As I see, it should represent a message if you send a newsletter with unsupported block type but different message to show instead "Skipped unsupported block type: { block_type }". Right now, its just an emptiness present, no message.

veljkho (approved):

PR #3532, opened 2021-05-25
Fix 1px line rendering bug in Outlook 2016 [MAILPOET-3448]

### LINKED ISSUE #3348: Outlook 2016 Rendering Bug (1 pixel(px) Horizontal Line Bug)
**Describe the bug**
For some layouts a 1 pixel line will appear if mails by MailPoet are opened with Outlook. This bug appears most often at the bottom of the table in which the text is contained, but sometimes will appear in other places in the template. Please found an extensive description here [https://support.politemail.com/hc/en-us/articles/360038012653-Outlook-1-pixel-px-Horizontal-Line-Bug](https://support.politemail.com/hc/en-us/articles/360038012653-Outlook-1-pixel-px-Horizontal-Line-Bug). 

**To reproduce**
1. Go to MailPoet and create any mail with multiple 1 or 2-column blocks; background color should differ from mail template
2. Put some dummy text into all blocks
3. Send that mail and open it via Outlook 2016 (other versions not verified)

**Actual behavior**
Certain combinations of font size and font family will trigger a rendering error that’s creates a 1px horizontal line of the email background color to appear.

**Expected behavior**
A clean layout without any additional lines.

**Screenshots**
See link above (in "Describe the Bug")

- WordPress version: 5.5.3 (shouldn't matter)
- PHP version: 7.2.34 (shouldn't matter)
- MailPoet version: 3.54.2

**Additional context**
It seems that MailPoet calculates uneven line-height values like "line-height:22.4px" that most likely will trigger the bug. In short, to  prevent the bug, MailPoet should output even numbers only. For that, 22.4px should be calculated to "22". A "23" should read "24" and so on. First proposed fix therefor would be something like this:

```
--- a/lib/Newsletter/Renderer/StylesHelper.php
+++ b/lib/Newsletter/Renderer/StylesHelper.php
@@ -134,7 +134,9 @@ class StylesHelper {
     if (!preg_match('/mailpoet_paragraph|h[1-4]/i', $selector)) return $style;
     $lineHeight = isset($style['lineHeight']) ? (float)$style['lineHeight'] : self::$defaultLineHeight;
     $fontSize = (int)$style['fontSize'];
-    $style['lineHeight'] = sprintf('%spx', $lineHeight * $fontSize);
+    $msoLineHeight = round($lineHeight * $fontSize);
+    if ($msoLineHeight % 2 == 1) $msoLineHeight++;
+    $style['lineHeight'] = sprintf('%spx', $msoLineHeight);
     return $style;
   }
```

Would be great if somebody could verify the behavior and include the fix in the next release :-). Thank you!

pavel-mailpoet:
[MAILPOET-3448], closes #3348

[MAILPOET-3448]: https://mailpoet.atlassian.net/browse/MAILPOET-3448

### DIFF
--- a/lib/Newsletter/Renderer/StylesHelper.php
+++ b/lib/Newsletter/Renderer/StylesHelper.php
@@ -16,6 +16,8 @@ class StylesHelper {
     'borderColor' => 'border-color',
     'borderRadius' => 'border-radius',
     'lineHeight' => 'line-height',
+    'msoLineHeightAlt' => 'mso-line-height-alt',
+    'msoFontSize' => 'mso-ansi-font-size',
   ];
   public static $font = [
     'Arial' => "Arial, 'Helvetica Neue', Helvetica, sans-serif",
@@ -147,7 +149,18 @@ class StylesHelper {
     if (!preg_match('/mailpoet_paragraph|h[1-4]/i', $selector)) return $style;
     $lineHeight = isset($style['lineHeight']) ? (float)$style['lineHeight'] : self::$defaultLineHeight;
     $fontSize = (int)$style['fontSize'];
+    $msoLineHeight = round($lineHeight * $fontSize);
+    if ($msoLineHeight % 2 === 1) {
+      $msoLineHeight++;
+    }
+    $msoFontSize = $fontSize;
+    if ($msoFontSize % 2 === 1) {
+      $msoFontSize++;
+    }
+    $style['msoLineHeightAlt'] = sprintf('%spx', $msoLineHeight);
+    $style = ['msoFontSize' => sprintf('%spx', $msoFontSize)] + $style;
     $style['lineHeight'] = sprintf('%spx', $lineHeight * $fontSize);
+
     return $style;
   }
 
--- a/tests/unit/Newsletter/Renderer/StylesHelperTest.php
+++ b/tests/unit/Newsletter/Renderer/StylesHelperTest.php
@@ -66,4 +66,24 @@ class StylesHelperTest extends \MailPoetUnitTest {
     expect(StylesHelper::getCustomFontsLinks($stylesWithoutCustomFonts))
       ->equals('');
   }
+
+  public function testItAddsMsoStyles() {
+    $styles = [
+      "fontSize" => "16px",
+      "lineHeight" => "1",
+    ];
+    $styles = StylesHelper::setStyle($styles, '.mailpoet_paragraph');
+    expect($styles)->stringContainsString('mso-ansi-font-size:16px;');
+    expect($styles)->stringContainsString('mso-line-height-alt:16px;');
+
+    $styles = [
+      "fontSize" => "17px",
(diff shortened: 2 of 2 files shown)


lib/Newsletter/Renderer/StylesHelper.php
```diff
@@ -147,7 +149,18 @@ public static function applyLineHeight($style, $selector) {
...
+    $msoLineHeight = round($lineHeight * $fontSize);
+    if ($msoLineHeight % 2 === 1) {
+      $msoLineHeight++;
+    }
+    $msoFontSize = $fontSize;
+    if ($msoFontSize % 2 === 1) {
+      $msoFontSize++;
+    }
+    $style['msoLineHeightAlt'] = sprintf('%spx', $msoLineHeight);
+    $style = ['msoFontSize' => sprintf('%spx', $msoFontSize)] + $style;
```
lysyjan, line 161: Why did you use the union operator between arrays instead of setting the value for the specific key?

lysyjan (commented):
👍🏻  It looks fine. I have only one question.

lib/Newsletter/Renderer/StylesHelper.php
pavel-mailpoet: It is because I wanted to achieve a specific order. This array is later used in a foreach, and the final CSS is built from it. I wanted the order of the CSS rules to match the [original issue](https://github.com/mailpoet/mailpoet/issues/3348). I don't know if that is important or not.

lysyjan (approved):

lib/Newsletter/Renderer/StylesHelper.php
NeosinneR: Just a note, the order of CSS rules inside a single selector/`style=""` attribute doesn't change anything. Only the specificity of a selector and selector's position can change which rule wins.

veljkho (approved):

lib/Newsletter/Renderer/StylesHelper.php
pavel-mailpoet: We don't use `style=""` here, we are using normal CSS declaration. And here is used the normal fontSize with the vendor-prefixed one and I thought the order is important for those.

That being said. I am not sure if it would make any difference with a different order. I tried to match what the original issue "specified".

fhwebcs:
Thank you! My first tryout looks very promising. Even though that there is a `line-height` rendering in header and footer block that is not processed by this fix (see [here](https://github.com/mailpoet/mailpoet/blob/0062a654b81d33280a36ed38510b1b6aa0e4fb90/lib/Newsletter/Renderer/Blocks/Header.php#L37) and [here](https://github.com/mailpoet/mailpoet/blob/0062a654b81d33280a36ed38510b1b6aa0e4fb90/lib/Newsletter/Renderer/Blocks/Footer.php#L37)). In best case it should be fixed here as well. But I'll try if this addtional fix is still needed.

PR #3533, opened 2021-05-25
 Migrate from node-sass to sass [MAILPOET-3634]

lysyjan:
[MAILPOET-3634]

[MAILPOET-3634]: https://mailpoet.atlassian.net/browse/MAILPOET-3634


assets/css/src/components-editor/_forms.scss
```diff
@@ -65,7 +67,7 @@ select.mailpoet_font-size {
     border-radius: $range-border-radius;
     cursor: pointer;
     height: $range-thumb-height;
-    margin-top: -1 * $range-thumb-height / 3;
+    margin-top: -1 * math.div($range-thumb-height, 3);
```
pavel-mailpoet, line 70: I am surprised you had to rewrite this line using `math.div`. I looked in the [documentation](https://sass-lang.com/documentation/operators/numeric) and it doesn't mention the operator for division doesn't work. What is the reason here?

lysyjan: The division operator will be deprecated, and there was a big warning for each usage. When I was finding the problem with a WordPress import, I need to solve each warning because the main error wasn't visible. When I made this change, I committed it.

pavel-mailpoet (approved):

veljkho (approved):

PR #3535, opened 2021-05-26
Refactor isSubscriberProcessed to Doctrine [MAILPOET-3267]

pavel-mailpoet:
[MAILPOET-3267]

@veljkho this is a refactor of "View in Browser" and statistics Tracking. So when testing this feature please focus on those two areas.

[MAILPOET-3267]: https://mailpoet.atlassian.net/browse/MAILPOET-3267


tests/integration/Newsletter/Sending/SendingQueuesRepositoryTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+  private function createSubscriber(): SubscriberEntity {
+    $subscriber = new SubscriberEntity();
+    $subscriber->setStatus(SubscriberEntity::STATUS_SUBSCRIBED);
+    $subscriber->setEmail('a@example.com');
+    $this->entityManager->persist($subscriber);
+    return $subscriber;
+  }
+
+  public function cleanup() {
```
lysyjan, line 94: I noticed that the entity SendingQueueEntity is missing in the cleanup.

tests/integration/Newsletter/ViewInBrowser/ViewInBrowserRendererTest.php
```diff
@@ -165,31 +166,38 @@ public function testItReusesRenderedNewsletterBodyWhenQueueExists() {
       $this->diContainer->get(Shortcodes::class),
       $this->diContainer->get(Renderer::class)
     );
+    /** @var SendingQueuesRepository $sendingQueueRepository */
+    $sendingQueueRepository = $this->diContainer->get(SendingQueuesRepository::class);
```
lysyjan, line 170: You get the SendingQueueRepository five times in different methods. It would be easier to store it one time in the constructor. What do you think?

lib/Newsletter/ViewInBrowser/ViewInBrowserRenderer.php
```diff
@@ -48,11 +43,16 @@ public function render(
...
+    SendingQueueEntity $queue = null
   ) {
     $wpUserPreview = $isPreview;
     if ($queue && $queue->getNewsletterRenderedBody()) {
-      $newsletterBody = $queue->getNewsletterRenderedBody('html');
+      $body = $queue->getNewsletterRenderedBody();
+      if (is_array($body)) {
+        $newsletterBody = $body['html'];
+      } else {
+        $newsletterBody = '';
```
lysyjan, line 54: I looked at the old implementation of the getter `getNewsletterRenderedBody`, and I think you made a mistake in the else statement. I guess that there should be `$newsletterBody = $body`. Or is there some reason for the empty string?

tests/integration/Router/Endpoints/TrackTest.php
```diff
@@ -131,6 +133,7 @@ public function testItFailsWhenSubscriberIsNotOnProcessedList() {
     $this->entityManager->persist($subscriber);
     $this->entityManager->flush();
     $data->subscriber->setId($subscriber->getId());
```
lysyjan, line 135: Isn't this line unnecessary now? This line sets subscriber ID, and the line below overwrites the whole subscriber.

lysyjan (commented):
It looks good 👍🏻 I have only a few notes.

tests/integration/Newsletter/Sending/SendingQueuesRepositoryTest.php
pavel-mailpoet: Do we have to clean it? I mean do we need to clean all of the data in all the tests?

lib/Newsletter/ViewInBrowser/ViewInBrowserRenderer.php
pavel-mailpoet: Basically, I did it because PHPStan was yelling at me. I would say that normally this would never be used. When a user displays a newsletter in a browser the `$body['html']` should always be set. If it is not something really went wrong so I guess the empty string is ok in this case.

tests/integration/Newsletter/Sending/SendingQueuesRepositoryTest.php
lysyjan: It is not necessary, but I thought that some forgotten data can affect other tests. I know that this entity probably not. I noticed it only and mentioned it.

lysyjan (approved):

veljkho (approved):

PR #3536, opened 2021-05-26
Add mailpoet_mailer_smtp_message filter

tylerdigital:
For our application, we need the ability to add some custom headers to outgoing SMTP messages. If you would please merge this PR, the filter will let us accomplish this. I imagine others would find this useful as a way to modify the message being sent (similar to WordPress core's default wp_mail filter)

### DIFF
--- a/lib/Mailer/Methods/SMTP.php
+++ b/lib/Mailer/Methods/SMTP.php
@@ -125,6 +125,7 @@ class SMTP {
     if (!empty($newsletter['body']['text'])) {
       $message = $message->addPart($newsletter['body']['text'], 'text/plain');
     }
+    $message = $this->wp->applyFilters('mailpoet_mailer_smtp_message', $message, $newsletter, $subscriber, $extraParams);
     return $message;
   }
 


tylerdigital:
@veljkho Sorry I'm not familiar with the process for submitting a PR here, it looks like CircleCI failed because of a Slack token being missing from the build process. I don't think this is something I can fix, but I'm happy to if you can point me in the right direction.

I'm just hoping to get this one-line commit in so we can use a filter rather than having to modify the plugin directly :)

pavel-mailpoet:
Hello @tylerdigital , 

sorry about the delay I totally missed your pull request. Thank you for it, we do appreciate it when people are improving the plugin. 

We are a bit hesitant here, we are refactoring the whole plugin and this area is yet to be refactored. We will introduce a BC break within a few months. There is another [hook](https://github.com/mailpoet/mailpoet/pull/3536/files#diff-06c86554e23278a01bdad85226daa551bfa25d355224017e09924865da188debR98) in this file. Could you use that instead of adding a new one? 

Pavel

tylerdigital:
Got it, I'll wait for the new version after you refactor

PR #3544, opened 2021-06-01
Refactor form bulk action to Doctrine [MAILPOET-3630]

pavel-mailpoet:
[MAILPOET-3630]

[MAILPOET-3630]: https://mailpoet.atlassian.net/browse/MAILPOET-3630

### DIFF
--- a/lib/Form/FormsRepository.php
+++ b/lib/Form/FormsRepository.php
@@ -4,7 +4,6 @@ namespace MailPoet\Form;
 
 use MailPoet\Doctrine\Repository;
 use MailPoet\Entities\FormEntity;
-use MailPoetVendor\Carbon\Carbon;
 
 /**
  * @extends Repository<FormEntity>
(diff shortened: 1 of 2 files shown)


lib/Form/FormsRepository.php
```diff
@@ -54,4 +54,49 @@ private function updateDeletedAt(FormEntity $form, ?Carbon $value) {
     $this->persist($form);
     $this->flush();
   }
+
+  public function bulkTrash(array $ids): int {
```
costasovo, line 58: I think it would be nice to reuse bulkTrash inside public function trash() so that we don't have 2 different implementations for single and bulk action. This applies also to bulkRestore and bulkDelete. That way we could get rid of `updateDeletedAt` method.

costasovo (changes requested):
I have just one suggestion about improving code reuse.

costasovo (approved):

veljkho (approved):

PR #3545, opened 2021-06-01
Wrap WC content with table [MAILPOET-3638]

lysyjan:
[MAILPOET-3638]

I found out that we need to wrap the WC content with another table. The reason is that WC [use another padding](https://github.com/woocommerce/woocommerce/blob/trunk/templates/emails/email-styles.php#L110-L116)

[MAILPOET-3638]: https://mailpoet.atlassian.net/browse/MAILPOET-3638

### DIFF
--- a/lib/DI/ContainerConfigurator.php
+++ b/lib/DI/ContainerConfigurator.php
@@ -373,6 +373,7 @@ class ContainerConfigurator implements IContainerConfigurator {
     $container->autowire(\MailPoet\WooCommerce\TransactionalEmails::class)->setPublic(true);
     $container->autowire(\MailPoet\WooCommerce\TransactionalEmails\Template::class)->setPublic(true);
     $container->autowire(\MailPoet\WooCommerce\TransactionalEmails\Renderer::class)->setPublic(true);
+    $container->autowire(\MailPoet\WooCommerce\TransactionalEmails\ContentPreprocessor::class)->setPublic(true);
     // WordPress
     $container->autowire(\MailPoet\WP\Emoji::class)->setPublic(true);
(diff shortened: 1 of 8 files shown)


costasovo (dismissed):
👍

veljkho (changes requested):
One issue fixed, new arrived:
The whole content is now narrowed. Should be stretched.
[Actual result](https://d.pr/i/lEEKpn)
[Expected result](https://d.pr/i/QL9Pgq)

So as it can be seen, text and the content have big space around, its like added big padding around it.

costasovo (approved):

veljkho (changes requested):
Unfortunately, it is not good enough.

I see some improvements but the width is not the same it was before, still needs to be stretched.
[Actual result](https://d.pr/i/3SKIO0)t vs [Expected result](https://d.pr/i/QL9Pgq)

There is another issue that has arrived during this PR, it's about `size:` & `print material:` attributes of the product, they should be moved a bit to the right.. check the expected result screenshot. We had this issue in the past and we fixed it, but seems again present in this pull request after changes.

I used Keyboard with size and print select options to test this.

lysyjan:
> There is another issue that has arrived during this PR, it's about `size:` & `print material:` attributes of the product, they should be moved a bit to the right.. check the expected result screenshot. We had this issue in the past and we fixed it, but seems again present in this pull request after changes.

@veljkho I think that this was an issue. When WC sends the email, then the product attributes are without padding on the left. In my opinion, the current state is correct.

costasovo (approved):

veljkho (approved):

PR #3549, opened 2021-06-02
Refactor data factories to doctrine [MAILPOET-3627]

pavel-mailpoet:
[MAILPOET-3627]

[MAILPOET-3627]: https://mailpoet.atlassian.net/browse/MAILPOET-3627

### DIFF
--- a/tests/acceptance/Settings/SettingsArchivePageCest.php
+++ b/tests/acceptance/Settings/SettingsArchivePageCest.php
@@ -40,9 +40,7 @@ class SettingsArchivePageCest {
     $segment3 = $segmentFactory->withName('SentNewsletters')->create();
     $newsletterFactory = new Newsletter();
     $newsletterFactory->withSubject('SentNewsletter')->withSentStatus()->withSendingQueue()->withSegments([$segment3])->create();
-    $newsletterFactory = new Newsletter();
     $newsletterFactory->withSubject('DraftNewsletter')->withDraftStatus()->withScheduledQueue()->withSegments([$segment3])->create();
-    $newsletterFactory = new Newsletter();
     $newsletterFactory->withSubject('ScheduledNewsletter')->withScheduledStatus()->withScheduledQueue()->withSegments([$segment3])->create();
     $pageTitle3 = 'SentNewsletterArchive';
     $pageContent3 = "[mailpoet_archive segments=\"{$segment3->getId()}\"]";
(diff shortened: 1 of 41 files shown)


tests/acceptance/Subscribers/ExportSubscribersCest.php
```diff
@@ -12,7 +12,9 @@ public function exportSubscribers(\AcceptanceTester $i) {
     $segment = $segmentFactory->withName($segmentName)->create();
     $subscriberFactory = new Subscriber();
     $subscriberFactory->withSegments([$segment])->withEmail('one@fake.fake')->create();
+    $subscriberFactory = new Subscriber();
```
lysyjan, line 15: I think that this doesn't seem right. The factory should work right when it is created only once.
I tried to run the test locally, and the problem is in the method `withSegments` because the function array_merge adds the segment again. After that, the `create` method tries to create an association twice for the same segment.

tests/acceptance/Settings/SettingsArchivePageCest.php
```diff
@@ -51,7 +53,9 @@ public function createArchivePageWithVariousStatusNewsletters(\AcceptanceTester
     $i->waitForText('SentNewsletter');
     $i->dontSee('DraftNewsletter');
     $i->dontSee('ScheduledNewsletter');
+    $newsletterFactory = new Newsletter();
     $newsletterFactory->withSubject('SentNewsletter2')->withDraftStatus()->withSendingQueue()->withSegments([$segment3])->create();
+    $newsletterFactory = new Newsletter();
```
lysyjan, line 58: This is a similar problem as at the Subscriber. The factory shouldn't be created more times. The method `withSegments` adds the segment again to the array. The solution can be to store the segment with a key which can be segment ID for example.

lysyjan (changes requested):
Good job! I have only two notes for new instances of factories.

tests/acceptance/Subscribers/ExportSubscribersCest.php
costasovo: We solved this issue in some newer data factories, by returning an updated clone for every change. See https://github.com/mailpoet/mailpoet/blob/master/tests/DataFactories/WooCommerceCustomer.php and look at update method. It would be nice if we could use it also in other factories so that we don't have to create a new instance of a factory all the time.

lysyjan (commented):
You forgot to remove new instances of factories.

lysyjan (approved):

veljkho (approved):

PR #3556, opened 2021-06-04
Gracefully handle missing WC transactional template [MAILPOET-3017]

pavel-mailpoet:
[MAILPOET-3017]

[MAILPOET-3017]: https://mailpoet.atlassian.net/browse/MAILPOET-3017

### DIFF
--- a/lib/WooCommerce/TransactionalEmailHooks.php
+++ b/lib/WooCommerce/TransactionalEmailHooks.php
@@ -73,6 +73,8 @@ class TransactionalEmailHooks {
     }
     $newsletter = $this->newsletterRepository->findOneById($this->settings->get(TransactionalEmails::SETTING_EMAIL_ID));
     if (!$newsletter) {
+      // the newsletter should always be present in the database, if it s not we shouldn't keep using this feature
+      // we need to recreate the newsletter and turn off the feature
       $this->transactionalEmails->init();
       $this->settings->set('woocommerce.use_mailpoet_editor', false);
     }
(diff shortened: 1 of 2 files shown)


lib/WooCommerce/TransactionalEmailHooks.php
```diff
@@ -48,7 +55,12 @@ public function useTemplateForWoocommerceEmails() {
   }
 
   private function getNewsletter() {
-    return Newsletter::findOne($this->settings->get(TransactionalEmails::SETTING_EMAIL_ID));
+    $newsletter = Newsletter::findOne($this->settings->get(TransactionalEmails::SETTING_EMAIL_ID));
+    if (!$newsletter) {
+      $this->transactionalEmails->init();
+      $this->settings->set('woocommerce.use_mailpoet_editor', false);
+    }
```
costasovo, line 62: Maybe we could add a comment here so that it is obvious that this is not an expected situation but the handling of an error caused by data that are missing. I bet someone would wonder why we turn the feature off.

costasovo (approved):

veljkho (changes requested):
Unable to receive any of the WooCommerce transactional emails after purchasing an item and changing the status of purchase later to Processing, Completed. I tested with the production build and it worked, then switched back to this build again and it still won't execute any of the transactional emails.

veljkho:
Somehow it is working now, going to merge.

veljkho (approved):

PR #3557, opened 2021-06-07
Fix SQL query for subscribers without list [MAILPOET-3647]

costasovo:
[MAILPOET-3647]

[MAILPOET-3647]: https://mailpoet.atlassian.net/browse/MAILPOET-3647


lysyjan (approved):
👍🏻

veljkho (approved):

PR #3571, opened 2021-06-22
Dynamic segment empty name [MAILPOET-3649][MAILPOET-3559][MAILPOET-3650]

lysyjan:
[MAILPOET-3649]
[MAILPOET-3559]
[MAILPOET-3650]

[MAILPOET-3649]: https://mailpoet.atlassian.net/browse/MAILPOET-3649
[MAILPOET-3559]: https://mailpoet.atlassian.net/browse/MAILPOET-3559

[MAILPOET-3650]: https://mailpoet.atlassian.net/browse/MAILPOET-3650

### DIFF
--- a/assets/js/src/segments/dynamic/form.tsx
+++ b/assets/js/src/segments/dynamic/form.tsx
@@ -15,6 +15,7 @@ import { FormFilterFields } from './form_filter_fields';
 import { isFormValid } from './validator';
 import { MinusIcon } from '../../common/button/icon/minus';
 import plusIcon from '../../common/button/icon/plus';
+import APIErrorsNotice from '../../notices/api_errors_notice';
 
 import {
   FilterRow,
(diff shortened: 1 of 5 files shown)


assets/js/src/segments/dynamic/form.tsx
```diff
@@ -64,6 +70,9 @@ export const Form: React.FunctionComponent<Props> = ({
               {MailPoet.I18n.t('name')}
             </label>
           </Heading>
+          {(errors.length > 0 && (
+            <APIErrorsNotice errors={errors.map((error) => ({ message: error }))} />
+          ))}
```
costasovo, line 75: I think that the APIErrorsNotice may actually display more types of errors (e.g. non unique form name) so we shouldn't display all within the form name input. I suggest moving those error notices at the very beginning of the form.

costasovo (changes requested):
Good job! I have just one remark.

costasovo (approved):
👍

veljkho (approved):

PR #3572, opened 2021-06-24
Remove segment ids from unconfirmed data on subscribe [MAILPOET-3663]

lysyjan:
The issue was caused by [the refactoring subscribing to Doctrine](https://github.com/mailpoet/mailpoet/pull/3440).
The reason is that we stored the segment ids as unconfirmed data. After that, a user clicked on the confirmation link, and the method createOrUpdate called the method resetSubscriptions that unsubscribe the user from the old segments.

[MAILPOET-3663]

[MAILPOET-3663]: https://mailpoet.atlassian.net/browse/MAILPOET-3663

### DIFF
--- a/lib/Subscribers/SubscriberSubscribeController.php
+++ b/lib/Subscribers/SubscriberSubscribeController.php
@@ -92,7 +92,8 @@ class SubscriberSubscribeController {
       throw new UnexpectedValueException($e->getMessage());
     }
 
-    $segmentIds = $this->getSegmentIds($form, $data);
+    $segmentIds = $this->getSegmentIds($form, $data['segments'] ?? []);
+    unset($data['segments']);
 
     $meta = $this->validateCaptcha($captchaSettings, $data);
     if (isset($meta['error'])) {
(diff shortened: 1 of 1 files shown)


lib/Subscribers/SubscriberSubscribeController.php
```diff
@@ -215,8 +215,9 @@ private function validateCaptcha($captchaSettings, $data): array {
     return $meta;
   }
 
-  private function getSegmentIds(FormEntity $form, array $data): array {
+  private function getSegmentIds(FormEntity $form, array &$data): array {
```
pavel-mailpoet, line 218: I am not a big fan of passing data using reference. When reading the code that calls this function, it is not obvious that some data are removed here. This function only needs `$data['segments']`, so why are we even passing the whole `$data` array in? Can we rewrite this so that we don't have to pass data using reference?

pavel-mailpoet (changes requested):

pavel-mailpoet (approved):

veljkho (approved):

PR #3583, opened 2021-07-02
Update dependencies [MAILPOET-3658]

costasovo:
[MAILPOET-3658]

After merging please notify developers on slack that in case they use `nvm` for managing node versions and they experience some errors in pre-commit hooks, they may need to create/update `~/.huskyrc` with:
```sh
# This loads nvm.sh and sets the correct PATH before running hook
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```

[MAILPOET-3658]: https://mailpoet.atlassian.net/browse/MAILPOET-3658


pavel-mailpoet (approved):

veljkho (changes requested):
[Fatal error](https://d.pr/i/5n0zuy) while trying to activate the plugin on qawp.net test site.
On another test site I got WordPress Error due to critical error, assuming the same issue.

lysyjan (approved):

veljkho (approved):

PR #3589, opened 2021-07-07
Subscribers counts caching [MAILPOET-3646]

lysyjan:
[MAILPOET-3646]

[MAILPOET-3646]: https://mailpoet.atlassian.net/browse/MAILPOET-3646

### DIFF
--- a/assets/js/src/common/subscribers_cache_message.tsx
+++ b/assets/js/src/common/subscribers_cache_message.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import MailPoet from 'mailpoet';
 import Button from 'common/button/button';
+import Notice from '../notices/notice';
 
 type Props = {
   cacheCalculation: string;
@@ -8,6 +9,7 @@ type Props = {
 
 export function SubscribersCacheMessage({ cacheCalculation }: Props): JSX.Element {
   const [loading, setLoading] = useState(false);
+  const [errors, setErrors] = useState([]);
   const datetimeDiff = new Date().getTime() - new Date(cacheCalculation).getTime();
   const minutes = Math.floor((datetimeDiff / 1000) / 60);
 
@@ -20,10 +22,8 @@ export function SubscribersCacheMessage({ cacheCalculation }: Props): JSX.Elemen
     }).done(() => {
       window.location.reload();
     }).fail((response) => {
-      MailPoet.Notice.error(
-        response.errors.map((error) => error.message),
-        { scroll: true }
-      );
+      setErrors(response.errors.map((error) => error.message));
+      setLoading(false);
     });
   };
 
@@ -43,6 +43,7 @@ export function SubscribersCacheMessage({ cacheCalculation }: Props): JSX.Elemen
         {MailPoet.I18n.t('recalculateNow')}
       </Button>
       <div className="mailpoet-gap" />
+      {errors.length > 0 && <Notice type="error">{errors.map((error) => <p key={error}>{error}</p>)}</Notice>}
     </div>
   );
 }
--- a/assets/js/src/common/subscribers_cache_message.tsx
+++ b/assets/js/src/common/subscribers_cache_message.tsx
@@ -1,0 +1,40 @@
+import React, { useState } from 'react';
+import MailPoet from 'mailpoet';
+import Button from 'common/button/button';
+
+type Props = {
+  cacheCalculation: string;
+};
+
+const handleRecalculate = () => {
+  window.location.reload();
+};
+
+const SubscribersCacheMessage = ({ cacheCalculation }: Props): JSX.Element => {
+  const [loading, setLoading] = useState(false);
+  const datetimeDiff = new Date().getTime() - new Date(cacheCalculation).getTime();
+  const minutes = Math.floor((datetimeDiff / 1000) / 60);
+  return (
+    <div className="mailpoet-subscribers-cache-notice">
+      {MailPoet.I18n.t('subscribersCountWereCalculated')}
+      &nbsp;
+      <abbr title={cacheCalculation}>{`${String(minutes)} ${String(MailPoet.I18n.t('subscribersMinutesAgo'))}`}</abbr>
+      <Button
+        className="mailpoet-subscribers-cache-notice-button"
+        type="button"
+        variant="secondary"
+        dimension="small"
+        onClick={() => {
+          setLoading(true);
+          handleRecalculate();
+        }}
+        withSpinner={loading}
+      >
+        {MailPoet.I18n.t('recalculateNow')}
+      </Button>
+      <div className="mailpoet-gap" />
+    </div>
+  );
+};
+
+export { SubscribersCacheMessage };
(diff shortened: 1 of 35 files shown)


lib/Segments/SegmentSubscribersRepository.php
```diff
@@ -24,12 +25,17 @@ class SegmentSubscribersRepository {
   /** @var FilterHandler */
   private $filterHandler;
 
+  /** @var TransientCache */
+  private $transientCache;
```
pavel-mailpoet, line 29: I am not sure if having the cache here in the repository is a good idea. You can convince me otherwise. I always thought repositories are in the code for SQL queries. I would add another layer on top of the repository and deal with the caching in there.

lib/Cron/Workers/SubscribersCountCacheRecalculation.php
```diff
@@ -0,0 +1,75 @@
...
+    WPFunctions $wp
+  ) {
+    parent::__construct($wp);
+    $this->transientCache = $transientCache;
+    $this->segmentsRepository = $segmentsRepository;
+    $this->segmentSubscribersRepository = $segmentSubscribersRepository;
+  }
+
+  public function processTaskStrategy(ScheduledTask $task, $timer) {
+    $segments = $this->segmentsRepository->findAll();
```
pavel-mailpoet, line 41: I am not sure how this works if this calculation goes always in the same order. Would the first segment be always up to date but the last one may fail to calculate because of the execution limit?

lib/Cron/Workers/SubscribersCountCacheRecalculation.php
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private function recalculateSegmentCache($timer, int $segmentId, ?SegmentEntity $segment = null): void {
+    $this->cronHelper->enforceExecutionLimit($timer);
+    $now = Carbon::now();
+    $item = $this->transientCache->getItem(TransientCache::SUBSCRIBERS_STATISTICS_COUNT_KEY, $segmentId);
+    if ($item === null || $now->diffInMinutes($item['created_at']) > self::EXPIRATION_IN_MINUTES) {
+      $this->transientCache->invalidateItem(TransientCache::SUBSCRIBERS_STATISTICS_COUNT_KEY, $segmentId);
+      if ($segment) {
+        $this->segmentSubscribersRepository->getSubscribersStatisticsCount($segment);
```
pavel-mailpoet, line 59: This is a bit weird, we are getting statistics, but we never save them? I understand that the function has a side-effect and stores the value in the cache, I am not sure if that is obvious enough and we will understand it always.

lib/AdminPages/PageRenderer.php
```diff
@@ -70,6 +76,9 @@ class_exists(Debugger::class)
       ) {
         DIPanel::init();
       }
+      if (is_admin() && $this->subscribersCountCacheRecalculation->shouldBeScheduled()) {
+        $this->subscribersCountCacheRecalculation->schedule();
```
pavel-mailpoet, line 80: I am not sure I understand this change. Why do we need to schedule this calculation in the renderer?

assets/js/src/common/subscribers_cache_message.tsx
```diff
@@ -6,14 +6,27 @@ type Props = {
...
+export function SubscribersCacheMessage({ cacheCalculation }: Props): JSX.Element {
   const [loading, setLoading] = useState(false);
   const datetimeDiff = new Date().getTime() - new Date(cacheCalculation).getTime();
   const minutes = Math.floor((datetimeDiff / 1000) / 60);
+
+  const handleRecalculate = () => {
+    setLoading(true);
+    MailPoet.Ajax.post({
+      api_version: MailPoet.apiVersion,
+      endpoint: 'settings',
```
pavel-mailpoet, line 18: I am not sure why 'settings'. That doesn't sound to me like the best place for subscribers count recalculation.

assets/js/src/common/subscribers_cache_message.tsx
```diff
@@ -6,14 +6,27 @@ type Props = {
...
+      action: 'recalculateSubscribersCountsCache',
+    }).done(() => {
+      window.location.reload();
+    }).fail((response) => {
+      MailPoet.Notice.error(
```
pavel-mailpoet, line 23: I think we should stop using these old notices. There is the React version that should be used nowadays. We should probably deprecate this old one.

pavel-mailpoet (changes requested):

assets/js/src/common/subscribers_cache_message.tsx
lysyjan: I thought that it is similar to the recalculation of the engagement score. So, I placed it in the settings. Which place is better for you?

lib/AdminPages/PageRenderer.php
lysyjan: The reason is that we don't need to recalculate cache when an admin doesn't need it. I liked that idea from Rosta's [comment](https://wp.me/pcNwfB-3B#comment-189). Maybe my solution is wrong.

lib/Cron/Workers/SubscribersCountCacheRecalculation.php
lysyjan: This problem should solve the check on line 56. When the cached item is up to date, it is skipped.

assets/js/src/common/subscribers_cache_message.tsx
pavel-mailpoet: Right, I didn't realize that. The engagement score is there because we call it from the settings page. If you think this is the best place, let's leave it here. I thought maybe the subscribers API would be better or segments maybe.

lib/AdminPages/PageRenderer.php
pavel-mailpoet: Right, I didn't remember that comment. I was surprised that it is in the renderer, that is all.

lib/Cron/Workers/SubscribersCountCacheRecalculation.php
lysyjan: You are right. I will create a service that will work directly with the repository. After that, I can remove the transient cache dependency from the repository.

assets/js/src/common/subscribers_cache_message.tsx
```diff
@@ -1,13 +1,15 @@
 import React, { useState } from 'react';
 import MailPoet from 'mailpoet';
 import Button from 'common/button/button';
+import Notice from '../notices/notice';
```
pavel-mailpoet, line 4: No need to change it now, but maybe it is better to use `import Notice from 'notices/notice';`

assets/js/src/common/subscribers_cache_message.tsx
```diff
@@ -43,6 +43,7 @@ export function SubscribersCacheMessage({ cacheCalculation }: Props): JSX.Elemen
         {MailPoet.I18n.t('recalculateNow')}
       </Button>
       <div className="mailpoet-gap" />
+      {errors.length > 0 && <Notice type="error">{errors.map((error) => <p key={error}>{error}</p>)}</Notice>}
```
pavel-mailpoet, line 46: If you used `APIErrorsNotice` instead of regular `Notice` the code would be much more simple here.

pavel-mailpoet (approved):

veljkho (approved):

PR #3591, opened 2021-07-13
Update AccessControl.php for WooCommerce

shoheitanaka:
Update AccessControl.php to allow managers in WooCommerce to configure Mailpoet setting.
I think that email management is necessary up to the shop manager.

### DIFF
--- a/lib/Config/AccessControl.php
+++ b/lib/Config/AccessControl.php
@@ -21,6 +21,7 @@ class AccessControl {
         'mailpoet_permission_access_plugin_admin',
         [
           'administrator',
+          'shop_manager',
           'editor',
         ]
       ),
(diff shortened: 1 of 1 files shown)


pavel-mailpoet:
Hello @shoheitanaka, thank you for your contribution. We are not very confident in making this change. It would apply to everyone and it would add the new permission to the existing WooCommerce stores. We are also not sure if a store manager is a right person to manage email marketing. 
But there are filters so you can use them to apply this change to your store. For example for settings you can use this code snippet:
```
    add_filter('mailpoet_permission_manage_settings', function () {
      return [
           'administrator',
           'shop_manager',
         ];
    });
```
and similar snippets for the others.

shoheitanaka:
Hi, Pavel. 
O.K. Thank you for your consideration. I will do that.

PR #3602, opened 2021-07-28
Fix some PHPStan errors and update PHPStan configuration accordingly [MAILPOET-3236]

rodrigoprimo:
In this PR, I have changed phpstan.neon and phpstan-tests.neon and removed the lines that were added to skip the following checks: checkGenericClassInNonGenericObjectType, checkMissingVarTagTypehint, and checkMissingTypehints. I have also fixed all the errors that PHPStan reported after I changed its configuration files.

Please note that before working on this, I was not familiar with those specific PHPStan checks, so there is a chance that I missed something.

Also, each commit contains all the relevant changes for a specific PHPStan check. It might be easier to review this PR commit by commit.

[MAILPOET-3236]

[MAILPOET-3236]: https://mailpoet.atlassian.net/browse/MAILPOET-3236

### DIFF
--- a/lib/Entities/NewsletterEntity.php
+++ b/lib/Entities/NewsletterEntity.php
@@ -131,25 +131,25 @@ class NewsletterEntity {
 
   /**
    * @ORM\OneToMany(targetEntity="MailPoet\Entities\NewsletterEntity", mappedBy="parent")
-   * @var NewsletterEntity[]|ArrayCollection
+   * @var ArrayCollection<int, NewsletterEntity>
    */
   private $children;
 
   /**
    * @ORM\OneToMany(targetEntity="MailPoet\Entities\NewsletterSegmentEntity", mappedBy="newsletter", orphanRemoval=true)
-   * @var NewsletterSegmentEntity[]|ArrayCollection
+   * @var ArrayCollection<int, NewsletterSegmentEntity>
    */
   private $newsletterSegments;
 
   /**
    * @ORM\OneToMany(targetEntity="MailPoet\Entities\NewsletterOptionEntity", mappedBy="newsletter", orphanRemoval=true)
-   * @var NewsletterOptionEntity[]|ArrayCollection
+   * @var ArrayCollection<int, NewsletterOptionEntity>
    */
   private $options;
 
   /**
    * @ORM\OneToMany(targetEntity="MailPoet\Entities\SendingQueueEntity", mappedBy="newsletter")
-   * @var SendingQueueEntity[]|ArrayCollection
+   * @var ArrayCollection<int, SendingQueueEntity>
    */
   private $queues;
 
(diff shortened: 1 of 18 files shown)


lib/Entities/NewsletterEntity.php
```diff
@@ -131,25 +131,25 @@ class NewsletterEntity {
 
   /**
    * @ORM\OneToMany(targetEntity="MailPoet\Entities\NewsletterEntity", mappedBy="parent")
-   * @var NewsletterEntity[]|ArrayCollection
+   * @var ArrayCollection<int, NewsletterEntity>
```
costasovo, line 134: I'm not sure about removing `NewsletterEntity[]`. 

`NewsletterEntity[]` is technically incorrect annotation but it is [a trick](https://stackoverflow.com/a/14725444) how to make code completion work in PHPStorm.

Without it we loose completion like this one:
Screen Shot on 2021-07-29 at 13-34-51

On the other hand the completion doesn't work when accessing Collection elements via Collection methods e.g.
`$this->children->get(0)->get...` even with the trick.

I tried adding back `NewsletterEntity[]` into an annotation and it seems that PHP Stan doesn't mind it so we could still allow the PHP Stan rule and also keep the trick.

I'm inclined to keep the trick there so that it is comfortable for developers to work with the collections. Doctrine collections [can be used as arrays](https://github.com/doctrine/common/blob/2.1.x/lib/Doctrine/Common/Collections/Collection.php#L46) so if we use them like that we could still have the advantage of code completion.

It is not a strong opinion though so I guess we could also ask others about their thoughts. What do you think @pavel-mailpoet , @lysyjan , @brezocordero ?

costasovo (commented):
Thanks for the PR. I added one comment about collection annotations.

lib/Entities/NewsletterEntity.php
pavel-mailpoet: I agree with you @costasovo

rodrigoprimo: Good point, @costasovo. I wasn't sure why we had `NewsletterEntity[]` in the docblock as I believe this is not the correct type for this property. Thanks for clarifying the reason.

When I first read your comment, I was torn. On the one hand, I understand the importance of auto-completion in the IDE, but on the other hand, I wasn't sure if we should use the incorrect type because our IDE doesn't support a specific PHPDoc feature.

I dig a little deeper and found that support for generics is coming in the next version of PHPStorm (https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-beta/), which I imagine will be released soon. The RC is already out: https://www.jetbrains.com/phpstorm/nextversion/. I believe this means that once the new version is released, auto-completion will work with just `@var ArrayCollection<int, NewsletterEntity>`.

In my opinion, this changes things a little bit, and I'm more inclined to suggest that we keep `@var ArrayCollection<int, NewsletterEntity>` instead of adding `NewsletterEntity[]` back. What do you and the others think?

costasovo: @rodrigoprimo That's great news! I din't know it. In that case I'm for going with correct annotations and no hacks.

costasovo (approved):

veljkho (approved):

PR #3605, opened 2021-07-29
Prevent deleting a list if attached to a form [MAILPOET-3661] 

brezocordero:
When a user tries to trash a list that is attached to a form, we show a message instead of allowing them to trash the list:
`List cannot be deleted because it’s used for {$formName} form`

Follows the same behavior as when deleting a list if attached to a newsletter.
[MAILPOET-3661] 

[MAILPOET-3661]: https://mailpoet.atlassian.net/browse/MAILPOET-3661

### DIFF
--- a/lib/Form/FormsRepository.php
+++ b/lib/Form/FormsRepository.php
@@ -27,6 +27,20 @@ class FormsRepository extends Repository {
       ->getResult();
   }
 
+  public function getNamesOfFormsForSegments(): array {
+    $allNonDeletedForms = $this->findAllNotDeleted();
+
+    $nameMap = [];
+    foreach ($allNonDeletedForms as $form) {
+      $blockSegmentsIds = $form->getSettingsSegmentIds();
+      foreach ($blockSegmentsIds as $blockSegmentId) {
+        $nameMap[(string)$blockSegmentId][] = $form->getName();
+      }
+    }
+
+    return $nameMap;
+  }
+
   public function count(): int {
     return (int)$this->doctrineRepository
       ->createQueryBuilder('f')
--- a/lib/Form/FormsRepository.php
+++ b/lib/Form/FormsRepository.php
@@ -34,7 +34,7 @@ class FormsRepository extends Repository {
     foreach ($allNonDeletedForms as $form) {
       $blockSegmentsIds = $form->getSettingsSegmentIds();
       foreach ($blockSegmentsIds as $blockSegmentId) {
-        $nameMap[(string)$blockSegmentId][] = $form->getName();
+        $nameMap[$blockSegmentId][] = $form->getName();
       }
     }
 
--- a/lib/Segments/SegmentsRepository.php
+++ b/lib/Segments/SegmentsRepository.php
@@ -8,6 +8,7 @@ use MailPoet\Entities\DynamicSegmentFilterData;
 use MailPoet\Entities\DynamicSegmentFilterEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SubscriberSegmentEntity;
+use MailPoet\Form\FormsRepository;
 use MailPoet\Newsletter\Segment\NewsletterSegmentRepository;
 use MailPoet\NotFoundException;
 use MailPoet\WP\Functions as WPFunctions;
@@ -23,12 +24,17 @@ class SegmentsRepository extends Repository {
   /** @var NewsletterSegmentRepository */
   private $newsletterSegmentRepository;
 
+  /** @var FormsRepository */
+  private $formsRepository;
+
   public function __construct(
     EntityManager $entityManager,
-    NewsletterSegmentRepository $newsletterSegmentRepository
+    NewsletterSegmentRepository $newsletterSegmentRepository,
+    FormsRepository $formsRepository
   ) {
     parent::__construct($entityManager);
     $this->newsletterSegmentRepository = $newsletterSegmentRepository;
+    $this->formsRepository = $formsRepository;
   }
 
   protected function getEntityClassName() {
@@ -169,8 +175,10 @@ class SegmentsRepository extends Repository {
   }
 
   public function bulkTrash(array $ids, string $type = SegmentEntity::TYPE_DEFAULT): int {
-    $activelyUsed = $this->newsletterSegmentRepository->getSubjectsOfActivelyUsedEmailsForSegments($ids);
-    $ids = array_diff($ids, array_keys($activelyUsed));
+    $activelyUsedInNewsletters = $this->newsletterSegmentRepository->getSubjectsOfActivelyUsedEmailsForSegments($ids);
+    $activelyUsedInForms = $this->formsRepository->getNamesOfFormsForSegments();
+    $activelyUsed = array_unique(array_merge(array_keys($activelyUsedInNewsletters), array_keys($activelyUsedInForms)));
+    $ids = array_diff($ids, $activelyUsed);
     return $this->updateDeletedAt($ids, new Carbon(), $type);
   }
 
--- a/lib/Segments/SegmentsRepository.php
+++ b/lib/Segments/SegmentsRepository.php
@@ -182,6 +182,10 @@ class SegmentsRepository extends Repository {
     return $this->updateDeletedAt($ids, new Carbon(), $type);
   }
 
+  public function doTrash(array $ids, string $type = SegmentEntity::TYPE_DEFAULT): int {
+    return $this->updateDeletedAt($ids, new Carbon(), $type);
+  }
+
   public function bulkRestore(array $ids, string $type = SegmentEntity::TYPE_DEFAULT): int {
     return $this->updateDeletedAt($ids, null, $type);
   }
--- a/tests/acceptance/Lists/ManageListsCest.php
+++ b/tests/acceptance/Lists/ManageListsCest.php
@@ -5,6 +5,7 @@ namespace MailPoet\Test\Acceptance;
 use MailPoet\DI\ContainerWrapper;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Segments\SegmentsRepository;
+use MailPoet\Test\DataFactories\Form;
 use MailPoet\Test\DataFactories\Newsletter;
 use MailPoet\Test\DataFactories\Segment;
 use MailPoet\Test\DataFactories\User;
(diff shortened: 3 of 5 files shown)


lib/Segments/SegmentsRepository.php
```diff
@@ -169,8 +175,10 @@ public function bulkDelete(array $ids, $type = SegmentEntity::TYPE_DEFAULT) {
   }
 
   public function bulkTrash(array $ids, string $type = SegmentEntity::TYPE_DEFAULT): int {
-    $activelyUsed = $this->newsletterSegmentRepository->getSubjectsOfActivelyUsedEmailsForSegments($ids);
-    $ids = array_diff($ids, array_keys($activelyUsed));
+    $activelyUsedInNewsletters = $this->newsletterSegmentRepository->getSubjectsOfActivelyUsedEmailsForSegments($ids);
+    $activelyUsedInForms = $this->formsRepository->getNamesOfFormsForSegments();
```
rodrigoprimo, line 179: It seems to me that calling `getNamesOfFormsForSegments()` here means that this method is called twice when deleting a single list and I don't think this is necessary. I understand that we need to call it here when bulk deleting lists. I wonder if it might make sense to split bulkTrash() into two methods. Something like:

```
function bulkTrash(...) {
  // filter out ids that we don't want to delete
  $this->doTrash(...);
}

function doTrash(...) { // maybe we can think of a better name for this method
  // actually delete lists
  // can be called by both bulkTrash() and Segments::trash()
}
```

What do you think?

tests/integration/API/JSON/v1/SegmentsTest.php
```diff
@@ -162,6 +163,62 @@ public function testItReturnsErrorWhenTrashingSegmentWithActiveNewsletter() {
...
+  public function testItCanTrashSegmentWithoutActiveForm() {
+    $form1 = new FormEntity('My Form');
+    $form1->setSettings([
+      'segments' => [(string)$this->segment3->getId()],
+    ]);
+    $this->entityManager->persist($form1);
+    $this->entityManager->flush();
+
+    $response = $this->endpoint->trash(['id' => $this->segment2->getId()]);
+    $this->entityManager->clear();
```
rodrigoprimo, line 210: Here comes another question from someone that is still learning Doctrine. Why do we need to call `clear()` here?

tests/acceptance/Lists/ManageListsCest.php
```diff
@@ -211,6 +212,32 @@ public function cantTrashOrBulkTrashActivelyUsedList(\AcceptanceTester $i) {
...
+    $i->login();
+    $i->amOnMailpoetPage('Lists');
+    $i->waitForText($listTitle, 5, '[data-automation-id="listing_item_' . $segment->getId() . '"]');
+    $i->clickItemRowActionByItemName($listTitle, 'Move to trash');
+    $i->waitForText("List cannot be deleted because it’s used for '{$formName}' form");
+    $i->seeNoJSErrors();
+    $i->checkOption('[data-automation-id="listing-row-checkbox-' . $segment->getId() . '"]');
+    $i->waitForText('Move to trash');
+    $i->click('Move to trash');
+    $i->waitForText('0 lists were moved to the trash.');
```
rodrigoprimo, line 238: I personally find it a bit confusing that we are silently ignoring the user's request to delete a few of the lists if they are still being used in a form or an e-mail. I would expect a more detailed message saying something like "X lists were moved to the trash. Unable to move one list because it is still being used in a form". My concern here is that, when deleting several lists at once, it is easy for the user to miss that some were not deleted as we are just showing a notification that indicates success. They have to know how many lists they tried to delete and compare that to the number of lists that were deleted and that we are showing in the message.

I thought it was worth mentioning this for discussion, but I'm not sure if we should do anything about it in this PR as I believe what I'm describing above applies more broadly to things that are not related to this ticket. For example, the same happens when deleting lists that are used in e-mails.

rodrigoprimo (commented):
Good job, @brezocordero. I tested and confirmed that this is working as expected. I left some comments for us to discuss some details.

tests/integration/API/JSON/v1/SegmentsTest.php
```diff
@@ -162,6 +163,62 @@ public function testItReturnsErrorWhenTrashingSegmentWithActiveNewsletter() {
     expect($response->errors[0]['message'])->equals("List cannot be deleted because it’s used for 'Subject' email");
   }
 
+  public function testItReturnsErrorWhenTrashingSegmentWithActiveForm() {
+    $form = new FormEntity('My Form');
+    $form->setSettings([
+      'segments' => [(string)$this->segment3->getId()],
+    ]);
+    $this->entityManager->persist($form);
+    $this->entityManager->flush();
```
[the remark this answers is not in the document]
brezocordero: Done!

tests/integration/API/JSON/v1/SegmentsTest.php
```diff
@@ -162,6 +163,62 @@ public function testItReturnsErrorWhenTrashingSegmentWithActiveNewsletter() {
...
+
+    $response = $this->endpoint->trash(['id' => $this->segment3->getId()]);
+    $this->entityManager->refresh($this->segment3);
```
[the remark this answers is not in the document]
brezocordero: I am also not familiar with Doctrine so take my answers with a grain of salt.
We call the method to reload the entity as we have made changes on the database.

lib/Form/FormsRepository.php
```diff
@@ -27,6 +27,20 @@ public function findAllNotDeleted(): array {
       ->getResult();
   }
 
+  public function getNamesOfFormsForSegments(): array {
+    $allNonDeletedForms = $this->findAllNotDeleted();
```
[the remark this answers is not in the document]
brezocordero: There is but I believe it is discouraged. The tables are mapped to the models and retrieving partial values means we get entities that are missing data. As the table is unlikely to get big, I would not optimize it.
Again, I am also not familiar with doctrine, [read about it on this article](https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/partial-objects.html#partial-objects), so happy to change it if I am wrong.

lib/Form/FormsRepository.php
```diff
@@ -27,6 +27,20 @@ public function findAllNotDeleted(): array {
...
+
+    $nameMap = [];
+    foreach ($allNonDeletedForms as $form) {
+      $blockSegmentsIds = $form->getSettingsSegmentIds();
+      foreach ($blockSegmentsIds as $blockSegmentId) {
+        $nameMap[(string)$blockSegmentId][] = $form->getName();
```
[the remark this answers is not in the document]
brezocordero: Removed. Good point.

lib/Segments/SegmentsRepository.php
brezocordero: I agree. I was hesitant to change it as this is my first look at the plugin code and I didn't want to cause a regression but it is not optimal.
I am also not sold on the function name `doTrash` but couldn't come up with anything better.

tests/integration/API/JSON/v1/SegmentsTest.php
brezocordero: And another answer from someone also not familiar with Doctrine:
`clear` would force reloading all the entities and it was not needed there. Changed it to refresh only the segment.

tests/acceptance/Lists/ManageListsCest.php
brezocordero: This is a good point. I will create a new issue to address it and we can discuss there how to handle the different cases.

tests/integration/API/JSON/v1/SegmentsTest.php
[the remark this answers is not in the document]
rodrigoprimo: I think there is still something that I'm not fully understanding and I will investigate it a little bit more when I have the time, but we don't need to hold this PR because of that.

The test pass without this line (which is expected, after all, we are not checking for any of the values inside `$this->segment3`). But what intrigues me and makes me wonder that `refresh()` is not reloading `$this->segment3` instead is doing something else, is that if I add the following code below line 175, the test still passes and I would expect it to fail:

```
$this->assertNull($this->segment3->getDeletedAt());
```

lib/Segments/SegmentsRepository.php
rodrigoprimo: Looks good to me. I think we can leave the name `doTrash()` for now as we couldn't come up with a better name.

What I had in mind is call `doTrash()` from inside `bulkTrash()` but the way it is works as well.

tests/acceptance/Lists/ManageListsCest.php
rodrigoprimo: Sounds good to me. Thanks!

tests/integration/API/JSON/v1/SegmentsTest.php
rodrigoprimo: Resolving this one as it is very similar to the conversation above (https://github.com/mailpoet/mailpoet/pull/3605#discussion_r680359953)

lib/Form/FormsRepository.php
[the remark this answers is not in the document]
rodrigoprimo: Thanks for pointing me to this article. I was not aware of it. To me, it makes sense to avoid using partial objects as the article suggests unless we really need to which is not the case here.

Going forward, for situations similar to this one, I will consider using `getArrayResult()` instead of `getResult()`. It seems like a way to retrieve just what we need without running into the partial objects problem.

rodrigoprimo (approved):
Good job and thanks for helping me learn more about Doctrine.

tests/integration/API/JSON/v1/SegmentsTest.php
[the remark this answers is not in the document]
brezocordero: Thanks for the questions, they are helping me dive deeper into Doctrine. I will investigate too and let you know what I find.

tests/acceptance/Lists/ManageListsCest.php
brezocordero: Issue created.

veljkho (changes requested):
Everything's good except one scenario:
- If I firstly just move a form to the trash (with the list attached) I will be able to remove the list completely. This is not good and would be expected to see the same warning message even if the form is laying in the trash (with the list attached). This means when permanently deleting a form with the list attached would then allow deleting the list only. This is working for newsletters for example.

brezocordero:
@veljkho Thanks for all the tests and sorry for the late reply. I missed your comment.

I am happy to make the change and only allow deleting the list if the form is permanently deleted but I followed the behavior we have with emails.

Today, if you move a welcome email or automatic email with a list attached to the trash, you can delete the list.
For newsletters, you can also delete a list if the newsletter is in the trash unless the newsletter is scheduled.
With post notifications, as they are always scheduled, they need to be deleted permanently in order to be able to delete a list.

What do you think?

veljkho (approved):
Hi @brezocordero 
I think its fine. Merging the pr. Thanks!

PR #3607, opened 2021-07-30
Use entities instead of old models in link shortcode filter [MAILPOET-3464]

lysyjan:
[MAILPOET-3464]

[MAILPOET-3464]: https://mailpoet.atlassian.net/browse/MAILPOET-3464

### DIFF
--- a/lib/Newsletter/Shortcodes/Categories/Link.php
+++ b/lib/Newsletter/Shortcodes/Categories/Link.php
@@ -87,9 +87,9 @@ class Link implements CategoryInterface {
         $url = $this->wp->applyFilters(
           'mailpoet_newsletter_shortcode_link',
           $shortcode,
-          $newsletterModel,
-          $subscriberModel,
-          $queueModel,
+          $newsletter,
+          $subscriber,
+          $queue,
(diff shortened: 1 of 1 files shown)


rodrigoprimo (commented):
Code changes look good to me.

The only thing that I wanted to check before we move this PR to QA is that this is a breaking change and has the potential to crash sites that might be using the filters and expecting the old objects instead of Doctrine entities. Is this a concern? How do we normally handle breaking changes like this one?

lysyjan:
@rodrigoprimo Good point! You are right. This is a breaking change, but our article in the KB contains a description that is equal to the state after merging this PR. It means that a user should expect Doctrine entities. We added [a deprecated shortcode notice](https://github.com/mailpoet/mailpoet/blob/master/lib/Util/Notices/DeprecatedShortcodeNotice.php) almost 8 months ago. I'm not sure about the best approach.

rodrigoprimo (approved):
Sorry for the delay to respond, @lysyjan. If as you said we already have in place code from a previous ticket to handle sites using the filter and expecting the old models, then we can move on with this PR.

veljkho (approved):

PR #3608, opened 2021-08-02
Allow import subscribers with empty IP address [MAILPOET-3665]

brezocordero:
Allow import subscribers with empty `confirmed_ip` or `subscribed_ip` when one of the columns is selected in the last step of the import.
Don't allow importing invalid IP addresses - still import the subscriber but remove the invalid IP address.

[MAILPOET-3665]

[MAILPOET-3665]: https://mailpoet.atlassian.net/browse/MAILPOET-3665

### DIFF
--- a/lib/Subscribers/ImportExport/Import/Import.php
+++ b/lib/Subscribers/ImportExport/Import/Import.php
@@ -205,9 +205,10 @@ class Import {
       }
       if (in_array($column, ['confirmed_ip', 'subscribed_ip'], true)) {
         $data = array_map(
-          function($index, $ip) use(&$invalidRecords, $validator) {
+          function($index, $ip) use($validator) {
             if (!$validator->validateIPAddress($ip)) {
-              $invalidRecords[] = $index;
+              // if invalid or empty, we allow the import but remove the IP
+              return null;
             }
             return $ip;
           }, array_keys($data), $data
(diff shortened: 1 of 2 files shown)


costasovo (approved):
Good job!

veljkho (changes requested):
Something is happening now. I could not import the list if I previously remove one or two subscribers from the list. I'm getting "Unable to save imported subscribers".

SS: https://d.pr/i/Kr1HiD

Steps (the easiest way)
1. Import the list that I'll give you on Slack (poke me pls)
2. After importing the list, remove 2 subscribers permanently (mailpoet10 and mailpoet11)
3. Now import the same list again and you'll experience the issue

Noting that the issue is not present on production, so probably caused within this pr.

brezocordero:
Hi @veljkho !
I have been able to reproduce the error with your list.
The error happens when a subscriber has an empty Confirmation Time.
You can create the subscriber but there is an error if you try to update it.

It is not caused by this PR but you don't see it in production because this PR allows you to insert subscribers with empty IPs and in your file, the subscribers without confirmation time are the same as the ones without IPs.

If you add confirmation times to the subscribers in your list, the error goes away.

To reproduce it on production, create a file with one subscriber, with all fields except for confirmation time.
Import it and then import it again, you will see the error the second time.

Could you confirm? I will then create a new issue for it.
Thanks!

veljkho (approved):
Hi @brezocordero 
Okay, then we have a bug on prod after pushing this to prod, and since does not look high prior, we can probably merge this one and proceed with the new ticket.

brezocordero:
> then we have a bug on prod after pushing this to prod

@veljkho The bug was already in prod, before publishing this PR but it was less visible because we were not importing the subscribers if the IPs were empty.

I have created the ticket: [MAILPOET-3751]

PR #3609, opened 2021-08-03
Outlook 2016 Rendering Bug in forwarded mail [MAILPOET-3468]

lysyjan:
[MAILPOET-3468]

[MAILPOET-3468]: https://mailpoet.atlassian.net/browse/MAILPOET-3468

QA: This PR contains a few rendering bug fixes, please take a look at rendering emails in different browsers. I have tried to compare master and this branch on Litmus. Unfortunately, I don't have access to Outlook 2016 with the possibility of redirecting.

### DIFF
--- a/lib/Config/Menu.php
+++ b/lib/Config/Menu.php
@@ -96,7 +96,7 @@ class Menu {
            . '|Raleway:400,400i,700,700i'
            . '|Permanent+Marker:400,400i,700,700i'
            . '|Pacifico:400,400i,700,700i';
-          echo '<!--[if !mso]><link href="https://fonts.googleapis.com/css?family=' . $fonts . '" rel="stylesheet"><![endif]-->';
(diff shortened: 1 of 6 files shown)


lib/Config/Menu.php
```diff
@@ -96,7 +96,7 @@ public function setup() {
            . '|Raleway:400,400i,700,700i'
            . '|Permanent+Marker:400,400i,700,700i'
            . '|Pacifico:400,400i,700,700i';
-          echo '<!--[if !mso]><link href="https://fonts.googleapis.com/css?family=' . $fonts . '" rel="stylesheet"><![endif]-->';
+          echo '<!--[if !mso]><!-- --><link href="https://fonts.googleapis.com/css?family=' . $fonts . '" rel="stylesheet"><!--<![endif]-->';
```
costasovo, line 99: This code is rendered in the browser. I think we should remove the HTML condition completely from this place and and echo the link without the condition

costasovo (changes requested):

costasovo (approved):

veljkho (approved):
Hi @lysyjan I dont have the access either. Just tested with Litmus.

PR #3610, opened 2021-08-03
Skip StatPage acceptance test when Premium is active [MAILPOET-3723]

lysyjan:
[MAILPOET-3723]

[MAILPOET-3723]: https://mailpoet.atlassian.net/browse/MAILPOET-3723

### DIFF
--- a/tests/_support/AcceptanceTester.php
+++ b/tests/_support/AcceptanceTester.php
@@ -345,7 +345,7 @@ class AcceptanceTester extends \Codeception\Actor {
     $i->cli(['plugin', 'deactivate', self::WOO_COMMERCE_SUBSCRIPTIONS_PLUGIN]);
   }
 
-  public function isPluginActive(string $plugin): bool {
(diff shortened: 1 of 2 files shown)


tests/_support/AcceptanceTester.php
```diff
@@ -345,6 +345,11 @@ public function deactivateWooCommerceSubscriptions() {
     $i->cli(['plugin', 'deactivate', self::WOO_COMMERCE_SUBSCRIPTIONS_PLUGIN]);
   }
 
+  public function isPluginActive(string $plugin): bool {
```
costasovo, line 348: I think it would be nicer to use a name like `checkPluginIsActive` so that it reads better when used with the tester variable ($i). `$i->isPluginActive` vs `$i->checkPluginIsActive`. What do you think?

costasovo (commented):
Good find! I left just one comment about function name.

costasovo (approved):

PR #3620, opened 2021-08-10
Protect against fatal error when rendering e-mail styles [MAILPOET-3731]

rodrigoprimo:
This PR protects the code that renders styles for e-mails against a fatal error that could happen if the variable $styles passed to Renderer::renderStyles() is not an array of arrays. To do that, the code will now check if each value of the array is an array. If not, we move on to the next value instead of calling StylesHelper::setStyle(). I also added type hints to StylesHelper::setStyle() and some of the methods that it calls to make it more explicit what it expects to receive and what it returns.

See related Jira ticket for testing instructions.

[MAILPOET-3731]

[MAILPOET-3731]: https://mailpoet.atlassian.net/browse/MAILPOET-3731

### DIFF
--- a/lib/Newsletter/Renderer/Renderer.php
+++ b/lib/Newsletter/Renderer/Renderer.php
@@ -175,6 +175,11 @@ class Renderer {
           $selector = '.mailpoet_content-wrapper';
           break;
       }
+
+      if (!is_array($style)) {
+        continue;
+      }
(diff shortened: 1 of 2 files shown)


costasovo (approved):
Nice catch!

veljkho (approved):

PR #3622, opened 2021-08-12
Update jsdom to version 16.7.0

rodrigoprimo:
This PR updates jsdom to version 16.7.0. It also updates tests/javascript_newsletter_editor/mochaTestHelper.js to use the new jsdom API as the old API was removed.

As usual, please note that I'm not an experienced JS developer, so I'm unsure if my approach to this update, especially the changes to mochaTestHelper.js, is correct. After creating this PR, I will add some inline comments with more details about some of the changes that I made.

@costasovo, in the ticket description, you mentioned that after updating jsdom, we might be able to remove some things from mochaTestHelper.js. I removed everything that it seemed to me that it wouldn't break the tests. Was there anything else on your mind?

Also, it seems to me that we are using jsdom in a way that is not recommended (https://github.com/jsdom/jsdom/wiki/Don't-stuff-jsdom-globals-onto-the-Node-global), but I opted to preserve things for now as it is working and this part of the code is used only for testing the e-mail editor which we plan to replace with Gutenberg eventually.

[MAILPOET-3678]

[MAILPOET-3678]: https://mailpoet.atlassian.net/browse/MAILPOET-3678

### DIFF
--- a/tests/javascript_newsletter_editor/mochaTestHelper.js
+++ b/tests/javascript_newsletter_editor/mochaTestHelper.js
@@ -3,7 +3,7 @@ var sinon = require('sinon');
 var sinonChai = require('sinon-chai');
 var chaiJq = require('chai-jq');
 var _ = require('underscore');
-var jsdom = require('jsdom/lib/old-api.js').jsdom;
+var { JSDOM } = require('jsdom');
 var URL = require('url').URL;
 var nodeCrypto = require('crypto');
 
@@ -15,16 +15,17 @@ global.sinon = sinon;
 global.URL = URL;
 
 if (!global.document || !global.window) {
-  global.document = jsdom('<html><head><script></script></head><body></body></html>', {
-    url: 'http://example.com',
-  }, {
-    FetchExternalResources: ['script'],
-    ProcessExternalResources: ['script'],
-    MutationEvents: '2.0',
-    QuerySelector: false,
-  });
+  global.document = new JSDOM(
+    '<html><head><script></script></head><body></body></html>',
+    {
+      url: 'http://example.com',
+      runScripts: 'dangerously',
+      resources: 'usable',
+      pretendToBeVisual: true,
+    }
+  ).window.document;
 
-  global.window = document.defaultView;
+  global.window = global.document.defaultView;
   global.navigator = global.window.navigator;
 }
 const testHelpers = require('./loadHelpers.js');
@@ -38,8 +39,9 @@ global._ = _;
 testHelpers.loadScript('tests/javascript_newsletter_editor/testBundles/vendor.js', global.window);
 const Handlebars = global.window.Handlebars;
 global.Handlebars = global.window.Handlebars;
-// Fix global access for Element. It is used in tinymce
+// Fix global access for Element and Node. It is used in tinymce
 global.Element = global.window.Element;
+global.Node = global.window.Node;
 // Fix global access for HTMLAnchorElement. It is used in FileSaver
 global.HTMLAnchorElement = global.window.HTMLAnchorElement;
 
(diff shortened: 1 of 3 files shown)


tests/javascript_newsletter_editor/mochaTestHelper.js
```diff
@@ -15,16 +15,17 @@ global.sinon = sinon;
 global.URL = URL;
 
 if (!global.document || !global.window) {
-  global.document = jsdom('<html><head><script></script></head><body></body></html>', {
-    url: 'http://example.com',
-  }, {
-    FetchExternalResources: ['script'],
```
rodrigoprimo, line 21: jsdom doesn't accept a third parameter anymore and those configurations were replaced with `runScripts` and `resources`.

tests/javascript_newsletter_editor/mochaTestHelper.js
```diff
@@ -15,16 +15,17 @@ global.sinon = sinon;
...
-  });
-
-  global.window = document.defaultView;
+  global.document = new JSDOM(
+    '<html><head><script></script></head><body></body></html>',
+    {
+      url: 'http://example.com',
+      runScripts: 'dangerously',
+      resources: 'usable',
+      pretendToBeVisual: true,
```
rodrigoprimo, line 24: I added this to get rid of some warnings saying that doesn't support requestAnimationFrame.

tests/javascript_newsletter_editor/mochaTestHelper.js
```diff
@@ -38,8 +39,9 @@ global._ = _;
 testHelpers.loadScript('tests/javascript_newsletter_editor/testBundles/vendor.js', global.window);
 const Handlebars = global.window.Handlebars;
 global.Handlebars = global.window.Handlebars;
-// Fix global access for Element. It is used in tinymce
+// Fix global access for Element and Node. It is used in tinymce
 global.Element = global.window.Element;
+global.Node = global.window.Node;
```
rodrigoprimo, line 44: Without this line, I was getting the error below. I'm not sure what are the implications of adding this here.

```
ReferenceError: Node is not defined
    at eval (webpack-internal:///./node_modules/tinymce/tinymce.js:1493:78)
```

tests/javascript_newsletter_editor/mochaTestHelper.js
```diff
@@ -58,7 +60,6 @@ global.interact = function () {
 };
 global.spectrum = function () { return this; };
 jQuery.fn.spectrum = global.spectrum;
-jQuery.fn.stick_in_parent = function () { return this; };
```
rodrigoprimo, line 61: Removed as the test work without it.

tests/javascript_newsletter_editor/mochaTestHelper.js
```diff
@@ -127,6 +118,8 @@ global.window.matchMedia = window.matchMedia || (
   }
 );
 
+global.window.fontsSelect = function () {};
```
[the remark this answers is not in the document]
costasovo: `views/newsletter/editor.html` is probably really not included. I see that we have some [loadTemplate](https://github.com/mailpoet/mailpoet/blob/master/tests/javascript_newsletter_editor/loadHelpers.js#L20-L26) function but that one loads only templates in `views/newsletter/templates/`. I think this workaround is fine.

tests/javascript_newsletter_editor/mochaTestHelper.js
costasovo: I assume this happens because tinymce access the Node and Element properties from global space so it works for browser but not in Node JS. I'm not sure about the implications either.

costasovo (approved):
👍 Looks good to me!

veljkho (approved):

PR #3624, opened 2021-08-13
Unify duplicated API error notices [MAILPOET-2844]

brezocordero:
To reproduce the duplicated notices without this PR:
1) Validate your MailPoet API Key
2) Schedule some type of email to happen soon. (I modified the scheduled time on table `wp_mailpoet_scheduled_tasks`)
3) Enter an invalid MailPoet key and click on validate. The notice on top appears when you are on Newsletters, Subscribers or Segments. If you are editing a newsletter, on step Send, the same notice is displayed twice.
4) Wait for the scheduled email to try going through. The second notice appears.

To validate the key from my local test site I used ngrok.
To connect my local Docker dev to ngrok I did the following:
1) Added this to wp-config.php
```
define( 'DOCKER_REQUEST_URL', ( ! empty( $_SERVER['HTTPS'] ) ? 'https://' : 'http://' ) . ( ! empty( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : 'localhost' ) );
define( 'WP_SITEURL', DOCKER_REQUEST_URL );
define( 'WP_HOME', DOCKER_REQUEST_URL );
```
2) Created a new domain on ngrok. (ngrok is already installed and authorized in my machine)
3) Started the tunnel with `ngrok http 8002 -subdomain=my-subdomain`

I have a few questions, added on comments on the files.

[MAILPOET-2844]

[MAILPOET-2844]: https://mailpoet.atlassian.net/browse/MAILPOET-2844


views/newsletters.html
```diff
@@ -295,7 +295,7 @@
     'schedule': __('Schedule'),
 
     'allSendingPausedHeader': __('All sending is currently paused!'),
-    'allSendingPausedBody': __('Your key to send with MailPoet is invalid.'),
+    'allSendingPausedBody': __('Your [link]API key[/link] to send with MailPoet is invalid.'),
```
brezocordero, line 298: @pavel-mailpoet Do you create new strings or is it ok to modify the exiting one?

views/segments.html
```diff
@@ -99,7 +99,7 @@
     'numberOfItemsSingular': __('1 item'),
     'numberOfItemsMultiple': __('%$1d items'),
     'allSendingPausedHeader': __('All sending is currently paused!'),
-    'allSendingPausedBody': __('Your key to send with MailPoet is invalid.'),
+    'allSendingPausedBody': __('Your [link]API key[/link] to send with MailPoet is invalid.'),
```
brezocordero, line 102: @pavel-mailpoet What about moving  'allSendingPausedHeader', 'allSendingPausedBody' and 'allSendingPausedLink' from newsletters.html, subscribers.html and segments.html to layout.html? Or do they need to be on each file?

assets/js/src/newsletters/send.jsx
```diff
@@ -478,10 +477,6 @@ class NewsletterSend extends React.Component {
           onChange={this.handleFormChange}
           onSubmit={this.handleSave}
         >
-          <InvalidMssKeyNotice
```
brezocordero, line 481: This duplicated notice was not part of the issue but saw it during the development.
DuplicatedAllSending

views/newsletters.html
pavel-mailpoet: It is fine to modify the existing one. However, it is discouraged to do if you can. Obviously in this case you cannot avoid changing the translation string. All of the translations will be lost and the translators will have to translate them again. That will take some time and users will see this note in English for some time.

views/segments.html
pavel-mailpoet: It is fine to move them to `layout.html`. Good idea.

pavel-mailpoet (changes requested):
It looks good. I am returning it back to you to unify the strings into layout.html

pavel-mailpoet (approved):

veljkho (approved):

PR #3630, opened 2021-08-17
Ignore Html2Text::convert errors in confirmation email [MAILPOET-3756]

NeosinneR:
[MAILPOET-3756]

[MAILPOET-3756]: https://mailpoet.atlassian.net/browse/MAILPOET-3756

### DIFF
--- a/lib/Subscribers/ConfirmationEmailMailer.php
+++ b/lib/Subscribers/ConfirmationEmailMailer.php
@@ -104,7 +104,10 @@ class ConfirmationEmailMailer {
     );
 
     //create a text version. @ is important here, Html2Text throws warnings
-    $text = @Html2Text::convert((mb_detect_encoding($body, 'UTF-8', true)) ? $body : utf8_encode($body));
+    $text = @Html2Text::convert(
+      (mb_detect_encoding($body, 'UTF-8', true)) ? $body : utf8_encode($body),
+      true
+    );
 
     // build email data
     $email = [


lib/Subscribers/ConfirmationEmailMailer.php
```diff
@@ -104,7 +104,7 @@ public function sendConfirmationEmail(Subscriber $subscriber) {
     );
 
     //create a text version. @ is important here, Html2Text throws warnings
-    $text = @Html2Text::convert((mb_detect_encoding($body, 'UTF-8', true)) ? $body : utf8_encode($body));
+    $text = @Html2Text::convert((mb_detect_encoding($body, 'UTF-8', true)) ? $body : utf8_encode($body), true);
```
pavel-mailpoet, line 107: There are two other places where we use `Html2Text::convert` should we update those as well? 
Also, would it be better to format this line a bit differently? 
```
$text = @Html2Text::convert(
  (mb_detect_encoding($body, 'UTF-8', true)) ? $body : utf8_encode($body)
  , true
);
```
Not a big deal, but I think it improves the readability.

NeosinneR: I updated the code for better readability. 

I don't know what's better in the other two places - being consistent or changing the least of code not to introduce bugs. I decided for the latter, but I'm fine changing it everywhere if you think consistency is more important.

pavel-mailpoet (approved):

veljkho (approved):

PR #3632, opened 2021-08-18
First pass at fixing reply-to when sending WP e-mails [MAILPOET-3707]

rodrigoprimo:
I'm creating this PR as a draft to get some feedback on the approach before continuing. I'm not happy with the current solution, but it seems to me that any other approach will require changes to a core part of MailPoet, the MailPoet\Mailer\Mailer class, is likely to take a significant chunk of time, and I'm not sure if it is worth it.

As described in the linked task, the problem that this PR tries to solve is that, when MailPoet is used to send WP e-mails, it overwrites the reply-to address, and whatever is set in WP is ignored.

At first, I thought this would be a simple fix, but it turns out it is not, as there is no easy way in the MailPoet\Mailer\Mailer class to just set the reply-to address. Currently, this is done inside \MailPoet\Mailer\Mailer::init(), which takes care of several things.

The solution that I'm proposing here is ugly, but it doesn't change how the internals of the MailPoet\Mailer\Mailer class. It works in most cases, but there is one unwanted side effect that I will explain below that might be problematic enough to justify looking for another solution.

Since I already spend a significant amount of time on this task, I decided to create this draft to get some feedback before deciding on the next steps. More information and some questions are in the in-line comments below.

[MAILPOET-3707]


lib/Mailer/WordPress/WordPressMailer.php
```diff
@@ -51,6 +51,10 @@ public function send() {
     ];
 
     $sendWithMailer = function ($mailer) use ($email, $address, $extraParams) {
+      // we need to call Mailer::init() for every single WP e-mail to make sure reply-to is set
+      $replyTo = $this->getReplyToAddress();
+      $mailer->init(false, false, $replyTo);
```
rodrigoprimo, line 56: Since I couldn't find a way to just set the reply-to address as this is currently done inside \MailPoet\Mailer\Mailer::init(), I opted to call it here directly, passing the reply-to address (before init() would be called internally when the code calls \MailPoet\Mailer\Mailer::send() in the line below).

Besides being a weird design pattern to have to call init() multiple times just to set the reply-to address, the other problem that I see with this approach is that the same \MailPoet\Mailer\Mailer instance is shared across different parts of MailPoet. So setting the reply-to address here has the potential to impact other e-mails, and I don't think there is an easy way around this other than refactoring \MailPoet\Mailer\Mailer, which I'm hesitant to do as it is a core part of the plugin. For an example of how this change impacts other parts of the plugin, see the test https://github.com/mailpoet/mailpoet/pull/3632/files#diff-f92c8ae77731d86590a35535f19be126061e371e196cca07718583607a84b675R345 (make sure to see the code comment at the top of the test method).

I wonder if there is another way to fix this that I'm missing?

tests/integration/Mailer/WordPress/WordpressMailerTest.php
```diff
@@ -16,6 +20,9 @@ class WordpressMailerTest extends \MailPoetTest {
   public function _before() {
     $this->subscribersRepository = $this->diContainer->get(SubscribersRepository::class);
     $this->subscribersRepository->truncate();
+
+    $settings = SettingsController::getInstance();
```
rodrigoprimo, line 24: This is just temporary and is used in a test that needs to be refactored (or more likely moved somewhere else).

tests/integration/Mailer/WordPress/WordpressMailerTest.php
```diff
@@ -2,10 +2,14 @@
 
 namespace MailPoet\Mailer\WordPress;
 
+use MailPoet\DI\ContainerWrapper;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Mailer\Mailer;
 use MailPoet\Mailer\MailerError;
 use MailPoet\Mailer\MetaInfo;
+use MailPoet\Models\Subscriber;
```
rodrigoprimo, line 10: The new classes that are added here and are not related to WordPressMailer are just temporary and used in a test that needs to be refactored (or more likely moved somewhere else).

tests/integration/Mailer/WordPress/WordpressMailerTest.php
```diff
@@ -268,6 +275,131 @@ public function testItAddSubscriberMetaInfo() {
...
+   */
+  public function testItPreservesWPReplyTo() {
+    list($mailer, $fallbackMailer) = $this->getMailerInstancesForReplyToTests();
+
+    $wpMailer = new WordPressMailer($mailer, $fallbackMailer, new MetaInfo, $this->subscribersRepository);
+    $wpMailer->addAddress('email@example.com', 'Full Name');
+    $wpMailer->addReplyTo('reply-to@example.com', 'Reply To');
+    $wpMailer->From = 'email-from@example.com';
+    $wpMailer->Body = 'Body';
+    $wpMailer->send();
```
rodrigoprimo, line 289: All the tests that I created here are sending the e-mails as I didn't find a way to check the reply-to address of the e-mails without actually sending them. There might be a way to achieve this using PHPUnit mocks, but I couldn't find it on my first try, and I didn't want to spend more time on this task before getting some feedback.

tests/integration/Mailer/WordPress/WordpressMailerTest.php
```diff
@@ -268,6 +275,131 @@ public function testItAddSubscriberMetaInfo() {
...
+
+    $mailer = $this->getPrivateProperty($wpMailer, 'mailer');
```
rodrigoprimo, line 291: Similar to what I mentioned above, I don't like this part of the tests that I wrote as well, but I settled with it, for now, to not spend too much time on this task before getting some feedback.

I had to create a hack using reflection to get the value of an private property to check if the reply-to address is correct, which is an anti-pattern for tests. But again, I don't see another way to write the tests without refactoring the Mailer class.

tests/integration/Mailer/WordPress/WordpressMailerTest.php
```diff
@@ -268,6 +275,131 @@ public function testItAddSubscriberMetaInfo() {
...
+
+    $this->assertEquals('staff@mailinator.com', $mailer->replyTo['reply_to_email']);
+    $this->assertEquals('Sender', $mailer->replyTo['reply_to_name']);
+  }
+
+  public function testWPEmailReplyToShouldntAffectOtherEmails() {
```
rodrigoprimo, line 345: This test demonstrates the current limitation of the approach proposed here, which I'm inclined to think is a dealbreaker. See the comments below for instructions on how to check the test results as it doesn't contain any assertion.

lib/Mailer/WordPress/WordPressMailer.php
pavel-mailpoet: > Besides being a weird design pattern to have to call init() multiple times just to set the reply-to address

Sorry, I am not sure I understand this note. Where do you call it multiple times? I can see it being called on this line, where is the other call?

> other problem that I see with this approach is that the same \MailPoet\Mailer\Mailer instance is shared across different parts of MailPoet

We can prevent that. All services are configured as shared and by default. But we can change it here https://github.com/mailpoet/mailpoet/blob/master/lib/DI/ContainerConfigurator.php#L248 and add `->setShared(false)` and a new instance will be created every time the class is used.

lysyjan: > Sorry, I am not sure I understand this note. Where do you call it multiple times? I can see it being called on this line, where is the other call?

I guess the idea is that the `init`method is called per each callback execution.
In my opinion, it is not ideal, but it is better than refactoring the `\MailPoet\Mailer\Mailer` class.  So, I am okay with the current solution.

rodrigoprimo: Thanks for your comments, @pavel-mailpoet, and @lysyjan.

> Sorry, I am not sure I understand this note. Where do you call it multiple times? I can see it being called on this line, where is the other call?

I could have been more clear here. What I mean is that in a single request we sometimes send more than one WP e-mail and thus we would be calling init() multiple times. And that seems to me to be against what was intended when the Mailer class was designed, as inside send(), init() is only called if mailerInstance is not set (https://github.com/mailpoet/mailpoet/blob/03400bc4f7e55f1f41139d030d2cb6ddde7dc025/lib/Mailer/Mailer.php#L60).

> We can prevent that. All services are configured as shared and by default. But we can change it here https://github.com/mailpoet/mailpoet/blob/master/lib/DI/ContainerConfigurator.php#L248 and add ->setShared(false) and a new instance will be created every time the class is used.

That is good to know. I will explore this option. I imagined that the DI Container had a way to change this behavior, but I'm a bit reluctant to change it as there is a chance that this particular service is shared by design, and changing this behavior could break some other part of the code. Anyway, I will look into it as this should be enough to fix the last problem that exists in the implementation that I shared in this PR.

> I guess the idea is that the initmethod is called per each callback execution.
In my opinion, it is not ideal, but it is better than refactoring the \MailPoet\Mailer\Mailer class. So, I am okay with the current solution.

Thanks for your input. I will look into using setShared(false) as Pavel suggested as I believe this might be enough to make everything work. If there are no unwanted side effects, I also prefer this approach over refactoring the \MailPoet\Mailer\Mailer class.

PR #3634, opened 2021-08-19
Refactor ImportExport API to Doctrine [MAILPOET-3625]

pavel-mailpoet:
[MAILPOET-3625]

[MAILPOET-3625]: https://mailpoet.atlassian.net/browse/MAILPOET-3625

### DIFF
--- a/lib/Newsletter/Sending/ScheduledTasksRepository.php
+++ b/lib/Newsletter/Sending/ScheduledTasksRepository.php
@@ -46,19 +46,15 @@ class ScheduledTasksRepository extends Repository {
       ->getResult();
   }
 
-  /**
-   * @param string[] $types
-   * @return ScheduledTaskEntity|null
-   */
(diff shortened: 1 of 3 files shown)


lib/Newsletter/Sending/ScheduledTasksRepository.php
```diff
@@ -46,6 +46,23 @@ public function findByNewsletterAndSubscriberId(NewsletterEntity $newsletter, in
       ->getResult();
   }
 
+  /**
+   * @param string[] $types
+   * @return ScheduledTaskEntity|null
+   */
+  public function findScheduledOrRunningTask(array $types = []): ?ScheduledTaskEntity {
```
lysyjan, line 53: This method returns an entity or null. Why did you decide to add a parameter of type array here? I think that when somebody calls this method with two types, it will end with an error.

lysyjan (approved):
Thanks 👍🏻

veljkho (approved):

PR #3636, opened 2021-08-19
Include open and click user agents in Export Personal Data tool [MAILPOET-3738]

rodrigoprimo:
This PR adds a new "User-Agent" field to e-mail clicks data and it also adds a new section with e-mail opens when the WordPress' Export Personal Data tool is used. I also did some refactoring in the code that generates data for e-mail clicks to avoid duplicating it to add the new e-mail opens section. In the process, since I didn't want to add code that interacts with Paris for e-mail opens, I changed parts of the e-mail clicks code to use Doctrine. I meant to refactor the tests to use Doctrine instead of Paris as well but ran out of time.

[MAILPOET-3738]

[MAILPOET-3738]: https://mailpoet.atlassian.net/browse/MAILPOET-3738

### DIFF
--- a/lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterClicksExporter.php
+++ b/lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterClicksExporter.php
@@ -49,6 +49,18 @@ class NewsletterClicksExporter {
       'name' => WPFunctions::get()->__('URL', 'mailpoet'),
       'value' => $row['url'],
     ];
+
+    if (!is_null($row['user_agent'])) {
+      $userAgent = $row['user_agent'];
+    } else {
+      $userAgent = WPFunctions::get()->__('Unknown', 'mailpoet');
+    }
+
+    $newsletterData[] = [
+      'name' => WPFunctions::get()->__('User-agent', 'mailpoet'),
+      'value' => $userAgent,
+    ];
+
     return [
       'group_id' => 'mailpoet-newsletter-clicks',
       'group_label' => WPFunctions::get()->__('MailPoet Emails Clicks', 'mailpoet'),
--- a/lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterClicksExporter.php
+++ b/lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterClicksExporter.php
@@ -2,29 +2,29 @@
 
 namespace MailPoet\Subscribers\ImportExport\PersonalDataExporters;
 
-use MailPoet\Models\StatisticsClicks;
+use MailPoet\Statistics\StatisticsClicksRepository;
 use MailPoet\WP\Functions as WPFunctions;
 
 class NewsletterClicksExporter extends NewsletterStatsBaseExporter {
-  protected $statsClass = StatisticsClicks::class;
+  protected $statsClassName = StatisticsClicksRepository::class;
 
   protected function getEmailStats(array $row) {
     $newsletterData = [];
     $newsletterData[] = [
       'name' => WPFunctions::get()->__('Email subject', 'mailpoet'),
-      'value' => $row['newsletter_rendered_subject'],
+      'value' => $row['newsletterRenderedSubject'],
     ];
     $newsletterData[] = [
       'name' => WPFunctions::get()->__('Timestamp of the click event', 'mailpoet'),
-      'value' => $row['created_at'],
+      'value' => $row['createdAt']->format("Y-m-d H:i:s"),
     ];
     $newsletterData[] = [
       'name' => WPFunctions::get()->__('URL', 'mailpoet'),
       'value' => $row['url'],
     ];
 
-    if (!is_null($row['user_agent'])) {
-      $userAgent = $row['user_agent'];
+    if (!is_null($row['userAgent'])) {
+      $userAgent = $row['userAgent'];
     } else {
       $userAgent = WPFunctions::get()->__('Unknown', 'mailpoet');
     }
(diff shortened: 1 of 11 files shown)


tests/integration/Subscribers/ImportExport/PersonalDataExporters/NewsletterOpensExporterTest.php
```diff
@@ -0,0 +1,124 @@
...
+      'user_agent_id' => $userAgentId,
+    ]);
+  }
+
+  public function _after() {
+    $this->truncateEntity(SubscriberEntity::class);
+    $this->truncateEntity(SendingQueueEntity::class);
+    $this->truncateEntity(NewsletterEntity::class);
+    $this->truncateEntity(UserAgentEntity::class);
+    ORM::raw_execute('TRUNCATE ' . StatisticsOpens::$_table);
```
pavel-mailpoet, line 122: I know you assigned this PR to @costasovo, But because he is on vacation, I took a quick look before he can prop

I don't understand why you are using the old models in the tests. Here you could use the same method `truncateEntity` as you did on the previous lines. And also, for creating the data, you used the old models. We shouldn't use the old models in the new code because we are creating new technical debt. Someone in the future would have to come back and do the refactor. I can see your note in the description:

> I meant to refactor the tests to use Doctrine instead of Paris as well but ran out of time.

But I am not sure I understand. What time do you have in mind? Is there a timebox on this feature?

rodrigoprimo: > I know you assigned this PR to @costasovo, But because he is on vacation, I took a quick look before he can prop

I opted to assign to Rosta even though he was on vacation as Friday was his last day off, and I'm not working on Fridays. So a review this week was not a problem. I also opted to do this since Jan is on vacation this week, and you are on L2 support. Anyway, thanks for jumping in and reviewing the PR, @pavel-mailpoet.

> I don't understand why you are using the old models in the tests. Here you could use the same method truncateEntity as you did on the previous lines. And also, for creating the data, you used the old models. We shouldn't use the old models in the new code because we are creating new technical debt. Someone in the future would have to come back and do the refactor.

That was a slip on my part. While working on this last week, I was under the false impression that there was not a `StatisticsOpenEntity` class which I can see now is not the case. I believe what happened is that to check for the existence of the class, I used the auto-complete feature of my IDE without realizing that the Paris models use the plural form and the Doctrine models use the singular form. So I was probably looking for `StatisticsOpensEntity`, and that is why I found nothing. I then opted to leave the code as is since I didn't want to change the focus of this task to work on creating a new Doctrine model.

Now that I can see that `StatisticsOpenEntity` exists, I pushed a new commit refactoring NewsletterClicksExporterTest.php and NewsletterOpensExporterTest.php.

> But I am not sure I understand. What time do you have in mind? Is there a timebox on this feature?

There was no timebox for this task. What I meant to say here is that I didn't have enough time to do the refactor (which I was wrongly assuming was a much more involved task than it actually is, as I explained above) before the end of my day. I thought it was not worth delaying this task because of a Doctrine related refactor since the functionality was finished.

lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterClicksExporter.php
```diff
@@ -49,6 +49,18 @@ private function exportNewsletter($row) {
       'name' => WPFunctions::get()->__('URL', 'mailpoet'),
       'value' => $row['url'],
     ];
+
+    if (!is_null($row['user_agent'])) {
+      $userAgent = $row['user_agent'];
+    } else {
+      $userAgent = WPFunctions::get()->__('Unknown', 'mailpoet');
```
pavel-mailpoet, line 29: You don't have to call `WPFunctions::get()->__` it was done automatically, but we normally use just `__`

lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterStatsBaseExporter.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace MailPoet\Subscribers\ImportExport\PersonalDataExporters;
+
+use MailPoet\Models\Subscriber;
+
+abstract class NewsletterStatsBaseExporter {
```
pavel-mailpoet, line 9: You don't need to rewrite it now. But we prefer composition over inheritance. So this would be just a regular class that would be used and called in `NewsletterOpensExporter` and `NewsletterClicksExporter`

pavel-mailpoet (approved):

veljkho (approved):

veljkho (dismissed):

lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterStatsBaseExporter.php
rodrigoprimo: I see. I don't have a strong preference. Could you point me to a place where I can learn more about the reasonings for preferring composition over inheritance?

lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterClicksExporter.php
rodrigoprimo: I will try to remember this. I think in this case I simply copied what was already being used in this file. Why we use `WPFunctions::get()` for all WP functions except `__()`?

veljkho (approved):

lib/Subscribers/ImportExport/PersonalDataExporters/NewsletterStatsBaseExporter.php
rodrigoprimo: Thanks for the additional details. I will take a look at the links that you mentioned.

PR #3637, opened 2021-08-20
Differentiate machine opens and clicks [MAILPOET-3736]

lysyjan:
[MAILPOET-3736]

I tried to test the performance on my local with an old mailpoet.com export. It looked fine, but I'm still slightly afraid about recreating indexes in bigger databases. Also, I had to add deleting old indexes because our migration tool doesn't support it.

[MAILPOET-3736]: https://mailpoet.atlassian.net/browse/MAILPOET-3736

### DIFF
--- a/lib/Config/Migrator.php
+++ b/lib/Config/Migrator.php
@@ -68,6 +68,7 @@ class Migrator {
     }
     $this->updateNullInUnsubscribeStats();
     $this->fixScheduledTasksSubscribersTimestampColumns();
+    $this->removeDeprecatedStatisticsIndexes();
     return $output;
   }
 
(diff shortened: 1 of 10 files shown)


lib/Config/Migrator.php
```diff
@@ -427,11 +428,12 @@ public function statisticsClicks() {
       'queue_id int(11) unsigned NOT NULL,',
       'link_id int(11) unsigned NOT NULL,',
       'user_agent_id int(11) unsigned NULL,',
+      'user_agent_type varchar(50) NULL,',
```
pavel-mailpoet, line 431: Would it be better to use an enum column here? This table was fixed length before this change. Now it isn't. I know there is a difference in MySQL. I don't remember exactly what the connotations are.

Also, the user agent strings are stored in the user agents table. Why is the type stored here? Why don't we store the type in the user agents table?

lysyjan: You are right with the column type. I would change the type on the column on tinyint with constants in the code.

> Also, the user agent strings are stored in the user agents table. Why is the type stored here? Why don't we store the type in the user agents table?

This was recommended in the ticket description, and it makes sense to me when we want to have a good performance on bigger sites. But we can't be sure that the join would have been a problem.

pavel-mailpoet (approved):

veljkho (approved):

PR #3642, opened 2021-08-25
Newsletter listing stats UI changes [MAILPOET-3739]

costasovo:
[MAILPOET-3739]

Note: Point 2c was already done in https://mailpoet.atlassian.net/browse/MAILPOET-3736

[MAILPOET-3739]: https://mailpoet.atlassian.net/browse/MAILPOET-3739

### DIFF
--- a/assets/js/src/common/listings/newsletter_stats.tsx
+++ b/assets/js/src/common/listings/newsletter_stats.tsx
@@ -25,27 +25,27 @@ const NewsletterStats = ({
   const openedDisplay = MailPoet.Num.toLocaleFixed(opened, 1);
   const clickedDisplay = MailPoet.Num.toLocaleFixed(clicked, 1);
 
-  let openedAndClickedStats = (
+  let clickedAndOpenedStats = (
     <div className="mailpoet-listing-stats-opened-clicked">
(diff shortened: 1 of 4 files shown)


assets/js/src/common/listings/newsletter_stats.tsx
```diff
@@ -25,23 +25,17 @@ const NewsletterStats = ({
...
+  let clickedAndOpenedStats = (
     <div className="mailpoet-listing-stats-opened-clicked">
       <div className="mailpoet-listing-stats-percentages">
-        {openedDisplay}
+        {clickedDisplay}
         %
         <br />
-        {clickedDisplay}
+        <span className="mailpoet-listing-stats-percentages-opens">{openedDisplay}</span>
         %
```
rodrigoprimo, line 35: I believe we should put this percentage symbol inside span.mailpoet-listing-stats-percentages-opens so that it is displayed without bold as well?

rodrigoprimo (changes requested):
Looks good to me. There is change one minor change that I believe we still need to do before approving this PR. See comment below.

veljkho (approved):

PR #3650, opened 2021-08-31
Machine opens segment [MAILPOET-3737]

pavel-mailpoet:
[MAILPOET-3737]

[MAILPOET-3737]: https://mailpoet.atlassian.net/browse/MAILPOET-3737


costasovo (commented):
I just have one question related to testing `opens.user_agent_type` for NULL value which I think is not necessary. See my comment.

costasovo (dismissed):

veljkho (changes requested):
Everything is verified except point 4.a:
I see 0 subscribers when I make a segment with that type, but it should show definitely some higher value, because when I make segment with machine-opened within specific newsletter, I could see 3 subscribers (and it is correct, confirmed via newsletter stats).

veljkho (approved):

PR #3656, opened 2021-09-04
Throttling also for logged-in users [MAILPOET-3790]

brezocordero:
Skips IP throttling check only for WordPress users with roles administrator or editor. 
Uses a similar [implementation as for captcha](https://github.com/mailpoet/mailpoet/blob/master/lib/Subscription/Captcha.php#L82-L89).

Adds a filter `mailpoet_subscription_throttling_exclude_roles` allowing to set which roles will be allowed to skip the throttling.

Test with:
`./do --test "test:integration --file=tests/integration/Subscription/ThrottlingTest.php"
`

[MAILPOET-3790]

[MAILPOET-3790]: https://mailpoet.atlassian.net/browse/MAILPOET-3790

### DIFF
--- a/lib/Subscription/Throttling.php
+++ b/lib/Subscription/Throttling.php
@@ -27,7 +27,7 @@ class Throttling {
 
     $subscriberIp = Helpers::getIP();
 
-    if ($subscriptionLimitEnabled && !$this->wp->isUserLoggedIn()) {
+    if ($subscriptionLimitEnabled && !$this->isUserExemptFromThrottling()) {
       if (!empty($subscriberIp)) {
         $subscriptionCount = $this->subscriberIPsRepository->getCountByIPAndCreatedAtAfterTimeInSeconds($subscriberIp, $subscriptionLimitWindow);
         if ($subscriptionCount > 0) {
@@ -69,5 +69,14 @@ class Throttling {
       'seconds' => $sec ? sprintf(__('%d seconds', 'mailpoet'), $sec) : '',
     ];
     return join(' ', array_filter($result));
+  }
+
+  private function isUserExemptFromThrottling(): bool {
+    if (!$this->wp->isUserLoggedIn()) {
+      return false;
+    }
+    $user = $this->wp->wpGetCurrentUser();
+    $roles = $this->wp->applyFilters('mailpoet_subscription_throttling_exclude_roles', ['administrator', 'editor']);
+    return !empty(array_intersect($roles, (array)$user->roles));
   }
 }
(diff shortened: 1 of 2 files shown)


rodrigoprimo (commented):
The code changes look good to me. Nice work!

There is one thing that I wonder if it is worth changing or not. We use just the IP address to identify a subscription attempt. We get the IP address from $_SERVER['REMOTE_ADDR']. The value of this variable can be the same for multiple users (because they are on the same network, or because the site uses something like Cloudflare and $_SERVER['REMOTE_ADDR'] resolves to a Cloudflare IP address).

If I'm not missing something, this means that we might inadvertently throttle single and valid attempts to subscribe from different logged-in users if they share the same IP address. I'm not sure if the benefits we get from this change (making it harder to use MailPoet to spam mailboxes with confirmation emails) are worth potentially making it harder for real users to subscribe to a list. I wonder if, when deciding whether to throttle or not for logged-in users, we could check not only the IP address but also the ID of the user?

I'm not sure if we should do this or not. Asking just to start a conversation for us to decide what is the best path forward.

brezocordero:
> we might inadvertently throttle single and valid attempts to subscribe from different logged-in users if they share the same IP address.
> I wonder if, when deciding whether to throttle or not for logged-in users, we could check not only the IP address but also the ID of the user?

@rodrigoprimo Thank you for the questions. Those are good and valid points. 

[WordPress core uses the user ID, instead of the IP, if available, to prevent comment flood.](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/comment.php#L887)
We would need to modify the Entity/DB model [SubscriberIPEntity](SubscriberIPEntity) to add the `user_id` and the function [throttle](https://github.com/mailpoet/mailpoet/blob/5a5ec1cc19b20d8d94f97f8b1df3162690e236df/lib/Subscription/Throttling.php#L22) to record it or use it when applicable.

@flakas or @NeosinneR what is your opinion on implementing the additional check for logged-in users?
I could also create a new issue and do it in a second step.

brezocordero:
@rodrigoprimo after talking with @flakas we will go ahead with the throttling using the IP.

We can revisit it in the future if needed.

rodrigoprimo (approved):

flakas:
@rodrigoprimo Just to provide some context, I am concerned about potential abuse stemming from attackers utilizing multiple WordPress accounts.

We know spambots can and do sign up and use WP accounts to generate spam.
The risk of the WP approach may be acceptable for content that remains on the site (comments usually don't trigger emails to anyone else, except maybe the admin), while subscription forms have more significant external impact potential and end up in spam for innocent victims.

veljkho (approved):

PR #3661, opened 2021-09-07
Add classes to Woo checkout opt-in to match other checkboxes [MAILPOET-3498]

costasovo:
I couldn't find some nice way of telling WooCommerce to use "standard" checkbox classes so I'm passing them via related options.
[MAILPOET-3498]

[MAILPOET-3498]: https://mailpoet.atlassian.net/browse/MAILPOET-3498

### DIFF
--- a/lib/WooCommerce/Subscription.php
+++ b/lib/WooCommerce/Subscription.php
@@ -82,6 +82,8 @@ class Subscription {
       [
         'type' => 'checkbox',
         'label' => $this->wp->escHtml($labelString),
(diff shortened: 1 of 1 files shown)


rodrigoprimo (approved):
Looks good to me!

> I couldn't find some nice way of telling WooCommerce to use "standard" checkbox classes so I'm passing them via related options.

Yeah, AFAIK, there is no way to do that so we have to approach it the way you did.

veljkho (approved):

PR #3663, opened 2021-09-08
Load custom fonts in chunks [MAILPOET-3791]

lysyjan:
We have a conflict with WooCommerce Payments plugin. When the file with custom fonts is bigger than 10kB, the payment form isn't loaded.

[MAILPOET-3791]

The fix is available on[ my testing site](https://lysyjan.wpcomstaging.com/).

[MAILPOET-3791]: https://mailpoet.atlassian.net/browse/MAILPOET-3791

### DIFF
--- a/lib/Form/Util/CustomFonts.php
+++ b/lib/Form/Util/CustomFonts.php
@@ -5,6 +5,7 @@ namespace MailPoet\Form\Util;
 use MailPoet\WP\Functions;
 
 class CustomFonts {
+  const FONT_CHUNK_SIZE = 25;
   const FONTS = [
     'Abril FatFace',
     'Alegreya',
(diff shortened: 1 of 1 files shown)


lib/Form/Util/CustomFonts.php
```diff
@@ -80,14 +81,16 @@ public function __construct(Functions $wp) {
   public function enqueueStyle() {
     $displayCustomFonts = $this->wp->applyFilters('mailpoet_display_custom_fonts', true);
     if ($displayCustomFonts) {
-      $this->wp->wpEnqueueStyle('mailpoet_custom_fonts', $this->generateLink());
+      foreach (array_chunk(self::FONTS, self::FONT_CHUNK_SIZE) as $key => $fonts) {
```
pavel-mailpoet, line 84: I think it would be better to leave a comment here explaining why are we using chunks here. 

I don't understand it either. How does that help? The fonts appear on the page anyway, don't they?

lysyjan: > I think it would be better to leave a comment here explaining why are we using chunks here.

Right, I will do it.

>I don't understand it either. How does that help? The fonts appear on the page anyway, don't they?

I don't know why, but when we load all fonts in one request, then the Stripe SDK doesn't work correctly, and the form from WC Payments isn't displayed. WC Payments gets all fonts from the website and adds them into the SDK. Maybe is the file size a problem. I'm not sure about the exact reason in the SDK.

pavel-mailpoet: > I don't know why, but when we load all fonts in one request, then the Stripe SDK doesn't work correctly, and the form from WC Payments isn't displayed.

Are you sure this really fixes the problem, then? Could it only work on your website and on a different hardware setup this would still be a problem? Maybe with a different laptop, browser? 

I am not sure if using this fix is a good idea. We [were already told](https://mailpoetp2.wordpress.com/2021/09/02/core-web-vitals-cwv-for-www-mailpoet-com-august-2021/) that we need to think about the frontend speed. This would create more requests and this change slows the page down.

pavel-mailpoet (approved):

veljkho (approved):

PR #3670, opened 2021-09-12
Make woocommerce subscription opt-in label translatable

davethehacker:
By changing to the `esc_html__` method, the checkbox label can be translated by manually creating a translation entry in a translation plugin of choice. If no matching translation is available, the behavior is unchanged.

### DIFF
--- a/lib/WP/Functions.php
+++ b/lib/WP/Functions.php
@@ -147,6 +147,10 @@ class Functions {
   public function escHtml($text) {
     return esc_html($text);
   }
+  
+  public function esc_html__($text, $domain) {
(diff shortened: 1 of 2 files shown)


lysyjan:
Hi there @davethehacker!
Thank you for your pull request.

We register all WordPress functions, which are used in our code, to [the Functions class](https://github.com/mailpoet/mailpoet/blob/master/lib/WP/Functions.php).
Please, can you add the function `esc_html__` to this class too?

We appreciate your help!

davethehacker:
> Please, can you add the function `esc_html__` to this class too?

Alright, I did that.

PR #3672, opened 2021-09-14
Re engagement email [MAILPOET-3763]

pavel-mailpoet:
[MAILPOET-3763]

[MAILPOET-3763]: https://mailpoet.atlassian.net/browse/MAILPOET-3763


assets/js/src/newsletters/types/re_engagement/re_engagement.tsx
```diff
@@ -1,21 +1,34 @@
...
+  settings: {
+    deactivate_subscriber_after_inactive_days: string;
+  }
+}
+
+declare let window: ReengagementWindow;
 
 export function NewsletterTypeReEngagement(): JSX.Element {
+  const [options] = useState({
+    afterTimeNumber: window.settings.deactivate_subscriber_after_inactive_days,
```
costasovo: I think we wanted to avoid adding new usages of the global window within components. What about adding the settings object to the MailPoet module? 

I think we already have a type definition for settings https://github.com/mailpoet/mailpoet/blob/2860db41e588400d8754dc6273fa7839b4f648d0/assets/js/src/settings/store/types.ts#L1-L137

assets/js/src/newsletters/send/re_engagement.tsx
```diff
@@ -0,0 +1,192 @@
...
+import GATrackingField from './ga_tracking';
+import SenderField from './sender_address_field';
+
+interface ReengagementWindow extends Window {
+  settings: {
+    deactivate_subscriber_after_inactive_days: string;
+  }
+}
+
+declare let window: ReengagementWindow;
```
costasovo, line 15: We could [reuse the type definition of settings object](https://github.com/mailpoet/mailpoet/blob/2860db41e588400d8754dc6273fa7839b4f648d0/assets/js/src/settings/store/types.ts#L1-L137) and also pass the settings object via MailPoet module as we do for other static values.

assets/js/src/newsletters/listings/re_engagement.jsx
```diff
@@ -11,9 +13,9 @@ import {
   checkMailerStatus,
 } from 'newsletters/listings/utils.jsx';
 import NewsletterTypes from 'newsletters/types';
-
-import classNames from 'classnames';
-import MailPoet from 'mailpoet';
+import ReactStringReplace from 'react-string-replace';
+import Tags from '../../common/tag/tags';
+import { ScheduledIcon } from '../../common/listings/newsletter_status';
```
costasovo, line 18: We prefer using the absolute path for imports like these

costasovo (changes requested):
Good job! I added just a couple of remarks.

costasovo (approved):
👍

veljkho (changes requested):
1. Getting blank page on email last step page.
2. I will mention also that if I set `12 months` in `Settings` > `Advanced` tab for _Stop sending to inactive subscribers_ I will see `11.166666666666666` pre-filled in [this field](https://d.pr/i/uGMgen). Then if I enter 13 number in the same field, I will see `12.166666666666666` in the warning message below. Is this expected somehow?
3. If I set `Never` in the `Settings` > `Advanced` tab, then I'll see [a warning message with 0 months](https://d.pr/i/BbABnT).

pavel-mailpoet:
@veljkho I fixed numbers 2. and 3. from your list. I cannot replicate the first item. So maybe it is fixed with the remaining changes. Or if not, please could you provide more details? When does that happen? Do you see any errors in the console?

costasovo (dismissed):

veljkho (changes requested):
I dont see anymore issue with the last email step 👍 
But I see now another issue. It's about a [warning message that is no anymore present](https://d.pr/i/5m3QHp). The same is with the last email step, no warning message present anymore. Everything else verified.

costasovo (dismissed):

veljkho (changes requested):
Previous issue fixed 👍 
Now issue with `12.66666666` is back! [Screenshot](https://d.pr/i/tYGvuN)
Interestingly, issue is not present for 6 or 3 months.. just for 12 months.

On another note, I don't know why I need now to enter [6 months to see the warning message for the user?](https://d.pr/i/zyxAj5)
Previously (also from the spec's image) the warning message has shown when I entered 7 value (of course this depends on what we have set in the settings). The same is happening for 3 months option, but interestingly not happening for 12 months option.

costasovo (dismissed):

veljkho (changes requested):
Sorry, the last issue is that there's [no warning message present if I set 12 value](https://d.pr/i/LFzKtu). It should start showing the warning message with equal of higher from what we've set in the settings, in this case 12 months. It is working with 3 and 6 months.

assets/js/src/newsletters/types/re_engagement/scheduling.tsx
```diff
@@ -25,6 +25,9 @@ export function Scheduling({
 }: Props): JSX.Element {
   const daysInPeriod = afterTimeType === 'weeks' ? 7 : 30;
   const daysSelected = Number(afterTimeNumber) * daysInPeriod;
+  let inactivePeriod = inactiveSubscribersPeriod;
+  // so that we can calculate the dates using
```
costasovo, line 29: I don't understand this comment. It feels like part of it is missing. I see from the commit message that it has something to do with `multiplication of 30` but I don't get why we need it. Could you please update the comment. Also it seems that if a user selects 361 days they won't see the warning. But I guess that is a tradeoff we could accept.

costasovo (approved):

veljkho (approved):

PR #3673, opened 2021-09-14
Add custom validation for names in forms [MAILPOET-3786]

lysyjan:
[MAILPOET-3786]

[MAILPOET-3786]: https://mailpoet.atlassian.net/browse/MAILPOET-3786

### DIFF
--- a/assets/js/src/public.jsx
+++ b/assets/js/src/public.jsx
@@ -1,11 +1,34 @@
 import MailPoet from 'mailpoet';
 import jQuery from 'jquery';
 import Cookies from 'js-cookie';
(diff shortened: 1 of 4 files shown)


pavel-mailpoet (approved):

veljkho (changes requested):
I was able to enter `www.domain.com/page`... without `HTTP` protocol. Can we forbid also without a protocol?

lysyjan:
@veljkho I think that after a discussion in Jira, we can accept the current solution.

veljkho (approved):

PR #3678, opened 2021-09-15
Update Webpack to v5 [MAILPOET-3214]

costasovo:
This PR updates Webpack and its modules.

I had to update the config file and fix a bunch of small issues. I tried to explain most of the changes in commit messages. So I
recommend reviewing it commit by commit. 

The most significant change is that I changed how we split code into chunks. Instead of the custom `admin_vendor_chunk` which had quite a complex test function for selecting modules to include I used a simple configuration for creating a commons.js chunk that contains all modules shared at least once. I found this simple config much more efficient than `admin_vendor_chunk`. It reduced the size of JS we serve by a couple of MBs (unzipped) and the size of plugin ZIP went down by 1.1MB.

Please note I'm not a Webpack expert and I had to check and learn a lot during working on this update so there might be some better approaches for some fixes I made.

There is also [a related PR in the premium repository](https://github.com/mailpoet/mailpoet-premium/pull/490). These 2 must be released together.

[MAILPOET-3214]

[MAILPOET-3214]: https://mailpoet.atlassian.net/browse/MAILPOET-3214


tasks/npm_post_install.sh
```diff
@@ -21,3 +21,7 @@ sed -i -- "s/_focusedField\.focus()/_focusedField\.trigger('focus')/g" node_modu
 # Github issue: https://github.com/WordPress/gutenberg/issues/32181
 sed -i -- "s/box-shadow: 0 0 0 2px-1px #007cba;/box-shadow: 0 0 0 calc(2px-1px) #007cba;/g" node_modules/@wordpress/block-editor/build-style/style.css
 sed -i -- "s/box-shadow: 0 0 0 var(--wp-admin-border-width-focus)-1px var(--wp-admin-theme-color);/box-shadow: 0 0 0 calc(var(--wp-admin-border-width-focus)-1px) var(--wp-admin-theme-color);/g" node_modules/@wordpress/block-editor/build-style/style.css
```
brezocordero, line 23:
```suggestion
```
Not related to the upgrade but do we still need this? https://github.com/WordPress/gutenberg/pull/32305/files

webpack.config.js
```diff
@@ -216,46 +258,25 @@ const adminConfig = {
...
-            // add admin/form_editor_legacy/newsletter_editor shared modules
-            const filteredChunks = chunks.filter((chunk) => {
-              return ['admin', 'newsletter_editor', 'form_editor', 'settings'].includes(chunk.name);
-            });
-            return filteredChunks.length > 1;
-          },
-          enforce: true,
-          chunks: (chunk) => ['admin_vendor', 'admin', 'newsletter_editor'].includes(chunk.name),
-          priority: 0,
+        commons: {
```
brezocordero, line 269: Looks like the custom `admin_vendor_chunk` was added to solve an issue with templates and translations being regenarated.
But I don't know how to check if this is still an issue. Maybe you have more context? https://github.com/mailpoet/mailpoet/commit/c934856f45d188b6074783b6cd002cd2e5918d74

package.json
```diff
@@ -127,7 +129,7 @@
     "@typescript-eslint/eslint-plugin": "^4.22.1",
     "@typescript-eslint/parser": "^4.22.1",
     "@wordpress/browserslist-config": "^4.0.0",
```
brezocordero, line 131:
```suggestion
    "@wordpress/browserslist-config": "^4.1.0",
```
Should we have the same version as in the premium plugin?

package.json
```diff
@@ -136,7 +138,7 @@
     "chai": "^4.3.4",
     "chai-jq": "^0.0.9",
     "clean-webpack-plugin": "^3.0.0",
```
brezocordero, line 140:
```suggestion
    "clean-webpack-plugin": "^4.0.0",
```
Should we have the same version as in the premium plugin?

webpack.config.js
```diff
@@ -378,19 +403,19 @@ const postEditorBlock = {
...
         seed: manifestSeed,
       })
     );
   }
-  // Clean output paths before build
-  if (config.output && config.output.path) {
-    del.sync([path.resolve(config.output.path, '**/*')]);
+  // Cleanup output path but only once to keep output files for subsequent builds
+  if (index === 0) {
+    config.plugins.unshift(new CleanWebpackPlugin());
```
brezocordero, line 418: Could we use https://webpack.js.org/configuration/output/#outputclean or are we on a use case that still needs   clean-webpack-plugin https://github.com/johnagan/clean-webpack-plugin/issues/197?

brezocordero (dismissed):
Great work @costasovo .
The config of the plugin is complex, so the upgrade is complex too.
Thanks for having detailed the changes in the commits, that was very useful.
You are probably more familiar with this Webpack config than I am but the changes looked good to me.
I have added few questions but it is more for my understanding.

tasks/npm_post_install.sh
costasovo: Good catch! I must have missed it when I was updating @wordpress dependencies. I'll remove it.

webpack.config.js
costasovo: I didn't know this option. Thanks! It will do the job!

package.json
costasovo: I decided to remove the plugin in favor of using output.clean setting.

webpack.config.js
costasovo: I don't have much more context. I haven't noticed any issues.

If I correctly understand [the original issue](https://github.com/webpack/webpack/issues/7230), it was happening when the bundle file was also an entry point. We use commons.js as a bundle with shared components and it isn't configured as an entry point and I think shouldn't be an entry point, but only contain modules that are used within some other entry points. I think that it is why we also need the admin_vendor entry point so that modules shared with the premium plugin are executed and exposed to the premium plugin.

webpack.config.js
costasovo: Hmm. After some testing, I found out that in free we probably can't use the clean-webpack-plugin or the output.clean option.

We need to share the same output path (assets/build/js) in more configs in order to generate a common manifest.json file. The cleanup plugin and also built-in output.clean setting does the cleanup on emitting and that causes that they delete files we want to keep even though I configured cleanup only for the first config. 

Using different output paths solves the problem with cleanup but it causes that we generate a bunch of manifest.json files instead of one.

As a simple workaround, I chose cleaning the output directory before we trigger the Webpack build in the Robofile.

costasovo:
@brezocordero Thanks for the review! Your comments were really helpful. I tried to address/answer all of them.

webpack.config.js
brezocordero: Sounds good to me! Thanks for testing the other option and for all the details

webpack.config.js
brezocordero: Thanks for taking the time to explain. Makes sense!

brezocordero (approved):

veljkho (approved):

PR #3680, opened 2021-09-15
Add bounce tab on stats page [PREMIUM-154]

lysyjan:
[PREMIUM-154]

[PREMIUM-154]: https://mailpoet.atlassian.net/browse/PREMIUM-154

### DIFF
--- a/views/newsletters.html
+++ b/views/newsletters.html
@@ -398,7 +398,7 @@
     'clickedLinks': __('Clicked Links'),
     'subscriberEngagement': __('Subscriber Engagement'),
     'googleAnalytics': __('GA campaign'),
-    'bounces': __('Bounces'),
+    'bounces': __('Bounces', 'A tab title for the list of bounces (w.wiki/45Qc)'),
 
     'premiumBannerCtaFree': __('Sign Up for Free', 'mailpoet'),
     'topBarLogoTitle': __('Back to section root'),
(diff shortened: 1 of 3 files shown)


views/newsletters.html
```diff
@@ -398,6 +398,7 @@
     'clickedLinks': __('Clicked Links'),
     'subscriberEngagement': __('Subscriber Engagement'),
     'googleAnalytics': __('GA campaign'),
+    'bounces': __('Bounces'),
```
pavel-mailpoet, line 401: I think this needs a comment for the translator. Imagine translating it to the Czech without any context. You wouldn't know what bounces they talk about and you would probably translate it like a ball bounces (https://d.pr/i/faPubH). That would be incorrect and possibly confusing for the user. 

We need to explain that this is an email that bounced and what that means. I would add a context like: "Those are emails that couldn't be delivered for some reason: email doesn't exist, or the mailbox is full." Maybe ask PM for the context.

pavel-mailpoet (changes requested):

views/newsletters.html
lysyjan: I asked PM for the context.
Thanks for the catch!

pavel-mailpoet (changes requested):
you need to use _x for the translation with a context

pavel-mailpoet (approved):

veljkho (approved):

PR #3688, opened 2021-09-20
Add confirmation before running integration tests on the dev site [MAILPOET-3755]

costasovo:
Instead of detecting the test environment, I tried to approach the problem from the other side. The idea is that a developer can mark an environment as "development site" and by doing so the environment will get protection against accidentally running DB cleanup within integration tests. Not everyone uses the MailPoet Dev Environment so I think that this way all developers could set it and use this protection.

There is a related [PR in MailPoet Dev Environment repository that adds the variable for the dev site container](https://github.com/mailpoet/mailpoet-dev-env/pull/43).

[MAILPOET-3755]

[MAILPOET-3755]: https://mailpoet.atlassian.net/browse/MAILPOET-3755

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -148,6 +148,12 @@ class RoboFile extends \Robo\Tasks {
   }
 
   public function testIntegration(array $opts=['file' => null, 'xml' => false, 'multisite' => false, 'debug' => false]) {
+    if (getenv('MAILPOET_DEV_SITE')) {
+      $run = $this->confirm("You are about to run tests on the development site. Your DB data will be erased. \nDo you want to proceed?", false);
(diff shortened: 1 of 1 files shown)


rodrigoprimo (approved):
Looks good to me! I like your approach of marking the dev env instead of detecting the test env.

I don't understand why Robo adds a question mark before the question itself (https://github.com/consolidation/Robo/blob/3.x/src/Common/IO.php#L214), but that is just a detail and outside of the scope of this PR.

rodrigoprimo (approved):
(I accidentally asked for a second review from myself :-))

veljkho (approved):

PR #3694, opened 2021-09-22
Refactor API subscriber fields methods to Doctrine [MAILPOET-3628]

pavel-mailpoet:
[MAILPOET-3628]

[MAILPOET-3628]: https://mailpoet.atlassian.net/browse/MAILPOET-3628

### DIFF
--- a/lib/API/MP/v1/CustomFields.php
+++ b/lib/API/MP/v1/CustomFields.php
@@ -1,0 +1,74 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\API\MP\v1;
+
+use MailPoet\CustomFields\ApiDataSanitizer;
+use MailPoet\Models\CustomField;
+
+class CustomFields {
+  /** @var ApiDataSanitizer */
+  private $customFieldsDataSanitizer;
+
+  public function __construct(ApiDataSanitizer $customFieldsDataSanitizer) {
+    $this->customFieldsDataSanitizer = $customFieldsDataSanitizer;
+  }
+
+  public function getSubscriberFields(): array {
+    $data = [
+      [
+        'id' => 'email',
+        'name' => __('Email', 'mailpoet'),
+        'type' => 'text',
+        'params' => [
+          'required' => '1',
+        ],
+      ],
+      [
+        'id' => 'first_name',
+        'name' => __('First name', 'mailpoet'),
+        'type' => 'text',
+        'params' => [
+          'required' => '',
+        ],
+      ],
+      [
+        'id' => 'last_name',
+        'name' => __('Last name', 'mailpoet'),
+        'type' => 'text',
+        'params' => [
+          'required' => '',
+        ],
+      ],
+    ];
+
+    $customFields = CustomField::selectMany(['id', 'name', 'type', 'params'])->findMany();
+    foreach ($customFields as $customField) {
+      $result = [
+        'id' => 'cf_' . $customField->id,
+        'name' => $customField->name,
+        'type' => $customField->type,
+      ];
+      if (is_serialized($customField->params)) {
+        $result['params'] = unserialize($customField->params);
+      } else {
+        $result['params'] = $customField->params;
+      }
+      $data[] = $result;
+    }
+
+    return $data;
+  }
+
+  public function addSubscriberField(array $data = []): array {
+    $customField = CustomField::createOrUpdate($this->customFieldsDataSanitizer->sanitize($data));
+    $errors = $customField->getErrors();
+    if (!empty($errors)) {
+      throw new APIException('Failed to save a new subscriber field ' . join(', ', $errors), APIException::FAILED_TO_SAVE_SUBSCRIBER_FIELD);
+    }
+    $customField = CustomField::findOne($customField->id);
+    if (!$customField instanceof CustomField) {
+      throw new APIException('Failed to create a new subscriber field', APIException::FAILED_TO_SAVE_SUBSCRIBER_FIELD);
+    }
+    return $customField->asArray();
+  }
+}
--- a/lib/API/MP/v1/CustomFields.php
+++ b/lib/API/MP/v1/CustomFields.php
@@ -3,14 +3,22 @@
 namespace MailPoet\API\MP\v1;
 
 use MailPoet\CustomFields\ApiDataSanitizer;
+use MailPoet\CustomFields\CustomFieldsRepository;
 use MailPoet\Models\CustomField;
 
 class CustomFields {
   /** @var ApiDataSanitizer */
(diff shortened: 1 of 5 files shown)


lib/API/MP/v1/CustomFields.php
```diff
@@ -41,18 +49,14 @@ public function getSubscriberFields(): array {
       ],
     ];
 
-    $customFields = CustomField::selectMany(['id', 'name', 'type', 'params'])->findMany();
+    $customFields = $this->customFieldsRepository->findAll();
```
rodrigoprimo, line 52: Before we were getting from the database just the columns 'id', 'name', 'type', 'params', and now we are getting all the fields from this particular table. This probably is not a big deal in this case as this table is unlikely to be bigger. But, as a good practice, what do you think of using another method that would let us specify which fields to retrieve and leave `findAll()` only for the instances when we want all the fields?

lib/API/MP/v1/CustomFields.php
```diff
@@ -41,18 +49,14 @@ public function getSubscriberFields(): array {
...
     foreach ($customFields as $customField) {
       $result = [
-        'id' => 'cf_' . $customField->id,
-        'name' => $customField->name,
-        'type' => $customField->type,
+        'id' => 'cf_' . $customField->getId(),
+        'name' => $customField->getName(),
+        'type' => $customField->getType(),
+        'params' => $customField->getParams(),
```
rodrigoprimo, line 58: Will $customField->getParams() handle both serialized and unserialized data correctly? I can see that it works for serialized data but I haven't checked unserialized data. Asking just to make sure that we don't need to treat the two cases differently as the old code did.

lib/API/MP/v1/CustomFields.php
```diff
@@ -64,15 +64,20 @@ public function getSubscriberFields(): array {
...
-    if (!empty($errors)) {
-      throw new APIException('Failed to save a new subscriber field ' . join(', ', $errors), APIException::FAILED_TO_SAVE_SUBSCRIBER_FIELD);
+    try {
+      $customField = $this->customFieldsRepository->createOrUpdate($this->customFieldsDataSanitizer->sanitize($data));
+    } catch (\Exception $e) {
+      throw new APIException('Failed to save a new subscriber field ' . join(', ', $e->getMessage()), APIException::FAILED_TO_SAVE_SUBSCRIBER_FIELD);
     }
-    $customField = CustomField::findOne($customField->id);
-    if (!$customField instanceof CustomField) {
+    $customField = $this->customFieldsRepository->findOneById($customField->getId());
```
rodrigoprimo, line 72: Why do we need to get the CustomFieldEntity again? I believe it was already returned by $this->customFieldsRepository->createOrUpdate().

tests/integration/API/MP/CustomFieldsTest.php
```diff
@@ -80,4 +81,30 @@ public function testItReturnsCustomFields() {
...
+      'name' => 'text custom field',
+      'type' => 'text',
+      'params' => [
+        'required' => '1',
+        'label' => 'text custom field',
+        'date_type' => 'year_month_day',
+      ],
+    ]);
+    expect($response)->array();
+    expect($response)->hasKey('name');
```
rodrigoprimo, line 96: What are your thoughts on making the assertions on this test more specific and check that the full response matches what was expected instead of just checking if the response is an array and has a 'name' key?

rodrigoprimo (commented):
Looks good to me. I left a few comments for us to discuss some details.

lib/API/MP/v1/CustomFields.php
pavel-mailpoet: This is the same as the old code was. So I left it there. We can remove it I guess. We usually did it to get all the fields that are generated by the database like `id` and `created_at`. But I am looking at Doctrine and it does fill the data automatically so I can remove this line.

tests/integration/API/MP/CustomFieldsTest.php
pavel-mailpoet: Well, my thinking was that if the array is not empty the method succeeded and so we don't need to assert all the fields like in the `get` method.

lib/API/MP/v1/CustomFields.php
pavel-mailpoet: I am not sure I understand correctly. The method `getParams` always returns `array` so I am not sure what to do here with unserialized data.

lib/API/MP/v1/CustomFields.php
pavel-mailpoet: I would argue it is a good practice in the Doctrine to load the complete entities for readability and maintainability purposes. If the entity is only partial, it can someone catch a surprise. If we need to optimise we can but I wouldn't do it when not needed.

rodrigoprimo: That makes sense

lib/API/MP/v1/CustomFields.php
rodrigoprimo: Maybe there is nothing that we need to do here. I just wanted to make sure that you have checked that this should work for all cases as the old code had an if condition to handle the data differently if it was serialized or not (https://github.com/mailpoet/mailpoet/pull/3694/files#diff-0958c60d6aae5b820e6f9c13761826370dd48694a799c74dd98573ee0e214e7bL91). It is possible that Doctrine is handling this for us behind the scenes or something else. I have not checked.

lib/API/MP/v1/CustomFields.php
rodrigoprimo: :+1:

tests/integration/API/MP/CustomFieldsTest.php
rodrigoprimo: I don't have a strong opinion on this one. I would've written an assertion to check if the method is returning exactly what we expect it to return, but if you prefer to leave it the way it is, it is fine as well.

My concern is that this test will pass even if in the future we inadvertently introduce a change that, for example, returns the key 'typo' instead of 'type' (ok, not the best example).

lib/API/MP/v1/CustomFields.php
pavel-mailpoet: I know that the old model worked differently. But the database contains always a serialised value: https://d.pr/i/fk8Q0p and the entity always returns [array](https://github.com/mailpoet/mailpoet/blob/8cfd47e6568d330067d722a51c2524254e5fc5f7/lib/Entities/CustomFieldEntity.php#L62). SO I am not sure what to check.

rodrigoprimo (approved):

veljkho (approved):

PR #3699, opened 2021-09-23
Refactor findDueByType() and findRunningByType() methods to use Doctrine instead of Paris [MAILPOET-2996]

rodrigoprimo:
This PR is the first one in a series of PRs to use ScheduledTaskEntity instead of ScheduledTask in CronWorkerRunner, SimpleWorker, and its child classes. Since those changes touch on a core part of MailPoet, I'm opting to break them into multiple somewhat small PRs.

This first one, refactors ScheduledTask::findDueByType() and ScheduledTask::findRunningByType() to use Doctrine instead of Paris. As part of this change, those two methods were moved to the class ScheduledTasksRepository. All the usages of those methods were updated.

To be able to create a PR just with the changes mentioned above, I had to create a helper method called convertTaskClass() inside CronWorkerRunner to convert a ScheduledTaskEntity object into a ScheduledTask object, as some parts of the code of this class still use the old model. This helper method will be removed in a subsequent PR once we finalize this refactor.

[MAILPOET-2996]

[MAILPOET-2996]: https://mailpoet.atlassian.net/browse/MAILPOET-2996


lib/Newsletter/Sending/ScheduledTasksRepository.php
```diff
@@ -87,6 +89,45 @@ public function findPreviousTask(ScheduledTaskEntity $task): ?ScheduledTaskEntit
...
+    if (is_null($status)) {
+      $queryBuilder->andWhere('st.status IS NULL');
+    } else {
+      $queryBuilder
+        ->andWhere('st.status = :status')
+        ->setParameter('status', $status);
+    }
+
+    if ($future) {
+      $queryBuilder->andWhere('st.scheduledAt >= :now');
```
costasovo, line 116: I noticed a small difference when comparing with [the old model code](https://github.com/mailpoet/mailpoet/blob/master/lib/Models/ScheduledTask.php#L189). In the old model, we use `whereGt` which is an equivalent for `>`. 

This is just a small difference but theoretically, it may change how the plugin triggers cron for jobs that are scheduled immediately using the current time. With `>` a significant portion of such jobs might be skipped in the current HTTP request (the one they were scheduled in) and triggered in a subsequent one.

Maybe I'm too much paranoid, but I would keep the behavior as close as possible to the current state. What do you think?

costasovo (commented):
Great job! Introducing these changes in the core gradually is a really great idea. 👍 
The PR looks good to me. I found just one thing which I'm not sure about. I left a comment.

lib/Newsletter/Sending/ScheduledTasksRepository.php
costasovo: Oh! I've just realized that it is the other way around :) Sorry! We trigger jobs cron for jobs that are due so this has nothing to do with the future jobs. But still, it might be better to keep those future and non-future intervals without an intersection.

rodrigoprimo: Good catch! I missed this when refactoring the code. I just pushed a new commit to fix this problem. Could you please review the PR again?

costasovo (approved):

veljkho (approved):

PR #3700, opened 2021-09-24
NPM update [MAILPOET-3724]

pavel-mailpoet:
I didn't update all dependencies. I left out `@wordpress` dependencies for a separate task. Some dependencies couldn't be updated because the versions are required by other dependencies.

Upgrading react-select is a separate [task](https://mailpoet.atlassian.net/browse/MAILPOET-3834).

[MAILPOET-3724]

[MAILPOET-3724]: https://mailpoet.atlassian.net/browse/MAILPOET-3724


package.json
```diff
@@ -151,14 +151,12 @@
     "exports-loader": "^3.0.0",
     "expose-loader": "^3.0.0",
     "grunt-cli": "^1.4.3",
-    "husky": "^6.0.0",
```
costasovo, line 154: This looks more like you removed husky instead of updating. I see only deletions even in package-lock.json. Is that expected?

costasovo (commented):
Looks good 👍 Nice that eslint forced adding those usecallback hooks 👍 

There is just one confusing commit message saying that husky was updated but it was uninstalled in the commit. I see that it was later added again so the final outcome is fine. It is just about the confusing commit message. What about changing the commit message from "updating" to "removing".

costasovo (approved):

veljkho (approved):

PR #3703, opened 2021-09-27
Fix flaky tests using  Scheduler::getScheduledTimeWithDelay [MAILPOET-3793]

brezocordero:
If the tests took too long to execute, they would fail because `Scheduler::getScheduledTimeWithDelay` would have a different time than the time set in the test. 

The tests could be made to fail consistently by inserting `sleep(60)` in between the function execution and the time we were setting `currentTime` within the test.

To solve this, I have added `WPFunctions` to the constructor, as an optional argument, so that we can mock `currentTime` inside of the tests.

The issue was originally to fix: `tests/integration/Newsletter/Scheduler/AutomaticEmailTest.php::testItSchedulesAutomaticEmailWhenConditionMatches`

I have also modified `WelcomeScheduler` class and the tests in `tests/integration/Newsletter/Scheduler/WelcomeTest.php`

[MAILPOET-3793]

[MAILPOET-3793]: https://mailpoet.atlassian.net/browse/MAILPOET-3793

### DIFF
--- a/lib/Newsletter/Scheduler/AutomaticEmailScheduler.php
+++ b/lib/Newsletter/Scheduler/AutomaticEmailScheduler.php
@@ -7,8 +7,18 @@ use MailPoet\Models\ScheduledTask;
 use MailPoet\Models\ScheduledTaskSubscriber;
 use MailPoet\Models\SendingQueue;
 use MailPoet\Tasks\Sending as SendingTask;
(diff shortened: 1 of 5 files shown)


rodrigoprimo (approved):
Looks good to me. Good job!

veljkho (approved):

PR #3705, opened 2021-09-28
Change SimpleWorker::prepareTaskStrategy() to use Doctrine [MAILPOET-3842]

rodrigoprimo:
This PR is a continuation of #3699. It should be reviewed only after #3699 is merged otherwise GitHub will display here all the commits from that PR as well. I decided to create this PR now in case #3699 is merged before I start my day tomorrow.

This PR changes the method prepareTaskStrategy() in SimpleWorker and all its child classes to use ScheduledTaskEntity instead of ScheduledTask. To do that, I also moved the method findCompletedByType() from ScheduledTask to ScheduledTaskEntity (as it is used in one of the modified child classes) and created a helper method for the tests to create ScheduledTaskEntity objects.

[MAILPOET-3842]

[MAILPOET-3842]: https://mailpoet.atlassian.net/browse/MAILPOET-3842

### DIFF
--- a/tests/DataFactories/ScheduledTask.php
+++ b/tests/DataFactories/ScheduledTask.php
@@ -15,12 +15,14 @@ use MailPoetVendor\Carbon\Carbon;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class ScheduledTask {
-  private $diContainer;
+  /**
+   * @var EntityManager
+   */
   private $entityManager;
 
   public function __construct() {
-    $this->diContainer = ContainerWrapper::getInstance();
-    $this->entityManager = $this->diContainer->get(EntityManager::class);
+    $diContainer = ContainerWrapper::getInstance();
+    $this->entityManager = $diContainer->get(EntityManager::class);
   }
 
   public function deleteAll() {
--- a/tests/DataFactories/ScheduledTask.php
+++ b/tests/DataFactories/ScheduledTask.php
@@ -8,15 +8,42 @@ use MailPoet\Cron\Workers\SendingQueue\Migration;
 use MailPoet\Cron\Workers\SubscriberLinkTokens;
 use MailPoet\Cron\Workers\UnsubscribeTokens;
 use MailPoet\Cron\Workers\WooCommercePastOrders;
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class ScheduledTask {
+  private $diContainer;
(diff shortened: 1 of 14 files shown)


tests/DataFactories/ScheduledTask.php
```diff
@@ -8,17 +8,44 @@
...
 use MailPoetVendor\Carbon\Carbon;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class ScheduledTask {
+  private $diContainer;
+  private $entityManager;
+
+  public function __construct() {
+    $this->diContainer = ContainerWrapper::getInstance();
+    $this->entityManager = $this->diContainer->get(EntityManager::class);
```
costasovo, line 23: I don't see any usage of `$this->diContainer` other than here in the constructor so I guess we might not need it as a property of this class. Or is here some reason why it is convenient that I'm missing?

tests/DataFactories/ScheduledTask.php
```diff
@@ -8,17 +8,44 @@
...
 use MailPoet\Cron\Workers\WooCommercePastOrders;
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class ScheduledTask {
+  private $diContainer;
+  private $entityManager;
```
costasovo, line 19: It would be nice to have doctypes for these properties.

costasovo (commented):
The previous PR has been merged but Github reports some conflicts probably because for some reason it can't recognise that first couple of commits are already merged in the main branch. So I guess the PR needs rebasing.

I've reviewed the new commits and it looks good to me. Good job!

I added just a notice about properties in the data factory in tests.

tests/DataFactories/ScheduledTask.php
rodrigoprimo: Done

tests/DataFactories/ScheduledTask.php
rodrigoprimo: Good point. I pushed a new commit changing `diContainer` to be a local variable inside the constructor.

rodrigoprimo:
Thanks for the review, @costasovo. I have implemented your suggestions and also fixed the conflicts (no idea why they were happening). Could you please check this PR again?

rodrigoprimo:
> no idea why they were happening

Ah, I think I know what happened. In my previous team, we would merge PRs into master instead of rebasing and then merging them. I forgot that at MailPoet we rebase and merge. I guess that when you rebased #3699 new hashes were created for the commits and that is why things didn't work the way I thought they would work.

costasovo (approved):

veljkho (approved):

PR #3709, opened 2021-09-30
Fix spacing between heading and paragraph in the ALC block [MAILPOET-3474]

lysyjan:
[MAILPOET-3474]

[MAILPOET-3474]: https://mailpoet.atlassian.net/browse/MAILPOET-3474

### DIFF
--- a/lib/Newsletter/Editor/PostContentManager.php
+++ b/lib/Newsletter/Editor/PostContentManager.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Newsletter\Editor;
 
+use MailPoet\Util\pQuery\pQuery;
 use MailPoet\WooCommerce\Helper as WooCommerceHelper;
 use MailPoet\WP\Functions as WPFunctions;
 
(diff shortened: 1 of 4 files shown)


lib/Newsletter/Editor/PostContentManager.php
```diff
@@ -68,7 +68,7 @@ public function filterContent($content, $displayType, $withPostClass = true) {
 
     $content = strip_tags($content, implode('', $tagsNotBeingStripped));
     if ($withPostClass) {
-      $content = str_replace('<p', '<p class="' . self::WP_POST_CLASS . '"', WPFunctions::get()->wpautop($content));
+      $content = preg_replace('/<p( class=\".+\")?/', '<p class="' . self::WP_POST_CLASS . '"', WPFunctions::get()->wpautop($content));
```
costasovo, line 71: I assume that we need to replace the class attribute here and make sure there are no 2 class attributes, correct?
I think this will not work if the class attribute will be placed other than first e.g. `<p id="anchor" class="custom-class">`.  So maybe we should remove the original class and then add the new one. What do you think?

It might be handy to have a comment here. I don't even know why we need to replace the class. So maybe if you learned why it is needed could you please add a comment so that the future us know?

I also noticed that there is [a test for the class](https://github.com/mailpoet/mailpoet/blob/b39dac75d6a729f5cff25a38e3d9b104f0711a34/tests/integration/Newsletter/Editor/PostContentManagerTest.php) so it would be nice to have this covered in the test.

lib/Newsletter/Renderer/Blocks/Text.php
```diff
@@ -120,8 +120,8 @@ public function convertParagraphsToTables($html) {
         $lineBreaks;
       if ($paragraph->hasClass(PostContentManager::WP_POST_CLASS)) {
         $paragraph->removeClass(PostContentManager::WP_POST_CLASS);
-        // if this element is followed by a paragraph, add double line breaks
-        $lineBreaks = ($nextElement && preg_match('/<p/i', $nextElement->getOuterText())) ?
+        // if this element is followed by a paragraph or heading, add double line breaks
+        $lineBreaks = ($nextElement && preg_match('/<(p|h[1-6]{1})/i', $nextElement->getOuterText())) ?
```
costasovo, line 124: Maybe this behavior could be tested within [the text block](https://github.com/mailpoet/mailpoet/blob/deb4d7833d0140a7a9381082146ec228d05897ca/tests/unit/Newsletter/Renderer/Blocks/TextTest.php) unit test.

costasovo (commented):
Good job finding the source of the issue! Could you please add a few tests for your changes?

costasovo (approved):

veljkho (approved):

PR #3718, opened 2021-10-05
Improve messages for the inactive premium plugin [MAILPOET-3719]

lysyjan:
[MAILPOET-3719]


lysyjan:
I didn't notice that @rodrigoprimo will be AFK this week. Can I ask you @brezocordero for the review?

PR #3747, opened 2021-10-21
Add limitation for scheduled_at [MAILPOET-3818]

lysyjan:
[MAILPOET-3818]

[MAILPOET-3818]: https://mailpoet.atlassian.net/browse/MAILPOET-3818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/css/src/components/_parsley.scss
+++ b/assets/css/src/components/_parsley.scss
@@ -40,3 +40,7 @@ textarea.parsley-error {
 .parsley-custom-error-message {
   color: #b94a48;
 }
+
+.mailpoet-form-errors {
+  padding: 0 16px;
+}
--- a/assets/js/src/newsletters/automatic_emails/events/event_scheduling.jsx
+++ b/assets/js/src/newsletters/automatic_emails/events/event_scheduling.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import MailPoet from 'mailpoet';
 import Selection from 'form/fields/selection.jsx';
 import Text from 'form/fields/text.jsx';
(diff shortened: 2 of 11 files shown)


veljkho (changes requested):
Everything's good except point 2 from the Specs.
First I scheduled both Welcome Email and Newsletter Email to be sent 6 years in the future, then I just uploaded the build zip from this PR and I still see that the both emails will be sent in 6 years. Is this only visual issue?

lysyjan:
The second point says:
> If there already exists an email, which should be scheduled outside of the MySQL TIMESTAMP range, use the maximum possible date to schedule it, instead of sending it immediately.

If I understand correctly, when the welcome email should be outside of the MySQL timestamp range, for example, `2040-01-01 12:12:12`, our plugin sends it immediately in the current state. So the fix should be to schedule the email at the maximal value in MySQL timestamp, which is `2038-01-19 03:14:07`.

> First I scheduled both Welcome Email and Newsletter Email to be sent 6 years in the future, then I just uploaded the build zip from this PR and I still see that the both emails will be sent in 6 years. Is this only visual issue?

I think that we don't want to reschedule user emails without notice. We can ask @NeosinneR for confirmation.

NeosinneR:
@lysyjan you are correct. I also added it to the specs as point 3. (for future reference).

> If there already exists an email, which is scheduled more than 5 years in the future, but not outside of the MySQL TIMESTAMP range, do nothing. Leave them scheduled as they were.

cc @veljkho

veljkho (approved):

PR #3750, opened 2021-10-22
Refactor saving newsletter template images [MAILPOET-2686]

costasovo:
I added comments into commit messages and code. I recommend reviewing by commits. 

### Note for QA
Before updating to this build import a couple of templates and send a couple of newsletters so that you have some template images to migrate after updating to the build.
With the new version, all newsletter templates images should be loaded from image files and the newsletter templates page should display faster.
I also altered how we take screenshots on retina screens so you can try saving a couple of newsletters as templates on a Mac (it is important that the browser window is opened on the retina screen) and check the quality of templates.

[MAILPOET-2686]

[MAILPOET-2686]: https://mailpoet.atlassian.net/browse/MAILPOET-2686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/js/src/common/thumbnail.ts
+++ b/assets/js/src/common/thumbnail.ts
@@ -11,6 +11,7 @@ import html2canvas from 'html2canvas';
 export const fromDom = async (element) => {
   const canvas = await html2canvas(element, {
     logging: false,
+    scale: 1, // Use a constant scale to prevent generating large images on Retina displays
   });
   return canvas.toDataURL('image/jpeg');
 };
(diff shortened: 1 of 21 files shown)


brezocordero (dismissed):

veljkho (changes requested):
Hi Rosta,

If I try to select any template for my newsletter, I'm getting [this error](https://d.pr/i/JwAMsN).

If I try to import a newsletter template, getting Unknown error. Could be the same reason.

costasovo:
@veljkho Looking at the error it seems that the DB migration that adds the new columns did not run. The migration runs during plugin activation or update. Is it possible that you somehow switched to this branch (or installed the build) without actually running update or activation?

veljkho (approved):

PR #3756, opened 2021-10-27
Add new re-engagement templates [MAILPOET-3764]

lysyjan:
[MAILPOET-3764]

[MAILPOET-3764]: https://mailpoet.atlassian.net/browse/MAILPOET-3764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


veljkho (changes requested):
Hi @lysyjan 
I can't see any template available, this is [what I see](https://d.pr/i/5pa7IT).
I also defined in `wp-config` file:
- define('MAILPOET_USE_CDN', true);
Still nothing present... Is this testable before production? Any helpful tips are welcomed.

lysyjan:
Hi @veljkho!
If I remember it correctly, CDN images are uploaded during the release process.
It means that previews of new templates are not visible before release.

Please, can you confirm that @NeosinneR?

NeosinneR:
@lysyjan yes, the images are uploaded to CDN during release. Optionally, it's possible to manually add them to WP CDN whenever even outside of the release, via SVN.

lysyjan:
> Optionally, it's possible to manually add them to WP CDN whenever even outside of the release, via SVN.

I didn't realize that. Thanks!

@veljkho If you want, I can do it.

veljkho:
Yes please @lysyjan , if you need access to my online site just let me know. Thanks!

lysyjan:
@veljkho Done! You can verify the preview of templates.

veljkho (changes requested):
Found only this issue below:
1. [Sort order is incorrect](https://d.pr/i/vnASYu), according to Specs it should be first _Confirm interest before deactivation_ and second one _Confirm your interest or unsubscribe

Additionally, is it expected to leave an option to activate re-engagement email by [clicking on status switcher in the listing](https://d.pr/i/t0MTsS) after I save the email without such required shortcode `[link:subscription_re_engage_url]`? Will the status switcher really activate such invalid email?

NeosinneR:
Just a quick note, the order of these two templates is not important, doesn't matter which is the first and which the second.

NeosinneR:
@veljkho 
> Additionally, is it expected to leave an option to activate re-engagement email by clicking on status switcher in the listing after I save the email without such required shortcode [link:subscription_re_engage_url]? Will the status switcher really activate such invalid email?

It shouldn't be possible to activate an email with missing `[link:subscription_re_engage_url]` no matter if from the last step or from the listing.

veljkho:
So that means that "activating email and seeing success message" for an email without such shortcode from the listing can be ignored in this situation or should we do something to prevent it from happening? @NeosinneR

lysyjan:
@veljkho @NeosinneR I forgot about the activation in the listing. I will fix it.

lysyjan:
@veljkho I added a check that an email contains the shortcode [link:subscription_re_engage_url] when a re-engagement email is activated from the listing.

costasovo (approved):

veljkho (approved):
Awesome, thanks! @lysyjan 
We should think of doing the same way for other emails that must include unsubscribe link.

PR #3771, opened 2021-11-02
Remove the method \UnitTester::generateSubscribers - [MAILPOET-3822]

triple0t:
[MAILPOET-3822](https://mailpoet.atlassian.net/browse/MAILPOET-3822)


rodrigoprimo (commented):
I can confirm that removing this method doesn't seem to have any unwanted side effects as it is not used anymore. But before moving on with this PR, I have a question about the other methods in this class. I know that the ticket only mentions `generateSubscribers()`, but I wonder if we should remove the whole class as it doesn't seem to be used anywhere, and we have a data factory to create subscribers if we need that in the integration and acceptance tests?

triple0t:
> I can confirm that removing this method doesn't seem to have any unwanted side effects as it is not used anymore. But before moving on with this PR, I have a question about the other methods in this class. I know that the ticket only mentions `generateSubscribers()`, but I wonder if we should remove the whole class as it doesn't seem to be used anywhere, and we have a data factory to create subscribers if we need that in the integration and acceptance tests?

Thanks for the feedback, Rodrigo.

I share a similar sentiment as well which is removing the Class since it appears it is not been used anywhere else.

On taking a deeper look, I noticed the following:

- I saw the file name has a generic title: **UnitTester.php** and I noticed a similarly titled file on the **mailpoet-premium** plugin.  I think it might have a special meaning and wasn’t sure if deleting it will have side effects.

- I do not know if a specific process has to be followed before we can delete a file or a Class.

Tagging in @pavel-mailpoet to ask for the way forward

pavel-mailpoet:
The class `UnitTester.php` has been in the codebase since 2015 when the first codeception setup was done. We can definitely remove all of the methods in there since we don't use them anywhere. We could keep the class itself for future use. But since we don't use it now I doubt we will start. So I would remove it. If we need we can always create it back. 

There is a similar class in the premium plugin, but I can see some usages so I would keep it in there.

triple0t:
> The class `UnitTester.php` has been in the codebase since 2015 when the first codeception setup was done. We can definitely remove all of the methods in there since we don't use them anywhere. We could keep the class itself for future use. But since we don't use it now I doubt we will start. So I would remove it. If we need we can always create it back.
> 
> There is a similar class in the premium plugin, but I can see some usages so I would keep it in there.

Noted.
Thanks for the quick response @pavel-mailpoet 
I will go ahead with deleting the file

triple0t:
Restore deleted UnitTester.php file due to lint check failing on Circleci.

We discovered that Codeception will regenerate the file during unit testing. Unfortunately, the regenerated file fails lint check which prevents the build from passing on Circleci.

I restored the file and left it empty.

rodrigoprimo (approved):
Good job!

veljkho (approved):

PR #3778, opened 2021-11-03
Introduce Newsletter block for Checkout block

mikejolley:
Introduces a new block called "MailPoet Marketing opt-in" that will be rendered below the email field in the Checkout block.

References:
- [Project Thread](https://woogrowp2.wordpress.com/2021/11/01/project-thread-integrate-automatewoo-and-mailpoet-with-checkout-block/)
- [Epic issue](woocommerce/woocommerce-gutenberg-products-block#5044)
- Note this PR does require some changes not yet merged in Blocks, [being worked on here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5045#issuecomment-958791601).

The block is force added to the Checkout Block layout underneath the contact/email field. When opt-in is disabled in MailPoet settings, you'll see a placeholder:

Screenshot 2021-11-03 at 15 38 00

When opt-in is enabled in MailPoet settings, you'll see a preview of the checkbox with the default opt-in message (from the `Checkbox opt-in message` setting):

Screenshot 2021-11-03 at 15 32 31

The message next is editable. Block placement is not. On the frontend you'll see whatever text you input, and the default status of the checkbox is based on your current subscription status, like the main checkout page:

Screenshot 2021-11-03 at 15 42 45

And that's about it. Subscription is handled via the StoreAPI when checkout is submitted.

I struggled a bit with the repo linting rules as it's not using `@wordpress/scripts` or a prettier config to format things automatically. Did I miss a step? I managed to get the linting rules passing manually but I did need to add some ignore rules inline. Let me know if anything needs changing.

## Testing Steps

**Regression testing:**

- With WooCommerce Blocks disabled, Gutenberg disabled, load this PR.
- There should be no errors whatsoever.
- In the editor, you shouldn't be able to insert "MailPoet Marketing Opt-in" block.
- Enable WooCommerce Blocks 6.2.0 or older, there should be no errors.

**Functionality testing:**

- Checkout and build trunk branch in https://github.com/woocommerce/woocommerce-gutenberg-products-block/ or [woocommerce-gutenberg-products-block.zip](https://github.com/woocommerce/automatewoo/files/7467470/woocommerce-gutenberg-products-block.zip).
- Before checking out this branch, create a new page and insert Checkout block in it, save.
- Checkout this branch and build
- With `Opt-in on checkout` off:
   - Edit the checkout block page
   - You should see the placeholder state below the email field
   - The button should link to MailPoet settings
   - Nothing should be visible on the frontend
- With `Opt-in on checkout` on:
    - View the page on the frontend 
    - You should see the marketing opt in checkbox below the email field.
    - The text should reflect what you have in settings, HTML is supported.
    - Edit the page, you should see the block in the editor.
    - Edit the text in the editor, your value should persist in frontend.
- With `Opt-in on checkout` on, test the checkout process as a customer:
    - With a guest account, opt-in and place order. Confirm it worked.  
    - With a guest account, do not opt-in and place order. Confirm the user was not subscribed.
    - Repeat with a logged in account.
    - Ensure that after subscription, the checkbox is checked by default.
    - Uncheck and place another order. Confirm user is unsubscribed.

## Changelog

> Enhancement - Add Marketing opt-in block for Checkout block.


lib/Config/Initializer.php
```diff
@@ -382,4 +389,12 @@ private function setupWoocommerceTransactionalEmails() {
       $this->wcTransactionalEmails->useTemplateForWoocommerceEmails();
     }
   }
+
+  private function setupWoocommerceBlocksIntegration() {
+    $wcEnabled = $this->wcHelper->isWooCommerceActive();
+    $wcBlocksEnabled = $this->wcHelper->isWooCommerceBlocksActive( '6.3.0-dev' );
```
mikejolley, line 395: This version may need updating when versions are finalised.

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
```diff
@@ -0,0 +1,109 @@
...
+    // See Subscription::subscribeOnCheckout
+    $subscriber = Subscriber::where('email', $order->get_billing_email())
+      ->where('is_woocommerce_user', 1)
+      ->findOne();
+
+    if (!$subscriber) {
+      return null;
+    }
+
+    $this->woocommerceSubscription->handleSubscriberOptin($subscriber, (bool)$request['extensions']['mailpoet']['optin']);
```
mikejolley, line 107: New method I added so the logic was reusable.

lib/WooCommerce/Subscription.php
```diff
@@ -116,7 +116,7 @@ private function getSubscriptionPresenceCheckField() {
     return str_replace('type="text', 'type="hidden"', $field);
   }
 
-  private function isCurrentUserSubscribed() {
+  public function isCurrentUserSubscribed() {
```
mikejolley, line 119: Needed access to this from the block initialisation class.

lib/WooCommerce/Subscription.php
```diff
@@ -148,44 +148,59 @@ public function subscribeOnCheckout($orderId, $data) {
...
+    return $this->handleSubscriberOptin( $subscriber, $checkoutOptinEnabled && $checkoutOptin );
+  }
+
+  /**
+   * Subscribe or unsubscribe a subscriber.
+   *
+   * @param Subscriber $subscriber Subscriber object
+   * @param bool $optin Opting in or opting out.
+   */
+  public function handleSubscriberOptin( Subscriber $subscriber, $optin = true ) {
```
mikejolley, line 169: Needed access to this logic from the block initialisation class so I made it more general use.

package.json
```diff
@@ -11,7 +11,7 @@
     "lint": "npm run lint6 && npm run lint-ts && npm run lint5 && npm run lint-tests",
     "lint6": "eslint -c .eslintrc.es6.json --max-warnings 0 'assets/js/src/**/*.jsx' 'tests/javascript/**/*.js'",
     "lint-ts": "eslint -c .eslintrc.ts.json --max-warnings 0 'assets/js/src/**/*.tsx' 'assets/js/src/**/*.ts'",
-    "lint5": "eslint -c .eslintrc.es5.json --max-warnings 0 'assets/js/src/**/*.js' 'webpack.config.js'",
+    "lint5": "eslint -c .eslintrc.es5.json --max-warnings 0 'assets/js/src/**/*.js'",
```
mikejolley, line 14: Feel free to modify this, but I needed the spread operator in the webpack config to support our build. These linting rules were too aggressive.

pavel-mailpoet:
@mikejolley your branch is quite old behind the current master, could you please rebase it and resolve the conflicts?

mikejolley:
Ok no problem, I've updated, it was just the lock file conflicting. I only started work 2 days ago on this so the branch should not be old.

assets/js/src/marketing_optin_block/block.tsx
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable react/react-in-jsx-scope */
+/* eslint-disable import/no-unresolved */
```
pavel-mailpoet, line 2: We really don't like `eslint-disable` statements. You can check our [contributing documentation](https://github.com/mailpoet/mailpoet/blob/master/CONTRIBUTING.md#disabling-linting-rules). Is this really necessary? We can see the WooCommerce modules are defined in weback. We use a similar approach in our premium plugin but we don't need to disable Eslint import check.

assets/js/src/marketing_optin_block/block.tsx
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable react/react-in-jsx-scope */
```
pavel-mailpoet, line 1: It would be nice to get rid of this statement, can we instead import react?

assets/js/src/marketing_optin_block/attributes.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * External dependencies
+ */
+import { getSetting } from '@woocommerce/settings';
+
+const { defaultText } = getSetting('mailpoet_data');
```
pavel-mailpoet, line 6: We try to refactor all `.js` files to typescript. All new files we create are in typescript. Could you have this file as `.ts` as well?

assets/js/src/marketing_optin_block/block.json
```diff
@@ -0,0 +1,27 @@
...
+      "type": "object",
+      "default": {
+        "remove": true,
+        "move": false
+      }
+    }
+  },
+  "parent": ["woocommerce/checkout-contact-information-block"],
+  "editorScript": "file:../../../dist/js/marketing_optin_block/marketing-optin-block.js",
+  "editorStyle": "file:../../../dist/js/marketing_optin_block/marketing-optin-block.css"
```
pavel-mailpoet, line 26: Could you please elaborate on how this works? Where would this file go? All of our built files are in the `dist` folder, would this path still be valid?

Also, for all of our built assets, we use a Webpack plugin that adds a random string into the file name. This prevents a caching problems when releasing a new version. Is this a concern here as well?

assets/js/src/marketing_optin_block/block.tsx
```diff
@@ -0,0 +1,50 @@
...
+/* eslint-disable import/no-unresolved */
+
+/**
+ * External dependencies
+ */
+import { CheckboxControl } from '@woocommerce/blocks-checkout';
+import { useState, useEffect } from '@wordpress/element';
```
pavel-mailpoet, line 8: I am not sure why are you using `useState` and `useEffect` from `'@wordpress/element'`. In other places, we use the React functions. What is the benefit?

assets/js/src/marketing_optin_block/block.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { getSetting } from '@woocommerce/settings';
+
+const { optinEnabled, defaultText, defaultStatus } = getSetting('mailpoet_data');
```
pavel-mailpoet, line 11: I am not sure what would happen if this file gets loaded but the WooCommerce plugin is not active? Would this still work?

assets/js/src/marketing_optin_block/block.tsx
```diff
@@ -0,0 +1,50 @@
...
+  if (!optinEnabled) {
+    return null;
+  }
+
+  return (
+    <CheckboxControl
+      checked={checked}
+      onChange={setChecked}
+    >
+      {/* eslint-disable-next-line react/no-danger */}
```
pavel-mailpoet, line 41: Could you please add a comment on why is this necessary?

assets/js/src/marketing_optin_block/edit.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const EmptyState = (): JSX.Element => (
+  <Placeholder
+    icon={<Icon icon={megaphone} />}
+    label={__('Marketing opt-in', 'mailpoet')}
+    className="wp-block-mailpoet-newsletter-block-placeholder"
+  >
+    <span className="wp-block-mailpoet-newsletter-block-placeholder__description">
+      {__(
+        'MailPoet marketing opt-in would be shown here if enabled. You can enable from the settings page.',
```
pavel-mailpoet, line 30: I am afraid this is not going to work. We don't use the standard WordPress translation function. You can check other places where is used `MailPoet.I18n.t`.

assets/js/src/marketing_optin_block/frontend.tsx
```diff
@@ -0,0 +1,17 @@
+/**
+ * External dependencies
+ */
+/* eslint-disable react/react-in-jsx-scope */
```
pavel-mailpoet, line 4: I don't think this file uses react, does it? Is this redundant? And the file itself can have a `.ts` extension.

lib/Config/Initializer.php
pavel-mailpoet: We don't write spaces in the parents, could you please remove them so that the code looks like this? 
```
$wcBlocksEnabled = $this->wcHelper->isWooCommerceBlocksActive('6.3.0-dev');
```

lib/PostEditorBlocks/MarketingOptinBlock.php
```diff
@@ -0,0 +1,83 @@
...
+    $this->options = $options;
+  }
+
+  /**
+   * The name of the integration.
+   *
+   * @return string
+   */
+  public function get_name() { // phpcs:ignore
+    return 'mailpoet';
```
pavel-mailpoet, line 29: Can you please explain where is this used? Should it be capitalised maybe? `MailPoet`

lib/PostEditorBlocks/MarketingOptinBlock.php
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * Register block scripts and assets.
+   */
+  public function initialize() {
+    $script_asset_path = Env::$assetsUrl . '/dist/js/marketing_optin_block/marketing-optin-block-frontend.asset.php';
+    $script_asset      = file_exists($script_asset_path)
```
pavel-mailpoet, line 37: In our coding standards we only have a single space on each side of `=`, could you please remove the extra spaces?

lib/PostEditorBlocks/MarketingOptinBlock.php
```diff
@@ -0,0 +1,83 @@
...
+   */
+  public function get_name() { // phpcs:ignore
+    return 'mailpoet';
+  }
+
+  /**
+   * Register block scripts and assets.
+   */
+  public function initialize() {
+    $script_asset_path = Env::$assetsUrl . '/dist/js/marketing_optin_block/marketing-optin-block-frontend.asset.php';
```
pavel-mailpoet, line 36: I am not sure where comes this file from? How is the file `marketing-optin-block-frontend.asset.php` created?

lib/PostEditorBlocks/MarketingOptinBlock.php
```diff
@@ -0,0 +1,83 @@
...
+    $script_asset      = file_exists($script_asset_path)
+      ? require $script_asset_path
+      : [
+        'dependencies' => [],
+        'version'      => Env::$version,
```
pavel-mailpoet, line 41: Same comment here, could you please remove the extra spaces?

lib/PostEditorBlocks/MarketingOptinBlock.php
```diff
@@ -0,0 +1,83 @@
+<?php
```
pavel-mailpoet, line 1: In the new files we add the statement `declare(strict_types=1);` Could you please add it here as well? And add types to all methods in this class?

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
```diff
@@ -0,0 +1,109 @@
+<?php
```
pavel-mailpoet, line 1: Similar comment here, could you please add the `declare(strict_types=1);` statement?

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
```diff
@@ -0,0 +1,109 @@
...
+use Automattic\WooCommerce\Blocks\Domain\Services\ExtendRestApi;
+use Automattic\WooCommerce\Blocks\Package;
+use Automattic\WooCommerce\Blocks\StoreApi\Schemas\CheckoutSchema;
+use MailPoet\Config\Env;
+use MailPoet\Models\Subscriber;
+use MailPoet\Settings\SettingsController;
+use MailPoet\WooCommerce\Subscription as WooCommerceSubscription;
+use MailPoet\WP\Functions as WPFunctions;
+
+// @todo Disable if blocks is inactive
```
pavel-mailpoet, line 14: Is there a task to finish this `@todo`? I don't want this to be here forever. Could this not being implemented cause any issues?

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
```diff
@@ -0,0 +1,109 @@
...
+    $this->woocommerceSubscription = $woocommerceSubscription;
+  }
+
+  public function init() {
+    $this->wp->addAction(
+      'woocommerce_blocks_checkout_block_registration',
+      [$this, 'registerCheckoutFrontendBlocks']
+    );
+    $this->wp->addAction(
+      '__experimental_woocommerce_blocks_checkout_update_order_from_request',
```
pavel-mailpoet, line 38: I can see this experimental Hook. What happens when the hook will be moved from the experimental status? Will you create a new pull request? Would it work for all WooCommerce versions?

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
```diff
@@ -0,0 +1,109 @@
...
+    if (!$order || !$order->get_billing_email()) {
+      return;
+    }
+
+    if (!isset($request['extensions']['mailpoet']['optin'])) {
+      return;
+    }
+
+    // See Subscription::subscribeOnCheckout
+    $subscriber = Subscriber::where('email', $order->get_billing_email())
```
pavel-mailpoet, line 99: Please don't use the old models. We are currently migrating from the old IdiORM to Doctrine. So please avoid the namespace `MailPoet\Models`. Instead use a subscribers repository:
```
$this->subscribersRepository->findOneBy(['email' => $order->get_billing_email(), 'isWoocommerceUser' => true]);
```
The repository is `use MailPoet\Subscribers\SubscribersRepository;`

lib/WooCommerce/Subscription.php
```diff
@@ -148,44 +148,59 @@ public function subscribeOnCheckout($orderId, $data) {
...
       ->where('is_woocommerce_user', 1)
       ->findOne();
+
     if (!$subscriber) {
       // no subscriber: WooCommerce sync didn't work
       return null;
     }
 
-    $checkoutOptinEnabled = (bool)$this->settings->get(self::OPTIN_ENABLED_SETTING_NAME);
+    $checkoutOptinEnabled = (bool)$this->settings->get( self::OPTIN_ENABLED_SETTING_NAME );
```
pavel-mailpoet, line 157: Could you please remove the extra spaces? It is multiple times in this file. I thought our PHP coding standard checker is configured better.

package.json
```diff
@@ -130,6 +130,8 @@
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.33.0",
     "@wordpress/browserslist-config": "^4.0.0",
+    "@wordpress/dependency-extraction-webpack-plugin": "^3.2.1",
```
pavel-mailpoet, line 133: Could you please explain why did you need this dependency?

package.json
```diff
@@ -130,6 +130,8 @@
...
+    "@wordpress/scripts": "^19.0.0",
```
pavel-mailpoet, line 134: Could you please why did you need to add this dependency?

pavel-mailpoet (changes requested):
Hi, 
we have a couple of questions and comments. In addition to those:

1. There are so many changes in the `package-lock.json` file. We are constantly fighting with the size of the plugin so I am afraid this could increase the plugin size. Also, I am afraid that the changes themselves could have unexpected side effects. Is it possible to avoid making so many changes to this file?
2. You completely changed the format of the `webpack.config.js` That is quite unfortunate. We try to use single quotes everywhere but you changed it to double quotes for some reason. Could you please revert that? It makes it hard to spot what is actually changed in this file.
3. If you run `./do qa` there are many errors in typescript, before we can merge this pull request those need to be addressed
4. There are no tests in this pull request. Could you please add some? There should be at least an Acceptance Test, you can maybe talk to Veljko and ask him to write the test for you later.

package.json
pavel-mailpoet: Sorry, for historical reasons we use some old Eslint rules. You could `Object.assign` instead of `...`, couldn't you?

assets/js/src/marketing_optin_block/block.tsx
mikejolley: I had to add this because of the excessive linting rules on commit this repository has implemented. They prohibit the use of dangerouslySet, and in this case, it's needed to support HTML formatting.

assets/js/src/marketing_optin_block/frontend.tsx
mikejolley: This is another case where your repository linting rules throw errors if react is not imported. We don't need react import here.

lib/PostEditorBlocks/MarketingOptinBlock.php
mikejolley: Do you have a phpcs/phpcbf config available so formatting is not a manual process?

assets/js/src/marketing_optin_block/block.tsx
mikejolley: The integration class is not initialized without WordPress and Blocks active, so this script would not be loaded.

assets/js/src/marketing_optin_block/block.json
senadir: this is file is consumed by WordPress native function `register_block_type`, it would use this path to load the correct files.
You can see that the file reference `dist`.

> Also, for all of our built assets, we use a Webpack plugin that adds a random string into the file name. This prevents a caching problems when releasing a new version. Is this a concern here as well?

We were not aware of that. It would be lovely to implement yet we're a bit constrained to actually do this. The above files have a `?ver=version` attached to them, as in the standard in WordPress, this should work just fine. Hashing in file names is superior but would the above block a bit complicated, I believe this can be a follow up?

lib/PostEditorBlocks/MarketingOptinBlock.php
mikejolley: This is generated on build by `@wordpress/dependency-extraction-webpack-plugin`. It creates a list of script dependencies that the block requires, which are external scripts loaded by WordPress and WooCommerce Blocks.

assets/js/src/marketing_optin_block/block.tsx
senadir: We follow the WordPress/Gutenberg standard of using core packages instead of directly referencing React, `@wordpress/element` is a light superset of React which allows the Gutenberg team to control which version of React is served (to avoid version conflict), offers extra functions, and allows you down the line to replace the react implementation if needed (like Jetpack using Preact).

I understand that your team uses React in your custom screens, but given this code is going to be served and consumed in Gutenberg, I feel it might less dangerous to just use `@wordpress/element` to avoid any version conflict.

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
mikejolley: If the __experimental hook was graduated to non-experimental in the Blocks plugin, deprecation would be added and this would need changing. I won't be creating new pull requests for this integration. It is my understanding that this work will be handed off to the MailPoet team once complete.

assets/js/src/marketing_optin_block/block.tsx
senadir: This code won't ever be reached if WooCommerce Blocks version 6.3.0 or up isn't enabled, so it's safe that way.
The block has a tight dependency on WooCommerce Checkout so it can't be used outside it if that's your question.

package.json
mikejolley: We're importing the WordPress webpack config from this package. We use this in blocks and other integrations, so I borrowed it for this.

package.json
mikejolley: See https://github.com/mailpoet/mailpoet/pull/3778#discussion_r742827507

lib/PostEditorBlocks/MarketingOptinBlock.php
senadir: `get_name` is used in several places, and is enqueued in the frontend as well, so the standard is to use snake_case.

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
senadir: Just to confirm that once we graduate this, `__experimental_woocommerce_blocks_checkout_update_order_from_request` would still function just as fine and the deprecation warning would outline any changes you might need to do.
We consider that hook fairly stable so far, so I doubt we will need to change anything.

lib/PostEditorBlocks/MarketingOptinBlock.php
mikejolley: This is the integration name, part of `IntegrationInterface`, and needs to be unique per extension. @senadir does case matter?

assets/js/src/marketing_optin_block/block.tsx
mikejolley: This would cause `@wordpress/dependency-extraction-webpack-plugin` to include `react` as a dependency when registering/enqueueing the block script. It's not neccessary.

lib/PostEditorBlocks/MarketingOptinBlock.php
senadir: name is used an array key and is exposed with settings data as `$name_data`. There is no requirement for it to be snake_case, but it's standard usage across plugins.

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
senadir: Thank you! is there a document I can check to know what to use? we were going by existing code tbh so I can't know what's old and what's new.

senadir:
> There are so many changes in the package-lock.json file. We are constantly fighting with the size of the plugin so I am afraid this could increase the plugin size. Also, I am afraid that the changes themselves could have unexpected side effects. Is it possible to avoid making so many changes to this file?

I understand that managing dependencies is an extremely hard task. The only dependencies we added are dev dependencies, which are not going to be included in the frontend or the final bundle size.
The number of changes in the package-lock.json is puzzling but is probably expected? we installed two new dependencies, using the node/npm version specified in nvm file, so it should follow what you have if that file is kept updated regularly (via a tool like renovate).

> You completely changed the format of the webpack.config.js That is quite unfortunate. We try to use single quotes everywhere but you changed it to double quotes for some reason. Could you please revert that? It makes it hard to spot what is actually changed in this file.

That's fine, do you have any sort of autofixer with eslint/prettier that can fix that? my editor isn't picking up anything in that file.

> If you run ./do qa there are many errors in typescript, before we can merge this pull request those need to be addressed

It seems a lot of errors are coming from node_modules, is there a way to limit the view to just src files?

> There are no tests in this pull request. Could you please add some? There should be at least an Acceptance Test, you can maybe talk to Veljko and ask him to write the test for you later.

I'm afraid that's outside the scope of our work as we don't have much knowledge of your codebase or how tests are written here and we have other extensions to do.

I think we provided the value of writing this integration ourselves given our knowledge with WooCommece Checkout/Gutenberg and the integration ourselves, which helps speed things up. We're happy to move forward with more questions about the integration, but I'm afraid we can't prioritize other things like writing tests.

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
mikejolley: @senadir looks like it was graduated already so let's swap.

assets/js/src/marketing_optin_block/block.tsx
costasovo: We added this rule to raise extra attention when using `dangerouslySetInnerHTML`. This may lead to some unexpected security issues. When we need to use it we should provide additional explanations as a code comment for our future selves.

If understand correctly this span is used as a label for the opt-in checkbox and we need to render HTML, so that a user can insert some links etc. Is that correct? Is that safe in terms of XSS? In our [current implementation for old-school checkout](https://github.com/mailpoet/mailpoet/blob/683c5fa01f0fbefeb14d6d8e4fd257303f6a5ffb/lib/WooCommerce/Subscription.php#L84), we use esc_html and we don't allow custom HTML. I understand that it is useful to support links etc. Maybe we could sanitize it by allowing some subset of tags. What about using something like [wp_kses ](https://developer.wordpress.org/reference/functions/wp_kses/)?

assets/js/src/marketing_optin_block/frontend.tsx
costasovo: I believe that in this particular case renaming the file to `frontend.ts` might be enough to get rid of the linter warning.

lib/PostEditorBlocks/MarketingOptinBlock.php
costasovo: Thanks for explaining!

lib/PostEditorBlocks/MarketingOptinBlock.php
costasovo: Thanks for explaining!

lib/PostEditorBlocks/MarketingOptinBlock.php
costasovo: Yes, we have (./do qa command uses it), but it seems this is not covered by our ruleset even though we thought it was :). Let's leave it as is. We will revisit it later.

lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
costasovo: We don't have a document about this. It is long-lasting effort we started about 2 years ago. We gradually replace the old ORM with Doctrine. We should add a comment about this to some docs for contributors. Thanks for mentioning it. We really want to avoid adding new usage of those old models, because it takes us effort to replace them later.

package.json
costasovo: Thanks for explaining!

costasovo (commented):
Thanks for the explanations and changes! Just so that you know I've been working on this review with @pavel-mailpoet so the replies are kind of from both of us.

> > If you run ./do qa there are many errors in typescript, before we can merge this pull request those need to be addressed
> 
> It seems a lot of errors are coming from node_modules, is there a way to limit the view to just src files?

I think this might be related to changes in package-lock.json. I will take a look into this later and let you know.

> > There are no tests in this pull request. Could you please add some? There should be at least an Acceptance Test, you can maybe talk to Veljko and ask him to write the test for you later.
> 
> I'm afraid that's outside the scope of our work as we don't have much knowledge of your codebase or how tests are written here and we have other extensions to do.

We understand. I guess we could add tests ourselves in a subsequent task.

> I think we provided the value of writing this integration ourselves given our knowledge with WooCommece Checkout/Gutenberg and the integration ourselves, which helps speed things up. We're happy to move forward with more questions about the integration, but I'm afraid we can't prioritize other things like writing tests.

Thank you for taking care of this integration! We understand that writing tests etc. is out of scope for you. I've just come across [a P2 post related to this PR](https://mailpoetp2.wordpress.com/2021/11/04/integrating-mailpoet-with-the-checkout-block/). So we will need to decide what are the next steps are.

assets/js/src/marketing_optin_block/block.tsx
mikejolley: This content is stored in the post (block attribute) so I believe kses is applied. Adding a note.

> What about using something like wp_kses

This isn't rendered on the PHP side.

mikejolley:
I've updated the PR and retested the functionality. I believe the remaining items are best left for your team, that is:

- i18n if you have no plans to use the WordPress package
- strict types (I based my class on existing classes without strict types)
- If a user is not synced e.g. a guest (logged out), subscription does not work because the user is not a woocommerce user. `MailPoet\WooCommerce\Subscription` does not handle this case either so I wasn't able to find an example to handle this. I added a `@todo` comment for this.

I will remove my assignment and post on p2 as well.

costasovo:
@mikejolley I've created [a new PR](https://github.com/mailpoet/mailpoet/pull/3813) and I'm closing this in favor of the new one.

In the new PR, I've rebased commits from this PR on top of the most recent changes in the plugin. I've also added a bunch of additional commits addressing issues mentioned in [the comment above](https://github.com/mailpoet/mailpoet/pull/3778#issuecomment-968827269) and also some additional fixes of issues discovered on CI or during testing.

Thanks again for preparing this integration!

PR #3783, opened 2021-11-04
Add info when form is inactive to subscription form block [MAILPOET-3896]

lysyjan:
[MAILPOET-3896]

[MAILPOET-3896]: https://mailpoet.atlassian.net/browse/MAILPOET-3896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/js/src/post_editor_block/subscription_form/edit.jsx
+++ b/assets/js/src/post_editor_block/subscription_form/edit.jsx
@@ -26,7 +26,7 @@ function Edit({ attributes, setAttributes }) {
         <option value="" disabled selected>{window.locale.selectForm}</option>
         {allForms.map((form) => (
           <option value={form.id}>
-            {form.name}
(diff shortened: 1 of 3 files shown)


costasovo (dismissed):

veljkho (changes requested):
Hey Jan,
I see that you've fixed issue but for MailPoet Subscription Form block and not MailPoet 3 Form block that I was referring. Is it fixable also for [MailPoet 3 Form block](https://d.pr/i/PEcRT8)? Is it part of our development? Otherwise, this is fixed for MailPoet Subscription Form block.

costasovo (approved):

veljkho (approved):

PR #3784, opened 2021-11-04
Remove remaining usages of NewsletterLink and deprecate class [MAILPOET-3816]

rodrigoprimo:
This is a draft PR to get feedback on an issue before I dive deep into it. In this PR, I replaced the remaining places where NewsletterLink was still used with NewsletterLinkEntity and deprecated the old model class.

The problem is that the integration test `testItLogsErrorWhenNewlyRenderedNewsletterBodyIsInvalid` is failing due to a change in behavior (https://app.circleci.com/pipelines/github/mailpoet/mailpoet/7812/workflows/2eb5dc19-9746-4e5b-80eb-fa21f618a598/jobs/148114/tests#failed-test-0). This test, sets an invalid `newsletter_rendered_body` to make sure that the code in \MailPoet\Cron\Workers\SendingQueue\Tasks\Newsletter::preProcessNewsletter() works correctly to detect an invalid body and stop processing the newsletter ([lib/Cron/Workers/SendingQueue/Tasks/Newsletter.php#L187](https://github.com/mailpoet/mailpoet/blob/c9636ebf1b57d53ffd1f94f4224226dcb9aadf26/lib/Cron/Workers/SendingQueue/Tasks/Newsletter.php#L187)).

But after replacing NewsletterLink with NewsletterLinkEntity in https://github.com/mailpoet/mailpoet/blame/090974dc6955e54ce4895edd4104181359c59c08/lib/Newsletter/Links/Links.php#L173, it is necessary to get an instance of the SendingQueueEntity instead of just passing the ID of the sending queue. The test now fails when calling `$this->sendingQueueRepository->findOneById()` as Doctrine returns an exception since the body of the newsletter is invalid.

@costasovo, before I dig into the code to understand the best way to approach this problem, I decided to open this PR as a draft to check with you if you have seen similar problems in the past and if you have any suggestions on how to approach it?

Thanks

[MAILPOET-3816]

[MAILPOET-3816]: https://mailpoet.atlassian.net/browse/MAILPOET-3816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/Config/Populator.php
+++ b/lib/Config/Populator.php
@@ -13,6 +13,7 @@ use MailPoet\Cron\Workers\SubscribersLastEngagement;
 use MailPoet\Cron\Workers\UnsubscribeTokens;
 use MailPoet\Entities\FormEntity;
 use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Entities\NewsletterLinkEntity;
 use MailPoet\Entities\NewsletterTemplateEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SendingQueueEntity;
@@ -700,8 +701,8 @@ class Populator {
     $wpdb->query(sprintf(
       $query,
       NewsletterLink::$_table,
-      NewsletterLink::INSTANT_UNSUBSCRIBE_LINK_SHORT_CODE,
-      NewsletterLink::UNSUBSCRIBE_LINK_SHORT_CODE
+      NewsletterLinkEntity::INSTANT_UNSUBSCRIBE_LINK_SHORT_CODE,
+      NewsletterLinkEntity::UNSUBSCRIBE_LINK_SHORT_CODE
     ));
   }
 
--- a/lib/Config/Populator.php
+++ b/lib/Config/Populator.php
@@ -22,7 +22,6 @@ use MailPoet\Entities\UserFlagEntity;
 use MailPoet\Form\FormsRepository;
 use MailPoet\Mailer\MailerLog;
 use MailPoet\Models\Newsletter;
-use MailPoet\Models\NewsletterLink;
 use MailPoet\Models\ScheduledTask;
 use MailPoet\Models\Segment;
 use MailPoet\Models\SendingQueue;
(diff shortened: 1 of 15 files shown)


costasovo (commented):
I see this required refactoring a bunch of tests. Great job! 👍 

As for your question about the failing test, if I understand correctly the SendingQueueEntity is needed [here](https://github.com/mailpoet/mailpoet/blame/090974dc6955e54ce4895edd4104181359c59c08/lib/Newsletter/Links/Links.php#L173) just to create the new NewsletterLinkEntity. I think that Doctrine has a nice feature that could be handy for this case. You can [ask entity manager for a reference of some entity](https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/advanced-configuration.html#reference-proxies). It will return a proxy and you can then use the proxy instead of the fully-loaded entity. I tried it and it fixed the test for me.

Actually, we could add `getReference` method to our [abstract repository class](https://github.com/mailpoet/mailpoet/blob/73a9bed483d62b98a6dad93e7901adc1f4746d33/lib/Doctrine/Repository.php). Something like this:
```php
public function getReference($id) {
    return $this->entityManager->getReference($this->getEntityClassName(), $id);
  }
```
and then you could just replace [findOneById](https://github.com/mailpoet/mailpoet/blame/090974dc6955e54ce4895edd4104181359c59c08/lib/Newsletter/Links/Links.php#L173) with `getReference` directly on the repository [here](https://github.com/mailpoet/mailpoet/blame/090974dc6955e54ce4895edd4104181359c59c08/lib/Newsletter/Links/Links.php#L173). 

What do you think about such an approach?

rodrigoprimo:
Thanks, @costasovo. I wasn't aware that it is possible to get just a reference to an entity without getting the whole thing. I appreciate your help in sharing your knowledge about Doctrine. I implemented exactly what you suggest and it seems to work well. This PR is now ready for a review.

costasovo (approved):

veljkho (approved):

veljkho (approved):

PR #3788, opened 2021-11-08
Refactor sending queue API to doctrine [MAILPOET-3876]

pavel-mailpoet:
[MAILPOET-3876]

[MAILPOET-3876]: https://mailpoet.atlassian.net/browse/MAILPOET-3876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

For QA @veljkho: This is changing the part of the plugin that starts the sending. So try to pause a newsletter that is being sent or resume it. Try to schedule a newsletter or send an email from the plugin.

### DIFF
--- a/lib/API/JSON/v1/SendingQueue.php
+++ b/lib/API/JSON/v1/SendingQueue.php
@@ -187,7 +187,7 @@ class SendingQueue extends APIEndpoint {
     $newsletter = $this->newsletterRepository->findOneById($newsletterId);
 
     if ($newsletter instanceof NewsletterEntity) {
-      $queue = $newsletter->getLatestQueue();
+      $queue = $newsletter->getLastUpdatedQueue();
 
       if (!$queue instanceof SendingQueueEntity) {
         return $this->errorResponse([
@@ -217,7 +217,7 @@ class SendingQueue extends APIEndpoint {
     $newsletter = $this->newsletterRepository->findOneById($newsletterId);
 
     if ($newsletter instanceof NewsletterEntity) {
-      $queue = $newsletter->getLatestQueue();
+      $queue = $newsletter->getLastUpdatedQueue();
 
       if (!$queue instanceof SendingQueueEntity) {
         return $this->errorResponse([
--- a/lib/API/JSON/v1/SendingQueue.php
+++ b/lib/API/JSON/v1/SendingQueue.php
@@ -120,14 +120,13 @@ class SendingQueue extends APIEndpoint {
     }
 
     WordPress::resetRunInterval();
-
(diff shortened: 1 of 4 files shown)


lib/API/JSON/v1/SendingQueue.php
```diff
@@ -177,20 +184,18 @@ public function pause($data = []) {
       ? (int)$data['newsletter_id']
       : false
     );
-    $newsletter = Newsletter::findOne($newsletterId);
+    $newsletter = $this->newsletterRepository->findOneById($newsletterId);
 
-    if ($newsletter instanceof Newsletter) {
-      $queue = $newsletter->getQueue();
+    if ($newsletter instanceof NewsletterEntity) {
+      $queue = $newsletter->getLatestQueue();
```
costasovo, line 190: I noticed that with this change we also change the logic of how is the queue fetched. The old model was using `updated_at` column ([#](https://github.com/mailpoet/mailpoet/blob/21839fc30b3fcdc57c9140998b33c325ecd7d33d/lib/Tasks/Sending.php#L141)), but `getLatestQueue` uses [order of ids](https://github.com/mailpoet/mailpoet/blob/3c5a368c2c6205a5bdaed1c4e023f7fa9147aa49/lib/Entities/NewsletterEntity.php#L461-L464). This should have probably very similar results, but maybe we could switch `getLatestQueue` to use the same logic or maybe better introduce a new getter `getLastUpdatedQueue()` so that we don't introduce some unexpected behaviour.

lib/API/JSON/v1/SendingQueue.php
```diff
@@ -209,19 +214,18 @@ public function resume($data = []) {
       ? (int)$data['newsletter_id']
       : false
     );
-    $newsletter = Newsletter::findOne($newsletterId);
-    if ($newsletter instanceof Newsletter) {
-      $queue = $newsletter->getQueue();
+    $newsletter = $this->newsletterRepository->findOneById($newsletterId);
+
+    if ($newsletter instanceof NewsletterEntity) {
+      $queue = $newsletter->getLatestQueue();
```
costasovo, line 220: Same as above. Maybe we should use `getLastUpdatedQueue()`

costasovo (changes requested):
Good job! I added a few comments.

costasovo (approved):
Looks good!

There are a few old models usages left, but they will be handled in subsequent tasks.

veljkho (approved):

PR #3793, opened 2021-11-09
TimestampListener should not use the same time object for different entities [MAILPOET-3870]

rodrigoprimo:
Besides changing TimestampListener to create copies of the Carbon object before passing them to the entities, in a separate, commit I also removed a few unused imports that existed in this class.

[MAILPOET-3870]

[MAILPOET-3870]: https://mailpoet.atlassian.net/browse/MAILPOET-3870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/Doctrine/EventListeners/TimestampListener.php
+++ b/lib/Doctrine/EventListeners/TimestampListener.php
@@ -27,11 +27,11 @@ class TimestampListener {
       && method_exists($entity, 'getCreatedAt')
       && !$entity->getCreatedAt()
     ) {
-      $entity->setCreatedAt($this->now);
+      $entity->setCreatedAt($this->now->copy());
     }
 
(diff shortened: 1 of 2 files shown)


lysyjan (commented):
It looks good to me.

I have noticed that we have a similar issue in LastSubscribedAtListener. Please, can you fix it too?

rodrigoprimo:
Good find, @lysyjan. I had missed that we had other listeners that suffer from the same problem. I created a new Jira ticket (https://mailpoet.atlassian.net/browse/MAILPOET-3946) and I will address this in a separate PR that I will assign to you.

lysyjan (approved):

veljkho (approved):

PR #3796, opened 2021-11-10
Attempt to convert dates using d/m/Y on imports [MAILPOET-3747]

brezocordero:
With this PR, when a site is using the `d/m/Y` format, we will parse `created_at` and `confirmed_at` using `datetime` and `d/m/Y` formats, then select the one with fewer errors.
When no errors parsing with both formats (or same amount of errors) we give preference to `datetime` format, for backwards compatibility.
We assume this will not be accurate in some cases, where dates could be parsed with both formats. For example 01/02/2021 could be first of February with d/m/Y or second of January with datetime. 

To test:
- Set the date on your site to d/m/Y format
- Create a subscribers import file using that format and dates that would be wrong using datetime. For example 31/12/2021
- Import the file, don't forget to select at least one field with dates to be imported as confirmation time (change the dropdown value from `ignore` to confirmation time)
- Check the records are imported properly in the subscribers database table and the dates are correct.

[MAILPOET-3747]

[MAILPOET-3747]: https://mailpoet.atlassian.net/browse/MAILPOET-3747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/Subscribers/ImportExport/Import/Import.php
+++ b/lib/Subscribers/ImportExport/Import/Import.php
@@ -237,9 +237,14 @@ class Import {
   }
 
   private function validateDateTime(array $data, array &$invalidRecords): array {
+    $siteUsesCustomFormat = WPFunctions::get()->getOption('date_format') === 'd/m/Y';
+    if ($siteUsesCustomFormat) {
+      return $this->validateDateTimeAttemptCustomFormat($data, $invalidRecords);
+    }
+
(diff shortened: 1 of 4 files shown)


costasovo (approved):
Good job!

veljkho (approved):

PR #3798, opened 2021-11-11
Fixed CSS is not applied when using form iframe - [MAILPOET-3871]

triple0t:
#### Background
See  [MAILPOET-3871](https://mailpoet.atlassian.net/browse/MAILPOET-3871)  for more information

#### Changes proposed in this Pull Request
 * Added default styles
 * Added support for custom fonts
 * Fixed iframe styling errors

**Before:**
Screenshot 2021-11-11 at 8 29 52 AM

Screenshot 2021-11-11 at 8 30 02 AM

**After:**

Screenshot 2021-11-11 at 8 28 26 AM

Screenshot 2021-11-11 at 8 28 58 AM

#### Testing instructions

* Create a form using any of the available form templates
* Save the form and copy iframe code
* Place the iframe to some other site. (Note you might need to update with or height attributes)
* Observe the fonts and other styles 

[MAILPOET-3871](https://mailpoet.atlassian.net/browse/MAILPOET-3871)

### DIFF
--- a/assets/css/src/components-public/_resets.scss
+++ b/assets/css/src/components-public/_resets.scss
@@ -1,0 +1,82 @@
+// basic default styling
+// to allow for consistent styling between mailpoet plugin and user's theme
+
+html {
+  box-sizing: border-box;
+}
+
+*,
+*:before,
+*:after {
+  box-sizing: inherit;
+}
+
+body {
+  color: rgb(40, 48, 61);
+  font-size: 16px;
+  font-weight: normal;
+  text-align: left;
+}
+
+html,
+body,
+p,
+ol,
+ul,
+li,
+dl,
+dt,
+dd,
+blockquote,
+figure,
+fieldset,
+form,
+legend,
+textarea,
+pre,
+iframe,
+hr,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: antialiased;
+}
+
+h1,
+.h1,
+h2,
+.h2,
+h3,
+.h3,
+h4,
+.h4,
+h5,
+.h5,
+h6,
+.h6 {
+  font-weight: normal;
+  letter-spacing: normal;
+}
+
+ul,
+ol {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+label {
+  font-family: Arial, Helvetica, sans-serif;
+  margin-bottom: 10px;
+}
+
+a {
+  background-color: transparent;
+  color: rgb(40, 48, 61);
+  cursor: pointer;
+  text-decoration-skip-ink: all;
+  text-decoration-thickness: 1px;
+  text-underline-offset: 3px;
+}
--- a/assets/css/src/components-public/_resets.scss
+++ b/assets/css/src/components-public/_resets.scss
@@ -11,72 +11,94 @@ html {
   box-sizing: inherit;
 }
 
-body {
-  color: rgb(40, 48, 61);
-  font-size: 16px;
-  font-weight: normal;
-  text-align: left;
-}
+.mailpoet_form {
+  body {
+    color: rgb(40, 48, 61);
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 16px;
+    font-weight: normal;
+    text-align: left;
+  }
 
-html,
-body,
-p,
-ol,
-ul,
-li,
-dl,
-dt,
-dd,
-blockquote,
-figure,
-fieldset,
-form,
-legend,
-textarea,
-pre,
-iframe,
-hr,
-h1,
-h2,
-h3,
-h4,
-h5,
-h6 {
-  -moz-osx-font-smoothing: grayscale;
-  -webkit-font-smoothing: antialiased;
-}
+  html,
+  body,
+  p,
+  ol,
+  ul,
+  li,
+  dl,
+  dt,
+  dd,
+  blockquote,
+  figure,
+  fieldset,
+  form,
+  legend,
+  textarea,
+  pre,
+  iframe,
+  hr,
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    color: rgb(40, 48, 61);
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 16px;
+    -moz-osx-font-smoothing: grayscale;
+    -webkit-font-smoothing: antialiased;
+    font-synthesis: none;
+    font-weight: normal;
+    letter-spacing: normal;
+    text-align: left;
+  }
 
-h1,
-.h1,
-h2,
-.h2,
-h3,
-.h3,
-h4,
-.h4,
-h5,
-.h5,
-h6,
-.h6 {
-  font-weight: normal;
-  letter-spacing: normal;
-}
+  h1,
+  .h1,
+  h2,
+  .h2,
+  h3,
+  .h3,
+  h4,
+  .h4,
+  h5,
+  .h5,
+  h6,
+  .h6 {
+    font-weight: normal;
+    letter-spacing: normal;
+    padding-top: 0;
+  }
 
-ul,
-ol {
-  font-family: Arial, Helvetica, sans-serif;
-}
+  ul,
+  ol {
+    font-family: Arial, Helvetica, sans-serif;
+  }
 
-label {
-  font-family: Arial, Helvetica, sans-serif;
-  margin-bottom: 10px;
-}
+  label {
+    font-family: Arial, Helvetica, sans-serif;
+    margin-bottom: 10px;
+  }
 
-a {
-  background-color: transparent;
-  color: rgb(40, 48, 61);
-  cursor: pointer;
-  text-decoration-skip-ink: all;
-  text-decoration-thickness: 1px;
-  text-underline-offset: 3px;
+  input {
+    font-family: Arial, Helvetica, sans-serif;
+  }
+
+  a {
+    background-color: transparent;
+    color: rgb(40, 48, 61);
+    cursor: pointer;
+    text-decoration-skip-ink: all;
+    text-decoration-thickness: 1px;
+    text-underline-offset: 3px;
+  }
+
+  button,
+  input[type='button'],
+  input[type='reset'],
+  input[type='submit'] {
+    text-transform: none;
+  }
 }
--- a/assets/css/src/components-public/_resets.scss
+++ b/assets/css/src/components-public/_resets.scss
@@ -1,25 +1,8 @@
 // basic default styling
 // to allow for consistent styling between mailpoet plugin and user's theme
 
-html {
-  box-sizing: border-box;
-}
-
-*,
-*:before,
-*:after {
-  box-sizing: inherit;
-}
-
-.mailpoet_form {
-  body {
-    color: rgb(40, 48, 61);
-    font-family: Arial, Helvetica, sans-serif;
-    font-size: 16px;
-    font-weight: normal;
-    text-align: left;
-  }
-
+.mailpoet_form_html,
+.mailpoet_form_iframe {
   html,
   body,
   p,
@@ -44,7 +27,6 @@ html {
   h4,
   h5,
   h6 {
-    color: rgb(40, 48, 61);
     font-family: Arial, Helvetica, sans-serif;
     font-size: 16px;
     -moz-osx-font-smoothing: grayscale;
@@ -67,7 +49,7 @@ html {
   .h5,
   h6,
   .h6 {
-    font-weight: normal;
+    font-weight: 700;
     letter-spacing: normal;
     padding-top: 0;
   }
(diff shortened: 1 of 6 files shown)


assets/css/src/components-public/_resets.scss
```diff
@@ -0,0 +1,82 @@
...
+}
+
+*,
+*:before,
+*:after {
+  box-sizing: inherit;
+}
+
+body {
+  color: rgb(40, 48, 61);
```
wxa, line 15: I'm just curious, what is the base for these styles, WP default admin theme?

wxa (changes requested):
In Firefox on Windows there are some differences between the form editor and the actual iframe on a blank page, some of them because of CSS resets (e.g. `font-weight`):

image

assets/css/src/components-public/_resets.scss
triple0t: @wxa  Yes, the base style was gotten from the WP default admin theme

lib/Form/Util/CustomFonts.php
```diff
@@ -87,12 +87,26 @@ public function enqueueStyle() {
       // When we load all custom fonts in one request, a form from WC Payments isn't displayed correctly.
       // It looks that the larger file size overloads the Stripe SDK.
       foreach (array_chunk(self::FONTS, self::FONT_CHUNK_SIZE) as $key => $fonts) {
-        $this->wp->wpEnqueueStyle('mailpoet_custom_fonts_' . $key, $this->generateLink($fonts));
```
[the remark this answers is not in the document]
triple0t: @wxa I totally understand. I wanted to reuse the generateLink function without having to create a new instance of the class, hence the static here.

My idea behind choosing the static route instead of the instance route is due to the function being a pure function. It requires no additional dependency to use. I believe it's actually much earlier to test as well.

This was supposed to be used here: `lib/Form/Widget.php` but unfortunately an instance of `lib/Form/Util/CustomFonts.php` class does not exist at that point of Widget class initialization on the ContainerWrapper::getInstance()-get method. 

I would have had to create a new instance of CustomFonts class providing all the additional dependencies. This is not really scalable as it is not making use of the Dependency injection mechanism and adding more dependencies to the CustomFonts class in the future will break the Widget class.

These were my reasoning for using the static function.

triple0t:
> In Firefox on Windows there are some differences between the form editor and the actual iframe on a blank page, some of them because of CSS resets (e.g. `font-weight`):
> 
> image

Is this to be expected. We actually have different styling between the Editor and the Form display.

This Ticket is made to fix the inconsistency between the Form display on the user site and in an iFrame.

Fixing the inconsistent styles between the Editor and the Form display would have to be done in a separate ticket. I suspect it might be a little bit more involved and i don't think it is okay to mix that with this fix.

> Some of them because of CSS resets (e.g. `font-weight`)
The CSS resets adds a very minimal styles to the form. If any of the form component for instance added a font-weight property, this will supersede the CSS reset.

> Link is dar color instead of white
The CSS resets adds a very minimal styles to the form. The Users' theme may override the link styles or the Form compoent may do that as well. We are currently not setting link styles from the Form compoent hence the user theme decides the link colors.

lib/Form/Util/CustomFonts.php
```diff
@@ -87,12 +87,26 @@ public function enqueueStyle() {
...
       }
     }
   }
 
-  private function generateLink(array $fonts): string {
+  public static function getCustomFontLink() {
+    $output = '';
+
+    foreach (array_chunk(self::FONTS, self::FONT_CHUNK_SIZE) as $key => $fonts) {
+      $output .= self::generateLink($fonts);
```
wxa, line 99: You are concatenating URLs here without any delimiter. Each one needs to be wrapped in a separate `<link>` tag, otherwise it won't work.

lib/Form/Util/CustomFonts.php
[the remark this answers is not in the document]
wxa: Thanks for explaining. 

>  unfortunately an instance of lib/Form/Util/CustomFonts.php class does not exist at that point of Widget class initialization on the ContainerWrapper::getInstance()-get method.

It's because the `Widget` class itself does not use dependency injection for some reason. If you set the `CustomFonts` class as public in `lib/DI/ContainerConfigurator`, it will be possible to explicitly get an instance of it in `Widget` via the container. Then the reasoning for not using an instance because of an overhead becomes irrelevant.

> My idea behind choosing the static route instead of the instance route is due to the function being a pure function. It requires no additional dependency to use. I believe it's actually much earlier to test as well.

Even if a static method has no dependencies, it's harder to stub it out when it's used in another class. I thought we had a coding rule for that, but I can't find it, so I will ask the team in Slack.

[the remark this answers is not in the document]
wxa: Update: as [discussed in Slack](https://a8c.slack.com/archives/C01G2EDTMQF/p1637136285030500), preferring instance methods over static ones looks to be a rule we missed in our coding guidelines. We'll update them.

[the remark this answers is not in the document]
triple0t: @wxa 
Noted.
I will update the PR.

lib/Form/Util/CustomFonts.php
triple0t: @wxa 
Not really.
The URL is not been concatenated here.

This will generate the full HTML link required in the `<link href>` element e.g the output would be `'https://fonts.googleapis.com/css?family=' . TOTAL_FONTS`

You can locate that here:  https://github.com/mailpoet/mailpoet/blob/1147ba6c8e6474ad9ed12c653d36365610fc56de/lib/Form/Util/CustomFonts.php#L109

This is the part that actually generates the HTML link: https://github.com/mailpoet/mailpoet/blob/1147ba6c8e6474ad9ed12c653d36365610fc56de/lib/Form/Util/CustomFonts.php#L105

wxa:
> This Ticket is made to fix the inconsistency between the Form display on the user site and in an iFrame.

Thanks for pointing out. These look closer. Still some differences due to a theme overriding certain styles. 

And I'm still not sure about the good of CSS resets, the same example with the bold font: in editor it's bold, on the site with a theme it's bold, in an iframe it's not - because of a CSS reset rule. If it wasn't for the rule, it would be bold by default, because it's an `h2`, after all. And that's just one example I noticed.

triple0t:
> > This Ticket is made to fix the inconsistency between the Form display on the user site and in an iFrame.
> 
> Thanks for pointing out. These look closer. Still some differences due to a theme overriding certain styles.
> 
> And I'm still not sure about the good of CSS resets, the same example with the bold font: in editor it's bold, on the site with a theme it's bold, in an iframe it's not - because of a CSS reset rule. If it wasn't for the rule, it would be bold by default, because it's an `h2`, after all. And that's just one example I noticed.

@wxa 
Yes, the Theme override takes precedence.

We are trying to make sure the iFrame form matches as close as possible to the Form display on the site.

Also, Different browser vendors provide different default styles for each HTML element. How it may appear on Chrome may not match that on Safari.

Would you suggest not resetting the Header tags (h1 - h6) font-weight in this case?

wxa:
> We are trying to make sure the iFrame form matches as close as possible to the Form display on the site.

That's the goal, but I'm not sure how much do CSS resets help that. They are overridden by form editor and WP theme respectively, so they are only applied to an iframe, where they override default browser styles. How does that help an embedded form be closer to an iframe, that's an open question.

> Would you suggest not resetting the Header tags (h1 - h6) font-weight in this case?

That would be appropriate, at the least.

lib/Form/Util/CustomFonts.php
triple0t: @wxa Yes, it is expected.

This makes use of one HTTP request to download all the required fonts. 

They are not multiple links but instead multiple font families, separated with a pipe character (|).

You can read more on that here: https://developers.google.com/fonts/docs/getting_started#overview

wxa: @triple0t 

> Yes, it is expected.

> This makes use of one HTTP request to download all the required fonts.

> They are not multiple links but instead multiple font families, separated with a pipe character (|).

There is no pipe character between URLs. There are three links merged together without any delimiter. Let me mark the issue visually for you:

image

triple0t: @wxa Wow, nice catch. Thanks for noticing this 🙌🏿.

triple0t:
> > We are trying to make sure the iFrame form matches as close as possible to the Form display on the site.
> 
> That's the goal, but I'm not sure how much do CSS resets help that. They are overridden by form editor and WP theme respectively, so they are only applied to an iframe, where they override default browser styles. How does that help an embedded form be closer to an iframe, that's an open question.
>
> > Would you suggest not resetting the Header tags (h1 - h6) font-weight in this case?
> 
> That would be appropriate, at the least.

Quick correction on this, the style will only be overridden if it does not exist on the Mailpoet plugin. Due to setting the style on the reset, the style applied supersedes the theme style. Also, we are sharing the same style between the Form display and the iFrame. We just had to account for styles not already set. 

Are you noticing a difference between the Form Display on the site (When using the MailPoet Subscription Form block) and the iFrame?

wxa:
@triple0t 

> Are you noticing a difference between the Form Display on the site (When using the MailPoet Subscription Form block) and the iFrame?

Yes, I do:

* Font is not bold;
* Text input font is different;
* Button text is capitalized.

image

triple0t:
> MailPoet Subscription Form

Noted.
I will take a look.
Thank you

triple0t:
I have fixed all the other issues but i am not able to fix the Text input font is different error.

The Font family selected is 'Cairo' but the rendered font is Montserrat

It makes no sense why that is happening.

This error is currently specific to the Twenty Sixteen theme and i can't replicate it on the other default theme versions.

Default themes

[2011](https://d.pr/i/jlGYKC)

[2016](https://d.pr/i/HWFL5M)

[2017](https://d.pr/i/MuPNit)

[2019](https://d.pr/i/MDtleE)

[2020](https://d.pr/i/ohKwDA)

[2021](https://d.pr/i/KrQQou)

assets/css/src/components-public/_resets.scss
```diff
@@ -11,72 +11,94 @@ html {
...
+  h6 {
+    color: rgb(40, 48, 61);
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 16px;
+    -moz-osx-font-smoothing: grayscale;
+    -webkit-font-smoothing: antialiased;
+    font-synthesis: none;
+    font-weight: normal;
+    letter-spacing: normal;
+    text-align: left;
```
wxa, line 55: Now the text is aligned left in the editor:

image

Should these resets be applied to the editor at all?

assets/css/src/components-public/_resets.scss
```diff
@@ -0,0 +1,104 @@
...
+
+html {
+  box-sizing: border-box;
+}
+
+*,
+*:before,
+*:after {
+  box-sizing: inherit;
+}
```
wxa, line 12: The `mailpoet-public.css` file may be included even on pages where there are no subscription forms. Anyway, styles applied to elements globally may affect page elements outside of MailPoet scope. Do we need such side-effects?

assets/css/src/components-public/_resets.scss
```diff
@@ -0,0 +1,104 @@
...
+  body {
+    color: rgb(40, 48, 61);
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 16px;
+    font-weight: normal;
+    text-align: left;
+  }
+
+  html,
+  body,
```
wxa, line 24: How `html` and `body` tags can be inside a `.mailpoet_form` class? This class is applied to a `div` inside the iframe `body`, as far as I can see. At the same time, we don't want to affect elements which don't belong to MailPoet, see my [other comment](https://github.com/mailpoet/mailpoet/pull/3798/files#r752046427).

wxa:
@triple0t Thank you for the changes.

> i am not able to fix the Text input font is different error.

> The Font family selected is 'Cairo' but the rendered font is Montserrat

That's minor, not a problem. However, I have made a few more remarks. I am still concerned about CSS resets, right now they still provide inconsistent styles between different views, at the same time they may affect other unrelated things if their scope isn't properly set. Have you checked that other form templates look consistently?

triple0t:
Hi @wxa 
Thanks for the review  🙌🏿


> That's minor, not a problem. However, I have made a few more remarks. I am still concerned about CSS resets, right now they still provide inconsistent styles between different views, at the same time they may affect other unrelated things if their scope isn't properly set. Have you checked that other form templates look consistently?

I understand and share your worries as well.

I will refactor the reset style to only match the iFrame

assets/css/src/components-public/_resets.scss
triple0t: Fixed this issue by scoping the styles to the Form Frontend

wxa:
@triple0t Looks better now! Thank you!

P.S.: I see the difference between the form block and shortcode placement methods (capitalized button text, bottom link color), maybe that can be solved in a later ticket:

image

veljkho (approved):

PR #3814, opened 2021-11-18
Remove re-engagement-email feature flag [MAILPOET-3859]

rodrigoprimo:
@costasovo, before reviewing this PR, please check the comment that I left on the Jira ticket.

@veljkho, when this PR gets to you for QA, please note that it should not be merged before #3822 (it is fine to include both PRs in the same release, but we should not include this one without including #3822).

[MAILPOET-3859]

[MAILPOET-3859]: https://mailpoet.atlassian.net/browse/MAILPOET-3859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/Cron/Daemon.php
+++ b/lib/Cron/Daemon.php
@@ -83,9 +83,7 @@ class Daemon {
     yield $this->workersFactory->createSubscribersEngagementScoreWorker();
     yield $this->workersFactory->createSubscribersLastEngagementWorker();
     yield $this->workersFactory->createSubscribersCountCacheRecalculationWorker();
-    if ($this->featureSwitch->isSupported(FeaturesController::RE_ENGAGEMENT_EMAIL)) {
-      yield $this->workersFactory->createReEngagementEmailsSchedulerWorker();
-    }
+    yield $this->workersFactory->createReEngagementEmailsSchedulerWorker();
     yield $this->workersFactory->createNewsletterTemplateThumbnailsWorker();
   }
 }
--- a/lib/Cron/Daemon.php
+++ b/lib/Cron/Daemon.php
@@ -3,7 +3,6 @@
 namespace MailPoet\Cron;
 
 use MailPoet\Cron\Workers\WorkersFactory;
-use MailPoet\Features\FeaturesController;
 
 class Daemon {
   public $timer;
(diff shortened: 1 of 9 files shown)


lib/Cron/Daemon.php
```diff
@@ -83,9 +83,7 @@ private function getWorkers() {
     yield $this->workersFactory->createSubscribersEngagementScoreWorker();
     yield $this->workersFactory->createSubscribersLastEngagementWorker();
     yield $this->workersFactory->createSubscribersCountCacheRecalculationWorker();
-    if ($this->featureSwitch->isSupported(FeaturesController::RE_ENGAGEMENT_EMAIL)) {
```
costasovo, line 86: I think that this was the only usage of `$this->featureSwitch` within the class. Could you please also remove the dependency from the constructor?

costasovo (changes requested):
Good job! I've added code comment!

Thanks for the heads up about the link to re-engagement page. I was able to replicate the same problem and it seems to me that there is some functionality missing. I'll share more details in [the Jira ticket](https://mailpoet.atlassian.net/browse/MAILPOET-3859).

rodrigoprimo:
Good catch, @costasovo. I pushed a new commit removing this property. I will ask for another review once I finish implementing the code to parse the new shortcode in a separate PR.

rodrigoprimo:
Since #3822 was approved, this PR is not ready for a final review.

costasovo (dismissed):
👍

veljkho (changes requested):
Everything's good except one thing:

When clicking the button in the email **[Yes, keep me subscribed!](https://d.pr/i/3TU4um)**, I see [404 page error](https://d.pr/i/xH48rr). This is because the button link is:

https: //domain-name.com/?mailpoet_router&endpoint=track&action=click&data=WzE1MzksImJmNzhhNjM1OTBlN2QwZDQwZDcxNmFmNjIxODVkNzUyIiwiMjgxOSIsIjE1MjVkMDk1ZDE4NyIsZmFsc2Vd

and where I landed is:

https: //domain-name.com/[link:subscription_re_engage_url]

Looks like we have shortcode in the URL. I tested the page in Settings > Basics.. where [it worked](https://d.pr/i/LdG9pv).

costasovo:
> When clicking the button in the email Yes, keep me subscribed!, I see 404 page error. 

@veljkho The issue happened because the shortcode processing was missing. We added it in [a separate PR](https://github.com/mailpoet/mailpoet/pull/3822). It was already merged so I've rebased this branch and the link should now work as expected. Can you please test it again?

veljkho (approved):
Thanks for letting me know @costasovo , this was the case. Feature is working as expected!

rodrigoprimo:
Thanks for stepping in and rebasing the branch, @costasovo!

PR #3815, opened 2021-11-22
Sync WooCommerce subscribers in cron batches [MAILPOET-3753]

lysyjan:
As a part of this pull request, I refactored the code into Doctrine.

Also, I was thinking a lot about preventing this cron job before performance issues. If you see a better solution or have any ideas, I would be glad about that.

### DIFF
--- a/lib/Segments/WooCommerce.php
+++ b/lib/Segments/WooCommerce.php
@@ -1,8 +1,10 @@
-<?php
+<?php declare(strict_types = 1);
 
 namespace MailPoet\Segments;
 
 use MailPoet\Config\Env;
+use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Entities\SubscriberSegmentEntity;
 use MailPoet\Models\ModelValidator;
 use MailPoet\Models\Segment;
 use MailPoet\Models\Subscriber;
@@ -12,9 +14,13 @@ use MailPoet\Subscribers\Source;
 use MailPoet\Subscribers\SubscribersRepository;
 use MailPoet\WooCommerce\Helper as WCHelper;
 use MailPoet\WP\Functions as WPFunctions;
-use MailPoetVendor\Idiorm\ORM;
+use MailPoetVendor\Carbon\Carbon;
+use MailPoetVendor\Doctrine\DBAL\Connection;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class WooCommerce {
+  public const BATCH_SIZE = 10000;
+
   /** @var SettingsController */
   private $settings;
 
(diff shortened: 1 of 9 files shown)


lib/Segments/WooCommerce.php
```diff
@@ -146,22 +167,35 @@ public function synchronizeGuestCustomer($orderId) {
...
+    }
 
-    return true;
+    return count($insertedUsersEmails);
+  }
+
+  public function resetSynchronization(): void {
+    $subscribersTable = $this->entityManager->getClassMetadata(SubscriberEntity::class)->getTableName();
+    $this->connection->executeQuery("
+      UPDATE {$subscribersTable}
```
costasovo, line 199: This updates all rows in subscribers table. I'm a bit afraid that this may cause some queuing of queries because of locking.

lib/Segments/WooCommerce.php
```diff
@@ -202,260 +236,285 @@ private function needsCollationChange(): bool {
...
 
-    return $insertedUsersEmails;
-  }
+    $usersEmails = $this->connection->executeQuery('
+      SELECT DISTINCT wppm.meta_value as email FROM `' . $wpdb->prefix . 'postmeta` wppm
+      JOIN `' . $wpdb->prefix . 'posts` p ON wppm.post_id = p.ID AND p.post_type = "shop_order"
+      WHERE wppm.meta_key = "_billing_email" AND wppm.meta_value != ""
+      ' . ($orderId ? ' AND p.ID = :orderId' : $subQuery) . '
+      ORDER BY wppm.meta_value
+      LIMIT :batchSize
```
costasovo, line 275: The subquery will return more and more emails as the sync progresses so in the last iteration for a shop with 100K email the query would return 90k emails and MySQL would have to compare the meta_value with all these string in the NOT IN condition.
```sql
AND wppm.meta_value NOT IN (
        SELECT email
        FROM {$subscribersTable}
        WHERE is_woocommerce_synced = 1
        ORDER BY email
    )
  ```
  I have an idea for a different approach for the batching. I think we could use orderId, which is a primary key in wp_post table. We could find the lowestOrderId and  highestOrderId and start fetching using range of orderids.
We could store the last processed orderId in the scheduled task meta.
 
 Here is an example of such a query:
 ```sql
 SELECT DISTINCT wppm.meta_value as email FROM wp_postmeta wppm
      JOIN  wp_posts p ON wppm.post_id = p.ID AND p.post_type = "shop_order"
      WHERE wppm.meta_key = "_billing_email" AND wppm.meta_value != ""
      AND p.ID > 1000 and p.ID < 4000
      ORDER BY wppm.meta_value;
 ```
 With this approcach we could remove  `is_woocommerce_synced` and also the potentialy problematic update in reset at the beginning.

costasovo (commented):
Good job! This is quite a huge refactor. 👏 
I have an idea for a different approach for batching. See my comments.

lib/Segments/WooCommerce.php
lysyjan: Thanks for this idea. These changes make sense to me. I will prepare additional changes.

costasovo (approved):
Great job! 👏

veljkho (approved):

PR #3820, opened 2021-11-24
New segment: subscribed to list [MAILPOET-3502]

pavel-mailpoet:
[MAILPOET-3502]

[MAILPOET-3502]: https://mailpoet.atlassian.net/browse/MAILPOET-3502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
+++ b/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
@@ -7,6 +7,14 @@ import Select from 'common/form/select/select';
 
 import { AnyValueTypes, WordpressRoleFormItem } from '../types';
 
+export function validateSubscribedToList(formItems: WordpressRoleFormItem): boolean {
+  return (
+    (formItems.operator !== AnyValueTypes.ANY)
+    && (formItems.operator !== AnyValueTypes.ALL)
+    && (formItems.operator !== AnyValueTypes.NONE)
+  );
+}
+
 type Props = {
   filterIndex: number;
 }
--- a/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
+++ b/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
@@ -1,0 +1,7 @@
+import React from 'react';
+
+export const SubscribedToList: React.FunctionComponent = () => (
+  <div>
+    sdaf
+  </div>
+);
--- a/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
+++ b/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
@@ -1,7 +1,49 @@
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useDispatch, useSelect } from '@wordpress/data';
 
-export const SubscribedToList: React.FunctionComponent = () => (
-  <div>
-    sdaf
-  </div>
-);
+import MailPoet from 'mailpoet';
+import { Grid } from 'common/grid';
+import Select from 'common/form/select/select';
+
+import { AnyValueTypes, WordpressRoleFormItem } from '../types';
+
+type Props = {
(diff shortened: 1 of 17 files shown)


assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
```diff
@@ -0,0 +1,100 @@
...
+        dimension="small"
+        isFullWidth
+        isMulti
+        placeholder={MailPoet.I18n.t('searchLists')}
+        options={options}
+        value={
+          filter(
+            (option) => {
+              if (!segment.segments) return undefined;
+              const segmentId = Number(option.value);
```
wxa, line 85: You shouldn't convert the value to `Number` here, because you'll compare it to an array of strings on the next line. Currently it will always return false, so pre-selected segments will be always empty.

assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
```diff
@@ -0,0 +1,100 @@
...
+              return segment.segments.indexOf(segmentId) !== -1;
+            },
+            options
+          )
+        }
+        onChange={(options: SelectOption[]): void => {
+          updateSegmentFilter(
+            { segments: options.map(compose([Number, get('value')])) },
```
wxa, line 93: Here you also won't need conversion to `Number` after fixing the issue above.

wxa (changes requested):
Besides the remarks below, I have one more: subscriber count for a segment is not recalculated after editing it - if it's not easy to fix, let me know, I'll create a new ticket for this.

assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
pavel-mailpoet: You are right, my thinking was to make everything nubmers because IDs of segments are stored in the database as strings. So I was thinking that we could maybe rather re-type the values to numbers in the server. What do you think?

wxa: Strings or numbers, it's up to you, really, it just needs to be consistent and backward-compatible. In the [subscriber score segment PR](https://github.com/mailpoet/mailpoet/pull/3821) I decided to store values as strings, so if you change to numbers, be sure not to break it.

pavel-mailpoet:
@wxa Thanks for the notes, I made the IDs numbers and everything seems to be working. including editing and recalculation.

assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_subscribed_to_list.tsx
```diff
@@ -0,0 +1,100 @@
...
+        (segment.operator !== AnyValueTypes.ANY)
+        && (segment.operator !== AnyValueTypes.ALL)
+        && (segment.operator !== AnyValueTypes.NONE)
+      )
+    ) {
+      updateSegmentFilter({ operator: AnyValueTypes.ANY }, filterIndex);
+    }
+  }, [updateSegmentFilter, segment, filterIndex]);
+  const options = staticSegmentsList.map((currentValue) => ({
+    value: currentValue.id.toString(),
```
wxa, line 58: There are unnecessary conversions: here you convert segment IDs to strings, later you convert them back to numbers (see my previous comments). All this can be get rid of.

wxa (approved):
Good job! Thank you for fixes! This branch and https://github.com/mailpoet/mailpoet/pull/3821 will have a conflict, so one of us will have to rebase and fix them. Depends on which one @veljkho merges first.

wxa (approved):

veljkho (changes requested):
Everything's good except:
- The [middle column is bigger](https://d.pr/i/iYEFqy) than the other two (1st and 3rd). We should have all three cols the same width.

wxa (approved):

veljkho (approved):

PR #3823, opened 2021-11-26
Unify tracking settings [MAILPOET-3185]

costasovo:
This PR removes "Open and click tracking" setting from MailPoet > Settings > Advanced Tab and "Enable Browser Cookies" setting from WooCommerce Tab. These settings are replaced with a new tracking setting which has 3 levels: `basic`, `partial`, and `full`.

Specification point `1.g) It’s up to the developer if we keep the same settings (and always update both when the tracking level changes) or create a new one`.

Because the values were stored in totally different properties (`tracking.enabled` and `woocommerce. accept_cookie_revenue_tracking.enabled`) within the settings I chose to use a new property `tracking.level`.

Choosing to use a new property meant that I needed to go through the codebase and replace all usages. I added a new service called `TrackingConfig` which holds the information about which tracking (so far we have email tracking and cookie tracking) is enabled for the selected levels of the tracking. I chose to add this service so that we can easily change or add what is tracked for specific levels without the need of migrating something in the DB.

### Note for QA
Apart from the functionality mentioned in the Jira ticket, I needed to do some changes to the WooCommerce Setup Screen which is displayed as the last step of the onboarding wizard or after someone activates WooCommerce for the first time.

Annotation on 2021-11-24 at 10-09-26

Newly when a user enables cookie tracking it switches tracking settings to the full mode and in case they disable it switches the setting to the partial mode or in case it was set to basic before it keeps it set to basic.

[MAILPOT-3185]


views/settings.html
```diff
@@ -290,6 +290,12 @@
     'testEmailBody': __('Yup, it works! You can start blasting away emails to the moon.'),
     'emailSent': __('The email has been sent! Check your inbox.'),
     'orCancel': __('or Cancel'),
+
+    'engagementTrackingTitle': __('Engagement analytics tracking'),
+    'engagementTrackingDescription': __('Choose the balance between subscriber privacy and deeper engagement insights that is right for your business.'),
+    'engagementTrackingFull': __('Full - additionally use browser cookies to improve the site and email engagement tracking and abandoned cart email accuracy.'),
```
pavel-mailpoet, line 291: I think it would look nicer if you used a dash here instead of a hyphen -. You can use n-dash: – or m-dash: —

pavel-mailpoet (changes requested):

pavel-mailpoet (approved):

veljkho (changes requested):
After deleting MailPoet and removing all the tables in MySQL (related to MP of course) and installing MailPoet from scratch again... I was [stuck at the last step of setup wizard](https://d.pr/i/Ey0wpv) where I unchecked the `enable cookie tracking` checkbox and clicked `Start using MailPoet`.

Other than this, everything seems good 👍 .

costasovo:
@pavel-mailpoet I added one commit addressing the issue reported by @veljkho.

pavel-mailpoet (approved):

veljkho (changes requested):
After disabling both `Open and click tracking` & `Enable browser cookies` and updating to this PR's build, I've seen the setting set to Full in Advanced tab but without a notice from the specs. According to Specs, it should show a notice if we set to Full for a user who didn't have options checked. Otherwise, we should switch to Basic option.

veljkho (approved):
DB migration needed to be performed, so it works actually. Merging this.

PR #3824, opened 2021-11-26
Allow multiple values in "purchased this product" segment [MAILPOET-3954]

lysyjan:
[MAILPOET-3954]

[MAILPOET-3954]: https://mailpoet.atlassian.net/browse/MAILPOET-3954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/Config/Migrator.php
+++ b/lib/Config/Migrator.php
@@ -7,6 +7,7 @@ use MailPoet\Entities\FormEntity;
 use MailPoet\Models\Newsletter;
 use MailPoet\Models\Subscriber;
 use MailPoet\Segments\DynamicSegments\Filters\UserRole;
+use MailPoet\Segments\DynamicSegments\Filters\WooCommerceProduct;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Util\Helpers;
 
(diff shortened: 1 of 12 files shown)


costasovo (approved):
Great job!

veljkho (changes requested):
1. The specs say: a) Change the name to `purchased product` but instead it is now `purchased products`.

2. The specs say: e) Make sure the existing segments with a single value continue working
This is not the case now, I've lost all subscribers who have bought a product, I had this segment before this change and it had 6 subscribers.. now says 0 subscribers for that segment.

lysyjan:
1. It is fixed. Thanks!
2. When you run migrations, it should be solved.

lib/Config/Migrator.php
```diff
@@ -730,4 +732,44 @@ private function migrateSerializedFilterDataToNewColumns(): bool {
 
     return true;
   }
+
+  private function migratePurchasedProductDynamicFilters(): bool {
+    global $wpdb;
+    // skip the migration if the DB version is higher than 3.74.2 or is not set (a new install)
+    if (version_compare($this->settings->get('db_version', '3.74.2'), '3.74.1', '>')) {
```
costasovo, line 739: We released 3.74.2 so we need to update the migration check.

costasovo (changes requested):
We need to update the version check in data migration.

costasovo (approved):

veljkho (approved):

PR #3828, opened 2021-11-30
Allow guest users in WooCommerce dynamic segments [MAILPOET-3576]

lysyjan:
[MAILPOET-3576]

[MAILPOET-3576]: https://mailpoet.atlassian.net/browse/MAILPOET-3576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
+++ b/tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types = 1);
 
 namespace MailPoet\Segments\DynamicSegments\Filters;
 
@@ -7,9 +7,10 @@ use MailPoet\Entities\DynamicSegmentFilterData;
 use MailPoet\Entities\DynamicSegmentFilterEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SubscriberEntity;
+use MailPoetVendor\Doctrine\DBAL\ForwardCompatibility\DriverStatement;
 
 class WooCommerceNumberOfOrdersTest extends \MailPoetTest {
-
+  /** @var WooCommerceNumberOfOrders */
   private $numberOfOrders;
 
   private $orders;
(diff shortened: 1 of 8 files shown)


tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types = 1);
```
rodrigoprimo, line 1: I have seen some test files with strict_types enabled and some with strict_types disabled. I'd be curious to learn more about why you added here and if there is a standard that we follow that says that we should enable strict_types in some circumstances.

tests/integration/Segments/DynamicSegments/Filters/WooCommerceTotalSpentTest.php
```diff
@@ -129,4 +134,15 @@ private function cleanUp(): void {
       $this->wp->wpDeletePost($orderId);
     }
   }
+
+  private function setIsWoocommerceUser(array $emails): void {
```
rodrigoprimo, line 138: I'm inclined to think that we should move this method somewhere else so that we don't have to repeat it here and in WooCommerceNumberOfOrdersTest. Have you considered this option?

rodrigoprimo (commented):
Looks good to me overall. I left two clarifying questions in the comments below.

tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
lysyjan: If I remember correctly, we would prefer to use strict types in new classes. So when I edit an older class, I enable it too sometimes.

tests/integration/Segments/DynamicSegments/Filters/WooCommerceTotalSpentTest.php
lysyjan: I was thinking about a method in the repository, but I don't see any usage. In my opinion, this case is rare, so I created this method twice.

rodrigoprimo:
Thanks for the clarifications, @lysyjan. Looks good to me. Unfortunately, now there is a conflict in a file. Could you please fix it before I approve this PR and move it to QA?

rodrigoprimo (approved):

veljkho (changes requested):
Everything looks good but when it comes to testing sending it doesn't send. It says it has been sent for the newsletter, but nothing has arrived in inbox. Sending to those lists doesn't work.

lysyjan:
I'm closing this PR, and I will open another with better performance.

PR #3833, opened 2021-12-01
Add translations downloading from translate.wordpress.com [MAILPOET-3808]

lysyjan:
This PR removes language packs from release builds and adds downloading from the site translate.wordpress.com.
The plugin downloads only languages that are used on the user site.

**Note for QA**
Downloading translation can be enforced by decreasing value `last_checked` in the table `wp_options`. The target row has `option_name` `_site_transient_update_plugins`. 

[MAILPOET-3808]

[MAILPOET-3808]: https://mailpoet.atlassian.net/browse/MAILPOET-3808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/Config/TranslationUpdater.php
```diff
@@ -77,10 +77,12 @@ private function getTranslationUpdates(array $locales): array {
       $requestBody['plugins'][$this->premiumSlug] = ['version' => $this->premiumVersion];
     }
 
+    // Three seconds, plus one extra second for every 10 plugins.
```
costasovo, line 80: I think you mean `for every 10 locales` don't you?

lib/Config/TranslationUpdater.php
```diff
@@ -77,10 +77,12 @@ private function getTranslationUpdates(array $locales): array {
...
+    $timeout = 3 + (int)(count($locales) / 10);
```
costasovo, line 81: Are 3 seconds enough? Should we keep it 10s or so to be safe?

costasovo (commented):
Good job! 

I think we could also remove code for `./do translations:pack` command. Or do you think we should keep it fo some reason?

lib/Config/TranslationUpdater.php
lysyjan: Yes, you are right.

lib/Config/TranslationUpdater.php
lysyjan: Good point! I used the pattern from the WC plugin. We can use 10 seconds as a minimum.

lysyjan:
> I think we could also remove code for ./do translations:pack command. Or do you think we should keep it fo some reason?

Right, we can remove it. I used it to test a possible API bug, but we don't need it anymore.
I will do it.

costasovo (dismissed):

mailpoet/build.sh
```diff
@@ -74,7 +64,6 @@ rm -rf vendor-prefixed/symfony/dependency-injection/Extension
 
 # Copy release folders.
 echo '[BUILD] Copying release folders'
-cp -Rf lang $plugin_name
```
costasovo, line 77: I think we still need to copy this folder because we distribute `.pot` file in it. Seehttps://mailpoet.atlassian.net/browse/MAILPOET-4028

costasovo (changes requested):
Looks good! I found one issue that I think needs to be addressed.

mailpoet/build.sh
lysyjan: Nice catch! I removed commit, which contains this change.

costasovo (approved):
👍

veljkho (approved):

PR #3834, opened 2021-12-02
Refactor ConfirmationEmailMailer to Doctrine

rodrigoprimo:
@brezocordero, I'm assigning this PR to you as you mentioned that you were working on another task to refactor API::subscribeToLists() to Doctrine. I believe I might need to make some adjustments to the changes that I made to this method depending on what you did.

The code changed here is used in a few places to send a confirmation email to a subscriber (not an exhaustive list):

- When an admin clicks on the link to resend the confirmation email on the page that lists all subscribers of a given list.
- When a WC customer purchase a new product and select the checkbox to subscribe to the WC list.
- When a new WP user is created.

When removing Subscriber::setError() as we discussed in the Slack channel, I opted to replace it with a call to `throw`. There are some disadvantages to this approach that I mentioned in more detail in the corresponding [commit message](https://github.com/mailpoet/mailpoet/commit/92185074428d7bde2afc78d1cb46a9dfcf5394bf). But settled with it as I felt it wasn't worth investing more time improving the solution.

[MAILPOET-3815]

[MAILPOET-3815]: https://mailpoet.atlassian.net/browse/MAILPOET-3815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/API/JSON/v1/Subscribers.php
+++ b/lib/API/JSON/v1/Subscribers.php
@@ -197,8 +197,8 @@ class Subscribers extends APIEndpoint {
 
   public function sendConfirmationEmail($data = []) {
     $id = (isset($data['id']) ? (int)$data['id'] : false);
-    $subscriber = Subscriber::findOne($id);
-    if ($subscriber instanceof Subscriber) {
+    $subscriber = $this->subscribersRepository->findOneById($id);
+    if ($subscriber instanceof SubscriberEntity) {
       if ($this->confirmationEmailMailer->sendConfirmationEmail($subscriber)) {
         return $this->successResponse();
       }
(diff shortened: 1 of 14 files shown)


brezocordero (approved):
Great job @rodrigoprimo !

The code looks good and the exception thrown instead of setErrors works well.
I have also tested a couple of cases from the UI, all worked as expected.

The descriptions in the commits made the review very easy. 
I tend to add comments on the PR instead but I like your approach and will mix it with mine.

veljkho (approved):

PR #3835, opened 2021-12-02
Refactor MPAPI subscribeToLists to Doctrine [MAILPOET-3813]

brezocordero:
This PR refactors 

- \MailPoet\API\MP\v1\API::subscribeToList 
- \MailPoet\API\MP\v1\API::subscribeToLists 

The methods are now in a new class `lib/API/MP/v1/Subscribers.php`
Tests have been moved and refactored from `tests/integration/API/MP/APITest.php` to `tests/integration/API/MP/SubscribersTest.php`

I have reused the class `lib/API/JSON/ResponseBuilders/SubscribersResponseBuilder.php` to build the response replacing 
`->withCustomFields()->withSubscriptions()->asArray();`
I have added a parameter to the function 'build' to add more data on subscriptions if we are calling it from MPAPI. 
This information is redundant but I wanted to avoid modifying current MPAPI response, [as documented here](https://github.com/mailpoet/mailpoet/blob/master/doc/api_methods/GetSubscriber.md).

There are two functions where the old models are still present that require refactoring other components.
* sendSubscriberNotification
* _sendConfirmationEmail : The old models on this function can be refactored with this PR: https://github.com/mailpoet/mailpoet/pull/3834

I have also replaced many of the fixtures for Subscribers and Segments on APITest.php with the Doctrine version.

To test:
```
./do --test "test:integration --file=tests/integration/API/MP/APITest.php"
./do --test "test:integration --file=tests/integration/API/MP/SubscribersTest.php"
```
You can also check that the response structure has not changed after the refactoring.
I haven't tested the integration with a 3rd party plugin but if you know of a good candidate, I wouldn't mind doing it.

[MAILPOET-3813]

[MAILPOET-3820]: https://mailpoet.atlassian.net/browse/MAILPOET-3820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-3813]: https://mailpoet.atlassian.net/browse/MAILPOET-3813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/API/JSON/ResponseBuilders/SubscribersResponseBuilder.php
+++ b/lib/API/JSON/ResponseBuilders/SubscribersResponseBuilder.php
@@ -69,12 +69,12 @@ class SubscribersResponseBuilder {
     ];
   }
 
-  public function build(SubscriberEntity $subscriberEntity): array {
+  public function build(SubscriberEntity $subscriberEntity, bool $isMPAPI = false): array {
     $data = [
       'id' => (string)$subscriberEntity->getId(),
       'wp_user_id' => $subscriberEntity->getWpUserId(),
       'is_woocommerce_user' => $subscriberEntity->getIsWoocommerceUser(),
-      'subscriptions' => $this->buildSubscriptions($subscriberEntity),
+      'subscriptions' => $this->buildSubscriptions($subscriberEntity, $isMPAPI),
       'unsubscribes' => $this->buildUnsubscribes($subscriberEntity),
       'status' => $subscriberEntity->getStatus(),
       'last_name' => $subscriberEntity->getLastName(),
@@ -93,21 +93,31 @@ class SubscribersResponseBuilder {
       'unsubscribe_token' => $subscriberEntity->getUnsubscribeToken(),
       'link_token' => $subscriberEntity->getLinkToken(),
     ];
-    $data = $this->buildCustomFields($subscriberEntity, $data);
-    return $data;
+
+    return $this->buildCustomFields($subscriberEntity, $data);
   }
 
-  private function buildSubscriptions(SubscriberEntity $subscriberEntity): array {
+  private function buildSubscriptions(SubscriberEntity $subscriberEntity, bool $isMPAPI = false): array {
     $result = [];
     $subscriptions = $this->subscriberSegmentRepository->findBy(['subscriber' => $subscriberEntity]);
     foreach ($subscriptions as $subscription) {
       $segment = $subscription->getSegment();
       if ($segment instanceof SegmentEntity) {
-        $result[] = [
+        $extraData = [];
+        if ( $isMPAPI ) {
+          $extraData = [
+            'id' => $subscription->getId(),
+            'subscriber_id' => (string)$subscriberEntity->getId(),
+            'created_at' => $subscription->getCreatedAt()->format(self::DATE_FORMAT),
+          ];
+        }
+        $result[] = array_merge(
+          $extraData,
+          [
           'segment_id' => (string)$segment->getId(),
           'status' => $subscription->getStatus(),
           'updated_at' => $subscription->getUpdatedAt()->format(self::DATE_FORMAT),
-        ];
+          ]);
       }
     }
     return $result;
--- a/lib/API/JSON/ResponseBuilders/SubscribersResponseBuilder.php
+++ b/lib/API/JSON/ResponseBuilders/SubscribersResponseBuilder.php
@@ -69,12 +69,12 @@ class SubscribersResponseBuilder {
     ];
   }
 
-  public function build(SubscriberEntity $subscriberEntity, bool $isMPAPI = false): array {
+  public function build(SubscriberEntity $subscriberEntity): array {
     $data = [
       'id' => (string)$subscriberEntity->getId(),
       'wp_user_id' => $subscriberEntity->getWpUserId(),
       'is_woocommerce_user' => $subscriberEntity->getIsWoocommerceUser(),
-      'subscriptions' => $this->buildSubscriptions($subscriberEntity, $isMPAPI),
+      'subscriptions' => $this->buildSubscriptions($subscriberEntity),
       'unsubscribes' => $this->buildUnsubscribes($subscriberEntity),
       'status' => $subscriberEntity->getStatus(),
       'last_name' => $subscriberEntity->getLastName(),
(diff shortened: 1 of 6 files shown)


tests/integration/API/MP/APITest.php
```diff
@@ -154,191 +98,21 @@ public function testItUsesMultipleListsSubscribeMethodWhenSubscribingToSingleLis
...
-    );
-    $wcSegment = Segment::createOrUpdate(
-      [
-        'name' => 'Default',
-        'type' => Segment::TYPE_WC_USERS,
-      ]
-    );
+    $defaultSegment = $this->getSegment();
+    $this->getSegment('WordPress', SegmentEntity::TYPE_WP_USERS);
+    $this->getSegment('WooCommerce', SegmentEntity::TYPE_WC_USERS);
```
brezocordero, line 111: I have changed the names of the segments to be different because of [MAILPOET-3998]

lib/API/JSON/ResponseBuilders/SubscribersResponseBuilder.php
```diff
@@ -93,21 +93,31 @@ public function build(SubscriberEntity $subscriberEntity): array {
...
-  private function buildSubscriptions(SubscriberEntity $subscriberEntity): array {
+  private function buildSubscriptions(SubscriberEntity $subscriberEntity, bool $isMPAPI = false): array {
     $result = [];
     $subscriptions = $this->subscriberSegmentRepository->findBy(['subscriber' => $subscriberEntity]);
     foreach ($subscriptions as $subscription) {
       $segment = $subscription->getSegment();
       if ($segment instanceof SegmentEntity) {
-        $result[] = [
+        $extraData = [];
+        if ( $isMPAPI ) {
```
pavel-mailpoet, line 107: Hmmm, the code sniffer failed :( we don't use spaces here `if ($isMPAPI) {`

lib/API/JSON/ResponseBuilders/SubscribersResponseBuilder.php
```diff
@@ -69,12 +69,12 @@ private function buildListingItem(SubscriberEntity $subscriber): array {
     ];
   }
 
-  public function build(SubscriberEntity $subscriberEntity): array {
+  public function build(SubscriberEntity $subscriberEntity, bool $isMPAPI = false): array {
```
pavel-mailpoet, line 72: Do you think we should use a different approach than the boolean argument? They are generally considered a [bad practice](https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3). 
But also if some other part of the code would require the extra data they would have to pass the argument `isMPAPI` even though it is a completely different part.

Do we even have to use the argument? What if we included the extra data always. I know it would consume a bit more memory, but I don't think it would be causing any problems. What do you think?

lib/API/MP/v1/Subscribers.php
```diff
@@ -0,0 +1,212 @@
...
+      throw new APIException(__('At least one segment ID is required.', 'mailpoet'), APIException::SEGMENT_REQUIRED);
+    }
+
+    if (empty($subscriberId)) {
+      throw new APIException(__('A subscriber is required.', 'mailpoet'), APIException::SUBSCRIBER_NOT_EXISTS);
+    }
+
+    // throw exception when subscriber does not exist
+    $subscriber = null;
+    if (is_int($subscriberId) || (string)(int)$subscriberId === $subscriberId) {
```
pavel-mailpoet, line 88: I wonder why do we need the double re-typing. It is quite an unusual pattern. I don't think I understand what it does. I know it was already in the code before, but I am confused. Do you know why is it here?

lib/API/MP/v1/Subscribers.php
```diff
@@ -0,0 +1,212 @@
...
+      $subscriber = $this->subscribersRepository->findOneById($subscriberId);
+    } else if (strlen(trim($subscriberId)) > 0) {
```
pavel-mailpoet, line 90: TIL: I didn't know we support this. I didn't know they can pass an email as an ID and it would work.

lib/API/MP/v1/Subscribers.php
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    if (!$subscriber) {
+      throw new APIException(__('This subscriber does not exist.', 'mailpoet'), APIException::SUBSCRIBER_NOT_EXISTS);
+    }
+
+    // throw exception when none of the segments exist
+    $foundSegments = $this->segmentsRepository->findBy(['id' => $listIds]);
+    if (!$foundSegments) {
+      $exception = WPFunctions::get()->_n('This list does not exist.', 'These lists do not exist.', count($listIds), 'mailpoet');
```
pavel-mailpoet, line 101: I didn't know this pattern works. I thought our translation model doesn't support `_n`. Good to know it works.

pavel-mailpoet (changes requested):
Hey Brezo, good job with the pull request, I have a couple of remarks. 

Next time could you please split the change into multiple commits? It is hard for me to review such a big change in a single diff. I think it would be easier for me if you moved the code from one place to another in one commit and then refactored it to doctrine in a subsequent commit(s). That would be easier for me.

lib/API/JSON/ResponseBuilders/SubscribersResponseBuilder.php
brezocordero: I like the idea of always sending the extra data, I guess I was being too cautious about not causing regressions in other parts of the plugin. I will make the change.

lib/API/MP/v1/Subscribers.php
brezocordero: To be honest, I don't know, I brought it over from the other function. I will take a second look and see if we can remove it.

brezocordero:
> Next time could you please split the change into multiple commits? It is hard for me to review such a big change in a single diff. 

I will split it next time. Thanks for reviewing it anyways.

pavel-mailpoet (approved):

veljkho (changes requested):
After clicking Recalculate now button I'm [getting this error](https://d.pr/i/MdvERj).
DB migration didn't help, just to mention what I tried.

brezocordero:
> After clicking Recalculate now button I'm [getting this error](https://d.pr/i/MdvERj). DB migration didn't help, just to mention what I tried.

@veljkho I can't reproduce it... Are there other logs you could give me?

brezocordero:
> After clicking Recalculate now button I'm [getting this error](https://d.pr/i/MdvERj). DB migration didn't help, just to mention what I tried.

@veljkho After debugging in staging, the error is coming from a WordPress User Role filter. The one inside "Ultra Complex Segment That Should Work". It fails on [this line](https://github.com/mailpoet/mailpoet/blob/d39d09dd72e19b680428ffe89a809cb31d81257e/lib/Segments/DynamicSegments/Filters/UserRole.php#L36) because the Filter contains an Array of roles, I believe that it is there from testing another PR.

It should also fail on master branch. My PR is not modifying the code that recalculates the subscribers count cache.

One test that could be useful is to test the API response consumed by another plugin. If you know of one.

veljkho:
Thanks @brezocordero for checking it, I'll assign the PR to myself and test it without having that segment.

veljkho (approved):

PR #3837, opened 2021-12-03
Allow multiple values in "WordPress user role" segment [MAILPOET-3955]

pavel-mailpoet:
[MAILPOET-3955]

[MAILPOET-3955]: https://mailpoet.atlassian.net/browse/MAILPOET-3955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:

2. Click on Segments tab
3. Click to `+ New Segment` button at the top
4. From there, test the segment creation and modification according to [MAILPOET-3955](https://mailpoet.atlassian.net/browse/MAILPOET-3955) specs in Jira ticket.
5. Make sure you follow the specs and that it behaves according to the specification.

### DIFF
--- a/lib/Segments/DynamicSegments/Filters/UserRole.php
+++ b/lib/Segments/DynamicSegments/Filters/UserRole.php
@@ -60,7 +60,14 @@ class UserRole implements Filter {
   private function createCondition(array $roles, string $operator, $parameterSuffix): string {
     $sqlParts = [];
     foreach ($roles as $key => $role) {
-      $sqlParts[] = '(wpusermeta.meta_value LIKE :role' . $key . $parameterSuffix . ')';
+      if ($operator === DynamicSegmentFilterData::OPERATOR_NONE) {
+        $sqlParts[] = '(wpusermeta.meta_value NOT LIKE :role' . $key . $parameterSuffix . ')';
+      } else {
+        $sqlParts[] = '(wpusermeta.meta_value LIKE :role' . $key . $parameterSuffix . ')';
+      }
+    }
+    if ($operator === DynamicSegmentFilterData::OPERATOR_NONE) {
+      return join(' AND ', $sqlParts);
     }
     return join(' OR ', $sqlParts);
   }
--- a/lib/Segments/DynamicSegments/Filters/UserRole.php
+++ b/lib/Segments/DynamicSegments/Filters/UserRole.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Segments\DynamicSegments\Filters;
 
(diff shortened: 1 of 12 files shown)


brezocordero (approved):
Great work @pavel-mailpoet 
Backwards compatibility works with the checks and conversion to Arrays.
The queries with the different operators are well-formed, all the changes on the UI are applied.

lib/Segments/DynamicSegments/Filters/UserRole.php
```diff
@@ -25,15 +26,48 @@ public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $fi
...
+      $role = [$role];
+    }
+    if (!$operator) {
+      $operator = DynamicSegmentFilterData::OPERATOR_ANY;
+    }
 
     $subscribersTable = $this->entityManager->getClassMetadata(SubscriberEntity::class)->getTableName();
-    $parameterSuffix = $filter->getId() ?? Security::generateRandomString();
-    return $queryBuilder->join($subscribersTable, $wpdb->users, 'wpusers', "$subscribersTable.wp_user_id = wpusers.id")
+    $parameterSuffix = ((string)$filter->getId()) . Security::generateRandomString();
```
brezocordero, line 42: Out of curiosity, why are we using this? Is it to avoid hitting the cache?

pavel-mailpoet: Hmm, I will try to explain, but it isn't straightforward, so please ask if you have more questions. Nope, it is not because of any cache. 

This is important when a customer connects multiple segments—we [pass the parameters](https://github.com/mailpoet/mailpoet/blob/master/lib/Segments/DynamicSegments/FilterHandler.php#L50) to the final query. But the customer can make a segment "(Subscribers in list A) OR (Subscribers in list B)". Then we need to pass the data "list A" and "list B" correctly to the final query. 

In other places, we have static parameter names ":segmentId," but that would pass it twice to the final query. Once for each sub-segment. And doctrine would only use "list A" in both queries, and the final query would look like: "(Subscribers in list A) OR (Subscribers in list A)". That means we need to make the parameters unique, so they are replaced correctly in the final big query. 

Does this make any sense?

brezocordero: It does, thanks for explaining.

veljkho (approved):

rohankacharegs:
GlobalStep Verified

PR #3839, opened 2021-12-06
Allow multiple values in purchased in this category segment - MAILPOET-3953

triple0t:
[MAILPOET-3953](https://mailpoet.atlassian.net/browse/MAILPOET-3953)

Testing instructions:

2. Click on Segments tab
3. Click to `+ New Segment` button at the top
4. From there, test the segment creation and modification according to [MAILPOET-3953](https://mailpoet.atlassian.net/browse/MAILPOET-3953) specs in Jira ticket. This change is related to `purchased in category` type of a segment.
5. Make sure you follow the specs and that it behaves according to the specification.


lib/Config/Migrator.php
```diff
@@ -772,4 +774,44 @@ private function migratePurchasedProductDynamicFilters(): bool {
 
     return true;
   }
+
+  private function migratePurchasedInCategoryDynamicFilters(): bool {
+    global $wpdb;
+    // skip the migration if the DB version is higher than 3.74.3 or is not set (a new install)
+    if (version_compare($this->settings->get('db_version', '3.74.3'), '3.74.2', '>')) {
```
costasovo, line 781: This PR won't make it to 3.74.3 so we will need to increase the versions in this check to 3.74.4 respectively 3.74.3.
So that we compare against the previous released version 3.74.3 and we use 3.74.4 as a default value in case of a new install.

costasovo (commented):
Great job! 👏  It worked as expected for me.

Thanks also for fixing the annoying [react warning in the message for the cached lists](https://github.com/mailpoet/mailpoet/pull/3839/files#diff-54b0040dd007bc792b07a470be0b200b4877a072923d96cbe313ba90759b52f0R37-R39)!

The change won't make it to release v3.74.3, so you will need to update the version check in migration. See my comment.

lib/Config/Migrator.php
triple0t: Noted

Thank you @costasovo 🙌🏿

costasovo (approved):
👍

veljkho (dismissed):

lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
```diff
@@ -30,44 +33,126 @@ public function __construct(
...
+      $this->applyTermTaxonomyJoin($queryBuilder, $parameterSuffix);
+
+    } elseif ($operator === DynamicSegmentFilterData::OPERATOR_ALL) {
+      $this->applyPostmetaJoin($queryBuilder, $subscribersTable);
+      $this->applyOrderItemsJoin($queryBuilder);
+      $this->applyOrderItemmetaJoin($queryBuilder);
+      $this->applyTermRelationshipsJoin($queryBuilder);
+      $this->applyTermTaxonomyJoin($queryBuilder, $parameterSuffix)
+      ->groupBy("{$subscribersTable}.id, items.order_id")
+      ->having('COUNT(items.order_id) = :count')
```
costasovo, line 62: @triple0t  Sorry for the late comment. Each parameter name needs to be a unique per filter. When we merge queries for multiple filters we also [need to merge the parameters](https://github.com/mailpoet/mailpoet/blob/master/lib/Segments/DynamicSegments/FilterHandler.php#L50-L59) and if the name is not unique the parameters override. So in this case, if somebody combines 2 these segments with `all` operand and selects different counts of categories only one count will be applied for both subqueries in the combined query.

I've just noticed the same issue in the purchased product segment which is already merged. I'll prepare a fix for the products segment. Could you please fix this PR. You just need to use 

```php
->having('COUNT(items.order_id) = :count' . $parameterSuffix)
->setParameter('count' . $parameterSuffix, count($categoryIds));
```

costasovo (changes requested):

lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
triple0t: @costasovo Noted 👍

costasovo (approved):
Thanks!

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3845, opened 2021-12-07
Move multi-condition segments to Premium [MAILPOET-3929]

wxa:
[MAILPOET-3929](https://mailpoet.atlassian.net/browse/MAILPOET-3929)


assets/js/src/segments/dynamic/form.tsx
```diff
@@ -150,21 +156,16 @@ export const Form: React.FunctionComponent<Props> = ({
...
-              const filters = segment.filters;
-              filters.push({
-                segmentType: SegmentTypes.WordPressRole,
-                action: SubscriberActionTypes.WORDPRESS_ROLE,
-              });
-              updateSegment({
-                filters,
-              });
-            }}
+            onClick={(e): void => addConditionAction(e, segment, updateSegment)}
```
pavel-mailpoet, line 159: Does the function `addConditionAction` really need the Event? Could we call the `preventDefault` function here and then not pass the event further?

pavel-mailpoet (commented):
Good job Alex, just a small question.

assets/js/src/segments/dynamic/form.tsx
wxa: You're right, it's not needed in this callback, I've removed it.

pavel-mailpoet (dismissed):

veljkho (changes requested):
Everything seems good except one thing:
I am able to make an advanced segment even with a warning "Your current MailPoet plan does not support advanced segments.".. I used **3 subscribers plan** so I'm getting a [warning about upgrade](https://d.pr/i/9E6fLO), and I am blocked to edit advanced segments nor to send newsletter to them, but I [am just able to make an advanced segment](https://d.pr/i/eC1gDy).

wxa:
@veljkho Fixed, please update both branches.

veljkho (dismissed):

veljkho (approved):

PR #3846, opened 2021-12-07
Recommend MailPoet in free users emails - MAILPOET-3981

triple0t:
[MAILPOET-3981](https://mailpoet.atlassian.net/browse/MAILPOET-3981)


lib/Config/ServicesChecker.php
```diff
@@ -132,6 +132,23 @@ public function isMailPoetAPIKeyPendingApproval(): bool {
     return $mssActive && $mssKeyValid && $mssKeyPendingApproval;
   }
 
+  public function isUserActivelyPaying(): bool {
+    $isPremiumKeyValid = $this->isPremiumKeyValid(false);
+    $premiumSupportTier = $this->settings->get('premium.premium_key_state.data.support_tier');
+    $isUserPayingForPremium = $premiumSupportTier === 'premium';
```
wxa, line 138: There is already a method for checking this: `$this->subscribersFeature->hasPremiumSupport()`. It's worth keeping such logic in one place by reusing the existing method.

I know that `ServicesChecker` and `SubscribersFeature` have some duplicate checks for legacy reasons. Originally `ServicesChecker` was used for handling license-related error notices, and `SubscribersFeature` for checking if the subscriber limit is reached, but over time their functionality started to overlap as both needed more granular checks for various license-related cases. Ripe for refactoring, it seems.

lib/Config/ServicesChecker.php
```diff
@@ -132,6 +132,23 @@ public function isMailPoetAPIKeyPendingApproval(): bool {
...
+
+    $mssActive = Bridge::isMPSendingServiceEnabled();
+    $isMssKeyValid = $this->isMailPoetAPIKeyValid(false);
+    $mssSupportTier = $this->settings->get('mta.mailpoet_api_key_state.data.support_tier');
+    $isUserPayingForMss = $mssSupportTier === 'premium';
```
wxa, line 143: Same as above: maybe it's better to create a method similar to `hasPremiumSupport()` in `MailPoet\Util\License\Features\Subscribers`.

wxa (changes requested):

lib/Config/ServicesChecker.php
triple0t: Noted.
I wasn’t aware of that method
Thanks for the review @wxa  🙌🏿

wxa (dismissed):

veljkho (approved):

veljkho (approved):

triple0t:
Closing due to Git Merge errors
New PR here: https://github.com/mailpoet/mailpoet/pull/3863

PR #3847, opened 2021-12-07
Remove makepot JS tooling [MAILPOET-4005]

JanJakes:
[MAILPOET-4005]

When working on [Use makepot tool from the free plugin](https://mailpoet.atlassian.net/browse/PREMIUM-167) I was able to [remove almost 200K LOC in almost 900 files in the premium plugin](https://github.com/mailpoet/mailpoet-premium/pull/513) due to removing our `makepot` tooling and its `node_modules`.

But we still have the same tooling in the **free plugin**.

We know that we can’t easily remove this and switch to WP CLI, for instance, because we extract strings from Twig and Handlebars templates using our custom code. After investigating I realized this custom code is only in PHP. What it actually does is:

1) PHP extracts all translation strings to POT.
2) JS parses the POT, changes some headers and outputs it. It also wraps it with Grunt.
3) We don’t seem to use other things from the Gruntfile anymore – we use RoboFile where we call Transifex directly and call GNU gettext utils via some scripts.

So I checked it out and was able to remove the whole JS part and `node_modules` also from the free plugin and keep just the PHP and make it work.

**What I did:**
1. Started to call makepot directly from PHP (the `grunt-makepot.php`).
2. Made POT file output as close as possible to the current state – commented out some POT comments that the JS was stripping away, made headers almost identical.
3. Moved the PHP code under `node_modules` up to `tasks/makepot`.
4. Removed all of the `node_modules` and JS code.

Apart from different line wraps and sorting the POT files before and after the changes are identical (I sorted them with a script and verified the diff).

---

I hope I didn't miss any usage of the Gruntfile but **it really seems to be no longer used**:
- [These lines](https://github.com/mailpoet/mailpoet/blob/728c80d08d568aae79fdfa01d72603da08bdf36f/RoboFile.php#L107-L130) in the `RoboFile.php` don't use it anymore.
- The `translations:push` command [uses `tx` directly](https://github.com/mailpoet/mailpoet/blob/728c80d08d568aae79fdfa01d72603da08bdf36f/RoboFile.php#L128).
- The `transifex_init.sh` [seems to be safe](https://github.com/mailpoet/mailpoet/blob/master/tasks/transifex_init.sh).
- The `pack_translations.sh` [seems to be safe as well](https://github.com/mailpoet/mailpoet/blob/master/tasks/pack_translations.sh) – uses `tx` and `msgfmt`.

[MAILPOET-4005]: https://mailpoet.atlassian.net/browse/MAILPOET-4005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wxa (approved):
@JanJakes Very cool optimization! Works a bit faster without the JS part.
I'm just curious, can the different wrapping of strings cause them to lose translations?

wxa (approved):

JanJakes:
> I'm just curious, can the different wrapping of strings cause them to lose translations?

@wxa I hope it's still OK since it's still a valid POT file. As the PHP code that creates it first comes from old wp-i18n-tools and Transifex seems to be commonly used for WP projects as well, I think this should be good. In some ways we're now closer to the "WP CLI" but a very old (and forked) one.

I made sure no strings are redundant or missing by commenting out the folding part in JS, sorting the strings and looking at their diff. The PHP sometimes folds lines as well although less than JS did. The content remains the same.

PR #3848, opened 2021-12-07
PHP 8.1 Fixes vol.1 [MAILPOET-3890]

costasovo:
This PR addresses some errors and deprecation warnings that I ran into while testing the plugin.

### Deprecation warnings caused by [return types that were added to the build-in classes](https://php.watch/versions/8.1/internal-method-return-types)

I fixed warnings that were coming from Idiorm and moved two abandoned libraries(`tburry/pquery` and `redcomponent/teststatistics`) into the plugin's codebase, and fixed the warnings. Both libraries are not heavily used, and were updated many years ago. So I find moving them into our codebase the most efficient solution. I also update doctrine libraries since I noticed that some PHP8.1 related commits were already merged, but those didn't address the deprecation warnings.

I updated [a Jira ticket for updating PHP dependencies](https://mailpoet.atlassian.net/browse/MAILPOET-4001) with instructions for checking these warnings. Hopefully some of them will be fixed by the time we get to it.

### Fixed [PDO bug 80808](https://bugs.php.net/bug.php?id=80808)
The fix caused that integers and floats are now fetched as integers and floats instead of strings. Fortunately, Doctrine/ORM already returned correct types in entities. But this change caused a couple of issues in cases where we use Doctrine/DBAL or Idiorm for fetching data. I fixed a couple of cases I found when playing with the plugin and when running integration tests. Another possible workaround could be using PDO attribute `PDO::ATTR_STRINGIFY_FETCHES` that would force that everything is fetched as string, but I was not sure about possible side effects so I rather chose to fix individual issues.

### [Deprecated strftime](https://php.watch/versions/8.1/strftime-gmstrftime-deprecated) 
I fixed use-cases in our codebase. Our usage was not dependent on locale so I used `date` function.

[MAILPOET-3890]

[MAILPOET-3890]: https://mailpoet.atlassian.net/browse/MAILPOET-3890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib-3rd-party/pquery/gan_parser_html.php
```diff
@@ -836,4 +836,4 @@ protected function parse_hierarchy($self_close = null) {
 	}
 }
 
-?>
\ No newline at end of file
+?>
```
pavel-mailpoet, line 839: We need to fix this, I think there is a white space and that could cause issues.

pavel-mailpoet (changes requested):

pavel-mailpoet (approved):

veljkho (approved):

PR #3849, opened 2021-12-08
Fix flaky TimestampListenerTest [MAILPOET-3983]

costasovo:
The problem was that the listener was replaced in the EntityManager only
in the first test case. For other test cases, replaceListeners didn't replace
anything because the original listener was not set.
[MAILPOET-3983]

[MAILPOET-3983]: https://mailpoet.atlassian.net/browse/MAILPOET-3983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo (approved):
Good job!

veljkho (approved):

PR #3850, opened 2021-12-09
Allow multiple values in "opened" segment [MAILPOET-3950]

pavel-mailpoet:
There is a lot of code shared with the "clicked" segment. But that is in a different [ticket](https://mailpoet.atlassian.net/browse/MAILPOET-3951). So I left some code in which will be deleted/refactored in an upcoming ticket.

[MAILPOET-3950]

[MAILPOET-3950]: https://mailpoet.atlassian.net/browse/MAILPOET-3950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:
2. Click to Add New Segment
3. Select `opened` segment type
4. Compare results from your testing with the specification from Jira ticket above (MAILPOET-3950)
5. Verify you see all the new improvements and that creating such segmentation works

### DIFF
--- a/lib/Segments/DynamicSegments/Filters/EmailAction.php
+++ b/lib/Segments/DynamicSegments/Filters/EmailAction.php
@@ -48,14 +48,24 @@ class EmailAction implements Filter {
 
   public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $filter): QueryBuilder {
     $filterData = $filter->getFilterData();
+    $operator = $filterData->getParam('operator');
     $action = $filterData->getAction();
+
+    if ($action === self::ACTION_NOT_OPENED) {
+      // for backward compatibility with old segments
+      $action = self::ACTION_OPENED;
+      $operator = DynamicSegmentFilterData::OPERATOR_NONE;
+    }
+
     $newsletterId = $filterData->getParam('newsletter_id');
     if ($newsletterId) {
+      // for backward compatibility with old segments
       $newsletters = [(int)$newsletterId];
     } else {
       $newsletters = $filterData->getParam('newsletters');
     }
-    $operator = $filterData->getParam('operator');
+
+
     $linkId = $filterData->getParam('link_id') ? (int)$filterData->getParam('link_id') : null;
     $parameterSuffix = (string)($filter->getId() ?? Security::generateRandomString());
 
@@ -69,7 +79,8 @@ class EmailAction implements Filter {
 
     $where = '1';
 
-    if (($action === self::ACTION_NOT_CLICKED) || ($action === self::ACTION_NOT_OPENED)) {
+    if (($action === self::ACTION_NOT_CLICKED)) {
+      // TODO remove this if branch in MAILPOET-3951 and merge it with the next one for ACTION_OPENED
       $queryBuilder = $queryBuilder->innerJoin(
         $subscribersTable,
         $statsSentTable,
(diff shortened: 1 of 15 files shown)


lib/Segments/DynamicSegments/Filters/EmailActionClickAny.php
```diff
@@ -0,0 +1,50 @@
...
+namespace MailPoet\Segments\DynamicSegments\Filters;
+
+use MailPoet\Entities\DynamicSegmentFilterEntity;
+use MailPoet\Entities\NewsletterLinkEntity;
+use MailPoet\Entities\StatisticsClickEntity;
+use MailPoet\Entities\SubscriberEntity;
+use MailPoetVendor\Doctrine\DBAL\Query\QueryBuilder;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
+
+class EmailActionClickAny implements Filter {
```
wxa, line 12: Is extraction of `EmailActionClickAny` to a new class related to refactoring of the "clicked" segment? Is there any specific reason why it's done in this ticket and not a follow-up one?

lib/Segments/DynamicSegments/Filters/EmailAction.php
```diff
@@ -75,36 +90,45 @@ public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $fi
...
       )->setParameter('newsletter' . $parameterSuffix, $newsletterId);
       $where .= ' AND stats.id IS NULL';
-    } else if ($action === self::ACTION_CLICKED_ANY) {
-      $excludedLinks = [
-        '[link:subscription_unsubscribe_url]',
-        '[link:subscription_instant_unsubscribe_url]',
-        '[link:newsletter_view_in_browser_url]',
-        '[link:subscription_manage_url]',
-      ];
+    } elseif (($action === EmailAction::ACTION_OPENED) && ($operator === DynamicSegmentFilterData::OPERATOR_NONE)) {
```
wxa, line 96: When I select more than one newsletter with the "none of" option, I get an SQL error while counting the number of subscribers:
> An exception occurred while executing 'SELECT IFNULL(SUM(            CASE WHEN wpdev_mailpoet_subscribers.deleted_at IS NULL              THEN 1 ELSE 0 END        ), 0) as `all`, IFNULL(SUM(            CASE WHEN wpdev_mailpoet_subscribers.deleted_at IS NOT NULL              THEN 1 ELSE 0 END        ), 0) as trash, IFNULL(SUM(            CASE WHEN wpdev_mailpoet_subscribers.status = ? AND wpdev_mailpoet_subscribers.deleted_at IS NULL              THEN 1 ELSE 0 END        ), 0) as ?, IFNULL(SUM(          CASE WHEN wpdev_mailpoet_subscribers.status = ? AND wpdev_mailpoet_subscribers.deleted_at IS NULL            THEN 1 ELSE 0 END        ), 0) as ?, IFNULL(SUM(          CASE WHEN wpdev_mailpoet_subscribers.status = ? AND wpdev_mailpoet_subscribers.deleted_at IS NULL            THEN 1 ELSE 0 END        ), 0) as ?, IFNULL(SUM(          CASE WHEN wpdev_mailpoet_subscribers.status = ? AND wpdev_mailpoet_subscribers.deleted_at IS NULL            THEN 1 ELSE 0 END        ), 0) as ?, IFNULL(SUM(          CASE WHEN wpdev_mailpoet_subscribers.status = ? AND wpdev_mailpoet_subscribers.deleted_at IS NULL            THEN 1 ELSE 0 END        ), 0) as ? FROM wpdev_mailpoet_subscribers INNER JOIN (SELECT DISTINCT wpdev_mailpoet_subscribers.id as inner_subscriber_id FROM wpdev_mailpoet_subscribers INNER JOIN wpdev_mailpoet_statistics_newsletters statssent ON wpdev_mailpoet_subscribers.id = statssent.subscriber_id AND statssent.newsletter_id IN (?, ?, ?) LEFT JOIN wpdev_mailpoet_statistics_opens stats ON statssent.subscriber_id = stats.subscriber_id AND stats.newsletter_id = (?, ?, ?) WHERE 1 AND stats.id IS NULL) a0 ON a0.inner_subscriber_id = wpdev_mailpoet_subscribers.id WHERE wpdev_mailpoet_subscribers.deleted_at IS NULL' with params ["subscribed", "subscribed", "unsubscribed", "unsubscribed", "inactive", "inactive", "unconfirmed", "unconfirmed", "bounced", "bounced", 18, 10, 7, 18, 10, 7]:SQLSTATE[21000]: Cardinality violation: 1241 Operand should contain 1 column(s)

wxa (changes requested):

lib/Segments/DynamicSegments/Filters/EmailActionClickAny.php
pavel-mailpoet: The reason was to make my job easier. There were so many options in the EmailAction segment that it was hard to update it. I had to add some more `if` branches and the code would become harder to read. I decided to refactor this filter to a separate file. It has nothing to do with the other filters anyway.

wxa (approved):
👍

veljkho (dismissed):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3856, opened 2021-12-09
Allow multiple values in "is in country" segment [MAILPOET-3952]

brezocordero:
This PR allows the selection of multiple values for the "is in country" segment and adds the operator None and Any.
It also changes the queries to use WC Lookup tables.

The tests have been modified to reflect the changes and there is a new integration test for the segment.
Because WooCommerce is not present for the integration tests, the PR also includes the creation of the lookup tables for the integration tests. The DB schema has been [copied from WC-Admin](https://github.com/woocommerce/woocommerce-admin/blob/49bc93f7e69d44c9da1cbd290a69092f6361002b/src/Install.php#L235).

To test:
* From the main branch, create a segment with a country.
* Switch to this branch
* Make sure the segment still works although the number of subscribers might be different if [this branch is not merged yet](https://github.com/mailpoet/mailpoet/compare/wc-guest-users).
* Edit the segment from MailPoet > Lists > Segments
* Check you can now select multiple countries
* Check it works with the None or Any operator
* Save the segment
* Check you can see the subscribers on the segment
* Repeat with a new segment
* Check that you can combine "is in country" with other filters.

New test:
`./do --test "test:integration --file=tests/integration/Segments/DynamicSegments/Filters/WooCommerceCountryTest.php"
`

[MAILPOET-3952]


lib/Segments/DynamicSegments/Filters/WooCommerceCountry.php
```diff
@@ -14,33 +16,70 @@ class WooCommerceCountry implements Filter {
...
+      'customer',
+      "$subscribersTable.email = customer.email $collation"
     )->innerJoin(
-      'postmeta',
-      $wpdb->postmeta,
-      'postmetaCountry',
-      "postmeta.post_id = postmetaCountry.post_id AND postmetaCountry.meta_key = '_billing_country' AND postmetaCountry.meta_value = :$countryFilterParam"
-    )->setParameter($countryFilterParam, $countryCode);
+      'customer',
+      $wpdb->prefix . 'wc_order_stats',
```
wxa, line 58: A note on using these lookup tables: on my local WooCommerce installation, which is quite old, some of these tables, e.g. `wc_order_stats`, were empty despite me having orders and customers. So a segment was giving me `0` subscribers no matter the settings. I had to submit new orders to see some data put into this table. Then the segment began to find subscribers, but only new ones. Can something akin to this happen with customer installations or is my case just weird?

wxa (commented):
Looks and works good. Just one question I wanted to address with you.

lib/Segments/DynamicSegments/Filters/WooCommerceCountry.php
brezocordero: Looks like your tables are out of sync with postmeta.
If you run `wp wc tool run regenerate_product_lookup_tables --user=1` do you get the data populated?

> Can something akin to this happen with customer installations or is my case just weird?

Hopefully, this is not a common scenario. We have started using the tables here to solve performance problems: https://github.com/mailpoet/mailpoet/pull/3862/files

wxa: > If you run wp wc tool run regenerate_product_lookup_tables --user=1 do you get the data populated?

The command run successfully, but the tables weren't populated with data. 

> Hopefully, this is not a common scenario.

Okay, let's assume this is a problem only with my WC installation, and not a widespread thing. And it is probably related to WooCommerce, not MailPoet.

wxa (approved):

veljkho (dismissed):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3859, opened 2021-12-10
More comparison options for custom field segment [MAILPOET-4008]

pavel-mailpoet:
[MAILPOET-4008]

[MAILPOET-4008]: https://mailpoet.atlassian.net/browse/MAILPOET-4008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/Segments/DynamicSegments/Filters/MailPoetCustomFields.php
+++ b/lib/Segments/DynamicSegments/Filters/MailPoetCustomFields.php
@@ -128,10 +128,10 @@ class MailPoetCustomFields implements Filter {
       $queryBuilder->andWhere("subscribers_custom_field.value != $valueParam");
       $queryBuilder->setParameter($valueParam, $value);
     } elseif ($operator === 'more_than') {
-      $queryBuilder->andWhere("subscribers_custom_field.value >= $valueParam");
+      $queryBuilder->andWhere("subscribers_custom_field.value > $valueParam");
       $queryBuilder->setParameter($valueParam, $value);
     } elseif ($operator === 'less_than') {
-      $queryBuilder->andWhere("subscribers_custom_field.value <= $valueParam");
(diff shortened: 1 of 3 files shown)


lib/Segments/DynamicSegments/Filters/MailPoetCustomFields.php
```diff
@@ -124,6 +124,15 @@ private function applyEquality(QueryBuilder $queryBuilder, DynamicSegmentFilterE
     if ($operator === 'equals') {
       $queryBuilder->andWhere("subscribers_custom_field.value = $valueParam");
       $queryBuilder->setParameter($valueParam, $value);
+    } elseif ($operator === 'not_equals') {
+      $queryBuilder->andWhere("subscribers_custom_field.value != $valueParam");
+      $queryBuilder->setParameter($valueParam, $value);
+    } elseif ($operator === 'more_than') {
+      $queryBuilder->andWhere("subscribers_custom_field.value >= $valueParam");
```
costasovo, line 131: Shouldn't we rather use only more than (>) instead of more than or equal (=>) here? I think a user can create more than or equal by combining two conditions (more than and equal). Alternatively, they could enter a lower value, but I would say it is not a good UX. What do you think?

costasovo (commented):
Looks good! I have just one question about the comparison.

costasovo (approved):

veljkho (changes requested):
I'm not getting any results when making a segment with `more than`, `less than` and/or `not equals`. Please refer to comments in Jira ticket for more details.

costasovo (approved):

veljkho (approved):

PR #3861, opened 2021-12-10
Add/multiple woo subscription segment [MAILPOET-3956]

costasovo:
[MAILPOET-3956]

[MAILPOET-3956]: https://mailpoet.atlassian.net/browse/MAILPOET-3956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:
1. Go to MailPoet > Lists
2. Click to add a New Segment
3. Select `has an active subscription segment` from the first dropdown
4. Follow the specs from the Jira ticket above and make sure everything is made according to Specification
5. After everything's tested and verified, try to disable `WooCommerce Subscription` plugin in My Plugins section and then to go back to MailPoet > Lists > Segments Tab... to check if you see this message "Activate the WooCommerce Subscriptions plugin to see the number of subscribers and enable the editing of this segment.". Because this message should be present for segments that contains "has an active subscription" segment type but the required plugin is not active. Please activate the subscription plugin at the end of testing.

### DIFF
--- a/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce_subscription.tsx
+++ b/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce_subscription.tsx
@@ -66,10 +66,6 @@ export const WooCommerceSubscriptionFields: React.FunctionComponent<Props> = ({ 
     ) {
       updateSegmentFilter({ operator: AnyValueTypes.ANY }, filterIndex);
     }
-    // Temporary BC fix
-    if (segment.product_id && !segment.product_ids) {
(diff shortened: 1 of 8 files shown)


brezocordero (dismissed):
All working as expected.
Good work.

veljkho (dismissed):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3862, opened 2021-12-13
Improve performance and allow guest users in WooCommerce dynamic segments [MAILPOET-3576]

lysyjan:
[MAILPOET-3576]

[MAILPOET-3576]: https://mailpoet.atlassian.net/browse/MAILPOET-3576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Test steps:
Info: With this pull request, we want to allow also non-registered users to be counted in the result when calculating segment. You need to make sure how many registered users are in `All Users` page, then to see how many Completed orders you have in WooCommerce > Orders.. and then to purchase few more with some random email. This way you'll know that you have enough orders that aren't purchased by registered users.
1. Go to All Users and see how many registered users are there
2. Go and purchase some items under Clothing category but as a guest user (log out and remember to click the checkbox in the checkout _"I would like to receive..."_), but make sure to enter email in the checkout such as `globalstep1@yopmail.com` (for the first order), `globalstep2@yopmail.com` (for the 2nd order)... and make 3 orders like that in a row.
3. Log in to admin account and go to WooCommerce > Orders page.. and mark as completed all those orders.
4. Go to MailPoet > Lists
5. Click to create New Segment
6. Select segment type `purchased in this category`
7. Select Clothing as the category
8. You should see a higher amount of subscribers compared to the amount of registered users (then it is verified, make a note of how many subscribers you see for your segment)
9. Let's make another test, save the segment like that and make another purchase with 4th email globalstep4@yopmail.com and mark it as completed.. then go to MailPoet > Lists > segments tab.. and click recalculate now button .. check if you see correct amount of subscribers for your created segment.. it should count also that new one subscriber.

Note for this test: If you want to make sure globalstep1..2.3.@yopmail.com are subscribed after purchasing an order, go to yopmail.com and just log in there with globalstep1, globalstep2..etc.. and click on the link inside confirmation email "Yes, let me in." in order to be successfully subscribed.

### DIFF
--- a/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
+++ b/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
@@ -5,6 +5,7 @@ namespace MailPoet\Segments\DynamicSegments\Filters;
 use MailPoet\Entities\DynamicSegmentFilterData;
 use MailPoet\Entities\DynamicSegmentFilterEntity;
 use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Util\DBCollationChecker;
 use MailPoet\Util\Security;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Doctrine\DBAL\Connection;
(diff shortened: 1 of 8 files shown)


pavel-mailpoet (dismissed):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

veljkho:
There are a lot of conflicted files, please resolve them @lysyjan . Thanks!

lysyjan:
@veljkho Done ✅

veljkho (approved):

PR #3867, opened 2021-12-15
Fix pQuery use statement [MAILPOET-4025]

lysyjan:
[MAILPOET-4025]

[MAILPOET-4025]: https://mailpoet.atlassian.net/browse/MAILPOET-4025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/Newsletter/Editor/StructureTransformer.php
+++ b/lib/Newsletter/Editor/StructureTransformer.php
@@ -3,8 +3,8 @@
 namespace MailPoet\Newsletter\Editor;
 
 use MailPoet\Util\DOM as DOMUtil;
-use MailPoet\Util\pQuery\DomNode;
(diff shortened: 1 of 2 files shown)


costasovo (approved):
Good find! I see that sometimes we can get TextNode and [that extends MailPoetVendor\pQuery\DomNode](https://github.com/mailpoet/mailpoet/blob/54a925f0642e11682a7329b6a04d361b4d6aa360/lib-3rd-party/pquery/gan_node_html.php#L2524) so we need to use MailPoetVendor\pQuery\DomNode here.

veljkho (approved):

PR #3869, opened 2021-12-15
Combine PHPStan configuration files into a single file [MAILPOET-4024]

rodrigoprimo:
[MAILPOET-4024]

[MAILPOET-4024]: https://mailpoet.atlassian.net/browse/MAILPOET-4024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/DataFactories/Form.php
```diff
@@ -89,9 +89,11 @@ public function withLastName() {
    * @return $this
    */
   private function addFormBlock(array $block) {
-    $body = unserialize($this->data['body']);
-    $body = array_merge([$block], $body);
-    $this->data['body'] = serialize($body);
+    if (is_string($this->data['body'])) {
```
rodrigoprimo, line 92: Noting here that I was not able to easily spot what is the difference between `phpstan-tests.neon` and `phpstan.neon` that made it necessary to update this file after combining both files (PHPStan reports the same errors for tests/DataFactories/Form.php when executed using the old version of `phpstan.neon`). Since the changes make sense, I thought it was not necessary to dig deep to understand what is causing this change.

pavel-mailpoet (approved):

veljkho (approved):

PR #3870, opened 2021-12-15
Fix import subscribers column header search [MAILPOET-3958]

johnolek:
[MAILPOET-3958](https://mailpoet.atlassian.net/browse/MAILPOET-3958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

Check the testing instructions inside Jira ticket.

This fixes a bug where the search at the top of columns on the subscriber import page weren't doing anything.

It also includes a fix for a [bug with the latest version of select2 and jQuery](https://github.com/select2/select2/issues/5993) where the search inputs weren't receiving focus when clicked.

Before:

https://user-images.githubusercontent.com/8656640/146259444-9aac664b-b094-4932-bdf8-3103963b6e16.mov

After:

https://user-images.githubusercontent.com/8656640/146259461-4e200002-74be-429c-b936-8630235a0436.mov

I believe this issue may have originally been caused by removing the `text` property from the data we supply to select2 (see 4045fba527a9acdafdcb08e346b80d0b85e2ed70), which means we're not complying with [select2's "required" data format](https://select2.org/data-sources/formats). Restoring those properties seems to have fixed the issue.

Sidenote: because I was testing this in the browser a lot, I ended up writing a little tempermonkey script so I didn't have to copy/paste my CSV data every time. Note that you still have to type something in the textarea to trigger the callback that activates the "Next step" button.

```javascript
// ==UserScript==
// @name         Fill Paste Method Input
// @namespace    http://tampermonkey.net/
// @version      0.1
// @description  try to take over the world!
// @author       You
// @match        http://localhost:8002/wp-admin/admin.php?page=mailpoet-import
// @icon         https://www.google.com/s2/favicons?domain=undefined.localhost
// @grant        none
// ==/UserScript==

(function() {
    'use strict';
    const pasteMethodOption = document.getElementById('import-paste-method');
    if (!pasteMethodOption) {
        return;
    }
    pasteMethodOption.click();
    const textarea = document.getElementById('paste_input');
    const content = "Email, First Name, Last Name\njohn@example.com, John, Olek";
    textarea.value = content;
    textarea.focus();
})();
```

### DIFF
--- a/assets/js/src/subscribers/importExport/import/step_data_manipulation/generate_column_selection.jsx
+++ b/assets/js/src/subscribers/importExport/import/step_data_manipulation/generate_column_selection.jsx
@@ -2,6 +2,36 @@ import jQuery from 'jquery';
 import MailPoet from 'mailpoet';
 
 export default () => {
+  // A custom matcher is required because we are using optgroups instead of single values.
+  // See https://select2.org/searching
+  const columnSearchMatcher = (params, data) => {
+    if (data.children === undefined) {
+      return null;
+    }
+    const searchTerm = (params.term ?? '').trim().toLowerCase();
+    if (searchTerm === '') {
+      return data;
+    }
+    // "children" are objects representing the individual options within an optgroup
+    // `name` is the label displayed to users
+    const matchedChildren = data.children.filter((child) => {
+      const label = child.name.toLowerCase();
+      return label.includes(searchTerm);
+    });
+    if (matchedChildren.length === 0) {
+      // Returning null prevent the entire optgroup from being displayed
+      return null;
+    }
+    // We can't change the original `data` object because select2 continues use it as the basis for
+    // future searches. In other words, if we simply modified `data`, it could prevent options
+    // that were previously filtered out from reappearing if the search term changes or gets
+    // cleared out.
+    const modifiedData = jQuery.extend(true, {}, data);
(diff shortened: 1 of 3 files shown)


assets/js/src/subscribers/importExport/import/step_data_manipulation/generate_column_selection.jsx
```diff
@@ -2,17 +2,12 @@ import jQuery from 'jquery';
...
 export default () => {
+  const select2Config = {
+    data: window.mailpoetColumnsSelect2,
+    width: '15em',
+  };
   jQuery('select.mailpoet_subscribers_column_data_match')
-    .select2({
-      data: window.mailpoetColumnsSelect2,
-      width: '15em',
-      templateResult(item) {
```
johnolek, line 9: FYI these callbacks are no longer needed because select2 automatically looks for a `text` property, which now exists again.

brezocordero (approved):
Good job and thanks for fixing the focus issue.
Works as expected and the code looks good.

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3871, opened 2021-12-16
Add Parsley validation to the manage form [MAILPOET-3928]

lysyjan:
[MAILPOET-3928]

[MAILPOET-3928]: https://mailpoet.atlassian.net/browse/MAILPOET-3928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/js/src/public.jsx
+++ b/assets/js/src/public.jsx
@@ -327,7 +327,12 @@ jQuery(($) => {
     $('.mailpoet_captcha_update').on('click', updateCaptcha);
 
     // Manage subscription form
-    $('.mailpoet-manage-subscription').on('submit', () => {
+    $('.mailpoet-manage-subscription').on('submit', (event) => {
+      if (!$(event.target).parsley().isValid()) {
+        event.preventDefault();
+        $(event.target).parsley().validate();
+        return;
+      }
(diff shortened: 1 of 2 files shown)


rodrigoprimo (commented):
Good job, @lysyjan. I can confirm that with the changes proposed here, the validation works on the "Manage subscription" page.

The only concern that I have is that @triple0t mentioned potential security issues. If there is any, I fear that client-side validation is not enough as it is usually pretty easy to bypass and we should validate and/or sanitize also on the server-side. I haven't checked the code deeply to verify potential security issues and whether or not it is worth implementing server-side validation. Generally speaking, I tend to err on the side of caution and assume it is worth doing it unless proven otherwise.

@triple0t, do you have any concrete security concerns in mind?

@lysyjan, what do you think?

lysyjan:
@rodrigoprimo Because a custom field can be without validation, I think the current solution is enough.

rodrigoprimo:
@triple0t, just a friendly reminder about the question above in case you missed it.

triple0t:
@rodrigoprimo @lysyjan I believe this is good to go.

rodrigoprimo (approved):

PR #3875, opened 2021-12-17
Ensure plugin activated before running tests

johnolek:
This PR ensures that the test environment's MailPoet plugin is activated before proceeding with any tests.

This might not be necessary, in which case I'm happy to close it out.


JanJakes:
Just linking a [related comment](https://wp.me/pcNwfB-7w#comment-439) here if anyone runs into this PR.

johnolek:
I'm marking this a draft for now because it's not clear at the moment where it would be best to include it. We may also want to discuss any other changes we want to make to the test suite before actually implementing anything.

PR #3878, opened 2021-12-20
Add check for unique name when creating segment [MAILPOET-3998]

johnolek:
This PR adds a check directly in [SegmentRepository](https://github.com/mailpoet/mailpoet/blob/90d479fdc7e25c42da3e15264c48f1f488ac8e2b/lib/Segments/SegmentsRepository.php) to verify that a given segment name is unique before attempting to persist a new segment to the database. It's important to have this check because the `mailpoet_segments` table has a [unique key constraint](https://github.com/mailpoet/mailpoet/blob/158d70dc7a43d5185c84db4493a66fa53cccd087/lib/Config/Migrator.php#L107) on the `name` column.

We were previously checking for uniqueness through the save controllers ([regular](https://github.com/mailpoet/mailpoet/blob/90d479fdc7e25c42da3e15264c48f1f488ac8e2b/lib/Segments/SegmentSaveController.php#L30) and [dynamic](https://github.com/mailpoet/mailpoet/blob/90d479fdc7e25c42da3e15264c48f1f488ac8e2b/lib/Segments/DynamicSegments/SegmentSaveController.php#L30)), which I believe are the only places where `createOrUpdate` is being called other than tests, so it's unlikely that this bug is something that could have been encountered in the wild.

I decided to refactor the handful of uniqueness checks we have in the codebase to all use the same code. This cuts down on a bit of duplication and allows us to eliminate some code in the controllers that can be handled through the repository instead. This may or may not be a good idea and I would very much like to get the reviewer's opinion on that.

I believe the biggest difference is that I changed the `InvalidArgumentException` to a `ConflictException` because it seemed more accurate [based on its description](https://github.com/mailpoet/mailpoet/blob/bf882e3c921a12a0f91f574e4d56969a1153a0c1/lib/exceptions.php#L91). I believe I updated all the places that were referencing the `InvalidArgumentException` but the reviewer may want to double check.

[MAILPOET-3998]

[MAILPOET-3998]: https://mailpoet.atlassian.net/browse/MAILPOET-3998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/API/JSON/v1/DynamicSegments.php
+++ b/lib/API/JSON/v1/DynamicSegments.php
@@ -2,12 +2,12 @@
 
 namespace MailPoet\API\JSON\v1;
 
-use InvalidArgumentException;
 use MailPoet\API\JSON\Endpoint as APIEndpoint;
 use MailPoet\API\JSON\Error;
 use MailPoet\API\JSON\Response;
 use MailPoet\API\JSON\ResponseBuilders\DynamicSegmentsResponseBuilder;
 use MailPoet\Config\AccessControl;
+use MailPoet\ConflictException;
 use MailPoet\Doctrine\Validator\ValidationException;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Listing\Handler;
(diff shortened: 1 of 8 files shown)


brezocordero (approved):
The refactor works well, the code is good.
 
Adding here the summary of our call:
- The case of updating a segment by the name without an ID can not happen via the UI, so it is not implemented on this PR
- The PR is instead a refactor that reduces duplication in the code
- The change in the Exception type doesn't modify any existing behavior or text presented to end users.

Tested by creating a new segment using an existing name and by duplicating the same segment twice.

veljkho (approved):

PR #3882, opened 2021-12-28
Create API endpoint to check if MailPoet welcome wizard is finished [MAILPOET-4042]

NeosinneR:
[MAILPOET-4042]

@costasovo I didn't add the new endpoint to public API docs because I don't think this is something other plugins should use. But I don't have a strong opinion on this, so if you think we should document it, I can do that.

For QA - this PR also includes a slight refactor for showing the welcome wizard, WooCommerce setup (when someone installs WC after MP is already set up), and MP2 migration step, so please make sure all are still working as they should.

[MAILPOET-4042]: https://mailpoet.atlassian.net/browse/MAILPOET-4042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/Readme.md
+++ b/doc/Readme.md
@@ -26,6 +26,7 @@ Class `\MailPoet\API\API` becomes available once MailPoet plugin is loaded by Wo
 * [Get Lists (getLists)](api_methods/GetLists.md)
 * [Get Subscriber (getSubscriber)](api_methods/GetSubscriber.md)
 * [Get Subscriber Fields (getSubscriberFields)](api_methods/GetSubscriberFields.md)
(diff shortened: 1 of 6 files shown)


costasovo (commented):
Thanks! Code looks good! As for documentation, I think the method might be handy also for other integrations. Basically for the similar same use case as ours. E.g. a subscription form integration could show warning that MailPoet is not fully configured which may cause confirmation emails would not be send.

veljkho (approved):

PR #3885, opened 2021-12-28
Improve segment form grid layout [MAILPOET-4039]

NeosinneR:
[MAILPOET-4039]

[MAILPOET-4039]: https://mailpoet.atlassian.net/browse/MAILPOET-4039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions are inside the Jira.

### DIFF
--- a/assets/css/src/components-plugin/_segments.scss
+++ b/assets/css/src/components-plugin/_segments.scss
@@ -5,9 +5,14 @@
   grid-column: span 2;
 }
 
-.mailpoet-segments-segments-section .mailpoet-grid-three-columns {
+.mailpoet-segments-grid {
+  grid-template-columns: 200px 1fr 1fr;
   padding-right: $grid-gap-large;
   position: relative;
+
+  @include respond-to(small-screen) {
+    grid-template-columns: 1fr;
+  }
 }
 
 .mailpoet-segments-description-section .mailpoet-form-textarea textarea {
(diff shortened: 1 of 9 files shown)


assets/css/src/components-plugin/_segments.scss
```diff
@@ -5,7 +5,8 @@
   grid-column: span 2;
 }
 
-.mailpoet-segments-segments-section .mailpoet-grid-three-columns {
+.mailpoet-segments-grid {
+  grid-template-columns: 200px 1fr 1fr;
```
costasovo, line 9: I noticed that this rule overrides the rule for mobile view defined for `mailpoet-grid-three-columns` and it results in worse UX on mobiles. There probably won't be many users who configure segments on mobile, but I guess it should be an easy fix.
Screenshot 2021-12-29 at 11 04 55

costasovo (changes requested):
Good job! I found one minor issue that probably would be nice to fix.

assets/css/src/components-plugin/_segments.scss
NeosinneR: Good catch! Fixed

costasovo (approved):
👍

rohankacharegs:
GlobalStep Verified

PR #3886, opened 2021-12-29
Add translations info and pot file [MAILPOET-4028]

costasovo:
[MAILPOET-4028]

[MAILPOET-4028]: https://mailpoet.atlassian.net/browse/MAILPOET-4028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -10,6 +10,9 @@ else
   ./do translations:pack
 fi
 
+echo '[BUILD] Generating translations .pot file'
+./do translations:build
+
 plugin_name='mailpoet'
 
 # Remove previous build.
--- a/mailpoet.php
+++ b/mailpoet.php
@@ -8,6 +8,8 @@
  * Author: MailPoet
  * Author URI: http://www.mailpoet.com
  * Requires at least: 5.3
+ * Text Domain: mailpoet
+ * Domain Path: /lang
  *
  * @package WordPress
  * @author MailPoet


lysyjan (approved):

build.sh
```diff
@@ -10,6 +10,9 @@ else
   ./do translations:pack
 fi
 
+echo '[BUILD] Generating translations .pot file'
+./do translations:build
+
```
JanJakes, line 15: @costasovo Why doesn't this revert [the original commit](https://github.com/mailpoet/mailpoet/pull/2824/commits/1da972239225c731587207703a3fa4a2f9c08626)? Seems like now we'll built this for all branches which may be unnecessary (and will slow down builds).

costasovo: @JanJakes I did it intensionally. We are about to remove the whole block for translations [in another PR](https://github.com/mailpoet/mailpoet/pull/3833/files#diff-4d2a8eefdf2a9783512a35da4dc7676a66404b6f3826a8af9aad038722da6823L3-L12). The `translations:build` is relatively fast (~2 seconds) and I think it would be convenient to have every build the same as the production so that we can send any of them to a customer (e.g.  on L2 support).

veljkho (approved):

PR #3887, opened 2021-12-29
Use "today" or "tomorrow" status when applicable [MAILPOET-3560]

johnolek:
[MAILPOET-3560]

On the Emails listing page, if the newsletter is scheduled to send
either today or tomorrow, the status column will now display "Today" or
"Tomorrow" respectively.

A user may still see the full date if desired by hovering their mouse
over the text, which will show the full date as a tooltip.

Testing Instructions:
1. Go to MailPoet > Emails
2. Create a new newsletter and schedule it for sending tomorrow (anytime tomorrow)
3. Create another new newsletter and schedule it for sending today (anytime but today)
4. Make sure you're on MailPoet > Emails page
5. Observe that you see Tomorrow and Today labels for those 2 scheduled newsletters
6. Hover over them and make sure you see dates as in the below video example

https://user-images.githubusercontent.com/8656640/147689762-20026445-5469-41d8-a349-31a3e8ac7152.mov

[MAILPOET-3560]: https://mailpoet.atlassian.net/browse/MAILPOET-3560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/common/listings/newsletter_status.tsx
+++ b/mailpoet/assets/js/src/common/listings/newsletter_status.tsx
@@ -2,9 +2,10 @@ import React from 'react';
 import MailPoet from 'mailpoet';
 import classNames from 'classnames';
 import {
-  isFuture, isPast, differenceInMinutes,
+  addDays, differenceInMinutes, isFuture, isPast,
 } from 'date-fns';
 import t from 'common/functions/t';
+import Tooltip from '../tooltip/tooltip';
 
 type NewsletterStatusProps = {
   scheduledFor?: Date;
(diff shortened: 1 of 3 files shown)


views/layout.html
```diff
@@ -123,6 +123,7 @@
 
   'close': __('Close'),
   'today': __('Today'),
+  'tomorrow': __('Tomorrow'),
```
johnolek, line 126: I couldn't find anywhere else we were using the string `'tomorrow`'. Did I miss it, or is this the first time we're using it?

If it is the first time, is this a good spot for it, and is there anything special I need to do to ensure it gets translated after the next release (if we're not already getting a translation from somewhere)?

wxa: > I couldn't find anywhere else we were using the string 'tomorrow'. Did I miss it, or is this the first time we're using it?
Looking at the `lang/mailpoet.pot` file which holds all strings for translation, I can't find it, so apparently it's the first time.

> If it is the first time, is this a good spot for it
If this string is used only on the newsletters page, then perhaps it's more suitable to put in in `views/newsletters.html`. 

> is there anything special I need to do to ensure it gets translated after the next release (if we're not already getting a translation from somewhere)?
No special actions needed: if the normal release procedure is followed, the string will be pushed for translators automatically before a release and translated within a week or so. For small strings we usually don't require translations in advance.

johnolek: Thanks for the detailed response! You definitely answered my questions about the translation process.

> If this string is used only on the newsletters page, then perhaps it's more suitable to put in in `views/newsletters.html`.

I was torn between the two places and ended up adding it here because it felt appropriate to include it next to `today`, and it seems like a generic enough word that it might be nice to simply have it in the top level layout file, but it would be trivial to move it there later if need be, so I'll move it to `views/newsletters.html` for now if you think that's the best way to go.

wxa (approved):
👍

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3889, opened 2022-01-03
Fix/marketing optin register [MAILPOET-4036]

costasovo:
I chose a bit different approach than suggested in the ticket.
The problem was that the WP function `register_block_type` didn't accept the JSON file in WP 5.7 and below. I replaced the function with `register_block_type_from_metadata` which is the one internally called in `register_block_type` in WP 5.8 but is available from WP 5.5.

In our plugin readme, we say we support WP 5.3, but the plugins we need for this integration require higher versions ([WooCommerce 6 requires WP 5.6](https://wordpress.org/plugins/woocommerce/) and [WooCommerce Blocks requires WP 5.8](https://wordpress.org/plugins/woo-gutenberg-products-block/)). So the integration shouldn't activate on an older WP than 5.5.

Note: When working on the fix I noticed a different issue with registering front-end assets. I fixed it in the second commit.

[MAILPOET-4036]


lysyjan (dismissed):
I like your approach. Good job!

veljkho (approved):

veljkho:
Please resolve conflicting files @costasovo so I can merge, thank you.

costasovo:
@veljkho Conflict resolved!

veljkho (approved):

PR #3890, opened 2022-01-06
Update block editor in the form editor [MAILPOET-4029][MAILPOET-3914]

costasovo:
[MAILPOET-4029] [MAILPOET-3914]

## Testing instructions
Info: This PR updates internals of the form editor.

1. Go to MailPoet > Forms
2. Please perform exploratory testing of the form editor, test complete functionalities.
3. There are 2 visible changes in UI/functionality that you should check also:

a. **New UI for color selection**
Screenshot 2022-01-06 at 16 42 15

b. **It is possible to set padding also to columns wrapper.**
Screenshot 2022-01-06 at 16 42 49

[MAILPOET-4029]: https://mailpoet.atlassian.net/browse/MAILPOET-4029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-3914]: https://mailpoet.atlassian.net/browse/MAILPOET-3914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pavel-mailpoet:
Perfect, thanks for the upgrade @costasovo the users will love the new padding feature.

pavel-mailpoet (dismissed):

rohankacharegs:
GlobalStep Verified

veljkho:
I think there's issue inside the form editor:
1. When you highlight the text with some color, it will show you in [black color](https://d.pr/i/dcqJCL) all the time. On the front-end, it will be in white color without background color (in case you added some bg color). Talking about highlighting text not coloring the paragraph block. There's some conflict between paragraph font-color/bg-color and highlight the text/adding highlighted background.
2. Paragraph's [default font-size inside editor](https://d.pr/i/G0cqaN) is different from [what's on the front-end](https://d.pr/i/2XNk0A). Looks way bigger font-size on the front-end. But, when user set some font-size inside editor it will be correctly reflected on the front-end, so this is just in case the user use `reset all` function or what it is about as default's font-size.
3. For some of the titles, it is all uppercased like in this [screenshot](https://d.pr/i/TO3T3a). Is this by a new design?

costasovo:
@pavel-mailpoet, @veljkho  I added two new commits to address issues 1) and 3) reported by @veljkho.

as for issue 1)  Good catch! It was a new feature that I missed when I was testing blocks after the update. I fixed it by adding support for the default color palette used in other color pickers. I think that in the future, we should improve it by adding support for [theme.json](https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-json/) so that the editor would use the color set defined by the theme (possibly with a fallback to the default palette for themes that don't support theme.json). Adding support for theme.json is a bit more complex and requires some additional research, but I think it would be nice to have. If you agree, I could prepare a Jira ticket. The editor is not a priority, so it will probably take some time until we get to it.

issue 2) I think it is not new behavior, but it is also present in the current version. Our philosophy was if some style isn't set, don't set anything, and let's use theme styles. The thing is that in the WP admin in the editor, we don't know what the font size on the front end is. Can you confirm that it is the same also in the current version of the editor? (Maybe I misunderstood something). 

issue 3) Good catch! There were some CSS changes in the post editor package. It should be fixed.

pavel-mailpoet (dismissed):

veljkho (changes requested):
1. Agree. Now it is partially fixed: When I highlight the text in paragraph/heading blocks with **white** color, it changes to black: https://d.pr/i/NJ23o6 Issue not present with other colors! (Only for highlighting with white - turns to black)
2. It is true. Issue also present in the production.
3. Verified.

veljkho (changes requested):
Unfortunately, we have again some issues:
Applying highlighted text in the form editor is working, but not reflected on the front-end.
This is happening for both applying text color and background color. Form editor is good.

Form editor [screenshot](https://d.pr/i/XHnTSw) vs Front-end [screenshot](https://d.pr/i/mCC6qF)

costasovo:
> Unfortunately, we have again some issues: Applying highlighted text in the form editor is working, but not reflected on the front-end. This is happening for both applying text color and background color. Form editor is good.
> 
> Form editor [screenshot](https://d.pr/i/XHnTSw) vs Front-end [screenshot](https://d.pr/i/mCC6qF)

@veljkho Good catch again! I tested it in form editor preview and I missed that our security data sanitization strips the tag used for the highlight when we save the form. Fixed!

pavel-mailpoet (approved):

veljkho (approved):

PR #3895, opened 2022-01-11
Fix a test that is broken on multisite

pavel-mailpoet:
The test is failing because of an error in the UI:
Usernames can only contain lowercase letters (a-z) and numbers.


lysyjan (approved):
There was one more wrong username. Good job!

PR #3902, opened 2022-01-11
Count non-customers in "none of" option of purchased product/in category segments [MAILPOET-4021]

brezocordero:
With this PR, when 'none' is selected, subscribers that never placed an order are also included on segments Purchased Products or Purchased in Category.

To test:
1. Go to MailPoet > Lists > Segment tab
2. Create a new segment
3. Select "purchased product", "none of" and a product (select one from your wish)
4. Save the segment
5. Create another segment
6. Select "purchased in category", "none of" and a category (any but remember which one you selected)
7. Save the segment
8. Now you have 2 segments that counts all subscribers who didn't purchase specific product or in specific category
9. Go and purchase some product (the one you selected) and make sure the subscriber is not counted in the segment list
10. Go and purchase some product (the one in category you selected) and make sure the subscriber isn't counted in the segment list
11. You could create also a segment with "purchased product" and "any of" type selecting the product which is in one of your lists, that will make sure that purchasing such product would add subscriber into that list and reducing it from one of your lists (that you previously created above)

Explanation: This test should ensure that segmentation can divide subscribers who have purchased specific products from those who never purchased it. Make sure to always press `Recalculate Now` button when you want to see the real standing of segments lists, e.g. after you place an order as a guest, registered user.. Lists need recalculation in order to show real results.

Note for developers: Integration tests for both segments:
```
./do --test "test:integration --file=tests/integration/Segments/DynamicSegments/Filters/WooCommercePurchasedTest.php"
./do --test "test:integration --file=tests/integration/Segments/DynamicSegments/Filters/WooCommerceCategoryTest.php"
```

[MAILPOET-4021]

[MAILPOET-4021]: https://mailpoet.atlassian.net/browse/MAILPOET-4021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
@@ -66,8 +66,6 @@ class WooCommerceCategory implements Filter {
         ->setParameter('count_' . $parameterSuffix, count($categoryIds));
 
     } elseif ($operator === DynamicSegmentFilterData::OPERATOR_NONE) {
-      $this->applyCustomerJoin($queryBuilder, $subscribersTable);
-      $this->applyOrderStatsJoin($queryBuilder);
       // subQuery with subscriber ids that bought products
       $subQuery = $this->createQueryBuilder($subscribersTable);
       $subQuery->select("DISTINCT $subscribersTable.id");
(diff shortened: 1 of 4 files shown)


lysyjan (dismissed):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3804, opened 2021-11-15
Update key display - [MAILPOET-1950]

triple0t:
[MAILPOET-1950]

[MAILPOET-1950]: https://mailpoet.atlassian.net/browse/MAILPOET-1950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3910, opened 2022-01-13
Mask the API key when displayed in the info page

rodrigoprimo:
This PR reverts the commits from #3804 and implements logic to mask half of the API key when it is displayed on the System Info page. There is also a bonus commit replacing the old Subscriber model used in an updated integration tests file with the Doctrine equivalent.

[MAILPOET-4009]

[MAILPOET-4009]: https://mailpoet.atlassian.net/browse/MAILPOET-4009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:
1. Go to MailPoet > Settings
2. Click on Key Activation tab
3. Verify the key is visible in full (without ***)
4. Then, go to MailPoet > Help
5. Click on System Info tab
6. Verify you see your key in half visible and half *** such as: `MailPoet Premium/MSS key: 12345678910111213****************`

### DIFF
--- a/mailpoet/assets/js/src/settings/pages/key_activation/key_activation.tsx
+++ b/mailpoet/assets/js/src/settings/pages/key_activation/key_activation.tsx
@@ -136,10 +136,10 @@ export default function KeyActivation() {
       />
       <Inputs>
         <Input
-          type="password"
+          type="text"
           id="mailpoet_premium_key"
           name="premium[premium_key]"
-          value={state.key ? 'premium_key' : ''}
+          value={state.key || ''}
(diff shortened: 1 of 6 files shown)


brezocordero (approved):
Works as expected.
Thanks for the bonus refactor!

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3911, opened 2022-01-13
Prevent tests from emailing [randomhex]@example.com [MAILPOET-4064]

johnolek:
Previously this test was using random email addresses in the form of
`bin2hex(random_bytes(7)) . '@example.com'`, the default address
generated by `\MailPoet\Test\DataFactories\Subscriber`. This was causing
the staff@mailpoet.com account to get banned by the MSS due to
`bad-domain-discard`s. Using the mailinator email address instead, like
the other sending tests, should prevent this from occurring.

[MAILPOET-4064]

[MAILPOET-4064]: https://mailpoet.atlassian.net/browse/MAILPOET-4064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/integration/Mailer/MailerTest.php
+++ b/mailpoet/tests/integration/Mailer/MailerTest.php
@@ -64,7 +64,7 @@ class MailerTest extends \MailPoetTest {
         getenv('WP_TEST_MAILER_MAILPOET_API') :
         '1234567890',
     ];
-    $this->subscriber = 'Recipient <mailpoet-phoenix-test@mailinator.com>';
+    $this->subscriber = 'Recipient <blackhole@mailpoet.com>';
     $this->newsletter = [
       'subject' => 'testing Mailer',
       'body' => [
@@ -209,7 +209,7 @@ class MailerTest extends \MailPoetTest {
 
     $subscriberFactory = new SubscriberFactory();
     $subscriber = $subscriberFactory
-      ->withEmail('mailpoet-phoenix-test@mailinator.com')
+      ->withEmail('blackhole@mailpoet.com')
       ->withFirstName('Recipient')
       ->create();
     $this->sender['address'] = 'staff@mailpoet.com';
--- a/mailpoet/tests/integration/Mailer/MailerTest.php
+++ b/mailpoet/tests/integration/Mailer/MailerTest.php
@@ -208,7 +208,10 @@ class MailerTest extends \MailPoetTest {
     }
 
     $subscriberFactory = new SubscriberFactory();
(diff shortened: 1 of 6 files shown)


tests/integration/Mailer/MailerTest.php
```diff
@@ -208,7 +208,10 @@ public function testItCanSendWhenSubscriberEntityIsPassed() {
     }
 
     $subscriberFactory = new SubscriberFactory();
-    $subscriber = $subscriberFactory->withFirstName('Recipient')->create();
+    $subscriber = $subscriberFactory
+      ->withEmail('mailpoet-phoenix-test@mailinator.com')
```
johnolek, line 212: I used mailpoet-phoenix-test@mailinator.com instead of blackhole@mailpoet.com because the other sending tests use the former, and those are all getting delivered fine. If we don't care about deliverability then we might want to change this to `blackhole@mailpoet.com` in both places. What do you think?

lysyjan: I would prefer our blackhole email address because we don't care about deliverability in this case.

johnolek: I just pushed an update that uses the blackhole@mailpoet.com address for both this test and the other sending test in the same file.

Are we actively checking emails sent to mailinator anywhere, or were we just using it because we didn't have the blackhole address set up yet? I'm seeing the mailpoet-phoenix-test@mailinator.com address used in the specific mailer method tests (`AmazonSESTest`, `MailPoetAPITest`, `PHPMailTest`, `SendGridTest`, and `SMTPTest`), and I'm wondering if it would be okay to update those to use the blackhole as well. At first glance it looks like that would be a safe change to make.

lysyjan: @johnolek I think that we can replace the `mailpoet-phoenix-test@mailinator.com` address. The blackhole email is supported from July 2021.

johnolek: I believe I've now replaced the mailinator address with the blackhole address everywhere @lysyjan

lysyjan (dismissed):

veljkho (approved):

veljkho:
Hi @johnolek 
Please rebase this as it contains conflicting files, I want to merge it. Thanks!

johnolek:
@veljkho I've rebased and I think this should be good to go. Please let me know if you need anything else.

veljkho (approved):

PR #3912, opened 2022-01-14
Use CircleCI's built-in parallelism [MAILPOET-3994]

johnolek:
This PR alters our test splitting behavior to use Circle CI's [built in parallelism](https://circleci.com/docs/2.0/parallelism-faster-jobs/#using-the-circleci-cli-to-split-tests) instead of manually splitting files.

Circle CI has the ability to split test files based on timing data from previous runs, creating `n` groups of files, where `n` is the value of the `parallelism`  parameter, and where each group should take approximately the same amount of time to run.

Another nice aspect is that the pipelines no longer need separate jobs for the different groups -- the parallel jobs are all contained in one parent job.

Before:
image

After:
image
image
image

Previously we had `16` separate jobs. I decided to go with `20` because that seems to result in a runtime of around 4 minutes, which is also about how long it takes to run the integration tests, which run at the same time. We could potentially speed up that entire step even more by parallelizing the integration tests as well and then increasing the `parallelism` value for the acceptance suite, but perhaps there's a good reason why we haven't split those tests yet? 

We would also experience diminishing returns with more parallelization because it usually takes 40-50 seconds for the "Spin up environment" and "Pull acceptance test docker images" steps, which both have to run for each parallel job.

I'm leaving this as a draft because I'm not sure if this is the way we want to go, and I'm not sure how best to test this with multisite to make sure it doesn't break anything. If this does look like a direction we would like to go, I could remove the code we were previously using to generate grouped acceptance tests if that sounds appropriate.

The main downside to this approach is that CircleCI bizarrely doesn't give you a way of re-running just one parallel job, so if there's one flaky test you have to re-run everything. There's already a [feature request](https://circleci.canny.io/cloud-feature-requests/p/rerun-failed-parallel-runs) to fix this.

[MAILPOET-3994]: https://mailpoet.atlassian.net/browse/MAILPOET-3994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pavel-mailpoet:
Hey, @johnolek I think it looks great. Next time I would recommend asking what other developers think about the approach. Either in slack or in a P2 post. 

I will approve this pull request but we should remove the [command](https://github.com/mailpoet/mailpoet/blob/09c2ce653c70141ee1b5bb2ac99e8436fcdf9a00/RoboFile.php#L648-L654) for split and the [whole tool](https://github.com/mailpoet/mailpoet/tree/21839fc30b3fcdc57c9140998b33c325ecd7d33d/tests/SplitTests)

johnolek:
> Next time I would recommend asking what other developers think about the approach. Either in slack or in a P2 post.

That makes a lot of sense and I definitely will do that next time. Thank you!

> I will approve this pull request but we should remove the command for split and the whole tool

I believe I've now removed everything that needs removing. Just let me know if I missed anything. Thanks for the review, @pavel-mailpoet!

pavel-mailpoet (approved):
Good Job @johnolek 

What do you think about the original JIRA ticket? Do you think it is still worth the effort splitting the long acceptance tests into smaller ones?

veljkho (approved):

johnolek:
> Do you think it is still worth the effort splitting the long acceptance tests into smaller ones?

@pavel-mailpoet  My initial instinct was to say no because I don't think that any of the longer tests or test files are currently acting as a bottleneck, especially since acceptance tests are now taking approximately the same amount of time as integration tests. However, if we wanted to improve run times even more, we could explore using a similar approach to parallelize the integration tests and then increase the parallelism parameter for acceptance tests.

Taking that approach would mean that the bottleneck would always be the one file that takes the longest to run. In our case it looks like that's `WooCommerceDynamicSegmentsCest`, which currently gets its own parallel job because it takes so long.

image

I would like to at least split up that file and maybe one or two of the other longest-running files in order to prevent them from being bottlenecks in the future.

PR #3913, opened 2022-01-14
Multiple clicked links in clicked segment filter and multiple emails in machine-opened filter [MAILPOET-3951] [MAILPOET-3957]

costasovo:
I did a small refactor while implementing the clicked links filter in which I splat code for click actions and open actions. After the refactor, I found that it would be much easier to add multiple emails to the machine-opened action filter than maintain the old behavior.

## Testing instructions
This PR contains a refactor that also touched `opened email` filters. So please also test this filter type. This change also includes data migration, so **you need to deactivate the plugin and re-activate it after uploading testing build** to trigger the migration. In the production build, it triggers automatically.

## Steps
1. Upload & install the MailPoet plugin version 3.75.1 ([click to download](https://downloads.wordpress.org/plugin/mailpoet.3.75.1.zip))
2. Go to MailPoet > Lists and create 7 separate segments with those filters:
`opened email`,
`not opened email`,
`machine-opened email`,
`clicked link (without selecting a link)`,
`clicked link (with a selected link)`, 
`not clicked link (without selecting a link)`,
`not clicked link (with a selected link)`
Tip: You can write in the segment description the total amount of subscribers found in the result, so you can compare later when updating the plugin to this build's version (migration process)
3. Update the plugin to this PR's build, download it from below as usual (remember to run the migration after updating by deactivating and re-activating the plugin (see ☝️))
4. Verify that all segments created in 3.75.1 were migrated correctly and that they show correct subscribers in the results (you can edit every segment to see if the calculation is correct from what you see when on the listing page)
6. Verify that the "machine-opened" segment filter works with multiple emails selected through `any of`/`all of` operators
7. Verify that the "clicked" segment filter works with multiple links selected through `any of`/`all of`/`none of` operators
8. Verify that the "opened" segment filter works with multiple emails selected through `any of`/`all of`/`none of` operators

[MAILPOET-3951](https://mailpoet.atlassian.net/browse/MAILPOET-3951)
[MAILPOET-3957](https://mailpoet.atlassian.net/browse/MAILPOET-3957)


assets/js/src/segments/dynamic/dynamic_segments_filters/email_statistics_opens.tsx
```diff
@@ -42,12 +43,14 @@ export const EmailOpenStatisticsFields: React.FunctionComponent<Props> = ({ filt
   useEffect(() => {
     if ((segment.operator !== AnyValueTypes.ANY)
         && (segment.operator !== AnyValueTypes.ALL)
-        && (segment.operator !== AnyValueTypes.NONE)
+        && ((segment.operator !== AnyValueTypes.NONE)
+          || (segment.action === EmailActionTypes.MACHINE_OPENED)
```
pavel-mailpoet, line 47: This condition seems a bit odd to me. Are you sure it is correct? Could you please leave a code comment on why do we need it this way? Or maybe even refactor it to a separate method and make it more readable...

costasovo: This is needed because machine-open doesn't support `none of`. But a good point it is not readable. I'll refactor it.

pavel-mailpoet (approved):

costasovo:
@pavel-mailpoet After rebasing on the new repo structure, I noticed an error on client-side validation for the email actions segment filter. I added one commit with a fix and also one commit with migration check update because the PR missed today's release.

pavel-mailpoet (approved):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3917, opened 2022-01-14
Add a check of production vendor packages for PHP 7 compatibility

brezocordero:
This PR includes a PHP lint job for `vendor` and `vendor-prefixed` folders that are included in the final build.

From the links below, WordPress.org plugin directory pre-commit hook lints on PHP7.4
Sources:
https://meta.trac.wordpress.org/ticket/3791#comment:36
https://wordpress.slack.com/archives/C02QB8GMM/p1619603847340700

The added job uses MailPoet WordPress 7.4 image. 
The lint function includes the files/folders that are present on the final build.
I have used [build.sh](https://github.com/mailpoet/mailpoet/blob/master/build.sh) as reference to exclude files/folders from the vendor-prefixed linted files.

[MAILPOET-3905]

[MAILPOET-3905]: https://mailpoet.atlassian.net/browse/MAILPOET-3905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,6 +84,11 @@ executors:
     docker:
     - image: mailpoet/wordpress:7.2_20211213.1
 
+  wpcli_php_max_wporg:
+    <<: *default_job_config
+    docker:
+      - image: mailpoet/wordpress:7.4_20210122.1
+
   wpcli_php_latest:
     <<: *default_job_config
     docker:
@@ -255,6 +260,14 @@ jobs:
   qa_php_oldest:
     executor: wpcli_php_oldest
     working_directory: /home/circleci/mailpoet/mailpoet
+    steps:
+      - attach_workspace:
+          at: /home/circleci/mailpoet
+      - run:
+          name: "QA PHP"
+          command: ./do qa:php
+  qa_php_max_wporg:
+    executor: wpcli_php_max_wporg
     steps:
       - attach_workspace:
           at: /home/circleci/mailpoet
@@ -485,6 +498,10 @@ workflows:
           requires:
             - build
       - qa_php_oldest:
+          <<: *slack-fail-post-step
+          requires:
+            - build
+      - qa_php_max_wporg:
           <<: *slack-fail-post-step
           requires:
             - build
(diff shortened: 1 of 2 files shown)


RoboFile.php
```diff
@@ -393,7 +439,7 @@ public function qaCodeSniffer(array $filesToCheck, $opts = ['severity' => 'all']
       'tests/_output',
       'tests/_support/_generated',
       'vendor',
-      'vendor-prefixed',
+      'vendor-prefixed/gregwar/captcha/demo',
```
wxa, line 442: Looks like this change is incorrect, we don't need to sniff the whole vendor-prefixed folder, it should still be ignored.

RoboFile.php
```diff
@@ -358,6 +364,46 @@ public function qaLint() {
...
+    $filesToCheckString = implode(' ', [
+      'lib/',
+      'lib-3rd-party/',
+      'vendor/composer',
+      'vendor/mtdowling',
+      'vendor/soundasleep',
+      'vendor-prefixed/',
+      'mailpoet.php',
+    ]);
+    $filesToExcludeString = '--exclude ' . implode(' --exclude ', [
```
wxa, line 378: Please add a code comment here that the list is taken from `build.sh`, so there will be less confusion if it needs to be updated.

wxa (changes requested):
Awesome job overall! It was able to catch the Doctrine error which was the cause of this task creation:

```
Syntax error found in 1 file

------------------------------------------------------------
Parse error: vendor-prefixed/doctrine/cache/lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php:24
    22|      */
    23|     public function __construct(
  > 24|         private string $key,
    25|         private mixed $value,
    26|         private bool $isHit,
Unexpected 'private' (T_PRIVATE), expecting variable (T_VARIABLE) in vendor-prefixed/doctrine/cache/lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php on line 24
```

Just a couple small nitpicks.

RoboFile.php
brezocordero: Done!

wxa (approved):

veljkho (approved):

PR #3919, opened 2022-01-18
Add Woo memberships segment [MAILPOET-3880]

wxa:
[MAILPOET-3880]

[MAILPOET-3880]: https://mailpoet.atlassian.net/browse/MAILPOET-3880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:
1. Go to Products and create a new product called `1-month membership` (add some $ value)
2. Go to WooCommerce > Memberships
3. Create one more membership plan and name it whatever (can be named `1-month membership` as well)
4. Make sure to add the length `1 month` and also to include created product `1-month membership` from the first step
5. Now that you have `1-month membership` plan and product created, go and purchase it with a few accounts (can be even a guest order)
6. Purchase also the one already created called `1-week membership` with some account
7. Now that you have purchased at least 2-3 orders, make sure the orders are marked as `completed` at WooCommerce > Orders page
8. Now you have data ready-made for testing, go to MailPoet > Lists
9. Create a new segment and verify it according to the specification from the Jira ticket above
10. Make sure that after saving such segment that you have some subscribers who have bought membership products are counted in the results (by clicking `view subscribers` link for your segment, click `recalculate now` button if you don't see them)

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -909,6 +909,15 @@ class RoboFile extends \Robo\Tasks {
     $this->say("Release '$version[name]' info was published on Slack.");
   }
 
+  public function downloadWooCommerceMembershipsZip($tag = null) {
+    if (!getenv('WP_GITHUB_USERNAME') && !getenv('WP_GITHUB_TOKEN')) {
+      $this->yell("Skipping download of WooCommerce Memberships", 40, 'red');
+      exit(0); // Exit with 0 since it is a valid state for some environments
+    }
+    $this->createGithubClient('woocommerce/woocommerce-memberships')
+      ->downloadReleaseZip('woocommerce-memberships.zip', __DIR__ . '/tests/plugins/', $tag);
+  }
+
   public function downloadWooCommerceSubscriptionsZip($tag = null) {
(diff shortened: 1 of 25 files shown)


lib/Models/Segment.php
```diff
@@ -166,6 +167,24 @@ public static function getWooCommerceSegment() {
     return $wcSegment;
   }
 
+  public static function getWooMembershipsSegment() {
```
brezocordero, line 170: Do we need the changes on this file? They don't seem to be used.

brezocordero (changes requested):
Great job @wxa !
* Code looks good
* The membership status check corresponds to this doc:
https://woocommerce.com/document/woocommerce-memberships-user-memberships/
* I have tested it and worked as expected.

Only one question to address.

lib/Models/Segment.php
wxa: Good catch! This is a leftover, I removed it.

brezocordero (approved):

veljkho (approved):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3923, opened 2022-01-19
Offer to download or activate the premium plugin - MAILPOET-3971

triple0t:
Offer to download or activate the premium plugin or upgrade plan in premium banner

[MAILPOET-3971](https://mailpoet.atlassian.net/browse/MAILPOET-3971)

Related to https://github.com/mailpoet/shop/pull/669

Testing instructions:
1. Make sure to downgrade Premium plugin to be 3.77.0 in order to test this pull request as this PR is on 3.77.0. Please head over [to this page to download](https://github.com/mailpoet/mailpoet-premium/releases) and upload 3.77.0 build. After that follow the next steps.
2. Go to MailPoet > Emails page
3. Click on statistics link for any of the sent newsletters to access the statistics page
4. On this page you could see the banner that is mentioned in the Jira ticket above. Please follow all three cases from the [Jira specification](https://mailpoet.atlassian.net/browse/MAILPOET-3971) to verify that the banner is present with the appropriate content is in the banner.
5. Three cases to test: With inactive Premium plugin; With not installed Premium plugin; With applied lower-tier subscription (go to call for testing p2 post to find the key to be applied before testing (under IMPORTANT NOTES BEFORE TESTING & SUBSCRIPTION KEYS section));

### DIFF
--- a/mailpoet/assets/js/src/common/premium_banner_with_upgrade/premium_banner_with_upgrade.tsx
+++ b/mailpoet/assets/js/src/common/premium_banner_with_upgrade/premium_banner_with_upgrade.tsx
@@ -1,0 +1,93 @@
+import React from 'react';
+import MailPoet from 'mailpoet';
+import PremiumRequired from 'common/premium_required/premium_required';
+import Button from 'common/button/button';
+import ReactStringReplace from 'react-string-replace';
+
+type Props = {
+  message: React.ReactNode;
+  actionButton: React.ReactNode;
+};
+
+interface BannerWindow extends Window {
+  mailpoet_has_valid_api_key: boolean;
+  mailpoet_has_valid_premium_key: boolean;
+  mailpoet_subscribers_count: number;
+  mailpoet_subscribers_limit: number | boolean;
+  mailpoet_subscribers_limit_reached: boolean;
+  mailpoet_premium_active: boolean;
+  mailpoet_premium_plugin_installed: boolean;
+  mailpoet_premium_plugin_download_url: string;
+  mailpoet_premium_plugin_activation_url: string;
+}
+
+declare let window: BannerWindow;
+
+const limitReached = window.mailpoet_subscribers_limit_reached;
+const limitValue = window.mailpoet_subscribers_limit;
+const subscribersCountTowardsLimit = window.mailpoet_subscribers_count;
+const premiumActive = window.mailpoet_premium_active;
+const hasValidApiKey = window.mailpoet_has_valid_api_key || window.mailpoet_has_valid_premium_key;
+const downloadUrl = window.mailpoet_premium_plugin_download_url;
+const activationUrl = window.mailpoet_premium_plugin_activation_url;
+const isPremiumPluginInstalled = window.mailpoet_premium_plugin_installed;
+
+const getBannerMessage = (translationKey) => {
+  const message = MailPoet.I18n.t(translationKey);
+  return (
+    <p>
+      {ReactStringReplace(
+        message,
+        /(\[subscribersCount]|\[subscribersLimit])/g,
+        (match) => ((match === '[subscribersCount]') ? subscribersCountTowardsLimit : limitValue)
+      )}
+    </p>
+  );
+};
+
+const getCtaButton = (translationKey, link = null, target = '_blank') => (
+  <Button
+    href={link}
+    target={target}
+    rel="noopener noreferrer"
+  >
+    {MailPoet.I18n.t(translationKey)}
+  </Button>
+);
+
+const PremiumBannerWithUpgrade: React.FunctionComponent<Props> = (
+  { message, actionButton }: Props
+) => {
+  let bannerMessage: React.ReactNode;
+  let ctaButton: React.ReactNode;
+
+  if (hasValidApiKey && !premiumActive) {
+    bannerMessage = getBannerMessage('premiumFeatureDescription');
+
+    ctaButton = isPremiumPluginInstalled
+      ? getCtaButton('premiumFeatureButtonActivatePremium', activationUrl, '_self')
+      : getCtaButton('premiumFeatureButtonDownloadPremium', downloadUrl);
+  } else if (limitReached) {
+    bannerMessage = getBannerMessage('premiumFeatureDescriptionSubscribersLimitReached');
+
+    const link = hasValidApiKey
+      ? MailPoet.MailPoetComUrlFactory.getUpgradeUrl()
+      : MailPoet.MailPoetComUrlFactory.getPurchasePlanUrl(subscribersCountTowardsLimit + 1);
+
+    ctaButton = getCtaButton('premiumFeatureButtonUpgradePlan', link);
+  } else {
+    // use the provided information
+    bannerMessage = message;
+    ctaButton = actionButton;
+  }
+
+  return (
+    <PremiumRequired
+      title={MailPoet.I18n.t('premiumFeature')}
+      message={bannerMessage}
+      actionButton={ctaButton}
+    />
+  );
+};
+
+export default PremiumBannerWithUpgrade;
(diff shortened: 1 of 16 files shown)


mailpoet/views/newsletters.html
```diff
@@ -62,6 +63,10 @@
       var mailpoet_newsletters_count = <%= newsletters_count %>;
       var mailpoet_send_transactional_emails = <%= json_encode(settings.send_transactional_emails == '1') %>;
       var mailpoet_settings = <%= json_encode(settings) %>;
+      var mailpoet_has_premium_support = <%= has_premium_support ? 'true' : 'false' %>;
```
pavel-mailpoet, line 66: I can see you duplicate all of those variables and translations in a couple of templates. Would it make sense to move them to layout and have them available everywhere?

mailpoet/lib/Config/ServicesChecker.php
```diff
@@ -161,4 +165,12 @@ public function getAnyValidKey(): ?string {
     }
     return null;
   }
+
+  public function generateHalfApiKey() {
```
pavel-mailpoet, line 169: I am not sure if the name is ideal. The fact that the key is split in the middle is only an implementation detail. I think the important part here is that the key is partial, would the better name be `generatePartialApiKey`?

Also, why don't you use type hints here?

mailpoet/lib/Config/ServicesChecker.php
```diff
@@ -25,6 +25,10 @@ public function __construct() {
     $this->subscribersFeature = ContainerWrapper::getInstance()->get(SubscribersFeature::class);
   }
 
+  public function isPremiumPluginActive() {
```
pavel-mailpoet, line 28: I am not sure about this method. It is called `isPremiumPluginActive` but it doesn't check if the premium plugin is active, it checks the license. So would it make sense to use a different name? Why do we even need this method? I think we already check about the existence of the premium plugin, don't we?

mailpoet/lib/Config/Installer.php
```diff
@@ -27,6 +28,14 @@ public function init() {
     WPFunctions::get()->addFilter('plugins_api', [$this, 'getPluginInformation'], 10, 3);
   }
 
+  public function generatePluginActivationUrl(string $plugin): string {
+    return WPFunctions::get()->adminUrl('plugins.php?' . implode('&', [
+      'action=activate',
+      'plugin=' . urlencode($plugin),
+      '_wpnonce=' . wp_create_nonce('activate-plugin_' . $plugin),
```
pavel-mailpoet, line 35: When calling the WordPress functions we always do it using the `WPFunctions` proxy class. Could you please create a method in there called `wpCreateNonce` and use it instead of directly calling the function here?

mailpoet/assets/js/src/common/premium_banner_with_upgrade/premium_banner_with_upgrade.tsx
```diff
@@ -0,0 +1,95 @@
...
+  mailpoet_subscribers_limit: number | boolean;
+  mailpoet_subscribers_limit_reached: boolean;
+  mailpoet_premium_active: boolean;
+  mailpoet_premium_plugin_installed: boolean;
+  mailpoet_premium_plugin_download_url: string;
+  mailpoet_premium_plugin_activation_url: string;
+  mailpoet_plugin_half_key: string;
+}
+
+declare let window: BannerWindow;
```
pavel-mailpoet, line 25: We have this task: https://mailpoet.atlassian.net/browse/MAILPOET-2943

I am not sure if it is outdated or still valid and the team still wants to go in that direction. If we don't want to add the window usage to the new code so that we don't have to remove it later. Would it make sense to move these properties to the `MailPoet` object and use it from there?

mailpoet/lib/Config/ServicesChecker.php
```diff
@@ -167,7 +167,7 @@ public function getAnyValidKey(): ?string {
   }
 
   public function generateHalfApiKey() {
-    $key = $this->getAnyValidKey();
+    $key = (string)($this->getAnyValidKey());
```
pavel-mailpoet, line 170: I wonder if amending a previous commit instead of creating a new one for the fix would be better? What do you think?

pavel-mailpoet (commented):
Looks good, I have a couple of comments, questions.

mailpoet/views/newsletters.html
triple0t: Yes, Thank you for this idea @pavel-mailpoet 👏🏾.
I had been looking for an easy and maintainable way to share the variables but I totally missed this line `<% extends 'layout.html' %>`.
For security reasons, I don't think I should include all of these variables in the layout.html file where it may/may not be publicly accessible (outside wp-admin)
So, I will create a new view file and extend that

mailpoet/lib/Config/ServicesChecker.php
triple0t: @pavel-mailpoet Noted 👍.

mailpoet/lib/Config/ServicesChecker.php
triple0t: @pavel-mailpoet  From what I could see in the codebase, the way we are checking if the Premium plugin is active is by checking for the existence of the `MAILPOET_PREMIUM_LICENSE` constant, done with `License::getLicense()`. I'm trying to extract that out, so we won't keep using `License::getLicense()`

mailpoet/lib/Config/Installer.php
triple0t: @pavel-mailpoet Noted 👍

mailpoet/assets/js/src/common/premium_banner_with_upgrade/premium_banner_with_upgrade.tsx
triple0t: @pavel-mailpoet Noted 👍

mailpoet/lib/Config/ServicesChecker.php
pavel-mailpoet: Oh, right, I thought the License is for the API key. That makes sense. Good job creating the method then :)

mailpoet/lib/Config/ServicesChecker.php
triple0t: @pavel-mailpoet I believe i added this line in a new commit: `Fix phpstan error`

pavel-mailpoet: I know. That is what I was trying to say. I wondered if it would be better to amend the commit that added this function instead of creating a new commit.

triple0t: @pavel-mailpoet Noted 👍

triple0t: @pavel-mailpoet  I'm not sure of a good way to make this change without destroying this branch and this PR. if this is not a blocker, I would like to leave it as it is and make a note of this for future reference.
Thanks

pavel-mailpoet: Sure, leave it, not a big deal. But I think with an interactive rebase you could make the change. But if you are not familiar with it keep the commit you could lose your work and that is not worth it. I make such changes using git fixup but that is for the future changes: https://gist.github.com/JanJakes/b4e7fbc2df0c5d2db07b1971c1ecc403#simplifying-fixups

triple0t: Noted @pavel-mailpoet

pavel-mailpoet (approved):

rohankacharegs:
GlobalStep Verified

veljkho (changes requested):
Everything's good except one scenario... when the limit has been reached (require plan upgrade). Here's the [screenshot](https://d.pr/i/4CybPf), I do not see the banner on the newsletter statistics page. I see the banner if I try to see the statistics of the subscriber (MailPoet > Subscribers > statistics link). For some reason, not displayed on the newsletter statistics only (in this case only when using lower tier subscription, require upgrade).

triple0t:
> Everything's good except one scenario... when the limit has been reached (require plan upgrade). Here's the [screenshot](https://d.pr/i/4CybPf), I do not see the banner on the newsletter statistics page. I see the banner if I try to see the statistics of the subscriber (MailPoet > Subscribers > statistics link). For some reason, not displayed on the newsletter statistics only (in this case only when using lower tier subscription, require upgrade).

@veljkho Awesome find 👏🏾. This has been fixed. The merge conflict has been fixed as well. Thank you

veljkho (approved):
Now it is good, thank you!

PR #3927, opened 2022-01-20
Update default deactivate subscriber frequency [MAILPOET-4061]

johnolek:
[MAILPOET-4061]

This PR updates the default inactive subscribers frequency from 6 months (180 days) to 12 months (365 days).

- The new default value should be automatically selected on the Advanced settings page if there is not a value saved for it in the database.
- Users who were previously using the default value of 180 days will be migrated to the new default of 365.
- Inactive subscribers should be recalculated after the setting changes, similar to how they are being recalculated if a user changes this setting themselves.

[MAILPOET-4061]: https://mailpoet.atlassian.net/browse/MAILPOET-4061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Config/Migrator.php
```diff
@@ -932,4 +935,29 @@ private function migrateEmailActionsFilters(): bool {
     }
     return true;
   }
+
+  private function updateDefaultInactiveSubscriberTimeRange(): bool {
+    // only run if the current database version is older than 3.78.0
+    if (version_compare((string)$this->settings->get('db_version', '3.78.0'), '3.78.0', '>=')) {
```
johnolek, line 941: This is a little bit different than how other migrations handle their version checks, such as [this recent check](https://github.com/mailpoet/mailpoet/blob/791f40fd7c2a04b8f2a2b4fc12807e12e67b56b4/mailpoet/lib/Config/Migrator.php#L865). 

The way we've been doing it seems like saying "Run the migration unless the user has a version greater than the release before the migration". My version is more like "Run the migration unless the user has a version matching or greater than the release with the migration" (assuming this migration will be released with version `3.78.0`).

With the current style, we have to update the checks for unreleased migrations if we have a patch release, like [this commit](https://github.com/mailpoet/mailpoet/commit/791f40fd7c2a04b8f2a2b4fc12807e12e67b56b4).

Is it not a safe assumption on my part that migrations for normal priority changes like this PR will go out with the next minor release version?

I feel like I'm missing something obvious.

mailpoet/lib/Config/Migrator.php
```diff
@@ -932,4 +935,29 @@ private function migrateEmailActionsFilters(): bool {
...
+      /**
+       * Ensure that inactive subscribers are recalculated as soon after the setting change as possible.
+       *
+       * This behavior is normally triggered when the setting is changed through the JSON API, i.e. when a user has
+       * changed the setting themselves through the UI. The `onInactiveSubscribersIntervalChange` method doesn't
+       * depend on any state, and calling it directly means avoiding code duplication.
+       */
+      $diContainer = ContainerWrapper::getInstance();
+      $apiSettings = $diContainer->get(Settings::class);
+      $apiSettings->onInactiveSubscribersIntervalChange();
```
johnolek, line 958: This doesn't feel like a great way of handling this. We're not using the DI container anywhere else in the migrator, and I'm not sure if there's a reason for that. Maybe more importantly than that, it seems like `onInactiveSubscribersIntervalChange` should live somewhere other than the `MailPoet\API\JSON\v1\Settings` class, but I don't think it makes sense to change that just for this migration.

There's also the possibility that the behavior of `\MailPoet\API\JSON\v1\Settings::onInactiveSubscribersIntervalChange` might change between when this is released and when a user actually runs the migration, so perhaps it would be better to duplicate whatever code necessary to ensure the migration's behavior stays consistent no matter what.

mailpoet/lib/Settings/SettingsController.php
```diff
@@ -12,7 +12,7 @@ class SettingsController {
   const DEFAULT_SENDING_METHOD = 'PHPMail';
   const DEFAULT_SENDING_FREQUENCY_EMAILS = 25;
   const DEFAULT_SENDING_FREQUENCY_INTERVAL = 5; // in minutes
-  const DEFAULT_DEACTIVATE_SUBSCRIBER_AFTER_INACTIVE_DAYS = 180;
+  const DEFAULT_DEACTIVATE_SUBSCRIBER_AFTER_INACTIVE_DAYS = 365;
```
johnolek, line 15: I considered changing this to be a string instead of an integer because it's a top level setting, i.e. it isn't inside an array. When array settings are saved to the database they're serialized, preserving type information on retrieval, but top level settings always get returned as a string and we need to be careful about type casting them in order to achieve consistent behavior.

At the end of the day I decided to leave it as an integer because that's what it is, and there are plenty of other places in our codebase where we need to typecast settings retrieved from the database. If we want to make a change it would probably be best to come up with a generic solution and apply it to all cases, but I'm curious to get the opinion of the reviewer(s).

mailpoet/lib/Util/Notices/InactiveSubscribersNotice.php
```diff
@@ -32,7 +32,7 @@ public function init($shouldDisplay) {
     }
 
     // don't display notice if user has changed the default inactive time range
-    $inactiveDays = $this->settings->get('deactivate_subscriber_after_inactive_days');
+    $inactiveDays = (int)$this->settings->get('deactivate_subscriber_after_inactive_days');
```
johnolek, line 35: I'm pretty sure the inactive subscriber notice hasn't been displaying as expected for the majority of users because this wasn't being cast to an integer. The only way the conditional below would be true is if there was no `deactivate_subscriber_after_inactive_days` setting saved in the user's database, which is only true for someone who's never gone into any MailPoet settings page and clicked the save button.

mailpoet/tests/DataFactories/Settings.php
```diff
@@ -181,12 +181,17 @@ public function withWooCommerceEmailCustomizerDisabled() {
   }
 
   public function withDeactivateSubscriberAfter3Months() {
-    $this->settings->set('deactivate_subscriber_after_inactive_days', 90);
+    $this->settings->set('deactivate_subscriber_after_inactive_days', '90');
```
johnolek, line 184: I decided to change this and the other `deactivate_subscriber_after_inactive_days` settings to be strings in order to simulate what we would expect to see in the real world. When we `set` something, the value gets put into an in-memory cache that sticks around for the remainder of the request. If we `get` it, we get the in-memory value, which could differ from the value we would get if we had to go to the database .

In other words, keeping these as integers would mean our tests would be getting integers when calling `$this->settings->get('deactivate_subscriber_after_inactive_days')`, but in the vast majority of cases (i.e. any time a value is actually retrieved from the database), it would be a string.

mailpoet/lib/Config/Migrator.php
lysyjan: In my opinion, you can refactor Migrator as a service. After that, when everything works well, you can inject the Settings class. But maybe it could be better to create a new service with this logic(only simple extract change events) to a new service. I guess we shouldn't inject the whole endpoint class.
I like your idea of avoiding code repetition. 👍🏻

lysyjan (commented):
Great job! I tried to reply to your comments.

mailpoet/lib/Config/Migrator.php
johnolek: @lysyjan Thanks for the in depth explanation. I was making an assumption that we wouldn't merge this branch into `master` until the next minor release, but it sounds like that assumption isn't accurate, or perhaps I'm still not understanding this 100%. I probably need to review the release process. In any case, I'm happy to rewrite this to follow the pattern of our other migrations.

johnolek: @lysyjan I just wanted to double check with you -- would the migration as I originally had it be problematic if we ensured that the code was released with version `3.78.0`? And is there a reason we would have included it in a `3.77.*` release? I just want to make sure I'm understanding the process and how the issue you described could have occurred.

mailpoet/lib/Settings/SettingsChangeHandler.php
```diff
@@ -0,0 +1,79 @@
...
+namespace MailPoet\Settings;
+
+use MailPoet\Cron\Workers\InactiveSubscribers;
+use MailPoet\Cron\Workers\WooCommerceSync;
+use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoetVendor\Carbon\Carbon;
+
+class SettingsChangeHandler {
```
johnolek, line 12: I didn't move everything over to this class, and @lysyjan I'm wondering which other methods, if any, are worth bringing over. Some things seemed a little less straightforward, like how `updateReEngagementEmailStatus` comes last and potentially alters the response of the API request.

I don't want to refactor this and leave it in a half-and-half state, but I did want to double check to see if you think some of the change logic needs to stay in `\MailPoet\API\JSON\v1\Settings`.

lysyjan: @johnolek I'm okay with the current state. It was a good step for using Migrator and the new class as DI dependencies. I wouldn't want to complicate more this pull request.

johnolek:
@lysyjan Since I won't be doing additional refactoring I believe this is (finally) ready for final review.

lysyjan (approved):

veljkho (approved):

PR #3929, opened 2022-01-21
Add "page view" tracking [MAILPOET-3897]

costasovo:
### Testing with cookie
1) Set tracking in MailPoet > Settings > Advanced to Full tracking
2) Send yourself an email with a link that goes to your site
3) Logout from the site and open the link (or use an anonymous window)
4) Check that cookie 'mailpoet_subscriber' is present (open console and click Application, then click the domain under Cookies and search for mentioned cookie
5) Reload the page
6) Check that cookie 'mailpoet_pave_view' is present (the same screen)
7) Go to admin and open stats for the subscriber who clicked the link and check that `Last engagement` (right under the heading) was updated

### Testing without cookie (for logged-in user)
1) Set tracking in MailPoet > Settings > Advanced to Partial tracking
2) Login to the site in an anonymous window and go to some front end page
3) Go to admin and open stats for the subscriber who clicked the link and check that `Last engagement` (right under the heading) was updated

### Check abandoned cart email
This change contains refactor of abandoned car rescheduling logic. Please check that abandoned cart emails work as expected. If the email is set to be sent 30 minutes after the last action it should be sent 30 minutes after the last subscriber's page visit. Please put any product to cart and wait for 30 mins in order to receive 'abandoned email' newsletter to your logged-in user's email address. Remember not to use testing site during that period.

[MAILPOET-3897]

[MAILPOET-3897]: https://mailpoet.atlassian.net/browse/MAILPOET-3897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Statistics/Track/SubscriberActivityTracker.php
```diff
@@ -0,0 +1,80 @@
...
+    $this->processTracking($subscriber);
+    return true;
+  }
+
+  private function processTracking(SubscriberEntity $subscriber): void {
+    $this->subscribersRepository->maybeUpdateLastEngagement($subscriber);
+    $this->pageViewCookie->setPageViewTimestamp($this->wp->currentTime('timestamp'));
+  }
+
+  private function shouldTrack() {
```
JanJakes, line 77: We could add a return typehint here, probably forgotten (and new methods from later commits too).

mailpoet/tests/acceptance/Subscribers/PageViewTrackingCookieCest.php
```diff
@@ -0,0 +1,31 @@
...
+
+  public function setCookieForSubscriberStoredInCookie(AcceptanceTester $i) {
+    $i->wantTo('Check page view cookie is set for subscriber saved in cookie');
+
+    $email = 'test-subscriber@example.com';
+    $subscriber = (new Subscriber())->withEmail($email)->create();
+
+    $i->amOnPage('/');
+    $i->cantSeeCookie(self::PAGE_VIEW_COOKIE_NAME);
+    $i->setCookie('mailpoet_subscriber', json_encode(['subscriber_id' => $subscriber->getId()]));
```
JanJakes, line 20: Is this needed? I mean... will the page view cookie be set even without subscriber cookie?

mailpoet/lib/Statistics/Track/SubscriberActivityTracker.php
```diff
@@ -58,23 +61,41 @@ private function processTracking(SubscriberEntity $subscriber): void {
...
+
   private function getSubscriber(): ?SubscriberEntity {
+    if ($this->activeSubscriber instanceof SubscriberEntity) {
+      return $this->activeSubscriber;
+    }
     $wpUser = $this->wp->wpGetCurrentUser();
     if ($wpUser->exists()) {
-      return $this->subscribersRepository->findOneBy(['wpUserId' => $wpUser->ID]);
+      $this->activeSubscriber = $this->subscribersRepository->findOneBy(['wpUserId' => $wpUser->ID]);
+      return $this->activeSubscriber;
```
JanJakes, line 110: I'm not sure if this type of caching (`$this->activeSubscriber`) is needed. Could we just call this function once and pass subscriber as an argument? You would then need to put more logic into `trackActivity()` itself, I guess... Not important though.

JanJakes (commented):
👍

mailpoet/tests/acceptance/Subscribers/PageViewTrackingCookieCest.php
costasovo: I implemented the cookie to set only if we can detect a subscriber either from a cookie or from logged-in WP User. So that's why I needed to set the subscriber cookie in the test. Do you think we should set always set the cookie even though there is no subscriber to track?

mailpoet/lib/Statistics/Track/SubscriberActivityTracker.php
costasovo: I added it as an optimization for the case when cookie tracking is not allowed and I need to use subscriber's last_engagement_at as a fallback value. The call of `$this->subscribersRepository->findOneBy(['wpUserId' => $wpUser->ID]);` always triggers an SQL query and because of how I structured the class it needs to call it twice. I guess I can refactor it so that the caching is not needed.

costasovo:
@JanJakes I added missing return type hints and refactored the tracker class so that it doesn't need caching of an active subscriber. Note that I kept the behavior that the cookie is set only when we can determine a subscriber. Please let me know if you think it would be better to set the cookie regardless we can determine a subscriber.

mailpoet/tests/acceptance/Subscribers/PageViewTrackingCookieCest.php
JanJakes: I see, makes sense, we don't need the cookie for anything else at the moment.

JanJakes (approved):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3933, opened 2022-01-25
Shortcode defaults are not being rendered in archives [MAILPOET-3965]

samnajian:


mailpoet/lib/Config/Shortcodes.php
```diff
@@ -187,9 +219,29 @@ public function renderArchiveDate(NewsletterEntity $newsletter) {
...
+      $this->dateCategory,
+      $this->linkCategory,
+      $this->newsletterCategory,
+      $this->subscriberCategory,
+      $this->wp
+    );
+    $shortcodeProcessor->setNewsletter($newsletter);
+
+    if (is_object($subscriber) && !is_a($subscriber, SubscriberEntity::class) && !empty($subscriber->id)) {
+      $subscriberEntity = $this->entityManager->find(SubscriberEntity::class, $subscriber->id);
```
brezocordero, line 235: I am not sure I understand this condition.
Could we remove the conversion to the old model `MailPoet\Models\Subscriber` [here](https://github.com/mailpoet/mailpoet/blob/master/mailpoet/lib/Config/Shortcodes.php#L144) and pass directly the `SubscriberEntity`?
Please take this as a question and not as something that should be done, maybe I am not understanding why we have the conversion.

brezocordero (changes requested):
@samnajian 
Awesome job on working on this complex fix.
I have only a question and a couple of nitpicking requests.

mailpoet/lib/Config/Shortcodes.php
```diff
@@ -187,9 +219,29 @@ public function renderArchiveDate(NewsletterEntity $newsletter) {
 
   public function renderArchiveSubject(NewsletterEntity $newsletter, $subscriber, SendingQueueEntity $queue) {
     $previewUrl = $this->newsletterUrl->getViewInBrowserUrl($newsletter, $subscriber, $queue);
+    /**
+     * An ugly workaround to make sure state is not shared via NewsletterShortcodes service
```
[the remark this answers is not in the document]
samnajian: @brezocordero  was wondering if linking a ticket in the free plugin codebase is fine?

mailpoet/lib/Config/Shortcodes.php
samnajian: @brezocordero there is `$previewUrl = $this->newsletterUrl->getViewInBrowserUrl($newsletter, $subscriber, $queue);` in `renderArchiveSubject` which is currently working with an instance of `SubscriberModel` and if I'm passing `SubscriberEntity` to `renderArchiveSubject` then I should adjust all the users or `MailPoet\Newsletter\Url:getViewInBrowserUrl` as well which seemed a bit too much for this PR

brezocordero (approved):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3934, opened 2022-01-26
Move CSS specific for mailpoet pages out from global admin CSS [MAILPOET-4080]

costasovo:
I found that [we load admin.css on every admin page](https://github.com/mailpoet/mailpoet/blob/master/mailpoet/lib/Config/Capabilities.php#L63-L74). The CSS file contains styles for MailPoet menu icon and styles for membership plugin integration.

To fix the problem of applying the CSS fix for menu icons on all admin pages, I moved the related rules into a different CSS bundle loaded only on mailpoet admin pages. While working on the task, I noticed that there is also a CSS rule for fixing the position of the HelpScout beacon (related [Jira ticket](https://mailpoet.atlassian.net/browse/MAILPOET-1247)) so I moved it together with the CSS for the icon image.

[MAILPOET-4080]


brezocordero (approved):
Nice job fixing two in one!
Code is good.
I tested visually (using the plugins on the issue) for the icons. Also tested the beacon display.
Works as expected.

veljkho (approved):

PR #3936, opened 2022-01-27
Move eslint configuration to the root directory [MAILPOET-3358]

lysyjan:
This pull request moves eslint configuration files to the specific directory. As a part of that, I could remove eslint dependencies from the `package.json` file. This should make updating eslint dependencies easier because they are shared with the premium plugin now.

Like the following steps, I would suppose new tasks for:
- Create a GH workflow with publishing eslint-config package
- Use a new package in the shop
- Unify eslint configurations for the premium and free plugin

I will be happy with each idea


costasovo (dismissed):
Great job! I tested it on the host and also from docker and it works as expected.

I see that there is still a different config for the premium plugin, but I assume this will be addressed by one of the follow-up tickets. 

The three tickets you mention make sense to me 👍

costasovo (approved):

PR #3937, opened 2022-01-27
Add mailpoet_subscription_before_subscribe hook [MAILPOET-3632]

websupporter:
Fixes [MAILPOET-3632]

This PR adds the `mailpoet_subscription_before_subscribe` in the SubscriberSubscribeController just before the subscriber gets stored. This hook could e.g. be used to validate the new subscriber by 3rd parties. A possible implementation could be

```
add_action(
	'mailpoet_subscription_before_subscribe',
	function(array $data, array $segmentIds, FormEntity $form) {
		if ($form->id() !== 10) {
			return
		}
		
		if ($data['some_custom_field'] === 'wrong') {
			throw new \MailPoet\UnexpectedValueException("You can not use 'wrong' in the some_custom_field.");
		}
	},
	10,
	3
);
```
I took the opportunity to add some unit tests for the SubscriberSubscribeController

[MAILPOET-3632]: https://mailpoet.atlassian.net/browse/MAILPOET-3632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/unit/Subscribers/SubscriberSubscribeControllerUnitTest.php
+++ b/mailpoet/tests/unit/Subscribers/SubscriberSubscribeControllerUnitTest.php
@@ -26,7 +26,8 @@ class SubscriberSubscribeControllerUnitTest extends \MailPoetUnitTest {
       SubscriberActions::class,
       [
         'subscribe' => Expected::never(),
-      ]
+      ],
+      $this
     );
     $subscriptionUrlFactory = Stub::makeEmpty(SubscriptionUrlFactory::class);
     $throttling = Stub::makeEmpty(SubscriptionThrottling::class);
@@ -47,7 +48,8 @@ class SubscriberSubscribeControllerUnitTest extends \MailPoetUnitTest {
       StatisticsFormsRepository::class,
       [
         'subscribe' => Expected::never(),
-      ]
+      ],
+      $this
     );
 
     $wp = Stub::make(
@@ -87,7 +89,8 @@ class SubscriberSubscribeControllerUnitTest extends \MailPoetUnitTest {
       SubscriberActions::class,
       [
         'subscribe' => Expected::never(),
-      ]
+      ],
+      $this
     );
     $subscriptionUrlFactory = Stub::makeEmpty(SubscriptionUrlFactory::class);
     $throttling = Stub::makeEmpty(
@@ -126,7 +129,8 @@ class SubscriberSubscribeControllerUnitTest extends \MailPoetUnitTest {
       StatisticsFormsRepository::class,
       [
         'subscribe' => Expected::never(),
-      ]
+      ],
+      $this
     );
     $wp = Stub::make(
       WPFunctions::class,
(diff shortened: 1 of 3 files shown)


mailpoet/tests/unit/Subscribers/SubscriberSubscribeControllerUnitTest.php
```diff
@@ -0,0 +1,494 @@
...
+        'subscribe' => Expected::never(),
+      ]
+    );
+
+    $wp = Stub::make(
+      WPFunctions::class,
+      [
+        'doAction' => Expected::never(),
+      ],
+      $this
```
wxa, line 58: Just a tip. I see you are using the third `$this` argument in stubs intermittently: sometimes it's here, sometimes it's not. There's a gotcha that expectations may not get executed if `$this` argument is omitted, it's not immediately obvious with `Expected::never()` because if it doesn't work it's kind of expected, but with `Expected::once()` etc. you may see the problem right away. So I would pay attention to that or even maybe build a habit of always adding this argument if there are any expectations, just to be safe.

mailpoet/tests/unit/Subscribers/SubscriberSubscribeControllerUnitTest.php
```diff
@@ -0,0 +1,494 @@
...
+          return $form;
+        },
+      ]
+    );
+    $statisticsFormsRepository = Stub::makeEmpty(StatisticsFormsRepository::class);
+    $wp = Stub::make(
+      WPFunctions::class,
+    [
+      'doAction' => function($receivedHook, $receivedData, $receivedSegmentIds, $receivedForm) use ($formFields, $segmentIds, $form) {
+        expect($receivedHook)->equals('mailpoet_subscription_before_subscribe');
```
wxa, line 469: It's great that you have added tests! One more idea is to test how throwing an exception from this hook interrupts the subscription process and returns the exception message. That would serve as a test for the spec 2:
> A message from the exception thrown from the filter should be displayed as an error within a form to a user who wants to subscribe

wxa (commented):
Great job and a solid first contribution! Congrats!

The example implementation you proposed would make a great documentation and help our users, because they often reuse code samples from the docs. I would also add an example of custom-validating the `$segmentIds` parameter for the sake of completeness.

mailpoet/tests/unit/Subscribers/SubscriberSubscribeControllerUnitTest.php
websupporter: Addressed in 11d934d

mailpoet/tests/unit/Subscribers/SubscriberSubscribeControllerUnitTest.php
websupporter: 7d345c5 Adds an unit test to make sure the subscription does not happen.
24d3808 Adds an integration test checking the same and whether the API output is expected.

websupporter:
> I would also add an example of custom-validating the $segmentIds parameter for the sake of completeness.
Will do when updating the KB

wxa (approved):

veljkho (approved):

PR #3942, opened 2022-01-31
Upgrade acceptance tests environment

brezocordero:
This PR upgrades the acceptance tests to use PHP8.0 for the running environment and WP-5.9-PHP8.0 for the WordPress image.

Uses the new images from:
https://github.com/mailpoet/wordpress-images/pull/28
https://github.com/mailpoet/wordpress-images/pull/29

A few tests have been updated to work both also with `seleniarm/standalone-chromium`, using `assertCssProperty`

The theme 2019 is no longer present on WP5.9. On this PR we conditionally load 2021 or 2019 depending on the CircleCI job.

To update your local tests:
Before running the new tests `./do d:d` and delete the  file `mailpoet/tests/_data/acceptancebackup.sql`

[MAILPOET-4014]

[MAILPOET-4014]: https://mailpoet.atlassian.net/browse/MAILPOET-4014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Forms/EditorAddDividerCest.php
+++ b/mailpoet/tests/acceptance/Forms/EditorAddDividerCest.php
@@ -53,8 +53,10 @@ class EditorAddDividerCest {
     $i->wantTo('Check divider on front end');
     $postUrl = $i->createPost('Title', 'Content');
     $i->amOnUrl($postUrl);
-    $i->assertAttributeContains('[data-automation-id="form_divider"]', 'style', 'height: 10px;');
-    $i->assertAttributeContains('[data-automation-id="form_divider"]', 'style', 'width: 10%;');
-    $i->assertAttributeContains('[data-automation-id="form_divider"]', 'style', 'border-top: 10px dotted black;');
+    $i->assertCssProperty('[data-automation-id="form_divider"]', 'height', '10px');
+    $i->assertAttributeContains('[data-automation-id="form_divider"]', 'style', 'width: 10%');
+    $i->assertCssProperty('[data-automation-id="form_divider"]', 'border-top-width', '10px');
+    $i->assertCssProperty('[data-automation-id="form_divider"]', 'border-top-style', 'dotted');
+    $i->assertCssProperty('[data-automation-id="form_divider"]', 'border-top-color', 'rgba(0, 0, 0, 1)');
   }
 }
--- a/mailpoet/tests/docker/codeception/docker-entrypoint.sh
+++ b/mailpoet/tests/docker/codeception/docker-entrypoint.sh
@@ -100,6 +100,10 @@ wp plugin activate woocommerce-subscriptions
 # activate theme
 wp theme activate twentytwentyone
 
+if [[ $CIRCLE_JOB == *"_oldest"* ]]; then
+  wp theme activate twentynineteen
+fi
+
 # print info about installed plugins
 wp plugin get woocommerce
 wp plugin get woocommerce-subscriptions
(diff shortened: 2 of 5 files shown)


mailpoet/tests/acceptance/Forms/EditorAddDividerCest.php
```diff
@@ -53,8 +53,10 @@ public function addDividerBlock(\AcceptanceTester $i) {
     $i->wantTo('Check divider on front end');
     $postUrl = $i->createPost('Title', 'Content');
     $i->amOnUrl($postUrl);
-    $i->assertAttributeContains('[data-automation-id="form_divider"]', 'style', 'height: 10px;');
-    $i->assertAttributeContains('[data-automation-id="form_divider"]', 'style', 'width: 10%;');
-    $i->assertAttributeContains('[data-automation-id="form_divider"]', 'style', 'border-top: 10px dotted black;');
+    $i->assertCssProperty('[data-automation-id="form_divider"]', 'height', '10px');
+    $i->assertAttributeContains('[data-automation-id="form_divider"]', 'style', 'width: 10%');
```
JanJakes, line 57: Is this just forgotten here or is there a valid reason not to use `$i-> assertCssProperty('[data-automation-id="form_divider"]', 'width', '10%');`?

mailpoet/tests/docker/codeception/docker-entrypoint.sh
```diff
@@ -97,6 +97,9 @@ sed -i "s/define( *'WP_DEBUG', false *);/$CONFIG/" /wp-core/wp-config.php
 wp plugin activate woocommerce
 wp plugin activate woocommerce-subscriptions
 
+# activate theme
+wp theme activate twentytwentyone
```
JanJakes, line 101: Can we/should we use 2022? (I think that's the default in WP 5.9.)

JanJakes (commented):
Awesome, just some tiny comments/questions.

mailpoet/tests/docker/codeception/docker-entrypoint.sh
brezocordero: We could and I tried. I decided not to do it because:

- We need to load a different theme when we want to test forms in widgets. 2022 does not have widget areas. So we would need more conditional actions depending on whether we are running acceptance-oldest and acceptance-latest.
- Some buttons and checkboxes on the front end become not interactable and the tests fail. The WC opt-in checkbox, the WC register button are a few examples. The tests pass manually, it is not an issue with our plugin. 

I can add a separate issue for it, I think it is not necessary to update the tests to 2022 now but we might need it when the work on the editor starts.

mailpoet/tests/acceptance/Forms/EditorAddDividerCest.php
brezocordero: It is not forgotten, I forgot to add a note on the PR. Thanks for noticing.

Because the value is computed when using `assertCssProperty`, the percentage becomes a number that depends on the theme/other elements on the page.
It didn't seem essential on the test, as we test it in the editor and we test the other properties on the front end, so I removed it.
Let me know what you think.

JanJakes: Oh, right, I see. It's not removed though, it stayed there in the original `assertAttributeContains` form in the diff but as it's probably property that doesn't get normalized by Chrome (except for the space that is there anyway), then I guess it works on both Chrome and Chromium and it's fine.

mailpoet/tests/docker/codeception/docker-entrypoint.sh
JanJakes: Thanks for the explanation. I don't know what are the plans or how much we need it so I'll leave it up to you to decide whether this needs a ticket in the backlog.

JanJakes (approved):
👍 As all modified files are related to tests only I think we don't need to wait for Veljko and I can merge it right away.

PR #3944, opened 2022-02-01
Incorrect use of $metas index in SendingQueue [MAILPOET-3943]

### LINKED ISSUE #3803: Incorrect use of $metas[0] index in individual newsletter processing
https://github.com/mailpoet/mailpoet/blob/34ef9fcbbf4e6fd681fec15f5736bd1741c315a8/lib/Cron/Workers/SendingQueue/SendingQueue.php#L290

This line includes the meta data for the first subscriber processed in every call to `sendNewsletter()`. Unlike the others arguments, `$metas` is not reset between iterations.

I believe the line can be modified to this:

```
['unsubscribe_url' => $unsubscribeUrls[0], 'meta' => end($metas)]
```

Alternatively, we can reset `$metas` like the remaining  arguments in that condition block like so:

```
 if ($processingMethod === 'individual') {
        $queue = $this->sendNewsletter(
          $queue,
          $preparedSubscribersIds[0],
          $preparedNewsletters[0],
          $preparedSubscribers[0],
          $statistics[0],
          $timer,
          ['unsubscribe_url' => $unsubscribeUrls[0], 'meta' => $metas[0]]
        );
        $preparedNewsletters = [];
        $preparedSubscribers = [];
        $preparedSubscribersIds = [];
        $unsubscribeUrls = [];
        $statistics = [];
$metas = [];
}
```

websupporter:
Fixes [MAILPOET-3943](https://mailpoet.atlassian.net/browse/MAILPOET-3943)
Fixes #3803

When processing the Queue in Cron/Workers/SendingQueue/SendingQueue, there are two distinct methods to do so: `bulk` and `individual`.

We loop through all subscribers and gather the individualized data like newsletter body for each subscriber. Those are stored in several arrays.

In case the processing method is `individual`, we send the newsletter immediately and resetting the mentioned arrays.

`$metas` did not get reset which resulted in #3803.

This PR adds a test to ensure the correct data gets send and resets the `$metas` array like the other arrays get resetted.

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -295,6 +295,7 @@ class SendingQueue {
         $preparedSubscribersIds = [];
         $unsubscribeUrls = [];
         $statistics = [];
+        $metas = [];
       }
     }
     if ($processingMethod === 'bulk') {
--- a/mailpoet/tests/integration/Cron/Workers/SendingQueue/SendingQueueTest.php
+++ b/mailpoet/tests/integration/Cron/Workers/SendingQueue/SendingQueueTest.php
@@ -445,8 +445,8 @@ class SendingQueueTest extends \MailPoetTest {
       Subscriber::create(),
       Subscriber::create(),
     ];
-    $subscribers[0]->set('id', 1);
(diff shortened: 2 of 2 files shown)


wxa (approved):
Looks good!

veljkho (approved):

PR #3947, opened 2022-02-02
Subscribers::listing endpoint does not find subscribers without segment who are trashed [MAILPOET-3975]

websupporter:
Fixes MAILPOET-3975

The Subscribers::listing endpoint failed to return the correct data, when filtering for trashed subscribers who are not attached to a segment.

This happened, because the SQL statement queried for subscribers who where `deletedAt IS NOT NULL AND deletedAt IS NULL`.

The `SegmentSubscribersRepository::addConstraintsForSubscribersWithoutSegment()` extended the query for subscribers and insisted, that these subscribers where not deleted and by doing so, it produced a conflict with the trash condition, which narrowed the query to only deleted subscribers. 455e6f9 removes the "is-not-deleted" condition in `addConstraintsForSubscribersWithoutSegment()`.

**What is affected by this change?**
`addConstraintsForSubscribersWithoutSegment()` is used by the Subscribers::listing endpoint. After the change, this listing endpoint returns the correct values as tested in 9e27be2.

Furthermore, `addConstraintsForSubscribersWithoutSegment()` is used by `SegmentSubscribersRepository::getSubscribersWithoutSegmentCount()`. A quick search through the codebase seemed to result in no method consumes `getSubscribersWithoutSegmentCount()`, but maybe this method is called in some opaque way a simple text search does not reveal.

### DIFF
--- a/mailpoet/lib/Segments/SegmentSubscribersRepository.php
+++ b/mailpoet/lib/Segments/SegmentSubscribersRepository.php
@@ -310,7 +310,6 @@ class SegmentSubscribersRepository {
           $queryBuilder->expr()->eq('ssg.status', ':statusSubscribed'),
           $queryBuilder->expr()->notIn('ssg.segment', $deletedSegmentsQueryBuilder->getDQL())
         ))
-      ->andWhere('s.deletedAt IS NULL')
       ->andWhere('ssg.id IS NULL')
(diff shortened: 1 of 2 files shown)


wxa (changes requested):
See my comment.

wxa (approved):

veljkho (approved):

PR #3948, opened 2022-02-02
Change default date and time when scheduling nl 4023

samnajian:


samnajian:
@costasovo Thanks for taking the time to review 💯 

> Thanks for preparing the PR! I tested the changes on my test site, and it works as expected 👍 Also, kudos for refactoring date.js to Typescript and enabling Typescript for the tests!
> 
> I noticed that PHPStan jobs failed on CI for both [PHP7](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/8701/workflows/19d5113c-9c03-4d15-b92c-1439df961a00/jobs/165411) and [PHP8](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/8701/workflows/19d5113c-9c03-4d15-b92c-1439df961a00/jobs/165412). It seems that DateTime needs to get a string as a first parameter, but you feed it with `string|int` . It needs to be fixed, so that builds pass.

Will take care of it, thanks for noticing :)


> The second commit contains refactor of date.js to Typescript and the condition change that ensures proper text in the submit button on the send page. That way, it is harder for me to see the change relevant to the commit message "Change submit button caption to publish/send based on date". It would have been easier if the commit had been divided into two commits. But it is just a detail.

Noted, will keep it in my for next commits/PRs

> Just out of curiosity, why do we need to update mocha? Does the new version contain something that was needed for Typescript?

As maintenance measures I'm used to updating devdependencies whenever I touch them, but this only done to devdependencies which are not directly influencing build artifacts.

> For some reason, the PR is marked as Draft. Is that on purpose?

I was a mistake on my side :) It's adjusted now

costasovo (approved):
LGTM!

rohankacharegs:
GlobalStep Verified

PR #3949, opened 2022-02-02
Update stats email to reflect redesign [MAILPOET-3324]

rodrigoprimo:
One of the first commits that I made in this PR was one to fix indentation issues in statsNotification.html. This might make it more convenient to review the changes in this PR commit by commit instead of everything at once.

I wasn't sure what is the best way to trigger a stats notification e-mail to test the changes that I implemented here. What I did was I changed the visibility of the method \MailPoet\Cron\Workers\StatsNotifications\Worker::constructNewsletter() to public and then I created a test in the WorkerTest.php file that would send the email and also write its contents to an HTML file. I also needed to change the constructor of the test class to define `$this->newsletterEntity = $newsletterEntity;`. Below is the snippet that I used for the test. Let me know if there is a better way to test this email (or emails in general).

```php
  public function testVerifyEmail() {
    $statsNotifications = $this->diContainer->get(Worker::class);
    $statsNotificationRepository = $this->diContainer->get(StatsNotificationsRepository::class);
    $statsNotificationEntity = $statsNotificationRepository->findOneById(1);
    $newsletter = $statsNotifications->constructNewsletter($statsNotificationEntity);
    $headers = ['From: test@test.com'];
    $headers_html = array_merge($headers, ['Content-Type: text/html; charset=UTF-8']);
    wp_mail('test@test.com', $newsletter['subject'] . ' (html)', $newsletter['body']['html'], $headers_html);
    wp_mail('test@test.com', $newsletter['subject'] . '(text)', $newsletter['body']['text'], $headers);
    file_put_contents('/var/www/html/email.html', $newsletter['body']['html']);
  }
``` 

Also, I should note that while working on this ticket I found a bug in the code that gets the number of bounced emails. So the percentage displayed in the email will be wrong until #3946 is merged.

[MAILPOET-3324]

[MAILPOET-3324]: https://mailpoet.atlassian.net/browse/MAILPOET-3324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/StatsNotifications/Worker.php
+++ b/mailpoet/lib/Cron/Workers/StatsNotifications/Worker.php
@@ -144,9 +144,9 @@ class Worker {
     $context = [
       'subject' => $subject,
       'preheader' => sprintf(_x(
-        '%1$s%% opens, %2$s%% clicks, %3$s%% unsubscribes in a nutshell.', 'newsletter open rate, click rate and unsubscribe rate', 'mailpoet'),
-        number_format($opened, 2),
+        '%1$s%% clicks, %2$s%% opens, %3$s%% unsubscribes in a nutshell.', 'newsletter open rate, click rate and unsubscribe rate', 'mailpoet'),
         number_format($clicked, 2),
+        number_format($opened, 2),
         number_format($unsubscribed, 2)
       ),
       'topLinkClicks' => 0,
--- a/mailpoet/lib/Cron/Workers/StatsNotifications/Worker.php
+++ b/mailpoet/lib/Cron/Workers/StatsNotifications/Worker.php
@@ -136,6 +136,7 @@ class Worker {
     $statistics = $this->newsletterStatisticsRepository->getStatistics($newsletter);
     $clicked = ($statistics->getClickCount() * 100) / $statistics->getTotalSentCount();
     $opened = ($statistics->getOpenCount() * 100) / $statistics->getTotalSentCount();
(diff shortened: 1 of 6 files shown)


costasovo (dismissed):
Good job! Thank you for finding and addressing the bugs you found!

As for testing during development, this sending is done in the background job so it is a bit tricky to test. I don't know any nice way of testing it. I like your approach. As an alternative approach that I would probably use (not knowing yours) I would send a newsletter locally, create some opens and clicks and then reschedule the scheduled task for the stats notification email to the past (and reset its status to scheduled if necessary). The email would be then sent by refreshing a site. When I need to activate xdebug and debug the background job I add `&XDEBUG_TRIGGER=1` into the cron URL [here](https://github.com/mailpoet/mailpoet/blob/master/mailpoet/lib/Cron/CronHelper.php#L168). It is probably not an ideal solution. I just wanted to share it.

rodrigoprimo:
Thanks for sharing an alternative solution to debugging those emails.

rodrigoprimo:
@costasovo, I implemented a few changes that @NeosinneR requested. Could you please review this PR again? More information about the changes in the Jira ticket.

@NeosinneR, here is a screenshot of the email after the changes you requested. Let me know if it looks good or if I missed something.

Screenshot from 2022-02-07 11-49-39

NeosinneR:
@rodrigoprimo Looks good. Thanks!

costasovo (approved):
LGTM! 👍

veljkho (approved):

PR #3950, opened 2022-02-02
Only store relevant meta for WooCommerce emails [MAILPOET-3518]

johnolek:
[MAILPOET-3518] please check testing instructions for 2 test cases inside Jira ticket.

This PR changes the way meta data is stored for automatic emails sent when a customer purchases a specified WooCommerce product or a product in a particular category. Previously we were storing all the product or category IDs associated with the order even if they weren't the IDs that triggered the newsletter. With this change, we only store the IDs that actually triggered the email, meaning the following scenario can occur:

1. There is an automatic email set up to send when a user purchases product ID 10
2. A customer makes a purchase that includes product IDs 10 and 15
3. The user receives the automatic email
4. The site admin updates the automatic email to be triggered when product ID 10 OR 15 is purchased
5. The same customer makes a purchase that includes product ID 15
6. The user receives the automatic email again

Previously, we would have stored product IDs 10 AND 15 at step 2 above, so the customer wouldn't receive the email in step 6 because it looked like that email had already been sent due to purchasing product 15 in step 2, even though that wasn't the case.

I refactored some of the existing tests in `PurchasedProductTest` because many of them required the same setup steps. I would request that whoever reviews this please double check that I didn't inadvertently alter the behavior of any existing tests.

[MAILPOET-3518]: https://mailpoet.atlassian.net/browse/MAILPOET-3518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/AutomaticEmails/WooCommerce/Events/PurchasedInCategory.php
+++ b/mailpoet/lib/AutomaticEmails/WooCommerce/Events/PurchasedInCategory.php
@@ -5,6 +5,7 @@ namespace MailPoet\AutomaticEmails\WooCommerce\Events;
 use MailPoet\AutomaticEmails\WooCommerce\WooCommerce;
 use MailPoet\DI\ContainerWrapper;
 use MailPoet\Logging\LoggerFactory;
+use MailPoet\Models\Newsletter;
 use MailPoet\Models\Subscriber;
 use MailPoet\Newsletter\AutomaticEmailsRepository;
 use MailPoet\Newsletter\Scheduler\AutomaticEmailScheduler;
@@ -126,18 +127,17 @@ class PurchasedInCategory {
     }
 
     $schedulingCondition = function($automaticEmail) use ($orderedProductCategories, $subscriber) {
-      $meta = $automaticEmail->getMeta();
+      $matchedCategories = $this->getProductCategoryIdsMatchingNewsletterTrigger($automaticEmail, $orderedProductCategories);
+      if (empty($matchedCategories)) {
+        return false;
+      }
 
-      if (empty($meta['option'])) return false;
       if ($this->repository->wasScheduledForSubscriber($automaticEmail->id, $subscriber->id)) {
-        $sentAllProducts = $this->repository->alreadySentAllProducts($automaticEmail->id, $subscriber->id, 'orderedProductCategories', $orderedProductCategories);
+        $sentAllProducts = $this->repository->alreadySentAllProducts($automaticEmail->id, $subscriber->id, 'orderedProductCategories', $matchedCategories);
         if ($sentAllProducts) return false;
       }
 
-      $metaCategories = array_column($meta['option'], 'id');
-      $matchedCategories = array_intersect($metaCategories, $orderedProductCategories);
-
-      return !empty($matchedCategories);
+      return true;
     };
 
     $this->loggerFactory->getLogger(self::SLUG)->addInfo(
@@ -152,7 +152,28 @@ class PurchasedInCategory {
       self::SLUG,
       $schedulingCondition,
       $subscriber->id,
-      ['orderedProductCategories' => $orderedProductCategories]
+      ['orderedProductCategories' => $orderedProductCategories],
+      [$this, 'metaModifier']
     );
+  }
+
+  public function metaModifier(Newsletter $automaticEmail, array $meta): array {
+    $orderedProductCategoryIds = $meta['orderedProductCategories'] ?? null;
+    if (empty($orderedProductCategoryIds)) {
+      return $meta;
+    }
+    $meta['orderedProductCategories'] = $this->getProductCategoryIdsMatchingNewsletterTrigger($automaticEmail, $orderedProductCategoryIds);
+
+    return $meta;
+  }
+
+  private function getProductCategoryIdsMatchingNewsletterTrigger(Newsletter $automaticEmail, array $orderedCategoryIds): array {
+    $automaticEmailMeta = $automaticEmail->getMeta();
+    if (empty($automaticEmailMeta['option'])) {
+      return [];
+    }
+    $emailTriggeringCategoryIds = array_column($automaticEmailMeta['option'], 'id');
+
+    return array_intersect($emailTriggeringCategoryIds, $orderedCategoryIds);
   }
 }
--- a/mailpoet/lib/Newsletter/Scheduler/AutomaticEmailScheduler.php
+++ b/mailpoet/lib/Newsletter/Scheduler/AutomaticEmailScheduler.php
@@ -25,6 +25,15 @@ class AutomaticEmailScheduler {
     foreach ($newsletters as $newsletter) {
       if ($newsletter->event !== $event) continue;
       if (is_callable($schedulingCondition) && !$schedulingCondition($newsletter)) continue;
+
+      /**
+       * $meta will be the same for all newsletters by default. If we need to store newsletter-specific meta, the
+       * $metaModifier callback can be used.
+       *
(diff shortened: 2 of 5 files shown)


mailpoet/lib/Newsletter/Scheduler/AutomaticEmailScheduler.php
```diff
@@ -19,12 +19,15 @@ public function __construct(
...
   }
 
-  public function scheduleAutomaticEmail($group, $event, $schedulingCondition = false, $subscriberId = false, $meta = false) {
+  public function scheduleAutomaticEmail($group, $event, $schedulingCondition = false, $subscriberId = false, $meta = false, $metaModifier = null) {
     $newsletters = Scheduler::getNewsletters(Newsletter::TYPE_AUTOMATIC, $group);
     if (empty($newsletters)) return false;
     foreach ($newsletters as $newsletter) {
       if ($newsletter->event !== $event) continue;
       if (is_callable($schedulingCondition) && !$schedulingCondition($newsletter)) continue;
+      if (is_callable($metaModifier)) {
```
johnolek, line 28: Having another callback seems kind of clunky but the meta data we want to store is based on the specific newsletter, and this was the cleanest approach I could come up with.

I also thought about adding a filter, something like the following:

`$meta = WPFunctions::get()->applyFilters($meta, $newsletter, $group, $event, etc)`

The individual events could then hook into the filter as part of their `init` methods.

I didn't know if we would want to expose a filter for a piece of functionality like this that probably wouldn't make sense to have other themes/plugins modifying.

I'm very open to other ideas for how to handle this and welcome suggestions.

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/PurchasedProductTest.php
```diff
@@ -290,83 +202,84 @@ public function _runTestItSchedulesEmailForState($state) {
...
-    // when 'woocommerce_order_status_$state' hook is triggered, an email should be scheduled
-    WPFunctions::get()->doAction('woocommerce_order_status_' . $state, $orderId);
-    $event->scheduleEmailWhenProductIsPurchased($orderId);
-    $scheduledTask = Sending::getByNewsletterId($newsletter->id);
-    expect($scheduledTask)->notEmpty();
-    $queue = $scheduledTask->queue();
-    expect($queue->getMeta())->equals(['orderedProducts' => [$incorrectProductId, $productId]]);
-    return $orderId;
+    // when 'woocommerce_order_status_$state' hook is triggered an email is scheduled when appropriate
+    WPFunctions::get()->doAction('woocommerce_order_status_' . $wooOrderState, $orderId);
```
johnolek, line 259: This is from the original version of the tests but I'm not understand why it was there to begin with. The original tests still passed with this line commented out.

mailpoet/lib/AutomaticEmails/WooCommerce/Events/PurchasedInCategory.php
```diff
@@ -126,18 +127,17 @@ public function scheduleEmail($orderId) {
...
-      $meta = $automaticEmail->getMeta();
+      $matchedCategories = $this->getProductCategoryIdsMatchingNewsletterTrigger($automaticEmail, $orderedProductCategories);
+      if (empty($matchedCategories)) {
+        return false;
+      }
 
-      if (empty($meta['option'])) return false;
       if ($this->repository->wasScheduledForSubscriber($automaticEmail->id, $subscriber->id)) {
-        $sentAllProducts = $this->repository->alreadySentAllProducts($automaticEmail->id, $subscriber->id, 'orderedProductCategories', $orderedProductCategories);
+        $sentAllProducts = $this->repository->alreadySentAllProducts($automaticEmail->id, $subscriber->id, 'orderedProductCategories', $matchedCategories);
```
johnolek, line 136: The original code checked for `$orderedProductCategories` here, whereas the original code for `PurchasedProduct` searched just for the matched categories, so I updated this one to be consistent with the `PurchasedProduct` version. Would this alter behavior in an undesirable way?

mailpoet/lib/Newsletter/Scheduler/AutomaticEmailScheduler.php
wxa: You're right John, it does seem clunky, but since the original `$meta` parameter was probably implied to be newsletter-agnostic, that is a viable way to tailor metadata to a specific newsletter without larger-scale refactoring of the event system. 

To make it easier to grasp, maybe it's worth adding a comment explaining why `$metaModifier` was needed, e.g.:
> The $meta argument is shared between all newsletters, to make it newsletter-specific, the $metaModifier callback can be passed.

mailpoet/lib/AutomaticEmails/WooCommerce/Events/PurchasedInCategory.php
wxa: At first glimpse it doesn't seem to have a negative impact on the behavior, the change is a logical improvement to me. Is this case covered by tests?

wxa (commented):
Looks good overall! Thanks for refactoring and improving tests. I just have a couple minor questions.

mailpoet/lib/Newsletter/Scheduler/AutomaticEmailScheduler.php
johnolek: I definitely agree a comment is in order here!

mailpoet/lib/AutomaticEmails/WooCommerce/Events/PurchasedInCategory.php
johnolek: It wasn't covered in the tests, and I've now added one. I'm glad I did because I'm pretty sure this represents a current prod bug:

1. Set up an email with a trigger of category 15 with a 1 day delay
2. Make a purchase with category 15
3. Email is queued and we mark that the newsletter was queued for category 15
4. Make a purchase with category 15 and 16
5. We see that the newsletter was already scheduled 
6. `$sentAllProducts` is `false`, because we haven't marked category 16 as sent
7. A duplicate email is queued

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/PurchasedProductTest.php
```diff
@@ -91,197 +91,109 @@ public function testItDoesNotScheduleEmailWhenCustomerIsNotAWCSegmentSubscriber(
...
+    // This was a cause for a duplicate email: https://mailpoet.atlassian.net/browse/MAILPOET-3254
+    $event2 = $this->createOrderEvent($subscriber, [1000, 12345]);
+    $event2->scheduleEmailWhenProductIsPurchased(13);
+    $queue2 = SendingQueue::where('newsletter_id', $email->id)->findMany();
+    expect($queue2)->count(count($queue1));
+  }
 
-    $event->scheduleEmailWhenProductIsPurchased($orderId);
-    $queue1 = SendingQueue::where('newsletter_id', $newsletter->id)->findMany();
+  public function testItDoesntCreateASecondSendingTaskWhenAlreadySentWhat() {
```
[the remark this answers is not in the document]
johnolek: My bad -- I even left in a stupid placeholder method name. I wanted to make sure that the test still passed using my refactored helpers. I've removed my duplicate and left the test that more closely matches the simpler behavior of the original. Sorry I didn't catch this!

wxa (approved):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3955, opened 2022-02-03
Storybook is not accessible in browser when ran on container [MAILPOET-4078]

samnajian:
### Description

This PR exposes the Storybook port number to the Docker host to make the Storybook accessible [Mailpoet-4078]

### How to test

- Run `./do storybook:watch`
- Confirm storybook is accessible under localhost:8083 

[MAILPOET-4078]

[MAILPOET-4078]: https://mailpoet.atlassian.net/browse/MAILPOET-4078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docker-compose.override.macos-sample.yml
+++ b/docker-compose.override.macos-sample.yml
@@ -6,8 +6,6 @@ services:
     image: mariadb:10.5.8
 
   wordpress:
-    ports:
-      - "8083:8083"
     volumes:
       - nfs-wordpress:/var/www/html
       - nfs-mailpoet:/var/www/html/wp-content/plugins/mailpoet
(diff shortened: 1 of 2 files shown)


docker-compose.override.macos-sample.yml
```diff
@@ -6,6 +6,8 @@ services:
     image: mariadb:10.5.8
 
   wordpress:
+    ports:
+      - "8083:8083" # Storybook port number, see package.json
```
costasovo, line 10: Do we need to repeat the port exposure here? It is already exposed in the main docker.compose.yml so I would say it is redundant.

costasovo (commented):
Thank you for the improvement! I have just one question in the comments.

costasovo (approved):
👍

veljkho (approved):

PR #3945, opened 2022-02-01
Enable Xdebug develop mode

rodrigoprimo:
In order to test the change proposed here, it is necessary to stop the dev environment, rebuild the wordpress service (`docker-compose build wordpress`), start the environment again, and check phpinfo() to confirm that `develop` is included in `xdebug.mode`. 

[MAILPOET-4098]

[MAILPOET-4098]: https://mailpoet.atlassian.net/browse/MAILPOET-4098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3960, opened 2022-02-07
Revert "Enable Xdebug develop mode" [MAILPOET-4121]

rodrigoprimo:
It seems that on some machines (maybe Macs?), enabling the Xdebug develop mode significantly impacts the performance of the WP site. On a quick investigation, our impression is that this is related to the DI container that we use. So for now, we are opting to revert this change. It is possible to enable the develop mode when needed using an environment variable that can be added to docker-compose.override.yml.

For more information, see this Slack thread: p1643895995581989-slack-C01G2EDTMQF

Reverts mailpoet/mailpoet#3945

[MAILPOET-4121]

[MAILPOET-4121]: https://mailpoet.atlassian.net/browse/MAILPOET-4121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo:
CircleCI is failing but I'm pretty sure it is not related to the changes here. I will keep trying to re-run it to see if it passes.

costasovo (approved):
Thank you!

PR #3961, opened 2022-02-08
Add acceptance tests for WC block checkout [MAILPOET-3937]

lysyjan:
[MAILPOET-3937]

[MAILPOET-3937]: https://mailpoet.atlassian.net/browse/MAILPOET-3937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -958,12 +958,8 @@ class RoboFile extends \Robo\Tasks {
   }
 
   public function downloadWooCommerceBlocksZip($tag = null) {
-    if (!getenv('WP_GITHUB_USERNAME') && !getenv('WP_GITHUB_TOKEN')) {
-      $this->yell("Skipping download of WooCommerce Blocks", 40, 'red');
-      exit(0); // Exit with 0 since it is a valid state for some environments
-    }
-    $this->createGithubClient('woocommerce/woocommerce-gutenberg-products-block')
-      ->downloadReleaseZip('woo-gutenberg-products-block.zip', __DIR__ . '/tests/plugins/', $tag);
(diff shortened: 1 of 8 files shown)


mailpoet/RoboFile.php
```diff
@@ -957,6 +957,15 @@ public function releasePublishSlack($version = null) {
...
 
+  public function downloadWooCommerceBlocksZip($tag = null) {
+    if (!getenv('WP_GITHUB_USERNAME') && !getenv('WP_GITHUB_TOKEN')) {
+      $this->yell("Skipping download of WooCommerce Blocks", 40, 'red');
+      exit(0); // Exit with 0 since it is a valid state for some environments
+    }
+    $this->createGithubClient('woocommerce/woocommerce-gutenberg-products-block')
+      ->downloadReleaseZip('woo-gutenberg-products-block.zip', __DIR__ . '/tests/plugins/', $tag);
+  }
+
```
costasovo, line 968: Just FYI: We added a tool for downloading free plugins directly from repo see https://github.com/mailpoet/mailpoet/blob/master/mailpoet/RoboFile.php#L978-L981
There is no need to change it. Just maybe next time, it might be easier to use and works without a token.

costasovo (changes requested):
It looks good! Great job! Could you please add the deactivation of the plugin to the extension of the plugin: https://github.com/mailpoet/mailpoet/blob/master/mailpoet/tests/_support/PluginsExtension.php

We deactivate all 3rd party plugins by default at the beginning of tests.

costasovo (approved):
LGTM! 👍

veljkho (approved):

PR #3963, opened 2022-02-08
Alter MSS notices on Status Page [MAILPOET-3909]

websupporter:
Fixes [MAILPOET-3909]

Sometimes, due to host configuration, an installation can not communicate with bridge.mailpoet.com

When you do not use the MailPoet sending service, there is currently no check in the status page to inform you about this issue:
image

A check is only performed on the status page after you have already connected.

Users try to connect using their API key, but get an error message. With this PR the user will get this information on the status page:
image
image

In this PR we check always, if we can connect, regardless of the question, if the installation is using the sending service. Depending on whether we use the service or not, different messages are shown to the user.

1. If the user is not connected,
1.1. but can reach the bridge: "Your installation can connect to our sending service"
1.2. and can not reach the bridge: "Currently, your installation can not reach the sending service. If you want to use our service please consult our [link]knowledge base article[/link] for troubleshooting tips."
2. If the user is connected,
2.1. but can't reach the bridge: "Connection unsuccessful. Please contact our technical support for assistance."
2.2. and can reach the bridge: "Connection successful."

The later cases are the wording as it was before. Not sure overall, if that's the best route to take.

[MAILPOET-3909]: https://mailpoet.atlassian.net/browse/MAILPOET-3909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wxa (approved):
Looks good. The only remark I have is that an always-on check can slow down the loading of the System Status page, or even make it inaccessible (e.g. if connection to bridge is very slow or times out). If this happens to be a problem, we'll need to implement a way to perform this check only if a customer wants so.

veljkho (approved):

PR #3968, opened 2022-02-11
Update eslint [MAILPOET-4114]

websupporter:
Fixes [MAILPOET-4114]

This PR updates Eslint and related packages in eslint-config (c266654)

Since those updates updated also the eslint rules, which we need to adhere, the following commits update the codebase accordingly.

I tried to fix one specific rule per commit. So, for example "[Fix eslint react/jsx-no-useless-fragment errors](https://github.com/mailpoet/mailpoet/pull/3968/commits/30607dd1dded8c3a4104db18beb577d7cae725b5)" will fix the jsx-no-useless-fragment issues.

[MAILPOET-4114]: https://mailpoet.atlassian.net/browse/MAILPOET-4114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/logs/list.tsx
```diff
@@ -28,7 +28,7 @@ type MessageProps = {
 
 function Message({ message, editing }: MessageProps):JSX.Element {
   if (!editing) {
-    return (<>{`${message.substr(0, 150)}…`}</>);
+    return (<div>{`${message.substr(0, 150)}…`}</div>);
```
JanJakes, line 31: This changes the DOM... could we simply do this?
```ts
return `${message.substr(0, 150)}…`;
```

mailpoet/assets/js/src/settings/pages/key_activation/key_activation.tsx
```diff
@@ -19,6 +19,70 @@ type KeyState = {
   is_approved: boolean;
 }
 
+function Messages(
+  state:KeyActivationState,
+  showPendingApprovalNotice:boolean,
+  activationCallback:()=>Promise<void>
```
JanJakes, line 25: Apart from the already mentioned `argument: type` spacing, the `=>` arrow should have one space on each side. Interesting that our eslint doesn't catch that. Anyway, Prettier will solve it.

mailpoet/assets/js/src/common/tabs/tab.tsx
```diff
@@ -14,7 +14,7 @@ function Tab({
   children,
 }: Props) {
   return (
-    children
+    <>{children}</> // eslint-disable-line react/jsx-no-useless-fragment
```
JanJakes, line 17: Why don't you leave the original `return children` here?

mailpoet/assets/js/src/common/listings/newsletter_status.tsx
```diff
@@ -28,7 +28,7 @@ function NewsletterStatus({
   const sent = (!scheduledFor || isPast(scheduledFor)) && processed >= total;
   const sentWithoutQueue = status === 'sent' && total === undefined;
   let percentage = 0;
-  let label = t('notSentYet');
+  let label : string|JSX.Element = t('notSentYet');
```
JanJakes, line 31: Doesn't `JSX.Element` already include `string`?

mailpoet/assets/js/src/logs/list.tsx
```diff
@@ -17,7 +17,7 @@ type Log = {
 
 export type Logs = Log[];
 
-function isCtrl(event: MouseEvent): boolean {
+function isCtrl(event: {ctrlKey ?:boolean, metaKey ?:boolean, altKey ?:boolean}): boolean {
```
JanJakes, line 20: Why `MouseEvent` doesn't work here? Is there a correct type somewhere (like `React.MouseEvent` or something)?

mailpoet/assets/js/src/newsletters/campaign_stats/page.tsx
```diff
@@ -65,7 +65,7 @@ function CampaignStatsPage({ match, history, location }: Props) {
         loading: false,
         item: response.data,
       });
-    }).fail((response) => {
+    }).fail((response:{errors:{message:string}[]}) => {
```
JanJakes, line 68: I think you could define an alias for this type as it's used in multiple places.

mailpoet/assets/js/src/date.ts
```diff
@@ -45,15 +45,15 @@ export const MailPoetDate = {
...
+    momentDate = Moment(date);
     if (options.offset === 0) momentDate = momentDate.utc();
     return momentDate.format(this.convertFormat(this.options.format));
   },
   toDate: function toDate(date: MomentInput, opts?: DateOptions): Date {
     const options = opts || {};
     this.init(options);
 
-    return Moment(date, this.convertFormat(options.parseFormat)).toDate();
+    return Moment(date).toDate();
```
JanJakes, line 56: What does this do? Seems like it could change the behavior somehow...

mailpoet/assets/js/src/form_editor/form_preview.ts
```diff
@@ -21,12 +21,12 @@ jQuery(($) => {
...
       }
       // Allow message processing only when send from editor's origin
-      const editorUrl = new URL($previewForm.data('editor-url'));
+      const editorUrl = new URL(<string> $previewForm.data('editor-url'));
       if (editorUrl.origin !== event.origin) {
         return;
       }
 
-      const formType = event.data.formType;
+      const formType = <string> event.data.formType;
```
JanJakes, line 29: I think we should stick to one syntax of type assertions and since we use `as` everywhere, it would be good to use it here as well.

mailpoet/assets/js/src/form_editor/translations.ts
```diff
@@ -1,11 +1,11 @@
...
 
-export default (moduleTranslations): void => {
+export default (moduleTranslations:string[]): void => {
   moduleTranslations.forEach((translations) => {
     const parsed = JSON.parse(translations);
     if (!parsed || !parsed.locale_data?.messages) {
       return;
     }
-    setLocaleData(parsed.locale_data.messages);
+    setLocaleData(<LocaleData>parsed.locale_data.messages);
```
JanJakes, line 9: Same here, I think we should stick to `as`.

mailpoet/assets/js/src/form_editor/utils/link_suggestions.tsx
```diff
@@ -2,7 +2,7 @@ import apiFetch from '@wordpress/api-fetch';
 import { addQueryArgs } from '@wordpress/url';
 import { decodeEntities } from '@wordpress/html-entities';
 import MailPoet from 'mailpoet';
-import type {WP_REST_API_Search_Results} from 'wp-types';
+import type { WP_REST_API_Search_Results } from 'wp-types';
```
JanJakes, line 5: Ah, I see this got somehow fixed in an unrelated commit. Not all spacing seems to be fixed though.

mailpoet/assets/js/src/notices/notice.tsx
```diff
@@ -36,7 +36,7 @@ function Notice({
     if (timeout) {
       timeoutRef.current = setTimeout(close, timeout);
     }
-    return () => (timeoutRef.current ? clearTimeout(timeoutRef.current) : null);
+    return () => (timeoutRef.current ? clearTimeout(timeoutRef.current as NodeJS.Timeout) : null);
```
JanJakes, line 39: Two things here:
1. `NodeJS.Timeout` will not be the correct type as this is run in browser where it doesn't exist. You can simply say "whatever `setTimeout` returns" - `ReturnType<typeof setTimeout>`.
2. You should avoid type assertions (`as`) as much as possible and type the original data, so `const timeoutRef = React.useRef<ReturnType<typeof setTimeout>>(null)` here.

mailpoet/assets/js/src/segments/dynamic/store/actions.ts
```diff
@@ -130,6 +130,6 @@ export function* handleSave(segmentId?: number): Generator<{
       messages.onCreate(segment);
     }
   } else {
-    yield setErrors(error);
+    yield setErrors(<string[]> error);
```
JanJakes, line 133: Same here (using `as`).

mailpoet/assets/js/src/segments/dynamic/store/controls.ts
```diff
@@ -31,7 +31,7 @@ export async function LOAD_SEGMENT({ segmentId }: { segmentId: number }): Promis
     });
     return {
       success: true,
-      res: convertSavedData(res.data),
+      res: convertSavedData(<{ [key: string]: string | number; }> res.data),
```
JanJakes, line 34: You can also use `Record<string, string | number>`.

mailpoet/assets/js/src/sending-paused-notices-resume-button.ts
```diff
@@ -55,6 +55,6 @@ const resumeSendingIfAuthorized = (fromAddress: string | null) => isValidFromAdd
 // jQuery-based notice implementation which doesn't trigger pure-JS added listeners
 jQuery(($) => {
   $(document).on('click', '.notice .mailpoet-js-button-resume-sending', (e) => (
-    resumeSendingIfAuthorized(e.target.value)
+    resumeSendingIfAuthorized(<string> e.target.value)
```
JanJakes, line 58: Also `as`.

mailpoet/assets/js/src/settings/store/actions/settings.ts
```diff
@@ -45,7 +45,7 @@ export function* saveSettings() {
     fromAddressModalCanBeShown: false,
   });
   yield setSettings(res.data);
-  yield setReEngagement(res.meta);
+  yield setReEngagement(<ReEngagement> res.meta);
```
JanJakes, line 48: Also `as` (and in all the cases below).

mailpoet/assets/js/src/subscribers/importExport/export.ts
```diff
@@ -158,7 +158,7 @@ jQuery(document).ready(() => {
       MailPoet.Modal.loading(false);
     }).done((response) => {
       const resultMessage = MailPoet.I18n.t('exportMessage')
-        .replace('%1$s', `<strong>${parseInt(response.data.totalExported, 10).toLocaleString()}</strong>`)
+        .replace('%1$s', `<strong>${parseInt(<string> response.data.totalExported, 10).toLocaleString()}</strong>`)
```
JanJakes, line 161: Doesn't `toLocaleString` already return string for sure?

eslint-config/package.json
```diff
@@ -8,6 +8,7 @@
   "devDependencies": {
     "@babel/eslint-parser": "^7.17.0",
     "@babel/preset-react": "^7.16.7",
+    "@types/jquery": "^3.5.13",
```
JanJakes, line 11: I think this belongs to the plugin, not to `eslint-config`.

mailpoet/assets/js/src/common/thumbnail.ts
```diff
@@ -3,12 +3,12 @@ import MailPoet from 'mailpoet';
...
 /**
- * Generates a thumbnail from a DOM element.
+ * Generates a thumbnail from a HTML element.
  *
- * @param  {DOMElement}      element
+ * @param  {HTMLElement}      element
  * @return {Promise<String>} DataURL of the generated image.
  */
-export const fromDom = async (element) => {
+export const fromDom = async (element:HTMLElement) => {
```
JanJakes, line 11: What's the issue with `DomElement`, would it be a wrong type here?

eslint-config/package.json
```diff
@@ -6,11 +6,12 @@
   "main": "index.js",
   "scripts": {},
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "^4.33.0",
-    "@typescript-eslint/parser": "^4.33.0",
-    "babel-eslint": "^10.1.0",
-    "eslint": "^7.32.0",
-    "eslint-config-airbnb": "^18.2.1",
+    "@babel/eslint-parser": "^7.17.0",
+    "@babel/preset-react": "^7.16.7",
```
JanJakes, line 10: Do we really need `@babel/preset-react` in the `eslint-config`? I'm not sure but I think this is related to compilation rather than linting and maybe we need only the parser here.

(We could also try if the default eslint's parser works and if it's faster or slower but yeah, not needed now.)

JanJakes (changes requested):
Great job! A ton of work and very nice chagnes!

I have some questions and comments and some small issues. Also some general notes:

1. There are a couple of commits named `package-lock.json`. Such a commit message says nothing, it would be good to actually keep this together with the respective `package.json` changes in a single commit. If you know which corresponds to which, feel free to squash them.
2. The [first commit](https://github.com/mailpoet/mailpoet/pull/3968/commits/c266654ef480f04102c12b51ae0ce445e887fc80) deletes `package-lock.json` altogether although the commit message says "Update eslint and related npm packages". This will essentially erase the history of the file. Maybe you can do an interactive rebase and run the installs again in the correct commits. I can also help if needed (a quick call or something).
3. Some commits include a bit of unrelated changes (fixing errors from previous commits, etc.) but we don't have to be pedantic here, it's just something to watch in the future.
4. When adding new eslint ignores, we should ("officially", see [this P2](https://mailpoetdev.wordpress.com/2021/03/19/disabling-linting-rules/)) explain why it's ignored. I know in this case it's a bit hard as you don't necessarily know it (you didn't write the original code).

All-in-all, great job. For the git issues we have a [this article](https://mailpoetdev.wordpress.com/2021/12/01/git-guidelines/) that links [proposed guidelines](https://gist.github.com/JanJakes/b4e7fbc2df0c5d2db07b1971c1ecc403#file-git-guidelines-md) and a [practical manual](https://gist.github.com/JanJakes/b4e7fbc2df0c5d2db07b1971c1ecc403#file-git-rebase-md) for interactive rebase to fix, merge, or otherwise change commit history. Again, if you need any more info or help, we can also jump on a call. (It's also not necessary to fix the history too much, it's just good to try it out and know how it works.)

mailpoet/assets/js/src/global.d.ts
```diff
@@ -5,6 +5,12 @@ declare module 'wp-js-hooks' {
   function applyFilters(name: string, ...args: any[]): any;
 }
 
+declare module 'mailpoet' {
+  export const I18n: {
+    t: (text: string) => string;
+  };
+}
+
```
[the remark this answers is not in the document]
websupporter: MailPoet.I18n.t() is not defined and thus could return `any`. If you use it in combination with methods, which expect `string` TypeScript complains. Afaik @samnajian is working on a shared global definition of MailPoet which would make this superfluous [MAILPOET-3140]

mailpoet/assets/js/src/form/fields/field.jsx
```diff
@@ -11,7 +11,7 @@ import jQuery from 'jquery';
 import PropTypes from 'prop-types';
 
 class FormField extends React.Component {
-  renderField = (data) => {
+  static renderField = (data) => {
```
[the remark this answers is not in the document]
websupporter: This is what I understood: https://eslint.org/docs/rules/class-methods-use-this, not a fan of it either :D I guess finding consensus in the team on that would be helpful.

mailpoet/assets/js/src/subscribers/importExport/export.ts
websupporter: The problem here is `parseInt` expecting string ([playground](https://www.typescriptlang.org/play?jsx=0&ts=4.5.4#code/EQVwzgpgBGAuBOBLAxrYBuAUAGwrKAZgPZEBcIAdgNYVEDuFUUAvFAIwBMAzFpgA4BDeJACSFWAApiRADTsADAEosg4RDGSAPHCQUA5gD5pctkvRA))

mailpoet/assets/js/src/form_editor/components/_stories/size_settings.tsx
```diff
@@ -1,57 +1,63 @@
 import React from 'react';
 import { action } from '_storybook/action';
+import { jsx } from '@emotion/react';
 import { SizeSettings } from '../size_settings';
+import JSX = jsx.JSX;
```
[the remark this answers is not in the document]
websupporter: Didn't see that. Seems like IDE auto import.

mailpoet/assets/js/src/common/tabs/tab.tsx
websupporter: Sorry, this was my bad. Because the commit you are viewing overwrote a commit I already did when trying to just return the children, which led to problems down the road, why I decided here to disable the rule is probably the sane path. 
image
This is the actual proposed diff when you merge.

mailpoet/assets/js/src/form_editor/form_preview.ts
websupporter: That makes sense. In general I prefer `<string>`, but it doesn't work in tsx files, this is why i switched in tsx files to `as string`. But makes sense to have it consistent.

mailpoet/assets/js/src/form/fields/field.jsx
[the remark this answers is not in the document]
JanJakes: Yeah, or let's just disable it for now (maybe the consensus is needed to actually introduce this rule).

mailpoet/assets/js/src/global.d.ts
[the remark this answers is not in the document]
JanJakes: Ah, OK, I didn't know. There will be some conflicts between those two branches 😀

mailpoet/assets/js/src/subscribers/importExport/export.ts
JanJakes: Ah, right, I misread the brackets.

mailpoet/assets/js/src/common/thumbnail.ts
websupporter: https://github.com/niklasvh/html2canvas/blob/master/src/index.ts#L19-L21
html2canvas wants explicitly an HTMLElement.

mailpoet/assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -81,7 +81,7 @@ type Props = {
 };
 
 function SetFromAddressModal({ onRequestClose, setAuthorizedAddress }: Props) {
-  const [address, setAddress] = useState(null);
+  const [address, setAddress] = useState('');
```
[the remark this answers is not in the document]
websupporter: Actually, I just tested, because I didn't like `useState<string>(null)` to much as well.
```
const [address, setAddress] = useState();
```
works just fine and imho improves readability.

mailpoet/assets/js/src/logs/list.tsx
websupporter: I agree that changing the DOM is not a perfect solution. I checked, there is no visual change though. I hesitate because of this [playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AoczAVwDsNgJa4AVJAZxgAoBKOALjgApAMoANAHQBRADZIQSWvADelOOrhEY1KM3wwOMMuQC+lKnQZM4mCBF5xV5DZqTbdcLgB4wAPi9snHAA9P7BfjwUZuRIAB6QsHAAJkiYKNQyWHakQA).

mailpoet/assets/js/src/common/listings/newsletter_status.tsx
websupporter: [Type 'string' is not assignable to type 'Element'.](https://www.typescriptlang.org/play?#code/FAYw9gdgzgLgBAIwIYCcBcApAygDQHQCiANgKYC2JE8AvHAOQwmx0DcQA)

mailpoet/assets/js/src/common/tabs/tab.tsx
JanJakes: Oh right, I think maybe we could use `allowExpressions`: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md#allowexpressions

mailpoet/assets/js/src/common/listings/newsletter_status.tsx
JanJakes: 👍

mailpoet/assets/js/src/form_editor/form_preview.ts
JanJakes: I think `<type>` is the older way to use it and the `as` was added later. Not sure why, maybe `<type>` can be confused with generics. Today, `as` seems to be more common to me.

mailpoet/assets/js/src/common/thumbnail.ts
JanJakes: 👍

mailpoet/assets/js/src/logs/list.tsx
JanJakes: This would be solved using `allowExpressions` as well (https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md#allowexpressions).

mailpoet/assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -143,7 +143,7 @@ function SetFromAddressModal({ onRequestClose, setAuthorizedAddress }: Props) {
             removeUnauthorizedEmailNotices();
             notices.success(getSuccessMessage(), { timeout: false });
           } catch (e) {
-            const error = e.errors && e.errors[0] ? e.errors[0] : null;
+            const error:error|null = e.errors && e.errors[0] ? e.errors[0] : null;
```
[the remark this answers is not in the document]
JanJakes: OK. Yep, `e` is `unknown` because at runtime (in the JS) there is no way to check for the type or catch only some types. There indeed should be a runtime check for the type you expect but `error` is TS-only type (not a runtime class) so this can't be done easily here.

mailpoet/assets/js/src/common/set_from_address_modal.tsx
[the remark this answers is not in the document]
JanJakes: In TS it is the standard way to use generics to type hooks such as `useState<StateType>()`. Note that `useState<StateType>()` will result in `StateType | undefined` while `useState<StateType>(null)` adds `null` as well. Therefore the variant without `null` (where `undefined` is the default) is usually preferred (unless `null` is needed).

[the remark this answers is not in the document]
websupporter: Thanks! Addressed in 182fc8d

mailpoet/assets/js/src/common/tabs/tab.tsx
websupporter: Addressed in [93bc295](https://github.com/mailpoet/mailpoet/pull/3968/commits/93bc295f4e34bc9d40492c77dcb4f11cbfcd9e48)

mailpoet/assets/js/src/newsletters/sending_status.jsx
```diff
@@ -76,21 +76,21 @@ const compareProps = (prev, next) => (
   && prev.params.id === next.params.id
 );
 
+const onRenderItem = (item) => (
+  <div>
+    { /* eslint-disable-next-line react/jsx-props-no-spreading */ }
+    <ListingItem {...item} />
```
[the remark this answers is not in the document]
websupporter: [e7fb8e1](https://github.com/mailpoet/mailpoet/pull/3968/commits/e7fb8e1c2c8028dddec19115046a0c2ccc0d7189) disables the rule, [e7a3a55](https://github.com/mailpoet/mailpoet/pull/3968/commits/e7a3a55fde9ccb7e0e21bedf2f749eb9cf4d40b4) removes the comments.

The premium plugin has no comments to remove and the TS eslint config already allows for props spreading.

mailpoet/assets/js/src/form_editor/components/_stories/size_settings.tsx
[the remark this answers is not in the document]
websupporter: Addressed in f996f3970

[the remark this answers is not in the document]
websupporter: Addressed in [f996f39](https://github.com/mailpoet/mailpoet/pull/3968/commits/f996f3970546be47ba63c980ff162754bc5cd58a)

mailpoet/assets/js/src/logs/list.tsx
websupporter: The problem for me originated here:
image

I tried to reproduce this in the playground, but I hadn't had any problem there. Not quite sure, what is causing it.

mailpoet/assets/js/src/newsletters/campaign_stats/page.tsx
websupporter: Addressed in 12eeced71

mailpoet/assets/js/src/date.ts
websupporter: Thanks! Good catch.

The original message was: `Unsafe argument of type `any` assigned to a parameter of type `boolean`  @typescript-eslint/no-unsafe-argument` which I attributed to the Moment declaration, where I missed the overload:
image

Instead it is related to `this.convertFormat()`returning _allegedly_ `any`, while it actually defines `string` as it's return value. I tried several things to explain this to TS but failed, that's why 04a935d66 ends up typecasting the output in the Moment constructor. Probably I am missing here something, why TS does not as expected:
image
image

mailpoet/assets/js/src/form/fields/field.jsx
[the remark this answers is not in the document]
websupporter: Addressed in 06a3ec411a9553fe26b70683c7cf61c95f01cac5

mailpoet/assets/js/src/settings/store/actions/settings.ts
websupporter: Addressed in 60f94f6ea

mailpoet/assets/js/src/sending-paused-notices-resume-button.ts
websupporter: Addressed in 60f94f6ea

mailpoet/assets/js/src/segments/dynamic/store/actions.ts
websupporter: Addressed in 60f94f6ea

mailpoet/assets/js/src/form_editor/translations.ts
websupporter: Addressed in 60f94f6ea

mailpoet/assets/js/src/form_editor/form_preview.ts
websupporter: Addressed in 60f94f6ea

mailpoet/assets/js/src/notices/notice.tsx
websupporter: Addressed in 7ca118a5364857f3ab95a64ba4783e3b82388716

mailpoet/assets/js/src/segments/dynamic/store/controls.ts
websupporter: Addressed in 2f0a9e80f

eslint-config/package.json
websupporter: Addressed in 487310f5b633a723ccc8b6cc0eb13a558231f8a6

eslint-config/package.json
websupporter: Addressed in 2f65af2473aa58b19175f37b6d5a138fce748011

mailpoet/assets/js/src/global.d.ts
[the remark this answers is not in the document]
JanJakes: MAILPOET-3140 has been merged so I guess we can drop this commit.

mailpoet/assets/js/src/global.d.ts
```diff
@@ -9,6 +9,19 @@ declare module 'mailpoet' {
...
+  export const premiumActive:boolean;
+  export const hasValidApiKey:boolean;
+  export const hasPremiumSupport:boolean;
+  export const Date: {
+    format: (dateString: string) => string;
+  };
+  export const apiVersion: string;
+  export const Ajax: {
+    post: (config: object) => any;
+  };
```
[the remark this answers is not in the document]
JanJakes: We can drop this as well, I guess (now that MAILPOET-3140 has been merged).

mailpoet/assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -79,7 +79,7 @@ type Props = {
 };
 
 function SetFromAddressModal({ onRequestClose, setAuthorizedAddress }: Props) {
-  const [address, setAddress] = useState('');
+  const [address, setAddress] = useState<string>();
```
JanJakes, line 82: Note that replacing the `('')` with `()` actually changes behavior – now it's `string | undefined` while previously it was `string` (only not typed). Not sure if it can break something.

mailpoet/assets/js/src/automation/api.tsx
```diff
@@ -63,7 +63,7 @@ export const useQuery = <T extends Data>(path: string, init?: RequestInit): Stat
 
   useEffect(
     () => {
-      void mutation();
+      void mutation(); // eslint-disable-line @typescript-eslint/no-floating-promises
```
JanJakes, line 66: This is strange. The `void` should avoid that error coming from `// eslint-disable-line @typescript-eslint/no-floating-promises`, that's why I allowed void [here](https://github.com/mailpoet/mailpoet/commit/8405659d070f110f49c55e44f84d45c56e4de703).

mailpoet/assets/js/src/sending-paused-notices-resume-button.ts
```diff
@@ -55,6 +55,7 @@ const resumeSendingIfAuthorized = (fromAddress: string | null) => isValidFromAdd
 // jQuery-based notice implementation which doesn't trigger pure-JS added listeners
 jQuery(($) => {
   $(document).on('click', '.notice .mailpoet-js-button-resume-sending', (e) : void => {
-    await resumeSendingIfAuthorized(e.target.value as string);
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    resumeSendingIfAuthorized(e.target.value as string);
```
JanJakes, line 59: And here adding `void` should help to avoid the eslint-disable, I guess.

mailpoet/assets/js/src/date.ts
JanJakes: @websupporter Is this maybe because `this.convertFormat` is a property on the Moment object? It's defined as:
```ts
export const MailPoetDate = {
  ...
  convertFormat: function convertFormat(format: string): string { ... }
  ...
}
```

In that case does it help to add `as const` to the object? Like:
```ts
export const MailPoetDate = {
  ...
  convertFormat: function convertFormat(format: string): string { ... }
  ...
} as const;
```

JanJakes (commented):
LGTM, a few more small notes.

mailpoet/assets/js/src/common/set_from_address_modal.tsx
JanJakes: Ahh, only now I see this was originally `useState(null);` – so we're just changing the default value from `null` to `undefined`, which should be fine 👍

mailpoet/assets/js/src/automation/api.tsx
websupporter: Worked. Not sure why it didn't before. [364b994](https://github.com/mailpoet/mailpoet/pull/3968/commits/364b994295e38917a7e0a01bdb855664dab0742a)

mailpoet/assets/js/src/sending-paused-notices-resume-button.ts
websupporter: Addressed in [364b994](https://github.com/mailpoet/mailpoet/pull/3968/commits/364b994295e38917a7e0a01bdb855664dab0742a)

JanJakes (approved):
@websupporter Slack is down but I see this is repushed so approving.

JanJakes:
@veljkho Since Slack is down I'll try to ask here – could we please handle this as the first thing that will be merged now, after the release, and give it a higher priority? Otherwise, we're risking conflicts and having to rebase this again, which can be non-trivial. It has also a [premium PR](https://github.com/mailpoet/mailpoet-premium/pull/541).

veljkho (approved):

PR #3970, opened 2022-02-14
Forms: per form cookies and custom expiry date [MAILPOET-3059]

brezocordero:
With this PR each form (of types popup, slide In, or Fixed bar) will have its own cookie.

It is possible to specify the expiry date in the form editor from the following values:
  - Always (session cookie)
  - 1 day
  - 3 days
  - 7 days (default)
  - 14 days
  - 30 days
  - 60 days
  - 90 days

Forms will not be displayed if the user is logged in and subscribed to any of the segments in the form

After subscription, the cookie is set with an expiration time of 182 days.

When this PR will be released, forms will be displayed again for visitors that had the previous cookie set up.

How to test:
* Create or edit a form, check that you can see the following dropdown when placement is `popup`, `slide in`, or `fixed bar`
ShowFormEvery

* Select one of the values on the dropdown, one list you are not subscribed to, any other required data and save the form.

* Log out, visit the home page, check that the form is displayed, subscribe. Check the cookie is set with 182 days expiration time.
* Remove the cookie, log in. Check the form is not displayed.
* Log out, visit the home page, check the form is displayed, dismiss the form, check the cookie is set with the expiration time selected in the editor.

Other things to test:
- Cookie with 182 days expiration time is also set on redirection after submission, with or without Captcha
- Cookie is not set when previewing the form
- Selecting `always` on the drop-down value will set a session cookie
- Cookie is removed when saving the form

[MAILPOET-3059]

[MAILPOET-3059]: https://mailpoet.atlassian.net/browse/MAILPOET-3059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/views/form/editor.html
```diff
@@ -103,6 +103,10 @@
   'formPlacementPlacementPositionLeft': _x('left', 'Placement of a slide in form, on the left or right side of the page'),
   'formPlacementPlacementPositionRight': _x('right', 'Placement of a slide in  form, on the left or right side of the page'),
   'formPlacementDelaySeconds': _x('%1s sec', 'times selection should be in the end "30 sec"'),
+  'formPlacementCookieExpiration': _x('Show form every', 'Label on a selection of different times'),
+  'formPlacementCookieExpirationAlways': _x('Always', 'times selection'),
```
brezocordero, line 107: I would have liked to use the [_nx](https://developer.wordpress.org/reference/functions/_nx/) function but didn't seem to work with our translation setup.

mailpoet/lib/Subscribers/SubscriberSubscribeController.php
```diff
@@ -153,6 +160,22 @@ public function subscribe(array $data): array {
...
   }
 
+  /**
+   * Checks if the subscriber is subscribed to any segments in the form
+   *
+   * @param  FormEntity       $form       The form entity
+   * @param  SubscriberEntity $subscriber The subscriber entity
+   * @return bool True if the subscriber is subscribed to any of the segments in the form
+   */
+  public function isSubscribedToAnyFormSegments(FormEntity $form, SubscriberEntity $subscriber): bool {
```
brezocordero, line 170: I couldn't find an existing function that would check if a subscriber is already subscribed to the segments of a form.
Let me know if there is one and I missed it!

mailpoet/lib/Form/DisplayFormInWPContent.php
```diff
@@ -160,6 +160,7 @@ private function getContentBellow(FormEntity $form, string $displayType): string
     $templateData['animation'] = $formSettings['form_placement'][$displayType]['animation'] ?? '';
     $templateData['fontFamily'] = $formSettings['font_family'] ?? '';
     $templateData['enableExitIntent'] = false;
+    $templateData['cookieFormExpirationTime'] = $formSettings['cookie_expiration_time'] ?? 365;
```
wxa: Is the value of 365 only for backwards compatibility? It's not possible to set such a value from the UI.

mailpoet/lib/Subscribers/SubscriberSubscribeController.php
wxa: I also could not find such a method. Maybe we had something in the past, but after the new form editor was launched, it's gone. Thanks for adding it!

mailpoet/views/form/editor.html
wxa: You're probably right. Even if you add it to template [i18n functions](https://github.com/mailpoet/mailpoet/blob/master/mailpoet/lib/Twig/I18n.php#L29), our makepot [extractor](https://github.com/mailpoet/mailpoet/blob/master/mailpoet/tasks/makepot/extract.php) may not be able to pick it up. And rewriting its regular expressions is a task not many would dare to do.

wxa (approved):
Looks good, works for me too! I have only insignificant remarks which shouldn't block the approval.

mailpoet/lib/Form/DisplayFormInWPContent.php
brezocordero: I switched 365 to 7 (new default) on a later commit and added a comment about the backwards compatibility.
Sorry for the confusion.

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #3977, opened 2022-02-16
Automation bootstrapping [MAILPOET-4135]

JanJakes:
[MAILPOET-4135]

This is an initial bootstrap for the **automation** feature. It includes:

1. Feature flag.
2. Admin page.
3. PHP namespace, initializer, DI.
4. Stricter PHPStan and CodeSniffer check as it will be a new code.
5. Migrator (some simple way to create/delete/recreate related tables).
6. REST API (based on WP REST API).
7. Exceptions (a new set compatible with WP REST API).
8. React app skeleton and basic API calling functionality.
9. Action Scheduler initialization.

It doesn't include yet:
1. Tests.
2. REST authorization.
3. Data validation.
4. Any automation functionality.

I recommend going commit-by-commit as I tried to make titles explanatory. Some commits include more detailed info as well.

[MAILPOET-4135]: https://mailpoet.atlassian.net/browse/MAILPOET-4135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/eslint-config/.eslintrc.ts.json
+++ b/eslint-config/.eslintrc.ts.json
@@ -54,6 +54,7 @@
     "no-use-before-define": 0, //use the typescript version
     "camelcase": 0, //use the typescript version
     "no-shadow": 0, //use the typescript version
+    "no-void": 0, // can conflict with @typescript-eslint/no-floating-promises
     "@typescript-eslint/naming-convention": [
       "error",
       { "selector": "classProperty", "format": [] },
--- a/mailpoet/assets/js/src/automation/api.tsx
+++ b/mailpoet/assets/js/src/automation/api.tsx
@@ -1,0 +1,64 @@
+import { useCallback, useEffect, useState } from 'react';
+
+const API_URL = '/wp-json/mailpoet/v1/automation';
+
+export const request = (path: string, init?: RequestInit): ReturnType<typeof fetch> => (
+  fetch(`${API_URL}/${path}`, init)
+);
+
+type Error<T> = {
+  response?: Response;
+  data?: T;
+};
+
+type State<T> = {
+  data?: T;
+  loading: boolean;
+  error?: Error<T>;
+};
+
+type Result<T> = [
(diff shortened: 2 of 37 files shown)


JanJakes:
@brezocordero

> I have a question about the API structure. Maybe it is too early for this, or it conflicts with the architecture we want to implement, but WP recommends using https://developer.wordpress.org/reference/classes/wp_rest_controller/

> Could we use the controller class? As an example, this is how WC uses it: https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/includes/rest-api

Good question. I followed the [Adding Custom Endpoints](https://developer.wordpress.org/rest-api/extending-the-rest-api/adding-custom-endpoints/) guide, and controllers have only a small mention at the bottom. When I got there, I had the basic setup almost done, and I decided to keep it that way for now. It's good to have input from someone more experienced with WP as I may not be aware of everything it provides.

I was, too, thinking that maybe we should use the controllers in the future but didn't want to spend more time on it now. They seem to solve some things but also bring some more complexity (the class has quite many responsibilities), so it may be nice to wrap them to something simpler to restrict ourselves a bit (as I tried to do with Request/Response/Endpoint).

That said, I do think we should definitely have a close look at them and probably adopt them (as they implement some features such as field selection, etc.). Not sure if we should use inheritance or composition (to be more restricted or simplify the interfaces). The example of WooCommerce may be very useful as well, thanks for that.

Hovewer, there will be an important consideration — the split between the free and the premium plugins. If we strive for a RESTful API (which I guess we should to go the WP way), then we'll probably need to have the ability to put methods for the same endpoint in different plugins (we may have `GET /worfklows` in the free plugin and `POST`/`PUT`/`DELETE` in premium). This can have some implications about whether or how we can use the controllers.

veljkho (approved):

PR #3978, opened 2022-02-17
Debounce dynamic segment calculations [MAILPOET-4018]

johnolek:
This PR alters the behavior of dynamic segment calculations to reduce the number of AJAX requests and, in turn, the amount of load we're putting on the database.

image
(the dynamic segment calculation page)

## Old behavior
A new admin-ajax.php request to the `dynamic_segments` endpoint's `getCount` method is initiated immediately every time the segment's configuration is changed.

This was causing issues when any of the calculations were long-running. The requests would keep piling up.

## New behavior
 A new request to will only be initiated 2 seconds after a user stops making changes. 

For example, a user could make 20 changes to the segment's configuration, and as long as there's no more than 2 seconds in between individual changes there will only be 1 ALAX request and it will start 2 seconds after the 20th change.

## Handling simultaneous requests
If a user still ends up with multiple simultaneous requests, this PR ensures we only display the results for the last one. 

For example, let's say a user initiates a complex request that takes 15 seconds to calculate. They then make a change  that make the calculation much simpler and only takes 1 second to complete. Even though the ALAX request won't start until 2 seconds later, it will complete before the first.

### Old behavior
The number of subscribers gets updated in the DOM after request 2 completes. ~12 seconds later request 1 completes and updates the DOM to show the count from that request, creating a mismatch between the configuration shown and the calculated subscriber number.

### New behavior
Only the latest request will cause an update to the subscriber count. The responses of older requests that complete later will be ignored.

## Notes
I decided to bypass the debounced `load` for the initial page load because I don't think it makes sense to wait at that point.

## Draft Status
This is a draft because I'm positive there's a better approach. I tried multiple approaches before this one, and I couldn't get any of them to work quite right. 

The debounced function needs to **not** be redefined every time the component re-renders. I tried using react's `useMemo` inside the component, but because `load` depends on variables that do change on every render, the function kept getting redefined anyway.

My react experience is mostly from back when class components were all the rage, and I've spent enough time on this now that I think it's best to just have someone with more experience point me in the right direction.

[MAILPOET-4018]


mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx
```diff
@@ -24,34 +24,62 @@ const SubscribersCounter: React.FunctionComponent = () => {
...
+    latestRequestIdRef.current += 1;
+    const requestId = latestRequestIdRef.current;
+    isRequestInFlight.current = true;
+
+    loadCount(loadItem).then((response) => {
+      isRequestInFlight.current = false;
+      if (deferredRequestRef.current) {
+        load(deferredRequestRef.current);
+        return;
+      }
```
JanJakes, line 49: I'm wondering whether the deferring logic is needed. It seems to me `debounce` should handle that already, shouldn't it? Or is it because the requests are very slow? It adds quite a bit of complexity but I guess there's not much we can do as we don't have a better logic in the codebase for this (like request cancellation).

mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx
```diff
@@ -24,34 +24,62 @@ const SubscribersCounter: React.FunctionComponent = () => {
...
+      if (requestId !== latestRequestIdRef.current) {
+        // Don't do anything with the response because a newer request has been initiated
+        return;
+      }
```
JanJakes, line 53: I don't have the context of what happens in the UI when the data is updated but maybe we could do `if (requestId < latestRequestIdRef.current)` here to avoid outdated requests doing anything but display some changes for non-outdated ones so at least some changes are displayed? Not sure, maybe not too important.

JanJakes (commented):
Looks good to me, I left just some small comments/questions. Generally, we should probably consider trying to adopt some libraries that can do this natively but that's a bigger task so for now, this fix seems good to me. (We'll likely need similar logic for the automation feature so I guess we'll explore more in there.)

mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx
johnolek: @JanJakes I went back and forth on this myself. My [original idea](https://mailpoet.atlassian.net/browse/MAILPOET-4018?focusedCommentId=21914) was that we could use deferring instead of debouncing. That way fast-to-calculate counts would come back right away, but with long-running queries there could only ever be one request at a time. We also wouldn't need to keep track of the order of requests.

@costasovo [thought](https://mailpoet.atlassian.net/browse/MAILPOET-4018?focusedCommentId=21915) it could be best to combine these two approaches, but I do think having both adds more complexity than its worth.

Combined, deferring only protects against what is probably an edge case where someone is making many changes to a dynamic segment that takes a very long time to calculate. Our timeout for these requests is 20 seconds, so worst case scenario, if you were very carefully pausing every 2 seconds or so, you could have around 10 simultaneous Ajax requests.

If the query itself is extremely inefficient and takes significantly more than 20 seconds to execute then this could be a slightly bigger deal. 

So I think we should go with just one or the other, and here are my thoughts I jotted down while considering both:

## Deferred Only

### Pros
- No dependency on third party library for `debounce`
- No need to track order of requests because they will only ever be made sequentially
- Users potentially see feedback much faster
- Less potential for initiating many simultaneous, long-running queries
- Potentially fewer total queries when working on segments with very long running count calculations

### Cons
- Will always execute at least 2 queries when user makes many changes in quick succession
- Could result in a much higher number of total queries if calculations are fast
- Could take longer to get results when going from a very complex segment to a very simple one (the complex query that would start immediately would need to complete before the simple one could start.

## Debounce Only

### Pros
- Potentially much fewer total DB queries when the calculations are fast
- User might see results faster when changing from a complex segment to a very simple one

### Cons
- User always has to wait at least 2 seconds to see a new count
- Easier to generate simultaneous long-running queries

Deciding between these choices could require some additional information
1. How long are the long-running queries that have been causing issues?
2. Is total query count something we care about?
3. What's the average amount of time it takes to calculate these counts? More or less than 2 seconds?
4. Is it more common for existing segments to increase or decrease in complexity when users edit them?

I think I'm overthinking this. 

I tried both ways and pushed the branches. Here are comparisons with this PR for reference.
[Deferred only](https://github.com/mailpoet/mailpoet/compare/master...update/MAILPOET-4018/deferred-only?expand=1)
[Debounce only](https://github.com/mailpoet/mailpoet/compare/master...update/MAILPOET-4018/deferred-only?expand=1)

I still think I prefer the defer-only approach, but I'd like go defer (🤣) to someone with more experience with this part of the app.

mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx
```diff
@@ -24,34 +24,62 @@ const SubscribersCounter: React.FunctionComponent = () => {
...
+      const finished = {} as SubscriberCount;
+      const errors = errorResponse.errors.map((error) => error.message);
+      finished.loading = false;
+      finished.count = undefined;
+      finished.errors = errors;
+      updateSubscriberCount(finished);
+    });
+  }
+
+  const debouncedLoadRef = useRef(debounce(load, 2000, { trailing: true }));
```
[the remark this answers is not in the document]
johnolek: My understanding is that `useCallback` expects an inline function, not a higher order function like a debounce function. There's a good discussion about it here: https://github.com/facebook/react/issues/19240#issuecomment-652945246. I was getting `error  React Hook useCallback received a function whose dependencies are unknown. Pass an inline function instead  react-hooks/exhaustive-deps`.

In my branch that only includes the debounce logic, I had to use `useCallback` for `load` as well as pass in updateSubscriberCount as a paramater (otherwise it would need to be in the dependency array and it was causing `load` to be different every time defeating the purpose), and then `useMemo` for `debouncedLoad`, in order to avoid an eslint error: https://github.com/mailpoet/mailpoet/blob/ca0d7b16227bd4ee022bf9787d2d7015a3c56e33/mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx#L34-L34

Here's one other place where we had to ignore a linting rule for a debounced function: https://github.com/mailpoet/mailpoet/blob/ca0d7b16227bd4ee022bf9787d2d7015a3c56e33/mailpoet/assets/js/src/form_editor/blocks/html/edit.jsx#L49-L49

Maybe it would be best to do that here as well so I can use the better `useCallback` approach? It just seems weird to have to ignore a rule like that.

mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx
johnolek: I'm not sure I completely follow. How would we get into a situation where there would be a difference between `requestId < latestRequestIdRef.current` and `requestId !== latestRequestIdRef.current`?

In any case, because this `return`s before calling `updateSubscriberCount`, the UI should continue to display the `Calculating segment size…` placeholder text until the latest request finishes, which makes sense to me.

I also [moved the code](http://github.com/mailpoet/mailpoet/blob/update/MAILPOET-4018/debounce_dynamic_segments/mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx#L76-L76) that starts displaying that placeholder text outside of `load`, because I wanted to avoid confusing situations like this:

1. User makes change at t=0 that will complete at t=3 (2 seconds delay + 1 second for the Ajax request)
2. User makes another change at t=2.5, which won't trigger the debounced `load` until t=4.5
3. At t=3, the first request comes back and updates the count. It would look like this number represents the count for the configuration on the screen
4. At t=4.5, 1.5 seconds later, the second debounced `load` function fires, and all of a sudden the form is recalculating without the user having made a change.

I could have moved out the incrementing of `latestRequestIdRef.current` instead  and passed the request ID as a parameter into `load`, but I thought it would be a better user experience to start showing "Calculating segment size…" as soon as a user does something that we know 100% will result in us eventually calculating the segment size. It's immediate feedback even though we know nothing will happen for at least 2 seconds. The only downside I could see is that it could make it look like all of our calculations are really slow.

mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx
JanJakes: Thanks for the detailed answer, I guess we can also keep both approaches, I see it seems to make sense as well.

mailpoet/assets/js/src/segments/dynamic/subscribers_counter.tsx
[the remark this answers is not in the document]
JanJakes: Oh, right, I missed that. If we don't expect `load` to be ever redefined even on a new component mount, then we can keep the `useRef`, I think.

JanJakes (approved):
LGTM, looks well thought out and tested so I'm OK with the current form 👍

JanJakes:
I didn't pass this to QA yet as we're two reviewers assigned here and maybe @lysyjan has some more observations.

lysyjan (approved):

rohankacharegs:
GlobalStep Verified

PR #3984, opened 2022-02-23
Fix Typescript and eslint warnings in cookie_settings.tsx [MAILPOET-4159]

costasovo:
There was an issue in argument type in the string replace function. Also, it seems that after changes introduced in [update eslint PR](https://github.com/mailpoet/mailpoet/pull/3968) we need to declare components using the function keyword see https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md.

[MAILPOET-4159]

[MAILPOET-4159]: https://mailpoet.atlassian.net/browse/MAILPOET-4159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brezocordero (approved):
Looks good! Thanks for the fix.

PR #3989, opened 2022-02-25
Add notice email limit reached [MAILPOET-4047]

lysyjan:
[MAILPOET-4047]

[MAILPOET-4047]: https://mailpoet.atlassian.net/browse/MAILPOET-4047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/global.d.ts
+++ b/mailpoet/assets/js/src/global.d.ts
@@ -90,4 +90,6 @@ interface Window {
   mailpoet_premium_plugin_download_url: string;
   mailpoet_premium_plugin_activation_url: string;
   mailpoet_plugin_partial_key: string;
+  mailpoet_email_volume_limit: string;
+  mailpoet_email_volume_limit_reached: string;
 }
--- a/mailpoet/assets/js/src/global.d.ts
+++ b/mailpoet/assets/js/src/global.d.ts
@@ -91,5 +91,5 @@ interface Window {
   mailpoet_premium_plugin_activation_url: string;
   mailpoet_plugin_partial_key: string;
   mailpoet_email_volume_limit: string;
-  mailpoet_email_volume_limit_reached: string;
+  mailpoet_email_volume_limit_reached: boolean;
 }
--- a/mailpoet/assets/js/src/listing/notices.jsx
+++ b/mailpoet/assets/js/src/listing/notices.jsx
@@ -29,6 +29,10 @@ function MailerError(props) {
   if (props.mta_log.error.operation === 'send' && props.mta_method === 'MailPoet' && MailPoet.hasInvalidMssApiKey) {
     return null;
   }
+  // do not display Email Volume Limit reached error twice
+  if (props.mta_method === 'MailPoet' && props.mta_log.error.operation === 'email_limit_reached') {
+    return null;
+  }
   if (props.mta_log.error.operation === 'migration') {
     return (
       <div className="mailpoet_notice notice notice-warning">
--- a/mailpoet/assets/js/src/notices/email_volume_limit_notice.tsx
+++ b/mailpoet/assets/js/src/notices/email_volume_limit_notice.tsx
@@ -1,0 +1,68 @@
+import React from 'react';
+import ReactStringReplace from 'react-string-replace';
+import ReactHtmlParser from 'react-html-parser';
+import MailPoet from 'mailpoet';
+import Notice from 'notices/notice';
+
+function EmailVolumeLimitNotice(): JSX.Element {
+  if (!MailPoet.emailVolumeLimitReached) return null;
+
+  const title = MailPoet.I18n.t('emailVolumeLimitNoticeTitle')
+    .replace('[emailVolumeLimit]', MailPoet.emailVolumeLimit);
+  const youReachedEmailVolumeLimit = MailPoet.I18n.t('youReachedEmailVolumeLimit')
+    .replace('[emailVolumeLimit]', MailPoet.emailVolumeLimit);
+  const upgradeLink = `https://account.mailpoet.com/orders/upgrade/${MailPoet.pluginPartialKey}`;
+  const refreshSubscribers = async () => {
+    await MailPoet.Ajax.post({
+      api_version: MailPoet.apiVersion,
+      endpoint: 'services',
+      action: 'recheckKeys',
+    });
+    window.location.reload();
+  };
+
+  const date = new Date();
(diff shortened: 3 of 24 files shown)


mailpoet/lib/Cron/Workers/KeyCheck/PremiumKeyCheck.php
```diff
@@ -25,6 +25,9 @@ public function checkProcessingRequirements() {
   }
 
   public function checkKey() {
+    // for phpstan because we set bridge property in the init function
+    if (!$this->bridge) return;
```
costasovo, line 29: What happens when someone forgets to call the init function? I mean wouldn't it be safer to throw an exception? I think that with this fix we actually allowed using the method when the class is not properly initialized.

mailpoet/assets/js/src/global.d.ts
```diff
@@ -90,4 +90,6 @@ interface Window {
   mailpoet_premium_plugin_download_url: string;
   mailpoet_premium_plugin_activation_url: string;
   mailpoet_plugin_partial_key: string;
+  mailpoet_email_volume_limit: string;
+  mailpoet_email_volume_limit_reached: string;
```
costasovo, line 94: The variable name sounds like this one should be boolean. Is the type correct?

mailpoet/assets/js/src/notices/email_volume_limit_notice.tsx
```diff
@@ -0,0 +1,68 @@
...
+import Notice from 'notices/notice';
+
+function EmailVolumeLimitNotice(): JSX.Element {
+  if (!MailPoet.emailVolumeLimitReached) return null;
+
+  const title = MailPoet.I18n.t('emailVolumeLimitNoticeTitle')
+    .replace('[emailVolumeLimit]', MailPoet.emailVolumeLimit);
+  const youReachedEmailVolumeLimit = MailPoet.I18n.t('youReachedEmailVolumeLimit')
+    .replace('[emailVolumeLimit]', MailPoet.emailVolumeLimit);
+  const upgradeLink = `https://account.mailpoet.com/orders/upgrade/${MailPoet.pluginPartialKey}`;
```
costasovo, line 14: We have [a factory for building such links](https://github.com/mailpoet/mailpoet/blob/3819dd774e53505616b548236b127d51db7230b4/mailpoet/assets/js/src/mailpoet_com_url_factory.js). It is available on [MailPoet.MailPoetComUrlFactory](https://github.com/mailpoet/mailpoet/blob/18b7139bec772fd4d4deef53fab7b87728062555/mailpoet/assets/js/src/mailpoet.ts#L17).
I think we should create the link using the factory since it adds a referral id that we want to have for every link that goes to the shop.

Saying that realized that we probably don't add the referral id in t[he messages we create in the MSS error mapper](https://github.com/mailpoet/mailpoet/blob/3819dd774e53505616b548236b127d51db7230b4/mailpoet/lib/Mailer/Methods/ErrorMappers/MailPoetMapper.php#L104-L151). So I guess we should add it there as well. Probably in a different task. Or maybe we could as @NeosinneR if the referral id is still needed. I haven't heard about the referrals for some time so maybe we could think of removing this functionality.

mailpoet/lib/API/JSON/v1/Services.php
```diff
@@ -211,6 +211,11 @@ public function recheckKeys() {
     $this->mssWorker->checkKey();
     $this->premiumWorker->init();
     $this->premiumWorker->checkKey();
+    // continue sending when it is paused
+    $mailerLog = MailerLog::getMailerLog();
+    if (isset($mailerLog['status']) && $mailerLog['status'] === MailerLog::STATUS_PAUSED) {
+      MailerLog::resumeSending();
+    }
```
costasovo, line 218: Is it safe to always resume the sending? Shouldn't we resume the sending based on the status returned by those checks?

mailpoet/assets/js/src/listing/notices.jsx
```diff
@@ -29,6 +29,10 @@ function MailerError(props) {
   if (props.mta_log.error.operation === 'send' && props.mta_method === 'MailPoet' && MailPoet.hasInvalidMssApiKey) {
     return null;
   }
+  // do not display Email Volume Limit reached error twice
+  if (props.mta_method === 'MailPoet' && props.mta_log.error.operation === 'email_limit_reached') {
+    return null;
+  }
```
costasovo, line 35: After hiding this duplicate message is it still necessary to have it? I mean it is not quite complex code to [create it in the MSS Error mapper](https://github.com/mailpoet/mailpoet/pull/3989/commits/8465035f3ce58ba1cb46156ab3f6e3969d0f228e#diff-a6737f295bd861863026c7f41274b9818006adda1f7683f5c4e374d3d52ae8beR177-R196) and now it seems to me that we will hide it anyway. Or is there some case when it will be displayed?

lysyjan: The error message is visible on the system status page. However, I'm not sure if it is necessary. What do you think @NeosinneR ?

mailpoet/lib/Cron/Workers/KeyCheck/PremiumKeyCheck.php
lysyjan: The exception was my second idea for this. I haven't changed the behavior, but I have wanted to add the annotation for the bridge property. I can use an exception here if you think that it is better.

mailpoet/assets/js/src/global.d.ts
lysyjan: True, I will fix it.

mailpoet/lib/API/JSON/v1/Services.php
lysyjan: Good point! I will change it.

mailpoet/assets/js/src/listing/notices.jsx
NeosinneR: If we already show an error message when the subscriber limit is reached on the System Status page, I'd also show an error message when an email volume limit is reached. Because both error messages indicate that there is something wrong with your plan or sending, the System Status page seems like a place to look for the reason for such an error.

costasovo: Aha. I forgot about the status page. 👍  Lets keep the message then

veljkho (approved):

PR #3997, opened 2022-03-04
Process arguments better when querying posts for automated latest content [MAILPOET-4082]

samnajian:
Fixes [MAILPOET-4082]

[MAILPOET-4082]: https://mailpoet.atlassian.net/browse/MAILPOET-4082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/JSON/v1/AutomatedLatestContent.php
+++ b/mailpoet/lib/API/JSON/v1/AutomatedLatestContent.php
@@ -4,23 +4,33 @@ namespace MailPoet\API\JSON\v1;
 
 use MailPoet\API\JSON\Endpoint as APIEndpoint;
 use MailPoet\Config\AccessControl;
+use MailPoet\Newsletter\AutomatedLatestContent as ALC;
+use MailPoet\Util\APIPermissionHelper;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoet\WP\Posts as WPPosts;
 
 class AutomatedLatestContent extends APIEndpoint {
-  /** @var \MailPoet\Newsletter\AutomatedLatestContent  */
+  /** @var ALC  */
   public $ALC;
+
+  /*** @var WPFunctions */
   private $wp;
+
+  /*** @var APIPermissionHelper */
+  private $permissionHelper;
+
   public $permissions = [
     'global' => AccessControl::PERMISSION_MANAGE_EMAILS,
   ];
 
   public function __construct(
-    \MailPoet\Newsletter\AutomatedLatestContent $alc,
+    ALC $alc,
+    APIPermissionHelper $permissionHelper,
     WPFunctions $wp
   ) {
     $this->ALC = $alc;
     $this->wp = $wp;
+    $this->permissionHelper = $permissionHelper;
   }
 
   public function getPostTypes() {
@@ -65,14 +75,24 @@ class AutomatedLatestContent extends APIEndpoint {
     return $this->successResponse(array_values($terms));
   }
 
+  /**
+   * @param \WP_Post[] $posts
+   * @return \WP_Post[]
+   */
+  private function getPermittedPosts($posts) {
+    return array_filter($posts, function ($post) {
+      return $this->permissionHelper->checkReadPermission($post);
+    });
+  }
+
   public function getPosts($data = []) {
     return $this->successResponse(
-      $this->ALC->getPosts($data)
+      $this->getPermittedPosts($this->ALC->getPosts($data))
     );
   }
 
   public function getTransformedPosts($data = []) {
-    $posts = $this->ALC->getPosts($data);
+    $posts = $this->getPermittedPosts($this->ALC->getPosts($data));
     return $this->successResponse(
       $this->ALC->transformPosts($data, $posts)
     );
@@ -83,7 +103,7 @@ class AutomatedLatestContent extends APIEndpoint {
     $renderedPosts = [];
 
     foreach ($data['blocks'] as $block) {
-      $posts = $this->ALC->getPosts($block, $usedPosts);
+      $posts = $this->getPermittedPosts($this->ALC->getPosts($block, $usedPosts));
       $renderedPosts[] = $this->ALC->transformPosts($block, $posts);
 
       foreach ($posts as $post) {
--- a/mailpoet/lib/API/JSON/v1/AutomatedLatestContent.php
+++ b/mailpoet/lib/API/JSON/v1/AutomatedLatestContent.php
@@ -3,6 +3,7 @@
 namespace MailPoet\API\JSON\v1;
 
 use MailPoet\API\JSON\Endpoint as APIEndpoint;
+use MailPoet\API\JSON\SuccessResponse;
 use MailPoet\Config\AccessControl;
 use MailPoet\Newsletter\AutomatedLatestContent as ALC;
 use MailPoet\Newsletter\BlockPostQuery;
@@ -77,29 +78,28 @@ class AutomatedLatestContent extends APIEndpoint {
   }
 
   /**
-   * @param \WP_Post[] $posts
-   * @return \WP_Post[]
+   * Fetches posts for Posts static block
    */
-  private function getPermittedPosts($posts) {
-    return array_filter($posts, function ($post) {
-      return $this->permissionHelper->checkReadPermission($post);
-    });
-  }
-
-  public function getPosts($data = []) {
+  public function getPosts(array $data = []): SuccessResponse {
     return $this->successResponse(
-      $this->getPermittedPosts($this->ALC->getPosts(new BlockPostQuery(['args' => $data])))
+      $this->getPermittedPosts($this->ALC->getPosts(new BlockPostQuery(['args' => $data, 'dynamic' => false])))
     );
   }
 
-  public function getTransformedPosts($data = []) {
+  /**
+   * Fetches products for Abandoned Cart Content dynamic block
+   */
+  public function getTransformedPosts(array $data = []): SuccessResponse {
     $posts = $this->getPermittedPosts($this->ALC->getPosts(new BlockPostQuery(['args' => $data])));
     return $this->successResponse(
       $this->ALC->transformPosts($data, $posts)
     );
   }
 
-  public function getBulkTransformedPosts($data = []) {
+  /**
+   * Fetches different post types for ALC dynamic block
+   */
+  public function getBulkTransformedPosts(array $data = []): SuccessResponse {
     $usedPosts = [];
     $renderedPosts = [];
 
@@ -114,5 +114,15 @@ class AutomatedLatestContent extends APIEndpoint {
     }
 
     return $this->successResponse($renderedPosts);
+  }
+
+  /**
+   * @param \WP_Post[] $posts
+   * @return \WP_Post[]
+   */
+  private function getPermittedPosts($posts) {
+    return array_filter($posts, function ($post) {
+      return $this->permissionHelper->checkReadPermission($post);
+    });
   }
 }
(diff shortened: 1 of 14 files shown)


mailpoet/lib/Newsletter/AutomatedLatestContent.php
```diff
@@ -65,8 +65,13 @@ public function getPosts($args, $postsToExclude = [], $newsletterId = false, $ne
       'post_type' => (isset($args['contentType'])) ? $args['contentType'] : 'post',
       'post_status' => (isset($args['postStatus'])) ? $args['postStatus'] : 'publish',
       'orderby' => 'date',
-      'order' => ($args['sortBy'] === 'newest') ? 'DESC' : 'ASC',
+      'order' => isset($args['sortBy']) && in_array($args['sortBy'], ['newest', 'DESC']) ? 'DESC' : 'ASC',
     ];
+
+    if (!$this->wp->isPostTypeViewable($parameters['post_type'])) {
+      throw new \InvalidArgumentException($this->wp->__('Invalid request parameters', 'mailpoet'));
```
JanJakes, line 75: I think we should throw here one of the [exceptions in the MailPoet namespace](https://github.com/mailpoet/mailpoet/blob/master/mailpoet%2Flib%2Fexceptions.php). Picking the correct one will also result in the correct HTTP code on the API (which I think should be 4xx, in this case). Now, this would probably result in error 500, which is not desired.

I'd suggest using this one:
```php
throw AccessDeniedException::create()
  ->withErrors([APIError::BAD_REQUEST => __('...', 'mailpoet')]);
```

Also the message could reflect that it's about access denied, I guess.

mailpoet/lib/WP/Functions.php
```diff
@@ -776,4 +776,8 @@ public function escUrlRaw(string $url, array $protocols = null): string {
   public function restUrl(string $path = '', string $scheme = 'rest'): string {
     return rest_url($path, $scheme);
   }
+
+  public function isPostTypeViewable(string $post_type): bool {
```
JanJakes, line 780: This is a tiny issue but the argument should be camel-cased (`$postType`).

mailpoet/lib/WP/Functions.php
```diff
@@ -780,4 +780,8 @@ public function restUrl(string $path = '', string $scheme = 'rest'): string {
   public function isPostTypeViewable(string $post_type): bool {
     return is_post_type_viewable($post_type);
   }
+
+  public function isPostStatusViewable(string $post_status): bool {
```
JanJakes, line 784: Same here (camel case).

mailpoet/tests/integration/Newsletter/Blocks/AbandonedCartContentTest.php
```diff
@@ -78,6 +78,14 @@ public function _before() {
       $this->wp->wpDeletePost((int)$product->ID);
     }
 
+
+    if (class_exists('\WC_Post_Types')) {
+      \WC_Post_Types::register_post_types();
```
JanJakes, line 83: I think we don't have WC in integration tests active at the moment so I'd remove the IF part and leave just the `else` branch now. What do you think?

mailpoet/lib/Newsletter/AutomatedLatestContent.php
```diff
@@ -65,8 +65,13 @@ public function getPosts($args, $postsToExclude = [], $newsletterId = false, $ne
       'post_type' => (isset($args['contentType'])) ? $args['contentType'] : 'post',
       'post_status' => (isset($args['postStatus'])) ? $args['postStatus'] : 'publish',
       'orderby' => 'date',
-      'order' => ($args['sortBy'] === 'newest') ? 'DESC' : 'ASC',
+      'order' => isset($args['sortBy']) && in_array($args['sortBy'], ['newest', 'DESC']) ? 'DESC' : 'ASC',
```
JanJakes, line 68: Was this broken? I don't know what the order params look like but is this a fix? It's probably something I'd normally extract to another commit to describe the fix in the commit message but of course no need to do it now.

JanJakes:
I'm also adding @websupporter for a review here as he was the one who reported this issue. Pls @websupporter, could you have a look if the added checks look good to you? Thanks!

mailpoet/lib/Newsletter/AutomatedLatestContent.php
samnajian: It was not broken per se, but it was wrong, it could easily force `DESC` to be interpreted as `ASC` and besides not checking if `sortBy` was set, I can put it to another commit

mailpoet/tests/integration/Newsletter/Blocks/AbandonedCartContentTest.php
samnajian: Having it here doesn't do any harms, and for now it also implies that we're better off having WC active but it may create benefits in future, I guess some of the flakey tests come from not using the correct functions to register stuff like this, this is only a guess though

JanJakes: @samnajian Right, but on the other hand, I don't think it's a good practice to have different possible flows in tests (if/else based on some circumstances). Again, WC is currently not there so it doesn't make sense to have an IF "in case it is there". Also (noticing just now), the post types should be removed in `cleanup`.

samnajian: I see your point, fine removing the other branch :)

JanJakes (approved):

websupporter:
I am not sure, if I would add those checks to `Newsletter/AutomatedLatestContent`. Imo the classes responsibility is to return posts based on query arguments. `API/JSON/v1/AutomatedLatestContent` on the other hand is responsible of returning this in a specific context. I think the checks should be there.

The current implementation breaks some features:
The endpoint is used to fetch drafts/scheduled and private posts in the Email editor:
image
image

Imo that makes sense: I draft a post. I draft a newsletter. I publish post. I send newsletter. Same for schedule and also private. So the evaluation should imo be on a user basis and not on the post status basis.

mailpoet/lib/API/JSON/v1/AutomatedLatestContent.php
```diff
@@ -65,14 +75,24 @@ public function getTerms($data = []) {
     return $this->successResponse(array_values($terms));
   }
 
+  /**
+   * @param int[]|\WP_Post[] $posts
+   * @return \WP_Post[]
+   */
+  private function getPermittedPosts($posts) {
+    return array_filter($posts, function ($post) {
+      return !is_numeric($post) && $this->permissionHelper->checkReadPermission($post);
```
websupporter, line 84: Why `!is_numeric()`? I mean, why expect numeric values at all? I know `get_posts()` can return `int[]` with `["fields" => "ids"]` which we don't use though. Maybe we could make sure, this does not happen for posts returned from ACL, so that we do not have this ambiguity trickling down in our code base?

samnajian: @websupporter since we were are not preventing the int[] output, this is needed per type definitions! not sure if we want to prevent this behaviour now, I can create a Jira ticket to check this in depth and probably refactor, what do you think?

websupporter (dismissed):
lgtm

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
```diff
@@ -0,0 +1,228 @@
...
+        return $post;
+      },
+    ]);
+    $apiPermissionHelper = $this->diContainer->get(APIPermissionHelper::class);
+    $this->alcAPI = new ALCAPI($alc, $apiPermissionHelper, $this->wp);
+
+  }
+
+  public function tearDown(): void {
+    parent::tearDown();
```
JanJakes, line 103: Is there a difference from using `_after`? I don't remember now...

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  private function loginWithRole(string $role): void {
+    $username = uniqid("testUser");
+    $email = "$username@test.com";
+    $existingUser = $this->wp->getUserBy("email", $email);
+
+    if ($existingUser) {
+      wp_delete_user($existingUser->ID);
+    }
```
JanJakes, line 119: Maybe this could be in before/after cleanups?

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
```diff
@@ -0,0 +1,228 @@
...
+    ];
+    $draftPost = [
+      'post_title' => 'Draft Post',
+      'post_content' => 'contents',
+      'post_status' => 'draft',
+    ];
+
+    array_map(function ($postarr){
+      wp_insert_post($postarr);
+    }, [$privatePost, $publicPost, $draftPost]);
```
JanJakes, line 206: Running `$privatePost = wp_insert_post([...])` directly would make it more consistent with the other test and would allow to get rid of this `array_map`.

mailpoet/lib/Util/APIPermissionHelper.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace MailPoet\Util;
+
+// \WP_REST_Posts_Controller is always available in WP runtime but not when building the plugin outside of WP env
+if (class_exists("\WP_REST_Posts_Controller")) {
```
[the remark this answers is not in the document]
samnajian: > Could we maybe do this (see below) instead of the conditional class definition?
> 
> ```
> require_once ABSPATH . '/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php';
> ```
> 
> We [already have a couple of these](https://github.com/mailpoet/mailpoet/search?q=require_once+ABSPATH). I'm curious how DI dump actually works when the class is only conditional 🤔

I actually tried the `require_once` approach first, but PHPStan was not happy with it, this if statement approach works and I've verified the `FreeCachedContainer ` has `APIPermissionHelper ` in the build artifact, but still would be happy to learn if you think this approach is error prone

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
samnajian: tearDown is after and `_before` is well before, na?

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
samnajian: No, this is done to make sure this user is not existing before trying to add it, but I'll add user cleanup anyway

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
JanJakes: I'm not sure I understand your answer but my question could be reformulated as "should we maybe use `_after` here?".

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
samnajian: 1) In this test I need access to `post_title` of all 3 created posts, removing these variables would mean I'll need to create another 3 variables which doesn't make any difference .
2) I'm not really sure what value  `consistency between different tests` can bring to us, nor do I know how getting rid of `array_map` can serve us

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
JanJakes: > No
Well, it can be in `_before` but I guess the username needs to be fixed and not random. It was a rhetorical question as we don't usually do these things in the tests themselves.

mailpoet/tests/integration/Newsletter/AutomatedLatestContentAPITest.php
JanJakes: 1. Right, that explains it.
2. Not sure about the need for such a comment, I'm only trying to simplify code.

websupporter (dismissed):
lgtm

JanJakes (approved):

veljkho (approved):

PR #3998, opened 2022-03-07
Button says "Send" instead of "Schedule" even if the schedule time is in the future MAILPOET-4163

samnajian:
### This PR fixed the issue with scheduling newsletters for future/past dates/times.

This PR initially makes some cleanup and then applies the fix as follows:
- Define NewsLetter type
- Define component props type
- Add acceptance test for past/future time

### Testing:
- Pick a newsletter and visit `Edit` page
- Visit the `Send` page by clicking `Next` on the the editor page
- Toggle the `schedule` option on
- Verify that the primary submit button this page reads `Send` for past dates/times and `Schedule` for future dates/times, note that time is calculated based on the time show in `Your website’s time is {time}`

[MAILPOET-4163]

[MAILPOET-4163]: https://mailpoet.atlassian.net/browse/MAILPOET-4163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/date.ts
+++ b/mailpoet/assets/js/src/date.ts
@@ -19,7 +19,7 @@ export const MailPoetDate : {
   full: (date: MomentInput) => string,
   time: (date: MomentInput) => string,
   convertFormat: (format: string) => string,
-  isInFuture: (dateString: string) => boolean
+  isInFuture: (dateString: string, currentTime: MomentInput) => boolean
 } = {
   version: 0.1,
   options: {},
@@ -178,5 +178,5 @@ export const MailPoetDate : {
 
     return convertedFormat.join('');
   },
-  isInFuture: (dateString: string): boolean => new Date(dateString).getTime() > Date.now(),
+  isInFuture: (dateString: string, currentTime: MomentInput): boolean => Moment(dateString).isAfter(currentTime, 's'),
 } as const;
--- a/mailpoet/assets/js/src/newsletters/models/newsletter.ts
+++ b/mailpoet/assets/js/src/newsletters/models/newsletter.ts
@@ -1,4 +1,20 @@
-type NewsLetterType = 'standard';
+export enum NewsletterType {
+  Standard = 'standard',
+  Automatic = 'automatic',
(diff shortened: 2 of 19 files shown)


mailpoet/assets/js/src/newsletters/models/newsletter.ts
```diff
@@ -1,4 +1,20 @@
-type NewsLetterType = 'standard';
+export enum NEWSLETTER_TYPE {
+  STANDARD = 'standard',
+  AUTOMATIC = 'automatic',
+  WELCOME = 'welcome',
+  NOTIFICATION = 'notification',
+  NOTIFICATION_HISTORY = 'notification_history',
+  WC_TRANSACTIONAL = 'wc_transactional',
+  RE_ENGAGEMENT = 're_engagement'
+}
```
JanJakes, line 9: This is just a stylistic note but we should preserve the same style as for other enums. This means `enum NewsletterType` and enum fields `Standard`, `Automatic`, etc.

Screen Shot 2022-03-11 at 13 41 32
Screen Shot 2022-03-11 at 13 42 28

mailpoet/assets/js/src/date.ts
```diff
@@ -178,5 +178,5 @@ export const MailPoetDate : {
 
     return convertedFormat.join('');
   },
-  isInFuture: (dateString: string): boolean => new Date(dateString).getTime() > Date.now(),
+  isInFuture: (dateString: string, currentTime: MomentInput): boolean => Moment(dateString).isAfter(currentTime, 's'),
```
JanJakes, line 181: So I guess we don't use this function in any other place?
Seems like `currentTime` could be optional. But if we don't use it elsewhere, maybe not too necessary.

JanJakes (commented):
LGTM, just tiny notes.

JanJakes (approved):
Good job!

veljkho (approved):

PR #4000, opened 2022-03-08
Automation prototyping [MAILPOET-4136]

JanJakes:
[MAILPOET-4136]

[MAILPOET-4136]: https://mailpoet.atlassian.net/browse/MAILPOET-4136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

This is an initial **prototype** for running automation workflows. It includes:

1. **Workflow** data definition, table, API to create them, etc.
2. **Workflow run** data definition, table, and related services.
3. **Step** definition and runner service.
4. **Trigger** interface and handler.
5. **Action** interface and runner service.
6. **Subject** interface for carrying data through the workflow.
7. **Integration** interface as a wrapper for added triggers, actions, and subjects (e.g. "MailPoet integration").
8. **MailPoet** integration with:
    - `Segment-subscribed` trigger (triggered when the state of a subscription to a segment becomes `subscribed`).
    - `Segment` subject.
    - `Subscriber` subject.
9. **Core** integration with:
    - `Wait` action.
10. A **testing button to create a workflow** in the UI.

To **create and test a workflow**:
1. Ensure you're running MailPoet under a domain name (`mailpoet.localhost`, `mp3.localhost`, etc.) so that WP Cron and Action Scheduler work.
2. Activate the automation feature on the experimental features page (`?page=mailpoet-experimental`).
3. Go to the automation page (`?page=mailpoet-automation`).
4. Click on `Create testing workflow`.
5. Go to the DB and change the workflow status to `active` (look for `wp_mailpoet_automation_workflows ` table).
6. Subscribe a new subscriber (e.g. via a subscription form).
7. Once the subscription is confirmed, the workflow will be triggered (see in `wp_mailpoet_automation_workflow_runs` and Action Scheduler tables).
8. The workflow only has a trigger and a 60s wait action. After the 60s the workflow run will be completed.

It **doesn't include** yet:
1. Tests.
2. Data validation.
3. Logging.
9. Connection to premium features.

I recommend going commit-by-commit as I tried to make them well-ordered and their titles and descriptions explanatory.

The APIs can be seen as drafts that can be significantly changed. I'll be happy for any feedback.

### DIFF
--- a/mailpoet/lib/Automation/Engine/Control/StepRunner.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepRunner.php
@@ -3,8 +3,11 @@
 namespace MailPoet\Automation\Engine\Control;
 
 use Exception;
+use MailPoet\Automation\Engine\Exceptions;
 use MailPoet\Automation\Engine\Exceptions\InvalidStateException;
 use MailPoet\Automation\Engine\Hooks;
+use MailPoet\Automation\Engine\Storage\WorkflowRunStorage;
+use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Automation\Engine\WordPress;
 use Throwable;
 
@@ -12,10 +15,20 @@ class StepRunner {
   /** @var WordPress */
   private $wordPress;
 
+  /** @var WorkflowRunStorage */
+  private $workflowRunStorage;
+
+  /** @var WorkflowStorage */
+  private $workflowStorage;
+
   public function __construct(
-    WordPress $wordPress
+    WordPress $wordPress,
+    WorkflowRunStorage $workflowRunStorage,
+    WorkflowStorage $workflowStorage
   ) {
     $this->wordPress = $wordPress;
+    $this->workflowRunStorage = $workflowRunStorage;
+    $this->workflowStorage = $workflowStorage;
   }
 
   public function initialize(): void {
@@ -43,6 +56,24 @@ class StepRunner {
       throw new InvalidStateException();
     }
 
-    // TODO: process step
+    $workflowRunId = $args['workflow_run_id'];
+    $stepId = $args['step_id'];
+
+    $workflowRun = $this->workflowRunStorage->getWorkflowRun($workflowRunId);
+    if (!$workflowRun) {
+      throw Exceptions::workflowRunNotFound($workflowRunId);
+    }
+
+    $workflow = $this->workflowStorage->getWorkflow($workflowRun->getWorkflowId());
+    if (!$workflow) {
+      throw Exceptions::workflowNotFound($workflowRun->getWorkflowId());
+    }
+
+    $step = $workflow->getStep($stepId);
+    if (!$step) {
+      throw Exceptions::workflowStepNotFound($stepId);
+    }
+
+    // TODO: process step based on its type
   }
 }
--- a/mailpoet/lib/Automation/Engine/Control/StepRunner.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepRunner.php
@@ -82,6 +82,12 @@ class StepRunner {
       throw Exceptions::workflowNotFound($workflowRun->getWorkflowId());
     }
 
+    // complete workflow run
+    if (!$stepId) {
+      $this->workflowRunStorage->updateStatus($workflowRunId, WorkflowRun::STATUS_COMPLETE);
+      return;
+    }
+
     $step = $workflow->getStep($stepId);
     if (!$step) {
       throw Exceptions::workflowStepNotFound($stepId);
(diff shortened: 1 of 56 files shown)


mailpoet/lib/Automation/Engine/Migrations/Migrator.php
```diff
@@ -0,0 +1,82 @@
...
+  /** @var string */
+  private $prefix;
+
+  /** @var wpdb */
+  private $wpdb;
+
+  public function __construct() {
+    global $wpdb;
+    $this->prefix = $wpdb->prefix . 'mailpoet_automation_';
+    $this->wpdb = $wpdb;
```
websupporter, line 18: This will most likely break on some hosts with advanced object caches (like WPE) and I think also in conjunction with plugins like W3 Total cache. I never really digged deep enough into why, but it seems, that sometimes $wpdb gets replaced by a new $wpdb later in the process and then we can't connect to the database anymore. We had the problem in the [CentralStock plugin](https://woocommerce.com/products/centralstock-for-woocommerce/) and I had it also with a client plugin. I injected a WpdbFactory instead, which always fetched the global `$wpdb` when it was needed.

JanJakes: @websupporter Oh, well... thanks, that's a good point. I have a note that we should write a wrapper class for WPDB anyway so we can handle query errors (and convert them to exceptions) so this is something we should handle in there as well.

mailpoet/lib/Automation/Engine/Storage/WorkflowRunStorage.php
```diff
@@ -0,0 +1,43 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Storage;
+
+use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Workflows\WorkflowRun;
+use wpdb;
+
+class WorkflowRunStorage {
```
websupporter, line 9: I would like to have an `updateWorkflowRun($workflowRun)` method (so not sure right now, what risks there might be involved), but at least an `updateSubjects($workflowRun, $subjects)` method, so actions are able to enrich/change contextual data

JanJakes: @websupporter I only added methods that the current code is using for now. I didn't try to create a "storage" service that covers all use cases, rather I added what I needed and thought we'll add more methods when we need them. Similarly for workflows — editing via the API will need to be implemented soon so we'll have to add that and think about the API.

We don't have an ORM here and so far it seemed to me these simple value objects I added served well. With editing, this may get more complicated and we'll have to decide whether the value objects should be editable via setters or an edit will mean a whole new value object and editing only particular DB columns will have specific storage methods.

mailpoet/lib/Automation/Engine/Control/StepRunner.php
```diff
@@ -0,0 +1,124 @@
...
+    if ($stepType === Step::TYPE_ACTION) {
+      $this->actionStepRunner->run($step, $workflow, $workflowRun);
+    } else {
+      throw new InvalidStateException();
+    }
+
+    $nextStepArgs = [
+      [
+        'workflow_run_id' => $workflowRunId,
+        'step_id' => $step->getNextStepId(),
```
websupporter, line 110: Instead of passing the next step as integer to the Action Scheduler, I would store it in the database I think.

This would allow us to quickly answer the question "How many users are in step `x` of the workflow"

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
```diff
@@ -0,0 +1,82 @@
...
+        steps longtext,
+        PRIMARY KEY (id)
+      );
+    ");
+
+    $this->runQuery("
+      CREATE TABLE {$this->prefix}workflow_runs (
+        id int(11) unsigned NOT NULL AUTO_INCREMENT,
+        workflow_id int(11) unsigned NOT NULL,
+        trigger_key int(11) unsigned NOT NULL,
```
brezocordero, line 40: There is a type mismatch here, `trigger_key` is a `String` in the `WorkflowRun` class.

brezocordero (changes requested):
Great work @JanJakes 
This is a solid prototype to build on it.

I found a small type issue and I have added a couple of comments to your checklist.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
JanJakes: Added this to the checklist.

mailpoet/lib/Automation/Engine/Control/StepRunner.php
JanJakes: @websupporter Good point. We definitely need to store such info. I actually wonder — will we need some "logs" table to store one row per workflow step? Thinking about it seems to me we'll need that — to have more info, history, possibility to debug, etc.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
JanJakes: @brezocordero Good catch! Checking what's stored in my DB now and there's just `0`. This column is not used anywhere yet but I thought it would be good to store which trigger started a workflow. I'll fix it.

PR #4002, opened 2022-03-08
Phpstan cache [MAILPOET-4055]

JanJakes:
[MAILPOET-4055]

[MAILPOET-4055]: https://mailpoet.atlassian.net/browse/MAILPOET-4055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/package.json
+++ b/mailpoet/package.json
@@ -32,7 +32,8 @@
     "**/newsletter_editor/**/*spec.js": "../eslint-config/node_modules/.bin/eslint -c ../eslint-config/.eslintrc.tests_newsletter_editor.json --max-warnings 0",
     "*.php": [
       "phplint",
-      "./do qa:code-sniffer"
+      "./do qa:code-sniffer",
+      "bash -c './do qa:phpstan'"
     ]
   },
   "dependencies": {
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -529,9 +529,6 @@ class RoboFile extends \Robo\Tasks {
     // PHPStan must be run out of main plugin directory to avoid its autoloading
     // from vendor/autoload.php where some dev dependencies cause conflicts.
     return $this->collectionBuilder()
-      // temp dir
-      ->taskExec('mkdir -p ' . __DIR__ . '/temp')
-      ->taskExec('rm -rf ' . __DIR__ . '/temp/phpstan')
       ->taskExec($task)
       ->rawArg(
         implode(' ', [
--- a/mailpoet/lib/Form/Block/BlockRendererHelper.php
+++ b/mailpoet/lib/Form/Block/BlockRendererHelper.php
@@ -143,7 +143,7 @@ class BlockRendererHelper {
     $rules = [];
     if (isset($formSettings['fontSize'])) {
       $rules[] = 'font-size: ' . trim($formSettings['fontSize']) . 'px;';
-      $rules[] = 'line-height: ' . trim($formSettings['fontSize']) * 1.2 . 'px";';
+      $rules[] = 'line-height: ' . (float)trim($formSettings['fontSize']) * 1.2 . 'px";';
     }
     if (isset($styles['bold'])) {
       $rules[] = 'font-weight: bold;';
--- a/mailpoet/tasks/phpstan/composer.json
+++ b/mailpoet/tasks/phpstan/composer.json
@@ -1,7 +1,7 @@
 {
     "require": {
         "php-stubs/woocommerce-stubs": "^6.0",
-        "phpstan/phpstan": "1.4.6",
+        "phpstan/phpstan": "1.4.8",
         "phpstan/phpstan-doctrine": "1.2.11",
         "phpstan/phpstan-phpunit": "1.0.0",
         "szepeviktor/phpstan-wordpress": "1.0.2"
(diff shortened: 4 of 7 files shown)


rodrigoprimo (commented):
Great work, @JanJakes. I'm excited to see this moving forward as it should save some time for all of us. I have a few questions just to double-check some things before moving this PR forward.

- Do we know what are the criteria that PHPStan uses to regenerate its cache? I ask this because if time is a factor we might run into some issues since we are adding the PHPStan check to the pre-commit hook. Say for example, the cache gets invalidated after a day. Then it is likely that the pre-commit hook will be slow often for most of us which was a concern that was raised in the P2 post.
- Is it safe to remove `hoa/console`? I see that it was added in 0cfcd6c18d61ef93b9e5bc3040014b062e933dd5 to make the codeception console work. This package is not required anymore by codeception? (cc @costasovo as you added it)

mailpoet/package.json
```diff
@@ -32,7 +32,8 @@
     "**/newsletter_editor/**/*spec.js": "../eslint-config/node_modules/.bin/eslint -c ../eslint-config/.eslintrc.tests_newsletter_editor.json --max-warnings 0",
     "*.php": [
       "phplint",
-      "./do qa:code-sniffer"
+      "./do qa:code-sniffer",
+      "bash -c './do qa:phpstan'"
```
rodrigoprimo, line 36: Just out of curiosity, what is the reason to use `bash -c './do qa:phpstan'` instead of just `./do qa:phpstan` as we are doing in the line above for PHPCS?

JanJakes: @rodrigoprimo I explained it [here](https://github.com/mailpoet/mailpoet/pull/4002/commits/9bb7b86c8a7ae4bcd5e236cebb3c0d01f36aa064), it's because lint-staged always passes the list of files.

JanJakes:
> Do we know what are the criteria that PHPStan uses to regenerate its cache?

It seems that `The full analysis is performed at least every 7 days.`: https://phpstan.org/user-guide/result-cache
So what do you think? Should we drop the pre-commit hook or keep it?

> Is it safe to remove hoa/console?

Ahh, damn. I searched for all usages within our project but never thought it could be an "optional" dependency that would not be explicitly registered anywhere. Well, that's a problem. The [hoa/console](https://github.com/hoaproject/Console) package seems to be completely abandoned so they will never fix this issue and I guess Codeception will have to remove it one day as well. I'll check locally if this is really a problem now and see what we can do.

mailpoet/package.json
rodrigoprimo: Ah, thanks. I missed the explanation in the commit message.

rodrigoprimo:
> It seems that The full analysis is performed at least every 7 days.: https://phpstan.org/user-guide/result-cache
So what do you think? Should we drop the pre-commit hook or keep it?

I suggest we keep it. I'm a big fan of having this check in the pre-commit hook. If it turns out PHPStan needs to regenerate the cache frequently and this becomes a problem, we can reevaluate this decision. It might also be worth checking with the rest of the team, or at least announcing this change when we merge this PR so that people can speak up if they find this change is impacting their workflow for some reason.

> Ahh, damn. I searched for all usages within our project but never thought it could be an "optional" dependency that would not be explicitly registered anywhere. Well, that's a problem. The [hoa/console](https://github.com/hoaproject/Console) package seems to be completely abandoned so they will never fix this issue and I guess Codeception will have to remove it one day as well. I'll check locally if this is really a problem now and see what we can do.

Yeah, I saw that the project is abandoned. I never used the Codeception console. I'm inclined to think that it is better to have PHPStan checks run fast than to keep an abandoned dependency. But let's check first if the Codeception console still depends on hoa/console and, if so, if there are people on our team that use it and ask them for their opinion.

JanJakes:
@rodrigoprimo So this got quite interesting... 😆 

First, I discovered that cleaning only PHPStan DI cache (and not the analysis cache) solves the issue for us and doesn't really slow down the runs, so I thought that would solve it.

Then, [I reported the issue to PHPStan](https://github.com/phpstan/phpstan/issues/6765) and while being requested to reproduce it on a small repo, I wasn't able to do so. Finally, I realized the issue no longer exists on the latest PHPStan that was released 4 days ago.

So yeah, the issue just disappeared. Good timing.

I just upgraded PHPStan and repushed the branch.

rodrigoprimo (approved):
I can confirm it is working, and we can move forward with this PR. What a crazy coincidence that a change in PHPStan that was just released helped us here. Thanks again for this change. I would suggest writing a P2 post announcing that we have added PHPStan to the git pre-commit hook so that others can let us know if they are ok with this change or not. Especially for us to monitor how often PHPStan will regenerate the cache.

As a sidenote, @costasovo, when you have the time, could you maybe check if we still need hoa/console? Since it has been abandoned, if we don't need it anymore, maybe we can remove it in a separate PR?

veljkho (approved):

costasovo:
> As a sidenote, @costasovo, when you have the time, could you maybe check if we still need hoa/console? Since it has been abandoned, if we don't need it anymore, maybe we can remove it in a separate PR?

I tried removing the package and without the package we are not able to use `$i->pauze();` ([interactive pauze](https://codeception.com/docs/02-GettingStarted#interactive-pause)). I use it from time to time, but I could live without it if necessary.
Screenshot 2022-03-14 at 13 13 43

rodrigoprimo:
@costasovo, thanks for checking. Now that Jan found a way to make the PHPStan cache work without removing the package, I think we can keep it.

PR #4004, opened 2022-03-08
Expand inactive subscriber criteria

brezocordero:
This PR changes the inactive subscriber criteria to:

d) The subscriber has not clicked or opened any emails in X months (as specified in settings), human or machine. Currently, we discard machine opens/clicks. (updated condition)
b) The subscriber has received at least 3 emails in the last X months (setting value) (updated condition)

It also removes old code on the tests and comments that were no longer true.

(c), (e), (f), and (g) are already implemented and covered by existing tests on: 
tests/integration/Statistics/Track/SubscriberActivityTrackerTest.php and tests/integration/Subscribers/InactiveSubscribersControllerTest.php

[MAILPOET-4081]

[MAILPOET-4081]: https://mailpoet.atlassian.net/browse/MAILPOET-4081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscribers/InactiveSubscribersController.php
+++ b/mailpoet/lib/Subscribers/InactiveSubscribersController.php
@@ -16,7 +16,7 @@ class InactiveSubscribersController {
 
   const UNOPENED_EMAILS_THRESHOLD = 3;
 
-  private $inactiveTaskIdsTableCreated = false;
+  private $processedTaskIdsTableCreated = false;
 
   /** @var SettingsRepository */
   private $settingsRepository;
@@ -79,24 +79,25 @@ class InactiveSubscribersController {
       return false;
     }
 
-    $inactiveTaskIdsTable = 'inactive_task_ids';
-    if (!$this->inactiveTaskIdsTableCreated) {
-      $inactiveTaskIdsTableSql = "
-        CREATE TEMPORARY TABLE IF NOT EXISTS {$inactiveTaskIdsTable}
+    // Temporary table with processed tasks from threshold date up to yesterday
+    $processedTaskIdsTable = 'inactive_task_ids';
+    if (!$this->processedTaskIdsTableCreated) {
+      $processedTaskIdsTableSql = "
+        CREATE TEMPORARY TABLE IF NOT EXISTS {$processedTaskIdsTable}
         (INDEX task_id_ids (id))
         SELECT DISTINCT task_id as id FROM {$sendingQueuesTable} as sq
           JOIN {$scheduledTasksTable} as st ON sq.task_id = st.id
           WHERE st.processed_at > :thresholdDate
           AND st.processed_at < :dayAgo
       ";
-      $connection->executeQuery($inactiveTaskIdsTableSql, [
+      $connection->executeQuery($processedTaskIdsTableSql, [
         'thresholdDate' => $thresholdDateIso,
         'dayAgo' => $dayAgoIso,
       ]);
-      $this->inactiveTaskIdsTableCreated = true;
+      $this->processedTaskIdsTableCreated = true;
     }
 
-    // Select subscribers who received at least a number of emails after  threshold date
+    // Select subscribers who received at least a number of emails after threshold date and subscribed before that
     $startId = (int)$startId;
     $endId = $startId + $batchSize;
     $inactiveSubscriberIdsTmpTable = 'inactive_subscriber_ids';
@@ -105,7 +106,7 @@ class InactiveSubscribersController {
       (UNIQUE subscriber_id (id))
       SELECT s.id FROM {$subscribersTable} as s
         JOIN {$scheduledTaskSubscribersTable} as sts USE INDEX (subscriber_id) ON s.id = sts.subscriber_id
-        JOIN {$inactiveTaskIdsTable} task_ids ON task_ids.id = sts.task_id
+        JOIN {$processedTaskIdsTable} task_ids ON task_ids.id = sts.task_id
       WHERE s.last_subscribed_at < :thresholdDate
         AND s.status = :status
         AND s.id >= :startId
--- a/mailpoet/lib/Subscribers/InactiveSubscribersController.php
+++ b/mailpoet/lib/Subscribers/InactiveSubscribersController.php
@@ -79,8 +79,6 @@ class InactiveSubscribersController {
       return false;
     }
 
-    // We take into account only emails which have at least one opening tracked
-    // to ensure that tracking was enabled for the particular email
     $inactiveTaskIdsTable = 'inactive_task_ids';
     if (!$this->inactiveTaskIdsTableCreated) {
       $inactiveTaskIdsTableSql = "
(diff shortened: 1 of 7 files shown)


mailpoet/lib/Subscribers/InactiveSubscribersController.php
```diff
@@ -79,8 +79,6 @@ private function deactivateSubscribers(Carbon $thresholdDate, int $batchSize, ?i
       return false;
     }
 
-    // We take into account only emails which have at least one opening tracked
-    // to ensure that tracking was enabled for the particular email
```
JanJakes, line 81: Yeah, this comment was confusing me in the previous commit. I'm also thinking the `inactive_task_ids` as a table name here is rather confusing as well. From what it seems this does is fetching all task IDs "from threshold date up to yesterday", which are only candidates in the inactivity period.

No need to change though, as this was not added now.

mailpoet/lib/Subscribers/InactiveSubscribersController.php
```diff
@@ -98,7 +96,7 @@ private function deactivateSubscribers(Carbon $thresholdDate, int $batchSize, ?i
       $this->inactiveTaskIdsTableCreated = true;
     }
 
-    // Select subscribers who received at least a number of tracked emails but didn't open any
+    // Select subscribers who received at least a number of emails after  threshold date
```
JanJakes, line 99: Another comment that was previously pretty confusing. It said `Select subscribers who received a recent tracked email but didn't open it but there is no check for "not opening it" as it only selects subscribers with a "recently received emails but subscribed non-recently".

Could we maybe improve the comment to sth like `Select subscribers who received at least a number of emails after threshold date and subscribed before that`?

JanJakes: I also really wonder why the part with checking engagement is not directly here — why there is yet another query below. Again, this is not a topic for this PR but I think we would save ourselves a query, a temporary table, and loading a lot of IDs to memory by just doing it all here:

```sql
SELECT s.id FROM {$subscribersTable} as s
  JOIN {$scheduledTaskSubscribersTable} as sts USE INDEX (subscriber_id) ON s.id = sts.subscriber_id
  JOIN {$inactiveTaskIdsTable} task_ids ON task_ids.id = sts.task_id
WHERE s.last_subscribed_at < :thresholdDate
  AND s.status = :status
  AND s.id >= :startId
  AND s.id < :endId
  AND s.last_engagement_at > :thresholdDate -- I added this
GROUP BY s.id
HAVING count(s.id) >= :unopenedEmailsThreshold
```

Note that I added only `AND s.last_engagement_at > :thresholdDate` and not the other columns as they seem to make no sense:
```sql
AND GREATEST(
  COALESCE(s.last_engagement_at, '0'),
  COALESCE(s.last_subscribed_at, '0'), -- this is excluded in WHERE above (s.last_subscribed_at < :thresholdDate)
  COALESCE(s.created_at, '0') -- this is also excluded by WHERE because created_at <= last_subscribed_at
) > :thresholdDate
```

I'll leave this question here for @costasovo, maybe, as it seems it would significantly simplify the code and probably also be more performant (no need for a temporary table, no need to join all again in the query below, simpler condition).

mailpoet/tests/integration/Subscribers/InactiveSubscribersControllerTest.php
```diff
@@ -28,6 +28,7 @@ class InactiveSubscribersControllerTest extends \MailPoetTest {
 
   const INACTIVITY_DAYS_THRESHOLD = 5;
   const PROCESS_BATCH_SIZE = 100;
+  const UNOPENED_EMAILS_THRESHOLD = 1;
```
JanJakes, line 31: I wonder if a test with `1` is a good test. It seems to be a bit of a "special case". Would it be hard to make the test use the actual `3` (and leave this parameter non-configurable in the method calls)?

JanJakes (commented):
LGTM! 👍 

I only added one question and some observations that are unrelated to this PR. I'd like @costasovo to have a look at them when he's back as it seems to me this the inactivation can be simpler.

mailpoet/tests/integration/Subscribers/InactiveSubscribersControllerTest.php
brezocordero: I will change it, true that 1 might not be the best case. It was the previous number and I didn't want to change all the tests.
It was also a way to test that changing the default also works, in case want to turn it into a setting, but I only need one test with that.

mailpoet/lib/Subscribers/InactiveSubscribersController.php
brezocordero: Excellent point. 
I didn't look at the last query with the engagement values and it does look like it can be simplified. I also prefer that @costasovo takes a look in case we are overseeing something.

Could we move forward and review the query next week or do you think we need to wait? The deactivation criteria are very restrictive at the moment.

JanJakes: > Could we move forward and review the query next week or do you think we need to wait?

Definitely. I know it's unrelated, I just noticed that...

JanJakes (approved):

veljkho (approved):

mailpoet/lib/Subscribers/InactiveSubscribersController.php
costasovo: @brezocordero, @JanJakes Sorry for the late answer! I think your observations are correct. We changed the code couple of times, and now it looks like we could include those conditions into the temporary table query.
I looked into the file's history, and here is the commit [that added the temporary tables](https://github.com/mailpoet/mailpoet/pull/2401/files#diff-5ba4c0669779ae299f0e1ae5a8cba775c10e69d2d31792f1c94569e03145b37bR104-R124). At that time, the second query contained another join. Hence, it seemed reasonable to have it separated. Later [we changed the second query](https://github.com/mailpoet/mailpoet/commit/616b8868fe3e4b89c9628980431dd84ebb6f8b5a#diff-5ba4c0669779ae299f0e1ae5a8cba775c10e69d2d31792f1c94569e03145b37bR105-R114) to work with `last_engagement_at`, so after this change, we probably no longer need the separate query for the ids.

But looking at this PR, I found something I'm not sure is the correct change. To take also machine opens into account for inactive detection, [we started to treat machine opens as user engagement](https://github.com/mailpoet/mailpoet/pull/4004/commits/a59bfdbf427562fd01bcf6f2844e027f9b516718) and we update subscribers' `last_egagement_at` also for machine opens. This may have some side effects e.g. subscribers with machine-opens only won't receive re-engagement emails. I'm not sure if that is expected/ok behavior. Maybe @NeosinneR could confirm it is ok on not? Another approach we could take here is reintroducing the join on `mailpoet_statistics_opens` in the second query, which selects ids for deactivation (somewhat revert [this change](https://github.com/mailpoet/mailpoet/commit/616b8868fe3e4b89c9628980431dd84ebb6f8b5a#diff-5ba4c0669779ae299f0e1ae5a8cba775c10e69d2d31792f1c94569e03145b37bR105-R114)).

costasovo: I discussed the change @NeosinneR on slack and the change ⬆️  is correct.

brezocordero: >  so after this change, we probably no longer need the separate query for the ids.

Thanks for confirming, I am working on another change that will touch the query and I will include the simplification.

PR #4007, opened 2022-03-11
Exclude orders not payed and on hold from calculating purchase segments [MAILPOET-4127]

websupporter:
Fixes [MAILPOET-4127]

To test:

1. Create Pending payment and On-hold orders.
2. Create a “Purchased this product” segment with this product
3. Customers should not appear in the segment
4. Create a “Purchased in category” segment with this product
5. Customers should not appear in the segment

[MAILPOET-4127]: https://mailpoet.atlassian.net/browse/MAILPOET-4127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
@@ -107,7 +107,7 @@ class WooCommerceCategory implements Filter {
         'orderStats',
         'customer.customer_id = orderStats.customer_id'
       )
-      ->andWhere("orderStats.status NOT IN ('wc-cancelled', 'wc-failed')");
+      ->andWhere("orderStats.status NOT IN ('wc-cancelled', 'wc-on-hold', 'wc-pending', 'wc-failed')");
   }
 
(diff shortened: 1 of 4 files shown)


brezocordero (changes requested):
@websupporter Code looks good and it works but there are other 3 segments where we also need to apply the changes:

- WooCommerceCountry.php
- WooCommerceProduct.php
- WooCommerceTotalSpent.php

If it gets too big, please create separate tickets and I will approve this PR.

websupporter:
Hi @brezocordero, thanks for looking into the PR

WooCommerce category is part of this PR or did I miss something?
For the other segments, I have created separate Jira issues, as I think the definition of what counts as an order will probably vary in those cases: [MAILPOET-4188], [MAILPOET-4187]

brezocordero (approved):
LGTM! 
Thanks for creating the other tickets.

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #4010, opened 2022-03-14
Add custom styles to CaptchaRenderer [MAILPOET-4134]

websupporter:
Fixes [MAILPOET-4134]

This PR utilizes the `Form\Util\Styles` class to apply the custom styles defined in a form also to the captcha form, thus rendering the error and success messages in the correct colors.

[MAILPOET-4134]: https://mailpoet.atlassian.net/browse/MAILPOET-4134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing Instructions:
1. Go to MailPoet > Forms
2. Edit one of the forms and make sure to set `Success Message Color` / `Error Message Color` colors to something like Black / Purple. Save the form at the end. Setting these colors can be found under Styles tab on the right.
3. Open Private window and go to front-end to trigger such form
4. Fill everything required and submit the form
5. When redirected to a new page to fill Captcha information, fill it and proceed further
6. Observe at the end the Success / Error message colors. Should reflect what you have set in the Styles settings. Previously, it was white all the time no matter what you set.

### DIFF
--- a/mailpoet/lib/Form/Util/Styles.php
+++ b/mailpoet/lib/Form/Util/Styles.php
@@ -108,7 +108,7 @@ class Styles {
 
     $typeSpecificStyles = $this->getFormTypeSpecificStyles($selector, $displayType);
 
-    $messagesStyles = $this->renderMessagesStyles($formSettings, $selector);
+    $messagesStyles = $this->renderFormMessageStyles($form, $selector);
 
     $additionalStyles = $selector . ' .mailpoet_paragraph.last {margin-bottom: 0} ';
     $media .= " @media (min-width: 500px) {{$selector} .last .mailpoet_paragraph:last-child {margin-bottom: 0}} ";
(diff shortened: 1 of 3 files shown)


wxa (approved):

rohankacharegs:
GlobalStep Verified

websupporter:
[daa8e04](https://github.com/mailpoet/mailpoet/pull/4010/commits/daa8e0414ab4578f5fd8e6f0bcbdede964294c29) extends the Styles class, so we are able to only render the message styles (color etc. of the message strings). The new public method is used in the CaptchaRenderer to add those styles only.

veljkho (approved):

veljkho (approved):

PR #3975, opened 2022-02-16
Revert "Add translations downloading from translate.wordpress.com [MAILPOET-3808]"

costasovo:
Reverts mailpoet/mailpoet#3833

PR #4013, opened 2022-03-15
Fix downloading translation language packs [MAILPOET-4144]

costasovo:
This PR contains changes from [the original PR](https://github.com/mailpoet/mailpoet/pull/3833) that were [reverted](https://github.com/mailpoet/mailpoet/pull/3975) due to issues detected in production. 

During debugging, I found a couple of issues:

1. MailPoet also has translations on translate.wordpress.org. These translations are incomplete, but they are installed automatically during the fresh plugin install before the plugin activation (hook is not active) and may be installed later if their change timestamp is newer than the timestamp of installed translations.
I addressed this issue by forcing a newer timestamp for translations that we download from translate.wordpress.com than installed translations from .org. Also, if we detect new translations available from .org for locales available to download from .com, we will remove them from the update queue so that they don't override installed translations from .org.

2. For the fresh installation of the plugin, WP downloads only incomplete translations from .org, and a user needs to go to the updates page and manually click install translations to get the correct translations from .com. The correct translations would be eventually downloaded with auto-update, but it takes some time.
I addressed the issue by adding code for checking and installing new translations into the plugin activation callback. It will be triggered together with migrator and populator.

### Locales that have language packs from translate.wordpress.com (for version 3.84.1)

Currently, we serve only language packs for locales that have more than 85% of strings translated

#### Locales
- sq
- ar
- ca
- da_DK
- nl_NL
- nl_NL_formal
- fr_CA
- fr_FR
- de_DE
- de_DE_formal
- de_CH
- el
- he_IL
- hu_HU
- it_IT
- ja
- pt_BR
- pt_BR
- ro_RO
- ru_RU
- sr_RS
- es_MX
- es_ES
- sv_SE
- tr_TR

### Testing with officially supported languages

1. Create a new testing site and switch it to some language that is officially supported by MailPoet (e.g., french, german)
2. Install the free plugin and the premium plugin and verify that the plugin is fully translated
3. Add another officially supported language to the site. Go to the updates page and update translations,
4. Verify that plugin is fully translated

### Testing with translations available only from .org (e.g,. czech, greek)
1. Create a new testing site and switch it to some language that is not officially supported by MailPoet (e.g.,  czech, greek)
2. Install the free plugin and the premium plugin and verify that the plugin is **partially** translated
3. Add another not officially supported language to the site. Go to the updates page and update translations.
4. Verify that plugin is **partially** translated

[MAILPOET-4144]

[MAILPOET-4144]: https://mailpoet.atlassian.net/browse/MAILPOET-4144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lysyjan (approved):
Great job!

JanJakes:
@costasovo Could you please rebase this on master? There are conflicts after https://github.com/mailpoet/mailpoet/pull/4012.

veljkho (approved):

PR #4014, opened 2022-03-15
Add transactional sending method to system info [MAILPOET-4148]

johnolek:
This PR adds the transactional sending method advanced setting to the system info tab on the Help page, as well as to the data we collect for analytics.

## Advanced Setting 
`/wp-admin/admin.php?page=mailpoet-settings#/advanced`
image

## System Info Tab 
`/wp-admin/admin.php?page=mailpoet-help#/systemInfo`
image

## Analytics Data
Viewing analytics data requires that the "Share anonymous data" advanced setting is set to `Yes`
image

The easiest way I found to inspect the analytics data was by looking at the `mailpoet_analytics_data` variable in the browser console.

image

However, this variable is not populated on every page load. It's created periodically whenever analytics are due to be sent in. You can make the plugin believe that it's time to send analytics again by deleting the `analytics_last_sent` setting from `wp_mailpoet_settings`.
image

[MAILPOET-4148]

[MAILPOET-4148]: https://mailpoet.atlassian.net/browse/MAILPOET-4148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Analytics/Reporter.php
+++ b/mailpoet/lib/Analytics/Reporter.php
@@ -108,6 +108,7 @@ class Reporter {
       'MailPoet Premium version' => (defined('MAILPOET_PREMIUM_VERSION')) ? MAILPOET_PREMIUM_VERSION : 'N/A',
       'Total number of subscribers' => $this->subscribersFeature->getSubscribersCount(),
       'Sending Method' => isset($mta['method']) ? $mta['method'] : null,
+      "Send all site's emails with" => $this->settings->get('send_transactional_emails') ? 'current sending method' : 'default WordPress sending method',
       'Date of plugin installation' => $this->settings->get('installed_at'),
       'Subscribe in comments' => (boolean)$this->settings->get('subscribe.on_comment.enabled', false),
       'Subscribe in registration form' => (boolean)$this->settings->get('subscribe.on_register.enabled', false),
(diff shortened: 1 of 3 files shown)


mailpoet/lib/Analytics/Reporter.php
```diff
@@ -108,6 +108,7 @@ public function getData() {
       'MailPoet Premium version' => (defined('MAILPOET_PREMIUM_VERSION')) ? MAILPOET_PREMIUM_VERSION : 'N/A',
       'Total number of subscribers' => $this->subscribersFeature->getSubscribersCount(),
       'Sending Method' => isset($mta['method']) ? $mta['method'] : null,
+      "Send all site's emails with" => $this->settings->get('send_transactional_emails') ? 'current sending method' : 'default WordPress sending method',
```
johnolek, line 111: I noticed that there doesn't appear to be an integration test for this class. Is that something that would be worth adding as part of this PR?

brezocordero:
@johnolek I am assigning it back to you in case you want to add the tests now.

Feel free to pass it on to QA if you decide not to do it as part of this issue.

johnolek:
@brezocordero thanks for the review! And thank you for suggesting to me however many weeks ago it was that testing steps are useful to have in PR descriptions! I probably wouldn't have included them if not for that, and I'm glad they were helpful.

I'm going to pass this particular PR on to QA. There are a couple of followup things I thought of (including those tests) and I think it makes the most sense to create a followup ticket.

veljkho (approved):

PR #4017, opened 2022-03-16
Refactor SubscribersFinder to use Doctrine instead of Paris [MAILPOET-3925]

rodrigoprimo:
[MAILPOET-3925]

[MAILPOET-3925]: https://mailpoet.atlassian.net/browse/MAILPOET-3925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Steps:
1. Go to MailPoet > Lists > Segments tab
2. Make sure you see some segment that counts at least 3 subscribers (or make a new segment instead)
3. Send simple newsletter to that segment and make sure they have received newsletter

### DIFF
--- a/mailpoet/lib/Segments/SubscribersFinder.php
+++ b/mailpoet/lib/Segments/SubscribersFinder.php
@@ -2,12 +2,12 @@
 
 namespace MailPoet\Segments;
 
+use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\ScheduledTaskSubscriberEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Entities\SubscriberSegmentEntity;
 use MailPoet\InvalidStateException;
-use MailPoet\Models\ScheduledTask;
 use MailPoetVendor\Doctrine\DBAL\Connection;
 use MailPoetVendor\Doctrine\DBAL\ParameterType;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
@@ -54,12 +54,12 @@ class SubscribersFinder {
   }
 
   /**
-   * @param ScheduledTask $task
+   * @param ScheduledTaskEntity $task
    * @param array<int>    $segmentIds
    *
    * @return float|int
    */
-  public function addSubscribersToTaskFromSegments(ScheduledTask $task, array $segmentIds) {
+  public function addSubscribersToTaskFromSegments(ScheduledTaskEntity $task, array $segmentIds) {
     // Prepare subscribers on the DB side for performance reasons
     $staticSegmentIds = [];
     $dynamicSegmentIds = [];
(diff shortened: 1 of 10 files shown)


mailpoet/lib/Segments/SubscribersFinder.php
```diff
@@ -46,35 +45,43 @@ private function findSubscribersInSegment(SegmentEntity $segment, $subscribersTo
...
-    foreach ($segments as $segment) {
-      if ($this->isStaticSegment($segment)) {
-        $staticSegments[] = $segment;
-      } elseif ($segment->type === SegmentEntity::TYPE_DYNAMIC) {
-        $dynamicSegments[] = $segment;
+    $staticSegmentIds = [];
+    $dynamicSegmentIds = [];
+    foreach ($segmentIds as $segment) {
+      $segment = $this->segmentsRepository->findOneById($segment);
+      if ($segment instanceof SegmentEntity) {
```
costasovo, line 68: Just a detail, but we could use early return here to save us from adding one level of nesting.
`if (!$segment instanceof SegmentEntity) continue;`

mailpoet/tests/DataFactories/Subscriber.php
```diff
@@ -97,6 +97,7 @@ public function withIsWooCommerceUser($isWooCustomer = true) {
    * @return $this
    */
   public function withSegments(array $segments) {
+    $this->segments = [];
```
costasovo, line 100: Good catch! This reminds me that we solved a similar issues caused by reusing the same instance of data factory by returning a clone in every `withXXX` method (e.g. https://github.com/mailpoet/mailpoet/blob/9f790efbf0f032fc32e264dc33450286b362976e/mailpoet/tests/DataFactories/WooCommerceProduct.php#L173-L179)

Maybe we could use the approach in all data factories. This note is not relevant to this PR since it would probably mean doing more changes in usages of factories. Just wanted to share the idea...

costasovo (approved):
Great job! The change is nicely separated into commits, so it was easy to follow.

Approved! I added one potentially/optionally actionable comment so I'm assigning it back to you but feel free to pass it forward directly to QA.

mailpoet/lib/Segments/SubscribersFinder.php
rodrigoprimo: That is a valid point. To be honest I'm not sure what version I prefer. I agree that having one less level of nesting helps with the readability of the code. On the other hand, I feel like "if nots" like `if (!$segment instanceof SegmentEntity)` are harder to grasp than "affirmative ifs" like `if ($segment instanceof SegmentEntity)`. Since I'm unsure I'm opting to leave the code as is. Happy to hear more about this if you feel like continuing this conversation.

mailpoet/tests/DataFactories/Subscriber.php
rodrigoprimo: I like this and I agree that we should have the same approach in all data factories. Should we maybe create a Jira ticket for it?

mailpoet/lib/Segments/SubscribersFinder.php
costasovo: In general, I prefer the early termination approach because it allows adding a couple separated checks. After them, I'm safe to proceed with the main task. In my opinion, this approach works nicely in a function body. I agree that negative conditions might be harder to read. Especially `instanceof` as the `!` is placed in a strange place. It always felt to me like a negation of the variable on the left side. So a good point about the affirmative if in this case! I'm maybe just more used to the early termination approach.

mailpoet/tests/DataFactories/Subscriber.php
costasovo: I agree about creating the Jira. I [asked other devs about their opinion on slack](https://a8c.slack.com/archives/C01GH835U90/p1647852662288569). If we get an agreement on the proposed change I'll prepare the ticket.

veljkho (approved):

PR #4018, opened 2022-03-16
Update wordpress packages for the form editor [MAILPOET-4113][MAILPOET-4124]

costasovo:
[MAILPOET-4113]

Updating the packages fixed also the issue with hidden block toolbar [MAILPOET-4124]

[MAILPOET-4113]: https://mailpoet.atlassian.net/browse/MAILPOET-4113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4124]: https://mailpoet.atlassian.net/browse/MAILPOET-4124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Steps:
1. Go to MailPoet > Forms
2. Perform adhoc test in the `form editor` / `subscribing to forms on the front-end` - everything should work as before.

Note: You can open front page in a private browser so you can see the form... form is mostly unavailable for logged in users. We don't have exact steps for this because we did update to the package for the form-editor and the bug might be anywhere inside... or not.


wxa (approved):

wxa:
I'm not sure if this is relevant to the task, but some rarely used styles are different in editor vs preview, e.g. "keyboard input" and "highlight" aren't applied to a heading:

изображение

wxa (commented):
Smoke-testing went fine. No console errors, basic functions seem to work. Didn't check all possible settings combinations, though. I have only one question.

costasovo:
Thank you for the thorough testing! I really appreciate it since the form editor is quite a complex app and it is easy to miss something.

> I'm not sure if this is relevant to the task, but some rarely used styles are different in editor vs preview, e.g. "keyboard input" and "highlight" aren't applied to a heading:
> 
> изображение

Good catch! I noticed the difference in "keyboard input" during testing, but I considered it ok-ish. The keyboard input wraps the text into "\<kbd\>" tag and the tag is rendered on the front end. The thing is that WP admin styles are different from the theme styles. 

Maybe we could push more for WYSIWIG here and copy WP admin CSS for the `kbd` tag and load it also on the front end or maybe we could introduce our own styling for the tag. cc @veljkho 

I tested also the highlight style and it worked for me. I tried using themes Twenty Twenty-One and Twenty Twenty-Two.
What theme do you use? Could you please check the behavior in the production version? 

In case it is not a regression I would prefer addressing the issue (actually both issues in case we decide to go for custom CSS for kbd) in a separate issue/PR. This PR contains changes in package-lock.json so it would be nice to have it in master so that we can continue with the rest of updates without conflicts.

wxa:
@costasovo
> I tested also the highlight style and it worked for me. 
I've re-tested, works for me too, so this point is called off.

> 
 I agree that this issue is minor, even if it's a regression, and can be fixed separately. I'll test

veljkho (approved):

PR #4020, opened 2022-03-17
Set an empty subscriber when no Subscriber found for ViewInBrowser

websupporter:
Fixes [MAILPOET-4190]

The problem in MAILPOET-4190 is we try to render the shortcode without a subscriber set, which just returns the plain shortcode [#](https://github.com/mailpoet/mailpoet/blob/9ea484c972000f861cfff3b9f613428a04aadd42/mailpoet/lib/Newsletter/Shortcodes/Categories/Subscriber.php#L37-L39). We rely on this behavior, as in some processes, we run our newsletter through the shortcode processes to prepare them and then again with the actual subscriber.

Therefore this PR proposes to use an empty `SubscriberEntity` when rendering the BrowserView and we do not have a subscriber at hand.

This PR also proposes to move all "find the subscriber"-logic into the `ViewInBrowserController::getSubscriber()`. So in case of no subscriber is found but we are in the preview mode, we try to find the subscriber for the logged in user now also in `ViewInBrowserController::getSubscriber()`.

[MAILPOET-4190]: https://mailpoet.atlassian.net/browse/MAILPOET-4190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:
1. Go to MailPoet > Pages > Add New Page
2. Place this shortcode and save the page: [mailpoet_archive]
3. Send some simple newsletter but place this shortcode [subscriber:firstname | default:Subscriber] in the title and somewhere in the newsletter content. Something like `Hi, [subscriber:firstname | default:Subscriber]`.
4. Log out
5. Go to the page you have created recently and make sure you see Hi, Subscriber in the title and in the content.
6. Log in
7. Check the same page that you see Hi, (your admin First Name).

### DIFF
--- a/mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserController.php
+++ b/mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserController.php
@@ -4,6 +4,7 @@ namespace MailPoet\Newsletter\ViewInBrowser;
 
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\SendingQueueEntity;
+use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Sending\SendingQueuesRepository;
 use MailPoet\Newsletter\Url as NewsletterUrl;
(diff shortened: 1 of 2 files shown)


costasovo (commented):
Good job! Nice that you found that we rely on running the shortcodes without a subscriber in the initial rendering during sending 👍 

Could you please extend the integration test for this class so that it covers the logic added in this PR? I think something similar like [the test case checking it loads WP user](https://github.com/mailpoet/mailpoet/blob/9f790efbf0f032fc32e264dc33450286b362976e/mailpoet/tests/integration/Newsletter/ViewInBrowser/ViewInBrowserControllerTest.php#L135-L154) should work.

websupporter:
Thanks @costasovo! Test added.

costasovo (approved):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #4021, opened 2022-03-18
Add subscriber lifetime emails count

brezocordero:
This PR:

- Adds a new column, `emails_count`, to the subscriber's table to store the number of emails received for the subscriber.
- Adds a SimpleWorker to calculate `emails_count`: 
  - The frequency is the same as InactiveSubscribers
  - If there was no previous run of the task, it will calculate the total count of emails until now - one day.
  - If there was a previous run of the task, it will increment the count of emails starting from previous run - one day.
- Adds a new condition for marking a subscriber as inactive: to have more than 10 `emails_count`

To see the new task running, make sure that tracking is enabled and that `Stop sending to inactive subscribers` setting is set to something else than `never`:
1) To create the column: Deactivate the plugin before switching to this branch and reactivate after switching
2) Check that the task is scheduled in the database (table wp_mailpoet_scheduled_tasks, type = `subscribers_email_count`)
3) Edit the task and set it in the past to make it run now
4) Check the column `emails_count` is populated

Also covered with automated tests.

[MAILPOET-4177]

[MAILPOET-4177]: https://mailpoet.atlassian.net/browse/MAILPOET-4177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Migrator.php
+++ b/mailpoet/lib/Config/Migrator.php
@@ -249,7 +249,7 @@ class Migrator {
       'engagement_score_updated_at timestamp NULL,',
       'last_engagement_at timestamp NULL,',
       'woocommerce_synced_at timestamp NULL,',
-      'emails_count int(11) unsigned NOT NULL DEFAULT 0, ',
+      'email_count int(11) unsigned NOT NULL DEFAULT 0, ',
       'PRIMARY KEY  (id),',
       'UNIQUE KEY email (email),',
       'UNIQUE KEY unsubscribe_token (unsubscribe_token),',
--- a/mailpoet/lib/Config/Migrator.php
+++ b/mailpoet/lib/Config/Migrator.php
@@ -249,6 +249,7 @@ class Migrator {
       'engagement_score_updated_at timestamp NULL,',
       'last_engagement_at timestamp NULL,',
       'woocommerce_synced_at timestamp NULL,',
+      'emails_count int(11) unsigned NOT NULL DEFAULT 0, ',
       'PRIMARY KEY  (id),',
       'UNIQUE KEY email (email),',
       'UNIQUE KEY unsubscribe_token (unsubscribe_token),',
--- a/mailpoet/lib/Cron/Workers/SubscribersEmailCount.php
+++ b/mailpoet/lib/Cron/Workers/SubscribersEmailCount.php
@@ -4,6 +4,8 @@ namespace MailPoet\Cron\Workers;
 
 use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Settings\SettingsController;
+use MailPoet\Settings\TrackingConfig;
 use MailPoet\Subscribers\SubscribersEmailCountsController;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
 
@@ -18,13 +20,36 @@ class SubscribersEmailCount extends SimpleWorker {
   /** @var EntityManager */
   private $entityManager;
 
+  /** @var SettingsController */
+  private $settings;
+
+  /** @var TrackingConfig */
+  private $trackingConfig;
+
   public function __construct(
     SubscribersEmailCountsController $subscribersEmailCountsController,
-    EntityManager $entityManager
+    EntityManager $entityManager,
+    SettingsController $settings,
+    TrackingConfig $trackingConfig
   ) {
     $this->subscribersEmailCountsController = $subscribersEmailCountsController;
     $this->entityManager = $entityManager;
+    $this->settings = $settings;
+    $this->trackingConfig = $trackingConfig;
     parent::__construct();
+  }
+
+  public function checkProcessingRequirements() {
+    if (!$this->trackingConfig->isEmailTrackingEnabled()) {
+      return false;
+    }
+
+    $daysToInactive = (int)$this->settings->get('deactivate_subscriber_after_inactive_days');
+    if ($daysToInactive === 0) {
+      return false;
+    }
+
+    return true;
   }
 
   public function processTaskStrategy(ScheduledTaskEntity $task, $timer) {
--- a/mailpoet/lib/Cron/Workers/SubscribersEmailCount.php
+++ b/mailpoet/lib/Cron/Workers/SubscribersEmailCount.php
@@ -1,0 +1,70 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Cron\Workers;
+
+use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Subscribers\SubscribersEmailCountsController;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
+
+class SubscribersEmailCount extends SimpleWorker {
+  const TASK_TYPE = 'subscribers_email_count';
+  const BATCH_SIZE = 1000;
+  const SUPPORT_MULTIPLE_INSTANCES = false;
+
+  /** @var SubscribersEmailCountsController */
+  private $subscribersEmailCountsController;
+
+  /** @var EntityManager */
+  private $entityManager;
+
+  public function __construct(
+    SubscribersEmailCountsController $subscribersEmailCountsController,
+    EntityManager $entityManager
+  ) {
+    $this->subscribersEmailCountsController = $subscribersEmailCountsController;
+    $this->entityManager = $entityManager;
+    parent::__construct();
+  }
+
+  public function processTaskStrategy(ScheduledTaskEntity $task, $timer) {
+    $previousTask = $this->findPreviousTask($task);
+    $dateFromLastRun = null;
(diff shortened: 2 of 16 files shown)


mailpoet/lib/Cron/Workers/SubscribersEmailCount.php
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private function findPreviousTask(ScheduledTaskEntity $task): ?ScheduledTaskEntity {
+    return $this->scheduledTasksRepository->findPreviousTask($task);
+  }
+
+  private function getHighestSubscriberId(): int {
+    $subscribersTable = $this->entityManager->getClassMetadata(SubscriberEntity::class)->getTableName();
+    $result = $this->entityManager->getConnection()->executeQuery("SELECT MAX(id) FROM $subscribersTable LIMIT 1;")->fetchNumeric();
+    /** @var int[] $result - it's required for PHPStan */
```
brezocordero, line 67: I could instead add an ignore on PHPStan baseline files. 
Not sure if we have a preferred option.

mailpoet/lib/Subscribers/InactiveSubscribersController.php
```diff
@@ -100,6 +101,7 @@ private function deactivateSubscribers(Carbon $thresholdDate, int $batchSize, ?i
     // Select subscribers who received at least a number of emails after threshold date and subscribed before that
     $startId = (int)$startId;
     $endId = $startId + $batchSize;
+    $lifetimeEmailsThreshold = self::LIFETIME_EMAILS_THRESHOLD;
     $inactiveSubscriberIdsTmpTable = 'inactive_subscriber_ids';
     $connection->executeQuery("
       CREATE TEMPORARY TABLE IF NOT EXISTS {$inactiveSubscriberIdsTmpTable}
```
brezocordero, line 107: I tried the simplification we talked about [here](https://github.com/mailpoet/mailpoet/pull/4004#discussion_r822392647) but the integration tests fail
I think it is better to create a separate issue if we want to investigate the simplification.

mailpoet/lib/Subscribers/SubscribersEmailCountsController.php
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      $resultQuery = $connection->executeQuery($processedTaskIdsTableSql, $queryParams);
+      $this->processedTaskIdsTableCreated = true;
+
+      if ($resultQuery->rowCount() === 0) return [0,0];
+    }
+
+    // Temporary table needed for UPDATE query
+    // mySQL does not allow to modify the same table used in the select
```
brezocordero, line 72: On my local env, I have MariaDB and I was able to UPDATE with one query, but it doesn't work with mySQL:
```
MailPoet\Subscribers\SubscribersEmailCountsControllerTest::testItCalculatesTotalSubscribersEmailCounts MailPoetVendor\Doctrine\DBAL\Exception\DriverException: An exception occurred while executing ' UPDATE mp_mailpoet_subscribers as s SET s.emails_count = + IFNULL( ( SELECT count(task_ids.id) from mp_mailpoet_subscribers s2 JOIN mp_mailpoet_scheduled_task_subscribers sts ON s2.id = sts.subscriber_id JOIN processed_task_ids task_ids ON task_ids.id = sts.task_id WHERE s2.id = s.id GROUP BY s.id ) ,0) WHERE s.id >= ? AND s.id <= ? ' with params [0, "3"]: SQLSTATE[HY000]: General error: 1093 You can't specify target table 's' for update in FROM clause
```

mailpoet/tests/integration/Cron/Workers/SubscribersLifetimeEmailCountTest.php
```diff
@@ -0,0 +1,178 @@
...
+use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Models\ScheduledTask;
+use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
+use MailPoet\Subscribers\SubscribersRepository;
+use MailPoet\Tasks\Sending;
+use MailPoet\Test\DataFactories\ScheduledTask as ScheduledTaskFactory;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoetVendor\Carbon\Carbon;
+
+class SubscribersLifetimeEmailCountTest extends \MailPoetTest {
```
brezocordero, line 18: I had to rename this file so that the test is executed after `SubscribersLastEngagementTest.php`
I haven't been able to figure out what is the side effect but if it runs before, the test `MailPoet\Cron\Workers\SubscribersLastEngagementTest::testItCanSetLastEngagementFromOpens` fails.

mailpoet/lib/Config/Migrator.php
```diff
@@ -249,6 +249,7 @@ public function subscribers() {
       'engagement_score_updated_at timestamp NULL,',
       'last_engagement_at timestamp NULL,',
       'woocommerce_synced_at timestamp NULL,',
+      'emails_count int(11) unsigned NOT NULL DEFAULT 0, ',
```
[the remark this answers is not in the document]
brezocordero: I have changed the name to `email_count`. I used `emails_count` because it is used in the stats but I agree that doesn't sound right.

mailpoet/lib/Subscribers/SubscribersEmailCountsController.php
```diff
@@ -0,0 +1,142 @@
...
+      SET s.emails_count = {$initUpdateValue} + IFNULL(sc.emails_count, 0)
+      WHERE s.id >= :startId
+      AND s.id <= :endId
+    ",
+      [
+        'startId' => $startId,
+        'endId' => $endId,
+      ]
+    );
+    $connection->executeQuery("DROP TABLE {$subscriberIdsEmailsCountTmpTable}");
```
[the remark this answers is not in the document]
brezocordero: @JanJakes I had some time during L2 support week and I have modified the query. 

I tested it on a site that had 10.000 subscribers and 9M rows in the wp_mailpoet_scheduled_task_subscribers table.

I need to make one last change because when it was split in multiple, I could know if there were no sending tasks and abort. I might have time to look at it today if there are no support requests.

mailpoet/lib/Subscribers/SubscribersEmailCountsController.php
```diff
@@ -0,0 +1,142 @@
...
+    );
+
+    $subscribersInRange = $result->fetchAllAssociative();
+
+    $countSubscribersInRange = count(array_map(
+      function ($id) {
+        return (int)$id['id'];
+      },
+      $subscribersInRange
+    ));
```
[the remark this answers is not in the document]
brezocordero: `SELECT MAX(s.id) AS max, COUNT(s.id) as count` would not work. 
I am using `LIMIT` in the query so that, if there are gaps in the subscribers IDs, we don't run an empty loop.
MAX(s.id) returns the max of the whole table and I need the MAX id of current batch.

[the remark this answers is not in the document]
JanJakes: Oh, right, that's true. But still I think it's true that `count($subscribersInRange)` would give the same result as the whole `count(array_map(...))` so I think the `array_map` is unnecessary here.

Or you could use a subquery:
```sql
SELECT COALESCE(MAX(ids.id), 0) AS max, COUNT(ids.id) as count FROM (
  SELECT s.id FROM wp_mailpoet_subscribers as s
  WHERE s.id >= 0
  ORDER BY s.id
  LIMIT 10
) ids
```

mailpoet/tests/integration/Cron/Workers/SubscribersLifetimeEmailCountTest.php
JanJakes: Do you think we should investigate this further? (E.g. create a ticket?)

JanJakes (dismissed):
Looks great! I approve but since I've added two more comments, I'm not moving it to QA yet and leaving it up to you, if you want to address them. It's only small details so no need to necessarily do anything.

mailpoet/tests/integration/Cron/Workers/SubscribersLifetimeEmailCountTest.php
brezocordero: Yes, I will create the ticket.

mailpoet/lib/Subscribers/SubscribersEmailCountsController.php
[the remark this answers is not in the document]
brezocordero: Thanks for the suggestions. Modified to use a subquery.

brezocordero:
@JanJakes had a chat with @NeosinneR who suggested not running the cron if the Inactive subscribers setting is disabled.
I have added a check in the cron Worker + test.
Would you mind having a last look?

JanJakes (approved):
Great! 👏

veljkho (approved):

PR #4024, opened 2022-03-22
Unique tokens for subscription form submissions [MAILPOET-2030]

alexpharris:
This PR alters subscription forms to always include a unique CSRF token. It also prevents these tokens from being reused.

### DIFF
--- a/mailpoet/lib/API/JSON/API.php
+++ b/mailpoet/lib/API/JSON/API.php
@@ -106,7 +106,13 @@ class API {
       $this->requestMethod === 'subscribe'
     );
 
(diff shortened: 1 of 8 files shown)


mailpoet/lib/API/JSON/API.php
```diff
@@ -100,13 +93,13 @@ public function setupAjax() {
       $this->setRequestData($_GET, Endpoint::TYPE_GET);
     }
 
-    $ignoreToken = (
```
alexpharris, line 103: I wasn't able to figure out why we were ignoring tokens if captchas were enabled, and I couldn't see a downside to removing this check.

PR #4025, opened 2022-03-22
Unique tokens for subscription form submissions [MAILPOET-2030]

johnolek:
This PR alters subscription forms to always include a unique CSRF token. It also prevents these tokens from being reused.

[MAILPOET-2030]

### DIFF
--- a/mailpoet/lib/API/JSON/API.php
+++ b/mailpoet/lib/API/JSON/API.php
@@ -106,7 +106,13 @@ class API {
       $this->requestMethod === 'subscribe'
     );
 
-    if (!$ignoreToken && $this->checkToken() === false) {
+    $nonceAction = 'mailpoet_token';
+    
+    if (isset($this->requestData['mailpoet_action']) && $this->requestData['mailpoet_action'] !== '') {
+      $nonceAction .= sprintf("_%s", $this->requestData['mailpoet_action']);
+    }
+
+    if (!$ignoreToken && $this->checkToken($nonceAction) === false) {
       $errorMessage = WPFunctions::get()->__("Sorry, but we couldn't connect to the MailPoet server. Please refresh the web page and try again.", 'mailpoet');
       $errorResponse = $this->createErrorResponse(Error::UNAUTHORIZED, $errorMessage, Response::STATUS_UNAUTHORIZED);
       return $errorResponse->send();
@@ -228,8 +234,8 @@ class API {
       $this->accessControl->validatePermission($permissions['global']);
   }
 
-  public function checkToken() {
-    return WPFunctions::get()->wpVerifyNonce($this->requestToken, 'mailpoet_token');
+  public function checkToken($action = 'mailpoet_token') {
+    return WPFunctions::get()->wpVerifyNonce($this->requestToken, $action);
   }
 
   public function setTokenAndAPIVersion() {
--- a/mailpoet/lib/API/JSON/API.php
+++ b/mailpoet/lib/API/JSON/API.php
@@ -93,13 +93,13 @@ class API {
       $this->setRequestData($_GET, Endpoint::TYPE_GET);
     }
 
-    $nonceAction = 'mailpoet_token';
-    
-    if (isset($this->requestData['mailpoet_action']) && $this->requestData['mailpoet_action'] !== '') {
-      $nonceAction .= sprintf("_%s", $this->requestData['mailpoet_action']);
-    }
+    $uniqueAction = $this->requestData['mailpoet_unique_action'] ?? '';
+    $isTokenValid = $uniqueAction !== ''
+      ? Security::checkTokenForUniqueAction($this->requestToken, $uniqueAction)
+      : Security::checkToken($this->requestToken);
 
-    if ($this->checkToken($nonceAction) === false) {
+
+    if (!$isTokenValid) {
       $errorMessage = WPFunctions::get()->__("Sorry, but we couldn't connect to the MailPoet server. Please refresh the web page and try again.", 'mailpoet');
       $errorResponse = $this->createErrorResponse(Error::UNAUTHORIZED, $errorMessage, Response::STATUS_UNAUTHORIZED);
       return $errorResponse->send();
@@ -219,10 +219,6 @@ class API {
     return(!empty($permissions['methods'][$requestMethod])) ?
       $this->accessControl->validatePermission($permissions['methods'][$requestMethod]) :
       $this->accessControl->validatePermission($permissions['global']);
-  }
-
-  public function checkToken($action = 'mailpoet_token') {
-    return WPFunctions::get()->wpVerifyNonce($this->requestToken, $action);
   }
 
   public function setTokenAndAPIVersion() {
--- a/mailpoet/lib/API/JSON/API.php
+++ b/mailpoet/lib/API/JSON/API.php
@@ -4,8 +4,6 @@ namespace MailPoet\API\JSON;
 
 use MailPoet\Config\AccessControl;
 use MailPoet\Exception;
-use MailPoet\Settings\SettingsController;
-use MailPoet\Subscription\Captcha;
 use MailPoet\Tracy\ApiPanel\ApiPanel;
 use MailPoet\Tracy\DIPanel\DIPanel;
 use MailPoet\Util\Helpers;
@@ -40,22 +38,17 @@ class API {
   /** @var WPFunctions */
   private $wp;
 
-  /** @var SettingsController */
-  private $settings;
-
   const CURRENT_VERSION = 'v1';
 
   public function __construct(
     ContainerInterface $container,
     AccessControl $accessControl,
     ErrorHandler $errorHandler,
-    SettingsController $settings,
     WPFunctions $wp
   ) {
     $this->container = $container;
     $this->accessControl = $accessControl;
     $this->errorHandler = $errorHandler;
-    $this->settings = $settings;
     $this->wp = $wp;
     foreach ($this->availableApiVersions as $availableApiVersion) {
       $this->addEndpointNamespace(
@@ -100,19 +93,13 @@ class API {
       $this->setRequestData($_GET, Endpoint::TYPE_GET);
     }
 
-    $ignoreToken = (
-      $this->settings->get('captcha.type') != Captcha::TYPE_DISABLED &&
-      $this->requestEndpoint === 'subscribers' &&
-      $this->requestMethod === 'subscribe'
-    );
-
     $nonceAction = 'mailpoet_token';
     
     if (isset($this->requestData['mailpoet_action']) && $this->requestData['mailpoet_action'] !== '') {
       $nonceAction .= sprintf("_%s", $this->requestData['mailpoet_action']);
     }
 
-    if (!$ignoreToken && $this->checkToken($nonceAction) === false) {
+    if ($this->checkToken($nonceAction) === false) {
       $errorMessage = WPFunctions::get()->__("Sorry, but we couldn't connect to the MailPoet server. Please refresh the web page and try again.", 'mailpoet');
       $errorResponse = $this->createErrorResponse(Error::UNAUTHORIZED, $errorMessage, Response::STATUS_UNAUTHORIZED);
       return $errorResponse->send();
--- a/mailpoet/lib/Form/DisplayFormInWPContent.php
+++ b/mailpoet/lib/Form/DisplayFormInWPContent.php
@@ -187,8 +187,11 @@ class DisplayFormInWPContent {
       $templateData['enableExitIntent'] = true;
     }
 
+    $uniqueFormAction = Security::generateRandomString();
+    $templateData['mailpoetAction'] = $uniqueFormAction;
+    
     // generate security token
-    $templateData['token'] = Security::generateToken();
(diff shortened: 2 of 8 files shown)


mailpoet/lib/API/JSON/API.php
```diff
@@ -100,13 +93,13 @@ public function setupAjax() {
...
-      $this->requestEndpoint === 'subscribers' &&
-      $this->requestMethod === 'subscribe'
-    );
+    $uniqueAction = $this->requestData['mailpoet_unique_action'] ?? '';
+    $isTokenValid = $uniqueAction !== ''
+      ? Security::checkTokenForUniqueAction($this->requestToken, $uniqueAction)
+      : Security::checkToken($this->requestToken);
+
 
-    if (!$ignoreToken && $this->checkToken() === false) {
```
johnolek, line 109: I couldn't think of a reason we would want to bypass this check, even if it is somewhat redundant. It was adding complexity, and if we can see that a token isn't valid, why would we want to let that request continue?

mailpoet/lib/Util/Security.php
```diff
@@ -31,11 +32,49 @@ public static function generateToken($action = 'mailpoet_token') {
...
+  }
+
+  public static function checkToken(string $token, string $action = 'mailpoet_token'): bool {
+    $isNonceValid = WPFunctions::get()->wpVerifyNonce($token, $action);
+
+    if (!$isNonceValid || !self::isActionUnique($action)) {
+      return $isNonceValid;
+    }
+
+    // Nonces that were generated for a unique action should truly only ever be used once
```
johnolek, line 58: This next part may be overkill, or there may be a more elegant way of handling it, but I really wanted to make sure that these security tokens were truly a single time use.

I was a little worried that this could result in tons of transients cluttering up users' options tables, but then I found that WordPress has a daily cron task to delete expired transients.

mailpoet/lib/API/JSON/API.php
JanJakes: I also have no idea. Found it [here](https://github.com/mailpoet/mailpoet/commit/caa062311221a54a18d3e67b8a06783f97d8c61e) but no reason why it was added. If the captcha works anyway, let's just check it always 👍

mailpoet/lib/Util/Security.php
```diff
@@ -31,6 +32,44 @@ public static function generateToken($action = 'mailpoet_token') {
     return WPFunctions::get()->wpCreateNonce($action);
   }
 
+  public static function generateTokenForUniqueAction(string $action) {
```
JanJakes, line 35: Not sure why we call it `generateToken` when internally we call `wpCreateNonce` (what I mean is we could use the "nonce" naming also outside to be clearer).

But I'm also not sure why the method was in this file. Maybe we could extract it now to a dedicated class to handle the nonces only?

mailpoet/lib/Util/Security.php
JanJakes: Yeah, it seems to be a bit unnecessary to me. The reason is, in other places, WP doesn't do that either (nonces are not one-time use) so us doing it doesn't seem to have that much sense to me. It's not possible to expire the nonces and https://codex.wordpress.org/WordPress_Nonces doesn't suggest attempting that.

So I'd stick to what WP recommends, as extending it to one-use would only add unnecessary complexity and possible performance issues on large sites. The security risk is questionable — if an attacker manages to get a nonce, they'd still need your session if I'm not mistaken.

mailpoet/lib/WP/Functions.php
```diff
@@ -563,6 +563,10 @@ public function wpLoginUrl($redirect = '', $forceReauth = false) {
     return wp_login_url($redirect, $forceReauth);
   }
 
+  public function wpNonceTick() {
+    return wp_nonce_tick();
+  }
+
```
JanJakes, line 569: This doesn't seem to be used, if I'm not mistaken.

mailpoet/lib/Form/DisplayFormInWPContent.php
```diff
@@ -187,8 +187,11 @@ private function getContentBellow(FormEntity $form, string $displayType): string
       $templateData['enableExitIntent'] = true;
     }
 
+    $uniqueFormAction = Security::generateRandomString();
+    $templateData['mailpoetAction'] = $uniqueFormAction;
```
JanJakes, line 191: It [seems that WP suggests](https://codex.wordpress.org/WordPress_Nonces) simply using IDs such as form ID in our case and I almost wanted to ask why we don't do that as well... but then I realized that we need the randomness for the guest users.

But that made me think... is it really correct to do this for guests? Many sites use CDNs, many pages may be cached, and when there's no session, the cache will break this mechanism. I'm not sure if this can work for guests at all and I just found WC was discussing the same: https://github.com/woocommerce/woocommerce/issues/2921

This makes me think whether the issue we're trying to solve is a valid issue. It doesn't seem so — nonces can prevent an attacker from doing a request on your behalf (e.g. when you're an admin) without you knowing that. But when you're a guest we don't care what they do on "your behalf" as you have no specific session tied to you.

I'm thinking this issue may be invalid in the first place. What do you think?

JanJakes (changes requested):
See my comment about caching. I think the reported issue unfortunately doesn't really make sense and we should not use nonces for guest users at all.

mailpoet/lib/Util/Security.php
johnolek: I agree that this is probably a bad solution since it's basically changing how nonces work, which could have unintended side effects and definitely adds more complexity than it's worth, especially since the issue itself doesn't seem valid as you point out in another comment.

mailpoet/lib/Util/Security.php
johnolek: Maybe we should just use the `wpCreateNonce` and `wpVerifyNonce` methods from `WPFunctions` directly wherever we need them? It doesn't seem like we need any special functionality unless we want to hardcode the action name to always be `mailpoet_token`, which doesn't seem like a great idea anyway.

mailpoet/lib/WP/Functions.php
johnolek: Good catch, thanks! I first tried to set the transient expiration based on this value before realizing it wasn't the correct thing to use.

mailpoet/lib/Form/DisplayFormInWPContent.php
johnolek: I agree that the issue as stated is probably invalid to begin with. I also know that sometimes these issues will continue to be flagged by security scanners and sometimes it's easier to change the code than to have to explain why a result is invalid. If we don't have any kind of obligation to prevent these kinds of issues (i.e. ones that aren't truly a problem) from coming up in scans I'm happy to avoid it.

Perhaps the reason why we were ignoring the token if the captcha is turned on is to avoid the caching issue?

I did also consider using form ID, or some kind of hash based on form data, and part of the reason I moved away from that was that I didn't want to clutter up `setupAjax` with more logic that's specific to a particular endpoint. That was a big part of the reason I wanted to get rid of the `$ignoreToken` logic

```php
    $ignoreToken = (
      $this->settings->get('captcha.type') != Captcha::TYPE_DISABLED &&
      $this->requestEndpoint === 'subscribers' &&
      $this->requestMethod === 'subscribe'
    );
```

We were including the `SettingsController` and had `use MailPoet\Subscription\Captcha;` specifically for one endpoint. I didn't want to have to import other classes just to retrieve the form if we wanted to generate an action based on more than the form ID. It might seem silly or like overkill to retrieve the form entity for the nonce, but one of the potential issues I was thinking about is this:

1) A form is created for segments 1 and 2
2) The form is rendered on a page and gets cached
3) The form is changed to subscribe users to segments 3, 4, and 5
4) A user sees the cached form and submits it

Would we still want the form to work if the segments changed? Users could potentially be signing up for something they weren't intending to. If we calculate an action based on the segment IDs then we could avoid that problem, though I don't know if it truly is a problem.

Overall I was wondering if it would make sense to reconsider how we use nonces in general. Perhaps it makes sense to hand over the responsibility of generating/checking nonces to the request endpoint classes themselves, since they would be better equipped to create action-specific actions instead of us always using `mailpoet_token`. The effort wouldn't be trivial though and I don't know that we would gain much from it.

Truly the biggest potential issue I see with our current implementation is that a person could more easily re-use a specific form by making direct API requests with `curl` or similar, and that's only an issue if captchas are disabled, which they aren't by default.

Anyway, that's a lot of text mostly to say that I agree that the original issue isn't valid as such, and maybe it would be best to change nothing at all because we're better off ignoring the tokens when captchas are enabled to avoid issues with cached forms.

mailpoet/lib/Util/Security.php
JanJakes: > Maybe we should just use the wpCreateNonce and wpVerifyNonce methods from WPFunctions directly wherever we need them?

True, also a possibility. I wrote that comment before thinking about the caching and validity of the ticket itself.

mailpoet/lib/Form/DisplayFormInWPContent.php
JanJakes: > sometimes it's easier to change the code than to have to explain why a result is invalid

I think the problem with this ticket, in particular, is that if we implement it, we will break forms on cached pages and it will happen in two ways (no matter whether captcha is used or not):

1. If a page is still in the cache and the nonce is expired, the form will not work.
2. If a user caches a nonce and we have the one-use policy, it will not work for another user (from the cache).

Additionally, the cache would still re-use that same nonce for multiple users (and we would have to make remove the one-use logic).

Therefore, it seems to me that we can't implement this as we would essentially create new problems with forms.

> Would we still want the form to work if the segments changed? Users could potentially be signing up for something they weren't intending to.

I'd say yes as this is a caching issue, not a security issue. I'm not sure what the form includes (if list IDs or just form ID) but it's something we can't really avoid anyway (a nonce would be cached too).

> Overall I was wondering if it would make sense to reconsider how we use nonces in general.

I agree, seems like we should use action-specific nonces, I'm not sure how often we use them now. But at the same time, it seems to me that for a REST API it's used like this: https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/

> Anyway, that's a lot of text mostly to say that I agree that the original issue isn't valid as such, and maybe it would be best to change nothing at all because we're better off ignoring the tokens when captchas are enabled to avoid issues with cached forms.

Yes, I'm afraid we can't merge it anyway, as mentioned above. I don't see captcha to be very related btw., as the nonces are for CSRF protection while captcha is against spambots. CSRF is dangerous when you're authenticated as someone could make your browser do unwanted actions, but doesn't seem to make any sense for guest users.

johnolek: > If a page is still in the cache and the nonce is expired, the form will not work.

Isn't this also true currently? I suppose only when captchas are disabled, and I suspect that's a very uncommon configuration.

> I'm not sure what the form includes (if list IDs or just form ID) but it's something we can't really avoid anyway (a nonce would be cached too).

In this case I was thinking we wouldn't include the segment IDs in the form itself, rather we would use those IDs to generate the nonce and also to check it, which would all be server side. For example, the nonce action might be `mailpoetSubscribeForm1Segments1,2`. At the time of the check it would be `mailpoetSubscribeForm1Segments3,4,5`. But like you said it's a totally separate issue.

> I agree, seems like we should use action-specific nonces, I'm not sure how often we use them now. But at the same time, it seems to me that for a REST API it's used like this: https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/

Thanks for pointing this out. If I'm understanding it correctly WP doesn't even check (or care) about nonces for logged out users, is that right? It makes sense since, as you've pointed out, CSRF protection doesn't really make much sense for logged out users.

Is there anything you think would be worthwhile to do at this point? Perhaps just getting rid of `Security`'s `generateToken` method and using our `WPFunctions` wrapper directly instead? We could also ignore tokens for all unauthenticated requests to match WordPress's behavior and avoid caching issues.

JanJakes: > Isn't this also true currently? I suppose only when captchas are disabled, and I suspect that's a very uncommon configuration.

I don't know. Is caching related to whether captchas are used? Is the captcha included on the first load or as an intermediate step after submit? I don't really remember how it works.

Seeing that WC went the same way to not using the tokens for guest sessions, I guess we could do the same.

> Is there anything you think would be worthwhile to do at this point?

I guess we can do the improvements you're suggesting. Maybe it would make the code more explicit about this if we ignore the tokens.

johnolek: > I don't know. Is caching related to whether captchas are used? Is the captcha included on the first load or as an intermediate step after submit? I don't really remember how it works.

What I meant is that it would only be an issue with captchas disabled because we're ignoring tokens in that case. If there's a cached form and the token is > 24 hours old, the nonce check will fail.

johnolek:
@JanJakes Thank you very much for your thoughtful review of this PR! Based on our discussion I think it makes the most sense to close this PR and say we won't do MAILPOET-2030, unless we can come up with a reason why this would actually ever be a problem and also come up with a solution that wouldn't break cached forms.

I created #4032 as followup to our discussion about using `generateToken` in the `Security` class.

JanJakes:
@johnolek Let's close this?

PR #4028, opened 2022-03-23
Fix Division by zero Error in Newsletter button - MAILPOET-4193

triple0t:
[MAILPOET-4193](https://mailpoet.atlassian.net/browse/MAILPOET-4193)

Testing:
* Create a new Newsletter or edit an already existing one
* Add the Button component 
* Set button **Height** to 0
* Save
* Click on Preview

Note: you can test before checking out to the branch to see the error

### DIFF
--- a/mailpoet/lib/Newsletter/Renderer/Blocks/Button.php
+++ b/mailpoet/lib/Newsletter/Renderer/Blocks/Button.php
@@ -22,7 +22,7 @@ class Button {
                     style="height:' . EHelper::escapeHtmlStyleAttr($element['styles']['block']['lineHeight']) . ';
                            width:' . EHelper::escapeHtmlStyleAttr($element['styles']['block']['width']) . ';
                            v-text-anchor:middle;"
-                    arcsize="' . round((int)$element['styles']['block']['borderRadius'] / (int)$element['styles']['block']['lineHeight'] * 100) . '%"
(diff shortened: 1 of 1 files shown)


websupporter:
Thanks @triple0t!
A bit nitpicky, [but `arcsize` is as far as I understand the border-radius for outlook emails](https://docs.microsoft.com/en-us/windows/win32/vml/msdn-online-vml-arcsize-attribute)?

The default border-radius is 5 Pixel, which instead of something very close to 0 or 0 would now lead to 500% (circle), which makes me wonder, if the email in outlook would look similar to my browser preview
image

triple0t:
Hi @websupporter 
Thanks for the review

> The default border-radius is 5 Pixel, which instead of something very close to 0 or 0 would now lead to 500% (circle), which makes me wonder, if the email in outlook would look similar to my browser preview

The Height of the button is effectively 0, (which mean the button may not even show) I don't think there’s any meaning in setting the border radius.
159857636-69885192-80b3-4391-98c6-3a571cbfdbe1

websupporter (approved):
I checked it quickly with [Litmus](https://litmus.com/folders/unsorted_emails/emails/6062901/checklist). There are differences, but those are not so much due to this change. Regarding the arcsize, I agree @triple0t

veljkho (approved):

PR #4031, opened 2022-03-24
Refactor MailPoet\Subscription\Pages to Doctrine

rodrigoprimo:


mailpoet/lib/Subscription/Pages.php
```diff
@@ -235,9 +246,11 @@ public function unsubscribe() {
           (int)$this->data['queueId']
         );
       }
-      $this->subscriber->status = Subscriber::STATUS_UNSUBSCRIBED;
-      $this->subscriber->save();
-      SubscriberSegment::unsubscribeFromSegments($this->subscriber);
```
rodrigoprimo, line 240: I have a question about the old behavior that for now I opted to preserve.

My understanding is that \MailPoet\Subscription\Pages::unsubscribe() is called when users click on the unsubscribe button of a given email that they receive for a particular list. Considering this, I would expect MailPoet to unsubscribe them just from this one list and not from all the lists, which seems to be the current behavior. Is there something that I'm missing, or is this a bug?

costasovo: I don't know for sure, but this is what I think:

The unsubscribe link changes global subscriptions status to `unsubscribed`, so keeping list level statuses as subscribed doesn't make much sense. The unsubscribe screen also allows users to go to the manage subscription page, and they can modify the global and list level subscription statuses. It is not an ideal state, but I think this is kind of related to issues with global/list subscription levels.

As I mentioned, I'm not sure. @NeosinneR could you please share your opinion here.

costasovo (approved):
Good job! Thank you for improving the test coverage of the functionality. 
I added a comment to your question about unsubscribing from segments. This PR doesn't change the behavior so I think it is ok to approve.

mailpoet/lib/Subscription/Pages.php
NeosinneR: @costasovo I agree with your thinking. I saw both approaches (unsubscribing completely or choosing which list to unsubscribe from) in different services, and I don't think there is a clear winner between them. Taking consistency into account, I'd keep our functionality the same. 

As you mentioned, subscribers can still go to Manage Subscription instead of Unsubscribing completely.

rohankacharegs:
GlobalStep Verified

PR #4032, opened 2022-03-25
Inline Security::generateToken for clarity [MAILPOET-2030]

johnolek:
This is a refactor to inline `Security::genreateToken`. The only thing Security::generateToken was providing was a default value for the $action, which created a pattern of using the same $action everywhere, which may not be the best way to go. It was obfuscating what we were really doing under the hood.

This is a followup to #4025, which we decided wouldn't make sense to implement.

[MAILPOET-2030]: https://mailpoet.atlassian.net/browse/MAILPOET-2030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/integration/API/JSON/APITest.php
+++ b/mailpoet/tests/integration/API/JSON/APITest.php
@@ -73,11 +73,15 @@ class APITest extends \MailPoetTest {
         expect($api instanceof JSONAPI)->true();
       }
     );
+    $wpStub = Stub::make(new WPFunctions, [
+      'wpVerifyNonce' => asCallable(function() {
+        return true;
+      })]);
     $api = Stub::makeEmptyExcept(
       $this->api,
       'setupAjax',
       [
-        'wp' => new WPFunctions,
+        'wp' => $wpStub,
         'processRoute' => Stub::makeEmpty(new SuccessResponse),
         'settings' => $this->container->get(SettingsController::class),
       ]
--- a/mailpoet/lib/API/JSON/API.php
+++ b/mailpoet/lib/API/JSON/API.php
@@ -9,7 +9,6 @@ use MailPoet\Subscription\Captcha;
 use MailPoet\Tracy\ApiPanel\ApiPanel;
 use MailPoet\Tracy\DIPanel\DIPanel;
 use MailPoet\Util\Helpers;
-use MailPoet\Util\Security;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Psr\Container\ContainerInterface;
 use Throwable;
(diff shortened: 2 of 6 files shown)


mailpoet/tests/integration/API/JSON/APITest.php
```diff
@@ -73,11 +73,15 @@ function($api) use (&$called) {
         expect($api instanceof JSONAPI)->true();
       }
     );
+    $wpStub = Stub::make(new WPFunctions, [
+      'wpVerifyNonce' => asCallable(function() {
+        return true;
+      })]);
     $api = Stub::makeEmptyExcept(
```
johnolek, line 80: The only reason this test was passing previously is that the check at http://github.com/mailpoet/mailpoet/blob/10e53ccd1be3ce3b83d287655bf553feb9e524c1/mailpoet/lib/API/JSON/API.php#L108-L108 was checking for false, and the empty stub for `checkToken` was returning `null`.

johnolek:
@JanJakes Please let me know if you can think of any other refactors or changes that you think would make sense here. We could definitely be more explicit about always ignoring tokens for logged out users and add explanations for why.

JanJakes (approved):
LGTM 👍 

> @JanJakes Please let me know if you can think of any other refactors or changes that you think would make sense here. We could definitely be more explicit about always ignoring tokens for logged out users and add explanations for why.

@johnolek Agreed. Do you have any idea how we could do that? Can we add a test that would fail if someone tries to add it again? From the failure it could be quite understandable that we don't expect the nonces to be used for guest users. If it's not easy, I guess we don't have to do it now.

JanJakes:
@johnolek Should we move this to QA or do you want to add some more changes?

johnolek:
@JanJakes Sorry for the delay in circling back about this! I've been going back and forth in my mind a little bit. I was wondering if we should make it so that any logged out API call would automatically skip the token check. Currently we handle the setup the same for both:

```php
    // ajax (logged in users)
    WPFunctions::get()->addAction(
      'wp_ajax_mailpoet',
      [$this, 'setupAjax']
    );

    // ajax (logged out users)
    WPFunctions::get()->addAction(
      'wp_ajax_nopriv_mailpoet',
      [$this, 'setupAjax']
    );
```

Ultimately it feels much too permissive to me -- I think it would be better to handle specific endpoints on a case by case basis as needed -- but I don't like the idea that it would be possible for someone else to make POST requests to all of our public endpoints without having to specify some kind of token. On the other hand, there currently could be some functionality that's broken on sites that use long lasting caches.

So maybe it would be best to move this forward for now and then research other endpoints that would be affected and see if we want to make additional changes?

JanJakes:
> So maybe it would be best to move this forward for now and then research other endpoints that would be affected and see if we want to make additional changes?

Makes sense. There are also longer-term plans to switch to WP REST (trying it out on automations first) so it may be kind of related.

JanJakes:
@johnolek Ahh, damn, there's a conflict. Otherwise, it's approved by me so after the conflict resolution feel free to pass it to QA.

johnolek:
@JanJakes Sorry for the delay on this. I think you may need to re-approve since I pushed changes?

JanJakes (approved):

veljkho (approved):

PR #4034, opened 2022-03-28
Add subscriber counts to MixPanel [MAILPOET-4204]

websupporter:
Fixes [MAILPOET-4204]

This PR adds the subscriber counts to our MixPanel data. Example: 
image

It utilizes the SubscriberListingRepository and therefore uses the same data source, the JSON API uses to display those data for the user.

The PR extends the constructor of the ListingDefintion by assigning some default values. The values chosen are the ones, which are used when visiting the Subscriber Listing page and not changing any value.

[MAILPOET-4204]: https://mailpoet.atlassian.net/browse/MAILPOET-4204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Analytics/Reporter.php
+++ b/mailpoet/lib/Analytics/Reporter.php
@@ -5,6 +5,7 @@ namespace MailPoet\Analytics;
 use MailPoet\Config\ServicesChecker;
 use MailPoet\Cron\CronTrigger;
 use MailPoet\Entities\DynamicSegmentFilterData;
+use MailPoet\Listing\ListingDefinition;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Segments\DynamicSegments\DynamicSegmentFilterRepository;
 use MailPoet\Segments\DynamicSegments\Filters\EmailAction;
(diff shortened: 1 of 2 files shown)


mailpoet/lib/Analytics/Reporter.php
```diff
@@ -189,6 +201,17 @@ public function getData() {
     return $result;
   }
 
+  private function subscriberProperties(): array {
+    $definition = new ListingDefinition();
+    $groups = $this->subscriberListingRepository->getGroups($definition);
+    $properties = [];
+    foreach ($groups as $group) {
+      $properties['Subscribers > ' . $group['label']] = (int)$group['count'];
```
lysyjan, line 209: In my opinion, you should use here a key `name` instead of `label` because the label can be translated by the user's language configuration. What do you think?

lysyjan (commented):
Good idea with SubscribeListingRepository 👍🏻  but I have noticed a minor issue with group labels.

lysyjan (approved):

PR #4036, opened 2022-03-28
Ensure we select today in the date picker [MAILPOET-4205]

websupporter:
Fixes [MAILPOET-4205]

At the end of the month it can happen that the current date (e.g. 28th) is twice visible in the date picker. In order to prevent trying to select the 28th of the previous month, this PR ensures, we pick the current date by switching the class selector


costasovo (approved):
Nice catch! 👍

veljkho (approved):

PR #4042, opened 2022-03-29
Downgrade doctrine [MAILPOET-4215]

costasovo:
Temporary downgrade doctrine packages to unblock release.

This PR is basically a reversion of [the commit with update](https://github.com/mailpoet/mailpoet/commit/ac0a9e7b6ff51d685b624b8dc53d1f3fca648b96) I just kept a few type casts in place.

[MAILPOET-4215]

PR #4044, opened 2022-03-30
Fix doctrine metadata loading errors [MAILPOET-4218]

costasovo:
This PR updates doctrine packages (after [the original update commit was reverted due to issues](https://github.com/mailpoet/mailpoet/pull/4042)) and fixes the issues.

In the production environment, we used to rely on Metadata MappingDriver not being used at all when all metadata are cached. The recent changes in `doctrine/orm` revealed that it is not the case and the driver is used in a [couple](https://github.com/doctrine/persistence/blob/b07e347a24e7a19a2b6462e00a6dff899e4c2dd2/src/Persistence/Mapping/AbstractClassMetadataFactory.php#L478) [of](https://github.com/doctrine/persistence/blob/b07e347a24e7a19a2b6462e00a6dff899e4c2dd2/src/Persistence/Mapping/AbstractClassMetadataFactory.php#L336) places regardless there is some cache. 

We were a bit lucky that we hadn't run into these issues earlier. We used to use [PHPDriver](https://github.com/mailpoet/mailpoet/blob/2668af08b7cbe37fb9fb528bae03c3468ebaf1af/mailpoet/lib/Doctrine/ConfigurationFactory.php#L47-L48) and it accitentally returned correct values that worked, but after [a change in doctrine/orm](https://github.com/doctrine/orm/pull/8285) the PHPDriver was no longer ok for all use-cases.

## Testing
It was causing a couple of issues, so please make sure that:
1) Admin is able to edit an existing newsletter and create a new one using a template
2) Admin is able to set a list to a subscriber (while editing a subscriber)
3) Customer is able to subscribe through MailPoet's form in the Anonymous/Private window

 [MAILPOET-4218]

[MAILPOET-4218]: https://mailpoet.atlassian.net/browse/MAILPOET-4218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Doctrine/CacheOnlyMappingDriver.php
+++ b/mailpoet/lib/Doctrine/CacheOnlyMappingDriver.php
@@ -1,0 +1,55 @@
+<?php declare(strict_types=1);
+
+namespace MailPoet\Doctrine;
+
+use MailPoet\RuntimeException;
+use MailPoetVendor\Doctrine\Persistence\Mapping\ClassMetadata;
+use MailPoetVendor\Doctrine\Persistence\Mapping\Driver\MappingDriver;
+use MailPoetVendor\Psr\Cache\CacheItemPoolInterface;
+
+/**
+ * Intended to be used in production environment where we rely on metadata cache for reading all metadata.
+ */
+class CacheOnlyMappingDriver implements MappingDriver {
+  /** @var string */
+  protected $cacheSalt = '__CLASSMETADATA__';
+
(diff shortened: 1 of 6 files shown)


JanJakes (approved):
Great job debugging this 👏

rohankacharegs:
GlobalStep Verified

PR #4047, opened 2022-04-01
Mailer Refactor - Part 1 [MAILPOET-4115]

costasovo:
This is the first PR related to refactoring Mailer. The goal of this first step was the extraction of factory functionality to a separate class.
Note: I was not able to follow the design suggested in the related research because I found a few more usages of the original `init` method and also found that we can't initialize/build Mailer directly in DI since the initialization may fail due incomplete configuration in settings. So I decided to use factory + "caching" of the default mailer.

### Testing
This change touches on core sending functionality. Please verify that all types of emails sent by the plugin work as expected. Please check the following areas of the plugin:

1. Standard Newsletters
2. Post Notifications
3. Welcome + Automatic (it should be enough to test only Welcome emails)
4. Subscription Confirmation emails (subscribing to a form and receiving confirmation email, clicking conf. link inside)
5. New Subscriber notification email (Check the checkbox to receive these notification at Settings > Basic tab)
6. Stats Notification emails (..Settings > Basic tab)
8. WooCommerce Transactional emails (While buying something on the shop and receiving WC's emails such as The order is in the process, completed..)

[MAILPOET-4115]

[MAILPOET-4115]: https://mailpoet.atlassian.net/browse/MAILPOET-4115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Mailer/Mailer.php
```diff
@@ -16,6 +18,12 @@ class Mailer {
   const METHOD_PHPMAIL = 'PHPMail';
   const METHOD_SMTP = 'SMTP';
 
+  public function __construct(
+    MailerMethod $mailerInstance
+  ) {
+    $this->mailerInstance = $mailerInstance;
```
JanJakes, line 24: Oh, this confused me for a while — let's rename the variable and property to `mailerMethod` or just `method` :)

mailpoet/lib/Mailer/MailerFactory.php
```diff
@@ -102,9 +102,7 @@ public function buildMailer(array $mailerConfig = null, array $sender = null, ar
       default:
         throw new InvalidStateException(__('Mailing method does not exist.', 'mailpoet'));
     }
-    $mailer = new Mailer();
-    $mailer->mailerInstance = $mailerInstance;
-    return $mailer;
+    return new Mailer($mailerInstance);
```
JanJakes, line 105: Also here, I'd suggest renaming of `$mailerInstance` to `$mailerMethod` or just `$method`.

JanJakes (commented):
Great job! 👏  Just a few teeny-tiny suggestions:
1. Maybe we could add `declare(strict_types = 1)` to all new files.
2. In new tests we may also add method return types (well, they are `void` so not much added value here).

costasovo:
@JanJakes Thanks for your review! I addressed most of your remarks. I added 4 new commits.

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #4050, opened 2022-04-04
Automation welcome email [MAILPOET-4191]

johnolek:
This PR adds an automation action for scheduling a welcome email. This is all currently hidden behind a feature flag and these changes will not have any effect on the UI.

[MAILPOET-4191]

[MAILPOET-4191]: https://mailpoet.atlassian.net/browse/MAILPOET-4191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:
1. Please test the Welcome emails sending.


mailpoet/lib/Automation/Integrations/MailPoet/Subjects/SubscriberSubject.php
```diff
@@ -25,8 +25,16 @@ public function __construct(
     $this->subscribersRepository = $subscribersRepository;
 
     $this->fields = [
-      // email
-      new Field(
+      'id' => new Field(
```
johnolek, line 28: The ticket specs say to use the email field for the send welcome email action, but some of the things we need to do (checking that the email hasn't already been scheduled for them, checking that they are in fact subscribed to the segment) seems to require a subscriber object or ID. I could look up the subscriber with their email, but I don't know why that would be preferable to simply using the subscriber ID.

Once again I think I'm not understanding the general idea when it comes to `Field`s. I'm wondering why we wouldn't do something like this:

```php
$this->fields = [
  new Field(
    'mailpoet:subscriber',
    Field::TYPE_SUBSCRIBER,
    __('Subscriber', 'mailpoet'),
    function() {
      return $this->getSubscriber();
    }
  ),
```

and then have actions take care of things like retrieving an email address or subscription status?

Do they need to be something that can be easily represented by text, like in a UI, and we shouldn't have actions retrieving additional data at all?

I like the idea of returning the subscriber entity because otherwise we would need to keep adding fields to `SubscriberSubject` any time a new action requires new data and that could get pretty messy. I'm sure there's something major that I'm missing here though.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+
+    // TODO: Are these necessarily "subscribed" status? Maybe there's a better way to check this
+    $subscriberSegments = $subscriber->getSegments();
+    $subscribedSegmentNames = array_map(function(SegmentEntity $segment) {
+      return $segment->getName();
+    }, $subscriberSegments);
+    
+    if (!in_array($segmentName, $subscribedSegmentNames)) {
+      throw InvalidStateException::create()->withMessage(__(sprintf("Subscriber with ID '%s' is not subscribed to segment '%s'.", $subscriberId, $segmentName), 'mailpoet'));
+    }
```
johnolek, line 90: This is obviously not a good way of handling this, especially the stuff I added for PHPStan. Is this something where the intention would have been that I should add a field to the `SubscriberSubject` that would allow retrieving a list of all the segment names a subscriber is subscribed to?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+    
+    $welcomeEmailId = (int)$step->getArgs()['welcomeEmailId'];
```
johnolek, line 92: Is it a valid assumption that this ID would come from the step?

Are the values for `getArgs` typed, or should we assume they're strings that will need to be cast appropriately?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+    $newsletter = $this->newslettersRepository->findOneById($welcomeEmailId);
+    if ($newsletter === null) {
+      throw NotFoundException::create()->withMessage(__(sprintf("Welcome Email with ID '%s' not found.", $welcomeEmailId), 'mailpoet'));
+    }
+
+    // This check also occurs in createWelcomeNotificationSendingTask, in which case the method returns null, but
+    // that's not the only thing that causes a return value of null, so let's check here so we can craft a more
+    // meaningful exception.
+    $previouslyScheduledNotification = $this->scheduledTasksRepository->findByNewsletterAndSubscriberId($newsletter, $subscriberId);
```
johnolek, line 101: I couldn't see a way around needing a subscriber's ID in this action in large part because of this check and the actual creation of the sending task on L106, which is why I added the ID field and avoided using the email field.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+    if (!$subscriberSubject instanceof SubscriberSubject) {
+      throw InvalidStateException::create()->withMessage(__('No mailpoet:subscriber subject provided.', 'mailpoet'));
+    }
+
+    $segmentSubject = $workflowRun->getSubjects()['mailpoet:segment'] ?? null;
+    if (!$segmentSubject instanceof SegmentSubject) {
+      throw InvalidStateException::create()->withMessage(__('No mailpoet:segment subject provided.', 'mailpoet'));
+    }
+
+    $globalSubscriberStatus = $subscriberSubject->getSubscriberStatusField()->getFactory()();
```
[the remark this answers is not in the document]
JanJakes: I guess that's my bad, just didn't think the concept through the end as I didn't need a subject yet (did some experiments with only "wait" actions). Surely, this doesn't look nice and I we should improve it.

Maybe adding `$field->getValue()` method? It won't be typed... (but maybe we could use PHPStan generics).

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
JanJakes: > Is it a valid assumption that this ID would come from the step?

I think so, yes. Again, I didn't really "use" it yet, just drafted, so it may need some changes.

> Are the values for getArgs typed, or should we assume they're strings that will need to be cast appropriately?

Having them saved as JSON, we can consider it typed but PHPStan doesn't know so the cast may be needed anyway for now.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
JanJakes: Makes sense, since it's a "welcome" email, subscriber should always exist.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+    // that's not the only thing that causes a return value of null, so let's check here so we can craft a more
+    // meaningful exception.
+    $previouslyScheduledNotification = $this->scheduledTasksRepository->findByNewsletterAndSubscriberId($newsletter, $subscriberId);
+    if (!empty($previouslyScheduledNotification)) {
+      throw InvalidStateException::create()->withMessage(__(sprintf("Newsletter with ID '%s' was previously scheduled for subscriber with ID '%s'.", $newsletter->getId(), $subscriberId)));
+    }
+
+    $sendingTask = $this->welcomeScheduler->createWelcomeNotificationSendingTask($newsletter, $subscriberId);
+    if ($sendingTask === null) {
+      // TODO: What exactly does this represent? I think it means the welcome email was configured to be triggered by a segment that has since been deleted. But in the case of this automation it seems like we shouldn't care about how the welcome email is configured. Do we need to be able to create welcome emails that have no explicit trigger segment? Basically something that says "This welcome email can only be triggered via an automation workflow"?
```
JanJakes, line 108: I probably didn't think it through with all the types of welcome emails. At the moment, I created only a [segment-subscribed action](https://github.com/mailpoet/mailpoet/blob/3320afe8f54e53884d3f7d9b3219fa271ba79a86/mailpoet/lib/Automation/Integrations/MailPoet/Triggers/SegmentSubscribedTrigger.php) so we may start with that and extend it further, maybe?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+
+    $segmentName = $segmentSubject->getNameField()->getFactory()();
+    $idField = $subscriberSubject->getSubscriberIdField();
+    $subscriberId = $idField->getFactory()();
+    if ($subscriberId === null) {
+      throw NotFoundException::create()->withMessage(__(sprintf("Subscriber with ID '%s' not found.", $subscriberId), 'mailpoet'));
+    }
+    
+    $subscriber = $this->subscribersRepository->findOneById($subscriberId);
+    // This is for PHPStan, which doesn't understand that retrieving the ID above means the subscriber entity exists
```
JanJakes, line 77: The check is fine, maybe not only for PHPStan, in case some data is invalid in the DB, etc.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+    if (!$subscriber instanceof SubscriberEntity) {
+      throw NotFoundException::create()->withMessage(__(sprintf("Subscriber with ID '%s' not found.", $subscriberId), 'mailpoet'));
+    }
+
+    // TODO: Are these necessarily "subscribed" status? Maybe there's a better way to check this
+    $subscriberSegments = $subscriber->getSegments();
+    $subscribedSegmentNames = array_map(function(SegmentEntity $segment) {
+      return $segment->getName();
+    }, $subscriberSegments);
```
JanJakes, line 86: They are not only "subscribed" status, I think. Also see how the `mailpoet_segment_subscribed` [was implemented](https://github.com/mailpoet/mailpoet/commit/86d934ba363b39bf36d9d090a7914ef833e24078) — checks for both subscriber status and subscriber-segment status will be needed.

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/SubscriberSubject.php
JanJakes: > The ticket specs say to use the email field for the send welcome email action... to simply using the subscriber ID.

Good point, in that case we may add a field for the ID. I think we can't use the subscriber entity itself though. The reason there are the primitives such as `Field::TYPE_STRING` and `Field::TYPE_ENUM` is that in the UI, we'll display them as respective form fields (e.g. a select box) and allow also filtering (e.g. a field of type integer is greater than some value). Having that... maybe we could have a helper method on the subject to fetch the subscriber or something but maybe also the first step can be fetching it by ID in the action, if we need the full entity.

> Do they need to be something that can be easily represented by text, like in a UI, and we shouldn't have actions retrieving additional data at all?

Yep, that's what I meant above.

> I like the idea of returning the subscriber entity because otherwise we would need to keep adding fields to SubscriberSubject any time a new action requires new data and that could get pretty messy. I'm sure there's something major that I'm missing here though.

Would ID work? Or some way to get the subscriber in an action but keep the basic fields for UI, conditions, filters, etc.?

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/SubscriberSubject.php
```diff
@@ -25,8 +25,16 @@ public function __construct(
...
+        'mailpoet:subscriber:email',
+        Field::TYPE_INTEGER,
+        __('Subscriber ID', 'mailpoet'),
+        function() {
+          return $this->getSubscriber()->getId();
+        }
+      ),
+      
+      'email' => new Field(
```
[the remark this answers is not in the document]
JanJakes: > It was a little unclear to me how fields were intended to be retrieved. Is it simply positional and we should reference a field like $subject->getFields()[1]?

I definitely didn't mean this to be positional :) and maybe you're right... that we could simply use the field key? 🤔  I guess that should work.

JanJakes (commented):
Great job so far! 👏

mailpoet/lib/Automation/Integrations/Core/Actions/Wait/ValidationResult.php
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types = 1);
```
johnolek, line 1: I like the idea of having each action define its own validation result class because it allows us to work with typed variables. Does this seem like a reasonable approach?

mailpoet/lib/Automation/Integrations/Core/Actions/Wait/Action.php
```diff
@@ -0,0 +1,60 @@
...
+      [
+        'workflow_run_id' => $workflowRun->getId(),
+        'step_id' => $step->getNextStepId(),
+      ],
+    ]);
+
+    // TODO: call a step complete ($id) hook instead?
+  }
+
+  public function validate(Workflow $workflow, Step $step, array $subjects = []): ValidationResult {
```
johnolek, line 46: It looks like [covariance](https://www.php.net/manual/en/language.oop5.variance.php) (return types more specific than a parent's) isn't fully supported until 7.4, so this is causing static analysis to [fail](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/9460/workflows/51fbb356-cce7-4f27-88e0-b5956fa8960b/jobs/173900). I would really like to be able to do this as written rather than defining the return type as `AbstractValidationResult` and then having to check if it's an instance of the more specific class that I know that it is. Is there a better way to handle this in a PHP 7.2-friendly way?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
```diff
@@ -0,0 +1,134 @@
...
+
+    return $result;
+  }
+
+  public function run(Workflow $workflow, WorkflowRun $workflowRun, Step $step): void {
+    $validationResult = $this->validate($workflow, $step, $workflowRun->getSubjects());
+
+    if (!$validationResult->isValid()) {
+      throw InvalidStateException::create()->withErrors($validationResult->getErrors());
+    }
```
johnolek, line 94: Once again, maybe it would make sense to extract this

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
```diff
@@ -0,0 +1,134 @@
...
+
+    $newsletter = $validationResult->getNewsletter();
+    $subscriber = $validationResult->getSubscriberSubject()->getSubscriber();
+    if (!$subscriber instanceof SubscriberEntity) {
+      throw InvalidStateException::create();
```
johnolek, line 99: I'll add specific error messages once we've settled on an approach. 

Will we want all error messages to be translatable strings? Or just the ones that can occur as part of the "static" validation process, i.e. ones that can be triggered in the UI?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public function getName(): string {
+    return __('Send Welcome Email', 'mailpoet');
+  }
+
+  public function validate(Workflow $workflow, Step $step, array $subjects = []): ValidationResult {
+    $result = new ValidationResult();
+    if (!isset($step->getArgs()['welcomeEmailId'])) {
+      $result->addError('welcomeEmailIdRequired', 'Step arguments did not include a welcomeEmailId.');
```
johnolek, line 56: I need to go back and make these translatable

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/SubscriberSubject.php
johnolek: I'm currently retrieving the subscriber and segment entities directly from the subjects, and I'm not sure I understand why it would be preferable to fetch them from an ID instead. I think I still need to wrap my head around exactly how the fields will work in the UI.

mailpoet/lib/DI/ContainerConfigurator.php
```diff
@@ -120,13 +120,14 @@ public function configure(ContainerBuilder $container) {
...
+    $container->autowire(\MailPoet\Automation\Integrations\Core\Actions\Wait\Action::class)->setPublic(true);
     $container->autowire(\MailPoet\Automation\Integrations\Core\CoreIntegration::class)->setPublic(true);
     // Automation - MailPoet integration
     $container->autowire(\MailPoet\Automation\Integrations\MailPoet\MailPoetIntegration::class)->setPublic(true);
-    $container->autowire(\MailPoet\Automation\Integrations\MailPoet\Subjects\SegmentSubject::class)->setPublic(true);
-    $container->autowire(\MailPoet\Automation\Integrations\MailPoet\Subjects\SubscriberSubject::class)->setPublic(true);
-     $container->autowire(\MailPoet\Automation\Integrations\MailPoet\Triggers\SegmentSubscribedTrigger::class)->setPublic(true);
+    $container->autowire(\MailPoet\Automation\Integrations\MailPoet\Subjects\SegmentSubject::class)->setPublic(true)->setShared(false);
+    $container->autowire(\MailPoet\Automation\Integrations\MailPoet\Subjects\SubscriberSubject::class)->setPublic(true)->setShared(false);
+     $container->autowire(\MailPoet\Automation\Integrations\MailPoet\Triggers\SegmentSubscribedTrigger::class)->setPublic(true)->setShared(false);
```
johnolek, line 129: I set the subjects to not be shared because they'll have state, i.e. specific entities loaded into them. Is this not actually necessary?

mailpoet/tests/DataFactories/Subscriber.php
```diff
@@ -124,6 +124,7 @@ public function create(): SubscriberEntity {
 
     foreach ($this->segments as $segment) {
       $subscriberSegment = new SubscriberSegmentEntity($segment, $subscriber, 'subscribed');
+      $subscriber->getSubscriberSegments()->add($subscriberSegment);
```
johnolek, line 127: This makes the subscriber factory match `\MailPoet\Subscribers\SubscriberSegmentRepository::createOrUpdate`. Prior to this change, calling subscriber->getSegments() after using this factory would always return an empty collection, even after using the `withSegments` method.

I'm no longer using a check that is affected by this, but I thought it would be good to fix this inconsistency in general.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+  public function run(Workflow $workflow, WorkflowRun $workflowRun, Step $step): void {
+    $subscriberSubject = $workflowRun->getSubjects()['mailpoet:subscriber'] ?? null;
+    if (!$subscriberSubject instanceof SubscriberSubject) {
+      throw InvalidStateException::create()->withMessage(__('No mailpoet:subscriber subject provided.', 'mailpoet'));
+    }
+
+    $segmentSubject = $workflowRun->getSubjects()['mailpoet:segment'] ?? null;
+    if (!$segmentSubject instanceof SegmentSubject) {
+      throw InvalidStateException::create()->withMessage(__('No mailpoet:segment subject provided.', 'mailpoet'));
+    }
```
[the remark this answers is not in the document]
johnolek: The changes I just pushed use a different convention that should allow for static validation in the UI.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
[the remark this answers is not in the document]
johnolek: Could you please point me in the direction of how to write such generics?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,119 @@
...
+    ScheduledTasksRepository $scheduledTasksRepository
+  ) {
+    $this->subscribersRepository = $subscribersRepository;
+    $this->welcomeScheduler = $welcomeScheduler;
+    $this->newslettersRepository = $newslettersRepository;
+    $this->scheduledTasksRepository = $scheduledTasksRepository;
+  }
+
+  public function getKey(): string {
+    return 'mailpoet:sendWelcomeEmail'; // TODO: casing for multi word keys? send-welcome-email? send_welcome_email?
```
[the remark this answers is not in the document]
johnolek: I like `send-welcome-email` too 🎉

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
johnolek: I believe I now have both of these checks in place, with a separate check for the global status and then using `findSubscribersInSegments` for the segment specific checks.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
```diff
@@ -0,0 +1,134 @@
...
+    if (!$segment instanceof SegmentEntity) {
+      throw InvalidStateException::create();
+    }
+
+    // Maybe unnecessary since findSubscribersInSegments is status-aware
+    if ($subscriber->getStatus() !== SubscriberEntity::STATUS_SUBSCRIBED) {
+      throw InvalidStateException::create();
+    }
+
+    $isSubscribedToSegment = $this->subscribersFinder->findSubscribersInSegments([$subscriber->getId()], [$segment->getId()]) !== [];
```
JanJakes, line 112: This is the check I'd change because the finder does a lot of queries (probably written for some batch jobs). So maybe I'd do something like this:

```php
// inject SubscriberSegmentRepository
$subscriberSegment = $this->subscriberSegmentRepository->findOneBy([
  'subscriber' => $subscriber,
  'segment' => $segment,
  'status' => SubscriberEntity::STATUS_SUBSCRIBED,
]);
```

Sth like this also works but may result in more queries:
```php
$subscriberSegment = $subscriber->getSubscriberSegments()->filter(
  function (SubscriberSegment $item) use ($subscriber) {
    return $item->getSubscriber() === $subscriber && $item->status === SubscriberEntity::STATUS_SUBSCRIBED;
  }
)[0] ?? null;
```

(This way we can also consider omitting the direct filtering by `SubscriberEntity::STATUS_SUBSCRIBED` and check it afterwards to have different message for segment not found vs. wrong status, but that's a tiny issue, I guess.)

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/SubscriberSubject.php
JanJakes: I guess that falls into what I meant by `Having that... maybe we could have a helper method on the subject to fetch the subscriber or something`. Seems good to me.

The only thing I wanted to stress here is that for some non-MailPoet steps or actions (like conditions and stuff), we'll need to extract all the data we want to allow to be used and/or filtered to fields as conditions should know nothing about `SubscriberEntity`. On the other hand the MailPoet actions definitely can know about it and use it.

mailpoet/lib/Automation/Integrations/Core/Actions/Wait/Action.php
JanJakes: Maybe PHPStan and IDEs would understand if you add `/** @return MyMoreSpecificType */` (keeping the PHP-native return type more general).

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
```diff
@@ -0,0 +1,134 @@
...
+        $result->addError('welcomeEmailNotFound', sprintf("Welcome Email with ID '%s' not found.", $welcomeEmailId));
+      } else {
+        $type = $newsletter->getType();
+        if ($type !== NewsletterEntity::TYPE_WELCOME) {
+          $result->addError('newsletterMustBeWelcomeType', sprintf("Newsletter must be a Welcome Email. Actual type for newsletter ID '%s' was '%s'.", $welcomeEmailId, $type));
+        } else {
+          $result->setNewsletter($newsletter);
+        }
+      }
+    }
```
JanJakes, line 70: I think I still have mixed feelings about the complete validation (meaning constructing results with all errors) using the validation result classes. I'm not sure if we'll ever need to display all these errors. It will be more like "offer this or that action after some kind of trigger or not". And I guess we can make use of the new validator.

There are two or three parts to this:

1. Validating the arg types (the code above). I think this could be done using [the validator](https://github.com/mailpoet/mailpoet/pull/4027). The question is how exactly — should we implement a `getArgsSchema` method for each action? Or should the actions inject the validator themselves?
2. Required subjects (the code below). I'm not sure if this needs a specific method to return required subject types or if it should be more "free-form" (like it is now) — I guess having a specific method is too restrictive, some actions may say "I need a subscriber and an order or a segment" and other complex stuff. But it kind of seems that for now, we could be fine with a simple true/false logic. Or do you see any use for more complex validation results?
3. Anything else. So if an action wants to validate that... I don't know... some plugin is installed, it may be good to allow that. So some kind of a "free-form" validation method may make sense, maybe covering 2. and 3.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
JanJakes: Good question. These (runtime) errors will be logged and probably displayable somewhere in the UI. So I guess in the WP world it means we should make them translatable.

JanJakes (commented):
I like where this is going and your thinking behind it. I guess I'd try to simplify the validation and make use of the new validator. See details in my comments.

mailpoet/lib/DI/ContainerConfigurator.php
johnolek: I was also having trouble coming up with a scenario in which this would actually cause a problem. In any case, we can always remove this when/if we come up with a different approach.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
johnolek: Thank you so much for this explanation! Using the subscribersFinder definitely seemed like overkill.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
johnolek: 1. I like the idea of using the validator for arguments, and my initial reaction is that I would prefer to have something like `getArgsSchema` be part of the interface. It seems like that would make it easier/more consistent to be able to figure out which actions are available for a given set of parameters. 

I'm thinking we should probably try to limit the scope of this PR for now and decide things like more complex validations stuff in a separate ticket. Does that sound appropriate to you?

2. The specific methods are almost certainly overkill. I think it was more that I wanted the certainty of the return types, as well as only having a single place that needs to know what the subject's `string` key is. For example, what if we have some validation method that says yes, `true`, all the required subjects are there, but then the `run` method looks for one of those subjects using `mailpoet:subscribers` instead of `mailpoet:subscriber` as the key?  But it's not like we truly need code to help out with situations like that, and the added complexity certainly isn't worth it, especially considering how much more restrictive it makes things.

3. I think it's a great idea to allow for free form validation that could check for just about anything. That may be something that we would want to handle in a separate ticket though?

mailpoet/lib/Automation/Integrations/Core/Actions/Wait/ValidationResult.php
johnolek: I agree overall that this seems unnecessary. I think the main potential benefit is that it could make it easier to write targeted tests, i.e. we would only need to ensure that validation fails a particular way when given a particular problem as opposed to having to set up all the data, try to run the action and then check the results/exceptions.

But as with my other comments, I think this isn't something that needs to be included in this PR, and it would make more sense to introduce additional complexity like this when/if there's a clear need for it.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
johnolek: Do we want to make all of the strings translatable now, or would it be better to wait until this feature is in more of a final state? I imagine things could change, and I don't know how big of a deal it is to ask our translators to translate strings that we might not end up using.

If these messages are going to be visible in the UI, should we get them approved first and have the approved messages be part of the ticket?

mailpoet/lib/Automation/Engine/Workflows/ActionInterface.php
```diff
@@ -2,10 +2,12 @@
 
 namespace MailPoet\Automation\Engine\Workflows;
 
-interface Action {
+interface ActionInterface {
```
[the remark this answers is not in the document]
johnolek: I definitely hated the renaming I did and I'm very happy to say that I've reverted those changes.

JanJakes (commented):
Great, looks good! 👏  I left only some tiny comments about a typo and coding style.

mailpoet/lib/Automation/Integrations/Core/Actions/Wait/Action.php
```diff
@@ -0,0 +1,60 @@
...
+  public function getKey(): string {
+    return 'core:wait';
+  }
+
+  public function getName(): string {
+    return __('Wait', 'mailpoet');
+  }
+
+  public function run(Workflow $workflow, WorkflowRun $workflowRun, Step $step): void {
+    $result = $this->validate($workflow, $step);
```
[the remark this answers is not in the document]
JanJakes: Right, that makes sense. Also not sure which of the options is better. I was kind of imagining possible 3rd-party actions and in that case, we are not able to ensure they call it as well but we know we call it from outside.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmail/Action.php
JanJakes: I'm not sure what's the process here. A couple of new strings should be OK but as far as the whole feature goes, we should probably let all the original strings have checked before the MVP goes out. I feel like we need to add some messages to the exceptions so it's clear for us, not necessarily translatable now, but then there's the risk way may forget to make them translatable later on.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendWelcomeEmailAction.php
```diff
@@ -0,0 +1,139 @@
...
+    }
+    
+    $segmentSubject = $subjects['mailpoet:segment'] ?? null;
+    $subscriberSubject = $subjects['mailpoet:subscriber'] ?? null;
+
+    return $segmentSubject instanceof SegmentSubject && $subscriberSubject instanceof SubscriberSubject;
+  }
+
+  private function getWelcomeEmailForStep(Step $step): NewsletterEntity {
+    if (!isset($step->getArgs()['welcomeEmailId'])) {
```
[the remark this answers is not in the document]
johnolek: Thank you for pointing out these things! I always appreciate when people point out "nitpicky" things like this, and I don't even like calling it nitpicky because I think they're important in their own way.

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/SubscriberSubject.php
```diff
@@ -25,8 +25,16 @@ public function __construct(
     $this->subscribersRepository = $subscribersRepository;
 
     $this->fields = [
-      // email
-      new Field(
+      'id' => new Field(
+        'mailpoet:subscriber:email',
```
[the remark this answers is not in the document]
johnolek: 😬 Whoops! Thank you for catching this. I guess I wasn't copying and pasting carefully enough.

mailpoet/lib/DI/ContainerConfigurator.php
```diff
@@ -125,9 +125,10 @@ public function configure(ContainerBuilder $container) {
...
```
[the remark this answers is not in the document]
johnolek: Not just GitHub, and I'm not sure how the extra space ended up there but I'm happy you caught it!

JanJakes (approved):

veljkho (approved):

PR #4052, opened 2022-04-05
Update release:publish to use mailpoet.pot file from CircleCI [MAILPOET-4173]

rodrigoprimo:
This PR changes the Robo command release:publish to use the mailpoet.pot file that is present in the zip file in the release branch of the CircleCI build. Before we were using the mailpoet.pot file present in the local repository, which meant it could contain undesired local changes that should not be included when creating a new MailPoet release.

A new Robo command was created to get the zip file from CircleCI called translations:get-pot-file-from-circle-ci. The old translations:build was preserved as it is used inside mailpoet/build.sh.

To test the changes here, I suggest commenting out all the commands that are called inside \RoboFile::releasePublish() after \RoboFile::translationsGetPotFileFromCircleCI() and then running `./do release:publish`. You can try the `release:publish` command before and after running `./do release:download-zip`. I'm not sure if there is a way to test the full `release:publish` command without actually creating a release. It would be nice to have a way to do that.

[MAILPOET-4173]

[MAILPOET-4173]: https://mailpoet.atlassian.net/browse/MAILPOET-4173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -111,12 +111,12 @@ class RoboFile extends \Robo\Tasks {
       )->run();
   }
 
-  public function translationsGetPotFileFromCircleCI() {
+  public function translationsGetPotFileFromBuild() {
     $potFilePathInsideZip = 'mailpoet/lang/mailpoet.pot';
     $potFilePath = 'lang/mailpoet.pot';
 
     if (!is_file(self::ZIP_BUILD_PATH)) {
-      $this->yell('mailpoet.zip file is missing. You must first download it from CircleCI using `./do release:download-zip`.', 40, 'red');
+      $this->yell('mailpoet.zip file is missing. You must first download it using `./do release:download-zip`.', 40, 'red');
       exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


mailpoet/RoboFile.php
```diff
@@ -111,6 +111,36 @@ public function translationsBuild() {
       )->run();
   }
 
+  public function translationsGetPotFileFromCircleCI() {
```
costasovo, line 114: It is just a small detail but I would suggest using a different name here e.g. `translationsGetPotFileFromReleaseZip`.
Currently, we download the release zip from CircleCi, but we may change the service or we decide to download the .zip from the release PR on Github so the name and also the error message may become obsolete/confusing.

rodrigoprimo: Fair point, @costasovo. What do you think of `translationsGetPotFileFromBuild()`?

rodrigoprimo:
@costasovo, FYI, while working on the same change for the premium plugin, I found a small bug without any real consequences in the Robo command that I created. I fixed the problem in commit 501be975ebf1469bc39fce11f4501017cca7b172.

costasovo (dismissed):
Thank you!

veljkho (approved):

PR #4053, opened 2022-04-05
Mailpoet 3992 to show a specific error message for pending approval users

samnajian:
This PR adds a separate message to the Emails listing page ( see attachment ) if the user's account is pending approval.

### To test:
1. Make sure you have MSS as sending method at `MailPoet` > `Settings` > `Send With` tab
2. Now go to `Plugins` > `Plugin File Editor` and make sure you see `MailPoet 3 (New)` as selected plugin.
Now locate the file [API.php](https://github.com/mailpoet/mailpoet/blob/9b71a7804cd7d3c8724c9ac877cc62def4d3a697/mailpoet/lib/Services/Bridge/API.php) under `Plugin Files` on the right sidebar, it should be under `lib` > `Services` > `Bridge` > `API.php` and add the following code into the beginning of the method [sendMessages](https://github.com/mailpoet/mailpoet/blob/9b71a7804cd7d3c8724c9ac877cc62def4d3a697/mailpoet/lib/Services/Bridge/API.php#L103) (exactly under "public function sendMessages($messageBody) {") Save it and continue the step 3.
```
    return [
      'status' => self::SENDING_STATUS_SEND_ERROR,
      'message' => MailerError::OPERATION_PENDING_APPROVAL,
      'code' => 403,
    ];
```
3. Tyy to send a newsletter and verify you see the notice in attachment on the listing page

Screenshot on 2022-04-06 at 16-14-59

NOTE: Please revert the changes in API.php by removing the code that you added just for this test, and save it.

[MAILPOET-3992]

[MAILPOET-3992]: https://mailpoet.atlassian.net/browse/MAILPOET-3992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Mailer/Methods/ErrorMappers/MailPoetMapper.php
+++ b/mailpoet/lib/Mailer/Methods/ErrorMappers/MailPoetMapper.php
@@ -196,10 +196,32 @@ class MailPoetMapper {
     return "{$message}<br/>";
   }
 
+  private function getPendingApprovalMessage(): string {
+    $message = __("Your subscription is currently [link]pending approval[/link].You’ll soon be able to send once our team reviews your account. In the meantime, you can send previews to your authorized emails.", 'mailpoet');
+    $message = Helpers::replaceLinkTags(
+      $message,
+      'https://kb.mailpoet.com/article/350-pending-approval-subscription',
+      [
+        'target' => '_blank',
+        'rel' => 'noopener noreferrer',
+        'data-beacon-article' => '5fbd3942cff47e00160bd248',
+      ],
+      'link'
+    );
+
+    return "{$message}<br/>";
+  }
+
   /**
- * Returns error $message and $operation for API::RESPONSE_CODE_CAN_NOT_SEND
- */
-  private function getCanNotSendError(array $result, array $sender): array {
+   * Returns error $message and $operation for API::RESPONSE_CODE_CAN_NOT_SEND
+   */
+  private function getCanNotSendError(array $result, $sender): array {
+    if ($result['message'] === MailerError::MESSAGE_PENDING_APPROVAL) {
+      $operation = MailerError::OPERATION_PENDING_APPROVAL;
+      $message = $this->getPendingApprovalMessage();
+      return [$operation, $message];
+    }
+
     if ($result['message'] === MailerError::MESSAGE_EMAIL_INSUFFICIENT_PRIVILEGES) {
       $operation = MailerError::OPERATION_INSUFFICIENT_PRIVILEGES;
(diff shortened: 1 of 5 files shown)


mailpoet/lib/Mailer/Methods/ErrorMappers/MailPoetMapper.php
```diff
@@ -211,4 +195,36 @@ private function getEmailVolumeLimitReachedMessage(): string {
 
     return "{$message}<br/>";
   }
+
+  /**
+ * Returns error $message and $operation for API::RESPONSE_CODE_CAN_NOT_SEND
+ */
```
lysyjan, line 217: I have noticed that this comment is formatted incorrectly. Please, can you fix it?

mailpoet/lib/Mailer/Methods/ErrorMappers/MailPoetMapper.php
```diff
@@ -196,10 +196,32 @@ private function getEmailVolumeLimitReachedMessage(): string {
     return "{$message}<br/>";
   }
 
+  private function getPendingApprovalMessage(): string {
+    $message = __("Your subscription is currently [link]pending approval[/link].You'll soon be able to send once our team reviews your account. In the meantime, you can send previews to your authorized emails.", 'mailpoet');
```
lysyjan, line 200: If I remember correctly, we prefer typographic apostrophes in English texts. It means `’` instead of `'`. Please, could you replace it here?

mailpoet/lib/Mailer/Methods/ErrorMappers/MailPoetMapper.php
```diff
@@ -196,10 +196,32 @@ private function getEmailVolumeLimitReachedMessage(): string {
...
+    $message = Helpers::replaceLinkTags(
+      $message,
+      'https://kb.mailpoet.com/article/350-pending-approval-subscription',
+      [
+        'target' => '_blank',
+        'rel' => 'noopener noreferrer',
+        'data-beacon-article' => '5fbd3942cff47e00160bd248',
+      ],
+      'link'
```
lysyjan, line 209: Just note that `link` is the default value for the parameter `$linkTag` when the parameter is not specified. So, it would be possible to don't use this parameter in this case. Maybe you know about that, and it's up to you if the value will stay here.

lysyjan (commented):
Good job! I have noticed only a few minor details.

lysyjan (approved):

PR #4064, opened 2022-04-08
Alter CookieCest test for multisite [MAILPOET-4060]

websupporter:
Fixes [MAILPOET-4060]

In a multisite, the tracking cookie is not set when you sign up to the site and you checked the newsletter checkbox. The problem is that at that point in the process the headers are already sent, so we can not send the Cookie from the server:
image

This PR alters the test and checks for the cookie once the user follows the link to wp-activate.php (at which point, the cookie will be set) and prevents the attempt to set a cookie.

[MAILPOET-4060]: https://mailpoet.atlassian.net/browse/MAILPOET-4060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscription/Registration.php
+++ b/mailpoet/lib/Subscription/Registration.php
@@ -109,7 +109,14 @@ class Registration {
       $segmentIds
     );
 
-    // start subscriber tracking (by email, we don't have WP user ID yet)
-    $this->subscriberHandler->identifyByEmail($email);
+
+    /**
+     * On multisite headers are already sent at this point, tracking will start
+     * once the user has activated his account at a later stage.
+     **/
+    if (!headers_sent()) {
+      // start subscriber tracking (by email, we don't have WP user ID yet)
+      $this->subscriberHandler->identifyByEmail($email);
+    }
   }
 }
--- a/mailpoet/tests/acceptance/Subscribers/SubscriberCookieCest.php
+++ b/mailpoet/tests/acceptance/Subscribers/SubscriberCookieCest.php
@@ -38,11 +38,17 @@ class SubscriberCookieCest {
       $i->checkOption('#mailpoet_subscribe_on_register');
       $i->click('Next');
       $i->waitForText('mutestuser is your new username');
-    }
 
-    // temporarily bypass the test when MULTISITE=1 as it fails in multisite mode ATM
-    if (getenv('MULTISITE')) {
-      return;
+      /**
+       * The tracking cookie will be set once the registrant has activated
+       * the wp_user account
+       **/
+      $i->amOnMailboxAppPage();
+      $i->waitForElement('.subject.unread', 10);
+      $i->click(Locator::contains('span.subject.unread', 'Activate'));
+      $i->switchToIframe('#preview-html');
+      $i->click(Locator::contains('a', 'wp-activate.php'));
+      $i->waitForText('Your account is now active');
     }
 
     // subscriber cookie should be set right after signup


mailpoet/lib/Util/Cookies.php
```diff
@@ -15,21 +15,39 @@ class Cookies {
...
+    $secure = isset($options['secure']) && $options['secure'] ? '; secure' : '';
+    printf(
+      "<script>document.cookie='%s=%s; expires=%s; path=%s; domain=%s%s%s'</script>",
+      esc_js($name),
+      wp_json_encode($value, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES),
+      esc_js($utcDate->format(\DATE_RFC1123)),
+      esc_js($options['path']),
+      esc_js($options['domain']),
+      esc_js($httpOnly),
+      esc_js($secure)
```
JanJakes, line 50: @websupporter I see I'm assigned for review. I'm on L2 support so I shouldn't be but given that it's Friday afternoon, I can leave it for Monday anyway.

But I immediately have to ask — there's really no other solution to this? Such as setting the cookies earlier in the code that uses it? This looks quite hacky (or WordPressy 😆) to me. I really would expect a cookie service to set cookies and not print out some data. Do we really have no other choice? 😬

JanJakes: So could we do `if (headers_sent()) { throw new SomeException(...); }` and make sure we never attempt that? I'm also worried about security and compatibility — for instance, what would `httpOnly` even mean in JS? Would it work? And if so, would it really be HTTP-only?

websupporter: > I'm on L2 support

JanJakes: > We could hook into before_signup_header and check ourselfs, if its a user signup and whether the result of wpmu_validate_user_signup() is ok. That's a viable path.

I don't see into the details of it at all but if there's a viable path that would avoid the print, I'd say let's try to go for it!

> While I like the idea, simply throwing will likely throw exceptions in our integration test suite. But we can probably mock around it 🤔

Maybe it's not very safe in general. Instead of just some warnings in log files, it could break some sites where some plugins output stuff very early, etc. So I don't know, maybe safer to keep as-is?

websupporter:
In order to be able to set our cookie, we need to hook earlier into the signup process in a multisite. This means, we need to evaluate for ourselfs if a request is indeed a user signup. The new method `checkMultsiteRegister()` extracts the logic, which is present in _wp-signup.php_ and which leads to a signup. The disadvantage is that we duplicate this logic and would need to reflect changes in the WordPress core. WordPress' stance towards backwards compatibility should make this a robust solution for quite a while though.

mailpoet/lib/Subscription/Registration.php
```diff
@@ -63,19 +63,39 @@ public function extendForm() {
...
-      if (
-        isset($_POST['mailpoet']['subscribe_on_register'])
-        && (bool)$_POST['mailpoet']['subscribe_on_register'] === true
-      ) {
-        $this->subscribeNewUser(
-          $result['user_name'],
-          $result['user_email']
-        );
-      }
+  public function checkMultsiteRegister() {
```
JanJakes, line 66: To understand this, I have a dumb question at first. Why don't we just register [the tracking part](https://github.com/mailpoet/mailpoet/pull/4064/files#diff-13238361a5ac4b7b8247fcc268aa39e82c46483b5b464abc224d5400dc0dc4deL113) later? Like `$this->wp->addAction(...)` instead of direct `$this->subscriberHandler->identifyByEmail($email);`?

(Or maybe even [the subscriber handler](https://github.com/mailpoet/mailpoet/blob/fc1f3e6dc265f5617c44b49ac80fb149f151c43a/mailpoet/lib/Statistics/Track/SubscriberHandler.php#L60) could do it?)

websupporter:
After a quick internal discussion, I have updated the PR description with what we ended up as solution.

JanJakes (approved):
Great! 👏

veljkho (approved):

PR #4075, opened 2022-04-12
Check for uniqueness when updating segment name [MAILPOET-4199]

johnolek:
This PR ensures that we run our own check for uniqueness on a segment's name when updating an existing segment so users don't end up seeing a doctrine exception if they attempt to update a segment's name to a name that already exists.

Throwing our own `ConflictException` means that  we'll display a more user-friendly message, one that we're already using and have translations for, because [we catch that exception](http://github.com/mailpoet/mailpoet/blob/f4e30789afc2ef66ab4fcf8a091f8c33000e47eb/mailpoet/lib/API/JSON/v1/Segments.php#L128-L132).

It's worth noting that the `ConflictException` message itself, which refers to creating a new segment, is never displayed to the user.

Before:
image

After:
image

[MAILPOET-4199]

[MAILPOET-4199]: https://mailpoet.atlassian.net/browse/MAILPOET-4199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Test steps inside the Jira ticket.

### DIFF
--- a/mailpoet/lib/Segments/SegmentsRepository.php
+++ b/mailpoet/lib/Segments/SegmentsRepository.php
@@ -126,7 +126,10 @@ class SegmentsRepository extends Repository {
       if (!$segment instanceof SegmentEntity) {
         throw new NotFoundException("Segment with ID [{$id}] was not found.");
       }
-      $segment->setName($name);
+      if ($name !== $segment->getName()) {
+        $this->verifyNameIsUnique($name, $id);
+        $segment->setName($name);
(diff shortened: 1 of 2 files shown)


rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #4076, opened 2022-04-13
Mailpoet 4202 unify js ts exports

samnajian:
This PR removes default exports and adjusts the users of these modules as well as tests accordingly.

### How to test:
Attention! This PR require you to download a build from the Premium repo as well and test them together.
Please check the [PR for premium plugin](https://github.com/mailpoet/mailpoet-premium/pull/570) and download the build also from there and upload to the test site, then download the build from this PR and upload as well. Then test the following:
1. Please go to MailPoet > Subscribers and perform importing subscribers' list

[MAILPOET-4202]

[MAILPOET-4202]: https://mailpoet.atlassian.net/browse/MAILPOET-4202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eslint-config/.eslintrc.es6.json
```diff
@@ -41,13 +41,15 @@
...
+    ],
+    "import/no-default-export": 1 // no default exports
   },
   "overrides": [
     {
       "files": ["*.spec.js"],
       "rules": {
-        "no-unused-expressions": "off"
+        "no-unused-expressions": "off",
+        "import/no-default-export": 0
```
costasovo, line 52: Do we need to allow the default export for tests? I don't see any default exports there, and `npm run lint` is locally passing for me. Can we remove this line?

costasovo (commented):
Nice job!

I think we can forbid default exports also for tests. See my comment.

I noticed that there are a lot of changes changing  `import classnames from 'classnames';` to `import classNames from 'classnames';` This is probably not necessary but it is not a big deal.
This was probably caused by the tool that did the fix. Am I right? Just out of curiosity what tool did you use?

samnajian:
> Nice job!
> 
> I think we can forbid default exports also for tests. See my comment.
> 
> I noticed that there are a lot of changes changing `import classnames from 'classnames';` to `import classNames from 'classnames';` This is probably not necessary but it is not a big deal. This was probably caused by the tool that did the fix. Am I right? Just out of curiosity what tool did you use?

No specific tool was used, at first because of too many errors and webpack/eslint going crazy I started to find the root cause and play with this import and at that point it was renamed, I guess I can revert them back now :)

costasovo (approved):
Thanks! 👍 I think there is no need to revert the change in `classnames` imports.

veljkho (approved):

PR #4080, opened 2022-04-14
Fix acceptance subscriber cookie test [MAILPOET-4249]

costasovo:
I found 2 issues:

1) The account activation email was not delivered to MailHog, and clicking on the activation link was not working. Fixed in the first commit. Please see the info in the commit message.
2) The cookie was not set on the activation page. After discussing it with @JanJakes I fixed the test by adding logging and verifying the cookie after logging in. Please see more info in the commit message of the second commit.

I temporarily activated the multisite test on this branch. Here is [a passing test run](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/9553/workflows/5928a457-d433-436e-aa8d-b501cfb4b180/jobs/174950).

[MAILPOET-4249]

[MAILPOET-4249]: https://mailpoet.atlassian.net/browse/MAILPOET-4249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/DataFactories/Settings.php
+++ b/mailpoet/tests/DataFactories/Settings.php
@@ -214,4 +214,8 @@ class Settings {
     $this->settings->set('installed_at', $date);
     return $this;
   }
+
+  public function withTransactionEmailsViaMailPoet() {
+    $this->settings->set('send_transactional_emails', true);
+  }
 }
--- a/mailpoet/tests/acceptance/Subscribers/SubscriberCookieCest.php
+++ b/mailpoet/tests/acceptance/Subscribers/SubscriberCookieCest.php
@@ -43,7 +43,7 @@ class SubscriberCookieCest {
       $i->waitForText('mutestuser is your new username');
 
       /**
-       * The tracking cookie will be set once the registrant has activated
+       * The tracking cookie will be set once the registrant has activated and logged into
        * the wp_user account
        **/
       $i->amOnMailboxAppPage();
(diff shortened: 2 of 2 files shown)


brezocordero (approved):
Nice work!

> I found that container used for WordPress site in acceptance test environment
is not configured to send emails to MailHog and that caused that WP transactional emails were not caught by MailHog.

@costasovo Was it different before? Should we change it and configure it to send to MailHog or there is no need for that?

websupporter:
thank you @costasovo :heart:

costasovo:
> Nice work!
> 
> > I found that container used for the WordPress site in the acceptance test environment
> > is not configured to send emails to MailHog and that caused that WP transactional emails were not caught by MailHog.
> 
> @costasovo Was it different before? Should we change it and configure it to send to MailHog or there is no need for that?

It has always been like that. So it is not a regression on the container. Actually we have [just recently fix that for the dev env](https://mailpoet.atlassian.net/browse/MAILPOET-3746).

> Should we change it and configure it to send to MailHog, or there is no need for that?

There is an easy workaround so I would say adding this configuration is just nice to have. Our [WP image](https://github.com/mailpoet/wordpress-images/blob/master/wp-5.9/php8.1/Dockerfile) feels to be a more general-purpose image, so maybe we would need to add an image just for the acceptance test app.

PR #4085, opened 2022-04-18
Fix failing tests of AutomatedLatestContent::getPosts in MU

samnajian:
The issue was that on the CI we are switching the blog to a blog with id other than 1 when testing on MU.

### The fix:
Since we are not fully supporting MU a switch to blog_id = 1 was added to run the MU test only against the first blog.

[[MAILPOET-4082]]

[MAILPOET-4082]: https://mailpoet.atlassian.net/browse/MAILPOET-4082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brezocordero (approved):
Thanks for the quick fix @samnajian 
Approving as the test run successfully here: https://app.circleci.com/pipelines/github/mailpoet/mailpoet/9582/workflows/61b3cdca-6fc6-46ed-a65e-0fd90cab03f6

PR #4088, opened 2022-04-19
Fix strings causing broken translations [MAILPOET-4185]

johnolek:
These quotes were causing issues with translations where quote styles are changed to anything other than double quotes. This was affecting at least German and French translations.

Because the translations get applied before the search/replace, changing to something other than standard double quotes meant that the search string would never be found, leaving the characters `%1$s` in the UI.

In order to prevent untranslated strings, the new strings were added in https://github.com/mailpoet/mailpoet/pull/4091.

[MAILPOET-4185]

[MAILPOET-4185]: https://mailpoet.atlassian.net/browse/MAILPOET-4185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


johnolek:
The only concern I personally have with this change is that it represents a change to an existing string. Do we have a process for getting strings translated prior to a release, like could we update the translations in Transifex manually? I wouldn't want this to break the existing translations for all languages while we wait to get new translations for the following release.

brezocordero (commented):
Great catch and investigation! It works and the code looks good.

Regarding your concern about the translations, one option would be to modify only welcome.jsx and the translations would work with any language.
We would display the quotes around the role for all languages but we would not need to modify the translations.
I don't know if we want to always display the quotes around the role, if we do, I believe it is a best practice to leave them for the translators.

What do you think?

johnolek:
> Regarding your concern about the translations, one option would be to modify only welcome.jsx and the translations would work with any language.
We would display the quotes around the role for all languages but we would not need to modify the translations.
I don't know if we want to always display the quotes around the role, if we do, I believe it is a best practice to leave them for the translators.

I tried this out myself before I opened this PR, and I don't think we want to display quotes at all because we're currently always replacing them with a `Tag` component, which stands out well on its own.

Here's what it looks like currently:
image

If we only update `welcome.jsx` we end up with visuals like the following, which I think look pretty bad:

image
image
image

It seems like we never really wanted the quotes to begin with, which is why I think we should remove them from `newsletters.html` as well. They're really not a meaningful part of the string because they're always getting replaced.

I was hoping we could try to prepare the translation ahead of the release. To be safe we could still include the old string in our codebase so we're only adding strings, not removing any. I was thinking something like this:

1. Update this PR to include the old strings as well as the new, something like this:
```
'welcomeEventSegment': __('Sent when someone subscribes to the list: %1$s.'),
'welcomeEventSegment_old': __('Sent when someone subscribes to the list: "%1$s".'),
```
2. Regenerate the `.pot` file for the plugin from this branch (freshly rebased on `master`) and upload it to transifex
3. Wait for the new translations to be ready
4. Merge in this PR
5. After this work has been released, we can remove the `_old` strings in a separate PR.

Does that seem like an ok approach?

It seems like this is something we might want to establish a process for in general. We'll run into this same issue any time we want to update a string that's already been translated, right? 

## Bad Option
Another option, which I don't actually like and probably shouldn't even bring up, would be to make zero code changes and update our current translations instead such that they always include standard double quotes. This has already been done with one of the problematic German translations.

image

That feels like more of a bandaid than a true fix though, and it would just make things confusing in the translations for languages like French/German to have just 2 strings that use a different quote style. Not to mention the fact that we would have to ensure that any translation into a new language with different quote styles would need to have these exceptions. Again, probably not even worth bringing up this idea.

brezocordero:
Thanks for the explanation and the screenshots.

> Does that seem like an ok approach?

It does, it is the less disrupting option.

What do you think of creating two PRs? One PR with `mailpoet/views/newsletters.html` containing the new and old strings.
Once they are translated, a second PR with your changes (that will remove the old strings).

Like that we don’t need to manually generate the .pot file, it would be slower but we can follow the regular release process.

It is only a suggestion to avoid doing manual changes but you can decide.

> Another option, which I don't actually like

😂  we can pass, thanks for outlining all the options.

johnolek:
@brezo The more I think about it the more I want to avoid having to manually upload a `.pot` file outside of the regular release process. 

So let's go with your idea of two PRs, and I was also thinking I could go into transifex and manually update all the problematic translations that we're aware of right now. This is technically the "bad option," but since those translations will eventually be going away anyway I'm ok with the bandaid approach, and it would fix the issue from the user perspective with the next release.

johnolek:
I opened #4091 to add the new strings.

johnolek:
The new strings are in Transifex but haven't been translated yet so I'm keeping this as a draft for now.

johnolek:
It looks to me like the new versions of the strings have been translated, so this should be ready to review/merge.

JanJakes (dismissed):

veljkho:
Hi @johnolek 
I don't see artifact build due to old PR, CircleCI removes artifact builds after 30 days of inactivity. Can you please pull/push latest changes from trunk to rerun the workflow again? Thanks.

johnolek:
@veljkho I rebased onto `trunk` locally and force pushed, but apparently that invalidated the prior approval from Jan. Are you able to approve yourself, or should I request another review from someone else?

veljkho (approved):

veljkho:
> @veljkho I rebased onto `trunk` locally and force pushed, but apparently that invalidated the prior approval from Jan. Are you able to approve yourself, or should I request another review from someone else?

No worries, I can approve and merge. Thank you!

PR #4091, opened 2022-04-20
Add new versions of strings for translations [MAILPOET-4185]

johnolek:
This is step 1 in a two step process to fix translations that change double quotes to something else. Once the translations for these new strings have been completed we can update the application code to use the new strings, change the find/replace, and remove the old strings. 

See #4088 for part 2.

Just for good measure, I ran `./do translations:build` and confirmed that these new strings do make it into the generated `.pot` file. Not that there was any reason to doubt that would be the case! I just wanted to see it for myself.

[MAILPOET-4185]

[MAILPOET-4185]: https://mailpoet.atlassian.net/browse/MAILPOET-4185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brezocordero (approved):
LGTM!

johnolek:
This PR is only adding new strings so they get picked up for translation. Because it doesn't make any changes to logic/other application code there shouldn't be anything for QA to test.

PR #4093, opened 2022-04-21
Update info about supported languages in readme.txt [MAILPOET-4265]

costasovo:
[MAILPOET-4265]

[MAILPOET-4265]: https://mailpoet.atlassian.net/browse/MAILPOET-4265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/readme.txt
+++ b/mailpoet/readme.txt
@@ -107,11 +107,14 @@ Please note:
 **Official translations**
 
 * Arabic
+* Catalan
 * Chinese
+* Czech
 * Danish
 * Dutch
 * French (FR)
 * German
+* Greek
 * Italian
 * Japanese
 * Mexican
@@ -127,9 +130,7 @@ Please note:
 
 * Albanian
 * British
-* Catalan
 * French (CA)
-* Greek
 * Hebrew
 * Hungarian
 * Norwegian


johnolek (approved):
These changes look good to me! I did have one question though. The slack message about this refers to "readme translations". Do we somehow distinguish between readme translations and translations of plugin strings when it comes to supported languages? I'm guessing not but just wanted to double check.

costasovo:
> These changes look good to me! I did have one question though. The slack message about this refers to "readme translations". Do we somehow distinguish between readme translations and translations of plugin strings when it comes to supported languages? I'm guessing not but just wanted to double check.

Sorry for the confusion, my bad. I added the link to slack discussion because the original spec was unclear about whether we want to list those translations under official translations or community translations. Here is the comment which led to the ticket: (p1650536233038229/1650535389.613479-slack-C01HUN2F7EH), and then I asked to which list we want to add them p1650539540334139/1650535389.613479-slack-C01HUN2F7EH.

veljkho (approved):

PR #4095, opened 2022-04-22
Update PHP dependencies that require minimum PHP 7.2.5 [MAILPOET-4227]

lysyjan:
[MAILPOET-4227]

[MAILPOET-4227]: https://mailpoet.atlassian.net/browse/MAILPOET-4227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/DatabaseInitializer.php
+++ b/mailpoet/lib/Config/DatabaseInitializer.php
@@ -1,24 +1,22 @@
-<?php
+<?php declare(strict_types = 1);
 
 namespace MailPoet\Config;
 
 use MailPoetVendor\Doctrine\DBAL\Connection;
-use MailPoetVendor\Psr\Container\ContainerInterface;
 
 class DatabaseInitializer {
-  private $diContainer;
+  /** @var Connection */
+  private $connection;
 
   public function __construct(
-    ContainerInterface $diContainer
+    Connection $connection
   ) {
-    $this->diContainer = $diContainer;
+    $this->connection = $connection;
   }
 
   public function initializeConnection() {
-    $connection = $this->diContainer->get(Connection::class);
-
     // pass the same PDO connection to legacy Database object
     $database = new Database();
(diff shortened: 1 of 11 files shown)


mailpoet/lib/Tracy/DIPanel/DIPanel.php
```diff
@@ -60,10 +61,11 @@ private function getServicesFromContainer($name) {
     $containerWrapper = ContainerWrapper::getInstance();
     $reflection = new \ReflectionProperty(ContainerWrapper::class, $name);
     $reflection->setAccessible(true);
+    /** @var Container $container */
     $container = $reflection->getValue($containerWrapper);
     $reflection = new \ReflectionProperty(get_class($container), 'services');
     $reflection->setAccessible(true);
-    return [$reflection->getValue($container), $container->getDefinitions()];
+    return [$reflection->getValue($container), $container->getServiceIds()];
```
brezocordero, line 68: @lysyjan It all works great but with this change, the Tracy panel does not populate the `Used In` column.

Is this required for the upgrade? Or is this to remove the PHPStan warnings?

I don't see it as blocking, we could create a separate ticket to investigate if you prefer.

brezocordero, line 68: @lysyjan It all works great but with this change, the Tracy panel does not populate the `Used In` column.

Is this required for the upgrade? Or is this to remove the PHPStan warnings?

I don't see it as blocking, we could create a separate ticket to investigate if you prefer.

lysyjan: @brezocordero I noticed an error in the debug panel but after deeper investigation today. It looks like it was caused by generated container cache.
Thanks for this catch. I will drop this commit.

veljkho (approved):

PR #4097, opened 2022-04-25
Improve message for confirmation emails [MAILPOET-4149]

costasovo:
[MAILPOET-4149]

[MAILPOET-4149]: https://mailpoet.atlassian.net/browse/MAILPOET-4149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:
1. Use a third-party sending method in `MailPoet` > `Settings` > `Send With…` (can be Your Host)
2. Disable signup confirmation in `MailPoet` > `Settings` > `Sign-up Confirmation`
3. Go to Subscribers page and try to `resend email confirmation` for any `Unconfirmed` subscriber (in case you don't have, edit one and set it to be unconfirmed status)
4. Verify that you see the error message correctly, test the link inside the sentence and make sure it leads to the settings page
Note: Additionally, enable the Sign-up Confirmation in the settings and try to resent the confirmation, make sure you have received it and click the link in email to confirm the subscription. Make sure the subscriber is subscribed again.

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Subscribers.php
+++ b/mailpoet/lib/API/JSON/v1/Subscribers.php
@@ -4,6 +4,7 @@ namespace MailPoet\API\JSON\v1;
 
 use MailPoet\API\JSON\Endpoint as APIEndpoint;
 use MailPoet\API\JSON\Error as APIError;
+use MailPoet\API\JSON\Response;
 use MailPoet\API\JSON\ResponseBuilders\SubscribersResponseBuilder;
 use MailPoet\Config\AccessControl;
 use MailPoet\Doctrine\Validator\ValidationException;
@@ -13,12 +14,14 @@ use MailPoet\Exception;
 use MailPoet\Listing;
 use MailPoet\Models\Subscriber;
 use MailPoet\Segments\SegmentsRepository;
+use MailPoet\Settings\SettingsController;
 use MailPoet\Subscribers\ConfirmationEmailMailer;
 use MailPoet\Subscribers\SubscriberListingRepository;
 use MailPoet\Subscribers\SubscriberSaveController;
 use MailPoet\Subscribers\SubscribersRepository;
 use MailPoet\Subscribers\SubscriberSubscribeController;
 use MailPoet\UnexpectedValueException;
+use MailPoet\Util\Helpers;
 use MailPoet\WP\Functions as WPFunctions;
 
 class Subscribers extends APIEndpoint {
@@ -53,6 +56,9 @@ class Subscribers extends APIEndpoint {
   /** @var SubscriberSubscribeController */
   private $subscribeController;
 
+  /** @var SettingsController */
+  private $settings;
+
   public function __construct(
     Listing\Handler $listingHandler,
     ConfirmationEmailMailer $confirmationEmailMailer,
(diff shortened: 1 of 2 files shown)


rodrigoprimo (approved):
Good job! The changes look good to me.

The only thing that I noticed that is not directly related to this ticket is that if the top of the page is not visible because the user scrolled down a bit, the error message is not visible when they click on the link to resend the confirmation email. I think this can be confusing as the users might miss that there was an error when sending the confirmation email. They don't get any visible feedback that there was a problem when they clicked the link.

What do you think? If you agree, maybe we should create a new ticket?

costasovo:
> The only thing that I noticed that is not directly related to this ticket is that if the top of the page is not visible because the user scrolled down a bit, the error message is not visible when they click on the link to resend the confirmation email. I think this can be confusing as the users might miss that there was an error when sending the confirmation email. They don't get any visible feedback that there was a problem when they clicked the link.

Aha! Good catch. I haven't realized that. This seems to be a general problem with our notices.

> What do you think? If you agree, maybe we should create a new ticket?

I think it is a UX problem that may cause confusion. Do you have some suggestions on how to solve the issue? I can think about scrolling up when a notice appears or making the notice sticky.

rodrigoprimo:
> I think it is a UX problem that may cause confusion. Do you have some suggestions on how to solve the issue? I can think about scrolling up when a notice appears or making the notice sticky.

@costasovo, I also thought about scrolling up when a notice appears. Just making the notice stay on the page instead of disappearing after some time is not enough in my opinion. That being said, maybe we can get input from a designer on this? I'm not sure what is the best way to handle UX problems in MailPoet.

costasovo:
> @costasovo, I also thought about scrolling up when a notice appears. Just making the notice stay on the page instead of disappearing after some time is not enough in my opinion. 

@rodrigoprimo I've checked how it works in WP core listings (users and posts) and clicking on item actions in those pages triggers a page reload and after the reload the page is scrolled to the top and notice is visible. So Scrolling up seems to be a good option how to make the behavior consistent with WP.

> That being said, maybe we can get input from a designer on this? I'm not sure what is the best way to handle UX problems in MailPoet.

Well in the past we usually asked @NeosinneR or discussed it in the product channel. In this case, we probably don't need input from a designer because the scrolling options seem to be reasonable. But I would like to learn where we should ask or how to get input from a designer in case we needed it. Do you know where to ask?

rodrigoprimo:
> So Scrolling up seems to be a good option how to make the behavior consistent with WP.

:+1: 

> Do you know where to ask?

Unforutantely I don't know what is the process for MailPoet.

NeosinneR:
I don't think there is any official process for requesting a design opinion on little things like this for MailPoet, so as Rosta mentioned, you can ask in #mailpoet-product or me directly and I can try to help. WooDesign, however, aims to have a dedicated designer for MailPoet in the medium-term and there is a chance, that Mikalai, which was hired for the Automation project might take this role once he is onboarded and Automation is progressing.

rohankacharegs:
GlobalStep Verified

veljkho:
We have issue with the wording,
Currently: Please enable it to resend confirmation emails or update your subscriber’s `statuz manuoually`.
Expected: Please enable it to resend confirmation emails or update your subscriber’s `status manually.`

veljkho (changes requested):
We have issue with the wording,
Currently: Please enable it to resend confirmation emails or update your subscriber’s `statuz manuoually`.
Expected: Please enable it to resend confirmation emails or update your subscriber’s `status manually`.

veljkho (approved):
Verified

PR #4106, opened 2022-04-28
Fix type error on lifetime emails count task

brezocordero:
If a site has inactive subscribers setting active, and the cron task that calculated the lifetime emails count does not finish during a cron run, on the next run, the cron will throw a type mismatch error.

The meta that stores the process and is used on the next run was being saved as a string and the function expects an int.

This is not easy to test manually but there are two new tests, one that covers that the task runs using the meta and another that tests it also works with the meta that was saved as a string.

[MAILPOET-4281]

[MAILPOET-4281]: https://mailpoet.atlassian.net/browse/MAILPOET-4281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing:
1. Go to MailPoet > Lists
2. Test segmentation (creation, updating, recalcualtion..) and ensure there are no issues. Please try to send a newsletter to created segment list.

### DIFF
--- a/mailpoet/lib/Cron/Workers/SubscribersEmailCount.php
+++ b/mailpoet/lib/Cron/Workers/SubscribersEmailCount.php
@@ -60,8 +60,8 @@ class SubscribersEmailCount extends SimpleWorker {
     }
 
     $meta = $task->getMeta();
-    $lastSubscriberId = $meta['last_subscriber_id'] ?? 0;
-    $highestSubscriberId = $meta['highest_subscriber_id'] ?? $this->getHighestSubscriberId();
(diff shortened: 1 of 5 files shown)


costasovo (approved):
👍

veljkho:
Hi @brezocordero Please rebase the PR because it requires latest master on it in order to be testable. Thanks!

brezocordero:
@veljkho rebased!

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #4111, opened 2022-05-03
Fix HelpTooltip component import MAILPOET-4288

samnajian:
Fixes [MAILPOET-4288]

## Notes:
- Having js/ts (x) files with same name caused wrong imports, this is mitigated by making the file names different.

[MAILPOET-4288]: https://mailpoet.atlassian.net/browse/MAILPOET-4288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/help-tooltip-helper.js
+++ b/mailpoet/assets/js/src/help-tooltip-helper.js
@@ -14,5 +14,3 @@ export const MailPoetHelpTooltip = {
     );
   },
 };
-
-export { Tooltip };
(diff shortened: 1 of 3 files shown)


costasovo (approved):
Good job! Thanks for renaming the file. Having two files like that was really confusing even the PHPStorm IDE showed them as the source and compiled output in the file tree.

veljkho (approved):

PR #4113, opened 2022-05-03
Cleanup cached files in dev env on install [MAILPOET-4262]

costasovo:
This PR introduces a new command `./do cleanup:cached-files` that removes all files from the `generated` folder and PHPStan cached files. It also adds the command as the last step of `./do install` command. 

The idea behind this is that after updating dependencies often the generated files become outdated and may cause strange and confusing errors.

[MAILPOET-4262]

[MAILPOET-4262]: https://mailpoet.atlassian.net/browse/MAILPOET-4262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo (approved):
Looks good to me!

veljkho (approved):

PR #4115, opened 2022-05-04
Introduce switch to use ExtendSchema for newer WooCommerce block installations  [MAILPOET-4276]

websupporter:
Fixes [MAILPOET-4276] and [MAILPOET-4208]

The `ExtendRestApi::class` is deprecated since WooCommerce Blocks  version 7.2.0 [#](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e192c1d378c57a5dec459d653334c3f0969775ba/src/Domain/Bootstrap.php#L288-L294).

Since we want to remain compatible also with older versions of the WooCommerce Blocks plugin, this PR introduces a PluginHelper class with which we can easily check the version of a currently installed plugin.

Depending on this version, we load either the `ExtendRestApi::class` or the `ExtendSchema::class`.

[MAILPOET-4276]: https://mailpoet.atlassian.net/browse/MAILPOET-4276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-4208]: https://mailpoet.atlassian.net/browse/MAILPOET-4208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
+++ b/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
@@ -5,12 +5,15 @@ namespace MailPoet\PostEditorBlocks;
 use Automattic\WooCommerce\Blocks\Domain\Services\ExtendRestApi;
 use Automattic\WooCommerce\Blocks\Package;
 use Automattic\WooCommerce\Blocks\StoreApi\Schemas\CheckoutSchema;
+use Automattic\WooCommerce\StoreApi\Schemas\ExtendSchema;
+use Automattic\WooCommerce\StoreApi\StoreApi;
 use MailPoet\Config\Env;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Models\Subscriber;
 use MailPoet\Segments\WooCommerce as WooSegment;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Subscribers\SubscribersRepository;
+use MailPoet\WooCommerce\Helper as WooHelper;
 use MailPoet\WooCommerce\Subscription as WooCommerceSubscription;
 use MailPoet\WP\Functions as WPFunctions;
 
@@ -30,18 +33,23 @@ class WooCommerceBlocksIntegration {
   /** @var SubscribersRepository */
   private $subscribersRepository;
 
+  /** @var WooHelper  */
+  private $wooHelper;
+
   public function __construct(
     WPFunctions $wp,
     SettingsController $settings,
     WooCommerceSubscription $woocommerceSubscription,
     WooSegment $wooSegment,
-    SubscribersRepository $subscribersRepository
+    SubscribersRepository $subscribersRepository,
+    WooHelper $wooHelper
   ) {
     $this->wp = $wp;
     $this->settings = $settings;
     $this->woocommerceSubscription = $woocommerceSubscription;
     $this->wooSegment = $wooSegment;
     $this->subscribersRepository = $subscribersRepository;
+    $this->wooHelper = $wooHelper;
   }
 
   public function init() {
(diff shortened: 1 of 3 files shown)


mailpoet/tasks/phpstan/phpstan.neon
```diff
@@ -42,12 +42,14 @@ parameters:
     - '/Parameter #1 \$cssOrXPath of method AcceptanceTester::moveMouseOver\(\) expects string\|null, array<string, string> given./'
     - '/Function expect invoked with 1 parameter, 0 required\./'
     - '/Call to method getName\(\) on an unknown class _generated\\([a-zA-Z])*Cookie/' # codeception generate incorrect return type in ../../tests/_support/_generated
+    - '/Call to static method container\(\) on an unknown class/'
```
websupporter, line 45: Our current version of the [woocommerce-stubs](https://github.com/php-stubs/woocommerce-stubs) do not know the new class. I tried upgrading our package, which would solve this problem, but this resulted in many other unrelated problems, I couldn't figure out quickly. Therefore I just ignore the problem for now.

Any idea?

lysyjan: Maybe we can update the woocommerce-stubs in a different task. What do you think?

websupporter: Will create an issue.

mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
```diff
@@ -30,18 +33,23 @@ class WooCommerceBlocksIntegration {
...
   ) {
     $this->wp = $wp;
     $this->settings = $settings;
     $this->woocommerceSubscription = $woocommerceSubscription;
     $this->wooSegment = $wooSegment;
     $this->subscribersRepository = $subscribersRepository;
+    $this->pluginHelper = $pluginHelper;
   }
 
   public function init() {
```
lysyjan, line 55: I have tested your build, and the deprecation warning still occurs because we need to use a specific hook in the init method by the WC Block version. In my opinion, we should focus on hooks:
`woocommerce_blocks_checkout_update_order_from_request`
`woocommerce_store_api_checkout_update_order_from_request`
We know that the older hook `woocommerce_blocks_checkout_update_order_from_request` doesn't work in version 7.2.0, but I think that it will be solved by using the newer hook.
What do you think about it?

lysyjan (changes requested):

mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
websupporter: Makes sense. Will do and take MAILPOET-4208 also into account for this PR.

websupporter: Addressed in [b0b73de](https://github.com/mailpoet/mailpoet/pull/4115/commits/b0b73dea16a4dbc3f253f665b75936d908b40a88). Thanks @lysyjan

lysyjan: @websupporter I apologize, but warnings are still presented. I think the problem is bundled WC Blocks plugin in the WC plugin. You can see it in the installed WC plugin in the folder `packages`.  In my opinion, the plugin helper doesn't know about those plugins. I have found [another helper](https://github.com/mailpoet/mailpoet/blob/9f39843189084d2349d4bfbbaa175a3e45d2eac4/mailpoet/lib/WooCommerce/Helper.php#L10-L18). Maybe we can use this helper instead of new? I haven't checked how the WC plugin loads the WC Blocks plugin yet.

lysyjan (approved):
LGTM!
Thank you for being so patient.

veljkho (approved):

PR #4116, opened 2022-05-04
Add handling for more 5xx errors from the bridge [MAILPOET-4209]

lysyjan:
I have selected @brezocordero because you created the ticket.

I considered which HTTP codes we should handle, and I added three.
In my opinion, this should be enough, but we can discuss it.

[MAILPOET-4209]

[MAILPOET-4209]: https://mailpoet.atlassian.net/browse/MAILPOET-4209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brezocordero (approved):
> I considered which HTTP codes we should handle, and I added three.
In my opinion, this should be enough

Agreed.

LGTM! Thanks for adding the test.

veljkho (approved):

PR #4119, opened 2022-05-06
Optimize WordPress cron trigger SQL [MAILPOET-4314]

costasovo:
This change optimizes the SQL query that we run in every request when we check whether we should trigger the cron background process.

After some time the majority of scheduled task records in the DB are completed tasks.
In our cron trigger check, we check we need the info about completed tasks only for the Migration task.
Excluding other workers' completed tasks significantly speeds up the query execution on sites with a long history. E.g. on mailpoet.com the query sped up from 0.6s to 0.003s.

[MAILPOET-4314]

[MAILPOET-4314]: https://mailpoet.atlassian.net/browse/MAILPOET-4314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Triggers/WordPress.php
+++ b/mailpoet/lib/Cron/Triggers/WordPress.php
@@ -330,13 +330,15 @@ class WordPress {
         count(*) as count,
         case when scheduled_at <= :now then :past else :future end as scheduled_in
       from $scheduledTasksTableName
-      where deleted_at is null
+      where deleted_at is null AND (status != :statusCompleted OR status IS NULL OR `type` = :typeMigration)
       group by type, status, scheduled_in";
 
     $stmt = $this->entityManager->getConnection()->prepare($sql);
     $stmt->bindValue('now', date('Y-m-d H:i:s', $this->wp->currentTime('timestamp')));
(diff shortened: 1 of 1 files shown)


websupporter:
Thank you for this PR!

I wonder if we need `deleted_at` at all and if that would have an impact. Did you have a look? Other than this question. LGTM!

costasovo:
> I wonder if we need `deleted_at` at all and if that would have an impact. Did you have a look?

I checked the impact on mailpoet.com and the output of EXPLAIN and execution time were the same with or without the `deleted_at` condition.
I found only four tasks with deleted_at set to some actual value so it seems that we use the deleted_at column just very occasionally. These four tasks belong to emails in the trash. 
I think it is safer to keep `deleted_at` in the condition because theoretically there might be a deleted task with the status that would invoke cron to run but the cron worker would skip the task because it is deleted so we would end up in an endless loop.

websupporter:
thanks for having a look.

veljkho (approved):

PR #4121, opened 2022-05-10
Add user role hook [MAILPOET-2834]

websupporter:
Fixes [MAILPOET-2834]

Current behavior:
When an administrator changes the role of a user on the profile page, we synchronize the user which means, we send an welcome email, if for the role a welcome email exists.
When a plugin updates a user using either `$user->add_role()` or `$user->set_role()` we do not synchronize the user and thus not send the welcome email.

This PR enables us, to listen for those changes as well.

Notes:
00c34867da8eabeea4bec177e3798ef839c64014 We use `$user->add_role()` in our tests which lead to side effects (where e.g. a WooCommerce user is now a WordPress user). The commit extends the `\WPUserWithExtraProps` we use in our integration tests and overwrites the default `add_role()` and does not fire the hook. Question to me would be, if we should rename `WPUserWithExtraProps` :thinking: 

MAILPOET-2834 has a fix, which suggested to fire `profile_update` on `add_user_role` (at least it provided it as a workaround for users facing the problem). I did not go down this route, as `profile_update` suggests, we are on the profile page and third parties might try to hook in with this assumption.

I did a quick test in combination with WooCommerce. My fear: `add_user_role()` might lead to a wrong attribution of the `source` in the `wp_mailpoet_subscribers` table. This seems not to be the case. `woocommerce_new_customer` is fired after `set_role` and `user_register`, so there is no alteration of behavior and the source will be overwritten e.g. `woocommerce` or in case of the checkout with `woocommerce_checkout`.

[MAILPOET-2834]: https://mailpoet.atlassian.net/browse/MAILPOET-2834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/integration/Segments/WPUserWithExtraProps.php
+++ b/mailpoet/tests/integration/Segments/WPUserWithExtraProps.php
@@ -1,7 +1,0 @@
-<?php
-
-namespace MailPoet\Test\Segments;
-
-class WPUserWithExtraProps extends \WP_User {
-  public $orderId;
-}
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -261,6 +261,11 @@ class Hooks {
       6, 2
     );
     $this->wp->addAction(
+      'add_user_role',
+      [$this->wpSegment, 'synchronizeUser'],
+      6, 1
+    );
+    $this->wp->addAction(
       'delete_user',
       [$this->wpSegment, 'synchronizeUser'],
       1
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -266,6 +266,11 @@ class Hooks {
       6, 1
     );
     $this->wp->addAction(
+      'set_user_role',
+      [$this->wpSegment, 'synchronizeUser'],
+      6, 1
+    );
+    $this->wp->addAction(
       'delete_user',
       [$this->wpSegment, 'synchronizeUser'],
       1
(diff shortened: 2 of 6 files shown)


websupporter:
To test:
I did set up a welcome email for the `contributor` role and installed the following mu plugin. Then I could just toggle between adding and removing the role with `http://test.local?test=<userId>`

The expected behavior: The first time I add the role, the welcome email should get send. The second time I add the role, the email should not get send again.

```
<?php

/**
 * Plugin Name: Test add_role sends welcome email
 **/

add_action(
	'init',
	function() {
		if (! isset($_GET['test'])) {
			return;
		}
		
		$user = new \WP_User($_GET['test']);
		if (! in_array('contributor', $user->roles)) {
			$user->add_role('contributor');
			wp_die('Role added');
		} else {
			$user->remove_role('contributor');
			wp_die('Role removed');
		}
	}
);
```

mailpoet/tests/integration/Segments/WPUserWithExtraProps.php
```diff
@@ -4,4 +4,20 @@
 
 class WPUserWithExtraProps extends \WP_User {
   public $orderId;
+
+  /**
+   * add_role contains the 'add_role' hook, which creates side effects,
+   * we do not want. Therefore, we overwrite this method.
```
JanJakes, line 10: Could you maybe explain it in the comment even a bit more? I'm afraid later on we'll forget what side effects and why we don't want them.

And relatedly:
> Question to me would be if we should rename WPUserWithExtraProps 🤔

I think yes, we can use some more generic name like `WPTestUser` or something like that.

JanJakes (commented):
LGTM! Just one comment about docs/naming.

mailpoet/tests/integration/Segments/WPUserWithExtraProps.php
websupporter: Thanks for the review @JanJakes. Addressed in [f616896](https://github.com/mailpoet/mailpoet/pull/4121/commits/f6168968b6fcf126e199e1093acbe2bca89daec6)

JanJakes (approved):

veljkho (approved):

PR #4124, opened 2022-05-10
Deprecate some Paris models [MAILPOET-4150]

rodrigoprimo:
This PR removes usages from test classes and then deprecates the following models:

- MailPoet\Models\DynamicSegmentFilter: simply deprecated as it was already not being used.
- MailPoet\Models\StatisticsClicks: calls in the test classes replaced with Doctrine code and model deprecated.
- MailPoet\Models\StatisticsOpens: calls in the test classes replaced with Doctrine code and model deprecated.
- MailPoet\Models\StatisticsWooCommercePurchases: calls in the test classes replaced with Doctrine code and model deprecated.

This PR is the first in a few that I plan to create to deprecate all the models that are used only in the tests and are listed in the ticket [MAILPOET-4150]. I'm opting to split this ticket into multiple PRs to make reviewing the code changes easier.

[MAILPOET-4150]

[MAILPOET-4150]: https://mailpoet.atlassian.net/browse/MAILPOET-4150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4150]: https://mailpoet.atlassian.net/browse/MAILPOET-4150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/integration/Cron/Workers/StatsNotifications/AutomatedEmailsTest.php
+++ b/mailpoet/tests/integration/Cron/Workers/StatsNotifications/AutomatedEmailsTest.php
@@ -5,10 +5,15 @@ namespace MailPoet\Cron\Workers\StatsNotifications;
 use Codeception\Stub;
 use MailPoet\Config\Renderer;
 use MailPoet\Cron\CronWorkerRunner;
+use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Entities\NewsletterOptionEntity;
+use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Entities\SendingQueueEntity;
+use MailPoet\Entities\StatisticsClickEntity;
+use MailPoet\Entities\StatisticsOpenEntity;
 use MailPoet\Mailer\Mailer;
 use MailPoet\Mailer\MailerFactory;
 use MailPoet\Mailer\MetaInfo;
-use MailPoet\Models\Newsletter;
 use MailPoet\Models\ScheduledTask;
 use MailPoet\Models\SendingQueue;
 use MailPoet\Models\StatisticsClicks;
@@ -17,7 +22,7 @@ use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Statistics\NewsletterStatisticsRepository;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Settings\TrackingConfig;
-use MailPoetVendor\Idiorm\ORM;
+use MailPoet\Test\DataFactories\Newsletter as NewsletterFactory;
 use PHPUnit\Framework\MockObject\MockObject;
 
 class AutomatedEmailsTest extends \MailPoetTest {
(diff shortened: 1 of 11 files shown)


mailpoet/tests/integration/Cron/Workers/StatsNotifications/AutomatedEmailsTest.php
```diff
@@ -134,9 +133,15 @@ public function testItDoesntRenderIfNoNewslettersFound() {
...
-    $this->createOpens($this->newsletter->getId(), 2);
+    $newsletter = $this->newsletterFactory
+      ->withSubject('Subject')
+      ->withWelcomeTypeForSegment(1)
+      ->withActiveStatus()
+      ->withSendingQueue(['count_processed' => 10])
+      ->create();
+
+    $this->createClicks($newsletter, 5);
+    $this->createOpens($newsletter, 2);
```
costasovo, line 144: It looks like this part of data preparation is the same in every test case. Shouldn't we move it to the _before function? Or do you prefer having it in each test for better readability?

costasovo (approved):
Great job! Thanks for refactoring the tests. This helps to move us faster in the doctrine project 👍 

I have added one question. Just out of curiosity.

veljkho (dismissed):

mailpoet/tests/integration/Cron/Workers/StatsNotifications/AutomatedEmailsTest.php
rodrigoprimo: I haven't noticed this repetition when refactoring this part of the code, to be honest.

My first instinct when I saw your question was to say that we should move it to _before() as this is where I believe all setup code should be to make the test methods smaller and easier to read. But then it occurred that it might also make sense to consider in this particular moving this setup code to a method called createNewsletterClicksAndOpens() or something like that and call this method from within the test methods. Just to make it more clear that those specific tests rely on the existence of a newsletter with some clicks and opens while the others don't. Just rambling here, would have to take a closer look to decide which of the two options I prefer.

Happy to hear your thoughts. That is a good question.

rodrigoprimo:
> Note: Since there will be more PRs I kept the ticket in the code review column on Jira.

Thanks!

mailpoet/tests/integration/Cron/Workers/StatsNotifications/AutomatedEmailsTest.php
costasovo: I like the idea of `createNewsletterClicksAndOpens` method as it solves the code duplicity and its usage in tests more explicitly declares that the test needs those data.

rodrigoprimo: @costasovo, since this PR is not merged yet, I went ahead and created the method createNewsletterClicksAndOpens() in c4174b4. @veljkho already approved this PR, and since I only did a small refactor in an integration test, I guess @costasovo can merge it directly after checking the changes that I made, provided there are no issues. But let me know if you prefer a different approach.

costasovo (approved):
Thanks for refactoring the test!

PR #4125, opened 2022-05-11
Replace swift mailer with PHPMailer [MAILPOET-4142]

lysyjan:
This PR aims to remove the dependency on the Swift Mailer library that is not maintained.

I tried to reuse code from PHPMailMethod and its error mapper. Also, it was necessary to make changes to the AmazonSES method. Unfortunately, I could not find or test all possible errors for SMTP.

I will be happy with each feedback or idea!

How to test:
Configure sending with SMTP or AmazonSES and send a newsletter. Don't forget to send previews etc.

[MAILPOET-4142]

[MAILPOET-4142]: https://mailpoet.atlassian.net/browse/MAILPOET-4142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions:
1. Go to MailPoet > Settings > Send with tab
2. Click to configure custom sending
3. Select SMTP (it should have already saved data for SMTP)
4. Save it
5. Send a few newsletters and test email previews


mailpoet/lib/Mailer/Methods/SMTP.php
```diff
@@ -53,89 +45,25 @@ public function __construct(
...
+  public function buildMailer(): PHPMailer {
+    $mailer = new PHPMailer(true);
+    $mailer->isSMTP();
+    $mailer->Host = $this->host; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+    $mailer->Port = $this->port; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+    $mailer->SMTPSecure = $this->encryption;
+    /** @phpstan-ignore-next-line - we cannot annotate the return type from a filter */
+    $mailer->SMTPOptions = $this->wp->applyFilters('mailpoet_mailer_smtp_options', []);
+    /** @phpstan-ignore-next-line - we cannot annotate the return type from a filter */
+    $mailer->Timeout = $this->wp->applyFilters('mailpoet_mailer_smtp_connection_timeout', self::SMTP_CONNECTION_TIMEOUT); // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
```
websupporter, line 60: Can't you do like `$mailer->Timeout = (int) $this->wp->applyFilters()`? Or is the problem something else?

mailpoet/lib/Mailer/WordPress/WordPressMailer.php
```diff
@@ -71,7 +73,7 @@ public function send() {
         $sendWithMailer($useFallback = true);
       } catch (\Exception $fallbackMailerException) {
         // throw exception passing the original (primary mailer) error
-        throw new \phpmailerException($e->getMessage(), $e->getCode(), $e);
+        throw new PHPMailerException($e->getMessage(), $e->getCode(), $e);
```
websupporter, line 76: These exception changes are because of the Namespace changes in WordPress core, right?

mailpoet/lib/Util/Notices/DeprecatedFilterNotice.php
```diff
@@ -0,0 +1,49 @@
+<?php
```
websupporter, line 1: declare strict_types

mailpoet/lib/Util/Notices/DeprecatedFilterNotice.php
```diff
@@ -0,0 +1,49 @@
...
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Util\Helpers;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+/**
+ * This can be removed after 2022-12-01
+ */
+class DeprecatedFilterNotice {
```
websupporter, line 12: Why not using `$this->wp->deprecatedHook`?

websupporter:
Thank you for this PR @lysyjan. Just a few remarks. Overall lgtm :bowing_man:

mailpoet/lib/Mailer/Methods/SMTP.php
lysyjan: I tried it, but it doesn't work. PHPstan reports an error `Cannot cast mixed to int.`
It makes sense to me because the filter can return anything.

mailpoet/lib/Util/Notices/DeprecatedFilterNotice.php
lysyjan: There are three main reasons:
1. If I am not wrong `deprecatedHook` works only when the hook is called.
2. We don't have anymore a class from Swift Mailer that is necessary as instance in filter parameter.
3. The filter is called in a background job, so a customer shouldn't notice

mailpoet/lib/Mailer/WordPress/WordPressMailer.php
lysyjan: This is because I changed the class `PHPMailerLoader`.

mailpoet/lib/Util/Notices/DeprecatedFilterNotice.php
lysyjan: Fixed in the latest commit

websupporter: Thanks!

websupporter (approved):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

lysyjan:
Hello @soderlind,
We have noticed that you are using the filter `mailpoet_mailer_smtp_transport_agent`. Unfortunately, we needed to replace this filter with a new filter `mailpoet_mailer_smtp_options`, because the Swift Mailer is not maintained anymore. Those changes will cause your plugin [mu-mailpoet-smtp](https://github.com/soderlind/mu-mailpoet-smtp) to stop working.

We apologize for the complications.

websupporter:
I will create also an issue on his repo.

PR #4128, opened 2022-05-12
Remove deprecated code after six months grace period [MAILPOET-4325]

rodrigoprimo:
[MAILPOET-4325]

[MAILPOET-4325]: https://mailpoet.atlassian.net/browse/MAILPOET-4325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/tests/integration/_bootstrap.php
```diff
@@ -30,50 +30,16 @@
...
-array_map($destroy, $models);
-
 $entityManager = ContainerWrapper::getInstance(WP_DEBUG)->get(EntityManager::class);
-foreach ($entities as $entity) {
-  $tableName = $entityManager->getClassMetadata($entity)->getTableName();
-  $connection->query('SET FOREIGN_KEY_CHECKS=0');
-  $connection->executeUpdate("TRUNCATE $tableName");
-  $connection->query('SET FOREIGN_KEY_CHECKS=1');
+$entitiesMeta = $entityManager->getMetadataFactory()->getAllMetadata();
+foreach ($entitiesMeta as $entityMeta) {
```
triple0t, line 37: Would this fetch the old Paris ORM models?

If not, Can we keep the function truncating the tables around till all the models have been migrated?

triple0t (commented):
Looks good.
I have some comments, kindly locate them below.

mailpoet/tests/integration/_bootstrap.php
rodrigoprimo: Thanks for the review, @triple0t! This doesn't fetch old Paris ORM models, but unless I'm missing something it is not a problem. My understanding is that what we want to do in this part of the code is truncate all the MailPoet tables. So it doesn't matter if we are getting all the tables from Paris or Doctrine as long as we get all the tables. As you pointed out above, some of the old Paris models are still being used indeed, but we already have correspondent Doctrine entities for those models.

As pointed out in the commit message, I opted to include this change here to make it easier when removing future models and to make sure the list of tables is always updated when we add new tables in the future.

What do you think?

mailpoet/tests/integration/_bootstrap.php
```diff
@@ -30,50 +30,16 @@
...
-  'ScheduledTask',
-  'ScheduledTaskSubscriber',
-  'SendingQueue',
-  'Subscriber',
-  'SubscriberCustomField',
-  'SubscriberSegment',
-  'StatisticsOpens',
-  'StatisticsClicks',
-  'StatisticsNewsletters',
-];
```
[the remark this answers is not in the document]
rodrigoprimo: Please see https://github.com/mailpoet/mailpoet/pull/4128#discussion_r872359240

mailpoet/tests/integration/_bootstrap.php
triple0t: Understood.
Thanks for the explanation, Rodrigo.

triple0t (approved):
LGTM

veljkho (approved):

PR #4130, opened 2022-05-13
Update js cookie mailpoet 4214

samnajian:
## This PR:

- Updates modules which use js-cookie to ts
- Adds latest js-cookie package and it's type definitions

[MAILPOET-4214]

[MAILPOET-4214]: https://mailpoet.atlassian.net/browse/MAILPOET-4214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/ajax.ts
```diff
@@ -59,7 +59,9 @@ export const MailPoetAjax = {
     token: null,
     data: {},
   },
-  post: function post(options): ResponseType {
+  post: function post<R = Response, ER = ErrorResponse>(
+    options,
+  ): ResponseType<R, ER> {
     return this.request('post', options);
   },
   get: function get(options): ResponseType {
```
rodrigoprimo, line 67: Why we don't need to change ResponseType here as well?

rodrigoprimo (commented):
I tested the forms functionality that uses js-cookies and confirmed that it is still working as expected. Code changes also look good to me. That being said, my knowledge about TS is very limited, so if you are unsure about something, please feel free to ask for a second review from another developer. If not, you can move this PR to QA.

I also left two in-line comments, but they are probably not blockers. Unless the to-do item is something that you were planning to add to this PR and forgot about it.

mailpoet/assets/js/src/public.tsx
```diff
@@ -48,7 +49,7 @@ jQuery(($) => {
     }
   }
 
-  $('.mailpoet_recaptcha').each((index, element) => {
+  $('.mailpoet_recaptcha').each((_, element) => {
```
rodrigoprimo, line 52: There is one more thing that I want to add that occurred to me after I submitted the review. Initially, I thought that `_` was the TS way to mark an unused parameter, but then I realized thanks to a conversation with @brezocordero that it is actually a prefix to mark an unused parameter (https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals). With that in mind, I would suggest we use a more explicit name like `_unused`. This should help those that are not super familiar with TS to understand what is going on in this like. What do you think?

mailpoet/assets/js/src/form_editor/store/controls.tsx
```diff
@@ -252,18 +268,17 @@ export const controls = {
...
 
   CALL_API,
 
   ENSURE_BROWSER_URL(actionData) {
+    const { formId } = actionData as Record<string, string>;
     let url = select('mailpoet-form-editor').getFormEditorUrl();
-    url = `${url}${actionData.formId}`;
-    if (window.location !== url) {
+    url = `${url}${formId}`;
+    if (window.location.href !== url) {
```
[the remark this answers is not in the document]
samnajian: Yes, that's a bug fix, added a separate commit for it :)

mailpoet/assets/js/src/public.tsx
rodrigoprimo: Thanks for changing this detail, Sam. I would suggest we update the code in all the places where `_` was used and instead of just in this line.

Also, I would recommend avoiding doing changes in the commits after the PR was submitted for review like it was done for `_` -> `_unused`. In my opinion, adding new commits is preferred in those situations. This because it is harder for the reviewer to see what changed and what didn't change in the original commits when force push is used. What do you think?

mailpoet/assets/js/src/global.d.ts
```diff
@@ -111,4 +112,8 @@ interface Window {
   mailpoet_date_display_format?: string;
   mailpoet_date_storage_format?: string;
   mailpoet_current_date_time?: string;
+  recaptcha?: any; // @Todo: add definition
```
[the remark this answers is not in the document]
rodrigoprimo: Similar to https://github.com/mailpoet/mailpoet/pull/4130#discussion_r881519245, for future PRs, I would recommend doing similar changes in a separate commit. Initially, I missed that the comment above was addressed and I thought you had missed it. But then going back to the original commit, I saw that this is now resolved.

rodrigoprimo:
Thanks for the changes, Sam. I left a few more comments above. Also, there are conflicts in public.jsx that we need to fix before moving this PR forward.

mailpoet/assets/js/src/public.tsx
samnajian: @rodrigoprimo thanks for the note, but I'm fine with both since even with an interactive rebase the reviewer is still able to see the older comment ( I should agree sometimes I get's harder to notice the improvement ), with this, said I'd agree with a separate commit after the PR has been submitted.

mailpoet/assets/js/src/form_editor/store/controls.tsx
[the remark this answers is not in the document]
samnajian: Good point @rodrigoprimo, I had tested with the new change and couldn't  find anything breaking, basically the previous state would always run `history.replaceState` but since the url in some cases was the same nothing would get triggered, with the added change the `history.replaceState` is only invoked when the url changes which will trigger route listeners, I'd say this change is pretty safe, but if you have confidence that this change can be causing issues I'm happy to revert it back, ts-ignore and create a bug ticket, what do you think?

samnajian:
> Thanks for the changes, Sam. I left a few more comments above. Also, there are conflicts in public.jsx that we need to fix before moving this PR forward.

Thanks for checking, master is rebased and the issues arising from the rebase are adjusted.

mailpoet/assets/js/src/form_editor/store/controls.tsx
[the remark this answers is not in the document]
rodrigoprimo: Thanks for the additional details. In this case, we can proceed with things as they are.

mailpoet/assets/js/src/public.tsx
rodrigoprimo: > I would suggest we update the code in all the places where _ was used and instead of just in this line.

I think we still need to update the above.

rodrigoprimo:
@samnajian, CircleCI is failing, could you please take a look?

rodrigoprimo:
@samnajian, I don't know what is happening, but I was not able to re-request a review so that I can approve this PR. Which is very unusual. Anyway, the code changes look good to me and we can move this to QA for @veljkho to check. An approval from him should be enough to unblock merging this PR if he doesn't find any issues.

veljkho (approved):

PR #4131, opened 2022-05-13
Automation UI with Gutenberg components [MAILPOET-4287]

JanJakes:
[MAILPOET-4287]


mailpoet/assets/js/src/automation/editor/components/header/more_menu.tsx
```diff
@@ -0,0 +1,35 @@
+import { MenuGroup } from '@wordpress/components';
+import { displayShortcut } from '@wordpress/keycodes';
+import { __, _x } from '@wordpress/i18n';
```
websupporter, line 3: just as a side note. not sure if we can with our current translation system utilize the javascript translation functions.

mailpoet/assets/js/src/automation/editor/components/icons/delay.tsx
```diff
@@ -0,0 +1,20 @@
+export function DelayIcon(): JSX.Element {
```
websupporter, line 1: i guess this depends on the actual implementation on how a `Step` should be rendered. I wonder if step specific files (like icons) should be organized separately. What I mean, to store them in a separate directory like `automation/steps/delay/icon.tsx`. Would help probably also with later having a workflow system with a different use case where e.g. a delay would not be needed.

mailpoet/assets/js/src/automation/editor/components/icons/plus.tsx
```diff
@@ -0,0 +1,14 @@
+export function PlusIcon(): JSX.Element {
```
websupporter, line 1: would it make sense to use @wordpress/icons to keep icons consistent with WordPress?

mailpoet/assets/js/src/automation/editor/components/inserter-listbox/listbox_group.tsx
```diff
@@ -0,0 +1,40 @@
+import { HTMLAttributes } from 'react';
+import { speak } from '@wordpress/a11y';
```
websupporter, line 2: regarding a11y. What I noticed, I couldn't navigate to the Workflow itself using `TAB` :thinking:

mailpoet/assets/js/src/automation/editor/components/header/more_menu.tsx
JanJakes: Yeah... 🤔 This is pretty much copied from the post editor itself and I guess the strings that are identical may be available (although not sure if they need to be loaded) and it would be strange to rely on that anyway.

I have a "JS translations" point in the [checklist draft P2](https://wordpress.com/post/mailpoetdev.wordpress.com/1276) so maybe we could address that separately?

mailpoet/assets/js/src/automation/editor/components/icons/delay.tsx
JanJakes: Makes sense. These things [are definitely rather mocked](https://github.com/mailpoet/mailpoet/pull/4131/commits/48fd9aa93c2bf5100a57a4a8b9b122331a0ae30d#diff-9c612c5c5cd17c514082571ad8f1a3469ddce34f3b89dd8080305bb59a90e9b6R8) for now. Maybe again something to improve on later? Should I add it to the checklist?

mailpoet/assets/js/src/automation/editor/components/header/more_menu.tsx
websupporter: Sure, just came to my mind when I saw it.

mailpoet/assets/js/src/automation/editor/components/inserter-listbox/listbox_group.tsx
JanJakes: It seems to work for me. TAB to get to the workflow, then arrow up/down to move within:

https://user-images.githubusercontent.com/141436/169025222-ce7e10bd-93e0-4e0a-8053-6ed4fd2fe073.mov

mailpoet/assets/css/src/components-automation-editor/_add-step-button.scss
```diff
@@ -0,0 +1,14 @@
+.mailpoet-automation-editor-add-step-button {
```
websupporter, line 1: regarding naming: how `mailpoet-` specific should it be?

mailpoet/assets/js/src/automation/editor/components/icons/plus.tsx
JanJakes: Do you mean the simple icons like "+" specifically or "all" icons? I think we won't find suitable icons for "delay" and things like that but I just added [a commit](https://github.com/mailpoet/mailpoet/pull/4131/commits/4c7b13f981dab333d8c5172ade49e999adc6aad5) to reuse the plus icon as I was able to get it very close to the design using the correct size and fill.

mailpoet/assets/js/src/automation/editor/components/icons/delay.tsx
websupporter: sounds good :)

mailpoet/assets/css/src/components-automation-editor/_add-step-button.scss
JanJakes: I've discussed this dilemma with Tautvidas and the decision was to prefix hooks and stuff with `mailpoet` for now. Seems like the requirement is "being able to extract it and make reusable quickly" rather than "having it completely reusable immediately". We don't have a more universal name at the moment (apart from "automation" maybe) so I guess we can keep the prefixes now (or omit the `mailpoet-` part).

mailpoet/assets/js/src/automation/editor/components/inserter-listbox/listbox_group.tsx
websupporter: found the issue. seems firefox does not like the css property https://github.com/mailpoet/mailpoet/pull/4131/files#r875785786

JanJakes: @websupporter Good catch! Makes sense, as I forgot the vendor-specific property there. Does [this](https://css-tricks.com/copy-the-browsers-native-focus-styles/) work for you or should I use just a fixed color? 🤔

websupporter: That should work :) :bow:

JanJakes: Or maybe we should do what WP does:

Screen Shot 2022-05-18 at 13 45 20

Not sure if it's safe to use those variables directly though...

websupporter: When I see this correctly, we have those variables, because we enqueue 'wp-edit-post' and the colors get overwritten, when I change the styles in my profile. I would like to see this for consistency and would think those variables fall under the "backwards compatibility" policy of WP.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -0,0 +1,69 @@
...
+    // Gutenberg styles
+    $this->wp->wpEnqueueStyle('wp-edit-post');
+    $this->wp->wpEnqueueStyle('wp-format-library');
+    $this->wp->wpEnqueueMedia();
+
+    $id = isset($_GET['id']) ? (int)$_GET['id'] : null;
+
+    $workflow = $id ? $this->workflowStorage->getWorkflow($id) : null;
+    $this->pageRenderer->displayPage('automation/editor.html', [
+      'workflow' => $workflow ? $this->buildWorkflow($workflow) : null,
```
websupporter, line 42: I assume a fallback when no workflow was found would be handled at a later stage.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -0,0 +1,69 @@
...
+      'sub_menu' => 'mailpoet-automation',
+      'api' => [
+        'root' => rtrim($this->wp->escUrlRaw($this->wp->restUrl()), '/'),
+        'nonce' => $this->wp->wpCreateNonce('wp_rest'),
+      ],
+    ]);
+  }
+
+  private function buildWorkflow(Workflow $workflow): array {
+    return [
```
websupporter, line 52: `Workflow::toArray()` does not work here?

mailpoet/lib/Config/Menu.php
```diff
@@ -412,6 +413,23 @@ public function setup() {
         'mailpoet-automation',
         [$this, 'automation']
       );
+
+      // Automation editor
+      $automationEditorPage = $this->wp->addSubmenuPage(
+        true,
+        $this->setPageTitle('Automation Editor'),
+        'Automation Editor',
+        AccessControl::PERMISSION_MANAGE_EMAILS,
```
websupporter, line 422: Afaik the default `capability` for the rest api endpoint is `administrator` while here it is `mailpoet_manage_emails` (defaults to admin and editor). I think this needs to be streamlined. https://github.com/mailpoet/mailpoet/blob/master/mailpoet/lib/Automation/Engine/API/Endpoint.php#L12-L14

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
JanJakes: Yeah, noting it.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
JanJakes: Not really. It's because `Workflow::toArray()` is for internal data saving, while here we're serving it to the output. What we may do here is to reuse a `GET /workflows` endpoint implementation internally, rather than duplicating it. Ultimately, it's what that is — just an optimization to avoid one extra request.

mailpoet/assets/js/src/automation/editor/components/inserter-listbox/listbox_group.tsx
JanJakes: Addressed in https://github.com/mailpoet/mailpoet/pull/4131/commits/e4c2da9ca1033204a142eb13c6fd0a214c9436a0.

mailpoet/lib/Config/Menu.php
JanJakes: 👍 Addressed in https://github.com/mailpoet/mailpoet/pull/4131/commits/7dc50e1454785c97ac033182cbb91f0909829dfe. For now I just used MailPoet's AccessControl class and noted we may extract it in the future.

mailpoet/assets/js/src/automation/editor/components/icons/plus.tsx
websupporter: > I was able to get it very close to the design

That's right, the PR is probably not the right place to bring this up. Thank you!

websupporter (approved):
Awesome! Thank you very much for working on this. LGTM

veljkho (approved):

PR #4137, opened 2022-05-17
memory limit when building translations

samnajian:
Remove memory limit when building translations


costasovo (approved):
Thank you for the fix!

PR #4138, opened 2022-05-17
Remove some PHPStan Level 6 errors

brezocordero:
This PR removes the following PHPStan level 6 errors from `MailPoet\Segments\DynamicSegments` (first and second commit) and from `MailPoet\Segments\SegmentsRepository` (third commit)

```
(Method|Property|Function) has no (return )?type specified.
(Method|Function) has parameter with no type (specified)
```

To test you can modify the level6 exclusion rule on phpstan.neon file to this:

```
    # exclude level 6 errors (but keep them for Automation)
    - '/(Method|Property|Function) (?!(MailPoet\\Automation\\)|(MailPoet\\Segments\\DynamicSegments\\)|(MailPoet\\Segments\\SegmentsRepository)).*has no (return )?type specified/'
    - '/(Method|Function) (?!(MailPoet\\Automation\\)|(MailPoet\\Segments\\DynamicSegments\\)|(MailPoet\\Segments\\SegmentsRepository)).*has parameter (\$[_A-Z]{1}[_a-z]+)? with no type (specified)?/i'
```
and run `./do qa:phpstan`

There are too many errors to be handled in one PR, I am working on a second PR and will create a follow-up issue for the remaining errors.

[MAILPOET-3720]

[MAILPOET-3720]: https://mailpoet.atlassian.net/browse/MAILPOET-3720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
+++ b/mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
@@ -153,23 +153,17 @@ class FilterFactory {
     return $this->emailAction;
   }
 
-  /**
-   * @return WooCommerceMembership
-   */
-  private function wooCommerceMembership() {
+  private function wooCommerceMembership(): WooCommerceMembership {
     return $this->wooCommerceMembership;
   }
 
-  /**
-   * @return WooCommerceSubscription
-   */
-  private function wooCommerceSubscription() {
+  private function wooCommerceSubscription(): WooCommerceSubscription {
     return $this->wooCommerceSubscription;
   }
 
   /**
    * @param ?string $action
-   * @return WooCommerceCategory|WooCommerceCountry|WooCommerceNumberOfOrders|WooCommerceProduct|WooCommerceTotalSpent
+   * @return Filter
    */
   private function wooCommerce(?string $action) {
     if ($action === WooCommerceProduct::ACTION_PRODUCT) {
--- a/mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
+++ b/mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
@@ -122,7 +122,11 @@ class FilterFactory {
     }
   }
 
-  private function userRole($action) {
+  /**
+   * @param ?string $action
+   * @return MailPoetCustomFields|SubscriberScore|SubscriberSegment|SubscriberSubscribedDate|UserRole
+   */
+  private function userRole(?string $action) {
     if ($action === SubscriberSubscribedDate::TYPE) {
       return $this->subscriberSubscribedDate;
     } elseif ($action === SubscriberScore::TYPE) {
@@ -135,7 +139,11 @@ class FilterFactory {
     return $this->userRole;
   }
 
-  private function email($action) {
+  /**
+   * @param ?string $action
+   * @return EmailAction|EmailActionClickAny|EmailOpensAbsoluteCountAction
+   */
+  private function email(?string $action) {
     $countActions = [EmailOpensAbsoluteCountAction::TYPE, EmailOpensAbsoluteCountAction::MACHINE_TYPE];
     if (in_array($action, $countActions)) {
       return $this->emailOpensAbsoluteCount;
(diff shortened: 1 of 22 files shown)


mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
```diff
@@ -145,15 +153,25 @@ private function email($action) {
     return $this->emailAction;
   }
 
+  /**
+   * @return WooCommerceMembership
+   */
   private function wooCommerceMembership() {
```
websupporter, line 159: In cases where we can use return types in the function defintion, shouldn't we use them?
E.g. `private function wooCommerceMembership() : WooCommerceMembership {}`

mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
```diff
@@ -145,15 +153,25 @@ private function email($action) {
...
     return $this->wooCommerceMembership;
   }
 
+  /**
+   * @return WooCommerceSubscription
+   */
   private function wooCommerceSubscription() {
```
websupporter, line 166: `private function wooCommerceSubscription():WooCommerceSubscription{}`

mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
```diff
@@ -145,15 +153,25 @@ private function email($action) {
...
     return $this->wooCommerceSubscription;
   }
 
-  private function wooCommerce($action) {
+  /**
+   * @param ?string $action
+   * @return WooCommerceCategory|WooCommerceCountry|WooCommerceNumberOfOrders|WooCommerceProduct|WooCommerceTotalSpent
+   */
+  private function wooCommerce(?string $action) {
```
websupporter, line 174: I wonder if it is better to be specific like `WooCommerceCategory|WooCommerceCountry|WooCommerceNumberOfOrders|WooCommerceProduct|WooCommerceTotalSpent` or just use the interface `Filter` as return type. What do you think?

mailpoet/lib/Segments/SegmentsRepository.php
```diff
@@ -48,6 +48,9 @@ protected function getEntityClassName() {
     return SegmentEntity::class;
   }
 
+  /**
+   * @return SegmentEntity|null
+   */
   public function getWPUsersSegment() {
```
websupporter, line 54: `public function getWPUsersSegment() : ?SegmentEntity`

mailpoet/lib/Segments/SegmentsRepository.php
```diff
@@ -172,12 +175,19 @@ public function createOrUpdate(
...
   }
 
-  public function bulkDelete(array $ids, $type = SegmentEntity::TYPE_DEFAULT) {
+  /**
+   * @param array  $ids   Array of segment ids to delete
+   * @param string $type  Type of segment to delete
+   *
+   * @return int Number of segments deleted
+   */
+  public function bulkDelete(array $ids, $type = SegmentEntity::TYPE_DEFAULT): int {
```
websupporter, line 184: I would just add `string $type=SegmentEntity::TYPE_DEFAULT` but all those comments are also just personal favorites right now :smile:

mailpoet/lib/Segments/SegmentsRepository.php
```diff
@@ -199,15 +209,15 @@ public function bulkDelete(array $ids, $type = SegmentEntity::TYPE_DEFAULT) {
...
-      return $entityManager->getConnection()->executeStatement("
-         DELETE s FROM $segmentTable s
-         WHERE s.`id` IN (:ids)
-         AND s.`type` = :type
-      ", [
-        'ids' => $ids,
-        'type' => $type,
-      ], ['ids' => Connection::PARAM_INT_ARRAY]);
+      $queryBuilder = $entityManager->createQueryBuilder();
+      $count = $queryBuilder->delete(SegmentEntity::class, 's')
```
websupporter, line 213: seems to work here, but would it make sense to typecast the return value of `->execute()`? 
image

websupporter:
Thank you @brezocordero for working on this :tada: I have just some remarks. LGTM :bow:

mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
brezocordero: Done!

mailpoet/lib/Segments/DynamicSegments/FilterFactory.php
brezocordero: Agreed and done.

mailpoet/lib/Segments/SegmentsRepository.php
brezocordero: I guess I got carried away 😅 
I have moved the types to the signature when possible and removed the comments. It is more consistent with the rest of the codebase.

mailpoet/lib/Segments/SegmentsRepository.php
brezocordero: It should return an integer with the number of rows affected because it is a delete query. `execute` can return different types depending on the type of query but in this case, I think it is safe not to cast it: https://github.com/doctrine/orm/blob/8f7701279de84411020304f668cc8b49a5afbf5a/lib/Doctrine/ORM/AbstractQuery.php#L1355

brezocordero:
Thanks @websupporter for the review. I have made the changes suggested except for the cast. I don't have a strong opinion on it, if you prefer I do, just let me know.

mailpoet/lib/Segments/SegmentsRepository.php
websupporter: thanks for looking into it!

websupporter (approved):
Thank you! LGTM

veljkho (approved):

PR #4149, opened 2022-05-20
Pregenerate twig templates

websupporter:
Fixes [MAILPOET-3985]

Twig generates templates on the fly, when they do not yet exist or the environment changes. So far, we have stored those files in uploads/mailpoet/cache.

This PR changes this behavior and generates the templates during build time. _If_ we want to go down this path, there are several considerations to make:

Template naming: The names of the template files depend on PHP version ([1](https://github.com/twigphp/Twig/blob/a1f8d3f91a0d28c258e81607e6711387a46f1705/src/Cache/FilesystemCache.php#L34), [2](https://github.com/twigphp/Twig/blob/a1f8d3f91a0d28c258e81607e6711387a46f1705/src/Environment.php#L263). This means if we generate templates in an PHP 8 environment, the templates would not be found, when the plugin runs on PHP 7.2. This PR therefore extends the twig core classes, which are responsible for this behavior, as we can not rely on it consistently. The solution feels "hacky" though.

While generating the templates, we are not in a WordPress context. Some of our [code](https://github.com/mailpoet/mailpoet/blob/3bd49fbc272964c93184a44a8bc5eed27f19ef93/mailpoet/lib/Util/APIPermissionHelper.php#L5-L8) can only be autoloaded in a WordPress environment this is why the robo command needs to define `ABSPATH` and why we need to download WordPress core when [building our environment in CircleCi](https://github.com/mailpoet/mailpoet/commit/bb7b05beaa12b1ac803c47a253962d194838f1b3) (Other suggestions would be very welcome).

For the twig functions, we introduce, we need to make sure, that we do not need to be in a completely loaded context when rendering the templates 5ced3c5e6e2061eae9a91b9e9c8e649af4976113

I still need to find a solution for the disabled test in https://github.com/mailpoet/mailpoet/commit/c9239158964c1ca33b018a119c714b5712fda21d


mailpoet/lib/Config/TwigEnvironment.php
```diff
@@ -0,0 +1,15 @@
...
+
+use MailPoetVendor\Twig\Environment;
+
+class TwigEnvironment extends Environment {
+
+
+  private $templateClassPrefix = '__TwigTemplate_';
+
+  public function getTemplateClass(string $name, int $index = null): string {
+    return $this->templateClassPrefix . \hash('sha256', $name) . (null === $index ? '' : '___' . $index);
```
JanJakes, line 18: I think it could be nice to add a short comment explaining why we're doing this. The commit message explains it nicely but having it in the code may be even better. A single short phrase could be enough.

mailpoet/lib/Config/TwigFileSystemCache.php
```diff
@@ -0,0 +1,21 @@
...
+    string $directory,
+    int $options = 0
+  ) {
+    $this->directory = \rtrim($directory, '\\/') . '/';
+    parent::__construct($directory, $options);
+  }
+
+  public function generateKey(string $name, string $className): string {
+    $hash = \hash('sha256', $className);
+    return $this->directory . $hash[0] . $hash[1] . '/' . $hash . '.php';
```
JanJakes, line 24: Do we need the `$hash[0] . $hash[1]` part? Feels like we don't have too many templates so maybe having all in a single directory could be enough?

We could again add a short code comment about why it's extended.

mailpoet/lib/Config/RendererFactory.php
```diff
@@ -11,9 +11,12 @@ class RendererFactory {
 
   public function getRenderer() {
     if (!$this->renderer) {
-      $caching = WPFunctions::get()->applyFilters('mailpoet_template_cache_enabled', !WP_DEBUG);
       $debugging = WP_DEBUG;
-      $this->renderer = new Renderer($caching, $debugging);
+      $this->renderer = new Renderer(
+        $debugging,
+        Env::$cachePath,
```
JanJakes, line 17: Maybe we can now hardcode this into the renderer as a private const or property? I guess it's no longer needed to be part of the config so we can simplify the parameters.

mailpoet/RoboFile.php
```diff
@@ -1013,6 +1017,22 @@ public function generateUnitOfData($generatorName = null) {
     $generator->run($generatorName);
   }
 
+  public function twigGenerateCache() {
+
+    $templatePath = __DIR__ . '/views/'; // \MailPoet\Config\Env::$viewsPath . '/'
+    define('ABSPATH', (dirname(__DIR__) . '/wordpress'));
```
JanJakes: The necessity of defining `ABSPATH` and loading WP here seems to be only related to [this line](https://github.com/mailpoet/mailpoet/blob/c992bdc2c0965876022da676d4d0a7ecdbac2d72/mailpoet/lib/Config/Renderer.php#L83). Could we make the asset URL evaluation more lazy so it's not needed at the cache build time? It seems it would simplify things.

mailpoet/lib/Twig/Analytics.php
```diff
@@ -9,32 +9,59 @@
...
       new TwigFunction(
         'is_analytics_public_id_new',
-        [$analytics, 'isPublicIdNew'],
+        [$this, 'isPublicIdNew'],
         ['is_safe' => ['all']]
       ),
     ];
   }
+
+  private function analytics() {
```
JanJakes: For a more descriptive name and a better consistency, I'd maybe call this `getAnalytics`.

JanJakes (commented):
LGTM! I see that this is more complex than I imagined but I still think it is worth it doing it.

mailpoet/lib/Config/TwigEnvironment.php
websupporter: [8645213](https://github.com/mailpoet/mailpoet/pull/4149/commits/8645213dcb244e7da72c9ec22851191d56717b03)

mailpoet/lib/Config/TwigFileSystemCache.php
websupporter: added comment in [8645213](https://github.com/mailpoet/mailpoet/pull/4149/commits/8645213dcb244e7da72c9ec22851191d56717b03)

> Do we need the $hash[0] . $hash[1] part?

I wanted to stick as close as possible with the original source here. Yes, I do not really think we need it as we do only have a few files, but as it is not a costly operation and might have solved also another issue, I do not yet see, I preferred to err on the cautious side.

mailpoet/lib/Config/RendererFactory.php
websupporter: `$debugging` is `\WP_DEBUG` and is subject to change. The paths can also vary depending on the system. I am not quite sure, we should get rid of those.

mailpoet/lib/Twig/Analytics.php
websupporter: [0665af1](https://github.com/mailpoet/mailpoet/pull/4149/commits/0665af1dc24c78f7edc471f41941172fdc329abd)

websupporter:
Thanks for your review @JanJakes and especially catching the `ABSPATH` solution! I updated the PR accordingly.

mailpoet/lib/Config/RendererFactory.php
JanJakes: It's now always in `generated`, which is a fixed path, I guess. But it's nothing important...

JanJakes (approved):
Great job!

veljkho (approved):

PR #4151, opened 2022-05-23
Fix issue of getting none-customer subs for # order filter [MAILPOET-4390]

samnajian:
Fixes [MAILPOET-4390]

[MAILPOET-4390]: https://mailpoet.atlassian.net/browse/MAILPOET-4390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Testing instructions inside Jira.

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
@@ -31,8 +31,8 @@ class WooCommerceNumberOfOrders implements Filter {
     global $wpdb;
     $subscribersTable = $this->entityManager->getClassMetadata(SubscriberEntity::class)->getTableName();
     $filterData = $filter->getFilterData();
-    $type = $filterData->getParam('number_of_orders_type');
-    $count = $filterData->getParam('number_of_orders_count');
+    $type = strval($filterData->getParam('number_of_orders_type'));
+    $count = intval($filterData->getParam('number_of_orders_count'));
     $days = $filterData->getParam('number_of_orders_days');
     $parameterSuffix = $filter->getId() ?? Security::generateRandomString();
     $collation = $this->collationChecker->getCollateIfNeeded(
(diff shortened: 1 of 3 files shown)


lysyjan (changes requested):
Please, fix the failing integration and acceptance tests.

mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrders.php
```diff
@@ -44,31 +44,55 @@ public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $fi
...
+      ->leftJoin(
+        'customer',
+        $wpdb->prefix . 'wc_order_stats',
+        'orderStats',
+        'customer.customer_id = orderStats.customer_id AND orderStats.date_created >= :date' . $parameterSuffix . ' AND orderStats.status NOT IN ("wc-cancelled", "wc-failed")'
+      );
+
+    $queryBuilder->add('join', [
+      $subscribersTable => [
+        'joinType' => $this-> shouldIncludeNoneCustomerSubscribers($type, $count) ? 'left' : 'inner',
```
lysyjan, line 61: In my opinion, you don't need to check which join is suitable. The inner join will be a good choice if you move the having clause to the subselect.
But be aware that you need to add a group by clause and select only emails.

I think that the final solution will have better readability.

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
```diff
@@ -142,4 +157,19 @@ private function cleanUp() {
     $this->connection->executeQuery("TRUNCATE TABLE {$wpdb->prefix}wc_customer_lookup");
     $this->connection->executeQuery("TRUNCATE TABLE {$wpdb->prefix}wc_order_stats");
   }
+
+  private function createSubscriber(): SubscriberEntity {
```
lysyjan, line 161: You don't need this logic. We have [a subscriber factory](https://github.com/mailpoet/mailpoet/blob/master/mailpoet/tests/DataFactories/Subscriber.php) that should cover your requirements. If some functionality is missing, don't hesitate to extend it. [Here](https://github.com/mailpoet/mailpoet/blob/master/mailpoet/tests/integration/API/JSON/v1/SendingTaskSubscribersTest.php#L61-L65) is an example of usage.

The factory is mostly used in acceptance tests, but I think that we want to avoid the code repetition as here.

lysyjan (changes requested):
I know that we discussed the query on Friday, but I would want to double-check the performance. So, please, can you verify it again?

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
```diff
@@ -2,15 +2,19 @@
 
 namespace MailPoet\Segments\DynamicSegments\Filters;
 
+use MailPoet\Segments\DynamicSegments\Filters\WooCommerceNumberOfOrders;
```
lysyjan, line 5: I also noticed that this line is unnecessary because the test is in the same namespace.

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceNumberOfOrdersTest.php
```diff
@@ -2,15 +2,19 @@
...
 use MailPoet\Entities\DynamicSegmentFilterEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Subscribers\Source;
+use MailPoet\Subscribers\SubscribersRepository;
 use MailPoetVendor\Carbon\Carbon;
 
 class WooCommerceNumberOfOrdersTest extends \MailPoetTest {
 
+  /* WooCommerceNumberOfOrders */
```
lysyjan, line 17: Another small detail is that the annotation should be `/** @var WooCommerceNumberOfOrders */`.

lysyjan (commented):
LGTM
I would want to ask if you made some performance comparison.

samnajian:
> LGTM I would want to ask if you made some performance comparison.

Good question @lysyjan 
I've run a comparison with a realistically big number of records and here is the result:
```
Correct query when including none-customer subscribers (i.e. leftJoin is used)
First run: 334 ms
Consecutive runs: ~234 ms

Wrong query:
First run: 191 ms
Consecutive runs: ~180 ms
```

lysyjan (approved):

rohankacharegs:
GlobalStep Verified

veljkho (approved):

PR #4153, opened 2022-05-23
Respect mailpoet_display_custom_fonts filter in Iframe MAILPOET-4247

websupporter:
Fixes [MAILPOET-4247]

[MAILPOET-4247]: https://mailpoet.atlassian.net/browse/MAILPOET-4247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Menu.php
+++ b/mailpoet/lib/Config/Menu.php
@@ -98,7 +98,7 @@ class Menu {
 
     if (
       !isset($_REQUEST['page'])
-      && sanitize_text_field(wp_unslash($_REQUEST['page'])) !== 'mailpoet-newsletter-editor'
+      || sanitize_text_field(wp_unslash($_REQUEST['page'])) !== 'mailpoet-newsletter-editor'
     ) {
       return;
     }
--- a/mailpoet/lib/Config/Menu.php
+++ b/mailpoet/lib/Config/Menu.php
@@ -22,6 +22,7 @@ use MailPoet\AdminPages\Pages\WelcomeWizard;
 use MailPoet\AdminPages\Pages\WooCommerceSetup;
 use MailPoet\DI\ContainerWrapper;
 use MailPoet\Features\FeaturesController;
+use MailPoet\Form\Util\CustomFonts;
 use MailPoet\Util\License\License;
 use MailPoet\WP\Functions as WPFunctions;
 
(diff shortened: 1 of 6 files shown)


triple0t (commented):
Well done @websupporter 
Just a note.

a. If I disable Google fonts with the filter
```php
add_filter('mailpoet_display_custom_fonts', function () {return false;});
```
The font assets won't be loaded ☑️

b. If I disable loading 3rd-party libraries (via settings > advanced) 
Screenshot 2022-05-24 at 6 28 52 PM
without using the filters

The font assets still load.  ⤫
I believe it should not be loading based on Spec item 3.

websupporter:
Hi @triple0t,
thanks a lot for reading the ticket thoroughly :see_no_evil: . Don't know, how I missed that. I updated the current PR accordingly.

mailpoet/lib/Config/Menu.php
```diff
@@ -82,36 +88,46 @@ public function setup() {
...
+    if (!self::isOnMailPoetAdminPage()) {
+      return;
+    }
+    $this->wp->doAction('mailpoet_conflict_resolver_styles');
+    $this->wp->doAction('mailpoet_conflict_resolver_scripts');
+
+    if (
+      !isset($_REQUEST['page'])
+      && sanitize_text_field(wp_unslash($_REQUEST['page'])) !== 'mailpoet-newsletter-editor'
+    ) {
```
triple0t: Hmm, is this not supposed to be an **OR**?

```PHP
!isset($_REQUEST['page'])
      || sanitize_text_field(wp_unslash($_REQUEST['page'])) !== 'mailpoet-newsletter-editor'
```

https://3v4l.org/1nPTX

Immediately the first part of the expression returns **false**, the second part is not evaluated. Hence, the code continues execution when it should not be doing so.

In this case:
`!isset($_REQUEST['page'])` does not have to equal  **'mailpoet-newsletter-editor'**

triple0t (commented):
Looks good.
Kindly find a note below.

Also, It would be nice if we could show a notice on the Email editor and Form editor page when 3rd party library or Google fonts are disabled

triple0t (approved):

veljkho (approved):

PR #4157, opened 2022-05-25
Bootstrap automation engine in premium plugin [PREMIUM-180]

alex-mailpoet:
[PREMIUM-180]

[PREMIUM-180]: https://mailpoet.atlassian.net/browse/PREMIUM-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Engine/Workflows/Workflow.php
```diff
@@ -30,14 +30,19 @@ class Workflow {
   /** @param Step[] $steps */
   public function __construct(
     string $name,
-    array $steps
+    array $steps,
+    int $id = null
```
JanJakes, line 35: Did you end up using it somewhere? Maybe I'll see it in the following PRs. I also "almost needed" it but then didn't add it to avoid these simple value objects becoming an ORM. I guess it's OK to add IDs and we may end up doing so anyway, just asking for the reason (that I may discover myself when moving on with the reviews).

JanJakes (commented):
LGTM! Just a tiny question.

mailpoet/lib/Automation/Engine/Workflows/Workflow.php
alex-mailpoet: These ID became necessary when I wrote a test where I created new `Workflow` and `WorkflowRun` objects. Their proper work depends on IDs being present, but before this change IDs could only be set in the JSON payload via API. Here's the test:
https://github.com/mailpoet/mailpoet-premium/pull/580/commits/cefba1c4ceaccf7c3a70acd97ebdfa2f47fab4f3#diff-ef262c33900febab317f387f447c59284156bf20fc126f69113e727447166330R76-R78

veljkho (approved):

PR #4159, opened 2022-05-25
Introducing Action Scheduler Cron Runner [MAILPOET-4274]

costasovo:
This PR introduces a new method for MailPoet's background job processing that uses the Action Scheduler in the background.

This is intended as a first step in migration towards using the Action scheduler. You can read more details about the proposed process in the related P2 post (pcNwfB-NX-p2).

The idea is that the daemon runner is wrapped in a single recurring action scheduler's action. 

### How it works
There is a trigger action (`mailpoet/cron/daemon-trigger`) that in every Action Scheduler run checks whether there are jobs to process. It uses the very same method as the current trigger, which does the same in every request.

In case the trigger action detects there is a job to do it schedules a daemon run action (`mailpoet/cron/daemon-run`) and spawns a new background action scheduler runner. The deamon run action runs for 10 seconds and after the run is completed it either schedules itself again or if there is no more work to do unschedules itself. rescheduling is done automatically by the Action Scheduler

The whole logic is implemented in `MailPoet\Cron\DaemonActionSchedulerRunner`.

I tried to comment on every commit. I recommend reviewing the change commit by commit.

### Test fixes
This PR contains a bunch of fixes in tests. When working on this task I needed to enable WP-Cron for the Acceptance Tests. This caused a bunch of flaky tests. Hopefully, I fixed all of them.

### Changes in UI

The method will be selected as the default option for new installs in MailPoet > Settings > Advanced
Screenshot 2022-05-25 at 14 51 14

There is additional status info on the MailPoet > Help > System Tab
Screenshot 2022-07-21 at 13 20 57

## Testing instructions
It would be great to test the overall functionality of the plugin with a focus on background jobs. Ideally, test on more hosts.

You need to go to Settings > Advanced and switch the task scheduler `WordPress built-in cron (recommended)`

- Does sending to a large number (over 10K) of subscribers work?
- You can test inactive subscribers' detection.
- Try scheduling some scheduled tasks to the past and observe they were executed.

### Testing build

I prepared a special build that, instead of real sending (through MSS, web host, and SMTP), only sleeps for some time and pretend the email was sent. The build can be used for testing how sending a large number of emails works without actually sending emails out.

Download >>> [mailpoet_mocked_sending.zip](https://github.com/mailpoet/mailpoet/files/9242315/mailpoet_mocked_sending.zip)

To be sure you have this build, you can check the plugin name in Plugins. It should say: `MailPoet 3 (New - Mocked Sending)`

[MAILPOET-4274]

[MAILPOET-4274]: https://mailpoet.atlassian.net/browse/MAILPOET-4274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Cron/DaemonActionSchedulerRunner.php
```diff
@@ -0,0 +1,124 @@
...
+    $this->wp->addAction('action_scheduler_after_process_queue', [$this, 'afterProcess']);
+    if (!as_has_scheduled_action(self::DAEMON_TRIGGER_SCHEDULER_ACTION)) {
+      as_schedule_recurring_action($this->wp->currentTime('timestamp'), 20, self::DAEMON_TRIGGER_SCHEDULER_ACTION);
+    }
+  }
+
+  /**
+   * In regular intervals checks if there are scheduled tasks to execute.
+   * In case there are tasks it spawns a recurring action.
+   * @return void
```
JanJakes: Can we use PHP return type instead? Also, there are more missing return types in this file.

mailpoet/lib/Cron/ActionScheduler/ActionScheduler.php
```diff
@@ -0,0 +1,19 @@
...
+namespace MailPoet\Cron\ActionScheduler;
+
+class ActionScheduler {
+  private const GROUP_ID = 'mailpoet-cron';
+
+  public function scheduleRecurringAction(int $timestamp, int $interval_in_seconds, $hook, $args = []): int {
+    return as_schedule_recurring_action($timestamp, $interval_in_seconds, $hook, $args, self::GROUP_ID);
+  }
+
+  public function unscheduleAction($hook, $args = []) {
```
JanJakes: A return type is missing. Also, `$args` should probably be an array as in the method below. Same goes for the method above, in fact.

mailpoet/lib/Cron/DaemonActionSchedulerRunner.php
```diff
@@ -0,0 +1,124 @@
...
+
+namespace MailPoet\Cron;
+
+use MailPoet\Cron\Triggers\WordPress;
+use MailPoet\WP\Functions as WPFunctions;
+
+class DaemonActionSchedulerRunner {
+  const DAEMON_RUN_SCHEDULER_ACTION = 'mailpoet/cron/daemon-run';
+  const DAEMON_TRIGGER_SCHEDULER_ACTION = 'mailpoet/cron/daemon-trigger';
+  const RUN_ACTION_SCHEDULER = 'mailpoet-cron-action-scheduler-run';
```
JanJakes: Could we improve the constant naming here? It seems like the first two are "daemon run" and "daemon trigger" actions so the "scheduler" seems to be quite extra there. Maybe `DAEMON_RUN_ACTION` and `DAEMON_TRIGGER_ACTION` are more understandable.

For the third constant, it's not quite clear what it is just by the name. It seems to be an "action" but a different one than the WP hook/action scheduler action. I actually think it would be better to extract this to a different class. What I mean is these things could go together:

```php
const RUN_ACTION = 'mailpoet-cron-action-scheduler-run';

public function init() {
  $this->wp->addAction('wp_ajax_nopriv_' . self::RUN_ACTION_SCHEDULER, [$this, 'runActionScheduler'], 0);
}

public function runActionScheduler() {
  // ...
}

public function ensureConcurrency(int $concurrency): int {
  // ...
}
```

It would also solve the issue that this class seems to do too many things at once.

Not sure how to name the new class... Seems like you call this by `triggerRemoteExecutor` so maybe something like `RemoteExecutorHandler`? I guess in that case we'd need to place both under a sub-namespace, such as:

```bash
mailpoet/lib/Cron/ActionScheduler/DaemonRunner.php
mailpoet/lib/Cron/ActionScheduler/RemoteExecutorHandler.php
```

Or something like that... maybe it can be split even more, not sure. For instance — shouldn't a trigger be a different service? Seems like this one is named "runner" but does sort of everything. When it has a single public method, then it's perfect 😄

mailpoet/lib/Cron/DaemonActionSchedulerRunner.php
```diff
@@ -0,0 +1,124 @@
...
+    if ($this->wordpressTrigger->checkExecutionRequirements()) {
+      sleep(2); // Add short sleep to ensure next action ready to be processed since minimal schedule interval is 1 second
+      $this->triggerRemoteExecutor();
+    } else {
+      as_unschedule_action(self::DAEMON_RUN_SCHEDULER_ACTION);
+    }
+  }
+
+  /**
+   * Spawns addition Action Scheduler runner
```
JanJakes: I noticed many of the comments in this file have typos and mistakes and can be hard to understand. Just quickly:

>  In regular intervals checks if there are scheduled tasks to execute.
>  In case there are tasks it spawns a recurring action.

"Spawn" doesn't seem to be a great word choice here. It seems to be more like "scheduling it". Thinking further, I think that method doesn't need a comment. The `as_schedule_recurring_action` line has a comment explaining it.

> Run daemon that processes scheduled tasks for limited time (default 20 seconds)

That default was changed later and I think that comment is unnecessary as well, especially if we split it to trigger and runner classes.

> This method is triggered by an ajax request.
> It creates ActionScheduler runner and process a batch actions that are ready to be processed

Action scheduler runner is also the class we're in, what you mean is a queue runner. There's also a typo `process` -> `processes`.

> After Action Scheduler finishes queue always check there is more work to do and in case there is trigger additional runner.

An IF is missing in the phrase and some commas too.

> // Add short sleep to ensure next action ready to be processed since minimal schedule interval is 1 second

An "is" is missing and some commas too.

I actually recommend carefully reading all the comments, removing and shortening as much as possible, and fixing the typos and understandability problems 🙏

mailpoet/lib/Cron/DaemonActionSchedulerRunner.php
```diff
@@ -0,0 +1,124 @@
...
+   * @see https://actionscheduler.org/perf/#increasing-initialisation-rate-of-runners
+   * @return void
+   */
+  private function triggerRemoteExecutor() {
+    $this->wp->addFilter('https_local_ssl_verify', '__return_false', 100);
+    $this->wp->wpRemotePost( $this->wp->adminUrl( 'admin-ajax.php' ), [
```
JanJakes: There are wrong spacings in the brackets. And I've seen it [a bit above](https://github.com/mailpoet/mailpoet/pull/4159/commits/de6c7f9c22b97d34ea5c5a2591f556467c0b3dd5#diff-cafd1ab4b6e1b51d2b7f837d1e31f2dd193968a895512ee38eb6449b9b36290fR79) too. Seems like we don't have sniffs for that 😬

mailpoet/views/help.html
```diff
@@ -73,6 +74,12 @@
   'updatedAt': __('Updated At'),
   'nothingToShow': __('Nothing to show.'),
   'preview': _x('Preview', 'Text of a link to email preview page.'),
+  'actionSchedulerStatus': __('WP Cron / Action Scheduler Status'),
+  'version': __('Version'),
+  'storage': __('Storage type'),
+  'latestActionSchedulerTrigger': _x('Next scheduler trigger run', 'Label for the next date and time of background job trigger run.'),
+  'latestActionSchedulerCompletedTrigger': _x('Last scheduler trigger run', 'Label for the last date and time of background job trigger run.'),
+  'latestActionSchedulerCompletedRun': _x('Last scheduler worker run', 'Label for the last date and time of background job worker run.'),
```
JanJakes: Coule we drop the "scheduler" part? It's confusing me a bit — "Next scheduler trigger run" is like "is it a scheduler or trigger or scheduler-trigger or is the trigger noun or a verb", etc. I think "Next trigger run" and "Next worker run" would be more understandable.

mailpoet/lib/API/JSON/v1/SendingQueue.php
```diff
@@ -232,4 +247,17 @@ public function resume($data = []) {
...
+   * This is done so that user immediately sees that email is sending and doesn't have to wait on WP Cron to start it.
+   */
+  private function triggerSending(NewsletterEntity $newsletter): void {
+    if (
+      $newsletter->getStatus() === NewsletterEntity::STATUS_SENDING
+      && $this->settings->get('cron_trigger.method') === CronTrigger::METHOD_ACTION_SCHEDULER
+    ) {
+      $this->actionSchedulerRunner->trigger();
+    }
+  }
```
JanJakes, line 262: Interesting. So here we need our trigger to be really executed, and not just Action Scheduler, right?

Because for Action Scheduler, [every admin request triggers the check by default](https://actionscheduler.org/faq/):
> By default, Action Scheduler is initiated by WP-Cron (and the 'shutdown' hook on admin requests).

mailpoet/tests/_support/DefaultsExtension.php
```diff
@@ -47,6 +47,9 @@ private function setupWordPress() {
     global $wp_rewrite; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
     $wp_rewrite->init(); // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
     $wp_rewrite->flush_rules(); // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+
+    // Make sure WP cron is not locked so that tests that rely on it don't timeout
+    delete_transient('doing_cron');
```
JanJakes, line 52: Does this somehow relate to the DB backup/restore logic?
I mean, if the cron is locked, then we would also backup the locked state and every test would start in that state, is that right?

Looks like a great catch! 👏

mailpoet/tests/_support/AcceptanceTester.php
```diff
@@ -677,4 +677,45 @@ public function canTestWithPlugin(string $pluginSlug): bool {
...
+      "DELETE FROM mp_options WHERE option_name = '_transient_doing_cron';", // Remove WP Cron lock
+    ]);
+
+    // Schedule MailPoet trigger in action scheduler to run immediately
+    $i->importSql([
+      "UPDATE mp_actionscheduler_actions SET scheduled_date_gmt = SUBTIME(now(), '01:00:00'), scheduled_date_local = SUBTIME(now(), '01:00:00') WHERE hook = 'mailpoet/cron/daemon-trigger' AND status = 'pending';",
+    ]);
+
+    //Page reload triggers WP Cron and it triggers Action Scheduler
+    $i->reloadPage();
```
JanJakes: I'm wondering if this needs to be this complex. Changing the Action Scheduler registration in WP cron doesn't seem like the best thing to do and think [a simple reload in admin should work as well](https://actionscheduler.org/faq/):

> By default, Action Scheduler is initiated by WP-Cron (and the 'shutdown' hook on admin requests).

So maybe it all could be reduced to requesting/visiting/reloading an admin page?

And what do you think about the WP CLI approach and having WP cron and Action Scheduler disabled?

mailpoet/lib/Config/Initializer.php
```diff
@@ -159,6 +159,9 @@ public function __construct(
   }
 
   public function init() {
+    // Initialize Action Scheduler. Needs to be called early otherwise some tools doesn't work properly
```
JanJakes: > Needs to be called early otherwise some tools doesn't work properly

Could you explain this a bit more? Most people won't know what tools and what to look for when touching this and you may forget it too.

mailpoet/tests/integration/_bootstrap.php
```diff
@@ -115,11 +116,14 @@ public function setUp(): void {
     $this->diContainer = ContainerWrapper::getInstance(WP_DEBUG);
     $this->connection = $this->diContainer->get(Connection::class);
     $this->entityManager = $this->diContainer->get(EntityManager::class);
+    // Cleanup scheduled tasks from previous tests and deactivate cron
```
JanJakes: Does this "deactivate" the cron or just switches the method?

mailpoet/lib/Cron/CronTrigger.php
```diff
@@ -42,10 +42,14 @@ public function __construct(
   public function init() {
     $currentMethod = $this->settings->get(self::SETTING_NAME . '.method');
     try {
+      if ($currentMethod === self::METHOD_ACTION_SCHEDULER) {
+        $this->cronActionScheduler->init();
+        return true;
+      } else {
+        $this->cronActionScheduler->deactivate();
```
JanJakes: We could also alternatively do the deactivation in the job run (like "IF current method is not AS, deactivate"). Would that be better/worse or similar?

mailpoet/lib/Cron/DaemonActionSchedulerRunner.php
```diff
@@ -28,6 +28,13 @@ class DaemonActionSchedulerRunner {
   /** @var ActionScheduler */
   private $actionScheduler;
 
+  /**
+   * The inital value is set based on default cron execution limit battle tested in MailPoet custom cron runner (an older version of the background processing).
+   * The default limit in PHP is 30s so it leaves 10 execution margin.
```
JanJakes: I find this long explanation quite unnecessary. I think saying "default is 20 seconds" is enough.

mailpoet/lib/Cron/DaemonActionSchedulerRunner.php
```diff
@@ -123,6 +130,17 @@ public function afterProcess() {
...
   }
 
+  /**
+   * This method is hooked into action_scheduler_maximum_execution_time_likely_to_be_exceeded
+   * and used to listen on how many execution time is needed.
+   * The execution limit is then used for the daemon run
+   */
+  public function storeRemainingExecutionLimit($likelyExceeded, $runner, $processedActions, $executionTime, $maxExecutionTime) {
+    $newLimit = floor(($maxExecutionTime - $executionTime) - self::EXECUTION_LIMIT_MARGIN);
+    $this->remainingExecutionLimit = intval(max($newLimit, 0));
```
JanJakes: If the input numbers are floats then maybe `remainingExecutionLimit` could also be float? That would avoid `floor` and `intval` calls.

JanJakes (commented):
Great job and amazing test fixes! 👏
I'm really happy you took the time to solve this.

I commented mostly on small issues. Overall I'd mainly suggest splitting the `DaemonActionSchedulerRunner` into multiple classes and checking the comments for clarity.

mailpoet/lib/Cron/DaemonActionSchedulerRunner.php
costasovo: Thanks for pointing that out! I think I'll split the class by actions Trigger and Run (+ maybe the remote executor runner). That way, I may also get better names for constants.

mailpoet/lib/API/JSON/v1/SendingQueue.php
costasovo: I added this based on my observations. After seeing your comment and researching a bit, I found that it is true that it is triggered by admin requests, but it happens [only once per minute](https://github.com/woocommerce/action-scheduler/blob/2d58fd29e9a17725009d78a029b0550112437a26/docs/index.md#how-it-works) ([related code](https://github.com/woocommerce/action-scheduler/blob/c2d775356f392150e667915c8a3d4736e9b99605/classes/ActionScheduler_QueueRunner.php#L102-L108)).
So without this trigger, it takes up to one minute to trigger the sending. I think it is important for the user experience that sending starts (or also potentially fails) immediately after being triggered.

mailpoet/tests/_support/DefaultsExtension.php
costasovo: Yes, it is related to DB backup/restore. But it is also a bit flaky, and took me some time before I noticed that this is happening 😄

mailpoet/tests/_support/AcceptanceTester.php
costasovo: Unfortunately, the admin page trigger works [also once per minute](https://github.com/woocommerce/action-scheduler/blob/2d58fd29e9a17725009d78a029b0550112437a26/docs/index.md#how-it-works).
But I guess I can try removing the lock for the action scheduler's admin page load trigger instead of rescheduling the WP Cron task. Removing the lock should be much simpler.

> And what do you think about the WP CLI approach and having WP cron and Action Scheduler disabled?

I think I tried that and ran into some issues, but I can't remember. Also, the triggering by wp-cron or the admin request is closer to what really happens.

mailpoet/lib/Cron/CronTrigger.php
costasovo: Sounds like a good idea. Also, the code would be at the same place 👍

costasovo: Moving it to the action itself would require always registering the action even if the method is not active. On the other hand running deactivate on every requests seems to trigger an SQL query. So I'm going to add `deactivateOnTrigger` method that will register the trigger action but with different callback, which only unschedules.

mailpoet/tests/_support/AcceptanceTester.php
costasovo: Ok. So I tried with the admin page trigger. It works, but in some tests (e.g. Woo purchase emails) we don't visit the admin. So I revisited the WP CLI approach. I found why it was not working, fixed that and now the code is much simpler.

costasovo:
@JanJakes I believe I've addressed all issues. I did all in new commits.

mailpoet/lib/Cron/DaemonActionSchedulerRunner.php
```diff
@@ -42,4 +48,17 @@ public function deactivate(): void {
...
+   * Note: We can't unschedule the actions directly inside the trigger action process, because the action is recurring and would schedule itself.
+   * We need to do it after the action scheduler process.
+   */
+  public function deactivateOnTrigger(): void {
+    $this->wp->addAction(DaemonTrigger::NAME, [$this, 'deactivateAfterProcess']);
+  }
+
+  public function deactivateAfterProcess(): void {
+    $this->wp->addAction('action_scheduler_after_process_queue', [$this, 'deactivate']);
+  }
```
JanJakes, line 68: Ahh, WP hooks are so hard and unintuitive to read 😞 But what can we do...
More to the point — do you think we should do a check of `$this->settings->get(self::SETTING_NAME . '.method')` for every `DaemonTrigger::NAME` in `action_scheduler_after_process_queue` and deactivate if needed? Or do you consider this deactivation "from outside" better?

JanJakes (approved):
Looks great! I added just a single question, nothing important.

mailpoet/lib/Cron/CronTrigger.php
JanJakes: > Moving it to the action itself would require always registering the action even if the method is not active.

Related to my [question](https://github.com/mailpoet/mailpoet/pull/4159#discussion_r930072008) — why would it always require registering the action? It seems to me that:
1. If `DaemonTrigger::NAME` is registered in Action Scheduler, and therefore running, we would detect and deactivate all scheduled Action Scheduler jobs in its `action_scheduler_after_process_queue`.
2. `DaemonTrigger::NAME` is **not** registered in Action Scheduler, then it has been already unregistered and we don't need to try to unregister it at all. Or is that not right? 🤔

costasovo: There are two things you need to register.
1) action in action scheduler
2) WordPress callback for the action
Currently both happens in the `$this->cronActionScheduler->init();` I used wrong wording by "registering the action" I meant the WP action (aka hook), not the scheduler action. Sorry for the confusion.

So for the case, the action is registered in the action scheduler, but the method is not active anymore in plugin settings I still need to register some WP action that will handle the deactivation. I can move the whole logic to `$this->cronActionScheduler->init();` and call it on every request, or the other place I found reasonable is the cron trigger as it handles logic around activating different methods.
In the next steps, we will remove the visitor method. So with that, we could get rid of the whole cron trigger and use only `$cronActionScheduler->init();`.
Does that make sense?

JanJakes: @costasovo Ahh, I get it — the scheduled action can't unregister itself because the code to handle that scheduled action is no longer registered. Makes sense.

JanJakes:
@costasovo GitHub reports a conflicting file. Is that expected or should I wait for a rebase before approving?

costasovo:
@JanJakes Yes, please wait for rebase! I rebased it yesterday and missed that it needs it again. Sorry for the inconvenience!

costasovo:
@JanJakes Now it is ready to review. I added 2 commits:
1) Refactoring of deactivation as we discussed over slack
2) Small fix in settings (a label had a wrong association to a radio button)

veljkho (approved):

PR #4164, opened 2022-05-27
Stop integration test from sending real email via AmazonSES MAILPOET-4394

samnajian:
Currently .env variables are not picked up for integration tests, this PR enables this.

[MAILPOET-4394]

[MAILPOET-4394]: https://mailpoet.atlassian.net/browse/MAILPOET-4394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/integration/Mailer/Methods/AmazonSESTest.php
+++ b/mailpoet/tests/integration/Mailer/Methods/AmazonSESTest.php
@@ -7,6 +7,7 @@ use MailPoet\Mailer\MailerError;
 use MailPoet\Mailer\Methods\AmazonSES;
 use MailPoet\Mailer\Methods\Common\BlacklistCheck;
 use MailPoet\Mailer\Methods\ErrorMappers\AmazonSESMapper;
+use MailPoet\WP\Functions as WPFunctions;
 
 class AmazonSESTest extends \MailPoetTest {
   public $extraParams;
@@ -51,7 +52,8 @@ class AmazonSESTest extends \MailPoetTest {
       $this->sender,
       $this->replyTo,
       $this->returnPath,
-      new AmazonSESMapper()
+      new AmazonSESMapper(),
+      new WPFunctions()
     );
     $this->subscriber = 'Recipient <blackhole@mailpoet.com>';
     $this->newsletter = [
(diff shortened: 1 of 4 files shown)


lysyjan (changes requested):
This pull request doesn't solve the issue. We need to continue with the discussion on Slack and decide how to deal with Amazon sending in integration tests.
Also, we should restore the removed sending test.

samnajian:
> This pull request doesn't solve the issue. We need to continue with the discussion on Slack and decide how to deal with Amazon sending in integration tests. Also, we should restore the removed sending test.

@lysyjan thanks for taking a look. As explained in the PR description this PR only enables the picking up the .env variables which is what's needed to be able to reproduce the issue documented in MAILPOET-4394 locally.

lysyjan:
@samnajian In my opinion, this pull request should also solve the reactivation and repair of the test. If not, I would want to know the next steps because you are assigned to the ticket in the Jira. When we move this pull request further, the test may stay inactive because the discussion in the slack has been almost two weeks without further response.

mailpoet/tests/integration/Mailer/Methods/AmazonSESTest.php
```diff
@@ -255,4 +258,33 @@ public function testItChecksBlacklistBeforeSending() {
     expect($result['error'])->isInstanceOf(MailerError::class);
     expect($result['error']->getMessage())->stringContainsString('AmazonSES has returned an unknown error.');
   }
+
+  public function testItCanSend() {
+    if (getenv('WP_TEST_MAILER_ENABLE_SENDING') !== 'true') $this->markTestSkipped();
```
lysyjan: I think that we don't need to skip this test anymore when the sending is mocked. What do you think?

lysyjan (commented):
It looks good. I have a question about skipping only.

mailpoet/tests/integration/Mailer/Methods/AmazonSESTest.php
samnajian: @lysyjan umm, yea makes sense, updating.

PR #4171, opened 2022-05-31
Add label to text input fields on form

brezocordero:
This PR adds `<label>` to text input fields in the forms for email, last name, and first name, in the front end.

If the text form has a placeholder, the label is not visible using the class `.screen-reader-text`
The forms also have a unique ID to use on the `for` attribute on the label.

To test, create a subscription form that is displayed on the front end.
Inspect the form and check that:
- The text inputs have a unique id
- Each text input has a label with a `for` attribute using the input id
- Labels on text input fields with placeholders use the class `.screen-reader-text`

[MAILPOET-4312]

[MAILPOET-4312]: https://mailpoet.atlassian.net/browse/MAILPOET-4312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Form/Block/BlockRendererHelper.php
+++ b/mailpoet/lib/Form/Block/BlockRendererHelper.php
@@ -104,28 +104,39 @@ class BlockRendererHelper {
 
   public function renderLabel(array $block, array $formSettings): string {
     $html = '';
+    $forId = '';
+
     if (
       isset($block['params']['hide_label'])
       && $block['params']['hide_label']
     ) {
       return $html;
     }
-    if (
-      isset($block['params']['label_within'])
-      && $block['params']['label_within']
-    ) {
-      return $html;
-    }
+
     $automationId = null;
     if (in_array($block['id'], ['email', 'last_name', 'first_name'], true)) {
       $automationId = 'data-automation-id="form_' . $block['id'] . '_label" ';
+      if (isset($formSettings['id'])) {
+        $forId = 'for="form_' . $block['id'] . '_' . $formSettings['id'] . '" ';
+      }
     }
+
     if (
       isset($block['params']['label'])
       && strlen(trim($block['params']['label'])) > 0
     ) {
+      $labelClass = 'class="mailpoet_' . $block['type'] . '_label" ';
+
+      if (
+        isset($block['params']['label_within'])
+        && $block['params']['label_within']
+      ) {
+        $labelClass = 'class="screen-reader-text" ';
+      }
+
       $html .= '<label '
-        . 'class="mailpoet_' . $block['type'] . '_label" '
+        . $forId
+        . $labelClass
         . $this->renderFontStyle($formSettings, $block['styles'] ?? [])
         . ($automationId ?? '')
         . '>';
--- a/mailpoet/lib/Form/Block/BlockRendererHelper.php
+++ b/mailpoet/lib/Form/Block/BlockRendererHelper.php
@@ -116,9 +116,10 @@ class BlockRendererHelper {
     $automationId = null;
     if (in_array($block['id'], ['email', 'last_name', 'first_name'], true)) {
       $automationId = 'data-automation-id="form_' . $block['id'] . '_label" ';
-      if (isset($formSettings['id'])) {
-        $forId = 'for="form_' . $block['id'] . '_' . $formSettings['id'] . '" ';
-      }
+    }
+
+    if (isset($formSettings['id'])) {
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Form/Block/Text.php
```diff
@@ -34,27 +34,34 @@ public function __construct(
...
     }
 
     if (in_array($block['id'], ['email', 'last_name', 'first_name'], true)) {
+
       $automationId = 'data-automation-id="form_' . $this->wp->escAttr($block['id']) . '" ';
+
+      if (isset($formSettings['id'])) {
+        $id = 'id="form_' . $this->wp->escAttr($block['id']) . '_' . $this->wp->escAttr($formSettings['id']) . '" ';
+      }
     }
```
triple0t: I'm curious why we are restricting **id** and **htmlFor** to only this array ['email', 'last_name', 'first_name'].
What about custom field text input types?

mailpoet/lib/Form/Block/BlockRendererHelper.php
```diff
@@ -104,28 +104,39 @@ public function getInputValidation(array $block, array $extraRules = [], ?int $f
...
-      return $html;
-    }
+
     $automationId = null;
     if (in_array($block['id'], ['email', 'last_name', 'first_name'], true)) {
       $automationId = 'data-automation-id="form_' . $block['id'] . '_label" ';
+      if (isset($formSettings['id'])) {
+        $forId = 'for="form_' . $block['id'] . '_' . $formSettings['id'] . '" ';
+      }
     }
```
triple0t: The same item here.

triple0t (commented):
Looks good @brezocordero  

Just a note.

Kindly find below

mailpoet/lib/Form/Block/Text.php
brezocordero: @triple0t Good question!
I am not sure of why the restriction is there. It is coming from this [commit](https://github.com/mailpoet/mailpoet/pull/2634/commits/bbfd137a3c2e514df66c774a815981dfbdf2d69b). 
There is another issue in the epic that targets form controls more widely, I have added a note about custom text fields there and specified better the scope of this issue.

mailpoet/lib/Form/Block/BlockRendererHelper.php
brezocordero: Answered above.

mailpoet/lib/Form/Block/Text.php
triple0t: Hi @brezocordero 
I mean this was added specifically in this commit https://github.com/mailpoet/mailpoet/pull/4171/commits/09ac83cf8efda93997a9ae16f44daf337aa8cdf5#diff-53657c1788edfbe586720f9a5024fd9578cac10853dd2c4602daea4dd2bcdb7aR48-R50
and I was wondering why you scoped the ID to work only under the array 
```
 if (isset($formSettings['id'])) {
        $id = 'id="form_' . $this->wp->escAttr($block['id']) . '_' . $this->wp->escAttr($formSettings['id']) . '" ';
  }
```

I can see [another ticket](https://mailpoet.atlassian.net/browse/MAILPOET-3729) in the backlog addressing this issue so I guess it's fine.

BTW, I can see the same errors report when I test with https://achecker.achecks.ca/checker/index.php (Mostly for forms with custom fields)
Screenshot 2022-06-03 at 9 49 19 PM
Screenshot 2022-06-03 at 9 49 27 PM

I will add these notes to the ticket and move this PR to QA.

Thank you

brezocordero: @triple0t I made the changes and updated the other ticket. Your point made more sense.

triple0t:
I discovered an issue with the class `.screen-reader-text`
The CSS class only exists within the WordPress environment and due to that, iFrame form labels were showing when they should not be visible.
I fixed this in https://github.com/mailpoet/mailpoet/pull/4171/commits/fc3345a50f88ffcb9b618efdef663934d0049eea

Screenshot 2022-06-06 at 9 40 34 AM

triple0t (approved):

mailpoet/lib/Form/Block/Text.php
triple0t: @brezocordero Thank you for working on it 👏🏾

veljkho (approved):

rohankacharegs:
GlobalStep Verified

PR #4172, opened 2022-06-01
do not override status when wp user list is deleted

cherbst:
this lead to confirmation email being sent even though
the user did not choose to be subscribed to any newsletter


costasovo:
@cherbst  Thanks for your contribution! If I understand correctly the issue is that the confirmation email was sent to a newly added user even though the "WordPress Users" list was disabled (in the trash). Correct? 

I think we can't fix the issue this way, because it may affect other subscription flows.

I tried to replicate the issue by disabling the list and adding users via admin, but it worked as expected and no confirmation email was sent. 

Could you please provide replication steps so that we can replicate the problem? Do you use any additional plugins for user management? Do you use WooCommerce?

JanJakes:
Closing as we have no additional information.

PR #4175, opened 2022-06-02
Create a workflow from a PHP template [MAILPOET-4264]

johnolek:
This PR adds the initial code required for creating a workflow from a template defined in PHP.

From a user's perspective, the only change is the addition of a new testing button on the automations page:

image

The created workflow will include 3 steps:
1. A segment signup trigger
2. A delay of 1 hour
3. Sending a welcome email

The created workflows will have a new status: `draft`. This is because we are not collecting any required parameters such as the specific email to send.

Note that this PR does *not* change anything about the sendWelcomeEmailAction. I'm assuming that we will address the changes required for that action in a separate ticket.

MAILPOET-4264


mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsCreateFromTemplateEndpoint.php
```diff
@@ -0,0 +1,33 @@
...
+  public function handle(Request $request): Response {
+    $data = $request->getParams();
+    $this->createWorkflowFromTemplateController->createWorkflow($data);
+    return new Response();
+  }
+
+  public static function getRequestSchema(): array {
+    return [
+      'name' => Builder::string()->required(),
+      'template' => Builder::string()->required(),
```
johnolek, line 30: Is there a way to require one of a handful of specific values without using a regular expression (i.e. `pattern()`)?

mailpoet/lib/Automation/Engine/Workflows/Workflow.php
```diff
@@ -55,6 +56,10 @@ public function getStatus(): string {
     return $this->status;
   }
 
+  public function setDraft(): void {
+    $this->status = self::STATUS_DRAFT;
+  }
```
johnolek: There's probably a better way of handling this than adding a public method like this. Maybe there should be a constructor argument for the status instead? Not sure what feels right.

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace MailPoet\Automation\Integrations\MailPoet\Templates;
+
+use MailPoet\Automation\Engine\Workflows\Step;
+use MailPoet\Automation\Engine\Workflows\Workflow;
+use MailPoet\Automation\Integrations\Core\Actions\WaitAction;
+use MailPoet\Automation\Integrations\MailPoet\Actions\SendWelcomeEmailAction;
+use MailPoet\Automation\Integrations\MailPoet\Triggers\SegmentSubscribedTrigger;
+
+class Templates {
```
johnolek: Please tell me if you can think of a better place to put this logic. I wanted it to exist somewhere outside the controllers though because it feels like something we may want to use outside the context of API requests, and it's more straightforward to test this way.

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
```diff
@@ -0,0 +1,71 @@
...
+    $triggerStep->setNextStepId($waitStep->getId());
+
+    $sendEmailStep = $this->sendEmailActionStep();
+    $waitStep->setNextStepId($sendEmailStep->getId());
+    
+    $steps = [
+      $triggerStep,
+      $waitStep,
+      $sendEmailStep,
+    ];
```
johnolek: I added `setNextStepId` because otherwise you kind of have to build workflows in reverse order (if you want to be able to include the nextStepId in the constructor). Maybe there's a better way of handling this too.

johnolek (commented):
Just a few things that I wanted to bring the reviewer's attention to

mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsCreateFromTemplateEndpoint.php
JanJakes: Seems like not yet. We'll be able to [use enums from WP 5.7](https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api.php#L1981). Now we have `oneOf` but WP's schema doesn't support `const` so we can't do that.

That said, I guess templates may be somehow dynamic in the future (meaning it may not be just an enum of fixed strings). At least with Premium, users will be able to save their own templates.

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  private function sendEmailActionStep(?int $newsletterId = null): Step {
+    return new Step($this->uniqueId(), Step::TYPE_ACTION, $this->sendWelcomeEmailAction->getKey(), null, [
+      'welcomeEmailId' => $newsletterId
+    ]);
+  }
+
+  private function uniqueId(): string {
+    return bin2hex(random_bytes(7));
```
JanJakes: Seems like [on the JS side we use 8 bytes](https://github.com/mailpoet/mailpoet/blob/609d80ca054fb054e4ab23d93022ad46c82e2fa8/mailpoet/assets/js/src/automation/id.tsx#L1).

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
```
JanJakes: Just noted this, it's strange, does `.editorconfig` work for you?

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
JanJakes: Putting it outside the controllers sounds good to me. I'm thinking if we can improve the naming.

The class basically generates a Workflow (with step IDs and stuff) so maybe it could be `...\Templates\WorkflowGenerator` or `WorkflowBuilder`?

If the class gets large, we may as well split it into multiple ones (`WaitStepGenerator`/`WaitStepBuilder`) so that could maybe work together...

mailpoet/lib/Automation/Engine/Workflows/Workflow.php
JanJakes: I think now `draft` should be the default. Or do we set it automatically to `inactive` when it's already valid? Not sure if that's needed. We may simply have only drafts and only allow activation when the workflow is valid. In that sense, we'd just rename "inactive" to "draft".

What do you think? Not sure about this.

JanJakes (commented):
LGTM! Just a few notes.

mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsCreateFromTemplateEndpoint.php
johnolek: That's pretty much what I thought -- thanks for confirming!

mailpoet/lib/Automation/Engine/Workflows/Workflow.php
johnolek: I definitely agree about changing the default to be `draft, but I think it will still be useful to have the 3 states.

Users will certainly want the ability to turn a specific automation on or off, so I like the idea of having `inactive` for that.

Some users will probably have a lot of workflows, and it may be helpful for them to be able to view just the drafts instead of seeing drafts and inactives together, for example if they have multiple that are a work in progress and they want to find all of those and finish them.

So perhaps a `draft` can only be manually transitioned into `active`, and then it can go back and forth between `active` and `inactive`. If we did that it might almost make sense to rename `inactive` to `paused`.

Then there's the question of what should the status be if a workflow becomes invalid somehow (e.g. a user deletes the segment the trigger is based on). Should the workflow be changed back to a draft? Should there be an `invalid` status? Maybe a question for another ticket.

For now I like the idea of keeping the 3 statuses and changing the default to `draft` but I'm open to other ideas.

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
johnolek: Thank you for the much better naming ideas! I'll definitely update this.

 I had also thought about splitting things out into separate classes but that seems like something that would be good to save for when it gets larger.

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
johnolek: Oof, I thought I caught these but apparently missed this one. 

I didn't have EditorConfig support enabled in PhpStorm, so it should be working for me now. I know I previously had this other setting enabled, but somehow it got turned off:
image

Thanks for catching this!

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
johnolek: I'm going to switch this to use [\MailPoet\Util\Security::generateRandomString](https://github.com/mailpoet/mailpoet/blob/4038985b16b5e77b0eabc82fd841b573b2bd9d72/mailpoet/lib/Util/Security.php) instead with a string length of 16, unless you think that would be a problem for whatever reason.

I noticed many instances of us using 7 bytes in PHP, but upon closer inspection it looks like all of those instances are in tests.

johnolek:
@JanJakes Thanks for the review! This should be ready for you to take a second look at, but please let me know if you disagree with any of my choices or if I missed anything.

mailpoet/lib/Automation/Integrations/MailPoet/Templates/Templates.php
JanJakes: Cool, Jan L. did a [great improvement of the .editorconfig](https://github.com/mailpoet/mailpoet/compare/trunk...editorconfig) as his meetup hackaton topic so soon it may become even more useful in PHPStorm.

JanJakes (approved):
Great job!

veljkho (approved):

PR #4177, opened 2022-06-03
Fix badge alignment in React selects [MAILPOET-4057]

johnolek:
Badges inside react selects are currently top aligned, sometimes appearing cut off. This PR reverts to aligning them in the middle vertically.

## Current
Screen Shot 2022-06-03 at 12 51 37 PM

## After PR
Screen Shot 2022-06-03 at 12 52 13 PM

I believe the issue was introduced in eecbeebafd39798d22b271fe4286d3fafaafca1f as part of MAILPOET-3997. It's unclear to me why it was necessary to change the vertical alignment for that PR. MAILPOET-3997 was a ticket to standardize the size of react selects, and changing the vertical alignment back to `middle` doesn't seem to affect the size of react selects that I can see. Maybe it has an effect when there is a certain amount or type of content in the select, or for certain browsers? @NeosinneR do you happen to remember?

MAILPOET-4057

Testing instructions
1. Check the badges and checkboxes, alignment should be fine now..
2. You can check those in Segments, Forms, Newsletters..

### DIFF
--- a/mailpoet/assets/css/src/generic/_forms-select.scss
+++ b/mailpoet/assets/css/src/generic/_forms-select.scss
@@ -134,8 +134,9 @@
   background: $color-tertiary-light !important;
 }
 
-.mailpoet-form-react-select-option > span {
-  vertical-align: middle;
+.mailpoet-form-react-select-option {
+  display: flex;
+  align-items: center;
 }
 
 .mailpoet-form-react-select-tag {
--- a/mailpoet/assets/css/src/generic/_forms-select.scss
+++ b/mailpoet/assets/css/src/generic/_forms-select.scss
@@ -135,7 +135,7 @@
 }
 
 .mailpoet-form-react-select-option > span {
-  vertical-align: top;
+  vertical-align: middle;
 }
 
 .mailpoet-form-react-select-tag {


NeosinneR (dismissed):
@johnolek I don't remember any specific reason I changed that, but as it works with `middle` I think we can keep it that way.

rohankacharegs:
@johnolek  **Not verified**

For the Firefox browser, "List" name and "Close" button appear misaligned on the "Segment" page.

### Screenshot:

johnolek:
I updated the approach to use flex box instead of vertical alignment, which I think should resolve the issue pointed out by @rohankacharegs.

johnolek:
@NeosinneR I know you're busy at the moment but you looked at this before and this is also a very small change. Let me know if you'd prefer that I assign it to someone else though. I tried using the mailpoet developers group but I don't think the auto assignment worked and now I can't remove the group as a requested reviewer.

NeosinneR:
@johnolek it's okay to assign me, although I'll only have time to take a look on Friday

NeosinneR:
LGTM!

NeosinneR (approved):

veljkho (approved):

PR #4183, opened 2022-06-08
Link segment badges to filtered subscribers page [MAILPOET-4244]

johnolek:
This PR turns segment badges into clickable links that take a user to the "All Subscribers" page with that segment selected.

Hovering over a badge reveals a tooltip telling the user what will happen if they click.

image

image

This PR also tweaks the logic for which segments are included in the dropdown on subscribers listing pages to include segments where the calculated subscriber count is 0. This is to ensure that the segment a user clicks on will definitely show up selected in the dropdown.

This change should affect segment badges anywhere they appear in the UI, including:

- Emails (all tabs + statistics page)
-- Note: Badges on the stats page will not work right now if the premium plugin is activated. I have a PR for the premium plugin here: https://github.com/mailpoet/mailpoet-premium/pull/585
- Forms (listing page)
- Subscribers

MAILPOET-4244

### DIFF
--- a/mailpoet/assets/css/src/components/_tooltips.scss
+++ b/mailpoet/assets/css/src/components/_tooltips.scss
@@ -3,6 +3,7 @@
   box-shadow: 0 2px 4px 0 rgba(229, 233, 248, .4), 0 5px 30px 0 $color-tertiary-light;
   max-width: 400px;
   padding: $grid-gap-half;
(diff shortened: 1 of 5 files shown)


triple0t (approved):
Well done
LGTM 👍

veljkho (approved):

PR #4189, opened 2022-06-09
Make sure Heading block parameters are defined before setting [MAILPOET-4411]

samnajian:
Fixes [[MAILPOET-4411](https://mailpoet.atlassian.net/browse/MAILPOET-4411)]


costasovo (approved):
Thank you for fixing it so quickly! 👍

veljkho (approved):

PR #4193, opened 2022-06-15
Create a command that can calculate number of code reviews

pavel-mailpoet:
This is a meetup hackathon project 
The command is `./do display:reviewers`

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -2,8 +2,8 @@
 
 // phpcs:disable PSR1.Classes.ClassDeclaration
 // phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
-
 use MailPoetVendor\Twig\Loader\FilesystemLoader as TwigFileSystem;
+use Robo\Symfony\ConsoleIO;
 
 class RoboFile extends \Robo\Tasks {
   const ZIP_BUILD_PATH = __DIR__ . '/mailpoet.zip';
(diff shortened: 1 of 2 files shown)


mailpoet/RoboFile.php
```diff
@@ -858,6 +858,44 @@ public function releasePublish($version = null) {
...
+   * This command displays how many pull request each person did recently
+   */
+  public function displayReviewers(ConsoleIO $io) {
+    $io->progressStart(2);
+    $freePluginGithubController = $this->createGitHubController();
+    $logins = $freePluginGithubController->calculateReviewers();
+
+    $io->progressAdvance();
+    $shopGithubController = $this->createGitHubController('shop');
+    $loginsShop = $shopGithubController->calculateReviewers();
```
triple0t: I can see you listed Shop here which should point to the Shop repository but unfortunately, this is currently pointing to the Premium repository.

You need to update the check [done here](https://github.com/mailpoet/mailpoet/blob/2fadd7baf32c4f5aba825fae85d2093f1bf2d752/mailpoet/tasks/release/GitHubController.php#L277-L279) which is used to determine the accurate API base repository [here](https://github.com/mailpoet/mailpoet/blob/2fadd7baf32c4f5aba825fae85d2093f1bf2d752/mailpoet/tasks/release/GitHubController.php#L38)

mailpoet/tasks/release/GitHubController.php
```diff
@@ -39,6 +39,57 @@ public function __construct(
...
   }
 
+  /**
+   * @return array{login: string, count: int}
+   */
+  public function calculateReviewers(): array {
+    // load pull requests
+    $response = $this->httpClient->get('pulls', [
+      'query' => [
+        'per_page' => '100', // 100 is maximum, if we want more we need to implement paging
```
triple0t, line 50: I understand why we are pulling only 100 pages but wanted to note, that this would show only the most recent count (Review) for a user.

e.g for your account, there wasn't any count at all for the MP Free repo (which means You haven't reviewed any PR recently).

triple0t (changes requested):
Hi @pavel-mailpoet 

I know what this command does: show the most recent number of reviews per user

but wanted to find out, what is the **aim** behind it?

Also, Would it be more useful to be able to filter by current opened PRs?

e.g I want to request a PR review but I don't want to look through GitHub for who's free. This would have been the perfect command to locate someone with more bandwidth (less number of reviews)

mailpoet/RoboFile.php
pavel-mailpoet: Thank you for noticing that, fixed.

mailpoet/tasks/release/GitHubController.php
pavel-mailpoet: Yes, this only pulls 100 pull requests. 

> for your account, there wasn't any count at all for the MP Free repo (which means You haven't reviewed any PR recently).

that is correct.

> Also, Would it be more useful to be able to filter by current opened PRs?

This is only the first version of the command. It can be improved, and we can build upon it if we want to. Eg. we can include the premium plugin and we can add some filters. We can display people with the capacity. However, I didn't have the time to build all the bells and whistles to the command. I only had time to implement this basic version. 

In the future, if we see the value in this command we can improve it and add some more functionality.

triple0t: @pavel-mailpoet Noted. 
Thank you for working on this 👏🏾

triple0t (approved):

veljkho (approved):

PR #4194, opened 2022-06-16
Do not autoreload twig templates [MAILPOET-4439]

websupporter:
Fixes [MAILPOET-4439]

Set to true auto_reload will attempt to overwrite existing files, which is to be avoided.

Current state:
When you download the plugin, you can see how, after loading some twig templates, the template files are regenerated and the owner etc. of those files change.

A useful breakpoint for xDebug would be vendor-prefixed/twig/twig/src/Environment.phpL151. You see how because of `isAutoReload()` you skip L153, which then produces a write in L159. This produces errors, when the directory/file is not writeable.

Solution:
This PR sets `auto_reload` to `false` which then does not produce the attempted write.

[MAILPOET-4439]: https://mailpoet.atlassian.net/browse/MAILPOET-4439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Renderer.php
+++ b/mailpoet/lib/Config/Renderer.php
@@ -27,7 +27,7 @@ class Renderer {
       [
         'cache' => new TwigFileSystemCache($cachePath),
         'debug' => $this->debuggingEnabled,
-        'auto_reload' => true,
+        'auto_reload' => false,
       ]
     );
 
--- a/mailpoet/lib/Config/Renderer.php
+++ b/mailpoet/lib/Config/Renderer.php
@@ -18,7 +18,8 @@ class Renderer {
   public function __construct(
     bool $debuggingEnabled,
     string $cachePath,
-    TwigFileSystem $fileSystem
+    TwigFileSystem $fileSystem,
+    bool $autoReload = false
   ) {
     $this->debuggingEnabled = $debuggingEnabled;
     $this->cachePath = $cachePath;
@@ -27,7 +28,7 @@ class Renderer {
       [
         'cache' => new TwigFileSystemCache($cachePath),
         'debug' => $this->debuggingEnabled,
-        'auto_reload' => false,
+        'auto_reload' => $autoReload,
       ]
     );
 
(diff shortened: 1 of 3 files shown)


mailpoet/lib/Config/Renderer.php
```diff
@@ -27,7 +27,7 @@ public function __construct(
       [
         'cache' => new TwigFileSystemCache($cachePath),
         'debug' => $this->debuggingEnabled,
-        'auto_reload' => true,
+        'auto_reload' => false,
```
triple0t: Hi @websupporter, Thank you for the PR.

This is just a quick confirmation.

Would running the `./do twig:generate-cache` command regenerate the twig cache files regardless of if the template was edited or not?

From the docs: 
> [auto_reload](https://symfony.com/doc/current/reference/configuration/twig.html#auto-reload)
type: boolean default: %kernel.debug%
> 
> If true, whenever a template is rendered, Symfony checks first if its source code has changed since it was compiled. If it has changed, the template is compiled again automatically.

and from the Robo command
https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/RoboFile.php#L1020-L1033

from my little understanding, it appears this would disable auto-compilation done by Twig.

Just wanted to confirm we are regenerating the cache files as it appears we are reusing the **same configuration for the Robo command**.

triple0t (commented):
Thanks for the PR.

Kindly locate a note below

mailpoet/lib/Config/Renderer.php
websupporter: Thank you for catching that @triple0t! The twig command runs at a point in time when the generated template file does not exist. But you are right: During development, this would not regenerate templates. Looking into it for a solution :thinking:

triple0t (approved):

veljkho (approved):

PR #4196, opened 2022-06-16
A11Y captcha [MAILPOET-4203]

websupporter:
Fixes [MAILPOET-4203]
Fixes also [MAILPOET-4514]

This PR enables you to listen to the captcha and hence a broader user base to solve it.

This PR adds ~215 KB to the current zip file.

With regards of testing:
* The captcha should now have a button which plays the audio. Check if the audio is playing. 
* Check if it corresponds with the current captcha. 
* Update the captcha with the new reload button. Did the audio also update?
* Can you navigate with the tabs properly and solve the captcha?
* Can you do so using a screenreader?
* When you enter the correct code does the screenreader give you the correct feedback.
* When you enter the wrong captcha code, does the screenreader give you the correct feedback?

[MAILPOET-4203]: https://mailpoet.atlassian.net/browse/MAILPOET-4203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-4514]: https://mailpoet.atlassian.net/browse/MAILPOET-4514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samnajian:
@websupporter This is what I see when testing and I the play button doesn't play any sound either.
Screen Shot 2022-06-16 at 16 54 35

mailpoet/assets/js/src/public.tsx
```diff
@@ -53,13 +53,40 @@ jQuery(($) => {
...
   function updateCaptcha(e?: Event) {
-    const captcha = $('img.mailpoet_captcha');
-    if (!captcha.length) return false;
-    const captchaSrc = captcha.attr('src');
-    const hashPos = captchaSrc.indexOf('#');
-    const newSrc = hashPos > 0 ? captchaSrc.substring(0, hashPos) : captchaSrc;
-    captcha.attr('src', `${newSrc}#${new Date().getTime()}`);
+    const captcha = document.querySelector('img.mailpoet_captcha');
+    const audioCaptcha = document.querySelector<HTMLAudioElement>(
+      '#mailpoet_captcha_player',
```
samnajian: If we have more than one form then we'll have multiple of these ids, on the other hand let's create a const for this selector like `exitIntentEvent`

mailpoet/assets/js/src/public.tsx
```diff
@@ -447,7 +474,8 @@ jQuery(($) => {
       });
     });
 
-    $('.mailpoet_captcha_update').on('click', updateCaptcha);
+    $('#mailpoet_captcha_update').on('click', updateCaptcha);
```
samnajian: Converting ids to class names would create more valid markup in case multiple captchas are added to the page

mailpoet/lib/Subscription/Captcha.php
```diff
@@ -98,6 +99,33 @@ private function isUserExemptFromCaptcha() {
     return !empty(array_intersect($roles, (array)$user->roles));
   }
 
+  public function renderAudio($sessionId, $return = false) {
+
+    $audioPath = Env::$assetsPath . '/audio/';
```
samnajian: Feels like we can benefit from memoizing this function since it's working with IO, what do you think?

samnajian (changes requested):

mailpoet/lib/Subscription/Captcha.php
websupporter: We exit inside the function.

mailpoet/assets/js/src/public.tsx
websupporter: > let's create a const for this selector

addressed in [e5a5a3f](https://github.com/mailpoet/mailpoet/pull/4196/commits/e5a5a3fdbdeccdc3e6891dea1d78b41bb889e313)

mailpoet/assets/js/src/public.tsx
websupporter: Addressed in [e5a5a3f](https://github.com/mailpoet/mailpoet/pull/4196/commits/e5a5a3fdbdeccdc3e6891dea1d78b41bb889e313)

One remark: For now, I just changed ids to classes. Currently there can't be two captchas rendered at the same page. To allow for that, some more refactoring and extraction would be needed, which would also need to take into account the `submitSubscribeForm()` which afaik would extend this PR quite a bit, as this method is not only concerned with the captcha form.

samnajian (dismissed):
Awesome feature, thanks 👍

veljkho (changes requested):
Nice try!
1. Sometimes it won't read the captcha (intermittent issue)
2. When reloading the captcha, audio plays the previous captcha
3. The audio doesn't say whether it was small or big letter (big problem)

websupporter:
@veljkho 's findings revealed a bigger problem with the prior attempt, where the audio and the image could get out of sync or the captcha string was not yet generated when attempting to load the audio. This happend, when the browser (was able to reproduce it in Chrome) attempted to load the audio before the image.

In order to be able to get around those problems, I created something like a middlelayer ["CaptchaPhrase"](https://github.com/mailpoet/mailpoet/blob/MAILPOET-4203-a11y-captcha/mailpoet/lib/Subscription/Captcha/CaptchaPhrase.php) which returns the phrase for a given type ("audio" or "image") based on the total_loads and how many times the type was already loaded.

The idea: If you first load the image and then want to get the phrase for the audio, you want the same phrase. If you load the audio and then again the audio, you want to generate a new phrase. Same goes for image.

Not sure if that's the best approach, but it works well and keeps audio and image in sync. While trying to integrate this approach, I realized that our former Captcha system operated with quite some bloated objects and classes. A "small" refactor ended up refactoring almost the whole system, but I think it is now better maintainable and extensible.

The former `Captcha` class had a lot of mixed responsibilities. Rendering the image, check if a certain captcha type is a Google captcha and whether you needed a captcha for a specific email. Also in the SubscriberSubscribeController had a lot of responsibilities, like also the validation of a Captcha.

The last commit in this PR rewrites the system. Everything has moved into the `MailPoet\Subscription\Captcha` namespace.  We have now validators, which validate a captcha. We have a `CaptchaRenderer` "rendering" audio/image.

`CaptchaPhrase` in charge of generating/returning the current phrase.

From the former `Captcha` class only the constants and a static constant check remained, which is now in `CaptchaConstants`. Most of the existing code could just move in one of the new classes.

Regarding dev review: I am sorry for this last commit, which got quite bloated.

*Regarding testing*
This rewrite extends a bit what needs to be tested, additionally to the above, what comes into my mind:
* Does Google ReCaptcha still work?
* Does `admin` role still skip the captcha?

mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php
```diff
@@ -0,0 +1,142 @@
...
+  ) {
+    $this->subscriptionUrlFactory = $subscriptionUrlFactory;
+    $this->captchaPhrase = $captchaPhrase;
+    $this->captchaSession = $captchaSession;
+    $this->wp = $wp;
+    $this->subscriberIPsRepository = $subscriberIPsRepository;
+  }
+
+  public function validate(array $data): bool {
+    $isBuiltinCaptchaRequired = $this->isRequired(isset($data['email']) ? $data['email'] : '');
```
samnajian: The default value is `null`, so maybe the else should set to `null` as well?

mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    $subscriptionCaptchaRecipientLimit = $this->wp->applyFilters('mailpoet_subscription_captcha_recipient_limit', 0);
+    if ($subscriptionCaptchaRecipientLimit === 0) {
+      return true;
+    }
+
+    // Check limits per recipient if enabled
+    if ($subscriberEmail) {
+      $subscriber = Subscriber::where('email', $subscriberEmail)->findOne();
```
samnajian: Why not `SubscriberEntity`? I think it's better to stick to Doctrine entities in new code so that we won't need to refactor Models out of them later on

mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php
```diff
@@ -0,0 +1,142 @@
...
+      if (
+        $subscriber instanceof Subscriber
```
samnajian: Wouldn't checking if `$subscriber` is defined  ( `if ($subscriber && ...)` )  suffice instead of checking if it's an instance of `Subscriber`? especially because of [this](https://phpgoodness.wordpress.com/2010/07/23/some-php-performance-myths/)

mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php
```diff
@@ -0,0 +1,142 @@
...
+          'refresh_captcha' => true,
+        ]
+      );
+    }
+
+    return true;
+
+  }
+
+  private function getSessionId() {
```
samnajian: I feel like this method belongs to `MailPoet\Subscription\Captcha:CaptchaSession` class, and the `init` method receives the `session id` and it has a logic to check if it needs to init or not, so this logic here is redundant

mailpoet/lib/Subscription/Captcha/Validator/RecaptchaValidator.php
```diff
@@ -0,0 +1,51 @@
...
+  public function validate(array $data): bool {
+
+    $captchaSettings = $this->settings->get('captcha');
+    if (empty($data['recaptchaResponseToken'])) {
+      throw new ValidationError(__('Please check the CAPTCHA.', 'mailpoet'));
+    }
+
+    $secretToken = $captchaSettings['type'] === CaptchaConstants::TYPE_RECAPTCHA_INVISIBLE ? $captchaSettings['recaptcha_invisible_secret_token'] : $captchaSettings['recaptcha_secret_token'];
+
+    $response = $this->wp->wpRemotePost('https://www.google.com/recaptcha/api/siteverify', [
```
samnajian: I'd suggest defining the URL in a const, what do you think?

samnajian (changes requested):

mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php
websupporter: Addressed in [3950064](https://github.com/mailpoet/mailpoet/pull/4196/commits/39500648c313caea6bd3c0ad0064363537b2c060)

mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php
websupporter: Thanks for that catch! Addressed in [c6cf795](https://github.com/mailpoet/mailpoet/pull/4196/commits/c6cf7956b072c15baf1dc1e61782b537a566e104)

mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php
websupporter: Addressed in [c6cf795](https://github.com/mailpoet/mailpoet/pull/4196/commits/c6cf7956b072c15baf1dc1e61782b537a566e104)

mailpoet/lib/Subscription/Captcha/Validator/RecaptchaValidator.php
websupporter: Addressed in [d86d690](https://github.com/mailpoet/mailpoet/pull/4196/commits/d86d690e9acfc68f90f660f1e67063078625578b)

mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php
websupporter: Addressed in [db4fc58](https://github.com/mailpoet/mailpoet/pull/4196/commits/db4fc584c4e90875fb4269778007f022ca932608)

samnajian (dismissed):
LGTM thanks!

veljkho (changes requested):
Everything's good except:
- As a not logged in user + having Google ReCaptcha (v2, visible/invisible) and performing subscribing to a form... I see forever loading wheel when tapping Subscribe button (and of course solving captcha)... so there's some conflict so I can't subscribe. PS. Even with Disabled captcha in MP's settings I'm getting the same problem. Checked this on production and issue is not present.
- If using Screen Reader and completing the subscription form (solving built-in captcha), I don't hear the success message so I know that I'm subscribed. No nothing after successfully subscribing. Can we include the success message at the end for the reader?

websupporter:
Thank you for your review @veljkho 

> As a not logged in user + having Google ReCaptcha (v2, visible/invisible) and performing subscribing to a form... I see forever loading wheel when tapping Subscribe button (and of course solving captcha)... so there's some conflict so I can't subscribe. PS. Even with Disabled captcha in MP's settings I'm getting the same problem. Checked this on production and issue is not present.

That should be resolved.

> If using Screen Reader and completing the subscription form (solving built-in captcha), I don't hear the success message so I know that I'm subscribed. No nothing after successfully subscribing. Can we include the success message at the end for the reader?

I am not quite sure what is causing this on your side. It did work for me. I raised the `aria-live` from `polite` to `assertive` and hope this helps :)

Aschepikov (approved):
GJ!

veljkho (approved):

PR #4197, opened 2022-06-16
Use original WordPress Mailer as fallback for transactional mailer [MAILPOET-4254]

costasovo:
This PR changes the fallback method for transactional emails when they are set to be sent using MailPoet. 
Prior to this change, we used PHPMailer with the default configuration as the fallback. Now we use PHPMailer as configured within the wp_mail function.

The problem it fixes is that on some hosts it is not possible to send with the PHP mail function which is used as default in PHPMailer and it is more likely that the WP PHPMailer would work.

### Testing instructions
1) Setup a WP site and make sure transactional emails work even without MailPoet (e.g., add a new WP user, create new Woo order).
2) Install MailPoet and set transactional emails to be sent via MailPoet (Settings > Advanced > The current sending method)
3) Misconfigure MailPoet sending settings so that it doesn't send (e.g. use an invalid API key or use SMTP with no existing host)
4) Make some action that triggers transactional email (e.g., add a new WP user, create new Woo order).
5) Verify that you received the email even when MailPoet is misconfigured

Also, it might be a good idea to test on an Atomic site, because that's where the issue was discovered.

[MAILPOET-4254]

[MAILPOET-4254]: https://mailpoet.atlassian.net/browse/MAILPOET-4254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Settings/TransactionalEmailsCest.php
+++ b/mailpoet/tests/acceptance/Settings/TransactionalEmailsCest.php
@@ -9,8 +9,10 @@ class TransactionalEmailsCest {
   public function sendTransactionalEmailFallback(\AcceptanceTester $i) {
     $i->wantTo('Check that transactional email are sent even when MailPoet sending doesn‘t send');
     $settings = new Settings();
+    $i->wantTo('Setup MailPoet to send transactional emails but having misconfigured SMTP settings.');
     $settings->withMisconfiguredSendingMethodSmtpMailhog();
     $settings->withTransactionEmailsViaMailPoet();
+    $i->wantTo('Create a new WP user and make sure transactional email were received');
     $i->cli(['user', 'create', 'john_doe', 'john_doe@example.com', '--send-email']);
     $i->amOnMailboxAppPage();
     $i->checkEmailWasReceived('New User Registration');
--- a/mailpoet/tests/acceptance/Settings/TransactionalEmailsCest.php
+++ b/mailpoet/tests/acceptance/Settings/TransactionalEmailsCest.php
@@ -10,7 +10,7 @@ class TransactionalEmailsCest {
     $i->wantTo('Check that transactional email are sent even when MailPoet sending doesn‘t send');
     $settings = new Settings();
     $i->wantTo('Setup MailPoet to send transactional emails but having misconfigured SMTP settings.');
-    $settings->withMisconfiguredSendingMethodSmtpMailhog();
+    $settings->withMisconfiguredSendingMethodSmtp();
     $settings->withTransactionEmailsViaMailPoet();
     $i->wantTo('Create a new WP user and make sure transactional email were received');
     $i->cli(['user', 'create', 'john_doe', 'john_doe@example.com', '--send-email']);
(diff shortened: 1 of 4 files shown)


mailpoet/tests/integration/Mailer/WordPress/WordpressMailerTest.php
```diff
@@ -43,9 +44,7 @@ public function testItFormatsTextNewsletterForMailer() {
         ],
       ]))
       ->willReturn(['response' => true]);
-    $fallbackMailer = $this->createMock(Mailer::class);
-    $fallbackMailer->expects($this->never())->method('send');
-    $mailerFactoryMock = $this->createMailerFactoryMock($mailer, $fallbackMailer);
```
costasovo, line 48: Every single test was checking that the fallback was not called when the primary mailer was successful. I kept the check only in one test case. I think it was redundant and the new approach requires more code to prepare the mock for the fallback.

mailpoet/tests/integration/Mailer/WordPress/WordpressMailerTest.php
```diff
@@ -218,35 +230,6 @@ public function testItThrowsOnUnknownContentType() {
     $wpMailer->send();
   }
 
-  public function testItTranslateExeceptionsToPhpmailerException() {
```
costasovo, line 221: I removed this test because it was a copy of the `testItThrowsOnError` test.

mailpoet/tests/acceptance/Settings/TransactionalEmailsCest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace MailPoet\Test\Acceptance;
+
+use MailPoet\Test\DataFactories\Settings;
+
+class TransactionalEmailsCest {
+
+  public function sendTransactionalEmailFallback(\AcceptanceTester $i) {
```
websupporter, line 9: Out of curiosity: Would this test succeed even with MailPoet deactivated? What I mean, is our acceptance test server configured in a way, that it would send emails with `wp_mail()` when MailPoet is deactivated? Because in that case I wonder about the meaningfulness of this test :thinking:

costasovo: Good question! It is configured to catch emails send by the PHP mail function (including wp_mail()) by MailHog. Actually, we added it just recently when we switched integration tests to the same infra (https://github.com/mailpoet/wordpress-images/pull/34).
To be sure I tested it and the test passes even when MailPoet is deactivated. 

But! I see that I forgot to ensure in the test that MailPoet is set to send transactional emails. 🤦 I need to add it.

So the goal of the test is that even though the MailPoet sending is misconfigured (it emulates also being blocked by MSS for spam reasons) the transaction emails are still delivered.

websupporter (dismissed):
Thank you for working on this @costasovo 

lgtm

mailpoet/tests/acceptance/Settings/TransactionalEmailsCest.php
costasovo: I've just realized that also the name `withMisconfiguredSendingMethodSmtpMailhog` is not correct. Because it is not Mailhog anymore it is just misconfigured and set to some non-extisting server. Let me update it.

websupporter (approved):
Thanks :bow: looks still good :smile:

veljkho (approved):

PR #4198, opened 2022-06-17
Automation listing prototype [MAILPOET-4404]

websupporter:
Fixes [MAILPOET-4404]

I quickly finished @JanJakes prototype
> What it now needs is just being mergeable to master. Things like paging don’t need to work and the design doesn’t have to be great. So what I’d do is probably just fix the last WIP commit so linters pass, maybe extract the listing component to a new file, and that would be it for now.

I added quickly some types, more columns to the table, mainly to see how this could work, created a small onboarding screen and our usual loading interface.

[MAILPOET-4404]: https://mailpoet.atlassian.net/browse/MAILPOET-4404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/list/automation-listing.tsx
```diff
@@ -0,0 +1,70 @@
...
+  loading,
+}: AutomationListingProps): JSX.Element {
+  const rows = workflows.map((workflow) => Row(workflow));
+
+  const headers = [
+    { key: 'name', label: __('Name', 'mailpoet') },
+    { key: 'subscribers', label: __('Subscribers', 'mailpoet') },
+    { key: 'status', label: __('Status', 'mailpoet') },
+    { key: 'edit', label: '' },
+    { key: 'more', label: '' },
```
johnolek: I know this PR isn't supposed to be a polished version or anything, but since column showing/hiding does actually work at the moment maybe we should give these labels so a user knows what they're toggling.

https://user-images.githubusercontent.com/8656640/175403988-393a63b5-a293-4178-aa63-9f334bb52602.mov

mailpoet/assets/js/src/automation/list/cells/More.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { WorkflowProps, WorkflowPropsShape } from '../workflow';
+
+export function More({ workflow }: WorkflowProps): JSX.Element {
+  return (
+    <EllipsisMenu
+      label={`Actions for ${workflow.name}`}
+      renderContent={() => (
+        <div>
+          <MenuItem onInvoke={() => {}}>
+            <p>{__('Rename automation', 'mailpoet')}</p>
```
johnolek: Should this be `Rename workflow`?

mailpoet/assets/js/src/automation/list/cells/Status.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const toggleStatus = () => {
+    // @Todo
+  };
+
+  return (
+    <div>
+      <Toggle onCheck={toggleStatus} />
+      {workflow.status === WorkflowStatus.ACTIVE
+        ? __('Active', 'mailpoet')
+        : __('Not active', 'mailpoet')}
```
johnolek: Maybe this has been discussed already, but after [MAILPOET-4264](https://mailpoet.atlassian.net/browse/MAILPOET-4264) we have active, inactive, and draft status. I don't know if we're going to keep all of those statuses for MVP, but if we want to have more than 2 we may need to rethink using a toggle here. 

I don't think this needs to be changed for this PR, mostly wondering if we should create a ticket or something.

mailpoet/assets/js/src/automation/list/automation-listing.tsx
```diff
@@ -0,0 +1,70 @@
...
+      title=""
+      isLoading={workflows.length === 0 || loading}
+      rows={rows}
+      headers={headers}
+      query={{ page: 2 }}
+      rowsPerPage={7}
+      totalRows={workflows.length}
+      hasSearch
+      actions={[
+        <ul className="subsubsub" style={{ width: '400px' }}>
```
johnolek: This will need a `key` to avoid a react warning

image

mailpoet/assets/js/src/automation/list/workflow.tsx
```diff
@@ -0,0 +1,22 @@
+import PropTypes from 'prop-types';
+
+export enum WorkflowStatus {
+  ACTIVE = 'active',
+  INACTIVE = 'inactive',
```
johnolek: Depending on how we want to handle statuses we may want to include `draft` here too

mailpoet/tasks/npm_post_install.sh
```diff
@@ -21,3 +21,5 @@ sed -i -- "s/_focusedField\.focus()/_focusedField\.trigger('focus')/g" node_modu
 sed -i -- "s/  lastKeyIndex = keyPath.length-1;/  var lastKeyIndex = keyPath.length-1;/g" node_modules/backbone.supermodel/build/backbone.supermodel.js
 sed -i -- "s/  key = keyPath\[i\];/  var key = keyPath\[i\];/g" node_modules/backbone.supermodel/build/backbone.supermodel.js
 
+# Fix @woocommerce/components having invalid CSS not parseable by SASS
+sed -i -- "s/[(]1fr[)]/1fr /g" node_modules/@woocommerce/components/build-style/style.css
```
johnolek, line 25: Any idea what the root cause of this issue is? It would be nice if we could link to a GitHub issue and say "we can remove this code when [issue here] is fixed."

I looked at the differences between the original and what we get after running this fix, and I think they're all `-ms-grid-columns` rules like this:

image

I spent a bit of time trying to figure out how it ended up like this to begin with but came up empty handed.

No worries if you're not sure yourself, @websupporter.

johnolek (changes requested):
@websupporter Sorry for the delay in reviewing this.

Overall I think this looks great! The only stuff I care about changing right now is avoiding the react `key` warning and doing something about the empty column labels for "edit" and "more".

mailpoet/assets/js/src/automation/list/cells/More.tsx
websupporter: Not according to the current Figma
image

mailpoet/tasks/npm_post_install.sh
websupporter: Sorry, I do not know :/

mailpoet/assets/js/src/automation/list/automation-listing.tsx
websupporter: Addressed in a6f4c06e8

mailpoet/assets/js/src/automation/list/automation-listing.tsx
websupporter: Addressed in 3604b0532

websupporter:
Hi @johnolek.

Thanks for the review. The issues have been addressed.

johnolek (dismissed):

mailpoet/tasks/npm_post_install.sh
JanJakes: I'm not aware of any existing issue, the file just contains invalid CSS (those lines you've found).

mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -13,71 +16,14 @@ function Workflows(): JSX.Element {
...
-          //}
-          //selected={ searchedLabels }
-          showClearButton
-          type="custom"
-          disabled={!data || loading || data.length === 0}
-          autocompleter={{}}
-        />,
-      ]}
-    />
+  const workflows: Workflow[] = data?.data ?? [];
```
JanJakes: This is a very cosmetic thing but probably better to add the type as a generic argument to `useQuery`.

mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -13,71 +16,14 @@ function Workflows(): JSX.Element {
...
+  if (loading) {
+    return <Loading />;
+  }
+  return workflows.length === 0 ? (
+    Onboarding()
```
JanJakes: I wonder if there's a reason why you didn't use `<Onboarding>` or whether this was just by mistake.

mailpoet/assets/js/src/automation/list/automation-listing.tsx
```diff
@@ -0,0 +1,70 @@
+import { Search, TableCard } from '@woocommerce/components';
+import PropTypes from 'prop-types';
```
JanJakes: @websupporter Please note that we don't use prop-types in  because:
1. They are not very useful when props are typed in TypeScript.
2. [We use a Babel plugin to compile TypeScript types to prop types automatically](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/.babelrc#L9). We may remove it in the future though as it seems to have no real benefits.

I know this may be confusing because you can find prop types in the old JS code and sometimes they might have persisted even during TS rewrites but the situation is no prop types in new TS code.

mailpoet/assets/js/src/automation/list/automation-listing.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { __ } from '@wordpress/i18n';
+import { Row } from './row';
+import { Workflow, WorkflowPropsShape } from './workflow';
+
+interface AutomationListingProps {
```
JanJakes: I think we usually use `type` for props so we should keep using them here for consistency.

mailpoet/assets/js/src/automation/list/automation-listing.tsx
```diff
@@ -0,0 +1,70 @@
+import { Search, TableCard } from '@woocommerce/components';
```
JanJakes: I guess for consistency with other listings the directory should be `listing`, not `list`.

We may also use `index.tsx` as the other related code does.

mailpoet/assets/js/src/automation/list/row.tsx
```diff
@@ -0,0 +1,31 @@
+import { Workflow } from './workflow';
+import { Subscribers } from './cells/Subscribers';
+import { Status } from './cells/Status';
+import { Name } from './cells/Name';
+import { Edit } from './cells/Edit';
+import { More } from './cells/More';
+
+export function Row(workflow: Workflow): object[] {
```
JanJakes: This doesn't seem to be a React component. Therefore, it should probably use lower-cased imperative naming, such as `getRow`.

mailpoet/assets/js/src/automation/list/cells/Edit.tsx
```diff
@@ -0,0 +1,19 @@
+import { __ } from '@wordpress/i18n';
+import PropTypes from 'prop-types';
+import { WorkflowProps, WorkflowPropsShape } from '../workflow';
+
+interface EditProps extends WorkflowProps {
+  label?: string;
+}
+export function Edit({ workflow, label = null }: EditProps): JSX.Element {
```
JanJakes: This actually makes label `string | undefined | null`. The default value can be omitted.

mailpoet/assets/js/src/automation/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,21 @@
+import { __ } from '@wordpress/i18n';
+import { CreateWorkflowFromTemplateButton } from '../testing';
+
+export function Onboarding() {
```
JanJakes: This misses a return type.

mailpoet/assets/js/src/automation/list/workflow.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const WorkflowPropsShape = {
+  id: PropTypes.number.isRequired,
+  name: PropTypes.string.isRequired,
+  status: PropTypes.string.isRequired,
+};
+
+export interface WorkflowProps {
+  workflow: Workflow;
+}
```
JanJakes: This seems to me rather like a strange place to put prop typings. The file defines some generic data types while props are React-component specific.

mailpoet/assets/js/src/automation/listing/cells/Status.tsx
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export function Status({ workflow }: Props): JSX.Element {
+  const toggleStatus = () => {
+    // @Todo
+  };
+
+  return (
+    <div>
+      <Toggle onCheck={toggleStatus} />
```
JanJakes: Seems like you're using a toggle component from the MailPoet design system. We need to use the toggle from `@wordpress/components`.

mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -1,36 +1,30 @@
 import ReactDOM from 'react-dom';
+import { Workflow } from './listing/workflow';
+import { AutomationListing } from './listing';
+import { Onboarding } from './onboarding';
+import { Loading } from '../common';
```
JanJakes: I don't think we should use the MailPoet-specific spinner here. There are some spinners in `@wordpress/components` or maybe in the Woo components but I think I'll just revert it and leave it for later.

JanJakes (commented):
@websupporter I realized this was not merged yet so I reviewed it a bit and added some stuff (see the changed commits).

cc @johnolek

JanJakes (approved):

mailpoet/assets/js/src/automation/automation.tsx
websupporter: Thanks @JanJakes, By mistake.

websupporter:
Thanks for your review @JanJakes. There were some helpful comments in it. Much appreciated.

PR #4204, opened 2022-06-21
Made checkboxes in a form accessible according to the WCAG-standards …

simsekb:
…by adding a fieldset around them and adding a legend describing the fieldset's purpose (name).

### DIFF
--- a/mailpoet/assets/css/src/components-public/_public.scss
+++ b/mailpoet/assets/css/src/components-public/_public.scss
@@ -42,12 +42,6 @@ $form-line-height: 1.4;
       font-size: $form-mobile-font-size !important;
       line-height: $form-line-height !important;
     }
-
-    fieldset {
-      border: 0;
-      margin: 0;
-      padding: 0;
-    }
   }
 
   &.mailpoet_form_successfully_send {
@@ -68,6 +62,12 @@ $form-line-height: 1.4;
 
   .mailpoet-has-font-size {
     line-height: $form-line-height;
+  }
+
+  fieldset {
+    border: 0;
+    margin: 0;
+    padding: 0;
   }
 }
 
--- a/mailpoet/assets/css/src/components-public/_public.scss
+++ b/mailpoet/assets/css/src/components-public/_public.scss
@@ -42,6 +42,12 @@ $form-line-height: 1.4;
       font-size: $form-mobile-font-size !important;
       line-height: $form-line-height !important;
     }
+
+    fieldset {
+      border: 0;
+      margin: 0;
+      padding: 0;
+    }
   }
 
   &.mailpoet_form_successfully_send {
@@ -554,10 +560,4 @@ h2.mailpoet-heading {
   top: 5px;
   width: auto;
   z-index: 100000;
-}
-
-fieldset {
-  border: 0;
-  margin: 0;
-  padding: 0;
 }
--- a/mailpoet/assets/css/src/components-public/_public.scss
+++ b/mailpoet/assets/css/src/components-public/_public.scss
@@ -555,3 +555,9 @@ h2.mailpoet-heading {
   width: auto;
   z-index: 100000;
 }
+
+fieldset {
+  border: 0;
+  margin: 0;
+  padding: 0;
+}
--- a/mailpoet/lib/Form/Block/BlockRendererHelper.php
+++ b/mailpoet/lib/Form/Block/BlockRendererHelper.php
@@ -34,7 +34,7 @@ class BlockRendererHelper {
       $rules['required'] = true;
       $rules['minlength'] = ModelValidator::EMAIL_MIN_LENGTH;
       $rules['maxlength'] = ModelValidator::EMAIL_MAX_LENGTH;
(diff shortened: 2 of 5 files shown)


simsekb:
I accidently commited another change into this branch..

costasovo (changes requested):
Thank you for your contribution! I'm sorry it took me so long to get to this!

The change makes sense. The thing is that adding wrapping those checkboxes into a fieldset would change the appearance forms which already exist on user sites.

Here is an example:

**1) Form before the change**
Screenshot 2022-07-11 at 15 21 07

**2) After switching to this branch**
Screenshot 2022-07-11 at 15 21 17

I think we could work around this by adding some CSS reset for the fieldset element inside mailpoet forms. The reset could be added into [a file with CSS rules used for the forms on the frontend](https://github.com/mailpoet/mailpoet/blob/7e59ad7f9061f2b9df285d2745ba858168661b57/mailpoet/assets/css/src/components-public/_public.scss).

Could you please add the CSS rules to ensure we don't change the appearance?

As for the second change, I'm not sure I fully understand the reason. Could you please explain a bit what was the problem?

simsekb:
> Thank you for your contribution! I'm sorry it took me so long to get to this!
> 
> The change makes sense. The thing is that adding wrapping those checkboxes into a fieldset would change the appearance forms which already exist on user sites.
> 
> Here is an example:
> 
> **1) Form before the change** Screenshot 2022-07-11 at 15 21 07
> 
> **2) After switching to this branch** Screenshot 2022-07-11 at 15 21 17
> 
> I think we could work around this by adding some CSS reset for the fieldset element inside mailpoet forms. The reset could be added into [a file with CSS rules used for the forms on the frontend](https://github.com/mailpoet/mailpoet/blob/7e59ad7f9061f2b9df285d2745ba858168661b57/mailpoet/assets/css/src/components-public/_public.scss).
> 
> Could you please add the CSS rules to ensure we don't change the appearance?
> 
> As for the second change, I'm not sure I fully understand the reason. Could you please explain a bit what was the problem?

Haha, no worries, it can happen to the best of us;)

Hmm, I thought I had that covered too, maybe our own MP-addon already had that one fix it. I'll commit a change for that one tommorow morning around 9-10 AM. As for the second change; when filling in an email field, there should be two errors. The one when the email is filled in incorrectly (is in a valid email format: 'type-message') and when the email field is not empty ('required-message'). When looking at the current situation, the 'error-message' overrules both situations and always shows the message: "Please specify a valid email address".

So when people with impaired vision enter an email address and it doesn't meet the requirements or the field is by any chance empty they will get a clear error saying what's wrong and will tell them what they should do to fill it in correctly. As for the current form, those users will have no idea what's wrong with the filled in email address.

mailpoet/assets/css/src/components-public/_public.scss
```diff
@@ -555,3 +555,9 @@ h2.mailpoet-heading {
   width: auto;
   z-index: 100000;
 }
+
+fieldset {
```
costasovo: This will change the global fieldset and possibly may influence other fieldsets on a site. Wouldn't it be better to restrict it only for fieldset inside `.mailpoet_form`? This way the rule will have also bigger specificity so lower not get overridden by a theme.

costasovo (commented):
Thanks for your quick reaction! I think it might be better to restrict the fieldset CSS reset so that it is applied only for the MailPoet forms.

mailpoet/assets/css/src/components-public/_public.scss
simsekb: You're absolutely right! I'm committing the change now.

costasovo (approved):
Thank you @simsekb! The PR looks good to me! We still haven't solved the issue with the public builds so I'll recreate the PR under my user (maybe also merge some commits). Same as I did with the previous PR.

Just want to let you know I'll get to this the next Monday. Thanks again for your contribution and your patience!

simsekb:
> Thank you @simsekb! The PR looks good to me! We still haven't solved the issue with the public builds so I'll recreate the PR under my user (maybe also merge some commits). Same as I did with the previous PR.
> 
> Just want to let you know I'll get to this the next Monday. Thanks again for your contribution and your patience!

Awesome! When do you think these changes would go into production?

simsekb:
> Thank you @simsekb! The PR looks good to me! We still haven't solved the issue with the public builds so I'll recreate the PR under my user (maybe also merge some commits). Same as I did with the previous PR.
> 
> Just want to let you know I'll get to this the next Monday. Thanks again for your contribution and your patience!

Awesome! When do you think these changes would go into production? @costasovo

costasovo:
@simsekb I'm closing this PR in favor of a new PR I've created: https://github.com/mailpoet/mailpoet/pull/4256. 
I also added a couple of additional commits for issues detected by tests on CI and a few things (e.g. it also affects manage subscription forms, legend input needs escaping) I missed in the original review.

alex-mailpoet:
@simsekb This change was released today in MailPoet 3.94.0.

PR #4208, opened 2022-06-22
Mailpoet 4222 update vulnerable dependencies

samnajian:
This PR addresses items in [MAILPOET-4222]

[MAILPOET-4222]: https://mailpoet.atlassian.net/browse/MAILPOET-4222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/package.json
```diff
@@ -180,24 +182,38 @@
...
     "wp-types": "^3.59.1"
+  },
+  "overrides": {
+    "prismjs": "$prismjs",
+    "webpack": "$webpack",
+    "react": "$react",
+    "underscore": "$underscore",
+    "trim": "$trim",
+    "glob-parent": "$glob-parent",
+    "meow": "$meow"
```
johnolek: @samnajian I haven't used `overrides` like this before. If I understand it correctly this will force any package that's referencing one of these to use the versions we're specifying in `dependencies` or `devDependencies`, is that right?

Would this explain why so many lines have been removed in `package-lock.json` compared to the number added? Perhaps there were lots of packages that were depending on different versions of these packages, and now that they're all referencing the same exact versions there are much fewer dependencies in general?

image

I'm just wondering if it's really necessary to include all of these in `overrides` and if it might cause problems down the line. Although this seems to be working fine now, what happens if we add a different package in the future that depends on `webpack` or `react` and doesn't work with the version we're using?

Please correct me if I'm not understanding this correctly.

It seems like we would want to keep this list as small as possible to reduce unexpected side effects in the future. Maybe we could document somewhere in this file, or the commit message(s), the specific packages that require us to include these overrides? 

Maybe we'll be able to remove some overrides in the future but it would be somewhat difficult to know which are okay to remove without more documentation (unless I'm missing something, in which case my apologies!).

johnolek (commented):
@samnajian I think this looks good, but I do have one question about the `overrides`, which I've included as a separate comment.

I think it might be good to get someone else, someone who's more familiar with the plugin's dependencies, to take a quick look at this just because I don't feel familiar enough to approve all these changes with confidence.

mailpoet/package.json
samnajian: Thanks for taking into it @johnolek.
Your analogy for the big number of changes is the package-lock.json is correct.

> I'm just wondering if it's really necessary to include all of these in overrides and if it might cause problems down the line. Although this seems to be working fine now, what happens if we add a different package in the future that depends on Webpack or React and doesn't work with the version we're using?

On one hand, most of the overrides ( other than underscore ) are made to mitigate issues arising from packages used by Storybook which means it's pretty safe for the production build, and also the overrides are basically forcing Storybook to use the versions we use for the main app. 
On the other hand, these overrides are done via reference to the versions installed in dependencies/devDependencies sections ( using variables instead of hard-coding versions ), so updating them in the future, for the most part, wouldn't mean that we need to touch the overrides as well.

I hope this explains the PR a bit better, let me know if you still think we'd need more supporting documentation and then maybe we can work on a solution.

johnolek: @samnajian Thanks for the explanation!

If the vulnerable packages related to storybook aren't included/used in the production build, is it necessary to override them at all right now?

Based on the following discussions, it looks like storybook will have a vulnerability-free dependency tree with version 7.0:
https://github.com/storybookjs/storybook/issues/14603
https://github.com/storybookjs/storybook/issues/15174

Looking at the current [storybook releases](https://github.com/storybookjs/storybook/releases), it looks like they're currently have version 7 alpha 6, so it probably won't be too long until we could just update to version 7.

It makes sense to me to use overrides for dependencies that are no longer maintained, like `backbone.supermodel`. That package hasn't seen an update in 7 years, and because it's depending on `underscore` with a version of [~1.5.2](https://github.com/tanqhnguyen/backbone.supermodel/blob/9780ba0c855cf16ca3900ead2027ec3974494432/package.json#L33) it's preventing us from updating to a newer minor version of `underscore`. Therefore our only viable option in that case is to use an override. Am I understanding that correctly?

Since storybook version 7 should no longer have any vulnerable dependencies in its tree, and because the vulnerabilities that currently exist should only ever affect our dev environment, it seems like it would be a better option to wait for that version.

But even if these dependencies are only used by storybook right now, isn't it possible that they might get referenced by some of our production dependencies in the future?

For example, could something like this happen?

1. We lock some package P to version `^6.0.1` in our overrides. Right now only storybook depends on it and nothing breaks by forcing the override.
2. Some months or years from now we update one of our prod dependencies Q, and it now depends on package P version `^7.0`.
3. Because of our overrides, package Q will be forced to use version `^6.0.1` instead of `^7.0`. This could cause some unexpected and difficult-to-debug behavior.

Perhaps that would be very unlikely, but it still seems like it would be best to minimize the risk of it happening at all, unless `npm` would tell us that something like this had happened.

Long story short, my impression is that the more overrides we have the more risk there is that we would see subtle problems in the future, and it could be difficult to know when it's safe to update those overrides. 

Please let me know if I'm misunderstanding anything and/or if you think there's a compelling reason to use overrides now for the storybook-related vulnerabilities instead of waiting to update to storybook version 7 when it's released.

samnajian: @johnolek thanks alot for a deep look into things.

> Long story short, my impression is that the more overrides we have the more risk there is that we would see subtle problems in the future, and it could be difficult to know when it's safe to update those overrides.

The first part is correct, but for the second part: later on for instance as we update the storybook ( per our usual recurring npm package update tickets like [MAILPOET-4257] ) we'll simply remove the overrides and run `npm audit` again which would tell us if we are green or not, the statement you made about future updates is always true and we should keep running npm audit after each npm package update and I don't think this ticket is going to change that.

> Please let me know if I'm misunderstanding anything and/or if you think there's a compelling reason to use overrides now for the storybook-related vulnerabilities instead of waiting to update to storybook version 7 when it's released.

I'd like to get rid of the vulnerabilities as soon as possible for obvious reasons and then enforce running npm audit after each package update and update the overrides since just like what you mentioned in a perfect world we should have no overrides at all and we can always drive towards perfection!

johnolek (dismissed):
Sorry for the delay in adding my approval, @samnajian. I think your explanation makes sense and thank you for the discussion we had about it. Since I don't have much experience with using overrides I do still think it might be good to get a second opinion from someone else on the team, but I'll leave that up to you, which is why I'm going to assign this back to you. Do feel free to move it along to QA.

samnajian:
> Sorry for the delay in adding my approval, @samnajian. I think your explanation makes sense and thank you for the discussion we had about it. Since I don't have much experience with using overrides I do still think it might be good to get a second opinion from someone else on the team, but I'll leave that up to you, which is why I'm going to assign this back to you. Do feel free to move it along to QA.

Thanks for all the great in-depth look @johnolek.
@rodrigoprimo I'll appreciate it if you could spend some time to give this a second look.

mailpoet/package.json
```diff
@@ -158,12 +158,13 @@
     "@types/wordpress__url": "^2.14.0",
     "@wordpress/browserslist-config": "^4.1.2",
     "@wordpress/dependency-extraction-webpack-plugin": "^3.4.1",
-    "@wordpress/scripts": "^22.3.0",
+    "@wordpress/scripts": "^23.3.0",
```
rodrigoprimo: I saw that you mentioned in the commit message that due to how we use this package, it is safe to update it here. That being said, the guide for updating @wordpress packages mentions that "Since we experienced that upgrading packages one by one may lead to ending up with multiple versions of different JS packages the recommended approach is removing all @wordpress/* packages and installing them again."

So could you please share more details on why we need to update this package? Maybe since it is a dev dependency it is fine to update it in another ticket together with the rest of the @wordpress packages?

rodrigoprimo (commented):
I don't have a strong opinion here, but I share @johnolek's concern about the added complexity of using overrides, and considering that Storybook is a dev dependency and it looks like they are going to release a new version soon, I would also suggest that we wait for the release instead of updating Storybook dependencies manually in this PR.

I also left a separate comment about @wordpress/scripts.

samnajian:
> I don't have a strong opinion here, but I share @johnolek's concern about the added complexity of using overrides, and considering that Storybook is a dev dependency and it looks like they are going to release a new version soon, I would also suggest that we wait for the release instead of updating Storybook dependencies manually in this PR.
> 

I've added a team handbook page to guide devs into working with overrides when updating packages.

mailpoet/package.json
samnajian: I guess in this case it only makes sense to update other packages starting with `@wordpress/`  in devDependencies, what do you think?

rodrigoprimo: I don't know to be honest, but I would assume it would be necessary to update all `@wordpress/` and not only those in `devDependencies`.

If there are no pressing reasons to update `@wordpress/scripts`, which seems to me to be the case, but please correct me if I'm wrong, I would suggest waiting and updating `@wordpress/scripts` the next time someone works on the ticket to update the `@wordpress/` packages instead of updating it in this PR.

rodrigoprimo:
This guide will be helpful if we decide to start using overrides for sure.

I apologize if my previous comment was not clear, but my suggestion was that we don't use it for now and wait for a new version of Storybook to be released, which I believe is more or less what @johnolek was suggesting. But, of course, I could be missing something, and it might make sense to use the overrides. I was expecting some input from you on whether you agree with the suggestion to wait for a new Storybook version or to continue the discussion on why it would be preferable to follow a different path and use overrides.

rodrigoprimo (dismissed):
LGTM

I chatted with Sam, and he clarified my concerns that I expressed in the previous review. This latest version of the changes proposed here use `overrides` only for packages that are regular dependencies (not dev dependencies) or that are required to be able to install the packages in the latest versions of npm.

veljkho:
Hey @samnajian 
There's one conflicting file. Can we do the rebase here?

samnajian:
> Hey @samnajian There's one conflicting file. Can we do the rebase here?

It's ready to go :)

veljkho (approved):

PR #4211, opened 2022-06-24
Introduce a SubjectFactory [MAILPOET-4449]

websupporter:
Fixes [MAILPOET-4449]

Depending on a specific integration, different subjects might exist for a WorkflowRun. This commit introduces the concept of a SubjectFactory, which can be registered by an integration and return a Subject depending on a given key.

The `Registry` with which integrations can register actions and triggers is used to also register a `SubjectFactory`. Such a factory returns an empty `Subject` for a given key, which the caller then can load. I use the `WorkflowRunStorage` as a caller. Here the subjects are instantiated. I do not really like this location, as it means, it spills over the knowledge on how to construct a subject. So far the storage just had to have an array of `$data`, now it needs to now that this array contains a `subject` property, which is a JSON encoded string, knowledge which resided only in the WorkflowRun so far, as encoding and decoding where done here. On the other hand I didn't want to inject the Registry or the existing SubjectFactories into the WorkflowRun itself, but that would be an alternative in my mind. Maybe something like
```
public static function fromArray(array $data, SubjectFactory ...$factories) : self { ... }
```

The `MailPoetIntegration` shows how such a factory can be registered and the `MailPoetSubjectFactory` how it could work.

I added an integration test, which demonstrates, how this integration is supposed to work. It creates a new workflow with a SegmentSubject and a SubscriberSubject and stores it in the database. Afterwards we test, if the subjects are correctly returned when fetched from the database.

[MAILPOET-4449]: https://mailpoet.atlassian.net/browse/MAILPOET-4449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Exceptions/UnexpectedSubjectType.php
+++ b/mailpoet/lib/Automation/Engine/Exceptions/UnexpectedSubjectType.php
@@ -1,0 +1,8 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Exceptions;
+
+class UnexpectedSubjectType extends RuntimeException {
+
+
+}
--- a/mailpoet/lib/Automation/Engine/Registry.php
+++ b/mailpoet/lib/Automation/Engine/Registry.php
@@ -3,6 +3,7 @@
 namespace MailPoet\Automation\Engine;
 
 use MailPoet\Automation\Engine\Workflows\Action;
+use MailPoet\Automation\Engine\Workflows\SubjectFactory;
 use MailPoet\Automation\Engine\Workflows\Trigger;
 
 class Registry {
@@ -11,6 +12,9 @@ class Registry {
 
   /** @var array<string, Action> */
   private $actions = [];
+
+  /** @var array<SubjectFactory> */
+  private $subjectFactories = [];
 
   public function addTrigger(Trigger $trigger): void {
     $key = $trigger->getKey();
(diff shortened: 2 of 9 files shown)


JanJakes (commented):
Nice catch!

I'm not sure about the solution though, it seems like the `WorkflowRun` was simply misused and that's what caused the error. I'd considered that a bug. I don't think `WorkflowStorage` should implement [such logic](https://github.com/mailpoet/mailpoet/pull/4211/files#diff-0a60b8fe1c407819c5edac9abbec31c2c63fd7cb8b1ac5f17319ec9f27d2a42cR54), it was meant to be a low-level data storage access class, and that's it. Injecting registry inside kind of seems to break the separation of concerns and create sort of a dependency cycle.

We already have the subjects registered as services (although I know that now via triggers only) so the only thing we should need now is to get the correct subject and call `load` with args. Registering them all again in a subject factory seems a bit duplicative to me. I'll have a look if the approach I'm describing can be done.

That said, We should probably consider allowing also other approaches to subjects than services in the future. I already had this in the [random ideas checklist](https://wordpress.com/post/mailpoetdev.wordpress.com/1276), mentioning `Can subjects be stateless? Are we able to provide a simple-enough factory-like API?`.

JanJakes:
@websupporter In the end I decided to keep the subject loading in the workflow run storage, for now, it's good for simplicity. Otherwise I just went for a simpler solution and added subjects to the registry. See https://github.com/mailpoet/mailpoet/pull/4220.

PR #4214, opened 2022-06-27
Add tagging subscribers on the edit page [MAILPOET-4440]

lysyjan:
This is the first pull request to add a new tag management feature.

**How to test**
1. Create/Edit a subscriber in the administration
2. Add/Remove tags
3. Check that tags are stored properly

Note: This PR contains a DB Migration. You need to run it before testing (e.g. deactivate and activate the plugin).

[MAILPOET-4440]

[MAILPOET-4440]: https://mailpoet.atlassian.net/browse/MAILPOET-4440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-plugin/_forms.scss
+++ b/mailpoet/assets/css/src/components-plugin/_forms.scss
@@ -219,6 +219,7 @@ progress::-moz-progress-bar {
   }
 }
 
+// We need to hide the label because it doesn't fit to our form design, and it's added automatically by Gutenberg component
 .mailpoet-form-field-tags label.components-form-token-field__label {
   display: none;
 }
--- a/mailpoet/assets/css/src/components-plugin/_forms.scss
+++ b/mailpoet/assets/css/src/components-plugin/_forms.scss
@@ -218,3 +218,7 @@ progress::-moz-progress-bar {
     font-style: italic;
   }
 }
+
+.mailpoet-form-field-tags label.components-form-token-field__label {
+  display: none;
+}
--- a/mailpoet/assets/css/src/mailpoet-admin.scss
+++ b/mailpoet/assets/css/src/mailpoet-admin.scss
@@ -3,3 +3,8 @@
 
 @import 'components-admin/menu';
 @import 'components-admin/3rd-party-plugins/members';
+
+// Generic
+// 3rd-party styles, can be overwritten.
+
+@import '../../../node_modules/@wordpress/components/build-style/style';
--- a/mailpoet/assets/css/src/mailpoet-admin.scss
+++ b/mailpoet/assets/css/src/mailpoet-admin.scss
@@ -3,8 +3,3 @@
 
 @import 'components-admin/menu';
 @import 'components-admin/3rd-party-plugins/members';
-
-// Generic
-// 3rd-party styles, can be overwritten.
-
-@import '../../../node_modules/@wordpress/components/build-style/style';
--- a/mailpoet/lib/AdminPages/Pages/Subscribers.php
+++ b/mailpoet/lib/AdminPages/Pages/Subscribers.php
@@ -76,9 +76,6 @@ class Subscribers {
   }
 
   public function render() {
-    // Gutenberg components styles
-    $this->wp->wpEnqueueStyle('wp-components');
-
     $installer = new Installer(Installer::PREMIUM_PLUGIN_SLUG);
     $pluginInformation = $installer->retrievePluginInformation();
 
--- a/mailpoet/lib/AdminPages/Pages/Subscribers.php
+++ b/mailpoet/lib/AdminPages/Pages/Subscribers.php
@@ -69,6 +69,9 @@ class Subscribers {
   }
 
   public function render() {
+    // Gutenberg components styles
+    $this->wp->wpEnqueueStyle('wp-components');
+
     $installer = new Installer(Installer::PREMIUM_PLUGIN_SLUG);
     $pluginInformation = $installer->retrievePluginInformation();
 
--- a/mailpoet/lib/AdminPages/Pages/Subscribers.php
+++ b/mailpoet/lib/AdminPages/Pages/Subscribers.php
@@ -6,6 +6,7 @@ use MailPoet\AdminPages\PageRenderer;
 use MailPoet\Cache\TransientCache;
 use MailPoet\Config\Installer;
 use MailPoet\Config\ServicesChecker;
+use MailPoet\Entities\TagEntity;
 use MailPoet\Form\Block;
 use MailPoet\Listing\PageLimit;
 use MailPoet\Models\CustomField;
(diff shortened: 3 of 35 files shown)


mailpoet/assets/css/src/components-plugin/_forms.scss
```diff
@@ -218,3 +218,7 @@ progress::-moz-progress-bar {
     font-style: italic;
   }
 }
+
+.mailpoet-form-field-tags label.components-form-token-field__label {
```
costasovo, line 223: I wonder why do we need to hide the label? Maybe this could use some code comment.

mailpoet/lib/AdminPages/Pages/Subscribers.php
```diff
@@ -69,6 +69,9 @@ public function __construct(
   }
 
   public function render() {
+    // Gutenberg components styles
+    $this->wp->wpEnqueueStyle('wp-components');
+
```
costasovo: I think we can't simply use styles this way. Currently, we bundle components from npm packages so when we use styles from WP we may get different results for different WP versions and also when the Gutenberg plugin is active.

Here is what happened when I the Gutenberg plugin:
Screenshot 2022-06-29 at 11 08 45

I think we have 2 options:
1) load both wp components and CSS from WordPress
2) bundle both from npm packages

mailpoet/lib/Subscribers/SubscriberSaveController.php
```diff
@@ -277,4 +294,27 @@ public function updateCustomFields(array $data, SubscriberEntity $subscriber): v
       $this->subscriberCustomFieldRepository->createOrUpdate($subscriber, $customField, $customFieldsMap[$customField->getId()]);
     }
   }
+
+  private function updateTags(array $data, SubscriberEntity $subscriber): void {
+    foreach ($subscriber->getSubscriberTags() as $subscriberTag) {
+      $tag = $subscriberTag->getTag();
+      if (!$tag) {
+        continue;
```
costasovo: If I'm not mistaken this happens in case the tag entity was deleted but the association is still saved in DB. Am I right?

If so. I think we could cleanup the association by calling `$subscriber->getSubscriberTags()->removeElement($subscriberTag);` instead of `continue`. What do you think?

mailpoet/tests/integration/Subscribers/SubscriberSaveControllerTest.php
```diff
@@ -75,6 +80,9 @@ public function testItCanUpdateASubscriber(): void {
       'segments' => [
         $segmentOne->getId(),
       ],
+      'tags' => [
```
costasovo, line 84: Just a suggestion. I think we could also test a case when we add and remove tags at the same time. We had some issues with similar cases in the past so I think it would be nice if the test protect us from such issues, which we may potentially introduce in some future refactors.

costasovo (changes requested):
Great job! It is nice to see we are adding this functionality!

I've found an issue with CSS when testing with the Gutenberg plugin active and added a few suggestions.

mailpoet/lib/Subscribers/SubscriberSaveController.php
lysyjan: >If I'm not mistaken this happens in case the tag entity was deleted but the association is still saved in DB. Am I right?

I think soo.

>If so. I think we could cleanup the association by calling $subscriber->getSubscriberTags()->removeElement($subscriberTag); instead of continue.

I agree with your proposal. I will change it.

mailpoet/assets/css/src/components-plugin/_forms.scss
lysyjan: I haven't found how to disable the label in the Gutenberg component, but we need to hide it due to the different form design. The label doesn't fit our form, and we would have duplicity here.

mailpoet/tests/integration/Subscribers/SubscriberSaveControllerTest.php
lysyjan: I considered it, but we don't test this behavior on segments, so I didn't do it. If it makes sense to you, I can extend it.

mailpoet/lib/AdminPages/Pages/Subscribers.php
lysyjan: Nice catch! I will try to solve it. Thanks for your proposals.

mailpoet/assets/css/src/components-plugin/_forms.scss
costasovo: Aha. Thanks! That makes sense. Could you please add a similar comment also to the code? Maybe the component will change and then we will know why this was done and it will be safe to remove it.

lysyjan:
@costasovo I addressed all your comments. I would like to do the interactive rebase after the successful review. Feel free to ping me with other questions.

mailpoet/assets/css/src/mailpoet-admin.scss
```diff
@@ -3,3 +3,19 @@
...
+@import '../../../node_modules/@wordpress/components/build-style/style';
+@import '../../../node_modules/@wordpress/edit-post/build-style/style';
+@import '../../../node_modules/@wordpress/edit-post/build-style/classic';
+@import '../../../node_modules/@wordpress/block-editor/build-style/style';
+@import '../../../node_modules/@wordpress/block-library/build-style/style';
+@import '../../../node_modules/@wordpress/block-library/build-style/theme';
+@import '../../../node_modules/@wordpress/block-library/build-style/editor';
+@import '../../../node_modules/@wordpress/format-library/build-style/style';
+@import '../../../node_modules/codemirror/lib/codemirror';
+@import '../../../node_modules/codemirror/theme/neo';
```
costasovo: Here we are moving all styles coming from WordPress packages to the main MailPoet admin CSS file. We use only a single component in the subscriber's form. Could we move only the @wordpress/components styles instead of moving all of them?

costasovo (dismissed):
LGTM! 👍

veljkho:
Everything's good, except one tiny thing:
- When I try to add multiple same tags but different casing like TEst, Test, TEST, test.. it will let me enter it but when saving will save the first one TEst only. This is okay to save only one, but not sure about the behavior of letting user add multiple same tags? SS https://d.pr/i/rYIkwU @lysyjan

lysyjan:
@veljkho Nice catch! I think that multiple tags should be created.

veljkho (changes requested):
Requesting changes as agreed above. Thanks!

costasovo (approved):
LGTM

veljkho (approved):

PR #4216, opened 2022-06-30
Feat/display a warning when using an unauthorized email address - MAILPOET-4245

triple0t:
[MAILPOET-4245](https://mailpoet.atlassian.net/browse/MAILPOET-4245)

**Testing**:

* Activate MSS sending method
* Visit Newsletter sending page
* If your sender's email address is authorized, you shouldn't get any message otherwise you would receive a message to authorized your email address
* Update your sender email address, check the message.
* Test with both authorized and unauthorized email addresses

Screenshot 2022-07-01 at 3 25 33 PM

### DIFF
--- a/mailpoet/assets/js/src/common/sender_email_address_warning.jsx
+++ b/mailpoet/assets/js/src/common/sender_email_address_warning.jsx
@@ -5,8 +5,33 @@ import ReactStringReplace from 'react-string-replace';
 const userHostDomain = window.location.hostname.replace('www.', '');
 const suggestedEmailAddress = `contact@${userHostDomain}`;
 
-function SenderEmailAddressWarning({ emailAddress, mssActive }) {
-  if (mssActive) return null;
+function SenderEmailAddressWarning({
+  emailAddress,
+  mssActive,
+  isEmailAuthorized,
+}) {
+  if (mssActive) {
+    if (!isEmailAuthorized) {
+      return (
+        <span className="mailpoet_error_item mailpoet_error">
+          {ReactStringReplace(
+            MailPoet.I18n.t('youNeedToAuthorizeTheEmail'),
+            '[email]',
+            () => emailAddress,
+          )}{' '}
+          <a
+            className="mailpoet-link"
+            href={`https://account.mailpoet.com/authorization?email=${emailAddress}`}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            {MailPoet.I18n.t('authorizeMyEmail')}
+          </a>
+        </span>
+      );
+    }
+    return null;
+  }
   const emailAddressDomain = emailAddress.split('@').pop().toLowerCase();
   if (window.mailpoet_free_domains.indexOf(emailAddressDomain) > -1) {
     return (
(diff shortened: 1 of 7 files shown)


mailpoet/assets/js/src/common/sender_email_address_warning.jsx
```diff
@@ -5,8 +5,33 @@ import ReactStringReplace from 'react-string-replace';
...
+      return (
+        <span className="mailpoet_error_item mailpoet_error">
+          {ReactStringReplace(
+            MailPoet.I18n.t('youNeedToAuthorizeTheEmail'),
+            '[email]',
+            () => emailAddress,
+          )}{' '}
+          <a
+            className="mailpoet-link"
+            href={`https://account.mailpoet.com/authorization?email=${emailAddress}`}
```
lysyjan, line 24: We have [a URL factory](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/mailpoet_com_url_factory.js) for similar purposes. Maybe we can add a method for this new URL type. What do you think about that?

mailpoet/assets/js/src/newsletters/send/sender_address_field.jsx
```diff
@@ -49,7 +49,7 @@ class SenderField extends Component {
           field={this.props.field}
           onValueChange={this.onChange}
         />
-        <div className="regular-text">
+        <div className="regular-text" style={{ marginTop: '2.5rem' }}>
```
lysyjan, line 62: Is this margin necessary? It looks better without it to me.

triple0t: Yeah, but that's to fix this issue https://user-images.githubusercontent.com/30554163/176655015-777259ac-5748-48b0-8f27-41befac015a9.png

mailpoet/assets/js/src/common/sender_email_address_warning.jsx
triple0t: That's definitely a very good suggestion. 
We will be changing the behavior of that link in this ticket https://mailpoet.atlassian.net/browse/MAILPOET-4300
We won't require linking to the Shop anymore 
So I thought to just keep it as it is for now

triple0t:
Hi @lysyjan 

Thanks for the review.

Please check my comments [here](https://github.com/mailpoet/mailpoet/pull/4216#discussion_r911789593) and [here](https://github.com/mailpoet/mailpoet/pull/4216#discussion_r911785225) 

Do let me know if I should make any more updates.

lysyjan:
@triple0t Thank you for the quick reactions. I will move the PR forward. Please update the image in the description because it's outdated. You can avoid future questions from QA.

lysyjan (dismissed):

veljkho (changes requested):
It is good, but one thing to ask.. Can we make it the same as it works in MailPoet > Settings > Basic tab? The difference is that when I start writing an email it immediately tells me that the email is not authorized.. until the email is fully written then the message disappears (in case it's authorized email).
I would expect when I finish writing email and then changing the cursor outside of the field to see if it is authorized or not.. that's how it works in the settings now.

triple0t:
> It is good, but one thing to ask.. Can we make it the same as it works in MailPoet > Settings > Basic tab? The difference is that when I start writing an email it immediately tells me that the email is not authorized.. until the email is fully written then the message disappears (in case it's authorized email).
> I would expect when I finish writing email and then changing the cursor outside of the field to see if it is authorized or not.. that's how it works in the settings now.

Thanks for the review and suggestions @veljkho 

That component does not offer that feature, but I have updated that in https://github.com/mailpoet/mailpoet/pull/4216/commits/77d6a3ad657535293fb7a41494baa07d641df61a

veljkho (approved):

PR #4218, opened 2022-07-01
Don't schedule next step if it's null [MAILPOET-4450]

alex-mailpoet:
[MAILPOET-4450]

[MAILPOET-4450]: https://mailpoet.atlassian.net/browse/MAILPOET-4450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Control/StepRunner.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepRunner.php
@@ -116,6 +116,11 @@ class StepRunner {
       return;
     }
 
+    // no need to schedule a new step if the next step is null
+    if (empty($nextStepArgs[0]['step_id'])) {
+      return;
+    }
+
     // enqueue next step
     $this->actionScheduler->enqueue(Hooks::WORKFLOW_STEP, $nextStepArgs);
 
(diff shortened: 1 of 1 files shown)


mailpoet/lib/Automation/Engine/Control/StepRunner.php
```diff
@@ -116,6 +116,11 @@ private function runStep(array $args): void {
       return;
     }
 
+    // no need to schedule a new step if the next step is null
+    if (empty($nextStepArgs[0]['step_id'])) {
```
websupporter, line 120: I think you need to set the status of the workflowrun to `WorkflowRun::STATUS_COMPLETE` which currently happens in L91 when no stepId was found.

alex-mailpoet: Good catch! I've added completion of the workflow run here.

websupporter (approved):

websupporter:
Thank you @alex-mailpoet. LGTM

JanJakes:
I think this one is also good to merge.

PR #4219, opened 2022-07-01
Remove old model usage from MailPoet\API\MP\v1::getLists() [MAILPOET-4292]

lysyjan:
[MAILPOET-4292]

[MAILPOET-4292]: https://mailpoet.atlassian.net/browse/MAILPOET-4292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/MP/v1/API.php
+++ b/mailpoet/lib/API/MP/v1/API.php
@@ -24,6 +24,9 @@ class API {
   /** @var CustomFields */
   private $customFields;
 
+  /** @var Segments */
+  private $segments;
+
   /** @var Subscribers */
   private $subscribers;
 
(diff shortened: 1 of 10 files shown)


mailpoet/lib/API/MP/v1/API.php
```diff
@@ -111,9 +119,21 @@ public function unsubscribeFromLists($subscriberId, array $listIds) {
     return $subscriber->withCustomFields()->withSubscriptions()->asArray();
   }
 
-  public function getLists() {
-    return Segment::where('type', Segment::TYPE_DEFAULT)
-      ->findArray();
+  public function getLists(): array {
+    $segments = $this->segmentsRepository->findBy(['type' => SegmentEntity::TYPE_DEFAULT]);
```
JanJakes: I guess this wasn't there previously either but should we add a default order here?

lysyjan: Good point! I will add it. Thanks

JanJakes (dismissed):

lysyjan:
After discussing with @pavel-mailpoet, we decided to move the main logic for this endpoint to the new class. I'm going to change it.

JanJakes (approved):

veljkho (approved):

PR #4220, opened 2022-07-01
Automation demo [MAILPOET-4454]

JanJakes:
This PR contains various (unrelated) automation fixes and improvements. See the commit titles. They were done in preparation for the early-stage automation demo.

**Testing:** No need to test yet, still pretty much prototyping.

[MAILPOET-4454]
[MAILPOET-4449]

Closes https://github.com/mailpoet/mailpoet/pull/4211.

[MAILPOET-4454]: https://mailpoet.atlassian.net/browse/MAILPOET-4454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4449]: https://mailpoet.atlassian.net/browse/MAILPOET-4449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
@@ -3,6 +3,7 @@
 namespace MailPoet\Automation\Engine\Builder;
 
 use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Exceptions\UnexpectedValueException;
 use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Automation\Engine\Workflows\Workflow;
 
@@ -27,6 +28,7 @@ class UpdateWorkflowController {
     }
 
     if (array_key_exists('status', $data)) {
+      $this->checkWorkflowStatus($data['status']);
       $workflow->setStatus($data['status']);
       $this->storage->updateWorkflow($workflow);
     }
@@ -36,5 +38,11 @@ class UpdateWorkflowController {
       throw Exceptions::workflowNotFound($id);
     }
     return $workflow;
+  }
+
+  private function checkWorkflowStatus(string $status): void {
+    if (!in_array($status, [Workflow::STATUS_ACTIVE, Workflow::STATUS_INACTIVE, Workflow::STATUS_DRAFT], true)) {
+      throw UnexpectedValueException::create()->withMessage(__(sprintf('Invalid status: %s', $status), 'mailpoet'));
+    }
   }
 }
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
@@ -1,0 +1,40 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Builder;
+
+use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Storage\WorkflowStorage;
+use MailPoet\Automation\Engine\Workflows\Workflow;
+
+class UpdateWorkflowController {
+  /** @var WorkflowStorage */
+  private $storage;
+
+  public function __construct(
+    WorkflowStorage $storage
+  ) {
+    $this->storage = $storage;
+  }
+
+  public function updateWorkflow(int $id, array $data): Workflow {
+    // TODO: data & workflow validation (trigger existence, graph consistency, etc.)
+    // TODO: new revisions when content is changed
+    // TODO: validation when status being is changed
+
+    $workflow = $this->storage->getWorkflow($id);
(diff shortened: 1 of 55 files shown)


mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
```diff
@@ -0,0 +1,40 @@
...
+    // TODO: new revisions when content is changed
+    // TODO: validation when status being is changed
+
+    $workflow = $this->storage->getWorkflow($id);
+    if (!$workflow) {
+      throw Exceptions::workflowNotFound($id);
+    }
+
+    if (array_key_exists('status', $data)) {
+      $workflow->setStatus($data['status']);
```
websupporter, line 32: I assume `status` is an enum and would get validated in the `ToDo` section above?

websupporter:
Thanks for this PR @JanJakes. One question, see above. Otherwise lgtm.

One question I have, but this PR might not be the right place. The current implementation of `Subject` allows only for one distinct Subject of the same type.

I could imagine use cases where we might want to have several subjects of the same type in the same run. At least from a general perspective.

Although we do not intend/need it now, but I could imagine e.g. a WorkflowRun with two `SubscriberSubject`s or `RecipientSubject`s or something similar. Afaik, the current architecture does not allow for that. What do you think?

websupporter (dismissed):

mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
JanJakes: It gets somewhat validated [here](https://github.com/mailpoet/mailpoet/pull/4220/files/48599edd5c8a0e021d20ce1d9cc12a7ee754e85b#diff-7d33db3284a65610c0bcd8679866b78534f97a059905da0df3a2373db24dafe0R35) but we don't have enums yet. Do you think we should validate it also inside the controller to be on the safe side? 🤔

I was also considering whether we should make the status change a dedicated endpoint (like `/workflow/<id>/status`) but looking at WP posts they don't do such a thing so I kept it this way for now.

JanJakes:
> Thanks for this PR @JanJakes. One question, see above. Otherwise lgtm.
> 
> One question I have, but this PR might not be the right place. The current implementation of `Subject` allows only for one distinct Subject of the same type.
> 
> I could imagine use cases where we might want to have several subjects of the same type in the same run. At least from a general perspective.
> 
> Although we do not intend/need it now, but I could imagine e.g. a WorkflowRun with two `SubscriberSubject`s or `RecipientSubject`s or something similar. Afaik, the current architecture does not allow for that. What do you think?

That's a great point, @websupporter, maybe we should make it an array, although then I'm not sure how the actions should approach that. Maybe a "send email" action can just throw an error if there are multiple subscribers and only one was expected. I added this [here](https://wordpress.com/post/mailpoetdev.wordpress.com/1276).

mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
websupporter: > Whether we should make the status change a dedicated endpoint

I think this is not needed. The general entity endpoint would want to be able to `PUT status:active` anyway imo (in case you later change the UI to a activate toggle and a save button, which would save all changes). To have a second endpoint for just the status seems not necessary for me. Besides what would be the payload? only `"active"|"inactive"`?

Since we do not have `enums` (though, I would like that) I think a simple
```
$workflow->setStatus($data['status'] === 'active' ? 'active':'inactive');
```
would be nice.

JanJakes: @websupporter [Enums are available from WP 5.7](https://github.com/WordPress/WordPress/blob/e70ac7004c8773249ec47fc56a73953043a6a877/wp-includes/rest-api.php#L1981) so I guess we may as well add them soon and use them in for automation and once we release the MVP, we'd have to bump minimum WP version.

As for this PR, I can just add a check that an existing value is being set for now.

websupporter: That sounds good. Thank you!

JanJakes: I added this: https://github.com/mailpoet/mailpoet/commit/cf752545d339de14804252603661de9f3f53d6b7

websupporter: Thank you!

websupporter (approved):

JanJakes:
Merging this as well as it should affect automations only.

PR #4222, opened 2022-07-04
Display tags on the subscribers listing page [MAILPOET-4441] 

lysyjan:
This pull request is the second in a row for the tags feature. The review should start from [this commit.](https://github.com/mailpoet/mailpoet/commit/584e225f95df0617916e7a57c4c9801a62f4498d)

For QA: This PR has to merge after [the first one](https://github.com/mailpoet/mailpoet/pull/4214) and should be rebased.

[MAILPOET-4441] 

[MAILPOET-4441]: https://mailpoet.atlassian.net/browse/MAILPOET-4441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/common/tag/_stories/tags.tsx
+++ b/mailpoet/assets/js/src/common/tag/_stories/tags.tsx
@@ -11,15 +11,22 @@ const segments = [
   { name: 'Unconfirmed' },
 ];
 
+const strings = ['First', 'Second', 'Third tag'];
+
 export function Segments() {
   return (
     <>
-      <Heading level={1}>Segments</Heading>
+      <Heading level={1}>Tags</Heading>
+      <Heading level={2}>Segments</Heading>
       <Tags segments={segments} dimension="large" />
       <div className="mailpoet-gap" />
       <Tags segments={segments}>
         <span>Prefix: </span>
       </Tags>
+      <Heading level={2}>Strings</Heading>
+      <Tags strings={strings} dimension="large" variant="good" />
+      <div className="mailpoet-gap" />
+      <Tags strings={strings} dimension="large" variant="good" isInverted />
     </>
   );
 }
--- a/mailpoet/assets/js/src/common/tag/tags.tsx
+++ b/mailpoet/assets/js/src/common/tag/tags.tsx
@@ -13,9 +13,18 @@ type Props = {
   dimension?: 'large';
   segments?: Segment[];
   strings?: string[];
+  variant?: 'average' | 'good' | 'excellent' | 'list' | 'unknown' | 'wordpress';
+  isInverted?: boolean;
 };
 
-function Tags({ children, dimension, segments, strings }: Props) {
+function Tags({
+  children,
+  dimension,
+  segments,
+  strings,
+  variant,
+  isInverted,
+}: Props) {
   return (
     <div className="mailpoet-tags">
       {children}
@@ -52,7 +61,12 @@ function Tags({ children, dimension, segments, strings }: Props) {
         })}
       {strings &&
         strings.map((string) => (
-          <Tag key={string} dimension={dimension} variant="list">
+          <Tag
+            key={string}
+            dimension={dimension}
+            variant={variant || 'list'} // due to backward compatibility we use `list` as the default value
+            isInverted={isInverted}
+          >
             {string}
           </Tag>
         ))}
--- a/mailpoet/assets/js/src/subscribers/list.jsx
+++ b/mailpoet/assets/js/src/subscribers/list.jsx
@@ -450,12 +450,7 @@ function SubscriberList({ match }) {
           <Tags segments={subscribedSegments} dimension="large" />
         </td>
         <td className="column" data-colname={MailPoet.I18n.t('tags')}>
-          <Tags
-            strings={subscriberTags}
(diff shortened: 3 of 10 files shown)


mailpoet/assets/js/src/common/tag/_stories/tags.tsx
```diff
@@ -11,15 +11,22 @@ const segments = [
   { name: 'Unconfirmed' },
 ];
 
+const strings = ['One', 'Two', 'Third tag'];
```
triple0t: Was this is intentional?

Why not 'First', 'Second', 'Third'? 
or
'One', 'Two', 'Three'?

mailpoet/assets/js/src/common/tag/tags.tsx
```diff
@@ -52,7 +61,12 @@ function Tags({ children, dimension, segments, strings }: Props) {
         })}
       {strings &&
         strings.map((string) => (
-          <Tag key={string} dimension={dimension} variant="list">
+          <Tag
+            key={string}
+            dimension={dimension}
+            variant={variant === undefined ? 'list' : variant} // due to backward compatibility we use `list` as the default value
```
triple0t: Hmmm
A question here, why are we specifically checking for `undefined`?

if we need to account for backward compatibility
why not
```
variant={variant ? variant : 'list' }
```

This would check for both undefined, null, and other falsely values

mailpoet/assets/js/src/subscribers/list.jsx
```diff
@@ -412,6 +417,13 @@ function SubscriberList({ match }) {
       });
     }
 
+    // Tags
+    if (subscriber.tags.length > 0) {
```
triple0t, line 421: Note here

I couldn’t locate the initial value for `subscriber.tags`.
Are we sure it would allow be an Array? even if it's an empty array `[]`?

An alternative would be checking for the presence of `subscriber.tags` and then `subscriber.tags.length > 0`

e.g.
```
if (subscriber.tags && subscriber.tags.length > 0) {
```

triple0t (commented):
Looks good.
Just some comments.

Kindly locate below

mailpoet/assets/js/src/common/tag/_stories/tags.tsx
lysyjan: I want a tag that contains a space in the name.

mailpoet/assets/js/src/common/tag/tags.tsx
lysyjan: That will work too. I will change it. Thank you

mailpoet/assets/js/src/common/tag/_stories/tags.tsx
triple0t: Noted.
In that case, can we use: 'First', 'Second', 'Third tag'?


Screenshot 2022-07-07 at 3 00 28 PM

mailpoet/assets/js/src/subscribers/list.jsx
lysyjan: The initial value gets from the Subscribers::listing endpoint, and I added tags in [this commit](https://github.com/mailpoet/mailpoet/pull/4222/commits/584e225f95df0617916e7a57c4c9801a62f4498d). As you can see tags are always an array.

mailpoet/assets/js/src/common/tag/_stories/tags.tsx
lysyjan: Sure, no problem.

mailpoet/assets/js/src/subscribers/list.jsx
triple0t: Noted.
Thank you for highlighting that.

lysyjan:
@triple0t I have addressed both fixes in [the last commit](https://github.com/mailpoet/mailpoet/pull/4222/commits/eb979caf1d2e930759d89f204ab3eab2ddb5a25b). It's a fixup commit because I assume that this branch will be rebased.

triple0t (dismissed):
LGTM 🎉

veljkho:
Hi @lysyjan 
I see some conflicting files.. probably after merging the 1st PR.

lysyjan:
@veljkho This was a little bit expected.  I am going to rebase it.

veljkho (changes requested):
Everything's fine, good job!
I have a suggestion to [make those tags smaller](https://d.pr/i/6n1BQ5).. they are currently so big to me, looking even bigger than list badges.. Also, tags are less important than lists, so I would make them smaller in size here in the listing. What do you think @lysyjan CC @NeosinneR

lysyjan:
@veljkho Thanks for pointing this out. I can use the same size as is in the score column. Do you agree with me @pavel-mailpoet?

pavel-mailpoet:
@lysyjan Sure, let's experiment with them. I have no preference. We can change them now or later if we want. Also, the colour will be probably different in the future (user editable)

triple0t (approved):

veljkho (approved):

PR #4227, opened 2022-07-06
Automation refinements [MAILPOET-4465]

JanJakes:
Connecting pieces of automations together, some fixes, and refinements.

Subjects/Fields will need further care, see https://a8c.slack.com/archives/C0313447YBC/p1657101012947749.

[MAILPOET-4465]

[MAILPOET-4465]: https://mailpoet.atlassian.net/browse/MAILPOET-4465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
@@ -12,7 +12,7 @@ function getIcon(step: StepType): JSX.Element | null {
     return <TriggerIcon />;
   }
 
-  if (step.key === 'core:wait') {
+  if (step.key === 'core:delay') {
     return <DelayIcon />;
   }
 
@@ -29,7 +29,7 @@ function getTitle(step: StepType): string {
     return 'Trigger';
   }
 
-  if (step.key === 'core:wait') {
+  if (step.key === 'core:delay') {
     return 'Delay';
   }
 
@@ -45,7 +45,7 @@ function getSubtitle(step: StepType): string {
   if (step.key === 'mailpoet:segment:subscribed') {
     return 'Subscribed to segment';
   }
-  if (step.key === 'core:wait') {
+  if (step.key === 'core:delay') {
     return `${step.args.seconds as number} seconds`;
   }
   if (step.key === 'mailpoet:send-welcome-email') {
--- a/mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
@@ -16,7 +16,7 @@ function getIcon(step: StepType): JSX.Element | null {
     return <DelayIcon />;
   }
 
-  if (step.key === 'mailpoet:send-welcome-email') {
+  if (step.key === 'mailpoet:send-email') {
     return <EmailIcon />;
   }
 
@@ -33,8 +33,8 @@ function getTitle(step: StepType): string {
     return 'Delay';
   }
 
-  if (step.key === 'mailpoet:send-welcome-email') {
-    return 'Send welcome email';
+  if (step.key === 'mailpoet:send-email') {
+    return 'Send email';
   }
 
   return '';
@@ -48,8 +48,8 @@ function getSubtitle(step: StepType): string {
   if (step.key === 'core:delay') {
     return `${step.args.seconds as number} seconds`;
   }
-  if (step.key === 'mailpoet:send-welcome-email') {
-    return `Email ID: ${step.args.welcomeEmailId as number}`;
+  if (step.key === 'mailpoet:send-email') {
+    return `Email ID: ${step.args.email_id as number}`;
   }
   return step.key;
 }
--- a/mailpoet/lib/AdminPages/Pages/AutomationEditor.php
+++ b/mailpoet/lib/AdminPages/Pages/AutomationEditor.php
@@ -37,7 +37,7 @@ class AutomationEditor {
     if (!$workflow) {
       $notice = new WPNotice(
         WPNotice::TYPE_ERROR,
-        __("Workflow not found.", 'mailpoet')
+        __('Workflow not found.', 'mailpoet')
       );
       $notice->displayWPNotice();
       $this->pageRenderer->displayPage('blank.html');
--- a/mailpoet/lib/AdminPages/Pages/AutomationEditor.php
+++ b/mailpoet/lib/AdminPages/Pages/AutomationEditor.php
@@ -8,6 +8,7 @@ use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Automation\Engine\Workflows\Step;
 use MailPoet\Automation\Engine\Workflows\Workflow;
 use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice as WPNotice;
 
 class AutomationEditor {
   /** @var WorkflowStorage */
@@ -33,8 +34,18 @@ class AutomationEditor {
     $id = isset($_GET['id']) ? (int)$_GET['id'] : null;
 
     $workflow = $id ? $this->workflowStorage->getWorkflow($id) : null;
+    if (!$workflow) {
+      $notice = new WPNotice(
+        WPNotice::TYPE_ERROR,
+        __("Workflow not found.", 'mailpoet')
+      );
+      $notice->displayWPNotice();
+      $this->pageRenderer->displayPage('blank.html');
+      return;
+    }
+
     $this->pageRenderer->displayPage('automation/editor.html', [
-      'workflow' => $workflow ? $this->buildWorkflow($workflow) : null,
+      'workflow' => $this->buildWorkflow($workflow),
       'sub_menu' => 'mailpoet-automation',
       'api' => [
         'root' => rtrim($this->wp->escUrlRaw($this->wp->restUrl()), '/'),
--- a/mailpoet/lib/Automation/Engine/Engine.php
+++ b/mailpoet/lib/Automation/Engine/Engine.php
@@ -14,6 +14,8 @@ use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Automation\Integrations\Core\CoreIntegration;
 
 class Engine {
+  const CAPABILITY_MANAGE_AUTOMATIONS = 'mailpoet_manage_automations';
+
   /** @var API */
   private $api;
 
--- a/mailpoet/lib/Automation/Engine/Migrations/Migrator.php
+++ b/mailpoet/lib/Automation/Engine/Migrations/Migrator.php
@@ -14,11 +14,13 @@ class Migrator {
 
   public function __construct() {
     global $wpdb;
-    $this->prefix = $wpdb->prefix . 'mailpoet_automation_';
+    $this->prefix = $wpdb->prefix . 'mailpoet_';
     $this->wpdb = $wpdb;
   }
 
   public function createSchema(): void {
+    $this->removeOldSchema();
+
     $this->runQuery("
       CREATE TABLE {$this->prefix}workflows (
         id int(11) unsigned NOT NULL AUTO_INCREMENT,
@@ -48,6 +50,7 @@ class Migrator {
   }
 
   public function deleteSchema(): void {
+    $this->removeOldSchema();
     $this->runQuery("DROP TABLE IF EXISTS {$this->prefix}workflows");
     $this->runQuery("DROP TABLE IF EXISTS {$this->prefix}workflow_runs");
 
@@ -67,8 +70,14 @@ class Migrator {
   }
 
   public function hasSchema(): bool {
-    $pattern = $this->wpdb->esc_like($this->prefix) . '%';
+    $pattern = $this->wpdb->esc_like("{$this->prefix}workflows") . '%';
     return $this->runQuery("SHOW TABLES LIKE '$pattern'") > 0;
+  }
+
+  private function removeOldSchema(): void {
+    $oldPrefix = $this->wpdb->prefix . 'mailpoet_automation_';
+    $this->runQuery("DROP TABLE IF EXISTS {$oldPrefix}workflows");
+    $this->runQuery("DROP TABLE IF EXISTS {$oldPrefix}workflow_runs");
   }
 
   private function runQuery(string $query): int {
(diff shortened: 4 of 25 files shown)


mailpoet/lib/Automation/Engine/Migrations/Migrator.php
```diff
@@ -14,11 +14,13 @@ class Migrator {
 
   public function __construct() {
     global $wpdb;
-    $this->prefix = $wpdb->prefix . 'mailpoet_automation_';
+    $this->prefix = $wpdb->prefix . 'mailpoet_';
```
johnolek, line 17: I just now discovered that we're allowing users to override the main plugin prefix with a filter (see https://mailpoet.atlassian.net/browse/MAILPOET-3798 and https://github.com/mailpoet/mailpoet/blob/7c690dfa07503b0d628c9b3cb0025411ee572563/mailpoet/lib/Config/Env.php#L61-L61). 

Should we use the same potentially-filtered value here and anywhere else we're defining table names?

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
```diff
@@ -14,11 +14,13 @@ class Migrator {
...
     $this->wpdb = $wpdb;
   }
 
   public function createSchema(): void {
+    $this->removeOldSchema();
```
johnolek, line 22: Will we want to remove this before we release MVP, or do we even need it at all since it should only affect people who have turned on the feature flag?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -93,33 +94,26 @@ public function run(Workflow $workflow, WorkflowRun $workflowRun, Step $step): v
...
+  private function getEmailForStep(Step $step): NewsletterEntity {
+    $emailId = $step->getArgs()['email_id'] ?? null;
+    if (!$emailId) {
       throw InvalidStateException::create();
     }
-    $newsletter = $this->newslettersRepository->findOneById($welcomeEmailId);
-    if ($newsletter === null) {
-      throw InvalidStateException::create()->withMessage(sprintf("Welcome Email with ID '%s' not found.", $welcomeEmailId));
-    }
-    $type = $newsletter->getType();
```
johnolek, line 132: Just commenting to say that I'm really happy to see this code going away. I didn't feel very good about it when I wrote it.

mailpoet/lib/Automation/Engine/Engine.php
```diff
@@ -14,6 +14,8 @@
 use MailPoet\Automation\Integrations\Core\CoreIntegration;
 
 class Engine {
+  const CAPABILITY_MANAGE_AUTOMATIONS = 'mailpoet_manage_automations';
```
johnolek, line 17: I'm not saying I disagree with extracting this permission, but I'm not sure I understand the point, especially when we still reference `AccessControl::PERMISSION_MANAGE_AUTOMATIONS`. Could you please give me a brief explanation of why this is helpful?

I appreciate that you're using the more WordPress-specific "capability" instead of "permission" in the name. Might it be confusing to have our codebase use both "permission" and "capability" to refer to the same thing though?

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -8,6 +8,7 @@
 use MailPoet\Automation\Engine\Workflows\Step;
 use MailPoet\Automation\Engine\Workflows\Workflow;
 use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice as WPNotice;
```
johnolek, line 11: Looking through the codebase it seems like we're all over the place in terms of how we refer to `MailPoet\WP\Notice`. Sometimes we don't alias it, sometimes we call it `WPNotice`, and sometimes we just use the fully qualified name without an import.

Should using it as `WPNotice` be the standard? Maybe we could have a maintenance ticket for making all of the references to this (and `WPFunctions` or other `MailPoet\WP\` namespaced classes) consistent.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -33,8 +34,18 @@ public function render() {
     $id = isset($_GET['id']) ? (int)$_GET['id'] : null;
 
     $workflow = $id ? $this->workflowStorage->getWorkflow($id) : null;
+    if (!$workflow) {
+      $notice = new WPNotice(
+        WPNotice::TYPE_ERROR,
+        __("Workflow not found.", 'mailpoet')
```
johnolek: Could be single quotes

mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
```diff
@@ -12,7 +12,7 @@ function getIcon(step: StepType): JSX.Element | null {
     return <TriggerIcon />;
   }
 
-  if (step.key === 'core:wait') {
+  if (step.key === 'core:delay') {
```
johnolek, line 15: Maybe already in progress but this line in the premium plugin will need to be updated: https://github.com/mailpoet/mailpoet-premium/blob/d3fa0dbb04f5d232fdf8f7617bf731ed5eec9768/tests/unit/REST/Automation/Workflows/WorkflowsPostTest.php#L31-L31

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/SegmentSubscribedTrigger.php
```diff
@@ -41,8 +41,18 @@ public function handleSubscription(SubscriberSegmentEntity $subscriberSegment):
...
+          'segment_id' => $segment->getId(),
+        ],
+      ],
+      [
+        'key' => SubscriberSubject::KEY,
+        'args' => [
+          'subscriber_id' => $subscriber->getId(),
+        ],
+      ],
     ]);
```
johnolek, line 56: Would it make sense to give the subjects unique keys here, allowing us to give them more meaningful descriptors and so we don't need to reference them by numeric index in actions? It could get kind of confusing because then we would have two kinds of subject keys.

mailpoet/lib/Automation/Engine/Workflows/WorkflowRun.php
```diff
@@ -87,6 +88,17 @@ public function getSubjects(string $key = null): array {
     return $this->subjects;
   }
 
+  public function requireSubject(string $key): Subject {
```
johnolek: I don't know that it's 100% obvious from the method name that this would throw an exception if there were 2 subjects with the provided key. Maybe it could be `requireSingleSubject`?

I still think it could be useful to have triggers use unique keys/names for their subjects, so if you need to have 2 subscriber subjects they would be differentiated by their unique keys, but maybe there's a problem with that approach I'm not thinking about.

I suppose it might be difficult to enforce the uniqueness of those keys, and if another trigger used the same key for a subject but used a different type of subject, it might look like an action would be compatible that actually isn't.

mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
```diff
@@ -0,0 +1,51 @@
+<?php declare(strict_types=1);
+
+namespace MailPoet\Newsletter\Scheduler;
```
johnolek, line 3: Could it make sense to include this under `MailPoet\Automation` to keep all the automation stuff together? Normally I wouldn't have even said anything, but the fully qualified name is so close to the existing `\MailPoet\Newsletter\Scheduler\AutomaticEmailScheduler` that I had to do a double take. It also seems like we've made an effort to keep most of the other automation functionality under that `MailPoet\Automation` namespace, including extracting that permission/capability.

mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
```diff
@@ -0,0 +1,51 @@
...
+
+  public function __construct(
+    EntityManager $entityManager,
+    WPFunctions $wp
+  ) {
+    $this->entityManager = $entityManager;
+    $this->wp = $wp;
+  }
+
+  public function createSendingTask(NewsletterEntity $email, SubscriberEntity $subscriber): ScheduledTaskEntity {
```
johnolek, line 31: Perhaps we should verify the `type` of the `$email` first to ensure someone like me doesn't accidentally try to use this service to schedule an `automatic` email instead of an `automation` email.

johnolek (changes requested):
These changes look really great overall. Thanks especially for rewriting my sendWelcomeEmail code!

I know that the conversation about multiple subjects is ongoing, and I've read the slack thread, so feel free to ignore my comments related to that, and I don't expect you to make any changes to that code in this PR.

There are some minor things I would like to see changed, and I had a few questions that I would really appreciate your feedback on. Please let me know if anything is unclear!

P.S. Thank you very much for being so well organized with your commits. You made it very easy to review this commit by commit, which is something I would like to do more of but is only really possible with commits like the ones you made in this PR.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
JanJakes: Great catch! That didn't occur to me at all.

This migrator class is temporary, for the current testing behind the feature flag, I guess. I've just created [a ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4466) to implement production-ready migrations and added a note about this filter.

I've added a note about this filter also to a [ticket for WPDB wrapper](https://mailpoet.atlassian.net/browse/MAILPOET-4192).

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
JanJakes: Yeah, definitely. In fact, [I think we'll replace the whole migrator](https://github.com/mailpoet/mailpoet/pull/4227#discussion_r915822314).

mailpoet/lib/Automation/Engine/Engine.php
JanJakes: > I'm not saying I disagree with extracting this permission, but I'm not sure I understand the point, especially when we still reference AccessControl::PERMISSION_MANAGE_AUTOMATIONS. Could you please give me a brief explanation of why this is helpful?

This is done to avoid referencing `AccessControl` from the automation code. The other way round should be OK.

> I appreciate that you're using the more WordPress-specific "capability" instead of "permission" in the name. Might it be confusing to have our codebase use both "permission" and "capability" to refer to the same thing though?

I think that's why I did `const PERMISSION_MANAGE_AUTOMATIONS = Engine::CAPABILITY_MANAGE_AUTOMATIONS;` inside the `AccessControl` class. That way it will be used consistently in the MP codebase, while automations will hold a better name.

Does that make sense or should I better change it?

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
JanJakes: Good point, I actually just copy-pasted from the first place I encountered. I think we should just rename these classes and throw away the aliases (such as `Functions` to `WPFunctions`, for instance).

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
JanJakes: 👍  I'm surprised code sniffer is fine with that. Fixed.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/SegmentSubscribedTrigger.php
JanJakes: I was thinking about creating a `SubjectBag`/`SubjectList`/... service to provide better access to this but still didn't really think through how having two subjects of the same type should be handled and what are all the implications. I guess at this stage such service can simply enforce max. one subject per key and we can solve this later. [As we discussed.](https://a8c.slack.com/archives/C0313447YBC/p1657101012947749)

mailpoet/lib/Automation/Engine/Workflows/WorkflowRun.php
JanJakes: Makes sense! As discussed, this will all probably change once again but for now I'm renaming it.

mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
JanJakes: I'm not sure, to me, it seems that this belongs to MailPoet "core", to the `Scheduler` namespace indeed. In `MailPoet\Automation\Integrations\MailPoet` we have only actions, triggers, subjects, and stuff like that.

The naming is not great but ultimately, "automation" should replace "automatic". It will take some time, of course. Maybe I'll just add the check as mentioned in https://github.com/mailpoet/mailpoet/pull/4227#discussion_r915163888?

mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
JanJakes: Fixed!

mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
JanJakes: Good catch! This is not verified in any way yet so the test passes. Fixing.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
johnolek: Sounds great!

mailpoet/lib/Automation/Engine/Engine.php
johnolek: I think this should be fine as-is. It's probably too late to fully change usages of `permission` to `capability` in the free plugin, especially the strings themselves since users are probably relying on those in production sites, but having `capability` in the automation codebase is nice.

I think I'm still getting used to the level of separation we're shooting for with the automation codebase and the plugins.

mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
johnolek: I agree that this does feel right to have in the core plugin, not sure what I was thinking last week.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
johnolek: That sounds like a great idea to me! Sounds like a good maintenance ticket. Should we create one?

johnolek (approved):
Thanks for the discussion and the changes. This looks good to me now!

veljkho (approved):

PR #4229, opened 2022-07-06
Feat/let users authorise their email addresses in the plugin - MAILPOET-4300

triple0t:
[MAILPOET-4300](https://mailpoet.atlassian.net/browse/MAILPOET-4300)

Please start reviewing from https://github.com/mailpoet/mailpoet/commit/4198bade172d6214f7360149eb02d52163358710

Depends on https://github.com/mailpoet/mailpoet/pull/4216

Testing:
* Test anywhere with links to MP Shop Email authorization page e.g. `/authorization?email=NEW_EMAIL`
* Confirm you can add a new email
* Confirm you get the confirmation email
* Confirm after clicking on the confirmation email and waiting for some seconds, the modal window update with new messages
* Confirm you can use the new email for sending newsletters

### DIFF
--- a/mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
+++ b/mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
@@ -122,6 +122,8 @@ function AuthorizeSenderEmailModal({
       return null;
     }
 
+    clearCurrentInterval(setIntervalId.current);
+
     makeApiRequest(senderEmailAddress)
       .then((res) => {
         const response = Boolean(res?.data);
@@ -130,6 +132,16 @@ function AuthorizeSenderEmailModal({
         if (response) {
           // if pending or already authorized perform the check ahead
           executeAction();
+
+          // start polling on success response
+          setIntervalStopTime.current = moment()
+            .add(STOP_POLLING_AFTER, 'hours')
+            .valueOf();
+
+          setIntervalId.current = setInterval(
+            executeAction,
+            1000 * SET_INTERVAL_SECONDS,
+          );
         }
       })
       .catch(() => {
@@ -137,15 +149,7 @@ function AuthorizeSenderEmailModal({
         setShowLoader(false);
       });
 
-    clearCurrentInterval(setIntervalId.current);
-    setIntervalStopTime.current = moment()
-      .add(STOP_POLLING_AFTER, 'hours')
-      .valueOf();
-
-    const invervalID = setInterval(executeAction, 1000 * SET_INTERVAL_SECONDS);
-    setIntervalId.current = invervalID;
-
-    return () => clearCurrentInterval(invervalID);
+    return () => clearCurrentInterval(setIntervalId.current);
   }, [senderEmailAddress]); // eslint-disable-line react-hooks/exhaustive-deps
 
   return (
--- a/mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
+++ b/mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
@@ -1,0 +1,86 @@
+import { useEffect, useState } from 'react';
+import ReactStringReplace from 'react-string-replace';
+import PropTypes from 'prop-types';
+import { MailPoet } from 'mailpoet';
+import { Modal } from 'common/modal/modal';
+import { Loader } from 'common';
+
+type ApiActionType = 'create' | 'confirm';
+
+/**
+ * @param {string} email - Email address
+ * @param {ApiActionType} type - action type
+ * @returns {Promise}
+ */
+const makeApiRequest = (email: string, type: ApiActionType = 'create') =>
+  MailPoet.Ajax.post({
+    api_version: MailPoet.apiVersion,
+    endpoint: 'settings',
+    action:
+      type === 'create'
+        ? 'authorizeSenderEmailAddress'
+        : 'confirmSenderEmailAddressIsAuthorized',
+    data: {
+      email,
+    },
+  });
+
+type Props = {
+  senderEmail: string;
+  onRequestClose: () => void;
+};
+
+function AuthorizeSenderEmailModal({ senderEmail, onRequestClose }: Props) {
+  const [apiResponseType, setApiResponseType] = useState<boolean>(null);
+
+  const senderEmailAddress = String(senderEmail).toLowerCase();
+
+  useEffect(() => {
+    if (!senderEmailAddress) {
+      return;
+    }
+
+    makeApiRequest(senderEmailAddress)
+      .then((res) => {
+        setApiResponseType(Boolean(res?.data));
+      })
+      .catch(() => {
+        setApiResponseType(false);
+      });
+  }, [senderEmailAddress]);
+
+  return (
+    <Modal
+      title={MailPoet.I18n.t('authorizeSenderEmailModalTitle').replace(
+        '[senderEmail]',
+        senderEmailAddress,
+      )}
+      onRequestClose={onRequestClose}
+      contentClassName="authorize-sender-email-modal"
+    >
+      {apiResponseType && (
+        <p>
+          {ReactStringReplace(
+            MailPoet.I18n.t('authorizeSenderEmailModalDescription'),
+            /\[bold\](.*?)\[\/bold\]/g,
+            (match, i) => (
+              <strong key={i}>{match}</strong>
+            ),
+          )}
+        </p>
+      )}
+
+      {apiResponseType === false && (
+        <p>{MailPoet.I18n.t('authorizeSenderEmailMessageError')}</p>
+      )}
+
+      <Loader size={64} />
+    </Modal>
+  );
+}
+
+AuthorizeSenderEmailModal.propTypes = {
+  senderEmail: PropTypes.string.isRequired,
+};
+
+export { AuthorizeSenderEmailModal };
--- a/mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
+++ b/mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
@@ -6,6 +6,7 @@ import moment from 'moment';
 import { MailPoet } from 'mailpoet';
 import { Modal } from 'common/modal/modal';
 import { Button, Loader } from 'common';
+import { isErrorResponse, ErrorResponse } from 'ajax';
 
 const SET_INTERVAL_SECONDS = 15;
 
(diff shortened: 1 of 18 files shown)


mailpoet/lib/Services/AuthorizedEmailsController.php
```diff
@@ -58,6 +64,37 @@ public function setFromEmailAddress(string $address) {
     $this->settings->set(self::AUTHORIZED_EMAIL_ADDRESSES_ERROR_SETTING, null);
   }
 
+  public function getAllAuthorizedEmailAddress() {
```
lysyjan: Only one of the three new methods has a return type. Is there a reason for that? I guess we want to have at least an annotation with type, but return type is better when it is possible.

mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
```diff
@@ -0,0 +1,86 @@
...
+    },
+  });
+
+type Props = {
+  senderEmail: string;
+  onRequestClose: () => void;
+};
+
+function AuthorizeSenderEmailModal({ senderEmail, onRequestClose }: Props) {
+  const [apiResponseType, setApiResponseType] = useState<boolean>(null);
```
lysyjan: I have a question, but do you think avoiding the null value as default is possible?

mailpoet/assets/js/src/common/sender_email_address_warning.jsx
```diff
@@ -10,24 +12,42 @@ function SenderEmailAddressWarning({
...
-            className="mailpoet-link"
-            href={`https://account.mailpoet.com/authorization?email=${emailAddress}`}
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            {MailPoet.I18n.t('authorizeMyEmail')}
-          </a>
-        </p>
+        <>
+          {showAuthorizedEmailModel && (
```
lysyjan: It looks like a typo here. Maybe it should be `showAuthorizedEmailModal` instead of `..Model`.

mailpoet/lib/Services/Bridge/API.php
```diff
@@ -180,7 +180,14 @@ public function getAuthorizedEmailAddresses(): ?array {
     return is_array($data) ? $data : null;
   }
 
-  public function createAuthorizedEmailAddress(string $emailAdress): bool {
+  /**
+   * Create Authorized Email Address
+   *
+   * returns true if done or an array of error messages
```
lysyjan: Have you considered a similar approach as we have in the `sendMessage` method? I think it would be better to have a strict return type.

mailpoet/lib/Services/Bridge/API.php
```diff
@@ -180,7 +180,14 @@ public function getAuthorizedEmailAddresses(): ?array {
...
+   * @param string $emailAdress
+   * @return bool|array
+   */
+  public function createAuthorizedEmailAddress(string $emailAdress) {
```
lysyjan: Here is a typo in the parameter name. It should be `$emailAddress`, right?

mailpoet/lib/Services/Bridge.php
```diff
@@ -125,7 +125,7 @@ public function getAuthorizedEmailAddresses($type = 'authorized'): array {
     if ($data && $type === 'all') {
       return $data;
     }
-    return $data ? $data[$type] : [];
+    return $data && isset($data[$type]) ? $data[$type] : [];
```
lysyjan: I guess that the first part of the condition is redundant here. Or am I missing something?

mailpoet/assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -23,18 +24,13 @@ const handleSave = (address: string | null) =>
...
     return MailPoet.I18n.t('setFromAddressEmailUnknownError');
   }
 
   if (error.error === 'unauthorized') {
-    const fromAddress = encodeURIComponent(address);
-    return MailPoet.I18n.t('setFromAddressEmailNotAuthorized').replace(
-      /\[link\](.*?)\[\/link\]/g,
-      `<a href="https://account.mailpoet.com/authorization?email=${fromAddress}" target="_blank" rel="noopener noreferrer">$1</a>`,
-    );
+    return ' ';
```
lysyjan: Should it be this string a space? I really don't know, so I want to ask you.

mailpoet/assets/js/src/common/set_from_address_modal.tsx
```diff
@@ -86,6 +82,10 @@ type Props = {
 
 function SetFromAddressModal({ onRequestClose, setAuthorizedAddress }: Props) {
   const [address, setAddress] = useState<string>();
+  const [showAuthorizedEmailModel, setShowAuthorizedEmailModel] =
```
lysyjan: Does this look like another typo? `Model` instead of `Modal`?

mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
```diff
@@ -81,47 +85,43 @@ function AuthorizeSenderEmailModal({
...
+      })
+      .then(() => {
+        setCreateEmailApiResponse(null);
+        setShowLoader(false);
+        setConfirmEmailApiResponse(true);
+        setAuthorizedAddress(senderEmailAddress);
+        removeUnauthorizedEmailNotices();
+      })
+      .catch(() => {
+        //
```
lysyjan: This empty comment looks suspicious. I would expect here at least a short sentence. What do you think about it?

mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
```diff
@@ -146,7 +146,7 @@ function AuthorizeSenderEmailModal({
     setIntervalId.current = invervalID;
 
     return () => clearCurrentInterval(invervalID);
-  }, [senderEmailAddress, setAuthorizedAddress]);
+  }, [senderEmailAddress]); // eslint-disable-line react-hooks/exhaustive-deps
```
lysyjan, line 182: I have read the comment about disabling eslint in the commit message, but we agreed that we want to have an explanation directly in the code. Please, can you change it? The full P2 post is [here](https://mailpoetdev.wordpress.com/2021/03/19/disabling-linting-rules/).

mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
```diff
@@ -1,9 +1,14 @@
...
 import PropTypes from 'prop-types';
+import moment from 'moment';
 import { MailPoet } from 'mailpoet';
 import { Modal } from 'common/modal/modal';
-import { Loader } from 'common';
+import { Button, Loader } from 'common';
+
+const SET_INTERVAL_SECONDS = 15;
+
+const STOP_POLLING_AFTER = 2; // hours
```
lysyjan: I'm a little bit confused about the new constants. I would expect a similar name pattern. Do you think that it is possible?

lysyjan (commented):
Hello @triple0t,
Great job with this pull request.
I have several remarks, but I think that nothing critical.

mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
triple0t: Yup, that's possible.
I wanted the constants naming to be declarative of its purpose.

Would `SET_INTERVAL_STOP_POLLING_AFTER_HOURS` be okay?

mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
triple0t: Noted

mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
triple0t: Noted.
I wasn’t aware of that. Thanks for linking the P2 🙌🏿

mailpoet/assets/js/src/common/sender_email_address_warning.jsx
triple0t: Yup, should definitely be a `Modal`.
Thanks for catching this 👏🏾

mailpoet/assets/js/src/common/set_from_address_modal.tsx
triple0t: Yeah. that should be a space otherwise a component below throws an error.
I will refactor it

mailpoet/assets/js/src/common/set_from_address_modal.tsx
triple0t: Yup, it should be a `Modal`
Thanks for catching this

mailpoet/lib/Services/AuthorizedEmailsController.php
triple0t: Noted.

mailpoet/lib/Services/Bridge.php
triple0t: I don't think that part of the condition is redundant.

Unlike the previous implementation, this includes `$data[$type]` where `$type` is now a variable. 

It's a good idea to confirm the value exists before accessing it otherwise we would get  an Undefined index error

mailpoet/lib/Services/Bridge/API.php
triple0t: Yeah, taking another look at this, I believe having a consistent return type would be much better.

This section will be updated.

mailpoet/lib/Services/Bridge/API.php
triple0t: Yup, should be `$emailAddress`. 
Thanks for catching it

mailpoet/assets/js/src/common/authorize_sender_email_modal.tsx
triple0t: May I ask why?

I'm using this as a sort of indeterminate state.

true represents a positive API response
false represents a negative API response
null (default) no response yet.

triple0t:
Thank you @lysyjan, for the in-depth review.

I have implemented your feedback.

mailpoet/lib/Services/Bridge.php
lysyjan: @triple0t I still think the condition would work without the first part (`$data`).
Take look at [#](https://onlinephp.io/?s=s7EvyCjg5eLlUimpLEhVsFVQT0lNy8xLTVG35uXS11dQSUksSQQKR8PFFWztgLxEdR0F9ST12FigsrLEoviU0twCjczi4tQSDbCWaLB5sZoK9grIfAUrhehYTWsA&v=7.4.30).

triple0t: Okay 👍

lysyjan (approved):
LGTM!

veljkho (approved):

PR #4230, opened 2022-07-07
Add tags dynamic filter [PREMIUM-196]

lysyjan:
Please start reviewing from https://github.com/mailpoet/mailpoet/commit/789e91413254c917aff3a690a420e42a952e14da

Depends on https://github.com/mailpoet/mailpoet/pull/4214

**How to test**:

1. Install the free plugin and the premium from https://github.com/mailpoet/mailpoet-premium/pull/596
2. Assign a tag to several subscribers
3. Create a new dynamic segment with tags - check if the count is correct
4. Try to deactivate the premium plugin
    a. it shouldn't be possible to view a count of subscribers or to edit the segment
    b. the new tag filter should be visible in the select, but it shouldn't be possible to use

[PREMIUM-196]

[PREMIUM-196]: https://mailpoet.atlassian.net/browse/PREMIUM-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
+++ b/mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
@@ -11,6 +11,7 @@ use MailPoet\Segments\DynamicSegments\Filters\MailPoetCustomFields;
 use MailPoet\Segments\DynamicSegments\Filters\SubscriberScore;
 use MailPoet\Segments\DynamicSegments\Filters\SubscriberSegment;
 use MailPoet\Segments\DynamicSegments\Filters\SubscriberSubscribedDate;
+use MailPoet\Segments\DynamicSegments\Filters\SubscriberTag;
 use MailPoet\Segments\DynamicSegments\Filters\WooCommerceCategory;
 use MailPoet\Segments\DynamicSegments\Filters\WooCommerceCountry;
 use MailPoet\Segments\DynamicSegments\Filters\WooCommerceMembership;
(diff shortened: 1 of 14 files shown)


samnajian:
Awesome feature, thanks @lysyjan
I get the following behaviour if the premium plugin is deactivated but I know the URL to the edit page, not sure if we can try to completely prevent the rendering of the edit page on the frontend ( at lease the second time I try for the frontend already knows this action is not allowed), however, I regard this mostly a nice to have rather than a must have since the feature works fine and there is no clear security/privilege issue in this.
https://d.pr/i/L5DkmP

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
```diff
@@ -125,6 +126,16 @@ private function createSubscriber(array $data): DynamicSegmentFilterData {
       if (!empty($data['operator'])) $filterData['operator'] = $data['operator'];
       return new DynamicSegmentFilterData(DynamicSegmentFilterData::TYPE_USER_ROLE, $data['action'], $filterData);
     }
+    if ($data['action'] === SubscriberTag::TYPE) {
```
samnajian, line 129: Seems like this method (`FilterDataMapper::map`) needs tests for all branches (and for this new branch) especially since it's rather big and convoluted, what do you think?

samnajian (changes requested):

lysyjan:
> I get the following behaviour if the premium plugin is deactivated but I know the URL to the edit page, not sure if we can try to completely prevent the rendering of the edit page on the frontend ( at lease the second time I try for the frontend already knows this action is not allowed), however, I regard this mostly a nice to have rather than a must have since the feature works fine and there is no clear security/privilege issue in this.

@samnajian We use this behavior for another dynamic filter that has dependencies on other plugins. I think it is fine.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
lysyjan: Good point. I will extend our tests.

samnajian (dismissed):

veljkho:
Hi @lysyjan 
One conflicting file was found. :)

lysyjan:
Hey @veljkho, this PR was rebased on the current `trunk`.

veljkho (approved):

PR #4231, opened 2022-07-07
Remove old model usage from MailPoet\Newsletter\Shortcodes\Categories\Link [MAILPOET-4361]

samnajian:
This PR removes old models from newsletter shortcode-link related files.

- The files using MailPoet\Newsletter\Shortcodes\Categories\Link where adjusted so that they are not using the old models either.
- Down the line NewslettersExporter had to get adjusted as well.
- Test where adjusted as well.

[MAILPOET-4361]

[MAILPOET-4361]: https://mailpoet.atlassian.net/browse/MAILPOET-4361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Shortcodes.php
+++ b/mailpoet/lib/Config/Shortcodes.php
@@ -18,7 +18,6 @@ use MailPoet\Segments\SegmentSubscribersRepository;
 use MailPoet\Subscribers\SubscribersRepository;
 use MailPoet\Subscription\Pages;
 use MailPoet\WP\Functions as WPFunctions;
-use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class Shortcodes {
   /** @var Pages */
@@ -39,9 +38,6 @@ class Shortcodes {
   /** @var NewslettersRepository */
   private $newslettersRepository;
 
-  /** @var EntityManager */
-  private $entityManager;
-
   /** @var Date */
   private $dateCategory;
 
@@ -61,7 +57,6 @@ class Shortcodes {
     SubscribersRepository $subscribersRepository,
     NewsletterUrl $newsletterUrl,
     NewslettersRepository $newslettersRepository,
-    EntityManager $entityManager,
     Date $dateCategory,
     Link $linkCategory,
     Newsletter $newsletterCategory,
@@ -73,7 +68,6 @@ class Shortcodes {
     $this->subscribersRepository = $subscribersRepository;
     $this->newsletterUrl = $newsletterUrl;
     $this->newslettersRepository = $newslettersRepository;
-    $this->entityManager = $entityManager;
     $this->dateCategory = $dateCategory;
     $this->linkCategory = $linkCategory;
     $this->newsletterCategory = $newsletterCategory;
@@ -235,7 +229,7 @@ class Shortcodes {
     $shortcodeProcessor->setQueue($queue);
     return '<a href="' . esc_attr($previewUrl) . '" target="_blank" title="'
       . esc_attr(__('Preview in a new tab', 'mailpoet')) . '">'
-      . esc_attr((string)$shortcodeProcessor->replace($queue->getNewsletterRenderedSubject())) .
+      . esc_attr((string)$shortcodeProcessor->replace($queue ? $queue->getNewsletterRenderedSubject() : '')) .
       '</a>';
   }
 }
--- a/mailpoet/lib/Config/Shortcodes.php
+++ b/mailpoet/lib/Config/Shortcodes.php
@@ -172,7 +172,6 @@ class Shortcodes {
     $newsletters = $this->newslettersRepository->getArchives($segmentIds);
 
     $subscriber = $this->subscribersRepository->getCurrentWPUser();
-    $subscriber = $subscriber ? Subscriber::findOne($subscriber->getId()) : null;
 
     if (empty($newsletters)) {
       return $this->wp->applyFilters(
@@ -216,7 +215,7 @@ class Shortcodes {
     );
   }
 
-  public function renderArchiveSubject(NewsletterEntity $newsletter, $subscriber, SendingQueueEntity $queue) {
+  public function renderArchiveSubject(NewsletterEntity $newsletter, ?SubscriberEntity $subscriber, ?SendingQueueEntity $queue) {
     $previewUrl = $this->newsletterUrl->getViewInBrowserUrl($newsletter, $subscriber, $queue);
     /**
      * An ugly workaround to make sure state is not shared via NewsletterShortcodes service
@@ -230,15 +229,9 @@ class Shortcodes {
       $this->subscriberCategory,
       $this->wp
     );
+
     $shortcodeProcessor->setNewsletter($newsletter);
-
-    if (is_object($subscriber) && !is_a($subscriber, SubscriberEntity::class) && !empty($subscriber->id)) {
-      $subscriberEntity = $this->entityManager->find(SubscriberEntity::class, $subscriber->id);
-    } else {
-      $subscriberEntity = new SubscriberEntity();
-    }
-
-    $shortcodeProcessor->setSubscriber($subscriberEntity);
+    $shortcodeProcessor->setSubscriber($subscriber ?? new SubscriberEntity());
     $shortcodeProcessor->setQueue($queue);
     return '<a href="' . esc_attr($previewUrl) . '" target="_blank" title="'
       . esc_attr(__('Preview in a new tab', 'mailpoet')) . '">'
--- a/mailpoet/lib/Newsletter/Shortcodes/Categories/Link.php
+++ b/mailpoet/lib/Newsletter/Shortcodes/Categories/Link.php
@@ -6,7 +6,6 @@ use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Models\SendingQueue;
-use MailPoet\Models\Subscriber as SubscriberModel;
 use MailPoet\Newsletter\Url as NewsletterUrl;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Settings\TrackingConfig;
@@ -49,7 +48,6 @@ class Link implements CategoryInterface {
     bool $wpUserPreview = false
   ): ?string {
     $subscriptionUrlFactory = SubscriptionUrlFactory::getInstance();
-    $subscriberModel = $this->getSubscriberModel($subscriber);
     $queueModel = $this->getQueueModel($queue);
 
     switch ($shortcodeDetails['action']) {
@@ -80,7 +78,7 @@ class Link implements CategoryInterface {
       case 'newsletter_view_in_browser_url':
         $url = $this->newsletterUrl->getViewInBrowserUrl(
           $newsletter,
-          $wpUserPreview ? null : $subscriberModel,
+          $wpUserPreview ? null : $subscriber,
           $queueModel,
           $wpUserPreview
         );
@@ -122,7 +120,6 @@ class Link implements CategoryInterface {
     SendingQueueEntity $queue = null,
     $wpUserPreview = false
   ): ?string {
-    $subscriberModel = $this->getSubscriberModel($subscriber);
     $queueModel = $this->getQueueModel($queue);
     $subscriptionUrlFactory = SubscriptionUrlFactory::getInstance();
     switch ($shortcodeAction) {
@@ -138,7 +135,7 @@ class Link implements CategoryInterface {
       case 'newsletter_view_in_browser_url':
         $url = $this->newsletterUrl->getViewInBrowserUrl(
           $newsletter,
-          $subscriberModel,
+          $subscriber,
           $queueModel,
           false
         );
@@ -170,14 +167,6 @@ class Link implements CategoryInterface {
     if ($queue instanceof SendingQueueEntity) {
       return $queue->getId();
     }
-    return null;
-  }
-
-  // temporary function until Links are refactored to Doctrine
-  private function getSubscriberModel(SubscriberEntity $subscriber = null): ?SubscriberModel {
-    if (!$subscriber) return null;
-    $subscriberModel = SubscriberModel::where('id', $subscriber->getId())->findOne();
-    if ($subscriberModel) return $subscriberModel;
     return null;
   }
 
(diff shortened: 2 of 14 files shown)


mailpoet/lib/Newsletter/Url.php
```diff
@@ -27,7 +27,7 @@ public function __construct(
   }
 
   public function getViewInBrowserUrl(
-    $newsletter,
+    NewsletterEntity $newsletter = null,
```
lysyjan: I think that this declaration should be `?NesletterEntity` without the default value. I checked all calls of this function, and the newsletter is always defined. So the first parameter should be mandatory.

mailpoet/lib/Newsletter/Url.php
```diff
@@ -46,14 +46,9 @@ public function getViewInBrowserUrl(
...
 
-  public function createUrlDataObject($newsletter, $subscriber, $queue, $preview) {
-    if ($newsletter instanceof NewsletterEntity) {
-      $newsletterId = (!empty($newsletter->getId())) ? (int)$newsletter->getId() : 0;
-      $newsletterHash = (!empty($newsletter->getHash())) ? $newsletter->getHash() : 0;
-    } else {
-      $newsletterId = (!empty($newsletter->id)) ? (int)$newsletter->id : 0;
-      $newsletterHash = (!empty($newsletter->hash)) ? $newsletter->hash : 0;
-    }
+  public function createUrlDataObject(NewsletterEntity $newsletter = null, $subscriber, $queue, $preview) {
```
lysyjan: This situation here is similar to my previous comment. I also noticed that this method is called only inside this class, so maybe we can change the visibility to private. What do you think?

mailpoet/lib/Newsletter/Url.php
lysyjan: You overlooked [a usage in the premium plugin](https://github.com/mailpoet/mailpoet-premium/blob/trunk/lib/API/JSON/v1/Stats.php#L85-L90). I guess we need to create a separate PR for this task.

mailpoet/lib/Config/Shortcodes.php
```diff
@@ -216,7 +215,7 @@ public function renderArchiveDate(NewsletterEntity $newsletter) {
     );
   }
 
-  public function renderArchiveSubject(NewsletterEntity $newsletter, $subscriber, SendingQueueEntity $queue) {
+  public function renderArchiveSubject(NewsletterEntity $newsletter, SubscriberEntity $subscriber = null, SendingQueueEntity $queue = null) {
```
lysyjan: Why did you change the second and the third parameter to optional when they were required? If you want to have them nullable, you can use the nullable type `?SubscriberEntity`.

mailpoet/lib/Config/Shortcodes.php
```diff
@@ -230,14 +229,9 @@ public function renderArchiveSubject(NewsletterEntity $newsletter, $subscriber,
...
       $this->wp
     );
-    $shortcodeProcessor->setNewsletter($newsletter);
-
-    if (is_object($subscriber) && !is_a($subscriber, SubscriberEntity::class) && !empty($subscriber->id)) {
-      $subscriberEntity = $this->entityManager->find(SubscriberEntity::class, $subscriber->id);
-    } else {
-      $subscriberEntity = new SubscriberEntity();
-    }
+    $subscriberEntity = $subscriber ?? new SubscriberEntity();
```
lysyjan: This variable is redundant because you can move this check to line 235, but it's only cosmetic detail.

mailpoet/lib/Newsletter/Url.php
```diff
@@ -5,39 +5,26 @@
...
+    LinkTokens $linkTokens
   ) {
     $this->linkTokens = $linkTokens;
-    $this->subscribersRepository = $subscribersRepository;
   }
 
   public function getViewInBrowserUrl(
     NewsletterEntity $newsletter = null,
-    $subscriber = false,
+    SubscriberEntity $subscriber = null,
```
lysyjan: I think that this declaration should be `?SubscriberEntity` with the default value `null`.

mailpoet/lib/Config/Shortcodes.php
```diff
@@ -236,7 +230,7 @@ public function renderArchiveSubject(NewsletterEntity $newsletter, SubscriberEnt
     $shortcodeProcessor->setQueue($queue);
     return '<a href="' . esc_attr($previewUrl) . '" target="_blank" title="'
       . esc_attr(__('Preview in a new tab', 'mailpoet')) . '">'
-      . esc_attr((string)$shortcodeProcessor->replace($queue->getNewsletterRenderedSubject())) .
+      . esc_attr((string)$shortcodeProcessor->replace($queue ? $queue->getNewsletterRenderedSubject() : "")) .
```
lysyjan: How is it possible that the `$queue` variable is empty now? This looks very suspicious to me. I guess that changes in the function header cause this.

mailpoet/lib/Newsletter/Shortcodes/Categories/Link.php
```diff
@@ -48,21 +47,26 @@ public function process(
...
           $wpUserPreview
         );
 
       case 'subscription_instant_unsubscribe_url':
         return self::processUrl(
           $shortcodeDetails['action'],
-          $subscriptionUrlFactory->getUnsubscribeUrl($wpUserPreview ? null : $subscriber, self::getSendingQueueId($queue)),
+          $subscriptionUrlFactory->getUnsubscribeUrl(
+            $wpUserPreview ? null : $subscriber,
+            $queue ? $queue->getId() : null
```
lysyjan, line 68: Have you considered changing an expected queue type in the function `getUnsubscribeUrl`? The method needs to have a similar logic. Maybe we can pass the whole entity instead of extracted id. What do you think? I know that more changes would be necessary, but the logic could look easier.

mailpoet/lib/Newsletter/Shortcodes/Categories/Link.php
```diff
@@ -106,7 +110,7 @@ public function process(
     }
   }
 
-  public function processUrl($action, $url, $queue, $wpUserPreview = false): string {
+  public function processUrl($action, $url, SendingQueueEntity $queue = null, $wpUserPreview = false): string {
```
lysyjan: You make the third parameter optional again. I think that this is not a good approach, and we should use the nullable type instead.

mailpoet/lib/Newsletter/Url.php
```diff
@@ -22,26 +22,26 @@ public function __construct(
...
+    $data = $this->createUrlDataObject($preview, $newsletter, $subscriber, $queue);
     return Router::buildRequest(
       ViewInBrowserEndpoint::ENDPOINT,
       ViewInBrowserEndpoint::ACTION_VIEW,
       $data
     );
   }
 
-  public function createUrlDataObject(NewsletterEntity $newsletter = null, SubscriberEntity $subscriber = null, $queue, $preview) {
+  public function createUrlDataObject(
```
lysyjan, line 36: You had to change the order of parameters because you didn't use nullable types. Also, the `$preview` parameter doesn't have a type here, but we know from the function above that it is of type boolean.

mailpoet/tests/integration/Config/ShortcodesTest.php
```diff
@@ -5,38 +5,51 @@
...
-    $newsletter = Newsletter::create();
-    $newsletter->type = Newsletter::TYPE_STANDARD;
-    $newsletter->status = Newsletter::STATUS_SENT;
-    $this->newsletter = $newsletter->save();
-    $queue = SendingTask::create();
-    $queue->newsletterId = $newsletter->id;
-    $queue->status = SendingQueue::STATUS_COMPLETED;
-    $this->queue = $queue->save();
+    $this->subscribersRepository = $this->diContainer->get(SubscribersRepository::class);
+    $this->newsletter = new NewsletterEntity();
```
lysyjan: Have you considered using [the Newsletter factory](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/DataFactories/Newsletter.php) here? In my opinion, entity creation is easier with it. We have a factory for `the ScheduledTaskEntity`, but maybe it doesn't cover your requirements here.

mailpoet/tests/integration/Config/ShortcodesTest.php
```diff
@@ -65,13 +78,15 @@ public function testItGetsArchives() {
...
-    expect($requestData['newsletter_hash'])->equals($this->newsletter->hash);
+    expect($requestData['newsletter_hash'])->equals($this->newsletter->getHash());
   }
 
   public function testItRendersShortcodeDefaultsInSubject() {
     $shortcodes = ContainerWrapper::getInstance()->get(Shortcodes::class);
-    $this->queue->newsletterRenderedSubject = 'Hello [subscriber:firstname | default:reader]';
-    $this->queue->save();
+    $this->queue->setNewsletterRenderedSubject('Hello [subscriber:firstname | default:reader]');
+    $this->entityManager->persist($this->queue);
```
lysyjan: You don't have to use the `persist` method when the record exists in the database, but it is only a detail.

mailpoet/tests/integration/Newsletter/Preview/SendPreviewControllerTest.php
```diff
@@ -36,7 +37,7 @@ public function _before() {
     $this->subscriptionUrlFactory = SubscriptionUrlFactory::getInstance();
     $newsletter = new NewsletterEntity();
     $newsletter->setType(NewsletterEntity::TYPE_STANDARD);
-    $newsletter->setSubject('My Standard Newsletter');
+    $newsletter->setSubject('My Standard Newsletter SendPreviewControllerTest');
```
lysyjan, line 39: Why did you change the subject? Is there a reason for that?

mailpoet/lib/Subscribers/ImportExport/PersonalDataExporters/NewslettersExporter.php
```diff
@@ -15,21 +16,31 @@ class NewslettersExporter {
...
-    $data = $this->exportSubscriber(Subscriber::findOne(trim($email)), $page);
+    $data = $this->exportSubscriber($this->subscribersRepository->findOneBy(['email' => trim($email)]), $page);
     return [
       'data' => $data,
       'done' => count($data) < self::LIMIT,
     ];
   }
 
-  private function exportSubscriber($subscriber, $page) {
+  private function exportSubscriber(SubscriberEntity $subscriber = null, $page) {
```
lysyjan: The first parameter should not be optional but only nullable because it is always set.

lysyjan (changes requested):
I know this is a complicated pull request because the work with shortcodes is not simple.
I noticed that you several times used the optional parameter with the default value `null` when the nullable type would be enough. You can see in the pull request that those changes have some consequences in other places.
Please, can you take a look at it?

mailpoet/lib/Newsletter/Shortcodes/Categories/Link.php
samnajian: As you stated more changes would be needed since the default value for `$queue` is `null`, that's why I decided to not try to change alot for this specific case
```  
public function process(
    array $shortcodeDetails,
    NewsletterEntity $newsletter = null,
    SubscriberEntity $subscriber = null,
    SendingQueueEntity $queue = null,
    string $content = '',
    bool $wpUserPreview = false
  ): ?string ```

Comment (lysyjan) [reply mailpoet/lib/Newsletter/Shortcodes/Categories/Link.php:68]: Okay, I'm fine with that.

File: mailpoet/tests/integration/Newsletter/Preview/SendPreviewControllerTest.php (hunk 12 above)
Comment (samnajian) [reply mailpoet/tests/integration/Newsletter/Preview/SendPreviewControllerTest.php:39]: made it easier for me to debug by making the subject more specific, and thought it can remain like this since it's a test and we may need to debug in future as well :)

File: mailpoet/lib/Config/Shortcodes.php [hunk 14]
```diff
@@ -215,7 +209,7 @@ public function renderArchiveDate(NewsletterEntity $newsletter) {
     );
   }
 
-  public function renderArchiveSubject(NewsletterEntity $newsletter, ?SubscriberEntity $subscriber, ?SendingQueueEntity $queue) {
+  public function renderArchiveSubject(NewsletterEntity $newsletter, ?SubscriberEntity $subscriber, SendingQueueEntity $queue) {
```
lysyjan: You changed the last parameter type from `?SendingQueueEntity` to `SendingQueueEntity`. Is there a reason for that? I checked the usage in the Shortcodes class, and it looks that [the queue can be null](https://github.com/mailpoet/mailpoet/blob/MAILPOET-4361-remove-old-model-usage-from-mail-poet-newsletter-shortcodes-categories-link-s/mailpoet/lib/Config/Shortcodes.php#L182-L190).

mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserRenderer.php
```diff
@@ -93,19 +93,15 @@ public function render(
...
-  private function prepareShortcodes($newsletter, $subscriber, $queue, $wpUserPreview) {
-    if ($queue instanceof SendingQueueEntity) {
-      $this->shortcodes->setQueue($queue);
-    }
-
-    if ($newsletter instanceof NewsletterEntity) {
-      $this->shortcodes->setNewsletter($newsletter);
-    }
-
+  private function prepareShortcodes(
```
lysyjan, line 96: This function is called only once inside the class, so I assume that all parameters should be mandatory, and the first cannot be `null`. I apologize that I overlooked it in the first review iteration.

lysyjan (commented):
I found two details, but you haven't created a PR in the premium plugin that is necessary because premium contains usages of Url::getViewInBrowserUrl(), which have changed types of parameters.

mailpoet/lib/Config/Shortcodes.php
samnajian: This is in contrast to this [former comment](https://github.com/mailpoet/mailpoet/pull/4231#discussion_r920915464), na? but I agree it can be empty/null likes it's always has been like that, PHPStan doesn't catch this because the nullable branch goes through hooks, last time I looked I couldn't find the nullable branch easer!

mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserRenderer.php
samnajian: I'm also really convinced defining this method makes much sense since it's not really adding much, keeping it though, so that when we refactor how data is passed to the shortcodes ([MAILPOET-4087]) service we can change this.

lysyjan (approved):

veljkho (approved):

PR #4233, opened 2022-07-07
Expose test db port to local dev machine

samnajian:
A not very common db port number is used so that it won’t
easily conflict with other possible instances of database

[MAILPOET-4468]

[MAILPOET-4468]: https://mailpoet.atlassian.net/browse/MAILPOET-4468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/docker/docker-compose.yml
+++ b/mailpoet/tests/docker/docker-compose.yml
@@ -104,6 +104,8 @@ services:
       MYSQL_PASSWORD: wordpress
     volumes:
       - /dev/shm:/dev/shm
+    ports:
+      - 4401:3306
 
   chrome:
     environment:


rodrigoprimo (approved):
Looks good to me. Maybe worth announcing this (and the new way integration tests are executed) to the other developers. I don't think we need QA for this, so, in my opinion, you can go ahead and merge this PR.

samnajian:
> Looks good to me. Maybe worth announcing this (and the new way integration tests are executed) to the other developers. I don't think we need QA for this, so, in my opinion, you can go ahead and merge this PR.

Thanks for taking a look, as you mentioned it makes sense to add a P2 for this, I'll try to take care of it next week :)

PR #4235, opened 2022-07-08
Remove MP2 migration code [MAILPOET-4376]

rodrigoprimo:
This PR removes all the code related to the MP2 migration:
- the main MP2 migration admin page (/wp-admin/admin.php?page=mailpoet-migration)
- a different step in the welcome wizard when the site contains MP2 migrated data
- a table that was created to correlate MP2 and MP3 data
- special logic in the code that handles inactive subscribers when they were migrated from MP2

It was also necessary to move a SQL query that changes `SQL_MODE` from `mailpoet/tests/_data/createMP2Tables.sql` to `mailpoet/tests/integration/Models/SubscriberSegmentTest.php`. See the message in the commit e8bd5e7761c5229b0b24af97f5c5871a8ceea5fc for more details.

**How to test this PR**
I would suggest checking that there is no MP2 code left and that the code that is being removed is not still used somewhere.

[MAILPOET-4376]

[MAILPOET-4376]: https://mailpoet.atlassian.net/browse/MAILPOET-4376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo:
This PR will be marked as ready for review in mid-August as by them MP2 won't be officially supported anymore.

rodrigoprimo:
Marking this PR as ready for review now, since the next release will already be in mid-August.

mailpoet/tests/integration/Models/SubscriberSegmentTest.php
```diff
@@ -15,6 +16,7 @@ class SubscriberSegmentTest extends \MailPoetTest {
 
   public function _before() {
     parent::_before();
+    ORM::raw_execute('SET SQL_MODE = "NO_AUTO_VALUE_ON_ZERO"');
```
rodrigoprimo, line 19: I opted to use Paris code here as this is a test class for a Paris model that we will eventually remove.

mailpoet/lib/Subscribers/InactiveSubscribersController.php
```diff
@@ -160,40 +146,24 @@ private function activateSubscribers(Carbon $thresholdDate, int $batchSize): int
...
-          AND s.status = :statusInactive
-          AND s2.id IS NOT NULL
-        GROUP BY s.id
-        LIMIT :batchSize
-      ", [
-        'thresholdDate' => $thresholdDate,
-        'statusInactive' => SubscriberEntity::STATUS_INACTIVE,
-        'batchSize' => $batchSize,
-      ], ['batchSize' => \PDO::PARAM_INT])->fetchAllAssociative();
-    }
```
costasovo, line 196: We are removing the code for migrating subscribers, but the migrated subscribers will still remain in the database. So I guess we couldn't remove this code just because it is no longer possible to migrate subscribers. Also, there still be a flow for MP2 users in which they will be able to migrate (Download the last MP3 version that supports migration, migrate and then update).

But on the other hand, we have recently [strengthened the criteria for marking users inactive](https://mailpoet.atlassian.net/browse/MAILPOET-4081), and IMO, the new criteria "receiving three emails during the detection period" should cover also the case of migrated subscribers. So, in the end, I'm ok with removing the code. 

I'm posting this comment to explain the situation and so that someone can verify my reasoning. @rodrigoprimo Do you think the above makes sense?

mailpoet/lib/Models/MappingToExternalEntities.php
```diff
@@ -2,12 +2,37 @@
 
 namespace MailPoet\Models;
 
+/**
+ * @deprecated This model is deprecated and there is no replacement.
+ * This class can be removed after 2023-01-06.
+ */
```
costasovo, line 8: I think that because the table is not present anymore, the usage of this model will, apart from the deprecation warnings, cause also PHP Errors when trying to save stuff to DB, etc. Maybe, in this case, we could make an exception and remove the model right away. What do you think?

costasovo (commented):
I'm happy to see we are getting rid of this. Overall I think the change is fine, and I couldn't find any other files we should remove. I added just two questions.

mailpoet/lib/Models/MappingToExternalEntities.php
rodrigoprimo: Please correct if I'm wrong, but I believe the table will still exist for sites that activated the plugin before this change. In this PR, we are stopping the creation of the table for new installs, but we are not dropping the table for sites that already have it. So for the odd site that still has this table and for some reason is using this model directly, I would prefer to deprecate the model instead of removing it. I think there is a chance no such site exists, but anyway, I would suggest we play it safe and follow the deprecation path instead of removing the model straight away.

mailpoet/lib/Subscribers/InactiveSubscribersController.php
rodrigoprimo: > We are removing the code for migrating subscribers, but the migrated subscribers will still remain in the database. So I guess we couldn't remove this code just because it is no longer possible to migrate subscribers.

My understanding is that although the subscribers originating from MP2 will still exist in the database, they will already have been marked as active so I guess it is not a problem?

> Also, there still be a flow for MP2 users in which they will be able to migrate (Download the last MP3 version that supports migration, migrate and then update). 

This is a case that I haven't considered and that I agree would be affected by removing this code. To me, considering what you shared below about the new criteria for marking subscribers as inactive and that this is a edge case, I don't think it is important enough to keep this code.

> But on the other hand, we have recently [strengthened the criteria for marking users inactive](https://mailpoet.atlassian.net/browse/MAILPOET-4081), and IMO, the new criteria "receiving three emails during the detection period" should cover also the case of migrated subscribers. So, in the end, I'm ok with removing the code.

That is good to know.

> I'm posting this comment to explain the situation and so that someone can verify my reasoning. @rodrigoprimo Do you think the above makes sense?

Makes sense to me. In my opinion, we can proceed with the removal of this code.

mailpoet/lib/Models/MappingToExternalEntities.php
costasovo: > Please correct if I'm wrong, but I believe the table will still exist for sites that activated the plugin before this change. 

Good point. Let's keep the model.

costasovo:
@rodrigoprimo There is a conflict in MP2Migrator.php. The file was removed in this PR, so I assume this one will be easy to fix.

rodrigoprimo:
Thanks for your comments, @costasovo. I believe I responded to all of them. When you have a moment, could you please check my answers?

I also had to rebase this branch due to a conflict. Luckily it was in one of the files that I'm deleting here (mailpoet/lib/Config/MP2Migrator.php), so there are no real code changes since the last time you reviewed this PR.

rodrigoprimo:
Ops, I missed that you were faster than me and replied to the my answers before my last comment :-)

I will wait for the CircleCI build to pass and assign this PR back to you. I already fixed the conflict.

costasovo (approved):

veljkho (approved):

PR #4244, opened 2022-07-12
Refactor MailPoet\Config\Shortcodes to Doctrine [MAILPOET-4307]

rodrigoprimo:
This PR replaces all models with Doctrine entities and repositories in the classes MailPoet\Config\Shortcodes and \MailPoet\Config\ShortcodesTest. 

In the process, it was necessary to change the type of a parameter passed to a WP filter. To avoid fatal errors for users that might be using this filter, a new filter was created, and the old filter was deprecated. A similar process already happened for this same filter (unfortunately) about six months ago. See more information about the discussion that lead to the decision to create a new filter here: https://github.com/mailpoet/mailpoet/pull/3665

**How to test this PR**

Check that the following shortcodes are still working and their behavior is exactly the same as it was before the changes proposed here: mailpoet_subscribers_count and mailpoet_archive

[MAILPOET-4307]

[MAILPOET-4307]: https://mailpoet.atlassian.net/browse/MAILPOET-4307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Shortcodes.php
+++ b/mailpoet/lib/Config/Shortcodes.php
@@ -6,7 +6,6 @@ use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Form\Widget;
-use MailPoet\Models\Subscriber;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Shortcodes\Categories\Date;
 use MailPoet\Newsletter\Shortcodes\Categories\Link;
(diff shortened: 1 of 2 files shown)


websupporter (dismissed):
lgtm. Sorry for the late review and thank you @rodrigoprimo

websupporter:
reassigned it to @rodrigoprimo, because the PR needs to be rebased.

rodrigoprimo:
@websupporter, thanks for the ping. I have rebased the files. I will ask you to please check this PR again.

What happened here is that in #4231 @samnajian removed the models from ShortcodesTest and also removed some of the calls to Paris code from Shortcodes (maybe because there was some dependency between the changes that he was making in MailPoet\Newsletter\Shortcodes\Categories\Link, which is the focus of the ticket that he worked on, and Shortcodes and ShortcodesTest).

I opted to preserve my version of ShortcodesTest as it seemed to me like the easiest way to keep the modifications related to the changes in the name of the filters and I tried to merge the changes that we both made to Shortcodes.

websupporter:
Thanks @rodrigoprimo. One question: I got curious why we use `WordPress::interceptFunction` in the tests. With step debugger, I never end up there and when I remove those functions, the tests are still green. I do not understand, why we attempt to intercept those functions and why I do not get stopped out there.
image

rodrigoprimo:
@websupporter, I don't understand it as well. When I was working on this PR, I saw those calls but didn't stop to investigate what they were doing. I simply updated the filter names. If the tests pass without WordPress::interceptFunction() and WordPress::releaseFunction() I guess we can remove them, but we could also create a ticket to investigate those functions and if there are other places where they are being called but are not needed. Let me know what you prefer.

websupporter:
Created a ticket: https://mailpoet.atlassian.net/browse/MAILPOET-4509

websupporter (approved):
lgtm. Thank you @rodrigoprimo

veljkho (approved):

PR #4247, opened 2022-07-13
Refactor WorkerTest, MetaInfoTest, ShortcodesTest and LinksTest to use Doctrine [MAILPOET-4379]

rodrigoprimo:
Most of the changes proposed in this PR are fairly straightforward and only affect a few integration test classes. The only exception is the commit that refactors MetaInfoTest. While refactoring this class, it was necessary to change the signature of \MailPoet\Mailer\MetaInfo::getNewsletterMetaInfo() which also meant changing \MailPoet\Cron\Workers\SendingQueue\SendingQueue::processQueue() and its test class. See the message in the commit a31a35265a4441b2d0b835d25644de03dbeab10c for more information.

Since there are changes here to a method related to sending, it is probably a good idea to check that this part is still working as expected.

[MAILPOET-4379]

[MAILPOET-4379]: https://mailpoet.atlassian.net/browse/MAILPOET-4379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -284,6 +284,8 @@ class SendingQueue {
       $subscriberEntity = $this->subscribersRepository->findOneById($subscriber->id);
       if ($subscriberEntity instanceof SubscriberEntity) {
         $metas[] = $this->mailerMetaInfo->getNewsletterMetaInfo($newsletter, $subscriberEntity);
+      } else {
+        $metas[] = [];
       }
 
       // keep track of values for statistics purposes
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -10,6 +10,7 @@ use MailPoet\Cron\Workers\SendingQueue\Tasks\Newsletter as NewsletterTask;
 use MailPoet\Cron\Workers\StatsNotifications\Scheduler as StatsNotificationsScheduler;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Entities\SubscriberEntity;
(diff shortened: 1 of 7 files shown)


mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -273,7 +280,12 @@ public function processQueue($queue, $newsletter, $subscribers, $timer) {
       $preparedSubscribersIds[] = $subscriber->id;
       // create personalized instant unsubsribe link
       $unsubscribeUrls[] = $this->links->getUnsubscribeUrl($queue, $subscriber->id);
-      $metas[] = $this->mailerMetaInfo->getNewsletterMetaInfo($newsletter, $subscriber);
+
+      $subscriberEntity = $this->subscribersRepository->findOneById($subscriber->id);
+      if ($subscriberEntity instanceof SubscriberEntity) {
```
alex-mailpoet, line 285: Can we fill `$metas[]` with `[]` if this condition returns `false`? Otherwise `$metas` would be out of sync with other arrays. I assume that same indexes should hold data for the same recipient.

rodrigoprimo: When working on this I assumed it was not necessary. This code shouldn't last long as it is only there while we are transitioning from Paris to Doctrine. Soon we should refactor \MailPoet\Cron\Workers\SendingQueue\SendingQueue::processQueue() and it will receive an array of subscriber entities and it won't be necessary to call findOneById() here anymore. I think that if we have a subscriber model and we can't find the equivalent subscriber entity we have much bigger problems.

But anyway, adding the else doesn't hurt and it won't create any problems. So I added it in a new commit. Could you please check this PR again?

alex-mailpoet (approved):
Thank you for making changes! Looks good to me.

veljkho (approved):

PR #4249, opened 2022-07-13
Mailpoet 4348 refactor personal data exporters to doctrine s

samnajian:
### How to test:

- Verify the data export works in `wp-admin/admin.php?page=mailpoet-export` and `wp-admin/export-personal-data.php`

[MAILPOET-4348]

[MAILPOET-4348]: https://mailpoet.atlassian.net/browse/MAILPOET-4348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Initializer.php
+++ b/mailpoet/lib/Config/Initializer.php
@@ -102,6 +102,9 @@ class Initializer {
   /** @var FeaturesController */
   private $featuresController;
 
+  /*** @var PersonalDataExporters */
+  private $personalDataExporters;
+
   const INITIALIZED = 'MAILPOET_INITIALIZED';
 
   public function __construct(
@@ -129,7 +132,8 @@ class Initializer {
     AssetsLoader $assetsLoader,
     Engine $automationEngine,
     MailPoetIntegration $automationMailPoetIntegration,
-    FeaturesController $featuresController
+    FeaturesController $featuresController,
+    PersonalDataExporters $personalDataExporters
   ) {
     $this->rendererFactory = $rendererFactory;
     $this->accessControl = $accessControl;
@@ -156,6 +160,7 @@ class Initializer {
     $this->automationEngine = $automationEngine;
     $this->automationMailPoetIntegration = $automationMailPoetIntegration;
     $this->featuresController = $featuresController;
+    $this->personalDataExporters = $personalDataExporters;
   }
 
   public function init() {
@@ -387,8 +392,7 @@ class Initializer {
   }
 
   public function setupPersonalDataExporters() {
-    $exporters = new PersonalDataExporters();
-    $exporters->init();
+    $this->personalDataExporters->init();
   }
 
   public function setupPersonalDataErasers() {
--- a/mailpoet/lib/Newsletter/Statistics/NewsletterStatisticsRepository.php
+++ b/mailpoet/lib/Newsletter/Statistics/NewsletterStatisticsRepository.php
@@ -5,14 +5,18 @@ namespace MailPoet\Newsletter\Statistics;
 use MailPoet\Doctrine\Repository;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\StatisticsBounceEntity;
 use MailPoet\Entities\StatisticsClickEntity;
+use MailPoet\Entities\StatisticsNewsletterEntity;
 use MailPoet\Entities\StatisticsOpenEntity;
 use MailPoet\Entities\StatisticsUnsubscribeEntity;
 use MailPoet\Entities\StatisticsWooCommercePurchaseEntity;
+use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Entities\UserAgentEntity;
 use MailPoet\WooCommerce\Helper as WCHelper;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
+use MailPoetVendor\Doctrine\ORM\Query\Expr\Join;
 use MailPoetVendor\Doctrine\ORM\QueryBuilder;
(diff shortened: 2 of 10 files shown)


mailpoet/lib/Config/Initializer.php
```diff
@@ -387,7 +393,7 @@ public function setupPrivacyPolicy() {
   }
 
   public function setupPersonalDataExporters() {
-    $exporters = new PersonalDataExporters();
+    $exporters = new PersonalDataExporters($this->subscribersRepository);
```
costasovo: I guess we could add the `PersonalDataExporters` class into [the DI container config](https://github.com/mailpoet/mailpoet/blob/1eef663ca3631837cc97571fd069216c8db92aef/mailpoet/lib/DI/ContainerConfigurator.php) and inject it into the Initializer instead of injecting the `SubscribersRepository`. I don't think there was a reason why we kept `PersonalDataExporters` instantiation inline. It was probably just a leftover.

mailpoet/lib/Config/Initializer.php
```diff
@@ -393,7 +401,10 @@ public function setupPrivacyPolicy() {
   }
 
   public function setupPersonalDataExporters() {
-    $exporters = new PersonalDataExporters($this->subscribersRepository);
+    $exporters = new PersonalDataExporters(
+      $this->subscribersRepository,
+      $this->customFieldsRepository
```
costasovo: Now we are injecting even more services so that we can build the `PersonalDataExporters`. So I would suggest moving it to DI container.

mailpoet/lib/Entities/SubscriberEntity.php
```diff
@@ -529,4 +529,11 @@ public function cleanupSubscriberSegments(): void {
       }
     });
   }
+
+  public function withCustomFiels(): SubscriberEntity {
+    foreach ($this->getSubscriberCustomFields() as $field) {
+      $this->{'cf_' . $field->getId()} = $field->getValue();
+    }
+    return $this;
```
costasovo: Could you please use a standard getter? We use the `withSomething...` approach in the old models, and it is something I (and I believe I'm not alone) dislike. It changes the shape of the object. Suddenly, an object of the same type may have different shapes. On top of that, these dynamic properties don't even respect camel case naming.
I see that it is used in `SubscriberExporter::getSubscriberExportData`. It feels like a proprietary thing just for the export because the export uses some naming in the format `cf_{id}`. So I would suggest moving this to the export class. Or maybe creating some getter `SubscriberEntity::getCustomFieldBuId()`.

Looking at it more closely it feels like we skip custom fields without values.
```php
foreach ($customFields as $customFieldId => $customFieldName) {
      $customFieldValue = $subscriber->{$customFieldId};	   
      if ($customFieldValue) {	
        $result[] = [
          'name' => $customFieldName,
          'value' => $customFieldValue,
        ];	      
      }	    
    }	   
```
So If I'm not missing anything you could simply use $subscriber->getSubscriberCustomFields() and iterate those.

mailpoet/lib/Newsletter/Statistics/NewsletterStatisticsRepository.php
```diff
@@ -102,6 +106,25 @@ public function getStatisticsMachineOpenCount(NewsletterEntity $newsletter): int
     return $result['cnt'] ?? 0;
   }
 
+  public function getAllForSubscriber(SubscriberEntity $subscriber): QueryBuilder {
```
costasovo: The name of the method suggests it returns an array of`NewsletterStatisticsEntity`, but it returns the query builder. I guess we could add optional limit and offset parametes, which I believe is more aligned with how we use repositories, or maybe rename the method so that it is more obvious it returns query builder.
What do you think?

costasovo (changes requested):
Thank you for refactoring this peace to Doctrine! I'm happy to see we are moving forward.
I found a couple of small issues. Please see my comments.

mailpoet/lib/Subscribers/ImportExport/PersonalDataExporters/SubscriberExporter.php
```diff
@@ -2,94 +2,132 @@
...
+        ? $subscriber->getCreatedAt()->format(DateTime::DEFAULT_DATE_TIME_FORMAT)
+        : '',
     ];
 
-    foreach ($customFields as $customFieldId => $customFieldName) {
-      $customFieldValue = $subscriber->{$customFieldId};
-      if ($customFieldValue) {
+    foreach ($subscriber->getSubscriberCustomFields() as $field) {
+      if (isset($this->getCustomFields()[$field->getId()])) {
         $result[] = [
```
costasovo, line 100: This also changes the logic. The old code was skipping custom fields with falsy values (`if ($customFieldValue)`). The new code checks if the custom field is defined in the list of all custom fields and exports also the falsy values and empty fields. Is this change intentional?

Thinking about it. The change might be ok because the current export is skipping falsy values, but they also carry information.

costasovo (commented):
Thanks for the fixes! I have one question about the logic change I noticed. See my comment.

mailpoet/lib/Subscribers/ImportExport/PersonalDataExporters/SubscriberExporter.php
costasovo: Maybe we could consult this ⬆️  with @pavel-mailpoet .

pavel-mailpoet: I think it is fine to keep those custom fields with false values in the export.

costasovo (approved):
LGTM! 👍

veljkho (approved):

PR #4250, opened 2022-07-13
Add automation workflow name editing [MAILPOET-4422]

alex-mailpoet:
[MAILPOET-4422]

[MAILPOET-4422]: https://mailpoet.atlassian.net/browse/MAILPOET-4422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
@@ -27,15 +27,20 @@ class UpdateWorkflowController {
       throw Exceptions::workflowNotFound($id);
     }
 
+    $changed = false;
+
     if (array_key_exists('name', $data)) {
-      $this->checkWorkflowName($data['name']);
       $workflow->setName($data['name']);
-      $this->storage->updateWorkflow($workflow);
+      $changed = true;
     }
 
     if (array_key_exists('status', $data)) {
       $this->checkWorkflowStatus($data['status']);
       $workflow->setStatus($data['status']);
+      $changed = true;
+    }
+
+    if ($changed) {
       $this->storage->updateWorkflow($workflow);
     }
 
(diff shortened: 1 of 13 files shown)


mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
```diff
@@ -40,6 +46,12 @@ public function updateWorkflow(int $id, array $data): Workflow {
     return $workflow;
   }
 
+  private function checkWorkflowName(string $name): void {
```
JanJakes: I think you can just use `Builder::string()->minLength(1)` [here](https://github.com/mailpoet/mailpoet/blob/3bbb662857959307fd895fdabb61bdf89a890f15/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsPutEndpoint.php#L34).

mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
```diff
@@ -27,6 +27,12 @@ public function updateWorkflow(int $id, array $data): Workflow {
       throw Exceptions::workflowNotFound($id);
     }
 
+    if (array_key_exists('name', $data)) {
+      $this->checkWorkflowName($data['name']);
+      $workflow->setName($data['name']);
+      $this->storage->updateWorkflow($workflow);
+    }
```
JanJakes, line 35: Now that we have two IFs here, we should probably extract `$this->storage->updateWorkflow($workflow);` below to run the query only once even in case both fields are updated.

JanJakes (commented):
Looks great! Just two tiny notes.

mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
alex-mailpoet: Done.

mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
alex-mailpoet: Fixed.

JanJakes (approved):

JanJakes:
Merging so I can rebase another branch (discussed with Veljko that this is safe to merge).

PR #4252, opened 2022-07-14
Mitigate fall flag reported for our dev dependencies MAILPOET-4475

samnajian:
Moved composer/composer to dev dependencies to prevent it from finding its was to production bundle.


JanJakes (approved):
Checked the ZIP, looks good!

PR #4253, opened 2022-07-15
Remove old model from the MailPoet\Subscription\Manage class [MAILPOET-4350]

lysyjan:
[MAILPOET-4350]

_How to test_
Verify that the form on the Manage page works properly.

[MAILPOET-4350]: https://mailpoet.atlassian.net/browse/MAILPOET-4350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscription/Manage.php
+++ b/mailpoet/lib/Subscription/Manage.php
@@ -147,8 +147,6 @@ class Manage {
     foreach ($subscriber->getSubscriberSegments() as $subscriberSegment) {
       $segment = $subscriberSegment->getSegment();
       if (!$segment) {
-        $this->subscriberSegmentRepository->remove($subscriberSegment);
-        $subscriber->getSubscriberSegments()->removeElement($subscriberSegment);
         continue;
       }
 
--- a/mailpoet/lib/Subscription/Manage.php
+++ b/mailpoet/lib/Subscription/Manage.php
@@ -2,20 +2,18 @@
 
 namespace MailPoet\Subscription;
 
+use MailPoet\CustomFields\CustomFieldsRepository;
 use MailPoet\Entities\StatisticsUnsubscribeEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Entities\SubscriberSegmentEntity;
 use MailPoet\Form\Util\FieldNameObfuscator;
-use MailPoet\InvalidStateException;
-use MailPoet\Models\CustomField;
-use MailPoet\Models\Subscriber;
-use MailPoet\Models\SubscriberSegment;
 use MailPoet\Newsletter\Scheduler\WelcomeScheduler;
 use MailPoet\Segments\SegmentsRepository;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Statistics\Track\Unsubscribes;
 use MailPoet\Subscribers\LinkTokens;
 use MailPoet\Subscribers\NewSubscriberNotificationMailer;
+use MailPoet\Subscribers\SubscriberSaveController;
 use MailPoet\Subscribers\SubscriberSegmentRepository;
 use MailPoet\Subscribers\SubscribersRepository;
 use MailPoet\Util\Url as UrlHelper;
@@ -43,6 +41,9 @@ class Manage {
   /** @var WelcomeScheduler */
   private $welcomeScheduler;
 
+  /** @var CustomFieldsRepository */
+  private $customFieldsRepository;
+
   /** @var SegmentsRepository */
   private $segmentsRepository;
 
@@ -52,6 +53,9 @@ class Manage {
   /** @var SubscriberSegmentRepository */
   private $subscriberSegmentRepository;
 
+  /** @var SubscriberSaveController */
+  private $subscriberSaveController;
+
   public function __construct(
     UrlHelper $urlHelper,
     FieldNameObfuscator $fieldNameObfuscator,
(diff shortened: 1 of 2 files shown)


mailpoet/lib/Subscription/Manage.php
```diff
@@ -132,34 +135,39 @@ public function onSave() {
...
     // Unsubscribe from all other segments already subscribed to
     // but don't change disallowed segments
-    foreach ($subscriber->subscriptions as $subscription) {
-      $segmentId = $subscription['segment_id'];
-      if ($allowedSegments && !in_array($segmentId, $allowedSegments)) {
+    foreach ($subscriber->getSubscriberSegments() as $subscriberSegment) {
+      $segment = $subscriberSegment->getSegment();
+      if (!$segment) {
+        $this->subscriberSegmentRepository->remove($subscriberSegment);
+        $subscriber->getSubscriberSegments()->removeElement($subscriberSegment);
```
johnolek: Is this necessary? I'm not very familiar with this code so I'm just trying to wrap my head around it. It seems like when we want to use the subscriber segments again we fetch them from the repository down on L168:171.

If the concern is about saving the subscriber with invalid segments it seems like [this code](https://github.com/mailpoet/mailpoet/blob/9867b8c7ad3e822836ebfca8df592c7a29debfc0/mailpoet/lib/Entities/SubscriberEntity.php#L524-L524) should take care of it.

I definitely don't mind having this in here if it's not strictly necessary! It's good to know that `$subscriber->getSubscriberSegments()` will be correct after this `foreach` loop. Again I just want to make sure I understand what's going on.

mailpoet/tests/integration/Subscription/ManageTest.php
```diff
@@ -64,59 +79,65 @@ public function testItDoesntRemoveHiddenSegmentsAndCanResubscribe() {
       $this->settings,
       $this->diContainer->get(NewSubscriberNotificationMailer::class),
       $this->diContainer->get(WelcomeScheduler::class),
+      $this->diContainer->get(CustomFieldsRepository::class),
```
johnolek: I think that this could all be replaced with a helper I added to the codebase a little while ago and I'm just curious to know your opinion on it. With the helper I think it would look like this:

```php
$manage = $this->getServiceWithOverrides(Manage::class, [
  'urlHelper' => Stub::make(UrlHelper::class, [
    'redirectBack' => null,
  ]),
  'fieldNameObfuscator' => Stub::make(FieldNameObfuscator::class, [
    'deobfuscateFormPayload' => function($data) {
      return $data;
    },
  ]),
  'linkTokens' => Stub::make(LinkTokens::class, [
    'verifyToken' => function($token) {
      return true;
    },
  ]),
  'settings' => $this->settings
]);
```

Do you have a preference for one approach vs the other?

Definitely not a required change!

mailpoet/tests/integration/Subscription/ManageTest.php
```diff
@@ -64,59 +79,65 @@ public function testItDoesntRemoveHiddenSegmentsAndCanResubscribe() {
...
-    $subscriber->withSubscriptions();
-    $subscriptions = array_map(function($s) {
-      return ['status' => $s['status'], 'segment_id' => $s['segment_id']];
-    }, $subscriber->subscriptions);
-    usort($subscriptions, function($a, $b) {
-      return $a['segment_id'] - $b['segment_id'];
-    });
-    expect($subscriber->status)->equals(Subscriber::STATUS_SUBSCRIBED);
+    $subscriber = $this->subscribersRepository->findOneById($this->subscriber->getId());
+    if (!$subscriber) throw new InvalidStateException();
```
johnolek: It feels a little strange to use an exception like this in a test. I'm assuming this is for PHPStan?

It would feel a little better to me to use something like `assert($subscriber instanceof SubscriberEntity);` which I think PHPStan understands, but it's not a big deal one way or another.

johnolek (dismissed):
@lysyjan this looks good to me overall! I had just one question about something I wasn't sure I was understanding. I left a couple other comments about minor things, nothing substantial as far as the PR goes.

I'm marking this approved because it looks good from what I can tell, but I'm going to leave it assigned to you in case you'd like to make any changes.

mailpoet/tests/integration/Subscription/ManageTest.php
lysyjan: If I am not wrong, we want to avoid using the `assert` function. We have [a ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4258) for that. But you are right that the exception looks a bit weird here, so I would change it to `$this->assertInstanceOf(SubscriberEntity::class, $subscriber);`. Do you agree with it?

mailpoet/lib/Subscription/Manage.php
lysyjan: Good point. I didn't realize that we have a code that solves it. I will remove it.

mailpoet/tests/integration/Subscription/ManageTest.php
lysyjan: We can simplify the code when it's possible. This pull request isn't big, so I would replace it with your new method.

mailpoet/tests/integration/Subscription/ManageTest.php
johnolek: @lysyjan I was familiar with wanting to remove `assert` from the codebase but I didn't realize that applied to our test code as well. Thanks for pointing that out!

The reason I suggested `assert` is because I tried `expect($subscriber)->isInstanceOf(SubscriberEntity::class);`, which PHPStan didn't seem to understand. If PHPStan does understand `$this->assertInstanceOf(SubscriberEntity::class, $subscriber);` I would definitely prefer that!

johnolek (approved):
Thanks for making those tweaks!

veljkho (approved):

PR #4256, opened 2022-07-19
Improve forms accessibility [MAILPOET-4471]

costasovo:
This PR was [originally submitted](https://github.com/mailpoet/mailpoet/pull/4204) by @simsekb.
I've recreated it because we still have the issue with public builds. I've also added a couple of additional fixes for issues that I missed when doing [the initial review on the original PR](https://github.com/mailpoet/mailpoet/pull/4204).

It changes two things:
1) Improves accessibility of radio buttons, list of segments, and checkboxes by grouping them in a fieldset (see [WAI grouping recommendations](https://www.w3.org/WAI/tutorials/forms/grouping/)). It affects forms that contain a list of segments block, radio buttons custom field block, or checkbox custom field block. It also affects the manage subscription form displayed on the manage subscription page. This PR also adds a CSS reset for fieldsets used inside MailPoets' forms, so the change should not affect the visual appearance of previously created forms.

2) Improves invalid email address error rendering. The error is now rendered next to the field and contains slightly different wording. `Please specify a valid email address.` > `This value should be a valid email.`

[MAILPOET-4471]

[MAILPOET-4471]: https://mailpoet.atlassian.net/browse/MAILPOET-4471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-public/_public.scss
+++ b/mailpoet/assets/css/src/components-public/_public.scss
@@ -67,8 +67,11 @@ $form-line-height: 1.4;
 
 /* Reset fieldset styles in form for backward compatibility. */
 .mailpoet_paragraph {
-  fieldset {
+  fieldset,
+  legend {
+    background: transparent;
     border: 0;
+    color: inherit;
     margin: 0;
     padding: 0;
   }
--- a/mailpoet/assets/css/src/components-public/_public.scss
+++ b/mailpoet/assets/css/src/components-public/_public.scss
@@ -63,6 +63,12 @@ $form-line-height: 1.4;
   .mailpoet-has-font-size {
     line-height: $form-line-height;
   }
+
+  fieldset {
+    border: 0;
+    margin: 0;
+    padding: 0;
+  }
 }
 
 .mailpoet_textarea {
--- a/mailpoet/assets/css/src/components-public/_public.scss
+++ b/mailpoet/assets/css/src/components-public/_public.scss
@@ -63,7 +63,10 @@ $form-line-height: 1.4;
   .mailpoet-has-font-size {
     line-height: $form-line-height;
   }
+}
 
+/* Reset fieldset styles in form for backward compatibility. */
+.mailpoet_paragraph {
   fieldset {
     border: 0;
     margin: 0;
--- a/mailpoet/lib/Form/Block/BlockRendererHelper.php
+++ b/mailpoet/lib/Form/Block/BlockRendererHelper.php
@@ -34,7 +34,7 @@ class BlockRendererHelper {
       $rules['required'] = true;
       $rules['minlength'] = ModelValidator::EMAIL_MIN_LENGTH;
       $rules['maxlength'] = ModelValidator::EMAIL_MAX_LENGTH;
-      $rules['type-message'] = __('This value should be a valid email', 'mailpoet');
+      $rules['type-message'] = __('This value should be a valid email.', 'mailpoet');
     }
 
     if (($blockId === 'first_name') || ($blockId === 'last_name')) {
--- a/mailpoet/lib/Form/Block/BlockRendererHelper.php
+++ b/mailpoet/lib/Form/Block/BlockRendererHelper.php
@@ -153,6 +153,37 @@ class BlockRendererHelper {
     return $html;
   }
 
+  public function renderLegend(array $block, array $formSettings): string {
+    $html = '';
+
+    if (
+      isset($block['params']['hide_label'])
+      && $block['params']['hide_label']
+    ) {
+      return $html;
+    }
+
+    if (
+      isset($block['params']['label'])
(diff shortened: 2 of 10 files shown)


mailpoet/lib/Form/Block/Checkbox.php
```diff
@@ -70,6 +71,8 @@ public function render(array $block, array $formSettings, ?int $formId = null):
       $html .= '</label>';
     }
 
+    echo '</fieldset>';
+
```
triple0t: Was this intentional?

mailpoet/assets/css/src/components-public/_public.scss
```diff
@@ -65,6 +65,15 @@ $form-line-height: 1.4;
   }
 }
 
+/* Reset fieldset styles in form for backward compatibility. */
+.mailpoet_paragraph {
```
triple0t, line 69: Is this class name okay?
This gives the impression that it's a paragraph not a fieldset

triple0t (commented):
Hi @costasovo 
Thanks for recreating the PR

I'm sorry for the delayed response and Apologies would not be able to review it due to my AFK

I took a quick look and it appears there's an error

Steps:
* Create a new post from the editor  i.e /wp-admin/post-new.php
* Add MailPoet subscription form
* Add a form with Checkbox and Radio custom fields (Create a new form if you don't have one)
* Check the console and notice the error



Screenshot 2022-07-22 at 1 03 05 PM

mailpoet/assets/css/src/components-public/_public.scss
costasovo: Yep, it is ok. I agree it is weird, but historically this is a class for wrapping MailPoet form inputs.

mailpoet/lib/Form/Block/Checkbox.php
costasovo: Nope! Good catch this need's to be fixed.

costasovo:
>  I took a quick look and it appears there's an error

Thanks for reporting the error. I think this is not related to the task because none of the changes affect the form editor. Actually, it is all about the front-end rendering. I'll try to replicate it.

mailpoet/assets/css/src/components-public/_public.scss
triple0t: Okay

mailpoet/lib/Form/Block/BlockRendererHelper.php
```diff
@@ -34,7 +34,7 @@ public function getInputValidation(array $block, array $extraRules = [], ?int $f
       $rules['required'] = true;
       $rules['minlength'] = ModelValidator::EMAIL_MIN_LENGTH;
       $rules['maxlength'] = ModelValidator::EMAIL_MAX_LENGTH;
-      $rules['error-message'] = __('Please specify a valid email address.', 'mailpoet');
+      $rules['type-message'] = __('This value should be a valid email', 'mailpoet');
```
lysyjan: I'm not sure about this message. The change makes sense to me, but the text has a slightly different stylization than other messages. But maybe we should address all messages and UX with error messages in a new task. I'm only curious about your opinion.

mailpoet/assets/css/src/components-public/_public.scss
```diff
@@ -63,6 +63,12 @@ $form-line-height: 1.4;
   .mailpoet-has-font-size {
     line-height: $form-line-height;
   }
+
+  fieldset {
```
lysyjan: I tried creating a testing form and found a problematic case. I guess it's caused by my specific WP theme ([Storefront](https://pcm.wordpress.org/themes/storefront/)), but our forms could look bad. Here is [an example](https://d.pr/i/Aj81Nc). In my opinion, we should reset the background color for the fieldset and the legend tag. Do you agree?

lysyjan (commented):
It looks good, but I have two notes about those changes.

mailpoet/assets/css/src/components-public/_public.scss
costasovo: Nice catch! Thank you! I agree.

mailpoet/lib/Form/Block/BlockRendererHelper.php
costasovo: That's a good point about the stylization. On the other hand, I like the new message more :) I don't have a strong opinion here.

lysyjan: Okay. Let's leave it as it is. Thanks

costasovo:
@lysyjan I believe I fixed the issue with Storefront theme.

lysyjan (approved):
👍🏻

veljkho (approved):

PR #4259, opened 2022-07-20
Fix PHPMailer on WP 5.6 in acceptance tests [MAILPOET-4481]

alex-mailpoet:
[MAILPOET-4481]

Use SMTP in PHPMailer on WP 5.6 because its Docker container
has a misconfigured PHP `mail()` function (no sendmail or alternative set up).

[MAILPOET-4481]: https://mailpoet.atlassian.net/browse/MAILPOET-4481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/docker/codeception/docker-entrypoint.sh
+++ b/mailpoet/tests/docker/codeception/docker-entrypoint.sh
@@ -118,6 +118,10 @@ if [[ ! -d "/wp-core/wp-content/plugins/woo-gutenberg-products-block" ]]; then
   unzip -q -o "$WOOCOMMERCE_BLOCKS_ZIP" -d /wp-core/wp-content/plugins/
 fi
 
+# Install a fix plugin for PHPMailer on WP 5.6
+cp /project/tests/docker/codeception/wp-56-phpmailer-fix.php /wp-core/wp-content/plugins/wp-56-phpmailer-fix.php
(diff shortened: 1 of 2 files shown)


JanJakes:
@alex-mailpoet Is the missing configuration in the official WP Docker image? And is it something we could fix in https://github.com/mailpoet/wordpress-images/blob/master/wp-5.6/php7.2/Dockerfile? Just wondering whether the workaround is needed. On the other hand, we may drop WP 5.6 soon, and maybe it works on 5.7; not sure.

alex-mailpoet:
@JanJakes 

> Is the missing configuration in the official WP Docker image? And is it something we could fix in https://github.com/mailpoet/wordpress-images/blob/master/wp-5.6/php7.2/Dockerfile?

Maybe we could fix it in our image, but this workaround is already done. Fixing it the other way would mean starting over and spending more time to fix a flaky test. Or do you think there can be other benefits in fixing the container?

JanJakes (approved):
@alex-mailpoet Makes sense, let's use this. Hopefully, we won't forget to remove this when bumping to WP 5.7.

veljkho (approved):

PR #4262, opened 2022-07-20
pnpm [MAILPOET-4485]

JanJakes:
This implements the switch from NPM to PNPM, sets up workspaces, and uses patch protocol for package modifications.

See: https://mailpoetdev.wordpress.com/?draft=3987
See also related premium and wordpress-images PR.

**Testing:** General smoke tests of everything frontend related. Check as many frontend screens as possible and check that nothing is broken on the frontend part. See also related premium PR.

[MAILPOET-4485]

[MAILPOET-4485]: https://mailpoet.atlassian.net/browse/MAILPOET-4485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.npmrc
```diff
@@ -1 +1,2 @@
+save-workspace-protocol=rolling
```
costasovo, line 1: Probably a dummy question. Maybe I don't understand it fully. Why do we prefer the rolling type?
From the [docs](https://pnpm.io/next/npmrc#save-workspace-protocol), I would say it only affects how it is saved in package.json. I would say the default settings that also save the version look more informative as I can see the version in every package.json.

package.json
```diff
@@ -9,5 +9,8 @@
   "devDependencies": {
     "lint-staged": "^12.3.7",
     "prettier": "2.6.2"
+  },
+  "volta": {
+    "node": "17.9.1"
```
costasovo, line 12: 👍

package.json
```diff
@@ -10,5 +10,10 @@
   "packageManager": "pnpm@7.5.2",
   "volta": {
     "node": "17.9.1"
+  },
+  "pnpm": {
+    "patchedDependencies": {
+      "@woocommerce/components@10.0.0": "patches/@woocommerce__components@10.0.0.patch"
```
costasovo: I learned that we can create these patches using https://pnpm.io/cli/patch and https://pnpm.io/cli/patch-commit. Do you have a link, or could you briefly explain a workflow for updating patch files? e.g. when we update @woocommerce/components@10.0.0 te e.g. 11.0.0, what should we do with the patch file? I guess I'm not the only one who doesn't know, so maybe the explanation or a link to a blog post or so could be helpful in the P2 post you are preparing.

costasovo (dismissed):
Great job! I'm very happy to see this change 👍 

I'm approving the PR, but I left a few questions.

I ran into an issue with Volta, and I googled t[hat it happened to you as well](https://github.com/volta-cli/volta/issues/1200). Can you share what workaround you are using? I have installed the corepack globally using npm, but it doesn't feel like the best solution. I guess using the version bundled with node would be better.

costasovo:
@JanJakes Assigning the PR back to you. There are some conflicts.

.npmrc
JanJakes: @costasovo I think that "rolling" is more intuitive because it sort of tells you that you'll get "whatever is in that workspace". An example:

1. The `mailpoet` workspace requires `eslint-config` as `workspace:1.0.0`.
2. Later on, we make changes to `eslint-config` and bump it to `2.0.0`.
3. Now `mailpoet` in fact uses `2.0.0` — with `workspace:` you always get the latest changes, including working changes, as it's just a symlink.
4. Now `workspace:1.0.0` in `mailpoet` is not really true... In fact, I'm not sure now, maybe `pnpm install` would fail with an error now? I can try it out. But in any way...
5. I think it may be clearer to use a rolling version, such as `workspace:^` to show that it simply takes anything from the `eslint-config` folder.

JanJakes:
> I ran into an issue with Volta, and I googled thttps://github.com/volta-cli/volta/issues/1200. Can you share what workaround you are using? I have installed the corepack globally using npm, but it doesn't feel like the best solution. I guess using the version bundled with node would be better.

Yes, I did. And there doesn't seem to be any fix from their side yet. My workaround was creating these 3 files in `~/.volta/bin`:

1. `~/.volta/bin/corepack`:
    ```
    DIR="$(dirname "$0")"
    NODE_VERSION="$(node --version)"
    "$DIR/../tools/image/node/${NODE_VERSION:1}/bin/corepack" "$@"
    ```
2. `~/.volta/bin/pnpm`:
    ```
    DIR="$(dirname "$0")"
    NODE_VERSION="$(node --version)"
    "$DIR/../tools/image/node/${NODE_VERSION:1}/bin/pnpm" "$@"
    ```
3. `~/.volta/bin/pnpx`:
    ```
    DIR="$(dirname "$0")"
    NODE_VERSION="$(node --version)"
    "$DIR/../tools/image/node/${NODE_VERSION:1}/bin/pnpx" "$@"
    ```

These will pass the respective commands to the Node.js install that's being used.

.npmrc
costasovo: Aha makes sense. In case it keeps outdated values in the "subproject" it seems better to go with rolling.

package.json
JanJakes: I'm not sure there's a workflow for that. The package version needs to be exact, and I think that's because the patched code can change any time, even with a patch fix. Therefore, when updating, you'll probably need to patch again on the current, updated code. Maybe it's also kind of supposed that after an update you often no longer need the patch.

Do you think this is a problem? I'm trying to find what Yarn does, as they've been using the patch protocol for a long time already.

.npmrc
JanJakes: > In case it keeps outdated values in the "subproject" it seems better to go with rolling.

Now I'm not fully sure, if a bump of the required package wouldn't rewrite the version in all packages that require it. I can give it a try.

JanJakes: @costasovo I found the context here: https://github.com/pnpm/pnpm/issues/4887

It seems that `rolling` is better for us.

costasovo: Thank you for finding the issue. So if I understand, the rolling is also more strict.

package.json
costasovo: I don't think it is a problem. The previous patch is a kind of guide for doing the new one so it seems straightforward. I just wanted to know what to do when I do the update.

costasovo (approved):
👍

veljkho (approved):

PR #4265, opened 2022-07-22
Feat/let users authorise their sender domain in the plugin - MAILPOET-4302

triple0t:
[MAILPOET-4302](https://mailpoet.atlassian.net/browse/MAILPOET-4302)

This PR provides the ability for a user to add and verify their sender domain from the plugin

It also includes a basic check for DMARC policy and gives a warning if it detects the domain has restrictive DMARC policies.

Testing:
* from the Settings page. test with unauthorized email addresses and with domains having  DMARC policy set to reject or quarantine. e.g @automattic.com or @mailpoet.com 
* From the Newsletter sending page. perform the same test.
* You should be able to authorized an email address, add a new sender domain and verify the sender domain

For QA:
Please retest this PR: https://github.com/mailpoet/mailpoet/pull/4229

Thank you

Screenshot 2022-07-22 at 10 16 27 AM

Screenshot 2022-07-22 at 10 16 14 AM


mailpoet/tests/integration/Services/BridgeTest.php
```diff
@@ -333,6 +333,167 @@ public function testItReturnsAnEmptyArrayIfNoNullForAuthorizedParam() {
...
+
+    // when param is not provided
+    $returnDataForNoArgs = [
+      'example.com' => $data[0]['dns']
+    ];
+
+    $api = Stub::make(new API(null), ['getAuthorizedSenderDomains' => $data], $this);
+    $this->bridge->api = $api;
+
+    $result = $this->bridge->getAuthorizedSenderDomains('all');
```
costasovo: I guess the parameter should be skipped, as the comment suggests.

mailpoet/tests/integration/Services/BridgeTest.php
```diff
@@ -333,6 +333,167 @@ public function testItReturnsAnEmptyArrayIfNoNullForAuthorizedParam() {
...
+    expect($result)->same($returnDataForNoArgs);
+  }
+
+  public function testItCanCreateSenderDomain() {
+    $apiKey = getenv('WP_TEST_MAILER_MAILPOET_API');
+    if (!$apiKey) {
+      $this->markTestSkipped("Skipping, 'WP_TEST_MAILER_MAILPOET_API' not set.");
+    }
+
+    $api = new API($apiKey, new WPFunctions());
```
costasovo: I think it would be better to mock API responses instead of calling the live API here. I assume the API endpoints are covered by tests on the bridge side.

mailpoet/lib/Util/DmarcPolicyChecker.php
```diff
@@ -0,0 +1,58 @@
+<?php
```
costasovo: We add `declare(strict_types = 1);` to all new PHP files

mailpoet/lib/Util/DmarcPolicyChecker.php
```diff
@@ -0,0 +1,58 @@
...
+
+    foreach ($cache as $value) {
+        $item = explode('=', $value);
+        $dKey = $item[0] ?? '';
+        $dValue = $item[1] ?? '';
+        $dmarcInfo[trim($dKey)] = trim($dValue);
+    }
+
+    // policy can either be reject or quarantine or none
+    $dmarcStatus = $dmarcInfo['sp'] ?? $dmarcInfo['p'] ?? self::POLICY_NONE;
```
costasovo: I would add brackets so that one doesn't have to think about the order when reading the code.

mailpoet/lib/Services/AuthorizedSenderDomainController.php
```diff
@@ -0,0 +1,68 @@
...
+    return $verifiedDomains;
+  }
+
+  /**
+   * Check Domain DMARC Policy
+   *
+   * returns `true` if domain has Retricted policy e.g policy === reject or quarantine
+   * otherwise returns `false`
+   */
+  public function isDomainDmarcRetricted(string $domain): bool {
```
costasovo: There is a small typo: `Retricted` > `Restricted`

mailpoet/lib/AdminPages/Pages/Newsletters.php
```diff
@@ -196,6 +202,8 @@ public function render() {
 
     $data['authorized_emails'] = $this->bridge->getAuthorizedEmailAddresses();
 
+    $data['verified_sender_domains'] = $this->senderDomainController->getVerifiedSenderDomains();
```
costasovo: It seems that we always try those even though the user might not use MSS. In such a case we return `null`, but we still call the API without API key. I think we should check here whether the sending method is MSS.

mailpoet/lib/AdminPages/Pages/Settings.php
```diff
@@ -97,6 +103,7 @@ public function render() {
       ],
       'built_in_captcha_supported' => $this->captcha->isSupported(),
       'authorized_emails' => $this->bridge->getAuthorizedEmailAddresses(),
+      'verified_sender_domains' => $this->senderDomainController->getVerifiedSenderDomains(),
```
costasovo: Same as the above. I think we should fetch verified domains only when a is using MSS.

mailpoet/assets/js/src/settings/pages/basics/default_sender.tsx
```diff
@@ -22,6 +37,36 @@ export function DefaultSender() {
     }
     setIsAuthorized(window.mailpoet_authorized_emails.includes(email));
   };
+
+  const checkSenderEmailDomain = (email: string) => {
+    const emailAddressDomain = email.split('@').pop().toLowerCase();
```
costasovo: I think we need to skip this check when sending method is not MSS.

mailpoet/lib/API/JSON/v1/Settings.php
```diff
@@ -261,13 +261,93 @@ public function checkDomainDmarcPolicy($data = []) {
...
+      ]);
+    }
+
+    $domain = strtolower(trim($domain));
+
+    try {
+      $response = $this->senderDomainController->createAuthorizedSenderDomain($domain);
+    } catch (\InvalidArgumentException $e) {
+      if (
+        $e->getMessage() === AuthorizedSenderDomainController::AUTHORIZED_SENDER_DOMAIN_ERROR_ALREADY_CREATED
```
costasovo, line 301: I think it would be nicer to use exception codes for the logic instead of messages, but it is not a big deal

mailpoet/assets/js/src/common/manage_sender_domain/manage_sender_domain.tsx
```diff
@@ -0,0 +1,100 @@
...
+          value: PropTypes.string,
+          type: PropTypes.string,
+          status: PropTypes.string,
+          message: PropTypes.string,
+        }),
+      ).isRequired,
+    }),
+  ).isRequired,
+  verifyDnsButtonClicked: PropTypes.func,
+};
```
costasovo: For other .tsx components, we only declare types for Props. I guess it is enough. I need to check it though.

costasovo (changes requested):
Wow! What a huge amount of work! I've found some small issues that need addressing.

mailpoet/assets/js/src/common/manage_sender_domain/manage_sender_domain.tsx
costasovo: I checked the component, and it seems that without the prop type checks don't trigger in runtime. So I'll keep the prototypes in this one.

JanJakes: Why do you need to check them at runtime? Do we catch those failures? If not, then I'm not sure if it's useful. Unknown data should be validated and sanitized somewhere, and then TS just can statically check if the rest is good. Duplicating it in PropTypes to make it crash sooner seems unnecessary to me.

costasovo: I'm not sure. Maybe it was handy for the author when developing the API and component at the same time. I'm just guessing. I will remove it.

JanJakes: Ahh, you're finishing this PR, now I get it, you don't know. I guess the only reason may be not everybody knows about this decision to use TS types only for props.

costasovo:
@lysyjan, I've fixed the issues I found during the code review. Well, except [this one](https://github.com/mailpoet/mailpoet/pull/4265#discussion_r929734378) as I considered just a minor issue and I found more things that needed some changes.

It might be worth checking the whole PR.

I've also fixed a couple of issues I found when working on the fixes.

### 1) I've improved the appearance of the errors on the send page
**Before**
The error would probably not fit in case translations were a bit longer.

Screenshot 2022-08-02 at 10 06 25


**After**

Screenshot 2022-08-02 at 10 07 28


### 2) When both errors were displayed, the link from the sender domain error was opening the wrong tab in modal
Now when a user clicks on the link in the authorized email error, the modal opens with the preselected Authorized Emails tab, and when a user clicks on the link in the sender domain error, it opens the sender domain settings tab

### 3) I added skipping checks of domain sender settings for free email addresses
MSS allows using free emails as senders. In those cases, a user can't change DNS (e.g. gmail.com), so we will not display the error.

### 4) A few minor UX/UI issues
* fixed initial width of modal by displaying the initial spinner in columns and unified size of the spinner
* fixed the appearance of copy to clipboard icons in the domain settings table
* added hiding of the sender error when user edits the sender address (we have the same behavior for the auth. email error)

### 5) I added a couple of basic acceptance tests for the new functionality

mailpoet/tests/integration/Util/DmarcPolicyCheckerTest.php
```diff
@@ -0,0 +1,55 @@
...
+
+  public function __construct() {
+    parent::__construct();
+    $this->dmarcPolicyChecker = new DmarcPolicyChecker();
+  }
+
+  public function testItReturnsNoneWhenDomainDoesNotHaveDmarc() {
+    $domain = 'example.com'; // has no TXT records
+    $result = $this->dmarcPolicyChecker->getDomainDmarcPolicy($domain);
+    expect($result)->equals('none');
```
lysyjan: I think that we could use constants from the DmarcPolicyChecker class in those comparisons.

mailpoet/lib/Util/DmarcPolicyChecker.php
```diff
@@ -0,0 +1,64 @@
...
+  const POLICY_REJECT = 'reject';
+  const POLICY_QUARANTINE = 'quarantine';
+
+  /**
+   * Lookup the Domain DMARC Policy
+   * returns reject or quarantine or none
+   */
+  public function getDomainDmarcPolicy(string $domain): string {
+
+    if (!$domain) {
```
lysyjan, line 16: I noticed that this case is not covered in the test, but it is nothing urgent.

mailpoet/lib/Services/Bridge.php
```diff
@@ -139,6 +139,57 @@ public function createAuthorizedEmailAddress(string $emailAdress) {
...
+    return $allSenderDomains;
+  }
+
+  /**
+   * Create a new Sender domain record
+   * returns an Array of DNS response or array of error
+   */
+  public function createAuthorizedSenderDomain(string $domain): array {
+    $data = $this
+    ->getApi($this->settings->get(self::API_KEY_SETTING_NAME))
```
lysyjan, line 179: It seems that the indentation here and in the following method is incorrect.

mailpoet/lib/Services/Bridge.php
```diff
@@ -139,6 +139,57 @@ public function createAuthorizedEmailAddress(string $emailAdress) {
...
+      // return an empty array if the provided domain can not be found
+      return $allSenderDomains[$domain] ?? [];
+    }
+
+    return $allSenderDomains;
+  }
+
+  /**
+   * Create a new Sender domain record
+   * returns an Array of DNS response or array of error
```
lysyjan, line 174: I have no idea how the response can look. An example or a link to the bridge documentation in the comment could be great here because our codebase and team are growing fast. This can be applicated to the other methods too.

mailpoet/lib/Services/Bridge/API.php
```diff
@@ -215,6 +217,89 @@ public function createAuthorizedEmailAddress(string $emailAddress): array {
     return ['status' => $isSuccess];
   }
 
+  /**
+   * Get a list of sender domains
+   * Fetched from API
+   */
+  public function getAuthorizedSenderDomains(): ?array {
```
lysyjan, line 226: This situation with examples is similar to the Bridge class.

mailpoet/lib/Services/Bridge/API.php
```diff
@@ -215,6 +217,89 @@ public function createAuthorizedEmailAddress(string $emailAddress): array {
...
+    $isSuccess = $code === 200;
+
+    if (!$isSuccess) {
+      $logData = [
+        'code' => $code,
+        'error' => is_wp_error($result) ? $result->get_error_message() : $rawResponseBody,
+      ];
+      $this->loggerFactory->getLogger(LoggerFactory::TOPIC_BRIDGE)->error('verifyAuthorizedSenderDomain API call failed.', $logData);
+
+      $fallbackError = sprintf($this->wp->__('An error has happened while performing a request, the server has responded with response code %d'), $code);
```
lysyjan: If I am not wrong, we use the translate function directly and not via our functions wrapper, right? Also, the second parameter with the domain is missing.

mailpoet/lib/Services/Bridge/API.php
```diff
@@ -215,6 +217,89 @@ public function createAuthorizedEmailAddress(string $emailAddress): array {
...
+
+      $errorData = is_array($responseBody) && isset($responseBody['error']) ? $responseBody['error'] : $fallbackError;
+      return ['error' => $errorData, 'status' => false];
+    }
+
+    return is_array($responseBody) ? $responseBody : [];
```
lysyjan: Is it a correct state when responseBody is not an array? Should we log it? TBH, I don't know, so I want to ask only.

mailpoet/tests/integration/Services/BridgeTest.php
```diff
@@ -333,6 +333,167 @@ public function testItReturnsAnEmptyArrayIfNoNullForAuthorizedParam() {
...
+    // when arg param is 'all'
+    $api = Stub::make(new API(null), ['getAuthorizedSenderDomains' => []], $this);
+    $this->bridge->api = $api;
+
+    $result = $this->bridge->getAuthorizedSenderDomains('all');
+    expect($result)->same([]);
+  }
+
+  public function testItReturnsSenderDomainsDnsRecords() {
+    $data = array (
```
lysyjan: Wau, longer array syntax, I haven't seen it for ages in the new code :)

I tried to check the current state after fixes, and this array is duplicated in the code. I guess we can reuse the same array from the mock. Is it possible?

mailpoet/lib/API/JSON/v1/Settings.php
```diff
@@ -246,6 +252,102 @@ public function confirmSenderEmailAddressIsAuthorized($data = []) {
...
+
+    if (!$domain) {
+      return $this->badRequest([
+        APIError::BAD_REQUEST => WPFunctions::get()->__('No sender domain specified.', 'mailpoet'),
+      ]);
+    }
+
+    $domain = strtolower(trim($domain));
+
+    $response = ['isDmarcPolicyRetricted' => $this->senderDomainController->isDomainDmarcRestricted($domain)];
```
lysyjan: Here is another typo in the array key `isDmarcPolicyRetricted`.

mailpoet/lib/API/JSON/v1/Settings.php
```diff
@@ -246,6 +252,22 @@ public function confirmSenderEmailAddressIsAuthorized($data = []) {
...
+
+    if (!$domain) {
+      return $this->badRequest([
+        APIError::BAD_REQUEST => WPFunctions::get()->__('No sender domain specified.', 'mailpoet'),
+      ]);
+    }
+
+    $domain = trim($domain);
+
+    $response = ['isDmarcPolicyRetricted' => $this->senderDomainController->isDomainDmarcRetricted($domain)];
```
lysyjan: Here is another typo in the array key `isDmarcPolicyRetricted`.

mailpoet/lib/API/JSON/v1/Settings.php
```diff
@@ -246,6 +252,22 @@ public function confirmSenderEmailAddressIsAuthorized($data = []) {
     return $this->successResponse($response);
   }
 
+  public function checkDomainDmarcPolicy($data = []) {
+    $domain = $data['domain'] ?? null;
+
+    if (!$domain) {
+      return $this->badRequest([
+        APIError::BAD_REQUEST => WPFunctions::get()->__('No sender domain specified.', 'mailpoet'),
```
lysyjan: We should use the translate function directly. Please, check the whole class here.

mailpoet/tests/integration/Services/AuthorizedSenderDomainControllerTest.php
```diff
@@ -18,41 +19,62 @@ class AuthorizedSenderDomainControllerTest extends \MailPoetTest {
...
   public function testItReturnsVerifiedSenderDomains() {
+    $this->skipTestsIfApiKeyIsMissing();
+
     $controller = $this->getController();
     $verifiedDomains = $controller->getVerifiedSenderDomains();
     expect($verifiedDomains)->same(['mailpoet.com']); // only this is Verified for now
   }
 
+  public function testItReturnsEmptyArrayWhenNoVerifiedSenderDomains() {
+    $expectaton = Expected::once([]); // with empty array
```
lysyjan: I guess this variable should name `expectation` instead of `expectaton`.

mailpoet/tests/integration/Services/AuthorizedSenderDomainControllerTest.php
```diff
@@ -65,8 +87,20 @@ public function testItReturnsFalseWhenDmarcIsNotEnabled() {
     expect($isRetricted)->same(false);
   }
 
-  private function getController(): AuthorizedSenderDomainController {
+  public function testItReturnsDmarcStatus() {
+    $controller = $this->getController();
+    $isRetricted = $controller->getDmarcPolicyForDomain('example.com');
```
lysyjan: Another typo is here in the restricted word.

mailpoet/lib/Services/AuthorizedSenderDomainController.php
```diff
@@ -23,37 +29,101 @@ public function __construct(
     $this->dmarcPolicyChecker = $dmarcPolicyChecker;
   }
 
+  /**
+   * Get the most recent cached record of Bridge::getAuthorizedSenderDomains
+   */
+  public function getDomainRecords($domain = ''): array {
```
lysyjan: We could use a type for the domain argument here.

mailpoet/lib/Services/AuthorizedSenderDomainController.php
```diff
@@ -23,37 +29,101 @@ public function __construct(
...
   /**
    * Get all Authorized Sender Domains
    *
    * Note: This includes both verified and unverified domains
    */
   public function getAllSenderDomains(): array {
-    $records = $this->bridge->getAuthorizedSenderDomains();
-    $domains = array_keys($records);
-    return $domains;
+    $records = $this->currentRecords = $this->bridge->getAuthorizedSenderDomains();
```
lysyjan: Is this correct? Do we override the `currentRecords` properly on each call? If yes, we can simplify this method.
The same logic is in the `getVerifiedSenderDomains` method.

mailpoet/tests/integration/Services/AuthorizedSenderDomainControllerTest.php
```diff
@@ -75,6 +76,75 @@ public function testItReturnsEmptyArrayWhenNoVerifiedSenderDomains() {
...
+    $controller = $this->getController($bridgeMock);
+    $controller->verifyAuthorizedSenderDomain('testdomain.com');
+  }
+
+  public function testVerifyAuthorizedSenderDomainThrowsForVerifiedDomains() {
+    $this->expectException(InvalidArgumentException::class);
+    $this->expectExceptionMessage('Sender domain already verified');
+
+    $domains = ['testdomain.com' => [
+      ['status' => 'valid'],
```
lysyjan, line 134: It's better to use constant here to avoid typos. But only for the next time.

mailpoet/tests/integration/Services/AuthorizedSenderDomainControllerTest.php
```diff
@@ -75,6 +76,75 @@ public function testItReturnsEmptyArrayWhenNoVerifiedSenderDomains() {
...
+    $errorMessage = 'This is a test message';
+    $this->expectException(InvalidArgumentException::class);
+    $this->expectExceptionMessage($errorMessage);
+
+    $domains = ['testdomain.com' => []];
+    $getSenderDomainsExpectaton = Expected::once($domains);
+    $verifySenderDomainsExpectaton = Expected::once(['error' => $errorMessage, 'status' => false]);
+
+    $bridgeMock = $this->make(Bridge::class, [
+      'getAuthorizedSenderDomains' => $getSenderDomainsExpectaton,
```
lysyjan: Another typo in the word `Expectaton`.

mailpoet/assets/js/src/common/check_sender_domain_dmarc_policy.ts
```diff
@@ -0,0 +1,54 @@
...
+    window.mailpoet_verified_sender_domains || []
+  ).includes(emailAddressDomain);
+  if (isDomainVerified) {
+    // do nothing if the email domain is verified
+    return false;
+  }
+
+  try {
+    const res = await makeApiRequest(emailAddressDomain);
+    const isDmarcPolicyRetricted = Boolean(res?.data?.isDmarcPolicyRetricted);
```
lysyjan: Another typo

lysyjan (commented):
Great job with those fixes and improvements!
I have noticed several details. It's up to you if you want to address them. I'm fine with moving this PR forward.

mailpoet/lib/Services/Bridge/API.php
costasovo: I'll add some logging

mailpoet/lib/Services/AuthorizedSenderDomainController.php
costasovo: Yep. This is done because of caching. `$this->currentRecords` is also used in `AuthorizedSenderDomainController::getDomainRecords`. But with this implementation, the method `getDomainRecords` will not work properly if the cache is not prepopulated. I'll refactor it a bit.

costasovo:
@lysyjan I addressed all your remarks [except this one optional](https://github.com/mailpoet/mailpoet/pull/4265#discussion_r937508951).

lysyjan (dismissed):

veljkho (changes requested):
Nice try!
When I enter an unauthorized sender domain, I just can't get the message "Email violates Sender Domain’s DMARC policy. Please set up sender authentication.". This happens only once when using moonbridge.com domain (and I have moonbridge.rs). If I try to use any other domain... I just can't get the possibility to authorize it (even the tab is missing in the popup).

Check this GIF: https://d.pr/i/EGgGqp
My authorized domain is: moonbridge.rs ✅ 
Using moonbridge.com I could get the proper response ✅  (for some reason it works with this domain, it's just a tld change)
Using any other domain, literally anything like moon.com, sdfds.com etc.. no response at all ❌

costasovo:
@veljkho Thanks for checking!

> Using any other domain, literally anything like moon.com, sdfds.com etc.. no response at all ❌

I tested the feature with your examples, and actually, I think this behavior is correct. The ticket is about displaying **a specific warning for domains that have set restrictive DMARC policies**.  In case the domain doesn't have set the restrictive policy, we do not display the warning (see point 4 on [the ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4302)). In your example, the domain `moonbridge.com` has set restrictive DMARC policy (`v=DMARC1;p=quarantine;sp=quarantine;adkim=r;aspf=r;pct=100;fo=0;rf=afrf;ri=86400`) but for domains `moon.com` or `moonbridge.bb` we can't find any DMARC policies at all so that means we treat the the policy as non-restrictive and we don't display the warning.

veljkho (approved):
Verified. I will retest this on release build.

veljkho (approved):

costasovo (approved):

PR #4269, opened 2022-07-26
Remove email signup step from Welcome Wizard [MAILPOET-4497]

lysyjan:
**How to test**
The Welcome Wizard can be raised by deleting a row with the name `version` from the table `wp_mailpoet_settings` or setting its value to `null`.

The wizard contains more alternatives depending on:
- WooCommerce plugin is active
- MailPoet key is empty/set

[MAILPOET-4497]

[MAILPOET-4497]: https://mailpoet.atlassian.net/browse/MAILPOET-4497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/views/welcome_wizard.html
```diff
@@ -4,7 +4,6 @@
 <script>
   var mailpoet_logo_url = '<%= cdn_url('welcome-wizard/mailpoet-logo.20200623.png') %>';
   var wizard_sender_illustration_url = '<%= cdn_url('welcome-wizard/sender.20200623.png') %>';
-  var wizard_email_course_illustration_url = '<%= cdn_url('welcome-wizard/email-course.20200623.png') %>';
```
samnajian, line 7: Was wondering if it makes sense to keep the image(s) in the repo, I guess we can remove them as well, what do you think?

lysyjan: Those images are not a part of the build, so I think we don't delete them. If I'm not wrong, they were uploaded to wordpress.org. You can see quite old images in [the folder](https://github.com/mailpoet/mailpoet/tree/trunk/mailpoet/plugin_repository/assets/welcome-wizard), so I guess that it's fine.

samnajian (approved):
LGTM!

veljkho (approved):

PR #4270, opened 2022-07-26
Add a link to the MailPoet settings page in the page that lists plugins [MAILPOET-4412]

rodrigoprimo:
@websupporter, I'm asking for your review as you mentioned in the ticket that it would be a good idea to create a test for this to prevent the same problem from happening again. Please correct me if I'm wrong, but my understanding is that the "Manage Plugin" button is only displayed when MailPoet is installed in the WP.com environment and we don't have an easy way to test MP using the WP.com interface. Because of that, I'm creating this PR without a test, but I'd be happy to create a test if still needed. Just creating an acceptance test to check that the "Settings" link works seemed unnecessary to me.

Also, I'm unsure if mailpoet/lib/Config/Hooks.php is the best file to add this hook, but it was the best that I could find.

[MAILPOET-4412]

[MAILPOET-4412]: https://mailpoet.atlassian.net/browse/MAILPOET-4412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -91,6 +91,7 @@ class Hooks {
     $this->setupPostNotifications();
     $this->setupWooCommerceSettings();
     $this->setupFooter();
+    $this->setupSettingsLinkInPluginPage();
   }
 
   public function initEarlyHooks() {
@@ -414,5 +415,20 @@ class Hooks {
 
   public function setFooter($text) {
     return '<a href="https://feedback.mailpoet.com/" rel="noopener noreferrer" target="_blank">Give feedback</a>';
+  }
+
+  public function setupSettingsLinkInPluginPage() {
+    $this->wp->addFilter(
+      'plugin_action_links_' . Env::$pluginPath,
+      [$this, 'setSettingsLinkInPluginPage']
+    );
+  }
+
+  public function setSettingsLinkInPluginPage($actionLinks) {
+    $customLinks = [
+      'settings' => '<a href="' . $this->wp->adminUrl('admin.php?page=mailpoet-settings') . '" aria-label="' . esc_attr__( 'View MailPoet settings', 'mailpoet' ) . '">' . esc_html__( 'Settings', 'mailpoet' ) . '</a>',
+    ];
+
+    return array_merge($actionLinks, $customLinks);
   }
 }


websupporter:
Hi @rodrigoprimo and thanks for the PR. Overall lgtm.

With regards to `Hooks`. Imo, it's the best spot for this hook :thinking: 

With regards to testing. My comment about testing in the issue was not about WP.com but WooCommerce. I was thinking of the following acceptance test:

`Given WooCommerce is active and MailPoet is freshly installed`
`Given I am on the "Set up marketing tools" page (test.dev/wp-admin/admin.php?page=wc-admin&task=marketing)`
`When I click Manage`
`Then I should see "Get ready to use MailPoet for WooCommerce"`
 
https://d.pr/v/wm8KWV

With regards to wp.com in my P2 plugin onboarding comment I was just wondering whether the "Manage Plugin" button there acts on the same hook or needs a different configuration.

rodrigoprimo:
Thanks for the additional details, @websupporter.

> My comment about testing in the issue was not about WP.com but WooCommerce. I was thinking of the following acceptance test

It is still not clear to me what is the relation between this PR and WooCommerce. So unless I'm missing something, the test you propose will fail because adding the link to the plugins page does not change the link that WooCommerce generates.

The Jira ticket links to a P2 post where there is a discussion about why users are not directed to the MailPoet settings page when they click on the "Manage plugin" button in WP.com. It was mentioned that WP.com relies on the existence of a Settings link on the plugins page which can be created using the `plugin_action_links_$plugin_name` filter.

WooCommerce is not mentioned, and on a quick check, it seems it doesn't use the same link to define where to direct users when they click on the "Manage" link presented in their dashboard. I believe the URL is hardcoded in WooCommerce code:

https://github.com/woocommerce/woocommerce/blob/a3a5168aeacc483fb378a0fafa331f7252ef179e/plugins/woocommerce/src/Admin/Marketing/InstalledExtensions.php#L273

So if we want to change that, I guess we would have to open an issue or PR in the WooCommerce repository.

websupporter (approved):
Thanks for clarification. lgtm :+1:

veljkho (approved):

PR #4223, opened 2022-07-04
Second pass at deprecating models NewsletterOption and NewsletterOptionField [MAILPOET-4389]

rodrigoprimo:
This PR is the second attempt to deprecate the models NewsletterOption and NewsletterOptionField. The first attempt #4126 had to be reverted in #4135 due to some code that was deprecated but still used in our codebase. We only caught this when #4126 was merged to trunk because it seems the error levels for the acceptance tests seem to differ between CircleCI builds for PRs and the nightly build. This will be addressed in a separate PR (see https://mailpoet.atlassian.net/browse/MAILPOET-4354). More information about the warning that happened when the original PR was merged can be found here: https://app.slack.com/client/T024FN1V2/C01H71ZLBGQ/thread/C01H71ZLBGQ-1652779279.594339

Initially, the plan was to work first on MAILPOET-4354 to consolidate the error levels between different CircleCI builds and then work on this PR. But I spent some time on MAILPOET-4354 and couldn't find anything easily, and I also noticed that the code that was causing the warning for the code changes proposed here was already removed from our code base.

The problem when this PR was introduced the first time was that it deprecated Newsletter::filterType(), but this method was still being used in Scheduler::getNewsletters(). That is not the case anymore, and filterType() was removed from getNewsletters() in https://github.com/mailpoet/mailpoet/commit/4ba5c6ce0cd309319173f8b4132a9f7ecb57f9e6#diff-1d581175bab2c67d713e5e2de493065591c1b93896e1061caf91a2f2c96e451eL70

[MAILPOET-4389]

[MAILPOET-4389]: https://mailpoet.atlassian.net/browse/MAILPOET-4389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4271, opened 2022-07-27
Fix deprecation warnings coming from NewsletterOption model [MAILPOET-4511][MAILPOET-4354]

costasovo:
By [deprecating NewsletterOption and NewsletterOptionField models ](https://github.com/mailpoet/mailpoet/pull/4223)we introduced a couple of deprecation warnings because they were still used in a couple of places.

This was not detected in acceptance tests because there was an issue in test configuration, and WP_DEBUG was disabled.
[WP_DEBUG affects error reporting in WordPress](https://github.com/WordPress/wordpress-develop/blob/6.0/src/wp-includes/load.php#L459-L479).

This PR fixes the deprecation warnings coming from two places in the plugin.  [MAILPOET-4511]

It also fixes the acceptance test configuration issue so that the acceptance test will start catching warnings etc. [MAILPOET-4354]

When I fixed the acceptance tests config, I also needed to fix EditExistingNewsletterCest.

[MAILPOET-4354]  [MAILPOET-4511]

[MAILPOET-4511]: https://mailpoet.atlassian.net/browse/MAILPOET-4511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4354]: https://mailpoet.atlassian.net/browse/MAILPOET-4354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4354]: https://mailpoet.atlassian.net/browse/MAILPOET-4354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4511]: https://mailpoet.atlassian.net/browse/MAILPOET-4511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -144,11 +144,16 @@ class SendingQueue {
       'sending queue processing',
       ['task_id' => $queue->taskId]
     );
+
     $newsletter = $this->newsletterTask->getNewsletterFromQueue($queue);
-    $newsletterEntity = $this->newslettersRepository->findOneById($newsletter->id);
-    if (!$newsletter || !$newsletterEntity) {
+    if (!$newsletter) {
       return;
     }
+    $newsletterEntity = $this->newslettersRepository->findOneById($newsletter->id);
+    if (!$newsletterEntity) {
+      return;
+    }
+
     // pre-process newsletter (render, replace shortcodes/links, etc.)
     $newsletter = $this->newsletterTask->preProcessNewsletter($newsletter, $queue);
     if (!$newsletter) {
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -70,7 +70,7 @@ class SendingQueue {
 
   /** @var ScheduledTasksRepository */
   private $scheduledTasksRepository;
-  
+
   /** @var SubscribersRepository */
   private $subscribersRepository;
 
@@ -145,7 +145,8 @@ class SendingQueue {
       ['task_id' => $queue->taskId]
     );
     $newsletter = $this->newsletterTask->getNewsletterFromQueue($queue);
-    if (!$newsletter) {
+    $newsletterEntity = $this->newslettersRepository->findOneById($newsletter->id);
+    if (!$newsletter || !$newsletterEntity) {
       return;
     }
     // pre-process newsletter (render, replace shortcodes/links, etc.)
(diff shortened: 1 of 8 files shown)


mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -145,7 +145,8 @@ private function processSending(SendingTask $queue, int $timer): void {
       ['task_id' => $queue->taskId]
     );
     $newsletter = $this->newsletterTask->getNewsletterFromQueue($queue);
-    if (!$newsletter) {
+    $newsletterEntity = $this->newslettersRepository->findOneById($newsletter->id);
```
JanJakes, line 152: This seems like this part needs to stay above:
```php
if (!$newsletter) {
  return;
}
```

Otherwise `$newsletter->id` won't exist.

mailpoet/tests/docker/codeception/docker-entrypoint.sh
```diff
@@ -129,15 +129,12 @@ wp plugin activate woocommerce-memberships
...
-CONFIG+="define('COOKIE_DOMAIN', \$_SERVER['HTTP_HOST']);\n"
-CONFIG+="define('DISABLE_WP_CRON', true);\n"
-
-sed -i "s/define( *'WP_DEBUG', !!getenv_docker('WORDPRESS_DEBUG', '')* );/$CONFIG/" /wp-core/wp-config.php
+# Set constants in wp-config.php
+wp config set WP_DEBUG true --raw
+wp config set WP_DEBUG_DISPLAY true --raw
+wp config set WP_DEBUG_LOG true --raw
+wp config set COOKIE_DOMAIN \$_SERVER[\'HTTP_HOST\'] --raw
+wp config set DISABLE_WP_CRON true --raw
```
JanJakes, line 137: Is this doable in one command to make it faster? Not important to change now though.

JanJakes (commented):
Great stuff! I found one slightly suspicious line.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
costasovo: 🤦 Thanks for catching it!

mailpoet/tests/docker/codeception/docker-entrypoint.sh
costasovo: If I understand correctly the command can be used only for one constant https://developer.wordpress.org/cli/commands/config/set/
Here is [the list of all config-related commands](https://developer.wordpress.org/cli/commands/config/) and there doesn't seem to be any batch set.

JanJakes (approved):

veljkho (approved):

PR #4272, opened 2022-07-27
Replace Paris models with Doctrine entities in InactiveSubscribers [MAILPOET-4344]

rodrigoprimo:
[MAILPOET-4344]

[MAILPOET-4344]: https://mailpoet.atlassian.net/browse/MAILPOET-4344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/InactiveSubscribers.php
+++ b/mailpoet/lib/Cron/Workers/InactiveSubscribers.php
@@ -3,10 +3,11 @@
 namespace MailPoet\Cron\Workers;
 
 use MailPoet\Entities\ScheduledTaskEntity;
-use MailPoet\Models\Subscriber;
+use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Settings\TrackingConfig;
 use MailPoet\Subscribers\InactiveSubscribersController;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class InactiveSubscribers extends SimpleWorker {
   const TASK_TYPE = 'inactive_subscribers';
@@ -22,14 +23,19 @@ class InactiveSubscribers extends SimpleWorker {
   /** @var TrackingConfig */
   private $trackingConfig;
 
+  /** @var EntityManager */
+  private $entityManager;
+
   public function __construct(
     InactiveSubscribersController $inactiveSubscribersController,
     SettingsController $settings,
-    TrackingConfig $trackingConfig
+    TrackingConfig $trackingConfig,
+    EntityManager $entityManager
   ) {
     $this->inactiveSubscribersController = $inactiveSubscribersController;
     $this->settings = $settings;
     $this->trackingConfig = $trackingConfig;
+    $this->entityManager = $entityManager;
     parent::__construct();
   }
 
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Cron/Workers/InactiveSubscribers.php
```diff
@@ -48,7 +54,17 @@ public function processTaskStrategy(ScheduledTaskEntity $task, $timer) {
...
+
+    if (isset($meta['max_subscriber_id'])) {
+      $maxSubscriberId = $meta['max_subscriber_id'];
+    } else {
+      $maxSubscriberId = $this->entityManager->createQueryBuilder()
+        ->select('MAX(s.id)')
+        ->from(SubscriberEntity::class, 's')
+        ->getQuery()
+        ->getSingleScalarResult();
+    }
```
johnolek: Would it be more appropriate to move this code into `\MailPoet\Subscribers\SubscribersRepository` and then `use` just the repository instead of the subscriber entity and entity manager?

I honestly don't know because I haven't worked much on doctrine myself, but it looks like most similar code appears in repository classes.

johnolek (changes requested):
@rodrigoprimo I added one question about where we want the query code to live.

Also the ticket says this:
> Because this task is very small and doesn’t have an integration test, please remove old models usages from the `SimpleWorkerTest`

but I don't see any changes to that file. Is that going to be a separate PR, or is there a reason that test doesn't need to be changed?

rodrigoprimo:
Thanks for the review, @johnolek.

> Would it be more appropriate to move this code into \MailPoet\Subscribers\SubscribersRepository and then use just the repository instead of the subscriber entity and entity manager?
> I honestly don't know because I haven't worked much on doctrine myself, but it looks like most similar code appears in repository classes.

Good question. I don't have much experience with Doctrine as well. It makes sense to me to keep the queries on the repository classes, so I added a new commit to implement this. But I don't know what is the best practice according to Doctrine. I noticed that in our codebase we don't call EntityManager::createQueryBuilder() exclusively inside the repository classes, but maybe we should? cc @lysyjan in case you have any input here, as I believe you have more experience with Doctrine than we do.

> but I don't see any changes to that file. Is that going to be a separate PR, or is there a reason that test doesn't need to be changed?

Thanks for flagging this. I saw it when reading the ticket description but then completely forgot about it. I added a new commit for SimpleWorkerTest.

I also added a third commit to improve InactiveSubscribersTest.php by using \MailPoetTest::getServiceWithOverrides() instead of instantiating InactiveSubscribers directly in multiple tests.

When you have a moment, could you please check this PR again?

johnolek (approved):
Thanks for making those updates, @rodrigoprimo! This looks good to me now and I'm approving, but I'm going to leave it up to you to move it to QA, especially since we might want to get someone like @lysyjan's input first.

rodrigoprimo:
Thanks, I will go ahead and move this PR to QA. We can always revisit the changes proposed here if needed once we get input from @lysyjan.

lysyjan:
The first input from @johnolek about moving the query builder creation to the repository seems reasonable. But, of course, it's nothing critical.

veljkho (approved):

PR #4275, opened 2022-07-29
Delay settings [MAILPOET-4418]

websupporter:
Fixes [MAILPOET-4418]

This PR introduces the delay settings.

The validation of the settings happens currently only in the run. I haven't introduced error handling in the UI yet, as there seems to be too many moving parts where and how we want to set those messages up and which buttons to disable etc.

An integration test has been added, to validate, the boundaries are respected during the workflow run.

This PR also introduces a `<ColoredIcon />` component to allow for the colored icons in the workflow view.

[MAILPOET-4418]: https://mailpoet.atlassian.net/browse/MAILPOET-4418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/editor/store/selectors.ts
```diff
@@ -47,6 +47,10 @@ export function getSelectedStep(state: State): Step | undefined {
   return state.selectedStep;
 }
 
+export function getStepType(state: State, type: string): StepType | undefined {
```
JanJakes: Isn't the second parameter rather `key` than `type`? Just thinking about naming here.

mailpoet/assets/js/src/automation/editor/store/types.ts
```diff
@@ -12,7 +12,8 @@ export type StepType = {
   group: StepGroup;
   title: string;
   description: string;
-  icon: ComponentType;
+  subtitle: (step: Step) => string;
```
JanJakes: Nice idea with allowing to modify the subtitle like this. One possible improvement (maybe) — could we allow it to return `JSX.Element` so that the subtitles can contain other nodes than just text as well?

mailpoet/assets/js/src/automation/integrations/core/steps/delay/edit.tsx
```diff
@@ -0,0 +1,56 @@
...
+            label=""
+            placeholder="Number"
+            value={(selectedStep.args.delay as string) ?? ''}
+            onChange={(rawValue) => {
+              const value: number =
+                rawValue.length === 0 || parseInt(rawValue, 10) < 1
+                  ? 1
+                  : parseInt(rawValue, 10);
+              dispatch(store).updateStepArgs(selectedStep.id, 'delay', value);
+            }}
```
JanJakes, line 37: What if we give it `type="number"` instead? It should then handle the data type, and we'll probably only need to check `> 0`. It will also render like this when focused:

Screen Shot 2022-07-29 at 8 54 40

mailpoet/assets/js/src/automation/testing.tsx
```diff
@@ -17,7 +17,8 @@ const createDelayStep = (nextStepId: string) => ({
   key: 'core:delay',
   next_step_id: nextStepId,
   args: {
-    seconds: 60,
+    delay: 1,
+    delay_type: 'HOURS',
```
JanJakes, line 21: I see you later convert this to seconds in PHP.

I'm thinking... What if we always save seconds already in the `args` field and let the component "decide" when to render hours, when weeks, etc.? I think it may slightly simplify the data type (and maybe its validation as well) and could give us more freedom changing the UI... or are there any issues with that approach?

(Oh, but I see now that would mean moving also the tests to JS 😬)

mailpoet/assets/js/src/automation/integrations/core/steps/delay/types/delayTypes.ts
```diff
@@ -0,0 +1,26 @@
...
+    plural: 'days',
+    value: 'DAYS',
+  },
+  {
+    label: 'Weeks',
+    singular: 'week',
+    plural: 'weeks',
+    value: 'WEEKS',
+  },
+];
```
JanJakes, line 25: We didn't solve translations yet, so this may just be a reflection for the future, but I'm thinking about making these translatable... Do you think we could use the [Intl API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) instead of translating all these forms (singular, plural, etc.)? After all, this won't work for Czech, for instance, as we have multiple plurals.

mailpoet/assets/css/src/components-automation-editor/_step.scss
```diff
@@ -33,3 +33,19 @@
 .mailpoet-automation-editor-step-subtitle {
   color: inherit;
 }
+
+.colored-icon {
```
JanJakes: I think we should add the `mailpoet-automation` prefix (we seem to do it in other places across the MailPoet codebase).

mailpoet/assets/js/src/automation/editor/components/icons/colored-icon.tsx
```diff
@@ -0,0 +1,26 @@
...
+      }}
+    >
+      <div
+        className="colored-icon-background"
+        style={{ backgroundColor: props.color }}
+      />
+      <Icon {...props} />
+    </div>
+  );
+}
```
JanJakes, line 30: I'll leave the comment here, but it may belong rather to the modified SVGs or the CSS of this component, I'm not sure. The issue is the icons seem to have wrong sizes now:

In the workflow, the `email` icon seems to be too small, compared to Figma, and the `delay` icon is somehow broken:
Screen Shot 2022-07-29 at 9 07 17

The correct should be:
Screen Shot 2022-07-29 at 9 10 42

Are also the background colors wrong?

In the sidebar, the delay icon seems to be too big now:
Screen Shot 2022-07-29 at 9 07 37

And as mentioned, the `delay` icon seems to be somehow damaged, it has even a different shape. Did you try to compress it somehow? Please let's revert such changes and only add necessary fixes for the colors, if possible.

mailpoet/assets/css/src/components-automation-editor/_step.scss
```diff
@@ -33,3 +33,19 @@
...
+  box-sizing: content-box;
```
JanJakes, line 39: I'm mentioning a problem with sizing below in the JS component. Not sure if the `content-box` sizing is related.

mailpoet/assets/css/src/components-automation-editor/_step.scss
```diff
@@ -33,3 +33,19 @@
...
+  padding: 10px;
+  position: relative;
+
+  .colored-icon-background {
+    border-radius: 50%;
+    height: 100%;
+    left: 0;
+    opacity: .1;
```
JanJakes: Ahh, this probably is why the email icon background has a different color than in Figma. I think we should keep full opacity to reflect the exact used color, if possible. We'll need two colors (foreground and background).

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/index.tsx
```diff
@@ -7,6 +7,8 @@ export const step: StepType = {
   group: 'actions',
   title: 'Send email',
   description: 'An email will be sent to subscriber',
-  icon: Icon,
+  subtitle: (data) => `Email ID ${data.args.email_id as string}`,
+  color: '#996800',
+  icon: Icon(),
```
JanJakes: I'm wondering why we now need to call the component function here. I'm a bit surprised it works as we're not in a React tree here yet. That's why I think we should pass component functions (like before) or render callbacks (like for the subtitle). I see this works now but still looks slightly suspicious to me.

mailpoet/assets/js/src/automation/integrations/core/steps/delay/edit.tsx
```diff
@@ -0,0 +1,56 @@
...
+      selectedStep: select(store).getSelectedStep(),
+    }),
+    [],
+  );
+
+  return (
+    <PanelBody opened>
+      <PlainBodyTitle title="Wait for" />
+      <Flex align="top">
+        <FlexItem style={{ flex: '1 1 0' }}>
```
JanJakes, line 25: Inline styles... I guess in this case, it's acceptable. Just thinking out loud. I'm surprisedWP components have such primitives, TIL, and grid seems to be experimental, as far as I can see.

Using a single parent `<div>` for the `<TextControl>` elements with a bit of CSS (display `grid` and `gap`, if needed) would simplify it a bit. But not important...

JanJakes (commented):
Great! Happy to see this moving!

I left a bunch of comments, but consider most of them just ideas/reflections, and feel free to address them or not. The most important is the icons seem to be broken a bit, and I'd use `type="number"` for the delay input.

mailpoet/assets/js/src/automation/editor/store/selectors.ts
websupporter: Addressed in [92bcdc9](https://github.com/mailpoet/mailpoet/pull/4275/commits/92bcdc93954ed8e9e5f6786750c6e68f86c696dc)

mailpoet/assets/js/src/automation/editor/store/types.ts
websupporter: :heart: Addressed in [536b582](https://github.com/mailpoet/mailpoet/pull/4275/commits/536b5828276e3788de0bc3e8cf739f694a833d0b)

mailpoet/assets/js/src/automation/integrations/core/steps/delay/edit.tsx
websupporter: :+1: Addressed in [5cec0ab](https://github.com/mailpoet/mailpoet/pull/4275/commits/5cec0abfa5e855ebd9cdac333e9b5fd3e07e7693)

mailpoet/assets/js/src/automation/testing.tsx
websupporter: So, when I would have 3600 seconds it would display 1 hour?

My problem would be: And if someone types in 7 days it would switch to 1 week? That likely feels strange and could lead to wrong inputs, when someone who typed in 7 days sees suddenly a 1 in the input field (without seeing the day switched to week) and changes it back to 7.

mailpoet/assets/css/src/components-automation-editor/_step.scss
websupporter: Addressed in [b455854](https://github.com/mailpoet/mailpoet/pull/4275/commits/b455854cda6b723906a9272563811b23dc0b2f4b)

mailpoet/assets/css/src/components-automation-editor/_step.scss
websupporter: Main commit to address this is [d403d7a](https://github.com/mailpoet/mailpoet/pull/4275/commits/d403d7ae91f2c6db2b81fbd35d11df307b507fa5)

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/index.tsx
websupporter: That was an odd one. I switched to @wordpress/components for the icon and was able to get away from `JSX.Element` and back to `ComponentType` ([a3893c4](https://github.com/mailpoet/mailpoet/pull/4275/commits/a3893c46d247eaf9842cdd1b27c63a419e9b6722))

mailpoet/assets/js/src/automation/editor/components/icons/colored-icon.tsx
websupporter: Icon switched back and some other improvements done.

mailpoet/assets/js/src/automation/integrations/core/steps/delay/types/delayTypes.ts
websupporter: :thinking: 
Did look into the Intl API, but not sure how the API could help us here. I propose for now to change the DateType object and add a `subtitle(value:number) => string` callback function there. Once we come to internationalization, we can pick it up from there I guess. [ced0355](https://github.com/mailpoet/mailpoet/pull/4275/commits/ced0355362c9dfbe22bf005c157f247da9e3076f)

websupporter:
Thanks for your feedback @JanJakes. I addressed most of the issues.

mailpoet/assets/js/src/automation/testing.tsx
JanJakes: Ah, right. We need to store it to display the UI correctly 👍

mailpoet/assets/js/src/automation/integrations/core/steps/delay/types/delayTypes.ts
JanJakes: @websupporter What I meant is the number format Intl API. Try this in your browser console:

```js
(5).toLocaleString('de', {
  style: 'unit',
  unit: 'second',
  unitDisplay: 'long',
});

// '5 Sekunden'
```

If we pass the language from MailPoet then this would solve it nicely, I guess. What do you think?

JanJakes (commented):
Looks good!

We may unify all of the `JSX.Element` vs. `ComponentType` vs. `ReactNode` vs. `string` later on, not sure which one will be the best fit. Let me know if you want to check the [Intl API as suggested here](https://github.com/mailpoet/mailpoet/pull/4275/files#r933206842) or whether I should hit approve.

mailpoet/assets/js/src/automation/integrations/core/steps/delay/types/delayTypes.ts
websupporter: I hesitate currently. As of know, we would need to hard code `'en'` anyway, because otherwise we get an English string with Czech words in it.

What I like about it, we can reduce now the "knowledge" of the `DelayType` to just the JavaScript unit and we do not need to move all texts into the `DelayType`. What I worry about is: I am not sure if there are strings in some languages where `5 hours` might be separated. What I mean: "Wait for `5 hours`" might correspond in another language to  a string like "The amount of `hours` you have to wait is `5`"

With using a JavaScript function we delegate a part of the translation to the domain of the browser and invite inconsistencies between browsers and other issues (Chrome does not yet support `locale X`, Firefox `locale Y`), we do not want to have and `_n()` could save us the headache I suppose.

I would suggest to return to this topic, when we have decided on translations @JanJakes

JanJakes: @websupporter OK, let's keep it as it is now. The `Intl` API is amazing if you need to display the time unit alone (without being embedded in a sentence) or a relative date for comments, etc. but it's true we have a slightly different case.

JanJakes (dismissed):

veljkho (approved):

veljkho (changes requested):
Nice try, one small UI issue found:

- If I enter any value in the empty field.. I won't see any label change in the workflow's badge for the Delay... only if I change Hours to something else I see the change. I would expect instantly if I enter any value in the empty field for numbers to see the `Delay for 3 hours` for example... [check this GIF](https://d.pr/i/gErS4f).

websupporter:
@JanJakes to account for the problem @veljkho reported I currently extend the `Templates\WorkflowBuilder` (which I missed to update after the new configuration values) in [50aa481](https://github.com/mailpoet/mailpoet/pull/4275/commits/50aa481df0b2167bb1e8dd8d947ec7e0b1f8d25a) to allow for configuring defaults in the template builder.

We do not yet have something like default values and a store action like `insertStep`. To create default values, I guess this would happen in such an action and would be handled, once we come to extending workflows?

I would imagine something like the following to be needed?
```
export type StepType = {
  key: string;
  group: StepGroup;
  title: string;
  ....

  configuration: {
    key: {
      default: "something",
      ...
    }
};
```

JanJakes (approved):
@websupporter I was thinking about the args and defaults quite a bit recently and I looked into Gutenberg attributes to align with. What Gutenberg does is they define a schema that is used in both PHP and JS and that led me to a solution I implemented in https://github.com/mailpoet/mailpoet/pull/4276/commits/352ce0b7b5f99fe4eabef7c52e71268a852f2e85. For now this is only the PHP part but, similarly to Gutenberg, we'll serialize all of this and make it available in JS as well. This is how I used the defaults in templates: https://github.com/mailpoet/mailpoet/pull/4276/commits/315e64d4f289c5648e7e6b83eb6a2f2448701f9d

For this branch, let's pass it on as you've done it and then I can rebase my branch on that. Next steps for args will be the JS part and their actual validation.

veljkho (approved):

PR #4276, opened 2022-07-29
Connect automation email with send email step [MAILPOET-4515]

JanJakes:
[MAILPOET-4515]

[MAILPOET-4515]: https://mailpoet.atlassian.net/browse/MAILPOET-4515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

To test, you need to deactivate and activate the plugin after checking out this branch. If installed using the ZIP build, the deactivation and activation should not be needed, I guess.

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
@@ -8,9 +8,10 @@ import { MailPoet } from '../../../../../mailpoet';
 export function DesignEmailButton(): JSX.Element {
   const [isSaving, setIsSaving] = useState(false);
 
-  const { selectedStep, workflowSaved } = useSelect(
+  const { selectedStep, workflowId, workflowSaved } = useSelect(
     (select) => ({
       selectedStep: select(store).getSelectedStep(),
+      workflowId: select(store).getWorkflowData().id,
       workflowSaved: select(store).getSelectedStep(),
     }),
     [],
@@ -27,6 +28,9 @@ export function DesignEmailButton(): JSX.Element {
       data: {
         type: 'automation',
         subject: '',
+        options: {
+          workflowId,
+        },
       },
     });
 
@@ -37,7 +41,7 @@ export function DesignEmailButton(): JSX.Element {
     );
 
     dispatch(store).save();
-  }, [selectedStep.id]);
+  }, [workflowId, selectedStep.id]);
 
   // This component is rendered only when no email ID is set. Once we have the ID
   // and the workflow is saved, we can safely redirect to the email design flow.
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
@@ -1,0 +1,62 @@
+import { useCallback, useEffect, useState } from 'react';
+import { dispatch, useSelect } from '@wordpress/data';
+import { plus } from '@wordpress/icons';
+import { Button } from '../../components/button';
+import { store } from '../../../../editor/store';
+import { MailPoet } from '../../../../../mailpoet';
+
+export function DesignEmailButton(): JSX.Element {
+  const [isSaving, setIsSaving] = useState(false);
+
+  const { selectedStep, workflowSaved } = useSelect(
+    (select) => ({
+      selectedStep: select(store).getSelectedStep(),
+      workflowSaved: select(store).getSelectedStep(),
+    }),
+    [],
+  );
+
+  const emailId = selectedStep?.args?.email_id as string | undefined;
+
+  const createEmail = useCallback(async () => {
+    setIsSaving(true);
+    const response = await MailPoet.Ajax.post({
+      api_version: window.mailpoet_api_version,
+      endpoint: 'newsletters',
+      action: 'create',
+      data: {
+        type: 'automation',
+        subject: '',
+      },
+    });
+
+    dispatch(store).updateStepArgs(
+      selectedStep.id,
+      'email_id',
+      parseInt(response.data.id as string, 10),
+    );
+
+    dispatch(store).save();
+  }, [selectedStep.id]);
+
+  // This component is rendered only when no email ID is set. Once we have the ID
+  // and the workflow is saved, we can safely redirect to the email design flow.
+  useEffect(() => {
+    if (emailId && workflowSaved) {
+      window.location.href = `admin.php?page=mailpoet-newsletter-editor&id=${emailId}`;
+    }
+  }, [emailId, workflowSaved]);
+
+  return (
+    <Button
+      variant="sidebar-primary"
+      centered
+      icon={plus}
+      onClick={createEmail}
+      isBusy={isSaving}
+      disabled={isSaving}
+    >
+      Design email
+    </Button>
+  );
+}
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
@@ -18,6 +18,7 @@ export function DesignEmailButton(): JSX.Element {
   );
 
   const emailId = selectedStep?.args?.email_id as string | undefined;
+  const workflowStepId = selectedStep.id;
 
   const createEmail = useCallback(async () => {
     setIsSaving(true);
@@ -30,18 +31,19 @@ export function DesignEmailButton(): JSX.Element {
         subject: '',
         options: {
           workflowId,
+          workflowStepId,
         },
       },
     });
 
     dispatch(store).updateStepArgs(
-      selectedStep.id,
+      workflowStepId,
       'email_id',
       parseInt(response.data.id as string, 10),
     );
 
     dispatch(store).save();
-  }, [workflowId, selectedStep.id]);
+  }, [workflowId, workflowStepId]);
 
   // This component is rendered only when no email ID is set. Once we have the ID
   // and the workflow is saved, we can safely redirect to the email design flow.
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
@@ -4,8 +4,19 @@ namespace MailPoet\Automation\Engine\Builder;
 
 use MailPoet\Automation\Engine\Data\Step;
 use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Registry;
 
 class UpdateStepsController {
+  /** @var Registry */
+  private $registry;
+
+  public function __construct(
+    Registry $registry
+  ) {
+    $this->registry = $registry;
+  }
+
   public function updateSteps(Workflow $workflow, array $data): Workflow {
     $steps = [];
     foreach ($data as $stepData) {
@@ -17,6 +28,11 @@ class UpdateStepsController {
   }
 
   private function processStep(array $data): Step {
+    $key = $data['key'];
+    $step = $this->registry->getStep($key);
+    if (!$step) {
+      throw Exceptions::workflowStepNotFound($key);
+    }
     return new Step(
       $data['id'],
       $data['type'],
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
@@ -5,21 +5,15 @@ namespace MailPoet\Automation\Engine\Builder;
 use MailPoet\Automation\Engine\Data\Step;
 use MailPoet\Automation\Engine\Data\Workflow;
 use MailPoet\Automation\Engine\Exceptions;
-use MailPoet\Automation\Engine\Hooks;
 use MailPoet\Automation\Engine\Registry;
 
 class UpdateStepsController {
-  /** @var Hooks */
-  private $hooks;
-
   /** @var Registry */
   private $registry;
 
   public function __construct(
-    Hooks $hooks,
     Registry $registry
   ) {
-    $this->hooks = $hooks;
     $this->registry = $registry;
   }
 
@@ -40,15 +34,12 @@ class UpdateStepsController {
       throw Exceptions::workflowStepNotFound($key);
     }
 
-    $stepData = new Step(
+    return new Step(
       $data['id'],
       $data['type'],
       $data['key'],
       $data['next_step_id'] ?? null,
       $data['args'] ?? null
     );
-
-    $this->hooks->doWorkflowStepBeforeSave($stepData);
-    return $stepData;
   }
 }
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Builder;
+
+use MailPoet\Automation\Engine\Workflows\Step;
+use MailPoet\Automation\Engine\Workflows\Workflow;
+
+class UpdateStepsController {
+  public function updateSteps(Workflow $workflow, array $data): Workflow {
+    $steps = [];
+    foreach ($data as $stepData) {
+      $step = $this->processStep($stepData);
+      $steps[$step->getId()] = $step;
+    }
+    $workflow->setSteps($steps);
+    return $workflow;
+  }
+
+  private function processStep(array $data): Step {
+    return new Step(
+      $data['id'],
+      $data['type'],
+      $data['key'],
+      $data['next_step_id'] ?? null,
+      $data['args'] ?? null
+    );
+  }
+}
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
@@ -5,15 +5,21 @@ namespace MailPoet\Automation\Engine\Builder;
 use MailPoet\Automation\Engine\Data\Step;
 use MailPoet\Automation\Engine\Data\Workflow;
 use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Hooks;
 use MailPoet\Automation\Engine\Registry;
 
 class UpdateStepsController {
+  /** @var Hooks */
+  private $hooks;
+
   /** @var Registry */
   private $registry;
 
   public function __construct(
+    Hooks $hooks,
     Registry $registry
   ) {
+    $this->hooks = $hooks;
     $this->registry = $registry;
   }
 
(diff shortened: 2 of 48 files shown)


mailpoet/lib/Automation/Engine/Registry.php
```diff
@@ -33,11 +37,31 @@ public function getSubjects(): array {
...
 
+  public function addStep(Step $step): void {
+    if ($step instanceof Trigger) {
+      $this->addTrigger($step);
+    } elseif ($step instanceof Action) {
+      $this->addAction($step);
+    }
+
+    // TODO: allow adding any other step implementations?
+  }
```
johnolek, line 57: Is this getting used anywhere right now? I think I'm still seeing calls directly to `addAction` or `addTrigger` instead and I'm not sure I understand how this method is/should be getting used.

mailpoet/lib/Automation/Engine/Hooks.php
```diff
@@ -2,10 +2,33 @@
...
+  }
+
   public const INITIALIZE = 'mailpoet/automation/initialize';
   public const API_INITIALIZE = 'mailpoet/automation/api/initialize';
   public const STEP_RUNNER_INITIALIZE = 'mailpoet/automation/step_runner/initialize';
   public const TRIGGER = 'mailpoet/automation/trigger';
   public const WORKFLOW_STEP = 'mailpoet/automation/workflow/step';
+
+  public const WORKFLOW_BEFORE_SAVE = 'mailpoet/automation/workflow/before_save';
+  public const WORKFLOW_STEP_BEFORE_SAVE = 'mailpoet/automation/workflow/step/before_save';
```
johnolek, line 25: As I was looking at these names, I was thinking to myself, "Would it be useful to append `_ACTION` or include `action` in the string somehow, like `mailpoet/automation/actions/workflow/step/before_save`, but then I realized that all of the hooks here are actions. It would also probably be excessive anyway.

That having been said, maybe it would make sense to name this class `Actions` instead of `Hooks`? I'm not sure if we're intentionally avoiding filters, but if we do want them in the future they could go into their own class.

This is very minor, but I think it might be just a tiny bit clearer for other developers who might want to use these actions.

mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
```diff
@@ -52,6 +58,7 @@ public function updateWorkflow(int $id, array $data): Workflow {
     }
 
     if ($changed) {
+      $this->hooks->doWorkflowBeforeSave($workflow);
```
johnolek, line 65: Is there a reason you don't want to have this inside `\MailPoet\Automation\Engine\Storage\WorkflowStorage::updateWorkflow` directly?

mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
```diff
@@ -33,12 +39,16 @@ private function processStep(array $data): Step {
...
+
+    $stepData = new Step(
       $data['id'],
       $data['type'],
       $data['key'],
       $data['next_step_id'] ?? null,
       $data['args'] ?? null
     );
+
+    $this->hooks->doWorkflowStepBeforeSave($stepData);
```
johnolek: This feels like a potentially misleading name. `processStep` is only called by `updateSteps`, which calls `$workflow->setSteps($steps);`, which is ultimately only changing the workflow steps in memory. It doesn't seem like a guarantee that `updateSteps` will definitely lead to a save, other than of course the fact that right now we [always save workflows after calling this method](https://github.com/mailpoet/mailpoet/blob/315e64d4f289c5648e7e6b83eb6a2f2448701f9d/mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php#L54-L63).

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -101,6 +101,24 @@ public function run(Workflow $workflow, WorkflowRun $workflowRun, Step $step): v
     }
   }
 
+  public function saveEmailSettings(Step $step): void {
+    if ($step->getKey() !== $this->getKey()) {
+      return;
+    }
```
johnolek: This might be a terrible idea, but one way we could avoid having to write these checks would be to create dynamic actions instead, where we insert the key as past of the action. We could even define it as a second action to maintain the ability to write callbacks that we want to run against every single step.

```php
  public function doWorkflowStepBeforeSave(Step $step): void {
    $this->wordPress->doAction('mailpoet/automation/workflow/step/before_save', $step);
    $this->wordPress->doAction(sprintf("mailpoet/automation/workflow/step/before_save/key=%s", $step->getKey()), $step);
  }
```

mailpoet/lib/Automation/Integrations/MailPoet/MailPoetIntegration.php
```diff
@@ -39,5 +39,8 @@ public function register(Registry $registry): void {
     $registry->addSubject($this->subscriberSubject);
     $registry->addTrigger($this->segmentSubscribedTrigger);
     $registry->addAction($this->sendEmailAction);
+
+    // sync step args (subject, preheader, etc.) to email settings
+    $registry->onBeforeWorkflowStepSave([$this->sendEmailAction, 'saveEmailSettings']);
```
johnolek: Is there a reason we would want this as a beforeSave hook instead of an afterSave? I know we don't have an afterSave right now and I'm wondering if that's something we would want.

I'm thinking it's possible for things to go wrong between this hook and the workflow itself getting saved. If the workflow fails to save, do we even want the updated email to save?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
```diff
@@ -0,0 +1,62 @@
+import { useCallback, useEffect, useState } from 'react';
+import { dispatch, useSelect } from '@wordpress/data';
+import { plus } from '@wordpress/icons';
+import { Button } from '../../components/button';
+import { store } from '../../../../editor/store';
+import { MailPoet } from '../../../../../mailpoet';
+
+export function DesignEmailButton(): JSX.Element {
```
johnolek, line 8: Not sure if this is the right place to put this comment, but there were a couple of things I noticed with the flow of creating/saving an email.

The first is that there's an initial warning about corrupted data when first viewing the editor.

image

And after saving it doesn't seem like it's redirecting correctly back to the workflow:

image

These are probably things that you're already aware of or there's some issue with my local environment, but I wanted to mention them just in case!

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -51,6 +59,15 @@ public function getName(): string {
     return __('Send email', 'mailpoet');
   }
 
+  public function getArgsSchema(): ObjectSchema {
+    return Builder::object([
+      'subject' => Builder::string()->default(__('Subject', 'mailpoet')),
+      'preheader' => Builder::string(),
+      'from_name' => Builder::string()->default($this->settings->get('sender.name')),
+      'email' => Builder::string()->default($this->settings->get('sender.address')),
```
johnolek, line 67: Is `'email'` too generic of a name for what I hope I'm correct in assuming is the "From" email address?

Also, is there a reason this doesn't include a newsletter ID? Maybe because the email is never present yet when the action is first created?

johnolek (changes requested):
I hope I did this PR justice. There were a lot of changes and a lot to wrap my head around, so I'm sorry if I didn't look at everything as closely as I should have.

The biggest thing for me is I'd like to know if the redirect back to workflow builder is working for you, because I haven't been able to get it to work for me locally. 

Many of my comments probably don't require any action, especially not in this PR, but are more things we might want to consider for later.

mailpoet/lib/Automation/Engine/Registry.php
JanJakes: I probably added this only for some kind of completeness and symmetry with getters.

The whole process that led me to adding the parent `Step` was that originally, instead of the `WORKFLOW_STEP_BEFORE_SAVE` hook, I was considering calling sth like `$step->onBeforeSave()` but then I realized we may end up with a huge interface over time and that some kind of events would be better and since we're in WP then actions/filters would be the most idiomatic. Having this done, I wanted to keep it because I think having the parent `Step` is correct in any way.

mailpoet/lib/Automation/Engine/Hooks.php
JanJakes: Good point. There's one thing I'm not sure about — should we use filters or actions or both? Or should we have actions for now and later convert them to filters if we allow modifying some of the data?

Thinking about it, I agree with the `_ACTION` suffix — that way, we can keep both actions and filters in the same class and choose action vs. filter freely. I already had `Hook naming` on the checklist (to review it before MVP) so I now changed it to `Hook naming (review the naming and actions vs. filters)`. It will be probably better to review separately than to try to solve it within this PR.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
JanJakes: That's actually a great idea! WP also has things like `activate_{$plugin}` so we can definitely do something similar.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
JanJakes: @johnolek

> The first is that there's an initial warning about corrupted data when first viewing the editor.

This will be solved by template selection step (discussed [here](https://a8c.slack.com/archives/C0313447YBC/p1659081183309909)). I'll create a ticket for that (today, hopefully).

> And after saving it doesn't seem like it's redirecting correctly back to the workflow:

Ahh, sorry, I forgot to mention that you need to deactivate and activate the plugin again to have the new newsletter option fields created in the DB. After that, the newsletters will have the workflow ID saved in them and it should work.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
JanJakes: I actually noticed this just yesterday and fixed it here: https://github.com/mailpoet/mailpoet/pull/4280/commits/dfee6fa39c30420aebd4ec0de4bb1112d9a2a564

I really like what type of details you notice 👏👍

mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
JanJakes: Hmm, not sure. I'm kind of thinking I'd like to have the storage classes very simple and kind of low-level, and maybe that it should be the controller on a higher level to add events like these. But I'm not 100% convinced either way.

I see a difference here from things like Doctrine events (where you really want to run something on every DB update) because this hook (in my thinking) was more like "call it exactly once and only when the workflow is explicitly updated through the update API" and it seemed to me that `UpdateWorkflowController` could be the right service with that responsibility.

Ultimately, we may also need a lower-level "update" that will not trigger this (e.g., something is updated by a cron, not by a user), and therefore I kind of think the storage classes could try to stay low-level.

mailpoet/lib/Automation/Integrations/MailPoet/MailPoetIntegration.php
JanJakes: My thinking was a bit the opposite — instead of this:

> I'm thinking it's possible for things to go wrong between this hook and the workflow itself getting saved. If the workflow fails to save, do we even want the updated email to save?

I was thinking, "if the hook fails for some reason, do we want to have such a workflow saved"? It also seemed to me safer to first sync the changes to the newsletter entity and only then update the workflow so we don't have an "inconsistent" state where the workflow is saved, but some related data is not saved yet.

mailpoet/lib/Automation/Engine/Hooks.php
JanJakes: Hmm, I'm looking at this, and I'm not sure. WP doesn't reflect action vs. filter in the name (https://adambrown.info/p/wp_hooks/hook). I know we can have it only in the constants but wouldn't it prevent us from simply switching from action to a filter in the future? Or I guess such a switch is not without a BC break anyway because a filter must always return something? 🤔

mailpoet/lib/Automation/Engine/Builder/UpdateStepsController.php
JanJakes: Good point! Fixed in https://github.com/mailpoet/mailpoet/pull/4276/commits/d6bc7ff7f155bf419fa56a44395755c024b89027.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
JanJakes: Implemented in https://github.com/mailpoet/mailpoet/pull/4276/commits/993489d66eb89514866fb85f262b89af2de05a1c.

mailpoet/lib/Automation/Engine/Hooks.php
johnolek: > should we use filters or actions or both? Or should we have actions for now and later convert them to filters if we allow modifying some of the data?

I think we'll ultimately just need to play this by ear. My gut reaction is that we should use actions where possible. We can even allow modification of data via actions since objects will be passed by reference, and we could expose public methods on objects that could be used to modify data if we wanted to, though that would arguably be moving a little too far away from WP standards. Filters seem to be more prone to problems, since it's possible for a callback to return a completely different type, or return nothing at all.

As for including `action` or `filter` in the names, I do think that's probably overkill. It could be a tiny nice-to-have to use a suffix for the constants, but in any case I don't think we need to make any changes right away.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
johnolek: I knew it must have been something in my local environment!

At first I thought it still wasn't working because I was still trying to use the newsletter that was created before the populator did its thing, but once I created a new newsletter it worked fine.

Not necessary for this PR at all, but a future enhancement might be to also store the corresponding step and have the editor automatically focus that step when you return, and maybe include a success notice about the email having been successfully saved/updated.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
johnolek: Awesome, thanks for making the update!

mailpoet/lib/Automation/Integrations/MailPoet/MailPoetIntegration.php
johnolek: I hadn't thought of it in the other direction like that. Excellent point!

I do still think the current implementation could potentially cause problems. 

Let's say I have a welcome email workflow. The trigger is signups to segment A, and the email subject is "Thanks for signing up for A!" I decide I don't want to send welcome emails for segment A anymore, but I do want to for segment B, so to be efficient I decide to edit the workflow instead of deleting it and creating a new one. I update the trigger to be for segment B, change the subject to be "Thanks for signing up for B!", and make some other change that somehow makes the workflow impossible to save. I see some error message saying, "Could not save workflow," and it happens to be the end of my work day so I sign off and tell myself I'll finish it tomorrow, not knowing that the email subject was actually changed. That evening 20,000 people sign up for segment A and get an email with the subject line, "Thanks for signing up for B!", and the very next day I get fired from my marketing job for incompetence.

Obviously this is an absurd and unlikely story. Is it a possibility in terms of the saving workflow right now though?

Part of me would love to wrap any database changes in a transaction so we could roll back if we encounter any errors, but that would certainly be overkill right now, and maybe the situation I outlined isn't even possible.

I think the current approach makes more sense than an `afterSave` so let's just keep this as is for now, though I am curious to hear if you think my scenario is possible or a concern at all.

mailpoet/lib/Automation/Engine/Builder/UpdateWorkflowController.php
johnolek: That makes perfect sense to me, thank you for the explanation!

johnolek:
@JanJakes Thank you as always for your thoughtful responses to my questions! I had just [one small comment](https://github.com/mailpoet/mailpoet/pull/4276#commitcomment-80067622) about one of the updates you made, otherwise this looks good to me.

Edit: looks like there are now some conflicts that will need to be resolved too.

mailpoet/lib/Automation/Engine/Hooks.php
JanJakes: > My gut reaction is that we should use actions where possible.
> Filters seem to be more prone to problems, since it's possible for a callback to return a completely different type, or return nothing at all.

Good point, I haven't thought about that. WP filters indeed allow absolutely anything, which is problematic. I initially almost started doing a custom event publisher/listener service but then I remembered it makes no sense — when we're in WP we probably should go with actions and filters.

I'm just thinking now to what extent we can wrap WP hooks and "type" them. That, of course, won't prevent anyone from using them directly, but we could have a higher-level safer way of using them. For now, I added a few helper methods (as you saw in this PR) but maybe that's all subject to change and improvement. Definitely leaving this open to further exploration.

mailpoet/lib/Automation/Integrations/MailPoet/MailPoetIntegration.php
JanJakes: Yes, that absolutely makes sense to me. For now, doing it either way can cause these inconsistencies and it's rather a point of view which one is better or worse.

As far as the hooks go, I think we should provide both — before and after. But for our implementation of the send email step we can explore what would be the safest:
1. We can decide whether we consider before or after safer.
2. We can try to use a transaction. Do we use them in the plugin at all? 🤔 I know there were some issues such as some "perf. optimizing" plugins converting all tables to MyISAM and such. But maybe a query with a transaction doesn't fail on MyISAM, it just may not end up being transactionally performed. Not sure.
3. Maybe we can force-sync the args to the newsletter entity before sending. Then we should probably use "after" save, as you suggest, or maybe even no hook at all. Noting this to the unorganized checklist.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/design_email_button.tsx
JanJakes: > Not necessary for this PR at all, but a future enhancement might be to also store the corresponding step and have the editor automatically focus that step when you return, and maybe include a success notice about the email having been successfully saved/updated.

Yes! I thought about focusing the step maybe, so I [implemented](https://github.com/mailpoet/mailpoet/pull/4276/commits/54a17943f7f1309e9d10af26271d3bb9a34af785), saving the step ID to newsletter options already. But the notice is a great idea as well! Again, noting it.

Thinking further, I wonder if we should explore other solutions (more user-friendly) than the redirects — maybe such as a modal element with an iframe or something like that 🤔

johnolek: I'm trying to imagine what I would prefer as a user and I don't really have a strong opinion. The way it's working now is certainly good for MVP, but I appreciate your drive to improve it if possible!

mailpoet/lib/Automation/Integrations/MailPoet/MailPoetIntegration.php
johnolek: Personally I like the idea of #3. Ensuring that we have the correct data in the newsletter at send time seems like the most important thing. And we definitely don't want to depend on transactions considering MyISAM, which I wasn't even thinking of. Another decision for the future!

johnolek (approved):
Thanks for the discussions and the tweaks, this looks good to me!

veljkho (approved):

PR #4277, opened 2022-07-29
Track if MailPoet is installed via WooCommerce Onboarding Wizard [MAILPOET-4491]

rodrigoprimo:
**How to test this PR**

- Check that the information about whether or not MP was installed via the WooCommerce Onboarding Wizard is displayed correctly on the System Info page: /wp-admin/admin.php?page=mailpoet-help#/systemInfo
- Check that the same information is also present in the data sent to Mixpanel. I used the following snippet to test that. There might be better ways to do the same.

```
add_action('init', function() {
  $reporter = \MailPoet\DI\ContainerWrapper::getInstance()->get(\MailPoet\Analytics\Reporter::class);
  var_dump($reporter->getData()); die;
});
```

See [this comment](https://mailpoet.atlassian.net/browse/MAILPOET-4491?focusedCommentId=22546) for more information on at which point of the WooCommerce onboarding wizard it is possible to install MailPoet.

[MAILPOET-4491]

[MAILPOET-4491]: https://mailpoet.atlassian.net/browse/MAILPOET-4491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/WooCommerce/Helper.php
+++ b/mailpoet/lib/WooCommerce/Helper.php
@@ -2,6 +2,9 @@
 
 namespace MailPoet\WooCommerce;
 
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\WP\Functions as WPFunctions;
+
 class Helper {
   public function isWooCommerceActive() {
     return class_exists('WooCommerce');
@@ -77,5 +80,22 @@ class Helper {
 
   public function getAllowedCountries(): array {
     return (new \WC_Countries)->get_allowed_countries() ?? [];
+  }
+
+  public function wasMailPoetInstalledViaWooCommerceOnboardingWizard(): bool {
+    $wp = ContainerWrapper::getInstance()->get(WPFunctions::class);
+    $installedViaWooCommerce = false;
+    $wooCommerceOnboardingProfile = $wp->getOption('woocommerce_onboarding_profile');
+
+    if (
+      !empty($wooCommerceOnboardingProfile)
+      && isset($wooCommerceOnboardingProfile['business_extensions'])
+      && is_array($wooCommerceOnboardingProfile['business_extensions'])
+      && in_array('mailpoet', $wooCommerceOnboardingProfile['business_extensions'])
+    ) {
+      $installedViaWooCommerce = true;
+    }
+
+    return $installedViaWooCommerce;
   }
 }
--- a/mailpoet/lib/WooCommerce/Helper.php
+++ b/mailpoet/lib/WooCommerce/Helper.php
@@ -88,7 +88,7 @@ class Helper {
     $wooCommerceOnboardingProfile = $wp->getOption('woocommerce_onboarding_profile');
 
     if (
-      !empty($wooCommerceOnboardingProfile)
+      is_array($wooCommerceOnboardingProfile)
       && isset($wooCommerceOnboardingProfile['business_extensions'])
       && is_array($wooCommerceOnboardingProfile['business_extensions'])
       && in_array('mailpoet', $wooCommerceOnboardingProfile['business_extensions'])
--- a/mailpoet/tests/integration/WooCommerce/HelperTest.php
+++ b/mailpoet/tests/integration/WooCommerce/HelperTest.php
@@ -25,12 +25,12 @@ class HelperTest extends \MailPoetTest {
   public function testGetDataMailPoetNotInstalledViaWooCommerceOnboardingWizard() {
     $this->assertFalse($this->helper->wasMailPoetInstalledViaWooCommerceOnboardingWizard());
 
-    $this->wp->addOption('woocommerce_onboarding_profile', ['business_extensions' => ['jetpack', 'mailchimp', 'another_plugin']]);
+    $this->wp->updateOption('woocommerce_onboarding_profile', ['business_extensions' => ['jetpack', 'mailchimp', 'another_plugin']]);
     $this->assertFalse($this->helper->wasMailPoetInstalledViaWooCommerceOnboardingWizard());
   }
 
   public function testGetDataMailPoetInstalledViaWooCommerceOnboardingWizard() {
-    $this->wp->addOption('woocommerce_onboarding_profile', ['business_extensions' => ['jetpack', 'mailchimp', 'mailpoet', 'another_plugin']]);
+    $this->wp->updateOption('woocommerce_onboarding_profile', ['business_extensions' => ['jetpack', 'mailchimp', 'mailpoet', 'another_plugin']]);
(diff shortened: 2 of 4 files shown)


mailpoet/lib/WooCommerce/Helper.php
```diff
@@ -78,4 +81,21 @@ public function getRawPrice($price, array $args = []) {
   public function getAllowedCountries(): array {
     return (new \WC_Countries)->get_allowed_countries() ?? [];
   }
+
+  public function wasMailPoetInstalledViaWooCommerceOnboardingWizard(): bool {
+    $wp = ContainerWrapper::getInstance()->get(WPFunctions::class);
```
rodrigoprimo, line 86: Just noting here that I opted to call ContainerWrapper::getInstance() instead of using dependency injection because the Helper class is instantiated directly in several places and I didn't want to expand too much the scope of this ticket and look into replacing the direct instantiations of the Helper class with DI all over or manually updating the instantiations to add the new dependency.

mailpoet/lib/WP/Functions.php
```diff
@@ -140,6 +140,10 @@ public function deleteCommentMeta($commentId, $metaKey, $metaValue = '') {
     return delete_comment_meta($commentId, $metaKey, $metaValue);
   }
 
+  public function addOption($option_name, $option_value) {
+    return add_option($option_name, $option_value);
```
johnolek: If we do want to keep this, would it be worthwhile to have this method match the WordPress signature exactly?

```php
function add_option( $option, $value = '', $deprecated = '', $autoload = 'yes' )
```

The only real benefit I can see would be in case we want to add an option that isn't autoloaded.

mailpoet/lib/WooCommerce/Helper.php
johnolek: Makes sense, thanks for the explanation!

mailpoet/tests/integration/WooCommerce/HelperTest.php
```diff
@@ -0,0 +1,36 @@
...
+    $this->wp = $this->diContainer->get(WPFunctions::class);
+    $this->helper = $this->diContainer->get(Helper::class);
+  }
+
+  public function _after() {
+    $this->wp->deleteOption('woocommerce_onboarding_profile');
+  }
+
+  public function testGetDataMailPoetNotInstalledViaWooCommerceOnboardingWizard() {
+    $this->assertFalse($this->helper->wasMailPoetInstalledViaWooCommerceOnboardingWizard());
```
johnolek, line 26: I noticed that we don't seem to use `$this->assertFalse` very much, favoring something like the following instead:

```php
expect($this->helper->wasMailPoetInstalledViaWooCommerceOnboardingWizard())->false();
```

Do you think it's worth changing the assertions in this file for consistency?

mailpoet/lib/WP/Functions.php
```diff
@@ -140,6 +140,10 @@ public function deleteCommentMeta($commentId, $metaKey, $metaValue = '') {
     return delete_comment_meta($commentId, $metaKey, $metaValue);
   }
 
+  public function addOption($option_name, $option_value) {
```
johnolek: It seems like we should be able to use the existing `updateOption` since it will create options if they don't already exist, but is there a reason you think `addOption` is more appropriate in this case?

mailpoet/lib/WooCommerce/Helper.php
```diff
@@ -78,4 +81,21 @@ public function getRawPrice($price, array $args = []) {
...
+    $installedViaWooCommerce = false;
+    $wooCommerceOnboardingProfile = $wp->getOption('woocommerce_onboarding_profile');
+
+    if (
+      !empty($wooCommerceOnboardingProfile)
```
johnolek: I might replace the `!empty` with `is_array`, just in case someone decided to come along and `update_option('woocommerce_onboarding_profile', new \stdClass())` or something equally as unlikely and ridiculous.

johnolek (changes requested):

mailpoet/lib/WP/Functions.php
rodrigoprimo: I completely forgot about `update_option()` when working on this :blush:

I agree that in this case, using `update_option()` is fine, so I have dropped the commit that created Functions::addOptions() and added a new commit to replace the calls to addOption() with updateOption().

mailpoet/lib/WP/Functions.php
rodrigoprimo: Yeah, I should have added the full signature and I would if we had decided to keep addOption() as more often than not it makes sense to add an option with autoload set to `no` to avoid performance problems due to bloated options that are unnecessarily autoloaded. Thanks for mentioning it.

mailpoet/tests/integration/WooCommerce/HelperTest.php
rodrigoprimo: I prefer to use the PHPUnit API over the Codeception API, but you raise a good point about consistency and I'd be happy to revisit my decision. My main reason for using the PHPUnit API is that I have found its error messages to be more useful in some cases. I don't remember any particular example, but cc @costasovo as we discussed this last year and he might remember. I tried to find where we had a conversation about this to link it here but I failed.

mailpoet/lib/WooCommerce/Helper.php
rodrigoprimo: Good point! I failed to consider this. I added a new commit making the change that you suggested.

mailpoet/tests/integration/WooCommerce/HelperTest.php
rodrigoprimo: Maybe it is the error messages when the expectations for stubs and mocks are not met, but I'm not sure.

johnolek: I prefer the style of the PHPUnit API myself, but I have an even stronger preference for consistency.

Ultimately, though, we use both styles in our tests and I'm totally fine keeping these as-is.

mailpoet/lib/WP/Functions.php
johnolek: Sounds good!

johnolek (approved):
Thanks for making those tweaks, @rodrigoprimo! Like I mentioned in my comment, there's no need (of course) to switch to using different assertions, but if you do want to I'm happy to hit the approve button again.

veljkho (approved):

PR #4281, opened 2022-08-02
Update Gutenberg JS packages [MAILPOET-4301]

lysyjan:
[MAILPOET-4301]

[MAILPOET-4301]: https://mailpoet.atlassian.net/browse/MAILPOET-4301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo (changes requested):
Excellent job! I really like the new inputs for colors! 👍 

I went through all blocks, and I found one issue we need to address.

The core/column block, which represents a column inside core/columns block, introduced a new setting called "Layout". It allows setting the width of content and width of the column.

We don't support this new settings in our data mapper [for the column block](https://github.com/mailpoet/mailpoet/blob/45e1db5893f46ce26a7a440fa74170380565fac6/mailpoet/assets/js/src/form_editor/store/blocks_to_form_body.jsx#L192-L216) (and also in the [mapper from form data to blocks](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/form_editor/store/form_body_to_blocks.jsx#L85-L161)) so it is not saved and also we don't support it in [the form column rederer on the frontend](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Form/Block/Column.php).

Screenshot 2022-08-03 at 10 36 45

I think we have two options here:
1) Try to disable the new settings [when we register the column block](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/form_editor/blocks/columns/column.jsx#L16-L17). By modifying the metadata's support property.
2) Add support for the new setting.

I find the new feature quite handy, so I would suggest exploring the second option first.
It means adding code to a couple of places (mappers, mappers tests and front end renderer), but it should be quite straight forward.

lysyjan:
@costasovo Thank you for this catch!
I decided to disable this feature because I wasn't able to use the `Wide` configuration properly. Maybe we can recheck this feature during the next update.

costasovo (dismissed):

samnajian:
FYI: Just rebased trunk since had branched from this branch and the new branch

veljkho (approved):

PR #4282, opened 2022-08-02
Fix notifications about new subscriber from the manage subscription page [MAILPOET-4518]

lysyjan:
[MAILPOET-4518]

[MAILPOET-4518]: https://mailpoet.atlassian.net/browse/MAILPOET-4518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscription/Manage.php
+++ b/mailpoet/lib/Subscription/Manage.php
@@ -153,7 +153,7 @@ class Manage {
       if ($allowedSegments && !in_array($segment->getId(), $allowedSegments)) {
         continue;
       }
-      if (!in_array($subscriberSegment->getId(), $segmentsIds)) {
+      if (!in_array($segment->getId(), $segmentsIds)) {
         $this->subscriberSegmentRepository->createOrUpdate(
           $subscriber,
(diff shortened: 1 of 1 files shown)


mailpoet/lib/Subscription/Manage.php
```diff
@@ -153,7 +153,7 @@ private function updateSubscriptions(SubscriberEntity $subscriber, array $subscr
       if ($allowedSegments && !in_array($segment->getId(), $allowedSegments)) {
         continue;
       }
-      if (!in_array($subscriberSegment->getId(), $segmentsIds)) {
+      if (!in_array($segment->getId(), $segmentsIds)) {
```
samnajian, line 156: Was wondering if we could adjust the existing tests to catch such and issue, what do you think?

samnajian (approved):
Let's proceed without test now and add the test later on.

veljkho (approved):

PR #4284, opened 2022-08-02
Move the Settings link before the Deactivation link [MAILPOET-4516]

lysyjan:
[MAILPOET-4516]

[MAILPOET-4516]: https://mailpoet.atlassian.net/browse/MAILPOET-4516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -424,9 +424,13 @@ class Hooks {
     );
   }
 
-  public function setSettingsLinkInPluginPage($actionLinks) {
+  /**
+   * @param array<string, string> $actionLinks
+   * @return array<string, string>
+   */
+  public function setSettingsLinkInPluginPage(array $actionLinks): array {
     $customLinks = [
-      'settings' => '<a href="' . $this->wp->adminUrl('admin.php?page=mailpoet-settings') . '" aria-label="' . esc_attr__( 'View MailPoet settings', 'mailpoet' ) . '">' . esc_html__( 'Settings', 'mailpoet' ) . '</a>',
+      'settings' => '<a href="' . $this->wp->adminUrl('admin.php?page=mailpoet-settings') . '" aria-label="' . $this->wp->escAttr(__('View MailPoet settings', 'mailpoet')) . '">' . $this->wp->escHtml(__('Settings', 'mailpoet')) . '</a>',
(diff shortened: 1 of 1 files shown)


rodrigoprimo (approved):
LGTM

mailpoet/lib/Config/Hooks.php
```diff
@@ -424,11 +424,15 @@ public function setupSettingsLinkInPluginPage() {
...
 
-  public function setSettingsLinkInPluginPage($actionLinks) {
+  /**
+   * @param array<string, string> $actionLinks
+   * @return array<string, string>
+   */
+  public function setSettingsLinkInPluginPage(array $actionLinks): array {
     $customLinks = [
-      'settings' => '<a href="' . $this->wp->adminUrl('admin.php?page=mailpoet-settings') . '" aria-label="' . esc_attr__( 'View MailPoet settings', 'mailpoet' ) . '">' . esc_html__( 'Settings', 'mailpoet' ) . '</a>',
+      'settings' => '<a href="' . $this->wp->adminUrl('admin.php?page=mailpoet-settings') . '" aria-label="' . $this->wp->escAttr(__('View MailPoet settings', 'mailpoet')) . '">' . $this->wp->escHtml(__('Settings', 'mailpoet')) . '</a>',
```
rodrigoprimo, line 433: @lysyjan I'm ok with using the internal functions and I already approved the PR, but just to make sure I'm not missing something, I thought that for translation functions we use WP functions directly?

lysyjan: I used internal functions for escaping the attribute and HTML, but for translation is still used WP function directly.

veljkho (approved):

mailpoet/lib/Config/Hooks.php
rodrigoprimo: Thanks for the input. I guess I see `esc_attr__()` and `esc_html__()` as translation functions that also take care of escaping and that is why I used the WP functions instead of the ones that we define. Do you think that is ok going forward? Should we document somewhere which WP functions we call directly and which we don't to avoid this kind of confusion going forward?

PR #4285, opened 2022-08-02
Remove old model usage from BatchIterator [MAILPOET-4357]

rodrigoprimo:
[MAILPOET-4357]

[MAILPOET-4357]: https://mailpoet.atlassian.net/browse/MAILPOET-4357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
+++ b/mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
@@ -62,11 +62,7 @@ class ScheduledTaskSubscribersRepository extends Repository {
       ->getQuery()
       ->getSingleScalarResult();
 
-    if (is_numeric($countSubscribers)) {
-      return (int)$countSubscribers;
-    } else {
-      return 0;
-    }
+    return intval($countSubscribers);
   }
 
   public function getSubscriberIdsBatchForTask(int $taskId, int $lastProcessedSubscriberId, int $limit): array {
--- a/mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
+++ b/mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
@@ -7,6 +7,7 @@ use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\ScheduledTaskSubscriberEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\InvalidStateException;
+use MailPoetVendor\Doctrine\ORM\QueryBuilder;
 
 /**
  * @extends Repository<ScheduledTaskSubscriberEntity>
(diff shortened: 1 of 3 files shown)


mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
```diff
@@ -53,4 +54,45 @@ public function createOrUpdate(array $data): ?ScheduledTaskSubscriberEntity {
...
+  public function countSubscriberIdsBatchForTask(int $taskId, int $lastProcessedSubscriberId): int {
+    $queryBuilder = $this->entityManager
+      ->createQueryBuilder()
+      ->select('count(sts.subscriber)');
+    $queryBuilder = $this->prepareSubscriberIdsBatchForTaskQuery($queryBuilder, $taskId, $lastProcessedSubscriberId);
+    $countSubscribers = $queryBuilder
+      ->getQuery()
+      ->getSingleScalarResult();
+
+    if (is_numeric($countSubscribers)) {
```
lysyjan: This if can be simplified by `return intval($countSubscribers);`.

mailpoet/lib/Tasks/Subscribers/BatchIterator.php
```diff
@@ -24,6 +28,7 @@ public function __construct(
     }
     $this->taskId = (int)$taskId;
     $this->batchSize = (int)$batchSize;
+    $this->scheduledTaskSubscribersRepository = ContainerWrapper::getInstance()->get(ScheduledTaskSubscribersRepository::class);
```
lysyjan, line 31: I'm not sure about this. For me, it looks like a hidden dependency. Did you consider passing the repository via constructor?

mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
```diff
@@ -53,4 +54,45 @@ public function createOrUpdate(array $data): ?ScheduledTaskSubscriberEntity {
...
+    $subscribersIds = $queryBuilder
+      ->orderBy('sts.subscriber', 'asc')
+      ->setMaxResults($limit)
+      ->getQuery()
+      ->getSingleColumnResult();
+
+    return $subscribersIds;
+  }
+
+  private function prepareSubscriberIdsBatchForTaskQuery(QueryBuilder $queryBuilder, int $taskId, int $lastProcessedSubscriberId): QueryBuilder {
```
lysyjan: This method could also contain creating a query builder because you avoid one-line code repetition, and the selection can be configured later, but it's a detail only.

mailpoet/tests/integration/Tasks/Subscribers/BatchIteratorTest.php
```diff
@@ -46,14 +57,6 @@ public function testItIterates() {
     foreach ($this->iterator as $batch) {
       $i++;
 
-      // process subscribers
```
lysyjan, line 49: I wonder if it's possible to remove this without replacement. I'm curious only.

lysyjan (commented):
I have only a few notes, nothing critical, I guess.

mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
rodrigoprimo: I opted not to use intval() as getSingleScalarResult() return type is `mixed` and I'm not sure if PHP will always be smart enough to convert whatever getSingleScalarResult() returns to a proper int. That being said, maybe I'm being over paranoid and the query is fairly simple and should always return an int due to the call to count(). What do you think?

mailpoet/lib/Tasks/Subscribers/BatchIterator.php
rodrigoprimo: Yeah, I considered passing the repository via the constructor. I opted not to as we instantiate BatchIterator directly in a few places and I didn't want to mess with that. I don't really see a problem with getting an instance of ScheduledTaskSubscribersRepository here directly and we can always change that in the future if needed. But I'd be happy to change it now if you see a reason for it.

mailpoet/tests/integration/Tasks/Subscribers/BatchIteratorTest.php
rodrigoprimo: I was trying to understand why we would need to change the value of the field `processed` for the subscribers inside this test. I couldn't figure this out, but in the process, I realized that the test didn't require it to pass (even before the changes that I made). So that is why it seems to me it is safe to remove this part of the code.

mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
rodrigoprimo: Good point, I was working under the incorrect assumption that I had to call `select()` before `from()`. I added a new commit implementing your suggestion. Thanks.

rodrigoprimo:
Thanks for the review, @lysyjan. I believe I responded to all your comments. Could you please check this PR again?

mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
lysyjan: The type `mixed` is in the annotation, but your query should return only the number. I think that `intval()` is secure here.

lysyjan:
@rodrigoprimo Thank you for addressing all comments. Please, use the `intval()` function in `ScheduledTaskSubscribersRepository.php`. The rest of the changes are fine.

rodrigoprimo:
Makes sense, @lysyjan.thanks for your input. I added a new commit. This PR is ready for a final review.

lysyjan (approved):

veljkho (approved):

PR #4286, opened 2022-08-02
Remove old model usage from the SubscriberLinkTokens worker [MAILPOET-4345]

rodrigoprimo:
A few things worth noting about this PR:

- There were no integration tests for SubscriberLinkTokens so I created a very basic test class with a single test.
- I also added a new method to the Subscriber data factory to be able to set the link_token when creating a subscriber.
- I used Doctrine to create a SQL query instead of using DQL as it seemed to me on a quick search that the SQL function MD5() is not available by default and it is not worth registering it as a custom function. Please let me know if I missed something and I should have taken a different approach.

[MAILPOET-4345]

[MAILPOET-4345]: https://mailpoet.atlassian.net/browse/MAILPOET-4345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SubscriberLinkTokens.php
+++ b/mailpoet/lib/Cron/Workers/SubscriberLinkTokens.php
@@ -2,11 +2,13 @@
 
 namespace MailPoet\Cron\Workers;
 
+use MailPoet\DI\ContainerWrapper;
 use MailPoet\Entities\ScheduledTaskEntity;
-use MailPoet\Models\Subscriber;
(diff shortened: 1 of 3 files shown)


costasovo (approved):
LGTM!

Thank you for improving the test coverage! As for the MD5, I agree that creating a custom function for this single case is not worth it.

veljkho (approved):

PR #4287, opened 2022-08-03
Refactor MP API addList to Doctrine [MAILPOET-4294]

costasovo:
[MAILPOET-4294]

[MAILPOET-4294]: https://mailpoet.atlassian.net/browse/MAILPOET-4294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/MP/v1/API.php
+++ b/mailpoet/lib/API/MP/v1/API.php
@@ -191,43 +191,7 @@ class API {
   }
 
   public function addList(array $list) {
-    // throw exception when list name is missing
-    if (empty($list['name'])) {
-      throw new APIException(
-        __('List name is required.', 'mailpoet'),
-        APIException::LIST_NAME_REQUIRED
-      );
-    }
-
-    // throw exception when list already exists
-    if (Segment::where('name', $list['name'])->findOne()) {
-      throw new APIException(
-        __('This list already exists.', 'mailpoet'),
-        APIException::LIST_EXISTS
-      );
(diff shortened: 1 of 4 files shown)


mailpoet/lib/API/MP/v1/API.php
```diff
@@ -191,43 +191,7 @@ public function addSubscriber(array $subscriber, $listIds = [], $options = []) {
...
-
-    // filter out all incoming data that we don't want to change, like type,
-    $list = array_intersect_key($list, array_flip(['name', 'description']));
-
-    // add list
-    $newList = Segment::create();
-    $newList->hydrate($list);
-    $newList->save();
-    if ($newList->getErrors() !== false) {
-      throw new APIException(
```
alex-mailpoet, line 218: A couple of these exceptions are missing from the refactored version. Do you think they are redundant or will Doctrine throw something similar in case of DB failures and there's no need to explicitly code them?

costasovo: A good catch! A doctrine would throw an exception but it would be a different type so technically a BC break. I'll add try/catch and convert the exception so that it is backward compatible.

alex-mailpoet (approved):
Looks good! Thanks!

veljkho (approved):

PR #4289, opened 2022-08-03
Workflow versioning

websupporter:
Fixes [MAILPOET-4430]

[MAILPOET-4430]: https://mailpoet.atlassian.net/browse/MAILPOET-4430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Migrations/Migrator.php
+++ b/mailpoet/lib/Automation/Engine/Migrations/Migrator.php
@@ -29,6 +29,14 @@ class Migrator {
         created_at timestamp NOT NULL,
         updated_at timestamp NOT NULL,
         deleted_at timestamp NULL,
+        PRIMARY KEY (id)
+      );
+    ");
+
+    $this->runQuery("
+      CREATE TABLE {$this->prefix}workflow_versions (
+        id int(11) unsigned NOT NULL AUTO_INCREMENT,
+        workflow_id int(11) unsigned NOT NULL,
         trigger_keys longtext NOT NULL,
         steps longtext,
         PRIMARY KEY (id)
@@ -53,6 +61,7 @@ class Migrator {
     $this->removeOldSchema();
     $this->runQuery("DROP TABLE IF EXISTS {$this->prefix}workflows");
     $this->runQuery("DROP TABLE IF EXISTS {$this->prefix}workflow_runs");
+    $this->runQuery("DROP TABLE IF EXISTS {$this->prefix}workflow_versions");
 
     // clean Action Scheduler data
     $this->runQuery("
--- a/mailpoet/lib/Automation/Engine/Migrations/Migrator.php
+++ b/mailpoet/lib/Automation/Engine/Migrations/Migrator.php
@@ -47,6 +47,7 @@ class Migrator {
       CREATE TABLE {$this->prefix}workflow_runs (
         id int(11) unsigned NOT NULL AUTO_INCREMENT,
         workflow_id int(11) unsigned NOT NULL,
+        version_id int(11) unsigned NOT NULL,
         trigger_key varchar(255) NOT NULL,
         status varchar(255) NOT NULL,
         created_at timestamp NOT NULL,
--- a/mailpoet/lib/Automation/Engine/Migrations/Migrator.php
+++ b/mailpoet/lib/Automation/Engine/Migrations/Migrator.php
@@ -39,6 +39,8 @@ class Migrator {
         workflow_id int(11) unsigned NOT NULL,
         trigger_keys longtext NOT NULL,
         steps longtext,
+        created_at timestamp NOT NULL,
+        updated_at timestamp NOT NULL,
         PRIMARY KEY (id)
       );
     ");
--- a/mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
+++ b/mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
@@ -10,43 +10,80 @@ use wpdb;
 
 class WorkflowStorage {
   /** @var string */
-  private $table;
+  private $workflowTable;
+  /** @var string */
+  private $versionsTable;
 
   /** @var wpdb */
   private $wpdb;
 
   public function __construct() {
     global $wpdb;
-    $this->table = $wpdb->prefix . 'mailpoet_workflows';
+    $this->workflowTable = $wpdb->prefix . 'mailpoet_workflows';
+    $this->versionsTable = $wpdb->prefix . 'mailpoet_workflow_versions';
     $this->wpdb = $wpdb;
   }
 
   public function createWorkflow(Workflow $workflow): int {
-    $result = $this->wpdb->insert($this->table, $workflow->toArray());
+    $result = $this->wpdb->insert($this->workflowTable, $this->workflowHeaderData($workflow));
     if (!$result) {
       throw Exceptions::databaseError($this->wpdb->last_error);
     }
-    return $this->wpdb->insert_id;
+    $id = (int)$this->wpdb->insert_id;
+    $this->insertWorkflowVersion($id, $workflow);
+    return $id;
   }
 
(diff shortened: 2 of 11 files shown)


mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -10,43 +10,80 @@
...
+FROM
+      $workflowTable as workflow,
+      $versionTable as version
+WHERE
+      version.workflow_id = workflow.id
+ORDER BY
+      version.id DESC
+LIMIT
+      0,1; ",
+      $id
```
JanJakes: Could we indent this similarly to normal PHP code (no left out-dent, sub-indenting by 2 spaces)? We seem to do that [in existing code](https://github.com/mailpoet/mailpoet/blob/c8278dde2a066f1f4602ef32151cc8d4ad22ca90/mailpoet/lib/Newsletter/NewslettersRepository.php#L268) and I guess it may look a tad bit nicer.

This applies also to other similar occurrences in this PR.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -10,43 +10,80 @@
...
+      version.steps,
+      version.trigger_keys
+FROM
+      $workflowTable AS workflow
+        LEFT JOIN
+         $versionTable as version ON (version.workflow_id=workflow.id)
+WHERE
+      version.id = (SELECT Max(id) FROM $versionTable WHERE workflow_id= version.workflow_id)
+ORDER BY
+      workflow.id DESC; ";
```
JanJakes: I was checking where we're using this, and it seems to be only [the `WorkflowsGetEndpoint`](https://github.com/mailpoet/mailpoet/blob/c8278dde2a066f1f4602ef32151cc8d4ad22ca90/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsGetEndpoint.php). Considering it's a "listing" endpoint and [seeing the data it uses now](https://github.com/mailpoet/mailpoet/blob/c8278dde2a066f1f4602ef32151cc8d4ad22ca90/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsGetEndpoint.php#L29), I suppose we could just keep this method touch only the `workflows` table, and maybe rename it to `getWorkflowsList` to indicate it doesn't return full workflows, only the listing data.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -73,9 +121,24 @@ public function getActiveTriggerKeys(): array {
...
-        "SELECT * FROM $table WHERE status = %s AND trigger_keys LIKE %s",
+        "
+SELECT
+       workflow.*,
+      version.id AS version_id,
+      version.steps,
+      version.trigger_keys
+FROM
+     $workflowTable AS workflow
+        LEFT JOIN
```
JanJakes: This is a small detail, but we may use just `JOIN` (inner join, I mean) as workflow without a record in `workflow_versions` would be invalid.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -85,4 +148,24 @@ public function getActiveWorkflowsByTrigger(Trigger $trigger): array {
       return Workflow::fromArray($workflowData);
     }, (array)$data);
   }
+
+  private function workflowHeaderData(Workflow $workflow): array {
```
JanJakes: To satisfy our naming conventions, I guess it needs a verb, such as `getWorkflowHeaderData`.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
```diff
@@ -47,6 +47,7 @@ public function createSchema(): void {
       CREATE TABLE {$this->prefix}workflow_runs (
         id int(11) unsigned NOT NULL AUTO_INCREMENT,
         workflow_id int(11) unsigned NOT NULL,
+        version_id int(11) unsigned NOT NULL,
```
JanJakes, line 52: Do we now need to store both `version_id` and `workflow_id`? Or... should we?
I guess in the `WorkflowRun` we can have both, but on the DB level, I'm not sure if it's needed. On the other hand, there's no harm saving both... or is there? 🤔  What do you think?

JanJakes (commented):
Great stuff! I have just a few comments/questions, nothing substantial.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
websupporter: Technically speaking the `workflow_id` would not be necessary. You can join the version data base with the unique `version_id` and conclude what is the `workflow_id`. But to me, it is a different table. Maybe I just want quickly to fetch `Select workflow.id, workflow.name, count(workflow_runs.id) as num from workflow, workflow_runs where workflow_runs.workflow_id = workflow.id` or similar queries, where I really do not need data from the version table at all. This column would enable me to do so. Also, in debugging scenarios, I think it helps, to have quickly a high level information what workflow a run belongs to.

The harm I see is in more bytes stored, but I think it is convenient information to have in this table.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
websupporter: I see why it would be beneficial to do so. On the other side, we switch explicitly from `LEFT JOIN` to `INNER JOIN` in the other queries to prevent invalid records being returned. And I think we should follow this thought here too. What do you think?

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
websupporter: Addressed in [f8785cd](https://github.com/mailpoet/mailpoet/pull/4289/commits/f8785cd3ad72147ab76f556977f32919d2619ea2)

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
websupporter: Addressed in [2ebd723](https://github.com/mailpoet/mailpoet/pull/4289/commits/2ebd7236d596787009393ed244de12682ab52e40)

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
websupporter: Let's keep it consistent :+1:
Addressed in [f937c5d](https://github.com/mailpoet/mailpoet/pull/4289/commits/f937c5d5c190c440ec2ffc7ba55bf65ee0296ce8)

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
JanJakes: @websupporter Ahh, right, we still need to use the `Workflow` data type. OK, let's keep it consistent. We can optimize it later by adding a read-only dedicated data type, such as `WorkflowListRecord`.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
JanJakes: @websupporter Makes sense! Let's keep it.

JanJakes (commented):
LGTM! There are some conflicts now, unfortunately. Could you rebase, @websupporter? Thanks!

websupporter:
@JanJakes thanks for the review. Rebased. Please have a brief look again, as it was a bigger rebase, but I think I got everything. :bow:

JanJakes (approved):

veljkho (approved):

PR #4293, opened 2022-08-04
Warnings about missing imports in webpack build [MAILPOET-4507]

samnajian:
Fixes [MAILPOET-4507]

[MAILPOET-4507]: https://mailpoet.atlassian.net/browse/MAILPOET-4507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/listing/components/cells/more.tsx
+++ b/mailpoet/assets/js/src/automation/listing/components/cells/more.tsx
@@ -1,4 +1,4 @@
-import { EllipsisMenu, MenuItem } from '@woocommerce/components';
+import { EllipsisMenu, MenuItem } from '@woocommerce/components/build';
 import { __ } from '@wordpress/i18n';
 import { Workflow } from '../../workflow';
 
--- a/mailpoet/assets/js/src/automation/listing/index.tsx
+++ b/mailpoet/assets/js/src/automation/listing/index.tsx
@@ -1,4 +1,4 @@
-import { Search, TableCard } from '@woocommerce/components';
+import { Search, TableCard } from '@woocommerce/components/build';
 import { Button, Dropdown, MenuGroup, MenuItem } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { getRow } from './get-row';
--- a/mailpoet/webpack.config.js
+++ b/mailpoet/webpack.config.js
@@ -13,6 +13,35 @@ const manifestSeed = {};
 
 // Base config
 const baseConfig = {
+  ignoreWarnings: [
+    (warnings) => {
+      // Todo: remove this if statement per MAILPOET-4544
+      if (
+        warnings &&
+        [
+          'AssetsOverSizeLimitWarning',
+          'EntrypointsOverSizeLimitWarning',
+          'NoAsyncChunksWarning',
+        ].includes(warnings.name)
+      ) {
+        return true;
+      }
+
+      // only show warnings when watching
+      if (process.env.WEBPACK_WATCH || process.argv.includes('--watch')) {
+        return false;
+      }
+
+      if (warnings) {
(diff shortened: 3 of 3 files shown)


lysyjan (changes requested):
@samnajian It looks that all acceptance tests are failing. Please, can you take a look?

samnajian:
> @samnajian It looks that all acceptance tests are failing. Please, can you take a look?

Sorry about that, I tried to avoid implementing the solution that is added now, but had to revert back my previous change.

JanJakes:
@samnajian Ahh, so this is what causes the warnings? Great catch! I just run into this. I see you tried `mainFields: ['main', 'module'],` in https://github.com/mailpoet/mailpoet/commit/8aeba2cd533a697c7e63ab7e6fdc8171fa895365, did it break something?

I'm just a bit worried we'll introduce the same issue again soon because IDEs won't autocomplete the `/build` variant. Alternatively, could we maybe alias the `/build` variant to the name without build? Not sure if that's possible. I've seen some aliased solutions [in the Webpack issue](https://github.com/webpack/webpack/issues/5756).

I'm also thinking whether we could make the builds fail if they have warnings 🤔 It would be great, and it would probably avoid us causing the same error again no matter the solution we choose. Not sure if there's such option though.

samnajian:
> @samnajian Ahh, so this is what causes the warnings? Great catch! I just run into this. I see you tried `mainFields: ['main', 'module'],` in [8aeba2c](https://github.com/mailpoet/mailpoet/commit/8aeba2cd533a697c7e63ab7e6fdc8171fa895365), did it break something?
> 
> I'm just a bit worried we'll introduce the same issue again soon because IDEs won't autocomplete the `/build` variant. Alternatively, could we maybe alias the `/build` variant to the name without build? Not sure if that's possible. I've seen some aliased solutions [in the Webpack issue](https://github.com/webpack/webpack/issues/5756).
> 
> I'm also thinking whether we could make the builds fail if they have warnings 🤔 It would be great, and it would probably avoid us causing the same error again no matter the solution we choose. Not sure if there's such option though.


Yes, unfortunately the `mainFields` definition doesn't work with all packages and some packages broke that's why even though the build was successful the artifact was broken, logically `module` should be what is used since we use `commonjs` modules, failing (bailing?) on warnings would make sense, I agree, but couldn't find a setting that would do that for us, maybe we can have another ticket to take a look and do it with an option or a bespoke solution?

JanJakes:
@samnajian Right, OK, I guess we can use this fix now, but we must keep an eye on this in automations or find a more robust solution.

samnajian:
> @samnajian Right, OK, I guess we can use this fix now, but we must keep an eye on this in automations or find a more robust solution.
@JanJakes 
[Just added a measure to exit when a warning happens](https://github.com/mailpoet/mailpoet/pull/4293/commits/918892821dabe2323986b0b6c1c1b8e2190792aa)

JanJakes:
@samnajian Awesome! And it indeed failed on a warning — oversize warning. I think you'll need to add `"performance": { "hints": false }` as well. Anyway, leaving you and @lysyjan to process this.

samnajian:
> @samnajian Awesome! And it indeed failed on a warning — oversize warning. I think you'll need to add `"performance": { "hints": false }` as well. Anyway, leaving you and @lysyjan to process this.
@JanJakes 
Created a ticket for that maybe we want to take a deeper look into it, I'll add your suggest as the last resort to that ticket

lysyjan (approved):
@JanJakes Thank you for looking into this.
It looks good to me now, so I'm moving this forward.

veljkho (approved):

PR #4294, opened 2022-08-05
Parse sender name from an email when the email is used as a name [MAILPOET-4501]

lysyjan:
[MAILPOET-4501]

[MAILPOET-4501]: https://mailpoet.atlassian.net/browse/MAILPOET-4501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Populator.php
+++ b/mailpoet/lib/Config/Populator.php
@@ -238,8 +238,12 @@ class Populator {
     }
 
     // set default sender info based on current user
+    $currentUserName = $currentUser->display_name ?: ''; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+    // parse current user name if an email is used
+    $senderName = explode('@', $currentUserName);
+    $senderName = reset($senderName);
     $defaultSender = [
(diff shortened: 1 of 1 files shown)


mailpoet/lib/Config/Populator.php
```diff
@@ -238,8 +238,12 @@ private function createDefaultSettings() {
     }
 
     // set default sender info based on current user
+    $currentUserName = $currentUser->display_name ?: ''; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+    // parse current user name if an email is used
+    $senderName = explode('@', $currentUserName);
+    $senderName = reset($senderName);
```
samnajian, line 244: I'd say we can do even more like checking for other [allowed characters](https://en.wikipedia.org/wiki/Email_address#Local-part) in email addresses and trying to use space (?) instead of them, what do you think?

lysyjan: I think this is unnecessary because parsing can happen only when the WP user that activates the plugin for the first time has an email address in the name. Concurrently the user can modify the name in the form input without any validation.

samnajian (approved):
LGTM!

veljkho (approved):

PR #4296, opened 2022-08-05
MAILPOET-4522 fix problems with marketing optin block and do watch js

samnajian:
Fixes [MAILPOET-4522]

[MAILPOET-4522]: https://mailpoet.atlassian.net/browse/MAILPOET-4522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/webpack.config.js
+++ b/mailpoet/webpack.config.js
@@ -518,24 +518,16 @@ function requestToHandle(request) {
   /* eslint-disable-next-line consistent-return, no-useless-return */
   return;
 }
-
 const marketingOptinBlock = Object.assign({}, wpScriptConfig, {
   name: 'marketing_optin_block',
   entry: {
-    'marketing-optin-block': path.resolve(
-      process.cwd(),
-      'assets/js/src/marketing_optin_block',
-      'index.tsx',
-    ),
-    'marketing-optin-block-frontend': path.resolve(
-      process.cwd(),
-      'assets/js/src/marketing_optin_block',
-      'frontend.ts',
-    ),
+    'marketing-optin-block': '/assets/js/src/marketing_optin_block/index.tsx',
+    'marketing-optin-block-frontend':
+      '/assets/js/src/marketing_optin_block/frontend.ts',
   },
   output: {
     filename: '[name].js',
-    path: path.resolve(process.cwd(), 'assets/dist/js/marketing_optin_block'),
+    path: path.join(__dirname, 'assets/dist/js/marketing_optin_block'),
   },
   module: Object.assign({}, wpScriptConfig.module, {
     rules: [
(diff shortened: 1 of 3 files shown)


mailpoet/webpack.config.js
```diff
@@ -562,10 +554,9 @@ const marketingOptinBlock = Object.assign({}, wpScriptConfig, {
     extensions: ['.js', '.jsx', '.ts', '.tsx'],
   },
   plugins: [
-    ...wpScriptConfig.plugins.filter(
-      (plugin) =>
-        plugin.constructor.name !== 'DependencyExtractionWebpackPlugin',
-    ),
+    ...wpScriptConfig.plugins.filter((plugin) => {
+      return plugin.constructor.pluginName === 'mini-css-extract-plugin';
+    }),
```
costasovo: Could you please elaborate a bit on why this change was needed? Maybe even a comment in the code explaining why we filter the plugins like that could be useful.

mailpoet/webpack.config.js
```diff
@@ -518,24 +518,16 @@ function requestToHandle(request) {
...
-      'assets/js/src/marketing_optin_block',
-      'frontend.ts',
-    ),
+    'marketing-optin-block': '/assets/js/src/marketing_optin_block/index.tsx',
+    'marketing-optin-block-frontend':
+      '/assets/js/src/marketing_optin_block/frontend.ts',
   },
   output: {
     filename: '[name].js',
-    path: path.resolve(process.cwd(), 'assets/dist/js/marketing_optin_block'),
```
costasovo, line 538: I just want to make sure I understand the change. The `process.cwd()` was causing the nesting, correct?

costasovo (commented):
Good job finding the issue! I have just a couple of questions to better understand the fix.

mailpoet/webpack.config.js
samnajian: Actually it was not this causing the nesting, but was looking weird compared to the rest of the config since we are always using `__dirname` in other parts of the config.

mailpoet/webpack.config.js
samnajian: `CopyWebpackPlugin` was the one causing the problems and with the previous config we were allowing that to be used from `wpScriptConfig` even though we were adding our custom one as well.

costasovo (approved):
Thanks for explaining!

veljkho (approved):

JanJakes:
@veljkho @samnajian This was created against the `js-gutenberg-update` branch and thus also merged into it. What now? Should we keep it there? I guess with PRs starting off of non-`trunk` branches we should wait requesting review before the parent is merged (and then rebase, to be sure).

samnajian:
@JanJakes while what you're saying is correct, I guess for this case we can also directly merge these changes into trunk, what do you think? should I create a PR against trunk since this is already merged into this [js-gutenberg-update](https://github.com/mailpoet/mailpoet/tree/js-gutenberg-update)?

JanJakes:
@samnajian Yep, I'd give priority to QAing the https://github.com/mailpoet/mailpoet/pull/4281, if possible, @veljkho 🙏

veljkho:
#4281 will be my next task today.

PR #4298, opened 2022-08-08
Fix type hint in fix-guzzle.php [PREMIUM-199]

alex-mailpoet:
[PREMIUM-199]

Without this fix the release command gives me an error:
```
ERROR: Uncaught TypeError: GuzzleHttp\http_build_query(): Argument #2 ($numeric_prefix) must be of type string, null given, called in /www/wp-content/plugins/mp3repo/mailpoet/vendor/guzzlehttp/guzzle/src/Client.php on line 452 and defined in /www/wp-content/plugins/mp3repo/mailpoet/vendor/guzzlehttp/guzzle/src/Client.php:28
```

[PREMIUM-199]: https://mailpoet.atlassian.net/browse/PREMIUM-199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tasks/fix-guzzle.php
+++ b/mailpoet/tasks/fix-guzzle.php
@@ -24,8 +24,8 @@ if (!file_exists($clientFilePath) || version_compare(phpversion(), '8.0.0') == -
 }
 $replacement = '
 // Updated by MailPoet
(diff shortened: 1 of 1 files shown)


samnajian (approved):
LGTM! Thanks for the fix.

PR #4303, opened 2022-08-09
Refactor MP API::unsubscribeFromLists to Doctrine [MAILPOET-4291]

costasovo:
[MAILPOET-4291]

[MAILPOET-4291]: https://mailpoet.atlassian.net/browse/MAILPOET-4291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/MP/v1/Subscribers.php
+++ b/mailpoet/lib/API/MP/v1/Subscribers.php
@@ -17,6 +17,9 @@ use MailPoet\Tasks\Sending;
 use MailPoet\WP\Functions as WPFunctions;
 
 class Subscribers {
+  const CONTEXT_SUBSCRIBE = 'subscribe';
+  const CONTEXT_UNSUBSCRIBE = 'unsubscribe';
+
   /** @var SettingsController */
   private $settings;
 
@@ -85,7 +88,7 @@ class Subscribers {
     $signupConfirmationEnabled = (bool)$this->settings->get('signup_confirmation.enabled');
 
     $subscriber = $this->findSubscriber($subscriberId);
-    $foundSegments = $this->getAndValidateSegments($listIds);
+    $foundSegments = $this->getAndValidateSegments($listIds, self::CONTEXT_SUBSCRIBE);
 
     $this->subscribersSegmentRepository->subscribeToSegments($subscriber, $foundSegments);
 
@@ -140,6 +143,14 @@ class Subscribers {
     }
 
     $this->subscribersRepository->refresh($subscriber);
+    return $this->subscribersResponseBuilder->build($subscriber);
+  }
+
+  public function unsubscribeFromLists($subscriberIdOrEmail, array $listIds): array {
+    $subscriber = $this->findSubscriber($subscriberIdOrEmail);
+    $foundSegments = $this->getAndValidateSegments($listIds, self::CONTEXT_UNSUBSCRIBE);
+    $this->subscribersSegmentRepository->unsubscribeFromSegments($subscriber, $foundSegments);
+
     return $this->subscribersResponseBuilder->build($subscriber);
   }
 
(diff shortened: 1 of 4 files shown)


mailpoet/lib/API/MP/v1/Subscribers.php
```diff
@@ -84,61 +87,8 @@ public function subscribeToLists(
...
-      $missingIds = array_values(array_diff($listIds, $foundSegmentsIds));
-      $exception = sprintf(
-        // translators: %s is a comma-separated list of list IDs.
-        _n('List with ID %s does not exist.', 'Lists with IDs %s do not exist.', count($missingIds), 'mailpoet'),
-        implode(', ', $missingIds)
-      );
-      throw new APIException(sprintf($exception, implode(', ', $missingIds)), APIException::LIST_NOT_EXISTS);
-    }
+    $subscriber = $this->findSubscriber($subscriberId);
+    $foundSegments = $this->getAndValidateSegments($listIds, self::CONTEXT_SUBSCRIBE);
```
johnolek, line 92: Do you think it might make sense to retrieve the segments before the subscriber just to be consistent with the old code? I doubt it really matters, but if someone made an API call with both an empty subscriber and no segment IDs, the previous code would return an error about requiring a segment while this would return an error about requiring a subscriber, right?

mailpoet/lib/API/MP/v1/Subscribers.php
```diff
@@ -223,4 +186,92 @@ protected function _sendConfirmationEmail(SubscriberEntity $subscriberEntity) {
...
+    }
+
+    return $subscriber;
+  }
+
+  /**
+   * @return SegmentEntity[]
+   * @throws APIException
+   */
+  private function getAndValidateSegments(array $listIds, string $context = self::CONTEXT_SUBSCRIBE): array {
```
johnolek: Do you think it might be good to remove the default value for the `$context`? It looks like you're explicitly including the value everywhere this method is invoked, which I prefer to relying on default values. No big deal either way though.

johnolek (changes requested):
@costasovo This looks good to me overall! I'm only marking this as requesting changes because of one very small question about a difference in the order we're retrieving data and how that could result in different error messages before/after the refactor in one edge case.

The ticket also mentions checking/refactoring the `\MailPoet\API\MP\v1\CustomFields` class, but it looks like that class was already refactored (b368a7659375532b6e044161dff4ba4ac2f4e2e0). I just wanted to triple check with you that there's nothing else we need to do with the `CustomFields` class to consider all the requirements of the ticket complete.

mailpoet/lib/API/MP/v1/Subscribers.php
costasovo: Good point. I'll remove it.

mailpoet/lib/API/MP/v1/Subscribers.php
costasovo: I was thinking that it was not a big deal. We can't fix it simply by changing the order of those methods because, in that case, we would change the order of subsequent checks. So I guess I could extract the basic checks of inputs and call them before so that we have consistent order of checks. 
I can imagine some plugin might be using an error message for its logic, so to be safe, I'll rather fix it.

costasovo:
@johnolek I've addressed your comments. Please see the two new commits.

> The ticket also mentions checking/refactoring the \MailPoet\API\MP\v1\CustomFields class, but it looks like that class was already refactored (https://github.com/mailpoet/mailpoet/commit/b368a7659375532b6e044161dff4ba4ac2f4e2e0). I just wanted to triple check with you that there's nothing else we need to do with the CustomFields class to consider all the requirements of the ticket complete.

I understood this mention of CustomFields as that I should get inspired by how the refactor was done for the custom field. It is mentioned in all MP API related tickets.

johnolek (approved):
Thanks for making those changes! And sorry about misinterpreting the comment in the ticket about the other code!

veljkho (approved):

PR #4305, opened 2022-08-09
Remove all usages of MailPoet\Tasks\Sending::getRunningQueues()[MAILPOET-4366]

lysyjan:
[MAILPOET-4366]

[MAILPOET-4366]: https://mailpoet.atlassian.net/browse/MAILPOET-4366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Entities/ScheduledTaskEntity.php
+++ b/mailpoet/lib/Entities/ScheduledTaskEntity.php
@@ -22,7 +22,6 @@ class ScheduledTaskEntity {
   const STATUS_PAUSED = 'paused';
   const STATUS_INVALID = 'invalid';
   const VIRTUAL_STATUS_RUNNING = 'running'; // For historical reasons this is stored as null in DB
-  const TYPE_SENDING = 'sending';
   const PRIORITY_HIGH = 1;
   const PRIORITY_MEDIUM = 5;
   const PRIORITY_LOW = 10;
--- a/mailpoet/lib/Entities/ScheduledTaskEntity.php
+++ b/mailpoet/lib/Entities/ScheduledTaskEntity.php
@@ -21,6 +21,7 @@ class ScheduledTaskEntity {
   const STATUS_PAUSED = 'paused';
   const STATUS_INVALID = 'invalid';
   const VIRTUAL_STATUS_RUNNING = 'running'; // For historical reasons this is stored as null in DB
+  const TYPE_SENDING = 'sending';
   const PRIORITY_HIGH = 1;
   const PRIORITY_MEDIUM = 5;
   const PRIORITY_LOW = 10;
--- a/mailpoet/lib/Entities/ScheduledTaskEntity.php
+++ b/mailpoet/lib/Entities/ScheduledTaskEntity.php
@@ -6,6 +6,7 @@ use DateTimeInterface;
 use MailPoet\Doctrine\EntityTraits\AutoincrementedIdTrait;
 use MailPoet\Doctrine\EntityTraits\CreatedAtTrait;
 use MailPoet\Doctrine\EntityTraits\DeletedAtTrait;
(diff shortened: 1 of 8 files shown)


mailpoet/lib/Newsletter/Sending/ScheduledTasksRepository.php
```diff
@@ -125,6 +125,25 @@ public function findFutureScheduledByType($type, $limit = null) {
...
+   * @return ScheduledTaskEntity[]
+   */
+  public function findRunningSendingTasks(?int $limit = null): array {
+    return $this->doctrineRepository->createQueryBuilder('st')
+      ->select('st')
+      ->join('st.sendingQueue', 'sq')
+      ->where('st.type = :type')
+      ->andWhere('st.status IS NULL')
+      ->andWhere('st.deletedAt IS NULL')
+      ->andWhere('sq.deletedAt IS NULL')
```
costasovo: I noticed that [the original query](https://github.com/mailpoet/mailpoet/blob/9b71a7804cd7d3c8724c9ac877cc62def4d3a697/mailpoet/lib/Tasks/Sending.php#L317-L326) was not checking `deletedAt` in the sending_queue table.
Do you think it is necessary to check it? I mean, we could avoid the join here and maybe make the query more efficient.

mailpoet/lib/Entities/ScheduledTaskEntity.php
```diff
@@ -21,6 +21,7 @@ class ScheduledTaskEntity {
   const STATUS_PAUSED = 'paused';
   const STATUS_INVALID = 'invalid';
   const VIRTUAL_STATUS_RUNNING = 'running'; // For historical reasons this is stored as null in DB
+  const TYPE_SENDING = 'sending';
```
costasovo: Other scheduled task type constants are declared in related worker classes (e.g. [#](https://github.com/mailpoet/mailpoet/blob/5373c811b2525de7a586e850a2c8fa9cfc04c168/mailpoet/lib/Cron/Workers/Bounce.php#L24), [#](https://github.com/mailpoet/mailpoet/blob/af396021db8ee4da756409d53c731b08c93c6926/mailpoet/lib/Cron/Workers/InactiveSubscribers.php#L12)).
So I would suggest declaring this type constant in [the SendingQueue worker](https://github.com/mailpoet/mailpoet/blob/8d4799b050bf058b8bb73ceb056c5f02ca0a33ec/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L32).
Also please note that this constant is already defined in https://github.com/mailpoet/mailpoet/blob/9b71a7804cd7d3c8724c9ac877cc62def4d3a697/mailpoet/lib/Tasks/Sending.php#L26
So maybe we could remove it from there.

costasovo (commented):
Good job! Overall looks good! I have just one remark and one question.

mailpoet/lib/Newsletter/Sending/ScheduledTasksRepository.php
lysyjan: You are right. We can remove the condition for sending queues.

mailpoet/lib/Entities/ScheduledTaskEntity.php
lysyjan: Thank you for this tip. I agree!

costasovo (dismissed):

veljkho (changes requested):
Seems not verified to me
I'm getting [call to undefined method](https://d.pr/i/3SDjUy) error when accessing `Subscribers Engagement` tab in newsletter's statistics page.

veljkho (approved):

veljkho:
Seems verified after rebasing it. Going to merge.

PR #4307, opened 2022-08-10
MAILPOET-4355 refactor usages of tasks\state

samnajian:
[MAILPOET-4355]

[MAILPOET-4355]: https://mailpoet.atlassian.net/browse/MAILPOET-4355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo:
@samnajian, the CircleCI build failed due to an error in an integration test. Judging by the name of the test that failed, it could be related to the changes suggested here. Could you please take a look before I start the review?

samnajian:
> @samnajian, the CircleCI build failed due to an error in an integration test. Judging by the name of the test that failed, it could be related to the changes suggested here. Could you please take a look before I start the review?

Just fixed, sorry didn't notice it earlier :(

mailpoet/lib/AdminPages/Pages/Help.php
```diff
@@ -117,4 +127,33 @@ private function getLatestActionSchedulerActionDate(string $hook, string $status
     }
     return null;
   }
+
+  private function buildTaskData(ScheduledTaskEntity $task): array {
```
rodrigoprimo: I'd be curious to hear your reasoning for moving buildTaskData() to Help instead of to ScheduledTasksRepository. I don't see a problem with this path, I'm just wondering.

mailpoet/lib/Newsletter/Sending/ScheduledTasksRepository.php
```diff
@@ -153,6 +154,43 @@ public function getCountsPerStatus(string $type = 'sending') {
...
+      ->setParameter('statuses', $statuses)
+      ->setMaxResults($limit);
+
+    if ($type) {
+      $tasksQuery = $tasksQuery->andWhere('st.type = :type')
+        ->setParameter('type', $type);
+    }
+
+    if (in_array(ScheduledTaskEntity::VIRTUAL_STATUS_RUNNING, $statuses)) {
+      $tasksQuery = $tasksQuery->orWhere('st.status IS NULL');
```
rodrigoprimo: The resulting query for this new version of this method is:

```
SELECT w0_.type AS type_0, w0_.status AS status_1, w0_.priority AS priority_2, w0_.scheduled_at AS scheduled_at_3, w0_.processed_at AS processed_at_4, w0_.meta AS meta_5, w0_.in_progress AS in_progress_6, w0_.reschedule_count AS reschedule_count_7, w0_.id AS id_8, w0_.created_at AS created_at_9, w0_.updated_at AS updated_at_10, w0_.deleted_at AS deleted_at_11
FROM wp_mailpoet_scheduled_tasks w0_
WHERE (w0_.status IN ('completed', 'scheduled', 'running') AND w0_.type = 'sending') OR w0_.status IS NULL
LIMIT 5
```

I could be wrong but my impression is that the logic inside the `where` part of the query is incorrect. I think it should be `w0_.status IN ('completed', 'scheduled', 'running') AND (w0_.type = 'sending' OR w0_.status IS NULL)` instead of `(w0_.status IN ('completed', 'scheduled', 'running') AND w0_.type = 'sending') OR w0_.status IS NULL`. Do you agree or am I missing something?

Another thing that I wonder is if we can have both `running` and `NULL` representing the `ScheduledTaskEntity::VIRTUAL_STATUS_RUNNING` status or just `NULL`. I believe it is just `NULL` but I'm saying this based only on this if condition. I haven't checked anything else.

mailpoet/tests/integration/AdminPages/HelpTest.php
```diff
@@ -0,0 +1,107 @@
...
+    $this->truncateEntity(NewsletterEntity::class);
+    $this->truncateEntity(SendingQueueEntity::class);
+  }
+
+  function _after() {
+    parent::_after(); // TODO: Change the autogenerated stub
+    $this->cleanup();
+  }
+
+}
```
rodrigoprimo: We are missing a new line at the of the file

mailpoet/tests/integration/AdminPages/HelpTest.php
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private function cleanup() {
+    $this->truncateEntity(ScheduledTaskEntity::class);
+    $this->truncateEntity(NewsletterEntity::class);
+    $this->truncateEntity(SendingQueueEntity::class);
+  }
+
+  function _after() {
+    parent::_after(); // TODO: Change the autogenerated stub
```
rodrigoprimo: I guess we can remove this autogenerated code comment

mailpoet/tests/integration/AdminPages/HelpTest.php
```diff
@@ -0,0 +1,107 @@
...
+  private $sendingQueuesRepository;
+
+  /**
+   * @var Help
+   */
+  private $helpPage;
+
+  public function _before() {
+    parent::_before();
+    $this->cleanup();
```
rodrigoprimo, line 36: Do we need to call cleanup() here considering that we are already calling it on _after()? AFAIK, removing everything from the database after the test is executed is enough and it is not necessary to do the same again before starting the execution of the next test.

mailpoet/tests/integration/AdminPages/HelpTest.php
```diff
@@ -0,0 +1,107 @@
...
+      Carbon::now()->addDay()
+    );
+    $data = $this->helpPage->buildTaskData($task);
+    expect($data['newsletter']['newsletter_id'])->equals(null);
+    expect($data['newsletter']['queue_id'])->equals(null);
+    expect($data['newsletter']['subject'])->equals(null);
+    expect($data['newsletter']['preview_url'])->equals(null);
+  }
+
+  public function createNewSendingQueue(?ScheduledTaskEntity $task, ?NewsletterEntity $newsletter, $renderedSubject = null) {
```
rodrigoprimo: This is just a detail, but I believe this method could be private as I don't see a reason to use it outside of this class. Also, I haven't looked into the details, but I wonder if instead of using this method we could create the newsletter calling the method `\MailPoet\Test\DataFactories\Newsletter::withSendingQueue()` which should create a sending queue?

mailpoet/lib/AdminPages/Pages/Help.php
```diff
@@ -128,10 +128,10 @@ private function getLatestActionSchedulerActionDate(string $hook, string $status
     return null;
   }
 
-  private function buildTaskData(ScheduledTaskEntity $task): array {
+  public function buildTaskData(ScheduledTaskEntity $task): array {
```
rodrigoprimo, line 131: Another small detail, but it seems to me that the title of this commit is a bit misleading. It mentions moving buildTaskData() to Pages\Help, but the method already exists in this class.

rodrigoprimo (commented):
Thanks for refactoring yet another class and getting us closer to removing Paris, Sam. I left a few inline comments with questions and suggestions. Could you please take a look when you have a moment?

mailpoet/lib/AdminPages/Pages/Help.php
samnajian: 1) The `buildTaskData`  is only used in Help.php ( probably not even usable elsewhere since it's very much coupled with what's expected in the template of this page)
2) Repository seems to be a good place to interact with DB and fetch data from it, but this method is simply changing the structure of the data ( other the side-effect `$queue = $this->sendingQueuesRepository->findOneBy(['task' => $task]);` that's there and is simply copied around)

mailpoet/tests/integration/AdminPages/HelpTest.php
samnajian: I've seen tests that leave stuff in DB and make other tests provide falls negatives or positives, and besides cleaning too much is better than cleaning too less! On the other hand your assumption is true, but only to the point if all tests clean after themselves, but in cases that some test suits fail to do so can get problematic and hard to debug in new tests.

mailpoet/tests/integration/AdminPages/HelpTest.php
samnajian: I agree that this can be private and basically data factories defined in [another PR](https://github.com/mailpoet/mailpoet/pull/4313/files#diff-19e3f169ce739f526e5702b016ea72e09f52fde590cf7fc91a4148753b6128ecR11) can be used here as well, but this method is assigning `$task` to `$queue` as well, see 
```  
   if ($task instanceof ScheduledTaskEntity) {
      $queue->setTask($task);
    }
```

mailpoet/lib/AdminPages/Pages/Help.php
samnajian: Thanks for noticing this @rodrigoprimo, adjusted the commit title.

mailpoet/lib/Newsletter/Sending/ScheduledTasksRepository.php
samnajian: 1) Thanks for noticing the issue in the query, [that's fixed now ](https://github.com/mailpoet/mailpoet/pull/4307/commits/cd61d8cfbc0b02e296b6d20e8f2b0cf88513513b)
2) Currently I've seen `NULL` as representing `running`, but there is no harm in checking for `running` as well if we end up persisting `running` to actually mean the task is in running status, see 
```
class ScheduledTaskEntity {
  const STATUS_COMPLETED = 'completed';
  const STATUS_SCHEDULED = 'scheduled';
  const STATUS_PAUSED = 'paused';
  const STATUS_INVALID = 'invalid';
  const VIRTUAL_STATUS_RUNNING = 'running'; // For historical reasons this is stored as null in DB
```
I'm also fine with removing `running` from the query, but it shouldn't be problematic.

mailpoet/tests/integration/AdminPages/HelpTest.php
rodrigoprimo: That is a fair point. My concern here, and I haven't measured anything so I'm talking anecdotally, is that by running the delete queries twice for each test we are making our integration test suite unnecessarily slower and I believe we should aim to make it as fast as possible. 

I would suggest that instead of running the cleanup code twice, we strive to find and fix the tests that are not cleaning up after themselves whenever we run into errors caused by such tests. Did you run into errors with your tests because of other tests that were not removing data after they were executed? If so, I'd be happy to help you to trace down which tests are causing the problem. Usually creating a MailPoetTest::_before() method that prints the result of a query that counts the number of existing entries of a given entity and then running the whole test suite is enough to find the tests that are leaving data behind. One caveat is that it is necessary to use `codecept_debug()` instead of the normal PHP functions like `var_dump()`.

If there are no errors when running the tests that you created without calling `cleanup()` inside `_before()`, in my opinion, we should remove it because of the added overhead.

mailpoet/tests/integration/AdminPages/HelpTest.php
rodrigoprimo: I'm not sure I follow why assigning `$task` to `$queue` as well is a problem in this particular case and it prevents using `withSendingQueue()` or `withScheduledQueue()`. We can always get the `$task` from the `$newsletter` and we might modify the data factory if we need to set specific options for the sending queue or task that are not possible with the current interface. But I don't think this is important enough and I'm code with leaving the code as is.

mailpoet/lib/Newsletter/Sending/ScheduledTasksRepository.php
rodrigoprimo: > Thanks for noticing the issue in the query, that's fixed now 

:+1: 

> Currently I've seen NULL as representing running, but there is no harm in checking for running as well if we end up persisting running to actually mean the task is in running status, see

I'm ok with leaving `running` as well.

rodrigoprimo:
Thanks for the changes, @samnajian. I marked all the conversations as resolved except for one where I left a comment and I would like to discuss the issue with you further.

rodrigoprimo (dismissed):

veljkho:
Hi @samnajian, there are some conflicts, could you please rebase?

veljkho (approved):

PR #4308, opened 2022-08-10
Rename workflow step [MAILPOET-4445]

websupporter:
Fixes [MAILPOET-4445]

A `Step` can now have a name. This name can be edited. In the UI a developer can use the new `<StepName step={step} />` to render the interface in the edit component of the step.

In MAILPOET-4445 we discussed briefly to make this edit button for all steps as a general setting, but the designs do not allow for it in the case of the Delay step, as where you would expect the name of the step there is currently the title of another setting. But with the `<StepName />` component, we can quickly extend this to the delay if needed.

[MAILPOET-4445]: https://mailpoet.atlassian.net/browse/MAILPOET-4445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-automation-editor/_panel.scss
+++ b/mailpoet/assets/css/src/components-automation-editor/_panel.scss
@@ -34,21 +34,17 @@
   }
 
   .mailpoet-automation-panel-plain-body-title-action {
-    margin-top: -13px;
+    margin-top: -10px;
     padding-right: 0;
     padding-top: 0;
   }
 }
 
-.mailpoet-step-name-input,
-.mailpoet-step-name-description,
-.mailpoet-step-name-title {
-  min-width: 250px;
+.mailpoet-step-name-popover {
+  margin-top: -25px;
+  padding: 8px;
 }
 
-.mailpoet-step-name-title {
-  font-size: 13px;
-  font-weight: normal;
-  line-height: 1.5;
-  margin: 1em 0;
+.mailpoet-step-name-input {
+  min-width: 208px;
 }
--- a/mailpoet/assets/css/src/components-automation-editor/_panel.scss
+++ b/mailpoet/assets/css/src/components-automation-editor/_panel.scss
@@ -20,3 +20,35 @@
     padding: 4px;
   }
 }
+
+.mailpoet-step-name-dropdown {
+  display: block;
+
+  h2 {
+    margin: 0;
+  }
+
+  .mailpoet-automation-panel-plain-body-title-text {
+    padding-left: 0;
+    padding-top: 0;
+  }
+
+  .mailpoet-automation-panel-plain-body-title-action {
+    margin-top: -13px;
+    padding-right: 0;
+    padding-top: 0;
+  }
+}
+
+.mailpoet-step-name-input,
+.mailpoet-step-name-description,
+.mailpoet-step-name-title {
+  min-width: 250px;
+}
+
+.mailpoet-step-name-title {
+  font-size: 13px;
+  font-weight: normal;
+  line-height: 1.5;
+  margin: 1em 0;
+}
--- a/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
@@ -41,7 +41,7 @@ export function StepName({ step }: Props): JSX.Element {
             }}
           />
           <p className="mailpoet-step-name-description">
-            Give the automation step a name that indicates it’s purpose. E.g
+            Give the automation step a name that indicates its purpose. E.g
             “Abandoned cart recovery”. <br />
             This name will be displayed only to you and not to the clients.
           </p>
--- a/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
@@ -19,7 +19,8 @@ export function StepName({ step }: Props): JSX.Element {
   return (
     <Dropdown
       className="mailpoet-step-name-dropdown"
-      position="bottom center"
+      contentClassName="mailpoet-step-name-popover"
+      position="bottom left"
       renderToggle={({ isOpen, onToggle }) => (
         <PlainBodyTitle
           title={step.name && step.name.length > 0 ? step.name : stepType.title}
@@ -34,22 +35,17 @@ export function StepName({ step }: Props): JSX.Element {
         </PlainBodyTitle>
       )}
       renderContent={() => (
-        <>
-          <h3 className="mailpoet-step-name-title">Step name</h3>
-          <TextControl
-            className="mailpoet-step-name-input"
-            placeholder={stepType.title}
-            value={step.name}
-            onChange={(value) => {
-              dispatch(store).updateStepName(step.id, value);
-            }}
-          />
-          <p className="mailpoet-step-name-description">
-            Give the automation step a name that indicates its purpose. E.g
-            “Abandoned cart recovery”. <br />
-            This name will be displayed only to you and not to the clients.
-          </p>
-        </>
+        <TextControl
+          label="Step name"
+          className="mailpoet-step-name-input"
+          placeholder={stepType.title}
+          value={step.name}
+          onChange={(value) => {
+            dispatch(store).updateStepName(step.id, value);
+          }}
+          help="Give the automation step a name that indicates its purpose. E.g
+            “Abandoned cart recovery”. This name will be displayed only to you and not to the clients."
+        />
       )}
     />
   );
--- a/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
@@ -24,7 +24,11 @@ export function StepName({ step }: Props): JSX.Element {
         <PlainBodyTitle
           title={step.name && step.name.length > 0 ? step.name : stepType.title}
         >
-          <TitleActionButton onClick={onToggle} aria-expanded={isOpen}>
+          <TitleActionButton
+            onClick={onToggle}
+            aria-expanded={isOpen}
+            aria-label="Edit step name"
+          >
             <Icon icon={edit} size={16} />
           </TitleActionButton>
         </PlainBodyTitle>
--- a/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
@@ -1,21 +1,18 @@
 import { Dropdown, TextControl } from '@wordpress/components';
 import { edit, Icon } from '@wordpress/icons';
-import { dispatch, useSelect } from '@wordpress/data';
 import { PlainBodyTitle } from './plain-body-title';
 import { TitleActionButton } from './title-action-button';
-import { store } from '../../store';
-import { Step } from '../workflow/types';
 
 type Props = {
-  step: Step;
+  currentName: string;
+  defaultName: string;
+  update: (value: string) => void;
 };
-export function StepName({ step }: Props): JSX.Element {
-  const { stepType } = useSelect(
-    (select) => ({
-      stepType: select(store).getStepType(step.key),
-    }),
-    [],
-  );
+export function StepName({
+  currentName,
+  defaultName,
+  update,
+}: Props): JSX.Element {
   return (
     <Dropdown
       className="mailpoet-step-name-dropdown"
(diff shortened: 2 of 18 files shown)


mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
```diff
@@ -33,11 +33,14 @@ function getIcon(step: StepData): JSX.Element | null {
 
 function getTitle(step: StepData): string {
   // mocked data
-  if (step.type === 'trigger') {
-    return 'Trigger';
+  if (step.key === 'mailpoet:segment:subscribed') {
+    return 'Subscribed to segment';
+  }
```
johnolek: Do we not want to keep this as `Trigger`? This seems to be duplicating the title and default name with a test automation, which is not what we have in the figma design.

In general it seems that the code in this branch is operating differently than I would expect, changing the top level title of a step when the figma seems to suggest we want that to stay the same and it's the subtitle that gets changed when you edit the name.

## Running this branch locally
image
image


## Figma
image
image

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
```diff
@@ -0,0 +1,52 @@
...
+            placeholder={stepType.title}
+            value={step.name}
+            onChange={(value) => {
+              dispatch(store).updateStepName(step.id, value);
+            }}
+          />
+          <p className="mailpoet-step-name-description">
+            Give the automation step a name that indicates it’s purpose. E.g
+            “Abandoned cart recovery”. <br />
+            This name will be displayed only to you and not to the clients.
```
johnolek: I know it's in the design, and I admire your dedication to reproducing the text from the design exactly, but can we at least fix the `it's` and make it `its`?

mailpoet/tests/integration/Automation/Integrations/Core/Actions/DelayActionTest.php
```diff
@@ -81,7 +81,7 @@ public function dataForTestItCalculatesDelayTypesCorrectly() : array {
    */
   public function testDelayActionInvalidatesOutsideOfBoundaries(int $delay, bool $expectation) {
 
-    $step = new Step("1", 'core:delay', 'core:delay', 'next-step', [
+    $step = new Step("1", null,'core:delay', 'core:delay', 'next-step', [
```
johnolek: Looks like there's a missing space after `null,` and there's the same issue in a few other spots (I'll spare you additional comments for the other instances)

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
```diff
@@ -0,0 +1,52 @@
...
+  );
+  return (
+    <Dropdown
+      className="mailpoet-step-name-dropdown"
+      position="bottom center"
+      renderToggle={({ isOpen, onToggle }) => (
+        <PlainBodyTitle
+          title={step.name && step.name.length > 0 ? step.name : stepType.title}
+        >
+          <TitleActionButton onClick={onToggle} aria-expanded={isOpen}>
```
johnolek: If we have `aria-expanded` I think it would also be appropriate to set an [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#buttons), otherwise I don't think it's clear to assistive technology what is actually open.

mailpoet/assets/css/src/components-automation-editor/_panel.scss
```diff
@@ -20,3 +20,35 @@
...
+
+  .mailpoet-automation-panel-plain-body-title-action {
+    margin-top: -13px;
+    padding-right: 0;
+    padding-top: 0;
+  }
+}
+
+.mailpoet-step-name-input,
+.mailpoet-step-name-description,
```
johnolek: I'm not seeing the step name description get the same styling that we have for the instructions in the automation renaming thing, such as the lighter grey text and slightly smaller font size.

## Figma
image

## My local
image

## My local automation renaming thing
image

mailpoet/assets/css/src/components-automation-editor/_panel.scss
```diff
@@ -20,3 +20,35 @@
...
+  .mailpoet-automation-panel-plain-body-title-text {
+    padding-left: 0;
+    padding-top: 0;
+  }
+
+  .mailpoet-automation-panel-plain-body-title-action {
+    margin-top: -13px;
+    padding-right: 0;
+    padding-top: 0;
+  }
```
johnolek: The figma shows a blue outline around this element in the active state, which I'm not seeing in my local.
## Figma
image

## Local
image

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { stepType } = useSelect(
+    (select) => ({
+      stepType: select(store).getStepType(step.key),
+    }),
+    [],
+  );
+  return (
+    <Dropdown
+      className="mailpoet-step-name-dropdown"
+      position="bottom center"
```
johnolek: This positioning doesn't seem to be working for me to reproduce what I'm seeing in Figma. I see a big gap between the action button and the form.

## Figma
image

## My local firefox
image

## My local Safari
image

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+export function StepName({ step }: Props): JSX.Element {
+  const { stepType } = useSelect(
+    (select) => ({
+      stepType: select(store).getStepType(step.key),
+    }),
+    [],
+  );
+  return (
+    <Dropdown
```
johnolek, line 17: This seems like an edge case, but it's possible to get into a state where you can't click outside of the input to close it.



https://user-images.githubusercontent.com/8656640/184240883-f1c2c0b2-b957-494e-8dc6-e3d4ac60689f.mov

mailpoet/assets/js/src/automation/editor/store/reducer.ts
```diff
@@ -50,6 +50,23 @@ export function reducer(state: State, action: Action): State {
...
+        workflowData: {
+          ...state.workflowData,
+          steps: {
+            ...state.workflowData.steps,
+            [action.stepId]: step,
+          },
+        },
+        selectedStep: step,
+      };
+    }
```
johnolek: This looks almost identical to `UPDATE_STEP_ARGS`. Is there a reason we can't simply use `UPDATE_STEP_ARGS` instead? And is there a reason this action doesn't mark the workflow as being unsaved like `UPDATE_STEP_ARGS` does on L87?
```
workflowSaved: false
```

johnolek (changes requested):
A lot of my comments are small style issues that for all I know could be due to an issue with my local environment, so if you're not able to reproduce some of those issues please disregard those comments.

My biggest concern is that the actual behavior seems to be different than my interpretation of the figma. Based on the designs it looks like we want to always keep the step type as the top title and the potentially-user-supplied-name underneath (more details in individual comments).

mailpoet/assets/css/src/components-automation-editor/_panel.scss
websupporter: It is in the `focus` state. I checked now with the Gutenberg editor. There we have the same behavior. Blue on focus but not on active state. I think, we should follow the precedence. 
image

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
websupporter: Can reproduce. This seems to be a more general issue with the `<DropDown />` component. It happens also in the dropdown where you can change the automation name. Not sure, if we can tackle this in this PR :thinking:

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
websupporter: To be honest, I am unsure of this. We can not really use `aria-controls` as the ID we could use is not yet in the DOM. Using `Tab` for navigation and screenreader: The behavior of `DropDown` is such, that the component sets the focus inside the DropDown, so `aria-controls` seems not really needed. Not sure about `aria-expanded`. It feels a bit useless as in the moment I this state changes to `true` I am inside the DropDown and moving outside (e.g. to focus again on the toggle button) means, the state is again `false`. But it is [mentioned of the documentation](https://developer.wordpress.org/block-editor/reference-guides/components/dropdown/) and we use it also here https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/automation/editor/components/header/document_actions.tsx#L85. So I would rather opt for it to stay. 

I asked [internally in the #a11y slack](https://a8c.slack.com/archives/C0FGK24F9/p1660283591739749), maybe they know a bit more.

websupporter:
Thank you so much for the review @johnolek!

> My biggest concern is that the actual behavior seems to be different than my interpretation of the figma

Thanks for pointing that out. I address this issue in 90a0f16. It lead me to a new conceptualization of where the name should be stored. As this area you are pointing to is rendered by the StepType::subtitle(), I decided to use the Step.args as the place where to store the data, which also means, we do not need a `UPDATE_STEP_NAME` in the reducer.

Regarding the layout issues I addressed those in b2bbc21a2

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
johnolek: Thanks for the explanation and for asking the question in slack! As of right now it looks like no one has responded. That being said, leaving it as-is makes sense to me for consistency if nothing else, and we can revisit if we find out this is an actual accessibility problem, which seems unlikely.

mailpoet/assets/css/src/components-automation-editor/_panel.scss
johnolek: That makes total sense! I imagine that's what was intended, but even if not consistency seems like what we should be shooting for.

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
johnolek: Agreed that would definitely be out of scope

mailpoet/lib/Automation/Integrations/MailPoet/Templates/WorkflowBuilder.php
```diff
@@ -58,7 +58,7 @@ public function welcomeEmailSequence(string $name): Workflow {
     $sendFirstEmailStep = $this->sendEmailActionStep();
     $firstDelayStep->setNextStepId($sendFirstEmailStep->getId());
 
-    $secondDelayStep = $this->delayStep( null,"HOURS");
+    $secondDelayStep = $this->delayStep( null, "HOURS");
```
johnolek: Only pointing this out because there are merge conflicts anyway that need to be resolved: there's still a leading space before `null` and we could use single quotes for the string. Same on line 55.

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
```diff
@@ -24,7 +24,11 @@ export function StepName({ step }: Props): JSX.Element {
         <PlainBodyTitle
           title={step.name && step.name.length > 0 ? step.name : stepType.title}
         >
-          <TitleActionButton onClick={onToggle} aria-expanded={isOpen}>
+          <TitleActionButton
+            onClick={onToggle}
+            aria-expanded={isOpen}
+            aria-label="Edit step name"
```
johnolek, line 28: Thanks for adding this!

johnolek:
Thanks for making those updates, @websupporter! I noticed one more styling thing, which is that the step name should be bold (`font-weight: 600` according to figma). Other than that (and one tiny comment) this looks good to me once you've resolved the merge conflicts.

mailpoet/lib/Automation/Integrations/MailPoet/Templates/WorkflowBuilder.php
websupporter: Why isn't this catched by our coding standards?

websupporter:
Thanks @johnolek, rebased and fixed :)

johnolek (approved):

mailpoet/lib/Automation/Integrations/MailPoet/Templates/WorkflowBuilder.php
johnolek: I was wondering the same thing myself! Maybe there's a PhpStan rule we should add? I don't know too much about PhpStan myself and after a little bit of googling I was no closer to figuring out how we would add a rule to catch this.

veljkho (approved):

PR #4312, opened 2022-08-11
Premium upgrade upsell modal [MAILPOET-4416][MAILPOET-4559]

JanJakes:
### Description
None

### Note for QA
Jira tickets [MAILPOET-4416] & [MAILPOET-4559]
1) Go to Automation page
2) Create a new Welcome workflow template and open it
3) On the right sidebar, click on Google Analytics tab
4) Click on the switcher to trigger the upsell modal
5) Please test all scenarios according to Jira specs

[MAILPOET-4416]: https://mailpoet.atlassian.net/browse/MAILPOET-4416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-4559]: https://mailpoet.atlassian.net/browse/MAILPOET-4559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Util/License/Features/Subscribers.php
```diff
@@ -18,22 +19,30 @@ class Subscribers {
   const PREMIUM_EMAIL_VOLUME_LIMIT_SETTING_KEY = 'premium.premium_key_state.data.email_volume_limit';
   const PREMIUM_EMAILS_SENT_SETTING_KEY = 'premium.premium_key_state.data.emails_sent';
   const PREMIUM_SUPPORT_SETTING_KEY = 'premium.premium_key_state.data.support_tier';
+  const SUBSCRIBERS_COUNT_CACHE_KEY = 'mailpoet_subscribers_count';
+  const SUBSCRIBERS_COUNT_CACHE_EXPIRATION_MINUTES = 60;
+  const SUBSCRIBERS_COUNT_CACHE_MIN_VALUE = 1000;
```
websupporter, line 24: In effect we cache an SQL query. Why don't we cache it always and start only above this threshold? To me it seems to introduce an unnecessary complication.

JanJakes: What do you mean by caching always? Moving it to a repository? Isn't it caching always now?

websupporter: I mean we cache the value only, when it is `> 1000` right? And I wonder why. https://github.com/mailpoet/mailpoet/pull/4312/files#diff-2d1730f301f3e90af93f2d5482c6446631bfc10e95b0743cba6ed3f0e4da4b2dR69-R71

mailpoet/assets/js/src/global.d.ts
```diff
@@ -102,7 +102,6 @@ interface Window {
   mailpoet_wp_week_starts_on: number;
   mailpoet_subscribers_counts_cache_created_at: string;
   mailpoet_shortcode_links: string[];
-  mailpoet_settings: any;
   mailpoet_tracking_config: string;
```
websupporter: This is not a string but an object. Maybe we can update it to the correct type `{level:"basic"|"partial" etc...}`?

websupporter (changes requested):
Hi @JanJakes, thanks for the PR.

Two general comments:

> When the user tries to add a step, we’ll show a popup that a paid plan is required.

Nothing happens when I click the plus button.

Upgrading with an older version
When I upgrade from an older version of the free plugin, it loads the latest premium plugin and I can not use the feature, see https://d.pr/v/xM2GEI

mailpoet/lib/Util/License/Features/Subscribers.php
JanJakes: Ahh, right. The idea was to provide immediate feedback on count changes when the plugin is freshly installed and tested. Some people may try to check if everything works and some delayed info could be confusing. I also thought it was less of a BC break so safer not to break stuff — simply caching only for large numbers of subscribers. Do you think we should remove that logic? I'd feel a bit safer about these changes keeping it, I guess.

JanJakes:
> When the user tries to add a step, we’ll show a popup that a paid plan is required.
> Nothing happens when I click the plus button.

Hmm, https://mailpoet.atlassian.net/browse/MAILPOET-4416 evolved to complex modal logic, and this is also covered by the last point in https://mailpoet.atlassian.net/browse/PREMIUM-194. I'd maybe rephrase MAILPOET-4416 to only implement the modal and keep the add-step modal to PREMIUM-194. Is that OK, @NeosinneR?

> Upgrading with an older version
> When I upgrade from an older version of the free plugin, it loads the latest premium plugin, and I can not use the feature, see https://d.pr/v/xM2GEI

That may be also a question for @NeosinneR. Do you think we should download the latest compatible premium version? Or solve this differently? Should we create another ticket for that?

NeosinneR:
> I'd maybe rephrase MAILPOET-4416 to only implement the modal and keep the add-step modal to PREMIUM-194. Is that OK, @NeosinneR?

Sounds good. Specs updated. 

> Do you think we should download the latest compatible premium version?

@JanJakes can we? Release service can return any premium plugin version, but how would you know which specific to ask for? Major and minor are the same, but in some cases, we released [the .1 version](https://github.com/mailpoet/mailpoet-premium/releases?q=&expanded=true) also for the premium. Those are especially cases when you don't want the .0. How about also updating free at the same time when the premium is downloaded?

mailpoet/assets/js/src/global.d.ts
JanJakes: Fixed in https://github.com/mailpoet/mailpoet/pull/4312/commits/4312b9aa9e71356f75b4da5249615eef2d512a31.

JanJakes:
> @JanJakes can we? Release service can return any premium plugin version, but how would you know which specific to ask for? Major and minor are the same, but in some cases, we released [the .1 version](https://github.com/mailpoet/mailpoet-premium/releases?q=&expanded=true) also for the premium. Those are especially cases when you don't want the .0. How about also updating free at the same time when the premium is downloaded?

@NeosinneR Good point. Could you pls create a ticket for this, so we don't forget to resolve it? We can leave it as WIP specs and discuss it further.

NeosinneR:
Created https://mailpoet.atlassian.net/browse/MAILPOET-4579

websupporter (dismissed):
lgtm. thank you @JanJakes

veljkho:
Hi @JanJakes please resolve conflicting files.

JanJakes:
@veljkho Rebased.

veljkho (approved):
Good job! @JanJakes ✅

PR #4314, opened 2022-08-11
MAILPOET-4371 remove old model usage from post notification scheduler and replace it with doctrine

samnajian:
[MAILPOET-4371]

[MAILPOET-4371]: https://mailpoet.atlassian.net/browse/MAILPOET-4371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
+++ b/mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
@@ -2,16 +2,20 @@
 
 namespace MailPoet\Newsletter\Scheduler;
 
+use MailPoet\Cron\Workers\SendingQueue\SendingQueue;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\NewsletterOptionEntity;
 use MailPoet\Entities\NewsletterOptionFieldEntity;
+use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Logging\LoggerFactory;
 use MailPoet\Newsletter\NewsletterPostsRepository;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Options\NewsletterOptionFieldsRepository;
 use MailPoet\Newsletter\Options\NewsletterOptionsRepository;
-use MailPoet\Tasks\Sending as SendingTask;
+use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
+use MailPoet\Newsletter\Sending\SendingQueuesRepository;
+use MailPoet\WP\DateTime;
 use MailPoet\WP\Posts;
 
 class PostNotificationScheduler {
@@ -43,12 +47,20 @@ class PostNotificationScheduler {
   /** @var Scheduler */
   private $scheduler;
 
+  /*** @var ScheduledTasksRepository */
+  private $scheduledTasksRepository;
+
+  /*** @var SendingQueuesRepository */
+  private $sendingQueuesRepository;
+
   public function __construct(
     NewslettersRepository $newslettersRepository,
     NewsletterOptionsRepository $newsletterOptionsRepository,
     NewsletterOptionFieldsRepository $newsletterOptionFieldsRepository,
     NewsletterPostsRepository $newsletterPostsRepository,
-    Scheduler $scheduler
+    Scheduler $scheduler,
+    ScheduledTasksRepository $scheduledTasksRepository,
+    SendingQueuesRepository $sendingQueuesRepository
   ) {
     $this->loggerFactory = LoggerFactory::getInstance();
     $this->newslettersRepository = $newslettersRepository;
@@ -56,6 +68,8 @@ class PostNotificationScheduler {
     $this->newsletterOptionFieldsRepository = $newsletterOptionFieldsRepository;
     $this->newsletterPostsRepository = $newsletterPostsRepository;
     $this->scheduler = $scheduler;
+    $this->scheduledTasksRepository = $scheduledTasksRepository;
+    $this->sendingQueuesRepository = $sendingQueuesRepository;
   }
 
   public function transitionHook($newStatus, $oldStatus, $post) {
@@ -95,7 +109,7 @@ class PostNotificationScheduler {
     }
   }
 
-  public function createPostNotificationSendingTask(NewsletterEntity $newsletter): ?SendingTask {
+  public function createPostNotificationSendingTask(NewsletterEntity $newsletter): ?ScheduledTaskEntity {
     $notificationHistory = $this->newslettersRepository->findSendigNotificationHistoryWithPausedTask($newsletter);
     if (count($notificationHistory) > 0) {
       return null;
@@ -105,7 +119,7 @@ class PostNotificationScheduler {
     if (!$scheduleOption) {
       return null;
     }
-    $nextRunDate = $this->scheduler->getNextRunDate($scheduleOption->getValue());
+    $nextRunDate = $this->scheduler->getNextRunDateTime($scheduleOption->getValue());
     if (!$nextRunDate) {
       return null;
     }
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Entities/ScheduledTaskEntity.php
```diff
@@ -26,6 +26,7 @@ class ScheduledTaskEntity {
   const PRIORITY_LOW = 10;
   const BASIC_RESCHEDULE_TIMEOUT = 5; // minutes
   const MAX_RESCHEDULE_TIMEOUT = 1440; // minutes
+  const DEFAULT_TASK_TYPE = 'sending';
```
lysyjan: We declare task type constants in workers. It was discussed https://github.com/mailpoet/mailpoet/pull/4305#discussion_r942309426  in another PR recently. Also, the sending type is not the default.

mailpoet/lib/Newsletter/Scheduler/Scheduler.php
```diff
@@ -81,4 +77,18 @@ public function getNewsletters(string $type, ?string $group = null): array {
   public function formatDatetimeString($datetimeString) {
     return Carbon::parse($datetimeString)->format('Y-m-d H:i:s');
   }
+
+  /**
+   * @return \DateTime|false
+   */
+  public function getNextRunDateTime($schedule, $fromTimestamp = false) {
```
lysyjan, line 84: Is this change necessary? I see that you created it probably due to storing `scheduledAt`. But maybe it would be easier to use `Carbon::createFromFormat` because we know the format. What do you think? You could also avoid formatting the date into logs.

mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
```diff
@@ -95,7 +108,11 @@ public function schedulePostNotification($postId) {
     }
   }
 
-  public function createPostNotificationSendingTask(NewsletterEntity $newsletter): ?SendingTask {
+  /**
```
lysyjan: I guess this annotation is redundant. In my opinion, the return type and parameter type are enough.

mailpoet/tests/integration/Newsletter/Scheduler/PostNotificationTest.php
```diff
@@ -63,22 +71,26 @@ public function testItCreatesPostNotificationSendingTask() {
     ]);
 
     // new queue record should be created
-    $queue = $this->postNotificationScheduler->createPostNotificationSendingTask($newsletter);
-    assert($queue instanceof SendingTask);
-    expect(SendingQueue::where('newsletter_id', $newsletter->getId())->findMany())->count(1);
-    expect($queue->newsletterId)->equals($newsletter->getId());
-    expect($queue->status)->equals(SendingQueue::STATUS_SCHEDULED);
+    $sendingQueue = $this->postNotificationScheduler->createPostNotificationSendingTask($newsletter);
+    assert($sendingQueue instanceof SendingQueueEntity);
```
lysyjan: We want to remove all asserts from our codebase. There already exists [a ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4258) for this.
If I'm not wrong, you can use `$this->assertInstanceOf()` to avoid errors from PHPStan.

You don't have to change the old code, but we shouldn't add more of those cases.

mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
```diff
@@ -95,7 +108,11 @@ public function schedulePostNotification($postId) {
...
+   * @param NewsletterEntity $newsletter
+   * @return SendingQueueEntity|null
+   */
+  public function createPostNotificationSendingTask(NewsletterEntity $newsletter): ?SendingQueueEntity {
```
lysyjan: This method has name `createPostNotificationSendingTask` and you return `SendingQueueEntity`. I would expect to return `ScheduledTaskEntity` instead. Or is there a reason for this? I checked it, and we use the returned value only in tests.

mailpoet/tests/integration/Newsletter/Scheduler/PostNotificationTest.php
```diff
@@ -63,22 +71,26 @@ public function testItCreatesPostNotificationSendingTask() {
...
+    assert($sendingQueue->getTask() instanceof ScheduledTaskEntity);
+    expect($this->sendingQueuesRepository->findBy(['newsletter' => $newsletter]))->count(1);
+    if ($sendingQueue->getNewsletter())
+      expect($sendingQueue->getNewsletter()->getId())->equals($newsletter->getId());
```
lysyjan: I think that this check shouldn't be behind a condition. Please, don't forget to check the rest of your changes.

mailpoet/tests/integration/Newsletter/Scheduler/PostNotificationTest.php
```diff
@@ -289,17 +304,25 @@ public function testSchedulerWontRunIfUnsentNotificationHistoryExists() {
...
     $notificationHistory->setParent($newsletter);
     $notificationHistory->setSubject($newsletter->getSubject());
     $this->newslettersRepository->persist($notificationHistory);
     $this->newslettersRepository->flush();
 
-    $sendingTask = SendingTask::create();
-    $sendingTask->newsletterId = $notificationHistory->getId();
-    $sendingTask->status = SendingQueue::STATUS_SCHEDULED;
-    $sendingTask->save();
+    $task = new ScheduledTaskEntity();
```
lysyjan: You can use [ScheduledTask factory](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/DataFactories/ScheduledTask.php) here.

mailpoet/tests/integration/Newsletter/Scheduler/PostNotificationTest.php
```diff
@@ -289,17 +304,25 @@ public function testSchedulerWontRunIfUnsentNotificationHistoryExists() {
...
+    $task->setType(ScheduledTaskEntity::DEFAULT_TASK_TYPE);
+    $task->setStatus( SendingQueueEntity::STATUS_SCHEDULED);
+    $this->scheduledTasksRepository->persist($task);
+    $this->scheduledTasksRepository->flush();
+
+    $this->sendingQueuesRepository->flush();
```
lysyjan: This flush is redundant.

mailpoet/tests/integration/Newsletter/Scheduler/PostNotificationTest.php
```diff
@@ -289,17 +304,25 @@ public function testSchedulerWontRunIfUnsentNotificationHistoryExists() {
...
+    $sendingQueue = new SendingQueueEntity();
```
lysyjan: I have created a SendingQueue factory in [another PR](https://github.com/mailpoet/mailpoet/pull/4305/commits/a0cf7a817de934f720a3c907f0c7848fd62dbb1a). Unfortunately, it is not merged yet.

lysyjan (changes requested):
I have a few remarks about your changes.

mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
samnajian: The reason is that we have `$sendingQueue->getTask()` but not a method to get the queue from `ScheduledTaskEntity` and the returned values are only used in test as you mentioned, still I'd agree the name doesn't match much, will adapt

lysyjan: @samnajian It's possible to change the unidirectional relation to bidirectional. I did it in a different PR in this commit https://github.com/mailpoet/mailpoet/pull/4305/commits/33fe72bb5d1dcef0a1eb9ed1167cb6fc3543a09b.
You can read how it works [here](https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/association-mapping.html#one-to-one-bidirectional).

You changed the method's name in the newest version, but other schedulers have the suffix `SendingTask`. So I think that we should preserve this pattern.

mailpoet/lib/Newsletter/Scheduler/Scheduler.php
samnajian: I'd say this is necessary since I'd normally expect a higher lower level class to give me `raw`  information and not formatted one, and besides why should we format and create a string out of `DateTime` and then use the string to create `DateTime` again? It doesn't make much sense even though it's pretty easy to do so.

mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
samnajian: Since you've added the code needed here in [this PR](https://github.com/mailpoet/mailpoet/pull/4305) it makes sense to either wait until the mentioned PR is merged or based [PR 4314](https://github.com/mailpoet/mailpoet/pull/4314) on it, what do you think?

lysyjan: @samnajian Yes, it makes sense.

mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
```diff
@@ -114,18 +128,32 @@ public function createPostNotificationSendingTask(NewsletterEntity $newsletter):
...
+    $this->scheduledTasksRepository->persist($sendingTask);
+    $this->scheduledTasksRepository->flush();
+
+    $sendingQueue = new SendingQueueEntity();
+    $sendingQueue->setNewsletter($newsletter);
+    $sendingQueue->setTask($sendingTask);
+    $this->sendingQueuesRepository->persist($sendingQueue);
+    $this->sendingQueuesRepository->flush();
+    $sendingTask->setSendingQueue($sendingQueue);
+    $this->scheduledTasksRepository->flush();
```
lysyjan: I guess this flush is redundant because the sending queue is the owner of the relation.

mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
```diff
@@ -114,18 +128,32 @@ public function createPostNotificationSendingTask(NewsletterEntity $newsletter):
...
+    if ($scheduledAt && $scheduledAt->format('Y-m-d H:i:s') === $nextRunDate->format('Y-m-d H:i:s')) {
       return null;
     }
 
-    $sendingTask = SendingTask::create();
-    $sendingTask->newsletterId = $newsletter->getId();
-    $sendingTask->status = SendingQueueEntity::STATUS_SCHEDULED;
-    $sendingTask->scheduledAt = $nextRunDate;
-    $sendingTask->save();
+    $sendingTask = new ScheduledTaskEntity();
```
lysyjan: This is a detail only, but the variable should be named `scheduledTask`. I'm fine with that, this is not a blocker.

lysyjan (commented):
Only one small detail.

lysyjan:
This PR should be merged after https://github.com/mailpoet/mailpoet/pull/4305

lysyjan (dismissed):
Thank you for the changes and patience.

veljkho (approved):

PR #4315, opened 2022-08-11
Prevent the user from leaving the settings page without saving [MAILPOET-4499]

rodrigoprimo:
I'm creating this PR as a draft first to get early feedback on the implementation before I finish it. As I don't have a lot of experience with React there is a big change I'm missing a lot of things.

A few questions/comments that I have:

- I opted to reuse the component UnsavedChangesNotice from the form editor and pass the name of the store when creating the component. I'm not entirely sure if that is a good idea or not.
- I don't know if I could use the spread operator instead of clone to avoid having to create the newState variable in https://github.com/mailpoet/mailpoet/compare/add-confirmation-message-to-settings-page?expand=1#diff-6dfcee87da4b090835dfe995637b6974c1079a63dbaea7aa426b42fac230ee4fR20.
- I'm using `SET_SETTINGS` to set hasUnsavedChanges to `false` instead of `SAVE_DONE` as settings are saved when the user verifies a new premium key but in this flow `SAVE_DONE` is not called and `SET_SETTINGS` is. For some reason that I still have to figure out, `SET_SETTING` is also called (I don't know where yet) when the user verifies the key causing a bug as `hasUnsavedChanges` is set to `true` but there are no unsaved changes.
- I might need to change the name of the selector `hasUnsavedChanges()` to something more generic as in the settings page we are checking for unsaved changes but also if the page is being saved. Please let me know if you have any suggestions.

[MAILPOET-4499]

[MAILPOET-4499]: https://mailpoet.atlassian.net/browse/MAILPOET-4499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/common/notices/unsaved_changes_notice.jsx
+++ b/mailpoet/assets/js/src/common/notices/unsaved_changes_notice.jsx
@@ -1,6 +1,6 @@
 import { useEffect } from 'react';
 import { useSelect } from '@wordpress/data';
-import { MailPoet } from '../../mailpoet';
+import { MailPoet } from 'mailpoet';
 
 export function UnsavedChangesNotice({ storeName }) {
   const hasUnsavedChanges = useSelect(
--- a/mailpoet/assets/js/src/common/notices/unsaved_changes_notice.jsx
+++ b/mailpoet/assets/js/src/common/notices/unsaved_changes_notice.jsx
@@ -1,10 +1,10 @@
 import { useEffect } from 'react';
-import { MailPoet } from 'mailpoet';
 import { useSelect } from '@wordpress/data';
+import { MailPoet } from '../../mailpoet';
 
-export function UnsavedChangesNotice() {
+export function UnsavedChangesNotice({ storeName }) {
   const hasUnsavedChanges = useSelect(
-    (sel) => sel('mailpoet-form-editor').hasUnsavedChanges(),
+    (sel) => sel(storeName).hasUnsavedChanges(),
     [],
   );
 
--- a/mailpoet/assets/js/src/form_editor/components/editor.jsx
+++ b/mailpoet/assets/js/src/form_editor/components/editor.jsx
@@ -21,7 +21,7 @@ import { Tutorial } from './tutorial';
 import { Sidebar } from './sidebar/sidebar';
 import { Inserter } from './inserter';
 import { Notices } from './notices.jsx';
-import { UnsavedChangesNotice } from './unsaved_changes_notice.jsx';
+import { UnsavedChangesNotice } from '../../common/notices/unsaved_changes_notice.jsx';
 import { FormStyles } from './form_styles.jsx';
 import { FormPreview } from './preview/preview';
 import { FormStylingBackground } from './form_styling_background.jsx';
@@ -133,7 +133,7 @@ export function Editor() {
                   )}
                   <div className="interface-interface-skeleton__content">
                     <Notices />
-                    <UnsavedChangesNotice />
+                    <UnsavedChangesNotice storeName="mailpoet-form-editor" />
                     <BlockSelectionClearer className="edit-post-visual-editor editor-styles-wrapper">
                       <BlockTools>
                         <BlockEditorKeyboardShortcuts />
--- a/mailpoet/assets/js/src/form_editor/components/editor.jsx
+++ b/mailpoet/assets/js/src/form_editor/components/editor.jsx
@@ -15,13 +15,13 @@ import {
 } from '@wordpress/block-editor';
 import { ShortcutProvider } from '@wordpress/keyboard-shortcuts';
 
+import { UnsavedChangesNotice } from 'common/notices/unsaved_changes_notice.jsx';
 import { fetchLinkSuggestions } from '../utils/link_suggestions';
 import { Header } from './header.jsx';
 import { Tutorial } from './tutorial';
 import { Sidebar } from './sidebar/sidebar';
 import { Inserter } from './inserter';
 import { Notices } from './notices.jsx';
-import { UnsavedChangesNotice } from '../../common/notices/unsaved_changes_notice.jsx';
 import { FormStyles } from './form_styles.jsx';
 import { FormPreview } from './preview/preview';
 import { FormStylingBackground } from './form_styling_background.jsx';
--- a/mailpoet/assets/js/src/settings/store/create_reducer.ts
+++ b/mailpoet/assets/js/src/settings/store/create_reducer.ts
@@ -1,4 +1,4 @@
-import { setWith, clone } from 'lodash';
+import { setWith, clone, isEqual } from 'lodash';
 import {
   State,
   Action,
@@ -24,7 +24,10 @@ export function createReducer(defaultValue: State) {
           clone,
         );
 
-        newState.save.hasUnsavedChanges = true;
+        newState.save.hasUnsavedChanges = !isEqual(
+          newState.data,
+          state.originalData,
+        );
 
         return newState;
       }
--- a/mailpoet/assets/js/src/settings/store/create_reducer.ts
+++ b/mailpoet/assets/js/src/settings/store/create_reducer.ts
@@ -24,7 +24,7 @@ export function createReducer(defaultValue: State) {
           clone,
         );
 
-        newState.hasUnsavedChanges = true;
+        newState.save.hasUnsavedChanges = true;
 
         return newState;
       }
@@ -32,16 +32,32 @@ export function createReducer(defaultValue: State) {
         return {
           ...state,
           data: normalizeSettings(action.value as Record<string, unknown>),
-          hasUnsavedChanges: false,
         };
       case 'SET_ERROR_FLAG':
         return { ...state, flags: { ...state.flags, error: !!action.value } };
       case 'SAVE_STARTED':
-        return { ...state, save: { inProgress: true, error: null } };
+        return {
+          ...state,
+          save: {
+            inProgress: true,
+            error: null,
+            hasUnsavedChanges: state.save.hasUnsavedChanges,
+          },
+        };
       case 'SAVE_DONE':
-        return { ...state, save: { inProgress: false, error: null } };
+        return {
+          ...state,
+          save: { inProgress: false, error: null, hasUnsavedChanges: false },
+        };
       case 'SAVE_FAILED':
-        return { ...state, save: { inProgress: false, error: action.error } };
+        return {
+          ...state,
+          save: {
+            inProgress: false,
+            error: action.error,
+            hasUnsavedChanges: state.save.hasUnsavedChanges,
+          },
+        };
       case 'SET_RE_ENGAGEMENT_NOTICE':
         return {
           ...state,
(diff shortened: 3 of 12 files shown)


rodrigoprimo:
Just noting here that some acceptance tests are failing and they seem to be failing due to the changes made here. I will check and fix the tests after I get the initial feedback on the proposed implementation.

mailpoet/assets/js/src/form_editor/components/editor.jsx
```diff
@@ -21,7 +21,7 @@ import { Tutorial } from './tutorial';
 import { Sidebar } from './sidebar/sidebar';
 import { Inserter } from './inserter';
 import { Notices } from './notices.jsx';
-import { UnsavedChangesNotice } from './unsaved_changes_notice.jsx';
+import { UnsavedChangesNotice } from '../../common/notices/unsaved_changes_notice.jsx';
```
costasovo: This is just a detail, but some time ago, we agreed that in cases when you need to go to the upper directory `../` we should rather use the absolute path (in this case `/common/notices/unsaved_changes_notice.jsx`).
With this, it is easier to move `editor.jsx`. We don't have a lint for that and I was not able to find a conversation about it. So it is probably more like a recommendation and we need to believe my memory :). Just mentioning it for the next time.

mailpoet/assets/js/src/common/notices/unsaved_changes_notice.jsx
```diff
@@ -1,10 +1,10 @@
 import { useEffect } from 'react';
-import { MailPoet } from 'mailpoet';
 import { useSelect } from '@wordpress/data';
+import { MailPoet } from '../../mailpoet';
```
costasovo: You can use absolute path import `import { MailPoet } from 'mailpoet'`;

mailpoet/assets/js/src/settings/store/create_reducer.ts
```diff
@@ -16,17 +16,23 @@ export function createReducer(defaultValue: State) {
...
-        return setWith(
+      case 'SET_SETTING': {
+        const newState = setWith(
           clone(state),
           ['data', ...action.path],
           action.value,
           clone,
         );
+
+        newState.hasUnsavedChanges = true;
```
costasovo: I would consider putting this under `state.save` so `newState.save.hasUnsavedChanges = true;`. I think it is closely related to saving as it holds the info about unsaved changes, and it resets when saving is done. I think it is a good practice (which we didn't always follow) to keep the root of the store transparent.

mailpoet/assets/js/src/settings/store/create_reducer.ts
```diff
@@ -16,17 +16,23 @@ export function createReducer(defaultValue: State) {
...
+
+        return newState;
+      }
       case 'SET_SETTINGS':
         return {
           ...state,
           data: normalizeSettings(action.value as Record<string, unknown>),
+          hasUnsavedChanges: false,
```
costasovo: I think understand why is the reset of `hasUnsavedChanges` done here, but I find it a bit confusing because without searching for usages of 'SET_SETTINGS' action just from naming I would assume that this action should set the `hasUnsavedChanges` to the opposite.
Maybe we could rename the action to `SET_SETTINGS_FROM_SERVER` or `SET_SAVED_SETTINGS` so that it is more obvious. Or another option could be resetting the flag on `SAVE_DONE` action. That would require yielding the action also [in `verifyMssKey` action](https://github.com/mailpoet/mailpoet/blob/d116ff360b8d73f0ae09429ae0c365ed30a54def/mailpoet/assets/js/src/settings/store/actions/mss_and_premium.ts#L14).
What do you think?

costasovo (commented):
Overall looks good. I added a few notes about imports and a suggestion about `SET_SETTINGS` action.

> I don't know if I could use the spread operator instead of clone to avoid having to create the newState

I think your approach is fine. In general, the reducer has to return a new object instead of modifying the original object. Cloning it once is enough. We don't need to clone it with every modification we do in a reducer function.

costasovo:
> I might need to change the name of the selector hasUnsavedChanges() to something more generic as in the settings page we are checking for unsaved changes but also if the page is being saved. Please let me know if you have any suggestions

I think my suggestion of moving the `hasUnsavedChanges` under `save` in the state may be also a solution to this.

mailpoet/assets/js/src/form_editor/components/editor.jsx
rodrigoprimo: Thanks for sharing this information. I believe the correct syntax is `common/notices/unsaved_changes_notice.jsx` (without the first slash). I added a new commit using absolute paths for all the relative paths that I had added.

mailpoet/assets/js/src/settings/store/create_reducer.ts
rodrigoprimo: Good point. You had mentioned it in our initial conversation and I completely forgot about it. I just implemented it in a new commit.

mailpoet/assets/js/src/settings/store/create_reducer.ts
rodrigoprimo: I agree that the name `SET_SETTINGS` can be misleading and it confused me when I first read the code. That being said, I opted to yield `SAVE_DONE` in `verifyPremiumKey()` as I wasn't sure how complex it would be to rename `SET_SETTINGS` and also just renaming it is not enough. When the user clicks on the `Verify` button, after `SET_SETTINGS` runs, `SET_SETTING` also runs changing state.save.hasUnsavedChanges to true.

rodrigoprimo:
@costasovo, I fixed the acceptance tests that were failing and this PR is now ready for a final review.

costasovo (dismissed):

veljkho (changes requested):
This is working as expected, but I have 2 scenarios that are suspicious to me:
1. When I do some change and revert it back to normal state like I did no change, I still see dialog that I'll lose changes if I leave...
2. When on `Settings` > `Send with` tab.. and when enabled `MSS`: Then, when you click on `Configure` link like you want to see `Other` sending method, and then just clicking `Cancel` link without any changes, then you'll see also dialog if you want to go away.

These 2 scenarios are bad UX, maybe not an issue as you've implemented the feature correctly to show for any change in the settings. I want to check with you if this is worth fixing and even fixable. Should it be fixed at all? GIF: https://d.pr/i/18LPMU

Thanks!

rodrigoprimo:
Sorry for the delay to reply, @veljkho.

> When I do some change and revert it back to normal state like I did no change, I still see dialog that I'll lose changes if I leave...

I agree that is bad UX. I discussed this with Rosta when we were talking about how to implement this solution and decided to leave it as is due to the complexity of tracking if the state changed back to the original state or not. But I will revisit this now and see if I can implement a solution that takes this into consideration.

> When on Settings > Send with tab.. and when enabled MSS: Then, when you click on Configure link like you want to see Other sending method, and then just clicking Cancel link without any changes, then you'll see also dialog if you want to go away.

I missed this case. I will fix it.

I should be able to work on this PR early next week.

rodrigoprimo:
@costasovo, I'm requesting a new review from you in this PR as I added a new commit to display the confirmation message only if there is actually a change in any of the settings to address one of the concerns that @veljkho raised.

> When on Settings > Send with tab.. and when enabled MSS: Then, when you click on Configure link like you want to see Other sending method, and then just clicking Cancel link without any changes, then you'll see also dialog if you want to go away.

@veljkho, the UI is not great here (and maybe we can create a new ticket to improve it), but I actually think we should preserve the current behavior of displaying the confirmation message after the user clicks on the `Configure` link. The reason for that is that, behind the scenes, when this link is clicked, the user is effectively changing the `Send with...` option from `MSS` to `Other` so there is indeed one unsaved change. What do you think?

rodrigoprimo:
@costasovo, please see the commit message of the last commit for more details on my reasoning for choosing the approach that I chose. I'm not sure if there is a better way to achieve the same.

mailpoet/assets/js/src/settings/store/selectors.ts
```diff
@@ -99,5 +99,8 @@ export function getReEngagementAction(state: State) {
 }
 
 export function hasUnsavedChanges(state: State) {
-  return state.save.hasUnsavedChanges || state.save.inProgress;
+  return (
+    (state.save.hasUnsavedChanges || state.save.inProgress) &&
```
costasovo: Did you consider adding `!_.isEqual(state.data, state.originalData)` into the reducer where we set the `state.save.hasUnsavedChanges` state to true? Theoretically, a selector hook might be called on every rerender of a component where it is used, so having that in the reducer, could be more efficient.

If we decide to keep the comparison of states here in the selector, do we still need to keep tracking the state `state.save.hasUnsavedChanges`?  After you added the comparison between the current and original data, it looks unnecessary and might even be confusing.

mailpoet/assets/js/src/settings/store/create_reducer.ts
```diff
@@ -16,13 +16,18 @@ export function createReducer(defaultValue: State) {
...
-        return setWith(
+      case 'SET_SETTING': {
+        const newState = setWith(
           clone(state),
           ['data', ...action.path],
           action.value,
           clone,
         );
+
+        newState.save.hasUnsavedChanges = true;
```
costasovo: Maybe instead of always setting true here, we could use `!_.isEqual(newState.data, state.originalData)` and avoid doing it in the selector method.

costasovo (commented):
@rodrigoprimo Thanks! I like the idea that we use the original data to detect a change. I think it might be nicer to do it in the reducer instead of the selector. Please see my other comments.

rodrigoprimo:
Thanks for your input, @costasovo.

> If we decide to keep the comparison of states here in the selector, do we still need to keep tracking the state state.save.hasUnsavedChanges? After you added the comparison between the current and original data, it looks unnecessary and might even be confusing.

I guess I got tunnel vision while working on this, focused too much on making it work, and didn't pause to consider that there was no point in keeping `state.save.hasUnsavedChanges` after I added the check in the selector.

In the end, I opted to follow your other suggestion and moved the check to the reducer. Please check this PR again.

costasovo (dismissed):
Good job! Thanks 👍 

> I guess I got tunnel vision while working on this, focused too much on making it work, and didn't pause to consider that there was no point in keeping state.save.hasUnsavedChanges after I added the check in the selector.

No worries! I think that also the fact that in redux, we divide the logic to action, reducer, and selectors kind of supports this tunnel vision as you don't see it all in one place.

rodrigoprimo:
Good point

veljkho:
Hey @rodrigoprimo , thanks for working on this. Now when I tested, I see that there's no anymore dialog when clicking `Cancel` link... however, I can see it when I try to move away from the settings page. Is this the fix for `Cancel` link? Asking because I am not sure... When the user cancels `Other` settings and then back to `Send with` screen.. `MSS` is again ticked as nothing has changed. Why do I see still dialog here if I want to move away?

> when this link is clicked, the user is effectively changing the Send with... option from MSS to Other so there is indeed one unsaved change.
As you've mentioned here, you might be right here so I don't see why the behavior has changed now to see dialog after clicking `Cancel` and then moving away with `MSS` ticked.

rodrigoprimo:
Thanks, @veljkho. I missed this problem with the "Or Cancel" button. It behaves differently than the rest of the page and that is why it was not working. It should be working now as expected. Could you please check again?

veljkho (approved):

PR #4316, opened 2022-08-12
Replace MailPoet\Models\Newsletter::createNotificationHistory() with Doctrine [MAILPOET-4310]

lysyjan:
[MAILPOET-4310]

[MAILPOET-4310]: https://mailpoet.atlassian.net/browse/MAILPOET-4310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/Scheduler.php
+++ b/mailpoet/lib/Cron/Workers/Scheduler.php
@@ -5,8 +5,10 @@ namespace MailPoet\Cron\Workers;
 use MailPoet\Cron\CronHelper;
 use MailPoet\Cron\CronWorkerScheduler;
 use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Entities\NewsletterSegmentEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SegmentEntity;
+use MailPoet\InvalidStateException;
 use MailPoet\Logging\LoggerFactory;
 use MailPoet\Models\Newsletter;
 use MailPoet\Models\ScheduledTask;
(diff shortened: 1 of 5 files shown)


mailpoet/lib/Cron/Workers/Scheduler.php
```diff
@@ -129,23 +150,25 @@ public function processPostNotificationNewsletter($newsletter, SendingTask $queu
...
-      $segments = $newsletterEntity->getSegmentIds();
-      if (empty($segments)) {
-        $this->loggerFactory->getLogger(LoggerFactory::TOPIC_POST_NOTIFICATIONS)->info(
-          'post notification no segments',
-          ['newsletter_id' => $newsletter->id, 'task_id' => $queue->taskId]
-        );
-        return $this->deleteQueueOrUpdateNextRunDate($queue, $newsletter);
-      }
+    if (!$newsletterEntity instanceof NewsletterEntity) {
+      throw new InvalidStateException();
```
johnolek, line 154: I don't believe the previous version of this code ever threw any exceptions. Could this cause an issue since this method is getting called in a `foreach` that's potentially processing multiple queues? In other words, could it throw an unhandled exception that would prevent the next newsletters from getting processed?

johnolek (commented):
Overall looks good to me! I just have one question about the new (I believe) exception that could get thrown.

mailpoet/lib/Cron/Workers/Scheduler.php
lysyjan: This is a good point, but the exception shouldn't be thrown at any time. We know a newsletter exists in the DB, but we need to load it for Doctrine. Because we refactor the code by parts, we sometimes need to have objects from the old and the new model.
When the whole class is refactored to Doctrine, this check can be removed because the new model will be passed from an argument.

If you still think this solution is too dangerous, I can wrap creating post notifications with `if ($newsletterEntity)` and remove this exception.

johnolek: That makes sense, and thanks for the explanation!

Knowing that we'll be able to remove this once the whole class gets refactored makes me feel much better.

johnolek (approved):

veljkho (approved):

PR #4319, opened 2022-08-16
Improve the DMARC error message - MAILPOET-4311

triple0t:
Screenshot 2022-08-16 at 10 00 29 AM

When the DMARC error message is displayed, add the sentence "Click here to start the authentication". This will be a link that opens the verify domain dialogue

[MAILPOET-4311](https://mailpoet.atlassian.net/browse/MAILPOET-4311)

Testing:
* Clear your errors and make sure you don't have any sending error
* Create a new newsletter. You may use any of the demo contents
* On the newsletter sending page, use an email domain with restricted DMARC policy e.g @automattic.com (You might have to authenticate the email address)
* Fill in the other fields and send the email. (I would recommend the use of a sending list with few subscribers e.g 1 or 2 subscribers)
* Wait a few minutes for the error message 
* Click on the "Click here to start the authentication" link
* Notice the sender domain verification modal.

### DIFF
--- a/mailpoet/assets/js/src/listing/notices.jsx
+++ b/mailpoet/assets/js/src/listing/notices.jsx
@@ -72,12 +72,33 @@ export function MailerError(props) {
       links[match],
       'text/xml',
     ).firstChild;
+
+    const listOfAttributeNames = link.getAttributeNames();
+    const allowedAttributesList = [
+      'target',
+      'rel',
+      'class',
+      'data-email',
+      'data-type',
+    ];
+
+    // include these custom attributes in the final link
+    const otherAttributes = listOfAttributeNames.reduce((acc, name) => {
+      if (allowedAttributesList.includes(name)) {
+        // react requires the class attribute to be named className.
+        return {
+          ...acc,
+          [name === 'class' ? 'className' : name]: link.getAttribute(name),
+        };
+      }
+      return { ...acc };
+    }, {});
+
     return (
       <a
         key={`a-${match}`}
         href={link.getAttribute('href')}
-        target={link.getAttribute('target')}
-        rel={link.getAttribute('rel')}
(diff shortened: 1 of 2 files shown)


mailpoet/assets/js/src/listing/notices.jsx
```diff
@@ -72,12 +72,33 @@ export function MailerError(props) {
...
+    // include these custom attributes in the final link
+    const otherAttributes = listOfAttributeNames.reduce((acc, name) => {
+      if (allowedAttributesList.includes(name)) {
+        // react requires the class attribute to be named className.
+        return {
+          ...acc,
+          [name === 'class' ? 'className' : name]: link.getAttribute(name),
+        };
+      }
+      return { ...acc };
```
samnajian: No need to create another object here, I guess you can simply return `acc`, what do you think?

mailpoet/assets/js/src/listing/notices.jsx
```diff
@@ -72,12 +72,33 @@ export function MailerError(props) {
...
+    }, {});
+
     return (
       <a
         key={`a-${match}`}
         href={link.getAttribute('href')}
-        target={link.getAttribute('target')}
-        rel={link.getAttribute('rel')}
+        {...otherAttributes}
```
samnajian: I feel like using `allowedAttributeNames` instead of `allowedAttributesList` and using `allowedAttributes` instead of `otherAttributes` would be a little easier to read understand what each variable holds, but it's just a suggestion!

samnajian (changes requested):
Works nicely, but a few short comments.

samnajian (approved):
LGTM! thanks :)

veljkho (approved):

PR #4321, opened 2022-08-16
Fix subscribers' segment filter [MAILPOET-4563]

lysyjan:
[MAILPOET-4563]

[MAILPOET-4563]: https://mailpoet.atlassian.net/browse/MAILPOET-4563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscribers/SubscriberListingRepository.php
+++ b/mailpoet/lib/Subscribers/SubscriberListingRepository.php
@@ -260,9 +260,7 @@ class SubscriberListingRepository extends ListingRepository {
     $queryBuilder
       ->select('s')
       ->from(SegmentEntity::class, 's');
-    if ($group === 'trash') {
-      $queryBuilder->andWhere('s.deletedAt IS NOT NULL');
-    } else {
+    if ($group !== 'trash') {
       $queryBuilder->andWhere('s.deletedAt IS NULL');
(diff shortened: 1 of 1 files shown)


triple0t (changes requested):
Thanks for the PR @lysyjan 

I do not believe this PR solves the issues highlighted in the ticket description.
i.e Item 4: `Now try to filter all trashed subscribers by the list (the one you know they are related to)`

e.g
Here is my Trashed list 

Trying to filter trashed subscribers via list
Screenshot 2022-08-17 at 1 04 45 PM
**Notice** the subscribers' list does not show up in the dropdown

Here is the full trashed list
mp3 localhost_wp-admin_admin php_page=mailpoet-subscribers


Can you please take another look?

lysyjan:
@triple0t You forgot to assign this PR back to me, and I overlooked your comment.
Unfortunately, this behavior depends on the segments cache. If you use the recalculate button, it should be okay.

triple0t (dismissed):
I can confirm clicking the "Recalculated now" button updates the list selector.

veljkho (approved):

PR #4322, opened 2022-08-16
Add tracking for MailPoet logo in free plan emails - MAILPOET-4543

triple0t:
[MAILPOET-4543](https://mailpoet.atlassian.net/browse/MAILPOET-4543)

Testing
* Make sure you are on the free plan or use a Free MSS subscription API key
* Create a new newsletter, add demo contents
* Send newsletter
* Notice the MailPoet logo at the bottom of the newsletter
* Click on it. You should be redirected to mailpoet.com with some URL query params

### DIFF
--- a/mailpoet/lib/Newsletter/Renderer/Renderer.php
+++ b/mailpoet/lib/Newsletter/Renderer/Renderer.php
@@ -227,7 +227,7 @@ class Renderer {
           'blocks' => [
             [
               'type' => 'image',
-              'link' => 'http://www.mailpoet.com',
+              'link' => 'https://www.mailpoet.com/?ref=free-plan-user-email&utm_source=free_plan_user_email&utm_medium=email',
               'src' => Env::$assetsUrl . '/img/mailpoet_logo_newsletter.png',
               'fullWidth' => false,
               'alt' => 'Email Marketing Powered by MailPoet',


rodrigoprimo (approved):
Nice work, @triple0t. Code changes look good to me, and I could confirm that the link to mailpoet.com now contains the new query params.

As a side note, I was not able to confirm that the value of `ref` is being added to all links in mailpoet.com that point to account.mailpoet.com as described in the Jira ticket. I'm not sure if I'm missing something while testing or if there is a problem. Anyway, even if there is a problem, it is unlikely to be related to the changes proposed here, but we should probably create another ticket regardless. What do you think, @triple0t?

For now, I'm leaving this PR approved but assigned to you. Feel free to move it to QA whenever you see fit.

triple0t:
Thank you for the review 🙌🏿 @rodrigoprimo

> As a side note, I was not able to confirm that the value of `ref` is being added to all links in mailpoet.com that point to account.mailpoet.com as described in the Jira ticket. I'm not sure if I'm missing something while testing or if there is a problem. Anyway, even if there is a problem, it is unlikely to be related to the changes proposed here, but we should probably create another ticket regardless. What do you think, @triple0t?

The value of `ref` is saved in the cookie for the page.
Screenshot 2022-08-17 at 11 04 08 AM


> As a side note, I was not able to confirm that the value of `ref` is being added to all links in mailpoet.com that point to account.mailpoet.com

I can confirm this as well. This is probably due to a bug in the site Theme code. I will create a ticket for that.

I will be moving this PR to QA.

samnajian:
Sorry to join the party pretty late, I was thinking maybe using `//` as prefix instead of `https//` or `http//` would be a better fit, since it would allow the link to always work without caring about the scheme,  what do you think?

NeosinneR:
@samnajian what benefit would it bring linking to `http://mailpoet.com` (when the user's website is on `http://`)? I see the value of using just `//` if you need to load external assets (styles, scripts) because the browser won't be complaining of using `http://` assets on `https://` website. But when linking outside, using the correct protocol can save 1 unnecessary redirect.

samnajian:
> @samnajian what benefit would it bring linking to `http://mailpoet.com` (when the user's website is on `http://`)? I see the value of using just `//` if you need to load external assets (styles, scripts) because the browser won't be complaining of using `http://` assets on `https://` website. But when linking outside, using the correct protocol can save 1 unnecessary redirect.

I see only two benefits in this
1) If for any reason our TLS expires even for a fraction of seconds this link would work
2) We can use this probable redirect ( if not done via our load balancer by only in application layer ) to record if the user is on https or not, but this benefit is behind lots of ifs 

Overall speaking this won't bring a big benefit.

veljkho (approved):

PR #4323, opened 2022-08-17
Workflow details [MAILPOET-4417]

websupporter:
### Description
None

### Note for QA
Jira ticket [MAILPOET-4417](https://mailpoet.atlassian.net/browse/MAILPOET-4417)
1) Go to Automation page
2) Click to add a new template and open it
3) Please follow the specification from Jira ticket and apply testing around those specs


alex-mailpoet:
Issues found:
* "Automation details" section is not collapsible;
* When I try to trash a workflow, I get "An error occurred!" error message, no indication of successful trashing given, status in the UI remains unchanged, although it is changed to `trash` in the DB;
* Trashed workflows are not indicated in listings in any way;
* When I try to open a trashed workflow, I get a blank page and there is an error in the console: `Uncaught TypeError: _mailpoet__WEBPACK_IMPORTED_MODULE_13__.MailPoet.urls is undefined`.

websupporter:
Hi @alex-mailpoet.

Thanks for your review!

> "Automation details" section is not collapsible;

addressed in 3ac687672

> When I try to trash a workflow, I get "An error occurred!" error message

I was able to reproduce the problem. Can you recreate the DB schema and try again?

> Trashed workflows are not indicated in listings in any way;

I would assume, this is part of [MAILPOET-4532](https://mailpoet.atlassian.net/browse/MAILPOET-4532), where the filters are added to the listing page.

> When I try to open a trashed workflow, I get a blank page and there is an error in the console: Uncaught TypeError: _mailpoet__WEBPACK_IMPORTED_MODULE_13__.MailPoet.urls is undefined.

That's odd. 8dca366 is supposed to add this object. I tried it with the release zip and it works. My current assumption is that the twig template did not update in your installation, but I do not know why. Here is the current flow with the release zip: https://d.pr/v/Ffg8dV

alex-mailpoet:
> My current assumption is that the twig template did not update in your installation

It seems like this was the cause. I have cleared Twig cache and things started to work. Thank you for the idea!

More issues:
* I see no trash restoration functionality (spec 3.b.i.1);
* Now, when I try to edit a trashed workflow, I get an intermediary blank page before I get redirected to the listing. In your video it's a blink of an eye, but on my local environment a white screen can last for a few seconds, which is not a great UX, the same can happen with users on slower connections. Avoiding this page would be an improvement, e.g. using a preloader or redirecting on the backend before the page load. But I assume this will be a rarely occurring scenario, especially after listing implementation ([MAILPOET-4532](https://mailpoet.atlassian.net/browse/MAILPOET-4532)), so maybe it's not worth wasting too much time on it.

websupporter:
> Avoiding this page would be an improvement, e.g. using a preloader or redirecting on the backend before the page load

That's a good point! Thank you, addressed in [7f2552e](https://github.com/mailpoet/mailpoet/pull/4323/commits/7f2552e561b32d4ce040748c0713c8d31cb36e05)

> I see no trash restoration functionality (spec 3.b.i.1);

I discussed with @NeosinneR to handle restoring functionality in a separate issue [MAILPOET-4574]. But I missed to update [MAILPOET-4417] accordingly. Did so now. Sorry. 

[MAILPOET-4574]: https://mailpoet.atlassian.net/browse/MAILPOET-4574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4417]: https://mailpoet.atlassian.net/browse/MAILPOET-4417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

alex-mailpoet:
Unfortunately the redirect doesn't work:
> Warning: Cannot modify header information - headers already sent by (output started at /www/wp-admin/includes/template.php:2546) in /www/wp-includes/pluggable.php on line 1421

When a MailPoet admin page code is executed, the WP admin menu is already output to the browser. You should redirect at an earlier stage or solve it in a hacky way like it's done on other admin pages: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/AdminPages/Pages/NewsletterEditor.php#L91-L96
I've tried the above solution, you get a grey blank screen instead of a white one with it, which is marginally better. It's your choice to improve it or revert back.

websupporter:
That's odd @alex-mailpoet. It worked for me, but I see where the headers are send, so I reverted.

alex-mailpoet:
> It worked for me, but I see where the headers are send, so I reverted.

Maybe you have output buffering on by default, this way headers can be set at any point.

alex-mailpoet (dismissed):
Looks good overall!

veljkho:
Hi @websupporter 
Please rebase the PR, I also see some failing acceptance not sure if that's flakiness, could be. Thanks.

veljkho (approved):

mailpoet/assets/js/src/automation/editor/components/workflow/index.tsx
```diff
@@ -4,6 +4,7 @@ import {
   __unstableUseCompositeState as useCompositeState,
 } from '@wordpress/components';
 import { useSelect } from '@wordpress/data';
+import { EditorNotices } from '@wordpress/editor';
```
JanJakes, line 7: @websupporter Ahh, I just needed to use notices and run into this. We're not using the `@wordpress/editor` package and therefore it should never be imported anywhere in the automation code! I wonder if should [add some lints for this](https://eslint.org/docs/latest/rules/no-restricted-imports) but maybe it's not worth it because extracting it to a package within the monorepo with its deps will solve it.

I fixed it in my branch: https://github.com/mailpoet/mailpoet/commit/ad3a3498bd5cb46f26d4fa6f78989b4b469397d7

(This was also the reason why the notices looked different than in WP.)

mailpoet/assets/js/src/automation/editor/components/actions/trash-button.tsx
```diff
@@ -0,0 +1,70 @@
...
+        });
+      })
+      .catch((): void => {
+        void createErrorNotice('An error occurred!', {
+          explicitDismiss: true,
+        });
+      });
+
+    return true;
+  };
```
JanJakes, line 49: @websupporter Another thing I noticed is that all of this should be an action in the editor store. Just to let you know, fixing in my branch.

PR #4325, opened 2022-08-18
Remove old model usage from MailPoet\WooCommerce\Subscription and WooCommerceBlocksIntegration [MAILPOET-4365]

rodrigoprimo:
## Description
N/A

## QA notes
Make sure that users can still subscribe to lists when using the WooCommerce and WooCommerce Blocks checkouts.

## Linked tickets
[MAILPOET-4365]

[MAILPOET-4365]: https://mailpoet.atlassian.net/browse/MAILPOET-4365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Entities/SubscriberEntity.php
+++ b/mailpoet/lib/Entities/SubscriberEntity.php
@@ -447,6 +447,12 @@ class SubscriberEntity {
     });
   }
 
+  public function getSubscribedSegments() {
+    $criteria = Criteria::create()
+      ->where(Criteria::expr()->eq('status', SubscriberEntity::STATUS_SUBSCRIBED));
+    return $this->getSubscriberSegments()->matching($criteria);
+  }
+
   /**
    * @return Collection<int, SubscriberCustomFieldEntity>
    */
--- a/mailpoet/lib/Entities/SubscriberEntity.php
+++ b/mailpoet/lib/Entities/SubscriberEntity.php
@@ -435,8 +435,16 @@ class SubscriberEntity {
   /**
    * @return Collection<int, SubscriberSegmentEntity>
    */
-  public function getSubscriberSegments() {
-    return $this->subscriberSegments;
+  public function getSubscriberSegments(?string $status = null) {
+    if (!is_null($status)) {
+      $criteria = Criteria::create()
+        ->where(Criteria::expr()->eq('status', SubscriberEntity::STATUS_SUBSCRIBED));
+      $subscriberSegments = $this->subscriberSegments->matching($criteria);
+    } else {
+      $subscriberSegments = $this->subscriberSegments;
+    }
+
+    return $subscriberSegments;
   }
 
   public function getSegments() {
@@ -445,12 +453,6 @@ class SubscriberEntity {
     })->filter(function ($segment) {
       return $segment !== null;
     });
-  }
-
-  public function getSubscribedSegments() {
-    $criteria = Criteria::create()
-      ->where(Criteria::expr()->eq('status', SubscriberEntity::STATUS_SUBSCRIBED));
-    return $this->getSubscriberSegments()->matching($criteria);
   }
 
   /**
--- a/mailpoet/lib/WooCommerce/Subscription.php
+++ b/mailpoet/lib/WooCommerce/Subscription.php
@@ -187,17 +187,23 @@ class Subscription {
     $checkoutOptinEnabled = (bool)$this->settings->get(self::OPTIN_ENABLED_SETTING_NAME);
     $checkoutOptin = !empty($_POST[self::CHECKOUT_OPTIN_INPUT_NAME]);
 
-    return $this->handleSubscriberOptin($subscriber, $checkoutOptinEnabled, $checkoutOptin);
+    $subscriberEntity = $this->subscribersRepository->findOneById($subscriber->id);
+
+    if ($subscriberEntity instanceof SubscriberEntity) {
+      return $this->handleSubscriberOptin($subscriberEntity, $checkoutOptinEnabled, $checkoutOptin);
+    }
   }
 
   /**
    * Subscribe or unsubscribe a subscriber.
    *
-   * @param Subscriber $subscriber Subscriber object
+   * @param SubscriberEntity $subscriberEntity Subscriber object
    * @param bool $checkoutOptinEnabled
    * @param bool $checkoutOptin
    */
-  public function handleSubscriberOptin(Subscriber $subscriber, bool $checkoutOptinEnabled, bool $checkoutOptin): bool {
+  public function handleSubscriberOptin(SubscriberEntity $subscriberEntity, bool $checkoutOptinEnabled, bool $checkoutOptin): bool {
+    $subscriber = Subscriber::findOne($subscriberEntity->getId());
+
     $wcSegment = Segment::getWooCommerceSegment();
     $moreSegmentsToSubscribe = (array)$this->settings->get(self::OPTIN_SEGMENTS_SETTING_NAME, []);
     $signupConfirmation = $this->settings->get('signup_confirmation');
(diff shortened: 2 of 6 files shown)


mailpoet/lib/Entities/SubscriberEntity.php
```diff
@@ -447,6 +447,12 @@ public function getSegments() {
     });
   }
 
+  public function getSubscribedSegments() {
```
lysyjan: I have a few notes about this new getter.
- There is no annotation with return type
- The name looks very similar to another getter `getSubscriberSegments`
- The name of this getter evokes in me that it returns an array of the SegmentEntity.
- Maybe adding a parameter with status would be better because it allows better reusability.

What do you think about it?

mailpoet/lib/WooCommerce/Subscription.php
```diff
@@ -229,60 +237,64 @@ public function handleSubscriberOptin(SubscriberEntity $subscriberEntity, bool $
...
-  private function subscribe(Subscriber $subscriber) {
-    $subscriber->status = SubscriberEntity::STATUS_SUBSCRIBED;
-    if (empty($subscriber->confirmedIp) && empty($subscriber->confirmedAt)) {
-      $subscriber->confirmedIp = Helpers::getIP();
-      $subscriber->setExpr('confirmed_at', 'NOW()');
+  private function subscribe(SubscriberEntity $subscriber) {
+    $subscriber->setStatus(SubscriberEntity::STATUS_SUBSCRIBED);
+    if (empty($subscriber->getConfirmedIp()) && empty($subscriber->getConfirmedAt())) {
+      $subscriber->setConfirmedIp(Helpers::getIP());
+      $subscriber->setConfirmedAt(new Carbon());
```
lysyjan, line 259: I think that we should use here a time from WP. Something like `Carbon::createFromTimestamp($this->wp->currentTime('timestamp'))`. Do you agree?

mailpoet/lib/WooCommerce/Subscription.php
```diff
@@ -229,60 +237,64 @@ public function handleSubscriberOptin(SubscriberEntity $subscriberEntity, bool $
...
+      $this->unsubscribesTracker->track((int)$subscriber->getId(), StatisticsUnsubscribeEntity::SOURCE_ORDER_CHECKOUT);
     }
   }
+
+  private function getOptinSegments(): array {
+    $segments = [];
+    $segmentsIds = (array)$this->settings->get(self::OPTIN_SEGMENTS_SETTING_NAME, []);
+
+    foreach ($segmentsIds as $segmentId) {
+      $segments[] = $this->entityManager->getReference(SegmentEntity::class, $segmentId);
```
lysyjan: Why did you use `getReference` here? This looks a little bit unsafe to me. If I am not wrong, `getRefernece` doesn't load the entity from the database, and it's possible to save invalid data with this approach. For example, when a segment in settings is deleted.
When you use `$this->segmentsRepository->findBy(['id' => $segmentIds])`, the code will be simpler and also it's possible to remove `EntityManager` from dependecies.

mailpoet/lib/WooCommerce/Subscription.php
```diff
@@ -169,12 +169,14 @@ public function isCurrentUserSubscribed() {
...
     }
+
     $wcSegment = $this->segmentsRepository->getWooCommerceSegment();
-    $subscriberSegment = SubscriberSegment::where('subscriber_id', $subscriber->getId())
-      ->where('segment_id', $wcSegment->getId())
-      ->findOne();
-    return $subscriberSegment instanceof SubscriberSegment
-      && $subscriberSegment->status === SubscriberEntity::STATUS_SUBSCRIBED;
+    $subscriberSegment = $this->subscriberSegmentRepository->findOneBy(
+      ['subscriber' => $subscriber->getId(), 'segment' => $wcSegment->getId()]
```
lysyjan, line 168: I'm not sure, but maybe you don't know it. Passing IDs here is not necessary because the whole entity is enough, but you don't have to change.

rodrigoprimo: I wasn't aware of that. Thanks for mentioning it.

mailpoet/lib/WooCommerce/Subscription.php
rodrigoprimo: Could you please elaborate on why you think we should use current_time() in this case? Is it because in this case we should respect whatever timezone is configured in WP? I haven't thought about it, it would be a change in the current behavior, but not necessarily a problem since what I'm suggesting is already changing the behavior anyway as the original code used MySQL's NOW().

mailpoet/lib/WooCommerce/Subscription.php
rodrigoprimo: I opted to use `getReference` to avoid hitting the database unnecessarily as we don't need the full entity here. But I missed that it was possible for users to delete segments that are added to the `woocommerce.segments` settings and that this would cause problems as you pointed out. I added a new commit using `findBy()` instead of `getReferece()` and simplifying the code.

rodrigoprimo: By the way, do you think we should create a ticket to add some protection against the user deleting segments that are part of `woocommerce.segments`?

mailpoet/lib/Entities/SubscriberEntity.php
rodrigoprimo: > The name looks very similar to another getter getSubscriberSegments

Yeah, I thought about this but couldn't think of a better name and went it with. But I agree that it has the potential to induce mistakes.

> Maybe adding a parameter with status would be better because it allows better reusability.

What if we add a new optional `$status` parameter to `getSubscriberSegments()`? This way we don't even need to create `getSubscribedSegments()`?

rodrigoprimo:
Thanks for the review, @lysyjan. I replied to your comments. Could you please take a look when you have a moment?

mailpoet/lib/WooCommerce/Subscription.php
lysyjan: > Could you please elaborate on why you think we should use current_time() in this case?

What I know is that we use `current_time()` when we want to store current time. If I'm not wrong, the reason is to respect the user's timezone. Here is [an example](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Subscription/Pages.php#L190). But I guess it's not critical here.

mailpoet/lib/WooCommerce/Subscription.php
lysyjan: >By the way, do you think we should create a ticket to add some protection against the user deleting segments that are part of woocommerce.segments?

👍🏻 I think that we can create a behavior that at least notifies a user about it.

mailpoet/lib/Entities/SubscriberEntity.php
lysyjan: >What if we add a new optional $status parameter to getSubscriberSegments()? This way we don't even need to create getSubscribedSegments()?

This sounds great to me! 👍🏻

mailpoet/lib/WooCommerce/Subscription.php
rodrigoprimo: Yeah, I don't think it is critical and it would change the current behavior which doesn't take into consideration the user timezone set on WP. So I'm inclined to think we don't need to modify it in this PR. Let me know if you agree or not.

mailpoet/lib/WooCommerce/Subscription.php
rodrigoprimo: https://mailpoet.atlassian.net/browse/MAILPOET-4597

mailpoet/lib/Entities/SubscriberEntity.php
rodrigoprimo: @lysyjan, I implemented the approach that we discussed above in https://github.com/mailpoet/mailpoet/pull/4325/commits/422b422d451294e58814528cd6c2b0bf816dcd35. Could you please check this PR again?

lysyjan (approved):
LGTM 👍🏻

Aschepikov (approved):
This is verified, good job.

PR #4327, opened 2022-08-18
Automation templates [MAILPOET-4538]

websupporter:
Fixes [MAILPOET-4538]

A `WorkflowTemplate entity has been added and a WorkflowTemplateStorage`, where the hardcoded templates can be found. With the help of a filter the Premium plugin can add its own templates.

The existing template logic has been rewired to use the storage now. Also, the `POST mailpoet/v1/workflows/create-from-template` endpoint has been altered and a new `GET mailpoet/v1/automation/workflow-templates` endpoint has been added.

[MAILPOET-4538]: https://mailpoet.atlassian.net/browse/MAILPOET-4538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/automation.tsx
+++ b/mailpoet/assets/js/src/automation/automation.tsx
@@ -81,11 +81,15 @@ function App(): JSX.Element {
       <Workflows />
       <div style={{ marginTop: 30, display: 'grid', gridGap: 8 }}>
         <CreateTestingWorkflowButton />
-        <CreateWorkflowFromTemplateButton template="delayed-email-after-signup">
+        <CreateWorkflowFromTemplateButton slug="simple-welcome-email">
           Create testing workflow from template (welcome email)
         </CreateWorkflowFromTemplateButton>
-        <CreateWorkflowFromTemplateButton template="welcome-email-sequence">
-          Create testing workflow from template (welcome sequence)
+        <CreateWorkflowFromTemplateButton slug="welcome-email-sequence">
+          Create testing workflow from template (welcome sequence, only premium)
+        </CreateWorkflowFromTemplateButton>
+        <CreateWorkflowFromTemplateButton slug="advanced-welcome-email-sequence">
+          Create testing workflow from template (advanced welcome sequence, only
+          premium)
         </CreateWorkflowFromTemplateButton>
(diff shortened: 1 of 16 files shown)


alex-mailpoet (dismissed):
LGTM!

JanJakes:
@websupporter I'm catching up on the automation progress, and I run into this. I scanned it only quickly so I'm not sure but did the defaults, implemented [here](https://github.com/mailpoet/mailpoet/commit/08e8e885024bb29dadd4aab2b76250e8fac6a3aa), get lost in the pull request? They are used in the email sidebar for some fields like sender name and address and reply to name and address.

websupporter:
Hello @JanJakes, thanks for chiming in and good catch! What do you think about [7142887](https://github.com/mailpoet/mailpoet/pull/4327/commits/7142887e515e7dab0f41d0ddb72428a1c0c8ad44) and [c34f998](https://github.com/mailpoet/mailpoet/pull/4327/commits/c34f998bfd29aade20f44fc512f5dbf46e5bd80b)?

JanJakes:
@websupporter Thanks for the quick update 👍 I have only one question — could we avoid making it static (so we don't have to use settings and other deps as a singleton)? After all, all of the required steps should be in the registry, so it should be doable by one of these:

1. Add a function to get step by class to the registry.
2. Change [these definitions](https://github.com/mailpoet/mailpoet/pull/4327/files#diff-674f11c160c506817632d24088d9ca99283e602ecd72920eb0e6834eb49ab43eR62-R66) to use step keys instead of classes.

For both options, you'd need to fetch the step instance from the registry somewhere [here](https://github.com/mailpoet/mailpoet/pull/4327/files#diff-d68fb40861791f28d1adf309fbf8adebf3621f7118914d2cfa9cda190dc81b6dR17-R35). Option 2. seems easier to do, I guess.

websupporter:
thank you @JanJakes. Updated the PR accordingly

JanJakes (approved):

veljkho (approved):

PR #4328, opened 2022-08-18
Don't requiring a list to save email draft [MAILPOET-4549]

samnajian:
### Description
This PR allows saving the newsletter as a draft without requiring a list.

- Adds the change.
- Cleans up how the form fields are prepared for the form.

### Note for QA
Jira ticket [MAILPOET-4549](https://mailpoet.atlassian.net/browse/MAILPOET-4549)
1) Go to Emails page
2) Start creating a new newsletter
3) Go to the last step ( step 4 - Send )
4) Verify you can "Save as draft and close" without selecting a list.

### Note for Dev
I tried to see if there is any better way of updating the `data-parsley-required` parameter, I'll appreciate any better approach suggested by the reviewer.

### DIFF
--- a/mailpoet/tests/acceptance/Newsletters/SaveNewsletterAsDraftCest.php
+++ b/mailpoet/tests/acceptance/Newsletters/SaveNewsletterAsDraftCest.php
@@ -5,10 +5,40 @@ namespace MailPoet\Test\Acceptance;
 class SaveNewsletterAsDraftCest {
   public function saveStandardNewsletterAsDraft(\AcceptanceTester $i) {
     $i->wantTo('Create standard newsletter and save as a draft');
-
-    $inactiveNewsletterUI = '.mailpoet-listing-row-inactive';
     $newsletterTitle = 'Testing Newsletter ' . \MailPoet\Util\Security::generateRandomString();
+
     $segmentName = $i->createListWithSubscriber();
+    $inactiveNewsletterUI = '.mailpoet-listing-row-inactive';
+
+    $this->startCreatingNewsletter($i, $newsletterTitle);
+
+    // step 4 - Choose list and send
+    $sendFormElement = '[data-automation-id="newsletter_send_form"]';
+    $i->waitForElement($sendFormElement);
+    $i->selectOptionInSelect2($segmentName);
+    $i->click('Save as draft and close');
+    $i->waitForText($newsletterTitle);
+    $this->assertNewsletterNotSent($i);
+  }
+
+  public function saveStandardNewsletterWithoutListAsDraft(\AcceptanceTester $i) {
+    $i->wantTo('Create standard newsletter and save as a draft without selecting subscribers list');
+    $newsletterTitle = 'Testing Newsletter ' . \MailPoet\Util\Security::generateRandomString();
+
+    $this->startCreatingNewsletter($i, $newsletterTitle);
+
+    // step 4 - Save as draft without selecting a list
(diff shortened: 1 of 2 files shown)


mailpoet/tests/acceptance/Newsletters/SaveNewsletterAsDraftCest.php
```diff
@@ -5,10 +5,40 @@
...
+  }
+
+  public function saveStandardNewsletterWithoutListAsDraft(\AcceptanceTester $i) {
+    $i->wantTo('Create standard newsletter and save as a draft without selecting subscribers list');
     $newsletterTitle = 'Testing Newsletter ' . \MailPoet\Util\Security::generateRandomString();
-    $segmentName = $i->createListWithSubscriber();
+
+    $this->startCreatingNewsletter($i, $newsletterTitle);
+
+    // step 4 - Choose list and send
```
rodrigoprimo: Sorry for the nitpick, but I guess we need to update this comment as the code below doesn´t choose a list.

rodrigoprimo (commented):
Looks good to me. Unfortunately, I don't know off the top of my head a better way to update the data-parsley-required parameter. I left just one comment with a nitpick.

mailpoet/tests/acceptance/Newsletters/SaveNewsletterAsDraftCest.php
samnajian: Updated the comment [in the last commit](https://github.com/mailpoet/mailpoet/pull/4328/commits/a5a75b4bb7ef71b7f5f809fb3f2ea39cd8aa9dff)

rodrigoprimo:
@samnajian, the acceptance test that was created in this PR is failing. Could you please take a look?

samnajian:
> @samnajian, the acceptance test that was created in this PR is failing. Could you please take a look?

@rodrigoprimo thanks for the ping, seems to be flaky and passed after a retry, still checking to see if I can get more stable

rodrigoprimo:
@samnajian, was a force push necessary? I understand that sometimes it is, but since we decided to avoid force pushes after the initial review of a PR, going forward, I would suggest documenting the reason for doing it here just so that the review is aware.

Right now there is no easy way for me to know what was changed after the last review. I'm assuming it was just something to make the acceptance test more stable, but I don't know for sure and I don't know what it was. Could you please add some information about it?

samnajian:
> @samnajian, was a force push necessary? I understand that sometimes it is, but since we decided to avoid force pushes after the initial review of a PR, going forward, I would suggest documenting the reason for doing it here just so that the review is aware.
> 
> Right now there is no easy way for me to know what was changed after the last review. I'm assuming it was just something to make the acceptance test more stable, but I don't know for sure and I don't know what it was. Could you please add some information about it?

@rodrigoprimo sorry about the force push, but I added no changes at all, just rebased trunk and pushed, and even without rebasing trunk a retry of the tests got passed, but I wanted to make sure a couple of runs don’t fail that’s why I triggered one by rebasing trunk into my branch

rodrigoprimo:
Thanks for the info, @samnajian. All good. Moving this PR to QA.

> I wanted to make sure a couple of runs don’t fail that’s why I triggered one by rebasing trunk into my branch

I guess for future PRs just triggering reruns from the CircleCI interface should be enough and there is no need to trigger them by pushing commits to the PR. Or is there any difference between the two approaches that I'm not aware of?

rodrigoprimo (approved):

veljkho (approved):

PR #4333, opened 2022-08-22
Check that changelog is not null before passing it to substr() [MAILPOET-4595]

rodrigoprimo:
Starting with PHP 8.1, substr() generates a warning if the first parameter is null. So, before calling it, we need to check if the changelog is null or not. See the Jira ticket for more details.

To test that the code still works even after the changes suggested here, you can add the following line to the foreach in \RoboFile::releaseCheckIssues() and run `./do release:check-issues`:

```
var_dump($issue['fields'][\MailPoetTasks\Release\JiraController::CHANGELOG_FIELD_ID]);
```

[MAILPOET-4595]

[MAILPOET-4595]: https://mailpoet.atlassian.net/browse/MAILPOET-4595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samnajian (approved):
LGTM!

veljkho (approved):

PR #4330, opened 2022-08-19
Remove old model usage from \MailPoet\Statistics\GATracking (S) [MAILPOET-4592]

rodrigoprimo:
[MAILPOET-4592]

[MAILPOET-4592]: https://mailpoet.atlassian.net/browse/MAILPOET-4592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4336, opened 2022-08-29
Remove old model usage from MailPoet\Cron\Workers\SendingQueue\Tasks\Newsletter - Part 1 [MAILPOET-4363]

rodrigoprimo:
## Description
Creating this PR as a draft while #4330 is not merged.

I'm opting to split [MAILPOET-4363] into multiple PRs to reduce the size of the code changes. More PRs to come once this one is merged.

## QA notes
Please perform exploratory testing around sending newsletters

## Linked tickets
[MAILPOET-4363]

[MAILPOET-4363]: https://mailpoet.atlassian.net/browse/MAILPOET-4363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4363]: https://mailpoet.atlassian.net/browse/MAILPOET-4363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -273,11 +273,17 @@ class SendingQueue {
     $statistics = [];
     $metas = [];
     foreach ($subscribers as $subscriber) {
+      $subscriberEntity = $this->subscribersRepository->findOneById($subscriber->id);
+
+      if (!$subscriberEntity instanceof SubscriberEntity) {
+        continue;
+      }
+
       // render shortcodes and replace subscriber data in tracked links
       $preparedNewsletters[] =
         $this->newsletterTask->prepareNewsletterForSending(
           $newsletter,
-          $subscriber,
+          $subscriberEntity,
           $queue
         );
       // format subscriber name/address according to mailer settings
@@ -288,12 +294,7 @@ class SendingQueue {
       // create personalized instant unsubsribe link
       $unsubscribeUrls[] = $this->links->getUnsubscribeUrl($queue, $subscriber->id);
 
-      $subscriberEntity = $this->subscribersRepository->findOneById($subscriber->id);
-      if ($subscriberEntity instanceof SubscriberEntity) {
-        $metas[] = $this->mailerMetaInfo->getNewsletterMetaInfo($newsletter, $subscriberEntity);
-      } else {
-        $metas[] = [];
-      }
+      $metas[] = $this->mailerMetaInfo->getNewsletterMetaInfo($newsletter, $subscriberEntity);
 
       // keep track of values for statistics purposes
       $statistics[] = [
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -177,7 +177,7 @@ class SendingQueue {
     // configure mailer
     $this->mailerTask->configureMailer($newsletter);
     // get newsletter segments
-    $newsletterSegmentsIds = $this->newsletterTask->getNewsletterSegments($newsletterEntity);
+    $newsletterSegmentsIds = $newsletterEntity->getSegmentIds();
     // Pause task in case some of related segments was deleted or trashed
     if ($newsletterSegmentsIds && !$this->checkDeletedSegments($newsletterSegmentsIds)) {
       $this->loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->info(
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -272,6 +272,9 @@ class SendingQueue {
     $unsubscribeUrls = [];
     $statistics = [];
     $metas = [];
+
+    $newsletterEntity = $this->newslettersRepository->findOneById($newsletter->id);
+
     foreach ($subscribers as $subscriber) {
       $subscriberEntity = $this->subscribersRepository->findOneById($subscriber->id);
 
@@ -282,7 +285,7 @@ class SendingQueue {
       // render shortcodes and replace subscriber data in tracked links
       $preparedNewsletters[] =
         $this->newsletterTask->prepareNewsletterForSending(
-          $newsletter,
+          $newsletterEntity,
           $subscriberEntity,
           $queue
         );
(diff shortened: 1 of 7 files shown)


mailpoet/tests/integration/Cron/Workers/SendingQueue/Tasks/NewsletterTest.php
```diff
@@ -243,38 +249,46 @@ public function testItSavesNewsletterPosts() {
...
+    $newsletter->setStatus('not_sent');
+    $this->newslettersRepository->persist($newsletter);
+    $this->newslettersRepository->flush();
     $this->newsletterTask->markNewsletterAsSent($newsletter, $queue);
-    $updatedNewsletter = Newsletter::findOne($newsletter->id);
-    assert($updatedNewsletter instanceof Newsletter);
-    expect($updatedNewsletter->status)->equals(Newsletter::STATUS_SENT);
-    expect($updatedNewsletter->sentAt)->equals($queue->processedAt);
+    $updatedNewsletter = $this->newslettersRepository->findOneById($newsletter->getId());
+    assert($updatedNewsletter instanceof NewsletterEntity);
```
lysyjan: I thought that we wanted to avoid using the `assert` function in our code base. It should be possible here by using `$this->assertInstanceOf`. There are more places with this. I won't comment on all of them.

mailpoet/lib/Cron/Workers/SendingQueue/Tasks/Newsletter.php
```diff
@@ -283,11 +289,20 @@ public function markNewsletterAsSent(NewsletterEntity $newsletter, $queue) {
     }
   }
 
-  public function getNewsletterSegments($newsletter) {
-    $segments = NewsletterSegmentModel::where('newsletter_id', $newsletter->id)
-      ->select('segment_id')
-      ->findArray();
-    return Helpers::flattenArray($segments);
+  public function getNewsletterSegments(NewsletterEntity $newsletter) {
+    $newsletterSegments = $this->newsletterSegmentRepository->findBy(['newsletter' => $newsletter]);
```
lysyjan: NewsletterEntity contains a method `getSegmentIds`. Do you think that it would be possible to use it here and simplify our code?

mailpoet/lib/Tasks/Sending.php
```diff
@@ -190,6 +193,13 @@ public function queue() {
     return $this->queue;
   }
 
+  public function getSendingQueueEntity(): SendingQueueEntity {
```
lysyjan, line 196: I guess this method may return null. But I'm not sure if we should solve it because it can be an edge case.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -272,12 +274,21 @@ public function processQueue($queue, $newsletter, $subscribers, $timer) {
...
+      if (!$subscriberEntity instanceof SubscriberEntity) {
+        continue;
+      }
+
       // render shortcodes and replace subscriber data in tracked links
       $preparedNewsletters[] =
         $this->newsletterTask->prepareNewsletterForSending(
-          $newsletter,
-          $subscriber,
+          $newsletterEntity,
```
lysyjan, line 297: I was wondering how this got through PHPStan because, if I am not wrong, the newsletter entity can be null here.
In the end, I found that the `newsletterTask` property is not annotated.

lysyjan (commented):
I have a few minor remarks, but I'm okay with moving this PR forward. It's up to you if you want to look at it.

mailpoet/tests/integration/Cron/Workers/SendingQueue/Tasks/NewsletterTest.php
rodrigoprimo: I failed to notice that some of the code that I moved around in this PR already contained calls to `assert()`. Thanks for pointing that out. I just added a new commit to remove all calls to `assert()` inside the integration tests that are modified in this PR. I left a call to `assert()` inside `SendingQueue.php` as it is not that trivial to replace and it will eventually get removed as we continue working on replacing Paris with Doctrine.

mailpoet/lib/Cron/Workers/SendingQueue/Tasks/Newsletter.php
rodrigoprimo: Good point. I missed that. I will create a new commit removing this method and using NewsletterEntity::getSegmentIds().

mailpoet/lib/Tasks/Sending.php
rodrigoprimo: I'm not sure as well, so if you are ok with that, I will leave the code as is. We will probably need to revisit this method when we get to the ticket to refactor the Sending class.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
rodrigoprimo: Good catch. I documented the properties and add an if to workaround the PHPStan error. The if is inside a foreach which is not really necessary but it was the easiest way that I found and it shouldn't be a big problem as far as I can see as we will revisit this soon as part of another Doctrine ticket.

rodrigoprimo:
@lysyjan, I addressed all of your comments. When you have a moment, could you please check this PR again?

(the CircleCI build failed but due to a flaky test that is likely unrelated to the changes suggested here, I'm rerunning it and will rerun again if it fails one more time)

lysyjan (approved):
Thank you for your answers and for addressing them!

Aschepikov (changes requested):
Hi. During the testing I found one issue:
I can resend the newsletter multiple times to the same subscribers. And I don't see a warning that this newsletter was sent. Newsletters from other versions cant be resent. Newsletters from this version can be resent in others. On the main page of emails, I can see that it sent 6 emails. 
image
But in the statistics, I can see that it was sent 12 times to 6 subscribers and I can edit and resend it for the third time
image

rodrigoprimo:
Hi, @Aschepikov. Thanks for testing this PR. I was not able to reproduce the problem that you mentioned. See the screencast below. Am I missing some steps to reproduce the issue? Thanks.

https://user-images.githubusercontent.com/77215/191339108-5c3a069a-2347-4d96-acb2-928d098c14f0.mp4

Aschepikov (approved):
Thank you, yes I can't reproduce it too, I have this broken newsletter and nothing else so it might be some edge case also present on the production, or even a server-side problem that has caused it.. and I will recheck it on the release build testing on Tuesday
 and I’ll be monitoring it in order to catch it once again and report it.

PR #4337, opened 2022-08-30
Add/remove workflow step [PREMIUM-194]

JanJakes:
Sorry for another huge PR. This got much bigger than I thought. I tried hard to make the commits nice so going commit-by-commit should make sense. Premium PR included as well.

**Testing:** With Premium activated and a valid Premium key, it should be possible to add/remove steps and triggers within a workflow and save the workflow.

[PREMIUM-194]
[MAILPOET-4585]

[PREMIUM-194]: https://mailpoet.atlassian.net/browse/PREMIUM-194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-4585]: https://mailpoet.atlassian.net/browse/MAILPOET-4585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/components/inserter-popover/index.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/inserter-popover/index.tsx
@@ -1,28 +1,62 @@
+import { useCallback, useRef, useState } from 'react';
 import { Popover } from '@wordpress/components';
-import { dispatch, useSelect } from '@wordpress/data';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
+import { Hooks } from 'wp-js-hooks';
+import { PremiumModal } from 'common/premium_modal';
 import { Inserter } from '../inserter';
+import { Item } from '../inserter/item';
 import { store } from '../../store';
 
 export function InserterPopover(): JSX.Element | null {
+  const popoverRef = useRef<HTMLDivElement>();
+  const [showModal, setShowModal] = useState(false);
   const { inserterPopover } = useSelect(
     (select) => ({
       inserterPopover: select(store).getInserterPopover(),
     }),
     [],
   );
+  const { setInserterPopover } = useDispatch(store);
 
-  const { setInserterPopover } = dispatch(store);
+  const onInsert = useCallback((item: Item) => {
+    const addStepCallback = Hooks.applyFilters(
+      'mailpoet.automation.workflow.add_step_callback',
+      () => {
+        setShowModal(true);
+      },
+    );
+    addStepCallback(item);
+  }, []);
 
   if (!inserterPopover) {
     return null;
   }
 
   return (
-    <Popover
-      anchorRect={inserterPopover.anchor.getBoundingClientRect()}
-      onClose={() => setInserterPopover(undefined)}
-    >
-      <Inserter />
-    </Popover>
+    <>
+      <Popover
+        ref={popoverRef}
+        anchorRect={inserterPopover.anchor.getBoundingClientRect()}
+        onClose={() => {
+          if (!showModal) {
+            setInserterPopover(undefined);
+          }
+        }}
+      >
+        <Inserter onInsert={onInsert} />
+      </Popover>
+
+      {showModal && (
+        <PremiumModal
+          onRequestClose={() => {
+            setShowModal(false);
+            popoverRef.current?.focus();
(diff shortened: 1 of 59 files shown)


mailpoet/lib/Automation/Engine/API/API.php
```diff
@@ -58,24 +58,29 @@ private function registerRoute(string $route, string $endpointClass, string $met
...
-        $endpoint = $this->endpointContainer->get($endpointClass);
-        return $endpoint->checkPermissions();
-      },
-      'args' => $schema,
-    ]);
+          return $endpoint->checkPermissions();
+        },
+        'args' => $schema,
+      ],
+      true // allow overriding
```
websupporter: It seems, allowing to overwrite in this way removes the free plugin `GET workflows` in favor of the the premium `POST workflows` route.

At least in my tests, setting the allow overwrite value to `false` recovers the `GET` endpoint

image

websupporter:
General question:
> it should be possible to add/remove [...] triggers

How can I do that? I miss somehow the obvious, but a completely empty workflow seems not to provide me a possibility and I do not see one in a workflow populated with steps
image
image

websupporter:
I wonder, now that there is a empty trigger necessary as a minimum for a workflow; in theory you can send `step:[]` to the API and a workflow is created without an empty trigger and basically useless. I think we should either throw an error from the API for such a case or fill an empty trigger as a default in case of `[]`

image

mailpoet/assets/js/src/automation/editor/components/inserter-popover/index.tsx
```diff
@@ -1,28 +1,66 @@
...
+import { useDispatch, useSelect } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
+import { Hooks } from 'wp-js-hooks';
+import { PremiumModal } from 'common/premium_modal';
 import { Inserter } from '../inserter';
-import { store } from '../../store';
+import { Item } from '../inserter/item';
+import { storeName } from '../../store';
 
 export function InserterPopover(): JSX.Element | null {
```
websupporter, line 11: `JSX.Element | null`

out of curiosity: Isn't `null` a valid `JSX.Element`?

mailpoet/lib/Automation/Engine/API/API.php
JanJakes: Yep, good catch! I need to overwrite the PUT route but not the others, damn. I guess I can use some filter to replace it instead. Checking...

JanJakes: Ahh, the fix is actually pretty simple! I'll drop this commit and make the premium route registrations run before the free ones!

mailpoet/assets/js/src/automation/editor/components/inserter-popover/index.tsx
JanJakes: It's [a bit compicated](https://stackoverflow.com/a/58123882/1236795), but no, `null` is not included in `JSX.Element`, it's only included in `React.ReactNode` but JSX element [seems to be the best consensus](https://github.com/typescript-cheatsheets/react/issues/129) for typing component return types.

JanJakes:
> I wonder, now that there is a empty trigger necessary as a minimum for a workflow; in theory you can send step:[] to the API and a workflow is created without an empty trigger and basically useless. I think we should either throw an error from the API for such a case or fill an empty trigger as a default in case of []

Good point! There are more things we don't validate yet, and I want to prepare a ticket for workflow validation, so I would like to include that. I'm not yet fully sure if we'll really need the empty triggers when we improve the workflow format to support multiple next steps (conditions, A/B test, etc.) but I want to look at it next week so I'll keep this in mind.

mailpoet/lib/Automation/Engine/API/API.php
websupporter: Thanks! Seems to do the trick 
image

websupporter (dismissed):
Thanks for the PRs. Really nice work. LGTM

veljkho (approved):

PR #4338, opened 2022-08-30
Automation listing UI updates 1/3 - MAILPOET-4532

johnolek:
This PR implements a number of UI tweaks to the automations listing page.

### DIFF
--- a/mailpoet/assets/css/src/components-automation-listing/_listing.scss
+++ b/mailpoet/assets/css/src/components-automation-listing/_listing.scss
@@ -1,3 +1,7 @@
 #mailpoet_automation * {
   box-sizing: border-box;
 }
+
+.mailpoet-automation-listing-heading {
+  margin-bottom: 16px;
+}
--- a/mailpoet/assets/css/src/components-automation-listing/_listing.scss
+++ b/mailpoet/assets/css/src/components-automation-listing/_listing.scss
@@ -1,5 +1,9 @@
 #mailpoet_automation * {
   box-sizing: border-box;
+
+  .mailpoet-add-new-button {
+    padding-right: 12px;
+  }
 }
 
 .mailpoet-filter-tab-panel {
--- a/mailpoet/assets/css/src/components-automation-listing/_listing.scss
+++ b/mailpoet/assets/css/src/components-automation-listing/_listing.scss
@@ -4,11 +4,18 @@
   .mailpoet-add-new-button {
     padding-right: 12px;
   }
+
+  .mailpoet-automation-listing {
+    /* Prevent border radius beneath tabs */
+    border-radius: 0 0 1px 1px;
+    /* Prevent overlapping box shadow since listing is rendered inside tabs */
+    clip-path: inset(-1px 0 0 0);
+  }
 }
 
 .mailpoet-filter-tab-panel {
   background-color: #fff;
-  border-radius: calc(2px - 1px);
+  border-radius: 1px;
   box-shadow: 0 0 0 1px rgba(0, 0, 0, .1);
   outline: none;
 
--- a/mailpoet/assets/css/src/components-automation-listing/_listing.scss
+++ b/mailpoet/assets/css/src/components-automation-listing/_listing.scss
@@ -2,6 +2,22 @@
   box-sizing: border-box;
 }
 
+.mailpoet-filter-tab-panel {
+  background-color: #fff;
+  border-radius: calc(2px - 1px);
+  box-shadow: 0 0 0 1px rgba(0, 0, 0, .1);
+  outline: none;
+
+  .count {
+    background-color: #f0f0f1;
+    border-radius: 12px;
+    font-size: 11px;
+    font-weight: 400;
+    margin-left: 6px;
+    padding: 2px 5px;
+  }
+}
+
 .mailpoet-automation-listing-heading {
   margin-bottom: 16px;
 }
--- a/mailpoet/assets/js/src/automation/automation.tsx
+++ b/mailpoet/assets/js/src/automation/automation.tsx
@@ -1,4 +1,5 @@
 import ReactDOM from 'react-dom';
+import { BrowserRouter } from 'react-router-dom';
 import { TopBarWithBeamer } from 'common/top_bar/top_bar';
 import { plusIcon } from 'common/button/icon/plus';
 import { Button, Flex } from '@wordpress/components';
@@ -101,24 +102,27 @@ function DeleteSchemaButton(): JSX.Element {
 
 function App(): JSX.Element {
   return (
-    <div>
-      <Workflows />
-      <div style={{ marginTop: 30, display: 'grid', gridGap: 8 }}>
-        <CreateEmptyWorkflowButton />
-        <CreateWorkflowFromTemplateButton slug="simple-welcome-email">
-          Create testing workflow from template (welcome email)
-        </CreateWorkflowFromTemplateButton>
-        <CreateWorkflowFromTemplateButton slug="welcome-email-sequence">
-          Create testing workflow from template (welcome sequence, only premium)
-        </CreateWorkflowFromTemplateButton>
-        <CreateWorkflowFromTemplateButton slug="advanced-welcome-email-sequence">
-          Create testing workflow from template (advanced welcome sequence, only
-          premium)
-        </CreateWorkflowFromTemplateButton>
-        <RecreateSchemaButton />
-        <DeleteSchemaButton />
+    <BrowserRouter>
+      <div>
+        <Workflows />
+        <div style={{ marginTop: 30, display: 'grid', gridGap: 8 }}>
+          <CreateEmptyWorkflowButton />
+          <CreateWorkflowFromTemplateButton slug="simple-welcome-email">
+            Create testing workflow from template (welcome email)
+          </CreateWorkflowFromTemplateButton>
+          <CreateWorkflowFromTemplateButton slug="welcome-email-sequence">
+            Create testing workflow from template (welcome sequence, only
+            premium)
+          </CreateWorkflowFromTemplateButton>
+          <CreateWorkflowFromTemplateButton slug="advanced-welcome-email-sequence">
+            Create testing workflow from template (advanced welcome sequence,
+            only premium)
+          </CreateWorkflowFromTemplateButton>
+          <RecreateSchemaButton />
+          <DeleteSchemaButton />
+        </div>
       </div>
-    </div>
+    </BrowserRouter>
   );
 }
 
--- a/mailpoet/assets/js/src/automation/automation.tsx
+++ b/mailpoet/assets/js/src/automation/automation.tsx
@@ -13,24 +13,26 @@ import {
 import { useMutation, useQuery } from './api';
 import { WorkflowListingNotices } from './listing/workflow-listing-notices';
 
-function Workflows(): JSX.Element {
+function Content(): JSX.Element {
   const { data, loading, error } = useQuery<{ data: Workflow[] }>('workflows');
-  let content;
 
   if (error) {
-    content = <div>Error: {error}</div>;
-  } else if (loading) {
-    content = <div>Loading workflows...</div>;
-  } else {
-    const workflows = data?.data ?? [];
-    content =
-      workflows.length < 1 ? (
-        <Onboarding />
-      ) : (
-        <AutomationListing workflows={workflows} loading={loading} />
-      );
+    return <div>Error: {error}</div>;
   }
 
+  if (loading) {
+    return <div>Loading workflows...</div>;
+  }
+
+  const workflows = data?.data ?? [];
+  return workflows.length > 0 ? (
+    <AutomationListing workflows={workflows} loading={loading} />
+  ) : (
+    <Onboarding />
+  );
+}
+
+function Workflows(): JSX.Element {
   return (
     <>
       <TopBarWithBeamer />
@@ -46,7 +48,7 @@ function Workflows(): JSX.Element {
           New automation
         </Button>
       </Flex>
-      {content}
+      <Content />
     </>
   );
 }
--- a/mailpoet/assets/js/src/automation/automation.tsx
+++ b/mailpoet/assets/js/src/automation/automation.tsx
@@ -44,6 +44,7 @@ function Workflows(): JSX.Element {
           }}
           icon={plusIcon}
           variant="primary"
+          className="mailpoet-add-new-button"
         >
           New automation
         </Button>
--- a/mailpoet/assets/js/src/automation/automation.tsx
+++ b/mailpoet/assets/js/src/automation/automation.tsx
@@ -1,4 +1,7 @@
 import ReactDOM from 'react-dom';
+import { TopBarWithBeamer } from 'common/top_bar/top_bar';
+import { plusIcon } from 'common/button/icon/plus';
+import { Button, Flex } from '@wordpress/components';
 import { Workflow } from './listing/workflow';
 import { AutomationListing } from './listing';
 import { Onboarding } from './onboarding';
@@ -11,20 +14,39 @@ import { WorkflowListingNotices } from './listing/workflow-listing-notices';
 
 function Workflows(): JSX.Element {
   const { data, loading, error } = useQuery<{ data: Workflow[] }>('workflows');
+  let content;
 
   if (error) {
-    return <div>Error: {error}</div>;
+    content = <div>Error: {error}</div>;
+  } else if (loading) {
+    content = <div>Loading workflows...</div>;
+  } else {
+    const workflows = data?.data ?? [];
+    content =
+      workflows.length < 1 ? (
+        <Onboarding />
+      ) : (
+        <AutomationListing workflows={workflows} loading={loading} />
+      );
   }
 
-  if (loading) {
-    return <div>Loading workflows...</div>;
-  }
-
-  const workflows = data?.data ?? [];
-  return workflows.length === 0 ? (
-    <Onboarding />
-  ) : (
-    <AutomationListing workflows={workflows} loading={loading} />
+  return (
+    <>
+      <TopBarWithBeamer />
+      <Flex className="mailpoet-automation-listing-heading">
+        <h1 className="wp-heading-inline">Automations</h1>
+        <Button
+          onClick={() => {
+            // TODO: link to templates page when implemented, MAILPOET-4533
+          }}
+          icon={plusIcon}
+          variant="primary"
+        >
+          New automation
+        </Button>
+      </Flex>
+      {content}
+    </>
   );
 }
 
--- a/mailpoet/assets/js/src/automation/listing/index.tsx
+++ b/mailpoet/assets/js/src/automation/listing/index.tsx
@@ -27,7 +27,7 @@ export function AutomationListing({ workflows, loading }: Props): JSX.Element {
           pageSearch.delete('paged');
         }
       });
-      history.replace({ search: pageSearch.toString() });
+      history.push({ search: pageSearch.toString() });
     },
     [pageSearch, history],
   );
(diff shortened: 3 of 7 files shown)


mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -79,24 +102,27 @@ function DeleteSchemaButton(): JSX.Element {
...
-        <CreateWorkflowFromTemplateButton slug="welcome-email-sequence">
-          Create testing workflow from template (welcome sequence, only premium)
-        </CreateWorkflowFromTemplateButton>
-        <CreateWorkflowFromTemplateButton slug="advanced-welcome-email-sequence">
-          Create testing workflow from template (advanced welcome sequence, only
-          premium)
-        </CreateWorkflowFromTemplateButton>
-        <RecreateSchemaButton />
-        <DeleteSchemaButton />
+    <BrowserRouter>
```
johnolek, line 108: I don't know if we want/need to wrap the entire App with the router

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -1,18 +1,90 @@
...
+import { Workflow, WorkflowStatus } from './workflow';
 
 type Props = {
   workflows: Workflow[];
   loading: boolean;
 };
 
+type FilterTab = {
+  name: string;
+  title?: JSX.Element;
```
johnolek: Only nullable because of how I build the tabs later, probably unnecessary.

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -21,56 +93,64 @@ export function AutomationListing({ workflows, loading }: Props): JSX.Element {
...
-                <MenuItem icon="trash">Move to trash</MenuItem>
-              </MenuGroup>
-            )}
+    <TabPanel
+      className="mailpoet-filter-tab-panel"
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore - the Tab type actually expects a string for titles but won't render HTML,
+      // making it very difficult to style the count badges. It seems to be compatible wtih JSX
+      // elements, however.
+      tabs={tabs}
```
johnolek: This might be one of the biggest issues. I'm not sure how we would get the styled badge counts if we have to use a normal string for the tab titles

mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -11,20 +14,39 @@ import { WorkflowListingNotices } from './listing/workflow-listing-notices';
...
+  } else if (loading) {
+    content = <div>Loading workflows...</div>;
+  } else {
+    const workflows = data?.data ?? [];
+    content =
+      workflows.length < 1 ? (
+        <Onboarding />
+      ) : (
+        <AutomationListing workflows={workflows} loading={loading} />
+      );
```
JanJakes: Nothing too important, but if you extract this to a component (can even be local to this file), you can use early returns instead of the non-const variable.

mailpoet/assets/js/src/automation/automation.tsx
JanJakes: My guess is the router only needs to sit above the highest place in the hierarchy where we're routing. That said, it's probably fine to have it somewhere very close to the root of the app or directly at the root, so I think this is good.

mailpoet/assets/css/src/components-automation-listing/_listing.scss
```diff
@@ -2,6 +2,22 @@
   box-sizing: border-box;
 }
 
+.mailpoet-filter-tab-panel {
+  background-color: #fff;
+  border-radius: calc(2px - 1px);
```
JanJakes: Isn't this `1px`? 😄 Or just a leftover from some experiments?

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -1,17 +1,90 @@
...
+
+  const updateUrlSearchString = (keysAndValues: Record<string, string>) => {
+    Object.keys(keysAndValues).forEach((key) => {
+      pageSearch.set(key, keysAndValues[key]);
+      if (['per_page', 'status'].includes(key)) {
+        pageSearch.set('paged', '1');
+      }
+    });
+    history.replace({ search: pageSearch.toString() });
+  };
```
JanJakes: Just to be sure, I'd wrap this in `useCallback()` so the function doesn't get redefined with each render. Not really important though, I guess.

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -1,17 +1,90 @@
...
+
+  const groupedWorkflows = {};
+  workflows.forEach((workflow) => {
+    if (!groupedWorkflows[workflow.status]) {
+      groupedWorkflows[workflow.status] = [];
+    }
+    groupedWorkflows[workflow.status].push(workflow);
+  });
```
JanJakes: Here, I'd probably use `useMemo` to avoid recomputing it on every render.

There's also a declarative way to write this, although not sure if more readable:
```ts
const groupedWorkflows = workflows.reduce(
  (grouped, workflow) => ({
    ...grouped,
    [workflow.status]: [...(grouped[workflow.status] ?? []), workflow],
  }),
  {},
);
```

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -1,17 +1,90 @@
...
+            {groupedWorkflows[tabInfo.status].length}
+          </span>
+        </>
+      );
+    } else {
+      tab.title = <span>{tabInfo.label}</span>;
+    }
+
+    tabs.push(tab);
+  });
```
JanJakes: Here, it may be nicer to extract the status to a component. I'd add `useMemo` again and maybe make it declarative again (with `tabLabels.map`).

Generally, there is a lot of logic directly in the component render body, but that's normal when prototyping. Later, it's better use add `useMemo` and `useCallback` or even better — extract some of it to a custom hook.

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -1,17 +1,90 @@
...
-  const rows = workflows.map((workflow) => getRow(workflow));
+  const history = useHistory();
+  const location = useLocation();
+  const pageSearch = new URLSearchParams(location.search);
+
+  const updateUrlSearchString = (keysAndValues: Record<string, string>) => {
+    Object.keys(keysAndValues).forEach((key) => {
+      pageSearch.set(key, keysAndValues[key]);
+      if (['per_page', 'status'].includes(key)) {
+        pageSearch.set('paged', '1');
```
JanJakes: One thing I noticed is WP doesn't seem to show `paged=1` for posts. They only start from `2` but for `1` they normalize it out.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: This [looks safe](https://github.com/WordPress/gutenberg/blob/39172d1aa22adfa0b337de6bb53801276f868065/packages/components/src/tab-panel/index.js#L87), at least for now. Maybe [the DefinitelyTyped types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fb66211010980641f01bcfe8ffbc0cc3d71989ca/types/wordpress__components/tab-panel/index.d.ts#L12) are even wrong.

There are also some [other fields](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fb66211010980641f01bcfe8ffbc0cc3d71989ca/types/wordpress__components/tab-panel/index.d.ts#L21) but I guess those help only in the tab body and not in the title?

JanJakes (commented):
Looks good! I think it's almost ready.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: This might be solvable also via CSS, using some pseudoelement like `::after` and passing in value from some data attribute, maybe. But it's weird and probably not very accessible.

mailpoet/assets/css/src/components-automation-listing/_listing.scss
johnolek: Haha, sorry about this! I was copying CSS rules from my browser to mimic the visual effects of the table. 

At first I thought there might just be a simple CSS class I could re-use, but it looks like the class name for these styles is computed, and I'm not sure where it's coming from or how I could re-use this manually. I'm sure there's a better way to handle this.

Here are the styles I'm talking about from my inspector:

image

mailpoet/assets/js/src/automation/automation.tsx
johnolek: Great idea, thanks for the suggestion!

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: Great catch! I'll delete the parameter instead of setting it to 1 in this case.

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: Good call on `useMemo`!

This is mostly a personal failing on my part, but I've never liked using `reduce`. It was hard for me to understand when I first started coding and I've never warmed up to it.

In this case I'm curious, would the performance be the same using the spread syntax compared to `push`? I did a little bit of testing and I think spread option is slower, but it probably wouldn't make any kind of noticeable difference without there being tons and tons of workflows to filter.

I tested with 15k workflows and your code took ~700ms to sort compared to ~1ms for using `.push`. The much, much slower part was actually the query to retrieve all the workflows in the first place because there's no `workflow_id` index on the `workflow_versions` table, so I'm going to add one.

I'm happy to change to something more declarative but I'd like to avoid the performance hit if possible.

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: I refactored this using `useMemo` and a `map`, but I'm not sure I completely follow what it would entail to convert it to its own component or use a custom hook. Could you please elaborate a little bit about what that might look like so I can further improve this?

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: I would prefer to avoid a CSS solution like that if possible, but I also don't want this to cause any issues, or really require the `@ts-ignore` at all.

> There are also some other fields but I guess those help only in the tab body and not in the title?

This matches my understanding. You can attach arbitrary key/value pairs but they're only intended to be used in the callback that renders the tab body.

I've refactored to an approach that doesn't require using JSX for the `title` (even though it's probably totally fine), but I doubt that it's doing things the "react" way. I would appreciate your guidance on how to make it better if you have any ideas!

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: I made this change, and also added `useMemo` for the `pageSearch` variable, but I'm not sure how much of a difference it will make. Won't it get redefined every time anyway since it will have `pageSearch` in its dependency array? Maybe it's useful to have for later, when we add other features and wouldn't want to redefine this callback if they make changes that don't affect the query string for the page?

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -1,76 +1,180 @@
...
+
+  const updateUrlSearchString = useCallback(
+    (keysAndValues: Record<string, string>) => {
+      Object.keys(keysAndValues).forEach((key) => {
+        pageSearch.set(key, keysAndValues[key]);
+        if (['per_page', 'status'].includes(key)) {
+          pageSearch.delete('paged');
+        }
+      });
+      history.push({ search: pageSearch.toString() });
```
johnolek: This is the one thing I'm still really struggling with. I'm not sure how to update the current tab in the TabPanel's state when a user hits back/forward. Do you have any ideas for how to handle this so users can actually navigate forwards/backwards through the tabs?

johnolek:
Thank you so much for the detailed review, @JanJakes! I think I've fixed some things and I've probably introduced other problems. Please let me know if you have any other ideas, especially for how to get history working.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: I checked the [TabPanel component](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/tab-panel), and weirdly enough, they provide only `initialTabName` but no `activeTabName` to make its state controlled. Maybe we should create an issue for that on their repo.

In the meantime, you can use a hack — nothing nice but very simple — add `key={pageSearch.get('status')}` to the `TabPanel` component. This will force-unmount and mount the component and together with the `initialTabName={pageSearch.get('status') || 'all'}` it will make it work.

JanJakes: On a different note, I'm not sure why do you check also for the status part in this case (`['per_page', 'status'].includes(key)`), those params seem unrelated, aren't they?

The loop seems to modify the original `pageSearch` object (which in React might not be the best to do) and in general seems a bit hard to read. I think what you mean is "take existing params + new params + delete paged when `1`" which can be written as:

```ts
const updateUrlSearchString = useCallback(
  (search: Record<string, string>) => {
    const params = new URLSearchParams({
      ...Object.fromEntries(pageSearch.entries()),
      ...search,
    });
    if (params.get('paged') === '1') {
      params.delete('paged');
    }
    history.push({ search: params.toString() });
  },
  [pageSearch, history],
);
```

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: > but I doubt that it's doing things the "react" way

Ahh, interesting idea, but yeah, this is really dangerous in React! You're creating elements under the React tree that are not known to React (not rendered within React's VDOM). Then any diffing will simply not count with those — they may be accidentally lost or cause some different problems due to undefined behavior. In this case, the previous `@ts-ignore` solution actually looks better and safer to me.

Again, we may try to create an issue [in the upstream repo to together with a PR changing the type](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/tab-panel/types.ts) (a funny thing I just noticed is they converted it to TS a few days ago).

But we should probably note these things in our checklist doc and do them post-MVP.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: The weird thing about React is that it kind of changes the semantics of the language, or at least how people understand it. A function suddenly must be expected to be called (rendered) many times, even for a single "render" from the user's perspective. That's not only non-performant but also very unintuitive, and it often causes developers to wrap a lot of stuff in `useMemo` and `useCallback` calls.

One may think that we should use these hooks only for some computationally expensive logic, but that's not the case either — when you want a stable reference to some object (for instance, to avoid re-rendering children components via changing props), then you need memoization as well.

This is why frameworks like [Solid.js](https://www.solidjs.com/) or [Svelte](https://svelte.dev/) are so much more performant, and, in many ways, better.

While in many cases, you may not need the hooks, I tend to use them rather more than less as a "good practice". As for the `pageSearch` — sure, in itself it would be a very quick computation, and if the identity of the object doesn't cascade many further renders, then it's probably OK without `useMemo`. But since the result is not a scalar value and it's identity would change with every render, it may be nice to keep the memo to prevent any possible issues.

Note that for a single "view" for the user, the component function is sometimes called multiple times — each `setState`, `useEffect`, or prop changes coming from parents can cause another re-render that the user won't notice (unless you start randomly losing focus in an editor, etc.). So if you type `console.log('render');` to some of our component bodies, you may see it logged multiple times. This stands true until advanced batching and some form of auto-memoization are in React itself.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: Good catch 😎  Reduce with spread inside surely is slower in cases like these because it has a complexity of `O(n^2)` (for each `n`, you need to iterate and reconstruct the already-existing array), compared to the possible `O(n)`, not to mention the extra temporary allocations. For a reasonably small amount of items, though, this is probably not measurable (as we'd never return thousands of workflows from the API). I would also guess the language optimizer could resolve these cases to something simpler, but as you've measured, it doesn't seem so.

In React state management, this approach is extremely common — most things are written declaratively to make sure (and visible) the proper objects and sub-objects have new instances and never modify the original ones. Often this is also solved using [Immer](https://github.com/immerjs/immer) and other libs.

Readability is probably partly a habit and personal preference. I guess that `reduce` in particular is a bit difficult, but otherwise, I got used to the declarative style quite a bit and find it readable.

In this case, use which option you prefer. Ultimately, both work if you don't modify the original object, and yes, yours may be more readable here.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: For instance, the workflow grouping can be extracted simply as follows:

```ts
const useGroupedWorkflows = (
  workflows: Workflow[],
): Record<string, Workflow[]> => {
  return useMemo(() => {
    const grouped = {};
    workflows.forEach((workflow) => {
      if (!grouped[workflow.status]) {
        grouped[workflow.status] = [];
      }
      grouped[workflow.status].push(workflow);
    });
    return grouped;
  }, [workflows]);
};
```

This is a custom hook, it's as simple as that. Thanks to `useMemo`, it returns a stable identity for the same `workflows` parameter.

---

As for tabs and table headers — they don't have any dependencies from the state or component body, so they could be simple module-level arrays (indeed ever allocated only once):

```ts
const tabs = [
  {
    name: 'all',
    title: 'All',
    className: 'mailpoet-tab-all',
  },
  {
    name: WorkflowStatus.ACTIVE,
    title: 'Active',
    className: 'mailpoet-tab-active',
  },
  {
    name: WorkflowStatus.INACTIVE,
    title: 'Inactive',
    className: 'mailpoet-tab-inactive',
  },
  {
    name: WorkflowStatus.DRAFT,
    title: 'Draft',
    className: 'mailpoet-tab-draft',
  },
  {
    name: WorkflowStatus.TRASH,
    title: 'Trash',
    className: 'mailpoet-tab-trash',
  },
] as const;
```

I even simplified it by inlining the `className`, seems more redable to me now. (Another nice thing is we can add the `as const` and TS will now understand that `label: 'Trash'` belongs to `className: 'mailpoet-trash-tab'` instad of inferring `label: string` and `className` string. But we don't need this in the code below so not important.)

Similarly for the table headers...

mailpoet/assets/css/src/components-automation-listing/_listing.scss
```diff
@@ -1,3 +1,34 @@
...
+
+  .mailpoet-add-new-button {
+    padding-right: 12px;
+  }
+
+  .mailpoet-automation-listing {
+    /* Prevent border radius beneath tabs */
+    border-radius: 0 0 1px 1px;
+    /* Prevent overlapping box shadow since listing is rendered inside tabs */
+    clip-path: inset(-1px 0 0 0);
```
JanJakes, line 12: I wonder whether it's a good idea to clip out the `box-shadow` rather than resetting the box shadow itself. Or maybe I'm not understanding what exactly this does. Can the listing have all borders and box shadows reset and can we apply them only on the outer tabs, maybe?

Anyway, the result looks good 👍 So maybe no need to fix anything here.

johnolek: The intention here is to clip out all of the box shadow except for 1px of shadow on the top.

It would be possible to use a border instead of a box shadow, but then I'm not sure if there's an elegant way to get the bottom border of the active tab to cover the border.

Using border:
image

Using box shadow with clipping:
image

I think it actually makes a pretty big difference zoomed out (this is just me switching between the two versions a few times):

https://user-images.githubusercontent.com/8656640/188711373-b6d5e405-60be-492d-b963-d428847ebe60.mov

I think I'll just leave this as-is for now and hopefully we can come up with a cleaner solution in the future when we refactor this into a proper component.

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: Thank you for the `key` suggestion! Not ideal, but better than having something non-functional.

> On a different note, I'm not sure why do you check also for the status part in this case (['per_page', 'status'].includes(key)), those params seem unrelated, aren't they?

The idea is to reset to the first page when a user clicks on a different tab or changes the number of results per page. If you're viewing page 2 of "All" and then click on "Inactive" you wouldn't necessarily want to immediately view page 2 of the "Inactive" tab. And if there aren't enough "Inactive" workflows for there to be a second page, the pagination controls won't even appear and you can end up seeing something like this:

image

It's similar but different for `per_page`. Let's say you have 240 workflows and you're viewing page 5/10 with 25 workflows per page. If you switch to 100 per page there are now only 3 total pages, so leaving `paged=5` would result in an empty table. 

We could try to dynamically calculate a logical new page number. For example we could figure out what the first workflow was on the page you were viewing previously and then switch to whichever page had that workflow with the new `per_page` value. If you were previously looking at workflows 101-125 then we could figure out we should show you 101-200 with the new `per_page`. That just seemed overly complicated and may not even be what users expect.

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: Very good points! I've converted back to the JSX for now.

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: Thank you very much for this detailed explanation!

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: > This is a custom hook, it's as simple as that

Is the benefit here that we would be able to extract the callback outside of this component entirely? If so where would be the best place for it to live? Or is there a different advantage that I'm not seeing?

johnolek: Thank you again for your detailed explanation! I wasn't familiar with `as const` before.

mailpoet/assets/css/src/components-automation-listing/_listing.scss
JanJakes: Ahh, OK! I didn't see this is more complex than just borders.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: > Is the benefit here that we would be able to extract the callback outside of this component entirely? If so where would be the best place for it to live? Or is there a different advantage that I'm not seeing?

I think in this case it's only about code organization, just giving that part of code a function name (hook name) and making it slightly more readable. I'd probably just put it in the same file. For custom hooks used in multiple places, we'd need to extract them to some specific files, I guess, but this one is used only locally.

React docs encourage the usage of custom hooks, it can partly separate business logic from rendering, but if it's not meant to be reused in multiple places, then there are not many more advantages than that. When naming conventions are met (the `use` prefix), linters also understand that it is a hook and enforce the [hook rules](https://reactjs.org/docs/hooks-rules.html).

So ultimately, it's just a tiny detail, nothing we'd have to do now.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: Ahh, thanks for the explanation, I totally missed that!

JanJakes (dismissed):

JanJakes (approved):

JanJakes:
LGTM! I did a rebase to resolve a conflict. Merging.

mailpoet/assets/js/src/automation/listing/index.tsx
johnolek: > So ultimately, it's just a tiny detail, nothing we'd have to do now

Still great to know, thanks for the explanation!

PR #4343, opened 2022-09-02
Enable Woo in integration test and used Woo code for creating orders [MAILPOET-4566]

costasovo:
This PR is a first step for enabling better testing of changes we need to do for supporting WooCommerce custom order tables (COT).

It removes custom-made mocks of WooCommerce classes and activates the WooCommere and other related plugins in the test environment.

It also removes custom SQL queries for creating Woo tables for tests. These are no longer needed because all Woo tables are created when we activate the plugin.

This PR also splits the integration test on CI into two jobs:

1. **integration_tests_base** - Run all tests not dependent on WooCommerce in an environment where WooCommerce and other plugins are not activated. This should (at least partially) protect against making some core parts of the plugin accidentally dependent on Woo.
2. **integration_tests_woo** - Run tests dependent on WooCommerce in an environment where the WooCommerce plugin is active and allows using WooCommerce code in the integration tests. These tests are distinguished using group `woo`.

NOTE: integration_tests_multisite and all integration tests in nightly build run within a single jog with active WooCommerce.
We don't run all possible combinations of configurations for every build. I believe that this setup covers possible cases sufficiently and, at the same time, doesn't increase the duration of builds.

The most important part of this PR is refactoring of tests that work with Woo orders so that they use `wc_create_order` and `wc_get_orders` to create/access orders instead of custom SQL or `wp_insert_post`. This should allow us to run those test with different configurations of the WooCommerce COT feature.

Note: The methods for creating, getting, deleting WooCommerce orders were added to [IntegrationTester](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/_support/IntegrationTester.php) helper class. After some exploration, I think that the more systemic solution would be refactoring the [WooCommerceOrder](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/DataFactories/WooCommerceOrder.php) data factory, which currently uses WP CLI. We could use the Woo code instead of the CLI, which would make it also faster. I found that the refactor would be a bit more complex task because we should also refactor the WooCommerceProduct data factory.
I'm happy to do the refactor, but I would prefer to do it within a separate ticket because this PR is already quite big, and changes should be sufficient to unblock the other tickets.

[MAILPOET-4566]

[MAILPOET-4566]: https://mailpoet.atlassian.net/browse/MAILPOET-4566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
```diff
@@ -269,6 +272,9 @@ public function testItSchedulesNewEmailWhenEmailAlreadySent() {
     $this->assertEquals($scheduled->getScheduledAt(), $expectedTime);
   }
 
+  /**
+   * @group woo
```
rodrigoprimo: To me, it is not immediately obvious why the other tests in this class are not part of the `woo` group. Could you please elaborate? I just skimmed through them so I could be missing something.

mailpoet/tests/docker/codeception/docker-entrypoint.sh
```diff
@@ -62,8 +62,6 @@ if [[ -z "${SKIP_DEPS}" ]]; then
   cd - >/dev/null
 fi
 
-# extra plugins are not supported for integration tests yest
-if [[ $TEST_TYPE == "acceptance" ]]; then
```
rodrigoprimo, line 66: I guess we don't need to install WooCommerce when skipping the woo group? I'm not sure if this will make a huge difference but it might make the tests setup process a little bit faster and it should be simple to implement? If it is not, I don't think it is worth it.

Maybe we can download the zip files for the plugins and activate them only if SKIP_PLUGINS is not true or we are not skipping the woo group?

mailpoet/tests/integration/Automation/Integrations/MailPoet/Actions/SendEmailActionTest.php
```diff
@@ -52,6 +56,8 @@ class SendEmailActionTest extends \MailPoetTest {
 
   public function _before() {
     parent::_before();
+    $this->cleanup();
```
rodrigoprimo, line 59: I believe the convention is to clean up the state after the test is executed and not before it (at least for PHPUnit, I'm not sure if things are different in Codeception). Would it make sense to do that here and delete the entries from the tables inside `_after()` instead of inside `_before()`?

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceCountryTest.php
```diff
@@ -136,12 +137,15 @@ private function createCustomerLookupData(array $data): void {
     $id = $connection->lastInsertId();
     $orderId = (int)$id + 1;
     $orderLookupTable = $wpdb->prefix . 'wc_order_stats';
+    $dateCreated = Carbon::now()->toDateTimeString();
     $connection->executeQuery("
-      INSERT INTO {$orderLookupTable} (order_id, status, customer_id)
+      INSERT INTO {$orderLookupTable} (order_id, status, customer_id, date_created, date_created_gmt)
```
rodrigoprimo: I wonder if we should/could use WooCommerce functions to insert data into the wp_wc_orders_stats table instead of using an insert query? I don´t think WC provides functions to manipulate wp_wc_orders_stats directly, but maybe we can achieve what we need by creating an order?

mailpoet/tests/integration/Segments/WooCommerceTest.php
```diff
@@ -660,7 +660,7 @@ private function cleanData(): void {
       DELETE FROM
         {$subscribersTable}
       WHERE
-        email LIKE 'user-sync-test%'
+        email LIKE 'user-sync-test%' OR email = ''
```
rodrigoprimo, line 664: Do we need to use WooCommerce functions instead of the direct queries below that delete orders and order meta?

https://github.com/mailpoet/mailpoet/pull/4343/commits/f8062f1ccd554d1494793568127bd0f2ddd47e15#diff-cd3b372ece0ad4e6702c5742311661aaff68b14de05e9fb8eb3fa5a5c4d50368L665-L683

I wonder if we need to do the same for \MailPoet\Test\DataFactories\WooCommerceOrder::create()?

https://github.com/mailpoet/mailpoet/blob/9f790efbf0f032fc32e264dc33450286b362976e/mailpoet/tests/DataFactories/WooCommerceOrder.php#L102

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceCategoryTest.php
```diff
@@ -34,7 +36,6 @@ public function _before(): void {
     $this->wooCommerceCategory = $this->diContainer->get(WooCommerceCategory::class);
     $this->subscribersRepository = $this->diContainer->get(SubscribersRepository::class);
 
-    Database::loadSQL('createWCLookupTables');
```
rodrigoprimo, line 37: This was the last use of the method Database::loadSQL(). Do you reckon we should remove it? I'm not sure as it could be useful in the future, but I'm inclined to say we should as we are not currently using it.

rodrigoprimo (commented):
Great job, @costasovo! I left some comments for us to explore a few points further.

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
costasovo: Aha. Nice catch! This is leftover. I'll remove it. I'll try to explain how it happened.
When I started adding the groups, I thought that I would add it on the test case level and pick only tests that really call some function from the WooCommerce codebase.
But then I realized that this is not how the semantics of the group works or, in other words, that the better semantics for the group would be that the tested functionality is related to Woo integration.

So it is still ok to mark just one test case or service that is just partially related to Woo integration. But in this case, the whole class was made because of Woo integration, so I marked the whole test class, but I forgot to remove the group from the test case I marked before.

mailpoet/tests/docker/codeception/docker-entrypoint.sh
costasovo: Good point! I'll add this improvement.

mailpoet/tests/integration/Automation/Integrations/MailPoet/Actions/SendEmailActionTest.php
costasovo: > I believe the convention is to clean up the state after the test is executed and not before it (at least for PHPUnit, I'm not sure if things are different in Codeception). Would it make sense to do that here and delete the entries from the tables inside _after() instead of inside _before()?

This is an interesting point. We have had a couple of conversations on this topic in the past (shame I couldn't find any on p2 or slack) and some of us (I remember @pavel-mailpoet being one of them) strongly preferred doing the cleanup in the before phase. The arguments for doing it before were (not sure I remember all of them):

1) It should be the responsibility of the test to prepare the environment for its run. This applies to both preparing testing data and cleaning data in DB tables it is about to use.
2) Doing the cleanup after the test implicates interdependence between tests.

I don't entirely agree with the second one because doing the cleanup after the run feels more natural to me (I clean my mess). But, on the other hand, I find cleaning before more error-prone as some tests may forget to clean up properly.

I don't have a strong opinion, and I usually add the cleanup to a private method and call it in both before and after. In this case, I forgot to add it to the after method.

So I would fix this and add it also to the after method. Would that be ok for you?

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceCountryTest.php
costasovo: Good point! I'll look into this. Maybe we could create an order + trigger sync to lookup tables.

mailpoet/tests/integration/Automation/Integrations/MailPoet/Actions/SendEmailActionTest.php
pavel-mailpoet: Since I was tagged I can share my opinion, but you can do whatever you think is better since I don't code much anymore.
1. If tests rely on emptying the database and we clean in the `_after` method we get flakey tests. Sometimes people forget to clean up properly and tests start failing when running the whole suite but pass if you run a single test. It is a pain to debug this kind of failures. And if that depends on the execution order, it is even more challenging to debug. So the tests are more stable if they don't expect anything and prepare the environment for them. 
2. If clean in both `_after` and `_before` we waste time. Running the cleaning queries takes time and if it is done in all tests it adds up and makes the tests run longer. 

So that is why I prefer if tests run the clean up before the execution.

mailpoet/tests/integration/Segments/WooCommerceTest.php
costasovo: Good catch! We already use the function for deleting tables from Woo in this test, so it is possible that doing this is just ambiguous.

> I wonder if we need to do the same for \MailPoet\Test\DataFactories\WooCommerceOrder::create()?

I think we need to do it.

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceCategoryTest.php
costasovo: I was thinking that it could be useful in the future. But you're right we don't use it so let's remove it.

mailpoet/tests/integration/Automation/Integrations/MailPoet/Actions/SendEmailActionTest.php
costasovo: > If clean in both _after and _before we waste time. Running the cleaning queries takes time, and if it is done in all tests it adds up and makes the tests run longer.

Yes. But as we are inconsistent in the approach adding to both `before` and `after` is the safest variant. The cleanup in the before section protects the test against ones that forget to cleanup (or were written only with cleanup in before), and adding cleanup in the after method protects tests that don't do cleanup in the before section.

So I guess this is something that needs to be discussed on the team level (I mean MailPoet team), and we need to unify the approach in existing tests. Until we use the same approach in all tests I would keep adding it to both.

costasovo:
@rodrigoprimo Thank you for the detailed review! I added 7 new commits. I believe they are addressing all the issues.

mailpoet/tests/integration/Automation/Integrations/MailPoet/Actions/SendEmailActionTest.php
rodrigoprimo: I'm ok with adding the cleanup() to both _before() and _after(), but I don't think it is needed in this case as, if I'm remembering correctly a check that I did last week, the test suite passes if we only call cleanup() inside _after().

> It should be the responsibility of the test to prepare the environment for its run. This applies to both preparing testing data and cleaning data in DB tables it is about to use.

I have a different view of the conventions regarding tests, but I'm talking off the top of my head so I might be misremembering things. I always saw things the opposite way as what was described above. Individual tests should make sure that they finish leaving the environment in the same state that they found it when they started.

> doing the cleanup after the run feels more natural to me (I clean my mess)

Same here

> Doing the cleanup after the test implicates interdependence between tests.

I also don't agree with this point.

> If tests rely on emptying the database and we clean in the _after method we get flakey tests.

I wouldn't call them flaky as they consistently fail or pass depending on whether they are being triggered individually or with the whole suite.

> Sometimes people forget to clean up properly and tests start failing when running the whole suite but pass if you run a single test. It is a pain to debug this kind of failures. And if that depends on the execution order, it is even more challenging to debug.

That is a fair point. In my opinion, we should address this by fixing the tests that are not cleaning up properly. I have fixed a few already by creating a MailPoetTest::_after() method that prints the number of entries in the affect tables and then running the whole test suite. This way it is not that hard to find the tests that are leaving entries in the database. But maybe there are other instances where it is not that easy to find the problematic tests for some reason?

As a side note, WP runs all the queries for a particular test inside a single transaction that is never committed. This way data is never written to the database and the tests run much faster. I don't know if we could do something similar for our tests.

mailpoet/tests/docker/codeception/docker-entrypoint.sh
rodrigoprimo: @costasovo, I'm sorry for the nitpick, but the indentation is incorrect for the if block below that installs WooCommerce.

rodrigoprimo:
Thanks for the changes, @costasovo. I resolved most of the conversations and left comments in two of them. I'm happy with whatever you choose regarding when to call cleanup(). I agree with you that, at some point, we should have a conversation with the whole team about this to settle on an approach.

mailpoet/tests/integration/Automation/Integrations/MailPoet/Actions/SendEmailActionTest.php
costasovo: I added the cleanup to the after method. So now we do it both before and after test.

> But maybe there are other instances where it is not that easy to find the problematic tests for some reason?

There are some tests that create WordPress users or add some hooks. Forgotten WP users may easily cause failing tests as a lot of them to work with subscribers, and WP users are automatically added to the WordPress users list.

costasovo:
@rodrigoprimo, I fixed the indentation issue. Thank you for the heads up!

rodrigoprimo (approved):

veljkho (approved):

PR #4344, opened 2022-09-02
Remove leftover MP2 migration file that is not needed anymore [MAILPOET-4612]

rodrigoprimo:
This file was loaded inside the MP2MigratorTest class that was removed (https://github.com/mailpoet/mailpoet/pull/4235/files#diff-fa71dea302390fbf7c04d9d882a5a66a70d3daa432883d4ba6951e2ec5abe287L123) and thus it is not needed anymore.

[MAILPOET-4612]

[MAILPOET-4612]: https://mailpoet.atlassian.net/browse/MAILPOET-4612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo (approved):
Good find 👍 ! I was looking for possible leftovers during the previous code review, but I missed that one.

veljkho (approved):

PR #4345, opened 2022-09-05
Improve error messages on the last step of sending [MAILPOET-4506]

lysyjan:
## Description
I decided to use a slightly different approach for validation segments compared to the ticket specification.
The error message is displayed under the select instead of a tooltip on the sending button.

## QA notes
1. Go to `MailPoet` > `Emails`
2. Click to create New simple Newsletter and go to the last step of the flow (4. Send step)
3. Select a list with zero subscribers
4. Try sending and observe the error message, compare it with the one from the Jira specs ticket
5. Check the styles from the specs and verify
6. Under the `Sender` section, change the existing email to some unauthorized email like example@moonbridge.rs and observe the results
7. Go to `MailPoet` > `Settings` > `Basics` tab and change the `Default Sender`'s `From` email to unauthorized and observe the results

## Linked tickets
[MAILPOET-4506]

[MAILPOET-4506]: https://mailpoet.atlassian.net/browse/MAILPOET-4506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/parsley-validators.jsx
+++ b/mailpoet/assets/js/src/parsley-validators.jsx
@@ -34,4 +34,27 @@ jQuery(($) => {
       en: 'An email can only be scheduled up to 5 years in the future. Please choose a shorter period.',
     },
   });
+
+  Parsley.addValidator('segmentsWithSubscribers', {
+    requirementType: 'string',
+    validateMultiple: (values, noSegmentWithSubscribersError) => {
+      const segments = window.mailpoet_segments || [];
+
+      let isValid = true;
+      values.forEach((segmentId) => {
+        const segment = segments.find((s) => s.id === segmentId);
+        if (segment && segment.subscribers === 0) {
+          isValid = false;
+        }
+      });
+
+      if (!isValid) {
+        return $.Deferred().reject(noSegmentWithSubscribersError);
+      }
+      return true;
+    },
+    messages: {
+      en: 'Please select a list with subscribers',
+    },
+  });
 });
--- a/mailpoet/assets/js/src/parsley-validators.jsx
+++ b/mailpoet/assets/js/src/parsley-validators.jsx
@@ -54,7 +54,7 @@ jQuery(($) => {
       return true;
     },
     messages: {
-      en: 'Please select a list with subscribers',
+      en: 'Please select a list with subscribers.',
     },
   });
 });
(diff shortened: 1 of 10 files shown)


mailpoet/assets/js/src/parsley-validators.jsx
```diff
@@ -34,4 +34,27 @@ jQuery(($) => {
...
+    validateMultiple: (values, noSegmentWithSubscribersError) => {
+      const segments = window.mailpoet_segments || [];
+
+      let isValid = true;
+      values.forEach((segmentId) => {
+        const segment = segments.find((s) => s.id === segmentId);
+        if (segment && segment.subscribers === 0) {
+          isValid = false;
+        }
+      });
```
johnolek: This will make the form appear invalid if you've selected any segments with 0 subscribers, including if you've selected other segments that do have subscribers, which isn't the intended behavior (I double checked with @NeosinneR). It should be valid/possible to send as long as there's 1 list selected that has subscribers.

johnolek (changes requested):

NeosinneR:
image

@lysyjan the change (from tooltip to error message) is okay, but seeing it in action, I see a couple of other UX issues.

1. The two error messages (list and email) look slightly different. Could you please unify them, with the authorized error being the correct one?
2. The green success state is really confusing. It is only shown on some of the inputs, and it's even shown on the input with an error message (I understand why, but the UX is subpar). Could you please completely remove parsley's success styling?
3. The authorized email error was designed to be on one line. There is definitely room, but it seems the holder for the error is not as wide as the input. Could you please make it 100% wide so the message fits on one line?

Let me know your thoughts, and I can also update the specs later.

lysyjan:
@NeosinneR Thank you for your notes. It makes sense to me, so I will improve the behavior. Please change the ticket description for easier validation by Veljko.

johnolek (dismissed):
LGTM!

Aschepikov (changes requested):
Hi. During the testing I found some issues:
**1. In the ticket margin-top: 6px; and in the pr and on the site it is 2**

image
image


**2. The error message "Please select a list with subscribers" should contain a dot at the end.** 

image
image

lysyjan:
@Aschepikov Thank you for your feedback!
1. This is intentional because it didn't look fine in some cases when I used `6px`. See an example below: 
image
What do you think about that @NeosinneR?
2. I fixed it.

NeosinneR:
@lysyjan right, I wasn't counting on errors on inputs with the small space, only with the larger space below them. If it looks good with 2px, that's fine by me.

Aschepikov (approved):
This is verified, good job.

PR #4347, opened 2022-09-06
Add possibility of tagging imported subscribers [MAILPOET-4558]

lysyjan:
## Description
This pull request extends the import form about a new input field. This input is optional and of [a type token field from Gutenberg](https://wp-gb.com/formtokenfield/). Selected tags are added to all imported subscribers.
 
## QA notes
**How to test:**

1. Prepare subscribers for import
2. Add a tag in the new text input field
   a. Select existing tag
   b. Type a not existing tag
3. Finish import
4. Check if all imported subscribers have selected tags
 
## Linked tickets
[MAILPOET-4558](https://mailpoet.atlassian.net/browse/MAILPOET-4558)

### DIFF
--- a/mailpoet/assets/js/src/subscribers/importExport/import/step_data_manipulation/assign_tags.tsx
+++ b/mailpoet/assets/js/src/subscribers/importExport/import/step_data_manipulation/assign_tags.tsx
@@ -27,12 +27,13 @@ export function AssignTags({
           {MailPoet.I18n.t('assignTagsDescription')}
         </p>
       </div>
-      <div className="mailpoet-settings-inputs">
+      <div className="mailpoet-settings-inputs mailpoet-import-tags">
         <TokenField
           name="tags"
           onChange={handleChange}
           suggestedValues={tags}
           selectedValues={selectedTags}
+          placeholder={MailPoet.I18n.t('addNewTag')}
         />
       </div>
     </>
--- a/mailpoet/assets/js/src/subscribers/importExport/import/step_data_manipulation/assign_tags.tsx
+++ b/mailpoet/assets/js/src/subscribers/importExport/import/step_data_manipulation/assign_tags.tsx
@@ -1,0 +1,40 @@
+import { MailPoet } from 'mailpoet';
+import { useCallback } from 'react';
+import { TokenField } from '../../../../common/form/tokenField/tokenField';
+
+interface Props {
+  selectedTags: string[];
+  setSelectedTags: (value) => void;
+}
+
+export function AssignTags({
+  selectedTags,
+  setSelectedTags,
+}: Props): JSX.Element {
+  const handleChange = useCallback(
+    ({ value }): void => {
+      setSelectedTags(value);
+    },
+    [setSelectedTags],
+  );
+
+  const tags = window.mailpoet_tags.map((tag) => tag.name);
+  return (
+    <>
+      <div className="mailpoet-settings-label">
+        {MailPoet.I18n.t('assignTagsLabel')}
+        <p className="description">
+          {MailPoet.I18n.t('assignTagsDescription')}
+        </p>
+      </div>
+      <div className="mailpoet-settings-inputs">
+        <TokenField
+          name="tags"
+          onChange={handleChange}
+          suggestedValues={tags}
+          selectedValues={selectedTags}
+        />
+      </div>
+    </>
+  );
+}
--- a/mailpoet/assets/js/src/subscribers/importExport/import/step_data_manipulation/assign_tags.tsx
+++ b/mailpoet/assets/js/src/subscribers/importExport/import/step_data_manipulation/assign_tags.tsx
@@ -18,7 +18,7 @@ export function AssignTags({
     [setSelectedTags],
   );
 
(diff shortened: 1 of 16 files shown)


mailpoet/lib/Subscribers/ImportExport/Import/Import.php
```diff
@@ -619,4 +634,42 @@ public function addSubscribersToSegments(array $subscribersIds, array $segmentsI
...
+            $tagId,
+            $this->createdAt,
+          ];
+        }, $subscriberIdsChunk));
+      }
+      $this->importExportRepository->insertMultiple(
+        SubscriberTagEntity::class,
+        $columns,
+        $data
+      );
```
costasovo: I see that we use multiple inserts in batches and we split them into chunks by splitting subscribers. I just want to point out that because of the nested for each we may actually generate multiple insert with more than self::DB_QUERY_CHUNK_SIZE items.
The chunk size is 100. 100 subscribers each with 100 tags will generate multiple inserts with 10000 rows.
This is probably not a big deal since there most probably won't be that many tags. If we want to be strict we modify the script so that it counts the number of generated rows and flushes them to DB each time they reach 100 and then last time at the end.

mailpoet/assets/js/src/subscribers/importExport/import/step_data_manipulation/assign_tags.tsx
```diff
@@ -0,0 +1,40 @@
...
+  setSelectedTags,
+}: Props): JSX.Element {
+  const handleChange = useCallback(
+    ({ value }): void => {
+      setSelectedTags(value);
+    },
+    [setSelectedTags],
+  );
+
+  const tags = window.mailpoet_tags.map((tag) => tag.name);
```
costasovo: We wanted to prevent adding more usages of `window` within the new code. I guess we could assign the tags on the MailPoet module as we do with many other items we pass from the server.

costasovo (commented):
Looks good. I added two small remarks.

mailpoet/lib/Subscribers/ImportExport/Import/Import.php
lysyjan: @costasovo I think you overlooked the custom solution for [inserting multiple rows in the importExportRepository](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php#L74-L102). This custom solution inserts all tags for the chunk size with one query. That means that 100 subscribers with 100 tags are inserted by one multi-insert.
I tried to test this solution on my local, and 1000 subscribers with 5 tags were stored by 10 queries.

mailpoet/assets/js/src/subscribers/importExport/import/step_data_manipulation/assign_tags.tsx
lysyjan: I addressed this in this commit [10059c1](https://github.com/mailpoet/mailpoet/pull/4347/commits/10059c165b5dc9b8b0da6cea5405fda86c3ee17e)

mailpoet/lib/Subscribers/ImportExport/Import/Import.php
costasovo: > I tried to test this solution on my local, and 1000 subscribers with 5 tags were stored by 10 queries.

@lysyjan I think that exactly this might potentially be the issue. Our constant self::DB_QUERY_CHUNK_SIZE is set to 100. My interpretation of the constant (please correct me if I'm wrong) is that we want to limit the number of rows in one multi-insert to 100.  Importing 1000 subscribers with 5 tags each in 10 queries means we had 500 rows in a single multi-insert query which is five times more than the chunk size defined by the constant. So the current implementation allows bigger chunks than defined in the constant and theoretically can lead to huge multi-insert queries.
I think that the expected behavior should be 1000 subscribers with 5 tags = 5000 rows -> so we need 50 queries (100 rows per query).

This may cause an issue only in some extreme cases like subscribers with hundreds of tags. The potential issue with such huge inserts could be that if SQL query becomes bigger than `max_allowed_packet`, MySQL will throw an error. https://www.mysqltutorial.org/mysql-insert-multiple-rows/

lysyjan: @costasovo Thank you for your explanation. It makes sense to me now. We have the same behavior for the segments. So, I would suggest changing this behavior for both cases. I could change the `foreach` order so that each query inserts only the exact count of records. Do you agree with my proposal?

costasovo: I haven't realized that it is doable by changing the order of the loops. If so, yes I agree.
An alternative solution could be storing rows in an array and each time it has 100 items we do the query + one query at the end.

lysyjan: @costasovo I changed it and added a commit with another fix.

costasovo (approved):
LGTM!

Aschepikov (approved):
This is verified, good job!

PR #4349, opened 2022-09-07
Make tags badges linkable to tags filtering [MAILPOET-4564]

lysyjan:
## Description
This pull request makes the tag badges on the subscribers list clickable. When a user clicks on the badge, the list should use a filter with this tag.
 
## QA notes
1. Go to Subscribers page
2. Make sure you edit a few subscribers and add a few tags to them
3. When on a Subscribers page, click on the tags separately and make sure you see subscribers with those tags
4. Check if anywhere else you see tag badges and if they are clickable
Note: The behavior should be equivalent to the list's badges.
 
## Linked tickets
[MAILPOET-4564]

[MAILPOET-4564]: https://mailpoet.atlassian.net/browse/MAILPOET-4564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/common/tag/tags.tsx
+++ b/mailpoet/assets/js/src/common/tag/tags.tsx
@@ -8,10 +8,18 @@ type Segment = {
   id?: string;
 };
 
+type SubscriberTag = {
+  id: string;
+  name: string;
+  subscriber_id: string;
+  tag_id: string;
+};
+
 type Props = {
   children?: ReactNode;
   dimension?: 'large';
   segments?: Segment[];
+  subscriberTags?: SubscriberTag[];
   strings?: string[];
   variant?: 'average' | 'good' | 'excellent' | 'list' | 'unknown' | 'wordpress';
   isInverted?: boolean;
@@ -21,6 +29,7 @@ function Tags({
   children,
   dimension,
   segments,
+  subscriberTags,
   strings,
   variant,
   isInverted,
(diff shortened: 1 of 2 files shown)


samnajian (commented):
While this works really good I noticed 2 items that can be improved:

1) The `Tags` can use some patterns like render props or HOC or alike to avoid the need to pass `subscriberTags` into it and then updating it to render these new tag variants, I see this could be done for other tag variants like `segments` or even `string`, please feel free to adjust this or delegate to another ticket.

2) After clicking a tag and having the listing filtered to the selected tag, the tag remains clickable and since clicking on it doesn't do anything anymore it feels like the UI is somehow broken, please see https://d.pr/i/K5y8eN, this is the same behavior with other tags variants likes `Lists`, so let's see if @NeosinneR thinks we should improve it in this ticket or leave it for later.

NeosinneR:
@samnajian good catch. As it also includes lists, I created a [separate ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4627) to improve it, and it's okay by me as it is currently implemented.

lysyjan:
@samnajian Because I want to avoid a bigger PR, I created [a maintenance ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4628). Thanks for your suggestions 👍

samnajian (approved):
Nice additions, thanks!

Aschepikov (approved):
This is verified, good job!

PR #4354, opened 2022-09-08
Add logging of workflow steps [MAILPOET-4463]

johnolek:
## Description

Adds a new type of automation record, workflow run logs, that store data about individual steps that have been run.

## Code review notes

Automation DB schema must be reset in order to get the new table.

## QA notes

QA step can be skipped.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

Automation devs, reset your automation DB schema.

[MAILPOET-4463]: https://mailpoet.atlassian.net/browse/MAILPOET-4463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -130,7 +130,7 @@ class StepHandler {
         throw $e;
       } finally {
         try {
-          $this->hooks->doWorkflowStepAfterRun($step, $log);
+          $this->hooks->doWorkflowStepAfterRun($log);
         } catch (Exception $e) {
           $log->addError($e);
         }
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -120,7 +120,7 @@ class StepHandler {
 
     $stepType = $step->getType();
     if (isset($this->stepRunners[$stepType])) {
-      $log = $this->createWorkflowRunLog($workflowRun, $step, $args);
+      $log = new WorkflowRunLog($workflowRun->getId(), $step->getId(), $args);
       try {
         $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
         $log->markCompleted();
@@ -134,7 +134,7 @@ class StepHandler {
         } catch (Exception $e) {
           $log->addError($e);
         }
-        $this->workflowRunLogStorage->updateWorkflowRunLog($log);
+        $this->workflowRunLogStorage->createWorkflowRunLog($log);
       }
     } else {
       throw new InvalidStateException();
@@ -161,17 +161,6 @@ class StepHandler {
 
     // enqueue next step
     $this->actionScheduler->enqueue(Hooks::WORKFLOW_STEP, $nextStepArgs);
-
     // TODO: allow long-running steps (that are not done here yet)
-  }
-
-  private function createWorkflowRunLog(WorkflowRun $workflowRun, Step $step, array $args): WorkflowRunLog {
-    $log = new WorkflowRunLog($workflowRun->getId(), $step->getId(), $args);
-    $logId = $this->workflowRunLogStorage->createWorkflowRunLog($log);
-    $workflowRunLog = $this->workflowRunLogStorage->getWorkflowRunLog($logId);
-    if (!$workflowRunLog instanceof WorkflowRunLog) {
-      throw new InvalidStateException();
-    }
-    return $workflowRunLog;
   }
 }
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -126,13 +126,13 @@ class StepHandler {
         $log->markCompletedSuccessfully();
       } catch (Throwable $e) {
         $log->markFailed();
-        $log->addError($e);
+        $log->setError($e);
         throw $e;
       } finally {
         try {
           $this->hooks->doWorkflowStepAfterRun($log);
         } catch (Throwable $e) {
-          $log->addError($e);
+          // Ignore integration errors
         }
         $this->workflowRunLogStorage->createWorkflowRunLog($log);
       }
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -123,8 +123,8 @@ class StepHandler {
       $log = new WorkflowRunLog($workflowRun->getId(), $step->getId(), $args);
       try {
         $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
-        $log->markCompleted();
-      } catch (Exception $e) {
+        $log->markCompletedSuccessfully();
+      } catch (\Exception $e) {
         $log->markFailed();
         $log->addError($e);
         throw $e;
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -124,14 +124,14 @@ class StepHandler {
       try {
         $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
         $log->markCompletedSuccessfully();
-      } catch (\Exception $e) {
+      } catch (Throwable $e) {
         $log->markFailed();
         $log->addError($e);
         throw $e;
       } finally {
         try {
           $this->hooks->doWorkflowStepAfterRun($log);
-        } catch (Exception $e) {
+        } catch (Throwable $e) {
           $log->addError($e);
         }
         $this->workflowRunLogStorage->createWorkflowRunLog($log);
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -120,7 +120,7 @@ class StepHandler {
 
     $stepType = $step->getType();
     if (isset($this->stepRunners[$stepType])) {
-      $log = new WorkflowRunLog($workflowRun->getId(), $step->getId(), $args);
+      $log = new WorkflowRunLog($workflowRun->getId(), $step->getId());
       try {
         $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
         $log->markCompletedSuccessfully();
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -6,9 +6,11 @@ use Exception;
 use MailPoet\Automation\Engine\Control\Steps\ActionStepRunner;
 use MailPoet\Automation\Engine\Data\Step;
 use MailPoet\Automation\Engine\Data\WorkflowRun;
+use MailPoet\Automation\Engine\Data\WorkflowRunLog;
 use MailPoet\Automation\Engine\Exceptions;
 use MailPoet\Automation\Engine\Exceptions\InvalidStateException;
 use MailPoet\Automation\Engine\Hooks;
+use MailPoet\Automation\Engine\Storage\WorkflowRunLogStorage;
 use MailPoet\Automation\Engine\Storage\WorkflowRunStorage;
 use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Automation\Engine\WordPress;
@@ -33,17 +35,27 @@ class StepHandler {
   /** @var array<string, StepRunner> */
   private $stepRunners;
 
+  /** @var WorkflowRunLogStorage */
+  private $workflowRunLogStorage;
+
+  /** @var Hooks */
+  private $hooks;
+
   public function __construct(
     ActionScheduler $actionScheduler,
     ActionStepRunner $actionStepRunner,
+    Hooks $hooks,
     WordPress $wordPress,
     WorkflowRunStorage $workflowRunStorage,
+    WorkflowRunLogStorage $workflowRunLogStorage,
     WorkflowStorage $workflowStorage
   ) {
     $this->actionScheduler = $actionScheduler;
     $this->actionStepRunner = $actionStepRunner;
+    $this->hooks = $hooks;
     $this->wordPress = $wordPress;
     $this->workflowRunStorage = $workflowRunStorage;
+    $this->workflowRunLogStorage = $workflowRunLogStorage;
     $this->workflowStorage = $workflowStorage;
   }
 
@@ -108,7 +120,22 @@ class StepHandler {
 
     $stepType = $step->getType();
     if (isset($this->stepRunners[$stepType])) {
-      $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
+      $log = $this->createWorkflowRunLog($workflowRun, $step, $args);
+      try {
+        $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
+        $log->markCompleted();
+      } catch (Exception $e) {
+        $log->markFailed();
+        $log->addError($e);
+        throw $e;
+      } finally {
+        try {
+          $this->hooks->doWorkflowStepAfterRun($step, $log);
+        } catch (Exception $e) {
+          $log->addError($e);
+        }
+        $this->workflowRunLogStorage->updateWorkflowRunLog($log);
+      }
     } else {
       throw new InvalidStateException();
     }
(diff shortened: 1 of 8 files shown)


mailpoet/lib/Automation/Engine/Control/StepHandler.php
```diff
@@ -108,7 +120,22 @@ private function handleStep(array $args): void {
 
     $stepType = $step->getType();
     if (isset($this->stepRunners[$stepType])) {
-      $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
+      $log = $this->createWorkflowRunLog($workflowRun, $step, $args);
+      try {
+        $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
+        $log->markCompleted();
+      } catch (Exception $e) {
```
JanJakes: We should probably use `Throwable` to catch also non-Exception based errors.

mailpoet/lib/Automation/Engine/Control/StepHandler.php
```diff
@@ -108,7 +120,22 @@ private function handleStep(array $args): void {
...
+        $log->markFailed();
+        $log->addError($e);
+        throw $e;
+      } finally {
+        try {
+          $this->hooks->doWorkflowStepAfterRun($step, $log);
+        } catch (Exception $e) {
```
JanJakes: Same here (`Throwable`, I mean).

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
```diff
@@ -0,0 +1,170 @@
...
+      'completed_at' => $this->completedAt ? $this->completedAt->format(DateTimeImmutable::W3C) : null,
+      'args' => Json::encode($this->args),
+      'errors' => Json::encode($this->errors),
+      'data' => Json::encode($this->data),
+    ];
+  }
+
+  public function markCompleted(): void {
+    $this->status = self::STATUS_COMPLETED;
+    $this->completedAt = new DateTimeImmutable();
```
JanJakes, line 117: Seeing we have `completedAt`, the question might be whether we should have `failedAt` as well. Or none of them, if `updatedAt` can handle that...

JanJakes (commented):
This looks good to me!

I added some small notes, and I have a question or two:

1. I see the logs are mutable. Did you consider making them immutable? Or would that create too much noise and too many records? (Like starting step record... step completed record.) Thinking further, it would probably be harder to determine which step succeeded, failed, etc.
2. Is both `$data` and `$errors` used? Not sure if I can spot it. Is maybe just one enough? Maybe `$data` and we'd be free to save errors or anything else?
3. I see you're also adding the possibility to save multiple errors (and using it). Isn't it simpler and sufficient to save just the first error and abort? I mean... if we want to allow integrations act on errors, maybe we can create some configurable error handlers in the future. Or do we already have a use case for this?

I might have missed some things you've thought trough, I scanned it rather quickly, so I hope my questions make sense.

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
johnolek: I think it makes sense to make these records immutable and don't want to have `updatedAt`. If I do end up keeping them mutable then I at least wouldn't want to have to rely on `updatedAt` to mean multiple things.

Maybe we could use something slightly more generic than `completedAt`, like `endedAt`? Or I suppose `completedAt` doesn't necessarily imply "completed successfully," so maybe we should just use `completedAt` in both places.

johnolek:
> I see the logs are mutable. Did you consider making them immutable? Or would that create too much noise and too many records? (Like starting step record... step completed record.) Thinking further, it would probably be harder to determine which step succeeded, failed, etc.

Immutable makes more sense for a log and I think it will allow the code to be much simpler. Thanks for bringing this up!

I don't think we would necessarily need separate records for starting and ending. We could create the record in memory, add whatever we want before/after the step runs, and only persist it to the database at the very end. I think that's what I'll try anyway, unless you can see any fundamental flaws with that approach.

> Is both $data and $errors used? Not sure if I can spot it. Is maybe just one enough? Maybe $data and we'd be free to save errors or anything else?

They are both used, and they probably aren't both necessary. The initial idea with $data is that it would be used primarily by integrations, but right now the code doesn't have a strong distinction between the two, and an integration could just as easily add an error as some piece of data.

I'm not sure if it makes a ton of sense to write code right now that is intended specifically for integrations, or to put a lot of effort into distinguishing between integration-provided data and data/errors coming from the automation itself. I'll need to give this some more thought.

> I see you're also adding the possibility to save multiple errors (and using it). Isn't it simpler and sufficient to save just the first error and abort? I mean... if we want to allow integrations act on errors, maybe we can create some configurable error handlers in the future. Or do we already have a use case for this?

This is a very good point and I'm not aware of any use cases that would require us to store multiple errors. The current version of the code does allow for the possibility of adding multiple errors: one from the execution of the step itself and another from any error occurring while running the WordPress action that exposes the log object. I don't think it's worthwhile to have both of those things though, especially because it doesn't really make sense to think of an error occurring as part of an action callback as an error with the step itself. I'll do my best to simplify this.

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
JanJakes: That's true, `completedAt` could work for both.

JanJakes:
> Immutable makes more sense for a log and I think it will allow the code to be much simpler. Thanks for bringing this up!

> I don't think we would necessarily need separate records for starting and ending. We could create the record in memory, add whatever we want before/after the step runs, and only persist it to the database at the very end. I think that's what I'll try anyway, unless you can see any fundamental flaws with that approach.

I almost wanted to say that by doing this, we may lose debugging info about the start of some problematic steps that will just fail and never log (timeout, out-of-memory, etc.), but then I realized that Action Scheduler already logs these things, so it's likely not necessary, and I like your idea about how to have a single record per step while keeping it immutable :+1:

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
johnolek: I've decided to simply use `completedAt` for both situations.

mailpoet/lib/Automation/Engine/Control/StepHandler.php
```diff
@@ -108,7 +120,22 @@ private function handleStep(array $args): void {
...
+        $log->markCompletedSuccessfully();
+      } catch (Throwable $e) {
+        $log->markFailed();
+        $log->setError($e);
+        throw $e;
+      } finally {
+        try {
+          $this->hooks->doWorkflowStepAfterRun($log);
+        } catch (Throwable $e) {
+          // Ignore integration errors
```
johnolek, line 135: I'm not sure if simply ignoring errors during the hook is what we want since it could make debugging very difficult for integration authors. @JanJakes can you think of a better way we could handle this, or does it seem fine for now?

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
```diff
@@ -0,0 +1,175 @@
...
+use DateTimeImmutable;
+use InvalidArgumentException;
+use MailPoet\Automation\Engine\Exceptions\InvalidStateException;
+use MailPoet\Automation\Engine\Exceptions\UnexpectedValueException;
+use MailPoet\Automation\Engine\Utils\Json;
+use Throwable;
+
+class WorkflowRunLog {
+
+  const STATUS_RUNNING = 'running';
```
johnolek, line 12: We might be able to just delete this status. I don't think it's meaningful, but if there's a use case for having it please let me know.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
```diff
@@ -60,12 +60,28 @@ public function createSchema(): void {
...
       );
     ");
+
+    $this->runQuery("
+      CREATE TABLE {$this->prefix}workflow_run_logs (
+        id int(11) unsigned NOT NULL AUTO_INCREMENT,
+        workflow_run_id int(11) unsigned NOT NULL,
+        step_id varchar(255) NOT NULL,
+        status varchar(255) NOT NULL,
+        started_at timestamp NOT NULL,
```
johnolek, line 71: Do you think it would be useful to have `created_at` in addition to `started_at`? I think they would be the same every time, unless we wanted `created_at` to correspond to when the record is actually saved, in which case I think it would always be the same as `completed_at`.

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
```diff
@@ -0,0 +1,174 @@
...
+      $newData = $this->getData();
+      $newData[$key] = $value;
+      $encoded = Json::encode($newData);
+      $decoded = Json::decode($encoded);
+      if ($decoded !== $newData) {
+        throw new InvalidArgumentException('$value must be serializable');
+      }
+    } catch (InvalidStateException | InvalidArgumentException | UnexpectedValueException $e) {
+      throw new InvalidArgumentException("Invalid data provided for key $key.");
+    }
```
johnolek: This is probably not the best way to handle this, but I do think it's important that we don't allow setting vales that we wouldn't be able to serialize/JSON encode and retrieve out of the database. Is there a more elegant way of handling this?

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
```diff
@@ -0,0 +1,174 @@
...
+  private $error;
+
+  /** @var array */
+  private $data;
+
+  /** @var string */
+  private $stepId;
+
+  /** @var array */
+  private $args;
```
johnolek: I wasn't sure if it would be meaningful to store the $args as part of the log or if we would be able to easily/reliably retrieve them from a related record, so I erred on the side of redundancy. Please let me know if you think we should remove this.

johnolek:
@JanJakes I'm relatively happy with this PR after making some changes and I think it's ready for a final review. I added a few comments about things I'm still unsure about, though I don't think any of them are show stoppers. That being said, I'm happy to make a few final tweaks before moving this forward!

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
JanJakes: I guess if we don't have a use for it yet, we can indeed delete it.

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
JanJakes: These are the step args? I guess they're not run-specific, only workflow-specific. And since we have versioning, we could find them through that, if needed (log -> workflow run -> workflow version). I'm not sure if we'll need them in the log, what do you think?

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
```diff
@@ -0,0 +1,174 @@
...
+  public function __construct(
+    int $workflowRunId,
+    string $stepId,
+    array $args,
+    int $id = null
+  ) {
+    $this->workflowRunId = $workflowRunId;
+    $this->stepId = $stepId;
+    $this->args = $args;
+    $this->status = self::STATUS_RUNNING;
```
JanJakes, line 47: Hmm, now I see that some kind of a "default" state might be needed. So we may as well keep this one.

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  public function getData(): array {
+    return $this->data;
+  }
+
+  /**
+   * @return DateTimeImmutable|null
+   */
+  public function getCompletedAt() {
```
JanJakes: You can use PHP-native return typehint, I guess.

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
JanJakes: What is it supposed to do, exactly? The `Json::encode()` function will throw a `InvalidStateException` if the input is not serializable, so that could be enough by itself, couldn't it? If you want to convert it to `InvalidArgumentException`, we can keep the catch + throw.

It may be also a bit confusing that you throw `new InvalidArgumentException('$value must be serializable')` and immediately below catch it, using a different message. As for `UnexpectedValueException`, that's not even thrown anywhere, is it?

I also just realized this is not used anywhere yet, but I guess it makes sense to have it.

mailpoet/lib/Automation/Engine/Migrations/Migrator.php
JanJakes: Yeah, I think I don't see a need for `created_at` at the moment.

mailpoet/lib/Automation/Engine/Storage/WorkflowRunLogStorage.php
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * @param int $workflowRunId
+   * @return WorkflowRunLog[]
+   * @throws InvalidStateException
+   */
+  public function getLogsForWorkflowRun(int $workflowRunId): array {
+    $table = esc_sql($this->table);
+    $query = $this->wpdb->prepare("SELECT * FROM $table WHERE workflow_run_id = %d", $workflowRunId);
```
JanJakes: This is just a tiny random idea that we may forget otherwise, but we could sort the logs in the order they were created. We can either use `started_at`, or (probably even better), simply `id`.

JanJakes (dismissed):
This looks good to me, and I consider it mergeable. I'll still leave it to you @johnolek, as I guess you may want to do some small tweaks based on our discussion but feel free to merge it if you need it in `trunk`; otherwise, I'll do so tomorrow.

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
johnolek: Yes, these are the step args. I should have made that clearer! I think that I'll remove them for now. The main concern would be if the step args could change before the step is run, or somehow get out of sync with what's in the workflow version, and that doesn't seem to be the case.

I also don't want to include more than what we know we'll need. It's not like it would be difficult to add step data back later if we come up with a specific use case for it. Let's keep is as simple as we can for now.

mailpoet/lib/Automation/Engine/Storage/WorkflowRunLogStorage.php
johnolek: Good call, thanks!

mailpoet/lib/Automation/Engine/Data/WorkflowRunLog.php
johnolek: I agree that this code is very confusing and I need to make it simpler.

> I also just realized this is not used anywhere yet, but I guess it makes sense to have it.

The main idea here is that integrations would be able to use this public method to store arbitrary data in the log, since the log object is getting exposed in a WordPress action.

The general idea with the check is I don't want to allow setting any kind of data that would end up being different after it's retrieved from the database and decoded.

Taking an example from one of the tests I wrote, you could have something like this:

```php
$badData = [
  function() { echo 'closures cannot be serialized'; }
];
$log->setData('badData', $badData);
```

This doesn't cause an error when given to `Json::encode`, but the encoded version just ends up being an empty object:

```json
{"badData":[{}]}
```

Trying to JSON encode a newsletter object causes a similar result (encoded without error as an empty object). And then there are cases where someone might try to store something like a stdClass, which works very well with encoding, but after decoding it it would be an array instead of a stdClass, which could be confusing.

If we wanted to allow storing more complex data, like objects, we could use PHP serialization instead of JSON, but that seems unnecessary and more likely to cause problems than anything.

Right now I'm leaning towards limiting the type of $value to just scalars, maybe including arrays that only have scalar values.

`UnexpectedValueException` would come from the [jsonNotObject](https://github.com/mailpoet/mailpoet/blob/MAILPOET-4463%2Fadd-step-logging/mailpoet/lib/Automation/Engine/Utils/Json.php#L25-L25) exception from decoding the JSON. I'm not sure if it would even be possible to trigger that in this case (seems doubtful) but I wanted to be exhaustive with the list of exceptions I was catching.

johnolek: I went ahead with allowing scalar values and arrays that exclusively contain scalar values. If this seems overly complex then I'm happy to change it.

johnolek:
@JanJakes Thanks again for the review! I just pushed some updates. Since this is very close I would say feel free to make any last tweaks that you might like yourself before merging. I would prefer for you to take one last look mainly at the new validation for allowed values for `setData`.

I did end up having to rebase onto trunk because of the new `Step` constructor signature.

JanJakes (approved):

PR #4355, opened 2022-09-08
Fix logic to decide which PHPStan baseline file to use [MAILPOET-4626]

rodrigoprimo:
This PR fixes a bug in the logic to decide which PHPStan baseline file to use that it was making the baseline file for PHP 8.1 not be used for PHP versions >= PHP 8.1.1. It also removes a few unmatched ignored errors from phpstan-8.1-baseline.neon that were uncovered when fixing the baseline load problem. See commit messages for more details on the changes.

I'm unsure if this PR needs to pass QA. I'm inclined to think it doesn't, but I will leave it for the reviewer to decide.

@triple0t I know you are on L2 support this week. So it is ok if you prefer to leave this review for next week. I chose you since you worked on this previously and I want to make sure I'm not missing something.

[MAILPOET-4626]

[MAILPOET-4626]: https://mailpoet.atlassian.net/browse/MAILPOET-4626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tasks/phpstan/phpstan-8.1-baseline.neon
+++ b/mailpoet/tasks/phpstan/phpstan-8.1-baseline.neon
@@ -9,17 +9,6 @@ parameters:
 			count: 1
 			path: ../../lib/Automation/Engine/Endpoints/Workflows/WorkflowTemplatesGetEndpoint.php
 
-
-		-
-			message: "#^Cannot cast string|void to string\\.$#"
-			count: 2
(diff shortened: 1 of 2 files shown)


triple0t (approved):
LGTM

PR #4356, opened 2022-09-08
Add method to detect if WooCommerce Custom Order Tables is enabled [MAILPOET-4567]

rodrigoprimo:
I think it is not necessary to send this PR to QA but I will leave this decision to the reviewer.

To test the code, I called it from one of the integration tests and played with the environment of the test by enabling and disabling WooCommerce Custom Order Tables via the web interface.

[MAILPOET-4567]

[MAILPOET-4567]: https://mailpoet.atlassian.net/browse/MAILPOET-4567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tasks/phpstan/phpstan.neon
+++ b/mailpoet/tasks/phpstan/phpstan.neon
@@ -52,6 +52,14 @@ parameters:
       message: "#^Cannot cast string|void to string\\.$#"
       count: 5
       path: ../../lib/Automation/Engine/Storage/WorkflowStorage.php
+    -
+      message: '/^Call to static method custom_orders_table_usage_is_enabled\(\) on an unknown class Automattic\\WooCommerce\\Utilities\\OrderUtil\.$/'
+      count: 1
+      path: ../../lib/WooCommerce/Helper.php
+    -
+      message: '/^Call to function method_exists\(\) with/'
+      count: 1
+      path: ../../lib/WooCommerce/Helper.php
   reportUnmatchedIgnoredErrors: true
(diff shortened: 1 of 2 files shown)


mailpoet/tasks/phpstan/phpstan.neon
```diff
@@ -52,6 +52,14 @@ parameters:
       message: "#^Cannot cast string|void to string\\.$#"
       count: 5
       path: ../../lib/Automation/Engine/Storage/WorkflowStorage.php
+    -
+      message: '/^Call to static method custom_orders_table_usage_is_enabled\(\) on an unknown class Automattic\\WooCommerce\\Utilities\\OrderUtil\.$/'
+      count: 1
+      path: ../../lib/WooCommerce/Helper.php
+    -
+      message: '/^Call to function method_exists\(\) with/'
```
rodrigoprimo, line 60: I was not able to easily find how to match the whole error string and that is why I added only part of the string. I was running into regex parse errors when trying to add the whole string and escape special characters.

costasovo: It mentions the function `method_exists` and also it is defined for `../../lib/WooCommerce/Helper.php` so I would say it is specific enough.

costasovo (approved):
LGTM! Since there is no testable change, I'll merge it.

PR #4357, opened 2022-09-09
Add New Pull Request Template

veljkho:
This pull request is to make a pull request template according to this [p2 post](https://wp.me/pcNwfB-19e-p2).


JanJakes (approved):
Thanks for organizing this!

PR #4359, opened 2022-09-09
Workflow data format [MAILPOET-4523]

JanJakes:
## Description

This PR implements a **new workflow format** bringing the following updates:
1. Possibility to represent multiple next steps (`next_step_id` string changed to `next_steps` array).
2. Possibility to store metadata along the edges of the workflow graph. To achieve that, `next_steps` stores an array of objects, currently `{ id: string }`, that can be extended in the future with any properties we might need.
3. Triggers are now ordered, so workflow rendering and evaluation are unambiguous and deterministic. For that, a new `root` step was added. This step is mandatory for every workflow, and its main task is to hold the array of `next_steps` at the moment. Triggers need no longer be searched for in the `steps` object.
4. Having direct `root` descendants that are not triggers is valid for a draft but invalid for an active workflow.
5. Due to point 3. and 4., the empty trigger placeholder (`core:empty`) is no longer needed, so it was removed. It is now inferable from the workflow structure which steps need a placeholder rendered before them (those that are direct descendants of `root` but are not of type `trigger`).
6. All of the basic step fields (all fields we currently have) are now mandatory. No `args` are represented by an empty object, no `next_steps` are represented by an empty array.
7. The workflow structure is described by a `WorkflowSchema` built using the [Validator](https://github.com/mailpoet/mailpoet/pull/4027). Semantic checks will be done separately as part of a validation task.

Here's an example of the new format:

```json
{
  "id": 1,
  "name": "Untitled",
  "status": "draft",
  "steps": {
    "root": {
      "id": "root",
      "type": "root",
      "key": "core:root",
      "args": {},
      "next_steps": [
        {"id": "4d65017d55a59e70"}
      ]
    },
    "4d65017d55a59e70": {
      "id": "4d65017d55a59e70",
      "type": "trigger",
      "key": "mailpoet:someone-subscribed",
      "args": {},
      "next_steps": [
        {"id": "22e137f29b8e4acb"}
      ],
    },
    "22e137f29b8e4acb": {
      "id": "22e137f29b8e4acb",
      "type": "action",
      "key": "core:delay",
      "args": {},
      "next_steps": [
        {"id": "8e45d26ad71e1f7a"}
      ],
    },
    "8e45d26ad71e1f7a": {
      "id": "8e45d26ad71e1f7a",
      "type": "action",
      "key": "mailpoet:send-email",
      "args": {},
      "next_steps": [],
    }
  }
}

```

---

Additionally, the PR includes some smaller fixes and improvements. The last commit is a simple move of the API from Automation to MailPoet root.

## Code review notes

Don't forget to reset all your automation DB schema if you're trying this out.

## QA notes

QA step can be skipped.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/619

## Linked tickets

[MAILPOET-4523]

## After-merge notes

Automation devs, reset your automation DB schema.

[MAILPOET-4523]: https://mailpoet.atlassian.net/browse/MAILPOET-4523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/editor/store/actions.ts
```diff
@@ -54,12 +54,12 @@ export function* save() {
...
     path: `/workflows/${workflow.id}`,
     method: 'PUT',
-    data: workflow,
+    data: { ...workflow },
   });
 
   return {
     type: 'SAVE',
-    workflow: data.data,
+    workflow: data?.data ?? workflow,
```
johnolek, line 67: When does this fallback come into play? And does it matter that it would be using the original workflow object?

mailpoet/lib/Automation/Engine/Validators/WorkflowSchema.php
```diff
@@ -0,0 +1,52 @@
...
+      'next_steps' => self::getNextStepsSchema()->required(),
+    ]);
+  }
+
+  public static function getRootStepSchema(): ObjectSchema {
+    return Builder::object([
+      'id' => Builder::string()->pattern('^root$'),
+      'type' => Builder::string()->pattern('^root$'),
+      'key' => Builder::string()->pattern('^core:root$'),
+      'args' => Builder::array()->maxItems(0)->maxItems(0),
```
johnolek: Is there a reason for having `maxItems(0)` twice?

mailpoet/lib/Automation/Engine/Exceptions.php
```diff
@@ -85,11 +85,11 @@ public static function workflowRunNotFound(int $id): NotFoundException {
       ->withMessage(sprintf(__("Workflow run with ID '%d' not found.", 'mailpoet'), $id));
   }
 
-  public static function workflowStepNotFound(string $id): NotFoundException {
+  public static function workflowStepNotFound(string $key): NotFoundException {
     return NotFoundException::create()
       ->withErrorCode(self::WORKFLOW_STEP_NOT_FOUND)
       // translators: %s is the ID of the workflow step.
```
johnolek: May want to update this comment as well to say that it's the key, not the ID, though it seems like it should be pretty clear from the error message itself

mailpoet/lib/Automation/Engine/Control/StepHandler.php
```diff
@@ -113,10 +113,11 @@ private function handleStep(array $args): void {
       throw new InvalidStateException();
     }
 
+    $nextStep = $step->getNextSteps()[0] ?? null;
     $nextStepArgs = [
       [
         'workflow_run_id' => $workflowRunId,
-        'step_id' => $step->getNextStepId(),
+        'step_id' => $nextStep ? $nextStep->getId() : null,
```
johnolek, line 120: It might be nice to update the check on line 130 to look at `$nextStep` instead of digging into `$nextStepArgs`.

mailpoet/lib/Automation/Engine/Exceptions/Exception.php
```diff
@@ -2,12 +2,14 @@
...
 
+use Exception as PhpException;
+use MailPoet\API\REST\Exception as RestException;
 use Throwable;
 
 /**
  * Frames all MailPoet Automation exceptions ("$e instanceof MailPoet\Automation\Exception").
  */
-abstract class Exception extends \Exception {
+abstract class Exception extends PhpException implements RestException {
```
johnolek, line 12: It seems a little bit strange on its face to have all Automation exceptions implement a `RestException` interface. Doesn't that kind of imply that we should only use these exceptions in code that's run as part of a REST request? I suppose that's probably true most of the time, unless action scheduler events aren't triggered as part of a REST request...

Anyway, I'm sure this is fine, it just struck me as a little strange when I first saw it.

johnolek (changes requested):
Overall looks great! Just a couple of extremely minor comments/questions, nothing close to any kind of show stopper.

mailpoet/assets/js/src/automation/editor/store/actions.ts
JanJakes: It can occur when the error handler implemented in https://github.com/mailpoet/mailpoet/pull/4359/commits/e31f7d66ffad8c4537b25fafc064db0fe3cabe52 catches an error and displays a notice. In such a case, I believe using the original workflow is probably correct because we were not able to save the workflow for some reason. The result will be something like this (still rendering the original workflow):

Screen Shot 2022-09-12 at 7 13 00

mailpoet/lib/Automation/Engine/Exceptions/Exception.php
JanJakes: Hmm, yeah, that's a good point, it does feel a bit strange now that all of these exceptions implement `MailPoet\API\REST\Exception`. I guess it partly comes from the extraction to `MailPoet\API\REST` while I wanted to simply keep the same behavior as before.

In practice, this is quite a handy thing to do — we don't have to think about some complex structure where some exceptions implement the interface, and some don't. It's also not "extending" a REST exception, only happens to implement a REST interface, which I guess still is slightly better.

We could maybe say that the `getStatusCode()`, `getErrorCode()`, and `getErrors()` are a universal thing we want to have for all exceptions, and then maybe the interface could be just `MailPoet\Exception` while the API would just "happen to know" how to process them. The weird part still is that the "status code" happens to correspond with HTTP status codes, so yeah, this is not a great abstraction either.

I don't know what's a good alternative here now. Designing an exception hierarchy is surprisingly hard, and it almost never feels perfect. And even when we have one, then it's really hard to have all devs use it correctly. I guess, ideally, anywhere deeper than in the endpoint (or maybe controller) classes, we should use just some generic exceptions, not REST-aware or anything, but that feels quite hard to do — we'd have to either catch them in endpoints and convert to error responses or they'd still need some kind of conversion to an HTTP status.

Do you have any ideas on how to improve this? And maybe, in general, the usage of exceptions in the code? I guess we don't have to solve it here and now, but it surely is one of those things that when we postpone it, we'd never do it, so maybe we could create a ticket or a P2 post to discuss how we could improve this.

mailpoet/lib/Automation/Engine/Validators/WorkflowSchema.php
JanJakes: Ehh, nope, just a typo or something, thanks for the catch. Fix: https://github.com/mailpoet/mailpoet/pull/4359/commits/69dce668736f69cc0f858153fe8c95402d809fa9

mailpoet/lib/Automation/Engine/Control/StepHandler.php
JanJakes: True! Fixed https://github.com/mailpoet/mailpoet/pull/4359/commits/a70b2606e0463890635a1c776f5f7c65aec72f75, thanks.

mailpoet/lib/Automation/Engine/Exceptions.php
JanJakes: :+1: But it's true that some of these comments are probably not necessary as it seems to be clear enough. Fix: https://github.com/mailpoet/mailpoet/pull/4359/commits/57fcd06bb3a5520ace5e1d93124974a151e0b894

mailpoet/lib/Automation/Engine/Exceptions/Exception.php
johnolek: Thanks for your thoughts on this! I'm sorry to say I don't have any great ideas for different approaches. I'll use this as an excuse:

> Designing an exception hierarchy is surprisingly hard, and it almost never feels perfect.

I think we should still create a backlog ticket for this if for no other reason than to already have one the next time this topic comes up, but it doesn't seem like a very high priority maintenance item even, especially since these exceptions are currently working just as we need them to. 

If you think it would be worth it at this point to collect other people's feedback and think a P2 would be fruitful, however, please don't let me stop you!

mailpoet/assets/js/src/automation/editor/store/actions.ts
johnolek: That makes perfect sense, thanks for the explanation!

johnolek (approved):

PR #4361, opened 2022-09-12
Improvements to Send Email step UI [MAILPOET-4622]

johnolek:
## Description

This PR makes some UI tweaks to the Send email Step sidebar when editing a workflow.

- Ensures the HelpScout icon doesn't cover any of the content in the lower right hand part of the automation sidebar (this should actually apply to any sidebar content, but right now Send email is probably the only one that generated enough content for this to be an issue for most browser sizes).
- Removes the email preview thumbnail
- Renames the header "Email content" to just "Email"
- Adds help text beneath Subject and Preheader fields informing users they can use MailPoet shortcodes inside those fields
    - If the user has HelpScout integration turned on, the link should open directly in the HelpScout widget thing.

## Code review notes

_N/A_

## QA notes

QA is not necessary since this only affects automations.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

### DIFF
--- a/mailpoet/assets/css/src/components-automation-editor/_panel.scss
+++ b/mailpoet/assets/css/src/components-automation-editor/_panel.scss
@@ -1,3 +1,7 @@
+.mailpoet-automation-sidebar {
+  padding-bottom: 100px;
+}
+
 .components-panel__body-title.mailpoet-automation-panel-plain-body-title {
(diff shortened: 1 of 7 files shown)


JanJakes (approved):
LGTM!

PR #4225, opened 2022-07-05
Bundle automation editor styles for now [MAILPOET-4451]

JanJakes:
We need to use older @wordpress/edit-site package (version 4.1.0)
to be compatible with the current @wordpress/edit-post (version 6.1.1).

[MAILPOET-4451]

[MAILPOET-4451]: https://mailpoet.atlassian.net/browse/MAILPOET-4451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4362, opened 2022-09-14
Automation template listing [MAILPOET-4533]

websupporter:
## Description

This PR adds the template listing page. You can reach the page by clicking the "New automation" button.

## Linked PRs
https://github.com/mailpoet/mailpoet-premium/pull/622

## Linked tickets

[MAILPOET-4533]

[MAILPOET-4533]: https://mailpoet.atlassian.net/browse/MAILPOET-4533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/mailpoet-automation-editor.scss
+++ b/mailpoet/assets/css/src/mailpoet-automation-editor.scss
@@ -1,4 +1,3 @@
-@import '../../../node_modules/@wordpress/components/build-style/style';
 @import '../../../node_modules/@wordpress/interface/build-style/style';
 @import '../../../node_modules/@wordpress/edit-site/build-style/style';
 @import '../../../node_modules/@wordpress/block-editor/build-style/style'; // for inserter styles
--- a/mailpoet/assets/css/src/mailpoet-automation-templates.scss
+++ b/mailpoet/assets/css/src/mailpoet-automation-templates.scss
@@ -23,6 +23,7 @@ ul.mailpoet-templates {
     height: 100%;
     padding: 24px 24px 26px;
     text-align: left;
+    width: 100%;
 
     &:hover {
       background: #fff;
@@ -51,5 +52,23 @@ ul.mailpoet-templates {
 
   p {
     margin: 8px 0 0;
+  }
+
+  .mailpoet-from-scratch {
+    button {
+      align-content: center;
+      border: 2px dashed #dcdcde;
+      display: grid;
+      grid-gap: 24px;
+      justify-items: center;
+
+      &:hover {
+        border: 2px dashed #dcdcde;
+      }
+    }
+
+    svg {
+      fill: #dcdcde;
+    }
   }
 }
--- a/mailpoet/assets/css/src/mailpoet-automation-templates.scss
+++ b/mailpoet/assets/css/src/mailpoet-automation-templates.scss
@@ -3,7 +3,7 @@
   max-width: 982px;
 }
 
-ul.mailpoet-templates {
+ul.mailpoet-automation-templates {
   clear: both;
   cursor: pointer;
   display: grid;
@@ -54,7 +54,7 @@ ul.mailpoet-templates {
     margin: 8px 0 0;
   }
 
-  .mailpoet-from-scratch {
+  .mailpoet-automation-from-scratch {
     button {
       align-content: center;
       border: 2px dashed #dcdcde;
--- a/mailpoet/assets/css/src/mailpoet-automation-templates.scss
+++ b/mailpoet/assets/css/src/mailpoet-automation-templates.scss
@@ -1,0 +1,55 @@
+.mailpoet-automation-templates-heading {
+  margin: auto;
+  max-width: 982px;
+}
+
+ul.mailpoet-templates {
+  clear: both;
+  cursor: pointer;
+  display: grid;
+  grid-gap: 32px;
+  grid-template-columns: repeat(auto-fill, 306px);
+  justify-content: center;
+  margin: auto;
+  max-width: 982px;
+  padding: 48px 0;
+
+  button.components-button {
+    background: #fff;
+    border: 1px solid #dcdcde;
+    border-radius: 4px;
+    cursor: pointer;
+    display: block;
+    height: 100%;
+    padding: 24px 24px 26px;
+    text-align: left;
+
+    &:hover {
+      background: #fff;
+      border: 1px solid #dcdcde;
+      box-shadow: 0 3px 6px rgba(0, 0, 0, .15);
+      color: inherit;
+    }
+
+    &:focus {
+      border-color: #2271b1;
+      box-shadow: 0 3px 6px rgba(0, 0, 0, .15);
+      color: inherit;
+    }
+  }
+
+  h2 {
+    background: transparent;
+    border: none;
+    color: #2271b1;
+    font-size: 14px;
+    font-weight: 600;
+    line-height: 21px;
+    margin: 0;
+    padding: 0;
+  }
+
+  p {
+    margin: 8px 0 0;
+  }
+}
--- a/mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
+++ b/mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
@@ -58,7 +58,7 @@ export function FromScratchButton(): JSX.Element {
           <p>{error}</p>
         </Notice>
       )}
-      <Button variant="primary" onClick={() => onClickScratchButton()}>
+      <Button variant="secondary" onClick={() => onClickScratchButton()}>
         {__('From scratch', 'mailpoet')}
       </Button>
       <FromScratchPremiumModal
--- a/mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
+++ b/mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
@@ -1,0 +1,97 @@
+import { Dispatch, SetStateAction, useCallback, useState } from 'react';
+import { __ } from '@wordpress/i18n';
+import { Button } from '@wordpress/components';
+import { Hooks } from 'wp-js-hooks';
+import { Icon, plusCircleFilled } from '@wordpress/icons';
+import { PremiumModal } from '../../../common/premium_modal';
+import { Notice } from '../../../notices/notice';
+
+type FromScratchPremiumModalProps = {
+  showModal: boolean;
+  setShowModal: Dispatch<SetStateAction<boolean>>;
+};
+
+function FromScratchPremiumModal({
+  showModal,
+  setShowModal,
+}: FromScratchPremiumModalProps): JSX.Element | null {
+  if (!showModal) {
+    return null;
+  }
+  return (
+    <PremiumModal
+      onRequestClose={() => {
+        setShowModal(false);
+      }}
+      tracking={{
+        utm_medium: 'upsell_modal',
+        utm_campaign: 'create_automation_from_scratch',
+      }}
+    >
+      {__('You cannot create automation from scratch.', 'mailpoet')}
+    </PremiumModal>
+  );
+}
+
+function fromScratchHook(callback: () => void, errorHandler: Dispatch<string>) {
+  const fromScratchCallback = Hooks.applyFilters(
+    'mailpoet.automation.templates.from_scratch_button',
+    () => {
+      callback();
+    },
+  );
+  fromScratchCallback(errorHandler);
+}
+
+export function FromScratchButton(): JSX.Element {
+  const [showModal, setShowModal] = useState(false);
+  const [error, errorHandler] = useState(null);
+  const onClickScratchButton = useCallback(() => {
+    fromScratchHook(() => {
+      setShowModal(true);
+    }, errorHandler);
+  }, []);
+  return (
+    <>
+      {error && (
+        <Notice type="error" closable timeout={false}>
+          <p>{error}</p>
+        </Notice>
+      )}
+      <Button variant="primary" onClick={() => onClickScratchButton()}>
+        {__('From scratch', 'mailpoet')}
+      </Button>
+      <FromScratchPremiumModal
+        showModal={showModal}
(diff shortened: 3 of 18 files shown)


mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsCreateFromTemplateEndpoint.php
```diff
@@ -19,8 +19,8 @@ public function __construct(
   }
 
   public function handle(Request $request): Response {
-    $this->createWorkflowFromTemplateController->createWorkflow((string)$request->getParam('slug'));
-    return new Response();
+    $workflow = $this->createWorkflowFromTemplateController->createWorkflow((string)$request->getParam('slug'));
+    return new Response($workflow->toArray());
```
JanJakes: Just a note, the `->toArray()` method is meant to be used for storing in the DB. This probably happens to be the same or almost the same at the moment, but the separation will likely be needed to maintain back compatibility of the API and handle differences between DB and API formats.

Currently, we have this quite badly done in the classes themselves (endpoints, pages), which is not reusable. I'd like to extract it to some `Mappers` namespace, but we don't have to to that now. At the moment, we can either keep this (and not forget to change it) or copy paste [the method](https://github.com/mailpoet/mailpoet/blob/7c3d3fbf12a1b3329ed76a17fcf8cd0f5c81bb6f/mailpoet/lib/AdminPages/Pages/AutomationEditor.php#L82) that the editor page is using.

mailpoet/lib/AdminPages/Pages/AutomationTemplates.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace MailPoet\AdminPages\Pages;
+
+use MailPoet\AdminPages\PageRenderer;
+use MailPoet\Automation\Engine\Data\WorkflowTemplate;
+use MailPoet\Automation\Engine\Storage\WorkflowTemplateStorage;
+use MailPoet\WP\Functions as WPFunctions;
+
+class AutomationTemplates {
```
JanJakes, line 10: Seeing this being a new admin page that requires a full page load, I'm kind of thinking... What if we included it in the automation listing React app and used React Router only to make the interaction faster? I know this may require some loading state or a spinner when the button is clicked but it will likely be more "slick" (and later on with preloading, instantaneous). Or would that cause any issues? Thinking about it, it's also a change that we can do later...

mailpoet/lib/Config/Menu.php
```diff
@@ -428,6 +429,16 @@ private function registerMailPoetMenu() {
...
+      // Automation templates
+      $this->wp->addSubmenuPage(
+        true,
+        $this->setPageTitle('Automation Templates'),
+        'Automation Templates',
+        AccessControl::PERMISSION_MANAGE_AUTOMATIONS,
+        'mailpoet-automation-templates',
+        [$this, 'automationTemplates']
+      );
+
```
JanJakes, line 441: When I'm on `page=mailpoet-automation-templates`, I don't see MailPoet menu expanded and the "Automations" menu highlighted. We seem to have some solution for this that is used [here](https://github.com/mailpoet/mailpoet/blob/9b30ed0d8b082e827fcc5ffea90ef0611832a49f/mailpoet/lib/AdminPages/Pages/AutomationEditor.php#L62), for instance.

mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { Icon, plusCircleFilled } from '@wordpress/icons';
+import { PremiumModal } from '../../../common/premium_modal';
+import { Notice } from '../../../notices/notice';
+
+type FromScratchPremiumModalProps = {
+  showModal: boolean;
+  setShowModal: Dispatch<SetStateAction<boolean>>;
+};
+
+function FromScratchPremiumModal({
```
JanJakes, line 14: For some reason, the modal renders differently than in other places.
Here:

Screen Shot 2022-09-14 at 10 27 40

Automation editor:

Screen Shot 2022-09-14 at 10 27 31

Is that some kind of an issue with how the modal handles styles, maybe? We need to be able to display this modal in the same way anywhere in the MailPoet admin.

mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const fromScratchCallback = Hooks.applyFilters(
+    'mailpoet.automation.templates.from_scratch_button',
+    () => {
+      callback();
+    },
+  );
+  fromScratchCallback(errorHandler);
+}
+
+export function FromScratchButton(): JSX.Element {
```
JanJakes, line 46: This seems to render as a primary button:

Screen Shot 2022-09-14 at 10 27 51

In the specs there seems to be secondary one (or how that variant is called).

mailpoet/assets/js/src/automation/templates/components/template-list-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </p>
+      </Notice>
+    );
+  }
+
+  return (
+    <li className="mailpoet-template-list-item">
+      {notice}
+      <Button
+        isBusy={loading}
```
JanJakes, line 48: I think there is no visible loading state currently, and the other buttons (from scratch item and button) don't have this. Not sure if this at least blocks multiple clicks but universally, it's kind of hard to prevent users from creating multiple workflows at this stage. What if we added a full-page overlay with a spinner (e.g. the [loader component](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/common/loader/loader.tsx), for now)? Not sure though, as this was not specified.

mailpoet/assets/css/src/mailpoet-automation-templates.scss
```diff
@@ -0,0 +1,74 @@
+.mailpoet-automation-templates-heading {
+  margin: auto;
+  max-width: 982px;
+}
+
+ul.mailpoet-templates {
```
JanJakes: I guess we should include the `automation` in the prefix, and maybe the same below for `mailpoet-from-scratch`.

mailpoet/lib/Automation/Engine/Builder/CreateWorkflowFromTemplateController.php
```diff
@@ -29,7 +29,11 @@ public function createWorkflow(string $slug): Workflow {
       throw UnexpectedValueException::create()->withMessage('Template not found.');
     }
 
-    $this->storage->createWorkflow($template->getWorkflow());
-    return $template->getWorkflow();
+    $workflowId = $this->storage->createWorkflow($template->getWorkflow());
+    $workflow = $this->storage->getWorkflow($workflowId);
+    if (!$workflow) {
+      throw UnexpectedValueException::create()->withMessage('Workflow not found.');
```
JanJakes: I think not having a workflow just after creating it should be an `InvalidStateException` from `MailPoet\Automation\Engine\Exceptions` rather than a 4xx `UnexpectedValueException`.

mailpoet/assets/js/src/automation/templates/components/template-list-item.tsx
JanJakes: Nevermind, this will be solved: https://a8c.slack.com/archives/C040TSG288Y/p1663149694514879?thread_ts=1663148422.788019&cid=C040TSG288Y

websupporter: As discussed, there will be a modal opening instead of a straight redirect, so I think it's ok to leave it for now.

mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsCreateFromTemplateEndpoint.php
websupporter: In [e2f2c05](https://github.com/mailpoet/mailpoet/pull/4362/commits/e2f2c056323dcf06ad502592c21235ab4065a98f) I opt for the copy paste method for now.

mailpoet/lib/AdminPages/Pages/AutomationTemplates.php
websupporter: > Or would that cause any issues?

Nothing comes to my mind tbh.

> Thinking about it, it's also a change that we can do later

I would like to defer this change, just in terms of keep moving.

mailpoet/lib/Config/Menu.php
websupporter: Good catch! Thank you! Addressed in [ada0ed3](https://github.com/mailpoet/mailpoet/pull/4362/commits/ada0ed34d763533e80f5fb3cf58e0be0dd9d0b5e)

mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
websupporter: The automation listing as well as the templates page did enqueue wp-component styles. I removed this in [d7140cf](https://github.com/mailpoet/mailpoet/pull/4362/commits/d7140cff91cb6251986e53e0a63400c677b78232)

mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
websupporter: Thank you. Addressed in [f6615ea](https://github.com/mailpoet/mailpoet/pull/4362/commits/f6615ea51537260e9a34adfe24e7f7e409f249da)

mailpoet/assets/css/src/mailpoet-automation-templates.scss
websupporter: Addressed in [2f42d2c](https://github.com/mailpoet/mailpoet/pull/4362/commits/2f42d2ceb69dd364672594f6c4d94e0bb6b47800)

mailpoet/lib/Automation/Engine/Builder/CreateWorkflowFromTemplateController.php
websupporter: Thank you! Fixed in [e40c459](https://github.com/mailpoet/mailpoet/pull/4362/commits/e40c4590446856712f5ec71d12cbed8050680939)

mailpoet/lib/Automation/Engine/Builder/CreateWorkflowFromTemplateController.php
```diff
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Automation\Engine\Builder;
 
+use _PHPStan_ae8980142\Nette\InvalidStateException;
```
JanJakes: Ups 😄 Your IDE did this, I guess.

mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
JanJakes: Ahh! And the removal is safe? Because we do use `@wordpress/components` so we need to load their styles somehow. In the editor, [we bundle them now](https://github.com/mailpoet/mailpoet/pull/4225/files) so I guess we need to bundle them also anywhere we use them and also within the modal?

It feels like removing these lines could break some existing styles, if `@wordpress/components` are used...

mailpoet/lib/Automation/Engine/Builder/CreateWorkflowFromTemplateController.php
websupporter: :facepalm: Addressed in [17e2f97](https://github.com/mailpoet/mailpoet/pull/4362/commits/17e2f979db39561b3ba63b702b08ce7b4c476540)

mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
websupporter: > In the editor, we bundle them now

We do so also in mailpoet-plugin.scss, which is loaded throughout the screens: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/css/src/mailpoet-plugin.scss#L24

I removed them now in 9fe003aa3 from the mailpoet-automation-editor.scss as they are not needed there.

What we witness is an update of the styles ([PR](https://github.com/WordPress/gutenberg/pull/40781)), This update affects Gutenberg version>13.7 while in [WordPress core the current Gutenberg version is 13.0](https://developer.wordpress.org/block-editor/contributors/versions-in-wordpress/).

So, with our current way, we divert from core and "enqueue" the latest styles (or the styles of the components version, we require). Not entirely sure, if that's the best way (thinking of 3rd parties potentially enqueuing wp-components)

mailpoet/assets/css/src/mailpoet-automation-editor.scss
```diff
@@ -1,4 +1,3 @@
-@import '../../../node_modules/@wordpress/components/build-style/style';
```
websupporter, line 1: See my comment here: https://github.com/mailpoet/mailpoet/pull/4362/files/788859ea6961b83c056bb2dfc53f6c9ed52eea76#r971524603

mailpoet/assets/js/src/automation/templates/components/from-scratch.tsx
JanJakes: Ahh, thanks for the explanation! Yeah, I guess at this point it's correct to "divert" in the styles because we're diverting also with the JS (bundling them instead of using [@wordpress/dependency-extraction-webpack-plugin](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-dependency-extraction-webpack-plugin/)) which we may fix in the future. It would require supporting only few recent versions of WP, and testing the UI across them. For now, we picked the "easier" way and did what's done in the form editor.

JanJakes (approved):

PR #4363, opened 2022-09-14
Feat/allow all emails from verified domains to be used for sending - MAILPOET-4601

triple0t:
## Description

In this PR, we made an update to our sending process.

Now, plugin users can send a Newsletter campaign from any email address of their verified domain.

## QA notes

Veljko to Alex notes:
2 domains are already authorized for testing on our test sites: `qawp.net` and `moonbridge.rs`
If you test positive scenario in this pr, you should use any email associated with above 2 domains, no errors/problems with sending etc.
If you test negative scenario in this pr, you should use any other email address not associated with above 2 domains. You should see warnings/errors in the settings page if you try to apply it for Default Sender's From field and/or on the last page when creating and sending newsletter (there's also From field there).

## Linked tickets

[MAILPOET-4601](https://mailpoet.atlassian.net/browse/MAILPOET-4601)

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Services.php
+++ b/mailpoet/lib/API/JSON/v1/Services.php
@@ -238,25 +238,27 @@ class Services extends APIEndpoint {
       return $this->createBadRequest(__('MailPoet Sending Service is not active.', 'mailpoet'));
     }
 
-    $authorizedEmails = $this->bridge->getAuthorizedEmailAddresses();
-
-    $verifiedDomains = $this->senderDomainController->getVerifiedSenderDomains(true);
-
-    if (!$authorizedEmails && !$verifiedDomains) {
-      return $this->createBadRequest(__('No FROM email addresses are authorized.', 'mailpoet'));
-    }
-
     $fromEmail = $this->settings->get('sender.address');
     if (!$fromEmail) {
       return $this->createBadRequest(__('Sender email address is not set.', 'mailpoet'));
     }
 
-    $arrayOfItems = explode('@', $fromEmail);
-    $emailDomain = array_pop($arrayOfItems);
+    $verifiedDomains = $this->senderDomainController->getVerifiedSenderDomainsIgnoringCache();
 
-    if (!$this->isItemInArray($fromEmail, $authorizedEmails) && !$this->isItemInArray($emailDomain, $verifiedDomains)) {
-      // translators: %s is the email address, which is not authorized.
-      return $this->createBadRequest(sprintf(__("Sender email address '%s' is not authorized.", 'mailpoet'), $fromEmail));
+    $arrayOfItems = explode('@', trim($fromEmail));
+    $emailDomain = strtolower(array_pop($arrayOfItems));
+
+    if (!$this->isItemInArray($emailDomain, $verifiedDomains)) {
+      $authorizedEmails = $this->bridge->getAuthorizedEmailAddresses();
+
+      if (!$authorizedEmails) {
+        return $this->createBadRequest(__('No FROM email addresses are authorized.', 'mailpoet'));
+      }
+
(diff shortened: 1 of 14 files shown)


mailpoet/lib/Services/AuthorizedEmailsController.php
```diff
@@ -191,4 +212,11 @@ private function validateAuthorizedEmail($authorizedEmails = [], $email = '') {
     $lowercaseAuthorizedEmails = array_map('strtolower', $authorizedEmails);
     return in_array(strtolower($email), $lowercaseAuthorizedEmails, true);
   }
+
+  private function validateEmailDomainIsVerified($verifiedDomains = [], $email = ''): bool {
```
costasovo: Just a detail, but could you please add type hints? We want to use type hints in the new code.

`validateEmailDomainIsVerified(array $verifiedDomains = [], string $email = '')`

mailpoet/lib/Services/AuthorizedSenderDomainController.php
```diff
@@ -46,14 +46,20 @@ public function getDomainRecords(string $domain = ''): array {
    *
    * Note: This includes both verified and unverified domains
    */
-  public function getAllSenderDomains(): array {
+  public function getAllSenderDomains(bool $skipCache = false): array {
```
costasovo: I find this type of boolean parameter not very readable. I would prefer creating a new public method that would say what it does in its name.

`$controller->getAllSenderDomains(true)`
vs
`$controller->getAllSenderDomainsIgnoringCache()`

What do you think?

mailpoet/lib/API/JSON/v1/Services.php
```diff
@@ -233,15 +239,22 @@ public function sendCongratulatoryMssEmail() {
     }
 
     $authorizedEmails = $this->bridge->getAuthorizedEmailAddresses();
-    if (!$authorizedEmails) {
+
+    $verifiedDomains = $this->senderDomainController->getVerifiedSenderDomains(true);
+
+    if (!$authorizedEmails && !$verifiedDomains) {
       return $this->createBadRequest(__('No FROM email addresses are authorized.', 'mailpoet'));
     }
```
costasovo: I think we could skip this `'No FROM email addresses are authorized.'` check. Because of the condition `!$authorizedEmails && !$verifiedDomains`, we need to always make two API requests. I think it would be nicer to take the same approach as in JS and check domains first, and it there are no domains, then check addresses or the other way around. This could save us a request in some casese. What do you think?

costasovo (changes requested):
Thanks for introducing this functionality! 👍 
I added a couple of remarks. Just details.

mailpoet/lib/Services/AuthorizedEmailsController.php
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/4363/commits/236b75eddf091510f9beaf12e3923629cced2ace

mailpoet/lib/Services/AuthorizedSenderDomainController.php
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/4363/commits/236b75eddf091510f9beaf12e3923629cced2ace

mailpoet/lib/API/JSON/v1/Services.php
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/4363/commits/236b75eddf091510f9beaf12e3923629cced2ace

costasovo (approved):
Thank you! 👍

Aschepikov (approved):
Good job!

PR #4366, opened 2022-09-15
Refactor MailPoet\Config\Populator class to doctrine [MAILPOET-4306]

rodrigoprimo:
## Description

This PR refactors the MailPoet\Config\Populator class to use Doctrine instead of Paris.

## Code review notes

To test the changes while I was coding I created a very simple PopulatorTest function and used Xdebug to check that the code in the Populator class was still working as expected. Sharing it here in case it helps:

```
  public function testBasic() {
    $this->populator->up();
  }
```

I'm unsure about the change to \MailPoet\Segments\SegmentsRepository::getWPUsersSegment(). I don't like this pattern of creating a segment inside a method that seems to be a getter judging by its name, but at the same, this is the pattern that we are already using for \MailPoet\Segments\SegmentsRepository::getWooCommerceSegment() and the equivalent methods in the Segment model class. Any input is appreciated.

## QA notes

It is not very easy to test most of the changes in this PR. The Populator class is used to populate the database when MailPoet is installed or updated. Some of the changes will be executed whenever MP is installed and are easier to test. Other changes are executed only when updating from a specific version and are harder to test.

The only one that I believe is easy to test is to check that the default, WooCommerce and WordPress segments are created when MP is installed.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4306]

## After-merge notes

_N/A_

[MAILPOET-4306]: https://mailpoet.atlassian.net/browse/MAILPOET-4306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Populator.php
+++ b/mailpoet/lib/Config/Populator.php
@@ -16,15 +16,16 @@ use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\NewsletterLinkEntity;
 use MailPoet\Entities\NewsletterTemplateEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\StatisticsFormEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Entities\UserFlagEntity;
 use MailPoet\Form\FormsRepository;
 use MailPoet\Mailer\MailerLog;
-use MailPoet\Models\Segment;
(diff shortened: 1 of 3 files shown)


lysyjan (approved):
> I don't like this pattern of creating a segment inside a method that seems to be a getter judging by its name

@rodrigoprimo I get your point. This approach is not very clean, but we copied it from the old model.
We can improve it as a maintenance ticket if you want.

The changes look good to me 👍

Aschepikov (approved):

Aschepikov (approved):
Good job!

PR #4367, opened 2022-09-16
Remove old model usage from WooCommercePastRevenues [MAILPOET-4377]

rodrigoprimo:
## Description

This PR removes all old models from WooCommercePastRevenues and replaces them with Doctrine code.

## Code review notes

_N/A_

## QA notes

I'm not sure if there is much for QA to do in this PR. It changes code that is used for the CLI command `./do gen:data past_revenues`

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4377]

## After-merge notes

_N/A_

[MAILPOET-4377]: https://mailpoet.atlassian.net/browse/MAILPOET-4377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
+++ b/mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
@@ -7,11 +7,11 @@ use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\NewsletterLinkEntity;
 use MailPoet\Entities\NewsletterSegmentEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Entities\ScheduledTaskSubscriberEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\StatisticsClickEntity;
 use MailPoet\Entities\StatisticsOpenEntity;
 use MailPoet\Entities\SubscriberEntity;
-use MailPoet\Models\ScheduledTaskSubscriber;
 use MailPoet\Models\SendingQueue;
 use MailPoet\Models\StatisticsNewsletters;
 use MailPoet\Models\Subscriber;
@@ -299,7 +299,7 @@ class WooCommercePastRevenues implements Generator {
     $connection->executeStatement("ALTER TABLE `" . SubscriberSegment::$_table . "` DISABLE KEYS");
     $connection->executeStatement("ALTER TABLE `" . $this->entityManager->getClassMetadata(NewsletterLinkEntity::class)->getTableName() . "` DISABLE KEYS");
     $connection->executeStatement("ALTER TABLE `" . $this->entityManager->getClassMetadata(ScheduledTaskEntity::class)->getTableName() . "` DISABLE KEYS");
-    $connection->executeStatement("ALTER TABLE `" . ScheduledTaskSubscriber::$_table . "` DISABLE KEYS");
+    $connection->executeStatement("ALTER TABLE `" . $this->entityManager->getClassMetadata(ScheduledTaskSubscriberEntity::class)->getTableName() . "` DISABLE KEYS");
     $connection->executeStatement("ALTER TABLE `" . SendingQueue::$_table . "` DISABLE KEYS");
     $connection->executeStatement("ALTER TABLE `" . $this->entityManager->getClassMetadata(StatisticsOpenEntity::class)->getTableName() . "` DISABLE KEYS");
     $connection->executeStatement("ALTER TABLE `" . $this->entityManager->getClassMetadata(StatisticsClickEntity::class)->getTableName() . "` DISABLE KEYS");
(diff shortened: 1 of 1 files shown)


mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
```diff
@@ -268,59 +285,66 @@ private function getBatchLog($dataType, $generatedCount) {
...
-    $subscriber = Subscriber::createOrUpdate([
-      'email' => $email,
-      'status' => $status,
-      'last_name' => $lastName,
-      'created_at' => $createdAtDate,
-    ]);
-    $subscriber->save();
-    $segment->addSubscriber($subscriber->id);
+  private function createSubscriber(string $email, string $lastName, string $createdAtDate, SegmentEntity $segment, $status = SubscriberEntity::STATUS_SUBSCRIBED): SubscriberEntity {
+    $subscriber = new SubscriberEntity();
```
samnajian: `MailPoet\Test\DataFactories\Subscriber` can be used to make the creation easier

mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
```diff
@@ -343,55 +367,57 @@ private function createProduct($name, $price, $categoryIds = [], $discount = 0)
...
+
     // Sending task
-    $task = ScheduledTask::createOrUpdate([
-      'type' => Sending::TASK_TYPE,
-      'status' => ScheduledTask::STATUS_COMPLETED,
-      'created_at' => $sentAt,
-      'processed_at' => $sentAt,
-    ]);
-    $task->save();
+    $task = new ScheduledTaskEntity();
```
samnajian, line 375: `MailPoet\Test\DataFactories\ScheduledTask` can facilitate creation

mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
```diff
@@ -343,55 +367,57 @@ private function createProduct($name, $price, $categoryIds = [], $discount = 0)
...
 
     // Sending queue
-    $queue = SendingQueue::createOrUpdate([
-      'task_id' => $task->id,
-      'newsletter_id' => $newsletter->getId(),
-      'count_total' => count($subscribersIds),
-      'count_processed' => count($subscribersIds),
-    ]);
-    $queue->save();
+    $queue = new SendingQueueEntity();
```
samnajian, line 384: `MailPoet\Test\DataFactories\SendingQueue` can facilitate here

mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
```diff
@@ -268,59 +285,66 @@ private function getBatchLog($dataType, $generatedCount) {
...
-    ORM::rawExecute("ALTER TABLE `" . Subscriber::$_table . "` ENABLE KEYS");
-    ORM::rawExecute("ALTER TABLE `" . SubscriberSegment::$_table . "` ENABLE KEYS");
-    ORM::rawExecute("ALTER TABLE `" . $this->entityManager->getClassMetadata(NewsletterLinkEntity::class)->getTableName() . "` ENABLE KEYS");
-    ORM::rawExecute("ALTER TABLE `" . ScheduledTask::$_table . "` ENABLE KEYS");
-    ORM::rawExecute("ALTER TABLE `" . ScheduledTaskSubscriber::$_table . "` ENABLE KEYS");
-    ORM::rawExecute("ALTER TABLE `" . SendingQueue::$_table . "` ENABLE KEYS");
-    ORM::rawExecute("ALTER TABLE `" . $this->entityManager->getClassMetadata(StatisticsOpenEntity::class)->getTableName() . "` ENABLE KEYS");
-    ORM::rawExecute("ALTER TABLE `" . $this->entityManager->getClassMetadata(StatisticsClickEntity::class)->getTableName() . "` ENABLE KEYS");
-    ORM::rawExecute("SET UNIQUE_CHECKS = 1;");
+    $connection->executeStatement("ALTER TABLE `{$prefix}posts` ENABLE KEYS");
```
samnajian, line 324: Despite these being old code, I'd suggest moving them to a different method with a meaningful name so that it's easier to know what these similar lines of code do.

samnajian (changes requested):

mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
rodrigoprimo: I agree with the general sentiment here but I got stuck in choosing a name and thus I will suggest, if you agree, that we leave this for the future. The reason why I got stuck is that I'm not even sure if running `DISABLE KEYS` and `ENABLE KEYS` is needed. My understanding is that it only has any effect when MyISAM is being used and I would assume that the majority of the WP installs where this can be executed (mostly dev environments) are using InnoDB. I don't think it is worth investing more time in this in the context of this PR.

mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
rodrigoprimo: Please correct me if I'm wrong but I don't think there is a way to set `createdAt` and `processed` using the ScheduledTask data factory. We could add it, but I would suggest we leave this for another PR as the current approach should work fine.

mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
rodrigoprimo: Similar to the above, I don't think there is a way to use the SendingQueue data factory to set `countTotal` and `countProcessed`.

mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
rodrigoprimo: I don't see any difference in this particular case in terms of making the creation easier, but at the same time, I'm not against it. I implemented this change in a new commit.

rodrigoprimo:
Thanks for the review, @samnajian. I believe I address all your comments. Could you please take a second look?

rodrigoprimo:
Thanks for your review, @samnajian. I believe I replied to all your comments. Could you please take a second look?

mailpoet/tests/DataGenerator/Generators/WooCommercePastRevenues.php
samnajian: I'm fine with leaving it to another PR, I only suggested doing it here since thought adding the missing part to `MailPoet\Test\DataFactories\ScheduledTask` are not very time intensive, and I leave this decision to you.

samnajian (approved):
LGTM!

veljkho (approved):

veljkho (approved):

PR #4368, opened 2022-09-16
COT Run WooCommerce tests with different WC/COT configurations [MAILPOET-4572]

costasovo:
## Description
This PR enables to running of integration and acceptance tests with different versions of the configuration of the COT feature.

### Downloading WooCommerce COT testing build

I added a Robo command for downloading the COT testing version of the WooCommerce plugin. This is a temporary command that currently downloads [COT Release testing build](https://github.com/woocommerce/woocommerce/releases/tag/feature-custom-order-table).
```
./do download:woo-commerce-cot-zip
```
**Note:** After downloading a new version of any plugin for tests, you need to destroy the test containers (`./do reset:test-docker`) so that the newly downloaded version is installed as we don't reinstall plugins every time for performance reasons.

### Running tests locally
I added two new flags for configuring the COT feature. Both can be used for integration and acceptance tests

- `--enable-cot`: When tests run with this flag, COT are created, and the shop is configured to use them. 
- `--enable-cot-sync`: When tests run with this flag, it enables sync ty from COT to the wp_posts. Note: without `--enable-cot` the flag is ignored.

e.g `./do test:integration --skip-deps --group woo --enable-cot --enable-cot-sync` runs Woo related integration tests with COT enabled and sync enabled.

I also added support for `--group` to the acceptance test and marked tests related to WooCommerce with this group.

### New jobs on Circle CI
I added three new acceptance test jobs and three new integration test jobs that use the WooCommerce COT release testing build and run in different configurations. These new jobs are set to run only on branches prefixed `cot-` and on `trunk`. Some of those jobs are also set to ignore failures (but you can still see what tests failed). This was done because we need to implement support for COT, and we don't want to have red builds. 
I created a follow-up ticket for enabling failure checks and allowing the jobs on all branches (MAILPOET-4648) after we fix all issues.

#### New jobs:
- acceptance_tests_woo_cot_sync // ingoring failures
- acceptance_tests_woo_cot_no_sync // ingoring failures
- acceptance_tests_woo_cot_off
- integration_test_woo_cot_sync // ingoring failures
- integration_test_woo_cot_no_sync // ingoring failures
- integration_test_woo_cot_off

I hope that from the names, it is clear what configurations are used.

## Code review notes
I ran into a problem when I was adding group woo on acceptance tests on Circle CI. Test splitting for parallel execution on Circle CI uses the group parameter. So I had to use a hack when running the `woo` group in acceptance tests. I search all test files that contain `/* @group woo` and pass those to the test split command. So when the group is used only for one test case on Circle CI, it will run all tests from a file. This is a bit suboptimal, but it seems to do the job.

## QA notes

All changes are related to tests and Circle CI. There is nothing to test in the plugin.

## Linked tickets

[MAILPOET-4572]

## After-merge notes

Ping developers on slack mailpoet-dev-code on slack and ask them to start adding acceptance tests related to WooCommerce to group  `woo`.

[MAILPOET-4572]: https://mailpoet.atlassian.net/browse/MAILPOET-4572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -713,6 +713,29 @@ workflows:
             - qa_php
       - integration_tests:
           <<: *slack-fail-post-step
+          group: woo
+          enable_cot: 1
+          enable_cot_sync: 0
+          allow_fail: 1
+          woo_core_version: woo-cot-beta # Temporarily force COT beta version
+          name: integration_test_woo_cot_no_sync
+          requires:
+            - unit_tests
+            - static_analysis_php8
+            - qa_js
+            - qa_php
+      - integration_tests:
+          <<: *slack-fail-post-step
+          group: woo
+          woo_core_version: woo-cot-beta # Temporarily force COT beta version
+          name: integration_test_woo_cot_off
+          requires:
+            - unit_tests
+            - static_analysis_php8
+            - qa_js
+            - qa_php
+      - integration_tests:
+          <<: *slack-fail-post-step
           skip_group: woo
           skip_plugins: 1
           name: integration_test_base
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -405,7 +405,12 @@ jobs:
             # Convert test result filename values to be relative paths because the circleci CLI's split command requires exact matches
             sed -i.bak 's#/wp-core/wp-content/plugins/mailpoet/##g' $CIRCLE_INTERNAL_TASK_DATA/circle-test-results/results.json
             # `circleci tests split` returns different values based on the container it's run on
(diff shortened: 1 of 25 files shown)


rodrigoprimo (approved):
Looks good to me, @costasovo. Great job! Merging directly as there is nothing for QA to check in this PR.

rodrigoprimo:
@costasovo, and just a reminder about the post-merge notes in case you are not already thinking about that.

PR #4371, opened 2022-09-19
Add and extend triggers [MAILPOET-4587]

websupporter:
## Description
In this PR the settings for the user registration trigger and the someone subscribed trigger are added.

## Code review notes

*Trigger::isTriggeredBy(WorkflowRun $workflowrun)*
With these settings in place, we need to validate if a given set of subjects are suitable to trigger a specific trigger. A user which subscribes to `list A` does not trigger the subscribed trigger with the setting of `subscribed to list B`. In our current setup, this can't happen in the default handle method, which is hooked into a specific action, because we do not yet have the trigger settings available. Therefore I added a `public isTriggeredBy(WorkflowRun $workflowRun) : bool;` which checks if a given WorkflowRun would have been triggered. This method is used shortly before the WorkflowRun gets stored to the database and the actions are stored in the ActionScheduler.

*`<FormTokenField />` does not handle key-value options*
This component only takes different text strings as suggestions. I followed roughly the [Gutenberg example](https://github.com/WordPress/gutenberg/blob/0b0f9e26cc6e315a505865998c5b341600a05492/packages/block-library/src/query/edit/inspector-controls/taxonomy-controls.js#L15-L34) on how to map strings to their IDs. I am not super happy about the solution, but otherwise, we need to create our own <FormFieldToken />  component it seems.

There is an odd `FormTokenField.Value`, but it is not used for suggestions and it seems, it does not really apply. See my `@ts-ignore` problems.

*`<FormTokenField />` label is currently @ts-ignore*
Although `label` exists and works with strings, `@types/wordpress__components` does not really allow for it. I use `@ts-ignore` for now.

## QA notes
`<FormTokenField />` and 'any'-types: As we want "any list" as default, the list can not be empty. When you have "any list" selected, you can click on `x` to remove it, but it does - obviously - not remove it. I didn't find a solution to hide the `x`, which would be nice though.

## After merge notes
Because of [1106c28](https://github.com/mailpoet/mailpoet/pull/4371/commits/1106c288e7f5656ed2aa56b42058fd51c63b06e5) your stored segment-subscribed triggers are no longer known and you will need to configure them again.

## Linked PRs
https://github.com/mailpoet/mailpoet-premium/pull/625

## Linked tickets
[MAILPOET-4587]

[MAILPOET-4587]: https://mailpoet.atlassian.net/browse/MAILPOET-4587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Engine/Workflows/Trigger.php
```diff
@@ -2,6 +2,17 @@
...
   public function registerHooks(): void;
+
+  /**
+   * Validate if the specific context of a run meets the
+   * settings of a given trigger.
+   *
+   * @param WorkflowRun $workflowRun
+   * @return bool
+   */
+  public function isTriggeredBy(WorkflowRun $workflowRun): bool;
```
JanJakes: I can't wrap my head around this, `trigger::isTriggeredBy($worfklowRun)` doesn't seem to make sense to me — it's not a workflow run that triggers a trigger; it's the other way round — a trigger triggers a workflow run. What is the reasoning here? Checking for required subjects or something like that?

EDIT: I found the implementation, and it's clearer to me now. I think we should simplify it, I'll comment more in the implementation part.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/list_panel.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const [segments, setSegments] = useState<FormTokenItem[]>([]);
+
+  const anyValue = {
+    id: 0,
+    name: __('Any list', 'mailpoet'),
+  };
+
+  useEffect(() => {
+    const getData = async () => {
+      const data = await MailPoetAjax.post({
```
JanJakes: Could we make this non-async, without the loaders in the sidebar? Or is that not so easy? Not sure what will be the performance impacts if we make all steps sync 🤔

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
```diff
@@ -0,0 +1,81 @@
+import { FormTokenField as WpFormTokenField } from '@wordpress/components';
+import { uniq } from 'lodash';
```
JanJakes: This is not important now, but could we use just a native `Set`, for instance?
I think there's a tendency to get rid of loadash/underscore in the JS ecosystem as modern JS offers most of that functionality quite out-of-the-box. In Gutenberg repo, I've seen such an initiative.

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  label,
+  selected,
+  suggestions,
+  onChange,
+  anyValue,
+  anyValueIsDefault,
+  ...props
+}: FormTokenFieldProps): JSX.Element {
+  if (anyValue) {
+    suggestions.push(anyValue);
```
JanJakes: Mutating props (well, params in general) can be dangerous. It's probably better to make it immutable (a new array). But, more importantly, this is an effect, and as such, you should never call it in the component body. It may end up adding the item many times instead of once.

mailpoet/assets/js/src/automation/integrations/mailpoet/index.tsx
```diff
@@ -1,6 +1,6 @@
 import { registerStepType } from '../../editor/store';
 import { step as SendEmailStep } from './steps/send_email';
-import { step as SomeoneSubscribedTrigger } from './steps/someone-subscribed';
+import { step as SomeoneSubscribedTrigger } from './steps/someone-subscribes';
```
JanJakes: We should probably also rename `SomeoneSubscribedTrigger` to `SomeoneSubscribesTrigger`.

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+          const hasAny =
+            newSelected.filter((item) => item.id === anyValue.id).length > 0;
+          const hasNone = allSelected.length === 0 && anyValueIsDefault;
+          if (hasAny || hasNone) {
+            // when "any item" has been selected, all other items are removed
+            // or when no item has been selected but "any item" is default, we store "any item"
+            onChange([anyValue]);
+            return;
+          }
+        }
```
JanJakes: Due to the [IF on line 30](https://github.com/mailpoet/mailpoet/pull/4371/commits/f4bde0a02b411648bace63f16310137e781d1863#diff-c993f976a5f0c987562cdc6ff34526cbe7d3d6fcc780207fbb204dd172528237R30), this whole block is unnecessary. It will work the same without it, I think.

And that's good — you could just ignore `anyItem` in the `onChange` callback, make it super-simple, and just watch the `selected` array with an effect such as:

```ts
useEffect(() => {
  if (selected.length === 0) {
    setSelected([...selected, anyValue]);
  }
  if (selected.contains(anyValue) && selected.length > 1) {
    setSelected([...selected.filter({ id }) !== anyValue.id]);
  }
}, [selected]);
```

(In this pseudocode I'm using `setSelected` but I mean the correct dispatch...)

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/list_panel.tsx
```diff
@@ -0,0 +1,91 @@
...
+  }, []);
+
+  const rawSelected = selectedStep.args?.segment_ids
+    ? (selectedStep.args.segment_ids as number[])
+    : [];
+  const selected = segments.filter((segment): boolean =>
+    rawSelected.includes(segment.id as number),
+  );
+  if (!selected.length) {
+    selected.push(anyValue);
```
JanJakes: As mentioned [above](https://github.com/mailpoet/mailpoet/pull/4371/commits/f4bde0a02b411648bace63f16310137e781d1863#diff-c993f976a5f0c987562cdc6ff34526cbe7d3d6fcc780207fbb204dd172528237R30), effects should never be done in a React component body, and it's often better to avoid mutability.

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
```diff
@@ -0,0 +1,81 @@
+import { FormTokenField as WpFormTokenField } from '@wordpress/components';
+import { uniq } from 'lodash';
+
+export type FormTokenItem = {
+  id: number | string;
```
JanJakes, line 4: Can't we convert it and use just one of those types?

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
JanJakes: A more troubling question is, I'm realizing, why do we need this, and what happens with two lists of the same name then?

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  name: string;
+};
+
+export type FormTokenFieldProps = Omit<
+  WpFormTokenField.Props,
+  'value' | 'suggestions' | 'onChange'
```
JanJakes, line 10: The `value` prop doesn't seem to by overwritten. Do we need to `omit` it?

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/SegmentSubscribedTrigger.php
```diff
@@ -13,12 +15,18 @@
 use MailPoet\WP\Functions as WPFunctions;
 
 class SegmentSubscribedTrigger implements Trigger {
```
JanJakes: You renamed the JS part, the PHP should be renamed as well. Same for the triggering hooks, local variables, etc.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/SegmentSubscribedTrigger.php
```diff
@@ -63,4 +71,29 @@ public function handleSubscription(SubscriberSegmentEntity $subscriberSegment):
...
+    $segmentSubject = $workflowRun->requireSingleSubject(SegmentSubject::class);
+    $segment = $segmentSubject->getSegment();
+    $stepArgs = $triggerData->getArgs();
+    if (!isset($stepArgs['segment_ids']) || !is_array($stepArgs['segment_ids'])) {
+      return false;
+    }
+    $segmentIds = $stepArgs['segment_ids'];
+    $anyList = 0;
+    return in_array($anyList, $segmentIds, true) || in_array($segment->getId(), $segmentIds, true);
+  }
```
JanJakes, line 90: So I think this should be just `isTriggeredBy(array $args)`. The workflow, trigger, and run checks are all redundant, the `TriggerHandler` takes care of that. It still feels a bit weird to return to the trigger we just used to trigger the workflow but I see why it's needed.

(Ultimately, maybe the architecture is wrong. Maybe instead of calling `mailpoet_segment_subscribed` somewhere in our code and then landing in `SegmentSubscribedTrigger`, we should be always calling just `mailpoet_automation_trigger` with a trigger name and args. Then the trigger class would receive a workflow run and could implement the `run` method similarly to actions. I'll think about it, no need to change now.)

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/UserRegistrationTrigger.php
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  private function getSegment(SubscriberEntity $subscriber): SegmentEntity {
+
+    $segments = $subscriber->getSubscriberSegments()->toArray();
+    if (!$segments) {
+      throw new InvalidStateException();
+    }
+    $segment = null;
+    foreach ($segments as $segment) {
```
JanJakes, line 111: You can also inject `SubscribersRepository` to this class, if it helps.

JanJakes (commented):
I have some comments and remarks but otherwise it's a great job and even better, with testws! 👏

There is one substantial issue that I'm realizing only now, at the end of the review, when testing — I think you picked the wrong UI component.

What I mean is that this is not a selectbox:

Screen Shot 2022-09-20 at 13 55 01

It doesn't work very well for the users — they need to type two letters of a list to discover it. It doen't show options, it's not a multi-select.

The specs even include a link to the correct component:

> Use @wordpress/components for the trigger settings; in particular, the [Dropdown Selector](https://wordpress.github.io/gutenberg/?path=/story/components-formtokenfield--dropdown-selector) will likely be useful here.

JanJakes:
@websupporter To be more specific on this:

> There is one substantial issue that I'm realizing only now, at the end of the review, when testing — I think you picked the wrong UI component.

Maybe it's the correct component, just misconfigured? It doesn't work as it does here: https://wordpress.github.io/gutenberg/?path=/story/components-formtokenfield--dropdown-selector

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/SegmentSubscribedTrigger.php
JanJakes: If it simplifies things, `isTriggeredBy(array $args, array $subjects)` could also work, maybe.

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
websupporter: user roles IDs are strings while list segment IDs are integers. I preferred to mark this distinctively. The diff to change that would look like this:
```
diff --git a/mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx b/mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
index edb20d8fc..64341c2a1 100644
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
@@ -1,7 +1,7 @@
 import { FormTokenField as WpFormTokenField } from '@wordpress/components';
 
 export type FormTokenItem = {
-  id: number | string;
+  id: string;
   name: string;
 };
 
diff --git a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/list_panel.tsx b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/list_panel.tsx
index eb41919d1..168032499 100644
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/list_panel.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/list_panel.tsx
@@ -15,14 +15,14 @@ export function ListPanel(): JSX.Element {
   );
 
   const rawSelected = selectedStep.args?.segment_ids
-    ? (selectedStep.args.segment_ids as number[])
+    ? (selectedStep.args.segment_ids as string[])
     : [];
 
   const validSegments = segments.filter(
     (segment) => segment.type === 'default',
   );
   const selected = validSegments.filter((segment): boolean =>
-    rawSelected.includes(segment.id as number),
+    rawSelected.includes(segment.id),
   );
   return (
     <PanelBody opened>
```
We would pretend the segment ID is a string, while it is not. JavaScript will still send an integer to the Rest API (so we are "typesafe" in PHP), but I preferred to spell out the real types in TypeScript and it feels to me otherwise, we would obscure that fact.

What do you think?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/list_panel.tsx
websupporter: Addressed in [c04a306](https://github.com/mailpoet/mailpoet/pull/4371/commits/c04a306efb701df4ca62d8ef2b1db841a3211b28)

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
websupporter: Problem outdated thanks to [710190b](https://github.com/mailpoet/mailpoet/pull/4371/commits/710190bd68357719381a9d577bc886a495f5a367)

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
websupporter: Problem outdated thanks to [710190b](https://github.com/mailpoet/mailpoet/pull/4371/commits/710190bd68357719381a9d577bc886a495f5a367)

mailpoet/assets/js/src/automation/integrations/mailpoet/index.tsx
websupporter: Addressed in e8a1f5dd7

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
websupporter: Problem outdated thanks to [710190b](https://github.com/mailpoet/mailpoet/pull/4371/commits/710190bd68357719381a9d577bc886a495f5a367)

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/list_panel.tsx
websupporter: Problem outdated thanks to [710190b](https://github.com/mailpoet/mailpoet/pull/4371/commits/710190bd68357719381a9d577bc886a495f5a367)

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
websupporter: Issue solved via [db83b9b](https://github.com/mailpoet/mailpoet/pull/4371/commits/db83b9bbc1755f2c31919ec4036b3b37bc830439). In [db83b9b](https://github.com/mailpoet/mailpoet/pull/4371/commits/db83b9bbc1755f2c31919ec4036b3b37bc830439) I align our props with the original FormTokenField.Props. I used `selected` where FormTokenField uses `value`. Now, we use `value` as well.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/SegmentSubscribedTrigger.php
websupporter: Addressed in [1106c28](https://github.com/mailpoet/mailpoet/pull/4371/commits/1106c288e7f5656ed2aa56b42058fd51c63b06e5)

websupporter:
Thank you for your review @JanJakes. I addressed most of the issues, please see my comments. 

> Maybe it's the correct component, just misconfigured

Addressed in [778cc06](https://github.com/mailpoet/mailpoet/pull/4371/commits/778cc0662249148d63a4dcf956813d9d60de0b1d)

The main difference was I did not use `__experimentalExpandOnFocus` which I do use now. The problem which occurs: For some reason, the select is now open as soon as the sidebar opens (and not just `onFocus`). I could not find the reason for this problem or a workaround :thinking:

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
websupporter: 710190b is a bit of a bigger change. Instead of having an "Any item" option, I changed the behavior to make "Any item" the placeholder. This solved my mentioned "Any item has a `x` to close it, but because it's default it does not get deleted"-problem. It also simplified much of the used logic and some of the problems, you are mentioning further down.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/SegmentSubscribedTrigger.php
websupporter: That is a good suggestion! Thank you. Addressed in [1a41901](https://github.com/mailpoet/mailpoet/pull/4371/commits/1a41901bf324435635221ea43f0fe917d678b69d)

mailpoet/assets/js/src/automation/integrations/mailpoet/components/form-token-field/index.tsx
JanJakes: Ahh, thanks for the clarification. So, in fact, it's more like a generic than a union since we won't mix strings and numbers in the same component instance. But I think this is fine; with generics we'd need to make the whole component generic, and I think it's not really worth it...

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -66,6 +76,7 @@ public function render() {
         'nonce' => $this->wp->wpCreateNonce('wp_rest'),
       ],
       'user_roles' => $roles->get_names(),
+      'segments' => $segments,
```
JanJakes, line 79: This is fine now but I'll take a note we should somehow allow integrations to "preload" stuff themselves.

JanJakes (approved):
> The main difference was I did not use __experimentalExpandOnFocus which I do use now. The problem which occurs: For some reason, the select is now open as soon as the sidebar opens (and not just onFocus). I could not find the reason for this problem or a workaround 🤔

That looks like a bug in the component. I was wondering why it doesn't occur in their Storybook and I think it was fixed so it should be fine once we update. [On this line](https://github.com/WordPress/gutenberg/blame/trunk/packages/components/src/form-token-field/index.tsx#L528) the version we're using doesn't have the `hasFocus()` check.

PR #4373, opened 2022-09-19
Workflow validation part 1 [MAILPOET-4629]

JanJakes:
## Description

This ticket implements the first part of workflow validation. On the way, it was necessary to improve and refactor the logic around subjects so we're able to validate steps and provide information about their compatibility with other steps.

The contents of this PR:
1. Universal workflow graph walker to traverse workflow graph and allow implementing validation on top of that (+ tests).
2. Implementation of various "rules" that validate the graph while it's being traversed (+ tests).
3. Running the validation on workflow creation and update.
4. Refactor subjects to separate data from services and define their step requirements. This introduces `Payload` (the data a subject is carrying) and `SubjectEntry` (subject, the service + payload). It also improves some architecture around it and removes logic that felt a bit misplaced from `WorkflowRun` and `WorkflowStorage`.
5. Introducing `StepRunArgs` for convenience and resolving subject data.

To be continued in the next ticket:
1. Validation of workflow steps before workflow activation.
2. Validation of step compatibility (if they can be in a specific order).
3. UI.

## Code review notes

The commit messages should tell the story. A lot of the bigger changes are moving stuff around. It's not too complex, I think.

## QA notes

QA step can be skipped.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/624

## Linked tickets

[MAILPOET-4629]

## After-merge notes

_N/A_

[MAILPOET-4523]: https://mailpoet.atlassian.net/browse/MAILPOET-4523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-4629]: https://mailpoet.atlassian.net/browse/MAILPOET-4629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -181,7 +181,7 @@ class StepHandler {
   private function getSubjectEntries(WorkflowRun $workflowRun, array $requiredSubjectKeys): array {
     $subjectDataMap = [];
     foreach ($workflowRun->getSubjects() as $data) {
-      $subjectDataMap[$data->getKey()] = array_merge($subjectDataMap[$data->getKey()], [$data]);
+      $subjectDataMap[$data->getKey()] = array_merge($subjectDataMap[$data->getKey()] ?? [], [$data]);
     }
 
     $subjectEntries = [];
--- a/mailpoet/lib/Automation/Engine/Control/StepHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/StepHandler.php
@@ -5,6 +5,7 @@ namespace MailPoet\Automation\Engine\Control;
 use Exception;
 use MailPoet\Automation\Engine\Control\Steps\ActionStepRunner;
 use MailPoet\Automation\Engine\Data\Step;
+use MailPoet\Automation\Engine\Data\SubjectEntry;
 use MailPoet\Automation\Engine\Data\WorkflowRun;
 use MailPoet\Automation\Engine\Data\WorkflowRunLog;
 use MailPoet\Automation\Engine\Exceptions;
@@ -14,6 +15,9 @@ use MailPoet\Automation\Engine\Storage\WorkflowRunLogStorage;
 use MailPoet\Automation\Engine\Storage\WorkflowRunStorage;
 use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Automation\Engine\WordPress;
+use MailPoet\Automation\Engine\Workflows\Action;
+use MailPoet\Automation\Engine\Workflows\Payload;
+use MailPoet\Automation\Engine\Workflows\Subject;
 use Throwable;
 
 class StepHandler {
@@ -22,6 +26,9 @@ class StepHandler {
 
   /** @var ActionStepRunner */
   private $actionStepRunner;
+
+  /** @var SubjectLoader */
+  private $subjectLoader;
 
   /** @var WordPress */
   private $wordPress;
@@ -45,6 +52,7 @@ class StepHandler {
     ActionScheduler $actionScheduler,
     ActionStepRunner $actionStepRunner,
     Hooks $hooks,
+    SubjectLoader $subjectLoader,
     WordPress $wordPress,
     WorkflowRunStorage $workflowRunStorage,
     WorkflowRunLogStorage $workflowRunLogStorage,
@@ -53,6 +61,7 @@ class StepHandler {
     $this->actionScheduler = $actionScheduler;
     $this->actionStepRunner = $actionStepRunner;
     $this->hooks = $hooks;
+    $this->subjectLoader = $subjectLoader;
     $this->wordPress = $wordPress;
     $this->workflowRunStorage = $workflowRunStorage;
     $this->workflowRunLogStorage = $workflowRunLogStorage;
@@ -122,7 +131,9 @@ class StepHandler {
     if (isset($this->stepRunners[$stepType])) {
       $log = new WorkflowRunLog($workflowRun->getId(), $step->getId());
       try {
-        $this->stepRunners[$stepType]->run($step, $workflow, $workflowRun);
+        $requiredSubjects = $step instanceof Action ? $step->getSubjectKeys() : [];
+        $subjectEntries = $this->getSubjectEntries($workflowRun, $requiredSubjects);
+        $this->stepRunners[$stepType]->run($workflow, $workflowRun, $step, $subjectEntries);
         $log->markCompletedSuccessfully();
       } catch (Throwable $e) {
         $log->markFailed();
@@ -162,5 +173,25 @@ class StepHandler {
     // enqueue next step
     $this->actionScheduler->enqueue(Hooks::WORKFLOW_STEP, $nextStepArgs);
     // TODO: allow long-running steps (that are not done here yet)
+  }
+
+  /** @return SubjectEntry<Subject<Payload>>[] */
+  private function getSubjectEntries(WorkflowRun $workflowRun, array $requiredSubjectKeys): array {
+    $subjectDataMap = [];
+    foreach ($workflowRun->getSubjects() as $data) {
+      $subjectDataMap[$data->getKey()] = array_merge($subjectDataMap[$data->getKey()], [$data]);
+    }
+
+    $subjectEntries = [];
+    foreach ($requiredSubjectKeys as $key) {
+      $subjectData = $subjectDataMap[$key] ?? null;
+      if (!$subjectData) {
+        throw Exceptions::subjectDataNotFound($key, $workflowRun->getId());
+      }
+      foreach ($subjectData as $data) {
+        $subjectEntries[] = $this->subjectLoader->getSubjectEntry($data);
+      }
+    }
+    return $subjectEntries;
   }
 }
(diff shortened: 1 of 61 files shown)


mailpoet/lib/Automation/Engine/Validation/WorkflowRules/ConsistentStepMapRule.php
```diff
@@ -0,0 +1,24 @@
...
+use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Validation\WorkflowGraph\WorkflowNode;
+use MailPoet\Automation\Engine\Validation\WorkflowGraph\WorkflowNodeVisitor;
+
+class ConsistentStepMapRule implements WorkflowNodeVisitor {
+  public function initialize(Workflow $workflow): void {
+    foreach ($workflow->getSteps() as $id => $step) {
+      if ($id !== $step->getId()) {
+        throw Exceptions::workflowStructureNotValid(__('TODO', 'mailpoet'));
```
websupporter: ToDo :)

mailpoet/tests/unit/Automation/Engine/Validation/WorkflowRules/NoDuplicateEdgesTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+require_once __DIR__ . '/WorkflowRuleTest.php';
+
+use MailPoet\Automation\Engine\Exceptions\UnexpectedValueException;
+use MailPoet\Automation\Engine\Validation\WorkflowGraph\WorkflowWalker;
+
+class NoDuplicateEdgesTest extends WorkflowRuleTest {
+  public function testItDetectsDuplicateEdges(): void {
+    $workflow = $this->createWorkflow([
+      'root' => ['a', 'a'],
```
websupporter, line 13: Is there an important reason for not allowing duplicate edges? For a `[yes,no]` switch, it does not make sense, but seems not to harm as well.

For a `[smaller_than_range,range,bigger_than_range]` it might be a valid option to union smaller and bigger in the same flow and the design of the step might not allow for a `[yes,no]`.

I just wonder, if I miss something.

websupporter:
image
In this `PUT` request I fire with a wrong property key. This seems not to bother us and the output is with the "expected" key, which means, we store it correctly. I wonder if we still should have a rule against such a case?

websupporter:
I think this PR is only concerned with the graph consistency. Will `type` and `args` be part of [MAILPOET-4659]?
image


[MAILPOET-4659]: https://mailpoet.atlassian.net/browse/MAILPOET-4659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mailpoet/lib/Automation/Engine/Exceptions.php
```diff
@@ -127,16 +133,66 @@ public static function subjectLoadFailed(string $key, array $args): InvalidState
...
       ->withErrorCode(self::WORKFLOW_STRUCTURE_MODIFICATION_NOT_SUPPORTED)
       ->withMessage(__("Workflow structure modification not supported.", 'mailpoet'));
   }
+
+  public static function workflowStructureNotValid(string $detail): UnexpectedValueException {
+    return UnexpectedValueException::create()
+      ->withErrorCode(self::WORKFLOW_STRUCTURE_NOT_VALID)
+      // translators: %s is a detailed information
+      ->withMessage(sprintf(__("Invalid workflow structure: %s", 'mailpoet'), $detail));
+  }
```
websupporter, line 183: It can be pretty hard to determine, which rule was actually broken. Once we debug on a translated site, it would be even harder. Maybe we could have an error code per rule?

Current output example, where I stumpled for a bit:
```
{
    "code": "mailpoet_automation_workflow_structure_not_valid",
    "message": "Invalid workflow structure: Step with ID 'A-bad-step' not found (referenced from 'tsucp1zbk5woswoc')",
    "data": {
        "status": 400
    }
}
``

File: mailpoet/tests/unit/Automation/Engine/Validation/WorkflowRules/NoDuplicateEdgesTest.php (hunk 2 above)
Comment (JanJakes) [reply mailpoet/tests/unit/Automation/Engine/Validation/WorkflowRules/NoDuplicateEdgesTest.php:13]: I think we can allow this in the future if we find a use case. For now, I disabled "everything", including splits and joins, that we will want to support as well. This rule is there just so we can allow particular constructs more granularly (e.g. splits for a yes/no without duplicate edges). So I guess the answer is we can surely allow it in the future.

**Comment by JanJakes** [comment]
@websupporter
> In this PUT request I fire with a wrong property key. This seems not to bother us and the output is with the "expected" key, which means, we store it correctly. I wonder if we still should have a rule against such a case?

Hmm, I'll need to check this. The `ConsistentStepMapRule` should take care of this.

> I think this PR is only concerned with the graph consistency. Will type and args be part of [MAILPOET-4659](https://mailpoet.atlassian.net/browse/MAILPOET-4659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)?

Yes, I started working on that but then went to refactor subjects and it was big enough, so I ended up splitting the task.

File: mailpoet/lib/Automation/Engine/Exceptions.php (hunk 3 above)
Comment (JanJakes) [reply mailpoet/lib/Automation/Engine/Exceptions.php:183]: One thing to note here is that these errors should never occur to us. Our frontend should never construct a workflow that would break those rules. If that happens, we have a bug. So it's mainly to protect against bugs and people playing with the API directly.

That said, we could add an error per rule... or maybe just some specific `rule` field in the `data`?

Comment (websupporter) [reply mailpoet/lib/Automation/Engine/Exceptions.php:183]: > or maybe just some specific rule field in the data

Whatever suits you best there. I do not have a strong opinion as long as we can quickly identify the broken rule.

File: mailpoet/lib/Automation/Engine/Control/StepHandler.php [hunk 4]
```diff
@@ -163,4 +176,24 @@ private function handleStep(array $args): void {
     $this->actionScheduler->enqueue(Hooks::WORKFLOW_STEP, $nextStepArgs);
     // TODO: allow long-running steps (that are not done here yet)
   }
+
+  /** @return SubjectEntry<Subject<Payload>>[] */
+  private function getSubjectEntries(WorkflowRun $workflowRun, array $requiredSubjectKeys): array {
+    $subjectDataMap = [];
+    foreach ($workflowRun->getSubjects() as $data) {
+      $subjectDataMap[$data->getKey()] = array_merge($subjectDataMap[$data->getKey()], [$data]);
```
websupporter: `$subjectDataMap[$data->getKey()] might not exist, maybe better:

`$subjectDataMap[$data->getKey()] = array_merge($subjectDataMap[$data->getKey()]??[], [$data]);`

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/ConsistentStepMapRule.php
JanJakes: Fixed.

mailpoet/lib/Automation/Engine/Exceptions.php
JanJakes: I think I prefer just adding the rule ID, but I'd move it to the follow-up ticket, if possible. Currently, we don't have an easy way of extending the error response data when converted from an exception but I'll probably need that in the follow-up ticket anyway.

mailpoet/lib/Automation/Engine/Control/StepHandler.php
JanJakes: Good catch, fixed.

mailpoet/lib/Automation/Engine/Exceptions.php
websupporter: sounds good to me.

websupporter (dismissed):
LGTM. Thank you

websupporter (approved):
LGTM.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/UserRegistrationTrigger.php
```diff
@@ -40,84 +42,52 @@ public function getArgsSchema(): ObjectSchema {
...
-    $segment = null;
-    foreach ($segments as $segment) {
-      $segment = $segment->getSegment();
-      if (!$segment || $segment->getType() !== SegmentEntity::TYPE_WP_USERS) {
-        continue;
-      }
-      break;
-    }
-    if (!$segment || $segment->getType() !== SegmentEntity::TYPE_WP_USERS) {
+  private function getWpSegment(SubscriberEntity $subscriber): SegmentEntity {
```
websupporter, line 85: @JanJakes Trying to create a new user results now in an error for me: 
image

PR #4376, opened 2022-09-20
Workflow listing actions — part 1 [MAILPOET-4540]

johnolek:
## Description

Implements actions in the automation listing table.

## Code review notes

This is the backend part, extracted from the original large PR.
A frontend PR where all the changes will be visible and testable will follow (some related comments are visible here).

From the original PR by @johnolek I made a few changes:

1. Did a rebase.
2. Extracted workflow response building to a mapper (the previous response object was not DI-injectable).
3. Extracted delete and duplicate logic to controllers. Now I'm thinking if most of the delete logic should also go out of the storage to a controller 🤔
4. Added regenerating step IDs on duplication. Just to be sure...
5. Added tests for both new endpoints.

## QA notes

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/634

## Linked tickets

[MAILPOET-4540]

## After-merge notes

_N/A_

[MAILPOET-4540]: https://mailpoet.atlassian.net/browse/MAILPOET-4540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-automation-listing/_listing.scss
+++ b/mailpoet/assets/css/src/components-automation-listing/_listing.scss
@@ -8,8 +8,6 @@
   .mailpoet-automation-listing {
     /* Prevent border radius beneath tabs */
     border-radius: 0 0 1px 1px;
-    /* Prevent overlapping box shadow since listing is rendered inside tabs */
-    clip-path: inset(-1px 0 0 0);
   }
 }
 
--- a/mailpoet/assets/js/src/automation/listing/components/cells/more.tsx
+++ b/mailpoet/assets/js/src/automation/listing/components/cells/more.tsx
@@ -13,12 +13,6 @@ export function More({ workflow }: Props): JSX.Element {
       renderContent={() => (
         <div>
           <MenuItem onInvoke={() => {}}>
-            <p>{__('Rename automation', 'mailpoet')}</p>
-          </MenuItem>
-          <MenuItem onInvoke={() => {}}>
-            <p>{__('Statistics', 'mailpoet')}</p>
-          </MenuItem>
-          <MenuItem onInvoke={() => {}}>
             <p>{__('Duplicate', 'mailpoet')}</p>
           </MenuItem>
           <MenuItem onInvoke={() => {}}>
--- a/mailpoet/assets/js/src/automation/listing/index.tsx
+++ b/mailpoet/assets/js/src/automation/listing/index.tsx
@@ -75,12 +75,17 @@ export function AutomationListing({ workflows, loading }: Props): JSX.Element {
   );
 
   const groupedWorkflows = useMemo<Record<string, Workflow[]>>(() => {
-    const grouped = {};
+    const grouped = {
+      all: [],
+    };
     workflows.forEach((workflow) => {
       if (!grouped[workflow.status]) {
         grouped[workflow.status] = [];
       }
       grouped[workflow.status].push(workflow);
+      if (workflow.status !== WorkflowStatus.TRASH) {
+        grouped.all.push(workflow);
+      }
     });
     return grouped;
   }, [workflows]);
@@ -88,10 +93,7 @@ export function AutomationListing({ workflows, loading }: Props): JSX.Element {
   const tabs = useMemo(
     () =>
       filterTabs.map((filterTab) => {
-        const count =
-          filterTab.name === 'all'
-            ? workflows.length
-            : (groupedWorkflows[filterTab.name] || []).length;
+        const count = (groupedWorkflows[filterTab.name] || []).length;
         return {
           name: filterTab.name,
           title:
@@ -106,13 +108,12 @@ export function AutomationListing({ workflows, loading }: Props): JSX.Element {
           className: filterTab.className,
         };
       }),
-    [workflows, groupedWorkflows],
+    [groupedWorkflows],
   );
 
   const tabRenderer = useCallback(
     (tab) => {
-      const filteredWorkflows: Workflow[] =
-        tab.name === 'all' ? workflows : groupedWorkflows[tab.name] ?? [];
+      const filteredWorkflows: Workflow[] = groupedWorkflows[tab.name] ?? [];
       const rowsPerPage = parseInt(pageSearch.get('per_page') || '25', 10);
       const currentPage = parseInt(pageSearch.get('paged') || '1', 10);
       const start = (currentPage - 1) * rowsPerPage;
--- a/mailpoet/lib/Automation/Engine/Builder/DeleteWorkflowController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/DeleteWorkflowController.php
@@ -21,6 +21,11 @@ class DeleteWorkflowController {
     if (!$workflow) {
       throw Exceptions::workflowNotFound($id);
     }
+
+    if ($workflow->getStatus() !== Workflow::STATUS_TRASH) {
+      throw Exceptions::workflowNotTrashed($id);
+    }
+
     $this->workflowStorage->deleteWorkflow($workflow);
     return $workflow;
   }
--- a/mailpoet/lib/Automation/Engine/Builder/DeleteWorkflowController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/DeleteWorkflowController.php
@@ -1,0 +1,27 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Builder;
+
+use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Storage\WorkflowStorage;
+
+class DeleteWorkflowController {
+  /** @var WorkflowStorage */
+  private $workflowStorage;
+
+  public function __construct(
+    WorkflowStorage $workflowStorage
+  ) {
+    $this->workflowStorage = $workflowStorage;
+  }
+
+  public function deleteWorkflow(int $id): Workflow {
+    $workflow = $this->workflowStorage->getWorkflow($id);
+    if (!$workflow) {
+      throw Exceptions::workflowNotFound($id);
+    }
+    $this->workflowStorage->deleteWorkflow($workflow);
+    return $workflow;
+  }
+}
--- a/mailpoet/lib/Automation/Engine/Builder/DuplicateWorkflowController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/DuplicateWorkflowController.php
@@ -1,0 +1,91 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Builder;
+
+use MailPoet\Automation\Engine\Data\NextStep;
+use MailPoet\Automation\Engine\Data\Step;
+use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Exceptions\InvalidStateException;
+use MailPoet\Automation\Engine\Storage\WorkflowStorage;
+use MailPoet\Automation\Engine\WordPress;
+use MailPoet\Util\Security;
+
+class DuplicateWorkflowController {
+  /** @var WordPress */
+  private $wordPress;
+
+  /** @var WorkflowStorage */
+  private $workflowStorage;
+
+  public function __construct(
+    WordPress $wordPress,
+    WorkflowStorage $workflowStorage
+  ) {
+    $this->wordPress = $wordPress;
+    $this->workflowStorage = $workflowStorage;
+  }
+
+  public function duplicateWorkflow(int $id): Workflow {
+    $workflow = $this->workflowStorage->getWorkflow($id);
+    if (!$workflow) {
+      throw Exceptions::workflowNotFound($id);
+    }
+
+    $duplicate = new Workflow(
+      $this->getName($workflow->getName()),
+      $this->getSteps($workflow->getSteps()),
+      $this->wordPress->wpGetCurrentUser()
+    );
+    $duplicate->setStatus(Workflow::STATUS_DRAFT);
+
+    $workflowId = $this->workflowStorage->createWorkflow($duplicate);
+    $savedWorkflow = $this->workflowStorage->getWorkflow($workflowId);
+    if (!$savedWorkflow) {
+      throw new InvalidStateException('Workflow not found.');
+    }
+    return $savedWorkflow;
+  }
+
+  private function getName(string $name): string {
+    // translators: %s is the original workflow name.
+    $newName = sprintf(__('Copy of %s', 'mailpoet'), $name);
+    $maxLength = $this->workflowStorage->getNameColumnLength();
+    if (strlen($newName) > $maxLength) {
+      $append = '…';
+      return substr($newName, 0, $maxLength - strlen($append)) . $append;
+    }
+    return $newName;
+  }
+
+  /**
+   * @param Step[] $steps
+   * @return Step[]
+  */
+  private function getSteps(array $steps): array {
+    $newIds = [];
+    foreach ($steps as $step) {
+      $id = $step->getId();
+      $newIds[$id] = $id === 'root' ? 'root' : $this->getId();
+    }
+
+    $newSteps = [];
+    foreach ($steps as $step) {
+      $newId = $newIds[$step->getId()];
+      $newSteps[$newId] = new Step(
+        $newId,
+        $step->getType(),
+        $step->getKey(),
+        $step->getArgs(),
+        array_map(function (NextStep $nextStep) use ($newIds): NextStep {
+          return new NextStep($newIds[$nextStep->getId()]);
+        }, $step->getNextSteps())
+      );
+    }
+    return $newSteps;
+  }
+
+  private function getId(): string {
+    return Security::generateRandomString(16);
+  }
+}
--- a/mailpoet/lib/Automation/Engine/Data/Workflow.php
+++ b/mailpoet/lib/Automation/Engine/Data/Workflow.php
@@ -191,8 +191,8 @@ class Workflow {
       }, Json::decode($data['steps'])),
       new \WP_User((int)$data['author'])
     );
-    $workflow->id = (int)($data['id'] ?? 0);
-    $workflow->versionId = (int)($data['version_id'] ?? 0);
+    $workflow->id = (int)$data['id'];
+    $workflow->versionId = (int)$data['version_id'];
     $workflow->status = $data['status'];
     $workflow->createdAt = new DateTimeImmutable($data['created_at']);
     $workflow->updatedAt = new DateTimeImmutable($data['updated_at']);
--- a/mailpoet/lib/Automation/Engine/Data/Workflow.php
+++ b/mailpoet/lib/Automation/Engine/Data/Workflow.php
@@ -191,8 +191,8 @@ class Workflow {
       }, Json::decode($data['steps'])),
       new \WP_User((int)$data['author'])
     );
-    $workflow->id = (int)$data['id'];
-    $workflow->versionId = (int)$data['version_id'];
+    $workflow->id = (int)($data['id'] ?? 0);
+    $workflow->versionId = (int)($data['version_id'] ?? 0);
     $workflow->status = $data['status'];
     $workflow->createdAt = new DateTimeImmutable($data['created_at']);
     $workflow->updatedAt = new DateTimeImmutable($data['updated_at']);
--- a/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDeleteEndpoint.php
+++ b/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDeleteEndpoint.php
@@ -1,0 +1,42 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Endpoints\Workflows;
+
+use MailPoet\API\REST\Request;
+use MailPoet\API\REST\Response;
+use MailPoet\Automation\Engine\API\Endpoint;
+use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Exceptions\InvalidStateException;
+use MailPoet\Automation\Engine\Storage\WorkflowStorage;
+use MailPoet\Validator\Builder;
+
+class WorkflowsDeleteEndpoint extends Endpoint {
+  /** @var WorkflowStorage */
+  private $workflowStorage;
+
+  public function __construct(
+    WorkflowStorage $workflowStorage
+  ) {
+    $this->workflowStorage = $workflowStorage;
+  }
+
+  public function handle(Request $request): Response {
+    $workflowId = $request->getParam('id');
+    if (!is_int($workflowId)) {
+      throw InvalidStateException::create();
+    }
+    $existingWorkflow = $this->workflowStorage->getWorkflow($workflowId);
+    if (!$existingWorkflow instanceof Workflow) {
+      throw InvalidStateException::create();
+    }
+    $this->workflowStorage->deleteWorkflow($existingWorkflow);
+
+    return new Response(null);
+  }
+
+  public static function getRequestSchema(): array {
+    return [
+      'id' => Builder::integer()->required(),
+    ];
+  }
+}
--- a/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDeleteEndpoint.php
+++ b/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDeleteEndpoint.php
@@ -5,32 +5,22 @@ namespace MailPoet\Automation\Engine\Endpoints\Workflows;
 use MailPoet\API\REST\Request;
 use MailPoet\API\REST\Response;
 use MailPoet\Automation\Engine\API\Endpoint;
-use MailPoet\Automation\Engine\Data\Workflow;
-use MailPoet\Automation\Engine\Exceptions\InvalidStateException;
-use MailPoet\Automation\Engine\Storage\WorkflowStorage;
+use MailPoet\Automation\Engine\Builder\DeleteWorkflowController;
 use MailPoet\Validator\Builder;
 
 class WorkflowsDeleteEndpoint extends Endpoint {
-  /** @var WorkflowStorage */
-  private $workflowStorage;
+  /** @var DeleteWorkflowController */
+  private $deleteController;
 
   public function __construct(
-    WorkflowStorage $workflowStorage
+    DeleteWorkflowController $deleteController
   ) {
-    $this->workflowStorage = $workflowStorage;
+    $this->deleteController = $deleteController;
   }
 
   public function handle(Request $request): Response {
-    $workflowId = $request->getParam('id');
-    if (!is_int($workflowId)) {
-      throw InvalidStateException::create();
-    }
-    $existingWorkflow = $this->workflowStorage->getWorkflow($workflowId);
-    if (!$existingWorkflow instanceof Workflow) {
-      throw InvalidStateException::create();
-    }
-    $this->workflowStorage->deleteWorkflow($existingWorkflow);
-
+    $workflowId = intval($request->getParam('id'));
+    $this->deleteController->deleteWorkflow($workflowId);
     return new Response(null);
   }
 
--- a/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDuplicateEndpoint.php
+++ b/mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDuplicateEndpoint.php
@@ -5,41 +5,28 @@ namespace MailPoet\Automation\Engine\Endpoints\Workflows;
 use MailPoet\API\REST\Request;
 use MailPoet\API\REST\Response;
 use MailPoet\Automation\Engine\API\Endpoint;
-use MailPoet\Automation\Engine\Data\Workflow;
-use MailPoet\Automation\Engine\Exceptions\InvalidStateException;
+use MailPoet\Automation\Engine\Builder\DuplicateWorkflowController;
 use MailPoet\Automation\Engine\Mappers\WorkflowMapper;
-use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Validator\Builder;
 
 class WorkflowsDuplicateEndpoint extends Endpoint {
   /** @var WorkflowMapper */
   private $workflowMapper;
 
-  /** @var WorkflowStorage */
-  private $workflowStorage;
+  /** @var DuplicateWorkflowController */
+  private $duplicateController;
 
   public function __construct(
-    WorkflowMapper $workflowMapper,
-    WorkflowStorage $workflowStorage
+    DuplicateWorkflowController $duplicateController,
+    WorkflowMapper $workflowMapper
   ) {
     $this->workflowMapper = $workflowMapper;
-    $this->workflowStorage = $workflowStorage;
+    $this->duplicateController = $duplicateController;
   }
 
   public function handle(Request $request): Response {
-    $workflowId = $request->getParam('id');
-    if (!is_int($workflowId)) {
-      throw InvalidStateException::create();
-    }
-    $existingWorkflow = $this->workflowStorage->getWorkflow($workflowId);
-    if (!$existingWorkflow instanceof Workflow) {
-      throw InvalidStateException::create();
-    }
-    $duplicateId = $this->workflowStorage->duplicateWorkflow($existingWorkflow);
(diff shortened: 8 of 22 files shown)


mailpoet/assets/css/src/components-automation-listing/_listing.scss
```diff
@@ -8,8 +8,12 @@
...
     /* Prevent border radius beneath tabs */
     border-radius: 0 0 1px 1px;
-    /* Prevent overlapping box shadow since listing is rendered inside tabs */
-    clip-path: inset(-1px 0 0 0);
+  }
+}
+
+.mailpoet-automation-ellipsis-menu {
+  .woocommerce-ellipsis-menu__popover {
+    opacity: 1;
```
johnolek: I can't figure out why the popover is getting an opacity of `0` and an incorrect position calculated when I have the wordpress components CSS enqueued, so this is a temporary workaround to make the dropdown at least visible while I test it. I could really use some advice on a proper fix though.

mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -1,21 +1,50 @@
...
   const { data, loading, error } = useQuery<{ data: Workflow[] }>('workflows');
+  const noticesRef = useRef([]);
+  const [notices, setNotices] = useState(noticesRef.current);
+  const addNotice = useCallback(
+    (notice) => {
+      noticesRef.current = [notice, ...noticesRef.current];
+      setNotices(noticesRef.current);
+    },
+    [noticesRef],
+  );
```
johnolek: Using a ref here is probably a code smell. I want to have a consistent way for the component and its children to add notices when certain things occur. If I don't use a ref, I sometimes end up with an old copy of `setNotices` that isn't aware of all the latest notices.

I noticed this with the Undo button for trashed automations. Its callback uses the old version of `addNotice`, so it would add its notice but it wouldn't retain the original one that had the Undo button in the first place.

I hope that makes sense. Like I said, this is probably a code smell and I'm sure there's a cleaner way to handle this.

mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -1,21 +1,50 @@
...
+          <p>{apiError.message}</p>
+        </Notice>,
+      );
+    },
+    [addNotice],
+  );
+
+  useEffect(() => {
+    errorHandler = noticeErrorHandler;
+  }, [noticeErrorHandler]);
```
johnolek: Is there a better way of doing this without using a variable in the outer scope? It doesn't seem like I would want to register the middleware as part of the component's lifecycle because it could occur multiple times, but I could be wrong about that.

mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -27,7 +56,14 @@ function Content(): JSX.Element {
 
   const workflows = data?.data ?? [];
   return workflows.length > 0 ? (
-    <AutomationListing workflows={workflows} loading={loading} />
+    <>
+      {notices}
```
johnolek: One thing to note is that this file  now has two different ways of generating notices: one based on URL params and this being the other. Once we've settled on an approach I was hoping to consolidate these to use the same approach.

mailpoet/assets/js/src/automation/listing/components/cells/more.tsx
```diff
@@ -1,29 +1,250 @@
...
+          },
+        );
+      }}
+    >
+      Undo
+    </Button>
+  );
+}
+
+export function More({ workflow, listingActions }: Props): JSX.Element {
```
johnolek: This all feels like a huge mess right now. I'm sure it should be possible to simplify or at least extract some things.

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -47,14 +48,62 @@ const tableHeaders = [
...
+
+  const addWorkflow = useCallback(
+    (newWorkflow: Workflow) => {
+      const newWorkflows = [newWorkflow, ...allWorkflows];
+      setAllWorkflows(newWorkflows);
+    },
+    [allWorkflows],
+  );
+
+  const listingActions = useMemo<ListingActions>(
```
johnolek: This feels kind of silly, like I'm recreating a worse version of redux. Maybe we should just be using redux/wp data?

I played around with the idea of just having a "refreshWorkflows" action that would re-fetch all the workflows from the server, but it ended up feeling clunky. That being said, this way feels somewhat brittle.

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -47,14 +48,62 @@ const tableHeaders = [
...
+  addNotice,
+}: Props): JSX.Element {
   const history = useHistory();
   const location = useLocation();
   const pageSearch = useMemo(
     () => new URLSearchParams(location.search),
     [location],
   );
 
+  const [allWorkflows, setAllWorkflows] = useState(workflows);
```
johnolek: I'm wondering if this is also a code smell, immediately turning a prop into state.

mailpoet/lib/Automation/Engine/API/WorkflowResponse.php
```diff
@@ -0,0 +1,41 @@
+<?php declare(strict_types = 1);
```
johnolek: This whole thing is probably unnecessary. I thought it would be nice to avoid the duplication we have throughout the app though.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -30,11 +30,22 @@ public function createWorkflow(Workflow $workflow): int {
...
     return $id;
   }
 
+  public function duplicateWorkflow(Workflow $workflow): int {
+    $data = $workflow->toArray();
+    $now = (new DateTimeImmutable())->format(DateTimeImmutable::W3C);
+    $data['created_at'] = $now;
+    $data['updated_at'] = $now;
+    $data['status'] = Workflow::STATUS_DRAFT;
+    $data['name'] = 'Copy of ' . $workflow->getName();
```
johnolek: This has the potential to cause an error for workflows with names close to the 255 limit

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -131,6 +142,39 @@ public function getActiveWorkflowsByTrigger(Trigger $trigger): array {
     }, (array)$data);
   }
 
+  public function deleteWorkflow(Workflow $workflow): void {
```
johnolek, line 136: I struggled with where this logic should live. On the one hand it doesn't feel right to have `WorkflowStorage` handle deleting runs and logs, but it also didn't seem appropriate to only have it take place in the API endpoint, because in an ideal world I think we would handle this type of thing directly at the database level by using foreign key constraints with cascading deletes, but since some of our users use MyISAM (and probably for other reasons) that's not a viable option.

I'm totally open to a different approach though! I thought about using some actions, like `beforeDeleteWorkflow` or something that would allow other parts of the application to respond to deletion events, but I wasn't sure where we would actually put the logic for deleting related records.

mailpoet/assets/js/src/automation/automation.tsx
JanJakes: I generally don't know how exactly we use the notices in the listing now, but I think the ideal would be to switch to `@wordpress/notices` and use their store. An example here: https://github.com/mailpoet/mailpoet/blob/2c159b16948de91dbc33f1320aa85a961b95a41c/mailpoet/assets/js/src/automation/editor/api-error-handler.tsx#L4

It doesn't need our store to be set up, we just need to render the notices somewhere, such as https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/automation/editor/components/notices/index.tsx.

But I don't know the current system, really. Can we reasonably easily do that?

As for the `addNotice`, that's weird. If `notices` is a dependency of the `useCallback`, then it should work, I guess. Maybe even safer is to use the prev state: `setNotices((prevState) => [notice, ...prevState])`, that should never mismatch them.

mailpoet/assets/js/src/automation/automation.tsx
JanJakes: Is it not possible to use `noticeErrorHandler` directly? The `useCallback` memoizes it. The global variable would cause problems if such a component is instantiated multiple times (which here it isn't but it's still a side-effect, I guess).

Now I realize what's the problem... you have a "local" state but want to use it in a middleware outside of the React tree... that probably won't work correctly and it may be source of some of the mentioned problems. Then the notices store would definitely help.

mailpoet/assets/js/src/automation/automation.tsx
JanJakes: Would be great to consolidate them to `@wordpress/notices`. Some may need params from URL because they're being rendered after a redirect though.

mailpoet/assets/js/src/automation/listing/components/cells/more.tsx
```diff
@@ -1,29 +1,250 @@
...
+  workflow,
+  previousStatus,
+  onComplete,
+}: UndoButtonProps): JSX.Element {
+  const [clicked, setClicked] = useState(false);
+  const [isBusy, setIsBusy] = useState(false);
+  return (
+    <Button
+      isLink
+      disabled={clicked}
```
JanJakes: Is it possible to use `disabled={isBusy}` and get rid of `clicked` state?

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: It is suspicious because it's only the state's default value. When the prop changes, it never mirrors that change to the state. You'd need to add `useEffect(() => setAllWorkflows(workflows), [workflows])` as well. Sometimes things like this may be needed, I guess.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: Yeah, I'll let you decide if you want to replace it with `@wordpress/data` now or if we should do it in another ticket. Now we don't have to hurry that much for the MVP deadline.

mailpoet/lib/Automation/Engine/API/WorkflowResponse.php
JanJakes: We absolutely need something like this! I was thinking about creating `Mappers` but this looks also nice. Not sure which is better. We may need to reuse some smaller units (such as building only workflow steps) so maybe some more granular mappers will be better than full responses.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
JanJakes: Should we increase the limit? Or truncate? I think 255 is fine but not sure if it's bytes or UTF8 characters...

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
JanJakes: I think it's fine, what can we do... We could inject other storages and reuse them but I'm not sure if a storage should reuse another one. We may end up with cyclic references.

mailpoet/assets/js/src/automation/listing/components/cells/more.tsx
johnolek: The idea was to try to limit people to only using the button once. I suppose we could just not render the button after it's been used, or maybe it would be even better to completely remove the notice after the undo operation completes.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
johnolek: I think 255 is fine as well, and is likely much more than anyone will ever use. I'm still going to truncate to 255 characters though, because I don't want to leave the possibility of an API error here that we could easily prevent.

mailpoet/assets/js/src/automation/automation.tsx
johnolek: I'm switching to using a store, but I'm not sure we can use `@wordpress/notices` without it looking weird because those notices use Gutenberg styles. Please let me know if you still think we should try to use them, though.

mailpoet/lib/Automation/Engine/API/WorkflowResponse.php
johnolek: I think some kind of `Mappers` would probably be nice. Maybe we could even set them up in such a way that API requests can include which data they want in the response, potentially speeding things up. For now I'm thinking I might revert this change in large part because it doesn't seem worth it to change premium just for this.

mailpoet/assets/js/src/automation/automation.tsx
```diff
@@ -129,8 +123,32 @@ function App(): JSX.Element {
...
+            void dispatch(storeName).addNotice(
+              <Notice type="error" closable timeout={false} key={Date.now()}>
+                <p>{(error as ApiError).message}</p>
+              </Notice>,
+            );
+          }
+          throw error;
+        }
+      },
+    );
```
johnolek: This should be extracted

mailpoet/assets/js/src/automation/listing/store/reducer.ts
```diff
@@ -0,0 +1,52 @@
...
+      };
+    case 'REMOVE_WORKFLOW':
+      return {
+        ...state,
+        workflows: state.workflows.filter(
+          (workflow: Workflow) => workflow.id !== action.workflow.id,
+        ),
+        workflowCount: state.workflows.filter(
+          (workflow: Workflow) => workflow.id !== action.workflow.id,
+        ).length,
```
johnolek: Is there a better way to handle this, like having REMOVE_WORKFLOW somehow trigger SET_WORKFLOW_COUNT? I'm repeating myself here because there was an eslint warning about declaring variables inside a case statement.

mailpoet/lib/AdminPages/Pages/Automation.php
```diff
@@ -36,6 +42,7 @@ public function render() {
         'root' => rtrim($this->wp->escUrlRaw($this->wp->restUrl()), '/'),
         'nonce' => $this->wp->wpCreateNonce('wp_rest'),
       ],
+      'workflowCount' => count($this->workflowStorage->getWorkflows()),
```
johnolek: I thought it would be worthwhile to have the count immediately on page load so we don't need a "loading" state before deciding to render either the onboarding component or the listing itself.  If that seems like a worthy cause I would add a method to workflowStorage to retrieve a count directly from the DB instead of loading them all into memory like this.

johnolek:
@JanJakes I would appreciate it if you could take a look at how I've started implementing a store for the listing page. There are still some other things that I need to take care of which is why I'm keeping this a draft, but it would be great to know if I need to change course at all.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -26,15 +26,39 @@ public function __construct() {
...
     return $id;
   }
 
+  public function duplicateWorkflow(Workflow $workflow): int {
+    $data = $workflow->toArray();
+    $now = (new DateTimeImmutable())->format(DateTimeImmutable::W3C);
+    $data['created_at'] = $now;
+    $data['updated_at'] = $now;
+    $data['status'] = Workflow::STATUS_DRAFT;
+    $prefix = 'Copy of ';
```
johnolek: needs translation

mailpoet/lib/AdminPages/Pages/Automation.php
JanJakes: @johnolek That's good. We can also consider inlining ("preloading") the first page here fully but this is fine now.

mailpoet/lib/Automation/Engine/Data/Workflow.php
```diff
@@ -191,8 +191,8 @@ public static function fromArray(array $data): self {
       }, Json::decode($data['steps'])),
       new \WP_User((int)$data['author'])
     );
-    $workflow->id = (int)$data['id'];
-    $workflow->versionId = (int)$data['version_id'];
+    $workflow->id = (int)($data['id'] ?? 0);
+    $workflow->versionId = (int)($data['version_id'] ?? 0);
```
JanJakes: A bit of a note for myself when looking at this. The change seems a bit dangerous. The `fromArray` is meant to be used when fetching data from the DB, and both ID and version ID should always be set. This seems to kind of silently convert it to `0`, which is invalid.

I guess that's related to https://github.com/mailpoet/mailpoet/pull/4376/files#diff-03a583fe00e732caecf03063cb72efe77e089b482fefee6934eaf01e9bcde84eR58, will have a look.

mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDeleteEndpoint.php
```diff
@@ -0,0 +1,42 @@
...
+
+  public function handle(Request $request): Response {
+    $workflowId = $request->getParam('id');
+    if (!is_int($workflowId)) {
+      throw InvalidStateException::create();
+    }
+    $existingWorkflow = $this->workflowStorage->getWorkflow($workflowId);
+    if (!$existingWorkflow instanceof Workflow) {
+      throw InvalidStateException::create();
+    }
```
JanJakes: TODO: Check exceptions/messages.

mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDeleteEndpoint.php
```diff
@@ -0,0 +1,42 @@
...
+
+use MailPoet\API\REST\Request;
+use MailPoet\API\REST\Response;
+use MailPoet\Automation\Engine\API\Endpoint;
+use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Exceptions\InvalidStateException;
+use MailPoet\Automation\Engine\Storage\WorkflowStorage;
+use MailPoet\Validator\Builder;
+
+class WorkflowsDeleteEndpoint extends Endpoint {
```
JanJakes, line 11: Note to self, maybe endpoint naming should be (singular vs. plural):

```bash
/workflows      WorkflowsGet/WorkflowsPost/...
/workflows/id   WorkflowPut/WorkflowDelete/...
```

mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDuplicateEndpoint.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+    $existingWorkflow = $this->workflowStorage->getWorkflow($workflowId);
+    if (!$existingWorkflow instanceof Workflow) {
+      throw InvalidStateException::create();
+    }
+    $duplicateId = $this->workflowStorage->duplicateWorkflow($existingWorkflow);
+    $duplicate = $this->workflowStorage->getWorkflow($duplicateId);
+    if (!$duplicate instanceof Workflow) {
+      throw InvalidStateException::create();
+    }
```
JanJakes: TODO: Check exceptions/messages.

mailpoet/lib/Automation/Engine/Engine.php
```diff
@@ -73,6 +75,8 @@ private function registerApiRoutes(): void {
       $api->registerGetRoute('workflows', WorkflowsGetEndpoint::class);
       $api->registerPutRoute('workflows/(?P<id>\d+)', WorkflowsPutEndpoint::class);
       $api->registerPostRoute('workflows/create-from-template', WorkflowsCreateFromTemplateEndpoint::class);
+      $api->registerPostRoute('workflows/(?P<id>\d+)/duplicate', WorkflowsDuplicateEndpoint::class);
+      $api->registerDeleteRoute('workflows/(?P<id>\d+)/delete', WorkflowsDeleteEndpoint::class);
```
JanJakes: The path should be only `workflows/(?P<id>\d+)`, I hope it works (that it doesn't override the POST, for instance).

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -26,15 +26,39 @@ public function __construct() {
...
+      : 255;
+    if (strlen($newName) > $maxLength) {
+      $truncateWith = '…';
+      $truncationLength = strlen($truncateWith);
+      $newName = substr($newName, 0, $maxLength - $truncationLength) . $truncateWith;
+    }
+    $data['name'] = $newName;
+    $duplicate = Workflow::fromArray($data);
+    return $this->createWorkflow($duplicate);
+  }
```
JanJakes: I'm wondering if duplicating a workflow should also change step IDs. It's not necessary (we never said they are universally unique) but it might be clearer and better for debugging, maybe 🤔

JanJakes (commented):
Just submitting the first batch of notes for myself (for the PHP part). Feel free to check them or respond as well, @johnolek, but feel free to ignore them, too.

I may split this into two smaller PRs (PHP & the fronted part), will see.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
JanJakes: One more TODO: the workflow author should now probably be the person who duplicated it.

mailpoet/lib/Automation/Engine/Endpoints/Workflows/WorkflowsDeleteEndpoint.php
JanJakes: Or `WorkflowsIdPut`... not so important now.

mailpoet/lib/Automation/Engine/Builder/DuplicateWorkflowController.php
```diff
@@ -0,0 +1,91 @@
...
+      throw new InvalidStateException('Workflow not found.');
+    }
+    return $savedWorkflow;
+  }
+
+  private function getName(string $name): string {
+    // translators: %s is the original workflow name.
+    $newName = sprintf(__('Copy of %s', 'mailpoet'), $name);
+    $maxLength = $this->workflowStorage->getNameColumnLength();
+    if (strlen($newName) > $maxLength) {
```
JanJakes, line 54: Here, if we were talking Unicode, we should actually use `mb_strlen`. But I don't know for sure... likely, for that to be completely correct, the users would need to use `utf8mb4` DB encoding, but we have no control over that.

So still thinking if 255 chars or potentially bytes, is OK, or if we should increase the column size 🤔

mailpoet/lib/Automation/Engine/Builder/DeleteWorkflowController.php
```diff
@@ -0,0 +1,27 @@
...
+  /** @var WorkflowStorage */
+  private $workflowStorage;
+
+  public function __construct(
+    WorkflowStorage $workflowStorage
+  ) {
+    $this->workflowStorage = $workflowStorage;
+  }
+
+  public function deleteWorkflow(int $id): Workflow {
```
JanJakes, line 19: Should we allow deleting only workflows that are trashed? 🤔

mailpoet/lib/Automation/Engine/Builder/DuplicateWorkflowController.php
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * @param Step[] $steps
+   * @return Step[]
+  */
+  private function getSteps(array $steps): array {
+    $newIds = [];
+    foreach ($steps as $step) {
+      $id = $step->getId();
+      $newIds[$id] = $id === 'root' ? 'root' : $this->getId();
```
websupporter, line 69: Out of curiosity: Do we generate new IDs only because, they are IDs and should be unique or is there a material reason for it I do miss? Or, to rephrase: Isn't it enough to be sure the IDs are unique in the same workflow.

If it is about: Maybe some day, we want to extract steps into its own table, I think we would need to check the uniqueness at that point anyway (if we would want to use this `getID()`-string in a `unique` column, given `root`).

I just wonder, if we could get rid of all that logic and just use the original `Step[]`s, especially as we return `$savedWorkflow` anyway, so the consumer of this method will get a copy and not the original `Step` objects.

mailpoet/lib/Automation/Engine/Builder/DuplicateWorkflowController.php
```diff
@@ -0,0 +1,91 @@
...
+  public function duplicateWorkflow(int $id): Workflow {
+    $workflow = $this->workflowStorage->getWorkflow($id);
+    if (!$workflow) {
+      throw Exceptions::workflowNotFound($id);
+    }
+
+    $duplicate = new Workflow(
+      $this->getName($workflow->getName()),
+      $this->getSteps($workflow->getSteps()),
+      $this->wordPress->wpGetCurrentUser()
```
websupporter, line 38: I think that is a good move, but wanted to highlight, that we do not copy the author, but transfer the authorship (potentially later on maybe CRUD capabilities) to the current user.

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsDuplicateEndpointTest.php
```diff
@@ -0,0 +1,107 @@
...
+    wp_set_current_user(0);
+    $data = $this->post(sprintf(self::ENDPOINT_PATH, $this->workflow->getId()));
+
+    $this->assertSame([
+      'code' => 'rest_forbidden',
+      'message' => 'Sorry, you are not allowed to do that.',
+      'data' => ['status' => 401],
+    ], $data);
+
+    $workflow = $this->workflowStorage->getWorkflow($this->workflow->getId());
```
websupporter, line 47: Should we check that `$this->workflowStorage->getWorkflow($this-workflow->getID()+1) === null`?

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsDuplicateEndpointTest.php
```diff
@@ -0,0 +1,107 @@
...
+    $expected = [
+      'id' => $id,
+      'name' => 'Copy of Testing workflow',
+      'status' => 'draft',
+      'created_at' => $createdAt->format(DateTimeImmutable::W3C),
+      'updated_at' => $updatedAt->format(DateTimeImmutable::W3C),
+      'activated_at' => null,
+      'author' => [
+        'id' => $user->ID,
+        'name' => $user->display_name
```
websupporter: Is missing a `,` in the end I think

mailpoet/lib/Automation/Engine/Mappers/WorkflowMapper.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace MailPoet\Automation\Engine\Mappers;
+
+use DateTimeImmutable;
+use MailPoet\Automation\Engine\Data\NextStep;
+use MailPoet\Automation\Engine\Data\Step;
+use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Storage\WorkflowStatisticsStorage;
+
+class WorkflowMapper {
```
websupporter, line 12: Maybe we should extract `WorkflowsGetEndpoint::buildWorkflow()` as well.

Under REST point of view, I would think `steps` should be also in `WorkflowsGetEndpoint`, but not so important right now.

Maybe a `public function buildReducedWorkflow(Workflow $workflow, WorkflowStatistics $statistics = null)`.

It makes sense to me, to consolidate this task in one class.

mailpoet/lib/Automation/Engine/Mappers/WorkflowMapper.php
```diff
@@ -0,0 +1,47 @@
...
+  /** @var WorkflowStatisticsStorage */
+  private $statisticsStorage;
+
+  public function __construct(
+    WorkflowStatisticsStorage $statisticsStorage
+  ) {
+    $this->statisticsStorage = $statisticsStorage;
+  }
+
+  public function buildWorkflow(Workflow $workflow): array {
```
websupporter, line 22: I think AdminPages/Pages/AutomationEditor.php could use this method as well.

mailpoet/lib/Automation/Engine/Builder/DuplicateWorkflowController.php
JanJakes: @websupporter I added a [note to myself](https://github.com/mailpoet/mailpoet/pull/4376#discussion_r994535062) when going through the code:

> I'm wondering if duplicating a workflow should also change step IDs. It's not necessary (we never said they are universally unique) but it might be clearer and better for debugging, maybe 🤔

You're right that this is not necessary, but my thinking was more like "just to be sure" not to mix up those two workflows somewhere together and make it clear for debugging, etc. Essentially, I wanted to prevent making duplicate workflow IDs explicitly (although we don't consider them universally unique).

mailpoet/lib/Automation/Engine/Builder/DuplicateWorkflowController.php
JanJakes: Yeah, [I thought this would be the correct behavior](https://github.com/mailpoet/mailpoet/pull/4376#discussion_r994554485). The person who duplicates a workflow is the one who creates the new one, I guess.

mailpoet/lib/Automation/Engine/Builder/DuplicateWorkflowController.php
websupporter: Ok :+1:

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsDuplicateEndpointTest.php
JanJakes: Good point, added: https://github.com/mailpoet/mailpoet/pull/4376/commits/4785813d508341e88459060b26798dee1328632b

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsDuplicateEndpointTest.php
JanJakes: 👍 https://github.com/mailpoet/mailpoet/pull/4376/commits/745790643498f06d4eb940b3c74e02ae4f090b58

mailpoet/lib/Automation/Engine/Mappers/WorkflowMapper.php
JanJakes: Yes! Implemented in: https://github.com/mailpoet/mailpoet/pull/4376/commits/03f9750aebf0e8b88ed8c3ff8303abfea88a7380

> Under REST point of view, I would think steps should be also in WorkflowsGetEndpoint, but not so important right now.

This is actually fine. We've been dealing with these issues in the past, and what seems to allow this difference is that these are different REST resources, in fact:

```bash
/workflows
/workflows/<id>
```

The former one always returns a list (can have filters, paging, etc.), and that can be a reduced item, the latter one always returns a full single `Workflow` resource, no matter the method (GET, POST, PUT, ...). Thinking about this, we seem to return `null` for `DELETE`. We could as well return the just deleted workflow, I guess...

mailpoet/lib/Automation/Engine/Mappers/WorkflowMapper.php
JanJakes: Yes yes! Great catch, done: https://github.com/mailpoet/mailpoet/pull/4376/commits/d16034def1e6ce82453d73a95c838a6cbc36a5df

mailpoet/lib/Automation/Engine/Mappers/WorkflowMapper.php
websupporter: Can we return a `204 No Content` for the `DELETE` endpoint? That would solve the question what status the returned workflow should have.

JanJakes: @websupporter Our API will always return something like `{"data": null}`, I think. It kind of is a "content," but I guess combining that with 204 would do no harm...? Returning the just-deleted entity is also an option, are there any use cases where that could be useful? E.g. for the UI to render something after the deletion? 🤔

(In any case I'd add it to the checklist rather than appending it here.)

websupporter: > E.g. for the UI to render something after the deletion

Thinking about it; yes, maybe something like a notice `Workflow "<name>" has been deleted`

> In any case I'd add it to the checklist rather than appending it here

:+1:

JanJakes: :+1: I agree, I think usually, you'll probably have the name just before deleting, but yeah, this could facilitate some UIs. Added to the checklist.

JanJakes:
@websupporter Thanks for the review, all is addressed, hopefully.

websupporter (approved):
lgtm. thank you

PR #4352, opened 2022-09-08
Update WooCommerce packages to follow the rule L-2 [MAILPOET-4615]

pavel-mailpoet:
[MAILPOET-4615]

[MAILPOET-4615]: https://mailpoet.atlassian.net/browse/MAILPOET-4615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4383, opened 2022-09-23
Fix nightly job and some flaky acceptance tests [MAILPOET-4662]

costasovo:
## Description
This PR fixes issues in some nightly jobs (**acceptance_oldest** and **acceptance_with_premium_latest** ) for the plugin and also attempts to fix the flakiness of a couple of tests.

My journey was full of surprises I can't keep to myself, and I hope it helps to explain the changes I made.

1. We were [using versions of WooCommerce and WooCommerce Blocks](https://github.com/mailpoet/mailpoet/pull/4352) that were not compatible with the WordPress version (5.6) we use in acceptance_oldest.  I [discussed the issue on slack](https://a8c.slack.com/archives/C01GH835U90/p1663760637325809), and the outcome was that until we update the minimal WP version to 5.8, we will use the plugins at the latest version that run on WP 5.6.
2. I found that after we switched to monorepo, the premium plugin was actually not really installed for any of the tests that are supposed to run with it.
3. After fixing the installation of the premium plugin, I ran into [an issue with the missing Doctrine Annotation Reader classes](https://output.circle-artifacts.com/output/job/67598f40-736b-452a-bbe5-9c3276a92bd1/artifacts/4/tests/_output/exceptions/error--2022-09-22--08-21--eef47c1f18.html). We delete those classes in the acceptance test to emulate the state in the production build. But we did delete those classes after plugin activation, which caused that there was the logged exception for an ajax request done by the action scheduler. We deleted those files during the request.
4. After fixing the issue with Annotation classes, I found that premium tests were failing [with some strange exception from WooCommerce subscription plugin](https://output.circle-artifacts.com/output/job/839aeccc-cac4-4d3d-b4be-08449e8f6495/artifacts/15/tests/_output/exceptions/error--2022-09-22--11-11--adf31beb0a.html). After debugging I found that this is related to how docker mounts volumes which caused that WP get_plugins function that is looking for plugins in the plugins directory ignored mailpoet-premium plugin. As a result, the premium plugin was listed in active plugins but not in installed plugins, and that was causing the warning (see [the surrounding code in the exception log](https://output.circle-artifacts.com/output/job/839aeccc-cac4-4d3d-b4be-08449e8f6495/artifacts/15/tests/_output/exceptions/error--2022-09-22--11-11--adf31beb0a.html)). I fixed that by mounting the volume to a different location and copying it to the plugins folder.  It is probably not the best solution, but it works.
5. After fixing the strange exception, the premium tests started to fail because the twig cache couldn't be written. We run the acceptance tests in dev mode with Tracy so that we can log errors nicely. As a side-effect, the Twig is configured to use `auto_reload`, which requires a writable `generated` folder (which was not writable because of what I did in 4)). I fixed that by adding write permissions to the folder.
6. I fixed 🤞 a couple of flaky tests I ran into. I tried to explain the fixes in commit messages.

## Code review notes
Here is a build where I also run the problematic jobs: https://app.circleci.com/pipelines/github/mailpoet/mailpoet/11417/workflows/a0df72e1-967c-4ad5-bf47-e668ec08dacc
Note: Some tests remain flaky, but I hope that overall the situation should get better after we merge these changes.

## QA notes
The only change in the production code is that I [added the automation-id data parameter to the name and description fields in the dynamic segment form]. This was needed to make a test more stable and should be an invisible change for users. (https://github.com/mailpoet/mailpoet/commit/855ed21564bea23735cf11e0306f8cf475d91253#diff-0239ab9f86833b2f6e4a75a5f66ff933b8243e8ab8168e157f3d5b684265a4c9)

## Linked tickets
[MAILPOET-4662]

[MAILPOET-4662]: https://mailpoet.atlassian.net/browse/MAILPOET-4662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


websupporter (approved):
:1st_place_medal: 

Thank you for digging into this s***t! Amazing work. LGTM.

veljkho (approved):

PR #4384, opened 2022-09-23
Update Black Friday notice [MAILPOET-4670]

costasovo:
## Description
This PR updates the Black Friday notice for the current year.

## Code review notes
The specification mentions: `Make sure the banner will be shown even for users who dismissed the last year’s Black Friday banner.`
The information about the dismissal of the banner was saved with 30 days expiration, the banner was displayed a year ago, so for this year, we should be fine.

## QA notes
For displaying the banner you need to change the system date, or maybe it would be easier [to edit the dates in the code](https://github.com/mailpoet/mailpoet/commit/e4f2761fbc7841dce2e67b8b575f60e429a73246#diff-53170429e3e13f729d6b3cc61518ccf33dd42d3fa49bf4333c0b653d0dec2c5bR25-R26).

## Linked tickets
[MAILPOET-4670]

[MAILPOET-4670]: https://mailpoet.atlassian.net/browse/MAILPOET-4670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Util/Notices/BlackFridayNotice.php
+++ b/mailpoet/lib/Util/Notices/BlackFridayNotice.php
@@ -22,8 +22,8 @@ class BlackFridayNotice {
 
   public function init($shouldDisplay) {
     $shouldDisplay = $shouldDisplay
-      && (time() <= strtotime('2021-11-30 12:00:00'))
-      && (time() >= strtotime('2021-11-24 12:00:00'))
+      && (time() <= strtotime('2022-10-07 12:00:00'))
+      && (time() >= strtotime('2022-10-03 12:00:00'))
       && !get_transient(self::OPTION_NAME);
     if ($shouldDisplay) {
       $this->display();
@@ -32,8 +32,8 @@ class BlackFridayNotice {
 
   private function display() {
     $subscribers = $this->subscribersRepository->countBy(['deletedAt' => null]);
-    $header = '<h3 class="mailpoet-h3">' . __('Save big on MailPoet – 40% off this Black Friday', 'mailpoet') . '</h3>';
-    $body = '<h5 class="mailpoet-h5">' . __('Our biggest ever sale is here! Save 40% on all annual plans and licenses until 8 am UTC, November 30. Terms and conditions apply.', 'mailpoet') . '</h5>';
+    $header = '<h3 class="mailpoet-h3">' . __('Get ready for Black Friday with 40% off MailPoet plans', 'mailpoet') . '</h3>';
+    $body = '<h5 class="mailpoet-h5">' . __('Save 40% on all annual plans and licenses until 2 pm UTC, 7 October. Terms & conditions apply.', 'mailpoet') . '</h5>';
     $link = "<p><a href='https://account.mailpoet.com/?s=$subscribers' class='mailpoet-button button-primary' target='_blank'>"
       . __('Shop now', 'mailpoet')
       . '</a></p>';
--- a/mailpoet/lib/Util/Notices/BlackFridayNotice.php
+++ b/mailpoet/lib/Util/Notices/BlackFridayNotice.php
@@ -22,8 +22,8 @@ class BlackFridayNotice {
 
   public function init($shouldDisplay) {
     $shouldDisplay = $shouldDisplay
-      && (time() <= strtotime('2022-10-07 12:00:00'))
-      && (time() >= strtotime('2022-10-03 12:00:00'))
+      && (time() >= strtotime('2022-10-03 14:00:00 UTC'))
+      && (time() <= strtotime('2022-10-07 14:00:00 UTC'))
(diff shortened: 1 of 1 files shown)


NeosinneR:
@veljkho just a note, please wait with merging, as we are still confirming with Marketing the exact sentences. They likely won't change, but just to be sure. But feel free to review everything else 😉

mailpoet/lib/Util/Notices/BlackFridayNotice.php
```diff
@@ -32,9 +32,9 @@ public function init($shouldDisplay) {
 
   private function display() {
     $subscribers = $this->subscribersRepository->countBy(['deletedAt' => null]);
-    $header = '<h3 class="mailpoet-h3">' . __('Save big on MailPoet – 40% off this Black Friday', 'mailpoet') . '</h3>';
-    $body = '<h5 class="mailpoet-h5">' . __('Our biggest ever sale is here! Save 40% on all annual plans and licenses until 8 am UTC, November 30. Terms and conditions apply.', 'mailpoet') . '</h5>';
-    $link = "<p><a href='https://account.mailpoet.com/?s=$subscribers' class='mailpoet-button button-primary' target='_blank'>"
+    $header = '<h3 class="mailpoet-h3">' . __('Get ready for Black Friday with 40% off MailPoet plans', 'mailpoet') . '</h3>';
+    $body = '<h5 class="mailpoet-h5">' . __('Save 40% on all annual plans and licenses until 2 pm UTC, October 7. Terms & conditions apply.', 'mailpoet') . '</h5>';
+    $link = "<p><a href='https://account.mailpoet.com/?s=$subscribers&billing=yearly' class='mailpoet-button button-primary' target='_blank'>"
       . __('Shop now', 'mailpoet')
```
johnolek: To match the specs we should capitalize both words and make this `Shop Now`

mailpoet/lib/Util/Notices/BlackFridayNotice.php
```diff
@@ -22,8 +22,8 @@ public function __construct(
 
   public function init($shouldDisplay) {
     $shouldDisplay = $shouldDisplay
-      && (time() <= strtotime('2021-11-30 12:00:00'))
-      && (time() >= strtotime('2021-11-24 12:00:00'))
+      && (time() >= strtotime('2022-10-03 12:00:00'))
+      && (time() <= strtotime('2022-10-07 12:00:00'))
```
johnolek: I've always been bad when it comes to thinking about time and timezones, so forgive me in advance if this comment doesn't make sense and while I think through it all in this comment.

Should we include UTC in the `strtotime` calls? If we don't then PHP will [default to using whatever the default timezone is](https://www.php.net/manual/en/function.strtotime.php), right? If I had my PHP's default timezone set to UTC-5, then this would evaluate to 

```php
strtotime('2022-10-07 12:00:00 -5:00')
> 1665162000
```

Whereas with UTC we get
```php
strtotime('2022-10-07 12:00:00 UTC')
> 1665144000
```

Since `time()` will always return UTC timestamps isn't it possible we would end up showing this notice after it's valid?

`time() <= strtotime('2022-10-07 12:00:00 America/Chicago')` should be equivalent to `time() <= strtotime('2022-10-07 17:00:00 UTC')`, so I think I could end up seeing this notice for 3 hours after the deal expires (again, only if my default PHP timezone were different -- strtotime doesn't care about the WP timezone).

johnolek (changes requested):
It's very possible that I'm incorrect about the time calculation comment!

mailpoet/lib/Util/Notices/BlackFridayNotice.php
NeosinneR: @johnolek for some reason I thought it's harder to work with timezones in PHP, so that's why I suggested a timezone-less approach with 12 hours of reserve on both ends. While officially the sale is from and to 2 pm UTC, the shop coupons operate on days, not hours, so the discount will be applied from midnight to midnight UTC, which should pretty much work out so everyone seeing the banner should be able to purchase it.

Nevertheless, I like your suggestion, so I changed it so it's visible during the official sale period from and to 2 pm UTC. Thanks!

(I made the change instead of Rosta, because I only now got the tracking parameters, which I added)

johnolek: Ahh, that makes sense. Somehow I missed/didn't realize that shop coupons are day based. Thanks for the explanation and the change!

johnolek (approved):
LGTM!

costasovo:
Thanks @NeosinneR for updating the PR!

veljkho (approved):

PR #4386, opened 2022-09-23
Remove old model usage from MailPoet\Cron\Workers\SendingQueue\Tasks\Newsletter - Part 2 [MAILPOET-4680]

rodrigoprimo:
## Description

This PR is a continuation of #4336 and it removes the remaining models from `MailPoet\Cron\Workers\SendingQueue\Tasks\Newsletter` and its test class.

## Code review notes

Please note that I opted to leave a call to the `SendingQueue` model in Newsletter.php: https://github.com/mailpoet/mailpoet/blob/ccbeb74a77da9cebe0ff82dc7c8cce5f7bc6cf94/mailpoet/lib/Cron/Workers/SendingQueue/Tasks/Newsletter.php#L203

I decided to do this as this call is tied to how the `Sending` class works and this class still uses the old `SendingQueue` model. Replacing the model in Newsletter.php with the corresponding entity without doing the same in `Sending` could cause inconsistencies as the same code path also uses the model within `Sending`. My suggestion is that we remove this remaining model from Newsletter.php in the ticket where we will address `Sending`.

Please let me know if you agree, and I will add a note to the corresponding ticket. 

## QA notes

This PR refactors code in a key part of MailPoet: the part responsible for sending e-mails. Please test sending different emails and make sure that they are delivered and that the subject and the body of the mails match what is expected.

## Linked PRs

#4336

## Linked tickets

[MAILPOET-4680]

## After-merge notes

This PR deprecates a filter and creates a new one. We should probably notify the users and update the documentation. See commit ccbeb74a77da9cebe0ff82dc7c8cce5f7bc6cf94 for more details.

[MAILPOET-4680]: https://mailpoet.atlassian.net/browse/MAILPOET-4680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -159,14 +159,9 @@ class SendingQueue {
       return;
     }
 
-    $newsletter = Newsletter::findOne($newsletterEntity->getId());
-    if (!$newsletter) {
-      return;
-    }
-
     // pre-process newsletter (render, replace shortcodes/links, etc.)
-    $newsletter = $this->newsletterTask->preProcessNewsletter($newsletter, $queue);
-    if (!$newsletter) {
+    $newsletterEntity = $this->newsletterTask->preProcessNewsletter($newsletterEntity, $queue);
+    if (!$newsletterEntity) {
       $this->loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->info(
         'delete task in sending queue',
         ['task_id' => $queue->taskId]
@@ -174,6 +169,12 @@ class SendingQueue {
       $queue->delete();
       return;
     }
+
+    $newsletter = Newsletter::findOne($newsletterEntity->getId());
+    if (!$newsletter) {
+      return;
+    }
+
     // clone the original object to be used for processing
     $_newsletter = (object)$newsletter->asArray();
     $_newsletter->options = $newsletterEntity->getOptionsAsArray();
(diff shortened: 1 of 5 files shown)


lysyjan (commented):
> My suggestion is that we remove this remaining model from Newsletter.php in the ticket where we will address Sending.

I am fine with it.

I can see several usages of MailPoet\Tasks\Sending. This class depends on the old model, and we should remove it too. However, this can be a more significant step, so we have to decide about the approach. Please, let me know what you prefer.

Nevertheless, the current changes look good to me.

rodrigoprimo:
@lysyjan, my suggestion would be that we move forward with this PR and address `MailPoet\Tasks\Sending` in a different PR/ticket.

lysyjan (approved):
@rodrigoprimo I'm approving it. Please, create a follow-up ticket before you move this PR forward. I would like us not to forget.

rodrigoprimo:
> Please, create a follow-up ticket before you move this PR forward. I would like us not to forget.

@lysyjan, what I have in mind is to address as part of https://mailpoet.atlassian.net/browse/MAILPOET-4375, if you agree, maybe we can just add a note to this ticket? Or do you reckon we should address it in a separate ticket?

lysyjan:
@rodrigoprimo I would address it in a separate ticket. It should be better for testing. Also, smaller tickets are more comfortable for developers, I guess.

rodrigoprimo:
@lysyjan, I'm unsure how easy it will be to address this in a separate ticket, but I guess it is worth trying. Here is the ticket: https://mailpoet.atlassian.net/browse/MAILPOET-4693.

Aschepikov (approved):
QA verified, good job

PR #4388, opened 2022-09-26
Update insertSubscribersFromOrders() query to work with WooCommerce Custom Orders Tables [MAILPOET-4577]

rodrigoprimo:
## Description

This PR updates the method `\MailPoet\Segments\WooCommerce::insertSubscribersFromOrders()` to use a different query if WooCommerce Custom Orders Tables is enabled. 

## Code review notes

_N/A_

## QA notes

Install MailPoet in a WooCommerce shop that already has a few orders placed by guest customers. Configure MailPoet to synchronize WooCommerce customers as MailPoet subscribers and check that subscribers are created for the WooCommerce guest customers. The functionality should continue working as expected with WooCommerce Custom Orders Table disabled (which is the default behavior). It still doesn't work with Woo COT enabled but that will be fixed in a subsequent PR when https://mailpoet.atlassian.net/browse/MAILPOET-4569 is implemented.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4577]

## After-merge notes

_N/A_

[MAILPOET-4577]: https://mailpoet.atlassian.net/browse/MAILPOET-4577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/WooCommerce/Helper.php
+++ b/mailpoet/lib/WooCommerce/Helper.php
@@ -3,6 +3,7 @@
 namespace MailPoet\WooCommerce;
 
 use MailPoet\DI\ContainerWrapper;
+use MailPoet\RuntimeException;
 use MailPoet\WP\Functions as WPFunctions;
 
 class Helper {
@@ -117,7 +118,7 @@ class Helper {
 
   public function getOrdersTableName() {
     if (!method_exists('\Automattic\WooCommerce\Internal\DataStores\Orders\OrdersTableDataStore', 'get_orders_table_name')) {
-      throw new \Exception('Cannot get orders table name when running a WooCommerce version that doesn\'t support custom order tables.');
+      throw new RuntimeException('Cannot get orders table name when running a WooCommerce version that doesn\'t support custom order tables.');
     }
 
     return \Automattic\WooCommerce\Internal\DataStores\Orders\OrdersTableDataStore::get_orders_table_name();
--- a/mailpoet/lib/WooCommerce/Helper.php
+++ b/mailpoet/lib/WooCommerce/Helper.php
@@ -114,4 +114,12 @@ class Helper {
 
     return $installedViaWooCommerce;
   }
+
+  public function getOrdersTableName() {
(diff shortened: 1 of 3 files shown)


mailpoet/lib/WooCommerce/Helper.php
```diff
@@ -114,4 +114,12 @@ public function wasMailPoetInstalledViaWooCommerceOnboardingWizard(): bool {
 
     return $installedViaWooCommerce;
   }
+
+  public function getOrdersTableName() {
+    if (!method_exists('\Automattic\WooCommerce\Internal\DataStores\Orders\OrdersTableDataStore', 'get_orders_table_name')) {
+      throw new \Exception('Cannot get orders table name when running a WooCommerce version that doesn\'t support custom order tables.');
```
costasovo: We have defined [a bunch of MailPoet exception classes](https://github.com/mailpoet/mailpoet/blob/1f34f57df5e330ee4baf844b013b419c26a9a662/mailpoet/lib/exceptions.php). We haven't replaced all usages of the default PHP exceptions but for new ones we should use our classes. Could you please repace the exception?

costasovo (changes requested):
Looks good! I just want to ask you to use another exception class for the exception. See my other comment.

mailpoet/lib/WooCommerce/Helper.php
rodrigoprimo: Thanks, @costasovo. I agree that using a more specific and MailPoet exception is better. I'm thinking using `RuntimeException` is fine in this case as the more specific exceptions set a HTTP error code which doesn't make much sense in this context as far as I can see. What do you think?

costasovo: @rodrigoprimo, I think the Runtime exception is ok. It is a situation that should not happen.

rodrigoprimo: @costasovo, done. I added a new commit using `RuntimeException`. Could you please take a final look?

costasovo (approved):
Thanks! 👍

Aschepikov (approved):
Well done!

PR #4391, opened 2022-09-27
Newsletter template UI updates for automation [MAILPOET-4520]

websupporter:
## Description
Updates the newsletter templates page:
* Adds MailPoet logo, which redirects to the automation editor.
* Adds Cancel button, which disconnects the current email_id in the step, so that the next time we go to the send email settings, we see "Design email" button again, instead of the "Edit email" button.
* Removes the Cancel button on the newsletter editor (discussed in [Slack](https://a8c.slack.com/archives/C0313447YBC/p1664192125708849?thread_ts=1664189210.533199&cid=C0313447YBC)) (dc94b3bdec558708fd8608a2f17e7956864668fc)
* Removes Tutorial button in editor (cbed281c427588281ed6b0ac4c90a61d5f5d9c31)
* There was a small JS error in the send email settings, where we tried to create a DOM structure like `<p><div>text</div></p>`. I fixed this in 5d090727a437a5db37b949752164d5b58159bb58

## Code review notes
In order to disconnect the email_id in the send_email settings, I created an additional $_GET parameter which will be read before the AutomationEditor renders. The basic idea is: Removing this ID is the responsibility of the Step. The Step interface therefore got a `performActions(StepData $step, array $action)`. The `?steps[<stepId>][<action>]=<value>` parameter enables us now, to target the data of specific steps and let the `Step` take care of what to do in case of an `action`.

I preferred that approach over a simple `$_GET['delete_mail']`. It seemed a bit more flexible, so that 3rd party steps might be able to perform similar actions on the editor page as well.

## QA notes
These are the things, which needs testing from my perspective:
* Did the Cancel button disappear in the editor?
* Does the cancel button actually cancel the current action, return to the automation editor and I can return to the template screen through the "Design email" button in the "send email"-step?
* Do I get redirected to the automation Editor when clicking the MailPoet logo in the top bar on the templates and the newsletter editor?
* When I create a normal newsletter: Does the logo redirect to the newsletter listing page?

## Linked PRs
https://github.com/mailpoet/mailpoet-premium/pull/627

## Linked tickets
[MAILPOET-4520]

[MAILPOET-4520]: https://mailpoet.atlassian.net/browse/MAILPOET-4520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/design_email_button.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/design_email_button.tsx
@@ -49,7 +49,7 @@ export function DesignEmailButton(): JSX.Element {
   // and the workflow is saved, we can safely redirect to the email design flow.
   useEffect(() => {
     if (emailId && workflowSaved) {
-      window.location.href = `admin.php?page=mailpoet-newsletter-editor&id=${emailId}`;
+      window.location.href = `admin.php?page=mailpoet-newsletters#/template/${emailId}`;
     }
   }, [emailId, workflowSaved]);
 
--- a/mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
+++ b/mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
@@ -59,10 +59,7 @@ const getEmailSendTitle = (emailType: string): string => {
   return typeMap[emailType] || MailPoet.I18n.t('stepNameSend');
 };
 
-const tutorialIcon = (emailType: string): JSX.Element | null => {
-  if (emailType === 'automation') {
-    return null;
-  }
+function TutorialIcon(): JSX.Element {
   return (
     <div>
       <a
@@ -87,7 +84,7 @@ const tutorialIcon = (emailType: string): JSX.Element | null => {
       <span id="beamer-empty-element" />
     </div>
   );
-};
+}
 
 const stepsListingHeading = (
   step: number,
@@ -125,7 +122,7 @@ const stepsListingHeading = (
         {' '}
       </h1>
       <div className="mailpoet-flex-grow" />
-      {tutorialIcon(emailType)}
+      {emailType !== 'automation' && <TutorialIcon />}
     </div>
   );
 };
--- a/mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
+++ b/mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
@@ -59,6 +59,36 @@ const getEmailSendTitle = (emailType: string): string => {
   return typeMap[emailType] || MailPoet.I18n.t('stepNameSend');
 };
 
+const tutorialIcon = (emailType: string): JSX.Element | null => {
+  if (emailType === 'automation') {
+    return null;
+  }
+  return (
+    <div>
+      <a
+        role="button"
+        onClick={displayTutorial}
+        className="mailpoet-top-bar-beamer"
+        title={MailPoet.I18n.t('topBarTutorial')}
+        tabIndex={0}
+        onKeyDown={(event) => {
+          if (
+            ['keydown', 'keypress'].includes(event.type) &&
+            ['Enter', ' '].includes(event.key)
+          ) {
+            event.preventDefault();
+            displayTutorial();
+          }
+        }}
+      >
+        <Icon icon={video} />
+        <span>{MailPoet.I18n.t('topBarTutorial')}</span>
+      </a>
+      <span id="beamer-empty-element" />
+    </div>
+  );
+};
+
 const stepsListingHeading = (
   step: number,
   emailType: string,
@@ -80,6 +110,7 @@ const stepsListingHeading = (
       MailPoet.I18n.t('stepNameDesign'),
     ];
   }
+
   return (
     <div className="mailpoet-top-bar" data-automation-id={automationId}>
       {showMailPoetLogo && <MailPoetLogoResponsive />}
@@ -94,28 +125,7 @@ const stepsListingHeading = (
         {' '}
       </h1>
       <div className="mailpoet-flex-grow" />
-      <div>
-        <a
-          role="button"
-          onClick={displayTutorial}
-          className="mailpoet-top-bar-beamer"
-          title={MailPoet.I18n.t('topBarTutorial')}
-          tabIndex={0}
-          onKeyDown={(event) => {
-            if (
-              ['keydown', 'keypress'].includes(event.type) &&
-              ['Enter', ' '].includes(event.key)
-            ) {
-              event.preventDefault();
-              displayTutorial();
-            }
-          }}
-        >
-          <Icon icon={video} />
-          <span>{MailPoet.I18n.t('topBarTutorial')}</span>
-        </a>
-        <span id="beamer-empty-element" />
-      </div>
+      {tutorialIcon(emailType)}
     </div>
   );
 };
--- a/mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
+++ b/mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
@@ -93,8 +93,8 @@ const stepsListingHeading = (
   step: number,
   emailType: string,
   automationId: string,
-  showMailPoetLogo: boolean,
   buttons: ReactNode,
+  onLogoClick?: () => void,
 ): JSX.Element => {
   const emailTypeTitle = getEmailTypeTitle(emailType);
   let stepTitles = [
@@ -113,7 +113,7 @@ const stepsListingHeading = (
 
   return (
     <div className="mailpoet-top-bar" data-automation-id={automationId}>
-      {showMailPoetLogo && <MailPoetLogoResponsive />}
+      <MailPoetLogoResponsive onClick={onLogoClick} />
       <HideScreenOptions />
       <Steps count={stepTitles.length} current={step} titles={stepTitles} />
       {buttons && (
@@ -135,7 +135,7 @@ export interface Props {
   emailType?: string;
   automationId?: string;
   location: Location;
-  showMailPoetLogo?: boolean;
+  onLogoClick?: () => void;
   buttons?: ReactNode;
 }
 
@@ -144,17 +144,18 @@ function ListingHeadingSteps({
   emailType,
   location,
   automationId,
-  showMailPoetLogo,
   buttons,
+  onLogoClick,
 }: Props): JSX.Element {
   const stepNumber = step || mapPathToSteps(location, emailType);
+
   if (stepNumber !== null) {
     return stepsListingHeading(
       stepNumber,
       emailType,
       automationId,
-      showMailPoetLogo,
       buttons,
+      onLogoClick,
     );
   }
   return null;
--- a/mailpoet/lib/AdminPages/Pages/AutomationEditor.php
+++ b/mailpoet/lib/AdminPages/Pages/AutomationEditor.php
@@ -7,6 +7,7 @@ use MailPoet\AdminPages\PageRenderer;
 use MailPoet\Automation\Engine\Data\NextStep;
 use MailPoet\Automation\Engine\Data\Step;
 use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Hooks;
 use MailPoet\Automation\Engine\Registry;
 use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Segments\SegmentsRepository;
@@ -45,6 +46,8 @@ class AutomationEditor {
 
   public function render() {
     $id = isset($_GET['id']) ? (int)$_GET['id'] : null;
+
+    $this->wp->doAction(Hooks::EDITOR_BEFORE_LOAD, (int)$id);
 
     $workflow = $id ? $this->workflowStorage->getWorkflow($id) : null;
     if (!$workflow) {
--- a/mailpoet/lib/Automation/Integrations/MailPoet/Hooks/AutomationEditorLoadingHooks.php
+++ b/mailpoet/lib/Automation/Integrations/MailPoet/Hooks/AutomationEditorLoadingHooks.php
@@ -60,11 +60,13 @@ class AutomationEditorLoadingHooks {
       }
 
       $this->newslettersRepository->bulkDelete([$emailId]);
+      $args = $step->getArgs();
+      unset($args['email_id']);
       $updatedStep = new Step(
         $step->getId(),
         $step->getType(),
         $step->getKey(),
-        array_merge($step->getArgs(), ['email_id' => 0]),
+        $args,
         $step->getNextSteps()
       );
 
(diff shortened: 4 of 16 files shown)


mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
```diff
@@ -59,6 +59,36 @@ const getEmailSendTitle = (emailType: string): string => {
   return typeMap[emailType] || MailPoet.I18n.t('stepNameSend');
 };
 
+const tutorialIcon = (emailType: string): JSX.Element | null => {
```
JanJakes: React component names must start with a capital letter. This is why you have to use it as `{tutorialIcon(emailType)}` instead of `<TutorialIcon ... />`.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -80,6 +84,42 @@ public function render() {
     ]);
   }
 
+  private function updateWorkflow(Workflow $workflow): Workflow {
+    $steps = $workflow->getSteps();
+
+    // Data is sanitized later.
+    // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+    $userData = isset($_GET['steps']) ? (array)wp_unslash($_GET['steps']) : [];
```
JanJakes: We shouldn't do this inside the WP REST API. The API doesn't differentiate the methods, they are all merged, and they should not be accessed via `$_GET` and similar. Some proxies may even convert some parameters to other ones for whatever reasons. But mainly, you completely bypassed all of our schema definitions and API docs (in future).

I find having this in the API (and the `Action::performActions`) a bit strange, and if possible, I'd try to avoid doing it and also avoid extending the interface just for this very special use case. I think it's better not to handle this case for now than doing this.

If we do want to solve it, I think it would be better if the PHP code of the MailPoet integration called something like "update my action args", which could be done via a hook, for instance. Or, maybe a simpler option, the cancel button can add a MailPoet-integration-specific URL parameter, such as `mailpoet-delete-email-id=...` and the MailPoet integration JS would catch it, update the store, and save.

JanJakes (commented):
I browsed this quickly to see if there are any changes to the architecture, and I have some concerns about the email deleting, as it is done now. See my comments.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/design_email_button.tsx
```diff
@@ -49,7 +49,7 @@ export function DesignEmailButton(): JSX.Element {
   // and the workflow is saved, we can safely redirect to the email design flow.
   useEffect(() => {
     if (emailId && workflowSaved) {
-      window.location.href = `admin.php?page=mailpoet-newsletter-editor&id=${emailId}`;
+      window.location.href = `admin.php?page=mailpoet-newsletters#/template/${emailId}`;
```
johnolek, line 53: This is pretty minor, but do we have a way to set a full page loading state when this button is clicked instead of only relying on the button's busy state?

[Figma loading state](https://www.figma.com/file/SAeK2GqSPbGsO7j5iQPT1j/Automations?node-id=4041%3A396229)
image

Right now the Design email button disappears as soon as the email is saved, and it's possible to click on "Edit content" until the redirect occurs, which can take a little while because I believe the separate workflow saving API call needs to complete first.

mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
```diff
@@ -94,28 +125,7 @@ const stepsListingHeading = (
...
-              displayTutorial();
-            }
-          }}
-        >
-          <Icon icon={video} />
-          <span>{MailPoet.I18n.t('topBarTutorial')}</span>
-        </a>
-        <span id="beamer-empty-element" />
-      </div>
+      {tutorialIcon(emailType)}
```
johnolek: Is there a reason you prefer this approach of creating a function to something like using a conditional inside the return, like this?

```js
      {emailType !== "automation" && (
        <div>
          <a
            role="button"
            onClick={displayTutorial}
            className="mailpoet-top-bar-beamer"
            title={MailPoet.I18n.t('topBarTutorial')}
            tabIndex={0}
            onKeyDown={(event) => {
              if (
                ['keydown', 'keypress'].includes(event.type) &&
                ['Enter', ' '].includes(event.key)
              ) {
                event.preventDefault();
                displayTutorial();
              }
            }}
          >
            <Icon icon={video} />
            <span>{MailPoet.I18n.t('topBarTutorial')}</span>
          </a>
          <span id="beamer-empty-element" />
        </div>
      )}
```

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
johnolek: I agree with @JanJakes about avoiding the `performActions` changes if possible.

Another thing to consider is that users might not necessarily hit the cancel button if they want to back out. In my opinion the button is pretty easy to miss as currently designed and implemented. I think there will be users who simply use their back button, or maybe they'll get distracted and click on a completely different menu item instead of choosing a template.

Maybe we could alter the behavior of the back button to include the same URL parameters as the cancel button, or have some kind of `onunload` event to make sure we delete these newsletters, but that seems like the wrong direction to approach this from.

Now that we have an intermediate step, could we defer the actual creation of the newsletter until someone selects a template? I'm guessing that would make it much more difficult to re-use the legacy template selector since it seems like it needs a newsletter ID in its route. But in any case, it seems like we shouldn't actually save the email ID to the email step until a user has selected a template if possible.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/design_email_button.tsx
```diff
@@ -49,7 +49,7 @@ export function DesignEmailButton(): JSX.Element {
   // and the workflow is saved, we can safely redirect to the email design flow.
   useEffect(() => {
     if (emailId && workflowSaved) {
```
johnolek: This might be more of a comment for @JanJakes because I believe he [wrote the original code](https://github.com/mailpoet/mailpoet/commit/8319c1e4b676c5e6399d16f0281d08e526c19a93#diff-8e2de473afa9b9ebdcbd86a22b1edcdb5352f6f289da587ea1363d685d3d4326R14). `workflowSaved` is always going to be true because it's actually getting the currently selected step.

```js
  const { selectedStep, workflowId, workflowSaved } = useSelect(
    (select) => ({
      selectedStep: select(storeName).getSelectedStep(),
      workflowId: select(storeName).getWorkflowData().id,
      workflowSaved: select(storeName).getSelectedStep(),
    }),
    [],
  );
```

When I change the definition to the actual `workflowSaved` state property, the redirect stops working.

I think this redirection behavior might currently depend on the fact that this effect fires again when the step (aka `workflowSaved` right now) changes to have an `email_id` defined in its `args`. In any case, when I try changing it to actually reference the `workflowSaved` data, I never see this component re-render with `workflowSaved` === `true`.

I don't know if it's related, but sometimes this redirect doesn't work for me. It seems to require something like the following:

1. Create a workflow with more than 1 email step
2. Create an email for one of the steps
3. Return to the workflow editor
4. Select the step that has an email associated with it (necessary step!)
5. Select a different email step without an email yet
6. Click the button to create an email
7. Observe that the redirect is not triggered

johnolek (changes requested):
Some of my comments aren't related to these actual changes but are things I think we might want to address as part of this ticket. But if you think it would make more sense to create separate tickets for anything I brought up that's fine too!

mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
websupporter: For me this became harder to read and understand what is happening. So I thought, I would have a simple function instead. Not happy about that function though. I think I prefer

```
{emailType !== 'automation' && <TutorialIcon />}
```

Does not make sense to me, to overload the icon function with responsibilities, addressed in [8fadc8a](https://github.com/mailpoet/mailpoet/pull/4391/commits/8fadc8ab75538e95d72cb20cb476aa7a8042c397)

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/design_email_button.tsx
websupporter: Thanks for pointing that out. I had a closer look. The problem is, the component would not render when an `email_id` is present. I am not quite sure, why the redirect happened at all, but it was not reliable. I have reworked the component and I think it works reliably now. [dd22635](https://github.com/mailpoet/mailpoet/pull/4391/commits/dd226354c2b6db36a5ba2a32869ec36743dfe150)

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/design_email_button.tsx
websupporter: We do not currently have such an overlay and I am hesitant to start on it without proper specification. With [dd22635](https://github.com/mailpoet/mailpoet/pull/4391/commits/dd226354c2b6db36a5ba2a32869ec36743dfe150) the "Design email"-button remains in place though.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
websupporter: Thank you @JanJakes and @johnolek for pointing that out. I think I found a better solution to the problem. [9b5ca6b](https://github.com/mailpoet/mailpoet/pull/4391/commits/9b5ca6b24e6f62cabe557e0d0517052c9e873391) adds a hook when the page is rendered and in this hook emails without a body will always be disconnected and in [cdb6d52](https://github.com/mailpoet/mailpoet/pull/4391/commits/cdb6d52fd948a48eb0f8650e1002931adf0f901d) deleted.

Therefore we do not need additional GET parameters or worry about the back button etc. Imo we can safely say, Newsletters with an empty body can always be deleted.

mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
websupporter: Addressed in [8fadc8a](https://github.com/mailpoet/mailpoet/pull/4391/commits/8fadc8ab75538e95d72cb20cb476aa7a8042c397)

websupporter:
Thank you very much @JanJakes and @johnolek for your reviews! I agree and have reworked the PR. Please feel free and have a look :)

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -46,6 +47,8 @@ public function __construct(
   public function render() {
     $id = isset($_GET['id']) ? (int)$_GET['id'] : null;
 
+    $this->wp->doAction(Hooks::EDITOR_BEFORE_LOAD, (int)$id);
```
johnolek, line 50: Are there cases where we would actually want this to fire before we confirm that a workflow exists? At first I thought maybe we should expose the workflow object directly instead of the ID, but then we would still need to load it from the database again in case any changes were made.

mailpoet/lib/Automation/Integrations/MailPoet/Hooks/AutomationEditorLoadingHooks.php
```diff
@@ -0,0 +1,82 @@
...
+      }
+      $newsletterEntity = $this->newslettersRepository->findOneById($emailId);
+      if ($newsletterEntity && $newsletterEntity->getBody() !== null) {
+        continue;
+      }
+      $updatedStep = new Step(
+        $step->getId(),
+        $step->getType(),
+        $step->getKey(),
+        array_merge($step->getArgs(), ['email_id' => 0]),
```
johnolek: Would it make more sense to remove the `email_id` completely? I think that would more closely match the original state of not having an associated email. I don't think this would cause a problem but it seems like an odd state to leave the step in unless we want to be able to tell that this occurred.

It would also mean not having to handle the special case in 8d5f941813e1ee823d53908a52e386d0b517071f

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/design_email_button.tsx
johnolek: That works for me!

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
johnolek: I like this solution much better! When I tested it locally with Firefox the back button didn't work for me. It doesn't reload the editor, so it's just stuck with the button in the busy state. Probably not a big deal though.

mailpoet/assets/js/src/newsletters/listings/heading_steps.tsx
johnolek: That looks much nicer to me too!

johnolek (dismissed):
Overall this looks good to me! I noticed one minor issue with the back button but I don't think it's a show stopper. I'm going to assign this over to @JanJakes for a final look just to make sure I didn't miss anything, if that's okay.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
websupporter: > Are there cases where we would actually want this to fire before we confirm that a workflow exists

At the moment, I can't think of any.

> we would still need to load it from the database again in case any changes were made

That was my thinking too.

mailpoet/lib/Automation/Integrations/MailPoet/Hooks/AutomationEditorLoadingHooks.php
websupporter: That would be something, I am interested to hear @JanJakes as well. I actually prefer having a permanent `email_id` property which is either `null` or `0` by default. But I might miss the reasoning behind not having the property by default.

JanJakes: @websupporter Sorry guys, I totally missed this was reassigned back to me.

Having/not having the email ID and having `0` vs `null` is rather a cosmetic issue, but it's true this it's not really compatible with what the step is doing (it doesn't exactly "undo" to the previous state without an email), so I agree with @johnolek it would make more sense to omit the `email` id in that case.

Note the validation as well, the email ID arg can be defined as `Builder::integer()->required()`. That doesn't allow for `null`, but in draft, this is not validated (only on activation). As for `0`, that seems a bit like a "hack" so I'd maybe avoid that, if possible. The `null` is fine, but at the moment it adds another state, it would make more sense only when combined with a real `null` default, such as `Builder::integer()->required()->default(null)` — and yeah, the default is invalid for the validation (which in this case is what we want).

websupporter: Addressed in 1389f6169

websupporter:
Thank you for your thoughts @JanJakes, I omit the `email_id` now in 1389f6169

JanJakes (approved):

PR #4393, opened 2022-09-27
Add tags tracking [MAILPOET-4583]

lysyjan:
## Description

This pull request extends tracking.

## Code review notes

_N/A_

## QA notes

New properties are visible in Mixpanel.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4583]

## After-merge notes

_N/A_

[MAILPOET-4583]: https://mailpoet.atlassian.net/browse/MAILPOET-4583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/analytics.js
+++ b/mailpoet/assets/js/src/analytics.js
@@ -51,8 +51,6 @@ export function mapFilterType(filter) {
     return 'subscribed date';
   if (filterType === 'woocommerce' && action === 'totalSpent')
     return 'total spent';
-  if (filterType === 'woocommerce' && action === 'totalSpent')
-    return 'total spent';
   if (filterType === 'userRole' && action === 'wordpressRole')
     return 'WordPress user role';
   if (filterType === 'userRole' && action === 'subscriberTag')
--- a/mailpoet/assets/js/src/analytics.js
+++ b/mailpoet/assets/js/src/analytics.js
@@ -17,44 +17,69 @@ export function mapFilterType(filter) {
   const action = filter.action;
   const filterType = filter.type;
 
-  if (filterType === 'email' && action === 'machineOpensAbsoluteCount')
-    return '# of machine-opens';
-  if (filterType === 'email' && action === 'opensAbsoluteCount')
-    return '# of opens';
-  if (filterType === 'woocommerce' && action === 'numberOfOrders')
-    return '# of orders';
-  if (filterType === 'email' && action === 'clicked') return 'clicked';
-  if (filterType === 'email' && action === 'clickedAny')
-    return 'clicked any email';
-  if (filterType === 'userRole' && action === 'subscriberScore') return 'score';
-  if (filterType === 'userRole' && action === 'subscribedToList')
-    return 'subscribed to list';
-  if (filterType === 'email' && action === 'opened') return 'opened';
-  if (filterType === 'email' && action === 'machineOpened')
-    return 'machine-opened';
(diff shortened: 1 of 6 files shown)


mailpoet/assets/js/src/analytics.js
```diff
@@ -12,3 +12,51 @@ export function getTrackingData() {
...
+  if (filterType === 'woocommerce' && action === 'purchasedCategory')
+    return 'purchased in category';
+  if (filterType === 'woocommerce' && action === 'purchasedProduct')
+    return 'purchased product';
+  if (filterType === 'userRole' && action === 'subscribedDate')
+    return 'subscribed date';
+  if (filterType === 'woocommerce' && action === 'totalSpent')
+    return 'total spent';
+  if (filterType === 'woocommerce' && action === 'totalSpent')
+    return 'total spent';
```
triple0t: This part is Repeated.

```
if (filterType === 'woocommerce' && action === 'totalSpent')
    return 'total spent';
  if (filterType === 'woocommerce' && action === 'totalSpent')
    return 'total spent';
```

mailpoet/assets/js/src/analytics.js
```diff
@@ -12,3 +12,51 @@ export function getTrackingData() {
...
+    return 'subscribed to list';
+  if (filterType === 'email' && action === 'opened') return 'opened';
+  if (filterType === 'email' && action === 'machineOpened')
+    return 'machine-opened';
+  if (filterType === 'woocommerceMembership' && action === 'isMemberOf')
+    return 'is active member of';
+  if (
+    filterType === 'woocommerceSubscription' &&
+    action === 'hasActiveSubscription'
+  )
```
triple0t: I would like to believe there's a better way to handle this block of code, but my brain is too tired to think right now.

I do think it makes readability a little challenging and makes updates tricky.

A quick suggestion I may offer:

Try to align and order by the same type of `filterType`
e.g
if filterType === 'email'

Show all filterType === 'email' before showing filterType === 'userRole' or filterType === 'woocommerce'

In this case, we have filterType === 'email', filterType === 'woocommerce' then filterType === 'email' again.

triple0t (dismissed):
Looks good, @lysyjan. Well done.

I have some comments, but I don't think they are major.

You may choose to go ahead and push to QA, or you may choose to update the PR.

I'm assigning the PR back to you to make your choice.

mailpoet/assets/js/src/analytics.js
samnajian: Sorry for jumping in, in these cases I really like the pattern used for [creating reducers](https://github.com/kolodny/redux-create-reducer#readme) which is used for [NgRX reducers](https://ngrx.io/guide/store/reducers) as well, which is basically only a hash of `case => reducers functions`, it's only a suggestion and maybe too big of a change for this case!

mailpoet/assets/js/src/analytics.js
lysyjan: Thanks for this catch!

lysyjan:
@triple0t Thank you for your suggestions. I added two commits with changes. I guess the readability is better now, but I want to avoid a robust solution that mentioned @samnajian.

triple0t (approved):
Looks good 💪🏿

Aschepikov (approved):
This is verified, good job

PR #4395, opened 2022-09-28
Add simple mixpanel data for automations [MAILPOET-4535]

websupporter:
## Description
Adds some basic mixpanel data

## QA notes
After making sure in [MAILPOET-4535] that the trigger based count is for active workflows only, I renamed it a bit to 
`Automation > Number of "<trigger>" active workflows` to highlight, we only count active workflows at this point.

## Linked tickets
[MAILPOET-4535]

[MAILPOET-4535]: https://mailpoet.atlassian.net/browse/MAILPOET-4535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4535]: https://mailpoet.atlassian.net/browse/MAILPOET-4535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Analytics/Reporter.php
+++ b/mailpoet/lib/Analytics/Reporter.php
@@ -2,9 +2,13 @@
 
 namespace MailPoet\Analytics;
 
+use MailPoet\Automation\Engine\Data\Step;
+use MailPoet\Automation\Engine\Data\Workflow;
+use MailPoet\Automation\Engine\Storage\WorkflowStorage;
 use MailPoet\Config\ServicesChecker;
 use MailPoet\Cron\CronTrigger;
 use MailPoet\Entities\DynamicSegmentFilterData;
+use MailPoet\Features\FeaturesController;
 use MailPoet\Listing\ListingDefinition;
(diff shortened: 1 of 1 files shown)


brezocordero (approved):
LGTM!
Code looks good.
I tested with no workflows and 1 draft. (I was not able to activate the draft on my local site)
Just to be sure, if there are no workflows at all, we still want to send the data, right?

websupporter:
> Just to be sure, if there are no workflows at all, we still want to send the data, right?

I would assume, in case there are no workflows, we want to send `0` as values.

//cc @NeosinneR

NeosinneR:
> I would assume, in case there are no workflows, we want to send `0` as values.

Correct

Aschepikov (approved):
This is verified, good job.

PR #4399, opened 2022-09-29
Confusing errors when activating creator and starter plan [MAILPOET-4464]

samnajian:
## Description

This PR basically makes sure that if the key is valid, but lacks MSS support (which currently happens for Creator plan), the messages about MSS not being activated is not shown, instead only the activated services are printed out on the screen.
Screen Shot 2022-10-10 at 16 32 48
Current state on trunk where if the key lacks MSS support

Screen Shot 2022-10-10 at 16 40 51
MSS error message not shown anymore if the key is valid for Premium plugin

## QA notes

Use keys with/without MSS support and verify the correct messages are shown and if the Creator key is valid, no error for MSS is shown.

Please note that the implementation covers the new signups/new verifications only, so if the user is old and their Creator plan key has already been added, they will still see the message about MSS not being active for them until they press on verify again.
 
## Linked tickets

[MAILPOET-4464]

[MAILPOET-4464]: https://mailpoet.atlassian.net/browse/MAILPOET-4464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/settings/store/actions/mss_and_premium.ts
+++ b/mailpoet/assets/js/src/settings/store/actions/mss_and_premium.ts
@@ -10,6 +10,7 @@ import {
 } from 'settings/store/types';
 import { updateKeyActivationState } from './key_activation';
 import { setSettings, setSetting } from './settings';
+import { getMssStatus } from '../make_default_state';
 
 export function* verifyMssKey(key: string) {
   const { success, error, res } = yield {
@@ -43,7 +44,10 @@ export function* verifyMssKey(key: string) {
     fields.mssStatus = MssStatus.VALID_MSS_NOT_ACTIVE;
   } else {
     yield setSettings(call.res.data);
-    fields.mssStatus = MssStatus.VALID_MSS_ACTIVE;
+    fields.mssStatus = getMssStatus(
+      call.res.data.mta.mailpoet_api_key_state.code === 200,
+      call.res.data,
+    );
   }
   return updateKeyActivationState(fields);
 }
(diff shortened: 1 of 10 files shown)


mailpoet/assets/js/src/settings/store/actions/mss_and_premium.ts
```diff
@@ -45,7 +45,9 @@ export function* verifyMssKey(key: string) {
   } else {
     yield setSettings(call.res.data);
     fields.mssStatus = getMssStatus(
-      call.res.data.mta.mailpoet_api_key_state.code === 200,
+      ['200', 200].includes(
+        call.res.data.mta.mailpoet_api_key_state.code as number | string,
+      ),
```
costasovo: Just curious. Could we use `parseInt(call.res.data.mta.mailpoet_api_key_state.code, 10) === 200` or maybe `Number(call.res.data.mta.mailpoet_api_key_state.code)` here?
That way, I believe, we could avoid using typecasting and Typescript should pick a correct type.

costasovo (commented):
Looks good!
I have just a couple of questions, and suggestions. But just tiny details.

1) In the bridge service integration test is a [test case that makes some assertions for all possible MSS key check results states](https://github.com/mailpoet/mailpoet/blob/5600a2fef0954593a9de86a34e7533b7640bffa9/mailpoet/tests/integration/Services/BridgeTest.php#L111-L123). Could you please extend it with the new state so that it is complete there.
2) I see that [the integration test for the Services API endpoint](https://github.com/mailpoet/mailpoet/blob/a006701d78d33e43f6b4b6dc67a54fc82bfdc736/mailpoet/tests/integration/API/JSON/v1/ServicesTest.php#L35-L156) has test cases for various MSS key states. Shouldn't we add a case also for this new case?

costasovo (approved):
Thank you! 👍

Aschepikov (approved):
GJ!

PR #4400, opened 2022-09-30
Workflow stats [MAILPOET-4673]

websupporter:
## Description
Introduces a `WorkflowStatistics` entity, which is used to show the statistics on the listing and editor page.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4673]

## After-merge notes

_N/A_

[MAILPOET-4673]: https://mailpoet.atlassian.net/browse/MAILPOET-4673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components/_automation_statistics.scss
+++ b/mailpoet/assets/css/src/components/_automation_statistics.scss
@@ -1,0 +1,53 @@
+.mailpoet-automation-stats {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  margin: auto;
+  max-width: 560px;
+
+  @at-root #mailpoet_automation_editor #{&} {
+    margin-bottom: 32px;
+  }
+
+  > li {
+    color: $color-wordpress-heading;
+    flex-grow: 1;
+    font-size: 12px;
+    font-weight: 600;
+    line-height: 16px;
+    padding: 0 16px;
+    position: relative;
+    text-align: center;
+
+    @at-root #mailpoet_automation_editor #{&} {
+      font-size: 14px;
+      line-height: 22px;
+    }
+
+    &:after {
+      align-items: center;
+      color: #a7aaad;
+      content: '›';
+      display: flex;
+      font-size: 20px;
+      font-weight: normal;
+      height: 100%;
+      justify-content: center;
+      position: absolute;
+      right: 0;
+      top: 0;
+    }
+
+    &:last-of-type:after {
+      content: '';
+    }
+
+    span {
+      color: #646970;
+      display: block;
+      font-size: 12px;
+      font-weight: 400;
+      line-height: 16px;
+    }
+  }
+}
--- a/mailpoet/assets/css/src/components/_automation_statistics.scss
+++ b/mailpoet/assets/css/src/components/_automation_statistics.scss
@@ -1,23 +1,33 @@
 .mailpoet-automation-stats {
   display: flex;
   flex-wrap: wrap;
-  justify-content: center;
+  justify-content: flex-start;
   margin: auto;
-  max-width: 560px;
 
   @at-root #mailpoet_automation_editor #{&} {
+    justify-content: center;
     margin-bottom: 32px;
+    max-width: 560px;
   }
 
-  > li {
+  .mailpoet-automation-stats-item {
     color: $color-wordpress-heading;
-    flex-grow: 1;
     font-size: 12px;
     font-weight: 600;
     line-height: 16px;
     padding: 0 16px;
     position: relative;
     text-align: center;
+
+    @at-root #mailpoet_automation_editor #{&} {
+      flex-grow: 1;
+    }
+
+    &:first-of-type {
+      @at-root #mailpoet_automation #{&} {
+        padding-left: 0;
+      }
+    }
 
     @at-root #mailpoet_automation_editor #{&} {
       font-size: 14px;
@@ -42,7 +52,7 @@
       content: '';
     }
 
-    span {
+    .mailpoet-automation-stats-label {
       color: #646970;
       display: block;
       font-size: 12px;
--- a/mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
@@ -1,0 +1,34 @@
+import { useSelect } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
+import { storeName } from '../../store';
+
+export function Statistics(): JSX.Element {
+  const { workflow } = useSelect(
+    (select) => ({
+      workflow: select(storeName).getWorkflowData(),
+    }),
+    [],
+  );
+  if (!workflow.stats.has_values) {
+    return <div />;
+  }
+
+  return (
+    <div>
+      <ul className="mailpoet-automation-stats">
+        <li className="mailpoet-automation-stats-entered">
+          <span>{__('Total Entered', 'mailpoet')}</span>
+          {new Intl.NumberFormat().format(workflow.stats.totals.entered)}
+        </li>
+        <li className="mailpoet-automation-stats-in-process">
+          <span>{__('Total Processing', 'mailpoet')}</span>
+          {new Intl.NumberFormat().format(workflow.stats.totals.in_progress)}
+        </li>
+        <li className="mailpoet-automation-stats-exited">
+          <span>{__('Total Exited', 'mailpoet')}</span>
+          {new Intl.NumberFormat().format(workflow.stats.totals.exited)}
+        </li>
+      </ul>
+    </div>
+  );
+}
--- a/mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
@@ -13,16 +13,22 @@ export function Statistics(): JSX.Element {
   return (
     <div>
       <ul className="mailpoet-automation-stats">
-        <li className="mailpoet-automation-stats-entered">
-          <span>{__('Total Entered', 'mailpoet')}</span>
+        <li className="mailpoet-automation-stats-item">
+          <span className="mailpoet-automation-stats-label">
+            {__('Total Entered', 'mailpoet')}
+          </span>
           {new Intl.NumberFormat().format(workflow.stats.totals.entered)}
         </li>
-        <li className="mailpoet-automation-stats-in-process">
-          <span>{__('Total Processing', 'mailpoet')}</span>
+        <li className="mailpoet-automation-stats-item">
+          <span className="mailpoet-automation-stats-label">
+            {__('Total Processing', 'mailpoet')}
+          </span>
           {new Intl.NumberFormat().format(workflow.stats.totals.in_progress)}
         </li>
-        <li className="mailpoet-automation-stats-exited">
-          <span>{__('Total Exited', 'mailpoet')}</span>
+        <li className="mailpoet-automation-stats-item">
+          <span className="mailpoet-automation-stats-label">
+            {__('Total Exited', 'mailpoet')}
+          </span>
           {new Intl.NumberFormat().format(workflow.stats.totals.exited)}
         </li>
       </ul>
--- a/mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
@@ -9,9 +9,6 @@ export function Statistics(): JSX.Element {
     }),
     [],
   );
-  if (!workflow.stats.has_values) {
-    return <div />;
-  }
 
   return (
     <div>
--- a/mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
+++ b/mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
@@ -4,7 +4,6 @@ namespace MailPoet\Automation\Engine\Data;
 
 class WorkflowStatistics {
 
-
   private $workflowId;
   private $versionId;
   private $entered;
--- a/mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
+++ b/mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
@@ -33,10 +33,6 @@ class WorkflowStatistics {
     return $this->entered;
   }
 
-  public function hasValues(): bool {
-    return $this->entered > 0;
-  }
-
   public function getInProgress(): int {
     return $this->inProgress;
   }
@@ -48,7 +44,6 @@ class WorkflowStatistics {
   public function toArray(): array {
     return [
       'workflow_id' => $this->getWorkflowId(),
-      'has_values' => $this->hasValues(),
       'totals' => [
         'entered' => $this->getEntered(),
         'in_progress' => $this->getInProgress(),
--- a/mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
+++ b/mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
@@ -1,0 +1,60 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Data;
+
+class WorkflowStatistics {
+
+
+  private $workflowId;
+  private $versionId;
+  private $entered;
+  private $inProgress;
+
+  public function __construct(
+    int $workflowId,
+    int $entered = 0,
+    int $inProcess = 0,
+    ?int $versionId = null
+  ) {
+    $this->workflowId = $workflowId;
+    $this->entered = $entered;
+    $this->inProgress = $inProcess;
+    $this->versionId = $versionId;
+  }
+
+  public function getWorkflowId(): int {
+    return $this->workflowId;
+  }
+
+  public function getVersionId(): ?int {
+    return $this->versionId;
+  }
+
+  public function getEntered(): int {
+    return $this->entered;
+  }
+
+  public function hasValues(): bool {
+    return $this->entered > 0;
+  }
+
+  public function getInProgress(): int {
+    return $this->inProgress;
+  }
+
+  public function getExited(): int {
+    return $this->getEntered() - $this->getInProgress();
+  }
+
+  public function toArray(): array {
+    return [
+      'workflow_id' => $this->getWorkflowId(),
+      'has_values' => $this->hasValues(),
+      'totals' => [
+        'entered' => $this->getEntered(),
+        'in_progress' => $this->getInProgress(),
+        'exited' => $this->getExited(),
+      ],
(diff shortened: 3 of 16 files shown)


mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
```diff
@@ -0,0 +1,60 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Data;
+
+class WorkflowStatistics {
+
+
```
johnolek: extra line break?

mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
```diff
@@ -0,0 +1,60 @@
...
+    return $this->versionId;
+  }
+
+  public function getEntered(): int {
+    return $this->entered;
+  }
+
+  public function hasValues(): bool {
+    return $this->entered > 0;
+  }
```
johnolek: This name seems a little ambiguous. Could something like `hasSubscribers` work? `hasValues` doesn't tell me much about what kind of values it has, and at the end of the day we're displaying these stats in the `Subscribers` column of the listing table which is why I suggested `hasSubscribers`. Not something that needs to change though!

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
```diff
@@ -0,0 +1,55 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Storage;
+
+use MailPoet\Automation\Engine\Data\WorkflowRun;
+use MailPoet\Automation\Engine\Data\WorkflowStatistics;
+
+class WorkflowStatisticsStorage {
```
johnolek, line 9: This doesn't feel like the right abstraction to me. Our other `Storage` classes all have unique tables and are responsible both for creating and retrieving data from those tables. That's what I assumed this would be, but all the data is based on workflow runs, and the `$table` property is even the same as the workflow runs class, which at first I thought might be a mistake.

The `getTotalRuns` and `getRunsWithStatuts` (sidenote: typo) methods feel like they belong in `WorkflowRunStorage`.

I'm not sure where the best place would be to create these stats objects instead, but it seems like it should be a class that receives `WorkflowRunStorage` as a dependency.

mailpoet/tests/integration/Automation/Engine/Storage/WorkflowStatisticsStorageTest.php
```diff
@@ -0,0 +1,192 @@
...
+      'two-two-zero' => [
+        2, 2,2,0, null
+      ],
+      'two-zero-two' => [
+        1, 2,0,2, null
+      ],
+    ];
+  }
+
+  public function testItSeperatesWorkflowRunsCorrectly() {
```
johnolek: should be `Separates`

mailpoet/tests/integration/Automation/Engine/Storage/WorkflowStatisticsStorageTest.php
```diff
@@ -0,0 +1,192 @@
...
+    $this->assertEquals(1, $statistics1->getInProgress());
+    $this->assertEquals(1, $statistics1->getExited());
+
+    $statistics2 = $this->testee->getWorkflowStats($workflow2->getId(), $workflow2->getVersionId());
+    $this->assertEquals(2, $statistics2->getInProgress());
+    $this->assertEquals(2, $statistics2->getExited());
+
+    $statistics3 = $this->testee->getWorkflowStats($workflow3->getId(), $workflow3->getVersionId());
+    $this->assertEquals(3, $statistics3->getInProgress());
+    $this->assertEquals(3, $statistics3->getExited());
```
johnolek: You might want to use different counts for the different statuses. Since there are always an equal number of running and exited, these assertions would still pass if I redefined `getExited` like this:

```php
  public function getExited(): int {
    return $this->getInProgress();
  }
```

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
```diff
@@ -0,0 +1,55 @@
...
+    global $wpdb;
+    $this->table = $wpdb->prefix . 'mailpoet_workflow_runs';
+    $this->wpdb = $wpdb;
+  }
+
+  public function getWorkflowStats(int $workflowId, int $versionId = null): WorkflowStatistics {
+    return new WorkflowStatistics(
+      $workflowId,
+      $this->getTotalRuns($workflowId, $versionId),
+      $this->getRunsWithStatuts(WorkflowRun::STATUS_RUNNING, $workflowId, $versionId),
```
johnolek: I don't think this is necessary for right now, but I was wondering what you think about the efficiency of using 2 separate queries for these things. Could it work to use a `GROUP BY` to get all counts by status and then calculate these values based on the results of the single query?

Right now I believe we're still returning all workflows when we render a listing page, so even if we had the pagination set to 10 per page, if we had 150 workflows we'd have to run an addition 300 queries for these stats, right?

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
```diff
@@ -0,0 +1,55 @@
...
+    $sql = !$versionId ?
+      $this->wpdb->prepare("SELECT count(ID) AS total_runs FROM $table WHERE workflow_id=%d", $workflowId) :
+      $this->wpdb->prepare("SELECT count(ID) AS total_runs FROM $table WHERE workflow_id=%d AND version_id=%d", $workflowId, $versionId);
+
+    $result = $this->wpdb->get_col($sql);
+    return $result[0] ? (int)$result[0] : 0;
+
+  }
+
+  private function getRunsWithStatuts(string $status, int $workflowId, int $versionId = null): int {
```
johnolek: I think we may want to add an index for `status` to the `mailpoet_workflow_runs` table so these queries won't be slow

mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
```diff
@@ -0,0 +1,34 @@
...
+export function Statistics(): JSX.Element {
+  const { workflow } = useSelect(
+    (select) => ({
+      workflow: select(storeName).getWorkflowData(),
+    }),
+    [],
+  );
+  if (!workflow.stats.has_values) {
+    return <div />;
+  }
```
johnolek: I don't think we necessarily want to display nothing if there are no values yet. The [figma](https://www.figma.com/file/SAeK2GqSPbGsO7j5iQPT1j/Automations?node-id=4041%3A535564) shows an editor that displays all 0s.

image

Since we don't have a similar rule on the listing page, it means we would be able to see more details on the listing page than the editor. I think a case could be made for using some kind of "no stats yet" placeholder on the listing page instead of showing a bunch of 0s but that's up to the design team.

In any case, if we don't want to show anything here in this case I think returning `null` would probably make more sense than an empty div. I do see that there was an empty div previously, I'm just not sure why it was there.

mailpoet/assets/css/src/components/_automation_statistics.scss
```diff
@@ -0,0 +1,53 @@
+.mailpoet-automation-stats {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  margin: auto;
```
johnolek, line 5: In the design, it looks like they want the "Entered" text to be aligned with the "Subscribers" column heading:

image

Currently the stats content is centered, which makes this look off especially with larger screen widths, but I think also a bit at smaller widths:

image

mailpoet/assets/css/src/components/_automation_statistics.scss
```diff
@@ -0,0 +1,53 @@
...
+  max-width: 560px;
+
+  @at-root #mailpoet_automation_editor #{&} {
+    margin-bottom: 32px;
+  }
+
+  > li {
```
johnolek: Not a big deal but it feels a tiny bit off to me to target an element like this instead of a class. Maybe all the stat items could have an additional shared class, e.g. `mailpoet-automation-stats-item mailpoet-automation-stats-entered`, so you could target that instead? Same for the `span` below.

I don't think it really matters that much, but it would make it easier to check at a glance where these styles are getting applied in the codebase.

johnolek (changes requested):
Nice work on this overall! I have a few comments here, some of which are very minor and shouldn't require any changes.

In terms of things where I would prefer to see some changes, I would say my priorities look something like this:

1. Ensure the new queries won't slow down the listing page when there are lots of workflows
2. Does it make sense to have a `WorkflowStatisticsStorage` if we don't have a separate table for the stats?

@JanJakes If you happen to have time I would appreciate your opinion on any of my comments just to make sure I'm not asking for anything that doesn't make sense!

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
JanJakes: @johnolek Just adding my thoughts, I guess you're right, and it's not a very big class, so I guess it's OK to have it in `WorkflowRunStorage`. On the other hand, as @websupporter did it, it results in smaller classes, which is nice.

I think in the Doctrine code, we also might have some more specific repositories acting on the same table as other repositories, I guess it's fine if the case makes sense. How would you guys like just renaming this class to `WorkflowRunStatisticsStorage` or `WorkflowRunStatsStorage`? Having the "run" there, it could make more sense.

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
JanJakes: Oh yeah, for the listings, it would be great to have a constant number of queries, if possible.

mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
JanJakes: Not sure about specs and Figma, just adding here that the decision to display stats may also depend on the workflow status, not sure. Does it make sense only for `active` workflows?

JanJakes (commented):
@johnolek Shared a few comments, nothing substantial.

@websupporter Looks good!

Btw. just a note on performance — it would be great to reduce the number of queries on the listings, as @johnolek pointed out, if possible, but for the future (let's not do premature optimization now), I think we could load the stats async in both the listing and the editor. That will require adding APIs and some UI placeholders, so really not something to do now.

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
websupporter: I would not _yet_ want to rename it to WorkflowRunStatistics. I made it a new class, as I imagine the entity will grow a bit, once we have more detailed statistics. I think in the end, the decision on the best abstraction etc. is probably made, once we have a broader insight in what the statistics all need to serve instead of just the totals and what is the best way to achieve that.

Currently, my thinking for naming is: These are the statistics for a Workflow. From a `WorkflowRunStatistic` I would expect statistics for a specific run like maybe `started_at`, `steps_done`... The Storage has been named accordingly. So for me the naming of the storage is not derived from the tables it queries but the entities it provides.

mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
websupporter: Thanks @johnolek, addressed in  [24f4352](https://github.com/mailpoet/mailpoet/pull/4400/commits/24f43520e2b4677ccdfd945d53b6895886e239f4).

> Does it make sense only for active workflows?

I would not do that as of now. 

1. Not specified in the ticket.
2. As an admin I might want to see how many entered my trashed or deactivated workflow to judge if I should activate it again or build a completely new one

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
websupporter: Good catch! Thanks: [b18b4d3](https://github.com/mailpoet/mailpoet/pull/4400/commits/b18b4d3317a869b2fec8da34cf1a8fa52ce3efee)

mailpoet/tests/integration/Automation/Engine/Storage/WorkflowStatisticsStorageTest.php
websupporter: With `testItSeparatesWorkflowRunsCorrectly` I just want to check whether the queries correctly account for the `workflow_id`. This is why I have different values per Workflow. If `getExited()` would change `testItCalculatesTotalsCorrectlyForSingleWorkflow` fails, which accounts for that. 

Thinking about it, I think on the contrary the `testItSeparatesWorkflowRunsCorrectly` test should be even more simple to avoid it failing in a scenario you describe, to improve the debugging experience. What do you think about 6fca21650

mailpoet/tests/integration/Automation/Engine/Storage/WorkflowStatisticsStorageTest.php
websupporter: Thank you! 8466e0f8d

mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
websupporter: > This name seems a little ambiguous

Actually that was by purpose, but I think, we should just drop this for now. The idea was to provide a quite general "Yes, I have some `values`" information. You have seen, how I used it. Now my thinking is: But having it so general, because this entity might evolve means, its meaning will change over time. How it is used currently is bound to total values...

Now, we could start like `hasSubscribers` or `hasTotals` and add more and more `hasX` methods or drop it and introduce specific methods, when they are actually needed.

[24f4352](https://github.com/mailpoet/mailpoet/pull/4400/commits/24f43520e2b4677ccdfd945d53b6895886e239f4)

mailpoet/lib/Automation/Engine/Data/WorkflowStatistics.php
websupporter: Fixed 9510018d0

mailpoet/assets/css/src/components/_automation_statistics.scss
websupporter: Addressed in [9b4fdbb](https://github.com/mailpoet/mailpoet/pull/4400/commits/9b4fdbb3cff3a222d39b60ecb812aec03801cd30)

mailpoet/assets/css/src/components/_automation_statistics.scss
websupporter: Addressed in [9b4fdbb](https://github.com/mailpoet/mailpoet/pull/4400/commits/9b4fdbb3cff3a222d39b60ecb812aec03801cd30)

websupporter:
Thank you @johnolek and @JanJakes for your reviews. Much appreciated.

websupporter:
I still need to address the query-issue.

> for the listings, it would be great to have a constant number of queries, if possible

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
johnolek: Makes sense to me! It sounds like I had the wrong idea about the "rules" for these `...Storage` classes.

mailpoet/tests/integration/Automation/Engine/Storage/WorkflowStatisticsStorageTest.php
johnolek: Ah, I see. I thought the separation was referring to run status as well. I like your change!

mailpoet/assets/js/src/automation/editor/components/workflow/statistics.tsx
johnolek: I agree that it makes sense to display the stats regardless of status, especially if we always display them on the listing page

mailpoet/lib/Automation/Engine/Storage/WorkflowStatisticsStorage.php
websupporter: Addressed in 4565af2 and 513166c49

websupporter:
Thank you for your review @johnolek, I addressed the open issues. Please have a look again :bow:

johnolek (approved):
@websupporter This definitely seems like it would work and I'm happy to move this forward as is. I personally found it a little hard to wrap my mind around some of the code at first, but that could just be sleep deprivation!

My only concern relates to workflow versions. I don't think the ticket specifies if we should be showing stats for all versions by default or just the most recent. In my mind it makes sense to use the most recent version because I wouldn't care how many people entered/exited an old version of a workflow if it was completely different in the previous version(s). Is this something that was discussed at all, or maybe I missed the requirement in the ticket?

In any case, I just want to be sure we're okay with defaulting to showing stats for all versions.  I know that it would be more complex if we wanted to do a specific version when retrieving all workflows, and maybe this is a question we can save for post-MVP.

JanJakes:
@johnolek @websupporter

> @websupporter This definitely seems like it would work, and I'm happy to move this forward as is. I personally found it a little hard to wrap my mind around some of the code at first, but that could just be sleep deprivation!

The "matrix" code seems a bit hard to process, but I don't know if there's a straightforward way to simplify it, and I would proceed to merge this now and improve later if needed.

> My only concern relates to workflow versions. I don't think the ticket specifies if we should be showing stats for all versions by default or just the most recent.

Ohh, great point! Let's bring @NeosinneR to this discussion, I guess. Should we merge it and create a new ticket, if needed?

websupporter:
> My only concern relates to workflow versions. I don't think the ticket specifies if we should be showing stats for all versions by default or just the most recent.

Would be interested in @NeosinneR opinion as well, but from a practical standpoint, to have only the most recent seems to me not the right approach. When I pause a workflow the stats would reset. When I change delay from 1 to 2 hours the stats would reset. I can see, how a user would want to see more detailed stats per version, to judge if one version performs better then another one, but I think on this high level I would want to have an aggregate.

JanJakes:
I'll proceed with merging this. If any changes are needed, we can create a ticket.

johnolek:
> When I pause a workflow the stats would reset. When I change delay from 1 to 2 hours the stats would reset. I can see, how a user would want to see more detailed stats per version, to judge if one version performs better then another one, but I think on this high level I would want to have an aggregate.

This is a great point, @websupporter! My example of someone making drastic changes to a workflow is probably much less likely than yours of someone making a small tweak.

PR #4403, opened 2022-09-30
Action scheduler cron fixes and improvements [MAILPOET-4684]

costasovo:
## Description

This PR addresses a couple of issues we identified on a customer page:
1) There were duplicate `mailpoet/cron/daemon-trigger` action. Probably created because of some race condition
2) In some cases, the `mailpoet/cron/daemon-run` action was running 20 times per minute.

To prevent these issue the PR introduces these changes:
1) Update of the Action Scheduler package from 3.4 to 3.5. Version 3.5 supports unique actions, and this should prevent creating duplicate actions.
2) Unscheduling all mailpoet cron actions during activation or update. This should clean up potential duplicate actions users have.
3) Cron will now pause its execution if the sending job is waiting for retry. The wait interval is 120, and the cron was running in a loop, not doing anything. This also affects the website visitors' type of cron.
4) Switched `mailpoet/cron/daemon-run` from a recurring action to a simple action. We now have complete control over the continuous scheduling of daemon's runs. This might also be less confusing for users as they should not see recurring actions scheduled to run once per second in the action scheduler logs.
5) Added protection against endless loops of the`mailpoet/cron/daemon-run` action. In case we detect that the deamon worked less than 2 seconds and there is still some execution time left, and the daemon still has more jobs to we won't trigger another `mailpoet/cron/daemon-run` action to run immediately, but we wait for the `mailpoet/cron/daemon-trigger` action to trigger it. So in case the daemon is stuck because of some issue we will run the cron only every 2 minutes instead of keeping it running continuously.
6) In order to make 5) work efficiently, I also made a change in the cron worker runner, and it will now immediately start processing tasks that were switched to running. Before the change, the worker just switched them to running, and they had to wait to be processed for the next iteration of the cron process.

## Code review notes

I tried to explain changes in commit messages. It is better to review commit by commit.

## QA notes

Please test sending to a bigger amount of subscribers (10K+). I've prepared a testing build that is faking sending emails. The process is running, but no emails go out. It can be used with MSS, SMTP, and sending with its own host methods. 
Download: [mailpoet-dummy-sending.zip](https://github.com/mailpoet/mailpoet/files/9683410/mailpoet-dummy-sending.zip)
You can observe action scheduler actions in WP Admin > Tools > Scheduled Actions. There should be only one `mailpoet/cron/daemon-trigger` in pending, and during sending you may observe that there will be logged `mailpoet/cron/daemon-run` actions approx every 20 seconds.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4684]

## After-merge notes

_N/A_

[MAILPOET-4684]: https://mailpoet.atlassian.net/browse/MAILPOET-4684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/ActionScheduler/Actions/DaemonRun.php
+++ b/mailpoet/lib/Cron/ActionScheduler/Actions/DaemonRun.php
@@ -7,6 +7,7 @@ use MailPoet\Cron\ActionScheduler\RemoteExecutorHandler;
 use MailPoet\Cron\CronHelper;
 use MailPoet\Cron\Daemon;
 use MailPoet\Cron\Triggers\WordPress;
+use MailPoet\Logging\LoggerFactory;
 use MailPoet\WP\Functions as WPFunctions;
 
 class DaemonRun {
@@ -32,6 +33,9 @@ class DaemonRun {
   /** @var ActionScheduler */
   private $actionScheduler;
 
+  /** @var LoggerFactory */
+  private $loggerFactory;
+
   /**
    * Default 20 seconds
    * @var float
@@ -47,7 +51,8 @@ class DaemonRun {
     WordPress $wordpressTrigger,
     CronHelper $cronHelper,
     RemoteExecutorHandler $remoteExecutorHandler,
-    ActionScheduler $actionScheduler
+    ActionScheduler $actionScheduler,
+    LoggerFactory $loggerFactory
   ) {
     $this->wp = $wp;
     $this->daemon = $daemon;
@@ -55,6 +60,7 @@ class DaemonRun {
     $this->cronHelper = $cronHelper;
     $this->remoteExecutorHandler = $remoteExecutorHandler;
     $this->actionScheduler = $actionScheduler;
+    $this->loggerFactory = $loggerFactory;
   }
 
   public function init(): void {
@@ -93,6 +99,10 @@ class DaemonRun {
     // If there was still some execution time left, the daemon should have been continued.
     $lastDurationWasTooShort = ($this->lastRunDuration < self::SHORT_DURATION_THRESHOLD) && ($this->remainingExecutionLimit > 0);
     if ($lastDurationWasTooShort) {
+      $this->loggerFactory->getLogger(LoggerFactory::TOPIC_CRON)->info('Daemon run ended too early!', [
+        'duration' => $this->lastRunDuration,
+        'remainingLimit' => $this->remainingExecutionLimit,
+      ]);
       return;
     }
     $this->actionScheduler->scheduleImmediateSingleAction(self::NAME);
--- a/mailpoet/lib/Cron/ActionScheduler/Actions/DaemonRun.php
+++ b/mailpoet/lib/Cron/ActionScheduler/Actions/DaemonRun.php
@@ -85,17 +85,21 @@ class DaemonRun {
    * After Action Scheduler finishes its work we need to check if there is more work and in case there is we trigger additional runner.
    */
   public function afterProcess(): void {
-    $lastDurationWasTooShort = ($this->lastRunDuration < self::SHORT_DURATION_THRESHOLD) && ($this->remainingExecutionLimit > 0);
-    // Schedule next action in case we have more jobs to do.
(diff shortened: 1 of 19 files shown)


mailpoet/lib/Cron/ActionScheduler/Actions/DaemonRun.php
```diff
@@ -79,12 +79,11 @@ public function getDaemonExecutionLimit(): float {
    */
   public function afterProcess(): void {
     if ($this->wordpressTrigger->checkExecutionRequirements()) {
+      $this->actionScheduler->scheduleImmediateSingleAction(self::NAME);
       // The automatic rescheduling schedules the next recurring action to run after 1 second.
       // So we need to wait before we trigger new remote executor to avoid skipping the action
```
JanJakes: I guess these comments are not relevant anymore? Even the sleep may not be needed, but I don't know, is it better to keep it for some reason? I guess it may be safer to keep it.

mailpoet/lib/Cron/ActionScheduler/Actions/DaemonRun.php
```diff
@@ -85,17 +85,21 @@ public function getDaemonExecutionLimit(): float {
...
     // If there was still some execution time left, the daemon should have been continued.
-    if ($this->wordpressTrigger->checkExecutionRequirements() && !$lastDurationWasTooShort) {
-      $this->actionScheduler->scheduleImmediateSingleAction(self::NAME);
-      // The automatic rescheduling schedules the next recurring action to run after 1 second.
-      // So we need to wait before we trigger new remote executor to avoid skipping the action
-      sleep(2);
-      $this->remoteExecutorHandler->triggerExecutor();
+    $lastDurationWasTooShort = ($this->lastRunDuration < self::SHORT_DURATION_THRESHOLD) && ($this->remainingExecutionLimit > 0);
+    if ($lastDurationWasTooShort) {
+      return;
```
JanJakes, line 106: This check is very cheap (comparing two numbers), so it could as well be the first one here since `checkExecutionRequirements` is more expensive. That said, this is just a safeguard and it shouldn't run often, so probably not important at all.

mailpoet/tests/integration/Cron/ActionScheduler/Actions/DaemonRunTest.php
```diff
@@ -118,5 +128,6 @@ private function cleanup(): void {
     $claimsTable = $wpdb->prefix . 'actionscheduler_claims';
     $wpdb->query('TRUNCATE ' . $claimsTable);
     $this->truncateEntity(ScheduledTaskEntity::class);
+    //$this->truncateEntity(LogEntity::class);
```
JanJakes: Forgotten?

JanJakes (commented):
Great job 👏 Really nice improvements!

mailpoet/lib/Cron/ActionScheduler/Actions/DaemonRun.php
costasovo: I'll keep it in this order. In the next commit, I added some logging, and I want to log only in case `$hasJobsToDo` is truthy so having this order is convenient to avoid some nested blocks.

mailpoet/lib/Cron/ActionScheduler/Actions/DaemonRun.php
costasovo: I agree it is safer to keep it. There is a similar [sleep in the action scheduler async runner](https://github.com/woocommerce/action-scheduler/blob/trunk/classes/ActionScheduler_AsyncRequest_QueueRunner.php#L91-L96). I'm not really sure why and how it helps, though.

JanJakes (dismissed):

Aschepikov (dismissed):
Verified. GJ!

Aschepikov:
@costasovo there are some conflicts, can you please resolve them?

Aschepikov (approved):
Verified

PR #4410, opened 2022-10-04
Skip WooCommerce Subscriptions tests when Woo Custom Orders Tables is enabled [MAILPOET-4695]

rodrigoprimo:
## Description

This PR changes a WooCommerce Subscriptions test to skip it when Woo Custom Orders Tables is enabled. To do that it was necessary to change how we forbid skipping tests when they are executed in the `trunk` and `release` branches (see commit message for more details).

I also created a branch that we can delete called `cot-skip-woocommerce-subscription-test-delme` to show that skipping the tests, with the exception of the ones in the accept list, is still not possible when using the two branches mentioned above: https://app.circleci.com/pipelines/github/mailpoet/mailpoet?branch=cot-skip-woocommerce-subscription-test-delme

We can delete this branch, after merging this PR.

## Code review notes

_N/A_

## QA notes

I don´t think QA is needed for this PR.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4695]

## After-merge notes

_N/A_

[MAILPOET-4695]: https://mailpoet.atlassian.net/browse/MAILPOET-4695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/_support/CheckSkippedTestsExtension.php
+++ b/mailpoet/tests/_support/CheckSkippedTestsExtension.php
@@ -12,7 +12,7 @@ class CheckSkippedTestsExtension extends Extension {
 
   public function checkSkippedTests(FailEvent $event) {
     $branch = getenv('CIRCLE_BRANCH');
-    $testName = $event->getTest()->getName();
+    $testName = $event->getTest()->getMetadata()->getName();
 
     // list of tests that are allowed to be skipped on trunk and release branches
     $allowedToSkipList = ['createSubscriptionSegmentForActiveSubscriptions'];
--- a/mailpoet/tests/_support/CheckSkippedTestsExtension.php
+++ b/mailpoet/tests/_support/CheckSkippedTestsExtension.php
@@ -15,7 +15,7 @@ class CheckSkippedTestsExtension extends Extension {
     $testName = $event->getTest()->getName();
 
     // list of tests that are allowed to be skipped on trunk and release branches
-    $allowedToSkipList = [];
+    $allowedToSkipList = ['createSubscriptionSegmentForActiveSubscriptions'];
 
     if (in_array($branch, ['trunk', 'release']) && !in_array($testName, $allowedToSkipList)) {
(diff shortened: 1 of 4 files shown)


mailpoet/tests/_support/CheckSkippedTestsExtension.php
```diff
@@ -1,19 +1,23 @@
...
-    Events::SUITE_AFTER => 'checkErrorsAfterTests',
+    Events::TEST_SKIPPED => 'checkSkippedTests',
   ];
 
-  public function checkErrorsAfterTests(SuiteEvent $e) {
+  public function checkSkippedTests(FailEvent $event) {
     $branch = getenv('CIRCLE_BRANCH');
-    $skipped = $e->getResult()->skipped();
-    if (in_array($branch, ['trunk', 'release']) && (count($skipped) !== 0)) {
+    $testName = $event->getTest()->getName();
```
costasovo: I think we could change this to: `$event->getTest()->getMetadata()->getName();` and that will make PHPStan happy and we won't need to add the ignore error.
The Codeception says that `getTest` returns `TestInterface`, which indeed doesn't define `getName`.

costasovo (commented):
Looks good! Thanks for improving the extension and switching it to the proper event.

I think we can get rid of the PHPStan error with a small change. Please see my other comment.

mailpoet/tests/_support/CheckSkippedTestsExtension.php
rodrigoprimo: @costasovo, good find. It didn't occur to me to check if there was another way to call `getName()`. I added a new commit implementing the change that you suggested. Can you please review this PR again?

costasovo (dismissed):
Good job! 👍 It's only tests related, so I'm going to merge it.

costasovo (approved):

PR #4411, opened 2022-10-06
Fix entity manager closed in cron [MAILPOET-4699]

websupporter:
## Description
Sometimes, it can happen (we do not know yet why) that the `newsletter_id` in the `sending_queues` table is zero. When fetched the `$newsletter` of the `SendingQueueEntity` is `null`. Somehow Doctrine started (we suspect a recent refactor) to recognize this as an alteration, which needed to be stored. So, it attempted to set `newsletter_id` to `null`. But the column is not nullable. This provoked an error and closed the EntityManager during the cron run.

Additionally, such a task with a `0`-newsletter remains running forever, now blocking effectively the cron system from working properly.

This PR attempts to fix the problem by
1. making the column nullable (28e9ac58654cc323ece903e4b87783e1beca57c1 and 28e9ac58654cc323ece903e4b87783e1beca57c1)
2. deleting a task when no newsletter is found, so to not run the task endlessly (335977a58d79aaa314c610b8089691672166c187)

While debugging the issue it was helpful to extend our logging to more precisely logging errors. We just logged the latest error in our cron system, but here, we had overlapping errors.

One job created an SQL error while the next one created an EntityManagerClosed error. With a97836c049bdfe5d4684c0276e585345e9af6d02 I improve the logging so that all errors will be catched and stored in the database.

## QA notes

Validate error
1. Use mailpoet v3.100, I would make it with a blank new installation
2. Execute the following two SQL commands: 
```
INSERT INTO `wp_mailpoet_sending_queues` (`id`, `task_id`, `newsletter_id`, `newsletter_rendered_body`, `newsletter_rendered_subject`, `subscribers`, `count_total`, `count_processed`, `count_to_process`, `meta`, `created_at`, `updated_at`, `deleted_at`) VALUES ('10000000', '10000000', '0', NULL, NULL, NULL, '0', '0', '0', NULL, NULL, CURRENT_TIMESTAMP, NULL);
```
```
INSERT INTO `wp_mailpoet_scheduled_tasks` (`id`, `type`, `status`, `priority`, `scheduled_at`, `processed_at`, `created_at`, `updated_at`, `deleted_at`, `in_progress`, `reschedule_count`, `meta`) VALUES ('10000000', 'sending', NULL, '5', '2022-10-04 10:47:00', NULL, '2022-10-04 10:47:00', '2022-10-04 10:47:00', NULL, NULL, '0', NULL); 
```
3. Run the cron
4. See the error message in the Help section. There can be multiple. At least one should say "EntityManager closed" or contain "Integrity constraint violation: 1048 Column 'newsletter_id' cannot be null"
5. Install the ZIP file which was provided in this run
6. To emulate, this ZIP file is version 3.100.1, you need to open the file `/mailpoet.php` and change the version in L23 from `3.100.0` to `3.100.1`
7. Go to wp-admin
8. Run the cron
9. Check that the entries you created in `wp_mailpoet_scheduled_tasks` and `wp_mailpoet_sending_queues` have been removed
10. The errors might still exist in the Help section, but you can see, how the "Last seen" date differs from the "Last run" date.
11. If the errors are visible: Create a newsletter and send it. I think afterwards the Last errors section should be empty.

## Merge notes
It seems the Doctrine refactor is continuing and we can't really merge this PR into trunk. In the end, we need to make an emergency release and cherry-pick anyway, if I see that correctly. And then we need to see, how to solve those merge conflicts. Not completely sure about the best process at the moment, as I just saw the merge conflict.

## Linked tickets

[MAILPOET-4699]

[MAILPOET-4699]: https://mailpoet.atlassian.net/browse/MAILPOET-4699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Migrator.php
+++ b/mailpoet/lib/Config/Migrator.php
@@ -218,7 +218,7 @@ class Migrator {
     $attributes = [
       'id int(11) unsigned NOT NULL AUTO_INCREMENT,',
       'task_id int(11) unsigned NOT NULL,',
-      'newsletter_id int(11) unsigned NOT NULL,',
+      'newsletter_id int(11) unsigned NULL,',
       'newsletter_rendered_body longtext,',
       'newsletter_rendered_subject varchar(250) NULL DEFAULT NULL,',
       'subscribers longtext,',
--- a/mailpoet/lib/Cron/CronWorkerRunner.php
+++ b/mailpoet/lib/Cron/CronWorkerRunner.php
@@ -3,6 +3,7 @@
 namespace MailPoet\Cron;
 
 use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Logging\LoggerFactory;
 use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
@@ -27,17 +28,22 @@ class CronWorkerRunner {
   /** @var ScheduledTasksRepository */
   private $scheduledTasksRepository;
 
+  /** @var LoggerFactory  */
+  private $loggerFactory;
+
   public function __construct(
     CronHelper $cronHelper,
     CronWorkerScheduler $cronWorkerScheduler,
     WPFunctions $wp,
-    ScheduledTasksRepository $scheduledTasksRepository
+    ScheduledTasksRepository $scheduledTasksRepository,
+    LoggerFactory $loggerFactory
   ) {
     $this->timer = microtime(true);
     $this->cronHelper = $cronHelper;
     $this->cronWorkerScheduler = $cronWorkerScheduler;
     $this->wp = $wp;
     $this->scheduledTasksRepository = $scheduledTasksRepository;
+    $this->loggerFactory = $loggerFactory;
   }
 
   public function run(CronWorkerInterface $worker) {
@@ -72,6 +78,10 @@ class CronWorkerRunner {
       }
     } catch (\Exception $e) {
       if (isset($task) && $task && $e->getCode() !== CronHelper::DAEMON_EXECUTION_LIMIT_REACHED) {
+        /**
+         * ToDo: Use \LoggerFactory::TOPIC_CRON as logger topic, once it is available
+         */
+        $this->loggerFactory->getLogger()->error($e->getMessage(), ['error' => $e]);
         $this->cronWorkerScheduler->rescheduleProgressively($task);
       }
       throw $e;
--- a/mailpoet/lib/Cron/Daemon.php
+++ b/mailpoet/lib/Cron/Daemon.php
@@ -3,6 +3,7 @@
 namespace MailPoet\Cron;
 
 use MailPoet\Cron\Workers\WorkersFactory;
+use MailPoet\Logging\LoggerFactory;
 
 class Daemon {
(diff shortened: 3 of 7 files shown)


mailpoet/lib/Cron/CronWorkerRunner.php
```diff
@@ -72,6 +73,8 @@ public function run(CronWorkerInterface $worker) {
       }
     } catch (\Exception $e) {
       if (isset($task) && $task && $e->getCode() !== CronHelper::DAEMON_EXECUTION_LIMIT_REACHED) {
+
+        LoggerFactory::getInstance()->getLogger()->error($e->getMessage(), ['error' => $e]);
```
JanJakes: I guess both of these loggers should define a topic and a message.

I've just seen a similar thing in a PR that introduced a `cron` loggin topic: https://github.com/mailpoet/mailpoet/pull/4403/commits/c30f6a385f264670d73cef67973a44bd3a82d80d

Unfortunately not merged yet.

Additionally, can we use DI for the logger factory in all the new cases? See the link above, how it's used there.

mailpoet/lib/Config/Migrator.php
```diff
@@ -664,6 +665,21 @@ private function sqlify($model, $attributes) {
     return implode("\n", $sql);
   }
 
+  private function updateNewsletterIdCanBeNullInSendingQueues() {
```
JanJakes: Just to be sure — we need to do this manually here because the delta DB will not detect that change?

JanJakes (commented):
Awesome!

mailpoet/lib/Cron/CronWorkerRunner.php
websupporter: > should define a topic

For the topic, I will add an Todo for now. Once its merged, we use cron.

> and a message

Isn't the message `$e->getMessage()`?

> can we use DI

Thanks for catching, will do.

mailpoet/lib/Config/Migrator.php
websupporter: Nice catch. No, it is not needed.

mailpoet/lib/Cron/CronWorkerRunner.php
JanJakes: > For the topic, I will add an Todo for now. Once its merged, we use cron.

I think you can add the constant as well. If you do it exactly on the same line as [here](https://github.com/mailpoet/mailpoet/pull/4403/commits/c30f6a385f264670d73cef67973a44bd3a82d80d#diff-eba49c637a6dbaf40e3ce6928b8c026f3e14aae04a8abf559beec797c0d09ef2R34), then I think the rebase of the other branch will even pass without conflicts — seeing both branches have the same change. Do you want to add it or prefer to do it later (an extra PR)?

> Isn't the message $e->getMessage()?

Oh, I missed that. I thought that we always pass in our own text, but now I see we seem to do that only for `info`, for error logs this is fine.

websupporter: later in an extra PR

JanJakes (approved):
👏

veljkho (approved):
Thanks so much for fixing this and writing detailed steps @websupporter !

PR #4413, opened 2022-10-06
Workflow validation UI [MAILPOET-4659]

JanJakes:
## Description

This PR:
1. Converts existing step and step arg validation to step validator rules, adds tests.
2. Introduces `StepValidationArgs` and adds rule for custom step validation via `validate` method.
3. Collects step validation errors for all steps and converts them to an error response.
4. Implements step order validation (provided and required subject dependencies).
5. Implements validation UI.
6. Fixes a few encountered random issues.

## Code review notes

_N/A_

## QA notes

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/632

## Linked tickets

[MAILPOET-4659]

## After-merge notes

_N/A_

[MAILPOET-4659]: https://mailpoet.atlassian.net/browse/MAILPOET-4659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Engine/Validation/WorkflowRules/UnknownStepRule.php
```diff
@@ -25,36 +22,31 @@ class WorkflowStepsValidator {
...
     $this->registry = $registry;
     $this->workflowStorage = $workflowStorage;
   }
 
-  public function validateSteps(Workflow $workflow): void {
-    foreach ($workflow->getSteps() as $step) {
-      $this->validateStep($workflow, $step);
-    }
+  public function initialize(Workflow $workflow): void {
+    $this->cachedExistingWorkflow = false;
```
websupporter, line 32: Just reading what `cachedExistingWorkflow` does. Should we do caching on that level or thinking about this at a deeper level like

```
class WorkflowStorageCache implements WorkflowStorage {

	private $dbStorage;
	
	public function __construct(WorkflowStorageDb $dbStorage) {
		$this->dbStorage = $dbStorage;
	}
	
	public function getById(int $id) {
		if (cached) {
			return cached;
		}
		
		cached = $this->dbStorage->getbyId($id);
		do_cache
		return cached;
	}
	
	public function update(Workflow) {
		delete_cache
		$this->dbStorage->update(Workflow);
	}
}
```

mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
```diff
@@ -91,6 +93,13 @@ export function Step({ step, isSelected }: Props): JSX.Element {
               : stepTypeData.title}
           </div>
         </div>
+        {error && (
+          <div className="mailpoet-automation-editor-step-footer">
+            <div className="mailpoet-automation-editor-step-error">
+              <Chip size="small">{__('Not set', 'mailpoet')}</Chip>
```
websupporter, line 99: Hmmm... `Not set`: Couldn't it be set to an invalid value as well (in theory)? Just thinking out loud, nothing to really act on I guess.

mailpoet/assets/js/src/automation/editor/components/header/errors.tsx
```diff
@@ -0,0 +1,169 @@
...
+        }
+        onMouseDown={() =>
+          // Catch and mark a mouse down event from an open popover with "undefined" to avoid closing it
+          // (automatically via click outside) and reopening it right after (via the onClick handler).
+          // The "onClose" method of the popover doesn't pass any events so we can't filter them.
+          setShowPopover((prevState) => (prevState ? undefined : prevState))
+        }
+        style={{ textDecoration: 'none', borderRadius: 99999 }}
+      >
+        <Chip>{stepErrors.length} issues</Chip>
```
websupporter, line 138: `_n( '1 issue', '%d issues', stepErrors.length, 'mailpoet' );`

So, we have an instance where WP translations would be really useful :)

mailpoet/assets/js/src/automation/editor/api-error-handler.tsx
```diff
@@ -16,6 +17,12 @@ export const registerApiErrorHandler = (): void =>
       } catch (error) {
         const errorObject = error as ApiError;
         const status = errorObject.data?.status;
+        const code = errorObject.code;
+
+        if (code === 'mailpoet_automation_workflow_not_valid') {
+          dispatch(storeName).setErrors({ steps: errorObject.data.errors });
+          return undefined;
```
websupporter, line 24: So `<Error />` pops up, but because of the early return no `Error message` which showed up will be shown. I guess it's a good thing and intended, but wanted to point that out.

websupporter (approved):
Overall, lgtm. Thanks for the PR @JanJakes 

You can merge when you like. No blocking issue found.

mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
JanJakes: Yep, that's how it was designed, and I think that's probably expected (@NeosinneR can confirm maybe) that when a step is invalid, we mark it as `Not set` anyway, kind of like `Not correctly set` or like `Not set well-enough to be activated, bro`. Then we'll display the error details somewhere so I guess this works.

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/UnknownStepRule.php
JanJakes: Good idea, we really could try to make sure some of the main entities are fetched only once per request. I added this to the P2 ideas draft.

mailpoet/assets/js/src/automation/editor/components/workflow/step.tsx
NeosinneR: As Jan said 👍 this way it also simplifies the implementation.

mailpoet/assets/js/src/automation/editor/api-error-handler.tsx
JanJakes: Yes, in this case, I think that's correct because, in the designs, there is no other "main" error message apart from the `X issues` chip. So one more mention of @NeosinneR whether we should consider also showing an in-editor notice saying sth like `Could not activate workflow because some steps are not fully set up...`?

NeosinneR: Without the in-editor notice, would the user just see the `2 issues` chip in the header with a popup of not set steps?

Editor_Default

JanJakes: @NeosinneR Yes, that's the current state. The popup auto-opens.

NeosinneR: Let's keep it like that, without global notice. I'd guess we'll receive feedback if that's not adequate.

PR #4415, opened 2022-10-07
Skip customer sync setting for new stores [MAILPOET-4652]

costasovo:
## Description
This PR changes how we display the WooCommerce setup screen. 

The screen for shops with no customers:
Screenshot 2022-10-07 at 10 39 04

The screen for shops with some customers (guest or registered).
Screenshot 2022-10-07 at 10 38 27

Note: During implementation, I found out that we need to save a value for the first checkbox event when it is not displayed. This is because it is then used in `Settings > WooCommerce > Subscribe old WooCommerce customers`, and the setting is used in Customer sync for [updating the status of subscribers who don't have confirmation records](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Segments/WooCommerce.php#L527-L551). I discussed it with @pavel-mailpoet in the Jira ticket, and we chose to use `false` since it is safer for list hygiene.

## Code review notes
I discussed the approach for fetching the subscribers count with the proton team.

## QA notes
The page can be displayed any time by going to URL: `wp-admin/admin.php?page=mailpoet-woocommerce-setup#/woocommerce`

1) Check the screen on with a fresh WooCommerce installation without any customers. You should see only one setting.
2) Create an order or register a customer and check the screen again and you should see both settings.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4652]

## After-merge notes

_N/A_

[MAILPOET-4652]: https://mailpoet.atlassian.net/browse/MAILPOET-4652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/settings/store/normalize_settings.ts
+++ b/mailpoet/assets/js/src/settings/store/normalize_settings.ts
@@ -185,7 +185,7 @@ export function normalizeSettings(data: Record<string, unknown>): Settings {
       }),
     }),
     mailpoet_subscribe_old_woocommerce_customers: asObject({
-      enabled: enabledRadio,
+      enabled: disabledCheckbox,
     }),
     premium: asObject({
       premium_key: text,
--- a/mailpoet/assets/js/src/wizard/steps/woo_commerce_step.jsx
+++ b/mailpoet/assets/js/src/wizard/steps/woo_commerce_step.jsx
@@ -8,7 +8,9 @@ import { YesNo } from '../../common/form/yesno/yesno';
 
 function WizardWooCommerceStep(props) {
   const [allowed, setAllowed] = useState(null);
-  const [importType, setImportType] = useState(null);
+  const [importType, setImportType] = useState(
+    props.showCustomersImportSetting === false ? 'unsubscribed' : null,
+  );
   const [submitted, setSubmitted] = useState(false);
 
   const submit = (event) => {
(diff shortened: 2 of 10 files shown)


rodrigoprimo (dismissed):
Looks good to me.

Aschepikov (commented):
Hi, @costasovo there is one issue with woocommerce settings in the 4s step in the wizard process.
In both cases (with woocommerce customers and without them in the 4s step I see just one setting, but the button ¨start using mailpoet¨ is frozen. 
In the 1 case (without wc customers) I can go to another page of the mailpoet and there I will see 1 setting for 4s step with working button.

https://user-images.githubusercontent.com/113185437/196198263-2db684a8-ec72-486c-bd23-3bedad286b85.mp4

In the 2 case (with wc customers) I can go to another page of the mailpoet too and there I will see 2 settings for 4s step with working button.

image

https://user-images.githubusercontent.com/113185437/196198322-1e92747a-c7ea-41e6-8620-71b54a2932b0.mp4

costasovo:
@JanJakes I added one new commit. I forgot to add the JS global variable to the wizard page.
@Aschepikov Good catch! I was testing only on the `mailpoet-woocommerce-setup` page and forgot to update data on the Wizard page. Let's wait for the code review of changes, and then please, you can check again.

JanJakes (approved):

Aschepikov (approved):
Verified, GJ!

PR #4421, opened 2022-10-10
Update \MailPoet\Segments\WooCommerce::updateNames() query to work with Woo HPOS [MAILPOET-4711]

rodrigoprimo:
## Description

This PR changes `\MailPoet\Segments\WooCommerce::updateNames()` to work with WooCommerce High Performance Orders Storage. To achieve that an `if` was added to use two different queries depending on whether Woo HPOS is enabled or not. It was also necessary to add a new Woo helper method to get the name of the new table that stores information about addresses.

## Code review notes

_N/A_

## QA notes

Test that MailPoet can correctly synchronize WooCommerce customers and create corresponding MailPoet subscribers. More specifically, check that the synchronization of the first and last names works as expected. To do that, install WooCommerce and create a few customers (/wp-admin/user-new.php). Then install MailPoet, activate the WooCommerce integration, and check that the corresponding subscribers were created. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4711]

## After-merge notes

_N/A_

[MAILPOET-4711]: https://mailpoet.atlassian.net/browse/MAILPOET-4711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/WooCommerce.php
+++ b/mailpoet/lib/Segments/WooCommerce.php
@@ -338,18 +338,43 @@ class WooCommerce {
     }
     $subscribersTable = $this->entityManager->getClassMetadata(SubscriberEntity::class)->getTableName();
 
-    $metaKeys = [
-      '_billing_first_name',
-      '_billing_last_name',
-    ];
-    $metaData = $this->connection->executeQuery("
-      SELECT post_id, meta_key, meta_value
-      FROM {$wpdb->postmeta}
-      WHERE meta_key IN (:metaKeys) AND post_id IN (:postIds)
-    ",
-      ['metaKeys' => $metaKeys, 'postIds' => array_values($orders)],
-      ['metaKeys' => Connection::PARAM_STR_ARRAY, 'postIds' => Connection::PARAM_INT_ARRAY]
-    )->fetchAllAssociative();
+    if ($this->woocommerceHelper->isWooCommerceCustomOrdersTableEnabled()) {
+      $addressesTableName = $this->woocommerceHelper->getAddressesTableName();
+      $metaData = [];
+      $results = $this->connection->executeQuery("
+        SELECT order_id, first_name, last_name
+        FROM {$addressesTableName}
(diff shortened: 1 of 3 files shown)


mailpoet/lib/Segments/WooCommerce.php
```diff
@@ -338,18 +338,43 @@ private function updateNames(array $orders): void {
...
-    $metaData = $this->connection->executeQuery("
-      SELECT post_id, meta_key, meta_value
-      FROM {$wpdb->postmeta}
-      WHERE meta_key IN (:metaKeys) AND post_id IN (:postIds)
-    ",
-      ['metaKeys' => $metaKeys, 'postIds' => array_values($orders)],
-      ['metaKeys' => Connection::PARAM_STR_ARRAY, 'postIds' => Connection::PARAM_INT_ARRAY]
-    )->fetchAllAssociative();
+    if ($this->woocommerceHelper->isWooCommerceCustomOrdersTableEnabled()) {
+      $addressesTableName = $this->woocommerceHelper->getAddressesTableName();
```
samnajian, line 342: Given that the method is already rather big and we are dealing with a paradigm shift on WC side, I was wondering what you think about creating a service dedicated to HPOS and using its methods instead of adding the changes directly where needed and making the logic even bigger?

rodrigoprimo: Thanks for the review, @samnajian. I'm not against the refactor that you are suggesting but at the same time I don't see a significant advantage in it, and considering that this is the last piece that is missing to support HPOS in this class, I'm inclined to prefer to not make this change at this point as it would mean revisiting all the other code changes that were already made in this class. This especially considering that ideally I would like to finish this project this week. Let me know your thoughts.

samnajian (approved):
LGTM!

Aschepikov (approved):
GJ!

PR #4422, opened 2022-10-11
Create a new API method getSubscribers [MAILPOET-4716]

lysyjan:
## Description

As a part of the CRM Jetpack integration, we want to extend our API. This is the first new method that could help the other developer integrate our plugin.

## Code review notes

I reused SubscriberListingRepository for a more straightforward implementation.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4716]

## After-merge notes

_N/A_

[MAILPOET-4716]: https://mailpoet.atlassian.net/browse/MAILPOET-4716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


doc/api_methods/GetSubscribers.md
```diff
@@ -0,0 +1,148 @@
...
+
+| Key            | Type         | Description                                                                                                       |
+| -------------- | ------------ | ----------------------------------------------------------------------------------------------------------------- |
+| status         | string       | Specific status of subscribers. One of values: `unconfirmed`, `subscribed`, `unsubscribed`, `bounced`, `inactive` |
+| list_id        | int          | List id or dynamic segment id                                                                                     |
+| min_updated_at | DateTime\int | DateTime object or timestamp of the minimal last update of subscribers                                            |
+
+## A subscriber data structure
+
+### Subscriber
```
costasovo: I have a suggestion. I guess we could link the subscriber output documentation from `getSubscriber` so that we don't have to copy it and maintain on two places. What do you think?
Something like: `To see the subscriber data structure, please check getSubscriber documentation.`

costasovo (commented):
Nice job! Thank you for fixing and updating the documentation. I have just one suggestion to discuss. Please see my comment.

woodyhayday:
Hey @lysyjan, 

Dropping in here from Avengers re: Jetpack CRM. First up, it's really helpful for you to add this endpoint, thank you very much.

I've tested it to replace the hacky backend approach we used temporarily in our [proof of concept PR](https://github.com/Automattic/zero-bs-crm/pull/2392), and it'll nearly perfectly work for that use case.

The only one request I'd have is that ideally we'd be able to see the count of a specific list without retrieving all of the subscribers. Perhaps this could be appended to the response with a key of `count` or similar. Does that make sense to you? Please let me know what you think about adding that, and thanks again.

pavel-mailpoet:
@woodyhayday Thanks for the feedback, the count endpoint will come as well, we just separated the work into two different tasks.

doc/api_methods/GetSubscribers.md
lysyjan: @costasovo Thank you for this idea. It makes sense to me. I will change it.

costasovo (approved):
LGTM!

PR #4423, opened 2022-10-12
Tests for automation REST endpoints [MAILPOET-4531]

websupporter:
## Description
Adds tests to the endpoints in the free plugin

## Linked tickets
[MAILPOET-4531]

[MAILPOET-4531]: https://mailpoet.atlassian.net/browse/MAILPOET-4531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
+++ b/mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
@@ -131,6 +131,13 @@ class WorkflowStorage {
     }, (array)$data);
   }
 
+  public function flush(): bool {
+    $workflowTable = esc_sql($this->workflowTable);
+    $versionTable = esc_sql($this->versionsTable);
+    return $this->wpdb->query("truncate $workflowTable;") === true &&
+      $this->wpdb->query("truncate $versionTable;") === true;
+  }
+
   private function getWorkflowHeaderData(Workflow $workflow): array {
     $workflowHeader = $workflow->toArray();
     unset($workflowHeader['steps']);
--- a/mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
+++ b/mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
@@ -131,7 +131,7 @@ class WorkflowStorage {
     }, (array)$data);
   }
 
-  public function flush(): bool {
+  public function truncate(): bool {
     $workflowTable = esc_sql($this->workflowTable);
     $versionTable = esc_sql($this->versionsTable);
     return $this->wpdb->query("truncate $workflowTable;") === true &&
--- a/mailpoet/tests/integration/REST/Automation/Workflows/WorkflowPutEndpointTest.php
+++ b/mailpoet/tests/integration/REST/Automation/Workflows/WorkflowPutEndpointTest.php
@@ -28,7 +28,7 @@ class WorkflowPutEndpointTest extends AutomationTest
     $this->workflowStorage = $this->diContainer->get(WorkflowStorage::class);
     $this->createWorkflow = $this->diContainer->get(CreateWorkflowFromTemplateController::class);
     $this->workflow = $this->createWorkflow->createWorkflow('simple-welcome-email');
-    assert($this->workflow instanceof Workflow);
+    $this->assertInstanceOf(Workflow::class, $this->workflow);
   }
 
   public function testGuestNotAllowed(): void {
@@ -49,14 +49,14 @@ class WorkflowPutEndpointTest extends AutomationTest
     ], $data);
 
     $workflow = $this->workflowStorage->getWorkflow($this->workflow->getId());
-    assert($workflow instanceof Workflow);
+    $this->assertInstanceOf(Workflow::class, $workflow);
     $this->assertSame('Simple welcome email', $workflow->getName());
   }
 
   public function testUpdateWorkflow(): void {
     $changes = [];
     $trigger = $this->workflow->getTrigger('mailpoet:someone-subscribes');
-    assert($trigger instanceof Step);
+    $this->assertInstanceOf(Step::class, $trigger);
     $changes[$trigger->getId()] = [
       'args' => [
         'segment_ids' => [1,2]
@@ -75,9 +75,9 @@ class WorkflowPutEndpointTest extends AutomationTest
     );
 
     $updatedWorkflow = $this->workflowStorage->getWorkflow($this->workflow->getId());
-    assert($updatedWorkflow instanceof Workflow);
+    $this->assertInstanceOf(Workflow::class, $updatedWorkflow);
     $updatedTrigger = $updatedWorkflow->getTrigger('mailpoet:someone-subscribes');
-    assert($updatedTrigger instanceof Step);
+    $this->assertInstanceOf(Step::class, $updatedTrigger);
 
     /** Ensure the old workflow does not already contain the values we attempt to change */
     $this->assertNotSame($changes[$trigger->getId()]['args'], $trigger->getArgs());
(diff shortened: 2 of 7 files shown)


mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -131,6 +131,13 @@ public function getActiveWorkflowsByTrigger(Trigger $trigger): array {
     }, (array)$data);
   }
 
+  public function flush(): bool {
```
JanJakes: Wouldn't `truncate` be a better method name? This can be confused with Doctrine's `flush`, which does something else (and much safer 😄) — it syncs changes from memory to a DB. I guess this could also be somewhere in the tests only but no strong opinion about where it should lie.

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsGetTest.php
```diff
@@ -24,4 +47,133 @@ public function testRequest(): void {
...
+    );
+  }
+
+  public function _after() {
+    $this->workflowStorage->flush();
+    foreach ($this->userIds as $userId) {
+      is_multisite() ? wpmu_delete_user($userId) : wp_delete_user($userId);
+    }
+    parent::_after();
+  }
```
JanJakes, line 175: I see some whitespacing issues here but I think that's due to https://a8c.slack.com/archives/C01GH835U90/p1663847701224009 so let's keep it and fix it there...

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsGetTest.php
```diff
@@ -24,4 +47,133 @@ public function testRequest(): void {
...
+      $this->assertEquals($status, $result['data'][0]['name']);
+      $this->assertEquals($status, $result['data'][0]['status']);
+    }
+  }
+
+  /**
+   * This small helper method can quickly assert strings or integers from
+   * the REST API.
+   */
+  private function assertWorkflowRestData($expectation, $data) {
```
JanJakes, line 107: Couldn't we just format the `$expectation` to exactly the array we expect to get and then check it's the same as the response? Would be kind of a manual "rewrite" of the response building methods, but I think it may be safer — if we change some Workflow property, for instance (a BC break), such a "hardcoded" test would catch it, while this may be more prone to carry the error as well.

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsCreateFromTemplateTest.php
```diff
@@ -35,4 +60,21 @@ public function testWorkflowsCreatedFromTemplatesAreCreatedInDraftStatus(): void
     expect($createdWorkflow->getStatus())->equals('draft');
   }
 
+  public function testWorkflowsCreatedFromTemplatesReturnsWorkflowId(): void {
+    $storage = ContainerWrapper::getInstance()->get(WorkflowStorage::class);
```
JanJakes: I guess you can use `$this->workflowStorage` here.

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowPutEndpointTest.php
```diff
@@ -0,0 +1,138 @@
...
+
+  /** @var Workflow */
+  private $workflow;
+
+  public function _before() {
+    parent::_before();
+    $this->workflowStorage = $this->diContainer->get(WorkflowStorage::class);
+    $this->createWorkflow = $this->diContainer->get(CreateWorkflowFromTemplateController::class);
+    $this->workflow = $this->createWorkflow->createWorkflow('simple-welcome-email');
+    assert($this->workflow instanceof Workflow);
```
JanJakes: I've seen multiple `assert(... instanceof ...)` throughout this PR. Could we use `$this->assertInstanceOf(...)` instead?

The reason is that we may add a sniff to ban `assert` in the future because in the `lib` code, `assert`s are marked as dangerous by some security plugins. And since our sniffs for the `lib` and `tests` are the same, this could cause some problems.

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsGetTest.php
```diff
@@ -4,11 +4,34 @@
...
+    parent::_before();
+    $this->workflowStorage = $this->diContainer->get(WorkflowStorage::class);
+    $userId = wp_insert_user([
+      'display_name' => 'test',
+      'roles' => ['editor'],
+      'email' => 'test@mailpoet.com',
+      'user_pass' => 'abc',
+      'user_login' => 'workflows-get-endpoint-test',
+    ]);
+    assert(is_numeric($userId));
```
JanJakes: And in this case, we can also use `$this->assertIsNumeric(...)`.

JanJakes (commented):
Good stuff!

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
websupporter: Good catch! Yes, definitely.

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsGetTest.php
websupporter: It is a bit a complicated helper function, that is true. When I wrote it expected more requests would be necessary to validate. Ended up being only two :laughing: 

I still like it. It always makes sure `steps` is not existent in every request. It checks that date values are at least readable as such. I did not want to compare exact dates, as that might become flaky. Another thought would have been to compare something like "updated_at>=two seconds ago", but opted for just checking we can read it.

The last bit of logic is the inner logic of `draft` status and `activated_at`. Thinking about it, the logic should be the other way around actually. In case `activated_at` is `null` it can only be draft. But an activated workflow can return to draft and still have the `activated_at` timestamp. I addressed that in a664c1f03.

So the idea of a bit more complex helper function is to check always a bit more. Usually, you would just compare exact values, which the method does, but why not ensure multiple times that dates are readable, `steps` are not send...

I am not quite sure, why this helper could be more prone to not catching BC breaking changes :thinking:

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsCreateFromTemplateTest.php
websupporter: Addressed in 0038088ee

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowPutEndpointTest.php
websupporter: Addressed in c95dc0ffe

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
websupporter: Addressed in afd1e0cf1

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsGetTest.php
websupporter: Addressed here: https://github.com/mailpoet/mailpoet/commit/c95dc0ffe7c959395012effca5c5885ad6f7ad2f#diff-b517c4a6320ddb55cbe3ff1aec93b5210bb4f3ed2845434e9d8eda69b105e6c6R31

websupporter:
Thanks for your review @JanJakes, replied and fixed most of it.

mailpoet/tests/integration/REST/Automation/Workflows/WorkflowsGetTest.php
JanJakes: I think what I wanted to say is that iterating the workflow data may result in unexpected data changes passing the test. If we internally change property `foo` from `int` to `string`, it will automatically appear as changed in the expectation (I think) and thus not catch the BC break on the API.

So there's a difference using:
```php
$this->assertSame(['foo' => 1], $response);
```

And:
```php
$this->assertSame($expectedDataObtainedFromWorkflow, $response);
```

But what I see now is that `$expectation` data is constructed pretty much manually so I guess it's good. And maybe this will belong to some "mapper" tests once we extract the response building logic out of the endpoints.

JanJakes (approved):
Thanks for the fixes!

PR #4424, opened 2022-10-12
Prevent displaying google analytics to free users [PREMIUM-200]

samnajian:
## Description

This PR makes sure we are not showing an enabled GA field to users without active premium plugins.

- Disables the field
- Opens up the modal to activate plugin plugin
- Removes the default value if the field is disabled
- Shows a `no-allowed`  cursor on hover of the disabled field, please note that this is added for all disabled fields.

## QA notes
Please follow the steps from the Jira ticket.

Note if you encounter problem:
Make sure you have uploaded the latest premium trunk if you hit the following error message when trying to activate the premium plugin using the upsell modal.
```
{error: "bad_request", message: "Interface 'MailPoet\Automation\Engine\Workflows\Action' not found"}
error: "bad_request"
message: "Interface 'MailPoet\\Automation\\Engine\\Workflows\\Action' not found" 
```

## Linked tickets

[PREMIUM-200]

[PREMIUM-200]: https://mailpoet.atlassian.net/browse/PREMIUM-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-plugin/_forms.scss
+++ b/mailpoet/assets/css/src/components-plugin/_forms.scss
@@ -232,3 +232,7 @@ progress::-moz-progress-bar {
 .mailpoet-form-field-tags label.components-form-token-field__label {
   display: none;
 }
+
+.mailpoet-form-field-disabled {
+  cursor: not-allowed;
+}
--- a/mailpoet/assets/js/src/form/fields/field.jsx
+++ b/mailpoet/assets/js/src/form/fields/field.jsx
@@ -2,6 +2,7 @@ import { Component } from 'react';
 import { FormFieldText } from 'form/fields/text.jsx';
 import jQuery from 'jquery';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 
 import { FormFieldTextarea } from 'form/fields/textarea.jsx';
 import { FormFieldSelect } from 'form/fields/select.jsx';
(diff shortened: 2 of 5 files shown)


lysyjan (approved):

Aschepikov (commented):
Hi, @pavel-mailpoet @NeosinneR  I think that the specs, that this google analytics premium feature is connected just to is there premium plugin of any version or no, are not correct. Because I can download mailpoet premium plugin from the internet of any version and in this case, I will see this google analytics  field without any subscription. As I understand it should work the same way as the banner in the statistics.
For example, I just download the premium version from the "[left](https://wp-club.ru/wordpress/wordpress-plaginy/mailpoet-premium/)" site and I don't have any key and this version is 3.80 but I can use google analytics
image
image
image
If I haven't real premium (just download it from the internet) but have google analytics acc and theoretically I can use it without mailpoet

pavel-mailpoet:
@Aschepikov, you are right. The same behaviour as the statistics would be better. But I am not too worried. Suppose someone wants to download a plugin full of malware from some unknown site they can. I think we can keep the behaviour implemented in this Pull Request for now. We have better stuff to do than worrying about users downloading pirated software.

JanJakes:
Just somehow got here; do you guys think the `not-allowed` cursor is a good idea? It is not common to do this for disabled elements because it makes them more prominent, and the cursor can feel "scary", especially on Windows 7.

I was a bit afraid seeing this:
> Shows a no-allowed cursor on hover of the disabled field, please note that this is added for all disabled fields.

Worried that we would start seeing this in the Gutenberg components. But from the code, it seems it applies only to `.mailpoet-form-field-disabled` so maybe it's "OK"? Anyway, just double-checking.

samnajian:
> Just somehow got here; do you guys think the `not-allowed` cursor is a good idea? It is not common to do this for disabled elements because it makes them more prominent, and the cursor can feel "scary", especially on Windows 7.
> 
> I was a bit afraid seeing this:
> 
> > Shows a no-allowed cursor on hover of the disabled field, please note that this is added for all disabled fields.
> 
> Worried that we would start seeing this in the Gutenberg components. But from the code, it seems it applies only to `.mailpoet-form-field-disabled` so maybe it's "OK"? Anyway, just double-checking.

@JanJakes yea checked before adding it to all fields of this type, so it's not added to Gutenberg.

PR #4425, opened 2022-10-13
Feat/feels forced to allow 3rd party libs to be able to contact support - MAILPOET-4657

triple0t:
## Description

Update the welcome wizard and remove the dependency between loading `3rd-party libraries` and `tracking (data sharing)`

## QA notes

To access the welcome wizard, visit {Site_url}/wp-admin/admin.php?page=mailpoet-welcome-wizard e.g. http://gs.qawp.net/wp-admin/admin.php?page=mailpoet-welcome-wizard

* In the wizard, disable `3rd-party libraries`, but try to enable `data sharing`. Notice the toggle is not dependent on each other.

## Linked tickets

[MAILPOET-4657](https://mailpoet.atlassian.net/browse/MAILPOET-4657)


mailpoet/views/welcome_wizard.html
```diff
@@ -26,22 +26,13 @@
...
-'welcomeWizardTrackingText': __('Gathering usage data allows us to make MailPoet better — the way you use MailPoet will be considered as we evaluate new features, judge the quality of an update, or determine if an improvement makes sense.'),
-'welcomeWizardTrackingList1': __('Any personal data'),
-'welcomeWizardTrackingList2': __('Email addresses'),
-'welcomeWizardTrackingList3': __('Login and passwords'),
-'welcomeWizardTrackingList4': __('Content of your emails'),
-'welcomeWizardTrackingList5': __('Open and click rates'),
-'welcomeWizardTrackingLink': __('Read more about what we collect.'),
+'welcomeWizardUsageTrackingStepLibs3rdPartyLabelNote': __('If enabled, we may load the Google Fonts library and [link]other 3rd-party libraries we use[/link].'),
+'welcomeWizardUsageTrackingStepLibs3rdPartyLabelNoteNote': __('Note'),
+'welcomeWizardUsageTrackingStepLibs3rdPartyLabel': __('Enable better-looking Google fonts in emails and show contextual help articles in MailPoet?'),
```
triple0t, line 35: Hi @lysyjan

For this part, with regards to Translations, would it be better to create a new key for the text string, or is it alright the way it is? 

I edited the contents of the text tag, and I'm wondering if it would show up in the Translation client as needing to be re-translated.

lysyjan: @triple0t If I am not wrong, our command extracts strings from the `__()` function. So we don't need to change the keys. [Here](https://mailpoetp2.wordpress.com/team-handbook/engineering-process/plugins-translations/) is a P2 page about how translations work.

triple0t: Thanks @lysyjan. I will check it out.

> our command extracts strings from the __() function.

To confirm, 
e.g 
**before**
```
'welcomeWizardUsageTrackingStepLibs3rdPartyLabelNoteNote': __('Important'),
```

**after**
```
'welcomeWizardUsageTrackingStepLibs3rdPartyLabelNoteNote': __('Note'),
```

The Key is the same `welcomeWizardUsageTrackingStepLibs3rdPartyLabelNoteNote` but the text content is not.

Here, the command will extract `'Note'` as the new text, right?

lysyjan: @triple0t Yes, this is exactly how I understand it.

triple0t: Awesome 👏, Thank you @lysyjan

lysyjan (approved):
LGTM

Aschepikov (approved):
GJ!

PR #4426, opened 2022-10-13
Users should be able to save an email as a draft even if the required fields are empty [MAILPOET-4591]

samnajian:
## Description

This PR completely disables validation when saving a draft.
- Disables the validation
- Removes redundant codes for selectively disabling validation

## Code review notes
Please note that already have an acceptance test for the list-selector, which covers the change as well, so no new acceptance is added.

## Linked PRs

[Old PR](https://github.com/mailpoet/mailpoet/pull/4328) were we disabled the validation only for the list selector.

## Linked tickets

[MAILPOET-4591]

[MAILPOET-4591]: https://mailpoet.atlassian.net/browse/MAILPOET-4591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/newsletters/send.jsx
+++ b/mailpoet/assets/js/src/newsletters/send.jsx
@@ -114,7 +114,6 @@ class NewsletterSendComponent extends Component {
       item: {},
       loading: true,
       thumbnailPromise: null,
-      isSavingDraft: false,
       showPremiumModal: false,
     };
   }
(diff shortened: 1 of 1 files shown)


lysyjan (dismissed):

Aschepikov (commented):
Hi, I found one issue with the validation of the sender when I am actually sending it.
 There Is no error about "`Not an authorized sender email address. [Authorize it now.]`" 
 
image
And I can send it from the incorrect sender 
image

For example now on prod 3.101 
image

samnajian:
> Hi, I found one issue with the validation of the sender when I am actually sending it. There Is no error about "`Not an authorized sender email address. [Authorize it now.]`"
> 
> image And I can send it from the incorrect sender image
> 
> For example now on prod 3.101 image


@Aschepikov Just noticed the behavior you reported happens on trunk, can you please confirm? If that's the case we need a separate ticket to fix the issue, I can include the fix for that in this PR as well though.

samnajian:
@Aschepikov I checked locally and it seems to work correctly, please note that some validations only work when MSS is selected as your sending method.

Aschepikov (approved):
Sorry my fault with mss, thank you, verified

PR #4427, opened 2022-10-13
Create a new API method getSubscribersCount [MAILPOET-4717]

lysyjan:
## Description

As a part of the CRM Jetpack integration, we want to extend our API. This is the second method that could help the other developer integrate our plugin.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4717]

## After-merge notes

_N/A_

[MAILPOET-4717]: https://mailpoet.atlassian.net/browse/MAILPOET-4717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/MP/v1/API.php
+++ b/mailpoet/lib/API/MP/v1/API.php
@@ -88,6 +88,10 @@ class API {
     return $this->subscribers->getSubscribers($filter, $limit, $offset);
   }
 
+  public function getSubscribersCount(array $filter = []): int {
+    return $this->subscribers->getSubscribersCount($filter);
+  }
+
   public function isSetupComplete() {
     return !(
       $this->changelog->shouldShowWelcomeWizard()
--- a/mailpoet/lib/API/MP/v1/Subscribers.php
+++ b/mailpoet/lib/API/MP/v1/Subscribers.php
@@ -247,18 +247,6 @@ class Subscribers {
     return $this->subscribersResponseBuilder->build($subscriber);
   }
 
-  /**
-   * @param array $filter {
-   *     Optional. Filters to retrieve subscribers.
-   *
-   *     @type string        $status       One of values: subscribed, unconfirmed, unsubscribed, inactive, bounced
-   *     @type int           $listId       id of a list or dynamic segment
-   *     @type \DateTime|int $minUpdatedAt DateTime object or timestamp of last update of subscriber.
-   * }
-   * @param int $limit
-   * @param int $offset
-   * @return array
-   */
   public function getSubscribers(array $filter, int $limit, int $offset): array {
     $listingDefinition = $this->buildListingDefinition($filter, $limit, $offset);
     $subscribers = $this->subscriberListingRepository->getData($listingDefinition);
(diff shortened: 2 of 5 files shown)


mailpoet/lib/API/MP/v1/API.php
```diff
@@ -88,6 +88,10 @@ public function getSubscribers(array $filter = [], int $limit = 50, int $offset
     return $this->subscribers->getSubscribers($filter, $limit, $offset);
   }
 
+  public function getSubscribersCount(array $filter = []): int {
```
samnajian, line 91: was wondering why this method is defined before the last method and not after it?

mailpoet/lib/API/MP/v1/Subscribers.php
```diff
@@ -260,6 +260,31 @@ public function unsubscribeFromLists($subscriberIdOrEmail, array $listIds): arra
...
+   * @param array $filter {
+   *     Optional. Filters to retrieve subscribers.
+   *
+   *     @type string        $status       One of values: subscribed, unconfirmed, unsubscribed, inactive, bounced
+   *     @type int           $listId       id of a list or dynamic segment
+   *     @type \DateTime|int $minUpdatedAt DateTime object or timestamp of last update of subscriber.
+   * }
+   * @return int
+   */
+  public function getSubscribersCount(array $filter): int {
```
samnajian: I think defining the filter as a class and passing a class instance would make filter definition and validation and sanitation ( as done in `buildListingDefinition`) more convenient, basically most parts of `buildListingDefinition` method can be encapsulated in that proposed class, what do you think?

mailpoet/lib/API/MP/v1/API.php
lysyjan: Two previous methods are `getSubscriber()` and `getSubscribers()`. It doesn't make sense to me put the new one at the end.

mailpoet/lib/API/MP/v1/Subscribers.php
lysyjan: I am not sure what you mean here.

If your idea is to pass the whole filter argument to the listing definition, that is not possible because the API uses `status` as a filter, but the listing uses it as a group. Also, we use internal `segment`, and the API expects `listId`

samnajian: For future reference: We've decided to leave it for now as currently, the number of filters is very limited, later on when we add more filter, it'll make more sense to pass the filters array in form of a class instance instead of an array.

costasovo: @samnajian, @lysyjan  I just want to add here that the $filter as the array was inspired by [get_posts](https://github.com/WordPress/WordPress/blob/c11ce53dceee9d8af402f16af23ab4aa61e6a074/wp-includes/post.php#L2380) and also a few older methods like addSubscriber accepts arrays.
Having it as a class has some advantages, like we can enforce types of parameters using type hints in setters.

> later on when we add more filter, it'll make more sense to pass the filters array in form of a class instance instead of an array.

Switching from array to class would be a BC break, and we will still need to support both array and class. The two new methods `getSubscribers` and `getSubscribersCount` were not released yet, so we still can make a decision. I don't have super strong opinion here. Array is used in WP and in older methods in our API, the class is better for types. So it is kind of alignment with WP and consistency VS types checking.

samnajian: Thanks for your input @costasovo especially since I didn't check if these methods are released or not yet, I personally like the advantage of type-checking as it means a lot of issues can be prevented just by my cold IDE, but that's also very opinionated when it gets to PHP. To me, an array is more suited for pure data, which doesn't carry logic or meaning, like a POJO and objects/class instances for anything that carries logic with itself.  I'll still leave this decision to @lysyjan since we can always refactor this to use class instances when the need is more vivid.

samnajian (approved):

veljkho (approved):

PR #4430, opened 2022-10-14
Remove babel plugin typescript to proptypes [MAILPOET-4521]

samnajian:
## Description

This PR removes `babel-plugin-typescript-to-proptypes` and updates the `tsx` files to remove usages `propTypes`.

## QA notes
Updated areas are as follows:

- Authorized sender email and domain modal 
- Forms that have `Tags` 
- Forms with inline styling that's used in form editor

 for the above-mentioned areas, the rendering and functionality should be checked.

## Linked PRs

The same changes on the premium plugin  [635](https://github.com/mailpoet/mailpoet-premium/pull/635)

## Linked tickets

[MAILPOET-4521]

[MAILPOET-4521]: https://mailpoet.atlassian.net/browse/MAILPOET-4521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/form/fields/tokenField.tsx
+++ b/mailpoet/assets/js/src/form/fields/tokenField.tsx
@@ -1,7 +1,16 @@
-import PropTypes from 'prop-types';
-import { TokenField } from 'common/form/tokenField/tokenField';
+import { TokenFieldProps, TokenField } from 'common/form/tokenField/tokenField';
+import { FormTokenItem } from '../../automation/integrations/mailpoet/components/form-token-field';
 
-function getItems(endpoint: string) {
+interface TokenFormFieldProps {
+  onValueChange: TokenFieldProps['onChange'];
+  item?: Record<string, FormTokenItem[]>;
+  field: Omit<TokenFieldProps, 'id' | 'onChange' | 'selectedValues'> & {
+    endpoint: string;
+    getName: (item: FormTokenItem) => string;
+  };
+}
+
+function getItems(endpoint: string): FormTokenItem[] {
   let items = [];
   if (typeof window[`mailpoet_${endpoint}`] !== 'undefined') {
     items = window[`mailpoet_${endpoint}`];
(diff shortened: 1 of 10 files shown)


mailpoet/assets/js/src/form/fields/tokenField.tsx
```diff
@@ -10,12 +19,15 @@ function getItems(endpoint: string) {
...
+function FormFieldTokenField(props: TokenFormFieldProps) {
+  const selectedValues: TokenFieldProps['selectedValues'] = Array.isArray(
+    props.item[props.field.name],
+  )
+    ? props.field.name &&
+      props.item[props.field.name].map((item) => props.field.getName(item))
     : [];
 
-  let suggestedValues = [];
+  let suggestedValues;
```
triple0t: Quick question:
May I know why you removed the default value here?

Looking through the rest of the code in that file, `suggestedValues` reassignment is covered within conditional checks. This might cause an error when both checks are false.

triple0t (commented):
Looks good.
I have a question. Kindly locate it below.

Also, I noticed a few more places `defaultProps` is being used in `.tsx` files. 
Screenshot 2022-10-18 at 10 35 27 AM
That won't be affected by this change, right?

mailpoet/assets/js/src/form/fields/tokenField.tsx
samnajian: thanks for noticing, since we have `else if ` statement the default should stay here, but the reason to touch it has been TS nagging about the `any[]` type assignment, fixed in bacbf08493a1c5f1718e2adebbc8f3e1d83144d0

triple0t (approved):

Aschepikov (approved):
Verified, GJ!

PR #4431, opened 2022-10-14
Feat/it is unclear what immediately means when setting up automatic emails - MAILPOET-4658

triple0t:
## Description

Update UI for the setup of the following emails
* Automatic emails: First purchase, Abandoned shopping cart, Purchased in category, Purchased product
* Post Notification

## Linked tickets

[MAILPOET-4658](https://mailpoet.atlassian.net/browse/MAILPOET-4658)

### DIFF
--- a/mailpoet/assets/js/src/newsletters/automatic_emails/events/event_scheduling.jsx
+++ b/mailpoet/assets/js/src/newsletters/automatic_emails/events/event_scheduling.jsx
@@ -6,6 +6,7 @@ import { MailPoet } from 'mailpoet';
 import { Selection } from 'form/fields/selection.jsx';
 import { FormFieldText } from 'form/fields/text.jsx';
 import { timeDelayValues } from 'newsletters/scheduling/common.jsx';
+import { Grid } from 'common/grid';
 
 const defaultAfterTimeType = 'immediate';
 const defaultAfterTimeNumber = 1;
@@ -142,10 +143,15 @@ class EventScheduling extends Component {
     const { event } = this.props;
     return (
       <>
-        <div className="mailpoet-grid-column mailpoet-flex">
+        <h4> {MailPoet.I18n.t('whenToSendMail')} </h4>
+
+        <Grid.CenteredRow className="mailpoet-re-engagement-scheduling">
           {this.displayAfterTimeNumberField()}
           {this.displayAfterTimeTypeOptions()}
-        </div>
+
+          {event.afterDelayText && <p>{event.afterDelayText}</p>}
+        </Grid.CenteredRow>
+
         <div className="mailpoet-form-errors" />
         <div className="mailpoet-gap" />
         {event.schedulingReadMoreLink && (
@@ -175,6 +181,7 @@ EventScheduling.propTypes = {
   onValueChange: PropTypes.func,
   event: PropTypes.shape({
     defaultAfterTimeType: PropTypes.string,
+    afterDelayText: PropTypes.string,
     timeDelayValues: PropTypes.objectOf(
       PropTypes.shape({
         text: PropTypes.string,
--- a/mailpoet/assets/js/src/newsletters/types/notification/notification.jsx
+++ b/mailpoet/assets/js/src/newsletters/types/notification/notification.jsx
@@ -80,7 +80,6 @@ class NewsletterNotificationComponent extends Component {
           <Heading level={4}>
             {MailPoet.I18n.t('selectEventToSendPostNotificationEmail')}
           </Heading>
-          <h5> {MailPoet.I18n.t('selectFrequency')} </h5>
 
           <NotificationScheduling
             item={this.state}
--- a/mailpoet/assets/js/src/newsletters/types/notification/notification.jsx
+++ b/mailpoet/assets/js/src/newsletters/types/notification/notification.jsx
@@ -77,7 +77,10 @@ class NewsletterNotificationComponent extends Component {
         />
 
         <Grid.Column align="center" className="mailpoet-schedule-email">
-          <Heading level={4}>{MailPoet.I18n.t('selectFrequency')}</Heading>
+          <Heading level={4}>
+            {MailPoet.I18n.t('selectEventToSendPostNotificationEmail')}
+          </Heading>
+          <h5> {MailPoet.I18n.t('selectFrequency')} </h5>
 
           <NotificationScheduling
             item={this.state}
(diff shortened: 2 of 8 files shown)


websupporter (dismissed):
lgtm. Thank you @triple0t!

Aschepikov (commented):
Hi, I have a question about changes in the post notifications, I didn't see anything new here.
image
image

And in the first purchase, it should be "after **the** first purchase" like in the ticket. Or "after first purchase" like in the pr 
image

Aschepikov (commented):
@triple0t  And there is a conflict, can you resolve it pls

triple0t:
Hey @Aschepikov, Thanks for the review.

> I have a question about changes in the post notifications, I didn't see anything new here.

compare-post-notification-image

Changes
* New header text: "When to send this post notification email?"
* Reduced the size of the previous header: "Select a frequency"

**Note**:  The post-notification part of the task does not require an after text, unlike the other sections, e.g., First purchase.


> And in the first purchase, it should be "after the first purchase" like in the ticket. Or "after first purchase" like in the pr

I think any of them is fine. 
Unless @pavel-mailpoet or @NeosinneR have other opinions.


> @triple0t And there is a conflict, can you resolve it pls

Noted.

pavel-mailpoet:
> Unless @pavel-mailpoet or @NeosinneR have other opinions.

I don't. It is fine for me.

NeosinneR:
> The post-notification part of the task does not require an after text, unlike the other sections, e.g., First purchase.

@triple0t now that we have the "When to send this post notification email?" title, the "Select a frequency" subtitle feels redundant - I think you can remove it.

Also, what do you think about adding "after publishing a post" for the "immediately" option (just here, doesn't have to be for specific periods), to explain what immediately means in this context (similarly as you added this explanation to other emails). 

> And in the first purchase, it should be "after the first purchase" like in the ticket. Or "after first purchase" like in the pr

Grammar-wise, there should be articles, as we are talking about specific or any subjects. So for each email, it should be
- `after abandoning the cart` (Abandoned cart)
- `after the first purchase` (First purchase)
- `after a purchase` (Purchased in product / category)
- `after publishing a post` (Post notification)

Aschepikov (commented):
Hi, it looks strange to me that everywhere this massage is separate and in the post notification it is in the list. Should we make it the same way as others?
image
image

And I found an issue with overlapped here in the first purchase 
image
image

triple0t:
Thank you for the Review, @Aschepikov 🙌🏿

> Hi, it looks strange to me that everywhere this massage is separate and in the post notification it is in the list. Should we make it the same way as others?

The implementation was based on @NeosinneR suggestions [here](https://github.com/mailpoet/mailpoet/pull/4431#issuecomment-1293682196)
> what do you think about adding "after publishing a post" for the "immediately" option (just here, doesn't have to be for specific periods)

I think it's okay for the post notifications.


> And I found an issue with overlapped here in the first purchase

Thank you for noticing the issue.

This has been there for a while now; unfortunately, the CSS fix for this can take anywhere from a few minutes to a few hours. I really don’t want to invest that much time in a minor fix right now. 

Could you please create a new ticket for that issue, and we can work on it later?

NeosinneR:
@triple0t 

> it looks strange to me that everywhere this massage is separate and in the post notification it is in the list. Should we make it the same way as others?

I agree with @Aschepikov. One of the reasons we are doing so many UI fixes in the onboarding improvements epic is that in the past, we weren't thorough and let these little inconsistencies be in the plugin. Many times, we create a follow-up ticket, but we'll never get to it, because there is always something more important than just fixing an "unimportant" UI element. I'd please ask you if this can be changed as @Aschepikov suggested, so it's consistent with other emails. 

> This has been there for a while now

Now, that we also added the phrase explaining when it's sent (e.g. "after the first purchase") the "later" can be removed from the options, as it no longer makes sense (this is also applicable to other emails, not just this one). And it would also solve the overlaying issue.

triple0t:
Hi @NeosinneR 

Here is an **example** of how the post notification screen would appear if implemented as @Aschepikov  suggested.

Immediately option
Screenshot 2022-11-04 at 3 12 51 PM

Once a day option
Screenshot 2022-11-04 at 3 13 08 PM

Weekly option
Screenshot 2022-11-04 at 3 13 16 PM

Monthly option
Screenshot 2022-11-04 at 3 13 27 PM

Monthly every _ option
Screenshot 2022-11-04 at 3 13 36 PM

Apart from the Immediately option, you may notice that the addition of the text "after publishing a post" for other options makes it challenging to understand what the options mean and can add to more confusion.

e.g.
I want to send a post notification **Weekly** on **Wednesday** by **2:00 AM**. What does "after publishing a post" mean here?
* Do I have to publish a post by 2:00 AM?


=====

I would still insist we go along with your initial suggestion [here](https://github.com/mailpoet/mailpoet/pull/4431#issuecomment-1293682196) 

> What do you think about adding "after publishing a post" for the "immediately" option (just here, doesn't have to be for specific periods) 

It makes more sense for the "immediately" option.

e.g Send  a post notification **immediately** after publishing a post

But not so much for the other options.



> Now, that we also added the phrase explaining when it's sent (e.g. "after the first purchase") the "later" can be removed from the options, as it no longer makes sense (this is also applicable to other emails, not just this one). And it would also solve the overlaying issue.

Noted. This will be done.

NeosinneR:
@triple0t I agree that it doesn't make sense to add also to other options. That's why I also suggested adding it only to the "immediately" option.

But what @Aschepikov was suggesting, and I also agree with, it would be nice to make it consistent with how it's implemented. In all other automatic emails, the "after X" is not part of the select box, but it's written outside of it. Like this:

198989159-f6f91dd6-5cf0-4435-b7f2-8b6332210303

So it would be good to have it like this also for "When to send this post notification email?" but only for the "Immediately" option. Other options with specific timing would be unchanged, like this:

199997896-1608cdb7-ab0f-47a8-827a-f5febd873401

triple0t:
> @triple0t I agree that it doesn't make sense to add also to other options. That's why I also suggested adding it only to the "immediately" option.
> 
> But what @Aschepikov was suggesting, and I also agree with, it would be nice to make it consistent with how it's implemented. In all other automatic emails, the "after X" is not part of the select box, but it's written outside of it. Like this:
> 
> 198989159-f6f91dd6-5cf0-4435-b7f2-8b6332210303
> 
> So it would be good to have it like this also for "When to send this post notification email?" but only for the "Immediately" option. Other options with specific timing would be unchanged, like this:
> 
> 199997896-1608cdb7-ab0f-47a8-827a-f5febd873401

Noted 👍

Aschepikov (approved):
Verified, GJ!

PR #4434, opened 2022-10-17
Automation email workflow acceptance test [MAILPOET-4530]

websupporter:
## Description
Adds an acceptance test.
1. Creates a "simple-welcome-email" workflow and activates it
2. Adds a new subscriber and checks if the email is received.

## Code review notes
*Two bugs needed to be fixed:*
d081fa1bd0d379c6b84179dda8ce8fc04103dec9: If you do not have any workflows, the statistics queries would produce an SQL error and the page would not load properly.

9a6bf39e33d5f6ffca66145905c41336699a705f: I think this was basically a mix up what `Workflow::getStep()` actually returns. It returns a `Data\Step`, but an `Integration\Step` was needed to receive the subjectKeys.

*A11Y*
I added some labels so I was able to address a button in the acceptance test: 8f89fc1d521204b6ed59520055c225c587af1d94

## QA notes
This is an automation ticket and can be merged without QA.

## Linked tickets

[MAILPOET-4530]
[MAILPOET-4732] - I wrote an issue for the SQL error, but I think we can quickly handle it in this PR together.

[MAILPOET-4530]: https://mailpoet.atlassian.net/browse/MAILPOET-4530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4732]: https://mailpoet.atlassian.net/browse/MAILPOET-4732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/_support/AcceptanceTester.php
+++ b/mailpoet/tests/_support/AcceptanceTester.php
@@ -706,6 +706,15 @@ class AcceptanceTester extends \Codeception\Actor {
     $i->cli(['action-scheduler', 'run', '--force']);
   }
 
+  public function triggerAutomationActionScheduler(): void {
+    $i = $this;
+    // Reschedule automation trigger action to run immediately
+    $i->importSql([
+      "UPDATE mp_actionscheduler_actions SET scheduled_date_gmt = SUBTIME(now(), '01:00:00'), scheduled_date_local = SUBTIME(now(), '01:00:00') WHERE hook = 'mailpoet/automation/workflow/step' AND status = 'pending';",
+    ]);
+    $i->cli(['action-scheduler', 'run', '--force']);
+  }
+
   public function isWooCustomOrdersTableEnabled(): bool {
     return (bool)getenv('ENABLE_COT');
   }
--- a/mailpoet/tests/acceptance/Automation/CreateEmailAutomationAndWalkThroughCest.php
+++ b/mailpoet/tests/acceptance/Automation/CreateEmailAutomationAndWalkThroughCest.php
@@ -1,0 +1,113 @@
+<?php
+
+namespace MailPoet\Test\Acceptance;
+
+use Facebook\WebDriver\WebDriverKeys;
+use MailPoet\Automation\Engine\Migrations\Migrator;
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\Features\FeaturesController;
+use MailPoet\Test\DataFactories\Features;
+use MailPoet\Test\DataFactories\Settings;
+
+class CreateEmailAutomationAndWalkThroughCest
+{
+  public function _before() {
+    // @ToDo Remove once MVP is released.
+    $features = new Features();
+    $features->withFeatureEnabled(FeaturesController::AUTOMATION);
+    $container = ContainerWrapper::getInstance();
+    $migrator = $container->get(Migrator::class);
+    $migrator->createSchema();
+
+    $settings = new Settings();
+    $settings->withCronTriggerMethod('Action Scheduler');
+  }
+
+  public function createEmailWorkflowAndReceiveAnAutomatedEmail(\AcceptanceTester $i) {
+    $i->wantTo('Create a workflow to send an email after a user subscribed');
+    $i->login();
+
+    $i->amOnMailpoetPage('Automation');
+    $i->see('Automations');
+    $i->waitForText('Scale your business with advanced automations');
+    $i->dontSee('Simple welcome email');
+    $i->dontSee('Active');
+    $i->dontSee('Entered');
+
+    $i->click('New automation');
+    $i->see('Choose your automation template');
+    $i->click('Simple welcome email');
+
+    $i->waitForText('Draft');
+    $i->click('Trigger');
+    $i->fillField('When someone subscribers to the following list(s):', 'Newsletter mailing list');
+    $i->click('Delay');
+    $i->fillField('Wait for', '5');
+
+    $i->click('Send email');
+    $i->fillField('“From” name','From');
+    $i->fillField('“From” email address','test@mailpoet.com');
+    $i->fillField('Subject','Automation-Test-Subject');
+
+    $i->click('Design email');
+    $i->waitForText('Newsletters');
+    $i->click('Newsletters');
+    $i->click('button[data-automation-id="select_template_0"]');
+    $i->waitForText('Design');
+    $i->click('Save and continue');
+
+    $i->waitForText('Draft');
+
+    $i->click('Send email');
+    $i->click('Reply to');
+    $i->fillField('“Reply to” name', 'Reply');
+    $i->fillField('“Reply to” email address', 'reply@mailpoet.com');
+
+    $i->click('Activate');
+    $i->waitForText('Are you ready to activate?');
+
+    // We use a selector to be specific about which Activate button we want to click.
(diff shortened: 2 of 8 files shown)


mailpoet/tests/_support/AcceptanceTester.php
```diff
@@ -52,6 +52,13 @@ class AcceptanceTester extends \Codeception\Actor {
   const WOO_COMMERCE_MEMBERSHIPS_PLUGIN = 'woocommerce-memberships';
   const WOO_COMMERCE_SUBSCRIPTIONS_PLUGIN = 'woocommerce-subscriptions';
 
+  public function activateFeature(string $feature) {
+    $feature = strtolower($feature);
+    $this->importSql([
+      'insert into mp_mailpoet_feature_flags (name, value) values ("'. $feature . '",1)',
+    ]);
+  }
+
```
JanJakes: We already have a data factory for that: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/DataFactories/Features.php

mailpoet/tests/acceptance/Automation/CreateEmailAutomationAndWalkThroughCest.php
```diff
@@ -0,0 +1,110 @@
...
+
+class CreateEmailAutomationAndWalkThroughCest
+{
+  public function _before() {
+    // @ToDo Remove once MVP is released.
+    $container = ContainerWrapper::getInstance();
+    $migrator = $container->get(Migrator::class);
+    $migrator->createSchema();
+
+    // Set the cron trigger to "WordPress", so we can trigger the cron by visiting the page.
```
JanJakes: Just an idea, using `$this->triggerMailPoetActionScheduler()` could probably work too.

mailpoet/tests/acceptance/Automation/CreateEmailAutomationAndWalkThroughCest.php
```diff
@@ -0,0 +1,110 @@
...
+    $i->waitForText('Simple welcome email');
+    $i->see('1 Entered');
+    $i->see('1 Processing');
+    $i->see('0 Exited');
+    $i->amOnMailboxAppPage();
+    $i->see('Inbox (0)');
+
+    // Jump the waiting time by scheduling the delay action to now.
+    $i->triggerAutomationActionScheduler();
+    $i->triggerAutomationActionScheduler();
```
JanJakes: Why is it necessary to call this twice? It's the number of steps in the workflow,  right? Maybe we could add some comments like:

```php
    $i->triggerAutomationActionScheduler(); // delay step
    $i->triggerAutomationActionScheduler(); // send email step
```

 Do we want/need to do any assertions in between those two steps?

JanJakes (commented):
This is awesome!

mailpoet/tests/_support/AcceptanceTester.php
websupporter: Thank you for pointing that out. Funny, I didn't end up this method in the end. Was a residual and I forgot about it. I switched to the DataFactory in the `_before` method of the test itself. [b5b8999](https://github.com/mailpoet/mailpoet/pull/4434/commits/b5b8999931476b1fd98c53a0bac8f2535fa8c6ca)

mailpoet/tests/acceptance/Automation/CreateEmailAutomationAndWalkThroughCest.php
websupporter: Thanks. Addressed in [9b286b3](https://github.com/mailpoet/mailpoet/pull/4434/commits/9b286b3d949a2ee01d3b921e776b81ab94fa5220)

mailpoet/tests/acceptance/Automation/CreateEmailAutomationAndWalkThroughCest.php
websupporter: Addressed in [9b286b3](https://github.com/mailpoet/mailpoet/pull/4434/commits/9b286b3d949a2ee01d3b921e776b81ab94fa5220)

websupporter: > Do we want/need to do any assertions in between those two steps?

I don't think so. To me the need to run it twice is already to much "looking into the black box". On the level of acceptance test, I don't think it makes sense to dive so deep for assertions.

Now I wonder about the database lookup I actually do [to check if we have a pending task in the Action Scheduler](https://github.com/mailpoet/mailpoet/pull/4434/files/8e50e4e608127de9aa9a54b3f001da8e9df371b6#diff-68667e9409836537b4425bd4f07c6aa5f5a7f5d26e4955eb233dbe196f5343f0R86), but I guess, we can leave it as it does not do any harm :see_no_evil:

websupporter:
I rebased current trunk, fixed the test, squashed former commits ([b5b8999](https://github.com/mailpoet/mailpoet/pull/4434/commits/b5b8999931476b1fd98c53a0bac8f2535fa8c6ca), [9b286b3](https://github.com/mailpoet/mailpoet/pull/4434/commits/9b286b3d949a2ee01d3b921e776b81ab94fa5220), [8e50e4e](https://github.com/mailpoet/mailpoet/pull/4434/commits/8e50e4e608127de9aa9a54b3f001da8e9df371b6)), which belong together.

Additionally: In the UI the old Activate button still existed when the Panel was open. This created a problem as the acceptance test tried to click this old button. In [3024a84](https://github.com/mailpoet/mailpoet/pull/4434/commits/3024a849142dcad91e115ae90c0808de6d7c7c08) the button is removed and we set the focus to the new activate button, which should also help with A11Y, as the focus is now where the action is.

mailpoet/assets/js/src/automation/editor/components/header/index.tsx
```diff
@@ -21,11 +22,19 @@ function ActivateButton({ onClick }): JSX.Element {
     [],
   );
 
+  const [activePanelOpen, setActivePanelOpen] = useState(false);
+  if (activePanelOpen) {
+    return null;
+  }
```
JanJakes: Removing the button seems a bit hacky to me. Now, there's a bug that if you hit "Cancel", you have no activation button. But I wouldn't do this, it feels like it's just for the tests. Does WP posts page do that?

(The accessibility part should probably work by having a correct focus and stuff, the element below not really being clickable due to being covered, or we can somehow improve it later, when we focus on a11y.)

websupporter: Thank you for catching that! Forgot about the `Cancel`-clickpath. I agree and changed the code. It still does set the autofocus, as it is still good to have, but in the acceptance test I use now simply the selector:

https://github.com/mailpoet/mailpoet/pull/4434/commits/c6d635c3070f9dc750d8c3d646349be1ef5b3325#diff-68667e9409836537b4425bd4f07c6aa5f5a7f5d26e4955eb233dbe196f5343f0R66-R71

JanJakes (approved):

PR #4435, opened 2022-10-18
Activate workflow panel [MAILPOET-4462]

websupporter:
## Description
Adds the `<ActivatePanel />`

## Linked tickets
[MAILPOET-4462]
## Before-merge notes
We should wait for #4434 to be merged, rebase and fix the acceptance test.

[MAILPOET-4462]: https://mailpoet.atlassian.net/browse/MAILPOET-4462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-automation-editor/_panel.scss
+++ b/mailpoet/assets/css/src/components-automation-editor/_panel.scss
@@ -100,6 +100,11 @@
 .mailpoet-automation-activate-panel__body {
   padding-left: 16px;
   padding-right: 16px;
+
+  .components-spinner {
+    display: block;
+    margin: 100px auto 0;
+  }
 }
 
 .mailpoet-automation-activate-panel__section {
--- a/mailpoet/assets/css/src/components-automation-editor/_panel.scss
+++ b/mailpoet/assets/css/src/components-automation-editor/_panel.scss
@@ -57,3 +57,73 @@
   color: #757575;
   font-style: italic;
 }
+
+.mailpoet-automation-activate-panel {
+  animation: mailpoet-automation-activate-panel-animation .1s forwards;
+  background: #fff;
+  border-left: 1px solid #ddd;
+  bottom: 0;
+  height: 100%;
+  left: auto;
+  overflow: auto;
+  position: fixed;
+  right: 0;
+  top: 0;
+  transform: translateX(100%);
+  width: 281px;
+  z-index: 999999;
+
+  button {
+    justify-content: center;
+    width: 100%;
+  }
+}
+
+.mailpoet-automation-activate-panel__header {
+  align-content: space-between;
+  align-items: center;
+  display: flex;
+  height: 61px;
+
+  .has-icon {
+    margin-left: auto;
+    width: auto;
+  }
+}
+
+.mailpoet-automation-activate-panel__header,
+.mailpoet-automation-activate-panel__section {
+  border-bottom: 1px solid #ddd;
+}
+
+.mailpoet-automation-activate-panel__header,
+.mailpoet-automation-activate-panel__body {
+  padding-left: 16px;
+  padding-right: 16px;
+}
+
+.mailpoet-automation-activate-panel__section {
+  margin-left: -16px;
+  margin-right: -16px;
+  padding: 16px;
+}
+
+.mailpoet-automation-activate-panel__header-activate-button,
+.mailpoet-automation-activate-panel__header-cancel-button {
+  flex-grow: 1;
+  max-width: 160px;
+}
+
+.mailpoet-automation-activate-panel__header-activate-button {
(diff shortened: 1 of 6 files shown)


mailpoet/assets/js/src/automation/editor/components/panel/activate-panel.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <Button
+          icon={closeSmall}
+          onClick={onClose}
+          label={__('Close', 'mailpoet')}
+        />
+      </div>
+
+      <div className="mailpoet-automation-activate-panel__body">
+        <div className="mailpoet-automation-activate-panel__section">
+          {__('"%s" is now live.', 'mailpoet').replace('%s', workflow.name)}
```
websupporter: Use `sprintf` from `wpi18n`

mailpoet/assets/js/src/automation/editor/components/panel/activate-panel.tsx
```diff
@@ -0,0 +1,99 @@
...
+            {__('Cancel', 'mailpoet')}
+          </Button>
+        </div>
+      </div>
+
+      <div className="mailpoet-automation-activate-panel__body">
+        <p>
+          <strong>{__('Are you ready to activate?', 'mailpoet')}</strong>
+        </p>
+        <p>{__('Double-check your settings before activating', 'mailpoet')}</p>
```
JanJakes: I think we can add a `.` at the end of the sentence since WP seems to do so for posts:

> Double-check your settings before publishing.

mailpoet/assets/css/src/components-automation-editor/_panel.scss
```diff
@@ -57,3 +57,73 @@
   color: #757575;
   font-style: italic;
 }
+
```
JanJakes, line 60: Are these styles taken from the Gutenberg repo? We could add a permalink above, for instance like [here](https://github.com/mailpoet/mailpoet/blob/3ea730a0b642b6c59875f0a0fec4e858d71315a7/mailpoet/assets/css/src/components-automation-editor/_step-card.scss#L1).

mailpoet/assets/js/src/automation/editor/components/panel/activate-panel.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }),
+    [],
+  );
+
+  const goToListings = () => {
+    window.location.href = MailPoet.urls.automationListing;
+  };
+
+  const { createErrorNotice } = useDispatch(noticesStore as StoreDescriptor);
+  void createErrorNotice(
```
JanJakes: There should be no business logic in the component body (that can re-render any number of times). This is a "side-effect" that causes other rendering so it results in the following error:

Screen Shot 2022-10-19 at 7 24 57

If you want to display the notice on component mount, you can do so with `useEffect` with an empty deps array (well, almost empty — having only the dispatch action):

```ts
useEffect(() => {
  void createErrorNotice(...);
}, [createErrorNotice]);
```

(Alternatively, it could be triggered [from within the activate action](https://github.com/mailpoet/mailpoet/blob/3ea730a0b642b6c59875f0a0fec4e858d71315a7/mailpoet/assets/js/src/automation/editor/store/actions.ts#L81) right after the `yield`).

JanJakes (changes requested):
Looks great!

A small idea — can we ("reasonably easily") replicate this loading state that WP posts have?

Screen Shot 2022-10-19 at 7 25 21

I guess it could just change the state of the component "onclick" and then `PostStep` would replace it.

mailpoet/assets/js/src/automation/editor/components/panel/activate-panel.tsx
websupporter: I like the idea to update it in the `active` method itself, so we have a "universal" behavior in case we activate somewhere else in the future as well. Addressed in f5989e412

websupporter:
> can we ("reasonably easily") replicate this loading state that WP posts have

That's a good idea. Addressed in 25ced434f

I differ from the post activating state in two ways:
* The cancel button is disabled, because we can't cancel this action at this point.
* The "Activating..." button `isBusy`, which replicates the `Update` button in the post editor and I wonder why the `Publishing...` button is not busy.

mailpoet/assets/css/src/components-automation-editor/_panel.scss
websupporter: The naming is a bit different and sometimes the selectors/selector-cascading. I wonder, should I align it more?

JanJakes (approved):

PR #4436, opened 2022-10-18
Update the confirmation page title - MAILPOET-4667

triple0t:
## Description

Update the confirmation page title

Show website title instead of list names

## Linked tickets

[MAILPOET-4667](https://mailpoet.atlassian.net/browse/MAILPOET-4667)

### DIFF
--- a/mailpoet/lib/Subscription/Pages.php
+++ b/mailpoet/lib/Subscription/Pages.php
@@ -354,28 +354,19 @@ class Pages {
     return $meta;
   }
 
-  private function getConfirmTitle() {
-    if ($this->isPreview()) {
-      $title = sprintf(
-        // translators: %s is a comma-separated list of segment names.
-        __("You have subscribed to: %s", 'mailpoet'),
-        'demo 1, demo 2'
-      );
-    } else {
-      $segmentNames = array_map(function($segment) {
-        return $segment->getName();
-      }, $this->subscriber->getSegments()->toArray());
+  private function getConfirmTitle(): string {
+    $wpSiteTitle = $this->wp->getBloginfo('name');
 
-      if (empty($segmentNames)) {
-        $title = __("You are now subscribed!", 'mailpoet');
-      } else {
-        $title = sprintf(
-          // translators: %s is a comma-separated list of segment names.
-          __("You have subscribed to: %s", 'mailpoet'),
-          join(', ', $segmentNames)
-        );
-      }
+    if (empty($wpSiteTitle)) {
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Subscription/Pages.php
```diff
@@ -354,28 +354,19 @@ public function setWindowTitleParts($meta = []) {
     return $meta;
   }
 
-  private function getConfirmTitle() {
-    if ($this->isPreview()) {
+  private function getConfirmTitle(): string {
+    $wpSiteTitle = $this->wp->getBloginfo('name');
```
samnajian, line 358: What do you think about using [get_blog_details](https://developer.wordpress.org/reference/functions/get_blog_details/) instead so that it's multisite compatible?

mailpoet/tests/acceptance/Settings/ConfirmConfirmationPageCest.php
```diff
@@ -12,7 +12,7 @@ public function confirmDefaultConfirmationPage(\AcceptanceTester $i) {
     $i->waitForText('MailPoet Page');
     $i->click('[data-automation-id="preview_page_link"]');
     $i->switchToNextTab();
-    $i->waitForText('You have subscribed to: demo 1, demo 2');
+    $i->see('You have subscribed to');
```
samnajian: Do you think it'd make sense to check for the site name in this assertion as well?

mailpoet/lib/Subscription/Pages.php
triple0t: I think I will leave it as it is.
I tested it, and this works with multisite as well. I was successfully able to fetch the Site title for each sub-site.
You may give it a try with this once code: 1062127.4:10894c879e10f9ddf659c21c45c318f5a5f858d481a7b4bd44e732b4b1f4924c (Check for newsletters, there’s a saved draft in each sub-site)

Also, [get_blog_details](https://developer.wordpress.org/reference/functions/get_blog_details/) appears to be [depreciated](https://make.wordpress.org/core/2016/11/04/multisite-focused-changes-in-4-7/)

mailpoet/tests/acceptance/Settings/ConfirmConfirmationPageCest.php
triple0t: Not necessarily. I think this is okay as well.
If we can see the string on the page, the test case should be allowed to pass

samnajian (approved):
LGTM thanks!

Aschepikov (approved):
Verified, GJ!

PR #4437, opened 2022-10-18
Workflow listing actions — part 2 [MAILPOET-4540]

JanJakes:
## Description

Based on the original PR by @johnolek, this:
1. Adds a `@wordpress/data`-based store for automation listing.
2. Uses `@wordpress/notices` store to render MailPoet notices (because those have WP listing styles).
3. Uses dropdown menu from `@wordpress/components` for listing actions.
4. Implements duplicate, trash, restore, and delete actions.
5. Implements notices and "undo" buttons.
6. Improves listing rendering and aligns it closer to specs.
7. Improves workflow stats rendering and aligns it closer to specs. Reuses a single component across editor/listing.

The commits also tell the story.

## Code review notes

_N/A_

## QA notes

## Linked PRs

Implements UI for https://github.com/mailpoet/mailpoet/pull/4376.

## Linked tickets

[MAILPOET-4540]

## After-merge notes

_N/A_

[MAILPOET-4540]: https://mailpoet.atlassian.net/browse/MAILPOET-4540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/listing/store/selectors.ts
```diff
@@ -1,6 +1,11 @@
 import { State } from './types';
 import { Workflow } from '../workflow';
+import { workflowCount } from '../../config';
 
 export function getWorkflows(state: State): Workflow[] {
   return state.workflows;
 }
+
+export function getWorkflowCount(state: State): number {
+  return state.workflows ? state.workflows.length : workflowCount;
```
websupporter, line 10: Curious: Why do we need to "import" the workflowCount. To me it looks like `return state.workflows ? state.workflows.length : 0;` would do the same.

mailpoet/assets/js/src/automation/listing/index.tsx
```diff
@@ -143,21 +147,15 @@ export function AutomationListing({ workflows, loading }: Props): JSX.Element {
               allowFreeTextSearch
```
websupporter, line 150: Not sure if it's related to this PR, but our search does currently not work. This happens, when I type something in the search box: 
image

mailpoet/assets/js/src/automation/listing/store/selectors.ts
websupporter: Ah: I think this prevents the `<Onboarding />` to show up while we just load the workflows.

mailpoet/assets/js/src/automation/listing/index.tsx
JanJakes: Yeah, we never implemented it. I'm thinking maybe we should hide it for the MVP 🤔

mailpoet/assets/js/src/automation/listing/store/selectors.ts
JanJakes: Yeah, the imported count tells you if you will have any workflows at all and you can decide to show either onboarding or a loader for the listing.

I was thinking to preload even the workflow list in the template but then I started to think about tabs and paging and stuff and just left the count as John did. In his version the count was a separate field in the state but it then creates too many states of "count" vs. "workflows" so I just used it in the selector when workflows are not loaded yet.

mailpoet/assets/css/src/components-automation/_statistics.scss
```diff
@@ -58,6 +59,10 @@
       font-size: 12px;
       font-weight: 400;
       line-height: 16px;
+
+      &.display-after {
+        order: 1;
+      }
```
websupporter: :clap: 

this is how to solve the problem from the number on one screen is on top, at the other on bottom. Didn't think about a CSS solution when working on it. Thank you!

websupporter (approved):
Thank you for the PR. Very nice changes. I love especially the fixes for the statistics component :heart:

PR #4443, opened 2022-10-20
Add new Subscriber hooks [MAILPOET-4727]

lysyjan:
## Description

This PR adds new hooks for better integration with other plugins. Here is a list:
- `mailpoet_subscriber_created(int $subscriberId)`
- `mailpoet_subscriber_updated(int $subscriberId)`
- `mailpoet_subscriber_deleted(int $subscriberId)`
- `mailpoet_multiple_subscribers_created(int $minCreatedAtTimestamp)`
- `mailpoet_multiple_subscribers_updated(int $minUpdatedAtTimestamp)`
- `mailpoet_multiple_subscribers_deleted(array $subscriberIds)`

## Code review notes

A new layer was created that contains stored changes from a repository and a listener, which are fired in the WP `shutdown` hook. The new layer should prevent multiple calls of update hooks. 

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4727]

## After-merge notes

_N/A_

[MAILPOET-4727]: https://mailpoet.atlassian.net/browse/MAILPOET-4727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/SubscriberChangesNotifier.php
+++ b/mailpoet/lib/Config/SubscriberChangesNotifier.php
@@ -68,4 +68,22 @@ class SubscriberChangesNotifier {
     // store id as a key and timestamp change as the value
     $this->deletedSubscriberIds[$subscriberId] = Carbon::createFromTimestamp($this->wp->currentTime('timestamp'), 'UTC');
   }
+
+  public function subscribersCreated(array $subscriberIds): void {
+    foreach ($subscriberIds as $subscriberId) {
+      $this->subscriberCreated($subscriberId);
+    }
+  }
+
+  public function subscribersUpdated(array $subscriberIds): void {
+    foreach ($subscriberIds as $subscriberId) {
+      $this->subscriberUpdated($subscriberId);
+    }
+  }
+
+  public function subscribersDeleted(array $subscriberIds): void {
+    foreach ($subscriberIds as $subscriberId) {
+      $this->subscriberDeleted($subscriberId);
+    }
+  }
 }
--- a/mailpoet/lib/Config/SubscriberChangesNotifier.php
+++ b/mailpoet/lib/Config/SubscriberChangesNotifier.php
@@ -8,13 +8,13 @@ use MailPoetVendor\Carbon\Carbon;
 
 class SubscriberChangesNotifier {
 
-  /** @var array<int, Carbon> */
+  /** @var array<int, int> */
   private $createdSubscriberIds = [];
 
-  /** @var array<int, Carbon> */
+  /** @var array<int, int> */
   private $deletedSubscriberIds = [];
 
-  /** @var array<int, Carbon> */
+  /** @var array<int, int> */
   private $updatedSubscriberIds = [];
 
   /** @var WPFunctions */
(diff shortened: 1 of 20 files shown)


mailpoet/lib/Config/SubscriberChangesNotifier.php
```diff
@@ -0,0 +1,113 @@
...
+    $this->notifyDeletes();
+  }
+
+  private function notifyCreations(): void {
+    if (count($this->createdSubscriberIds) === 1) {
+      foreach ($this->createdSubscriberIds as $subscriberId => $updatedAt) {
+        $this->wp->doAction(SubscriberEntity::HOOK_SUBSCRIBER_CREATED, $subscriberId);
+      }
+    } elseif ($this->createdSubscriberIds) {
+      $minTimestamp = min($this->createdSubscriberIds);
```
cleacos, line 41: This array could be empty, and `min` requires at least one element. E_WARNING.

mailpoet/lib/Config/SubscriberChangesNotifier.php
```diff
@@ -0,0 +1,113 @@
...
+      unset($this->updatedSubscriberIds[$subscriberId]);
+    }
+
+    if (count($this->updatedSubscriberIds) === 1) {
+      foreach ($this->updatedSubscriberIds as $subscriberId => $updatedAt) {
+        $this->wp->doAction(SubscriberEntity::HOOK_SUBSCRIBER_UPDATED, $subscriberId);
+      }
+    } elseif ($this->updatedSubscriberIds) {
+      $minTimestamp = min($this->updatedSubscriberIds);
+      if ($minTimestamp) {
```
cleacos, line 60: Same here for a empty array

mailpoet/lib/Config/SubscriberChangesNotifier.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  private function notifyDeletes(): void {
+    if (count($this->deletedSubscriberIds) === 1) {
+      foreach ($this->deletedSubscriberIds as $subscriberId => $updatedAt) {
+        $this->wp->doAction(SubscriberEntity::HOOK_SUBSCRIBER_DELETED, $subscriberId);
+      }
+    } elseif ($this->deletedSubscriberIds) {
+      $this->wp->doAction(SubscriberEntity::HOOK_MULTIPLE_SUBSCRIBERS_DELETED, array_keys($this->deletedSubscriberIds));
```
cleacos, line 72: if there are no deleted subscribers... we will get the hook `HOOK_MULTIPLE_SUBSCRIBERS_DELETED` after calling to the `notify()` method, won't we?

cleacos:
I took a look at the code, and it looks good. Woody will test it in our PR. Thank you.

lysyjan:
@costasovo Feel free to reassign this PR if you are too busy.

mailpoet/lib/Config/SubscriberChangesNotifier.php
```diff
@@ -32,39 +33,39 @@ public function notify() {
...
+    foreach ($this->deletedSubscriberIds as $subscriberId => $deletedAt) {
+      $this->wp->doAction(SubscriberEntity::HOOK_SUBSCRIBER_DELETED, $subscriberId, $deletedAt->getTimestamp());
     }
   }
 
   public function subscriberCreated(int $subscriberId): void {
-    // to avoid duplicities we use id as a key
-    $this->createdSubscriberIds[$subscriberId] = $subscriberId;
+    // store id as a key and timestamp change as the value
+    $this->createdSubscriberIds[$subscriberId] = Carbon::createFromTimestamp($this->wp->currentTime('timestamp'), 'UTC');
```
costasovo: I think that the resulting timestamp might not be UTC here.
WP function `current_time` takes a second argument `$gmt` which is false by default.
If the $gmt is false it adds the offset from WP option.  [See the source](https://developer.wordpress.org/reference/functions/current_time/#source)

costasovo (changes requested):
Thanks for implementing those hooks! I like that the logic for notifying is in the notification service.

I think I've found a potential issue with a timestamp of a change. Please see my comment.

I see that the PR covers changes done for individual Subscriber entities via the event listener and also bulk changes done via the Subscribers Repository. But I think we have more places where we update Subscribers places that are not covered by this PR.

- [WordPress users sync](https://github.com/mailpoet/mailpoet/blob/24cca1ba04a9ab78250e62f7ec62aff78ba217b1/mailpoet/lib/Segments/WP.php#L211-L225)
- [Subscribers Import](https://github.com/mailpoet/mailpoet/blob/daa5575d876cc1087bae806f5b300992dd875e94/mailpoet/lib/Subscribers/ImportExport/Import/Import.php#L490-L546)
- [WooCommerce Customers batch sync](https://github.com/mailpoet/mailpoet/blob/1010b64c05dba38f62cad0933ec32e7e2a83836f/mailpoet/lib/Segments/WooCommerce.php#L173-L193)

There might be some more. Do we plan to cover those later?

lysyjan:
@costasovo I added new commits that should cover changes in import and synchronization from WP and WC.

costasovo (approved):
LGTM! 👍

veljkho (approved):

PR #4447, opened 2022-10-21
Validation UI [MAILPOET-4700]

websupporter:
## Notes
The error messages from the WP Rest API are not usable for our us. In this PR we map the error codes now against a set of strings to produce more user friendly error messages ([7b956c1](https://github.com/mailpoet/mailpoet/pull/4447/commits/7b956c1450b9d7aa37c5ebdb1a5a405043895916))

We loose quite a bit of information by this change, but I think given translations of those messages, there is no good reliable way to extract this information from the message strings themselves. Furthermore, I am not sure, we need this information at all.

## Linked tickets
[MAILPOET-4700]

[MAILPOET-4700]: https://mailpoet.atlassian.net/browse/MAILPOET-4700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Validator/ValidationException.php
```diff
@@ -10,8 +10,10 @@ class ValidationException extends UnexpectedValueException {
   protected $wpError;
 
   public static function createFromWpError(WP_Error $wpError): self {
-    $exception = self::create()
-      ->withMessage($wpError->get_error_message());
```
JanJakes: Should we maybe keep both the message and errors?

mailpoet/lib/Validator/Validator.php
```diff
@@ -25,11 +25,28 @@ public function __construct(
...
+    if (!isset($schema['properties'])) {
+      $result = $this->validateAndSanitizeValueFromSchema($value, $schema, $paramName);
+      if ($result instanceof WP_Error) {
+        throw ValidationException::createFromWpError($result);
+      }
+      return $result;
+    }
+
+    $combinedError = new WP_Error();
+    foreach ($schema['properties'] as $propertyKey => $definition) {
```
JanJakes: I think this should actually go outside of `mailpoet/lib/Validator/Validator.php` somewhere in the automation code, only reusing the validator from outside.

The reason here is that objects can have some validation rules on the whole object itself, but here we're breaking it  (what I mean is things like `minProperties`, etc.). Or... what do you think?

mailpoet/lib/Validator/Validator.php
```diff
@@ -25,11 +25,28 @@ public function __construct(
...
+      $result = $this->validateAndSanitizeValueFromSchema($value, array_merge($schema, ['properties' => [$propertyKey => $definition]]), $paramName);
+      if ($result instanceof WP_Error) {
+        $combinedError->add($propertyKey, $result->get_error_message(), $result->get_error_data());
+      }
+    }
```
JanJakes: I think we should also handle `additionalProperties` and `patternProperties`. Or don't they require special treatment? With regards to my comment above (that we could reuse the `mailpoet/lib/Validator/Validator.php` from outside), it may not be needed to solve now.

Maybe we can put this logic back to Validator itself once we're able to make sure all per-object rules work, that we validate all types of properties correctly, etc.? And leave it outside for now?

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/ValidStepRule.php
```diff
@@ -45,9 +46,19 @@ public function visitNode(Workflow $workflow, WorkflowNode $node): void {
       try {
         $rule->visitNode($workflow, $node);
       } catch (UnexpectedValueException $e) {
-        $this->errors[$stepId] = ['step_id' => $stepId, 'message' => $e->getMessage()];
+        if (!isset($this->errors[$stepId])) {
+          $this->errors[$stepId] = ['step_id' => $stepId, 'message' => $e->getMessage(), 'fields' => []];
+        }
```
JanJakes, line 51: This IF is to save the first error per field? Just asking if I understand.

mailpoet/assets/img/icons/warning.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="-2 -2 24 24">
```
JanJakes: You mentioned this is included in `@wordpress/icons`. In that case, we don't need the SVG, do we? Or did you want to use the image from Figma?

JanJakes: Aaaah, that's due to the usage in CSS? In that case, if we need an SVG anyway, I guess we could use the icon from Figma, maybe.

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/ValidStepRule.php
```diff
@@ -63,6 +69,61 @@ public function visitNode(Workflow $workflow, WorkflowNode $node): void {
...
+          case "rest_invalid_email":
+            return __('This value is not an email address.', 'mailpoet');
+          case "rest_invalid_ip":
+            return __('This value is not an IP address.', 'mailpoet');
+          case "rest_invalid_uuid":
+            return __('This value is not an UUID.', 'mailpoet');
+          default:
+            return $errorCode;
+        }
+      },
```
JanJakes, line 118: I was thinking... can we improve the existing messages just by replacing some substrings? But then I realized nope, because it would not be translatable. But... this doesn't seem to cover things like `value must be at most 3 characters long.`. It also sets some messages to empty strings.

The validator is a carefully designed WP schema wrapper, and we must be very cautious not to bend it to something else just because of a specific feature. It relates to my previous comments about whether we should use it from the outside for now, maybe it would be safer.

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/AtLeastOneTriggerRule.php
```diff
@@ -0,0 +1,28 @@
...
+
+class AtLeastOneTriggerRule implements WorkflowNodeVisitor {
+  public const RULE_ID = 'at-least-one-trigger';
+
+  public function initialize(Workflow $workflow): void {
+    foreach ($workflow->getSteps() as $step) {
+      if ($step->getType() === 'trigger') {
+        return;
+      }
+    }
```
JanJakes: You could also avoid this iteration by moving the check to `visitNode` and throwing the exception in `complete`. That said, I know in reality the performance difference will likely be unmeasurable...

JanJakes (commented):
This looks great! I have just a few questions/observations.

mailpoet/lib/Validator/ValidationException.php
websupporter: You mean something like `self::create()->withMessage()->withWpError($wpError)`?

Currently I am not sure if we need it and we can always add it later on I guess.

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/ValidStepRule.php
websupporter: Yes.

mailpoet/assets/img/icons/warning.svg
websupporter: Yes. For me the question is more, do we align with `@wordpress/icons` or Figma at that point from I guess a product perspective. @NeosinneR related Slack was here: https://a8c.slack.com/archives/C0313447YBC/p1666332588178519

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/AtLeastOneTriggerRule.php
websupporter: Addressed in a0ed13bd3

mailpoet/assets/img/icons/warning.svg
NeosinneR: @websupporter I like @JanJakes suggestion that if we need to add the SVG into codebase anyway, let's stick to Figma. But if we can use the existing `@wordpress/icons`, let's choose that.

websupporter: Addressed in [e5b6aec](https://github.com/mailpoet/mailpoet/pull/4447/commits/e5b6aec4beca57082c8102a0e87434beca44e18e)

websupporter:
The Validator is now behaving like before. I just added a `public function validateArray` which can validate the array from `$schema->toArray()` directly. The ValidStepArgsRule validates now each property, catches the `ValidationException` with the `WP_Error` and retrieves the information.

> It also sets some messages to empty strings

I did not find a good string to map `rest_additional_properties_forbidden`, `rest_too_few_properties` and `rest_too_many_properties`, this is why they are empty. To me at least, those seem to be errors, a user can not really produce through the UI.

I also think, sometimes, it is better the JS implementation decides the best message for the user. An example is the error when no email_id was assigned yet. Before ec69810:
image

After ec69810:
image

mailpoet/lib/Validator/Validator.php
```diff
@@ -25,7 +25,18 @@ public function __construct(
...
+  }
+
+  /**
+   * Strict validation & sanitization implementation.
+   * It only coerces int to float (e.g. 5 to 5.0).
+   *
+   * @param mixed $value
+   * @return mixed
+   */
+  public function validateArray(array $schema, $value, string $paramName = 'value') {
```
JanJakes: 👍, just thinking about the naming here... Maybe this could be confused with "validating an array" rather than "validating anything but having the schema specified as array". But not sure how to name it better.

Another thing is that the `Validator` wrapper has some constraints on the JSON schema in the array, and those constraints are enforced with the typed `Schema`. So this method is sort of "unsafe" in terms of anything that can be passed into it. So maybe naming like `rawValidate` or similar would catch this better?

(On the other hand, we can remove this method, if we fully implement the new validation logic inside the validator one day.)

JanJakes (commented):
@websupporter LGTM! Added a comment but not really necessary to address this now so let me know if we should proceed merging or you want to do some changes.

mailpoet/lib/Validator/Validator.php
websupporter: Addressed in[9a11683](https://github.com/mailpoet/mailpoet/pull/4447/commits/9a11683d416d6b448902255f0abfc0beca0ffe06)

JanJakes (approved):

PR #4448, opened 2022-10-21
Add beta badge [MAILPOET-4728]

websupporter:
## Code review notes
Didnt think too much about "code quality", as this is a temporary code snippet, we will remove soon.

## Linked tickets

[MAILPOET-4728]

[MAILPOET-4728]: https://mailpoet.atlassian.net/browse/MAILPOET-4728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Menu.php
+++ b/mailpoet/lib/Config/Menu.php
@@ -92,6 +92,23 @@ class Menu {
 
     $this->registerMailPoetMenu();
 
+    // @ToDo Remove Beta once Automation is no longer beta.
+    $this->wp->addAction('admin_head', function () {
+      echo '<style>
+#adminmenu .toplevel_page_mailpoet-newsletters a[href="admin.php?page=mailpoet-automation"] {
+  display: flex;
+  gap: 8px;
+}
+.mailpoet-beta-badge {
+  background: #F0F0F1;
+  border-radius: 50px;
+  padding: 0 12px;
+  color: #1D2327;
+  font-weight: normal;
+}
+</style>';
+    });
+
     if (!self::isOnMailPoetAdminPage()) {
       return;
     }
@@ -413,7 +430,8 @@ class Menu {
       $this->wp->addSubmenuPage(
         self::MAIN_PAGE_SLUG,
         $this->setPageTitle('Automation'),
-        'Automation',
+        // @ToDo Remove Beta once Automation is no longer beta.
+        '<span>Automation</span><span class="mailpoet-beta-badge">Beta</a>',
         AccessControl::PERMISSION_MANAGE_EMAILS,
         'mailpoet-automation',
         [$this, 'automation']
--- a/mailpoet/lib/Config/Menu.php
+++ b/mailpoet/lib/Config/Menu.php
@@ -436,9 +436,9 @@ class Menu {
     }
     $this->wp->addSubmenuPage(
       self::MAIN_PAGE_SLUG,
-      $this->setPageTitle(__('Automation', 'mailpoet')),
+      $this->setPageTitle(__('Automations', 'mailpoet')),
       // @ToDo Remove Beta once Automation is no longer beta.
-      '<span>' . esc_html__('Automation', 'mailpoet') . '</span><span class="mailpoet-beta-badge">Beta</a>',
+      '<span>' . esc_html__('Automations', 'mailpoet') . '</span><span class="mailpoet-beta-badge">Beta</a>',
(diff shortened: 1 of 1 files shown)


mailpoet/lib/Config/Menu.php
```diff
@@ -413,7 +430,8 @@ private function registerMailPoetMenu() {
       $this->wp->addSubmenuPage(
         self::MAIN_PAGE_SLUG,
         $this->setPageTitle('Automation'),
-        'Automation',
+        // @ToDo Remove Beta once Automation is no longer beta.
+        '<span>Automation</span><span class="mailpoet-beta-badge">Beta</a>',
```
JanJakes: Could you pls make the `Automation` string translatable, @websupporter? I won't do this in my branch, so we don't create conflicts.

Additionally, we can ask @NeosinneR if the menu item should be `Automation` or `Automations` and where in the menu (which order) it should lie.

NeosinneR: Let's go with Automations, as all other menu items are written in plural. And positioned below Emails. 

As for translations - are other items translated? If so, I'd also translate this one.

websupporter: Menu items: Some where translateable, others not. I made them all translateable in [fc80f73](https://github.com/mailpoet/mailpoet/pull/4448/commits/fc80f73ea27c55f189b9b0374169d50e43b56e6b)

The Automations menu item has now been moved [db8fc2d](https://github.com/mailpoet/mailpoet/pull/4448/commits/db8fc2d9b55fd036ecfef5ad5784eb707afe0154) and renamed [d4259fa](https://github.com/mailpoet/mailpoet/pull/4448/commits/d4259fa5d9b364e12a7ac2e9595477ea457c6f19)

mailpoet/lib/Config/Menu.php
```diff
@@ -401,51 +420,56 @@ private function registerMailPoetMenu() {
...
+  private function registerAutomationMenu() {
+
+    if (!$this->featuresController->isSupported(FeaturesController::AUTOMATION)) {
+      return;
     }
+    $this->wp->addSubmenuPage(
+      self::MAIN_PAGE_SLUG,
+      $this->setPageTitle(__('Automations', 'mailpoet')),
+      // @ToDo Remove Beta once Automation is no longer beta.
+      '<span>' . esc_html__('Automations', 'mailpoet') . '</span><span class="mailpoet-beta-badge">Beta</a>',
```
johnolek: Is there something I'm missing or should this be a closing `</span>` instead of `</a>`? Not sure if it matters -- it renders fine in Firefox anyway, but I'm not sure if that's because it's automatically closing the tag or if WordPress is fixing it or what, and I don't know if all browsers will behave the same way.

mailpoet/lib/Config/Menu.php
```diff
@@ -401,51 +420,56 @@ private function registerMailPoetMenu() {
...
+      $this->setPageTitle(__('Automation Templates', 'mailpoet')),
+      esc_html__('Automation Templates', 'mailpoet'),
+      AccessControl::PERMISSION_MANAGE_AUTOMATIONS,
+      'mailpoet-automation-templates',
+      [$this, 'automationTemplates']
+    );
+
+    // add body class for automation editor page
+    $this->wp->addAction('load-' . $automationEditorPage, function() {
+      $this->wp->addAction('admin_body_class', function ($classes) {
```
johnolek: I know this was already here and you just switched to using an early return, but I think this should be `addFilter` instead of `addAction`.

johnolek (changes requested):
There's one small thing with the HTML for the badge that I'd like you to take a look at. As I mention in my comment I don't know if it actually matters.

mailpoet/lib/Config/Menu.php
websupporter: Thank you! That was not meant with "I didn't care too much about code quality" :see_no_evil: Addressed in 39fe683ef

mailpoet/lib/Config/Menu.php
websupporter: Good catch. Addressed in 677c9ee19

websupporter:
Thank you @johnolek. I addressed both issues.

johnolek (approved):
Looks good to me!

PR #4450, opened 2022-10-24
Deactivate workflow part one [MAILPOET-4730]

websupporter:
## Description
This implements the first part of the deactivation workflow. To my understanding MAILPOET-4731 will implement the backend logic (I think we need a different status for `deactivating`) as well as the missing pieces of the frontend for the deactivation workflow

## Code review notes
Because both issues (4730 and 4731) are so close aligned, I left some ToDo comments, where I think the decision about e.g. naming should be made in 4731.

## Linked tickets
[MAILPOET-4730]

[MAILPOET-4730]: https://mailpoet.atlassian.net/browse/MAILPOET-4730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
@@ -1,5 +1,6 @@
+import { useState } from 'react';
 import { Button, NavigableMenu, TextControl } from '@wordpress/components';
-import { useDispatch, useSelect } from '@wordpress/data';
+import { dispatch, useDispatch, useSelect } from '@wordpress/data';
 import { PinnedItems } from '@wordpress/interface';
 import { __ } from '@wordpress/i18n';
 import { DocumentActions } from './document_actions';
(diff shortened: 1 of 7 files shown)


mailpoet/assets/js/src/automation/editor/components/header/index.tsx
```diff
@@ -57,6 +59,43 @@ function SaveDraftButton(): JSX.Element {
...
+    (select) => ({
+      hasUsersInProgress:
+        select(storeName).getWorkflowData().stats.totals.in_progress > 0,
+    }),
+    [],
+  );
+
+  const toggleModal = () => {
+    setShowDeactivateModal(!showDeactivateModal);
+  };
```
JanJakes: I slightly fear using `toggle` logic in these cases, in case something runs for any random reason twice.

In other words, I think the actions you have here are only "open modal" (the button) and close modal (via `onClose`). There's no reason that the `Deactivate` button actually toggles the modal because a click outside of it will close it anyway.

JanJakes (commented):
👏

mailpoet/assets/js/src/automation/editor/components/header/index.tsx
websupporter: Addressed in [beee962](https://github.com/mailpoet/mailpoet/pull/4450/commits/beee9625396475455b25809e287124de51967a15)

websupporter:
Thank you for the Review @JanJakes. Addressed the issue in [beee962](https://github.com/mailpoet/mailpoet/pull/4450/commits/beee9625396475455b25809e287124de51967a15)

JanJakes (approved):

PR #4455, opened 2022-10-25
Add Email Preview link [MAILPOET-4631]

websupporter:
[MAILPOET-4631]

[MAILPOET-4631]: https://mailpoet.atlassian.net/browse/MAILPOET-4631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/edit_newsletter.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/edit_newsletter.tsx
@@ -105,10 +105,12 @@ export function EditNewsletter(): JSX.Element {
         variant="secondary"
         centered
         isBusy={fetchingPreviewLink}
+        disabled={fetchingPreviewLink}
         onClick={async () => {
           setFetchingPreviewLink(true);
           const link = await retrievePreviewLink(emailId);
           window.open(link as string, '_blank');
+          setFetchingPreviewLink(false);
         }}
       >
         Preview
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/edit_newsletter.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/edit_newsletter.tsx
@@ -5,9 +5,31 @@ import { Button } from '../../../components/button';
 import { storeName } from '../../../../../editor/store';
 import { MailPoet } from '../../../../../../mailpoet';
 
+const emailPreviewLinkCache = {};
+const retrievePreviewLink = async (emailId, callback) => {
+  if (
+    emailPreviewLinkCache[emailId] &&
+    emailPreviewLinkCache[emailId].length > 0
+  ) {
+    callback(emailPreviewLinkCache[emailId]);
+    return;
+  }
+  const response = await MailPoet.Ajax.post({
+    api_version: window.mailpoet_api_version,
+    endpoint: 'newsletters',
+    action: 'get',
+    data: {
+      id: emailId,
+    },
+  });
+  emailPreviewLinkCache[emailId] = response?.meta?.preview_url ?? '';
+  callback(emailPreviewLinkCache[emailId]);
+};
+
 export function EditNewsletter(): JSX.Element {
   const [redirectToTemplateSelection, setRedirectToTemplateSelection] =
     useState(false);
+  const [previewLink, setPreviewLink] = useState('');
 
   const { selectedStep, workflowId, workflowSaved } = useSelect(
     (select) => ({
@@ -20,6 +42,10 @@ export function EditNewsletter(): JSX.Element {
 
   const emailId = selectedStep?.args?.email_id as number | undefined;
   const workflowStepId = selectedStep.id;
+
+  if (emailId && !previewLink) {
+    void retrievePreviewLink(emailId, setPreviewLink);
+  }
 
   const createEmail = useCallback(async () => {
     setRedirectToTemplateSelection(true);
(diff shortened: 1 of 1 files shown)


mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/edit_newsletter.tsx
```diff
@@ -5,9 +5,31 @@ import { Button } from '../../../components/button';
 import { storeName } from '../../../../../editor/store';
 import { MailPoet } from '../../../../../../mailpoet';
 
+const emailPreviewLinkCache = {};
```
johnolek, line 8: Using a local memory cache like this feels a little bit off to me in a react/redux context even though I think it works perfectly fine as far as I can tell. I know you don't want to make duplicate API calls every time an email step is opened for the n+1th time though. 

Might it make sense to store the retrieved preview URLs in redux instead to persist the data between mounts?

I'm not sure it fully makes sense to me to to fire off an API call every single time an email step is opened for the first time just to get the preview URL. It's not exactly an expensive operation or anything, and it doesn't block anything else, but what percentage of the time will users actually want to use that link?

If someone is switching quickly between tabs you could also end up in a situation where the email step component unmounts before the ajax request finishes, and then the `setPreviewLink` callback won't work and might emit a warning, though probably not with react in production mode.

I might consider having the preview button work more similarly to the create button, where clicking the button fires off the ajax request to get the preview URL (or retrieves it from the cache/redux/wherever) and then opens a new tab, just to prevent unnecessary API calls.

Anyway, this does work fine for now, just options to consider. @JanJakes please let us know if you can think of any actual issues we might encounter from using a cache like this, since you're much more familiar with the implications of this kind of stuff from a react perspective.

johnolek (dismissed):
This looks good to me, but I left a comment with some ideas for alternative approaches we could consider, though I'm not sure there would be much benefit to either of them at the moment. Up to you if you would like to make any modifications!

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/edit_newsletter.tsx
websupporter: I do not like the solution as well. For the feature itself, I thought creating its own store or extending an existing one is a bit too much though (at least for now). But open to suggestions :)

JanJakes: @johnolek @websupporter Usually, using a module global to store data for React components is risky because when you have multiple instances of the same component, they would override each others data. But in this case, it may actually work to it's advantage — multiple instances of the same component could read from the same cache.

But there's another issue — the `retrievePreviewLink` is called in the component body, not wrapped in an `useEffect` which will very likely produce JS errors in some cases:
```ts
  if (emailId && !previewLink) {
    void retrievePreviewLink(emailId, setPreviewLink);
  }
```

Plus, even if the above is fixed, multiple usages of the same component are likely to trigger the request in parallel.

Other ideas:
1. Fetch preview URL on click, then use it, as @johnolek said.
2. Add a new non-public URL for "email preview by ID" that would redirect to the preview URL.
3. Replicate the URL generation logic in JS (we may forget to sync it with PHP, if ever changed).
4. Save `preview_url` in args (but this can break old emails if we ever changed the logic somehow).

What do you think? I'm not sure what I prefer, option 2. seems kind of reasonable, I guess.

websupporter: Thanks for your feedback. I opted for the first option in [857ab8f](https://github.com/mailpoet/mailpoet/pull/4455/commits/857ab8fbf080889fa03115a4e80f21de0784124e)

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/edit_newsletter.tsx
```diff
@@ -80,7 +101,16 @@ export function EditNewsletter(): JSX.Element {
       >
         Edit content
       </Button>
-      <Button variant="secondary" centered>
+      <Button
+        variant="secondary"
+        centered
+        isBusy={fetchingPreviewLink}
+        onClick={async () => {
+          setFetchingPreviewLink(true);
```
johnolek, line 110: You never reset this to `false`, so the button remains in a busy state until the email step component is unmounted, i.e. when you return to the editor from the preview tab the button still looks busy.

You might also want to add a `disabled` attribute to prevent duplicate API calls, though it doesn't change the behavior so it's not that big of a deal.

I noticed something interesting, which is that the local state seems to persist between email steps if you switch immediately from one to another. Maybe React isn't fully unmounting the component in this case, instead just changing the props? Here's a quick screen recording in case I'm not making sense.

https://user-images.githubusercontent.com/8656640/198082552-69080264-053f-4235-b4ab-434238bfab25.mov

It's not unique to the changes from this PR, but it is more noticeable because `fetchingPreviewLink` stays `true`. Another example is if you click on a "Design Email" button for one email step and then immediately switch to a different email step (before the API call completes), the "Design Email" button appears busy for that step too (which is actually probably a good thing because it doesn't make sense to trigger two email creations at the same time).

Anyway, tldr I think this would be good to go if you just change `fetchingPreviewLink` back to `false` once it's done.

johnolek (changes requested):

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/edit/edit_newsletter.tsx
websupporter: Good catch! Thanks! Addressed in [c7a548d](https://github.com/mailpoet/mailpoet/pull/4455/commits/c7a548de92e9449f91d475250eb7219e82f3b87f)

johnolek (approved):
Looks good to me!

PR #4459, opened 2022-10-25
Let users display forms on the product lists - MAILPOET-4663

triple0t:
## Description

Add support for displaying forms on the WooCommerce Shop product listing page

## QA notes

* Make sure to select the Shop product listing page or show forms on all pages

## Linked tickets

[MAILPOET-4663](https://mailpoet.atlassian.net/browse/MAILPOET-4663)

### DIFF
--- a/mailpoet/lib/Form/DisplayFormInWPContent.php
+++ b/mailpoet/lib/Form/DisplayFormInWPContent.php
@@ -117,12 +117,11 @@ class DisplayFormInWPContent {
   private function displayFormInListingPage(): bool {
     $displayCheck = $this->wp->applyFilters('mailpoet_display_form_in_product_listing', true);
 
-    if (function_exists('wc_get_page_id')) {
-      $shopPageId = wc_get_page_id('shop');
-      $this->wooShopPageId = $shopPageId && $shopPageId >= 0 ? $shopPageId : null;
-      if ($displayCheck && !is_null($this->wooShopPageId) && $this->wp->isPage($shopPageId)) {
-        return true;
-      }
+    $shopPageId = $this->wp->wcGetPageId('shop');
+    $this->wooShopPageId = $shopPageId && $shopPageId >= 0 ? $shopPageId : null;
+
+    if ($displayCheck && !is_null($this->wooShopPageId) && $this->wp->isPage($this->wooShopPageId)) {
+      return true;
     }
 
     return $displayCheck && $this->wp->isArchive() && $this->wp->isPostTypeArchive('product');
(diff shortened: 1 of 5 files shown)


mailpoet/lib/Form/DisplayFormInWPContent.php
```diff
@@ -104,11 +106,28 @@ private function shouldDisplay(): bool {
...
     if (!$this->wp->isSingle() && !$this->wp->isPage()) $result = $this->wp->applyFilters('mailpoet_display_form_is_single', false);
+
+    if ($this->displayFormInListingPage()) $result = true;
+
     $noFormsCache = $this->wp->getTransient(DisplayFormInWPContent::NO_FORM_TRANSIENT_KEY);
     if ($noFormsCache === '1') $result = false;
     return $result;
   }
 
+  private function displayFormInListingPage(): bool {
```
costasovo: The code is very specific for Woo products page, but the name of the method is quite generic. I think it would be nicer to name it `displayFormInProductListPage` or somewhat similar. What do you think?

mailpoet/lib/Form/DisplayFormInWPContent.php
```diff
@@ -117,12 +117,11 @@ private function shouldDisplay(): bool {
   private function displayFormInListingPage(): bool {
     $displayCheck = $this->wp->applyFilters('mailpoet_display_form_in_product_listing', true);
 
-    if (function_exists('wc_get_page_id')) {
-      $shopPageId = wc_get_page_id('shop');
-      $this->wooShopPageId = $shopPageId && $shopPageId >= 0 ? $shopPageId : null;
-      if ($displayCheck && !is_null($this->wooShopPageId) && $this->wp->isPage($shopPageId)) {
-        return true;
-      }
+    $shopPageId = $this->wp->wcGetPageId('shop');
```
costasovo: We have [a WooCommerce Helper class](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/WooCommerce/Helper.php) that is used the same way as WPFunctions. Could you please move the wrapped function into the Woo helper so that it is together with other Woo funcitons?

costasovo (changes requested):
Looks good! The code for display logic is getting quite complex, so thanks a lot for extending the test coverage.

I added 2 remarks with small improvements.

mailpoet/lib/Form/DisplayFormInWPContent.php
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/4459/commits/d51ae1859d890471ba4efd3529380d10fa45ce37

mailpoet/lib/Form/DisplayFormInWPContent.php
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/4459/commits/d51ae1859d890471ba4efd3529380d10fa45ce37

costasovo (approved):
LGTM 👍

Aschepikov (approved):
Verified, GJ!

PR #4460, opened 2022-10-26
Make the labels in form editor consistent - MAILPOET-4650

triple0t:
## Description

Make the labels in the form editor consistent. 

## Linked tickets

[MAILPOET-4650](https://mailpoet.atlassian.net/browse/MAILPOET-4650)

### DIFF
--- a/mailpoet/assets/js/src/form_editor/components/form_settings/form_placement_options/settings_panels/placement_settings.tsx
+++ b/mailpoet/assets/js/src/form_editor/components/form_settings/form_placement_options/settings_panels/placement_settings.tsx
@@ -220,7 +220,9 @@ export function PlacementSettings({
         />
       </div>
       <div>
-        <p>{MailPoet.I18n.t('displayOnCategories')}</p>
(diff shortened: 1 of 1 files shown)


websupporter (changes requested):
Thank you @triple0t 

Looks good to me. Just one remark:
image


I think we can get rid of the CSS definitions for `.form-editor-sidebar-heading`

triple0t:
Thanks for the review, @websupporter  🙌🏿

> I think we can get rid of the CSS definitions for `.form-editor-sidebar-heading`

I initially thought this was coming from Gutenberg and might be used by other components, so I left it alone.

Taking a deeper look, It would appear the CSS class was used for specific purposes (I noticed it was defined in two separate files). Rather than removing it, I would instead add the class back.

I will push a new commit to reflect that now.

websupporter (approved):
lgtm. Thank you @triple0t

Aschepikov (approved):
Verified, GJ!

PR #4462, opened 2022-10-26
Deactivate 'update/activate' buttons when status is active/deactivting

websupporter:
## Linked tickets
[MAILPOET-4744]

[MAILPOET-4744]: https://mailpoet.atlassian.net/browse/MAILPOET-4744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
@@ -1,5 +1,10 @@
 import { useState } from 'react';
-import { Button, NavigableMenu, TextControl } from '@wordpress/components';
+import {
+  Button,
+  NavigableMenu,
+  TextControl,
+  Tooltip,
+} from '@wordpress/components';
 import { dispatch, useDispatch, useSelect } from '@wordpress/data';
 import { PinnedItems } from '@wordpress/interface';
 import { __ } from '@wordpress/i18n';
@@ -19,36 +24,62 @@ import {
 //   https://github.com/WordPress/gutenberg/blob/0ee78b1bbe9c6f3e6df99f3b967132fa12bef77d/packages/edit-site/src/components/header/index.js
 
 function ActivateButton({ onClick, label }): JSX.Element {
-  const { errors } = useSelect(
+  const { errors, isDeactivating } = useSelect(
     (select) => ({
       errors: select(storeName).getErrors(),
+      isDeactivating:
+        select(storeName).getWorkflowData().status ===
+        WorkflowStatus.DEACTIVATING,
     }),
     [],
   );
 
-  return (
+  const button = (
     <Button
       variant="primary"
       className="editor-post-publish-button"
       onClick={onClick}
-      disabled={!!errors}
+      disabled={isDeactivating || !!errors}
     >
       {label}
     </Button>
   );
+
+  if (isDeactivating) {
+    return (
+      <Tooltip
+        text={__(
+          'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality',
+          'mailpoet',
+        )}
+      >
+        {button}
+      </Tooltip>
+    );
+  }
+
+  return button;
 }
 
 function UpdateButton(): JSX.Element {
   const { save } = useDispatch(storeName);
 
   return (
-    <Button
-      variant="primary"
-      className="editor-post-publish-button"
-      onClick={save}
+    <Tooltip
+      text={__(
+        'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality',
+        'mailpoet',
+      )}
     >
-      {__('Update', 'mailpoet')}
-    </Button>
+      <Button
+        variant="primary"
+        className="editor-post-publish-button"
+        onClick={save}
+        disabled
+      >
+        {__('Update', 'mailpoet')}
+      </Button>
+    </Tooltip>
   );
 }
 
--- a/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
@@ -68,6 +68,24 @@ function ActivateButton({ onClick, label }): JSX.Element {
 function UpdateButton(): JSX.Element {
   const { save } = useDispatch(storeName);
 
+  const { workflow } = useSelect(
+    (select) => ({
+      workflow: select(storeName).getWorkflowData(),
+    }),
+    [],
+  );
+
+  if (workflow.stats.totals.in_progress === 0) {
+    return (
+      <Button
+        variant="primary"
+        className="editor-post-publish-button"
+        onClick={save}
+      >
+        {__('Update', 'mailpoet')}
+      </Button>
+    );
+  }
   return (
     <Tooltip
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
--- a/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
@@ -48,6 +48,10 @@ function ActivateButton({ onClick, label }): JSX.Element {
   if (isDeactivating) {
     return (
       <Tooltip
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        // The following error seems to be a mismatch. It claims the 'delay' prop does not exist, but it does.
+        delay={0}
         text={__(
           'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality.',
           'mailpoet',
@@ -66,6 +70,10 @@ function UpdateButton(): JSX.Element {
 
   return (
     <Tooltip
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      // The following error seems to be a mismatch. It claims the 'delay' prop does not exist, but it does.
+      delay={0}
       text={__(
         'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality.',
         'mailpoet',
--- a/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/header/index.tsx
@@ -49,7 +49,7 @@ function ActivateButton({ onClick, label }): JSX.Element {
     return (
       <Tooltip
         text={__(
-          'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality',
+          'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality.',
           'mailpoet',
         )}
       >
@@ -67,7 +67,7 @@ function UpdateButton(): JSX.Element {
   return (
     <Tooltip
       text={__(
-        'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality',
+        'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality.',
         'mailpoet',
       )}
     >
--- a/mailpoet/assets/js/src/automation/editor/index.tsx
+++ b/mailpoet/assets/js/src/automation/editor/index.tsx
@@ -47,7 +47,7 @@ const showInserterSidebar = false;
  *
  * see MAILPOET-4744
  */
-function updateingActiveWorkflowNotPossible() {
+function updatingActiveWorkflowNotPossible() {
   const workflow = globalSelect(storeName).getWorkflowData();
   if (
     ![WorkflowStatus.ACTIVE, WorkflowStatus.DEACTIVATING].includes(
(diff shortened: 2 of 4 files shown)


mailpoet/assets/js/src/automation/editor/components/header/index.tsx
```diff
@@ -19,36 +24,62 @@ import {
...
     >
       {label}
     </Button>
   );
+
+  if (isDeactivating) {
+    return (
+      <Tooltip
+        text={__(
+          'Editing an active workflow is temporarily unavailable. We are working on introducing this functionality',
```
johnolek: Missing a period at the end of the second sentence (as is the one down on line 70).

johnolek (changes requested):
Just a tiny detail in terms of actual specs, but the tooltip text is missing a period at the end of the second sentence.

Although these changes achieve everything the ticket is asking for, I think the UX could be improved and I'd like to get @NeosinneR 's opinion on some of my thoughts:

1. I think it should be very clear on page load that it won't be possible to save any changes without deactivating first, otherwise users might start making changes and then get frustrated when they discover they can't update without deactivating first. 

It's true that we're disabling the `Update` button, but this is a perfectly reasonable expectation on page load because no changes have been made yet so there wouldn't be anything to update. I think a user would reasonably expect that the `Update` button would become active once they start making changes, which isn't the case, and also wouldn't be immediately obvious.

2. We're not locking down any other functionality. If it's not possible to save changes I wouldn't expect to be able to actually edit step settings or add steps or anything else that would need to be saved.

3. The tooltip from `@wordpress/components` has a delay that I think makes all of this behavior even less obvious. It looks like the [latest version of the tooltip](https://wordpress.github.io/gutenberg/?path=/story/components-tooltip--default) lets you specify a `delay` prop, and would allow a value of `0` to show it immediately, but we're on an older version of `@wordpress/components` and I don't think we have the option of specifying a `delay`. I think the default delay we're working with is 700ms.

I also noticed that the tooltip will not render at all if you click on the element, no matter how long you wait after clicking. Your cursor has to leave the element and go back in before it will display (at least in my testing in firefox).

I made this little video showing how an impatient user might end up having a hard time seeing the explanation about why they can't update their workflow:

https://user-images.githubusercontent.com/8656640/198153724-b82ba239-91b6-426d-a059-b16f71cd4c32.mov

4. Is there any reason we shouldn't allow updating if there aren't any users in progress? This is especially relevant because we [only show the deactivation warning modal](https://github.com/mailpoet/mailpoet/blob/4b6bfab3a65199b7f344a53298814088e7901cc6/mailpoet/assets/js/src/automation/editor/components/header/index.tsx#L109-L112) when there are users in progress. 

If I were to activate a workflow and then immediately notice something I want to be different, I would have to:
a. make the change
b.  see that the `Update` button is disabled
c. hopefully hover long enough to see the tooltip explaining why it's disabled
d. deactivate the workflow (almost certainly without seeing the warning modal, unless someone entered the workflow right away)
e. Press `Update & Activate`.
f. Press the second `Activate` button

I might not feel confident as a user that I was doing things correctly after going through all these steps, and at best it's pretty cumbersome and it would leave me wondering as a user why the MailPoet team wasn't able to implement updating an already-active workflow if all I needed to do was turn it off and on again.

Long story short, I think we could improve this quite a bit by:
1) Allowing updates as long as there aren't currently any users in progress (we might want to check for in-progress users at save time, in case users entered the workflow after the editor loaded)
2) Displaying a message/notice somewhere obvious on page load explaining that the workflow will need to be deactivated before it can be updated (if there are users in progress) instead of relying on a tooltip on the update button, which I think users will be most likely to encounter after they've already made changes they want to save.

There are other things we could do that might be nice, like disabling/making read-only the form fields for step settings, but I wouldn't want to do anything more complex like that because I know we'll be revisiting this post-MVP.

Maybe this PR wasn't the best place for this comment since the changes do achieve everything specified in the ticket, and of course feel free to disagree if you don't think these concerns are problematic enough to warrant any changes!

mailpoet/assets/js/src/automation/editor/components/header/index.tsx
websupporter: Addressed in [f55a3b3](https://github.com/mailpoet/mailpoet/pull/4462/commits/f55a3b31b821b2a624b113e1beea9c45134d309d)

websupporter:
> and I don't think we have the option of specifying a delay

I checked, and we do have the delay property:
https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4019.16.0/packages/components/src/tooltip/index.js#L114-L121

but it seems `@types/wordpress` does not reflect them properly.

I added `delay 0` in [128d7d4](https://github.com/mailpoet/mailpoet/pull/4462/commits/128d7d47217120f2ede7199cb367d24534947f9a)

websupporter:
Thank you for your review and especially on your UI warnings. I think they are warranted.

My thinking: 
1. `Editing an active workflow is temporarily unavailable`. What does "temporarily" mean? Will we fix it in 1 week, 1 month, 1 year?
2. During this time span: Is it more harmful to have the current behavior (edit active workflow) or this disabled update UI?
3. Would showing a explaining modal when the editor loads help?

I hope temporarily means max 1 month. The problem is: If you edit the newsletter of an activate workflow users who enter the send email step during this time will get an imperfect newsletter saved via autosave.

The fix imo is an additional state: `paused`, which the Workflow enters when the editor opens. But this brings lots of questions: You click "Activate" and you see "Paused"... Another option might be to introduce something like WordPress `_edit_lock` meta value, which when set would block `WorkflowRun`s to run.

But I am curious about the opinion of @NeosinneR

websupporter:
Thinking a bit more about the issue: If it is supposed to fix our "autosave"-problem with the newsletter; this issue will not fix it at all, because the autosave happens in the newsletter-editor and not the automation-editor

NeosinneR:
Some explanation first, then off to specific notes: This ticket is really meant to be temporary, even if it might introduce worse UX in terms of editing an active workflow, but it should prevent inconsistencies of what happens when you edit an active workflow with subscribers in it. Right after the MVP, we'll work on improving this, but it's not a trivial change, so I went with the least complex way to "fix" this for MVP. 

@johnolek 
> Allowing updates as long as there aren't currently any users in progress (we might want to check for in-progress users at save time, in case users entered the workflow after the editor loaded)

This sounds good. If easy to do, let's do it. 

@websupporter 
> Editing an active workflow is temporarily unavailable. What does "temporarily" mean? Will we fix it in 1 week, 1 month, 1 year?

You are right, that it's more in terms of weeks. I'd definitely want to get it done by the end of the year.

> Would showing a explaining modal when the editor loads help?

The tooltip that is shown when you hover the disabled "Update" button, can we show it on page load? It would be dismissed when the user hovers over it or the button and then goes away with the mouse.

> If it is supposed to fix our "autosave"-problem with the newsletter

No, and I'm not even worried about users changing emails inside the workflow, even if they can't save the workflow itself. I see more harm in changing the workflow tree and needing to react somehow to subscribers in it. 

---

So to reiterate, does this ticket as originally written make sense in the context of a temporary solution to prevent changing the workflow tree with subscribers in it? I agree that it's not ideal, but at the same time we are not saying it needs to work perfectly, and this is a beta product. 

I'd still do small improvements as John mentioned
- showing the tooltip immediately and not with a delay
- if easy to detect there is no one inside, allow updating active workflow (checked at the time of saving)
- showing the tooltip on the page load, if there is anyone in the workflow at the time of page load

websupporter:
> showing the tooltip immediately and not with a delay

done here: acf0dd188cb1211a96dfcd9a8bb2fe884e196fd6

> showing the tooltip on the page load

showing the tooltip is not so easy. I decided to go with a snackbar message. I think, it looks better as well for the page load at least:
image
instead of
image
[5b90216](https://github.com/mailpoet/mailpoet/pull/4462/commits/5b90216432a2f60802a3187b72efe67df3d1f95b)

> if easy to detect there is no one inside, allow updating active workflow (checked at the time of saving)

Backend check: [2aea392](https://github.com/mailpoet/mailpoet/pull/4462/commits/2aea392fff70e595f66d526100a2379af07f2f9a)

Frontend: [cd46fa8](https://github.com/mailpoet/mailpoet/pull/4462/commits/cd46fa86db95c0dd24550b4c6dc67ab58bc3b0ab)

Regarding backend: If a user entered the workflow while the editor is open and you click "Update" the editor will see a message "Unknown error" or something. I didn't find an easy way to print the error message. Since it's just  couple of weeks and more of an edge case, I left it as is.

websupporter:
With those changes, I needed to change premium as well: https://github.com/mailpoet/mailpoet-premium/pull/647

mailpoet/assets/js/src/automation/editor/index.tsx
```diff
@@ -42,7 +75,7 @@ function Editor(): JSX.Element {
     showIconLabels,
     workflow,
   } = useSelect(
-    (select) => ({
+    () => ({
```
johnolek: It feels strange to remove the `select` from the `useSelect` callback and instead rely on the imported `select`, but I'm not familiar enough with redux/wordpress data to know if there's a functional difference or if it could cause issues, so I'll ping @JanJakes for a more informed opinion.

mailpoet/assets/js/src/automation/editor/index.tsx
```diff
@@ -69,6 +103,10 @@ function Editor(): JSX.Element {
     setShowActivatePanel(!showActivatePanel);
   };
 
+  if (isBooting) {
+    updateingActiveWorkflowNotPossible();
+    setIsBooting(false);
+  }
```
johnolek: I think that instead of using state here it would be more idiomatic to use `useEffect` with an empty dependency array (or perhaps a dependency array that won't change). Again I'm not sure if this might have any kind of negative effect and I know this is just a temporary solution (which seems to be working as intended locally for me), so I'll once again defer to @JanJakes.

johnolek (commented):
This seems to be working correctly for me locally, but there are a couple of changes I wasn't sure about and just wanted to confirm before approving the PR (see comments).

mailpoet/assets/js/src/automation/editor/index.tsx
websupporter: For context: `select` got imported here
https://github.com/mailpoet/mailpoet/pull/4462/commits/5b90216432a2f60802a3187b72efe67df3d1f95b#diff-644bd4c7e44dd87a74660f1465dcf0bfc1b11437d02a7b52ce4e001de1eccb96R6

I use it in the new `updateingActiveWorkflowNotPossible()` method. Codesniffer yelled, so I just removed it. Another way would be to use another variable name I guess.

My thinking in the end: The code in this PR will be removed quite soon :crossed_fingers:

mailpoet/assets/js/src/automation/editor/index.tsx
JanJakes: @johnolek That's a good catch. Yes, it should be in `useEffect`, the dependency will be `isBooting`, I guess, and the linters may force you to add `updateingActiveWorkflowNotPossible` as well, which is fine, since it's a module-global, which will always have the same identity.

@websupporter Even if this works, it's fragile and sometimes React can throw errors that the state was updated during a render (not in an effect).

I also noticed `updateing` -> should be `updating` but probably not important in a "temporary" code.

Btw. this is one of the things I really dislike about React. They don't invent a new "language" like Svelte, but they heavily bend JS itself and force us to use unintuitive constructs. In SolidJS, this code would be perfectly valid. Not only valid, it would actually be the correct way to do it.

mailpoet/assets/js/src/automation/editor/index.tsx
JanJakes: @websupporter Why did you remove it, in fact? You can just rename it, can't you? Either the import or the local variable. I'm not familiar with the internals of `useSelect` so I don't know what happens when you do this, but I've never seen it used like that in Gutenberg, so I think it would be safer to use the select passed to the hook, as @johnolek points out.

mailpoet/assets/js/src/automation/editor/index.tsx
websupporter: Thank you both for your input.

useEffect: [ae1d2da](https://github.com/mailpoet/mailpoet/pull/4462/commits/ae1d2da4203a880cd7c5f153ee04150b4b42a015)

Typo: [6119835](https://github.com/mailpoet/mailpoet/pull/4462/commits/611983506af2ed0d9261c77c5711b754912591a6)

mailpoet/assets/js/src/automation/editor/index.tsx
websupporter: Thank you. Updated in [dd8a252](https://github.com/mailpoet/mailpoet/pull/4462/commits/dd8a2525ae105aa3ed10be3c74abcecbda0980a2)

johnolek (approved):
Thanks for the updates, this looks good to me!

PR #4465, opened 2022-10-27
Add new API list methods [MAILPOET-4752]

lysyjan:
## Description

Add new MP API methods `listDelete` and `listUpdate` for easier integration with Jetpack CRM.

## Linked tickets

[MAILPOET-4752]

[MAILPOET-4752]: https://mailpoet.atlassian.net/browse/MAILPOET-4752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/api_methods/UpdateList.md
+++ b/doc/api_methods/UpdateList.md
@@ -1,0 +1,38 @@
+[back to list](../Readme.md)
+
+# Add Subscriber
+
+## `array updateList(array $list)`
+
+This method provides functionality for updating a list name or description.
+
+It returns the updated list. See [Get Lists](GetLists.md) for a list data structure description.
+
+## Arguments
+
+### `$list` (required)
+
+An associative array which contains list data.
+
+| Property               | Type         | Limits    | Description                |
+| ---------------------- | ------------ | --------- | -------------------------- |
+| id (required)          | string       | 11 chars  | A id of the list.          |
+| name (required)        | string       | 90 chars  | A name of the list.        |
+| description (optional) | string\|null | 250 chars | A description of the list. |
+
+## Error handling
+
+All expected errors from the API are exceptions of class `\MailPoet\API\MP\v1\APIException`.
+Code of the exception is populated to distinguish between different errors.
+
+An exception of base class `\Exception` can be thrown when something unexpected happens.
+
+Codes description:
+
+| Code | Description                                  |
+| ---- | -------------------------------------------- |
+| 5    | The list was not found by id                 |
+| 14   | Missing list name                            |
+| 15   | Trying to use a list that is already used    |
+| 18   | Missing list id                              |
+| 19   | The list couldn’t be updated in the database |
--- a/doc/api_methods/UpdateList.md
+++ b/doc/api_methods/UpdateList.md
@@ -4,7 +4,7 @@
 
 ## `array updateList(array $list)`
 
-This method provides functionality for updating a list name or description.
+This method provides functionality for updating a list name or description. Only lists of type 'default' are supported.
 
 It returns the updated list. See [Get Lists](GetLists.md) for a list data structure description.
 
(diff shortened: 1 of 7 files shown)


doc/api_methods/UpdateList.md
```diff
@@ -0,0 +1,38 @@
...
+
+An exception of base class `\Exception` can be thrown when something unexpected happens.
+
+Codes description:
+
+| Code | Description                                  |
+| ---- | -------------------------------------------- |
+| 5    | The list was not found by id                 |
+| 14   | Missing list name                            |
+| 15   | Trying to use a list that is already used    |
```
brezocordero:
```suggestion
| 15   | Trying to use a list name that is already used    |
```

mailpoet/lib/API/MP/v1/Segments.php
```diff
@@ -27,31 +39,125 @@ public function getAll(): array {
...
+            'mailpoet'
+          )
+        ),
+        APIException::LIST_USED_IN_FORM
       );
     }
 
     try {
-      $segment = $this->segmentsRepository->createOrUpdate($data['name'], $data['description'] ?? '');
+      $this->segmentsRepository->bulkDelete([$listId]);
```
brezocordero, line 123: If I try to delete a list that can not be deleted, because type !== 'default', ex: 'WooCommerce Customers', the list is not deleted but the API returns true.

One option would be to the check the return of `$this->segmentsRepository->bulkDelete([$listId]);` 
if > 0 return true, 
if not, throw APIException::FAILED_TO_DELETE_LIST (or a generic Exception that would convert into APIException::FAILED_TO_DELETE_LIST on the catch block)

What do you think? 

You can't delete dynamic segments, is this expected?

doc/api_methods/UpdateList.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Arguments
+
+### `$list` (required)
+
+An associative array which contains list data.
+
+| Property               | Type         | Limits    | Description                |
+| ---------------------- | ------------ | --------- | -------------------------- |
+| id (required)          | string       | 11 chars  | A id of the list.          |
```
brezocordero, line 19: I got a type error when using a string:

`wp> \MailPoet\API\API::MP('v1')->updateList(['id'=>'2','name'=>'superWoo'])
TypeError: MailPoet\Segments\SegmentsRepository::createOrUpdate(): Argument #5 ($id) must be of type ?int, string given,`

mailpoet/lib/API/MP/v1/Segments.php
```diff
@@ -27,31 +39,125 @@ public function getAll(): array {
...
+
+  public function updateList(array $data): array {
+    // firstly validation on list id
+    $this->validateSegmentId((string)($data['id'] ?? ''));
+
+    // secondly validation on list name
+    $this->validateSegmentName($data);
+
+    try {
+      $segment = $this->segmentsRepository->createOrUpdate(
```
brezocordero, line 67: I was able to update segments that you can not update on the interface, like 'WooCommerce customers.
Is this expected?

```
wp> \MailPoet\API\API::MP('v1')->updateList(['id'=>2 ,'name'=>'superWoo'])
=> array(7) {
  ["id"]=>
  string(1) "2"
  ["name"]=>
  string(8) "superWoo"
  ["type"]=>
  string(17) "woocommerce_users"
  ["description"]=>
  string(0) ""
  ["created_at"]=>
  string(19) "2021-07-29 12:34:17"
  ["updated_at"]=>
  string(19) "2022-10-28 00:28:01"
  ["deleted_at"]=>
  NULL
}
```

`updateList` allows you to update dynamic segments, `deleteList` does not let you delete dynamic segments.
Not sure which one is correct. If this is expected, let's add a note on the documentation.

brezocordero (changes requested):
@lysyjan nice work!

I found one type mismatch and some behavior that is different from what you can do on MailPoet>Lists. I don't know if the differences are expected, in that case, maybe updating the documentation is enough.

mailpoet/lib/API/MP/v1/Segments.php
lysyjan: Thank you, Brezo! Nice catch. MP API should allow default lists to be deleted because it only supports creating default lists.

mailpoet/lib/API/MP/v1/Segments.php
lysyjan: I will add checking the list type. 👍

lysyjan:
@brezocordero Thank you for your excellent review. I added three new commits that should cover all issues.

brezocordero (approved):
Looks good!

PR #4467, opened 2022-10-28
Migrations [MAILPOET-4466]

JanJakes:
## Description

Implements new `MailPoet\Migrations` and fully replaces`MailPoet\Config\Migrator` (and a part of `MailPoet\Config\Populator`).

The new migrator is executed in the same places as the previous one (in activator). It also adds a few commands for development:

```bash
./do migrations:new     # creates a new migration file
./do migrations:status  # displays detailed info about migrations (must be run in Docker)
./do migrations:run     # runs all new migrations (must be run in Docker)
```

## Code review notes

I recommend looking commit-by-commit, hopefully it's clear that way.

## QA notes

Please test plugin upgrade from older versions such as `3.26.0`, `3.31.1`, `3.38.1`, `3.42.0`, `3.46.13`, `3.47.6`, `3.48.1`, `3.49.1`, `3.60.4`, `3.63.0`, `3.67.0`, `3.73.0`, `3.73.2`, `3.74.2`, `3.75.1`, `3.77.1`, `3.78.0`, `3.99.0`, `3.91.1`.

It is not necessary to test all of them! Two or three random picks should do the job. Maybe not so random, in fact — I recommend including one of the older versions in the mix, that will execute most of the database migration logic.

Note that automations migration is not added yet as we still have the feature flat in place.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4466]

## After-merge notes

Ping @JanJakes to inform the team about new migrations.

[MAILPOET-4466]: https://mailpoet.atlassian.net/browse/MAILPOET-4466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JanJakes:
Still thinking about these two points:

1. Should we rerun failed and timed-out migrations the next time they are executed (e.g. plugin is activated)? It may actually be a safer approach in the WP world, I guess.
2. WP CLI should maybe respect the lock that the activator uses. It's only for development, though, so nothing too important.

brezocordero (commented):
@JanJakes 
I installed the build on an Atomic site with 100.1, it worked but after resetting the plugin I get the following error and can't access the plugin:
`Migration class "MailPoet\Migrations\index" not found`

JanJakes:
@brezocordero Ahh, great catch, thanks! Fixed in https://github.com/mailpoet/mailpoet/pull/4467/commits/6c25edbd9e5450493dd8d8c668ce1830535eb832.

brezocordero (commented):
Works well.
Tested:
- upgrading from the latest version and from 3.56
- resetting the plugin
- re-running migrations by changing the name of the file
- The cli commands

I checked the tests and went through the commits. 
It looks good.

Just one question @JanJakes :
Could we have a case where a long migration is still running and because there is no limit to retries we run it again and again?

JanJakes:
@brezocordero Thanks for the thorough testing!

> Could we have a case where a long migration is still running and because there is no limit to retries we run it again and again?

As it is used now, we should be safe. That's because [the migrator is only ever called from the Activator](https://github.com/mailpoet/mailpoet/blob/6c25edbd9e5450493dd8d8c668ce1830535eb832/mailpoet/lib/Config/Activator.php#L74) and [the Activator handles locking](https://github.com/mailpoet/mailpoet/blob/6c25edbd9e5450493dd8d8c668ce1830535eb832/mailpoet/lib/Config/Activator.php#L54) — meaning it will never run multiple times in parallel.

I was wondering if we should add some additional safeguards here, but as it is handled now, this indeed looks enough. In development and when using the CLI, you could run into this, but on the other hand, the CLI command will not exit until the migration is done.

brezocordero (dismissed):

JanJakes:
@brezocordero I just noticed and fixed a tiny issue, see https://github.com/mailpoet/mailpoet/pull/4467/commits/0147f45c127d984e64fa3d91b28bac5ccbfb6de8, totally missed that before.

Aschepikov (approved):
Verified, GJ!

PR #4470, opened 2022-10-31
Mailpoet  confirmation before delete

websupporter:
## Description
This PR remove the `mailpoet.automation.workflow.delete_step_callback` filter and adds the `mailpoet.automation.workflow.step.more-controls` filter.

The `mailpoet.automation.workflow.step.more-controls` filters an `Record<string,JSX.Element>`, so that third parties can alter and add items to the More menu of a single step.

This enables us to realize [MAILPOET-4674]

## Code review notes
This PR suggests to render the items of the `DropdownMenu` in a different way. My current problem is: I want to be able to close the menu in within a menu item. If a user clicks cancel, not only the delete dialog is supposed to go away, but the menu should close. I did not find a proper way to realize that and hope, there is a better way :)

## Linked PRs
https://github.com/mailpoet/mailpoet-premium/pull/641

## Linked tickets
[MAILPOET-4674]

[MAILPOET-4674]: https://mailpoet.atlassian.net/browse/MAILPOET-4674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4674]: https://mailpoet.atlassian.net/browse/MAILPOET-4674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/components/workflow/step-more-menu.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/workflow/step-more-menu.tsx
@@ -1,25 +1,16 @@
-import { useState } from 'react';
-import { Dropdown, DropdownMenu, MenuItem } from '@wordpress/components';
+import { useState, Fragment } from 'react';
+import { DropdownMenu } from '@wordpress/components';
 import { useSelect } from '@wordpress/data';
 import { moreVertical, trash } from '@wordpress/icons';
 import { __ } from '@wordpress/i18n';
 import { Hooks } from 'wp-js-hooks';
 import { PremiumModal } from 'common/premium_modal';
 import { Step as StepData } from './types';
-import { storeName } from '../../store';
+import { MoreControls, storeName } from '../../store';
 
 type Props = {
   step: StepData;
 };
-
-// I do not like this hack.
-// What I want to achieve: With the lazyOnClose method, I want to be able to control to close
-// the menu from within a menuitem.
-// But there must be a better way then storing currentProps like that.
-let currentProps: Dropdown.RenderProps | null = null;
-function setCurrentProps(current) {
-  currentProps = current;
-}
 
 export function StepMoreMenu({ step }: Props): JSX.Element {
   const { stepType } = useSelect(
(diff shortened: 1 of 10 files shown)


mailpoet/assets/js/src/automation/editor/components/workflow/step-more-menu.tsx
```diff
@@ -1,44 +1,66 @@
...
 };
 
+// I do not like this hack.
+// What I want to achieve: With the lazyOnClose method, I want to be able to control to close
+// the menu from within a menuitem.
+// But there must be a better way then storing currentProps like that.
+let currentProps: Dropdown.RenderProps | null = null;
+function setCurrentProps(current) {
+  currentProps = current;
+}
```
JanJakes: Hmm, it seems to me that this is needed only because the menu doesn't autoclose anymore. And that seems to be caused by the custom rendering of the dropdown children.

The problem with this "custom" approach seems to be that the rendering differs from what the component does. It also differs visually. Before:

Screenshot 2022-10-31 at 13 49 15

After:

Screenshot 2022-10-31 at 13 50 13

Could we keep using the "controls" of the component, as it is probably intended to be used? I worry that this customization could break some edge cases (indeed, like auto-close) or focus management/accessibility.

I actually ran into the same issue [here](https://github.com/mailpoet/mailpoet/blob/4c65374a0c9a1fc4a3f8a5a7011c4d5dda578148/mailpoet/assets/js/src/automation/listing/components/cells/actions.tsx#L19-L20), example usage [here](https://github.com/mailpoet/mailpoet/blob/4c65374a0c9a1fc4a3f8a5a7011c4d5dda578148/mailpoet/assets/js/src/automation/listing/components/menu/delete.tsx#L17).

So my idea is — could you make that filter process an array of objects such as `{ key, control, slot }`? That would keep using the "controls" attribute and fix this closing issue as well.

(Last side note, I guess we're using underscores in action/filter names but I can recheck them all at once later this week, maybe.)

websupporter:
Thank you for your review and improvements @JanJakes. I followed the recommendations.

 In [6e5d28f](https://github.com/mailpoet/mailpoet/pull/4470/commits/6e5d28f714742ea03f4d9bca1ca6edc9c3c07c21) I took the liberty to go through all filters and specified expected return types, which we can use then e.g. in premium to be more type safe. Let me know what you think.

JanJakes (dismissed):
LGTM!

JanJakes (approved):

PR #4473, opened 2022-11-01
Automation fixes [MAILPOET-4757]

JanJakes:
## Description

Fixes and improves workflow validation and various other small things.

## Code review notes

_N/A_

## QA notes

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/645

## Linked tickets

[MAILPOET-4757]

## After-merge notes

_N/A_

[MAILPOET-4757]: https://mailpoet.atlassian.net/browse/MAILPOET-4757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


websupporter:
I can now save empty reply_to values. Is this ok?
image

What I also found, but maybe for another issue: We do not properly check for `is_email()` I think:

image

Other than that. Thank you @JanJakes. lgtm.

JanJakes:
@websupporter

> I can now save empty reply_to values. Is this ok?

I think this is OK, maybe even good, as long as after page reload it will have the toggle closed like "not set".

> What I also found, but maybe for another issue: We do not properly check for is_email() I think:

True that, I can try to add it.

JanJakes:
@websupporter Rebased and updated based on https://a8c.slack.com/archives/C0313447YBC/p1667417933160259, hopefully it behaves now.

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/AtLeastOneTriggerRule.php
```diff
@@ -25,6 +25,11 @@ public function visitNode(Workflow $workflow, WorkflowNode $node): void {
   }
 
   public function complete(Workflow $workflow): void {
+    // run full step validation only for active workflows
+    if ($workflow->getStatus() !== Workflow::STATUS_ACTIVE) {
```
websupporter: should we also validate `Workflow::STATUS_DEACTIVATING`. I think in the end no, because no new user can enter the workflow and the current runs are on a different version. But thinking out loud, maybe I miss something.

mailpoet/assets/js/src/automation/integrations/mailpoet/types/filters.tsx
```diff
@@ -2,6 +2,11 @@
  * The types in this file document the expected return types of specific
  * filters.
  */
+import { State } from '../../../editor/store';
+import { Step } from '../../../editor/components/workflow/types';
+
+// mailpoet.automation.send_email.create_step
+export type SendEmailCreateStepType = (step: Step, state: State) => Step;
```
websupporter: I do not see this type being used in free or premium

websupporter:
Thank you @JanJakes, just two short questions in the comments :)

mailpoet/assets/js/src/automation/integrations/mailpoet/types/filters.tsx
JanJakes: Oh, yeah, I created it first and then didn't use it, it seems. The issue is I don't assign it to a variable anywhere, anything I could type simply. Do you think I should delete the type or try to use it somehow?

mailpoet/lib/Automation/Engine/Validation/WorkflowRules/AtLeastOneTriggerRule.php
JanJakes: That's a good point. This applies to many of the rules. I think I could add a method on the workflow itself, something like `$workflow->needsFullValidation()` or some better name, if anyone has a better idea. Then it would all be in one place.

JanJakes:
@websupporter Finally ready, I hope. New commits start here https://github.com/mailpoet/mailpoet/pull/4473/commits/ec03b00eb5d9799d94e7b5b73af6df471a521bbd.

websupporter (approved):
lgtm

PR #4475, opened 2022-11-01
Use the new api for unsubscribe headers [MAILPOET-4702]

samnajian:
## Description

This PR replaces the `list_unsubscribe` with the new `unsubscribe` object when sending the email body to MSS for sending.

## QA notes

Send a newsletter email and verify you see the `unsubscribe` button/link provided by your email viewer.

## Linked tickets

[MAILPOET-4702]

[MAILPOET-4702]: https://mailpoet.atlassian.net/browse/MAILPOET-4702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Notes:
Please refrain from merging this PR once approved, get in touch with @samnajian before merging

### DIFF
--- a/mailpoet/lib/Mailer/Methods/MailPoet.php
+++ b/mailpoet/lib/Mailer/Methods/MailPoet.php
@@ -10,7 +10,7 @@ use MailPoet\Mailer\Methods\ErrorMappers\MailPoetMapper;
 use MailPoet\Services\AuthorizedEmailsController;
 use MailPoet\Services\Bridge;
 use MailPoet\Services\Bridge\API;
-use MailPoet\Util\Url as UrlUtil;
+use MailPoet\Util\Url;
 
 class MailPoet implements MailerMethod {
   public $api;
@@ -27,8 +27,8 @@ class MailPoet implements MailerMethod {
   /** @var BlacklistCheck */
   private $blacklist;
 
-  /*** @var UrlUtil */
-  private $urlUtil;
+  /*** @var Url */
+  private $url;
 
   public function __construct(
     $apiKey,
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Util/Url.php
```diff
@@ -69,4 +69,8 @@ public function redirectWithReferer($url = null) {
     }
     exit();
   }
+
+  public function isUsingHttps(string $url): bool {
+    return wp_parse_url($url, PHP_URL_SCHEME) === 'https';
```
lysyjan: We use WP functions through are wrapper, as you can see [here](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/AdminPages/PageRenderer.php#L119).

mailpoet/lib/Mailer/Methods/MailPoet.php
```diff
@@ -10,6 +10,7 @@
 use MailPoet\Services\AuthorizedEmailsController;
 use MailPoet\Services\Bridge;
 use MailPoet\Services\Bridge\API;
+use MailPoet\Util\Url as UrlUtil;
```
lysyjan: I apologize, but I'm still surprised about this alias because it looks a little bit weird to me.
I tried to check the code in our plugin, and we use an alias `UrlHelper` or no alias.
But it's up to you :) Feel free to skip this.

lysyjan (commented):
I have noticed only a detail. The second comment about the alias is up to you. You can ignore it.
I was also considering tests, but you cover the new states in the old test cases, so I'm okay with that.

mailpoet/lib/Mailer/Methods/MailPoet.php
samnajian: the naming is basically like converting `Util\Url ` to `UrlUtil` i'm fine with your suggestion as well :)

lysyjan (approved):

Aschepikov (approved):
Verified, GJ!

PR #4476, opened 2022-11-02
Add 'Give feedback' beacon [MAILPOET-4755]

websupporter:
## Code review notes
This PR adds a new `current_page` variable to be used in twig. It then compares if the current page is an automation page and replaces the existing beacon ID with the "Give feedback" beacon ID.

## Linked tickets
[MAILPOET-4755]

[MAILPOET-4755]: https://mailpoet.atlassian.net/browse/MAILPOET-4755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/AdminPages/PageRenderer.php
+++ b/mailpoet/lib/AdminPages/PageRenderer.php
@@ -114,6 +114,7 @@ class PageRenderer {
     $subscribersCacheCreatedAt = $this->transientCache->getOldestCreatedAt(TransientCache::SUBSCRIBERS_STATISTICS_COUNT_KEY) ?: Carbon::now();
 
     $defaults = [
+      'current_page' => sanitize_text_field(wp_unslash($_GET['page'] ?? '')),
       'site_name' => $this->wp->wpSpecialcharsDecode($this->wp->getOption('blogname'), ENT_QUOTES),
       'site_url' => $this->wp->siteUrl(),
       'site_address' => $this->wp->wpParseUrl($this->wp->homeUrl(), PHP_URL_HOST),
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -423,8 +423,8 @@ class Hooks {
     );
   }
 
-  public function setFooter($text) {
-    return '<a href="https://feedback.mailpoet.com/" rel="noopener noreferrer" target="_blank">Give feedback</a>';
+  public function setFooter() {
+    return '<a href="https://feedback.mailpoet.com/" rel="noopener noreferrer" target="_blank">' . esc_html__('Give feedback', 'mailpoet') . '</a>';
   }
 
   public function setupSettingsLinkInPluginPage() {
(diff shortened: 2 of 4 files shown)


mailpoet/lib/AdminPages/PageRenderer.php
```diff
@@ -114,6 +114,7 @@ public function displayPage($template, array $data = []) {
     $subscribersCacheCreatedAt = $this->transientCache->getOldestCreatedAt(TransientCache::SUBSCRIBERS_STATISTICS_COUNT_KEY) ?: Carbon::now();
 
     $defaults = [
+      'current_page' => sanitize_text_field(wp_unslash($_GET['page'] ?? '')),
```
johnolek, line 117: Would it be better to put these WP functions in our wrapper and then call them through that? I'm honestly not sure what the risk is (if any) to calling WP functions directly instead of using `$this->wp` like the rest of the file

johnolek (changes requested):
Looks good to me, but you still need to hide the standard "Give feedback" link in the bottom lefthand corner of the listing and templates pages to meet the ticket requirements.

mailpoet/lib/AdminPages/PageRenderer.php
websupporter: Throughout our code base, we do not use the WPFunctions for `sanitize_text_field` and `wp_unslash`. Both methods are not reflected in WPFunctions either. I think at the moment the code sniffer would react and warn about user input not being sanitized, if we would use WPFunctions.

websupporter:
Thank you for your review @johnolek 

> hide the standard "Give feedback" link

Ah! Good catch. Now I understand the screenshot in the issue. [8ca9eab](https://github.com/mailpoet/mailpoet/pull/4476/commits/8ca9eabac4b858d0c149b6378f5cd9feff0027ef) returns now an empty string.

@NeosinneR 
> Hide the existing “Give feedback” link on the Automations page.

Can be done in two ways. This is a filter where we overwrite the default string `<span id="footer-thankyou">Thank you for creating with <a href="https://wordpress.org/">WordPress</a>.</span>` with our feedback string. So two possibilities, return the default or an empty string. I opted for empty string, but FYI, maybe return default is desired.

NeosinneR:
An empty string is fine by me 👍

mailpoet/lib/Config/Hooks.php
```diff
@@ -423,8 +423,12 @@ public function setupFooter() {
     );
   }
 
-  public function setFooter($text) {
-    return '<a href="https://feedback.mailpoet.com/" rel="noopener noreferrer" target="_blank">Give feedback</a>';
+  public function setFooter(): string {
```
johnolek, line 426: I don't think there's any functional difference, but if you're removing the unused `$text` you could update the `addFilter` call above to either omit the last `1` parameter or pass `0`. But `1` is the default anyway so it shouldn't matter at all.

johnolek (approved):
looks good to me!

PR #4477, opened 2022-11-02
Feat/move the list management from settings - MAILPOET-4669

triple0t:
## Description

Move the list management from the settings page.

* Remove the field “Subscribers can choose from these lists:” from settings.
* Add new checkbox “Show this list on the Manage Subscription page” to the list edit page
* Migrate previous data

## Code review notes

Note: This PR includes a DB migration

## Linked tickets

[MAILPOET-4669](https://mailpoet.atlassian.net/browse/MAILPOET-4669)

### DIFF
--- a/mailpoet/lib/Entities/SegmentEntity.php
+++ b/mailpoet/lib/Entities/SegmentEntity.php
@@ -68,10 +68,10 @@ class SegmentEntity {
   private $averageEngagementScoreUpdatedAt;
 
   /**
-   * @ORM\Column(type="integer", nullable=true, options={"default" : 0})
-   * @var int|null
+   * @ORM\Column(type="boolean")
+   * @var bool
    */
-  private $displayInManageSubscriptionPage;
+  private $displayInManageSubscriptionPage = false;
 
   public function __construct(
     string $name,
@@ -162,12 +162,12 @@ class SegmentEntity {
     $this->averageEngagementScoreUpdatedAt = $averageEngagementScoreUpdatedAt;
   }
 
-  public function getDisplayInManageSubscriptionPage(): ?int {
+  public function getDisplayInManageSubscriptionPage(): bool {
     return $this->displayInManageSubscriptionPage;
   }
 
-  public function setDisplayInManageSubscriptionPage(int $value): void {
-    $this->displayInManageSubscriptionPage = $value;
+  public function setDisplayInManageSubscriptionPage(bool $state): void {
+    $this->displayInManageSubscriptionPage = $state;
   }
 
   /**
--- a/mailpoet/lib/Entities/SegmentEntity.php
+++ b/mailpoet/lib/Entities/SegmentEntity.php
@@ -67,6 +67,12 @@ class SegmentEntity {
    */
   private $averageEngagementScoreUpdatedAt;
 
+  /**
+   * @ORM\Column(type="integer", nullable=true, options={"default" : 0})
+   * @var int|null
+   */
+  private $displayInManageSubscriptionPage;
+
   public function __construct(
     string $name,
     string $type,
@@ -154,6 +160,14 @@ class SegmentEntity {
 
   public function setAverageEngagementScoreUpdatedAt(?\DateTimeInterface $averageEngagementScoreUpdatedAt): void {
     $this->averageEngagementScoreUpdatedAt = $averageEngagementScoreUpdatedAt;
+  }
+
+  public function getDisplayInManageSubscriptionPage(): ?int {
+    return $this->displayInManageSubscriptionPage;
+  }
+
+  public function setDisplayInManageSubscriptionPage(int $value): void {
+    $this->displayInManageSubscriptionPage = $value;
   }
 
   /**
--- a/mailpoet/assets/js/src/form/form.jsx
+++ b/mailpoet/assets/js/src/form/form.jsx
@@ -24,8 +24,19 @@ class FormComponent extends Component {
       this.loadItem(this.props.params.id);
     } else {
       setImmediate(() => {
+        const defaultValues =
+          jQuery('.mailpoet_form').mailpoetSerializeObject();
+        const checkboxField =
+          Array.isArray(this.props.fields) &&
+          this.props.fields.length > 0 &&
+          this.props.fields.find(
+            (field) => field?.type === 'checkbox' && field?.isChecked,
+          );
+        if (checkboxField && checkboxField.name) {
+          defaultValues[checkboxField.name] = '1';
+        }
         this.setState({
-          item: jQuery('.mailpoet_form').mailpoetSerializeObject(),
+          item: defaultValues,
         });
       });
     }
--- a/mailpoet/assets/js/src/segments/form.jsx
+++ b/mailpoet/assets/js/src/segments/form.jsx
@@ -28,6 +28,7 @@ const fields = [
         'showInManageSubscriptionPageTip',
       ),
     },
+    isChecked: true, // default checked state
   },
 ];
 
--- a/mailpoet/assets/js/src/segments/form.jsx
+++ b/mailpoet/assets/js/src/segments/form.jsx
@@ -19,6 +19,16 @@ const fields = [
     type: 'textarea',
     tip: MailPoet.I18n.t('segmentDescriptionTip'),
   },
+  {
+    name: 'showInManageSubscriptionPage',
+    label: MailPoet.I18n.t('showInManageSubscriptionPage'),
+    type: 'checkbox',
+    values: {
+      showInManageSubscriptionPage: MailPoet.I18n.t(
+        'showInManageSubscriptionPageTip',
+      ),
+    },
+  },
 ];
 
 const messages = {
--- a/mailpoet/assets/js/src/settings/pages/basics/manage_subscription.tsx
+++ b/mailpoet/assets/js/src/settings/pages/basics/manage_subscription.tsx
@@ -1,11 +1,10 @@
 import ReactStringReplace from 'react-string-replace';
 import { t } from 'common/functions';
 import { useSetting } from 'settings/store/hooks';
-import { Inputs, Label, PageSelect, SegmentsSelect } from 'settings/components';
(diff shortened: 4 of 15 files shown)


mailpoet/lib/Config/Migrator.php
```diff
@@ -122,6 +124,7 @@ public function segments() {
       'deleted_at timestamp NULL,',
       'average_engagement_score FLOAT unsigned NULL,',
       'average_engagement_score_updated_at timestamp NULL,',
+      'display_in_manage_subscription_page int(1) NULL DEFAULT 0,',
```
costasovo: I think it would be nicer to use `tinyint(1)` for such a boolean flag. Using `int(1)` would still use the same size as `int(11)`. Also, we may not need to make the column nullable, I guess by using the default 0, we initially set all to 0.

mailpoet/lib/Config/Migrator.php
```diff
@@ -1003,4 +1006,48 @@ private function setDefaultValueForLoadingThirdPartyLibrariesForExistingInstalls
 
     return true;
   }
+
+  private function migrateSegmentDisplaySettingsOnManageSubscriptionPage(): bool {
+    // skip the migration if the DB version is higher than 3.91.1 or is not set (a new installation)
```
costasovo: The comment here seems to be outdated. We are also planning an emergency release, so I guess we will need to increase also the tested version here from `3.101.1` to `3.102.1`.

mailpoet/lib/Config/Migrator.php
```diff
@@ -1003,4 +1006,48 @@ private function setDefaultValueForLoadingThirdPartyLibrariesForExistingInstalls
...
+      WHERE table_name = %s AND column_name = 'display_in_manage_subscription_page';
+     ", $segmentsTable));
+
+    if (empty($displayInManageSubscriptionPageColumnExists)) {
+      $addNewColumnQuery = "
+        ALTER TABLE `$segmentsTable`
+        ADD `display_in_manage_subscription_page` int(1) NULL DEFAULT 0;
+      ";
+      $wpdb->query($addNewColumnQuery);
+    }
```
costasovo: You've added the column to the table definition and it will be created via `dbDelta`. Here is [the code that does that](https://github.com/mailpoet/mailpoet/blob/357dede83d2c4c8224cc4ee24011946fb47f4c9b/mailpoet/lib/Config/Migrator.php#L84-L87).
So this block of code for adding the column is not needed here.

costasovo (changes requested):
Looks good! I found a few details that might need addressing. Please see my comments.

The specification mentions, "The standard lists have this field enabled by default".
I think that with the current implementation when a user wants to add a new list and clicks `+ New list` the checkbox is unchecked. So I would consider checking the box for the new lists.

mailpoet/lib/Config/Migrator.php
triple0t: Noted

mailpoet/lib/Config/Migrator.php
triple0t: Noted

mailpoet/lib/Config/Migrator.php
triple0t: Thanks for the note, @costasovo. 🙌🏿

When testing, the column was not created for me. There was probably an issue with the way I ran the migrator.
I have a question: Does this code have a side effect?

costasovo: > When testing, the column was not created for me. There was probably an issue with the way I ran the migrator.

The migrator runs in 2 cases)
1) On every request, the plugin checks if the version is up to date.
2) During plugin activation

So to trigger it locally, you need to deactivate and reactivate the plugin or go to `mailpoet_settings_table` and decrease the value for `db_version`.

> I have a question: Does this code have a side effect?

I'm not sure what you mean. I would prefer removing the code for adding the column so that we don't have the column definition in two places.

mailpoet/lib/Config/Migrator.php
```diff
@@ -1008,29 +1008,14 @@ private function setDefaultValueForLoadingThirdPartyLibrariesForExistingInstalls
   }
 
   private function migrateSegmentDisplaySettingsOnManageSubscriptionPage(): bool {
-    // skip the migration if the DB version is higher than 3.91.1 or is not set (a new installation)
-    if (version_compare($this->settings->get('db_version', '3.102.0'), '3.101.1', '>')) {
+    // skip the migration if the DB version is higher than 3.102.0 or is not set (a new installation)
+    if (version_compare($this->settings->get('db_version', '3.102.1'), '3.102.0', '>')) {
```
costasovo: After yesterday's emergency release, we need to update this code to
`version_compare($this->settings->get('db_version', '3.102.2'), '3.102.1', '>')`

costasovo (changes requested):
Thank you for addressing my notes! Unfortunately, we will need to update versions in the migration version check because of yesterday's emergency release.

mailpoet/lib/Config/Migrator.php
JanJakes: Hmm, but if https://github.com/mailpoet/mailpoet/pull/4467 gets merged first (which I think it should), then this should be rewritten to a migration instead.

JanJakes: Just checked with @Aschepikov, and the migrations should be ready on Monday. I'll publish a P2 about how to use them once they are merged. It's super easy. This code will need to be moved to a migration.

triple0t: Honestly, the migrator part still confuses me a little bit 😩. Thanks for the note @costasovo 

@JanJakes Super excited about the new migration workflow. I will take a look and rebase on the branch. Thank you for jumping in and dropping a comment

JanJakes: @triple0t I'll drop you a draft of the P2 in a DM.

costasovo: @JanJakes Thanks for stepping in! I was thinking that the new migrations are related only to the new tables related to automations.
@triple0t  I'll assign the PR back to you since it will require the rebase.

mailpoet/lib/Entities/SegmentEntity.php
```diff
@@ -67,6 +67,12 @@ class SegmentEntity {
    */
   private $averageEngagementScoreUpdatedAt;
 
+  /**
+   * @ORM\Column(type="integer", nullable=true, options={"default" : 0})
+   * @var int|null
+   */
+  private $displayInManageSubscriptionPage;
+
```
costasovo: Hmm, this is a boolean flag, correct? The doctrine has a boolean type that should be used for such stuff. Please see an [example usage in SubscriberEntity](https://github.com/mailpoet/mailpoet/blob/2d835cdec1df34608f8043f0df7125cbe9e0e93c/mailpoet/lib/Entities/SubscriberEntity.php#L53-L57).
I'm sorry I didn't notice that in the initial review. Could you please change the type to boolean here?
If we expect to support more values then true/false, then please ignore this comment.

costasovo (commented):
Looks good! Thanks for refactoring the migration to the new system.
I noticed one thing I missed in the original review, that I believe would be nice to address. I'm sorry! Please take a look at my comment.

mailpoet/lib/Entities/SegmentEntity.php
triple0t: Thanks for the review @costasovo 
It's 1 or 0 (integer). 
We could typecast it to boolean as the column shouldn’t hold any other value.

mailpoet/lib/Entities/SegmentEntity.php
```diff
@@ -68,8 +68,8 @@ class SegmentEntity {
   private $averageEngagementScoreUpdatedAt;
 
   /**
-   * @ORM\Column(type="integer", nullable=true, options={"default" : 0})
-   * @var int|null
+   * @ORM\Column(type="boolean", nullable=true)
+   * @var bool|null
```
costasovo: Is there a reason why we need the column to be nullable?
In migration, we set the default to 0, so during the alter command, all rows will be set to 0.

I think we could get rid of nullability (add `not null` to the migration ALTER command and remove nullable from here). Then we could set default directly to the property.
```php
private $displayInManageSubscriptionPage = false;
```
and also the getter `getDisplayInManageSubscriptionPage` would always return a boolean.

What do you think?

costasovo (commented):
Thank you for dealing with the boolean value!

I have one more comment about the nullability of the column.

costasovo (dismissed):
Thank you for implementing all my remarks and for your patience. 👍

Aschepikov (commented):
Hi, I found one tiny issue with the dot in the message
image

And one issue with the migration, steps to reproduce:
1. I installed 3.102 from https://github.com/mailpoet/mailpoet/releases and chose 2 lists in the settings 
image
2. Then I installed 3.102.1 from this pr, I saw a message in the settings and the checkbox in the lists (after deleting the cache), but in the manage subscription, I saw all lists, like my selection with the previous plugin was reset. 
image
3. I installed the previous plugin 3.102 again and I saw that my selection of lists was working  
image

triple0t:
Thank you for the review, @Aschepikov 

> Hi, I found one tiny issue with the dot in the message

Noted. Will be fixed.

> And one issue with the migration, steps to reproduce:

I suspect an issue with the migrations on your site.

Could you test this workflow again on a new site (or a fresh WordPress site setup)?

To fix the migration issue on your site.
1. Checkout to this branch (or download the release zip file)
2. Check the `wp_mailpoet_migrations` table, and confirm you have an entry with this name: `Migration_20221108_140545`. If the completed_at column of the entry has a date, please set it to NULL. If you do not have this entry, Skip to the next step
3. Visit the `wp_mailpoet_settings` table, locate the `db_version` entry and edit it. You may reduce the DB version by 1, Save
4. Reload the manage subscription page

Aschepikov (approved):
Hi, thank you, yes on the new site everything works as expected. On my site, I couldn't solve the problem with your steps.

PR #4478, opened 2022-11-02
Implement 1 click post unsubscribe [MAILPOET-4703]

samnajian:
## Description

This PR makes the unsubscription endpoint acknowledge POST requests for 1-click unsubscribe and unsubscribe the user without confirmation

## QA notes

1) Send out a newsletter using MSS ( sending preview to email won't work )
2) Using the unsubscribe button shown by the email viewer application of your choice you should be able to unsubscribe.

- Please note that for this test your website should be using HTTPS
- Ther showing of unsubscribe button/link in the email viewer is added in https://github.com/mailpoet/mailpoet/pull/4475

## Linked PRs
[4475](https://github.com/mailpoet/mailpoet/pull/4475)

## Linked tickets

[MAILPOET-4703]

[MAILPOET-4703]: https://mailpoet.atlassian.net/browse/MAILPOET-4703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Notes:
Please refrain from merging this PR once approved, get in touch with @samnajian before merging

### DIFF
--- a/mailpoet/lib/WP/Functions.php
+++ b/mailpoet/lib/WP/Functions.php
@@ -848,7 +848,7 @@ class Functions {
     return is_wp_error($value);
   }
 
-  public function isSiteUsingHttps(): bool {
+  public function wpIsSiteUrlUsingHttps(): bool {
     return wp_is_site_url_using_https();
   }
 }
--- a/mailpoet/lib/WP/Functions.php
+++ b/mailpoet/lib/WP/Functions.php
@@ -847,4 +847,8 @@ class Functions {
   public function isWpError($value): bool {
     return is_wp_error($value);
   }
+
+  public function isSiteUsingHttps(): bool {
+    return wp_is_site_url_using_https();
+  }
 }
--- a/mailpoet/tests/integration/Mailer/Methods/MailPoetAPITest.php
+++ b/mailpoet/tests/integration/Mailer/Methods/MailPoetAPITest.php
@@ -128,16 +128,32 @@ class MailPoetAPITest extends \MailPoetTest {
   public function testItCanAddExtraParametersToMultipleMessages() {
     $newsletters = array_fill(0, 10, $this->newsletter);
     $subscribers = array_fill(0, 10, $this->subscriber);
+
+    $unsubscribeUrls = array_merge(array_fill(0, 5, 'http://example.com'), array_fill(0, 5, 'https://example.com'));
+    $unsubscribeUrls = array_map(function ($url, $index){
+      return "$url/$index";
+    }, $unsubscribeUrls, array_keys($unsubscribeUrls));
+
+    $oneClickUrls = array_fill(0, 10, 'https://oneclick.com');
+    $oneClickUrls = array_map(function ($url, $index){
+      return "$url/$index";
+    }, $oneClickUrls, array_keys($oneClickUrls));
+
     $extraParams = [
-      'unsubscribe_url' => array_fill(0, 10, 'http://example.com'),
-      'one_click_unsubscribe' => array_fill(0, 10, 'http://oneclick.com'),
+      'unsubscribe_url' => $unsubscribeUrls,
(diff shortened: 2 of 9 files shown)


mailpoet/lib/WP/Functions.php
```diff
@@ -847,4 +847,8 @@ public function getAvailableLanguages(?string $dir = null): array {
   public function isWpError($value): bool {
     return is_wp_error($value);
   }
+
+  public function isSiteUsingHttps(): bool {
```
lysyjan: We use the exact name of the function, but it is converted from snake case to camel case.
So it should be `wpIsSiteUrlUsingHttps`.

mailpoet/tests/integration/Router/Endpoints/SubscriptionTest.php
```diff
@@ -18,18 +19,22 @@ class SubscriptionTest extends \MailPoetTest {
...
+    $this->request = $this->diContainer->get(Request::class);
   }
 
   public function testItDisplaysConfirmPage() {
     $pages = Stub::make(Pages::class, [
       'wp' => $this->wp,
       'confirm' => Expected::exactly(1),
     ], $this);
-    $subscription = new Subscription($pages, $this->wp, $this->captcha);
+    $subscription = new Subscription($pages, $this->wp, $this->captcha, $this->request);
```
lysyjan, line 37: You can simplify this by calling `$this->getServiceWithOverrides()`. It is possible to apply that to each test method.

mailpoet/tests/integration/Mailer/Methods/MailPoetAPITest.php
```diff
@@ -129,6 +130,7 @@ public function testItCanAddExtraParametersToMultipleMessages() {
     $subscribers = array_fill(0, 10, $this->subscriber);
     $extraParams = [
       'unsubscribe_url' => array_fill(0, 10, 'http://example.com'),
+      'one_click_unsubscribe' => array_fill(0, 10, 'http://oneclick.com'),
```
lysyjan: I'm considering if we should improve this test method. If I'm not wrong, each subscriber should have a different one-click unsubscribe URL, but I don't see any check on it now. What do you think about it?

lysyjan (commented):
Good job! I have a few minor remarks only.

mailpoet/tests/integration/Mailer/Methods/MailPoetAPITest.php
samnajian: While your assumption is correct, it highly depends on what we try to test, this test is not testing the structure of the URL, on the other hand the structure remains like `unsubscribe_url`, that's why the test is not adjusted to check for the url structure based on subscriber, but still since the `one_click_unsubscribe` should only be used when the site is using https, this is tested.

mailpoet/tests/integration/Router/Endpoints/SubscriptionTest.php
samnajian: I'm not getting how using this method would make it more simple, at least for these cases.

lysyjan: > I'm not getting how using this method would make it more simple, at least for these cases.

If somebody changes dependencies in the future, then he doesn't need to update those lines. But when you think that it's not necessary I'm fine with that.

samnajian: I'd rather leave it like this, I'm more inclined to like that when changing something breaks tests, and the developer has to update the tests, it's a way of keeping informed about the tests, at least for me.

lysyjan (dismissed):

Aschepikov (approved):
GJ!

PR #4480, opened 2022-11-03
Update Woo Blocks integration code [MAILPOET-4774]

costasovo:
## Description
We were using some deprecated classes in the Woo Blocks integration code. 
This PR updates classes used in the integration and also prevents enabling the integration for older versions of Woo Blocks plugin.

Some [users reported that they get errors ](https://wordpress.org/support/topic/fatal-error-after-update-to-3-102-0/)when the plugin was trying to load the deprecated ChecknoutSchema class. We were not able to replicate the error, but assuming that because the other WooCommerce Blocks classes used in the code did not throw the error, the issue might be related to [backward compatible aliases](https://github.com/woocommerce/woocommerce-blocks/blob/ecc423ea0ec94009369d495dd4494b06405bdc85/src/StoreApi/deprecated.php#L28) that, for some unknown reason, don't work for some users.

To prevent Fatal errors in case someone is using the old WooCommerce blocks plugin, I updated the version check when we activate the integration. I chose the oldest supported Woo Blocks version (8.0.0) to be the same as [the WooCommerce Blocks version used in WooCommerce 6.8.0](https://plugins.trac.wordpress.org/browser/woocommerce/tags/6.8.0/packages/woocommerce-blocks/readme.txt), which is the oldest Woo we support by L-2 rule.

## Code review notes
I also needed to update Woo Stubs for PHP stan. When updating it I manually changed composer.json and prettier reformated whitespaces. I noticed it after I prepared the PR, but I guess that is not a big issue.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4774]

## After-merge notes

_N/A_

[MAILPOET-4774]: https://mailpoet.atlassian.net/browse/MAILPOET-4774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tasks/phpstan/phpstan.neon
+++ b/mailpoet/tasks/phpstan/phpstan.neon
@@ -42,8 +42,6 @@ parameters:
     - '/Parameter #1 \$cssOrXPath of method AcceptanceTester::moveMouseOver\(\) expects string\|null, array<string, string> given./'
     - '/Function expect invoked with 1 parameter, 0 required\./'
     - '/Call to method getName\(\) on an unknown class _generated\\([a-zA-Z])*Cookie/' # codeception generate incorrect return type in ../../tests/_support/_generated
-    - '/Call to static method container\(\) on an unknown class/'
-    - '/Class Automattic\\WooCommerce\\StoreApi\\Schemas\\ExtendSchema not found./'
     -
       message: "#^Cannot cast string|void to string\\.$#"
       count: 5
@@ -52,10 +50,6 @@ parameters:
       message: "#^Cannot cast string|void to string\\.$#"
       count: 5
       path: ../../lib/Automation/Engine/Storage/WorkflowStorage.php
-    -
-      message: '/^Call to static method custom_orders_table_usage_is_enabled\(\) on an unknown class Automattic\\WooCommerce\\Utilities\\OrderUtil\.$/'
-      count: 1
-      path: ../../lib/WooCommerce/Helper.php
     -
       message: '/^Call to static method get_orders_table_name\(\) on an unknown class Automattic\\WooCommerce\\Internal\\DataStores\\Orders\\OrdersTableDataStore\.$/'
       count: 1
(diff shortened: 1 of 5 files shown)


mailpoet/tasks/phpstan/phpstan.neon
```diff
@@ -42,8 +42,6 @@ parameters:
     - '/Parameter #1 \$cssOrXPath of method AcceptanceTester::moveMouseOver\(\) expects string\|null, array<string, string> given./'
     - '/Function expect invoked with 1 parameter, 0 required\./'
     - '/Call to method getName\(\) on an unknown class _generated\\([a-zA-Z])*Cookie/' # codeception generate incorrect return type in ../../tests/_support/_generated
-    - '/Call to static method container\(\) on an unknown class/'
```
rodrigoprimo, line 45: It is not immediately obvious to me which changes in this commit prompted the need to update the phpstan.neon file to remove a few of the ignore errors. Could you please elaborate?

rodrigoprimo (approved):
@costasovo, looks good to me. I left just one comment about something that I did not fully understand but it is probably not a blocker.

Since we are not able to reproduce the error, I imagine it is probably a good idea for QA to thoroughly check MP's integration with the WooCommerce Blocks before we merge this PR.

mailpoet/tasks/phpstan/phpstan.neon
costasovo: When I changed the class in the integration, I got new warnings from the PHPStan complaining that the `Automattic\WooCommerce\StoreApi\Schemas\V1\CheckoutSchema` doesn't exist.
So I could add a new ignore rule to the phpstan.neon, but I checked the WooCommerce Stubs and saw that the newer version contains the class. 
After I updated the Woo Stubs, some previously thrown errors were not thrown anymore because after updating the stubs PHPStan knows the new stuff in Woo.

rodrigoprimo: That makes sense. Thanks for the clarification. Moving this PR to QA.

Aschepikov (approved):
Verified, GJ!

PR #4474, opened 2022-11-01
Add WooCommerce Subscriptions integration tests to the list of tests that can be skipped [MAILPOET-4765]

rodrigoprimo:
## Description

When we started skipping the WooCommerce Subscriptions tests, we forgot to add them to the list of tests that can be skipped when running the tests using the trunk and release branches. This PR adds those tests to this list. Please see the ticket description and the commit messages for more details.

I created a temporary branch to show that skipping the tests is actually working:

https://app.circleci.com/pipelines/github/mailpoet/mailpoet?branch=cot-allow-tests-to-be-skipped-delme

It contains the following commit:

https://github.com/mailpoet/mailpoet/commit/9512c2001336f89cba677b3b33d923cd4c953171

The branch can be deleted after the reviewer checks the result of its CircleCI build.

## Code review notes

_N/A_

## QA notes

I don't think QA is needed for this PR.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4765]

## After-merge notes

_N/A_

[MAILPOET-4765]: https://mailpoet.atlassian.net/browse/MAILPOET-4765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4482, opened 2022-11-03
Revert "Add WooCommerce Subscriptions integration tests to the list of tests that can be skipped [MAILPOET-4765]"

rodrigoprimo:
Reverting mailpoet/mailpoet#4474 as it broke trunk: https://app.circleci.com/pipelines/github/mailpoet/mailpoet/11939/workflows/676f247e-c3b2-4d69-ab92-d191ffd06150/jobs/203672

On a quick look, it seems that for some reason Codeception returns `null` when `$event->getTest()->getMetadata()->getName()` is called for integration tests and that is making the code that was added in this PR always return an error when there is a skipped test in the integration tests. I will investigate this more carefully, but reverting the PR meanwhile to restore the CircleCI builds.

### DIFF
--- a/mailpoet/tests/_support/CheckSkippedTestsExtension.php
+++ b/mailpoet/tests/_support/CheckSkippedTestsExtension.php
@@ -15,12 +15,7 @@ class CheckSkippedTestsExtension extends Extension {
     $testName = $event->getTest()->getMetadata()->getName();
 
     // list of tests that are allowed to be skipped on trunk and release branches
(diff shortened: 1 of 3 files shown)


costasovo (approved):
👍

rodrigoprimo:
Merging even though CircleCI is failing as I tried to rerun it twice already and got errors in different acceptance tests. I'm pretty sure the errors are not related to the changes here.

PR #4483, opened 2022-11-03
Automation: Fatal error in user registration [MAILPOET-4773] [MAILPOET-4772]

websupporter:
## Description
This PR solves two issues at once.

The User Registered Trigger gets rewritten. It will now trigger, once a registered users confirms his subscription. This solved both issues at once.

## Code review notes

I was not sure about the best way to place `<a>` elements in a translated string. I used `ReactStringReplace` which I found in our code base as a solution.

I needed to extend the Segments\WP class. The problem was: When the confirmation email option was `false` our hook would not fire. I added the hook in [c74925a](https://github.com/mailpoet/mailpoet/pull/4483/commits/c74925aac375b6526d239bcc99cc7c837f945125).

I also needed to refresh the copy of the subscriber of the database in the trigger here: [148301c](https://github.com/mailpoet/mailpoet/pull/4483/commits/148301c482213dc4e602e607f7a5dadd84ddc464)

I think probably some model/doctrine issue again. The WP class runs on multiple hooks, which are fired during user registration. In some of those (add_role) the user does not yet exist and so I needed a fresh copy from the database once the WPUser was correctly set.

## Linked tickets
[MAILPOET-4773] [MAILPOET-4772]

[MAILPOET-4773]: https://mailpoet.atlassian.net/browse/MAILPOET-4773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4772]: https://mailpoet.atlassian.net/browse/MAILPOET-4772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/wp-user-registered/edit/role_panel.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/wp-user-registered/edit/role_panel.tsx
@@ -1,10 +1,30 @@
 import { PanelBody } from '@wordpress/components';
 import { dispatch, useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
+import ReactStringReplace from 'react-string-replace';
 import { storeName } from '../../../../../editor/store';
 import { PlainBodyTitle } from '../../../../../editor/components/panel';
 import { userRoles } from './role';
 import { FormTokenField } from '../../../components/form-token-field';
+
+function SettingsInfoText(): JSX.Element {
+  return (
+    <p>
+      {ReactStringReplace(
+        __(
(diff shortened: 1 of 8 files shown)


brezocordero (changes requested):
@websupporter as we talked:
Please check the use case when a WP user is added via WP-Admin
No fatal error and I only got the email once but the workflow run twice.

brezocordero (approved):
LGTM!

PR #4488, opened 2022-11-04
Prevent flakiness when comparing time in NewsletterTest [MAILPOET-4710]

costasovo:
## Description

This commit fixes the flakiness by comparing timestamps with delta 1 second instead of comparing date time strings. 

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4710]

## After-merge notes

_N/A_

[MAILPOET-4710]: https://mailpoet.atlassian.net/browse/MAILPOET-4710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo (approved):
Looks good to me. I was not aware that you could pass a second parameter to `equals()` to set a delta. Thanks for showing me that. Let's hope that 1 second is enough, and if we continue to see flakiness in this test, we can increase the delta.

rodrigoprimo:
I will merge this PR as I don't think it requires QA as it just changes a integration test.

PR #4489, opened 2022-11-04
Link the default privacy policy from the default form templates - MAILPOET-4651

triple0t:
## Description

Link the default privacy policy from the default form templates

Add the replacePrivacyLinkTags method for Form Templates. 

## QA notes

Set privacy policy page here: /wp-admin/options-privacy.php

## Linked tickets

[MAILPOET-4651](https://mailpoet.atlassian.net/browse/MAILPOET-4651)

### DIFF
--- a/mailpoet/lib/Form/Templates/FormTemplate.php
+++ b/mailpoet/lib/Form/Templates/FormTemplate.php
@@ -151,4 +151,17 @@ EOL;
   protected function replaceLinkTags($source, $link, $attributes = [], $linkTag = false): string {
     return Helpers::replaceLinkTags($source, $link, $attributes, $linkTag);
   }
+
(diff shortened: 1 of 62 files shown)


brezocordero (changes requested):
Hi @triple0t 
As we talked, please check if we need to [esc_url](https://developer.wordpress.org/reference/functions/esc_url/)
Thanks!

triple0t:
> As we talked, please check if we need to [esc_url](https://developer.wordpress.org/reference/functions/esc_url/)

@brezocordero Fixed.

Thank you

brezocordero (approved):
Works great.
Thanks for making the change.

Aschepikov (approved):
Verified, GJ!

PR #4490, opened 2022-11-04
Improve stability of AbandonedCartTest [MAILPOET-4723]

costasovo:
## Description

This PR should fix the flakiness of the AbandonedCartTest integration test.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4723]

## After-merge notes

_N/A_

[MAILPOET-4723]: https://mailpoet.atlassian.net/browse/MAILPOET-4723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/_support/IntegrationTester.php
+++ b/mailpoet/tests/_support/IntegrationTester.php
@@ -98,4 +98,18 @@ class IntegrationTester extends \Codeception\Actor {
   public function uniqueId($length = 10): string {
     return Security::generateRandomString($length);
   }
+
+  /**
+   * Compares two DateTimeInterface objects by comparing timestamp values.
+   * $delta parameter specifies tolerated difference
+   */
+  public function assertEqualDateTimes(DateTimeInterface $date1 = null, DateTimeInterface $date2 = null, int $delta = 0) {
+    if (!$date1 instanceof DateTimeInterface) {
+      throw new \Exception('$date1 is not DateTimeInterface');
+    }
+    if (!$date2 instanceof DateTimeInterface) {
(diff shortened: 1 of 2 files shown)


mailpoet/tests/_support/IntegrationTester.php
```diff
@@ -98,4 +98,18 @@ public function deleteTestWooOrders() {
   public function uniqueId($length = 10): string {
     return Security::generateRandomString($length);
   }
+
+  /**
+   * Compares two DateTimeInterface objects by comparing timestamp values.
+   * $delta parameter specifies tolerated difference
+   */
+  public function assertEqualDateTimes(DateTimeInterface $date1 = null, DateTimeInterface $date2 = null, int $delta = 0) {
```
lysyjan: Can I ask why you use all parameters as optional with default values?
In my opinion, the first two parameters can be nullable without a default value. Because if I use this function without `date2`, it throws an exception. Or am I missing something?

I like this idea 👍

lysyjan (commented):
I have a question about the new assert parameters.
Maybe there is a reason for your solution.

mailpoet/tests/_support/IntegrationTester.php
costasovo: @lysyjan Good catch! That's exactly what I wanted to do, but my hands were probably faster than my head... Updated!

lysyjan (approved):
👍

PR #4493, opened 2022-11-04
Prevent flakiness when comparing time in AbandonedCartTest [MAILPOET-4781]

rodrigoprimo:
## Description

When comparing scheduledAt dates there might be one second difference.

This commit fixes the flakiness by comparing timestamps with delta 1 second instead of comparing date time strings.

Recently, we have seen two of the modified tests fail because of this problem:

https://app.circleci.com/pipelines/github/mailpoet/mailpoet/11995

I'm opting to update all four tests that compared the value of scheduledAt for good measure as they all could suffer from the same problem.

## Code review notes

_N/A_

## QA notes

I don't think QA is needed for this PR as it only changes a couple of integration tests.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4781]

## After-merge notes

_N/A_

[MAILPOET-4781]: https://mailpoet.atlassian.net/browse/MAILPOET-4781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
+++ b/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
@@ -205,7 +205,9 @@ class AbandonedCartTest extends \MailPoetTest {
     $scheduledTasks = $this->scheduledTasksRepository->findAll();
     $this->assertCount(1, $scheduledTasks);
     $this->assertEquals($scheduledTasks[0]->getStatus(), ScheduledTaskEntity::STATUS_SCHEDULED);
-    $this->assertEquals($scheduledTasks[0]->getScheduledAt(), $expectedTime);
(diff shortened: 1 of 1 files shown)


lysyjan (commented):
We already merged [a PR](https://github.com/mailpoet/mailpoet/pull/4490/files) with the same topic.
We can close this pull request and mark the ticket as won't do. @rodrigoprimo What do you think?

rodrigoprimo:
I agree, @lysyjan. @costasovo was faster than me :-)

I should have searched for tickets in Jira before creating [MAILPOET-4781](https://mailpoet.atlassian.net/browse/MAILPOET-4781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ).

PR #4499, opened 2022-11-07
Fix acceptance test UserRegistrationTriggered [MAILPOET-4782]

veljkho:
## Description

Fixing acceptance test for both single and multi-site scenarios as multi-site require now different approach due to the different design of the registration flow. Now, in multi-site, there's different text on the `Submit` button and after registering different text to verify. Additionally removed - from the username names as this is another different rule for multi-side registering to include only letters and numbers.

## Code review notes

This approach works, but I was using try/catch method and I hope it's fine.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4782](https://mailpoet.atlassian.net/browse/MAILPOET-4782)

## After-merge notes

_N/A_

### DIFF
--- a/mailpoet/tests/acceptance/Automation/UserRegistrationTriggerCest.php
+++ b/mailpoet/tests/acceptance/Automation/UserRegistrationTriggerCest.php
@@ -120,7 +120,7 @@ class UserRegistrationTriggerCest
       $i->click('#mailpoet_subscribe_on_register');
     }
     $i->click('Next');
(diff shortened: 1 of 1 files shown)


rodrigoprimo (approved):
Looks good to me. I'm not sure if the try/catch block is the best alternative here, but it works and I don't think it is worth delaying the release preparation even further. We can always revisit and improve this part of the test in the future.

PR #4502, opened 2022-11-07
Use `workflow_triggers` table instead of an inline JSON

JanJakes:
## Description

Use `workflow_triggers` table instead of an inline JSON array that is not indexable.

## Code review notes

_N/A_

## QA notes

Automations only.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4778]

## After-merge notes

_N/A_

[MAILPOET-4778]: https://mailpoet.atlassian.net/browse/MAILPOET-4778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
+++ b/mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
@@ -11,7 +11,7 @@ use wpdb;
 
 class WorkflowStorage {
   /** @var string */
-  private $workflowTable;
+  private $workflowsTable;
 
   /** @var string */
   private $versionsTable;
@@ -24,7 +24,7 @@ class WorkflowStorage {
 
   public function __construct() {
     global $wpdb;
-    $this->workflowTable = $wpdb->prefix . 'mailpoet_workflows';
+    $this->workflowsTable = $wpdb->prefix . 'mailpoet_workflows';
     $this->versionsTable = $wpdb->prefix . 'mailpoet_workflow_versions';
     $this->triggersTable = $wpdb->prefix . 'mailpoet_workflow_triggers';
     $this->wpdb = $wpdb;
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -211,4 +242,46 @@ private function insertWorkflowVersion(int $workflowId, Workflow $workflow): voi
...
+
+  private function insertWorkflowTriggers(int $workflowId, Workflow $workflow): void {
+    $triggerKeys = [];
+    foreach ($workflow->getSteps() as $step) {
+      if ($step->getType() === Step::TYPE_TRIGGER) {
+        $triggerKeys[] = $step->getKey();
+      }
+    }
+
+    if (!$triggerKeys) {
```
websupporter: Not sure, maybe we should delete existing trigger_keys when no trigger_keys are found. Just to have the correct representation in the database.

To reproduce: If you delete the trigger in your draft and store the draft, the old keys remain in the database.

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
```diff
@@ -94,42 +105,48 @@ public function getWorkflows(array $status = null): array {
...
-    $triggerKeys = [];
-    foreach ($result as $item) {
-      /** @var string[] $keys */
-      $keys = Json::decode($item);
-      $triggerKeys = array_merge($triggerKeys, $keys);
-    }
-    return array_unique($triggerKeys);
+    $query = (string)$this->wpdb->prepare(
+      "
+        SELECT DISTINCT triggers.trigger_key
```
websupporter: `triggers` as a table alias seems quite close to the `CREATE TRIGGER` command. It gets strangely highlighted by my PHPStorm. I didnt find any reserved word `triggers` in MySql, but maybe we should either escape or rename the alias, just to be on the safe side?

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
JanJakes: Oh, of course, totally missed that 🤦 Thanks!

mailpoet/lib/Automation/Engine/Storage/WorkflowStorage.php
JanJakes: 👍 I'm also thinking about using short single-letter aliases. I think we're doing that in many parts of the MP codebase and it seems quite readable.

websupporter (approved):
lgtm

PR #4511, opened 2022-11-09
Fix user registration test [MAILPOET-4785]

websupporter:
## Description

When you register as a user in a multisite, you need to activate your account, before a `WP_User` is created. Actions like `user_register` fire only after you have activated your account. This PR therefore goes through the whole registration process.

The second change that was needed: The activation takes place on `wp-activate.php`, but there only network-wide plugins can be active, as you are not in the context of a specific site/blog. So, the test ensures, mailpoet is active network-wide.

The last change: The activation email gets send with `wp_mail()`. For some reason `wp_mail()` did not send. We had already a plugin in place, which made sure `wp_mail()` would send with SMTP in a WP 5.6 environment. I changed this to always send via SMTP.

## QA notes

run `./do test:acceptance-multisite --skip-deps --file=tests/acceptance/Automation/UserRegistrationTriggerCest.php` to verify the test works on multisite.

## Linked tickets

[MAILPOET-4785]

[MAILPOET-4785]: https://mailpoet.atlassian.net/browse/MAILPOET-4785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Automation/UserRegistrationTriggerCest.php
+++ b/mailpoet/tests/acceptance/Automation/UserRegistrationTriggerCest.php
@@ -49,10 +49,12 @@ class UserRegistrationTriggerCest
     $this->workflowStorage = $this->container->get(WorkflowStorage::class);
     $this->workflowRunStorage = $this->container->get(WorkflowRunStorage::class);
     $this->workflowRunLogStorage = $this->container->get(WorkflowRunLogStorage::class);
+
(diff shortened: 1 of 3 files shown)


JanJakes (approved):
Works and LGTM!

PR #4512, opened 2022-11-09
Prevent leaving unsaved automation

brezocordero:
## Description

Add a pop-up requesting confirmation if leaving the workflow editor with unsaved changes

## Code review notes

_N/A_

## QA notes

To test:
- With feature flag
- Edit a workflow or create a new one
- Make a modification
- Reload the page or click on another page
- Check there is a pop up asking for confirmation to leave with unsaved changes

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4776]

## After-merge notes

_N/A_

[MAILPOET-4776]: https://mailpoet.atlassian.net/browse/MAILPOET-4776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/index.tsx
+++ b/mailpoet/assets/js/src/automation/editor/index.tsx
@@ -72,6 +72,25 @@ function updatingActiveWorkflowNotPossible() {
   );
 }
 
+function onUnload(event) {
+  if (!globalSelect(storeName).getWorkflowSaved()) {
+    // eslint-disable-next-line no-param-reassign
+    event.returnValue = __(
+      'There are unsaved changes that will be lost. Do you want to continue?',
+      'mailpoet',
+    );
+    return event.returnValue;
+  }
+  return '';
+}
+
+function useConfirmUnsaved() {
+  useEffect(() => {
+    window.addEventListener('beforeunload', onUnload);
+    return () => window.removeEventListener('beforeunload', onUnload);
+  }, []);
+}
+
 function Editor(): JSX.Element {
   const {
     isFullscreenActive,
@@ -91,6 +110,8 @@ function Editor(): JSX.Element {
   );
   const [showActivatePanel, setShowActivatePanel] = useState(false);
   const [isBooting, setIsBooting] = useState(true);
+
+  useConfirmUnsaved();
 
   useEffect(() => {
     if (!isBooting) {
--- a/mailpoet/tests/acceptance/Automation/ConfirmLeaveWhenUnsavedChangesCest.php
+++ b/mailpoet/tests/acceptance/Automation/ConfirmLeaveWhenUnsavedChangesCest.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace MailPoet\Test\Acceptance;
+
+use MailPoet\Automation\Engine\Migrations\Migrator;
+use MailPoet\DI\ContainerWrapper;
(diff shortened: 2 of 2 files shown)


mailpoet/assets/js/src/automation/editor/index.tsx
```diff
@@ -72,26 +72,45 @@ function updatingActiveWorkflowNotPossible() {
   );
 }
 
+function useConfirmUnsaved(isSaved) {
+  function onUnload(event) {
+    if (!isSaved) {
+      event.returnValue = __('changesNotSaved', 'mailpoet'); // eslint-disable-line no-param-reassign
```
JanJakes: This seems to try to fetch a string from the legacy translation system. Could we put here the string value itself? Not sure, if we want the `The changes you made may not be saved` or something else (`...may be lost`?). @NeosinneR could answer that, I guess.

mailpoet/assets/js/src/automation/editor/index.tsx
```diff
@@ -72,26 +72,45 @@ function updatingActiveWorkflowNotPossible() {
...
+      return event.returnValue;
+    }
+    return '';
+  }
+
+  useEffect(() => {
+    window.addEventListener('beforeunload', onUnload);
+    return () => window.removeEventListener('beforeunload', onUnload);
+  });
```
JanJakes: I think having no deps array here will result in adding and removing the listener on every re-render. I think there should be a dependency array. But it's interesting the linter doesn't say anything 🤔

It may not be a big deal in this case but I think the correct dependency array for the `useEffect` call is `[onUnload]`. But since the function is redefined on each render, it won't help. It would also need `useCallback` to keep a stable identity:

```ts
const onUnload = useCallback(() => ..., [isSaved]);
```

But thinking further, even that is problematic, as when the `onUnload` changes, the `removeEventListener` will fail to remove it on unmount as it will be a different function. I think here is a very similar issue: https://stackoverflow.com/questions/53750619/event-listener-is-not-removed-in-useeffect-call-within-function-based-component

So I'm thinking... can we extract `onUnload` to the global level, remove its parameter and use `select(...)` in it instead? Haven't tried that approach, just thinking whether something along those lines could work.

JanJakes (commented):
Thanks! It works nicely! I shared some thoughts about the custom hook.

mailpoet/assets/js/src/automation/editor/index.tsx
NeosinneR: If we are sure, that the changes will be lost, I'd be direct about that and not mention just the possibility of losing the data. `There are unsaved changes that will be lost. Do you want to continue?` @brezocordero @JanJakes

brezocordero: I have changed the text although most of the time it will use [the default text from the browser](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#compatibility_notes).

mailpoet/assets/js/src/automation/editor/index.tsx
brezocordero: Done. Thanks for the explanation.

JanJakes (approved):
Works and LGTM, thanks!

PR #4513, opened 2022-11-09
Do not activate MSS unless key is valid for sending [MAILPOET-4767]

johnolek:
## Description

This PR stops the plugin from activating the MSS sending method when a user verifies a key that does not include MSS sending (i.e. creator plans).

It also stops us from making an unnecessary API call attempting to send a congratulatory email in those cases.

## Code review notes

_N/A_

## QA notes

This ticket does not include disabling MSS upon verification or other UI changes we might want to make when a user has a creator plan activated. I will create a new ticket or tickets to track some of these other changes.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4767](https://mailpoet.atlassian.net/browse/MAILPOET-4767)

## After-merge notes

_N/A_


costasovo (approved):
Thanks for adding this improvement!

Aschepikov (approved):
GJ!

PR #4516, opened 2022-11-09
Revert "Use the new api for unsubscribe headers [MAILPOET-4702]"

Aschepikov:
Reverts mailpoet/mailpoet#4475


pavel-mailpoet:
@Aschepikov why was the pull request reverted?

Aschepikov:
@pavel-mailpoet  Hi, I accidentally merged [this pr](https://github.com/mailpoet/mailpoet/pull/4475) which is linked with [this one](https://github.com/mailpoet/mailpoet/pull/4478) They are tested, but I shouldn't merge it without asking 
[@Sam Najian](https://a8c.slack.com/team/U023HUZP875)

With Sam we decided to revert it but one of the tests was failing all over again, so he told that it is ok to leave it in the trunk, that it would be fine as well, as the bridge supports both, but if I want to keep this in the trunk I need to confirm it with the mailpoet-dev or mailpoet-dev-code guys

PR #3813, opened 2021-11-18
Add Checkout block Integration [MAILPOET-3920]

costasovo:
This PR is a continuation of [the original PR](https://github.com/mailpoet/mailpoet/pull/3778) opened by the Rubik team. I've rebased the PR on top of the latest changes and added a bunch of commits with various changes/fixes. I tried to describe the intention of each change in a commit message or code comments.

Note there is a subsequent task for [adding acceptance tests](https://mailpoet.atlassian.net/browse/MAILPOET-3937).

[MAILPOET-3920]

For convenience, I'm adding **the description from the original PR (slightly updated)**:

Introduces a new block called "MailPoet Marketing opt-in" that will be rendered below the email field in the Checkout block.

References:
- [Project Thread](https://woogrowp2.wordpress.com/2021/11/01/project-thread-integrate-automatewoo-and-mailpoet-with-checkout-block/)
- [Epic issue](woocommerce/woocommerce-gutenberg-products-block#5044)

The block is force added to the Checkout Block layout underneath the contact/email field. When opt-in is disabled in MailPoet settings, you'll see a placeholder:

Screenshot 2021-11-03 at 15 38 00

When opt-in is enabled in MailPoet settings, you'll see a preview of the checkbox with the default opt-in message (from the `Checkbox opt-in message` setting):

Screenshot 2021-11-03 at 15 32 31

The message next is editable. Block placement is not. On the frontend you'll see whatever text you input, and the default status of the checkbox is based on your current subscription status, like the main checkout page:

Screenshot 2021-11-03 at 15 42 45

And that's about it. Subscription is handled via the StoreAPI when checkout is submitted.

I struggled a bit with the repo linting rules as it's not using `@wordpress/scripts` or a prettier config to format things automatically. Did I miss a step? I managed to get the linting rules passing manually but I did need to add some ignore rules inline. Let me know if anything needs changing.

## Testing Steps

**Regression testing:**

- With WooCommerce Blocks disabled, Gutenberg disabled, load this PR.
- There should be no errors whatsoever.
- In the editor, you shouldn't be able to insert "MailPoet Marketing Opt-in" block.
- Enable WooCommerce Blocks 6.2.0 or older, there should be no errors.

**Functionality testing:**

- Install [WooCommerce Blocks plugin (v6.3.0 or above)](https://wordpress.org/plugins/woo-gutenberg-products-block/).
[...]

PR #4517, opened 2022-11-09
Prevent Store API validation errors when the 'optin' field has a null value

malithsen:
## Description

This PR is intended to solve an issue between [WooPay](https://woocommerce.com/woopay/) and Mailpoet for WooCommerce so that merchants can use both products together.

When an order gets created via the Store API from WooPay, the 'optin' field expected by Mailpoet will be null. Right now, mailpoet only accepts boolean values, so the validation fails. This results in the order not being created.

Mailpoet already handles null value inputs and correctly sanitizes them as false values. 

More info on this issue can be found here: https://wp.me/pdpOw2-1uq#comment-1687

## Code review notes

_N/A_

## QA notes

This PR has been tested along with WooPay to make sure the WooPay order flow is functional. 
To make there are no regressions in the Marketing Opt-In option, please follow the functional testing instructions in this PR: https://github.com/mailpoet/mailpoet/pull/3813 
## Linked PRs

_N/A_

## Linked tickets

https://github.com/Automattic/woopay/issues/1416

## After-merge notes

_N/A_

### DIFF
--- a/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
+++ b/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
@@ -120,7 +120,7 @@ class WooCommerceBlocksIntegration {
           return [
             'optin' => [
               'description' => __('Subscribe to marketing opt-in.', 'mailpoet'),
-              'type' => 'boolean',
+              'type' => array( 'boolean', 'null'),
             ],
           ];
         },


costasovo:
@malithsen I'm closing this PR in favor of https://github.com/mailpoet/mailpoet/pull/4524, which I did as a workaround for our failing CI build for PRs from forks.

PR #4520, opened 2022-11-10
Shorten test and run action-scheduler hooks

brezocordero:
## Description

This is a small try to make the test faster and see if we get fewer timeouts. 

I have targeted the hooks on the action scheduler we need to run for the test. I don't think there will be many jobs on the scheduler but it can't hurt.

I have also removed a few lines on a test that I considered not essential and reused existing products instead of creating new ones.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4746]

## After-merge notes

_N/A_

[MAILPOET-4746]: https://mailpoet.atlassian.net/browse/MAILPOET-4746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samnajian (approved):
LGTM!

PR #4531, opened 2022-11-14
Add check for mail function and display notice if disabled - MAILPOET-4760

triple0t:
## Description

Check if the `mail` function is available and display a notice if not.

**Note**: We only display the notice if the user is sending with the host method (**PHPMail**) and the `mail` function is disabled

## Code review notes

How to disable `mail` function.

The mail function can be disabled from the php.ini file.

On the MailPoet Dev environment, you may do that here: [dev/php.ini](https://github.com/mailpoet/mailpoet/blob/trunk/dev/php.ini#L5)
By setting `disable_functions = 'mail'`

```
upload_max_filesize = 500M
memory_limit = 1024M
post_max_size = 500M
sendmail_path = /usr/bin/msmtp -C /etc/msmtprc -t
sendmail_from = 'wordpress@mp3.localhost'
disable_functions = 'mail'
```

Note: Remember to rebuild your docker image: `docker-compose build wordpress`

## QA notes

Test on a site with a restricted `mail` function (e.g., Pressable) and others without restriction (e.g., Jurassic ninja)

* Activate the plugin, and check MailPoet pages.
* Switch the sending method from settings (/wp-admin/admin.php?page=mailpoet-settings#/mta/other)
* Select SMTP, Click on Activate button
* Reload the page. 
* Confirm Notice is not displayed.
* Switch the sending method back to "Your web host/web server", Click on the Activate button
* Reload the page. 

## Linked tickets

[MAILPOET-4760](https://mailpoet.atlassian.net/browse/MAILPOET-4760)

## After-merge notes

_N/A_

### DIFF
--- a/mailpoet/lib/Config/Activator.php
+++ b/mailpoet/lib/Config/Activator.php
@@ -134,7 +134,7 @@ class Activator {
   private function checkForDisabledMailFunction() {
     $sendingMethodSet = $this->settings->get('mta.method', false);
 
-    if ($sendingMethodSet === 'PHPMail' || is_bool($sendingMethodSet)) {
+    if ($sendingMethodSet === 'PHPMail') {
       // check for valid mail function
       $this->settings->set(DisabledMailFunctionNotice::QUEUE_DISABLED_MAIL_FUNCTION_CHECK, true);
     }
--- a/mailpoet/lib/Config/Activator.php
+++ b/mailpoet/lib/Config/Activator.php
@@ -7,6 +7,7 @@ use MailPoet\Cron\CronTrigger;
 use MailPoet\InvalidStateException;
 use MailPoet\Migrator\Migrator;
 use MailPoet\Settings\SettingsController;
+use MailPoet\Util\Notices\DisabledMailFunctionNotice;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Doctrine\DBAL\Connection;
 
@@ -81,6 +82,8 @@ class Activator {
 
     $localizer = new Localizer();
     $localizer->forceInstallLanguagePacks($this->wp);
+
+    $this->checkForDisabledMailFunction();
   }
 
   public function deactivate() {
@@ -125,6 +128,15 @@ class Activator {
         'date' => date('Y-m-d H:i:s'),
       ];
       $this->settings->set('updates_log', $updatesLog);
+    }
+  }
+
+  private function checkForDisabledMailFunction() {
+    $sendingMethodSet = $this->settings->get('mta.method', false);
+
+    if ($sendingMethodSet === 'PHPMail' || is_bool($sendingMethodSet)) {
+      // check for valid mail function
+      $this->settings->set(DisabledMailFunctionNotice::QUEUE_DISABLED_MAIL_FUNCTION_CHECK, true);
     }
   }
 
--- a/mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
+++ b/mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Mailer\Mailer;
+use MailPoet\Settings\SettingsController;
+use MailPoet\Util\Helpers;
+use MailPoet\Util\License\Features\Subscribers as SubscribersFeature;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class DisabledMailFunctionNotice {
+
+  const OPTION_NAME = 'disabled-mail-function';
+
+  /** @var SettingsController */
+  private $settings;
+
+  /** @var WPFunctions */
+  private $wp;
+
+  /** @var SubscribersFeature */
+  private $subscribersFeature;
+
+  public function __construct(
+    WPFunctions $wp,
+    SettingsController $settings,
+    SubscribersFeature $subscribersFeature
+  ) {
+    $this->settings = $settings;
+    $this->wp = $wp;
+    $this->subscribersFeature = $subscribersFeature;
+  }
+
+  public function init($shouldDisplay) {
+    $shouldDisplay = $shouldDisplay && $this->checkRequirements();
+    if ($shouldDisplay) {
+      $this->display();
+    }
+  }
+
+  private function checkRequirements(): bool {
+    $sendingMethod = $this->settings->get('mta.method', false);
+    $isPhpMailSendingMethod = $sendingMethod === Mailer::METHOD_PHPMAIL;
+
+    $functionName = 'mail';
+    $isMailFunctionDisabled = $this->isFunctionDisabled($functionName);
+
+    return $isPhpMailSendingMethod && $isMailFunctionDisabled;
+  }
+
+  private function isFunctionDisabled(string $function): bool {
+    $result = function_exists($function) && is_callable($function, false);
+    return !$result;
+  }
+
+  private function display() {
+    $header = $this->getHeader();
+
+    $body = $this->getBody();
+
+    $button = $this->getConnectMailPoetButton();
+
+    $message = $header . $body . $button;
+
+     Notice::displayWarning($message, '', self::OPTION_NAME, false);
+  }
+
+  private function getHeader(): string {
+    return '<h4>' . __('Get ready to send your first campaign.', 'mailpoet') . '</h4>';
+  }
+
+  private function getBody(): string {
+    $bodyText = __('Connect your website with MailPoet, and start sending for free. Reach inboxes, not spam boxes. [link]Why am I seeing this?[/link]', 'mailpoet');
+
+    $bodyWithReplacedLink = Helpers::replaceLinkTags($bodyText, 'https://kb.mailpoet.com/article/396-disabled-mail-function', [
+      'target' => '_blank',
+    ]);
+
+    return '<p>' . $bodyWithReplacedLink . '</p>';
+  }
+
+  private function getConnectMailPoetButton(): string {
+    $subscribersCount = $this->subscribersFeature->getSubscribersCount();
+    $buttonLink = "https://account.mailpoet.com/?s={$subscribersCount}&utm_source=mailpoet&utm_medium=plugin&utm_campaign=disabled_mail_function";
+    $link = $this->wp->escAttr($buttonLink);
+    return '<p><a target="_blank" href="' . $link . '" class="button button-primary">' . __('Connect MailPoet', 'mailpoet') . '</a></p>';
+  }
+}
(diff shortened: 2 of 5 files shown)


mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
```diff
@@ -0,0 +1,93 @@
...
+    $this->subscribersFeature = $subscribersFeature;
+  }
+
+  public function init($shouldDisplay): ?string {
+    $shouldDisplay = $shouldDisplay && $this->checkRequirements();
+    if (!$shouldDisplay) {
+      return null;
+    }
+
+    return $this->display();
```
rodrigoprimo, line 51: Just a nitpick but I noticed that we are returning something in this method but we don't do anything with what is returned when init() is called. Do we need to return something?

rodrigoprimo (dismissed):
Looks good to me. I left one minor comment that is just a detail and probably doesn't need to be addressed. I will leave it to you to decide whether you want to address it or move this PR to QA.

I also noticed that the KB page still doesn't exist, but I'm assuming the plan is to create it before this PR is released.

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
triple0t: We are not using the returned data in the class.
It is being used [here](https://github.com/mailpoet/mailpoet/blob/b63e19b7d260957013a37a501de2e709e2cb480c/mailpoet/tests/integration/Util/Notices/DisabledMailFunctionNoticeTest.php#L54-L61) for testing. i.e confirming we get the expected notice message

triple0t:
Thank you for the review, @rodrigoprimo 

> I also noticed that the KB page still doesn't exist, but I'm assuming the plan is to create it before this PR is released.

Yes, the KB page would be created before the PR is released.

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
rodrigoprimo: Ah, I missed that it is being used for testing. I'm not a huge fan of class behavior that is only used in the tests, but I see that this was already used in the other notice classes and I don't think we need to address this here. I will move this PR to QA.

mailpoet/lib/Config/Activator.php
```diff
@@ -128,6 +131,15 @@ public function updateDbVersion() {
     }
   }
 
+  private function checkForDisabledMailFunction() {
+    $sendingMethodSet = $this->settings->get('mta.method', false);
+
+    if ($sendingMethodSet === 'PHPMail' || is_bool($sendingMethodSet)) {
```
rodrigoprimo: It is not clear to me why we want to set `DisabledMailFunctionNotice::QUEUE_DISABLED_MAIL_FUNCTION_CHECK` to true if `$sendingMethodSet` is bool. Could you please explain? If it is not something obvious that I'm missing, I wonder if maybe it is worth adding a inline comment with an explanation?

mailpoet/lib/Config/Activator.php
```diff
@@ -81,6 +82,8 @@ private function processActivate(): void {
 
     $localizer = new Localizer();
     $localizer->forceInstallLanguagePacks($this->wp);
+
+    $this->checkForDisabledMailFunction();
```
rodrigoprimo, line 86: Is checking this here on the plugin activation enough? What happens if the host changes the server configuration and `mail()` starts to malfunction or the site is migrated to a different host with different `mail()` configuration? Maybe we don't want to cover those cases because the cons (the complexity in the implementation) outweigh the pros?

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
```diff
@@ -90,4 +130,87 @@ private function getConnectMailPoetButton(): string {
...
+    } catch (\Exception $e) {
+      $errorMessage = $e->getMessage();
+      return !$this->checkForErrorMessage($errorMessage);
+    }
+
+    return true;
+  }
+
+  private function checkForErrorMessage($errorMessage): bool {
+    $phpmailerError = 'Could not instantiate mail function';
```
rodrigoprimo, line 211: Is there any chance that in some servers the misconfigured `mail()` function might return a different error message? Is this possible that this message is returned in a different language? Do we want to display the notice if the `mail()` function returns any other error?

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
```diff
@@ -42,13 +52,43 @@ public function init($shouldDisplay): ?string {
...
+   * Check MisConfigured Function
+   *
+   * This method will cause this class to only display the notice if the settings option
+   *
+   * disabled_mail_function_check === true
+   * or
+   * queue_disabled_mail_function_check === true
+   *
+   */
+  public function checkMisConfiguredFunction(): bool {
```
rodrigoprimo: The name of this method seems misleading to me. When I read it, I thought this method would perform the check for the misconfigured function instead of checking if the code needs to check for the misconfigured function. Maybe we can find a more descriptive name? Maybe `shouldCheckMisconfiguredFunction()` or something along those lines?

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
```diff
@@ -0,0 +1,216 @@
...
+use MailPoet\Mailer\MailerFactory;
+use MailPoet\Settings\SettingsController;
+use MailPoet\Util\Helpers;
+use MailPoet\Util\License\Features\Subscribers as SubscribersFeature;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class DisabledMailFunctionNotice {
+
+  const OPTION_NAME = 'disabled_mail_function_check';
```
rodrigoprimo: Maybe now that we have two different settings it would be better to name this constant something more specific than `OPTION_NAME`?

rodrigoprimo (commented):
Good job, @triple0t. I left some comments for us to discuss a few points. Thanks.

mailpoet/lib/Config/Activator.php
triple0t: >  It is not clear to me why we want to set DisabledMailFunctionNotice::QUEUE_DISABLED_MAIL_FUNCTION_CHECK to true if $sendingMethodSet is bool

By default, we set the sending method to PHPMail, or the user may change it to something else.

Ideally, there should always be a value when we call that method.
If there’s ever a case where the sending method is not set (i.e. is bool), then check if the mail() function is enabled.

mailpoet/lib/Config/Activator.php
triple0t: We also make the check when the user switches the sending method. [Link here](https://github.com/mailpoet/mailpoet/pull/4531/files#diff-399dff53352c832df3eb84219a2e7faf14176cfc04a8613825a9ee7810829df7R366-R378)

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
triple0t: > Is this possible that this message is returned in a different language

Possible. The [error message](https://github.com/WordPress/WordPress/blob/master/wp-includes/PHPMailer/PHPMailer.php#L1951-L1953) is defined in the PHPMailer library [link here](https://github.com/WordPress/WordPress/blob/master/wp-includes/PHPMailer/PHPMailer.php#L2281). I suppose the error message can be translated. But, it appears WP defaulted to en. (Which is the default from the Library: [1](https://github.com/WordPress/WordPress/blob/master/wp-includes/PHPMailer/PHPMailer.php#L4135), [2](https://github.com/WordPress/WordPress/blob/master/wp-includes/PHPMailer/PHPMailer.php#L2375))

I tested with a WordPress instance in another language (Fr), and the error message returned in English.

>  Is there any chance that in some servers the misconfigured mail() function might return a different error message? Do we want to display the notice if the mail() function returns any other error?

Nope, we do not. There might be other issues or a temporary error.

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
triple0t: Noted.

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
triple0t: Noted.

triple0t:
Hi @rodrigoprimo, The PR has been updated. Thank you for the review.

mailpoet/lib/Config/Activator.php
rodrigoprimo: Thanks for the explanation

mailpoet/lib/Config/Activator.php
rodrigoprimo: But I guess checking when the user switches the sending method won't cover the cases mentioned above, right? Or am I missing something?

I'm inclined to think the complexity of covering them doesn't justify the benefits, but I just wanted to make sure this was considered and that is why I asked.

mailpoet/lib/Util/Notices/DisabledMailFunctionNotice.php
rodrigoprimo: > I tested with a WordPress instance in another language (Fr), and the error message returned in English.

Thanks for checking. Good to know that it is always in English.

rodrigoprimo (approved):
@triple0t, sorry for the delay to respond. This PR looks good to me. There is still one comment that I left unresolved, but it is just a detail and I'm inclined to think we don't need to address it. I just want to make sure it is a conscious decision. Feel free to move this PR to QA or address the cases that I mentioned in the unresolved comment.

mailpoet/lib/Config/Activator.php
triple0t: @rodrigoprimo 

> But I guess checking when the user switches the sending method won't cover the cases mentioned above, right? Or am I missing something?

It should. 
Switching the sending method would trigger the check for valid mail function, which also checks the mail function is appropriately configured.

There are currently two trigger methods:
 1. When the plugin is activated 
 2. When the sending method is updated, and PHPMail is selected.
 
Please note: We only check for valid mail() function when the PHPMail sending method is active.

> What happens if the host changes the server configuration (while PHPMail is active) and mail() starts to malfunction, or the site is migrated to a different host with different mail() configuration?

Difficult to catch this and trigger the check without a lot of extra overhead work. As you pointed out, the complexity would be quite high and probably not something we should do in this ticket.

But toggling the sending method (between another sending method and PHPMail) should trigger the check.

rodrigoprimo: > Difficult to catch this and trigger the check without a lot of extra overhead work. As you pointed out, the complexity would be quite high and probably not something we should do in this ticket.

Yeah, that is what I was referring to when I said that switching the sending method wouldn't cover the cases that I mentioned above. Anyway, we are in agreement that we should not cover this case. Thanks for confirming this. Moving this PR to QA.

Aschepikov (approved):
GJ!

PR #4537, opened 2022-11-16
Add command for rerunning the latest project workflow on CircleCI [PREMIUM-207]

lysyjan:
## Description

When we release premium and free plugins together, our first builds for the premium plugin on Circle CI fail. This PR adds a command that reruns the latest Circle CI workflow for a specific project. The new command is also used in the prepare command.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[PREMIUM-207](https://mailpoet.atlassian.net/browse/PREMIUM-207)

## After-merge notes

_N/A_

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -1116,21 +1116,15 @@ class RoboFile extends \Robo\Tasks {
     $this->say("Release '$version[name]' info was published on Slack.");
   }
 
-  public function releaseRerunCircleWorkflow(string $project) {
-    $project = strtoupper($project);
-    $supportedProjects = [
-      \MailPoetTasks\Release\CircleCiController::PROJECT_PREMIUM,
-      \MailPoetTasks\Release\CircleCiController::PROJECT_MAILPOET,
-    ];
-    if (!in_array($project, $supportedProjects, true)) {
-      throw new \Exception('Unsupported project');
-    }
+  public function releaseRerunCircleWorkflow(string $project = null) {
     $circleciController = $this->createCircleCiController();
     $result = $circleciController->rerunLatestWorkflow($project);
+    // Sometimes can be useful to know which Circle project workflow was restarted
+    $project = $project ? " for the project '{$project}'" : '';
     if (!$result) {
-      $this->yell("Circle Workflow for the project '{$project}' was not restarted", 40, 'red');
+      $this->yell("Circle Workflow{$project} was not restarted", 40, 'red');
     }
-    $this->say("Circle Workflow for the project '{$project}' was started from the beginning");
+    $this->say("Circle Workflow{$project} was started from the beginning");
   }
 
   public function downloadWooCommerceBlocksZip($tag = null) {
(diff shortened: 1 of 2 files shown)


mailpoet/RoboFile.php
```diff
@@ -1116,6 +1119,17 @@ public function releasePublishSlack($version = null) {
...
 
+  public function releaseRerunCircleWorkflow(string $project = null) {
+    $circleciController = $this->createCircleCiController();
+    $result = $circleciController->rerunLatestWorkflow($project);
+    // Sometimes can be useful to know which Circle project workflow was restarted
+    $project = $project ? " for the project '{$project}'" : '';
+    if (!$result) {
+      $this->yell("Circle Workflow{$project} was not restarted", 40, 'red');
+    }
+    $this->say("Circle Workflow{$project} was started from the beginning");
```
rodrigoprimo: I guess we need to put this line inside an else condition or finish the execution of this method when yell() is called to avoid having both the error and the success message being printed when there is an error?

rodrigoprimo (commented):
Overall the code changes look good to me, @lysyjan. I left a comment with a small detail that I believe we should fix before merging this. Besides that, I wonder if there is a way to test the changes in this PR before I approve it? IMO, it would be good if we could test this before using it in the next release next week.

mailpoet/RoboFile.php
lysyjan: Good catch! I fixed it in the latest commit.

lysyjan:
@rodrigoprimo You can test the new command by following these steps:
1. Change the branch name on [this line](https://github.com/mailpoet/mailpoet/pull/4537/files#diff-7fd98d4fe52b90d96a196f1ff8179ad830d024eb35e06f3e14b18e51a1116b53R171)
2. Execute command `./do release:rerun-circle-workflow`.

I have noticed now that I haven't added the command to the premium repository. At this moment, we don't need it, but it might be nice to have the release commands consistent. What do you think?

mailpoet/RoboFile.php
rodrigoprimo: Thanks

rodrigoprimo:
> You can test the new command by following these steps

Thanks for the testing instructions. I was able to force CircleCI to restart the build for a test branch called `delme`. The commands seem to be working as expected.

> I have noticed now that I haven't added the command to the premium repository. At this moment, we don't need it, but it might be nice to have the release commands consistent. What do you think?

I don't have a strong opinion here, but I'm inclined to think that if we don't need it at this moment, we shouldn't include it. We can always include it later if needed. I'm also unsure which build job we would want to rerun when running `./do release:prepare` for the premium plugin. Let me know if I'm missing something. I will leave this PR as approved. Feel free to move it to QA if you think that is needed (probably not), merge it, or continue the conversation about adding the command to the premium repository.

rodrigoprimo (approved):

lysyjan:
@rodrigoprimo Thank you. I'm okay with the current state, and I will merge it.

PR #4541, opened 2022-11-18
Prevent errors due to wrapped WC function [MAILPOET-4834]

johnolek:
## Description

This PR ensures that our WC helper's wrapped version of `wc_get_page_id` will never throw an error due to an incorrect return value. See the linked ticket for an example of how this was affecting a user.

## Code review notes

_N/A_

## QA notes

This shouldn't have any effect on functionality, but it might still be best to test the functionality from MAILPOET-4663, which introduced this code.

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/4459 introduced some of this code

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-4834

## After-merge notes

_N/A_

### DIFF
--- a/mailpoet/lib/Form/DisplayFormInWPContent.php
+++ b/mailpoet/lib/Form/DisplayFormInWPContent.php
@@ -124,7 +124,7 @@ class DisplayFormInWPContent {
     $displayCheck = $this->wp->applyFilters('mailpoet_display_form_in_product_listing', true);
 
     $shopPageId = $this->woocommerceHelper->wcGetPageId('shop');
-    $this->wooShopPageId = $shopPageId && $shopPageId >= 0 ? $shopPageId : null;
+    $this->wooShopPageId = $shopPageId && $shopPageId > 0 ? $shopPageId : null;
 
     if ($displayCheck && !is_null($this->wooShopPageId) && $this->wp->isPage($this->wooShopPageId)) {
       return true;
--- a/mailpoet/lib/WooCommerce/Helper.php
+++ b/mailpoet/lib/WooCommerce/Helper.php
@@ -63,8 +63,8 @@ class Helper {
   }
 
   public function wcGetPageId(string $page): ?int {
-    if (function_exists('wc_get_page_id')) {
-      return wc_get_page_id($page);
+    if ($this->isWooCommerceActive()) {
+      return (int)wc_get_page_id($page);
     }
     return null;
   }


mailpoet/lib/Form/DisplayFormInWPContent.php
```diff
@@ -123,8 +123,8 @@ private function shouldDisplay(): bool {
   private function displayFormInProductListPage(): bool {
     $displayCheck = $this->wp->applyFilters('mailpoet_display_form_in_product_listing', true);
 
-    $shopPageId = $this->woocommerceHelper->wcGetPageId('shop');
-    $this->wooShopPageId = $shopPageId && $shopPageId >= 0 ? $shopPageId : null;
+    $shopPageId = (int)$this->woocommerceHelper->wcGetPageId('shop');
+    $this->wooShopPageId = $shopPageId && $shopPageId > 0 ? $shopPageId : null;
```
johnolek, line 127: I wasn't sure why we were allowing values of `0` previously. I don't think WordPress posts can have post IDs of `0`, so I made this slight change just to ensure there's no possibility of setting `wooShopPageId` to `0`. Please let me know if I shouldn't have touched this logic.

johnolek:
@triple0t I would love your input on this one since you worked on https://github.com/mailpoet/mailpoet/pull/4459

mailpoet/lib/Form/DisplayFormInWPContent.php
triple0t: that's okay

mailpoet/lib/WooCommerce/Helper.php
```diff
@@ -62,7 +62,7 @@ public function wcGetProduct($theProduct = false) {
     return wc_get_product($theProduct);
   }
 
-  public function wcGetPageId(string $page): ?int {
+  public function wcGetPageId(string $page) {
     if (function_exists('wc_get_page_id')) {
       return wc_get_page_id($page);
     }
```
triple0t: > To prevent this from being a problem, we could either remove the ?int return type hint or always coerce the value returned from wc_get_page_id to be an integer.

May I ask why you decided not to go with the second option (always coerce the value returned from wc_get_page_id to be an integer)?

triple0t (dismissed):
Looks good. Well done.

I'm approving this PR, but I have a little question.

Kindly locate it below.

mailpoet/lib/WooCommerce/Helper.php
johnolek: Of course! Honestly I was torn between the two options. Removing the type hint felt a tiny bit more appropriate to me because if we do find ourselves in an edge case where `wc_get_page_id` is defined by something other than WooCommerce and returns something other than an integer, coercing the value would mean our wrapper would return a different type than the wrapped function, which just feels inconsistent. Not having the return type hint makes it clearer to me that we can't actually be 100% sure what `wc_get_page_id` will return.

Would your preference be for keeping the type hint and coercing the value in the wrapper?

triple0t: > we do find ourselves in an edge case where wc_get_page_id is defined by something other than WooCommerce and returns something other than an integer

We should not return either string or integer if WooCommerce is not installed.  
The function is meant to get a WooCommerce page ID by name. If WooCommerce is not installed and activated, we should return null.

That would prevent some weird bugs down the road.

If wc_get_page_id was defined by something other than WooCommerce itself, there is a high chance the behavior would be different.



> Would your preference be for keeping the type hint and coercing the value in the wrapper?

Yes. 

We won’t have to coerce the function return value everywhere else it's been used. e.g. like you did [here](https://github.com/mailpoet/mailpoet/blob/20f79c45cf4f8ccf65b0db94ddfd15f2f51b9a01/mailpoet/lib/Form/DisplayFormInWPContent.php#L126)

johnolek: @triple0t I ended up going a slightly different route. I replaced the check to see if the function `wc_get_page_id` is defined to instead check if WC is installed and active, which should ensure that the version of `wc_get_page_id` that exists actually comes from WC. I left in the return type hint, and I also included a type cast just in case something changes in WC's implementation in the future. Does this seem reasonable to you?

One thing I struggled with while thinking about this is consistency. Many of our other wrapped WC functions (`getWoocommerceCurrency`, `get_woocommerce_currency_symbol`, `wcLightOrDark` etc) don't include return type hints, and they don't even include checks for the existence of the function or if WC is active. Instead it looks like we either use these functions in contexts where we know WC must be active (like WC hooks), or we include checks to `isWooCommerceActive` in the application code before we call the helper. 

Part of me thinks the existing convention makes a little more sense. Why call these methods in the first place unless we know WC is installed/active? If we return `null` if WC isn't active that's just one more thing we might have to check for depending on what the original signature looked like.

In any case, I'm probably overthinking this and spending too much time writing this comment, but I do think it's worth thinking about what convention makes the most sense for these wrapped functions in general.

triple0t (approved):
LGTM

mailpoet/lib/WooCommerce/Helper.php
triple0t: > @triple0t I ended up going a slightly different route. I replaced the check to see if the function wc_get_page_id is defined to instead check if WC is installed and active, which should ensure that the version of wc_get_page_id that exists actually comes from WC. I left in the return type hint, and I also included a type cast just in case something changes in WC's implementation in the future. Does this seem reasonable to you?

Yup. It does.

> One thing I struggled with while thinking about this is consistency. Many of our other wrapped WC functions (getWoocommerceCurrency, get_woocommerce_currency_symbol, wcLightOrDark etc) don't include return type hints, and they don't even include checks for the existence of the function or if WC is active. Instead it looks like we either use these functions in contexts where we know WC must be active (like WC hooks), or we include checks to isWooCommerceActive in the application code before we call the helper.
>
> Part of me thinks the existing convention makes a little more sense. Why call these methods in the first place unless we know WC is installed/active? If we return null if WC isn't active that's just one more thing we might have to check for depending on what the original signature looked like.
>
> In any case, I'm probably overthinking this and spending too much time writing this comment, but I do think it's worth thinking about what convention makes the most sense for these wrapped functions in general.

I definitely agree with you.

Aschepikov (approved):
GJ!

PR #4546, opened 2022-11-21
Add empty homepage [MAILPOET-4824]

costasovo:
## Description

This PR adds a feature switch and an empty homepage.

Screenshot 2022-11-21 at 11 18 12

Apart from these changes, I updated some links (e.g. in logo or redirect after reinstalling, redirect after finishing wizard) to go to the Homepage instead of the Emails page

## Code review notes

I updated the final ticket for removing the feature flag to include also a refactoring of the `public static $mainPageSlug` back to constant. We will no longer need to dynamically its value.

## QA notes

To activate the homepage go to /wp-admin/admin.php?page=mailpoet-experimental and activate the feature called `homepage`

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4824]

## After-merge notes

_N/A_

[MAILPOET-4824]: https://mailpoet.atlassian.net/browse/MAILPOET-4824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/views/homepage.html
+++ b/mailpoet/views/homepage.html
@@ -1,0 +1,5 @@
+<% extends 'layout.html' %>
+
+<% block content %>
+<h1>MailPoet Homepage</h1>
+<% endblock %>
--- a/mailpoet/assets/js/src/global.d.ts
+++ b/mailpoet/assets/js/src/global.d.ts
@@ -142,4 +142,5 @@ interface Window {
     name: string;
   }[];
   mailpoet_cdn_url: string;
+  mailpoet_main_page_slug: string;
 }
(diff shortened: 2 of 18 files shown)


mailpoet/views/homepage.html
```diff
@@ -0,0 +1,5 @@
+<% extends 'layout.html' %>
+
+<% block content %>
+<h1>MailPoet Homepage</h1>
```
samnajian, line 4: Not sure if we leave this untranslated at this stage!

samnajian (commented):
Looking to me other than the added comment about untranslated string.

mailpoet/views/homepage.html
costasovo: This is just placeholder content. The whole homepage is hidden behind the feature flag. It is to be removed once we start adding real content. I decided to keep it there so that it is easier to verify that the homepage template loads.
I can remove it, but I wouldn't make it translatable since it will be removed soon. Do you think I should rather remove it?

samnajian: No need to change it if it's only a placeholder.

samnajian (approved):
LGTM!

Aschepikov (approved):
GJ!

PR #4550, opened 2022-11-23
Add plugins version to logs

pavel-mailpoet:
## Linked tickets

[MAILPOET-4735]

[MAILPOET-4735]: https://mailpoet.atlassian.net/browse/MAILPOET-4735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Logging/LoggerFactory.php
+++ b/mailpoet/lib/Logging/LoggerFactory.php
@@ -65,15 +65,15 @@ class LoggerFactory {
 
   /**
    * @param string $name
-   * @param bool $attachProcessors
+   * @param bool $attachOptionalProcessors
    *
    * @return \MailPoetVendor\Monolog\Logger
    */
-  public function getLogger($name = 'MailPoet', $attachProcessors = WP_DEBUG) {
+  public function getLogger($name = 'MailPoet', $attachOptionalProcessors = WP_DEBUG) {
     if (!isset($this->loggerInstances[$name])) {
       $this->loggerInstances[$name] = new \MailPoetVendor\Monolog\Logger($name);
 
-      if ($attachProcessors) {
+      if ($attachOptionalProcessors) {
         // Adds the line/file/class/method from which the log call originated
         $this->loggerInstances[$name]->pushProcessor(new IntrospectionProcessor());
         // Adds the current request URI, request method and client IP to a log record
--- a/mailpoet/lib/Logging/LoggerFactory.php
+++ b/mailpoet/lib/Logging/LoggerFactory.php
@@ -82,6 +82,9 @@ class LoggerFactory {
         $this->loggerInstances[$name]->pushProcessor(new MemoryUsageProcessor());
       }
 
+      // Adds the plugin's versions to the log, we always want to see this
+      $this->loggerInstances[$name]->pushProcessor(new PluginVersionProcessor());
+
       $this->loggerInstances[$name]->pushHandler(new LogHandler(
         $this->logRepository,
         $this->entityManager,
--- a/mailpoet/lib/Logging/PluginVersionProcessor.php
+++ b/mailpoet/lib/Logging/PluginVersionProcessor.php
@@ -1,10 +1,11 @@
-<?php declare (strict_types=1);
+<?php declare (strict_types = 1);
 
 namespace MailPoet\Logging;
 
 use MailPoet\Config\Env;
+use MailPoetVendor\Monolog\Processor\ProcessorInterface;
 
-class PluginVersionProcessor {
+class PluginVersionProcessor implements ProcessorInterface {
   public function __invoke(array $record): array {
     $record['extra']['free_plugin_version'] = Env::$version;
     $record['extra']['premium_plugin_version'] = defined('MAILPOET_PREMIUM_VERSION') ? MAILPOET_PREMIUM_VERSION : 'premium not installed';
--- a/mailpoet/lib/Logging/PluginVersionProcessor.php
+++ b/mailpoet/lib/Logging/PluginVersionProcessor.php
@@ -1,0 +1,13 @@
+<?php declare (strict_types=1);
+
+namespace MailPoet\Logging;
+
+use MailPoet\Config\Env;
+
+class PluginVersionProcessor {
+  public function __invoke(array $record): array {
+    $record['extra']['free_plugin_version'] = Env::$version;
+    $record['extra']['premium_plugin_version'] = defined('MAILPOET_PREMIUM_VERSION') ? MAILPOET_PREMIUM_VERSION : 'premium not installed';
+    return $record;
+  }
+}
--- a/mailpoet/tests/unit/Logging/LoggerFactoryTest.php
+++ b/mailpoet/tests/unit/Logging/LoggerFactoryTest.php
@@ -39,13 +39,13 @@ class LoggerFactoryTest extends \MailPoetUnitTest {
   public function testItAttachesProcessors() {
     $logger1 = $this->loggerFactory->getLogger('logger-with-processors', true);
     $processors = $logger1->getProcessors();
-    expect($processors)->notEmpty();
+    expect(count($processors))->greaterThan(1);
   }
 
   public function testItDoesNotAttachProcessors() {
     $logger1 = $this->loggerFactory->getLogger('logger-without-processors', false);
     $processors = $logger1->getProcessors();
-    expect($processors)->isEmpty();
+    expect($processors)->count(1);
   }
 
   public function testItAttachesHandler() {
--- a/mailpoet/tests/unit/Logging/LoggerFactoryTest.php
+++ b/mailpoet/tests/unit/Logging/LoggerFactoryTest.php
@@ -42,7 +42,7 @@ class LoggerFactoryTest extends \MailPoetUnitTest {
     expect(count($processors))->greaterThan(1);
   }
 
-  public function testItDoesNotAttachProcessors() {
+  public function testItSkipsOptionalProcessors() {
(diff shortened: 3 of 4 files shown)


mailpoet/lib/Logging/LoggerFactory.php
```diff
@@ -82,6 +82,9 @@ public function getLogger($name = 'MailPoet', $attachProcessors = WP_DEBUG) {
         $this->loggerInstances[$name]->pushProcessor(new MemoryUsageProcessor());
       }
 
+      // Adds the plugin's versions to the log, we always want to see this
+      $this->loggerInstances[$name]->pushProcessor(new PluginVersionProcessor());
```
rodrigoprimo, line 86: One thing that it is not clear to me is why we don't want to always include the three processors that already existed and why we want to always have the `PluginVersionProcessor`? Asking just in case there is a scenario where the code expects to run without any processors. I'm not sure if that is the case or not.

Also, if now we always include `PluginVersionProcessor` maybe we should change the name of the variable `$attachProcessors` to `$attachOptionalProcessors` or something like that?

mailpoet/tests/unit/Logging/LoggerFactoryTest.php
```diff
@@ -39,13 +39,13 @@ public function testItReturnsInstance() {
   public function testItAttachesProcessors() {
     $logger1 = $this->loggerFactory->getLogger('logger-with-processors', true);
     $processors = $logger1->getProcessors();
-    expect($processors)->notEmpty();
+    expect(count($processors))->greaterThan(1);
   }
 
   public function testItDoesNotAttachProcessors() {
     $logger1 = $this->loggerFactory->getLogger('logger-without-processors', false);
```
rodrigoprimo, line 46: Similar to a previous comment, I believe we should change the name of this logger and the name of this test to reflect that this logger actually includes one processor, but doesn't include "optional" processors. What do you think?

mailpoet/lib/Logging/PluginVersionProcessor.php
```diff
@@ -0,0 +1,13 @@
+<?php declare (strict_types=1);
+
+namespace MailPoet\Logging;
+
+use MailPoet\Config\Env;
+
+class PluginVersionProcessor {
```
rodrigoprimo: Any reason why this processor doesn't extend `ProcessorInterface` as the other processors?

rodrigoprimo (commented):
Overall the code changes look code to me, and I could confirm that the version of both plugins is included in the entry that is added to wp_mailpoet_log. I left a few comments with some details for us to discuss.

mailpoet/lib/Logging/PluginVersionProcessor.php
pavel-mailpoet: There are two reasons: 
- it is not mandatory to implement the interface, the method `pushProcessor` only requires a `callable`
- I tried that, but PHPStan complained about the `Record`:
```
  15     Parameter $record of method
         MailPoet\Logging\PluginVersionProcessor::__invoke() has invalid type
         MailPoetVendor\Monolog\Processor\Record.
  15     Return type (array) of method
         MailPoet\Logging\PluginVersionProcessor::__invoke() should be
         compatible with return type (MailPoetVendor\Monolog\Processor\Record)
         of method
         MailPoetVendor\Monolog\Processor\ProcessorInterface::__invoke()
```
I tried to fix it for some time but was unable. Rosta tried to help me but even he couldn't figure it out. So I removed the `ProcessorInterface` from this class.

mailpoet/lib/Logging/LoggerFactory.php
pavel-mailpoet: Good question. I don't know. Originally, this was to save some data and only keep what is important in the log. 

For the `PluginVersion` I think it is useful to have it always not only when `WP_DEBUG` is enabled because we want to know when did the customer produce each log. But maybe it is not worth differentiating and we could include always everything. What do you think?

mailpoet/lib/Logging/PluginVersionProcessor.php
rodrigoprimo: Thanks for the additional information, @pavel-mailpoet.

You probably saw that already, but I guess the errors are happening because of those two lines that use PHPStan specific documentation:

https://github.com/Seldaek/monolog/blob/2.4.0/src/Monolog/Processor/ProcessorInterface.php#L26-L27

On a quick look to me, it seems that the values of @phpstan-param and @phpstan-return are simply wrong, as the method signature clearly indicates that it receives an array and not Record. Have you considered implementing the interface and ignoring the PHPStan errors? I slightly prefer this approach as it would mean it would be easier to catch changes to the interface in the future (errors would happen on compile time instead of runtime). For example, this interface already changed in the latest version of Monolog:

https://github.com/Seldaek/monolog/blob/main/src/Monolog/Processor/ProcessorInterface.php#L26

What are your thoughts?

mailpoet/lib/Logging/LoggerFactory.php
rodrigoprimo: I'm not sure. I'm inclined to think we should not change that unless we have a good reason to (for example, support tells us that always having this information available would be helpful) and we check that it won't add a significant overhead.

pavel-mailpoet: This information is not for supporters. This is for L2. We already had this problem when a customer claimed there is an error but we fixed it already in the latest version. So we weren't sure if the problem was fixed or if the customer did not update when the error occurred for them. 

If we only include the plugin version when `WP_DEBUG` is true, we wouldn't know what happens.

mailpoet/lib/Logging/PluginVersionProcessor.php
pavel-mailpoet: > You probably saw that already, but I guess the errors are happening because of those two lines that use PHPStan specific documentation:

correct

> On a quick look to me, it seems that the values of @phpstan-param and @phpstan-return are simply wrong, as the method signature clearly indicates that it receives an array and not Record. Have you considered implementing the interface and ignoring the PHPStan errors? I slightly prefer this approach as it would mean it would be easier to catch changes to the interface in the future (errors would happen on compile time instead of runtime). For example, this interface already changed in the latest version of Monolog:

I am not a big fan of ignoring errors. It adds unnecessary stuff to the source code and can hide real errors. 

I am still not sure why should we implement the interface. The `pushProcessor` does not require it. It accepts anything `callable`. So I am not worried if the interface changes.

mailpoet/lib/Logging/LoggerFactory.php
rodrigoprimo: Sorry, my previous comment was not clear. I was not referring to putting the plugin processor behind the WP_DEBUG flag. I was referring to keeping the other processors behind the WP_DEBUG flag or not. That is the part that I'm not sure if we should change or not due to the reasons that I shared above.

mailpoet/lib/Logging/PluginVersionProcessor.php
rodrigoprimo: Correct me if I'm wrong, but although the interface is not explicitly required when `pushProcessor()` is called, it seems to me it is implicitly required in other places. For example, when the processor is invoked the type of the parameter that is passed to it matters and without using the interface we won't know at compile time if the type changes and we might run into runtime errors. Considering this, I'm inclined to think that it is best to use the interface and ignore the specific PHPStan errors than to not use it. But it is a tradeoff and I understand your point about not being a fan of ignoring errors. In this case, I don't mind it so much as we are ignoring an error that is not our fault, it seems to be a problem in Monolog's code documentation.

Thoughts?

mailpoet/lib/Logging/LoggerFactory.php
pavel-mailpoet: Sorry @rodrigoprimo I am quite lost after this long discussion. I am quite unsure what do you suggest. Is there anything I should change to get this Pull request moving, or do you think it is good as is?

mailpoet/lib/Logging/PluginVersionProcessor.php
pavel-mailpoet: Let me confirm this with you first. This is the change you suggest I make, right? https://d.pr/i/Jz7D6P – it makes the error go away, and PHPStan is green. 

I am not sure how it helps here, but if it helps could you please let me understand?

rodrigoprimo: >  This is the change you suggest I make, right?

This is close to what I had in mind, but instead of using `@phpstan-ignore-next-line`, we could add the error to `phpstan.neon`. This way, we are ignoring a specific error instead of all errors in the next line, and we get notified by PHPStan when the error is not present anymore or changes.

> I am not sure how it helps here, but if it helps could you please let me understand?

To me, the main advantage, as I tried to elaborate above, is that this allows us to extend `ProcessorInterface`, and thus we get errors at compile time instead of runtime if the interface changes. Happy to chat synchronously about this if you would like to.

mailpoet/lib/Logging/LoggerFactory.php
rodrigoprimo: To try to summarize my suggestions:

- You asked me if I think we should always include all the processors regardless of the value of the `WP_DEBUG` and I'm inclined to think that we should not change this part of the logic and keep it as is.
- I'm suggesting we change the name of the variable `$attachProcessors` as now there is one processor that is always included. One possible replacement is `$attachOptionalProcessors`.

rodrigoprimo (approved):
LGTM!

veljkho (approved):

PR #4569, opened 2022-11-25
Fix flaky WooCommercePurchaseTest [MAILPOET-4854]

costasovo:
## Description

I found that the test was relying on the click and order Datetime to be in the same second. The click was created before the order, and sometimes the times differed, causing the flakiness. I fixed this by moving the order created at the time to the past by a minute.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4854]

## After-merge notes

_N/A_

[MAILPOET-4854]: https://mailpoet.atlassian.net/browse/MAILPOET-4854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lysyjan (approved):
Good job!

PR #4570, opened 2022-11-25
Try select option in select 2 multiple times [MAILPOET-4555]

lysyjan:
## Description

This PR adds retries when we want to select an option in select2. This change should improve our test's stability with select2. 

## Code review notes

I also increased some waiting times because this test doesn't work correctly on my local. 

## QA notes

I guess we can skip QA here when it is approved.

## Linked tickets

[MAILPOET-4555]

[MAILPOET-4555]: https://mailpoet.atlassian.net/browse/MAILPOET-4555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo (approved):
LGTM! I hope this will finally fix this test.🤞

PR #4573, opened 2022-11-25
Add react error boundary into js applications in the admin [MAILPOET-4706]

samnajian:
## Description

This PR introduces `ErrorBoundary` components and adds it to the following applications/modules:

-  subscribers/subscribers.jsx
-  newsletters/newsletters.jsx
-  newsletters_editor/initialize.jsx 
- segments/segments.jsx
- forms/forms.jsx
- form_editor/form_editor.jsx
- help/help.jsx
- subscribers/importExport/import.js
- subscribers/importExport/export.ts
- wizard/wizard.jsx
- experimental_features/experimental_features.jsx
- logs/logs
- automation/automation.tsx (very minimal, mostly left to MAILPOET-4907)

## Code review notes

The idea behind where to add and where not to add is as follows:
- The boundary is added to components that have a bit more complexity and are bug prone in case data or inputs are malformed
- For the components with multiple usages, using `withBoundary` in the definition source is favored.
- In most cases a custom `displayName` is added to components so to be able to customize the displayed name and make it easier to find the throwing component [despite React internally trying to infer it](https://reactjs.org/docs/react-component.html#displayname).

## QA notes

A casual walk-through should suffice, given the tests pass.

## Linked tickets

[MAILPOET-4706]

[MAILPOET-4706]: https://mailpoet.atlassian.net/browse/MAILPOET-4706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/common/error_boundary/error_boundary.tsx
+++ b/mailpoet/assets/js/src/common/error_boundary/error_boundary.tsx
@@ -1,0 +1,68 @@
+import { Children, Component } from '@wordpress/element';
+import { ComponentType } from 'react';
+import { getComponentDisplayName } from './utils';
+
+type ErrorBoundaryState = {
+  hasError: boolean;
+  error?: string;
+};
+
+export type ErrorBoundaryProps = {
+  onError?: (Error_Boundary_State) => ComponentType;
+};
+
+export class ErrorBoundary extends Component<
+  ErrorBoundaryProps,
+  ErrorBoundaryState
+> {
+  constructor(props: ErrorBoundaryProps) {
+    super(props);
+    this.state = {
+      hasError: false,
+      error: '',
+    };
+  }
+
+  static getDerivedStateFromError(error: Error & { fileName?: string }) {
+    return {
+      hasError: true,
+      error: `${error?.toString()} \nFile name: ${
+        error?.fileName ?? 'No fileName reported'
+      } \nStack trace: ${error?.stack ?? 'No stack trace reported'}`,
+    };
+  }
+
+  componentDidCatch(error, errorInfo) {
+    // eslint-disable-next-line no-console
+    console.log('logging', { error, errorInfo });
+  }
+
+  render() {
+    const { onError } = this.props;
+    const { hasError, error } = this.state;
+    if (hasError) {
+      if (onError) {
+        return onError(this.state);
+      }
+      return (
+        <>
+          <h1>
+            The application{' '}
+            <strong>
+              {Children.map(this.props.children, (child) =>
+                getComponentDisplayName(child as ComponentType),
+              ).join(', ')}{' '}
+            </strong>
+            encountered an error
+          </h1>
+          <p>
+            Please report the following error to{' '}
+            <a href="https://www.mailpoet.com/support/">MailPoet support</a>
+          </p>
+          <pre>{error}</pre>
+        </>
+      );
+    }
+    return this.props.children;
+  }
+}
--- a/mailpoet/assets/js/src/automation/editor/components/header/document_actions.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/header/document_actions.tsx
@@ -21,7 +21,7 @@ const Dropdown: ComponentType<
   }
 > = WpDropdown;
 
-export function DocumentActions({ children }): JSX.Element {
+function DocumentActions({ children }): JSX.Element {
   const { automationName, automationStatus, showIconLabels } = useSelect(
     (select) => ({
       automationName: select(storeName).getAutomationData().name,
@@ -101,3 +101,6 @@ export function DocumentActions({ children }): JSX.Element {
     </div>
   );
 }
+
+DocumentActions.displayName = 'DocumentActions';
+export { DocumentActions };
--- a/mailpoet/assets/js/src/automation/editor/components/header/errors.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/header/errors.tsx
@@ -9,6 +9,7 @@ import {
 import { useDispatch, useSelect } from '@wordpress/data';
 import { createContext } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
+import { ErrorBoundary } from 'common';
 import { Chip } from '../chip';
 import { ColoredIcon } from '../icons';
 import {
@@ -69,6 +70,8 @@ function StepError({ stepId }: StepErrorProps): JSX.Element {
     </CompositeItem>
   );
 }
+
+StepError.displayName = 'StepError';
 
 export function Errors(): JSX.Element | null {
   const [showPopover, setShowPopover] = useState(false);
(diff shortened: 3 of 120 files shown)


mailpoet/assets/js/src/common/error_boundary/error_boundary.tsx
```diff
@@ -0,0 +1,69 @@
...
+            encountered an error
+          </h1>
+          <p>
+            Please report the following error to{' '}
+            <a href="https://www.mailpoet.com/support/">MailPoet support</a>
+          </p>
+          <pre>{error}</pre>
+        </>
+      );
+      return <h1>Something went wrong.</h1>;
```
costasovo: I think this line is a dead code.

costasovo (commented):
Nice! Thank you for going thoroughly through the whole plugin code!
I tested it by throwing some random errors in different parts of the app, and it works nicely.
Getting a screenshot like this 👇 from our users instead of a white screen of death will be helpful. Thank you @samnajian for picking this task!
Screenshot 2022-12-09 at 12 52 34

I found one tiny issue and there is a conflict so I assigning back to you @samnajian.
This PR touches a lot of files so it would be nice to get it to merge soon. I would still appreciate a second pair of eyes on this. @JanJakes could you please check (at least brefly) this PR?

costasovo:
@samnajian Thanks for fixing the dead code. There is still a conflict in `welcome_wizard_controller.jsx` that needs to be addressed.

samnajian:
> @samnajian Thanks for fixing the dead code. There is still a conflict in `welcome_wizard_controller.jsx` that needs to be addressed.

Ops sorry for the inconvenience! on it.

costasovo (dismissed):

JanJakes (commented):
This looks great! 👏

I only have small observations/discussions:

1. This PR adds `displayName` to various components within the codebase. This may be a bit inconvenient when one needs to wrap the component under an error boundary with a new component or move things around. What if we made it the error boundary prop isntead and used it as `<ErrorBoundary displayName="SomeName">` and `withBoundary(Component, "SomeName")`? That would probably also keep the related parameters closer together.
2. The error boundaries seem to be very granular; what's a good approach to how to decide where to use it? Isn't this maybe even too granular? I'm just a bit worried that as the code evolves, people will not use it consistently. For instance, the `DocumentActions` component in the automation editor is a subcomponent of the header. Would an error fit in there? And how to decide whether we should be wrapping this component particularly?
3. Related to the previous point, it seems we have no boundaries close to the app root in some cases. I guess that can be useful to render the error info for any other uncaught errors — maybe we could add one boundary to each React app, close to its top level? E.g., to places like the form [editor index](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/form_editor/form_editor.jsx) and similar.

Personally, I'd maybe even go as far as to use it close to the root only, just for simplicity. On support, we'd still get the error messages (rather than "white screen" reports), so it's still useful. But I'm not sure, let me know what are your thoughts. Does @costasovo have any ideas or preferences here?

samnajian:
> This looks great! 👏
> 
> I only have small observations/discussions:
> 
>     1. This PR adds `displayName` to various components within the codebase. This may be a bit inconvenient when one needs to wrap the component under an error boundary with a new component or move things around. What if we made it the error boundary prop isntead and used it as `<ErrorBoundary displayName="SomeName">` and `withBoundary(Component, "SomeName")`? That would probably also keep the related parameters closer together.
> 
>     2. The error boundaries seem to be very granular; what's a good approach to how to decide where to use it? Isn't this maybe even too granular? I'm just a bit worried that as the code evolves, people will not use it consistently. For instance, the `DocumentActions` component in the automation editor is a subcomponent of the header. Would an error fit in there? And how to decide whether we should be wrapping this component particularly?
> 
>     3. Related to the previous point, it seems we have no boundaries close to the app root in some cases. I guess that can be useful to render the error info for any other uncaught errors — maybe we could add one boundary to each React app, close to its top level? E.g., to places like the form [editor index](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/form_editor/form_editor.jsx) and similar.
> 
> 
> Personally, I'd maybe even go as far as to use it close to the root only, just for simplicity. On support, we'd still get the error messages (rather than "white screen" reports), so it's still useful. But I'm not sure, let me know what are your thoughts. Does @costasovo have any ideas or preferences here?

Thanks for the great ideas @JanJakes 

1) While I see your point, I'd say the module that defines the component knows how to define the `displayName` as well, it's more or less like the classical `defaultProps` to me, so I'd say it belongs to the defining module not the user, the user would just say: Ok let's use this component `X` but if it breaks, I don't want it to break other branches!, so let's put its exceptions in a boundary!

2) This is by nature a rather hard decision to make to find out where to put the boundary and/or where not, my understanding is whenever an error is likely to happen, but even if it happens, the rest of the page/application remainds useful and can deliver their service, with this definition I'd prefer keeping the boundaries as granular as possible so that a tiny exception in a part of the application doesn't render the whole application unusable, with this definition I see the error reporting purpose secondary, as we already have other means of knowing what's breaking and this PR only adds another method.

3) Please see no.2

costasovo:
Thanks, @JanJakes for the great input!
Point 3 caught my eye. I think having the error boundary on the root level is very important to eliminate the white screen of death.
I looked into a couple of places in our codebase and found that we will need to do some refactors to be able to catch everything. For example in [form editor app bootstrap](https://github.com/mailpoet/mailpoet/blob/ab828b27087abbd0843fce1d07787eb6b0fa75f1/mailpoet/assets/js/src/form_editor/form_editor.jsx#L29-L35) a lot of stuff that potentially may cause erorr is done outside of React and the error boundary. [Automation bootstrap](https://github.com/mailpoet/mailpoet/blob/eb71dd8a68ee747ad10709460f0964b1beca9bf8/mailpoet/assets/js/src/automation/automation.tsx#L82-L92) or [settings bootstrap](https://github.com/mailpoet/mailpoet/blob/d116ff360b8d73f0ae09429ae0c365ed30a54def/mailpoet/assets/js/src/settings/index.tsx#L15-L19) are similar cases. So when the form data mapping fail with error we will still get the white screen of death. 
This is imo out of scope of this task but I think we should address that in the future as well. Maybe we could move these initial calls into an effect inside the main component.

JanJakes:
@samnajian @costasovo So I guess ultimately, we can have both — more granular boundaries as well as top-level ones to prevent pretty much any white screen of death and get some error message from the users. I guess that can be another task. We should probably announce to the rest of the team that we have error boundaries and maybe some kind of "use them, when..." instructions. One idea could be something like "use it for any route (that changes rendered views, such as tabs) or significantly complex component" or such.

JanJakes (dismissed):

JanJakes:
@costasovo Do you think we should address point 3 now or maybe create a ticket? Otherwise, we can move this on, I guess.

costasovo:
> @costasovo Do you think we should address point 3 now or maybe create a ticket? Otherwise, we can move this on, I guess.



@JanJakes I will prepare a new ticket. I added it to my todo list!

Aschepikov:
@samnajian There are some conflicts, can you pls resolve them

Aschepikov (approved):
Verified, gj!

samnajian:
Thanks for the ping @Aschepikov, the conflict is resolved.

PR #4582, opened 2022-11-30
Simplify <Tags> component and add wrapper components [MAILPOET-4628]

websupporter:
## Description
To reduce the complexity of the `<Tags />` component, I decided only `tags` would be a valid property for the tags. I added `<StringTags>` which would transform `strings` to `tags` and use the `<Tags />` component. `<SegmentTags />` would transform `segments` to `tags` and `<SubscriberTags />` would transform `subscribers`. These new components are now used in the code base instead of the generic `<Tags />` component.

## QA notes
This PR touches the `<Tags />` element. The behavior/appearance should not change. Some examples for those `<Tags />`:
image
image

## Linked tickets
[MAILPOET-4628]

[MAILPOET-4628]: https://mailpoet.atlassian.net/browse/MAILPOET-4628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/common/tag/_stories/tags.tsx
+++ b/mailpoet/assets/js/src/common/tag/_stories/tags.tsx
@@ -1,5 +1,5 @@
 import { Heading } from 'common/typography/heading/heading';
-import { Tags } from '../tags';
+import { SegmentTags, StringTags } from '../tags';
 
 export default {
   title: 'Tags',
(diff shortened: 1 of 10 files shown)


samnajian (approved):
LGTM! thanks

Aschepikov (approved):
QA verified, good job!

PR #4586, opened 2022-11-30
Feat/let users design nicer confirmation emails - MAILPOET-4649

triple0t:
## Description

Let users design nicer confirmation emails.

This PR will add support for designing and customizing confirmation emails with the newsletter email editor.

## QA notes

* Visit the settings page : `/wp-admin/admin.php?page=mailpoet-settings#/signup`
* Click on the toggle "Enable visual subscription confirmation emails"
* Click on the "Save settings" button
* Click on "Open template editor."
* You may edit the template or not. Click on save.
* Close the editor
* Subscribe to a Form on the site.
* You should receive the confirmation email with the new HTML template

## Linked tickets

[MAILPOET-4649](https://mailpoet.atlassian.net/browse/MAILPOET-4649)


alex-mailpoet (dismissed):
Good job! Looks good to me and works!

Aschepikov (commented):
Hi, I have an issue with the error "All emails must include an "Unsubscribe" link.
Add a footer widget to your email to continue". In this case, I don't have a footer in the content and I think that in the confirmation email it should be included by default if it is necessary or we need to make an exception to this type of email because now with the default confirmation email we don't have this unsubscribe buttons. 
image
Also, if I delete the subscribe button, I won't see an additional error about `Don't forget to include the activation link in the email`” from the ticket.
image
I have this error just in the preview but the text is different
image
And even without this button, this email is autosaving so I can receive this confirmation email without any button. Maybe we need to make this button like order information in the "Edit template for WooCommerce emails" where we cant  delete it
image
And there is a dot in the title 
image
We don't have this dot in the woocommerce customizer

triple0t:
I had a nagging suspicion I forgot something on this PR
but I was in a rush, and it skipped my mind
Thank you for catching this issue, @Aschepikov 🙌🏿

Aschepikov:
Thank you @triple0t , about this button with the activation link and autosaving. Yes now it is not autosaving always, but it is still autosaving when we are doing something with the button with the activation link, even if we are deleting it. 
image
 In that case, it is autosaving, but the confirmation email can't be sent without an activation link so when I am trying to resend it I can see the error that is something wrong with my sending method. I think we should fix it because for me wasn't obvious that the form was autosaved without a button and that the confirmation emails are not sent. 
image

image

triple0t:
@Aschepikov 

> Thank you @triple0t , about this button with the activation link and autosaving. Yes now it is not autosaving always, but it is still autosaving when we are doing something with the button with the activation link, even if we are deleting it.

On the Email Editor, Please perform a **Hard Refresh (reload)** and test again.
This will clear your browser cache.

Please note: We can not prevent the user from deleting the button block (at least not with the provided email template). 
  
       
Performing hard refresh
Chrome and Windows:
To hard refresh on Google Chrome on Windows, there are two ways you can do it:

* Hold down Ctrl and click the Reload button.
* Or Hold down Ctrl and press F5.

Chrome and Mac:
To do a hard refresh on Google Chrome on Mac, there are two ways that you can do it:

* Hold down Shift and click the Reload button.
* Or Hold down Command, Shift and click the ‘R’ key.

More information [here](https://fabricdigital.co.nz/blog/how-to-hard-refresh-your-browser-and-clear-cache).

Aschepikov:
@triple0t  I am checking this pr in incognito mode with a clean refresh every time, also I deleted all cache via application in inspect and settings in chrome, I am sure that it has autosaving when I am deleting this button

triple0t:
@Aschepikov 
> @triple0t I am checking this pr in incognito mode with a clean refresh every time, also I deleted all cache via application in inspect and settings in chrome, I am sure that it has autosaving when I am deleting this button

hmmm, okay, lemme add an extra commit.
Thanks for checking.

triple0t:
Hi @Aschepikov
Can you try again?

Kind reminder to download the new [release](https://output.circle-artifacts.com/output/job/9d9024d9-b341-4c6c-830d-241519a5a5c5/artifacts/0/home/circleci/mailpoet/mailpoet/mailpoet.zip)

Aschepikov (approved):
Verified, GJ!

PR #4587, opened 2022-11-30
1-click unsubscribe should be recorded as click stat as well [MAILPOET-4862]

samnajian:
## Description

- With this PR a 1-click unsubscribe records a click stat
- Also some previous wrong tests for 1-click business logic are fixed
- Per talk with @flakas we decided to keep using `[link:subscription_instant_unsubscribe_url]` for 1-click unsubscribe as well and not introduce a new shortcode.

## Code review notes

Please note the PR created for the premium plugin as mentioned below.

## QA notes

- Send out newsletter using MSS
- Click the 1-click unsubscribe button in your email viewer application
- Make sure a click is recorded in the DB (`wp_mailpoet_statistics_clicks`) and on the stats page `Unsubscribe link` with `1` click count for the given newsletter is shown

## Linked PRs
[662](https://github.com/mailpoet/mailpoet-premium/pull/662)

## Linked tickets

[MAILPOET-4862]

[MAILPOET-4862]: https://mailpoet.atlassian.net/browse/MAILPOET-4862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscription/Pages.php
+++ b/mailpoet/lib/Subscription/Pages.php
@@ -3,12 +3,16 @@
 namespace MailPoet\Subscription;
 
 use MailPoet\Config\Renderer as TemplateRenderer;
+use MailPoet\Cron\Workers\StatsNotifications\NewsletterLinkRepository;
+use MailPoet\Entities\NewsletterLinkEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\StatisticsUnsubscribeEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Form\AssetsController;
 use MailPoet\Newsletter\Scheduler\WelcomeScheduler;
+use MailPoet\Newsletter\Sending\SendingQueuesRepository;
 use MailPoet\Settings\TrackingConfig;
+use MailPoet\Statistics\StatisticsClicksRepository;
 use MailPoet\Statistics\Track\SubscriberHandler;
 use MailPoet\Statistics\Track\Unsubscribes;
 use MailPoet\Subscribers\LinkTokens;
@@ -82,6 +86,15 @@ class Pages {
   /** @var SubscriberSegmentRepository */
   private $subscriberSegmentRepository;
 
+  /*** @var NewsletterLinkRepository */
+  private $newsletterLinkRepository;
+
+  /*** @var StatisticsClicksRepository */
+  private $statisticsClicksRepository;
+
+  /*** @var SendingQueuesRepository */
+  private $sendingQueuesRepository;
+
   public function __construct(
     NewSubscriberNotificationMailer $newSubscriberNotificationSender,
     WPFunctions $wp,
@@ -98,7 +111,10 @@ class Pages {
     TrackingConfig $trackingConfig,
     EntityManager $entityManager,
     SubscriberSaveController $subscriberSaveController,
-    SubscriberSegmentRepository $subscriberSegmentRepository
+    SubscriberSegmentRepository $subscriberSegmentRepository,
+    NewsletterLinkRepository $newsletterLinkRepository,
+    StatisticsClicksRepository $statisticsClicksRepository,
+    SendingQueuesRepository $sendingQueuesRepository
   ) {
     $this->wp = $wp;
     $this->newSubscriberNotificationSender = $newSubscriberNotificationSender;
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Subscription/Pages.php
```diff
@@ -230,14 +249,23 @@ public function confirm() {
   public function unsubscribe(string $method): void {
     if (
       !$this->isPreview()
-      && ($this->subscriber !== null)
-      && ($this->subscriber->status !== SubscriberEntity::STATUS_UNSUBSCRIBED)
+      && (!is_null($this->subscriber))
+      && ($this->subscriber->getStatus() !== SubscriberEntity::STATUS_UNSUBSCRIBED)
```
costasovo, line 253: Wow. So this was working thanks to the [BC code we added to the entity](https://github.com/mailpoet/mailpoet/blob/926620e8f82ca691f1c11f6e973a2f86a6117522/mailpoet/lib/Entities/SubscriberEntity.php#L193-L203). Nice catch! Hmm maybe we should consider starting throwing deprecation warnings inside the BC code.

mailpoet/lib/Subscription/Pages.php
```diff
@@ -230,14 +249,23 @@ public function confirm() {
...
     ) {
       if ($this->trackingConfig->isEmailTrackingEnabled() && isset($this->data['queueId'])) {
+        $queueId = (int)$this->data['queueId'];
+
+        if ($method === StatisticsUnsubscribeEntity::METHOD_ONE_CLICK) {
+          /**
+           * With 1-click method, redirect shouldn't happen that's why the click state should be directly recorded
+           */
+          $this->updateClickStatistics($queueId);
```
costasovo, line 262: Shouldn't the click tracking be done on a different level, like within the code where we track the rest of the clicks? Or is it not possible for some reason? I don't have in-depth knowledge of the one-click unsubscribe feature, so maybe there is a good reason.

costasovo (commented):
Looks good! I have just one question. We might be leaking click-tracking logic to the already bloated Pages class. Is that necessary?

mailpoet/lib/Subscription/Pages.php
samnajian: The one-click endpoint should not redirect, that's why I'm not using the the ` track/click` endpoint, which finally redirects to the endpoint defined in `link_hash`.

mailpoet/lib/Subscription/Pages.php
samnajian: I agree with the deprecation warning idea! we may have this issue in other parts of the code base or may even introduce in future.

samnajian:
> Looks good! I have just one question. We might be leaking click-tracking logic to the already bloated Pages class. Is that necessary?

I agree that it's not the best place to add such a tracking, even though [we already have tracking](https://github.com/mailpoet/mailpoet/blob/a8e25d45a5e48b608613ab482dd1e4f6e62a0117/mailpoet/lib/Subscription/Pages.php#L265) in this file, the reason for adding the new `updateClickStatistics` to this file was 1) This tracking needs data which are already prepared and ready in this class and I didn't want to fetch the same data in another place as well, specially since it involves fetching them from the DB 2) As we've already talked, I'm adding a maintenance ticket for this class, and thought under that ticket we can put things where they belong. [Ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4898) added.

mailpoet/lib/Subscription/Pages.php
costasovo: Aha! So this one-click unsubscribe URL is never pointed to the tracking endpoint and in every email, we send the URL that points directly to the unsubscribe page. Correct?

costasovo:
> I agree that it's not the best place to add such a tracking, even though [we already have tracking](https://github.com/mailpoet/mailpoet/blob/a8e25d45a5e48b608613ab482dd1e4f6e62a0117/mailpoet/lib/Subscription/Pages.php#L265) in this file, the reason for adding the new updateClickStatistics to this file was 1) This tracking needs data which are already prepared and ready in this class and I didn't want to fetch the same data in another place as well, specially since it involves fetching them from the DB 2) As we've already talked, I'm adding a maintenance ticket for this class, and thought under that ticket we can put things where they belong. [Ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4898) added. 

Got it! I forgot our discussion about the maintanence ticket. I see a potential issue when we decide to add some new data (e.g. we allow admin to add custom tracking data to links via a hook) we will need to make sure these data are added also to this one click URL which we want to track similarly to the other links but without  passing through link tracking. But we haven't add new data to tracking for ages so hopefully this won't be an issue.

costasovo (approved):
LGTM! 👍

Aschepikov (approved):
GJ!

PR #4590, opened 2022-12-01
Change title and description of the key activation settings tab [MAILPOET-4749]

rodrigoprimo:
## Description

This PR changes the text of the title and the description of the key activation settings tab to hopefully make it more clear the meaning of the key and what they should do if they don't have one or if they are unsure where they can get their key.

## Code review and QA notes

1. Go to /wp-admin/admin.php?page=mailpoet-settings#/premium
2. Check that the title, the description and the links in the page match what was described in the ticket.

Please note that one of the links uses the `s` parameter to pass to `account.mailpoet.com` the number of subscribers. The shop is currently ignoring this parameter due to a bug that will be addressed separately: p1669906982764479-slack-C01GH735066

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4749]

## After-merge notes

_N/A_

[MAILPOET-4749]: https://mailpoet.atlassian.net/browse/MAILPOET-4749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/settings/pages/key_activation/key_activation.tsx
+++ b/mailpoet/assets/js/src/settings/pages/key_activation/key_activation.tsx
@@ -86,6 +86,34 @@ type Props = {
   subscribersCount: number;
 };
 
+const premiumTabDescription = ReactStringReplace(
+  t('premiumTabDescription'),
+  /\[link\](.*?)\[\/link\]/g,
+  (text) => (
+    <a
+      href="https://account.mailpoet.com/account?utm_source=plugin&utm_medium=settings&utm_campaign=activate-existing-plan&ref=settings-key-activation"
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      {text}
+    </a>
+  ),
+);
+
+const premiumTabGetKey = ReactStringReplace(
+  t('premiumTabGetKey'),
+  /\[link\](.*?)\[\/link\]/g,
+  (text) => (
+    <a
+      href="https://account.mailpoet.com/account?utm_source=plugin&utm_medium=settings&utm_campaign=activate-existing-plan&ref=settings-key-activation"
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      {text}
+    </a>
+  ),
+);
+
 export function KeyActivation({ subscribersCount }: Props) {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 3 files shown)


mailpoet/assets/js/src/settings/pages/key_activation/key_activation.tsx
```diff
@@ -147,7 +152,53 @@ export function KeyActivation() {
...
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  {text}
+                </a>
+              ),
+            )}
+            <br />
+            <br />
+            {ReactStringReplace(
```
samnajian: I was wondering if assigning these two regex replacements to a constant outside of the render cycle and simply using them would be better, what do you?  These first replacements are not bound to any `prop` so in practice they are simply a constant by nature and don't need to re-calculate them in each render.

samnajian (commented):
I guess it's also a good time to introduce some utility like the following ( it's just a suggestion )
```
const translateForLink = (key: string, Link: ComponentType<{text: string}>) => ReactStringReplace(
  MailPoet.I18n.t(key),
  /\[link\](.*?)\[\/link\]/g,
  (text) => (
   <Link text={text}/>
  ),
);
```
Or even with more optional arguments so that we won't need to know about the regex or the underlying `ReactStringReplace`, what do you think?

rodrigoprimo:
I just rebased this PR per p1670235909191389-slack-C01H71ZLBGQ

mailpoet/assets/js/src/settings/pages/key_activation/key_activation.tsx
rodrigoprimo: Fair point. I pushed a new commit implementing this.

rodrigoprimo:
Thanks for the review, @samnajian. I implemented one of your suggestions and I'm suggesting that we address the second one separately. Could you please check this PR again?

> I guess it's also a good time to introduce some utility like the following ( it's just a suggestion )

I published a P2 post (pcNwfB-1AU-p2) to start a discussion about this, and my suggestion is that we address it in a separate ticket.

samnajian (approved):
LGTM! thanks

Aschepikov (approved):
GJ!

PR #4591, opened 2022-12-02
Automation StepType can be undefined [MAILPOET-4889]

websupporter:
## Description
When I open an automation with a third party step and the plugin, which provides this step is deactivated, the editor does not load properly:
image

This PR fixes the issue:
image

## Code review notes

_N/A_

## QA notes
[hello-automation.zip](https://github.com/mailpoet/mailpoet/files/10139481/hello-automation.zip)

To reproduce:
You can install this plugin. It will provide you a "Hello Automation"-Action. Add this action to an automation and save the automation with your current `trunk`. Deactivate the plugin and reload the edtior: You should the the error.

Use the MailPoet version of this PR and the error should be gone.

Note:
There is a second problem. If you reload the editor with the plugin activated, you will run into a similar issue, which will not be addressed in this PR but subject of an upcoming P2 post.

## Linked tickets
[MAILPOET-4889]

## Linked PR
https://github.com/mailpoet/mailpoet-premium/pull/663

[MAILPOET-4889]: https://mailpoet.atlassian.net/browse/MAILPOET-4889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
@@ -2,7 +2,6 @@ import { __ } from '@wordpress/i18n';
 import { chartBar } from '@wordpress/icons';
 import { Hooks } from 'wp-js-hooks';
 import { MoreControlType, StepMoreControlsType } from '../../../types/filters';
-import { StepType } from '../../../editor/store';
 import { Step } from '../../../editor/components/automation/types';
 
 const emailStatisticsControl = (step: Step): MoreControlType => {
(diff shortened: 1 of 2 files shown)


mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
```diff
@@ -33,9 +33,9 @@ export function registerStepControls() {
     (
       controls: StepMoreControlsType,
       step: Step,
-      stepType: StepType,
+      stepType: StepType | undefined,
     ): StepMoreControlsType => {
-      if (stepType.key === 'mailpoet:send-email') {
+      if (stepType?.key === 'mailpoet:send-email') {
```
JanJakes: What if we drop `stepType` from this filter and use `step.key` instead?
The thought process is that even an unknown step must have a `key`. And if someone needs a step type, it may be fetchable using a selector, I guess.

websupporter: That's a good idea. Rebased and fixed in da2dd49. Please be aware of https://github.com/mailpoet/mailpoet-premium/pull/663 as da2dd49 alters the signature of the hook.

JanJakes (approved):
Thanks!

Aschepikov (approved):
GJ!

PR #4593, opened 2022-12-02
Fix issue when sql_mode ANSI_QUOTES is enabled [MAILPOET-4887] 

costasovo:
## Description
Some time ago [we started treating ANSI_QUOTES mode as incompatible](https://mailpoet.atlassian.net/browse/MAILPOET-3579), and we now [remove it from sql_mode for DB session in the connection factory](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Doctrine/ConnectionFactory.php#L67-L70).

When solving a ticket on L2 where a customer complained about errors with missing mailpoet tables, I learned that WP core also has a list of [incompatible modes](https://github.com/WordPress/WordPress/blob/73bd837100ffdbe779277287532dc1dc352df6f0/wp-includes/class-wpdb.php#L648-L655), but ANSI_QUOTES is not among them. In our [initial migration](https://github.com/mailpoet/mailpoet/blob/caa76983e4baaa86d613c05101295cd8df224086/mailpoet/lib/Migrations/Migration_20221028_105818.php#L106), we use `dbDelta` function which works with `$wpdb` so we need to make sure that our migration code is ANSI_QUOTES compatible. The first commit contains fixes of the migration SQL so that it works with ANSI_QUOTES enabled. 

I searched all direct usages of $wpdb in the plugin and it seems that they are all compatible with the ANSI_QUOTES mode.

Btw. I think that this issue may be responsible for the occasional issues with missing DB tables we see on support.

Another issue I found was that the ANSI_QUOTE mode can be activated using [ANSI mode which is a shortcut for grouping a couple of modes](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_ansi). I addressed it in the second commit.

We will need to look more deeply into SQL modes especially as we are planning to write $wpdb compatible driver for Doctrine. This will force us to support the same modes as WP does. I plan to write a P2 about it.

## Code review notes

_N/A_

## QA notes
Configure your site to use ANSI mode and try to install MailPoet plugin and test basic functionality.
You can set the mode by going into adminer/PHPMyAdmin
and run 
```sql
 SET @@global.sql_mode = 'REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE,ONLY_FULL_GROUP_BY,ANSI,STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION';
```
This will enabe probably all possible problematic modes.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4887]

## After-merge notes

_N/A_

[MAILPOET-4887]: https://mailpoet.atlassian.net/browse/MAILPOET-4887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


websupporter (commented):
Do you have an idea, if we can prevent somehow the introduction of quotes in the future? I currently have none tbh. If we do not find an automated prevention, maybe a P2 would be useful to at least spread the knowledge. What do you think?

costasovo:
@websupporter Here is a P2 post about the issue https://mailpoetdev.wordpress.com/2022/12/02/mysql-modes-in-mailpoet-vs-wpdb/
I also created a Jira ticket (https://mailpoet.atlassian.net/browse/MAILPOET-4892) for making plugin queries compatible with the mode. Within the ticket, I suggest we set our test DB to use the mode so in case we introduce wrong quotes, tests should fail.
Hmm, thinking about it. I guess I could configure the test environment (integration and acceptance test) to use the mode within this PR. What do you think?

websupporter:
sounds good to me :)

costasovo:
@websupporter I added four new commits:

1) Configured the docker dev env to use ANSI and FULL_GROUP_BY modes
2) Configured tests environment to use ANSI and FULL_GROUP_BY modes
3) Added code that prints active SQL modes when test environment start  (see [here](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/12454/workflows/d8210073-2788-4a0a-a4f9-d090f481c9d4/jobs/213015/parallel-runs/0/steps/0-106) you need to scroll a bit)
4) Fixed a few occurrences where we had incompatible queries. Good that I switched the test configuration.

I added both ANSI and FULL_GROUP_BY, because these two we switch off. Please see also commit notes for additional details.

websupporter (approved):
Thank you very much @costasovo. lgtm.

veljkho (approved):
Checked integration and acceptance tests for this specific sql_mode and they are passing. I was unable to enable ansi_quotes on my test sites due to lack of privileges (the sql code won't run).

PR #4599, opened 2022-12-06
Use method from acceptance tester [MAILPOET-4899]

lysyjan:
## Description

Add triggering of the action scheduler as it is in a similar method in the acceptance tester.

## QA notes

We can skip QA here.

## Linked tickets

[MAILPOET-4899]

[MAILPOET-4899]: https://mailpoet.atlassian.net/browse/MAILPOET-4899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo (approved):
Thanks! Good idea to add the comment to the method 👍 The next time, we won't need to figure out why it was added.

PR #4601, opened 2022-12-07
Surface confirmation email sending errors to subscribers and admins [MAILPOET-4736]

costasovo:
## Description
In the current version of the plugin, [we silently ignore errors when a confirmation email fails](https://github.com/mailpoet/mailpoet/blob/12afcfb656a825f8d30d1647d1fdd6b9be4b3db5/mailpoet/lib/Subscribers/SubscriberActions.php#L105-L109). So subscribers don't get any feedback that something went wrong when they submit a subscription form and site admin doesn't know that sending doesn't work. This PR addresses these two issues.

When confirmation email sending fails, a subscriber should see the following error.

Screenshot 2022-12-07 at 16 13 59

When confirmation email sending fails three times (with at least 2 minutes interval in between those failures), the site admin should see a notice on the emails page.

Screenshot 2022-12-07 at 10 52 34

## Code review notes

The PR should be easier to review going commit by commit

#### Testing instructions
1. Misconfigure sending (use an invalid MSS key or just point SMTP to an invalid host)
2. Make sure subscription confirmation is enabled in the plugin settings
3. Try to add a subscriber via a subscription form and verify the correct error message (Note that the subscriber still will be added to the DB but they will see the error message)
4. After adding fails wait 2 minutes (maybe more because of IP throttling) and try to add another one and verify the correct error message
5. Repeat the step for (4)
6. After three failed attempts, you should see in WP Admin MailPoet > Emails a notice with info about the sending issue

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4736]

## After-merge notes

_N/A_

[MAILPOET-4736]: https://mailpoet.atlassian.net/browse/MAILPOET-4736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Mailer/MailerLog.php
+++ b/mailpoet/lib/Mailer/MailerLog.php
@@ -201,7 +201,7 @@ class MailerLog {
     $mailerLog['transactional_email_last_error_at'] = time();
     $mailerLog['transactional_email_error_count'] = ($mailerLog['transactional_email_error_count'] ?? 0) + 1;
     self::updateMailerLog($mailerLog);
-    if ($mailerLog['transactional_email_error_count'] > self::RETRY_ATTEMPTS_LIMIT) {
+    if ($mailerLog['transactional_email_error_count'] >= self::RETRY_ATTEMPTS_LIMIT) {
       self::pauseSending($mailerLog);
     }
   }
--- a/mailpoet/lib/Mailer/MailerLog.php
+++ b/mailpoet/lib/Mailer/MailerLog.php
@@ -16,7 +16,9 @@ use MailPoet\Settings\SettingsController;
  *   "status": ?string,
  *   "retry_attempt": ?int,
  *   "retry_at": ?int,
- *   "error": ?MailerLogError
+ *   "error": ?MailerLogError,
+ *   "transactional_email_last_error_at": ?int,
+ *   "transactional_email_error_count": ?int,
  * }
  */
 
@@ -56,6 +58,8 @@ class MailerLog {
       'retry_attempt' => null,
       'retry_at' => null,
       'error' => null,
+      'transactional_email_last_error_at' => null,
+      'transactional_email_error_count' => null,
     ];
     $settings = SettingsController::getInstance();
     $settings->set(self::SETTING_NAME, $mailerLog);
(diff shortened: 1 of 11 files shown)


mailpoet/lib/Mailer/MailerLog.php
```diff
@@ -172,6 +178,34 @@ public static function processError(
...
+    $ignoreErrorThreshold = time() - (2 * 60); // 2 minutes ago
+    // We want to log the error max one time per 2 minutes
+    if ($lastErrorTime && $lastErrorTime > $ignoreErrorThreshold) {
+      return;
+    }
+    $mailerLog = self::setError($mailerLog, $operation, $errorMessage, $errorCode);
+    $mailerLog['transactional_email_last_error_at'] = time();
+    $mailerLog['transactional_email_error_count'] = ($mailerLog['transactional_email_error_count'] ?? 0) + 1;
+    self::updateMailerLog($mailerLog);
+    if ($mailerLog['transactional_email_error_count'] > self::RETRY_ATTEMPTS_LIMIT) {
```
rodrigoprimo: > When confirmation email sending fails three times (with at least 2 minutes interval in between those failures), the site admin should see a notice on the emails page.

Since we are using `>` here instead of `>=`, we are pausing sending and showing the admin notice when the confirmation email sending fails four times instead of three times. This is just a detail, and I think either way works, I'm mentioning it just for consistency. If we decide to keep it the way it is, we might want to update the PR and the ticket descriptions.

rodrigoprimo (dismissed):
Looks good to me. Good job!

I'm approving this PR, but I left one inline comment that might or might not require a small code change. Please decide which approach you prefer and either move this PR to QA or request another review after changing the code.

mailpoet/lib/Mailer/MailerLog.php
costasovo: @rodrigoprimo Nice catch! I'll update the condition.

costasovo:
@rodrigoprimo I added a new commit with a fix of the condition.

rodrigoprimo (approved):

Aschepikov (approved):
GJ!

PR #4603, opened 2022-12-07
Changes to the first step of the Welcome Wizard [MAILPOET-4815]

rodrigoprimo:
## Description

This PR implements a few changes to the first step of the Welcome Wizard as described in the ticket.

## Code review notes

There is some repetition in `welcome_wizard_controller.jsx` and `woocommerce_controller.jsx` (`finishWizard()` and `updateSettings()`). I did not address this as the repetition was already there before I started this PR and it was not clear to me how to address it in a React way. Please let me know if you have any suggestions and if you think I should change this part.

## QA notes

Go to /wp-admin/admin.php?page=mailpoet-welcome-wizard#/steps/1 to access the first step of the Welcome Wizard and see the changes proposed in this PR.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4815]

## After-merge notes

_N/A_

[MAILPOET-4815]: https://mailpoet.atlassian.net/browse/MAILPOET-4815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo:
@samnajian, I refactored three of the four JSX files that were changed in this PR to TSX. I didn't refactor `woocommerce_controller.jsx` as I plan to touch this file on #4605. Could you please review this PR again?

Please note that this is one of the first times that I have refactored JSX files to TSX, so I'm not sure about all the changes that I made and it is likely I missed something.

mailpoet/assets/js/src/wizard/steps/sender_step.tsx
```diff
@@ -52,47 +72,32 @@ function WelcomeWizardSenderStep(props) {
               name="senderAddress"
               type="text"
               placeholder="john@doe.com"
-              value={props.sender ? props.sender.address : ''}
+              value={sender ? sender.address : ''}
```
samnajian, line 74: You can replace this with [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) as well

mailpoet/assets/js/src/wizard/welcome_wizard_controller.tsx
```diff
@@ -71,31 +81,29 @@ function WelcomeWizardStepsController(props) {
...
+      redirect(step),
+    );
   }, [redirect, sender, step]);
 
   const skipSenderStep = useCallback(
     (e) => {
       e.preventDefault();
       setLoading(true);
-      updateSettings(
+      void updateSettings(
```
samnajian: Not sure what value does this `void` operator add in these case, ts seems like to be happy without it as well, and I guess removing it doesn't change the logic either.

samnajian (commented):
Thanks for the great work and js->ts conversions as well, I've just added some minor comments.

mailpoet/assets/js/src/wizard/steps/sender_step.tsx
rodrigoprimo: That is good to know. I don't think it is necessary to create a commit to change it in this case, but I will try to remember next time. Let me know if you disagree and you think I should change it in this PR.

samnajian: It's fine to leave it like this.

mailpoet/assets/js/src/wizard/welcome_wizard_controller.tsx
rodrigoprimo: Without the `void` ESLint returns the following error:

```
mailpoet/mailpoet/assets/js/src/wizard/welcome_wizard_controller.tsx
  63:5  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
```

My assumption is that since we are already handling errors inside `updateSettings()`, it is safe to use `void` here instead of using `catch()` or the second parameter of `then()`, but I'm not 100% sure. What do you think?

samnajian: I'd prefer prefixing with `await` to better expose the exceptions if they happen specially since as a user we don't know if and how the exceptions are handled inside `updateSettings`, besides this function `submitTracking` is stalling the application by setting `setLoading(true);` so by nature it's sync and it's fine to await it.

rodrigoprimo: I don't know if I'm missing something as I'm not super familiar the all the concepts here, but I don't think I can simply use `await` as it should be used only inside an async function or at the top level of a module. Neither is the case for the places where I used `void`. Happy to chat synchronously about this tomorrow if it is easier so that we can find a solution.

rodrigoprimo: Thanks for the information you shared on Slack, Sam. This PR is ready for a final review.

samnajian (dismissed):
LGTM, thanks for the updates.

Aschepikov (dismissed):
GJ!

Aschepikov:
@rodrigoprimo Hi, There are some conflicts can you pls resolve them

rodrigoprimo:
@samnajian, could you please take a final look at this PR after the rebasing and the subsequent fixes? Thanks.

samnajian (approved):
LGTM! thanks for the updates

Aschepikov (approved):
Verified, gj!

PR #4602, opened 2022-12-07
Changes to the second step of the Welcome Wizard [MAILPOET-4816]

rodrigoprimo:
## Description

This PR implements a few changes to the second step of the Welcome Wizard as described in the ticket. 

## Code review notes

_N/A_

## QA notes

Go to `/wp-admin/admin.php?page=mailpoet-welcome-wizard#/steps/2` to access the second step of the Welcome Wizard and see the changes proposed in this PR.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4816]

## After-merge notes

_N/A_

[MAILPOET-4816]: https://mailpoet.atlassian.net/browse/MAILPOET-4816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4605, opened 2022-12-08
Redesign the WooCommerce step of the welcome wizard [MAILPOET-4817]

rodrigoprimo:
## Description

This PR implements a few changes to the WooCommerce step of the Welcome Wizard as described in the Jira ticket.

## Code review notes

I'm unsure about the approach that I used in the commit to change the order of the steps (especially around the way that I'm passing the current step and redirectToNextStep to the WooCommerce controller). Please let me know if there is a more React way to achieve the same.

## QA notes

Go to /wp-admin/admin.php?page=mailpoet-welcome-wizard#/steps/3 to see the changes. Please note that the change to the horizontal line was made in a different PR (#4602) and is only visible in the WooCommerce step when this branch is combined with the from the other PR.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4817]

## After-merge notes

_N/A_

[MAILPOET-4817]: https://mailpoet.atlassian.net/browse/MAILPOET-4817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/wizard/welcome_wizard_controller.jsx
+++ b/mailpoet/assets/js/src/wizard/welcome_wizard_controller.jsx
@@ -6,7 +6,7 @@ import { MailPoet } from 'mailpoet';
 import { WelcomeWizardSenderStep } from './steps/sender_step.jsx';
 import { WelcomeWizardUsageTrackingStep } from './steps/usage_tracking_step.jsx';
 import { WelcomeWizardPitchMSSStep } from './steps/pitch_mss_step.jsx';
-import { WooCommerceController } from './woocommerce_controller.jsx';
+import { WooCommerceController } from './woocommerce_controller';
 import { WelcomeWizardStepLayout } from './layout/step_layout.jsx';
 
 import { createSenderSettings } from './create_sender_settings.jsx';
--- a/mailpoet/assets/js/src/wizard/welcome_wizard_controller.jsx
+++ b/mailpoet/assets/js/src/wizard/welcome_wizard_controller.jsx
@@ -150,7 +150,11 @@ function WelcomeWizardStepsController(props) {
         ) : null}
 
         {stepName === 'WizardWooCommerceStep' ? (
-          <WooCommerceController isWizardStep />
+          <WooCommerceController
+            isWizardStep
+            currentStep={step}
+            redirectToNextStep={redirect}
+          />
         ) : null}
       </StepsContent>
     </>
--- a/mailpoet/assets/js/src/wizard/welcome_wizard_controller.jsx
+++ b/mailpoet/assets/js/src/wizard/welcome_wizard_controller.jsx
@@ -152,8 +152,7 @@ function WelcomeWizardStepsController(props) {
         {stepName === 'WizardWooCommerceStep' ? (
           <WooCommerceController
             isWizardStep
-            currentStep={step}
-            redirectToNextStep={redirect}
+            redirectToNextStep={() => redirect(step)}
           />
         ) : null}
       </StepsContent>
--- a/mailpoet/assets/js/src/wizard/woocommerce_controller.jsx
+++ b/mailpoet/assets/js/src/wizard/woocommerce_controller.jsx
@@ -5,7 +5,11 @@ import { StepsContent } from 'common/steps/steps_content.tsx';
 import { WizardWooCommerceStep } from './steps/woo_commerce_step.jsx';
 import { WelcomeWizardStepLayout } from './layout/step_layout.jsx';
 
-function WooCommerceController({ isWizardStep = false }) {
+function WooCommerceController({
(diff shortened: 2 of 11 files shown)


mailpoet/assets/js/src/wizard/welcome_wizard_controller.jsx
```diff
@@ -150,7 +150,11 @@ function WelcomeWizardStepsController(props) {
         ) : null}
 
         {stepName === 'WizardWooCommerceStep' ? (
-          <WooCommerceController isWizardStep />
+          <WooCommerceController
+            isWizardStep
+            currentStep={step}
+            redirectToNextStep={redirect}
```
costasovo: I think you don't need to pass both `currentStep` and `redirectToNextStep`. You could use an anonymous function and save one property. But it is just a detail.
```js
redirectToNextStep={() => redirect(step)}
```

costasovo (dismissed):
@rodrigoprimo Good job! I added a comment with an idea of how we could get rid of one of the props on the WooCommerceStep. I don't think it is necessary to fix that, but I'm assigning it back to you and letting you decide. Please feel free to move the PR forward.

mailpoet/assets/js/src/wizard/welcome_wizard_controller.jsx
rodrigoprimo: I like your suggestion and I haven't considered it before. I just added a commit that implements your idea. Thanks.

mailpoet/assets/js/src/wizard/woocommerce_controller.jsx
```diff
@@ -71,10 +82,12 @@ function WooCommerceController({ isWizardStep = false }) {
 
 WooCommerceController.propTypes = {
   isWizardStep: PropTypes.bool,
+  redirectToNextStep: PropTypes.shape,
```
rodrigoprimo: @costasovo, one thing that I forgot to ask before is about the way that I defined the type of this parameter. Is just using `PropTypes.shape` here considered enough? Maybe there is a way to better defined what this parameter is about?

costasovo: @rodrigoprimo Good question! I missed that. Hmm, I think we used to have lint check for that, but maybe we removed that as we are shifting toward TS. Ideally [as we discussed](https://a8c.slack.com/archives/C01GH835U90/p1667207368379109), we should try to rewrite .jsx files to Typescript when we touch them when we work on a feature. In such a case, we would remove the proptypes definitions and replace them with TS type definitions where we don't allow the `object` type.

Sometimes it is not easy to refactor to TS, so it is ok to keep the JSX file. What I wanted to say is that we will fix it during the rewrite to TS, and the component is not reused in many places (compared to some UI components), so I think we can live with the `shape`. But feel free to add a more specific definition of the property, or you can try refactoring to Typescript. Here is [a commit where I recently refactored o component on a similar level of abstraction](https://github.com/mailpoet/mailpoet/pull/4609/commits/55baa581a4433ea9acee5c9cebfc905439d88078) you may need to do similar changes if you decide to give it a try. I need to practice TS as well, so I'm happy to assist.

costasovo (dismissed):
@rodrigoprimo  Good job! I'm approving the PR as is and I'm assigning it back to you. Feel free to move this forward. Please see my other comment.

mailpoet/assets/js/src/wizard/woocommerce_controller.jsx
rodrigoprimo: @costasovo, I completely forgot about refactoring JSX files to TSX. I added a couple of commits with my attempt in this conversation. This is one of the first times that I have done this, so there is a chance that there is something wrong. Could you please review this PR again?

Unfortunately, I had to rebase this PR and then force-push the commits, so the hash of the commits that you already reviewed changed. Also, I did not refactor steps_numbers.jsx and welcome_wizard_controller.jsx as I addressed those in #4603.

For some reason, GitHub is not picking up that woo_commerce_step.jsx was renamed to woocommerce_step.tsx and is showing woocommerce_step.tsx as a new file. My local git client is correctly identifying the rename and showing only the specific changes in the file making it easier to review.

> Here is [a commit where I recently refactored o component on a similar level of abstraction](https://github.com/mailpoet/mailpoet/pull/4609/commits/55baa581a4433ea9acee5c9cebfc905439d88078) you may need to do similar changes if you decide to give it a try.

Thanks, having an example was super helpful.

costasovo (approved):
Thank you for refactoring those components to TS!

Aschepikov (approved):
GJ!

PR #4610, opened 2022-12-13
Update tiny mce to version 6 [MAILPOET-4213]

### LINKED ISSUE #4167: Cannot use a shortcode as a url in links
**Describe the bug**
I selected some text in a text block and created a link. I was unable to use spaces in the URL input to insert a shortcode unless I insert it in the middle of existing text.

**To reproduce**

1. Create a text block.
2. Select some text.
3. Click "link" button
4. Start entering a shortcode `[custom:something url="helloworld"]`
5. Cannot enter a space, however, if you go back a character and input a space, you can insert spaces.

**Actual behavior**
Complicated to enter shortcodes in urls

**Expected behavior**
Able to enter shortcodes in urls.

samnajian:
## Description

This PR updates TinyMCE used in email editor.

## QA notes

Please test the text block inside newsletter editor and make sure all the functions works and renders properly after sending the newsletter. Tiny MCE is about all those functions to bold the text, make it italic, alignments...

## Linked tickets

[MAILPOET-4213], closes #4167

[MAILPOET-4213]: https://mailpoet.atlassian.net/browse/MAILPOET-4213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/newsletter_editor/behaviors/TextEditorBehavior.ts
+++ b/mailpoet/assets/js/src/newsletter_editor/behaviors/TextEditorBehavior.ts
@@ -4,23 +4,24 @@
  * Adds TinyMCE text editing capabilities to a view
  */
 import Marionette from 'backbone.marionette';
-import tinymce from 'tinymce/tinymce';
+import tinymce, { RawEditorOptions } from 'tinymce/tinymce';
 import { BehaviorsLookup } from 'newsletter_editor/behaviors/BehaviorsLookup';
 import { App } from 'newsletter_editor/App';
 import { tinyMceAdd } from 'newsletter_editor/tinymce/mailpoet_shortcodes.js';
-
 
 // TinyMCE theme and plugins
 import 'tinymce/themes/silver';
 import 'tinymce/plugins/code';
 import 'tinymce/plugins/link';
 import 'tinymce/plugins/lists';
+import 'tinymce/plugins/paste';
 
 import './tinymce_icons';
 
-var BL = BehaviorsLookup;
+const configurationFilter = (originalConfig: RawEditorOptions) =>
+  originalConfig;
 
-BL.TextEditorBehavior = Marionette.Behavior.extend({
+BehaviorsLookup.TextEditorBehavior = Marionette.Behavior.extend({
   defaults: {
     selector: '.mailpoet_content',
     toolbar1: 'bold italic link unlink forecolor mailpoet_shortcodes',
@@ -30,9 +31,7 @@ BL.TextEditorBehavior = Marionette.Behavior.extend({
     invalidElements: 'script',
     blockFormats: 'Paragraph=p',
     plugins: 'link mailpoet_shortcodes',
-    configurationFilter: function configurationFilter(originalConfig) {
-      return originalConfig;
-    },
+    configurationFilter,
   },
   initialize: function initialize() {
     this.listenTo(App.getChannel(), 'dragStart', this.hideEditor);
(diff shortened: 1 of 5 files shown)


mailpoet/assets/js/src/newsletter_editor/behaviors/TextEditorBehavior.ts
```diff
@@ -9,12 +9,12 @@ import { BehaviorsLookup } from 'newsletter_editor/behaviors/BehaviorsLookup';
 import { App } from 'newsletter_editor/App';
 import { tinyMceAdd } from 'newsletter_editor/tinymce/mailpoet_shortcodes.js';
 
+
```
rodrigoprimo: I guess this extra newline was added by accident?

mailpoet/assets/js/src/newsletter_editor/behaviors/TextEditorBehavior.ts
```diff
@@ -82,14 +80,14 @@ BL.TextEditorBehavior = Marionette.Behavior.extend({
...
         plugins: this.options.plugins,
 
-        setup: function setup(editor) {
+        setup: (editor) => {
           // Store the editor instance
-          that.tinymceEditor = editor;
-          editor.on('change', function onChange() {
-            that.view.triggerMethod('text:editor:change', editor.getContent());
+          this.tinymceEditor = editor;
+          editor.on('change', () => {
```
rodrigoprimo, line 86: Does converting the file to TS require replacing `function foo() {}` with `() => {}` or do we follow some standard to use the latter instead of the former?

I don't think you need to change this, but I'm asking mostly because I'm not a fan of the arrow function syntax in most cases as it makes it a bit harder to read the code in my opinion. My impression is that any developer, regardless of their programming language background, can easily understand `function foo() {}` or even `function() {}`, but the same is not true for arrow functions.

rodrigoprimo (commented):
Looks good to me. Good job!

I left two inline comments, but they are just details and not blockers.

mailpoet/assets/js/src/newsletter_editor/behaviors/TextEditorBehavior.ts
samnajian: converting to ts doesn't directly require converting these functions to arrow function, but in the former code were defining [that](https://github.com/mailpoet/mailpoet/pull/4610/files#diff-ef944cf4cc1566facf2c6d659dacb51dd061c3630a8c9ca7c9c40ccbabd590a5L46)  to be able to use the `this` from the upper context, but this kind of context binding is not allowed in eslint rules defined for our ts files, so I had two options:
1) Use arrow function, since arrow functions automatically are bound to their parent
2) Keep these functions and bind context using `bind(this)`
I chose the first option since it's more concise and ES6+ way of binding to the parent context.
Arrow functions are now more or less the new norm, at least in ES6+ JS world, the only justification is not using them can be that we can name the older function syntax, but we miss this possibility when using arrow functions.

mailpoet/assets/js/src/newsletter_editor/behaviors/TextEditorBehavior.ts
samnajian: I'm a bit confused with this comment, since no newline has been added to `line 12` as I see it. https://d.pr/i/J5WoIh

rodrigoprimo: Ah, I see what happened here. I reviewed this PR commit by commit, and I made this comment when an extra newline was added in https://github.com/mailpoet/mailpoet/pull/4610/commits/3a6740342becab43099dde22960a1219e55db5a9#diff-ef944cf4cc1566facf2c6d659dacb51dd061c3630a8c9ca7c9c40ccbabd590a5R12. But I missed that this newline was removed in the next commit: https://github.com/mailpoet/mailpoet/pull/4610/commits/7bec395200d19dae7d32853a4a4131050883360c#diff-ef944cf4cc1566facf2c6d659dacb51dd061c3630a8c9ca7c9c40ccbabd590a5L12

mailpoet/assets/js/src/newsletter_editor/behaviors/TextEditorBehavior.ts
rodrigoprimo: Thanks for taking the time to provide the additional context and for explaining why you opted to use arrow functions.

rodrigoprimo (approved):

Aschepikov (approved):
QA verified, gj!

PR #4617, opened 2022-12-16
Separate text domains in phpcs [MAILPOET-4913]

websupporter:
## Description
Introduces two ruleset defintions and a shared one. The free plugin will use `free-ruleset.xml` while the premium will use `premium-ruleset.xml`

## Code review notes
I was thinking about changing the `phpcs` command itself in the RoboFile, but I think it is better to operate with those distinct ruleset.xml. This helps to keep all rules at one place and I think it is clearer to follow than having some rules in a ruleset and other defintions in the command line.

## QA notes

Just changing the definition of the coding standards. Nothing for QA here.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/669

## Linked tickets
[MAILPOET-4913]

## After-merge notes

_N/A_

[MAILPOET-4913]: https://mailpoet.atlassian.net/browse/MAILPOET-4913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


johnolek (dismissed):
@websupporter This looks good to me!

Should remove the `WordPress.WP.I18n` rule definition from `shared-ruleset.xml` since it's not getting used?

I also have a more general question: is it necessarily a bad thing to allow the `mailpoet` text domain in the premium plugin? Couldn't there be cases where we would want to re-use already-translated strings from the free plugin (and make sure they stay in sync)? And it's not possible to use premium without free so it seems like it would be safe.

If you think it makes sense to remove the rule from the shared ruleset please go ahead with that, otherwise feel free to merge.

websupporter:
@johnolek 

> Is it necessarily a bad thing to allow the mailpoet text domain in the premium plugin? Couldn't there be cases where we would want to re-use already-translated strings from the free plugin (and make sure they stay in sync)? And it's not possible to use premium without free so it seems like it would be safe.

It is not necessarily a bad thing. We allow for example for the `woocommerce` text domain. History showed that we often use `mailpoet` in premium where we should use `mailpoet-premium` and we had to fix those issues in different PRs, e.g. https://github.com/mailpoet/mailpoet-premium/pull/657

This is why we decided to not allow `mailpoet` in the premium plugin.

[1aa9359](https://github.com/mailpoet/mailpoet/pull/4617/commits/1aa93593667b5f767c5df41ee8d0b5fc86dff7a3) removes the text domain rule from the shared ruleset.

johnolek (approved):

johnolek:
@websupporter Thanks for the update and explanation! I'm going to leave it up to you when you want to merge these PRs.

PR #4618, opened 2022-12-16
Improve comparison stability in test [MAILPOET-4886]

lysyjan:
## Description

_N/A_

## Code review notes

I replaced two comparisons with one that should be more stable because it uses rounded seconds.

## QA notes

QA can be skipped because it's only a tiny change in a test.

## Linked tickets

[MAILPOET-4886]

[MAILPOET-4886]: https://mailpoet.atlassian.net/browse/MAILPOET-4886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo (approved):
👍

PR #4620, opened 2022-12-22
Feat/create an empty landing page - MAILPOET-4795

triple0t:
## Description

This PR creates an empty landing page.

Content will be added in upcoming PRs.

## Code review notes

You can access the page here: `wp-admin/admin.php?page=mailpoet-landingpage`

You need to enable experimental features here: `/wp-admin/admin.php?page=mailpoet-experimental`

## QA notes

Hmm 🤔, The easiest way to test this PR would be with a new site (site without previous MailPoet install), but the feature is currently hidden behind an experimental flag, so it would not work straight out of the box.

Testing steps:
* Set `version` to NULL on `mailpoet-settings` table (from the Database) 
<details>
<summary>SQL query</summary>
<p>

```sql

 UPDATE `wp_mailpoet_settings` SET
`value` = NULL,
`updated_at` = now()
WHERE `name` = 'version';

```
</p>
</details>

* Visit `/wp-admin/admin.php?page=mailpoet-experimental`, enable **Landingpage**
* Visit the plugins page (`/wp-admin/plugins.php`). Deactivate the MailPoet plugin
* Activate the MailPoet plugin
* You should be redirected to the landing page

## Linked PRs

* **First**: https://github.com/mailpoet/mailpoet/pull/4620
* Second: https://github.com/mailpoet/mailpoet/pull/4626
* Third: https://github.com/mailpoet/mailpoet/pull/4628

## Linked tickets

[MAILPOET-4795](https://mailpoet.atlassian.net/browse/MAILPOET-4795)

[MAILPOET-4795]: https://mailpoet.atlassian.net/browse/MAILPOET-4795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Changelog.php
+++ b/mailpoet/lib/Config/Changelog.php
@@ -109,6 +109,8 @@ class Changelog {
   }
 
   public function redirectToLandingPage() {
+    if (isset($_GET['activate-multi'])) return; // do not redirect when activated with bulk activation mode
+
     if ($this->shouldShowLandingPage() && !$this->isLandingPage()) {
       $this->urlHelper->redirectTo(
         $this->wp->adminUrl('admin.php?page=mailpoet-landingpage')
--- a/mailpoet/lib/Config/Changelog.php
+++ b/mailpoet/lib/Config/Changelog.php
@@ -121,11 +121,26 @@ class Changelog {
   public function maybeRedirectToLandingPage() {
     if ($this->isWelcomeWizardPage()) return; // do not redirect when on welcome wizard page
 
-    $this->redirectToLandingPage();
+    if ($this->shouldShowLandingPage()) {
+      $this->redirectToLandingPage();
+      return;
+    }
+
+    $this->maybeRedirectToWelcomeWizard(); //TODO: Remove when landing page is out of experimental
   }
 
   private function setupNewInstallation() {
     $this->settings->set('show_congratulate_after_first_newsletter', true);
+  }
+
+  private function maybeRedirectToWelcomeWizard() {
+    if ($this->shouldShowWelcomeWizard() && !$this->isWelcomeWizardPage()) {
+      if ($this->isLandingPage()) return; // do not redirect when on landing page
+
+      $this->urlHelper->redirectWithReferer(
+        $this->wp->adminUrl('admin.php?page=mailpoet-welcome-wizard')
+      );
+    }
   }
 
   private function isWelcomeWizardPage() {
--- a/mailpoet/lib/Config/Changelog.php
+++ b/mailpoet/lib/Config/Changelog.php
@@ -74,7 +74,7 @@ class Changelog {
     $version = $this->settings->get('version');
     if ($version === null) {
       $this->setupNewInstallation();
-      $this->maybeRedirectToWelcomeWizard();
+      $this->maybeRedirectToLandingPage();
     }
     $this->checkWooCommerceListImportPage();
     $this->checkRevenueTrackingPermissionPage();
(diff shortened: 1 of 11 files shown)


mailpoet/lib/Config/Changelog.php
```diff
@@ -98,12 +108,24 @@ public function shouldShowRevenueTrackingPermissionPage() {
       && $this->wp->currentUserCan('administrator');
   }
 
+  public function redirectToLandingPage() {
+    if (isset($_GET['activate-multi'])) return; // do not redirect when activated with bulk activation mode
+
+    if ($this->shouldShowLandingPage() && !$this->isLandingPage()) {
```
samnajian, line 114: `$this->shouldShowLandingPage()` is checked in the caller as well, was wondering if we need both?

mailpoet/lib/Config/Initializer.php
```diff
@@ -253,11 +255,17 @@ public function init() {
...
+      $this,
+      'afterPluginActivation',
+    ]);
+
     $this->hooks->initEarlyHooks();
   }
 
   public function runActivator() {
     try {
+      $this->wpFunctions->addOption(self::PLUGIN_ACTIVATED, true);
```
samnajian: Feels like a bit too soon to set plugin activated, maybe `Activator::processActivate` is a better place? and since we are deleting this option in `afterPluginActivation`, wouldn't `WPFunctions::doAction` be a better replacement, as in `afterPluginActivation` the `mailpoet_plugin_activated` option finally is deleted while the plugin is still active, I'm afraid it may get confusing in future, what do you think?

mailpoet/lib/Config/Changelog.php
triple0t: Hmm, True. I could remove it from the caller, but we do need it here

mailpoet/lib/Config/Initializer.php
triple0t: > Feels like a bit too soon to set plugin activated, maybe Activator::processActivate is a better place?

No, I believe this is okay.
The action is not executed here (on registerActivationHook); it is done in [the admin_init](https://github.com/mailpoet/mailpoet/blob/47ea982eb44984dddacfe7398a938cf7561b697f/mailpoet/lib/Config/Initializer.php#L258-L261) hook.


> since we are deleting this option in afterPluginActivation, wouldn't WPFunctions::doAction be a better replacement, as in afterPluginActivation the mailpoet_plugin_activated option finally is deleted while the plugin is still active, I'm afraid it may get confusing in future, what do you think?

The `mailpoet_plugin_activated` option acts as a flag here to indicate a plugin activation action.

From [WP Docs](https://developer.wordpress.org/reference/functions/register_activation_hook/#process-flow), it is recommended not to perform the type of action we are trying to do here (redirect to another page) on `registerActivationHook` (which includes `Activator::processActivate` as well).

> If you are interested in doing something just after a plugin has been activated it is important to note that the hook process performs an instant redirect after it fires. So it is impossible to use [add_action()](https://developer.wordpress.org/reference/functions/add_action/) or [add_filter()](https://developer.wordpress.org/reference/functions/add_filter/) type calls until the redirect has occurred

While it's possible to use `WPFunctions::doAction`, It would not work for our case as the `registerActivationHook` requires the  instant redirect to be successful. That would not happen if we hook to `WPFunctions::doAction` on `registerActivationHook` and redirect the user to the landing page.

This process is more of a walkaround for getting it to work correctly.

> I'm afraid it may get confusing in future

Would you be okay with linking to this conversation and the [WP Docs](https://developer.wordpress.org/reference/functions/register_activation_hook/#process-flow) in a comment on the code?

samnajian: Thanks for the elaborate explanation @triple0t, I think adding a link to the conversion on `WP Docs` sounds like a good idea to reduce future confusion.

samnajian (dismissed):
LGTM, thanks for the updates.

Aschepikov:
Hi @triple0t When I am activating the plugin and redirecting to the landing page I am stuck on it and I can't load any other mailpoet page or do something else in the plugin. Is it expected behavior for this ticket when a version is NULL on mailpoet-settings table? I can't test it with fresh installing because of the experimental feature so it is hard to say for me that it will work ok with a real case. Wizard process on this site is finished.

Aschepikov:
https://user-images.githubusercontent.com/113185437/211064799-bc81c8a1-6488-4a65-9495-431e0d4ba6c4.mp4

Aschepikov:
Ok, I understood how it works. QA verified, GJ!

Aschepikov (approved):
QA verified, GJ!

mailpoet/lib/Config/Initializer.php
```diff
@@ -319,6 +327,22 @@ public function initialize() {
...
+   * and https://github.com/mailpoet/mailpoet/pull/4620#discussion_r1058210174
+   * @return void
+   */
+  public function afterPluginActivation() {
+    if (!$this->wpFunctions->isAdmin() || !defined(self::INITIALIZED) || !$this->wpFunctions->getOption(self::PLUGIN_ACTIVATED)) return;
+
+    $this->changelog->redirectToLandingPage();
+
+    // done with afterPluginActivation actions
+    $this->wpFunctions->deleteOption(self::PLUGIN_ACTIVATED);
```
fseabra47am, line 343: @triple0t the method `redirectToLandingPage` is exiting the code earlier and this method - `deleteOption` - isn't getting reached and the option persists.
This behavior makes the redirect to "/wp-admin/admin.php?page=mailpoet-landingpage" occurs whenever any plugin is activated.
I've opened an issue: https://github.com/mailpoet/mailpoet/issues/5014

triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/5020
Thank you, @fseabra47am

PR #4627, opened 2022-12-30
Use wordpress i 18 n in all of our codebase [MAILPOET-4908]

samnajian:
## Description

This PR uses @wordpress/i18n package in the js/ts code related to Newsletters and the common code ( the directory mailpoet/assets/js/src/common ). 

## Code review notes
There will be a follow up ticket to take care of rest of the codebase.

## QA notes

Please switch your site's language to a language that has a good coverage and check if the plugin remains translated on different pages

## Linked tickets

[MAILPOET-4908]

## After-merge notes

Please run the following commands to keep the translations current for the release to work fine:
- ` ./do translations:build `
- ` ./do translations:push `
- Run `./do release:version-get-next` to get the next version to use in the next step
- `./do translations:prepare-language-packs {version from previous step}`

[MAILPOET-4908]: https://mailpoet.atlassian.net/browse/MAILPOET-4908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/RoboFile.php
```diff
@@ -153,7 +153,7 @@ public function translationsBuild() {
       ->taskExec("php -d memory_limit=-1 tasks/makepot/makepot-views.php . > lang/mailpoet.pot")
 
       // PHP, JS/TS
-      ->taskExec("vendor/bin/wp i18n make-pot --merge --slug=mailpoet --domain=mailpoet --exclude=$exclude --headers=$headers . lang/mailpoet.pot")
+      ->taskExec("php -d memory_limit=-1 $(which wp) i18n make-pot --merge --slug=mailpoet --domain=mailpoet --exclude=$exclude --headers=$headers . lang/mailpoet.pot")
```
JanJakes: @samnajian Hey! Sorry to jump in, I was just browsing through and this caught my eye. Apart from fixing the memory limit, this seems to break the command anywhere outside Docker container by changing `vendor/bin/wp` to `$(which wp)`. Could we revert that change? Thanks!

mailpoet/assets/js/src/newsletters/listings/re_engagement.jsx
```diff
@@ -260,9 +267,13 @@ class NewsletterListReEngagementComponent extends Component {
...
 
-    const sendingFrequency = MailPoet.I18n.t('reEngagementSettings')
+    const sendingFrequency = _x(
+      '{$count} {$frequency} after inactivity',
+      'example: "5 months after inactivity"',
+      'mailpoet',
+    )
       .replace('{$count}', newsletter.options.afterTimeNumber)
-      .replace('{$frequency}', MailPoet.I18n.t(frequencyKey));
+      .replace('{$frequency}', __(frequencyKey, 'mailpoet'));
```
lysyjan: When there is used the variable `frequencyKey` I think it should contain messages instead of the old translation keys. For example, the initialization should be changed to `let frequencyKey = _x('month', 'month in the sentence "1 month after inactivity"')`. Do you agree or I miss something?

mailpoet/assets/js/src/newsletters/send.tsx
```diff
@@ -460,8 +480,8 @@ class NewsletterSendComponent extends Component<
           ) {
             this.context.notices.success(
               <p>
-                {MailPoet.I18n.t('automaticEmailActivated').replace(
-                  '%1s',
+                {sprintf(
```
lysyjan, line 486: I want to ask why you used `sprintf` instead of `replace`, as we use it in other places.

mailpoet/assets/js/src/newsletters/send/congratulate/success_pitch_mss.tsx
```diff
@@ -98,25 +108,13 @@ export function PitchMss(props: Props): JSX.Element {
               window.open(props.purchaseUrl);
               next();
             }}
-          >
```
lysyjan, line 97: You removed here the button with the thanks message. I want to ask you why? Because there is no comment and it is a part of the huge commit.

mailpoet/assets/js/src/newsletter_editor/behaviors/MediaManagerBehavior.js
```diff
@@ -63,6 +63,13 @@ BL.MediaManagerBehavior = Marionette.Behavior.extend({
 
     MediaManager = window.wp.media.view.MediaFrame.Select.extend({
       initialize: function () {
+        /**
+         * As of importing @wordpress/i18n, and since it's using lodash, we need to reset window._ back to
+         * underscore for the media manage to work.
+         * The media manager is using _.each and changes the context using the 3 parameter, however _.each of
+         * lodash doesn't accept a third parameter.
+         */
+        window._ = _;
```
jsnajdr, line 72: This issue is caused not by using `underscore` by itself, but by exposing the module as `window._` in the webpack config, using the `expose-loader` plugin. I'm curious why MailPoet is doing this, instead of just using underscore internally, without making it visible outside its webpack bundle?

Also, `underscore` is a standart part of WordPress. MailPoet could externalize its usages, and specify it as a dependency when registering MailPoet scripts.

jsnajdr: Since version 4.17.0, `@wordpress/i18n` doesn't depend on Lodash any more. By upgrading in the `package.json`, from the currently used 4.14.0, you could eliminate this Lodash issue completely.

mailpoet/mailpoet.php
```diff
@@ -2,7 +2,7 @@
 
 /*
  * Plugin Name: MailPoet
- * Version: 4.8.0
+ * Version: 4.8.1-beta.1
```
lysyjan: I want to ask why those changes. I assume for testing, but I want to be sure.

mailpoet/assets/js/src/subscribers/subscribers.jsx
```diff
@@ -30,5 +30,6 @@ function App() {
 const container = document.getElementById('subscribers_container');
 
 if (container) {
+  registerTranslations();
```
lysyjan, line 33: I'm wondering if there is some system solution for registering translations. Do you have any idea?
When I'm checking this commit, there is a lot of code repetition. At the same time, I'm thinking about what can happen when a developer forgets to add calling this function.

mailpoet/assets/js/src/newsletters/send/congratulate/success_pitch_mss.tsx
lysyjan: This was addressed later in another commit. You can ignore it.

lysyjan (commented):
Good job Sam! It's a huge pull request, and I tried to check it thoroughly.
So please take a look at my minor remarks and questions and also at the comments from other developers.

mailpoet/mailpoet.php
samnajian: It's added for local and CI to work properly, and I left it in the PR since it won't do any harm and the triggering of language-pack preparation, i.e. `./do translations:prepare-language-packs {version}` needs a version, for which this version is used so that the local/CI testing doesn't update the translations on translate.wordpress.com with the ones for this PR.

mailpoet/assets/js/src/subscribers/subscribers.jsx
samnajian: basically, this is needed per page/application, and since we have a single application per each menu entry, this is needed, but we need to do it once per each menu entry, so while I see your concern, the probability of it happening often is very low.

mailpoet/assets/js/src/newsletters/send.tsx
samnajian: I could have left it like that, but since this `sprintf` is provided by the `@wordpress/i18n` package and mentioned in examples, I wanted to give it a try and mention it here, also a ticket is created to adjust our current translations with custom placeholders and start using `sprintf`, this will enable us to get rid of https://www.npmjs.com/package/react-string-replace from our dependencies and start using [a package  that's already installed](https://www.npmjs.com/package/sprintf-js)

mailpoet/assets/js/src/newsletters/listings/re_engagement.jsx
samnajian: thanks for catching it, updated in 768e7b6e77be21e57ca468246810fd384c34b581

mailpoet/assets/js/src/newsletters/send.tsx
jsnajdr: `sprintf` is a standard companion of `__` in the `@wordpress/i18n` library, and it's also used in Gutenberg, so the migration makes sense. If it's systematic, replacing all `.replace` calls.

mailpoet/assets/js/src/newsletters/listings/re_engagement.jsx
```diff
@@ -249,22 +249,31 @@ class NewsletterListReEngagementComponent extends Component {
       (match, i) => <SegmentTags segments={newsletter.segments} key={i} />,
     );
 
-    let frequencyKey = 'reEngagementFrequencyMonth';
+    let frequency = _x(
```
lysyjan, line 252: I have noticed @samnajian  that you didn't use the third parameter `domain` here.
It is not used also in other calls in this commit.

samnajian: ops! I had done a double check for other commits, but copy/pasted these from template files and missed adding the domain! thanks for the catch. Add them in fc504c2ac95b28a4ff39b93751b80c8bcddd02e5

mailpoet/assets/js/src/newsletter_editor/behaviors/MediaManagerBehavior.js
samnajian: As we have another [PR](https://github.com/mailpoet/mailpoet/pull/4786) taking care of the dependencies update, I opted to leave this update to that PR and then adjust after that PR is merged to trunk in a follow up PR, specially since we have 2 instances of such a workaround.

lysyjan (dismissed):
Great job! Thank you!

samnajian:
(i) Latest forced push interactive rebase was to fix the merge conflicts

mailpoet/assets/js/src/newsletter_editor/behaviors/MediaManagerBehavior.js
JanJakes: @samnajian @jsnajdr The [Gutenberg update PR](https://github.com/mailpoet/mailpoet/pull/4786) was already merged so let's try to remove that workaround.

mailpoet/assets/js/src/newsletters/send.tsx
JanJakes: Yeah, we could actually create a ticket to adopt `sprintf` everywhere. But maybe it makes even more sense to accompany it with the adoption of [@wordpress/element](https://github.com/WordPress/gutenberg/blob/trunk/packages/element/README.md) and using [createInterpolateElement](https://github.com/WordPress/gutenberg/blob/trunk/packages/element/README.md#createinterpolateelement) instead of [react-string-replace](https://github.com/iansinnott/react-string-replace). Or we can create two separate tickets.

mailpoet/assets/js/src/newsletter_editor/behaviors/MediaManagerBehavior.js
JanJakes: @jsnajdr Thanks for the details!

> I'm curious why MailPoet is doing this, instead of just using underscore internally, without making it visible outside its webpack bundle?

Good question. We'll need to investigate, it's probably been like that for ages.

> Also, underscore is a standart part of WordPress. MailPoet could externalize its usages, and specify it as a dependency when registering MailPoet scripts.

Makes sense, I think with underscore, this would work well. We're not quite there yet to externalize all of our `@wordpress` deps but getting closer by updating our deps and adopting better WP/WC version support policies (L-2 or even L-1) where we could reasonably cover them with test automation.

mailpoet/assets/js/src/newsletters/send.tsx
samnajian: I've already [created](https://mailpoet.atlassian.net/browse/MAILPOET-5117) a ticket for such cases , it entails changing the placeholders in the translatable strings, other than that I would have replaced the `.replace` with `sprintf` in this PR.

mailpoet/assets/js/src/newsletter_editor/behaviors/MediaManagerBehavior.js
samnajian: > @samnajian @jsnajdr The [Gutenberg update PR](https://github.com/mailpoet/mailpoet/pull/4786) was already merged so let's try to remove that workaround.

I'd do it in a separate PR, and base it on this branch, thanks for the heads up.

JanJakes: @samnajian Sure, anything you prefer. Just unfortunately, merging of the upgrade caused conflicts in this PR.

Aschepikov (approved):
GJ!

PR #4631, opened 2023-01-03
Improve logging errors for JSON API [MAILPOET-4104]

lysyjan:
## Description

Because sometimes can be hard to debug our plugin, we want to improve API logs. Those changes store an error stack in our log table and the PHP error log.

## Linked tickets

[MAILPOET-4104]

[MAILPOET-4104]: https://mailpoet.atlassian.net/browse/MAILPOET-4104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/JSON/API.php
+++ b/mailpoet/lib/API/JSON/API.php
@@ -4,6 +4,7 @@ namespace MailPoet\API\JSON;
 
 use MailPoet\Config\AccessControl;
 use MailPoet\Exception;
+use MailPoet\Logging\LoggerFactory;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Subscription\Captcha\CaptchaConstants;
 use MailPoet\Tracy\ApiPanel\ApiPanel;
@@ -42,6 +43,9 @@ class API {
   /** @var SettingsController */
   private $settings;
 
+  /** @var LoggerFactory */
+  private $loggerFactory;
+
   const CURRENT_VERSION = 'v1';
 
   public function __construct(
@@ -49,6 +53,7 @@ class API {
     AccessControl $accessControl,
     ErrorHandler $errorHandler,
     SettingsController $settings,
+    LoggerFactory $loggerFactory,
     WPFunctions $wp
   ) {
     $this->container = $container;
@@ -62,6 +67,7 @@ class API {
         $availableApiVersion
       );
     }
+    $this->loggerFactory = $loggerFactory;
   }
 
   public function init() {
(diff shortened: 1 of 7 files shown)


JanJakes (commented):
At a first sight, it seemed to me that the ticket doesn't follow the specs — it's all about using `error_log` but this implements using the logging table. It also differs from what's implemented on the WP REST API (and doesn't unify it).

I see that the `error_log` was added in the last commit; however, I still feel like the part that uses the logger service might be problematic. Both the [ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4104) and the [original post](https://mailpoetdev.wordpress.com/2022/01/19/visibility-of-php-errors-in-our-json-api/) seem to focus on using a PHP-native error log rather than our logging table. It was even briefly discussed in the comments.

So it feels to me that using our logging mechanism may go beyond what was proposed before, and maybe we should discuss that. I see some potential problems there:
1. It may result in a large number of logs. That said, sending can probably cause that too. But in the case of the API, some malicious users or bots could do that.
2. Saving the full request payload essentially means anybody can save anything to your WP database. That opens the doors to very easy DDoS attacks, database storage overflows, etc.

I think I'd prefer sticking to the `error_log` approach but we can also see what others think about this.

lysyjan:
@JanJakes Thank you for your insight!

I agree with your arguments. But I read the comments in the P2 post, and I still think the possibility of detailed logging on API can be helpful. We could change the log level to the warning. That would allow us to reduce the risk because the feature would be active only when a user or L2 engineer changes the logging settings in our plugin. What do you think?

I'm okay with using the error_log function only if you consider it too dangerous.

JanJakes:
> We could change the log level to the warning.

@lysyjan That's a great idea! So when the log level is "Errors only," warnings are not included?

I like that idea. Additionally — do you think we should store the whole request payload or would the exception be enough? What do you think? No strong opinions either way.

lysyjan:
> So when the log level is "Errors only," warnings are not included?

@JanJakes Exactly. Warnings can be enabled by the logging option `Everything`.

> Additionally — do you think we should store the whole request payload or would the exception be enough? What do you think?

I think that an exception is not enough. I prefer the whole stack trace, but because it can be too much, we can log only the other values and add the stack trace later when it's needed.

JanJakes:
@lysyjan Logging stack trace is fine, I think. What I meant is if we need to log the request payload.

lysyjan:
@JanJakes I changed the level logging into warnings and removed the request data from the context. All changes are added as new commits and start at commit https://github.com/mailpoet/mailpoet/pull/4631/commits/641ccff213e7722df589caf6b9a2eb2e1e5406c4.

JanJakes (approved):

Aschepikov (approved):
GJ!

PR #4633, opened 2023-01-04
Use translate function on errors from bridge [MAILPOET-4639]

lysyjan:
## Description

This PR uses the WP translate function on errors from the bridge. It should explain to our users, who use a different language, where is the exact issue with sending.

## Code review notes

I tried to find all possible MSS errors that can be rendered in messages, but if I missed something, please let me know.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4639]

## After-merge notes

_N/A_

[MAILPOET-4639]: https://mailpoet.atlassian.net/browse/MAILPOET-4639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Services/Bridge/API.php
+++ b/mailpoet/lib/Services/Bridge/API.php
@@ -152,12 +152,7 @@ class API {
       $response = ($this->wp->wpRemoteRetrieveBody($result)) ?
         $this->wp->wpRemoteRetrieveBody($result) :
         $this->wp->wpRemoteRetrieveResponseMessage($result);
-      return [
-        'status' => self::SENDING_STATUS_SEND_ERROR,
-        'error' => $response,
-        'message' => $this->getTranslatedErrorMessage($response),
-        'code' => $responseCode,
-      ];
+      return $this->createErrorResponse((int)$responseCode, $response, self::SENDING_STATUS_SEND_ERROR);
     }
     return ['status' => self::RESPONSE_STATUS_OK];
   }
@@ -231,12 +226,8 @@ class API {
       // translators: %d is the error code.
       $fallbackError = sprintf(__('An error has happened while performing a request, the server has responded with response code %d', 'mailpoet'), $responseCode);
 
-      return [
-        'status' => self::RESPONSE_STATUS_ERROR,
-        'code' => $responseCode,
-        'error' => is_array($errorResponseData) && isset($errorResponseData['error']) ? $errorResponseData['error'] : $fallbackError,
-        'message' => is_array($errorResponseData) && isset($errorResponseData['error']) ? $this->getTranslatedErrorMessage($errorResponseData['error']) : $fallbackError,
-      ];
+      $error = is_array($errorResponseData) && isset($errorResponseData['error']) ? $errorResponseData['error'] : $fallbackError;
+      return $this->createErrorResponse((int)$responseCode, $error);
     }
 
     return ['status' => self::RESPONSE_STATUS_OK];
@@ -292,12 +283,8 @@ class API {
       // translators: %d will be replaced by an error code
       $fallbackError = sprintf(__('An error has happened while performing a request, the server has responded with response code %d', 'mailpoet'), $responseCode);
 
-      return [
-        'status' => self::RESPONSE_STATUS_ERROR,
-        'code' => $responseCode,
-        'error' => is_array($responseBody) && isset($responseBody['error']) ? $responseBody['error'] : $fallbackError,
-        'message' => is_array($responseBody) && isset($responseBody['error']) ? $this->getTranslatedErrorMessage($responseBody['error']) : $fallbackError,
-      ];
+      $error = is_array($responseBody) && isset($responseBody['error']) ? $responseBody['error'] : $fallbackError;
+      return $this->createErrorResponse((int)$responseCode, $error);
     }
 
     if (!is_array($responseBody)) {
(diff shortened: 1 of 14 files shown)


costasovo (changes requested):
@lysyjan  Good job! 
The PR covers errors from the messages API. But when looking into [the ticket description](https://mailpoet.atlassian.net/browse/MAILPOET-4639) we should also add translations for the newly added endpoints that the bridge proxies to the shop (adding sender domains, adding authorized emails).

These errors are processed in the Bridge API class ([create sender domain](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Services/Bridge/API.php#L272), [create auth. email](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Services/Bridge/API.php#L214), etc.)
These requests [are proxied to the shop](https://github.com/mailpoet/services-bridge/blob/master/middleware/shop_proxy.rb) so you need to look for the error messages in the shop's codebase (some [are in the route file](https://github.com/mailpoet/shop/blob/master/routes/hooks/sending/v1/index.js#L159), [for some you need to look into services](https://github.com/mailpoet/shop/blob/master/services/user/authorized_sending_email.js#L118-L134)).

If you prefer, this could also be done in a follow-up ticket/PR.

lysyjan:
@costasovo I tried to rework this PR from the very beginning. However, I could overlook something because I didn't work on the new features with the activation email/domain yet.

costasovo (commented):
Good job! Thanks a lot for improving the error formats 👍

I have a couple of ideas that I believe could improve it a bit further. Please see my comments. 
What do you think? If you think they are good, we don't need to add those to this PR we can create an additional ticket.

mailpoet/lib/Services/Bridge/API.php
```diff
@@ -7,6 +7,10 @@
 use WP_Error;
 
 class API {
+  const AUTHORIZED_EMAIL_STATUS_OK = 'ok';
+  const AUTHORIZED_EMAIL_STATUS_ERROR = 'authorized_email_error';
+  const AUTHORIZED_DOMAIN_STATUS_OK = 'ok';
+  const AUTHORIZED_DOMAIN_STATUS_ERROR = 'authorized_domain_error';
   const SENDING_STATUS_OK = 'ok';
   const SENDING_STATUS_CONNECTION_ERROR = 'connection_error';
   const SENDING_STATUS_SEND_ERROR = 'send_error';
```
costasovo: Just an idea. Do we need too many special `ok` statuses? I mean the caller knows which method was called so maybe we could live with a common `ok` status shared by all endpoints. We could also have a common error status shared by all endpoints + a couple of special error statuses if needed.

I think we could change:
```php
  const AUTHORIZED_EMAIL_STATUS_OK = 'ok';
  const AUTHORIZED_EMAIL_STATUS_ERROR = 'authorized_email_error';
  const AUTHORIZED_DOMAIN_STATUS_OK = 'ok';
  const AUTHORIZED_DOMAIN_STATUS_ERROR = 'authorized_domain_error';
  const SENDING_STATUS_OK = 'ok';
  const SENDING_STATUS_CONNECTION_ERROR = 'connection_error';
  const SENDING_STATUS_SEND_ERROR = 'send_error';
```

to:
```php
  const RESPONSE_STATUS_OK = 'ok';
  const RESPONSE_STATUS_ERROR = 'error';
  const SENDING_STATUS_CONNECTION_ERROR = 'connection_error';
  const SENDING_STATUS_SEND_ERROR = 'send_error';
```

mailpoet/lib/Services/Bridge/API.php
```diff
@@ -196,26 +220,29 @@ public function createAuthorizedEmailAddress(string $emailAddress): array {
...
+      $fallbackError = sprintf(__('An error has happened while performing a request, the server has responded with response code %d', 'mailpoet'), $responseCode);
 
-      $errorData = is_array($errorResponseData) && isset($errorResponseData['error']) ? $errorResponseData['error'] : $fallbackError;
-      return ['error' => $errorData, 'status' => false];
+      return [
+        'status' => self::AUTHORIZED_EMAIL_STATUS_ERROR,
+        'code' => $responseCode,
+        'error' => is_array($errorResponseData) && isset($errorResponseData['error']) ? $errorResponseData['error'] : $fallbackError,
+        'message' => is_array($errorResponseData) && isset($errorResponseData['error']) ? $this->getTranslatedErrorMessage($errorResponseData['error']) : $fallbackError,
+      ];
```
costasovo: I see we construct the array of this shape in a couple of places. I really like the idea of a unified shape of the error response. 

I think we could introduce a private method (e.g. `formatErrorResponse` or some better name). This would increase the probability that methods we add in the future will use the same shape because the developer will see there is a method for formatting errors.

The translation mapping returns the original error when there is no translation so the `formatErrorResponse` method could handle adding the `message` key with translation. Also, with the common response status (see previous comment), it could have the status parameter optional defaulting to `self::RESPONSE_STATUS_ERROR`

mailpoet/lib/Services/Bridge/API.php
lysyjan: I will change it. Good idea. I was considering something similar but didn't have the final idea.

mailpoet/lib/Services/Bridge/API.php
lysyjan: Good point. I will address it in a new commit. 👍

costasovo (approved):
Thanks! 👍

Aschepikov (approved):
GJ!

PR #4635, opened 2023-01-06
Fix switch languages test [MAILPOET-4909]

costasovo:
## Description

This PR adds logging of failed calls when fetching translation data from translate.wordpress.com fails and adds per-request caching of the API calls.

The caching should fix the flakiness issue caused by the occasional 429 errors. When debugging the calls we made during the test, I found that we called the API four times, and it was called twice per request because of the WP action we hooked into. This should also reduce unnecessary calls for our users.

## Code review notes

_N/A_

## QA notes
Please test that switching languages works as expected.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4909]

## After-merge notes

_N/A_

[MAILPOET-4909]: https://mailpoet.atlassian.net/browse/MAILPOET-4909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/TranslationUpdater.php
+++ b/mailpoet/lib/Config/TranslationUpdater.php
@@ -26,6 +26,9 @@ class TranslationUpdater {
   /** @var string|null */
   private $premiumVersion;
 
+  /** @var array  */
+  private $requestCache = [];
+
   public function __construct(
     WPFunctions $wpFunctions,
     string $freeSlug,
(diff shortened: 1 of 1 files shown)


JanJakes (approved):
@costasovo LGTM! Just one question:

> The caching should fix the flakiness issue caused by the occasional 429 errors. When debugging the calls we made during the test, I found that we called the API four times, and it was called twice per request because of the WP action we hooked into. This should also reduce unnecessary calls for our users.

Does this mean the all the calls are done correctly? In other words — the cache doesn't just hide some problematic underlying implementation, or does it?

Thanks for looking into this!

costasovo:
> Does this mean the all the calls are done correctly? In other words — the cache doesn't just hide some problematic underlying implementation, or does it?

Well. Kind of. We hook into `pre_set_site_transient_update_plugins` action which is called by WP when saving update data for plugins. We don't control how many times this is called, and in some cases, WP calls this hook two times in one request. I'm not aware of any other hook we could use instead (WooCommerce uses the same one), and at the same time, I don't want to prefetch the API data and cache it for more than one request to have the ability to regenerate packages, etc.

JanJakes:
@costasovo Thanks for the explanation! The fix makes sense to me, given we can't use any other hook or control when the method is called.

Aschepikov (approved):
GJ!

PR #4636, opened 2023-01-06
Create a coupon block in the email editor [MAILPOET-4678]

samnajian:
## Description

This PR adds simple the Coupon block and takes care of the following:
1) Adding the block to the email
2) Adding default config
3) Generating a coupon for each added block and then persisting the coupon id along with the block settings
4) Generating a coupon for `automatic` emails but not persisting the coupon id with the block so the the next rendering/send out of the email gets a new coupon.

## QA notes:
- The feature is behind a feature flag
- Please add the coupon to and email and check if the coupon is generated when it's send out and not in preview
- In preview we show only a placeholder of the coupon, unless the newsletter already has a coupon generated and assigned to it
- The validation as of now only happens on frontend as this is the simple implementation of the feature and more validation are to be added (especially on backend side) in other tickets
- The only email time that will get a dynamic coupon on sending (dynamic means the coupon is not saved to the coupon block setting, but only generated once the email is sent out), are the `automatic` email types, one example of those are `abandoned cart` emails, the rest of the email types, get a coupon code generated on the first send out, and for the following send outs the same coupon code will be used
- The coupon code that's generated should have the email subject and ID as description of the coupon, and is viewable under WooCommerce -> Coupons menu.
- 
## Linked tickets

[MAILPOET-4678]

[MAILPOET-4678]: https://mailpoet.atlassian.net/browse/MAILPOET-4678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/newsletter_editor/blocks/coupon.ts
```diff
@@ -0,0 +1,213 @@
...
+  className: 'mailpoet_block mailpoet_coupon_block mailpoet_droppable_block',
+  getTemplate: () => window.templates.couponBlock,
+  onDragSubstituteBy: () => Module.CouponWidgetView,
+  behaviors: _.extend({}, base.BlockView.prototype.behaviors, {
+    ShowSettingsBehavior: {},
+  }),
+  initialize(...args) {
+    base.BlockView.prototype.initialize.apply(this, args);
+
+    // Listen for attempts to change all dividers in one go
```
costasovo: Shouldn't `all dividers` be `all coupons` here?

mailpoet/assets/js/src/global.d.ts
```diff
@@ -183,4 +183,5 @@ interface Window {
     subscribers_count: number;
   };
   templates?: Record<string, string>;
+  is_wc_active?: boolean;
```
costasovo, line 186: We already have that on window see https://github.com/mailpoet/mailpoet/blob/a6bb3809ca363e2902c2eb42602845f266c0358f/mailpoet/assets/js/src/global.d.ts#L92
This one is also available on MailPoer modude under `MailPoet.isWoocommerceActive`.

Could you please use the exisitng value. Also when adding a new global value on window please prefix it with `mailpoet_` to prevent potential conflicts. 

Note: I noticed that there are also `window.templates` without the prefix. I see we have it like that historically. Would be nicer to have `mailpoet_templates` but I would say the change is out of the scope of this change.

mailpoet/lib/Newsletter/Renderer/Blocks/Coupon.php
```diff
@@ -0,0 +1,59 @@
+<?php // phpcs:ignore SlevomatCodingStandard.TypeHints.DeclareStrictTypes.DeclareStrictTypesMissing
```
costasovo: It is a new file so please add ` declare(strict_types = 1);` instead of the ignore comment for the code sniffer

mailpoet/lib/WooCommerce/CouponPreProcessor.php
```diff
@@ -0,0 +1,60 @@
...
+    $chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789';
+    $length = strlen($chars);
+    return sprintf(
+      "%s-%s-%s",
+      substr($chars, rand(0, $length - 5), 4),
+      substr($chars, rand(0, $length - 8), 7),
+      substr($chars, rand(0, $length - 5), 4)
+    );
+  }
+}
```
costasovo: Just a suggestion. We already have a util method for generating a random code. See https://github.com/mailpoet/mailpoet/blob/926620e8f82ca691f1c11f6e973a2f86a6117522/mailpoet/lib/Util/Security.php#L36 so maybe you could use this one instead of implementing own.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
```diff
@@ -0,0 +1,60 @@
...
+
+    return $coupons;
+  }
+
+  private function generateCoupon(string $couponCode = null): int {
+    $coupon = new \WC_Coupon();
+    $code = $couponCode ?? $this->generateRandomCode();
+    $coupon->set_code($code);
+
+    return $coupon->save();
```
costasovo, line 85: What happens if we accidentally generate an existing code? Will `coupon->save` throw an exception or return an error? I just want to ensure we don't end up with an error in preprocessing that would block sending.

From what I experienced when testing, when we set coupon type percent and coupon value more than 100, there will be an error.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
```diff
@@ -65,6 +66,8 @@ private function addOrUpdateCoupon(array $couponBlock, NewsletterEntity $newslet
     }
     $coupon->set_discount_type($couponBlock['discountType']);
     $coupon->set_amount($couponBlock['amount']);
+    $expiration = (new DateTime())->getCurrentDateTime()->modify("+{$couponBlock['expiryDay']} day")->getTimestamp();
+    $coupon->set_date_expires($expiration);
```
costasovo, line 75: Just want to double-check check I understand correctly. This updates the expiration each time when the plugin pre-process a newsletter. So for the welcome email, it prolongs expiration every time when the newsletter is sent. Correct? I don't think it is a big issue. Just want to make sure we prolong the expiration so that the new subscribers have a valid coupon. Maybe an integration test would be nice for that.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
```diff
@@ -91,4 +91,11 @@ private function generateRandomCode(): string {
   private function shouldGenerateCoupon(array $block): bool {
     return empty($block['couponId']);
   }
+  
+  /**
+   * For some renders/send outs the coupon id shouldn't be persisted along the coupon block
+   */
+  private function shouldPersist(NewsletterEntity $newsletter): bool {
+    return $newsletter->getType() !== NewsletterEntity::TYPE_AUTOMATIC;
```
costasovo, line 112: Aha, so that means that for automatic emails, we will generate a coupon for every sending? Correct?
I think that TYPE_AUTOMATION emails and maybe also TYPE_WELCOME potentially also  TYPE_RE_ENGAGEMENT and TYPE_CONFIRMATION_EMAIL_CUSTOMIZER have the same behavior that they can be sent after some event.

costasovo (changes requested):
Thanks for implementing this feature! Good job! This is a really big feature.

I've added some comments. 

I run into an issue when testing:

It is possible to add random values for amount. For example, when wanted to use the percent coupon type and passed value higher than 100 the sending failed with and error `Invalid discount amount`. I'm not sure but I think that the error was coming grom Woo code for saving a coupon. I was testing with a welcome email and the email ended up stuck in sending state, which is an issue because this would block the background job. We need to make sure we can recover when there is an error when saving the coupon and also, it would be nice to restrict the amount to numbers only and don't allow more than 100 for percent coupons. I noticed in the ticket comments that some validation will be done in follow-up ticket. But since this is not hidden behind a feature flag we should at least make sure that invalid values will not block sending.

Screenshot 2023-01-13 at 11 18 44

I would also suggest adding some tests for the new functionality:
1) The CouponPreProcessor contains a lot of logic that I think really needs an integration test
2) We have [unit tests for block renderers](https://github.com/mailpoet/mailpoet/tree/82aeb898540e726ec6a5d177d70d9031202d1e06/mailpoet/tests/unit/Newsletter/Renderer/Blocks) so I would consider adding test for this block as well
3) We have ["component level" test for newsletter editor blocks](https://github.com/mailpoet/mailpoet/tree/d116ff360b8d73f0ae09429ae0c365ed30a54def/mailpoet/tests/javascript_newsletter_editor/newsletter_editor/blocks). I would suggest adding tests for the new block.
4) I would add a very basic acceptance test (checking block can be added to the email content, setting panel can be opened and some value can be set and coupon saved)

mailpoet/assets/js/src/global.d.ts
samnajian: nice catch, that's a left over, will get removed.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
samnajian: That's correct, it'll throw, I was under the impression to leave this to the validation ticket, but your `it won't get released under a feature flag` is valid and I'll manage this.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
samnajian: yes, so basically the actual expiration is the one of the last send out, adding a test makes sense.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
samnajian: The implementation of this method is kept simple, mostly like a placeholder, since we have a ticket to dynamically create coupons for the newsletters you mentioned ( or even more?)

mailpoet/assets/js/src/global.d.ts
samnajian: I was also not comfortable with keeping the name `template` but wanted to refrain from adding even more changes.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
samnajian: I have no strong opinion about this, fine with using `generateRandomString` if you think that's better.

mailpoet/assets/js/src/global.d.ts
costasovo: Let's keep the template as is. It has been like that for ages.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
costasovo: Using `generateRandomString` makes sense to me so that we don't re-create similar functionality we already have.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
costasovo: Adding test is a good idea. I think it would be nice to cover whole class with an integration test. It seems that the class handles all the logic about coupon creation. I assume we will be adding more in the follow-up tasks.

mailpoet/lib/WooCommerce/CouponPreProcessor.php
costasovo: Aha. If it is a placeholder, maybe we could always return true and add a comment that in the future, it will return true for dynamic coupons. Or maybe just add a comment. When I saw the code it was suspicious. Why only automatic emails?

mailpoet/tests/unit/WooCommerce/CouponPreProcessorTest.php
```diff
@@ -24,6 +25,10 @@ public function __construct(
   ) {
     parent::__construct($name, $data, $dataName);
 
+    WordPress::interceptFunction('wp_timezone', function() {
+      new \DateTimeZone('UTC');
+    });
```
costasovo, line 30: Hmm, I was not aware we had this! TIL 👍

costasovo (dismissed):
Thanks for adding the validation and the CouponProcessor test.

Aschepikov:
Hi, @samnajian I found some issues, some of them can be connected with block settings or will be fixed in the next pr, take a look at them pls.

1. After coupon adding there is a console error
image

2. The percent symbol is moving with scrolling
image
image

3. I can see just a code of the colors but usually we can see color box like in the button settings and we have an issue with the border position
image
In the button settings it is like
image

4. Issue with the positioning of the text when there is a long name of font
image
image
It can be connected with the whole block I can make a ticket
image

5. When we are changing from the fixed cart discount to the fixed product discount and back, the coupon amount field is green, but with percentage discount is white, why
image
image
image
image
And every time when we are changing the coupon amount within validation it is green but why it is green after changing from the fixed cart discount to the fixed product discount and back

6. Coupon amount is not changing when we are changing the discount type so we can set an amount which is not validated and we don't see validation error 
image
image
With wrong validation after that, the sending of the email doesnt working 
image

7. There are no font changes on the preview or in the email
image
image

8. I can't send the second newsletter with coupons, just to make copies of the first, and even if I will change some data in the coupon settings it will not change them, but if I will make a copy and delete all of the coupons and make them again is sending much longer and has the same coupon code but new data
image
image
image
image
image

samnajian:
@Aschepikov Thanks for the quick feedback, much appreciated, some of the issues you reported were because of the js error on render, that’s fixed and pushed now, so please give it another check and those rendering issues are gone, apart from that, please note that the none-coupon settings are 100% the same as the ones for the button block, so the issues are the same ( things like longer font name causing the font-size selector to get cut ) and we’ll be tackling the design in a separate ticket anyways, so you can ignore these

mailpoet/assets/js/src/newsletter_editor/components/save.js
```diff
@@ -53,6 +53,8 @@ Module.save = function () {
     if (json.body.blockDefaults) {
       delete json.body.blockDefaults.woocommerceHeading;
       delete json.body.blockDefaults.woocommerceContent;
+      delete json.body.blockDefaults?.coupon.couponId;
+      delete json.body.blockDefaults?.coupon.code;
```
costasovo: I think here we should also check the coupon is set `delete json.body.blockDefaults?.coupon?.code`;

costasovo (dismissed):
LGTM! 👍

costasovo (approved):

Aschepikov (approved):
QA verified

PR #4641, opened 2023-01-10
Sanitize data on save [MAILPOET-4942]

samnajian:
[MAILPOET-4942]

[MAILPOET-4942]: https://mailpoet.atlassian.net/browse/MAILPOET-4942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/MP/v1/Segments.php
+++ b/mailpoet/lib/API/MP/v1/Segments.php
@@ -42,7 +42,9 @@ class Segments {
     $this->validateSegmentName($data);
 
     try {
-      $segment = $this->segmentsRepository->createOrUpdate($data['name'], $data['description'] ?? '');
+      $name = sanitize_text_field($data['name']);
+      $description = isset($data['description']) ? sanitize_textarea_field($data['description']) : '';
+      $segment = $this->segmentsRepository->createOrUpdate($name, $description);
     } catch (\Exception $e) {
       throw new APIException(
         __('The list couldn’t be created in the database', 'mailpoet'),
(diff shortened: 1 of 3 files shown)


mailpoet/lib/API/MP/v1/Segments.php
```diff
@@ -42,7 +42,9 @@ public function addList(array $data): array {
     $this->validateSegmentName($data);
 
     try {
-      $segment = $this->segmentsRepository->createOrUpdate($data['name'], $data['description'] ?? '');
+      $name = sanitize_text_field($data['name']);
```
xknown, line 45: `sanitize_text_field` is just a wrapper around `strip_tags` which might not do what you might expect. For example:
```
> return [sanitize_text_field('hola < aa> "'), sanitize_text_field('hola > aa "')]
array (
  0 => 'hola < aa> "',
  1 => 'hola > aa "',
)
```
If this behavior is fine, please feel free to ignore this comment :)

costasovo (approved):
Good job!

mailpoet/lib/API/MP/v1/Segments.php
costasovo: We still need to add esc_attr when printing the value on the output, so I guess this is fine cc @samnajian.

Aschepikov (approved):
GJ!

PR #4642, opened 2023-01-10
Fix multiple form display on pages with Woocommerce [products] shortcode - MAILPOET-4885, MAILPOET-4820

triple0t:
## Description

Fix multiple form display on pages with Woocommerce [products] shortcode

## QA notes

* Create a pop-up form. Enable display on all pages
* Create a page with [products] shortcode.
* View the page
* Notice the pop-up form is displayed once.

Note: Please QA [MAILPOET-4820](https://mailpoet.atlassian.net/browse/MAILPOET-4820) as well.

## Linked tickets

[MAILPOET-4885](https://mailpoet.atlassian.net/browse/MAILPOET-4885)

[MAILPOET-4820](https://mailpoet.atlassian.net/browse/MAILPOET-4820)

[MAILPOET-4820]: https://mailpoet.atlassian.net/browse/MAILPOET-4820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4885]: https://mailpoet.atlassian.net/browse/MAILPOET-4885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4820]: https://mailpoet.atlassian.net/browse/MAILPOET-4820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Form/DisplayFormInWPContent.php
+++ b/mailpoet/lib/Form/DisplayFormInWPContent.php
@@ -87,7 +87,7 @@ class DisplayFormInWPContent {
    * @param mixed $content
    * @return string|mixed
    */
-  public function display($content = null) {
+  private function display($content = null) {
     if (!is_string($content) || !$this->shouldDisplay()) return $content;
 
     $forms = $this->getForms();
@@ -125,6 +125,7 @@ class DisplayFormInWPContent {
     // this code ensures that we display the form only on a page which is related to single post
     if (!$this->wp->isSingle() && !$this->wp->isPage()) $result = $this->wp->applyFilters('mailpoet_display_form_is_single', false);
 
+    // Ensure form does not show up multiple times when called from the woocommerce_product_loop_end filter
     if ($this->inWooProductLoop) $result = $this->displayFormInProductListPage();
 
     $noFormsCache = $this->wp->getTransient(DisplayFormInWPContent::NO_FORM_TRANSIENT_KEY);
--- a/mailpoet/lib/Form/DisplayFormInWPContent.php
+++ b/mailpoet/lib/Form/DisplayFormInWPContent.php
@@ -59,6 +59,8 @@ class DisplayFormInWPContent {
 
   private $wooShopPageId = null;
 
+  private $inWooProductLoop = false;
+
   public function __construct(
     WPFunctions $wp,
     FormsRepository $formsRepository,
@@ -101,6 +103,16 @@ class DisplayFormInWPContent {
     return $result;
   }
 
+  public function contentDisplay($content = null) {
+    $this->inWooProductLoop = false;
+    return $this->display($content);
+  }
+
+  public function wooProductListDisplay($content = null) {
+    $this->inWooProductLoop = true;
+    return $this->display($content);
+  }
+
   private function shouldDisplay(): bool {
     $result = true;
     // This is a fix Yoast plugin and Shapely theme compatibility
@@ -113,7 +125,7 @@ class DisplayFormInWPContent {
     // this code ensures that we display the form only on a page which is related to single post
     if (!$this->wp->isSingle() && !$this->wp->isPage()) $result = $this->wp->applyFilters('mailpoet_display_form_is_single', false);
 
-    if ($this->displayFormInProductListPage()) $result = true;
+    if ($this->inWooProductLoop) $result = $this->displayFormInProductListPage();
 
     $noFormsCache = $this->wp->getTransient(DisplayFormInWPContent::NO_FORM_TRANSIENT_KEY);
     if ($noFormsCache === '1') $result = false;
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -206,11 +206,11 @@ class Hooks {
     );
     $this->wp->addFilter(
       'the_content',
-      [$this->displayFormInWPContent, 'display']
+      [$this->displayFormInWPContent, 'contentDisplay']
     );
     $this->wp->addFilter(
       'woocommerce_product_loop_end',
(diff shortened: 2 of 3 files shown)


mailpoet/lib/Form/DisplayFormInWPContent.php
```diff
@@ -113,7 +125,7 @@ private function shouldDisplay(): bool {
     // this code ensures that we display the form only on a page which is related to single post
     if (!$this->wp->isSingle() && !$this->wp->isPage()) $result = $this->wp->applyFilters('mailpoet_display_form_is_single', false);
 
-    if ($this->displayFormInProductListPage()) $result = true;
+    if ($this->inWooProductLoop) $result = $this->displayFormInProductListPage();
```
lysyjan, line 129: I need help understanding this change. Please can you explain it? Also, a comment in the code can be helpful because this logic is very complex.

mailpoet/lib/Form/DisplayFormInWPContent.php
```diff
@@ -101,6 +103,16 @@ public function display($content = null) {
...
   }
 
+  public function contentDisplay($content = null) {
+    $this->inWooProductLoop = false;
+    return $this->display($content);
+  }
+
+  public function wooProductListDisplay($content = null) {
+    $this->inWooProductLoop = true;
+    return $this->display($content);
```
lysyjan, line 113: We could change the display method from `public` to `private` to avoid calling it without setting the new property `inWooProductLoop`. But I am not sure how important is to have this property configured properly. What do you think?

lysyjan (commented):
This pull request should fix two issues in our form rendering, but I don't see any changes or improvements in tests. Did you forget, or are those changes not testable? I have to ask because it's a very complex class with significant functionality.

mailpoet/lib/Form/DisplayFormInWPContent.php
triple0t: This method `displayFormInProductListPage()`, checks if we can display forms on Woo Shop page.

The bug we had previously: This function `shouldDisplay()` was being called within the `woocommerce_product_loop_end` filter and was always returning true (because that was the default value) instead of false or true depending on the result of the `displayFormInProductListPage()` method check.

For example

When you have the [products] shortcode in multiple locations on the page. This filter `woocommerce_product_loop_end` would be called at the end of each of those product loops. 

if `shouldDisplay()` returns true for each of those product loops, the form would be displayed multiple times.

The update here ensures the form should only display depending on the result of the `displayFormInProductListPage()` method check (if we are within the `woocommerce_product_loop_end` filter).

mailpoet/lib/Form/DisplayFormInWPContent.php
triple0t: We can't change the method from `public` to `private`. It is being used here https://github.com/mailpoet/mailpoet/blob/40b0914d3c6864906de015f48ec8bf536a87abbb/mailpoet/lib/Config/Hooks.php#L211-L214

>  But I am not sure how important is to have this property configured properly.

We need the property in the check here https://github.com/mailpoet/mailpoet/blob/40b0914d3c6864906de015f48ec8bf536a87abbb/mailpoet/lib/Form/DisplayFormInWPContent.php#L128

Which help determine if we are within the `woocommerce_product_loop_end` filter or not.

triple0t:

> This pull request should fix two issues in our form rendering, but I don't see any changes or improvements in tests. Did you forget, or are those changes not testable? I have to ask because it's a very complex class with significant functionality.

I haven’t figured out a good way to write the tests.

I'm planning on working on a follow-up ticket to write the acceptance tests

mailpoet/lib/Form/DisplayFormInWPContent.php
lysyjan: @triple0t I may have written it confusingly, but I thought changing `public function display($content = null)` to `private function display($content = null)`. Never mind, because it is a detail only.

lysyjan (commented):
> I haven’t figured out a good way to write the tests.
> I'm planning on working on a follow-up ticket to write the acceptance tests

There is an existing unit test. Do you think it is possible to extend it? 
I am afraid that a follow-up ticket with acceptance tests will stay in our backlog for months without a touch.

mailpoet/lib/Form/DisplayFormInWPContent.php
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/4642/commits/e8109c66ce2cf711091e346643997006f41ed4c3

mailpoet/lib/Form/DisplayFormInWPContent.php
triple0t: Comment added in https://github.com/mailpoet/mailpoet/pull/4642/commits/e8109c66ce2cf711091e346643997006f41ed4c3

triple0t:
> There is an existing unit test. Do you think it is possible to extend it?
> I am afraid that a follow-up ticket with acceptance tests will stay in our backlog for months without a touch.

Yeah, I understand.

I have updated the tests

lysyjan (dismissed):
Thank you @triple0t 👍

Aschepikov (approved):
GJ!

PR #4644, opened 2023-01-11
Add deactivation survey [MAILPOET-4686]

lysyjan:
## Description

This PR adds a modal with a survey when a user deactivates our plugin.

I tried to unify a design with our plugin, but there are some restrictions because it's a 3rd party survey.
Here is the current state

## Code review notes

_N/A_

## QA notes

 1. Deactivate the plugin and check the modal with the survey
   a. Reply to the survey and check your answer in the result (the link is in the Jira ticket)
   b. Skip the survey and check that plugin is deactivated

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4686]

## After-merge notes

Delete results from the poll.

[MAILPOET-4686]: https://mailpoet.atlassian.net/browse/MAILPOET-4686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/views/deactivationPoll/index.html
+++ b/mailpoet/views/deactivationPoll/index.html
@@ -9,7 +9,6 @@
           var formOpen = false;
 
           if (!deactivateLink || !overlay || !closeButton) {
-            console.log('early return');
             return;
           }
 
--- a/mailpoet/views/deactivationPoll/index.html
+++ b/mailpoet/views/deactivationPoll/index.html
@@ -1,13 +1,17 @@
-<div class="mailpoet-deactivate-survey-modal" id="mailpoet-deactivate-survey">
+<div class="mailpoet-deactivate-survey-modal" id="mailpoet-deactivate-survey-modal">
   <div class="mailpoet-deactivate-survey-wrap">
     <div class="mailpoet-deactivate-survey">
-
       <script type="text/javascript">
-        window.addEventListener('load', function(event) {
+        window.addEventListener('load', function() {
           var deactivateLink = document.querySelector('#the-list [data-slug="mailpoet"] span.deactivate a');
-          var overlay = document.querySelector('#mailpoet-deactivate-survey');
+          var overlay = document.querySelector('#mailpoet-deactivate-survey-modal');
           var closeButton = document.querySelector('#mailpoet-deactivate-survey-close');
           var formOpen = false;
+
+          if (!deactivateLink || !overlay || !closeButton) {
+            console.log('early return');
+            return;
+          }
 
           deactivateLink.addEventListener('click', function (event) {
             event.preventDefault();
@@ -33,8 +37,7 @@
         var pd_callback = function(json) {
           var obj = JSON.parse(json);
           var deactivateLink = document.querySelector('#the-list [data-slug="mailpoet"] span.deactivate a');
-          var overlay = document.querySelector('#mailpoet-deactivate-survey');
-
+          var overlay = document.querySelector('#mailpoet-deactivate-survey-modal');
           if (obj.result === 'already-registered' || obj.result === 'registered') {
             overlay.style.display = 'none';
             location.href = deactivateLink.getAttribute('href');
--- a/mailpoet/views/deactivationPoll/index.html
+++ b/mailpoet/views/deactivationPoll/index.html
@@ -1,0 +1,51 @@
+<div class="mailpoet-deactivate-survey-modal" id="mailpoet-deactivate-survey">
+  <div class="mailpoet-deactivate-survey-wrap">
+    <div class="mailpoet-deactivate-survey">
+
+      <script type="text/javascript">
+        window.addEventListener('load', function(event) {
+          var deactivateLink = document.querySelector('#the-list [data-slug="mailpoet"] span.deactivate a');
+          var overlay = document.querySelector('#mailpoet-deactivate-survey');
+          var closeButton = document.querySelector('#mailpoet-deactivate-survey-close');
+          var formOpen = false;
+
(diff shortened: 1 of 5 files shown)


mailpoet/views/deactivationPoll/index.html
```diff
@@ -0,0 +1,51 @@
+<div class="mailpoet-deactivate-survey-modal" id="mailpoet-deactivate-survey">
```
triple0t: Would it be possible to use the same attribute value?

I noticed for class, we have **"mailpoet-deactivate-survey-modal"** and for id, we have **"mailpoet-deactivate-survey"**.

Using two different attribute values for the same HTML element feels confusing

mailpoet/views/deactivationPoll/index.html
```diff
@@ -0,0 +1,51 @@
...
+  <div class="mailpoet-deactivate-survey-wrap">
+    <div class="mailpoet-deactivate-survey">
```
triple0t, line 3: Case in point here.

We are using "mailpoet-deactivate-survey" for the class attribute which is used as the Id for the parent element

mailpoet/views/deactivationPoll/index.html
```diff
@@ -0,0 +1,51 @@
...
+
+      <script type="text/javascript">
+        window.addEventListener('load', function(event) {
```
triple0t: Can we remove this `event` variable if it's unused?

mailpoet/views/deactivationPoll/index.html
```diff
@@ -0,0 +1,51 @@
...
+          var deactivateLink = document.querySelector('#the-list [data-slug="mailpoet"] span.deactivate a');
+          var overlay = document.querySelector('#mailpoet-deactivate-survey');
```
triple0t: Can we add an early return here if `deactivateLink` does not exist?

triple0t (commented):
Looks good. I have a few questions

1. How long before the survey answers are reflected in the poll? I checked this [poll](https://mailpoet.atlassian.net/browse/MAILPOET-4686?focusedCommentId=23503) but could not locate my response.

2. I noticed the Survey questions do not pass through the MailPoet translation function. Is there an option to show different questions for different locals, or can the question be dynamically translated?

3. Can a dismiss button be added to the model? 
e.g, something like this?
Screenshot 2023-01-11 at 1 12 01 PM

I'm conflicted on what action it should perform: either close the modal (and do nothing) **or** perform the same function as the "Skip survey and deactivate MailPoet" button.

What do you think?

mailpoet/views/deactivationPoll/index.html
lysyjan: For a bigger part, I reused the old code from the previous survey mentioned in the Jira task.
If you think that it's confusing, I can change it.

mailpoet/views/deactivationPoll/index.html
lysyjan: Yes, we can. Good point.

mailpoet/views/deactivationPoll/index.html
lysyjan: 👍 Yes, we can.

lysyjan:
@triple0t Thank you for your feedback 👍 

I added a commit addressing your comments.

> 1. How long before the survey answers are reflected in the poll? I checked this [poll](https://mailpoet.atlassian.net/browse/MAILPOET-4686?focusedCommentId=23503) but could not locate my response.

It's almost immediately. If you were connected via VPN, it could block your vote because I voted via VPN too. I saw some restrictions in the poll settings. You can try a new anonymous tab. I also deleted all responses, so it should be easy to check if it works properly. We could disable those restrictions in the poll settings.

> 2. I noticed the Survey questions do not pass through the MailPoet translation function. Is there an option to show different questions for different locals, or can the question be dynamically translated?

Crowdsignal allows customized UI translation as a button etc., but the translation of answers and the question is impossible, or I haven't found it. The non-translated question in the code should be visible only when JS is disabled. I think that we are fine with the current state because it doesn't make sense to me to mix more languages in the modal. Are you agree?

> 3. Can a dismiss button be added to the model?

I considered it, but when both buttons can close this modal, and the escape function can close it too, I didn't add it.
The old survey didn't contain it, but we can improve our new solution. I would expect that the button has the same functionality as the escape key.
@pavel-mailpoet  What do you think?

triple0t (dismissed):

mailpoet/views/deactivationPoll/index.html
```diff
@@ -0,0 +1,54 @@
...
+    <div class="mailpoet-deactivate-survey">
+      <script type="text/javascript">
+        window.addEventListener('load', function() {
+          var deactivateLink = document.querySelector('#the-list [data-slug="mailpoet"] span.deactivate a');
+          var overlay = document.querySelector('#mailpoet-deactivate-survey-modal');
+          var closeButton = document.querySelector('#mailpoet-deactivate-survey-close');
+          var formOpen = false;
+
+          if (!deactivateLink || !overlay || !closeButton) {
+            console.log('early return');
```
triple0t: Hi @lysyjan 
Noting the console log here. 
You can ignore if this is intentional

lysyjan: Thank you! I removed it in a new commit.

lysyjan:
@Aschepikov Conflict resolved

Aschepikov (approved):
GJ!

PR #4650, opened 2023-01-12
Redesign the MSS step of the welcome wizard [MAILPOET-4818]

rodrigoprimo:
## Description

This PR implements the new MSS step of the Welcome Wizard. It is a big PR as it involves design and functional changes to the MSS step, some JSX to TSX refactoring, and refactoring of the code to validate an MSS key so that it can be used both in the settings page and in the wizard. Check the Jira ticker for more details on the specific changes.

## Code review notes

@samnajian already helped with a review of a draft version of this PR and helped me confirm a few of the decisions that I made when implementing this PR. But he did not check the most recent commits, and anyway, I believe a full review is valid. I would suggest reviewing this PR commit by commit. I tried my best to make the commits granular and the commit messages relevant to make the reviewing process easier.

I left one inline comment with a detail that I'm not sure about.

## QA notes

This PR makes significant changes to both the welcome wizard and the "Key Activation" tab of the settings page. The only visible change in the "Key Activation" tab is that now the "Verify" button is disabled when there is no key. The other changes were refactors without any visible change to the user.

Please test both pages carefully as there are a lot of code changes in this PR.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4818]

## After-merge notes

_N/A_

[MAILPOET-4818]: https://mailpoet.atlassian.net/browse/MAILPOET-4818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/wizard/steps/pitch_mss_step.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+function WelcomeWizardPitchMSSStep({
+  subscribersCount,
+  finishWizard,
+}: WelcomeWizardPitchMSSStepPropType): JSX.Element {
+  const { path } = useRouteMatch();
+
+  return (
+    <Switch>
```
rodrigoprimo, line 17: I ended up placing the nested routes here. It was the easy place to add it, but I'm not sure if I should have placed everything on wizard.tsx where the parent routes are located.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <Route path={`${path}/part/2`}>
+        <MSSStepSecondPart finishWizard={finishWizard} />
+      </Route>
+      <Route path={`${path}/part/3`}>
+        <MSSStepThirdPart finishWizard={finishWizard} />
+      </Route>
+      <Route path={[`${path}`, `${path}/part/1`]}>
```
rodrigoprimo: Another thing that I don't like about this is the fact that I had to use two paths here and because of the root path, I had to place the route for the first past last.

mailpoet/assets/js/src/wizard/welcome_wizard_controller.tsx
```diff
@@ -60,13 +60,17 @@ function WelcomeWizardStepsController({
       });
   }
 
-  async function finishWizard() {
+  const finishWizard = useCallback(async (redirect_url = null) => {
```
rodrigoprimo: I know we discussed moving this function to its own component, but I didn't implement it because it was not clear to me how to reuse the `loading` state that is defined inside `updateSettings()` and used in the `WelcomeWizardStepsController` component.

I agree that I should do something about it, as passing this function around the components doesn't look good.

rodrigoprimo:
@samnajian, I'm asking for your review on the draft of this PR to get some early feedback on the implementation. Especially around the use of the React Router to implement the sub-pages (that I'm calling parts but I'm not super happy with this name) of the MSS step. Any feedback that you have is welcome. I left some inline comments with the questions that I have. Thanks.

mailpoet/assets/js/src/newsletters/send/congratulate/success_pitch_mss.tsx
```diff
@@ -3,10 +3,19 @@ import { MailPoet } from 'mailpoet';
 
 import { Heading } from 'common/typography/heading/heading';
 import { WelcomeWizardStepLayoutBody } from '../../../wizard/layout/step_layout_body.jsx';
-import {
-  Controls,
-  FreeBenefitsList,
-} from '../../../wizard/steps/pitch_mss_step';
+import { List } from '../../../common/typography/list/list';
+import { Button } from '../../../common';
```
samnajian: Suggestion: You can import directly from `common` like `import {something} from 'common'`, and defining an index file in `typography` folder should allow unifying importing `Heading` and `List`.

mailpoet/assets/js/src/newsletters/send/congratulate/success_pitch_mss.tsx
```diff
@@ -3,10 +3,19 @@ import { MailPoet } from 'mailpoet';
...
+
+function FreeBenefitsList(): JSX.Element {
```
samnajian, line 7: Suggestion: I'd leave the type inference to TS in cases it can infer the type, and this is one these cases.

mailpoet/assets/js/src/wizard/welcome_wizard_controller.tsx
samnajian: I ended up extracting `updateSettings`, `wizard-redesign-mss-step-different-look` branch includes the proposed change, however it's better to test it together since I'm  not exactly sure how the feature should behave.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step.tsx
samnajian: I would do the same since the sub-routes belong to this component.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step.tsx
samnajian: Well if we want both of these routes to render them same component, then this is the way to go, but the other way around would be to redirect the route path to `${path}/par/1`, something like 
```
<Route exact path={`${path}`}>
  <Redirect to={`${path}/part/1`} /> 
</Route>

```

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+function MSSStepFirstPart({
+  subscribersCount,
+  finishWizard,
+}: MSSStepFirstPartPropType): JSX.Element {
+  const history = useHistory();
+  const { step } = useParams<{ step: string }>();
+
+  const mailpoetAccountUrl = MailPoet.MailPoetComUrlFactory.getPurchasePlanUrl(
```
samnajian: If i'm not mistaken `mailpoetAccountUrl` is not bound to state, so it can get defined outside of the component.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <div className="mailpoet-gap" />
+      <div className="mailpoet-gap" />
+
+      <Button
+        isFullWidth
+        href={mailpoetAccountUrl}
+        target="_blank"
+        rel="noopener noreferrer"
+        onClick={(event) => {
+          event.preventDefault();
```
samnajian: I'd define this as a named function to: 
1) Refrain from defining the function in each render
2) Adding a meaningful name to the function would allow other developers to quickly understand what's being done and why

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+      <div className="mailpoet-gap" />
+      <p>{MailPoet.I18n.t('welcomeWizardMSSFirstPartSubtitle')}</p>
+      <div className="mailpoet-gap" />
+
+      <div className="mailpoet-welcome-wizard-mss-list">
+        <List>
+          <li>{MailPoet.I18n.t('welcomeWizardMSSList1')}</li>
+          <li>{MailPoet.I18n.t('welcomeWizardMSSList2')}</li>
+          {subscribersCount < 1000 ? (
```
samnajian: `subscribersCount` is passed down too often while the value is a global, it can either come directly from window or using a utility function.

samnajian (commented):
Great work, please see my comments.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step.tsx
rodrigoprimo: Thanks for your input. I think it is much cleaner than what I had and the `<Redirect>` component is what I was missing to be able to implement it.

I had to change `<Redirect to={`${path}/part/1`} />` to `<Redirect to={`/steps/${step}/part/1`} />` and call `useParams()` to get the value of `step`. I believe this is because `<Redirect>` expects a URL and not the route match string.

Since this is the review of a draft PR, to avoid polluting the commit history and since this PR will need to be reviewed again, I'm opting to amend the commits that already existed with the changes instead of creating new commits.

mailpoet/assets/js/src/wizard/welcome_wizard_controller.tsx
rodrigoprimo: @samnajian, maybe you forgot to push the branch? I'm not able to find it.

mailpoet/assets/js/src/newsletters/send/congratulate/success_pitch_mss.tsx
rodrigoprimo: Thanks for the suggestion. I implemented it in 53c78ce6ae5ff4fe1a1ffb62b7ed13d97e01c0e9.

mailpoet/assets/js/src/newsletters/send/congratulate/success_pitch_mss.tsx
rodrigoprimo: Good to know that this is an option. Due to time constraints, and since I assume having the type is not an issue, I'm not planning to implement this change in this PR (please let me know if you think that I should). But I will keep this in mind going forward.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
rodrigoprimo: Done

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
rodrigoprimo: Makes sense. fd5a43a5fd3a10c7a8391a6c578ff77665da0b64 is my attempt at implementing it. When you have a moment, it would be great if you could take a look and share your thoughts with me.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
rodrigoprimo: Done, thanks

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/second_part.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useEffect } from 'react';
+import { useSelector } from 'settings/store/hooks';
+import { OwnEmailServiceNote } from './own_email_service_note';
+
+function MSSStepSecondPart(): JSX.Element {
+  const history = useHistory();
+  const { step } = useParams<{ step: string }>();
+  const state = useSelector('getKeyActivationState')();
+
+  useEffect(() => {
```
rodrigoprimo, line 16: I'm unsure if using `useEffect()` here is the best way to change to the next part of this step when the user adds a valid key. This has the downside that the code runs when the components mount for the first time which is not necessary in this case.

rodrigoprimo:
I considered adding an acceptance test to cover the basic flow of the wizard but I'm not sure if there is an MSS key I could use to pass the key validation. Thoughts?

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
costasovo: The attempt in https://github.com/mailpoet/mailpoet/commit/fd5a43a5fd3a10c7a8391a6c578ff77665da0b64 makes sense. In general, this makes sense. But in this particular case, I don't think that redefining the function for each render is a big concern here because I don't expect the component to re-render many times. Another solution could be using [useCallback](https://beta.reactjs.org/reference/react/useCallback)

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/own_email_service_note.tsx
```diff
@@ -0,0 +1,77 @@
...
+        )}
+      </p>
+      {confirmationModalIsOpen && (
+        <Modal
+          className="mailpoet-welcome-wizard-confirmation-modal"
+          title={MailPoet.I18n.t('welcomeWizardMSSConfirmationModalTitle')}
+          onRequestClose={closeConfirmationModal}
+        >
+          <p>
+            {window.mailpoet_mail_function_enabled
```
costasovo: I see that this variable is available for every admin page and set to window in layout.html. I would consider adding it as a static property to the MailPoet module [as we do with other such variables](https://github.com/mailpoet/mailpoet/blob/05202d572f8c6475d3704e7d0df25d9a6dfadc2b/mailpoet/assets/js/src/mailpoet.ts#L15-L74).

We want to change how we pass the variables to JS, so accessing them through the module should make future changes easier. You can find some context in this ticket: https://mailpoet.atlassian.net/browse/MAILPOET-2943

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/second_part.tsx
```diff
@@ -42,7 +42,7 @@ function MSSStepSecondPart({
               key={index}
               target="_blank"
               rel="noreferrer"
-              href="https://account.mailpoet.com/"
+              href="https://account.mailpoet.com/?ref=plugin-wizard&utm_source=plugin&utm_medium=onboarding&utm_campaign=purchase"
```
costasovo, line 39: I remember that we used to use a service for building URLs that link to the account page. [The service adds](https://github.com/mailpoet/mailpoet/blob/05202d572f8c6475d3704e7d0df25d9a6dfadc2b/mailpoet/assets/js/src/mailpoet_com_url_factory.js)  ref parameter in case the site is run by an affiliate partner. Maybe the we don't use it any more. I remember seeing a discussion about it, but I couldn't find it. If I remember correctly you were one of the participants in so maybe you know the current state.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
```diff
@@ -49,7 +47,7 @@ function MSSStepFirstPart({
         <List>
           <li>{MailPoet.I18n.t('welcomeWizardMSSList1')}</li>
           <li>{MailPoet.I18n.t('welcomeWizardMSSList2')}</li>
-          {subscribersCount < 1000 ? (
+          {window.mailpoet_subscribers_count < 1000 ? (
```
costasovo: Please use `MailPoet.subscribersCount` we already have this one defined in the MailPoet module.

mailpoet/assets/js/src/common/premium_key/key_messages/service_unavailable_messages.tsx
```diff
@@ -1,4 +1,4 @@
-import { MailPoet } from 'mailpoet';
+import { MailPoet } from '../../../mailpoet';
```
costasovo: I'm not sure this change was intentional. Normally we use `import { MailPoet } from 'mailpoet';`.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/second_part.tsx
costasovo: This is interesting. We have a state in store and steps state in route 🤔 I don't know what is the best practice synchronizing this. I think we can go with this solution.

costasovo (commented):
Great job! This was a big one. Thank you for all the refactorings and cleanups you did!

I added a few suggestions and questions. Only tiny details.

I tested the new page and when redirected to the shop's page I got and error message on the shop. Maybe it is a known issue and probably related to the shop, but I rather share it here. I can provide an URL that triggers the error.

Screenshot 2023-01-31 at 11 28 41

rodrigoprimo:
> I tested the new page and when redirected to the shop's page I got, and error message on the shop. Maybe it is a known issue and probably related to the shop, but I rather share it here. I can provide an URL that triggers the error.

@costasovo, I was not able to reproduce this error. Could you please share the problematic URL and where you are seeing it?

Also, a friendly reminder about this comment in case you missed it: https://github.com/mailpoet/mailpoet/pull/4650#issuecomment-1409410248.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/second_part.tsx
rodrigoprimo: Ok, thanks for checking.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/own_email_service_note.tsx
rodrigoprimo: That is good point and thanks for linking to the related ticket. I implemented your suggestion in ac9046cd078a776d080a275527dea36e8027d74b.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
rodrigoprimo: See above: https://github.com/mailpoet/mailpoet/pull/4650#discussion_r1091977707

mailpoet/assets/js/src/common/premium_key/key_messages/service_unavailable_messages.tsx
rodrigoprimo: I had this fixed already but forgot to push that commit with the fix before asking for the review. Apologies for the noise.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/second_part.tsx
rodrigoprimo: I remember this discussion as well but like you I'm not able to find it. I have a vague memory that the conclusion to not use it because it is not currently working properly, but I could be wrong. Checking the service quickly, I was not able to find a method that I could use to generate the same URL that I'm generating here.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
rodrigoprimo: Thanks for the additional information. I'm assuming you are ok with leaving the code as is, but please let me know if you think I should refactor it to use `useCallback` for some reason.

rodrigoprimo:
Thanks for the review, @costasovo. I believe I addressed all your comments. Could you please take a second look?

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/first_part.tsx
costasovo: The current code is ok.

mailpoet/assets/js/src/wizard/steps/pitch_mss_step/second_part.tsx
costasovo: I found it ([#](https://mailpoet.atlassian.net/browse/MAILPOET-4749?focusedCommentId=23366))! This time I searched using the name of the service. It seems we don't need to use it. So we are fine. Maybe we could get rid of it. I'll ask on slack.

costasovo (commented):
Thank you for updating the PR and for reminding me about the acceptance test question. I saw it, but forgot about it when checking commits. Sorry about that!

> I considered adding an acceptance test to cover the basic flow of the wizard but I'm not sure if there is an MSS key I could use to pass the key validation. Thoughts?

Please check [the acceptance test we have for the key validation page in the settings](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/acceptance/Settings/AddSendingKeyCest.php#L13-L16). It tests basically the same functionality and solves obtaining the key.

As for the error in the shop. I'll share the URL via PM on slack.

rodrigoprimo:
@pavel-mailpoet, @costasovo found an issue in the link to point users to create a new account in the shop. We believe should be fixed in the shop and not in the plugin, and I would like your help to confirm that before we create a new ticket.

Here is a simplified version of the link that we use in the "Connect MailPoet" button in the first part of the MSS step on the welcome wizard:

https://account.mailpoet.com/?s=SUBSCRIBER_COUNT&g=starter

The problem is that if SUBSCRIBER_COUNT for a given site is bigger than 1000, the shop will load with the following error: `This plan is currently not available`. The link works if we replace `g=starter` with `g=business` for sites with more than a thousand subscribers. But we both believe that we should address this in the shop as this is where we know what are the subscriber limits for each type of plan. Maybe it is even possible to redirect a user with more than a thousand subscribers from `?s=SUBSCRIBER_COUNT&g=starter` to `?s=SUBSCRIBER_COUNT&g=business`? What do you recommend as the next step? Should we block this PR while this issue is not fixed?

cc @triple0t for awareness since you are the DRI for this project as we might block this PR

pavel-mailpoet:
@rodrigoprimo thank you for the ping. What I would propose is:

- remove the arguments from the link here in this pull request and unblock it – we don't want to send users to a page with an error message
- create a follow-up ticket with the following:
  - fix for the shop that would accept both arguments
  - update this code to re-introduce the argument back

What do you think?

rodrigoprimo:
@pavel-mailpoet, sounds good to me. I will create the tickets and update this PR.

rodrigoprimo:
@costasovo, I implemented Pavel's suggestion, created the Jira tickets, and added a commit to this PR to remove the parameters from the link. Unfortunately, I had to rebase most of the commits and force push due to a conflict in `global.d.ts`.

To not further delay this PR, I plan to check if I can write an acceptance test for the welcome wizard in a separate ticket.

Could you please do a final review of this PR?

costasovo (approved):
👍 Thank you for preparing the tickets and unblocking the PR! We can add the acceptance test in a separate PR.

Aschepikov (approved):
GJ!

PR #4653, opened 2023-01-13
Display the PHP outdated version warning for sites running PHP <= 7.3

rodrigoprimo:
## Description

@NeosinneR, I assigned this PR to you as the code changes are fairly straightforward and the main question that I have is related to the specs of this ticket. I wonder if we want to update the message that is displayed in the notice for users running PHP 7.3 or if we are fine with it? The message currently says "Your website is running on PHP 7.3.0 which MailPoet does not officially support".

Please correct me if I'm wrong, but my understanding is that we still support PHP 7.3, but don't recommend that sites run it. Let me know if it is fine to proceed with this PR the way it is or if I should change the message to something else.

## Code review notes

_N/A_

## QA notes

- Check you can see a PHP outdated version warning in all MailPoet versions when the site is running PHP <= 7.3 and that the same warning is not displayed for sites running PHP >= 7.4.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4933]

## After-merge notes

_N/A_

[MAILPOET-4933]: https://mailpoet.atlassian.net/browse/MAILPOET-4933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Util/Notices/PHPVersionWarnings.php
+++ b/mailpoet/lib/Util/Notices/PHPVersionWarnings.php
@@ -23,7 +23,7 @@ class PHPVersionWarnings {
 
   public function display($phpVersion) {
     // translators: %s is the PHP version
-    $errorString = __('Your website is running on PHP %s which MailPoet does not officially support. Read our [link]simple PHP upgrade guide.[/link]', 'mailpoet');
+    $errorString = __('Your website is running an outdated version of PHP (%s), on which MailPoet might stop working in the future. We recommend upgrading to 7.4 or greater. Read our [link]simple PHP upgrade guide.[/link]', 'mailpoet');
     $errorString = sprintf($errorString, $phpVersion);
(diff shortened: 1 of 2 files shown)


NeosinneR:
@rodrigoprimo good point on the message, and I agree that it's not entirely correct. How about changing it to 
> Your website is running on PHP 7.3.0, which is no longer officially supported, and MailPoet might stop working in the future.

I changed it to mention it's no longer supported from PHP's side (reached end-of-life), and also that MailPoet might stop working, which might be a stronger incentive to update. What do you think?

rodrigoprimo:
Documenting here that I chatted with @NeosinneR and this is the final version of the message that we will use:

> Your website is running an outdated version of PHP (7.3.0), on which MailPoet might stop working in the future. We recommend upgrading to 7.4 or greater.

NeosinneR (approved):

Aschepikov (approved):
GJ!

PR #4655, opened 2023-01-16
Update subscriber action [PREMIUM-215]

JanJakes:
## Description

This PR contains placeholders for the "update subscriber" action and some fixes and improvements. The action itself is implemented in the premium PR.

## Code review notes

This PR also implements simple "contexts" for automation integrations — that is, defining data that will be loaded in the automation editor.

## QA notes

Please, test the new "update subsriber" action in automations.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/676

## Linked tickets

[PREMIUM-215]

## After-merge notes

_N/A_

[PREMIUM-215]: https://mailpoet.atlassian.net/browse/PREMIUM-215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/integrations/mailpoet/steps/update-subscriber/index.tsx
```diff
@@ -0,0 +1,33 @@
+import { __ } from '@wordpress/i18n';
+import { postAuthor } from '@wordpress/icons';
+import { StepType } from '../../../../editor/store/types';
+import { PremiumModalForStepEdit } from '../../../../../common/premium_modal';
+import { LockedBadge } from '../../../../../common/premium_modal/locked_badge';
+
+export const step: StepType = {
+  key: 'mailpoet:update-subscriber',
+  group: 'actions',
+  title: __('Update subscriber', 'mailpoet'),
```
johnolek, line 10: I wanted to ask about something I [brought up](https://github.com/mailpoet/mailpoet/pull/4617#pullrequestreview-1221225471) in my review of the PR that enforces different text domains for free/premium. Do we end up needing to maintain two separate translations for this and the [corresponding title](https://github.com/mailpoet/mailpoet-premium/pull/676/files#diff-7a774574f0f5132757d75fdd3ca31514d54014acf94815e5385e55270c34e715R26) in premium, and is that a pain as far as our translators are concerned? I ask because this seems like it will probably be a common pattern with many of the premium actions we'll be adding.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -108,4 +109,12 @@ private function buildRegistry(): array {
     }
     return ['steps' => $steps];
   }
+
+  private function buildContext(): array {
+    $data = [];
+    foreach ($this->registry->getContextFactories() as $key => $factory) {
+      $data[$key] = $factory();
+    }
```
johnolek, line 107: If the idea is to allow any plugin to use this, should we surround this with a `try/catch` so one plugin can't ruin the fun for everyone?

johnolek (dismissed):
Looks good! The `try/catch` comment isn't something that I think we would need to change even if it would be a potential problem since obviously no one else will be using this right now, but feel free to update if you think it makes sense.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/update-subscriber/index.tsx
JanJakes: @johnolek My thinking is that, in general, it makes sense to check for the text domains. In the past, we've repeatedly had wrong text domains in the premium plugin, and usually, we had no idea until someone noticed that some premium strings were not translated.

As for this particular case of automation actions in the free and the premium plugin, I think we could address it in a different way. We allow using the free plugin step definition and overriding it. Something like:

```ts
export const step = (freePluginStep: StepType): StepType => ({ title: freePluginStep.title, ...});
```

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
JanJakes: That's a good point! I'm just thinking about how to address a failure...

Should we display a notice in the editor that loading some integration failed? I guess that might be the best, but it will require propagating the errors to the UI somehow.

To add more context to the usage of these contexts, I think we could/should make it async so that the editor loads as fast as possible and the contexts are fetched with an API request asynchronously. Then, I guess, that async request could also contain an array of integration errors that we could convert into notices in the UI.

If that sounds good to you, I will compile it into a ticket so it doesn't get lost here. What do you think?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/update-subscriber/index.tsx
johnolek: I like the sounds of that approach much better than changing the text domain stuff. Great idea!

No need to implement it for this PR unless you really want to.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
johnolek: That sounds like a good solution to me, especially ensuring the UI loads as fast as possible. 

Initially I was thinking that maybe some simple logging would be sufficient in the case of errors, but realistically we might have people writing integrations who may not think to check logs. It might be good to include some logs with more details (e.g. a stack trace) that would be messy to have in the UI.

tldr that all sounds great to me!

johnolek (approved):
This all looks good to me!

Aschepikov (approved):
GJ!

PR #4660, opened 2023-01-17
Add a new item to the WooCommerce Task List with a link to the MP wizard [MAILPOET-4814]

rodrigoprimo:
## Description

This PR adds a new item to the WooCommerce Task List with a link to the MP wizard (see ticket description for more details). It also adds a new Mixpanel event to track when the user arrives to the MP wizard via the WooCommerce task list item.

## Code review notes

- To check that Mixpanel tracking is working you can add `debug: true` to [/mailpoet/assets/js/lib/analytics.js#L9](https://github.com/mailpoet/mailpoet/blob/c368b130b84ca66dcee6b91bd99eb754de8cd330/mailpoet/assets/js/lib/analytics.js#L9) (see https://developer.mixpanel.com/docs/javascript#debug-mode).
- Since Mixpanel is not enabled before the user finishes the wizard, I had to store the information that we want to send an event to Mixpanel in a setting and use it later in a different request. I have some doubts about this implementation, and I left a few questions in inline comments.

## QA notes

- Go to /wp-admin/admin.php?page=wc-admin
- If the user never completed the MP Welcome Wizard, check that there is a link under `Things to do next` pointing the user to the MP wizard. If the user completed the wizard, the same link should point to the newsletters page. In the future, we will update this to point to the MP homepage (https://mailpoet.atlassian.net/browse/MAILPOET-4926).
- Check that the MP Welcome Wizard still works as expected.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4814]

## After-merge notes

_N/A_

[MAILPOET-4814]: https://mailpoet.atlassian.net/browse/MAILPOET-4814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/newsletters/listings/standard.jsx
+++ b/mailpoet/assets/js/src/newsletters/listings/standard.jsx
@@ -187,25 +187,6 @@ class NewsletterListStandardComponent extends Component {
     };
   }
 
-  componentDidMount() {
-    // Send event that the MP wizard was loaded via the WooCommerce home page to Mixpanel and delete the corresponding setting.
-    // We need to do this after the user completes the wizard, since before that tracking is not enabled.
-    if (window.mailpoet_track_wizard_loaded_via_woocommerce) {
-      MailPoet.trackEvent(
-        'User opened the MailPoet setup task in WooCommerce > Home',
-        {
-          'WooCommerce version': MailPoet.woocommerceVersion,
-        },
-      );
-      MailPoet.Ajax.post({
-        api_version: window.mailpoet_api_version,
-        endpoint: 'settings',
-        action: 'delete',
-        data: 'send_event_that_wizard_was_loaded_via_woocommerce',
-      });
-    }
-  }
-
   renderItem = (newsletter, actions, meta) => {
     const rowClasses = classnames(
       'manage-column',
--- a/mailpoet/assets/js/src/newsletters/listings/standard.jsx
+++ b/mailpoet/assets/js/src/newsletters/listings/standard.jsx
@@ -200,8 +200,8 @@ class NewsletterListStandardComponent extends Component {
       MailPoet.Ajax.post({
         api_version: window.mailpoet_api_version,
         endpoint: 'settings',
-        action: 'set',
-        data: { send_event_that_wizard_was_loaded_via_woocommerce: 0 },
+        action: 'delete',
+        data: 'send_event_that_wizard_was_loaded_via_woocommerce',
       });
     }
   }
--- a/mailpoet/assets/js/src/newsletters/listings/standard.jsx
+++ b/mailpoet/assets/js/src/newsletters/listings/standard.jsx
@@ -187,6 +187,25 @@ class NewsletterListStandardComponent extends Component {
     };
   }
 
+  componentDidMount() {
+    // Send event that the MP wizard was loaded via the WooCommerce home page to Mixpanel and delete the corresponding setting.
+    // We need to do this after the user completes the wizard, since before that tracking is not enabled.
+    if (window.mailpoet_track_wizard_loaded_via_woocommerce) {
+      MailPoet.trackEvent(
+        'User opened the MailPoet setup task in WooCommerce > Home',
+        {
+          'WooCommerce version': MailPoet.woocommerceVersion,
+        },
+      );
+      MailPoet.Ajax.post({
+        api_version: window.mailpoet_api_version,
+        endpoint: 'settings',
+        action: 'set',
+        data: { send_event_that_wizard_was_loaded_via_woocommerce: 0 },
+      });
+    }
+  }
+
   renderItem = (newsletter, actions, meta) => {
     const rowClasses = classnames(
       'manage-column',
--- a/mailpoet/lib/WooCommerce/MailPoetTask.php
+++ b/mailpoet/lib/WooCommerce/MailPoetTask.php
@@ -1,0 +1,74 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\WooCommerce;
+
+use Automattic\WooCommerce\Admin\Features\OnboardingTasks\Task;
+use MailPoet\Config\Menu;
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\Settings\SettingsController;
+
+/**
+ * MailPoet task that is added to the WooCommerce homepage.
+ */
+class MailPoetTask extends Task {
+  /**
+   * @return string
+   */
+  public function get_id() { // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    return 'mailpoet_task';
+  }
+
+  /**
+   * @return string
+   */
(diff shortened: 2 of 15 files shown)


mailpoet/assets/js/src/newsletters/listings/standard.jsx
```diff
@@ -187,6 +187,25 @@ class NewsletterListStandardComponent extends Component {
     };
   }
 
+  componentDidMount() {
+    // Send event that the MP wizard was loaded via the WooCommerce home page to Mixpanel and delete the corresponding setting.
+    // We need to do this after the user completes the wizard, since before that tracking is not enabled.
+    if (window.mailpoet_track_wizard_loaded_via_woocommerce) {
```
rodrigoprimo: I'm not sure if this is the best place to put the code that sends the Mixpanel event and updates the setting.

mailpoet/assets/js/src/newsletters/listings/standard.jsx
```diff
@@ -187,6 +187,25 @@ class NewsletterListStandardComponent extends Component {
...
+      MailPoet.trackEvent(
+        'User opened the MailPoet setup task in WooCommerce > Home',
+        {
+          'WooCommerce version': MailPoet.woocommerceVersion,
+        },
+      );
+      MailPoet.Ajax.post({
+        api_version: window.mailpoet_api_version,
+        endpoint: 'settings',
+        action: 'set',
```
rodrigoprimo: Ideally, I would like to remove the setting instead of changing its value. But it seems there is no method in our API to do that and I was not sure if I should add one. Thoughts?

I also considered using WP options, but it seems there is no WP REST API endpoint for options.

mailpoet/assets/js/src/newsletters/listings/standard.jsx
```diff
@@ -187,6 +187,25 @@ class NewsletterListStandardComponent extends Component {
     };
   }
 
+  componentDidMount() {
```
rodrigoprimo: I opted not to refactor this file to TSX as I already spent way more time than I wanted on this ticket. I tried to do the refactor but ran into some issues that were not straightforward to fix and decided to leave it to another moment.

mailpoet/lib/WooCommerce/MailPoetTask.php
```diff
@@ -0,0 +1,74 @@
...
+use MailPoet\Settings\SettingsController;
+
+/**
+ * MailPoet task that is added to the WooCommerce homepage.
+ */
+class MailPoetTask extends Task {
+  /**
+   * @return string
+   */
+  public function get_id() { // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
```
costasovo: It is just a detail, but could we use a return type hint instead of the PHP doc return parameter? Tyhehints are checked runtime PHPdoc might become outdated. Or is there some reason/risk why we can't use type hints?

mailpoet/assets/js/src/newsletters/listings/standard.jsx
costasovo: I see that [the SettingsController has a method for deleting a setting](https://github.com/mailpoet/mailpoet/blob/926620e8f82ca691f1c11f6e973a2f86a6117522/mailpoet/lib/Settings/SettingsController.php#L117) so we could add the endpoint.

mailpoet/assets/js/src/newsletters/listings/standard.jsx
costasovo: Hmm. That is tricky. I'm currently working on the new homepage so after the homepage lands Emails will no longer be the page where we send users from the wizard.

I think we might not to put it anywhere in React, but maybe a simple `document.addEventListener('DOMContentLoaded', trackWizardLoaded)` could work. It can be placed into a separate file just needs to be added to [webpack_admin_index.tsx](https://github.com/mailpoet/mailpoet/blob/671c7eed7db526acb21e78cb8b3e6ff381e31685/mailpoet/assets/js/src/webpack_admin_index.tsx) so that it is bundled within admin JS which we include on every MailPoet admin page.
The DOMContentLoaded event should be enough to make sure `MailPoet.trackEvent` is available. We don't need to wait for Mixpanel to load because this is handled inside the `MailPoet.trackEvent` ([#](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/analytics_event.js#L68-L75)).

What do you think about such an approach? Would that work for you?

costasovo (commented):
Good stuff! Thank you for preparing a fix for the issue with incorrect doctypes in Woo.

I added one comment and tried to answer your questions.

mailpoet/lib/WooCommerce/MailPoetTask.php
rodrigoprimo: Good point. I used an example class from a WooCommerce extension and didn't pause to reflect on the fact that I could use return type hints instead of docblocks precisely for the reason that you mentioned. Thanks for pointing that out. It is now fixed.

mailpoet/assets/js/src/newsletters/listings/standard.jsx
rodrigoprimo: I considered this option, but I wasn't sure if that was a desirable endpoint or not. Since you also think it is a good approach, I added it in a new commit.

mailpoet/assets/js/src/newsletters/listings/standard.jsx
rodrigoprimo: I went with the approach that you suggested.

The only downside that I see is that we are adding this script to every single page. That is not a big deal since it is the admin and it applies only to MP pages. I considered adding a check to load it only in the home page, but I thought this check could be a bit brittle and break in the future due to changes that we might implement. So I used only the DOMContentLoaded event.

rodrigoprimo:
Thanks for the review, @costasovo. I implemented the changes that you suggested. Could you please take another look at this PR?

mailpoet/assets/js/src/newsletters/listings/standard.jsx
costasovo: > The only downside that I see is that we are adding this script to every single page.

Good point! But in most cases, it only checks if `window.mailpoet_track_wizard_loaded_via_woocommerce` is set so I think it is not a big deal.

costasovo (dismissed):
LGTM!

mailpoet/assets/js/src/newsletters/listings/standard.jsx
rodrigoprimo: Yeah, I agree

Aschepikov (dismissed):
GJ!

Aschepikov:
Hi, @rodrigoprimo there are some conflicts, can you resolve them pls

rodrigoprimo:
@Aschepikov, thanks for the ping. I fixed the conflict. CircleCI is now failing but it doesn't seem to be related to this PR as trunk is also failing with the same error.

Aschepikov (approved):
GJ!

PR #4662, opened 2023-01-19
Add new shortcode site:homepate_url and change rendering of site:homepage_link [MAILPOET-4936]

lysyjan:
## Description

This PR changes the behavior of the shortcode `[site:homepage_link]` and unifies it with the other link shortcodes. For this change was created a new shortcode `[site:homepage_url]`.

## Code review notes

_N/A_

## QA notes

When you want to test new settings for confirmation emails, you need to remove the old one from the mailpoet settings table. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4936]

## After-merge notes

_N/A_

[MAILPOET-4936]: https://mailpoet.atlassian.net/browse/MAILPOET-4936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Shortcodes/Categories/Site.php
+++ b/mailpoet/lib/Newsletter/Shortcodes/Categories/Site.php
@@ -35,8 +35,8 @@ class Site implements CategoryInterface {
       case 'homepage_link':
         return sprintf(
           '<a target="_blank" href="%s">%s</a>',
-          $this->wp->getBloginfo('url'),
-          $this->wp->getBloginfo('name')
+          $this->wp->escUrl($this->wp->getBloginfo('url')),
+          $this->wp->escHtml($this->wp->getBloginfo('name'))
         );
 
       default:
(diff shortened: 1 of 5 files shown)


mailpoet/lib/Newsletter/Shortcodes/Categories/Site.php
```diff
@@ -29,9 +29,16 @@ public function process(
...
+      case 'homepage_url':
         return $this->wp->getBloginfo('url');
 
+      case 'homepage_link':
+        return sprintf(
+          '<a target="_blank" href="%s">%s</a>',
+          $this->wp->getBloginfo('url'),
+          $this->wp->getBloginfo('name')
+        );
+
```
brezocordero, line 41: @lysyjan are we displaying the result of the shortcodes on the browser?

If yes, are they escaped later on? If they are displayed and not escaped, could we use `esc_url` and `esc_html`?

Please ignore the comment if it doesn't apply to this case.

lysyjan: Good point! What I tested was without an issue, but I added escaping functions on both strings in the last commit.

brezocordero (approved):
LGTM!

Aschepikov (approved):
GJ!

PR #4663, opened 2023-01-19
Fix ManageSegmentsCest:deleteExistingSegment - MAILPOET-4881

triple0t:
## Description

Fix ManageSegmentsCest:deleteExistingSegment flaky test

## Linked tickets

[MAILPOET-4881](https://mailpoet.atlassian.net/browse/MAILPOET-4881)

[MAILPOET-4881]: https://mailpoet.atlassian.net/browse/MAILPOET-4881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Segments/ManageSegmentsCest.php
+++ b/mailpoet/tests/acceptance/Segments/ManageSegmentsCest.php
@@ -172,7 +172,8 @@ class ManageSegmentsCest {
     $i->wantTo('Empty trash from other segments');
     $i->waitForElementClickable('[data-automation-id="empty_trash"]');
     $i->click('[data-automation-id="empty_trash"]');
-    $i->waitForText('1 segment was permanently deleted.');
+    $i->waitForNoticeAndClose('1 segment was permanently deleted.');
+    $i->wait(10); // wait a moment for redirect loading
     $listingAutomationSelector = '[data-automation-id="listing_item_' . $segment1->getId() . '"]';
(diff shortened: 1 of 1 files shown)


mailpoet/tests/acceptance/Segments/ManageSegmentsCest.php
```diff
@@ -172,7 +172,8 @@ public function deleteExistingSegment(\AcceptanceTester $i) {
     $i->wantTo('Empty trash from other segments');
     $i->waitForElementClickable('[data-automation-id="empty_trash"]');
     $i->click('[data-automation-id="empty_trash"]');
-    $i->waitForText('1 segment was permanently deleted.');
+    $i->waitForNoticeAndClose('1 segment was permanently deleted.');
+    $i->wait(10); // wait a moment for redirect loading
```
lysyjan: We usually use the wait function with a parameter value of one or two seconds. Do you think that ten seconds are necessary?
We could use a lower value and increase it when this test stays flaky. What do you think?

triple0t: I agree with you. Fixed [here](https://github.com/mailpoet/mailpoet/pull/4663/commits/58809db5534a2671d410db7a66e4cdc9929ad559)

lysyjan (approved):

PR #4665, opened 2023-01-23
WC Tracker - campaign revenue tracking [MAILPOET-5014]

costasovo:
## Description
We want to find out how MailPoet campaigns help to generate revenue in Woo stores.
In order to do so, we want to use WC Tracker for tracking campaign revenue stats data we already store in the `mailpoet_statistics_woocommerce_purchases` table.

### Tracked data
Here is an example of the data that will be added to the tracker data:
```JSON
{
    "campaigns_count": 476,
    "currency": "CZK",
    "campaign_191_revenue": 1200.5,
    "campaign_191_orders_count": 12,
    "campaign_191_type": "standard",
    "campaign_192_revenue": 600,
    "campaign_192_orders_count": 6,
    "campaign_192_type": "standard",
    "campaign_194_revenue": 500,
    "campaign_194_orders_count": 5,
    "campaign_194_type": "welcome",
    "campaign_195_revenue": 800,
    "campaign_195_orders_count": 8,
    "campaign_195_type": "automation",
    "campaign_196_revenue": 600,
    "campaign_196_orders_count": 6,
    "campaign_196_type": "automatic",
    "campaign_197_revenue": 600.5,
    "campaign_197_orders_count": 6,
    "campaign_197_type": "notification",
    "campaign_201_revenue": 700,
    "campaign_201_orders_count": 7,
    "campaign_201_type": "notification"
}
```

`"campaign_XXX_revenue": 1200.0` - Sum of revenue generated by the campaign
`"campaign_XXX_type": "standard"` - Campaign type
`"campaign_XXX_orders_count": 12` - Number of orders that were created after users visited the site via the campaign

Regarding the format, please see [this comment](https://github.com/mailpoet/mailpoet/pull/4665#issuecomment-1401610401).

#### Notes
1. In the DB table, we save revenue for all clicks a subscriber did a couple of days before making an order. In the tracking data, I take into account only the first click so that every order is included only once.
2. `campaign_revenues` contains only campaigns that generated some orders.
3. We save revenue for post notifications (e.g., weekly post notification) emails separately for each sending. I grouped those data under post-notification parent so that all post notifications emails are under a single campaign.
4. In the future, I would like to extend the campaign data with some additional details (e.g., type of an automatic email, type of dynamic segment used, post notifications frequency, and sent count)

## Code review notes

_N/A_

## QA notes

There are no changes in UI.
In case you want to check what is tracked (you need to have WP CLI):
1) Setup page with Woo and MailPoet
2) Create a subscriber and add to a list
3) Send a newsletter to the list
4) Click on a link from the received email
5) Make an order when being signed in using the subscribers email (or use the email in the order)
6) Go to a command line and use: `wp wc tracker snapshot --format=yaml`
7) You should see the data from above being printed among other tracker data

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5014]

## After-merge notes

We need to update Woo usage tracking page https://href.li/?https://woocommerce.com/usage-tracking/

[MAILPOET-5014]: https://mailpoet.atlassian.net/browse/MAILPOET-5014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -89,6 +89,7 @@ class Hooks {
     $this->setupWooCommerceUsers();
     $this->setupWooCommercePurchases();
     $this->setupWooCommerceSubscriberEngagement();
+    $this->setupWooCommerceTracking();
     $this->setupImageSize();
     $this->setupListing();
     $this->setupSubscriptionEvents();
@@ -377,6 +378,16 @@ class Hooks {
       [$this->hooksWooCommerce, 'updateSubscriberEngagement'],
       7
     );
+  }
+
+  public function setupWooCommerceTracking() {
+    if ($this->wp->getOption('woocommerce_allow_tracking', 'no') === 'yes') {
+      $this->wp->addFilter(
+        'woocommerce_tracker_data',
+        [$this->hooksWooCommerce, 'addTrackingData'],
+        10
+      );
+    }
   }
 
   public function setupImageSize() {
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -381,13 +381,11 @@ class Hooks {
   }
 
   public function setupWooCommerceTracking() {
-    if ($this->wp->getOption('woocommerce_allow_tracking', 'no') === 'yes') {
-      $this->wp->addFilter(
-        'woocommerce_tracker_data',
-        [$this->hooksWooCommerce, 'addTrackingData'],
-        10
-      );
-    }
+    $this->wp->addFilter(
+      'woocommerce_tracker_data',
+      [$this->hooksWooCommerce, 'addTrackingData'],
+      10
+    );
   }
 
   public function setupImageSize() {
--- a/mailpoet/lib/Newsletter/NewslettersRepository.php
+++ b/mailpoet/lib/Newsletter/NewslettersRepository.php
@@ -203,7 +203,11 @@ class NewslettersRepository extends Repository {
       'product_purchased_in_category_emails_count' => $analyticsMap[NewsletterEntity::TYPE_AUTOMATIC][PurchasedInCategory::SLUG] ?? 0,
       'abandoned_cart_emails_count' => $analyticsMap[NewsletterEntity::TYPE_AUTOMATIC][AbandonedCart::SLUG] ?? 0,
     ];
-    $data['campaigns_count'] = $data['welcome_newsletters_count'] + $data['notifications_count'] + $data['automatic_emails_count'] + $data['automation_emails_count'] + $data['re-engagement_emails_count'] + $data['sent_newsletters_count'];
+    // Count all campaigns
+    $analyticsMap[NewsletterEntity::TYPE_AUTOMATIC] = array_sum($analyticsMap[NewsletterEntity::TYPE_AUTOMATIC] ?? []);
+    // Post notification history is not a campaign, we count only the parent notification
+    unset($analyticsMap[NewsletterEntity::TYPE_NOTIFICATION_HISTORY]);
+    $data['campaigns_count'] = array_sum($analyticsMap);
     return $data;
   }
 
--- a/mailpoet/lib/Newsletter/NewslettersRepository.php
+++ b/mailpoet/lib/Newsletter/NewslettersRepository.php
@@ -189,10 +189,12 @@ class NewslettersRepository extends Repository {
       }
     }
 
-    return [
+    $data = [
       'welcome_newsletters_count' => $analyticsMap[NewsletterEntity::TYPE_WELCOME] ?? 0,
       'notifications_count' => $analyticsMap[NewsletterEntity::TYPE_NOTIFICATION] ?? 0,
       'automatic_emails_count' => array_sum($analyticsMap[NewsletterEntity::TYPE_AUTOMATIC] ?? []),
+      'automation_emails_count' => $analyticsMap[NewsletterEntity::TYPE_AUTOMATION] ?? 0,
+      're-engagement_emails_count' => $analyticsMap[NewsletterEntity::TYPE_RE_ENGAGEMENT] ?? 0,
       'sent_newsletters_count' => $analyticsMap[NewsletterEntity::TYPE_STANDARD] ?? 0,
       'sent_newsletters_3_months' => $this->getStandardNewsletterSentCount(Carbon::now()->subMonths(3)),
       'sent_newsletters_30_days' => $this->getStandardNewsletterSentCount(Carbon::now()->subDays(30)),
@@ -201,6 +203,8 @@ class NewslettersRepository extends Repository {
       'product_purchased_in_category_emails_count' => $analyticsMap[NewsletterEntity::TYPE_AUTOMATIC][PurchasedInCategory::SLUG] ?? 0,
       'abandoned_cart_emails_count' => $analyticsMap[NewsletterEntity::TYPE_AUTOMATIC][AbandonedCart::SLUG] ?? 0,
     ];
+    $data['campaigns_count'] = $data['welcome_newsletters_count'] + $data['notifications_count'] + $data['automatic_emails_count'] + $data['automation_emails_count'] + $data['re-engagement_emails_count'] + $data['sent_newsletters_count'];
+    return $data;
   }
 
   /**
--- a/mailpoet/lib/Statistics/StatisticsWooCommercePurchasesRepository.php
+++ b/mailpoet/lib/Statistics/StatisticsWooCommercePurchasesRepository.php
@@ -51,7 +51,7 @@ class StatisticsWooCommercePurchasesRepository extends Repository {
     $revenueStatsTable = $this->entityManager->getClassMetadata(StatisticsWooCommercePurchaseEntity::class)->getTableName();
     $newsletterTable = $this->entityManager->getClassMetadata(NewsletterEntity::class)->getTableName();
 
-    return $this->entityManager->getConnection()->executeQuery('
+    $data = $this->entityManager->getConnection()->executeQuery('
       SELECT
         SUM(swp.order_price_total) AS revenue,
         COALESCE(n.parent_id, n.id) AS campaign_id,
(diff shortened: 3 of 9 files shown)


menakas:
Thanks for this proactive work to enable yourself for the hands-on @costasovo 

>I'm not sure whether it is a good idea to push campaigns this way. The data need to be transformed, but I'm not familiar with the possibilities of the transformation. From the Looker perspective, I imagined I would have data from campaigns as dimensions (+ data identifying the store so I could group those by the store) and metrics for order count, revenue sum, and average orders count. The campaigns are created by users, so the number of campaigns differs per site.


As far as I understand this code, this data will be received in the Tracker in this format for each check-in.

`wctracker.tracking_data`

<table>
<thead>
  <tr>
    <th>id</th>
    <th>url</th>
    <th>email</th>
    <th>date_checked</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>123456789 </td>
    <td>example.com</td>
    <td>admin@example.com</td>
    <td>2023-01-01 01:02:03</td>
  </tr>
</tbody>
</table>


`wctracker.tracking_data_meta`
<table>
<thead>
  <tr>
    <th>id</th>
    <th>tracking_data_id </th>
    <th>meta_key</th>
    <th>meta_value</th>
    <th>dateymd</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>00001</td>
    <td>123456789 </td>
    <td>extensions_mailpoet_campaigns_count </td>
    <td>32</td>
    <td>20230101</td>
  </tr>
  <tr>
    <td>00002</td>
    <td>123456789</td>
    <td>extensions_mailpoet_currency</td>
    <td>USD</td>
    <td>20230101</td>
  </tr>
</tbody>
</table>

However, it would be nice to have the rest of the values as well in this form, instead of an array. It is not impossible if we receive an array, but it greatly simplifies the transformations if the rest of the values can be received in the following or similar associative format ( you can choose the format, as long as it is associative). 
<table>
 <tr>
    <td>00003</td>
    <td>123456789</td>
    <td>extensions_mailpoet_campaign_id_191_revenue</td>
    <td>1200.0</td>
    <td>20230101</td>
  </tr>
  <tr>
    <td>00004</td>
    <td>123456789</td>
    <td>extensions_mailpoet_campaign_id_191_type</td>
    <td>standard</td>
    <td>20230101</td>
  </tr>
  <tr>
    <td>00005</td>
    <td>123456789</td>
    <td> extensions_mailpoet_campaign_id_191_orders_count</td>
    <td>12</td>
    <td>20230101</td>
  </tr>
</tbody>
</table>

An example is [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-tracker.php#L533-L541).

If, for some reason, you're unable to do this, it is still okay. We should be able to transform it to the format that we want.

This could be transformed to see something like below on Looker.

<table>
<thead>
  <tr>
    <th>site url</th>
    <th>campaign id</th>
    <th>campaign type</th>
    <th>month</th>
    <th>revenue</th>
    <th> orders</th>
    <th></th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>example.com</td>
    <td>191</td>
    <td>standard</td>
    <td>2023-01</td>
    <td>1200</td>
    <td>12</td>
    <td></td>
  </tr>
  <tr>
    <td>example.com</td>
    <td>192</td>
    <td>standard</td>
    <td>2023-01</td>
    <td>...</td>
    <td>...</td>
    <td></td>
  </tr>
</tbody>
</table>

Or this

<table>
<thead>
  <tr>
    <th>site url</th>
    <th>campaign type</th>
    <th>month</th>
    <th>Total revenue</th>
    <th> Total orders</th>
    <th></th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>example.com</td>
    <td>standard</td>
    <td>2023-01</td>
    <td>2500</td>
    <td>23</td>
    <td></td>
  </tr>
  <tr>
    <td>example.com</td>
    <td>welcome</td>
    <td>2023-01</td>
    <td>...</td>
    <td>...</td>
    <td></td>
  </tr>
</tbody>
</table>

>Would it be possible to use the campaign_id combined with a site domain as a composite campaign id so that we can track how a campaign performed over time?

Yes. The explore can be filtered by the campaign id and the the monthly trend can be visualised.

menakas:
@costasovo 
One way you could test if the snapshots would be fine is the way outlined in woocommerce/woocommerce#30010

You may also call `WC_Tracker::send_tracking_data(true) ` to force a snapshot send.

costasovo:
@menakas Thank you for the tips. I updated the format and updated the description of the PR with the new format.

mailpoet/lib/WooCommerce/Tracker.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\WooCommerce;
+
+class Tracker {
+  /**
+   * @param array $data
```
JanJakes: Technically, you seem to be expecting `mixed` here. Not very important, though.

mailpoet/lib/Statistics/StatisticsWooCommercePurchasesRepository.php
```diff
@@ -46,4 +46,25 @@ public function createOrUpdateByClickDataAndOrder(StatisticsClickEntity $click,
     $statistics->setSubscriber($click->getSubscriber());
     $this->flush();
   }
+
+  public function getRevenuesByCampaigns() {
```
JanJakes: We could add a return type, I guess.

mailpoet/lib/Statistics/StatisticsWooCommercePurchasesRepository.php
```diff
@@ -46,4 +46,25 @@ public function createOrUpdateByClickDataAndOrder(StatisticsClickEntity $click,
...
+    return $this->entityManager->getConnection()->executeQuery('
+      SELECT
+        SUM(swp.order_price_total) AS revenue,
+        COALESCE(n.parent_id, n.id) AS campaign_id,
+        (CASE WHEN n.type = :notification_history_type THEN :notification_type ELSE n.type END) AS campaign_type,
+        COUNT(order_id) as orders_count
+      FROM ' . $revenueStatsTable . ' swp
+        JOIN ' . $newsletterTable . ' n ON
+          n.id = swp.newsletter_id
+          AND swp.click_id IN (SELECT MIN(click_id) FROM ' . $revenueStatsTable . ' ss GROUP BY order_id)
```
JanJakes, line 65: Related to the comment [higher above](https://github.com/mailpoet/mailpoet/pull/4665/commits/2882171d8e606075ae7154ab7e0de5d951d53136#diff-6dca911f02a3f27ae96ba1f6af9bf490665746f2fca55456aca037726bbedd39R20), what you're doing here is to count each purchase ever only once, is that so?

I know that originally, the task was specified as tracking each purchase at most once per newsletter, which means it can be counted multiple times. Am I correct to assume the subquery wants to prevent that?

In any case, I think we should add a short comment explaining that. E.g.:

```
// The "SELECT MIN(click_id)..." subquery is used to count each purchase only once.
// In the UI, we count a purchase to multiple newsletters if clicks from multiple newsletters occured.
```

mailpoet/lib/Config/Hooks.php
```diff
@@ -379,6 +380,16 @@ public function setupWooCommerceSubscriberEngagement() {
     );
   }
 
+  public function setupWooCommerceTracking() {
+    if ($this->wp->getOption('woocommerce_allow_tracking', 'no') === 'yes') {
+      $this->wp->addFilter(
+        'woocommerce_tracker_data',
```
JanJakes: Just out of curiosity, as I'm unfamiliar with the tracker — do you know if this is executed by action scheduler or visitor requests?

mailpoet/lib/Newsletter/NewslettersRepository.php
```diff
@@ -201,6 +203,8 @@ public function getAnalytics(): array {
       'product_purchased_in_category_emails_count' => $analyticsMap[NewsletterEntity::TYPE_AUTOMATIC][PurchasedInCategory::SLUG] ?? 0,
       'abandoned_cart_emails_count' => $analyticsMap[NewsletterEntity::TYPE_AUTOMATIC][AbandonedCart::SLUG] ?? 0,
     ];
+    $data['campaigns_count'] = $data['welcome_newsletters_count'] + $data['notifications_count'] + $data['automatic_emails_count'] + $data['automation_emails_count'] + $data['re-engagement_emails_count'] + $data['sent_newsletters_count'];
```
JanJakes: Should we just use `array_sum($analyticsMap)`, maybe? That would include any possible new future email types. What do you consider better or safer?

mailpoet/lib/WooCommerce/Tracker.php
```diff
@@ -39,6 +44,7 @@ public function addTrackingData($data): array {
       $analyticsData = $this->newslettersRepository->getAnalytics();
       $data['extensions']['mailpoet'] = [
         'campaigns_count' => $analyticsData['campaigns_count'],
+        'currency' => $this->wooHelper->getWoocommerceCurrency(),
```
JanJakes: I think in the purchase statistics, we also include currencies. If a currency is changed or multiple are mixed for some reason, it will not match what will be tracked here. That said, maybe it's quite an edge case.

I'm not fully sure what we're doing in the UI, but I think maybe we consider only those states whose currency equals the currently set currency. That might be an option for a simple improvement, although I guess that may mean saving the currency per campaign or something.

JanJakes (commented):
LGTM! Just some small observations.

mailpoet/lib/WooCommerce/Tracker.php
costasovo: Good idea! I'll filter only revenues that are in the current shop currency.

mailpoet/lib/Statistics/StatisticsWooCommercePurchasesRepository.php
costasovo: > Am I correct to assume the subquery wants to prevent that?

Correct. I send totals per campaign and later want to aggregate using e.g. using campaign type. So I want to make sure an order is counted only once.

mailpoet/lib/Config/Hooks.php
costasovo: If I'm not mistaken, it uses WP Cron `wp_schedule_single_event`

costasovo:
@JanJakes Thank you for the review and helpful observations! I addressed them in four new commits. Assigning the PR back to you.

JanJakes (approved):

Aschepikov (approved):
GJ!

PR #4673, opened 2023-01-24
Add homepage upsell component [MAILPOET-4830]

lysyjan:
## Description

Add a new upsell section on the new homepage.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4830]

## After-merge notes

_N/A_

[MAILPOET-4830]: https://mailpoet.atlassian.net/browse/MAILPOET-4830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/css/src/components-homepage/_upsell.scss
```diff
@@ -0,0 +1,68 @@
...
+      max-width: 270px;
+    }
+  }
+}
+
+@media (max-width: 600px) {
+  .mailpoet-homepage-upsell {
+    background: $color-white;
+  }
+}
```
costasovo: Just a suggestion. Do we need to use these specific breakpoints? I always try to reuse [the mixins that we have defined for media](https://github.com/mailpoet/mailpoet/blob/9f790efbf0f032fc32e264dc33450286b362976e/mailpoet/assets/css/src/mixins/_breakpoints.scss#L13-L33). With the mixins we make all parts of page changes with the same breakpoints. I'm ok with specific values in case we need those for a reason.

costasovo (commented):
Good job! Thank you for improving the Subscribers factory for tests 👍 

I left one comment with a question and have a couple of suggestions:
1) The component should be usable in other places in the admin, so I would move the JS and CSS source of the component from the `homepage` directory to `common`. This can be also done at the time we decide to really reuse it. I'm not sure what is better.
2) I noticed that the PNG file you added has 76 KB. I tried to run it through an image optimizer, and it was reduced to 23 KB.  I thought that Figma exports optimized images, but it seems not always be the case. I'm uploading the optimized image here. I would suggest using the smaller one
upsell-illustration

mailpoet/assets/css/src/components-homepage/_upsell.scss
lysyjan: @costasovo I wanted it to look as good as possible in different widths. But using existing breakpoints indeed makes sense.

lysyjan:
Here are reactions to your suggestions, thanks for them!
> The component should be usable in other places in the admin, so I would move the JS and CSS source of the component from the homepage directory to common. This can be also done at the time we decide to really reuse it. I'm not sure what is better.

I prefer doing that at the moment of reusing. I expect other details here, like adding a property or something like that. A wrapper or width specification was also missing when I tried to use this component. Does it make sense to you? I can move JS and CSS if you think that is a better option.

> I noticed that the PNG file you added has 76 KB. I tried to run it through an image optimizer, and it was reduced to 23 KB. I thought that Figma exports optimized images, but it seems not always be the case. I'm uploading the optimized image here. I would suggest using the smaller one

Thank you, I will replace it in a new commit.

costasovo:
> I prefer doing that at the moment of reusing. I expect other details here, like adding a property or something like that. A wrapper or width specification was also missing when I tried to use this component. Does it make sense to you? I can move JS and CSS if you think that is a better option.

@lysyjan You are right! It makes better sense to do it when we know more details. So I'm totally ok with keeping it as is for now.

lysyjan:
@costasovo I addressed your suggestions in the last two commits.

costasovo (approved):
👍 Thanks for addressing all the comments!

Aschepikov (approved):
GJ!

PR #4679, opened 2023-01-25
Fix Form Editor scrollbars [MAILPOET-4607]

NeosinneR:
## Description

Remove the redundant close button in the Form Editor sidebar header, which was causing vertical scrollbars to be present in all areas (sidebar, content and settings). 

Also added animation on the toggle button in the top bar when opening/closing the sidebar.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4607]

## After-merge notes

_N/A_

[MAILPOET-4607]: https://mailpoet.atlassian.net/browse/MAILPOET-4607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-form-editor/_block-editor.scss
+++ b/mailpoet/assets/css/src/components-form-editor/_block-editor.scss
@@ -129,3 +129,11 @@ h2 {
     transform: rotate(45deg);
   }
 }
+
+// Hide block selector header with close button on desktops
+
+@include respond-to(not-small-screen) {
+  .edit-post-editor__inserter-panel-header {
+    display: none;
+  }
+}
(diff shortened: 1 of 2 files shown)


costasovo (changes requested):
Thank you for the fix!

I found that without the button, there is no way (at least I couldn't find it) to close the inserter on a small screen.

Screenshot 2023-01-25 at 8 29 44

I don't expect users to design forms fully on mobile, but they may want to adjust display settings and labels or even add a simple field. Maybe instead of removing the button, we could just hide it on a big screen?

NeosinneR:
@costasovo nice catch. I updated the PR to only hide it on big screen, so it's still visible on small screens

costasovo (approved):
Thank you! 👍

Aschepikov (approved):
GJ!

PR #4690, opened 2023-01-26
Allow editor to access automations [MAILPOET-5024]

websupporter:
## Description
This PR enables editors to use the automation feature.

## QA
When you test this PR, deactivate and activate the plugin, so the capabilities of the editor get updated. Test with the editor role.

## Linked tickets
[MAILPOET-5024]

[MAILPOET-5024]: https://mailpoet.atlassian.net/browse/MAILPOET-5024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## After merge
Deactivate and activate plugin, when you want to reap the benefits immediately.

### DIFF
--- a/mailpoet/lib/Config/AccessControl.php
+++ b/mailpoet/lib/Config/AccessControl.php
@@ -67,6 +67,7 @@ class AccessControl {
         'mailpoet_permission_manage_automations',
         [
           'administrator',
+          'editor',
         ]
       ),
     ];
--- a/mailpoet/tests/integration/REST/Automation/Automations/AutomationPutEndpointTest.php
+++ b/mailpoet/tests/integration/REST/Automation/Automations/AutomationPutEndpointTest.php
@@ -31,6 +31,24 @@ class AutomationPutEndpointTest extends AutomationTest {
     $this->assertInstanceOf(Automation::class, $this->automation);
   }
 
+  public function testEditorIsAllowed(): void {
+    wp_set_current_user($this->editorUserId);
+    $data = $this->put(
+      sprintf(self::ENDPOINT_PATH, $this->automation->getId()),
+      [
+        'json' => [
+          'name' => 'Test',
+        ],
+      ]
+    );
+
+    $this->assertSame("Test", $data['data']['name']);
+
+    $automation = $this->automationStorage->getAutomation($this->automation->getId());
+    $this->assertInstanceOf(Automation::class, $automation);
+    $this->assertSame('Test', $automation->getName());
+  }
+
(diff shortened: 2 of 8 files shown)


mailpoet/tests/integration/REST/Automation/Automations/AutomationPutEndpointTest.php
```diff
@@ -31,6 +31,29 @@ public function _before() {
...
+          'name' => 'Test',
+        ],
+      ]
+    );
+
+    $this->assertNotSame([
+      'code' => 'rest_forbidden',
+      'message' => 'Sorry, you are not allowed to do that.',
+      'data' => ['status' => 401],
+    ], $data);
```
JanJakes: This check doesn't seem to make much sense given the check just below (and in all of the other occurrences) but I guess no harm in keeping it either.

mailpoet/lib/Config/AccessControl.php
```diff
@@ -13,7 +12,7 @@ class AccessControl {
   const PERMISSION_MANAGE_SUBSCRIBERS = 'mailpoet_manage_subscribers';
   const PERMISSION_MANAGE_FORMS = 'mailpoet_manage_forms';
   const PERMISSION_MANAGE_SEGMENTS = 'mailpoet_manage_segments';
-  const PERMISSION_MANAGE_AUTOMATIONS = Engine::CAPABILITY_MANAGE_AUTOMATIONS;
+  const PERMISSION_MANAGE_AUTOMATIONS = self::PERMISSION_MANAGE_EMAILS;
```
JanJakes: We seem to be quite granular in the roles in general (managing subscribers, forms, segments, ...). Seeing that, isn't removing the automations role a bit of an inconsistency? Wouldn't it be better to just add the `editor` role as a default (same as for emails and for plugin admin)?

websupporter: > Wouldn't it be better to just add the editor role as a default (same as for emails and for plugin admin)?

~~Tbh, I tried that. For some reason, it did not work as expected~~

Ahhh! Capabilties are updated during plugin update and not on runtime! Yes, I agree, that would be better. Preparing...

websupporter:
Thank you @JanJakes! Reset and recommitted.

JanJakes (approved):

Aschepikov (approved):
GJ!

PR #4693, opened 2023-01-26
Focusable close buttons on forms [MAILPOET-4877]

johnolek:
## Description

This PR makes the close buttons of forms focusable and interactive via the keyboard (hitting space or enter will activate the button when it's focused). It also moves the close button after the form HTML so it receives focus last.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets
[MAILPOET-4877]

## After-merge notes

_N/A_

[MAILPOET-4877]: https://mailpoet.atlassian.net/browse/MAILPOET-4877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-public/_public.scss
+++ b/mailpoet/assets/css/src/components-public/_public.scss
@@ -343,10 +343,12 @@ div.mailpoet_form_popup {
 .mailpoet_form_close_icon {
   cursor: pointer;
   display: block;
+  height: 20px;
   margin: 0 0 0 auto;
   position: absolute;
   right: 10px;
   top: 10px;
+  width: 20px;
   z-index: 100002;
 }
 
--- a/mailpoet/views/form/front_end_form.html
+++ b/mailpoet/views/form/front_end_form.html
@@ -27,15 +27,6 @@
       data-editor-url="<%= editor_url %>"
     <% endif %>
   >
-    <% if form_type == 'popup' or form_type == 'fixed_bar' or form_type == 'slide_in' %>
-      <img
-        class="mailpoet_form_close_icon"
-        alt="close"
-        width=20
-        height=20
-        src='<%= image_url("form_close_icon/" ~ close_button_icon ~ ".svg") %>'
-      >
-    <% endif %>
 
     <style type="text/css">
      <%= styles|raw %>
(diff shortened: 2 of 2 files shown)


websupporter:
Thank you @johnolek for working on this.

Two minor remarks:

When using a screen reader it read "Close [image]". In case of a `<button>Close</button>` it would read "Close [push button]", which would make it clear that the user can interact with the element. So I was wondering, if we can use a `<button>` instead of an ``. If there is a reason, why you want to use ``, I would suggest the `alt` attribute to be more specific like `Click to close` or `Push button to close` or something like that.

And can we make the `close` string translatable?

johnolek:
Very good points, @websupporter! I decided to use an input with a type='image' because I didn't want to have to hide the default button styling. Does this seem like a decent approach? 

There are some other accessibility concerns I was thinking about too, though perhaps better left for another ticket. I was thinking it could be nice to automatically move focus to the popup when it appears so that a user could immediately start tabbing through the form, and then when closing the form it would be good to restore focus to the last active element before the form appeared. However, I think these changes might not make sense for all form types, or if there are multiple forms on the same page, so it could require some thinking to get it just right. Please let me know if you can think of any other low hanging fruit that would be good to include in this PR though.

websupporter:
> I decided to use an input with a type='image' because I didn't want to have to hide the default button styling. Does this seem like a decent approach?

LGTM. and my screen reader reads now "Close [push button]"

> I was thinking it could be nice to automatically move focus to the popup when it appears so that a user could immediately start tabbing through the form

I would not see a real benefit. What I think is worth arguing is, _when_ should the popup be read out loud; before you start reading the content or after? In the DOM it is currently placed after the content (as we filter `the_content` and append it). I think an argument could be made to read it first or to give an option to the admin, but not sure about the actual benefits.

Should we move the focus forcefully? I don't think so. I was thinking about a "modal"-trap as well (shift the focus to the modal and trap it inside until you close). Such a behavior makes sense, when a modal opens for a critical information or decision inside an app. At least from an a11y POV, this seems not to be the case here. Not sure, if it does from a conversion POV.

Feel free to open a ticket, where this can be discussed more broadly and from different perspectives.

Thank you for working on it @johnolek

websupporter (approved):
lgtm

johnolek:
Thanks for your additional thoughts, @websupporter! The way I was thinking about it was more like, "How can we make the user experience as similar as possible for sighted people and people using screen readers?" If a modal form basically forces a sighted user to interact with it before they can read the rest of the content, should we make it the same for screen readers?

That being said, I'm happy to leave things how they are for now. I think it could become rather complex to achieve that kind of parity and it probably isn't worth it considering how it would probably be a jarring experience for a screen reader no matter what.

Aschepikov (approved):
GJ!

PR #4698, opened 2023-01-27
Prevent captcha audio to get out of sync when loaded range wise [MAILPOET-5032]

websupporter:
## Description
In Safari the build in captcha does not always work. The reason is that Safari sends three requests to load the captcha
* 1 normal request for the image
* 1 request where it just requests the first byte of the audio
* 1 request where it requests the whole audio

This broke the logic when to create a new captcha phrase. This PR fixes this issue.

A second issue was discovered where the reload button would not properly work in Safari. This PR fixes also this issue.

## QA notes
I was able to reproduce the issues on Mac Ventura and Safari 16

## Linked tickets

[MAILPOET-5032]

[MAILPOET-5032]: https://mailpoet.atlassian.net/browse/MAILPOET-5032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
+++ b/mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
@@ -23,7 +23,7 @@ class CaptchaRenderer {
   public function renderAudio($sessionId, $return = false) {
 
     $audioPath = Env::$assetsPath . '/audio/';
-    $phrase = $this->phrase->getPhraseForType('audio', $sessionId);
+    $phrase = $this->phrase->getPhraseForType($this->determineAudioType(), $sessionId);
     $audio = null;
     foreach (str_split($phrase) as $character) {
       $file = $audioPath . strtolower($character) . '.mp3';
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
```diff
@@ -45,6 +45,14 @@ public function renderAudio($sessionId, $return = false) {
     exit;
   }
 
+  private function determineAudioType(): string {
+    $type = 'audio';
+    if (isset($_SERVER['HTTP_RANGE'])) {
+      $type .= '-range-' . sanitize_text_field(wp_unslash($_SERVER['HTTP_RANGE']));
```
lysyjan, line 51: We usually wrap all WP functions in our class [Functions.php](https://github.com/mailpoet/mailpoet/blob/d9eca55189e12f5f5700b09f6da5906535741694/mailpoet/lib/WP/Functions.php). We could add functions `wp_unslash` and `sanitize_text_field` too. What do you think?

lysyjan (commented):
It looks good! I have just one small remark.

mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
websupporter: Thanky @lysyjan,
we don't do this for some methods like `sanitize_text_field` and `wp_unslash`, because we use the default WordPress coding standards, which would otherwise complain that we do not sanitize those fields.

References:
Usages of `sanitize_text_field`: https://github.com/mailpoet/mailpoet/search?q=sanitize_text_field
Existence of `sanitizeTextField`: none; https://github.com/mailpoet/mailpoet/search?q=sanitizeTextField

lysyjan: 👍 Thank you for the explanation. I didn't realize that.

lysyjan (approved):

Aschepikov (approved):
GJ!

PR #4701, opened 2023-01-27
Homepage resources [MAILPOET-4829]

costasovo:
## Description
This PR adds a section with educational resources to the homepage.

Screenshot 2023-01-27 at 15 06 41

## Code review notes
I recommend checking the PR commit by commit.

If you want to test locally in dev env and want to display images in the section, set `define( 'MAILPOET_USE_CDN', false );` in wp-config.php.

## QA notes
The images will not be displayed before release. Please ping @costasovo or some other dev to push the images from this PR to SVN.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4829]

## After-merge notes

_N/A_

[MAILPOET-4829]: https://mailpoet.atlassian.net/browse/MAILPOET-4829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/homepage/components/resources.tsx
```diff
@@ -0,0 +1,13 @@
+import { MailPoet } from 'mailpoet';
+import { ContentSection } from './content-section';
+
+export function Resources(): JSX.Element {
```
samnajian, line 8: I'd suggest refraining from explicitly setting the return type and letting TS type inference do its job

mailpoet/assets/js/src/homepage/components/resources.tsx
```diff
@@ -52,7 +56,48 @@ export function Resources(): JSX.Element {
...
+        ) : (
+          <Icon icon={chevronLeft} />
+        )}
+
+        {activePage < Math.ceil(posts.length / 2) ? (
+          <a
+            href="#"
+            onClick={(e) => {
+              e.preventDefault();
+              setActivePage(activePage + 1);
```
samnajian: Defining this as a function can help reduce the number of functions created in the render cycle and besides this can be reused on line 74 as well

samnajian (commented):
Looking great, left some suggestions/comments

mailpoet/assets/js/src/homepage/components/resources.tsx
costasovo: @samnajian Can you please explain why it is better to leave it on TS?

I'm asking because I see that we use this return type in almost all new components. Personally, I think that it is a good idea to explicitly define return types for simple functions where the type plays a role as a guard so that, for example, a function that is expected to return a number wouldn't suddenly start returning null. But I'm not sure about the components.

samnajian: I'd agree 100% with normal functions/methods, so that we don't need to check the function to know what is returns, but for components, they are well, components! the return type is clear us and TS, and adding explicit types is simply adding code for no special benefit, they are redundant, and besides TS is trying hard to be more smart in inferring types and nag while building ( and we watch the code with TS ), so leaving the job to TS type inference is safe and reduces the number of characters we end up writing, still please note all of these is simply a suggestion and explicitly adding return types doesn't have any known negative effects.

mailpoet/assets/js/src/homepage/components/resources.tsx
costasovo: Good point! I found that in order to avoid creating anonymous functions, I needed to add two functions because in the second place I need to set `activePage - 1`, and creating a new function that would accept that as a parameter would mean calling in in an anonymous function. I added those in a new commit.

mailpoet/assets/js/src/homepage/components/resources.tsx
costasovo: Thank you! I understand. So I choose to keep those types to be consistent with most of our codebase.

costasovo:
@samnajian Thank you for your review! I addressed the note about anonymous functions in a new commit.

samnajian (dismissed):
LGTM, thanks!

Aschepikov:
Hi, I have some tiny issues to correct
1. Images are wrong for the first articles, you need to swap them
image
2. No dot in the Create an Email: Types of Campaigns description 
image

costasovo:
Nice catch @Aschepikov! Thank you!
@samnajian I added two new commits. The first one addresses the issues found by @Aschepikov, and in the second one, I had to update an acceptance test which started to fail because of a change in scroll position after I uploaded images to CDN.

samnajian (approved):
Thanks for the fixes.

Aschepikov (approved):
Thank you, GJ!

PR #4707, opened 2023-01-31
Add new test Settings basics

veljkho:
## Description

Adding a new test for covering Settings - Basics saving only

## Code review notes

Additionally some changes:
- added `default` to tests so they can be run individually.
- renamed `login` to `authenticate`, as this was a more standard name for what the function does.
- deleted tests from the `nightly` scenario, will be `// TBD` for now
- increased timing for the tests, this will be configured at the end of the project, we don't know the exact time wanted atm.

Executing tests locally is not set yet properly (no data to test against). However, executing against `qawp.net` site is possible, so please ping me if you'd like to test that way.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4960](https://mailpoet.atlassian.net/browse/MAILPOET-4960)

[MAILPOET-4960]: https://mailpoet.atlassian.net/browse/MAILPOET-4960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/utils/helpers.js
+++ b/mailpoet/tests/performance/utils/helpers.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-shadow */
 /* eslint-disable import/no-unresolved */
+/* eslint-disable import/no-default-export */
 /**
  * Internal dependencies
  */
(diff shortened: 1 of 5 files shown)


mailpoet/tests/performance/utils/helpers.js
```diff
@@ -1,16 +1,20 @@
...
-export function login(page) {
+export function authenticate(page) {
   // Enter login credentials and login
-  page.locator('input[name="log"]').type(`${adminUsername}`);
-  page.locator('input[name="pwd"]').type(`${adminPassword}`);
+  Promise.all([
+    page.waitForNavigation({ waitUntil: 'networkidle' }),
+    page.locator('input[name="log"]').type(`${adminUsername}`),
+    page.locator('input[name="pwd"]').type(`${adminPassword}`),
+  ]);
```
JanJakes: The promise here is not handled in any way. Did you mean to `await` it, for instance?

JanJakes (commented):
> Additionally some changes:

Generally, our Git rule is to put each isolated change into its own commit. No need to change it now but it makes a clearer PR and Git history.

I have one question about a promise in the code.

mailpoet/tests/performance/utils/helpers.js
veljkho: Sorry for Git, I will follow the rule from now on.
For the Promise, should be some leftover from attempts to fix some race condition, but it works with/without Promise fine in my tests so I removed it. Thanks for the review!

JanJakes (approved):

PR #4714, opened 2023-02-03
Run workflow multiple times per subscriber [MAILPOET-4966]

websupporter:
## Description
Adds a small toggle to our two triggers which enables us to run an automation only once per subscriber:
image

## Code review notes
1. I moved some helper functions in the IntegrationTester class. Seemed more appropriate there.
2. Do you think we should just remove the "previouslyScheduled"-check in SendEmail or check if - in case there is are previously scheduled emails if they are all send? I opted for removing it completely, but maybe there is a good reason for not sending the email a 2nd time when we haven't yet send the first one :thinking: 
3. I call the extra db column `subject_hash`but I do not actually hash the value, as I do not see a benefit in doing so. But currently I lack of a better name.

## QA notes

Check if  a subscriber gets an email multiple times when the setting of the trigger is accordingly and vice versa.

## Linked tickets
[MAILPOET-4966]

## After-merge notes

_N/A_

[MAILPOET-4966]: https://mailpoet.atlassian.net/browse/MAILPOET-4966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/someone-subscribes/edit/index.tsx
@@ -1,5 +1,11 @@
 import { ListPanel } from './list_panel';
+import { RunOnlyOncePanel } from '../../../../shared/run-only-once-panel';
 
 export function Edit(): JSX.Element {
-  return <ListPanel />;
+  return (
+    <>
+      <RunOnlyOncePanel />
+      <ListPanel />
+    </>
+  );
 }
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/wp-user-registered/edit/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/wp-user-registered/edit/index.tsx
@@ -1,5 +1,11 @@
 import { RolePanel } from './role_panel';
+import { RunOnlyOncePanel } from '../../../../shared/run-only-once-panel';
 
 export function Edit(): JSX.Element {
-  return <RolePanel />;
+  return (
+    <>
+      <RunOnlyOncePanel />
+      <RolePanel />
+    </>
+  );
 }
--- a/mailpoet/assets/js/src/automation/integrations/shared/run-only-once-panel/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/shared/run-only-once-panel/index.tsx
@@ -1,0 +1,35 @@
+import { PanelBody, ToggleControl } from '@wordpress/components';
+import { dispatch, useSelect } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
+import { storeName } from '../../../editor/store';
+import { PlainBodyTitle } from '../../../editor/components';
+
+export function RunOnlyOncePanel(): JSX.Element {
+  const { selectedStep } = useSelect(
+    (select) => ({
+      selectedStep: select(storeName).getSelectedStep(),
+    }),
+    [],
+  );
+  const checked: boolean = selectedStep.args?.run_multiple_times as boolean;
+  return (
+    <PanelBody opened>
+      <PlainBodyTitle title={__('Automation setting', 'mailpoet')} />
+
+      <ToggleControl
+        label={__(
+          'A subscriber can enter this automation multiple times.',
+          'mailpoet',
+        )}
+        checked={checked}
+        onChange={(value) => {
+          dispatch(storeName).updateStepArgs(
+            selectedStep.id,
+            'run_multiple_times',
+            value,
+          );
+        }}
+      />
+    </PanelBody>
+  );
+}
--- a/mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
@@ -61,7 +61,13 @@ class TriggerHandler {
         $entry->getPayload();
       }
 
-      $automationRun = new AutomationRun($automation->getId(), $automation->getVersionId(), $trigger->getKey(), $subjects);
+      $automationRun = new AutomationRun(
+        $automation->getId(),
+        $automation->getVersionId(),
+        $trigger->getKey(),
+        $subjects,
+        $trigger->getSubjectHash($subjectEntries)
+      );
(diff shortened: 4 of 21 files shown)


JanJakes:
@websupporter I managed only a quick look today, but I'll reply at least to some of the main questions:

> 1. I moved some helper functions in the IntegrationTester class. Seemed more appropriate there.

Do we already use a similar pattern, or is that new? Do we have any data factory code reuse in the integration tests? I'm just thinking that if this is new, we may as well introduce some kind of wrapper class to have something like `$this->dataFactory->createAutomation(...)` and being able to split the factories to multiple files. Or maybe traits could work, or inheritance (like a base class for automation tests), not sure.

> 2. Do you think we should just remove the "previouslyScheduled"-check in SendEmail or check if - in case there is are previously scheduled emails if they are all send? I opted for removing it completely, but maybe there is a good reason for not sending the email a 2nd time when we haven't yet send the first one 🤔

I'm not aware of any specific reasons, but we need to be very cautious not to send duplicates due to spam detection. If there's any case where that step would know for sure "this is not another run but a weird duplicate", then it's still better to throw an exception than sending it. But I guess we're unable to detect that now.

> 3. I call the extra db column `subject_hash` but I do not actually hash the value, as I do not see a benefit in doing so. But currently I lack of a better name.

The question is what exactly does it do. At a first sight, I'd call it a `subject_key`, but that might be confused with our other "keys" in automations so `subject_index`, maybe. That said, I have some more questions about that below.

---

I'll add one general question here. How's this supposed to work with **multiple triggers**? What I mean is — if trigger A has the new checkbox enabled, trigger B disabled, and they both produce a subscriber (or are even the same trigger), what should happen then? Is this always per-trigger, meaning that you can enter once via trigger A, but multiple times via trigger B?

So far so good, I guess, but what if both trigger A and B have the new checkbox disabled? That would still enable entering the workflow twice — once using each trigger.

I'm asking because it's important for the implementation, as well as for the data storage. It's sort of the equivalent of asking "if we use a `unique` index in the DB, which columns of the automation runs table should it include?" — no matter if we use the index or not, it's expressed at query time.

And going even a bit further — what if I want to implement a trigger that would have two of such checkboxes for two different subjects? Maybe that's unlikely and confusing for users, just thinking...

websupporter:
> Do we already use a similar pattern, or is that new?

We have some helper functions there e.g. for WooCommerce. E.g. the `createWooCommerceOrder()` method is used here: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/FirstPurchaseTest.php#L239-L240

I see, we do not have a clear line between this Tester and the DataFactories. E.g. it exists a data factory for WooCommerce orders, but this factory is only used in the acceptance tests. Whereas the Subscriber DataFactory is also used in the integration tests.

Thinking about it, a Data Factory is the better approach if you think how many different data factory methods would need to go into the tester.

> But I guess we're unable to detect that now.

That is also my feeling atm.

> multiple triggers

I discussed this briefly with @neosinner in the ticket [MAILPOET-4966]. (Reading it again, I see I missed realizing another specification, which was produced in this discussion, but thats another topic)

My problem right now is, I do not really see how a good archtiecture to move this to being an "automation setting" rather than an "trigger setting". What I mean by that is, currently I do not see a way, which would solve this problem on an Automation level.

For me, the task is currently solveable in a mental model which says: The automation is a house and the trigger is a entrance door. The subject or subject combination is something like the key and on a "per-door" basis you can say a specific key works only once. Once we come to multiple doors, we would have to establish the correct mental model, phrasing. The label "Automation setting" and "A subscriber can enter this automation multiple times" therefore activates the wrong mental model.

But I would love it, if we could find a good way to really solve the issue on "per automation" basis.

[MAILPOET-4966]: https://mailpoet.atlassian.net/browse/MAILPOET-4966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

JanJakes:
@websupporter

> Whereas the Subscriber DataFactory is also used in the integration tests.

Ohh, really? I thought those data factories were only for acceptance tests. But maybe someone tried to use it in integration tests as well, and it worked 😆 I'm not sure if those particular data factories are a good fit for integrations tests, haven't thought about that. What I meant is more like "should we extract those factories to some other files", but ultimately, we can always leave it for a later refactor.

> I discussed this briefly with @NeosinneR in the ticket [MAILPOET-4966](https://mailpoet.atlassian.net/browse/MAILPOET-4966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ).

I browsed the specs to search for that, but I missed the comments below. I guess, to make it reasonably simple and understandable in the UI, it makes sense to keep what we have now without solving all the edge cases. I was thinking if we can improve the wording somehow, but looking at it from the per-trigger perspective, the wording makes sense.

---

Adding to the discussion for some future per-automation settings, it might be doable to say something like _"run this automation only once per these subject instances"_, and maybe the subjects could even expose the settings or at least some defaults.

But I guess it's hard to keep this reasonably simple in the UI, and it makes total sense to have it per-trigger for all workflows with a single trigger. So perhaps we can just improve the UX once we add multi-trigger support, to clearly explain what's happening 🤔 

Anyway, thanks for the explanation, I'll look at the code a bit closer now.

[MAILPOET-4966]: https://mailpoet.atlassian.net/browse/MAILPOET-4966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

JanJakes:
I have one more question, and I guess we could bring @NeosinneR here as well, I think this wasn't really in the ticket.

At the moment, the new setting is in its own section above the trigger settings. Wouldn't it be better to move below the list selection (as it seems to be a bit more advanced setting) and maybe to the trigger settings (as it's per-trigger setting) or renaming the title to something else (like "Repeated runs" or how to say it better)?

What do you think? Or do you prefer the current implementation (below)?

Screenshot 2023-02-14 at 11 07 13

PR #4721, opened 2023-02-08
Migration fixes [MAILPOET-5054]

JanJakes:
## Description

Includes two fixes resulting in errors like https://wordpress.org/support/topic/migration-class-mailpoetmigrationsmigration_20221108_140545-not-found/:
1. Always skip unknown migrations, not only when they are completed (e.g. if a migration is marked as failed in the DB but doesn't exist at all, it should not be retried).
2. Cast empty errors to `NULL`. This is a fix for 3rd party plugins corrupting our data: https://wordpress.org/support/topic/null-values-are-exported-as-empty-strings/

## Code review notes

_N/A_

## QA notes

1. Install and activate the plugin.
3. Go to the `<prefix>_mailpoet_migrations` table and change some `error` records from `NULL` to an empty string.
4. Deactivate and activate the plugin. You should see no errors.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5054]

## After-merge notes

_N/A_

[MAILPOET-5054]: https://mailpoet.atlassian.net/browse/MAILPOET-5054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Migrator/Cli.php
+++ b/mailpoet/lib/Migrator/Cli.php
@@ -109,6 +109,8 @@ class Cli {
 
       WP_CLI::log("MIGRATIONS:\n");
       $table = array_map(function (array $data): array {
+        $data['name'] .= $data['unknown'] ? ' (unknown)' : '';
+        unset($data['unknown']);
         return array_map(function ($field) {
           return $field === null ? '' : $field;
         }, $data);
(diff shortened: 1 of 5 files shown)


johnolek (approved):
This looks good to me. Thanks so much for your help with this and the quick fix!

Aschepikov (approved):
GJ!

PR #4733, opened 2023-02-15
Change design of the coupon block [MAILPOET-4890]

lysyjan:
## Description

This PR primarily refactors the coupon block settings to React because it simplifies designing the settings panel.
Also, it was necessary to extend or add some functionality.

## Code review notes

I would recommend reviewing the PR by commits.
Unfortunately, I had to remove one of the test cases for coupon block in newsletter editor tests.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4890]

## After-merge notes

Don't forget to extend the ticket with test improvements.

[MAILPOET-4890]: https://mailpoet.atlassian.net/browse/MAILPOET-4890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/AdminPages/Pages/NewsletterEditor.php
```diff
@@ -136,6 +144,8 @@ public function render() {
       'is_wc_transactional_email' => $newsletterId === $woocommerceTemplateId,
       'is_confirmation_email_template' => $newsletterId === $confirmationEmailTemplateId,
       'is_confirmation_email_customizer_enabled' => (bool)$this->settings->get('signup_confirmation.use_mailpoet_editor', false),
+      'product_categories' => $this->wpPostListLoader->getWooCommerceCategories(),
+      'products' => $this->wpPostListLoader->getProducts(),
```
triple0t, line 148: Suggestion: Why not move these two into the [woocommerceData](https://github.com/mailpoet/mailpoet/blob/d06447b48bff6ec6a2c29088c9e2ab91030e258f/mailpoet/lib/AdminPages/Pages/NewsletterEditor.php#L112) section?

This would prevent any error that product_categories & products could cause. e.g, when WooCommerce is not available

mailpoet/assets/js/src/newsletter_editor/blocks/coupon/usage_restriction.tsx
```diff
@@ -0,0 +1,275 @@
...
+type Post = {
+  id: number;
+};
+
+type Props = {
+  getValueCallback: (name: string) => string | boolean | Backbone.Collection;
+  setValueCallback: (
+    name: string,
+    value: string | boolean | Backbone.Collection,
+  ) => void;
```
triple0t: I noticed we repeat this Type declaration in a few other files. I would suggest creating a reusable Type that you can import and use everywhere else.

mailpoet/assets/js/src/newsletter_editor/blocks/coupon/usage_limits.tsx
```diff
@@ -0,0 +1,80 @@
...
+import Backbone from 'backbone';
+import { Component } from '@wordpress/element';
+import { MailPoet } from 'mailpoet';
+
+type Props = {
+  getValueCallback: (name: string) => string | boolean | Backbone.Collection;
+  setValueCallback: (
+    name: string,
+    value: string | boolean | Backbone.Collection,
+  ) => void;
```
triple0t: Same here as well

mailpoet/assets/js/src/newsletter_editor/blocks/coupon/usage_limits.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return (
+      <Panel>
+        <PanelBody
+          title={MailPoet.I18n.t('usageLimits')}
+          className="mailpoet-coupon-block-usage-limits"
+          initialOpen={false}
+        >
+          <PanelRow>
+            <TextControl
+              className="mailpoet_field_coupon_usage_limits_per_coupon"
```
triple0t: Does this `className` serve any purpose here?
If not, May I suggest removing it?

mailpoet/assets/js/src/newsletter_editor/blocks/coupon/usage_limits.tsx
```diff
@@ -0,0 +1,80 @@
...
+              type="number"
+              onChange={(usageLimit) => {
+                this.setValueCallback('usageLimit', usageLimit);
+                this.setState({ usageLimit });
+              }}
+            />
+          </PanelRow>
+          <PanelRow>
+            <TextControl
+              className="mailpoet_field_coupon_usage_limits_per_user"
```
triple0t: Same here

mailpoet/assets/js/src/newsletter_editor/blocks/coupon/usage_restriction.tsx
```diff
@@ -0,0 +1,275 @@
...
+            <Selection
+              field={{
+                ...productsField,
+                name: 'productIds',
+                selected: () =>
+                  this.state.productIds.map((product: Post) => product.id),
+              }}
+              onValueChange={this.handleSelection}
+            />
+          </PanelRow>
```
triple0t, line 182: Can this component take all the products on a site?
I noticed you removed the pagination. This means this component will contain all the products available on the site.

triple0t (commented):
Such an incredible amount of work. Well done, @lysyjan.

I have a few comments.

1. EmailRestrictions does not allow multiple emails
   Screenshot 2023-02-16 at 1 24 40 PM - emailRestrictions

   Screenshot 2023-02-16 at 2 01 15 PM

   As you may notice, we are unable to save multiple emails in the `emailRestrictions` field. This should be allowed.

2. Product ids and Categories id not saving.
I assume you have to update these [methods](https://github.com/mailpoet/mailpoet/blob/d06447b48bff6ec6a2c29088c9e2ab91030e258f/mailpoet/lib/WooCommerce/CouponPreProcessor.php#L121-L127)
   Screenshot 2023-02-16 at 1 37 12 PM

3. Error in the console
   Screenshot 2023-02-16 at 1 27 31 PM - console
   I got this error when I tried editing some input fields. e.g, Minimum spend

4. Kind reminder to create a follow-up ticket for the tests removed in this [commit](https://github.com/mailpoet/mailpoet/pull/4733/commits/d06447b48bff6ec6a2c29088c9e2ab91030e258f)

5. "Display options" button/link is really easy to miss. Are we sure we want to keep it hidden by default? 

   Screenshot 2023-02-16 at 1 14 02 PM - Display options

6. Can we duplicate the "Back to selection" button/link? I suggest we add to the selection panel's top and bottom. It's really easy to miss as well

   Screenshot 2023-02-16 at 1 16 13 PM - Back to selection

lysyjan:
@triple0t Thank you for the detailed review with nice catches. Those are big changes that I wanted to move forward with.
I want to address some of your findings from the list.

> 1. EmailRestrictions does not allow multiple emails

I tried it again on my local, and I think the space isn't allowed behind the comma. In my opinion, it's fine. I also tested the behavior in the WC plugin, and they allow the spaces, but after saving are removed. This would mean more work for this PR, and if you think that is important, I prefer addressing it in another ticket.
Email restirctions

> 2. Product ids and Categories id not saving.

Nice catch. I will fix it.

> 3. Error in the console

This can happen when the initial value is `null` or `undefined`. I will fix it. Thanks.

> 4. Kind reminder to create a follow-up ticket for the tests removed in this [commit](https://github.com/mailpoet/mailpoet/pull/4733/commits/d06447b48bff6ec6a2c29088c9e2ab91030e258f)

I'm still thinking about those tests. We have an existing ticket for test improvements. I will add it to after merge notes.

> 5. "Display options" button/link is really easy to miss. Are we sure we want to keep it hidden by default?

This is based on the design in Figma. I made it a little bit different and followed our current approach in the other blocks.

> 6. Can we duplicate the "Back to selection" button/link? I suggest we add to the selection panel's top and bottom. It's really easy to miss as well

Yes, we can. But I would expect some spacing or separator here.

@pavel-mailpoet Please, can you tell us your opinion on points 5 and 6?
For me, those are details that can be decided and improved later.

pavel-mailpoet:
@lysyjan 

5. "Display options" is consistent with ALC and seems fine to me so I would keep it as you did.

6. I agree with you. If it is easy, you can duplicate the button, but please add some margin so it looks good.


Btw. nice animation you did when switching to and from the display option. If this was not the legacy editor I would probably create a new ticket to add more animations when switching between other tabs, but it is not worth for the old editor.

lysyjan:
>Btw. nice animation you did when switching to and from the display option. If this was not the legacy editor I would probably create a new ticket to add more animations when switching between other tabs, but it is not worth for the old editor.

@pavel-mailpoet It's reused style from the other blocks.

mailpoet/lib/AdminPages/Pages/NewsletterEditor.php
lysyjan: I have checked the code, and if I am not wrong, the result is empty when WC is not installed. So it looks safe to me.

Product and categories are here because the selection component reads them by the property `endpoint`. I used this approach because is also used in the dynamic segments form.

mailpoet/assets/js/src/newsletter_editor/blocks/coupon/usage_restriction.tsx
lysyjan: I noticed it during the implementation, but in the end, I forgot. I will change it.

mailpoet/assets/js/src/newsletter_editor/blocks/coupon/usage_limits.tsx
lysyjan: I added those classes automatically because I didn't know if I needed them in one of my following improvements, such as design or validation. I can remove them.

mailpoet/lib/AdminPages/Pages/NewsletterEditor.php
lysyjan: > Can this component take all the products on a site?
I noticed you removed the pagination. This means this component will contain all the products available on the site.

I will continue this topic here because it is related. I am aware of it, and we use this approach in dynamic segments.
But I agree it is not the best solution for sites with a higher amount of products. I wanted to change my solution and use the property `remoteQuery`([#](https://github.com/mailpoet/mailpoet/blob/d116ff360b8d73f0ae09429ae0c365ed30a54def/mailpoet/assets/js/src/form/fields/selection.jsx#L150-L182)). Unfortunately, I cannot map the response from the endpoint `automatedLatestContent` without significant changes as it is [here](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/newsletter_editor/blocks/coupon.ts#L434-L447).

I prefer to create a new ticket because this PR covers a lot of other changes. The previous solution contained this benefit, but I don't want a huge PR and want to unblock the rest of the tasks. Do you agree?

lysyjan:
@triple0t I tried to address all your remarks.
Unfortunately, when I tried to add `Back to selection link`, it didn't look good, in my opinion. I would prefer a tip from a designer here. I guess that when we prepare a demo we will have time for those details.

mailpoet/lib/AdminPages/Pages/NewsletterEditor.php
triple0t: Yeah, that is fine with me.

triple0t (approved):
LGTM.
Well done 👏

Aschepikov (approved):
GJ!

PR #4735, opened 2023-02-15
Fix mixpanel conflict [MAILPOET-5065]

### LINKED ISSUE #4712: JS conflict when another plugin loads mixpanel script
**Describe the bug**
There are some other plugins that load MixPanel to track some metrics and if that happen, we can see the following error in console

image

and this is happening because we define MailPoet global variable here:
https://github.com/mailpoet/mailpoet/blob/05202d572f8c6475d3704e7d0df25d9a6dfadc2b/mailpoet/assets/js/src/mailpoet.ts#L83

and before that we call this analytics_event
https://github.com/mailpoet/mailpoet/blob/05202d572f8c6475d3704e7d0df25d9a6dfadc2b/mailpoet/assets/js/src/mailpoet.ts#L8-L9

which calls this function globally:
https://github.com/mailpoet/mailpoet/blob/47e46aeebed11268b4a5ade6b84223a24ca98ea3/mailpoet/assets/js/src/analytics_event.js#L93

So I believe the main problem comes from this condition:
https://github.com/mailpoet/mailpoet/blob/47e46aeebed11268b4a5ade6b84223a24ca98ea3/mailpoet/assets/js/src/analytics_event.js#L82

it should be something like that:
```
if (typeof window.mixpanel === 'object' && typeof window.MailPoet === 'object') {
```

In the code u depended on that your mixpanel script will not be loaded until u define this global variable MailPoet but we need to make sure that this global variable is there before accessing and altering it.

What do u think?
If it's OK, I can create this small PR to fix it.

**To reproduce**

1. Use the following code inside your wordpress
```
function ahmed_add_mixpanel_code() {
	?>
	<!-- start Mixpanel --><script type="text/javascript">(function(e,a){if(!a.__SV){var b=window;try{var c,l,i,j=b.location,g=j.hash;c=function(a,b){return(l=a.match(RegExp(b+"=([^&]*)")))?l[1]:null};g&&c(g,"state")&&(i=JSON.parse(decodeURIComponent(c(g,"state"))),"mpeditor"===i.action&&(b.sessionStorage.setItem("_mpcehash",g),history.replaceState(i.desiredHash||"",e.title,j.pathname+j.search)))}catch(m){}var k,h;window.mixpanel=a;a._i=[];a.init=function(b,c,f){function e(b,a){var c=a.split(".");2==c.length&&(b=b[c[0]],a=c[1]);b[a]=function(){b.push([a].concat(Array.prototype.slice.call(arguments,
			0)))}}var d=a;"undefined"!==typeof f?d=a[f]=[]:f="mixpanel";d.people=d.people||[];d.toString=function(b){var a="mixpanel";"mixpanel"!==f&&(a+="."+f);b||(a+=" (stub)");return a};d.people.toString=function(){return d.toString(1)+".people (stub)"};k="disable time_event track track_pageview track_links track_forms register register_once alias unregister identify name_tag set_config reset people.set people.set_once people.increment people.append people.union people.track_charge people.clear_charges people.delete_user".split(" ");
			for(h=0;h<k.length;h++)e(d,k[h]);a._i.push([b,c,f])};a.__SV=1.2;b=e.createElement("script");b.type="text/javascript";b.async=!0;b.src="undefined"!==typeof MIXPANEL_CUSTOM_LIB_URL?MIXPANEL_CUSTOM_LIB_URL:"file:"===e.location.protocol&&"//cdn.mxpnl.com/libs/mixpanel-2-latest.min.js".match(/^\/\//)?"https://cdn.mxpnl.com/libs/mixpanel-2-latest.min.js":"//cdn.mxpnl.com/libs/mixpanel-2-latest.min.js";c=e.getElementsByTagName("script")[0];c.parentNode.insertBefore(b,c)}})(document,window.mixpanel||[]);
	</script><!-- end Mixpanel -->
	<?php
}
add_action( 'admin_print_scripts', 'ahmed_add_mixpanel_code' );
```
2. Visit MailPoet settings page
3. Check console, you will see an error and the page will not load at all.

**Actual behavior**
Console error that prevents the settings page from being loaded at all.

**Expected behavior**
Settings page should appear without any error in the console.

**Screenshots**
Added above.

- WordPress version: **6.1.1**
- PHP version: PHP **7.4.32** (cli) (built: Sep 29 2022 22:24:52) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Xdebug v3.1.5, Copyright (c) 2002-2022, by Derick Rethans
    with Zend OPcache v7.4.32, Copyright (c), by Zend Technologies
- MailPoet version: **4.6.0**

**Additional context**

costasovo:
## Description
When another plugin or theme loads mixpanel library it may cause that MailPoet's pages crash with an error.

See details in https://github.com/mailpoet/mailpoet/issues/4712

The initializeMixpanelWhenLoaded expects window.MailPoet to be defined. Instead of calling initializeMixpanelWhenLoaded as a side-effect of the import we can export the function and make sure it is called really after window.MailPoet is defined. 

## Code review notes

_N/A_

## QA notes

Please verify that Mixpanel tracking works. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5065] , Closes https://github.com/mailpoet/mailpoet/issues/4712

## After-merge notes

_N/A_

[MAILPOET-5065]: https://mailpoet.atlassian.net/browse/MAILPOET-5065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/analytics_event.js
+++ b/mailpoet/assets/js/src/analytics_event.js
@@ -78,7 +78,7 @@ function cacheEvent(forced, name, data, options, callback) {
   }
 }
 
-function initializeMixpanelWhenLoaded() {
+export function initializeMixpanelWhenLoaded() {
   if (typeof window.mixpanel === 'object') {
     exportMixpanel();
     trackCachedEvents();
@@ -89,5 +89,3 @@ function initializeMixpanelWhenLoaded() {
 
 export const MailPoetTrackEvent = _.partial(cacheEvent, false);
 export const MailPoetForceTrackEvent = _.partial(cacheEvent, true);
-
-initializeMixpanelWhenLoaded();
--- a/mailpoet/assets/js/src/mailpoet.ts
+++ b/mailpoet/assets/js/src/mailpoet.ts
@@ -5,8 +5,11 @@ import { MailPoetDate } from './date';
 import { MailPoetAjax } from './ajax';
 import { MailPoetModal } from './modal';
 import { MailPoetNotice } from './notice';
-// side effect - extends MailPoet object in initializeMixpanelWhenLoaded
-import { MailPoetForceTrackEvent, MailPoetTrackEvent } from './analytics_event';
+import {
+  MailPoetForceTrackEvent,
+  MailPoetTrackEvent,
+  initializeMixpanelWhenLoaded,
+} from './analytics_event';
 import { MailPoetNum } from './num';
 import { MailPoetHelpTooltip } from './help-tooltip-helper';
 import { MailPoetIframe } from './iframe';
@@ -82,3 +85,6 @@ declare global {
 
 // Expose MailPoet globally
 window.MailPoet = MailPoet;
+
+// initializeMixpanelWhenLoaded needs to be called after window.MailPoet is defined.
+initializeMixpanelWhenLoaded();


samnajian (dismissed):
LGTM, I'd suggest adding the comment ```make sure it is called after window.MailPoet is defined.```  (or alike, it's just a part of your commit message) before invoking ` initializeMixpanelWhenLoaded ` to prevent future issues of the same kind

costasovo:
@samnajian I updated the original commit and added the comment

samnajian (approved):
LGTM, thanks!

Aschepikov (approved):
GJ!

PR #4736, opened 2023-02-15
Campaign IDs for emails sent via MSS [MAILPOET-5034]

johnolek:
## Description

This PR generates campaign IDs for emails, stores them as part of the sending queue, and includes them in the `meta` when performing sends. 

Even though the campaignId is currently only relevant for MSS sends, this PR creates and stores one for all sending queues. This is both for simplicity and because it's theoretically possible to switch from  a non-MSS sending method to MSS in the middle of processing a queue.

## Code review notes

Campaign IDs should be different if the content for the same newsletter changes, so I decided to calculate a hash right after preprocessing the newsletter.

The sending queue seemed like a good place to store the campaignId because we have a very specific version of the rendered email at that point. The campaignId can also change for a specific newsletter ID, so storing the campaignId on the newsletter table seemed like a bad idea because we wouldn't be able to easily determine which newsletter corresponds to an old campaignId if it does change. I also wanted to avoid any database migrations if possible.

This will not add campaign IDs for preview emails.

## QA notes

I think the easiest way to test this will be to send a variety of emails and verify that you see unique campaign IDs stored in the `meta` column of the `mailpoet_sending_queues` table. 

For automatic or automation emails, the campaign ID should stay the same from send to send unless the content changes (see the ticket for exact details). This doesn't really apply to post notification emails since the content changes every time.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5034](https://mailpoet.atlassian.net/browse/MAILPOET-5034)

## After-merge notes

_N/A_

[MAILPOET-5034]: https://mailpoet.atlassian.net/browse/MAILPOET-5034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -20,6 +20,7 @@ use MailPoet\Models\StatisticsNewsletters as StatisticsNewslettersModel;
 use MailPoet\Models\Subscriber as SubscriberModel;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
+use MailPoet\Newsletter\Sending\SendingQueuesRepository;
 use MailPoet\Segments\SegmentsRepository;
 use MailPoet\Segments\SubscribersFinder;
 use MailPoet\Subscribers\SubscribersRepository;
@@ -78,6 +79,9 @@ class SendingQueue {
   /** @var SubscribersRepository */
   private $subscribersRepository;
 
+  /** @var SendingQueuesRepository */
+  private $sendingQueueRepository;
+
   public function __construct(
     SendingErrorHandler $errorHandler,
     SendingThrottlingHandler $throttlingHandler,
@@ -92,6 +96,7 @@ class SendingQueue {
     ScheduledTasksRepository $scheduledTasksRepository,
     MailerTask $mailerTask,
     SubscribersRepository $subscribersRepository,
+    SendingQueuesRepository $sendingQueuesRepository,
     $newsletterTask = false
   ) {
     $this->errorHandler = $errorHandler;
@@ -109,6 +114,7 @@ class SendingQueue {
     $this->links = $links;
     $this->scheduledTasksRepository = $scheduledTasksRepository;
     $this->subscribersRepository = $subscribersRepository;
+    $this->sendingQueueRepository = $sendingQueuesRepository;
   }
 
   public function process($timer = false) {
@@ -160,11 +166,32 @@ class SendingQueue {
       return;
     }
 
+    $campaignId = null;
+
+    $afterPreProcessFilter = function(array $renderedNewsletters, NewsletterEntity $renderedNewsletterEntity) use ($newsletterEntity, &$campaignId) {
+      if ($newsletterEntity !== $renderedNewsletterEntity || !isset($renderedNewsletters['text'])) {
+        return;
+      }
+      $textVersion = $renderedNewsletters['text'];
+      $campaignId = $this->calculateCampaignId($newsletterEntity, $textVersion);
+      return $renderedNewsletters;
+    };
+
+    // This filter fires during preProcessNewsletter, after some initial rendering but before any shortcodes are replaced.
+    $this->wp->addFilter('mailpoet_sending_newsletter_render_after_pre_process', $afterPreProcessFilter, 10, 2);
+
     // pre-process newsletter (render, replace shortcodes/links, etc.)
     $newsletterEntity = $this->newsletterTask->preProcessNewsletter($newsletterEntity, $queue);
+
+    $this->wp->removeFilter('mailpoet_sending_newsletter_render_after_pre_process', $afterPreProcessFilter);
+
     if (!$newsletterEntity) {
       $this->deleteTask($queue);
       return;
+    }
+
+    if ($campaignId) {
+      $this->sendingQueueRepository->addCampaignId($queue->getSendingQueueEntity(), $campaignId);
     }
 
     $newsletter = Newsletter::findOne($newsletterEntity->getId());
@@ -387,6 +414,10 @@ class SendingQueue {
       $statistics,
       $timer
     );
+  }
+
+  public function calculateCampaignId(NewsletterEntity $newsletter, string $textBody): string {
+    return substr(md5(implode('|', [$newsletter->getId(), $textBody, $newsletter->getSubject()])), 0, 16);
   }
 
   /**
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -301,6 +301,9 @@ class SendingQueue {
     $statistics = [];
     $metas = [];
     $oneClickUnsubscribeUrls = [];
+    $sendingQueueEntity = $queue->getSendingQueueEntity();
+    $sendingQueueMeta = $sendingQueueEntity->getMeta() ?? [];
+    $campaignId = $sendingQueueMeta['campaignId'] ?? null;
 
     $newsletterEntity = $this->newslettersRepository->findOneById($newsletter->id);
 
@@ -331,7 +334,11 @@ class SendingQueue {
       $unsubscribeUrls[] = $this->links->getUnsubscribeUrl($queue->id, $subscriberEntity);
       $oneClickUnsubscribeUrls[] = $this->links->getOneClickUnsubscribeUrl($queue->id, $subscriberEntity);
 
-      $metas[] = $this->mailerMetaInfo->getNewsletterMetaInfo($newsletter, $subscriberEntity);
+      $metasForSubscriber = $this->mailerMetaInfo->getNewsletterMetaInfo($newsletter, $subscriberEntity);
+      if ($campaignId) {
+        $metasForSubscriber['campaign_id'] = $campaignId;
+      }
+      $metas[] = $metasForSubscriber;
 
       // keep track of values for statistics purposes
       $statistics[] = [
@@ -370,7 +377,11 @@ class SendingQueue {
         $preparedSubscribers,
         $statistics,
         $timer,
-        ['unsubscribe_url' => $unsubscribeUrls, 'meta' => $metas, 'one_click_unsubscribe' => $oneClickUnsubscribeUrls]
+        [
+          'unsubscribe_url' => $unsubscribeUrls,
+          'meta' => $metas,
+          'one_click_unsubscribe' => $oneClickUnsubscribeUrls,
+        ]
       );
     }
     return $queue;
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -20,7 +20,6 @@ use MailPoet\Models\StatisticsNewsletters as StatisticsNewslettersModel;
 use MailPoet\Models\Subscriber as SubscriberModel;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
-use MailPoet\Newsletter\Sending\SendingQueuesRepository;
 use MailPoet\Segments\SegmentsRepository;
 use MailPoet\Segments\SubscribersFinder;
 use MailPoet\Subscribers\SubscribersRepository;
@@ -79,9 +78,6 @@ class SendingQueue {
   /** @var SubscribersRepository */
   private $subscribersRepository;
 
-  /** @var SendingQueuesRepository */
-  private $sendingQueueRepository;
-
   public function __construct(
     SendingErrorHandler $errorHandler,
     SendingThrottlingHandler $throttlingHandler,
@@ -96,7 +92,6 @@ class SendingQueue {
     ScheduledTasksRepository $scheduledTasksRepository,
     MailerTask $mailerTask,
     SubscribersRepository $subscribersRepository,
-    SendingQueuesRepository $sendingQueuesRepository,
     $newsletterTask = false
   ) {
     $this->errorHandler = $errorHandler;
@@ -114,7 +109,6 @@ class SendingQueue {
     $this->links = $links;
     $this->scheduledTasksRepository = $scheduledTasksRepository;
     $this->subscribersRepository = $subscribersRepository;
-    $this->sendingQueueRepository = $sendingQueuesRepository;
   }
 
   public function process($timer = false) {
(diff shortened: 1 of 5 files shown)


mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -160,13 +166,34 @@ private function processSending(SendingTask $queue, int $timer): void {
...
+      return $renderedNewsletters;
+    };
+
+    // This filter fires during preProcessNewsletter, after some initial rendering but before any shortcodes are replaced.
+    $this->wp->addFilter('mailpoet_sending_newsletter_render_after_pre_process', $afterPreProcessFilter, 10, 2);
+
     // pre-process newsletter (render, replace shortcodes/links, etc.)
     $newsletterEntity = $this->newsletterTask->preProcessNewsletter($newsletterEntity, $queue);
+
+    $this->wp->removeFilter('mailpoet_sending_newsletter_render_after_pre_process', $afterPreProcessFilter);
```
johnolek: It feels like there is probably a better way to handle this, but I haven't been able to come up with one. 

I originally thought I could have the calculateCampaignId method do some rendering itself, but if you have the renderer render the same newsletter twice it may not produce the same results, e.g. it will skip ALC posts that were previously rendered.

Ultimately the idea of using this filter appealed to me because I think the newsletter is in just the right state. Shortcodes haven't been replaced yet, which I think is good for what we're trying to achieve.

I could have done some refactoring, but lots of this code is going to need refactoring when we convert it to doctrine, and I'm honestly nervous about changing too much in any of this code.

That being said, I'm very much open to better ideas!

johnolek:
@JanJakes I would love to get your opinion on the approach I've taken, and I would be especially happy if you have any ideas that would achieve the same result more gracefully.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
JanJakes: This looks good to me, the only "why don't you just" question I'll try will be this one:

Why not implementing this within the `preProcessNewsletter` function that you're wrapping here via shortcodes? In other words — can we put the code to the place from which this hook is invoked? Or would that be a worse spot to place this?

JanJakes (commented):
LGTM!

> Campaign IDs should be different if the content for the same newsletter changes, whether it's the body content or the subject line, so I decided to calculate a hash right after we render the newsletter that is based on newsletter ID, the text version of the rendered body (because it doesn't seem all that relevant if only some styles change), and the subject of the newsletter.

This sounds like a great solution to me!

> It wouldn't be difficult to change the application code to only generate IDs for MSS sends, but it was proving a bit annoying to update the tests and I wanted to validate the approach before putting in too much time.

Is there any harm in always generating campaign IDs, independently of the sending method? For simplicity, I think we could do that.

JanJakes:
@johnolek I'll add one more question. Just would like to know your thoughts about content-base campaign IDs (as implemented) vs. random ones. I like your solution, just want to know the context. Do you think generating random campaign IDs would be harder because we'd need to catch all the possible cases that would cause its invalidation?

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
johnolek: Such a simple question and yet something that I didn't properly consider!

Mainly I was thinking that it only makes sense to compute the value when we're sure a send is about to occur, and I (incorrectly, I believe) assumed that `preProcessNewsletter` might be used in other situations.

I think [this](https://github.com/mailpoet/mailpoet/blob/MAILPOET-5034%2Fgenerate_campaign_ids_for_new_email_campaigns/mailpoet/lib/Newsletter/NewsletterSaveController.php#L415-L415) is the only other usage, and long story short I think by not including this campaignId code directly in `preProcessNewsletter` we could end up sending the wrong campaignId for newsletters that get scheduled and then the content gets updated.

Thanks so much for pointing this out! I'll move the code over there and triple check that it won't cause any issues.

johnolek:
@JanJakes Thanks very much for taking a look!

>Is there any harm in always generating campaign IDs, independently of the sending method? For simplicity, I think we could do that.

The only harm would be storing extra unnecessary data in the database, but honestly it's not much compared to the `newsletter_rendered_body` we're already storing in the same row. I think a lot of sends don't store anything in the `meta` right now and for some reason it felt compelling to maintain that behavior.

That being said, I think there's a bigger risk in _not_ always generating a campaign ID. I could be wrong about this, but someone could switch sending methods after a send has started, right? Like they send an email to 10,000 subscribers using their host at 20 emails/5 minutes, then decide to switch to MSS halfway through. In that case we would be missing the campaign ID after the switch, unless we added even more code to detect that case and correct it, which seems silly.

> Just would like to know your thoughts about content-base campaign IDs (as implemented) vs. random ones. I like your solution, just want to know the context. Do you think generating random campaign IDs would be harder because we'd need to catch all the possible cases that would cause its invalidation?

Yes, pretty much exactly that.

Sorry in advance for the number of thoughts I'm about to share -- this got out of hand. I don't mean to bore you with the details, but if I'm making incorrect assumptions or misunderstanding anything it would be very helpful to know. And maybe this will spark some ideas that wouldn't have otherwise occurred to you.

Also, you asked!

Please don't feel a need to respond to any specific parts of these ramblings.

## Thoughts

Invalidation is only relevant for automatic emails because newsletter ID is taken into account in the calculation (I confirmed that we want a different `campaignId` if you duplicate a standard newsletter, make zero changes, and send it again), so the best example I can think of is any kind of automatic/automation email that includes an ALC block. Assuming no other changes, if it's sent to multiple people with post id `1`, the campaign ID should be the same for each email, regardless of when it's sent.

At what point would we catch it if the post query returns post id `2` instead of `1`? If `campaignId` was a property of the newsletter, we wouldn't be able to simply rely on saving/updating hooks and update it on save, because no changes would be necessary for the rendered newsletter content to be different.

There's also the matter of returning to a previous state. If the ALC block is updated to once again render post `1` instead of `2`, then the `campaignId` should also revert, assuming the post content itself hasn't changed. Maybe more of a nice to have than a requirement, but if we did want to do this using random IDs we'd need to somehow look up the correct previous `campaignId` if one existed, which could mean having to store extra data for lookups. The digest approach avoids all of that.

We could track the post IDs that would be rendered for any email with an ALC block, and update the `campaignId` any time they change, but that wouldn't be enough because the title/content of the included post could change between sends.

Long story short, the `campaignId` should change when the content changes, and there are lots of ways the content can change, including lots of filters that could be set to run conditionally.

We do store the `newsletter_rendered_body` in the database, and we could compare that value with the one from the previous sending queue for the same newsletter ID if one exists. The main downside I see to this approach is links. If click tracking is enabled, the `newsletter_rendered_body` has values like `[mailpoet_click_data]-849e4ce37319` instead of actual URLs, and the values will change from send to send for the same automat(ed|ic) email even if the URL stays the same, which makes sense.

This is why I think we need to rely on the content of the email at the point of the `mailpoet_sending_newsletter_render_after_pre_process` filter. Of course it would be wasteful to store all that content in addition to the `text` version we're already saving in the `newsletter_rendered_body`, which leads me to the idea of calculating a hash, and at that point we may as well just use that hash as the `campaignId` instead of using it to determine if we should calculate a new random one.

Another small benefit I see to this approach is there's no need for any migrations.

I also like the fact that this will detect changes even if a user changes the content directly in the database, which might not be possible/easy using other approaches.

## Limitations

I think the code in this PR would work fine for 96+% of cases, but here are some limitations I've been thinking about:

### Any changes not reflected in the text version
Using the text version of the email should catch most changes we care about, but not all.

#### Images
The text version includes the alt text for an image. Of course multiple images can have the same alt text, and most users will be viewing the HTML version of the email, where a different image with the same alt text could make a world of difference.

I could update the calculation to parse the html version, extract any `src` attributes for `img` tags, and include them in the MD5 calculation.

I suppose the image file itself could be replaced with another with the same name. Clearly the only solution here* is to locate all the image files on the server and calculate their MD5s or file sizes and include those in the `campaignId` calculation as well.

*I don't actually think this is worth doing.

#### Major style changes
Changing the font size from 16px to 15px doesn't seem like a good reason to regenerate a campaignId, but changing all the text to be comic sans and using outrageous colors does. It would be near impossible to come up with reasonable rules for "allowable" changes though, not to mention a huge pain to implement and probably not worthwhile.

It's also possible that things like column layouts could change while not affecting the text version of the email.

My initial thought was that we could err on the conservative side and simply include the newsletter's `updatedAt` in the `campaignId` calculation, which would work for image changes too, but that would also capture simple status changes, like disabling/re-enabling a welcome email, so I don't have any good ideas for these right now.

### Certain shortcodes
The current solution doesn't render any shortcodes. I think this makes sense for almost all the default shortcodes (I don't think we care if the total number of subscribers, `[newsletter:total]`, changes, or any of the date shortcodes, and we can't even render any subscriber-specific shortcodes). The only default shortcodes that make sense to me to consider would be the site title/URL. The `campaignId` should be different if an email is sent before and after a site migration, and the site title is relevant content.

Users can define whatever custom shortcodes they want, and they could render very different content from send to send of the same email.

I doubt custom shortcodes are very common in the first place, and the site shortcodes are very unlikely to change, so overall I'm not too concerned about this, but if it becomes an issue we could consider rendering a specific subset of shortcodes before calculating the campaignId.

We could also include the site URL/title in the MD5 calculation directly instead of rendering shortcodes. This would just leave custom shortcodes to worry about.

### Other content changes from filters - will be resolved!
The current code hooks into the `mailpoet_sending_newsletter_render_after_pre_process` filter with a priority of `10`, meaning it could be missing content changes that come from later hooks. Another benefit to moving the calculation directly into `preProcessNewsletter` means we can perform the `campaignId` calculation on the final version returned from the filter without having to worry about priorities.

Thanks again for your suggestion to move that code!

###  Super edge cases
These probably aren't worth thinking about at all, and yet I wrote this whole section about them.

#### MD5 collisions
It's always possible we'll get the same MD5 value for different content. The fact that we're taking the first 16 hex characters makes this _slightly_ more possible than using the full MD5, but collisions still seems extraordinarily unlikely unless there's something I don't understand about this hashing algorithm. This would only be an issue for newsletters with different IDs since we're expecting to regularly get the same hash for the same email if nothing relevant changes.

We could mitigate this by making campaign IDs include more than just a hash. For example, we could also include the newsletter ID directly in the campaignId, like `{newsletterId}|{md5}`. If we definitely need to avoid non-alphanumeric characters in the `campaignId` we could arbitrarily start the hash portion with some specific letter. Then instead of 16 characters of MD5, we would add as many characters as needed to get up to 16. Since IDs are unsigned `int`s in the DB with a maximum value of `4294967295`, so 10 digits max, that would still give us 5 digits of MD5 on the end, which would honestly probably be plenty.

##### Well actually...

Could it be useful to include a newsletter ID in the campaign ID? I'm thinking about how sometimes when investigating sending issues it's hard to tell which specific newsletter generated bounces/complaints. We can see the `emailType` and `subject`, and when it was sent, but it would be really nice if we also had the newsletter ID for when support reaches out to users to help fix issues. Of course it would be better to just include this as a separate meta field instead of combining it with the campaignId.

#### Full reinstalls that happen to exactly recreate old emails
A site that has used MailPoet and sent emails with the MSS, then did a full reinstall of MailPoet (resetting the ID auto increment), then recreate and send an email that just so happens to be exactly the same content/subject as the pre-reinstall newsletter with the same ID. Right now that email would incorrectly get the same `campaignId` as the original. Maybe I should include `createdAt` in the `campaignId` calculation, or the newsletter's own `hash` value...

## Final Thoughts
I think I've been overthinking this.

JanJakes:
> That being said, I think there's a bigger risk in not always generating a campaign ID.

Agreed. I think I'd simply include them in all cases.

> Invalidation

I agree with your solution, thanks for all the info! I think I meant more like if we had something like versioning in automations (where we create a new version on every save, if something has changed), then with each new version we could generate a new fully random string. But since we have no such thing for emails, it's not really meaningful to even consider that, and it wouldn't even revert to a previous campaign ID when a change was reverted.

> Images

I see you've already solved those! Well, then we have the probably most common edge case covered!

> Major style changes

Probably not worth trying to solve now. Using `updatedAt` in the hash would break the case of making a change and reverting it. If we don't need to cover that case, then using `updatedAt` probably means we don't need to hash the content at all — ID & `updatedAt` would suffice in such case.

> Certain shortcodes

I think your solution is great, especially for treating shortcodes reasonably, and I wouldn't worry about the edge cases.

> MD5 collisions

I wouldn't worry about them. As for including the newsletter ID, yeah, it's probably better to leave it to another meta field.

You can also consider [other hashing algorithms](https://www.php.net/manual/en/function.hash.php), but most of them are probably an overkill. We could also convert the HEX hash to some variant of base32 (in case we want URL-safe and case-insensitive strings), but it seems to be an overkill as well, given that it's probably not a single function call in PHP.

> Full reinstalls that happen to exactly recreate old emails

If the email and the site are the same, it seems reasonable to me, that the campaign ID remains the same.

That said, do we prefer different campaign IDs on different sites? What I mean is — should we include some sort of site identifier in the hash? Seems like it might be better to have different campaign IDs on two different sites that both do an empty install and send the same template without any changes. What's the MSS' point of view here? Does it expect every campaign ID from every customer to be unique? I imagine that would be way easier to use and filter by then.

johnolek:
> That said, do we prefer different campaign IDs on different sites?

This wasn't covered in the specs, but in a [comment](https://mailpoet.atlassian.net/browse/MAILPOET-5033?focusedCommentId=23598) on another ticket it was confirmed that it doesn't matter if two different sites use the same ID. There are more details in [this google doc](https://docs.google.com/document/d/1iry0Pyq2OJ2jdBAdIf8EPNBx9XuM2fFNpLctyo0XIUw/edit?usp=sharing). Basically the MSS side of things will handle generating a globally unique ID that includes the campaign ID.

As for the hashing algorithm, I wasn't able to identify a compelling reason to use something other than MD5. If you have any suggestions though, I'm all ears!

Other than those things, anything you'd like to see changed/updated in this PR @JanJakes?

JanJakes (approved):
Great PR!

> Other than those things, anything you'd like to see changed/updated in this PR @JanJakes?

No, looks great to me. Given that we don't need to have the IDs globally unique, I think there's nothing to add. Or do you want to add a site identifier to the hash anyway? Approving and leaving it up to you, probably nothing important now. Thanks!

johnolek:
Thanks for the review, @JanJakes! I don't think the campaignId is an appropriate place for a site/newsletter ID. If those things were meaningful to us we could always include them separately in the meta. I'm happy to leave this as is for now.

@oz This PR includes the sending of `campaign_id`s in the [meta](https://github.com/mailpoet/mailpoet/pull/4736/files#diff-3a26b2d8faf9cc01efd5aef47b058c088c0de01b8074c3be7cefd9adb77fbaaaR313). Do you have any concerns about us merging this in before the MSS has been updated to handle `campaign_id`s? I wanted to double check because we didn't include that in the specs for the ticket.

JanJakes (approved):

oz:
> @oz This PR includes the sending of `campaign_id`s in the [meta](https://github.com/mailpoet/mailpoet/pull/4736/files#diff-3a26b2d8faf9cc01efd5aef47b058c088c0de01b8074c3be7cefd9adb77fbaaaR313). Do you have any concerns about us merging this in before the MSS has been updated to handle `campaign_id`s? I wanted to double check because we didn't include that in the specs for the ticket.

Thanks for checking.

We haven't started work on the bridge API to [support this new parameter](https://github.com/mailpoet/services-bridge/issues/106) yet. Meanwhile having an extraneous "meta" parameter will not break posting: the API will ignore it.

Also cc @mhadidg 😎

Aschepikov (approved):
GJ!

PR #4737, opened 2023-02-16
Add setting to run a workflow only once. [MAILPOET-4966]

websupporter:
## Description

Solves [MAILPOET-4966]

## Code review notes

As we discussed, for now we show the setting only when a certain circumstances are given: No trigger, or trigger with subscriber subject ([3039181](https://github.com/mailpoet/mailpoet/pull/4737/commits/3039181454210a5245003af35ae0e3a1786c0e62)). I therefore extended the `Step` definition in React: [7fb717f](https://github.com/mailpoet/mailpoet/pull/4737/commits/7fb717f89495b166a88ba0d3e534ae4ad222ad01), which made it necessary to update also the Add Step process in premium, to transfer this information to the automation data.

I moved the subjects of a run to its own data table and I added a `hash`, which is indexed in order to query by hash. The hash is a `md5` hash, which limits the length of the string. I chose to do so as this enables us to use `longtext` for the subject arguments.

As I test whether `Automation::getMeta()` et.al work as expected, I added also some more tests for the `Automation` class.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/683

## Linked tickets

[MAILPOET-4966]

[MAILPOET-4966]: https://mailpoet.atlassian.net/browse/MAILPOET-4966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4966]: https://mailpoet.atlassian.net/browse/MAILPOET-4966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/components/automation/types.ts
+++ b/mailpoet/assets/js/src/automation/editor/components/automation/types.ts
@@ -32,4 +32,5 @@ export type Automation = {
     };
   };
   steps: Record<string, Step> & { root: Step };
+  meta: Record<string, unknown>;
 };
--- a/mailpoet/assets/js/src/automation/editor/components/automation/types.ts
+++ b/mailpoet/assets/js/src/automation/editor/components/automation/types.ts
@@ -7,7 +7,6 @@ export type NextStep = {
 export type Step = {
   id: string;
   type: 'root' | 'trigger' | 'action';
-  subject_keys: string[];
   key: string;
   args: Record<string, unknown>;
   next_steps: NextStep[];
--- a/mailpoet/assets/js/src/automation/editor/components/automation/types.ts
+++ b/mailpoet/assets/js/src/automation/editor/components/automation/types.ts
@@ -7,6 +7,7 @@ export type NextStep = {
 export type Step = {
   id: string;
   type: 'root' | 'trigger' | 'action';
+  subject_keys: string[];
   key: string;
   args: Record<string, unknown>;
   next_steps: NextStep[];
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateAutomationController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateAutomationController.php
@@ -66,7 +66,7 @@ class UpdateAutomationController {
       }
     }
 
-    $automation->deleteMetas();
+    $automation->deleteAllMetas();
     if (array_key_exists('meta', $data)) {
       foreach ($data['meta'] as $key => $value) {
         $automation->setMeta($key, $value);
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateAutomationController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateAutomationController.php
@@ -66,8 +66,8 @@ class UpdateAutomationController {
       }
     }
 
-    $automation->deleteAllMetas();
     if (array_key_exists('meta', $data)) {
+      $automation->deleteAllMetas();
       foreach ($data['meta'] as $key => $value) {
         $automation->setMeta($key, $value);
       }
--- a/mailpoet/lib/Automation/Engine/Builder/UpdateAutomationController.php
+++ b/mailpoet/lib/Automation/Engine/Builder/UpdateAutomationController.php
@@ -66,6 +66,13 @@ class UpdateAutomationController {
       }
     }
 
+    $automation->deleteMetas();
+    if (array_key_exists('meta', $data)) {
+      foreach ($data['meta'] as $key => $value) {
+        $automation->setMeta($key, $value);
+      }
+    }
+
     $this->hooks->doAutomationBeforeSave($automation);
 
     $this->automationValidator->validate($automation);
--- a/mailpoet/lib/Automation/Engine/Data/Automation.php
+++ b/mailpoet/lib/Automation/Engine/Data/Automation.php
@@ -190,7 +190,7 @@ class Automation {
     return $this->meta[$key] ?? null;
   }
 
-  public function getMetas(): array {
+  public function getAllMetas(): array {
     return $this->meta;
   }
 
@@ -209,7 +209,7 @@ class Automation {
     $this->setUpdatedAt();
   }
 
-  public function deleteMetas(): void {
+  public function deleteAllMetas(): void {
     $this->meta = [];
     $this->setUpdatedAt();
   }
--- a/mailpoet/lib/Automation/Engine/Data/Automation.php
+++ b/mailpoet/lib/Automation/Engine/Data/Automation.php
@@ -45,7 +45,7 @@ class Automation {
   /** @var array<string, Step> */
   private $steps;
 
-  /** @var array<string, Mixed> */
+  /** @var array<string, mixed> */
   private $meta = [];
 
   /** @param array<string, Step> $steps */
--- a/mailpoet/lib/Automation/Engine/Data/Automation.php
+++ b/mailpoet/lib/Automation/Engine/Data/Automation.php
@@ -45,6 +45,9 @@ class Automation {
   /** @var array<string, Step> */
   private $steps;
 
+  /** @var array<string, Mixed> */
+  private $meta = [];
+
   /** @param array<string, Step> $steps */
   public function __construct(
     string $name,
@@ -171,11 +174,44 @@ class Automation {
           return $step->toArray();
         }, $this->steps)
       ),
+      'meta' => Json::encode($this->meta),
     ];
   }
 
   private function setUpdatedAt(): void {
     $this->updatedAt = new DateTimeImmutable();
+  }
+
+  /**
+   * @param string $key
+   * @return mixed|null
+   */
+  public function getMeta(string $key) {
+    return $this->meta[$key] ?? null;
+  }
+
+  public function getMetas(): array {
+    return $this->meta;
+  }
+
+  /**
+   * @param string $key
+   * @param mixed $value
+   * @return void
+   */
+  public function setMeta(string $key, $value): void {
+    $this->meta[$key] = $value;
+    $this->setUpdatedAt();
+  }
+
+  public function deleteMeta(string $key): void {
+    unset($this->meta[$key]);
+    $this->setUpdatedAt();
+  }
+
+  public function deleteMetas(): void {
+    $this->meta = [];
+    $this->setUpdatedAt();
   }
 
   public static function fromArray(array $data): self {
@@ -193,6 +229,8 @@ class Automation {
     $automation->createdAt = new DateTimeImmutable($data['created_at']);
     $automation->updatedAt = new DateTimeImmutable($data['updated_at']);
     $automation->activatedAt = $data['activated_at'] !== null ? new DateTimeImmutable($data['activated_at']) : null;
+
+    $automation->meta = $data['meta'] ? Json::decode($data['meta']) : [];
     return $automation;
   }
 }
--- a/mailpoet/lib/Automation/Engine/Data/Subject.php
+++ b/mailpoet/lib/Automation/Engine/Data/Subject.php
@@ -27,10 +27,15 @@ class Subject {
     return $this->args;
   }
 
+  public function hash(): string {
+    return md5($this->getKey() . Json::encode($this->getArgs()));
+  }
+
   public function toArray(): array {
     return [
       'key' => $this->getKey(),
       'args' => Json::encode($this->getArgs()),
+      'hash' => $this->hash(),
     ];
   }
 
--- a/mailpoet/lib/Automation/Engine/Data/Subject.php
+++ b/mailpoet/lib/Automation/Engine/Data/Subject.php
@@ -2,6 +2,8 @@
 
 namespace MailPoet\Automation\Engine\Data;
 
+use MailPoet\Automation\Engine\Utils\Json;
+
 class Subject {
   /** @var string */
   private $key;
@@ -28,11 +30,11 @@ class Subject {
   public function toArray(): array {
     return [
       'key' => $this->getKey(),
-      'args' => $this->getArgs(),
+      'args' => Json::encode($this->getArgs()),
     ];
   }
 
   public static function fromArray(array $data): self {
-    return new self($data['key'], $data['args']);
+    return new self($data['key'], Json::decode($data['args']));
   }
 }
--- a/mailpoet/lib/Automation/Engine/Data/Subject.php
+++ b/mailpoet/lib/Automation/Engine/Data/Subject.php
@@ -27,7 +27,7 @@ class Subject {
     return $this->args;
   }
 
-  public function hash(): string {
+  public function getHash(): string {
     return md5($this->getKey() . serialize($this->getArgs()));
   }
 
@@ -35,7 +35,7 @@ class Subject {
     return [
       'key' => $this->getKey(),
       'args' => Json::encode($this->getArgs()),
-      'hash' => $this->hash(),
+      'hash' => $this->getHash(),
     ];
   }
 
--- a/mailpoet/lib/Automation/Engine/Data/Subject.php
+++ b/mailpoet/lib/Automation/Engine/Data/Subject.php
@@ -28,7 +28,7 @@ class Subject {
   }
 
   public function hash(): string {
-    return md5($this->getKey() . Json::encode($this->getArgs()));
+    return md5($this->getKey() . serialize($this->getArgs()));
   }
 
   public function toArray(): array {
--- a/mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
+++ b/mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
@@ -121,7 +121,7 @@ class AutomationRunStorage {
     $table = esc_sql($this->table);
     $subjectTable = esc_sql($this->subjectTable);
 
-    $sql = "SELECT count(runs.id) as count from $table as runs
+    $sql = "SELECT count(DISTINCT runs.id) as count from $table as runs
       JOIN $subjectTable as subjects on runs.id = subjects.automation_run_id
       WHERE runs.automation_id = %d
       AND subjects.hash = %s";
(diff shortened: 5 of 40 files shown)


mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
```diff
@@ -11,28 +11,60 @@ class AutomationRunStorage {
...
+  /** @var string */
+  private $subjectTable;
+
   /** @var wpdb */
   private $wpdb;
 
   public function __construct() {
     global $wpdb;
     $this->table = $wpdb->prefix . 'mailpoet_automation_runs';
+    $this->subjectTable = $wpdb->prefix . 'mailpoet_automation_run_subjects';
```
JanJakes, line 24: Would it make sense to create an `AutomationRunSubjectStorage` and just call it from this class for better code separation or... meh?

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
```diff
@@ -11,28 +11,60 @@ class AutomationRunStorage {
...
   public function createAutomationRun(AutomationRun $automationRun): int {
-    $result = $this->wpdb->insert($this->table, $automationRun->toArray());
+    $automationTableData = $automationRun->toArray();
+    unset($automationTableData['subjects']);
+    $result = $this->wpdb->insert($this->table, $automationTableData);
+    if ($result === false) {
+      throw Exceptions::databaseError($this->wpdb->last_error);
+    }
+    $automationRunId = $this->wpdb->insert_id;
+    $subjectTableData = $automationRun->toArray()['subjects'];
```
JanJakes: If you move this above the `unset` higher up, you won't need to call `toArray()` twice and it should save a bit of CPU.

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
```diff
@@ -41,14 +73,39 @@ public function getAutomationRun(int $id): ?AutomationRun {
...
+    if (!is_array($automationRuns) || !$automationRuns) {
+      return [];
+    }
+
+    $automationRunIds = array_map(
+      function(array $runData): int {
+        return (int)$runData['id'];
+      },
+      $automationRuns
+    );
```
JanJakes, line 97: I guess you also could use `array_column(...)`. It won't give you the cast to `int` but given how it's later used, maybe it's fine.

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
```diff
@@ -41,14 +73,39 @@ public function getAutomationRun(int $id): ?AutomationRun {
...
+
+    $sql = sprintf("SELECT * FROM $subjectTable WHERE automation_run_id in (%s) order by automation_run_id", implode(',', array_map(function() {return '%d';
+
+    }, $automationRunIds)));
```
JanJakes: This looks like sort of temporary formatting 😄

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
```diff
@@ -41,14 +73,39 @@ public function getAutomationRun(int $id): ?AutomationRun {
...
+
+    $query = (string)$this->wpdb->prepare($sql, ...$automationRunIds);
+    $subjects = $this->wpdb->get_results($query, ARRAY_A);
+
+    return array_map(
+      function(array $runData) use ($subjects): AutomationRun {
+        $runData['subjects'] = array_filter(
```
JanJakes: Be careful with `array_filter` as it makes gaps in the array indexes and thus a numeric array becomes associative. The safe path is to always wrap them with `array_values`. Or maybe we could simplify the search logic by preparing a map first...

mailpoet/lib/Migrations/Migration_20230215_050813.php
```diff
@@ -0,0 +1,73 @@
...
+    $tableName = esc_sql($wpdb->prefix . 'mailpoet_automation_runs');
+    if (!$this->columnExists($tableName, 'subjects')) {
+      return;
+    }
+
+    $sql = "SELECT id,subjects FROM $tableName where subjects is not null";
+    $results = $wpdb->get_results($sql, ARRAY_A);
+    if (is_array($results) && $results) {
+      return;
+    }
```
JanJakes: I am not sure if these checks are needed, but this would mean that something failed to migrate, but `return` here will make the migration successful. In such case, it should end with an error, I think.

mailpoet/lib/Migrations/Migration_20230215_050813.php
```diff
@@ -0,0 +1,73 @@
...
+    $this->subjectsMigration();
+  }
+
+  private function subjectsMigration(): void {
+    $this->createTable('automation_run_subjects', [
+      '`automation_run_id` int(11) unsigned NOT NULL',
+      '`key` varchar(191)',
+      '`args` longtext',
+      'index (automation_run_id)',
+    ]);
```
JanJakes, line 34: I just noticed a potential problem 😬 We used to have the subjects saved as an array, thus ordered. Here, we'll lose that ability. Can that be a problem anywhere? I guess unless we allow multiple subjects of the same type, we're fine. So maybe we don't have to solve it now at all. The question is, whether we should also make `UNIQUE (automation_run_id, key)` for now...

mailpoet/lib/Migrations/Migration_20230215_050813.php
```diff
@@ -0,0 +1,73 @@
...
+
+use MailPoet\Migrator\Migration;
+
+class Migration_20230215_050813 extends Migration {
+  public function run(): void {
+    $this->subjectsMigration();
+  }
+
+  private function subjectsMigration(): void {
+    $this->createTable('automation_run_subjects', [
```
JanJakes, line 25: There seems to be no primary key in this table. Do we already have any such tables? I'm just thinking, it might be safer to add an autogenerated ID column...

mailpoet/lib/Automation/Engine/Data/Automation.php
```diff
@@ -45,6 +45,9 @@ class Automation {
   /** @var array<string, Step> */
   private $steps;
 
+  /** @var array<string, Mixed> */
```
JanJakes: Is `Mixed` a type? Should be `mixed`, I guess. But PHPStan doesn't seem to complain 🤔

mailpoet/lib/Automation/Engine/Data/Automation.php
```diff
@@ -171,13 +174,46 @@ public function toArray(): array {
...
 
+  /**
+   * @param string $key
+   * @return mixed|null
+   */
+  public function getMeta(string $key) {
+    return $this->meta[$key] ?? null;
+  }
+
+  public function getMetas(): array {
```
JanJakes: The naming of these methods reveals that `meta` is not a great name 😄
But if you want to keep it for consistency with post meta, then maybe we could rename the plural methods to `getAllMeta`, `deleteAllMeta`, etc.

mailpoet/tests/integration/Automation/Engine/Data/AutomationTest.php
```diff
@@ -0,0 +1,54 @@
...
+    $this->assertEmpty($automation->getMetas());
+  }
+
+  public function _after() {
+    $this->storage->truncate();
+  }
+}
+{
+
+}
```
JanJakes: Something happened here 😄 `{ } } { }`

mailpoet/lib/Automation/Engine/Builder/UpdateAutomationController.php
```diff
@@ -66,6 +66,13 @@ public function updateAutomation(int $id, array $data): Automation {
       }
     }
 
+    $automation->deleteMetas();
```
JanJakes: Doesn't this belong inside that IF below? Seems like otherwise it would have the side effects of deleting the meta when anything else is changed.

mailpoet/lib/Automation/Engine/Data/Subject.php
```diff
@@ -27,10 +27,15 @@ public function getArgs(): array {
     return $this->args;
   }
 
+  public function hash(): string {
+    return md5($this->getKey() . Json::encode($this->getArgs()));
```
JanJakes: I think for the use case of hashing, it's safer to use `serialize` over JSON. The reason is that JSON can (in theory) contain arbitrary whitespace, have arbitrary order on object keys, etc.

mailpoet/assets/js/src/automation/editor/components/automation/types.ts
```diff
@@ -7,6 +7,7 @@ export type NextStep = {
 export type Step = {
   id: string;
   type: 'root' | 'trigger' | 'action';
+  subject_keys: string[];
```
JanJakes: Why add it here? It seems to me that this data doesn't belong in the step instance, only to the [registry as you did below](https://github.com/mailpoet/mailpoet/pull/4737/commits/7fb717f89495b166a88ba0d3e534ae4ad222ad01#diff-cb1312bbc63f2b97cea4601c15449850fb79af2ea6b93a30137a6734cdd7a439R15). Could we use it only from there?

mailpoet/lib/Automation/Integrations/MailPoet/Hooks/CreateAutomationRunHook.php
```diff
@@ -0,0 +1,49 @@
...
+    $this->wp->addAction(Hooks::AUTOMATION_RUN_CREATE, [$this, 'createAutomationRun'], 5, 2);
+  }
+
+  public function createAutomationRun(bool $result, StepRunArgs $args): bool {
+    if (!$result) {
+      return $result;
+    }
+
+    $automation = $args->getAutomation();
+    $runOnlyOnce = $automation->getMeta('run_automation_once');
```
JanJakes: Could we maybe reflect the MP-specific behavior in the meta name? What I mean is we could name it something like this: `mailpoet:run-once-per-subscriber`

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
```diff
@@ -108,6 +109,26 @@ function(array $subjectData) use ($runData): bool {
     );
   }
 
+  /**
+   * @param Automation $automation
+   * @return int
+   */
+  public function countRunsForAutomationAndSubject(Automation $automation, Subject $subject): int {
```
JanJakes: I guess more consistent naming with what we use in repositories is `getCountByAutomationAndSubject` or `getCountForAutomationAndSubject`.

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
```diff
@@ -108,6 +109,26 @@ function(array $subjectData) use ($runData): bool {
...
+    $table = esc_sql($this->table);
+    $subjectTable = esc_sql($this->subjectTable);
+
+    $sql = "SELECT count(runs.id) as count from $table as runs
```
JanJakes: Might be safer to use `COUNT (DISTINCT ...)` in case the JOIN ever multiplies rows.

mailpoet/lib/Automation/Engine/Data/Subject.php
```diff
@@ -25,14 +27,19 @@ public function getArgs(): array {
     return $this->args;
   }
 
+  public function hash(): string {
```
JanJakes: Probably just nitpicking, but `getHash` seems more consistent with our naming.

JanJakes (commented):
Insane amount of work! Just wow, @websupporter! I threw in some ideas and observations, but overall looks good!

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
websupporter: 91b30c519

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
websupporter: 0c8416fae

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
websupporter: 75ac7a9b2

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
websupporter: 7188de08f

mailpoet/lib/Automation/Engine/Data/Automation.php
websupporter: 84b99e4de

mailpoet/lib/Automation/Engine/Data/Automation.php
websupporter: 0e115020f

mailpoet/tests/integration/Automation/Engine/Data/AutomationTest.php
websupporter: 616179edc

mailpoet/lib/Automation/Engine/Data/Subject.php
websupporter: ed7f38ade

mailpoet/lib/Automation/Integrations/MailPoet/Hooks/CreateAutomationRunHook.php
websupporter: 8e95388

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
websupporter: 979c51c5e

mailpoet/lib/Automation/Engine/Data/Subject.php
websupporter: 16b9180b2

mailpoet/lib/Migrations/Migration_20230215_050813.php
websupporter: The way I currently see it: I would strongly suggest against the need to have subjects ordered. Where I see multiple subjects of the same type: E.g. send an email to multiple addresses. But the order of the SubscriberSubject does not matter in this case.

Where I think multiple subjects of the same type would be misused: Use one of those `EmailSubjects` as `From` and one as `To`. I think using `from => $emailSubjects[0], to => $emailSubjects[1]` is just prone to produce bugs. I would prefer `class FromSubject` and `class ToSubject` or `$emailSubject->role() === 'sender'` or something along those lines than :crossed_fingers: `[0]` is really `From` and every developer is aware of the dangers of mixing it up there.

Which then makes me even more: For now, we do not need `UNIQUE` as we do not think of those arrays as ordered lists.

If we should come to another conclusion later on, we have now 75dd778ec which actually would bring some kind of order back :thinking:

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
websupporter: Not sure. Mostly because of the `JOIN` we need to calculate the number of runs.

Atm I do not see the need tbh.

mailpoet/lib/Migrations/Migration_20230215_050813.php
websupporter: You are right. `dropSubjectColumn()` would only be executed once `moveSubjectData()` is done. So it can't happen like `moveSubjectData()` is not finished but `dropSubjectColumn()` gets called. Fixed in 61c856ea2

mailpoet/lib/Migrations/Migration_20230215_050813.php
websupporter: 75dd778ec

mailpoet/lib/Automation/Engine/Builder/UpdateAutomationController.php
websupporter: 6d6cb32f3

mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php
websupporter: b2bcb234b

mailpoet/assets/js/src/automation/editor/components/automation/types.ts
websupporter: Thank you for pointing that out! 1f3e94b3d

websupporter:
Thank you @JanJakes for your thorough review. Much appreciated. I addressed your comments (see above). :heart:

mailpoet/lib/Migrations/Migration_20230215_050813.php
JanJakes: @websupporter Thanks for the explanation, that makes a lot of sense. We can keep it "unordered" for now, at least for the purposes of "use the first subject for this and the second for that".

But maybe for the purposes of internal API consistency, I'd add `ORDER BY id` on the query side? What I mean is that calling `getAutomationRunsForAutomation` multiple times would return the same stable results. So, adding the `id` to the query:

```sql
SELECT * FROM $subjectTable WHERE automation_run_id in (%s) order by automation_run_id, id -- added "id"
```

We won't run into multiple subjects of the same type for a long time, I guess, but it would be nice for the current ones to get consistent order in the same automation, e.g., 1. `subscriber`, 2. `segment` on every call of that method.

JanJakes (commented):
LGTM! Thanks! I added only one small comment about the subject ordering.

mailpoet/lib/Migrations/Migration_20230215_050813.php
websupporter: Thanks @JanJakes! Added suggestion in 5d476714c

JanJakes (approved):
Thanks! 👏

Aschepikov (approved):
GJ!

websupporter:
comment to see if jira picks up on the status change.

PR #4740, opened 2023-02-16
Merge checkout opt-ins from AutomateWoo and MailPoet [MAILPOET-5057]

alex-mailpoet:
## Description

The solution builds on the fact that WooCommerce actions used by MailPoet are run before actions used by AutomateWoo, so MP can reliably control AW's checkout opt-in behavior.

## Code review notes

_N/A_

## QA notes

**Testing instructions:**
1. Install & activate AutomateWoo;
2. Set up WooCommerce checkout opt-ins in both MP and AW settings;
3. Try to order a product on your site, see if there's only one checkbox (MailPoet's) displayed on the checkout page;
4. Check the checkbox and see that opt-in works for both MP and AW (in the Opt-ins submenu).
5. Install the [Plugins Load Order](https://wordpress.org/plugins/plugins-load-order/) plugin, check that the above-mentioned scenario works regardless of what plugin loads first, MP or AW.

Please test other cases in combination:
* AW active/disabled;
* Opt-in checkbox active/disabled (should display AW's checkbox in the latter case);
* Opt-in checkbox checked/unchecked.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5057]

## After-merge notes

_N/A_

[MAILPOET-5057]: https://mailpoet.atlassian.net/browse/MAILPOET-5057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/WooCommerce/Subscription.php
+++ b/mailpoet/lib/WooCommerce/Subscription.php
@@ -92,6 +92,7 @@ class Subscription {
   }
 
   public function extendWooCommerceCheckoutForm() {
+    $this->hideAutomateWooOptinCheckbox();
     $inputName = self::CHECKOUT_OPTIN_INPUT_NAME;
     $checked = $this->isCurrentUserSubscribed();
     if (!empty($_POST[self::CHECKOUT_OPTIN_INPUT_NAME])) {
@@ -183,6 +184,7 @@ class Subscription {
   }
 
   public function subscribeOnCheckout($orderId, $data) {
+    $this->triggerAutomateWooOptin();
     if (empty($data['billing_email'])) {
       // no email in posted order data
       return null;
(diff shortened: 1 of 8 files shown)


mailpoet/lib/WooCommerce/Subscription.php
```diff
@@ -252,6 +254,28 @@ public function handleSubscriberOptin(SubscriberEntity $subscriber, bool $checko
     return true;
   }
 
+  private function hideAutomateWooOptinCheckbox() {
+    if (!$this->wp->isPluginActive('automatewoo/automatewoo.php')) {
+      return false;
```
lysyjan: The return value from this method is never used. I guess we can use `return` and declare the return type to `void`. 
The same approach can be used for the following method.
What do you think?

lysyjan (commented):
Good job!
I just have one small comment, and I would like to ask if you considered adding an acceptance test. Maybe it is too complicated because it depends on another plugin.

mailpoet/lib/WooCommerce/Subscription.php
alex-mailpoet: Sure, we can do that, I have made the change.

alex-mailpoet:
> if you considered adding an acceptance test. Maybe it is too complicated because it depends on another plugin.

Yes, it turned out to be complicated because it involved another plugin, I ended up spending a lot of time on it, but I've finally managed to add an acceptance test. Main issues were:
* Not having access to the AW repository, I had to ask its developers to provide access for our CI GitHub account;
* The AW checkout opt-in worked locally, but didn't work on CI, it turned out I had to access its plugin admin UI after installation, or default settings wouldn't take effect.

lysyjan (approved):
Thank you! Good job

Aschepikov:
Hi @alex-mailpoet when I have both plugins active with checked checkboxes I see them both on the checkout page
image

Aschepikov (approved):
GJ!

PR #4745, opened 2023-02-20
Add a new k6 test: Subscribers filtering by status, list and search subscriber

veljkho:
## Description

More details in the linked Jira ticket below.

In this task, I wanted to add a new k6 tests:

- Subscribers filtering by subscribed status
- Subscribers filtering by a list
- Searching for a specific subscriber

Testable against qawp.net site (local site is empty, no testing data)

Additionally, I added a few changes:

- Renamed Automattic to automatic.
- Added adminEmail in config file
- Added eslint ignore for __ENV lines in config.js, we use it currently for running tests but not inside the test files so it triggers error when you want to commit

## Code review notes

After reviewing the code, please try running this individual test locally against qawp.net test site using the following command:

Note: Before running these, make sure to change the admin password in `config.js` from `password` to **reveal the password on the secret store** (search for `qawp`).

- ./do test:performance --url=https://qawp.net tests/performance/tests/subscribers-filtering.js --head
- ./do test:performance --url=https://qawp.net --scenario pullrequests

Expected result when running this test case: https://d.pr/i/kmSawi
Expected result when running scenarios: https://d.pr/i/iJpVqh

## Linked tickets

[MAILPOET-4955](https://mailpoet.atlassian.net/browse/MAILPOET-4955)

[MAILPOET-4955]: https://mailpoet.atlassian.net/browse/MAILPOET-4955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/tests/subscribers-filtering.js
+++ b/mailpoet/tests/performance/tests/subscribers-filtering.js
@@ -42,6 +42,16 @@ export function subscribersFiltering() {
       })
 
       .then(() => {
+        page.locator('[data-automation-id="filters_subscribed"]').click();
+        page.waitForSelector('[data-automation-id="filters_subscribed"]');
+        check(page, {
+          'subscribers filter is visible': page
+            .locator('[data-automation-id="listing_filter_segment"]')
+            .isVisible(),
+        });
+      })
+
+      .then(() => {
         page
           .locator('[data-automation-id="listing_filter_segment"]')
           .selectOption('3');
@@ -56,7 +66,7 @@ export function subscribersFiltering() {
 
       .then(() => {
         page.waitForNavigation({ waitUntil: 'networkidle' });
-        page.locator('#search_input').type(adminEmail);
+        page.locator('#search_input').type(adminEmail, { delay: 50 });
         sleep(2);
         page.waitForSelector('[data-automation-id="filters_subscribed"]');
         check(page, {
--- a/mailpoet/tests/performance/tests/subscribers-filtering.js
+++ b/mailpoet/tests/performance/tests/subscribers-filtering.js
@@ -16,13 +16,17 @@ import {
   thinkTimeMin,
   thinkTimeMax,
   headlessSet,
+  timeoutSet,
   adminEmail,
 } from '../config.js';
 import { authenticate } from '../utils/helpers.js';
 /* global Promise */
 
 export function subscribersFiltering() {
-  const browser = chromium.launch({ headless: headlessSet });
+  const browser = chromium.launch({
+    headless: headlessSet,
+    timeout: timeoutSet,
+  });
   const page = browser.newPage();
 
   group('Subscribers - Filter subscribers', function subscribersFiltering() {
@@ -55,7 +59,7 @@ export function subscribersFiltering() {
         page
           .locator('[data-automation-id="listing_filter_segment"]')
           .selectOption('3');
-        sleep(2);
+        page.waitForSelector('.mailpoet-listing-no-items');
         page.waitForSelector('[data-automation-id="filters_subscribed"]');
         check(page, {
           'subscribers filter is visible': page
@@ -67,7 +71,7 @@ export function subscribersFiltering() {
       .then(() => {
         page.waitForNavigation({ waitUntil: 'networkidle' });
         page.locator('#search_input').type(adminEmail, { delay: 50 });
-        sleep(2);
+        page.waitForSelector('.mailpoet-listing-no-items');
         page.waitForSelector('[data-automation-id="filters_subscribed"]');
         check(page, {
           'subscribers filter is visible': page
@@ -82,7 +86,7 @@ export function subscribersFiltering() {
       });
   });
 
-  sleep(randomIntBetween(`${thinkTimeMin}`, `${thinkTimeMax}`));
+  sleep(randomIntBetween(thinkTimeMin, thinkTimeMax));
 }
 
 export default function subscribersFilteringTest() {
--- a/mailpoet/tests/performance/tests/subscribers-filtering.js
+++ b/mailpoet/tests/performance/tests/subscribers-filtering.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-shadow */
 /* eslint-disable import/no-unresolved */
 /* eslint-disable import/no-default-export */
 /**
@@ -29,7 +28,7 @@ export function subscribersFiltering() {
   });
   const page = browser.newPage();
 
-  group('Subscribers - Filter subscribers', function subscribersFiltering() {
+  group('Subscribers - Filter subscribers', () => {
     page
       .goto(`${baseURL}/wp-admin/admin.php?page=mailpoet-subscribers`, {
         waitUntil: 'networkidle',
--- a/mailpoet/tests/performance/tests/subscribers-filtering.js
+++ b/mailpoet/tests/performance/tests/subscribers-filtering.js
@@ -1,0 +1,80 @@
+/* eslint-disable no-shadow */
+/* eslint-disable import/no-unresolved */
+/* eslint-disable import/no-default-export */
+/**
+ * External dependencies
+ */
+import { sleep, check, group } from 'k6';
+import { chromium } from 'k6/x/browser';
+import { randomIntBetween } from 'https://jslib.k6.io/k6-utils/1.1.0/index.js';
+
+/**
+ * Internal dependencies
+ */
+import {
+  baseURL,
+  thinkTimeMin,
+  thinkTimeMax,
+  headlessSet,
+  adminEmail,
+} from '../config.js';
+import { authenticate } from '../utils/helpers.js';
+/* global Promise */
+
+export function subscribersFiltering() {
+  const browser = chromium.launch({ headless: headlessSet });
+  const page = browser.newPage();
+
+  group('Subscribers - Filter subscribers', function subscribersFiltering() {
+    page
+      .goto(`${baseURL}/wp-admin/admin.php?page=mailpoet-subscribers`, {
+        waitUntil: 'networkidle',
+      })
+
+      .then(() => {
+        authenticate(page);
+      })
+
+      .then(() => {
+        return Promise.all([
+          page.waitForNavigation({ waitUntil: 'networkidle' }),
+        ]);
+      })
+
+      .then(() => {
+        page
+          .locator('[data-automation-id="listing_filter_segment"]')
+          .selectOption('3');
+        sleep(2);
+        page.waitForSelector('[data-automation-id="filters_subscribed"]');
+        check(page, {
+          'subscribers filter is visible': page
+            .locator('[data-automation-id="listing_filter_segment"]')
+            .isVisible(),
+        });
+      })
+
+      .then(() => {
+        page.waitForNavigation({ waitUntil: 'networkidle' });
+        page.locator('#search_input').type(adminEmail);
+        sleep(2);
+        page.waitForSelector('[data-automation-id="filters_subscribed"]');
+        check(page, {
+          'subscribers filter is visible': page
+            .locator('[data-automation-id="listing_filter_segment"]')
+            .isVisible(),
+        });
+      })
+
+      .finally(() => {
+        page.close();
+        browser.close();
+      });
+  });
+
+  sleep(randomIntBetween(`${thinkTimeMin}`, `${thinkTimeMax}`));
+}
+
+export default function subscribersFilteringTest() {
+  subscribersFiltering();
+}
--- a/mailpoet/tests/performance/README.md
+++ b/mailpoet/tests/performance/README.md
@@ -28,7 +28,7 @@ Automated k6 performance tests for MailPoet. To be used for benchmarking perform
 
 ### Installing k6
 
-You don't need to install it - it's Automattic! :)
+You don't need to install it - it's automatic!
 
 To execute the tests, use the following command:
 
--- a/mailpoet/tests/performance/config.js
+++ b/mailpoet/tests/performance/config.js
@@ -5,7 +5,7 @@ export const headlessSet = ['true', '1'].includes(
 export const scenario = __ENV.SCENARIO; // eslint-disable-line
 
 export const adminUsername = 'admin';
-export const adminPassword = 'uYNOVMJVd@!';
+export const adminPassword = 'password';
 export const adminEmail = 'test@test.com';
 
 export const thinkTimeMin = '1';
--- a/mailpoet/tests/performance/config.js
+++ b/mailpoet/tests/performance/config.js
@@ -1,11 +1,12 @@
-export const baseURL = __ENV.URL || 'http://localhost:9500';
+export const baseURL = __ENV.URL || 'http://localhost:9500'; // eslint-disable-line
 export const headlessSet = ['true', '1'].includes(
-  `${__ENV.HEADLESS || 'true'}`.toLowerCase(),
+  `${__ENV.HEADLESS || 'true'}`.toLowerCase(), // eslint-disable-line
 );
-export const scenario = __ENV.SCENARIO;
+export const scenario = __ENV.SCENARIO; // eslint-disable-line
 
 export const adminUsername = 'admin';
-export const adminPassword = 'password';
+export const adminPassword = 'uYNOVMJVd@!';
+export const adminEmail = 'test@test.com';
 
 export const thinkTimeMin = '1';
 export const thinkTimeMax = '4';
--- a/mailpoet/tests/performance/config.js
+++ b/mailpoet/tests/performance/config.js
@@ -3,6 +3,7 @@ export const headlessSet = ['true', '1'].includes(
   `${__ENV.HEADLESS || 'true'}`.toLowerCase(), // eslint-disable-line
 );
 export const scenario = __ENV.SCENARIO; // eslint-disable-line
+export const timeoutSet = '2m';
 
 export const adminUsername = 'admin';
 export const adminPassword = 'password';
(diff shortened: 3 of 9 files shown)


mailpoet/tests/performance/tests/subscribers-filtering.js
```diff
@@ -0,0 +1,90 @@
...
+        });
+      })
+
+      .finally(() => {
+        page.close();
+        browser.close();
+      });
+  });
+
+  sleep(randomIntBetween(`${thinkTimeMin}`, `${thinkTimeMax}`));
```
johnolek: I think you should be able to do this:

```suggestion
  sleep(randomIntBetween(thinkTimeMin, thinkTimeMax));
```

mailpoet/tests/performance/tests/subscribers-filtering.js
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable no-shadow */
+/* eslint-disable import/no-unresolved */
+/* eslint-disable import/no-default-export */
+/**
+ * External dependencies
+ */
+import { sleep, check, group } from 'k6';
+import { chromium } from 'k6/x/browser';
+import { randomIntBetween } from 'https://jslib.k6.io/k6-utils/1.1.0/index.js';
```
johnolek, line 8: It might be nice to include helpers like this directly in our codebase, maybe in `helpers.js`?, just so we don't have to rely on fetching dependencies over the internet.

mailpoet/tests/performance/tests/subscribers-filtering.js
```diff
@@ -0,0 +1,90 @@
...
+      .finally(() => {
+        page.close();
+        browser.close();
+      });
+  });
+
+  sleep(randomIntBetween(`${thinkTimeMin}`, `${thinkTimeMax}`));
+}
+
+export default function subscribersFilteringTest() {
```
johnolek, line 91: I don't see `subscribersFilteringTest` used anywhere, but I do see `newsletterListing` used. Is this export getting used implicitly somehow, or is it necessary at all?

mailpoet/tests/performance/tests/subscribers-filtering.js
```diff
@@ -0,0 +1,90 @@
...
+            .locator('[data-automation-id="listing_filter_segment"]')
+            .isVisible(),
+        });
+      })
+
+      .then(() => {
+        page
+          .locator('[data-automation-id="listing_filter_segment"]')
+          .selectOption('3');
+        sleep(2);
```
johnolek: I'm not sure I understand this sleep. Is it to ensure the ajax request has started? Maybe instead you could wait for the Loading subscribers text?
image

mailpoet/tests/performance/tests/subscribers-filtering.js
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable no-shadow */
```
johnolek: Is this necessary?

johnolek (changes requested):
Hey @veljkho, I'm not as familiar with the k6 stuff as I probably should be. This is my first time trying to do anything with it, and I wasn't able to get it to run. I was getting output like this:

```
~/Workspace/mailpoet (k6/add-new-test-subscribers-filtering*) » ./do test:performance                                                                                                                                                                johnoleksowicz@Work-MacBook-Pro
 [Exec] Running COMPOSE_HTTP_TIMEOUT=200 docker-compose run --rm -it setup in /var/www/html/wp-content/plugins/mailpoet/tests/performance
sh: 1: docker-compose: not found
 [Exec]  Exit code 127  Time 0.022s
 [Exec] Running php /var/www/html/wp-content/plugins/mailpoet/tools/xk6browser.php run --env 'URL=' --env 'HEADLESS=true' --env 'SCENARIO=' /var/www/html/wp-content/plugins/mailpoet/tests/performance/scenarios.js in /var/www/html/wp-content/plugins/mailpoet

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

  execution: local
     script: /var/www/html/wp-content/plugins/mailpoet/tests/performance/scenarios.js
     output: -

  scenarios: (100.00%) 2 scenarios, 2 max VUs, 5m30s max duration (incl. graceful stop):
           * nightlytests: 3 iterations for each of 1 VUs (maxDuration: 5m0s, exec: nightly, gracefulStop: 30s)
           * pullrequests: 1 iterations for each of 1 VUs (maxDuration: 2m0s, exec: pullRequests, gracefulStop: 30s)

ERRO[0000] GoError: launching browser: exec: no command
        at reflect.methodValueCall (native)
        at newsletterListing (file:///var/www/html/wp-content/plugins/mailpoet/tests/performance/tests/newsletter-listing.js:19:36(7))
        at pullRequests (file:///var/www/html/wp-content/plugins/mailpoet/tests/performance/scenarios.js:59:2(4))
        at native  executor=per-vu-iterations scenario=pullrequests source=stacktrace
```

Is there something else I need to do to install the k6 dependencies?

I have a couple of small questions/suggestions for the PR itself. Please let me know if anything isn't clear.

Edit: do you have credentials for the test site shared in the secret store? I saw a link to quickforget in an older PR but the link is expired.

mailpoet/tests/performance/tests/subscribers-filtering.js
```diff
@@ -0,0 +1,90 @@
...
+          'subscribers filter is visible': page
+            .locator('[data-automation-id="listing_filter_segment"]')
+            .isVisible(),
+        });
+      })
+
+      .then(() => {
+        page
+          .locator('[data-automation-id="listing_filter_segment"]')
+          .selectOption('3');
```
johnolek, line 60: Are we concerned about making these tests general enough to run against multiple instances? Hardcoding the filter ID might not be the best way to go if so, but if you have a specific reason for using this filter, or there's no easy way to make it general, this should be fine for now.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: I don't think it is, I could remove it if you think the same. Maybe some leftover from the beginning.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: This is by default and according to documentation to use it this way if working with k6.
Maybe only for the 3rd one, random number, but since i use it already in all tests one separate PR would be necessary for this change. I would leave it as-is in this PR.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: For some reason it won't work when choosing the list by its name, I will try it once again.
The list `Newsletter mailing list` is always the 3rd by every MailPoet installation, that's why I used it as safe bet for now.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: I can try that, yes. The reason was to trigger this loading state, and adding sleeps is widely accepted in performance testing, it is even recommended to use it.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: I will try that, thanks.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: The name of the function is required by Linter, can't be without a name and that's why I have to name all those functions and use `subscribersFiltering` in scenarios.

veljkho:
> I'm not as familiar with the k6 stuff as I probably should be. This is my first time trying to do anything with it, and I wasn't able to get it to run. I was getting output like this:
> Is there something else I need to do to install the k6 dependencies?

You should try running `./do test:performance` or any other command inside `mailpoet` folder (do `cd mailpoet` before running).
I will update the PR with the commands to run both `scenarios` and this individual test case, sorry I should have done that.

I'm glad to see you are interested in k6 performance and I will help you understand the project @johnolek

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: But I found out that it is important to have in every test because it is used to disable eslint error for having 2 same `subscribersFiltering` names in the test, one for the name of the test and the second in the bottom export function.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: Nope, still seeing [this error](https://d.pr/i/9No8fN) when using `({ label: 'Newsletter mailing list (19)' })` and I need to mention how much subscribers are there so the list name is not just `Newsletter mailing list` unfortunately. I'll keep the `3` for now, maybe I have to report a potential bug to k6 for this since it's still beta.

mailpoet/tests/performance/tests/subscribers-filtering.js
johnolek: Sounds good! Definitely not a big deal.

mailpoet/tests/performance/tests/subscribers-filtering.js
johnolek: What I meant is that I don't see this default export being referenced anywhere. In `scenarios.js` where you have this:

```js
import { subscribersFiltering } from './tests/subscribers-filtering.js';
```

you're importing a named import, i.e. the function that you define as an export on line 25 of `subscribers-filtering.js`:

```js
export function subscribersFiltering() {
```

If you were using the default export there wouldn't be the curly braces.

Long story short, does the code still work if you completely remove the default export?

mailpoet/tests/performance/tests/subscribers-filtering.js
johnolek: My concern was that the sleep of 2 seconds might not always be enough time in some rare cases to trigger the loading state, which could lead to unexpected results. Probably not likely, or maybe not even possible. I just always prefer to wait for something specific if it's feasible instead of relying on wait times. Not only can it be more reliable, but it also makes the test code more self documenting -- someone reading it would immediately understand the state you're waiting for.

It's interesting to hear that sleeps are encouraged in performance testing, though, and I'm interested to hear how it might be useful in this case other than waiting to trigger the loading state.

mailpoet/tests/performance/tests/subscribers-filtering.js
johnolek: The linter is throwing the error because the inner function, the group callback, has the same function name as the top level function:

```js
export function subscribersFiltering() { // <-----------
  const browser = chromium.launch({
    headless: headlessSet,
    timeout: timeoutSet,
  });
  const page = browser.newPage();

  group('Subscribers - Filter subscribers', function subscribersFiltering() { // <-----------
```

I think it should be possible to use an anonymous arrow function like this, avoiding the error:

```js
group('Subscribers - Filter subscribers', () => {
```

This would also be more idiomatic for the style of javascript we have in the codebase.

mailpoet/tests/performance/tests/subscribers-filtering.js
johnolek: Good point about it always being the third list!

I agree it's best to avoid having to know the number of subscribers.

What I was wondering is if there's an easy way to say, "Select the 3rd option in this dropdown". You could maybe try the [press](https://k6.io/docs/javascript-api/k6-browser/api/locator/press/) method to press the down arrow twice and then hit enter to select it? I'm not sure since I haven't been able to get this running myself.

It's also totally fine to leave it as is, especially knowing that `3` should  almost always work for any installation.

johnolek:
> You should try running ./do test:performance or any other command inside mailpoet folder (do cd mailpoet before running).

@veljkho I did try running the command in the `mailpoet` directory and the root directory, and neither worked for me. 

I noticed I had to change `docker-compose` to `docker compose` to get those commands to work for me. Maybe I have an out of date version of `docker-compose`? Even after changing that, clearing out the `tools/vendor` directory and trying again, I still get the same error.

```
ERRO[0000] GoError: launching browser: exec: no command
```

Unfortunately I don't have a lot of time I can dedicate to this right now, and googling for the error message isn't turning up anything helpful.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: Good point. So if I remove the default export the code is working but then not executing it in `scenarios.js`.
I tried also adding `default` to the test function and it worked to execute it but again not in `scenarios.js`.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: Yes I agree with you, I don't want to use sleeps and will try to avoid it as much as possible. With much larger sleeps of 5 to 10 seconds, that's how they slow down the tests but we have mission to have this test suite running fast. I can ask this question about using sleeps often in tests in some of the upcoming meetings with k6 happy hours.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: > group('Subscribers - Filter subscribers', () => {

This one worked fine and I made changes, thanks!

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: Yes, for now selecting value 3 as it will be always `Newsletter mailing list` is fine for now until I figure out what's wrong with selecting a label. I have it on my notes of improving existing code after the project ends.

veljkho:
> > You should try running ./do test:performance or any other command inside mailpoet folder (do cd mailpoet before running).
> 
> @veljkho I did try running the command in the `mailpoet` directory and the root directory, and neither worked for me.
> 
> I noticed I had to change `docker-compose` to `docker compose` to get those commands to work for me. Maybe I have an out of date version of `docker-compose`? Even after changing that, clearing out the `tools/vendor` directory and trying again, I still get the same error.
> 
> ```
> ERRO[0000] GoError: launching browser: exec: no command
> ```
> 
> Unfortunately I don't have a lot of time I can dedicate to this right now, and googling for the error message isn't turning up anything helpful.

I'm wondering if the correct xk6browser binary file was downloaded for your machine, the error you see says that it is unable to run the chrome browser. Could happen due to wrong binary file, please check your folder name in `/tools/vendor/`.. should be `xk6-browser-v0.7.0-macos-amd64` (if you're on Mac with Intel CPU)

mailpoet/tests/performance/tests/subscribers-filtering.js
johnolek: I think I understand now. I didn't realize that if you want to run a specific test directly it depends on there being a default export.

It doesn't work to change it to `export default function subscribersFiltering() {...` because, I believe, it no longer creates a named export, and `scenarios.js` is explicitly importing named exports.

I think the best way to do this would be to keep the function defined the way it is now, `export function subscribersFiltering() {...`, and then for the default export, instead of creating a new function that just calls the first one, you can export the function itself as the default like this:

```js
export default subscribersFiltering;
```

This seemed to work for me anyway!

mailpoet/tests/performance/tests/subscribers-filtering.js
johnolek: FYI, I did some experimentation of my own, and the only way I could find to select the `nth` option in a select list was to first evaluate some javascript to return an array of all the options:

```js
      .then(() => {
        const options = page.evaluate(() => {
          const select = document.querySelector(
            '[data-automation-id="listing_filter_segment"]',
          );
          return Array.from(select.options).map((option) => option.value);
        });
        console.log({options});
        page
          .locator('[data-automation-id="listing_filter_segment"]')
          .selectOption(options[2]);
```

The `options` array looked like this for the test site:
`["","without-list","3","7","8","9","10","11","2","1"]`

So selecting `options[2]` would be equivalent to selecting `'3'`.

This may or may not be helpful!

johnolek (dismissed):
@veljkho The reason the tests weren't running for me was something that should have been very obvious: I didn't have chrome itself installed. I thought the binary would include a full version of chromium but the tests didn't start working for me until I installed the chrome app itself. 

Thanks for making the updates, and I included one last note about exports, but it's definitely nothing that needs to be addressed so I'm marking this as approved. Feel free to merge it as-is or if you'd like to make a change to the exports you're welcome to do that first.

mailpoet/tests/performance/tests/subscribers-filtering.js
veljkho: I tried your suggestion but was unable to run the test in `scenarios.js`. It worked only by executing the individual test.
So to confirm, I tried with:
`export default subscribersFiltering;`
to be placed at the top bottom of the test and executing scenarios.
also tried with:
`export default subscribersFiltering();`
just in case, but didn't work.

I will continue as-is and if you feel it really worked for you to execute scenarios as well, please ping me on Slack to make a separate PR for that change. Thanks for very good review of the PR!

veljkho:
Thanks @lysyjan for helping me fix conflicting issues with this PR, I have checked everything and the PR is ready for merging. Please approve the PR so I can merge, @johnolek has approved the changes already, no new changes added after his last approval.

lysyjan (approved):

PR #4747, opened 2023-02-20
Coupon block error handling log errors for the newsletter stop sending for this newsletter [MAILPOET-4983]

samnajian:
## Description

This PR catches issues thrown by the Coupon block in the rendering/processing phase, logs, and notifies the user about it so that they can fix the issue and continue.

## QA notes

You'll need to make a newsletter with a coupon block run into a problem, one way to do this is adding a newsletter with a coupon block, but when on the `Send` page, just before clicking on the `Send` button, go to plugins' page and deactivate WooCommerce, this should prevent the email from being sent and make you end up with a listing like the attached screenshot.
https://d.pr/i/lZf3sV

## Linked tickets
[MAILPOET-4983]

[MAILPOET-4983]: https://mailpoet.atlassian.net/browse/MAILPOET-4983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/common/listings/_stories/newsletter_status.tsx
+++ b/mailpoet/assets/js/src/common/listings/_stories/newsletter_status.tsx
@@ -2,6 +2,7 @@ import { addHours, subHours } from 'date-fns';
 import { MailPoet } from 'mailpoet';
 import { NewsletterStatus } from '../newsletter_status';
 import { Heading } from '../../typography/heading/heading';
+import { NewsletterStatus as NewsletterStatusEnum } from '../../../newsletters/models';
 
 MailPoet.I18n.add('notSentYet', 'Not sent yet!');
 MailPoet.I18n.add('sent', 'Sent');
@@ -57,7 +58,7 @@ export function NewsletterStatuses() {
       <div className="mailpoet-gap" />
 
       <Heading level={3}>Sent without queue</Heading>
-      <NewsletterStatus status="sent" />
+      <NewsletterStatus status={NewsletterStatusEnum.Sent} />
     </>
   );
 }
--- a/mailpoet/assets/js/src/common/listings/_stories/newsletter_status.tsx
+++ b/mailpoet/assets/js/src/common/listings/_stories/newsletter_status.tsx
@@ -1,8 +1,8 @@
 import { addHours, subHours } from 'date-fns';
 import { MailPoet } from 'mailpoet';
-import { NewsletterStatus } from '../newsletter_status';
-import { Heading } from '../../typography/heading/heading';
 import { NewsletterStatus as NewsletterStatusEnum } from 'newsletters/models';
+import { NewsletterStatus } from '../newsletter_status';
+import { Heading } from '../../typography';
 
 MailPoet.I18n.add('notSentYet', 'Not sent yet!');
 MailPoet.I18n.add('sent', 'Sent');
--- a/mailpoet/assets/js/src/common/listings/_stories/newsletter_status.tsx
+++ b/mailpoet/assets/js/src/common/listings/_stories/newsletter_status.tsx
@@ -2,7 +2,7 @@ import { addHours, subHours } from 'date-fns';
 import { MailPoet } from 'mailpoet';
 import { NewsletterStatus } from '../newsletter_status';
 import { Heading } from '../../typography/heading/heading';
-import { NewsletterStatus as NewsletterStatusEnum } from '../../../newsletters/models';
+import { NewsletterStatus as NewsletterStatusEnum } from 'newsletters/models';
 
 MailPoet.I18n.add('notSentYet', 'Not sent yet!');
 MailPoet.I18n.add('sent', 'Sent');
--- a/mailpoet/assets/js/src/common/listings/newsletter_status.tsx
+++ b/mailpoet/assets/js/src/common/listings/newsletter_status.tsx
@@ -6,7 +6,7 @@ import { Tooltip } from '../tooltip/tooltip';
 import {
   NewsLetter,
   NewsletterStatus as NewsletterStatusEnum,
-} from '../../newsletters/models';
+} from 'newsletters/models';
 
 type CircularProgressProps = {
   percentage: number;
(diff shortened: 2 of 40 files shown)


mailpoet/lib/Newsletter/Renderer/Preprocessor.php
```diff
@@ -57,6 +70,17 @@ public function process(NewsletterEntity $newsletter, $content, bool $preview =
 
     $contentBlocks = $this->couponPreProcessor->processCoupons($newsletter, $contentBlocks, $preview);
 
+    try {
+      $contentBlocks = $this->couponPreProcessor->processCoupons($newsletter, $contentBlocks, $preview);
+    } catch (NewsletterProcessingException $e) {
```
triple0t: This looks repeated. This [line](https://github.com/mailpoet/mailpoet/blob/57636955e7566e432fcf2e2559941c0586c39ea0/mailpoet/lib/Newsletter/Renderer/Preprocessor.php#L71) and this [line](https://github.com/mailpoet/mailpoet/blob/57636955e7566e432fcf2e2559941c0586c39ea0/mailpoet/lib/Newsletter/Renderer/Preprocessor.php#L74) are the same

mailpoet/assets/js/src/common/listings/_stories/newsletter_status.tsx
```diff
@@ -2,6 +2,7 @@ import { addHours, subHours } from 'date-fns';
 import { MailPoet } from 'mailpoet';
 import { NewsletterStatus } from '../newsletter_status';
 import { Heading } from '../../typography/heading/heading';
+import { NewsletterStatus as NewsletterStatusEnum } from '../../../newsletters/models';
```
triple0t: I believe this can be replaced with
```suggestion
import { NewsletterStatus as NewsletterStatusEnum } from 'newsletters/models';
```

mailpoet/assets/js/src/common/listings/newsletter_status.tsx
```diff
@@ -3,6 +3,10 @@ import classnames from 'classnames';
 import { addDays, differenceInMinutes, isFuture, isPast } from 'date-fns';
 import { t } from 'common/functions/t';
 import { Tooltip } from '../tooltip/tooltip';
+import {
+  NewsLetter,
+  NewsletterStatus as NewsletterStatusEnum,
+} from '../../newsletters/models';
```
triple0t: This can be replaced as well.
```suggestion
} from 'newsletters/models';
```

mailpoet/tests/integration/Logging/LogHandlerTest.php
```diff
@@ -34,6 +34,7 @@ public function testItCreatesLog() {
       'context' => [],
       'channel' => 'name',
       'datetime' => $time,
+      'message' => 'some log message',
```
triple0t, line 37: I thought this `message` parameter was optional.
Should the tests fail for optional parameters?

mailpoet/assets/js/src/newsletters/listings/queue_status.tsx
```diff
@@ -1,34 +1,28 @@
...
-import { APIErrorsNotice } from 'notices/api_errors_notice.tsx';
+import { APIErrorsNotice } from 'notices/api_errors_notice';
 import { Button } from 'common/button/button';
 import { NewsletterStatus } from 'common/listings/newsletter_status';
 import { withBoundary } from '../../common';
+import {
+  NewsLetter,
+  NewsletterStatus as NewsletterStatusEnum,
+} from '../models';
+import { confirmAlert } from '../../common/confirm_alert';
```
triple0t: Kindly update
```suggestion
import { confirmAlert } from 'common/confirm_alert';
```

mailpoet/assets/js/src/newsletters/listings/queue_status.tsx
```diff
@@ -54,11 +47,31 @@ function QueueSending({ newsletter }) {
       .fail((response) => setErrors(response.errors));
   };
 
+  const resumeSending = async () => {
+    setErrors([]);
+    await requestResume();
+  };
+
+  const confirmAndResumeSending = async () => {
+    setErrors([]);
```
triple0t: This is interesting.
Why do you want to clear the errors after clicking the button, not after the prompt is confirmed?

triple0t (commented):
Well done.

Works well. Just a few notes

mailpoet/assets/js/src/newsletters/listings/queue_status.tsx
samnajian: yea, you're right, that's a mistake, thanks for catching it

mailpoet/lib/Newsletter/Renderer/Preprocessor.php
samnajian: nice catch, should be from a merge conflict resolution.

mailpoet/tests/integration/Logging/LogHandlerTest.php
samnajian: It's actually `Nullable` in db but not optional, it's always set when using [`Logger::addRecord` ](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Logger.php#L321)

triple0t (dismissed):
Thank you for fixing the issues highlighted.

I noticed a little bug I wanted to inform you.

When the Newsletter has the status `corrupt` and sending is resumed.
The Newsletter does not re-process the coupon block.

Screenshot 2023-02-28 at 10 05 23 AM

I'm approving for now. Please decide how to proceed

triple0t (approved):

Aschepikov:
HI @samnajian, can we make this error message viewable until the problem will be resolved? 
In my case, I see the error 
image
Then I am resuming the sending
image
Confirming
image
Reloading the page 
image
And after that I don't see any messages 
image
It is not critical but if you can it will be better to make it viewable all the time while the problem is here.
P.S. I checked it a few times, sometime you need to resume the sending 1 time sometimes 2.

samnajian:
@Aschepikov as I understand, you've resumed and then refreshed the page? in this case the notice should go away because you have triggered sending again and in the new run there is no problem yet, specially since the blue button reads `pause`, i.e sending is in progress., what do you think?

Aschepikov:
From that side yes, I agree with you. If it will be paused again with error its good

Aschepikov:
image
@samnajian  In my case it was paused again without error and after reloading no error to0

Aschepikov:
I think it is ok, it is not the usual case, and after our automatic reloading of the list all errors are available again, I am approving it

Aschepikov (approved):

PR #4748, opened 2023-02-21
Order status changed trigger [MAILPOET-4935]

websupporter:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs
https://github.com/mailpoet/mailpoet-premium/pull/686

## Linked tickets

[MAILPOET-4935]

## After-merge notes

_N/A_

[MAILPOET-4935]: https://mailpoet.atlassian.net/browse/MAILPOET-4935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/AdminPages/Pages/AutomationEditor.php
```diff
@@ -85,7 +92,12 @@ public function render() {
         'root' => rtrim($this->wp->escUrlRaw(admin_url('admin-ajax.php')), '/'),
       ],
       'user_roles' => $roles->get_names(),
-    ]);
+    ];
+
+    if (class_exists('Woocommerce')) {
+      $variables['woo_order_status'] = \wc_get_order_statuses();
+    }
```
JanJakes: For consistency, could we use `MailPoet\WooCommerce\Helper:isWooCommerceActive()` and probably also call `wc_get_order_statuses` via a helper?

I think it could also be nice to nest this (and maybe make it nullable instead of existing/not-existing?), sth like:
```php
  'woocommerce' => [
    'order_statuses' => ...,
  ],
```

Also seeing the `wc_get_order_statuses`, I guess the variable name should be `order_statuses` rather than `order_status`.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/order_status_changed/edit/order_status.ts
```diff
@@ -0,0 +1,25 @@
+import { OrderStatusOptions } from '../../../../../types/filters';
+import { Hooks } from '../../../../../../hooks';
+
+declare global {
+  interface Window {
+    mailpoet_woo_order_status: Record<string, string>;
+  }
+}
+export const COMPLETED_ORDER_STATUS = 'wc-completed';
+export const orderStatus: OrderStatusOptions = Hooks.applyFilters(
```
JanJakes: Could we improve the naming? Something like `getOrderStatusOptions` or what exactly does it return...

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/order_status_changed/edit/order_status_panel.tsx
```diff
@@ -0,0 +1,52 @@
...
+      />
+      {showPremiumModal && (
+        <PremiumModal
+          onRequestClose={() => setShowPremiumModal(false)}
+          tracking={{
+            utm_medium: 'upsell_modal',
+            utm_campaign: 'trigger_order_status_changed',
+          }}
+        >
+          {__('You cannot change the status.', 'mailpoet')}
```
JanJakes: I'm not sure if this is a very friendly message. We usually used something like "Doing ... is a premium feature.". What do you think?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/order_status_changed/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  title: __('Order status changed', 'mailpoet'),
+  foreground: '#2271b1',
+  background: '#f0f6fc',
+  description: __(
+    'Start the automation when an order is created or changed to a specific status.',
+    'mailpoet',
+  ),
+  subtitle: () => _x('Trigger', 'noun', 'mailpoet'),
+  icon: () => (
+    <div style={{ width: '100%', height: '100%', scale: '1.12' }}>
```
JanJakes: I think the icon needs some size tuning:

Screenshot 2023-03-07 at 15 07 56

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/order_status_changed/edit/order_status.ts
```diff
@@ -0,0 +1,25 @@
...
+      const status = {
+        value,
+        label: window.mailpoet_woo_order_status[value],
+        isDisabled: value !== COMPLETED_ORDER_STATUS,
+      };
+      return status;
+    })
+    .reduce((acc, curr) => {
+      acc[curr.value] = curr;
+      return acc;
```
JanJakes: Looking further, I think a plain JS object has no order of properties defined. That may (or may not 😄) result in strange behavior in some browsers. I think we should either use an array or a `Map` (which preservers order as well).

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/OrderSubject.php
```diff
@@ -0,0 +1,43 @@
...
+use MailPoet\Automation\Engine\Integration\Subject;
+use MailPoet\Automation\Integrations\MailPoet\Payloads\OrderPayload;
+use MailPoet\NotFoundException;
+use MailPoet\Validator\Builder;
+use MailPoet\Validator\Schema\ObjectSchema;
+
+/**
+ * @implements Subject<OrderPayload>
+ */
+class OrderSubject implements Subject {
```
JanJakes, line 16: My natural question here is "should we create a new `WooCommerce integration`?"... but then I'm also thinking we may need to use a MP subscriber at some point 🤔 

I guess in the ideal case, this should go to WC integration, and MP could "derive" subscriber subject from the order somehow...?

(And then there is the question... what will happen with AutomateWoo?)

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/OrderSubject.php
```diff
@@ -0,0 +1,43 @@
...
+
+  const KEY = 'woocommerce:order';
```
JanJakes, line 18: Aha! So it really seems to be "non-mailpoet" 😄

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/OrderSubject.php
```diff
@@ -0,0 +1,43 @@
...
+    return Builder::object([
+      'order_id' => Builder::integer()->required(),
+    ]);
+  }
+
+  public function getPayload(SubjectData $subjectData): Payload {
+    $id = $subjectData->getArgs()['order_id'];
+    $order = wc_get_order($id);
+    if (!$order instanceof \WC_Order) {
+      // translators: %d is the ID.
```
JanJakes: Haha, we can also tell them "is the _order_ ID", but I guess that clear anyway.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/OrderStatusChangedTrigger.php
```diff
@@ -0,0 +1,138 @@
...
+  private function findSubscriber(\WC_Order $order): ?SubscriberEntity {
+    $subscriber = $this->subscribersRepository->findOneBy(['email' => $order->get_billing_email()]);
+    if ($subscriber instanceof SubscriberEntity) {
+      return $subscriber;
+    }
+
+    $subscriber = $this->subscribersRepository->findOneBy(['wpUserId' => $order->get_user_id()]);
+    if ($subscriber instanceof SubscriberEntity) {
+      return $subscriber;
+    }
```
JanJakes, line 127: It's probably a small edge case, but shouldn't the `wpUserId` go first, so we try to match the most accurately possible? That said, it could then happen that an email would be sent to a different email address than the one that was used during the order, right? 🤔

mailpoet/lib/Automation/Integrations/MailPoet/Payloads/OrderPayload.php
```diff
@@ -0,0 +1,33 @@
...
+  public function getEmail(): string {
+    return $this->order->get_billing_email();
+  }
+
+  public function getId(): int {
+    return $this->order->get_id();
+  }
+
+  public function getFormattedOrderStatus(): string {
+    return 'wc-' . $this->order->get_status();
```
JanJakes: Why is the `wc-` prefix needed? Why not using the Woo-original naming?

mailpoet/lib/Automation/Integrations/MailPoet/Payloads/CustomerPayload.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace MailPoet\Automation\Integrations\MailPoet\Payloads;
+
+use MailPoet\Automation\Engine\Integration\Payload;
+
+class CustomerPayload implements Payload {
+
+
+
+  private $customer;
```
JanJakes, line 10: You seem to add more and more empty lines with every new class 😂

Could you also pls type the properties? Here and in all the other new classes? Otherwise, `$this->property` is not typed.

mailpoet/lib/Automation/Integrations/MailPoet/ContextFactory.php
```diff
@@ -18,9 +18,14 @@ public function __construct(
...
     return [
       'segments' => $this->getSegments(),
+      'woo_order_status' => $this->getOrderStatuses(),
     ];
   }
 
+  private function getOrderStatuses(): array {
+    return (function_exists('wc_get_order_statuses')) ? wc_get_order_statuses() : [];
+  }
+
```
JanJakes, line 39: Then the first commit of this PR could be dropped, right? And I think that this should probably go to a WC integration, if we're able to do that and we should use the WC helper here.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/OrderStatusChangedTrigger.php
```diff
@@ -0,0 +1,138 @@
...
+      OrderSubject::KEY,
+    ];
+  }
+
+  public function validate(StepValidationArgs $args): void {
+    // TODO: Implement validate() method.
+  }
+
+  public function registerHooks(): void {
+    add_action(
```
JanJakes: Probably a WP helper should be used here.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/OrderStatusChangedTrigger.php
```diff
@@ -0,0 +1,152 @@
...
+    return [
+      SubscriberSubject::KEY,
+      OrderSubject::KEY,
+      OrderStatusChangeSubject::KEY,
+      CustomerSubject::KEY,
+    ];
+  }
+
+  public function validate(StepValidationArgs $args): void {
+    // TODO: Implement validate() method.
```
JanJakes: I guess we should remove `TODO` if it needs no further validation.

JanJakes (commented):
Good stuff! But there are some things that need a bit of polishing:

1. Use WP and WC helpers, the PR has a mixture of global methods and the helpers.
2. There was some data added to the page variables, then moved elsewhere, but not really reverted.
3. I guess we should create a new "WooCommerce" integration and think about "deriving" some MP subjects from those. We can also leave it "hardcoded" in the trigger now, I guess.
4. Some properties don't have types annotated. I guess

mailpoet/lib/Automation/Integrations/MailPoet/Payloads/OrderPayload.php
websupporter: Woo itself is inconsistent. But I do this sanitization now in the `isTriggeredBy` itself.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/OrderStatusChangedTrigger.php
websupporter: > email would be sent to a different email address than the one that was used during the order, right? 

Yes. They can differ.

mailpoet/lib/Automation/Integrations/MailPoet/Subjects/OrderSubject.php
websupporter: added :)

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/OrderStatusChangedTrigger.php
websupporter: https://github.com/mailpoet/mailpoet/blob/MAILPOET-4935-order-trigger/mailpoet/lib/Automation/Integrations/MailPoet/Triggers/OrderStatusChangedTrigger.php#L77

mailpoet/lib/Automation/Integrations/MailPoet/ContextFactory.php
websupporter: is gone.

mailpoet/lib/Automation/Integrations/MailPoet/Payloads/CustomerPayload.php
websupporter: :see_no_evil:  cleaned up.

mailpoet/lib/Automation/Integrations/MailPoet/Triggers/OrderStatusChangedTrigger.php
websupporter: gone.

mailpoet/lib/AdminPages/Pages/AutomationEditor.php
websupporter: thanks! adjusted accordingly.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/order_status_changed/edit/order_status.ts
websupporter: done. https://github.com/mailpoet/mailpoet/blob/MAILPOET-4935-order-trigger/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/order_status_changed/edit/order_status.ts#L6

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/order_status_changed/edit/order_status_panel.tsx
websupporter: Thanks! Changed to 

"Changing the status is a premium feature."

https://github.com/mailpoet/mailpoet/blob/MAILPOET-4935-order-trigger/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/order_status_changed/edit/order_status_panel.tsx#L61

websupporter:
closing in favor of new pr

PR #4750, opened 2023-02-22
Add a new k6 test: Adding a new subscriber

veljkho:
## Description

Adding a new k6 performance test for measuring performance while adding a new subscriber.

Additionally,
- updated `xk6-browser` binary version to its latest 8.1 version
- increased max duration time for scenarios
- added a new helper method `selectInSelect2` for react select component

## Code review notes

For executing this test locally, please run the command inside `/mailpoet/` folder:

Note: Since we don't have any data locally yet, please change the password in `config.js` from `password` to the correct one for the site qawp.net (copy/paste the password from the secret store: search for `qawp`)

- ./do test:performance --url=https://qawp.net tests/performance/tests/subscribers-adding.js

The expected result you should see is: [screenshot](https://d.pr/i/wGwhEe).
If you don't see this result at first try, please run it one more time (the test is not flaky, but the server and site qawp.net can sometimes show 502 error on a first page load).

To watch the test run in a real browser add this argument `--head` in the end of the test command above.

## Linked PRs

This branch was created from this pull request #4745

## Linked tickets

[MAILPOET-4956]

[MAILPOET-4956]: https://mailpoet.atlassian.net/browse/MAILPOET-4956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/config.js
+++ b/mailpoet/tests/performance/config.js
@@ -9,5 +9,9 @@ export const adminUsername = 'admin';
 export const adminPassword = 'password';
 export const adminEmail = 'test@test.com';
 
+export const firstName = 'John';
+export const lastName = 'Doe';
+export const defaultListName = 'Newsletter mailing list';
+
 export const thinkTimeMin = '1';
 export const thinkTimeMax = '4';
--- a/mailpoet/tests/performance/scenarios.js
+++ b/mailpoet/tests/performance/scenarios.js
@@ -5,6 +5,7 @@ import { newsletterListing } from './tests/newsletter-listing.js';
 import { subscribersListing } from './tests/subscribers-listing.js';
 import { settingsBasic } from './tests/settings-basic.js';
 import { subscribersFiltering } from './tests/subscribers-filtering.js';
+import { subscribersAdding } from './tests/subscribers-adding.js';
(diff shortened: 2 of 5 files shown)


mailpoet/tests/performance/config.js
```diff
@@ -9,5 +9,9 @@ export const adminUsername = 'admin';
 export const adminPassword = 'password';
```
samnajian, line 9: I was thinking maybe we could rename this file to `config.sample.js`, just like what we do for `.env` file, and ask the devs to copy it to `.config.js`, especially since we'll need to update and add the correct password, this will make sure we're not committing the password to the repo by mistake.

samnajian (commented):
Other than the comment, the rest look great, thanks :)

mailpoet/tests/performance/config.js
veljkho: Hi @samnajian , thanks for reviewing. I understand the point, but the password should always be `password` (maybe in the future if we will have some external site with some password). In one of the next PRs I will add data to a local site so we can test against local always.

samnajian (dismissed):
LGTM! nice addition, thanks.

lysyjan (approved):

PR #4751, opened 2023-02-23
Improve messages about pending approval review [MAILPOET-4528]

lysyjan:
## Description

_N/A_

## Code review notes

I added a new parameter for the messages component if the success class can be used. Maybe there is a better approach, but I did not realize it. 

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4528]

## After-merge notes

_N/A_

[MAILPOET-4528]: https://mailpoet.atlassian.net/browse/MAILPOET-4528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/common/premium_key/key_messages/key_messages.tsx
+++ b/mailpoet/assets/js/src/common/premium_key/key_messages/key_messages.tsx
@@ -6,7 +6,7 @@ type KeyValidMessageProps = { canUseSuccessClass: boolean };
 function KeyValidMessage({ canUseSuccessClass }: KeyValidMessageProps) {
   return (
     <div
-      className={classnames('mailpoet_success_item mailpoet_success_item', {
+      className={classnames('mailpoet_success_item', {
         mailpoet_success: canUseSuccessClass,
       })}
     >
--- a/mailpoet/assets/js/src/common/premium_key/key_messages/key_messages.tsx
+++ b/mailpoet/assets/js/src/common/premium_key/key_messages/key_messages.tsx
@@ -1,9 +1,15 @@
+import classnames from 'classnames';
 import { MailPoet } from 'mailpoet';
 import { useSelector } from 'settings/store/hooks';
 
-function KeyValidMessage() {
+type KeyValidMessageProps = { canUseSuccessClass: boolean };
+function KeyValidMessage({ canUseSuccessClass }: KeyValidMessageProps) {
   return (
-    <div className="mailpoet_success_item mailpoet_success_item mailpoet_success">
+    <div
+      className={classnames('mailpoet_success_item mailpoet_success_item', {
+        mailpoet_success: canUseSuccessClass,
+      })}
+    >
       {MailPoet.I18n.t('premiumTabKeyValidMessage')}
     </div>
   );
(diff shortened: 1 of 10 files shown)


mailpoet/assets/js/src/common/premium_key/key_messages/key_messages.tsx
```diff
@@ -1,9 +1,15 @@
...
 import { MailPoet } from 'mailpoet';
 import { useSelector } from 'settings/store/hooks';
 
-function KeyValidMessage() {
+type KeyValidMessageProps = { canUseSuccessClass: boolean };
+function KeyValidMessage({ canUseSuccessClass }: KeyValidMessageProps) {
   return (
-    <div className="mailpoet_success_item mailpoet_success_item mailpoet_success">
+    <div
+      className={classnames('mailpoet_success_item mailpoet_success_item', {
```
johnolek: I'm pretty sure there's no need to repeat this class name and that it was accidentally added in https://github.com/mailpoet/mailpoet/commit/77f800c799198fb95cf8994902ba342ad40d67e1#diff-85288b20ad01b082b87d06270e63545393c135cd1abd37e4f0ae4fbcde4e90fbR6. Even though you didn't introduce it in this PR could we remove it assuming it's not necessary?

mailpoet/lib/Util/Notices/PendingApprovalNotice.php
```diff
@@ -0,0 +1,78 @@
...
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Mailer\Mailer;
+use MailPoet\Settings\SettingsController;
+use MailPoet\Util\Helpers;
+use MailPoet\WP\Notice as WPNotice;
+
+class PendingApprovalNotice {
+
+  const OPTION_NAME = 'dismissed-pending-approval-notice';
```
johnolek: I'm a little confused about the naming here. This is called an option name, and it has the word `dismissed` in it, but it's not used as an option and it doesn't seem to be dismissable like the others listed in `ajaxDismissNoticeHandler`.

mailpoet/lib/Util/Notices/PendingApprovalNotice.php
```diff
@@ -0,0 +1,78 @@
...
+    $message = Helpers::replaceLinkTags(
+      $message,
+      'https://www.mailpoet.com/support/',
+      [
+        'target' => '_blank',
+      ],
+      'link4'
+    );
+
+    WPNotice::displayError($message, '', self::OPTION_NAME, true, false);
```
johnolek: This is more of a question about the original specs. Do we really want this to be an error? It feels like more of a warning to me. There's not necessarily anything for the user to do right away (unless it's been more than 48 hours), and it's a normal part of the process, not something that's gone wrong.

johnolek (dismissed):
This looks good to me! I had a few minor questions, but as-is this definitely meets the specs so I'll leave it up to you if you want to make any tweaks.

mailpoet/lib/Util/Notices/PendingApprovalNotice.php
lysyjan: @johnolek Thank you for this remark. I was thinking about it too, and ultimately, I followed the image from the specification.
@NeosinneR Do you agree with it, or should we use a warning that contains yellow color instead of the red one?

NeosinneR: @johnolek @lysyjan Good call. Let's go with a warning.

lysyjan:
@johnolek Hey! I have added new commits addressing your remarks! Please take a look. Thank you.

johnolek (approved):
Thank you for the tweaks, @lysyjan!

Aschepikov (approved):
GJ!

PR #4752, opened 2023-02-23
Update shop links on the Upgrade page [MAILPOET-5081]

costasovo:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

Note: While working on the PR, I noticed that the logic for displaying the upgrade page in the sidebar menu is dependent if the premium plugin is installed. The starter plan doesn't have access to the premium plugin, so it makes sense and I kept it that way. I added the note because you may have the premium plugin installed e.g., from the previous testing.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5081]

## After-merge notes

_N/A_

[MAILPOET-5081]: https://mailpoet.atlassian.net/browse/MAILPOET-5081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/views/upgrade.html
+++ b/mailpoet/views/upgrade.html
@@ -16,9 +16,9 @@
         <a
           target="_blank"
           <% if (has_valid_api_key) %>
-            href="https://account.mailpoet.com/upgrade?utm_source=plugin&utm_medium=premium&utm_campaign=upgrade"
+            href="<%= 'https://account.mailpoet.com/orders/upgrade/' ~ plugin_partial_key ~ '?utm_source=plugin&utm_medium=premium&utm_campaign=upgrade' | escape('html_attr') %>"
           <% else %>
-            href="<%= "https://account.mailpoet.com/?s=" ~ subscriber_count ~ "&g=business&billing=monthly&email=" ~ current_wp_user.user_email ~ '&utm_source=plugin&utm_medium=premium&utm_campaign=purchase' | escape('html_attr') %>"
+            href="<%= "https://account.mailpoet.com/?s=" ~ subscriber_count ~ "&g=business&billing=monthly&email=" ~ current_wp_user.user_email ~ '&utm_source=plugin&utm_medium=premium&utm_campaign=purchase' |  escape('html_attr') %>"
           <% endif %>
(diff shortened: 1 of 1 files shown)


lysyjan (approved):
LGTM

Aschepikov (commented):
I have a question about links. They are working well but all the same way, the link get started under the business plan and get started under the creator plan,
image
 and buy now under the agency 
image
 in my case are redirecting me to the page with upgrading my creator plan 
image
, it is a little bit confusing if I am choosing to buy now under the business plan but going to the creators upgrading, but in this case, I still can select the business tab if I want, so it is not critical. Logically upgrade link should redirect me to the upgrade plan, the get started link under the business plan should redirect me to the upgrade my plan to business and so on
image
Should we change it or is the same link for all buttons ok?

costasovo:
@Aschepikov The change is that once a user enters a valid API key all links change to:  https://account.mailpoet.com/upgrade/{hash}. The logic in selecting the correct plan or group is in the shop. But the shop should pick the upgrade group based on the currently active plan. So if the currently active plan in the plugin is business the link should always go to business.

Aschepikov:
image
Usually, there are links to the different plans if I still don't have one.

Aschepikov:
Ok, if the shop should pick the upgrade group based on the currently active plan everything works ok, gj!

Aschepikov (approved):

costasovo:
@Aschepikov But I see your point! I checked in the shop and it doesn't support any parameter on the upgrade page that would allow overriding the group.
@NeosinneR What do you think about adding support for a parameter in the upgrade link so that it can preselect the product group no matter what is the group of the currently active plan?

NeosinneR:
@Aschepikov @costasovo good thinking. I plan to post a Creative Studio request to overhaul this page because it's a wall of text, and I doubt people are reading this. So I'd like to make it way more sellable. Depending on whether we'll also keep the Creator plan here or focus on pushing just the Business plan, we can add the support to the shop.

PR #4755, opened 2023-02-24
Remove Sending factory usage from tests [MAILPOET-4369]

lysyjan:
## Description

_N/A_

## Code review notes

This PR should remove usages `Sending::createFromScheduledTask()` and `Sending::createFromQueue()` from integration tests. Unfortunately, I didn't remove them from `AbandonedCartContentTest` because it would need to change the implementation in the class `AbandonedCartContent`, which is not a part of this task.

## QA notes

I guess we can skip QA here because those changes are only integration tests.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4369]

## After-merge notes

_N/A_

[MAILPOET-4369]: https://mailpoet.atlassian.net/browse/MAILPOET-4369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Entities/ScheduledTaskEntity.php
+++ b/mailpoet/lib/Entities/ScheduledTaskEntity.php
@@ -10,6 +10,7 @@ use MailPoet\Doctrine\EntityTraits\SafeToOneAssociationLoadTrait;
 use MailPoet\Doctrine\EntityTraits\UpdatedAtTrait;
 use MailPoetVendor\Doctrine\Common\Collections\ArrayCollection;
 use MailPoetVendor\Doctrine\Common\Collections\Collection;
+use MailPoetVendor\Doctrine\Common\Collections\Criteria;
 use MailPoetVendor\Doctrine\ORM\Mapping as ORM;
 
 /**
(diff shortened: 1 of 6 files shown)


mailpoet/lib/Entities/ScheduledTaskEntity.php
```diff
@@ -211,6 +212,19 @@ public function getSubscribers(): Collection {
...
+   * @param int $processed ScheduledTaskSubscriberEntity::PROCESSED_* constant
+   * @return SubscriberEntity[]
+   */
+  public function getSubscribersByProcessed(int $processed): array {
+    $criteria = Criteria::create()
+      ->where(Criteria::expr()->eq('processed', $processed));
+    $subscribers = $this->subscribers->matching($criteria)->map(function (ScheduledTaskSubscriberEntity $taskSubscriber): ?SubscriberEntity {
+      return $taskSubscriber->getSubscriber();
+    });
+    return array_filter($subscribers->toArray());
```
websupporter, line 225: Why `array_filter()`?

websupporter:
Thank you @lysyjan.

> I didn't remove them from AbandonedCartContentTest because it would need to change the implementation in the class AbandonedCartContent, which is not a part of this task.

Is this covered by another issue? Did a quick search, but didn't find it.

Over all, lgtm to me! Just one small question. Thank you very much.

mailpoet/lib/Entities/ScheduledTaskEntity.php
lysyjan: Because `ScheduledTaskSubscriberEntity::getSubscriber()` can return `null` or `SubscriberEntity` and PHPStan reports an error here.

lysyjan:
@websupporter We have [a ticket](https://mailpoet.atlassian.net/browse/MAILPOET-4375) that should cover removing all usages of Sending class. I expect that we will need more minor follow-up tickets here, but they will probably be created during the implementation of the bigger one. Does it make sense to you?

mailpoet/lib/Entities/ScheduledTaskEntity.php
websupporter: gotcha.

websupporter (dismissed):
Yes, it does. Thank you. Can you quickly resolve the conflict. Doesnt seem too complicated. Afterwards you can merge imo.

lysyjan:
@websupporter I have rebased it and solved conflicts.

websupporter (approved):
lgtm

websupporter:
Thank you @lysyjan

PR #4756, opened 2023-02-24
Add check that function error_log exists [MAILPOET-5026]

lysyjan:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5026]

## After-merge notes

_N/A_

[MAILPOET-5026]: https://mailpoet.atlassian.net/browse/MAILPOET-5026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/JSON/API.php
+++ b/mailpoet/lib/API/JSON/API.php
@@ -281,7 +281,9 @@ class API {
 
   private function logError(Throwable $e): void {
     // logging to the php log
-    error_log((string)$e); // phpcs:ignore Squiz.PHP.DiscouragedFunctions
+    if (function_exists('error_log')) {
+      error_log((string)$e); // phpcs:ignore Squiz.PHP.DiscouragedFunctions
+    }
     // logging to the MailPoet table
     $this->loggerFactory->getLogger(LoggerFactory::TOPIC_API)->warning($e->getMessage(), [
       'requestMethod' => $this->requestMethod,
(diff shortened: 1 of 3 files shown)


samnajian (approved):
LGTM

veljkho (approved):

JanJakes:
Oh, wow, I'll never stop being surprised by all the cases that we need to handle 🙀

Do you, @lysyjan and @samnajian think, that to avoid this in the future, it might be better to "polyfill" such functions when bootstrapping the plugin? If we do something like this, then we will never forget to use an IF:

```php
// a dummy polyfill that does nothing
if (!function_exists('error_log')) {
  function error_log(string $error) {}
}
```

samnajian:
@JanJakes I agree that polyfilling is probbaly a better approach, thanks for the idea, @lysyjan we can add it as an improvement, what do you think?

JanJakes:
If you guys agree with this, we could create a ticket for that, and have a look, if we use IFs for any other functions that tend to be disabled.

lysyjan:
@JanJakes @samnajian I like this proposal!

JanJakes:
https://mailpoet.atlassian.net/browse/MAILPOET-5108

PR #4757, opened 2023-02-27
Fix InactiveSubscribers cron worker [MAILPOET-5084]

lysyjan:
## Description

_N/A_

## Code review notes

Doctrine returns a string from the query when `getSingleScalarResult ()` is used, and usage of [is_int](https://www.php.net/manual/en/function.is-int.php) always returned `false` on a string. This caused our method in the repository always returned zero.
This behavior was the reason why the InactiveSubscribers worker was marked as `completed` during the first iteration.
## QA notes

1. More than 1000 subscribers are needed
2. Change the value of settings `Stop sending to inactive subscribers` (Settings > Advanced). This schedule the cron worker immediately.
3. Verify that the last worker-run with the type `inactive_subscribers` contains a different value from `{"last_subscriber_id":1000}` in the meta column. This can be done in MySQL(Adminer etc.).

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5084]

## After-merge notes

_N/A_

[MAILPOET-5084]: https://mailpoet.atlassian.net/browse/MAILPOET-5084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscribers/SubscribersRepository.php
+++ b/mailpoet/lib/Subscribers/SubscribersRepository.php
@@ -372,7 +372,7 @@ class SubscribersRepository extends Repository {
       ->getQuery()
       ->getSingleScalarResult();
 
-    return is_int($maxSubscriberId) ? $maxSubscriberId : 0;
+    return intval($maxSubscriberId);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


triple0t (approved):
LGTM

Aschepikov (approved):
GJ!

PR #4767, opened 2023-03-02
Add a new k6 test: Newsletter searching and filtering by a list

veljkho:
## Description

I will add 2 tests in a single PR because they are very related and to speed up the test execution.

## Code review notes

If you'd like to test it locally, this is the command:

cd to `/mailpoet/` first
- `./do test:performance --url=https://qawp.net tests/performance/tests/newsletter-searching.js`

add `--head` to the end of the command above if you want to see it visually and not in headless mode.

In case it fails at first try, run it once again. Logging to WP admin may not work 100%, or the server where qawp.net is hosted can show 502 error sometimes.

## Linked tickets

[MAILPOET-4952] & [MAILPOET-4953]

[MAILPOET-4952]: https://mailpoet.atlassian.net/browse/MAILPOET-4952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4953]: https://mailpoet.atlassian.net/browse/MAILPOET-4953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/tests/newsletter-searching.js
+++ b/mailpoet/tests/performance/tests/newsletter-searching.js
@@ -1,0 +1,69 @@
+/* eslint-disable import/no-unresolved */
+/* eslint-disable import/no-default-export */
+/**
+ * External dependencies
+ */
+import { sleep, check, group } from 'k6';
+import { chromium } from 'k6/x/browser';
+import { randomIntBetween } from 'https://jslib.k6.io/k6-utils/1.1.0/index.js';
+
+/**
+ * Internal dependencies
+ */
+import {
+  baseURL,
+  thinkTimeMin,
+  thinkTimeMax,
+  headlessSet,
+  timeoutSet,
+} from '../config.js';
+import { authenticate } from '../utils/helpers.js';
+/* global Promise */
+
+export function newsletterSearching() {
+  const browser = chromium.launch({
+    headless: headlessSet,
+    timeout: timeoutSet,
+  });
+  const page = browser.newPage();
+
+  group('Emails - Search for a newsletter', () => {
+    page
+      .goto(`${baseURL}/wp-admin/admin.php?page=mailpoet-newsletters`, {
+        waitUntil: 'networkidle',
+      })
+
+      .then(() => {
+        authenticate(page);
+      })
+
+      .then(() => {
+        return Promise.all([
+          page.waitForNavigation({ waitUntil: 'networkidle' }),
+        ]);
+      })
+
+      .then(() => {
+        page.locator('#search_input').type('Newsletter 1st', { delay: 50 });
+        page.waitForSelector('.mailpoet-listing-no-items');
+        page.waitForSelector('[data-automation-id="listing_filter_segment"]');
+        page.waitForLoadState('networkidle');
+        check(page, {
+          'newsletter is found': page
+            .locator('.mailpoet-listing-title')
+            .innerText('Newsletter 1st'),
+        });
+      })
+
+      .finally(() => {
+        page.close();
+        browser.close();
+      });
+  });
+
+  sleep(randomIntBetween(thinkTimeMin, thinkTimeMax));
+}
+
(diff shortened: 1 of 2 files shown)


mailpoet/tests/performance/tests/newsletter-searching.js
```diff
@@ -0,0 +1,83 @@
...
+} from '../config.js';
+import { authenticate } from '../utils/helpers.js';
+/* global Promise */
+
+export function newsletterSearching() {
+  const browser = chromium.launch({
+    headless: headlessSet,
+    timeout: timeoutSet,
+  });
+  const page = browser.newPage();
```
brezocordero, line 28:
```
  const browser = chromium.launch({
    headless: headlessSet,
    timeout: timeoutSet,
  });
  const page = browser.newPage();
```
One suggestion would be to extract this to a helper function that returns the browser and page, I noticed it is the same for all tests.
Maybe include also the authentication.

mailpoet/tests/performance/tests/newsletter-searching.js
```diff
@@ -0,0 +1,83 @@
...
+
+      .then(() => {
+        authenticate(page);
+      })
+
+      .then(() => {
+        return Promise.all([
+          page.waitForNavigation({ waitUntil: 'networkidle' }),
+        ]);
+      })
```
brezocordero, line 44:
```suggestion
      .then(() => page.waitForNavigation({ waitUntil: 'networkidle' }))
```
I am not familiar with k6 but would it work to remove the Promise.all as there is only one?

mailpoet/tests/performance/tests/newsletter-searching.js
```diff
@@ -0,0 +1,83 @@
...
+        });
+      })
+
+      .finally(() => {
+        page.close();
+        browser.close();
+      });
+  });
+
+  sleep(randomIntBetween(thinkTimeMin, thinkTimeMax));
```
brezocordero, line 78: One more suggestion:
This is only one line but as it is used in all tests, you could also extract it to a helper function.

brezocordero (commented):
@veljkho The test works, great work on setting this up!
I have some small suggestions for the code but nothing blocking, let me know what you think.

I did get an error when running the test, not sure why, the test did work.

```
Error: Strange wp-config.php file: wp-settings.php is not loaded directly.
Error: Strange wp-config.php file: wp-settings.php is not loaded directly.
```

mailpoet/tests/performance/tests/newsletter-searching.js
veljkho: I see this as a widely accepted behavior to leave in every test and not in a helper.js, I don't feel necessary to be in the helper. If you have some example how it would be called in the helper, please let me know.

mailpoet/tests/performance/tests/newsletter-searching.js
veljkho: If I leave it without `Promise.all` the test will brake at that step so the only solution I have currently working 100% is to leave inside `Promise.all`. 
However, I am working on a different way to authenticate which will be without logging to WP Admin, so this will be removed along with `authenticate` method.

mailpoet/tests/performance/tests/newsletter-searching.js
veljkho: The same as with `browser` and `page`. Since this is a small amount of code in one line, I would keep it without helper.
However, if I have more `group` options in a single test, I need to divide the groups with such `sleep`. It makes the test more clear with it also. If you agree, I would leave it as-is.

brezocordero (approved):
LGTM!

PR #4770, opened 2023-03-03
Homepage - Remove feature switch [MAILPOET-4831]

costasovo:
## Description

Draft because it waits for https://github.com/mailpoet/mailpoet/pull/4701 to get merged.

This PR:
1) Removes the feature switch and makes the homepage available as the default main page in the plugin.
2) Adds migrations for older users (set states for the task list and the engagement section)
3) Add tracking of clicks in the tasks list

## Code review notes

_N/A_

## QA notes
Please test that the behavior for existing users is correct. See details in [MAILPOET-4831]

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4831]

## After-merge notes

_N/A_

[MAILPOET-4831]: https://mailpoet.atlassian.net/browse/MAILPOET-4831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4831]: https://mailpoet.atlassian.net/browse/MAILPOET-4831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/homepage/components/discovery-task.tsx
+++ b/mailpoet/assets/js/src/homepage/components/discovery-task.tsx
@@ -1,7 +1,7 @@
 import { Icon } from '@wordpress/components';
 import { check } from '@wordpress/icons';
-import { MailPoet } from 'mailpoet';
 import classnames from 'classnames';
+import { trackCtaAndRedirect } from 'homepage/tracking';
 
 type Props = {
   title: string;
@@ -23,16 +23,7 @@ export function DiscoveryTask({
   isDone,
 }: Props): JSX.Element {
   const handleTaskClick = () => {
-    MailPoet.trackEvent(
-      'Home Page Task',
-      {
-        ctaLabel: slug,
-      },
-      { send_immediately: true },
-      () => {
-        window.location.href = link;
-      },
-    );
+    trackCtaAndRedirect('Home Page Task', slug, link);
   };
   return (
     <li
--- a/mailpoet/assets/js/src/homepage/components/discovery-task.tsx
+++ b/mailpoet/assets/js/src/homepage/components/discovery-task.tsx
@@ -5,6 +5,7 @@ import classnames from 'classnames';
 
 type Props = {
   title: string;
+  slug: string;
   link: string;
   imgSrc: string;
   isDone: boolean;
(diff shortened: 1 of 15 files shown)


samnajian (approved):
LGTM!

Aschepikov:
Hi @costasovo 
The product discovery section is visible on an old site with all tasks completed. 
image
In the ticket, I see that if the all tasks are finished we should hide them. 
image

costasovo:
> In the ticket, I see that if the all tasks are finished we should hide them.

@Aschepikov Can you please share how you tested it? I'm asking because the "hiding" of the product discovery section happens in migration, and because of that, there might be some cases when testing feature branch build zip that may cause the migration didn't trigger.

1) In case the site was once updated to the new version and then downgraded and then updated again it will cause the migration will not run again, and the section will remain visible. So for testing, you need a fresh site that never had this build installed before.
2) Or maybe the migration just didn't run because the feature branch build has the same version as the current production build. You can verify it by looking into `wp_mailpoet_migrations` table and checking that there is a row with the name `Migration_20230131_121621` and no errors. If the row is missing, the migration didn't run. You can manually trigger migrations by deactivating and reactivating the plugin.

Could you please check if 1) or 2) didn't cause that the section remained visible during testing?

Aschepikov:
@costasovo No it was the first install and I see this migration in my db. It's hard to say why it was so because I can't repeat it without reinstalling, I will try to find another old site and check it
image

Aschepikov:
@costasovo Yep I checked it on another very old site and I don't see this task list, GJ! But then I installed woocommerce on this site and I am not sure, should this task list be visible in this case or not?

Aschepikov:
On one side we have two additional tasks that we need to complete for a better understanding of this integration with woocommerce, on the other side user is not a beginner. What do you think @costasovo

Aschepikov (approved):

Aschepikov:
https://a8c.slack.com/archives/C01H71ZLBGQ/p1678712499572369 discussion here

PR #4780, opened 2023-03-08
Enforce minimum WooCommerce version

pavel-mailpoet:
## Description

After an issue on support where a user had an outdated WooCommerce version, a fatal PHP error on the site led to the site crashing. Now we will not allow installing MailPoet if WC is ancient (< 6.4), and we will display a warning for outdated version (< 7.2)

## Linked tickets

[MAILPOET-5085]

[MAILPOET-5085]: https://mailpoet.atlassian.net/browse/MAILPOET-5085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Util/Notices/WooCommerceVersionWarning.php
+++ b/mailpoet/lib/Util/Notices/WooCommerceVersionWarning.php
@@ -1,0 +1,64 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Config\Env;
+use MailPoet\Util\Helpers;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class WooCommerceVersionWarning {
+  const OPTION_NAME = 'dismissed-woo-version-outdated-notice';
+  const DISMISS_NOTICE_TIMEOUT_SECONDS = 2592000; // 30 days
+
+  /** @var WPFunctions */
+  private $wp;
+
+  public function __construct(
+    WPFunctions $wp
+  ) {
+    $this->wp = $wp;
+  }
+
+  public function init($shouldDisplay) {
+    if (!is_plugin_active('woocommerce/woocommerce.php')) {
+      return;
+    }
+    $woocommerceVersion = get_plugin_data(WP_PLUGIN_DIR . '/woocommerce/woocommerce.php')['Version'];
+    $requiredWooCommerceVersion = $this->getRequiredWooCommerceVersion();
+    if ($shouldDisplay && $this->isOutdatedWooCommerceVersion($woocommerceVersion, $requiredWooCommerceVersion)) {
+      $this->display($requiredWooCommerceVersion);
+    }
+  }
+
+  public function isOutdatedWooCommerceVersion($woocommerceVersion, $requiredWooCommerceVersion): bool {
+    return version_compare($woocommerceVersion, $requiredWooCommerceVersion, '<') && !get_transient(self::OPTION_NAME);
+  }
+
+  private function display($requiredWooCommerceVersion) {
+    // translators: %s is the PHP version
+    $errorString = __('MailPoet plugin requires WooCommerce version %s or newer. Please update your WooCommerce plugin version, or read our [link]instructions[/link] for additional options on how to resolve this issue. ', 'mailpoet');
+    $errorString = sprintf($errorString, $requiredWooCommerceVersion);
+    $error = Helpers::replaceLinkTags($errorString, 'https://kb.mailpoet.com/article/152-minimum-requirements-for-mailpoet-3#woocommerce-version', [
+      'target' => '_blank',
+    ]);
+
+    $extraClasses = 'mailpoet-dismissible-notice is-dismissible';
+
+    Notice::displayWarning($error, $extraClasses, self::OPTION_NAME);
+  }
+
+  public function disable() {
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Util/Notices/WooCommerceVersionWarning.php
```diff
@@ -0,0 +1,64 @@
...
+    WPFunctions $wp
+  ) {
+    $this->wp = $wp;
+  }
+
+  public function init($shouldDisplay) {
+    if (!is_plugin_active('woocommerce/woocommerce.php')) {
+      return;
+    }
+    $woocommerceVersion = get_plugin_data(WP_PLUGIN_DIR . '/woocommerce/woocommerce.php')['Version'];
```
costasovo: I think it would be better to add  the`get_plugin_data` to WPFunctions. So that we are consistent.

mailpoet/mailpoet.php
```diff
@@ -52,6 +53,17 @@ function mailpoet_deactivate_plugin() {
   return;
 }
 
+// Check for minimum supported WooCommerce version
+if (function_exists('is_plugin_active') && is_plugin_active('woocommerce/woocommerce.php')) {
+  $woocommerceVersion = get_plugin_data(WP_PLUGIN_DIR . '/woocommerce/woocommerce.php')['Version'];
```
costasovo, line 61: I have a slight concern about the performance impact of this. This requires parsing an initial 8KB of the PHP file on every request. Unfortunately, WC_VERSION is not available at this point so I don't know any better alternative.
Does the get_plugin_data function cover file_exists check? I'm asking in case someone uses a different directory for Woo.

mailpoet/lib/Util/Notices/WooCommerceVersionWarning.php
```diff
@@ -0,0 +1,64 @@
...
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Config\Env;
+use MailPoet\Util\Helpers;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class WooCommerceVersionWarning {
+  const OPTION_NAME = 'dismissed-woo-version-outdated-notice';
+  const DISMISS_NOTICE_TIMEOUT_SECONDS = 2592000; // 30 days
```
costasovo, line 12: Using days feels a bit suboptimal here. If we increase the minimal version users may see the warning even 30 days after that. What about adding (only or additionally) the woo version into the transient key? e.g `dismissed-woo-version-outdated-notice-7.4`

mailpoet/mailpoet.php
```diff
@@ -28,6 +28,7 @@
 ];
 
 const MAILPOET_MINIMUM_REQUIRED_WP_VERSION = '5.9';
+const MAILPOET_MINIMUM_REQUIRED_WOOCOMMERCE_VERSION = '6.4';// Older versions lead to fatal errors
```
costasovo, line 31: What will be the mechanism for updating this version?
I suggest unifying it with the version we use for [the oldest_acceptance tests](https://github.com/mailpoet/mailpoet/blob/trunk/.circleci/config.yml#L838), and then we could update both regularly. Or do we want to keep these two separately?

mailpoet/lib/Util/Notices/WooCommerceVersionWarning.php
costasovo: Also, it would be helpful to prefix the transient key with `mailpoet-` so that it is easy to filter it in the DB in case of debugging.

costasovo (commented):
Thanks for adding this check. I added a few questions and suggestions.

mailpoet/mailpoet.php
pavel-mailpoet: It is a different version, in the oldest acceptance tests. Here we enforce 6.4, but in acceptance tests, we have 6.8.

mailpoet/mailpoet.php
pavel-mailpoet: > I'm asking in case someone uses a different directory for Woo

I don't think you can do that. That would fail even the test above and it wouldn't get to this line.

> I have a slight concern about the performance impact of this.

Hmm, that could be an issue, because I can't find any other way to get the version.

pavel-mailpoet: I measured the impact of this call: https://d.pr/i/K4RKDz

costasovo (dismissed):
LGTM! Thanks for including the changes and checking the performance impact of get_plugin_version!

Aschepikov:
Hi, @pavel-mailpoet I have an issue with WooCommerce versions < 6.4.
If I will install wc < 6.4. and then the zip plugin from this pr or if I will install mailpoet from here and then wc < 6.4 It will cause the fatal error because mailpoet will be activated 
image
image
image

pavel-mailpoet:
@costasovo Can you please re-check this pull request? I discovered that the functions `is_plugin_active` and `get_plugin_data` may not be present in some circumstances. So I ensured they are [loaded](https://wordpress.stackexchange.com/questions/15310/if-plugin-is-active-check-if-plugin-is-enabled-or-not).

costasovo (commented):
LGTM! 👍

Aschepikov (approved):
GJ!

PR #4785, opened 2023-03-09
Ensure users know they're deactivating email [MAILPOET-4397]

johnolek:
## Description

This PR aims to decrease the likelihood that users will accidentally deactivate an automated email without realizing it.

- The confirmation shown on the listing page is now triggered when clicking on the subject line of the email in addition to when a user clicks "Edit"
- There is a confirmation when entering the editor itself, which uses the same string as the confirmation modal. If the user hits cancel, they're returned to the listing page for the type of email they were trying to edit. This is useful for the following situations:
  - A user visits the editor from a bookmark
  - A user opens the editor in a new tab
  -  A user activates an inactive email using the toggle on the listing page and then clicks either the edit button or the subject line to enter the editor
- The notice stating that the email was deactivated now uses the warning style instead of success, and it does not dismiss itself automatically.   

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4397](https://mailpoet.atlassian.net/browse/MAILPOET-4397)

## After-merge notes

_N/A_

[MAILPOET-4397]: https://mailpoet.atlassian.net/browse/MAILPOET-4397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/newsletter_editor/initializer.jsx
+++ b/mailpoet/assets/js/src/newsletter_editor/initializer.jsx
@@ -91,6 +91,16 @@ const initializeEditor = (config) => {
         newsletter.queue &&
         newsletter.queue.status === null
       ) {
+        // Users can confirm before visiting the page
+        let pauseConfirmed = getUrlParam('pauseConfirmed') === 'yes';
+        if (!pauseConfirmed) {
+          // eslint-disable-next-line no-alert
+          pauseConfirmed = window.confirm(MailPoet.I18n.t('confirmEdit'));
+        }
+        if (!pauseConfirmed) {
+          window.location = `admin.php?page=mailpoet-newsletters#/${newsletter.type}`;
+          return;
+        }
         MailPoet.Ajax.post({
           api_version: window.mailpoet_api_version,
           endpoint: 'sending_queue',
@@ -116,7 +126,7 @@ const initializeEditor = (config) => {
         newsletterTypesWithActivation.includes(newsletter.type) &&
         newsletter.status === 'active'
       ) {
-        // Users can confirm deactivation before visiting the page
+        // Users can confirm before visiting the page
         let deactivationConfirmed =
           getUrlParam('deactivationConfirmed') === 'yes';
         if (!deactivationConfirmed) {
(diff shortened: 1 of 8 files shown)


mailpoet/assets/js/src/newsletter_editor/initializer.jsx
```diff
@@ -138,6 +113,19 @@ const initializeEditor = (config) => {
         newsletterTypesWithActivation.includes(newsletter.type) &&
         newsletter.status === 'active'
       ) {
+        // Users can confirm deactivation before visiting the page
```
brezocordero: Should we add the check [before pausing newsletters ](https://github.com/mailpoet/mailpoet/pull/4785/files#diff-4e317ae921510cd1cac6b3bf3f58e7b6cc9cc5af674f9e752e56c720173a73faL118)that are sending?

brezocordero (commented):
@johnolek Nice work!
I have looked a the code and tested the different cases. It all works as expected.
I like how you cover all the different cases.
I only have one question, it is a bit of an edge case (newsletter is sending and they don't trigger the confirmation when clicking the title or edit link). Let me know what you think.

mailpoet/assets/js/src/newsletter_editor/initializer.jsx
johnolek: Great point, @brezocordero! I do think it's worth it to cover that edge case, and I've implemented it in fb00cf094fdf788fb663791ecb0c6b7b4b6991e6. I also changed the success notice to a system (i.e. warning) so it won't auto-dismiss. Let me know if it looks ok or if you think of anything else that could improve this PR.

brezocordero (dismissed):
Looks great! 
Thanks for making the changes.

Aschepikov (dismissed):
GJ!

Aschepikov:
@johnolek  HI, there are some conflicts, can you resolve them pls

Aschepikov (approved):
GJ!

PR #4791, opened 2023-03-14
Add new k6 test Complex Segment

veljkho:
## Description

In this PR, I'm going to add a new k6 test, but also to optimize the existing scenario and tests by adding `await` and moving `sleep` into `finally`.
Added a new helper function `selectInReact` and updated existing `authentication` method.

Now the tests are run sequentially when executing scenarios, but also way more stable.

## Code review notes

Test is not possible to be run locally since it requires Premium plugin active. Would like to keep the test in the base and run it against online site with Premium active on a nightly basis.

## Linked tickets

[MAILPOET-4958]

[MAILPOET-4958]: https://mailpoet.atlassian.net/browse/MAILPOET-4958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/config.js
+++ b/mailpoet/tests/performance/config.js
@@ -14,4 +14,4 @@ export const lastName = 'Doe';
 export const defaultListName = 'Newsletter mailing list';
 
 export const thinkTimeMin = '1';
-export const thinkTimeMax = '4';
+export const thinkTimeMax = '3';
--- a/mailpoet/tests/performance/scenarios.js
+++ b/mailpoet/tests/performance/scenarios.js
@@ -60,20 +60,20 @@ if (scenario) {
 }
 
 // All the tests ran for pull requests
-export function pullRequests() {
-  newsletterListing();
-  subscribersListing();
-  settingsBasic();
-  subscribersFiltering();
-  subscribersAdding();
-  formsAdding();
-  newsletterSearching();
-  newsletterSending();
-  listsViewSubscribers();
+export async function pullRequests() {
+  await newsletterListing();
+  await subscribersListing();
+  await settingsBasic();
+  await subscribersFiltering();
+  await subscribersAdding();
+  await formsAdding();
+  await newsletterSearching();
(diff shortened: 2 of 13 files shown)


triple0t (commented):
It's not getting past here for me.
Screenshot 2023-03-16 at 3 40 44 PM


There were other issues as well.
e.g
It got stuck on the landing page
It got stuck on the welcome wizard

veljkho:
Hi @triple0t 
Sorry about that, our local env is not ready yet.
If you execute the test against online site, using this command:
`./do test:performance --url=https://qawp.net tests/performance/tests/lists-complex-segment.js --head`
It will work. However, you have to change the password in `config.js` to (reveal the password in the secret store, search for `qawp`)

Please let me know if any questions.

triple0t:
> Sorry about that, our local env is not ready yet. If you execute the test against online site, using this command: `./do test:performance --url=https://qawp.net tests/performance/tests/lists-complex-segment.js --head` It will work. However, you have to change the password in `config.js` to (reveal the password in the secret store, search for `qawp`)
> 
> Please let me know if any questions.

It works. Well done.

May I suggest you use `.env` values for the username and password in the `config.js` file

triple0t (dismissed):
LGTM

veljkho:
Thanks! @triple0t  It is question if we could use such a test since this requires a premium plugin. I started a Slack thread recently in #mailpoet-dev. Will try to put inside `.env`.

veljkho:
Hi @triple0t 
Please take another look as I added a new commit to exclude this test from scenario.js execution, since this test requires Premium plugin and probably won't be possible to test locally. I will try to migrate those tests in Premium repo instead, or just find a way to execute locally in the free plugin, that's why I want to keep the test in the base.

triple0t (approved):
Okay 👍

PR #4793, opened 2023-03-14
Prefix Html2Text as we do for other vendors [MAILPOET-5128]

samnajian:
## Description

This PR add Html2Text via prefixer.

## QA notes

Technically we should be fine if automatic tests pass.

## Linked tickets

[MAILPOET-5128]

[MAILPOET-5128]: https://mailpoet.atlassian.net/browse/MAILPOET-5128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -479,7 +479,6 @@ class RoboFile extends \Robo\Tasks {
       'lib-3rd-party/',
       'vendor/composer',
       'vendor/mtdowling',
-      'vendor/soundasleep',
       'vendor-prefixed/',
       'mailpoet.php',
     ]);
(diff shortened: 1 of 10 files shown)


mailpoet/RoboFile.php
```diff
@@ -479,7 +479,6 @@ public function qaLintBuild() {
       'lib-3rd-party/',
       'vendor/composer',
       'vendor/mtdowling',
-      'vendor/soundasleep',
```
rodrigoprimo, line 482: Checking the commit message, it is not clear to me why it was necessary to include `vendor/soundasleep` in the PHP linter check, and now that it was moved to vendor prefixed, it is not necessary anymore. Could you please explain? In my opinion, this is the kind of information that is useful to include in the commit message.

rodrigoprimo (commented):
Looks good to me. I just left one comment with a clarifying question.

mailpoet/RoboFile.php
samnajian: Just for the records: I'm going to add this line back.

rodrigoprimo (approved):

PR #4794, opened 2023-03-15
Add order status change trigger [MAILPOET-4935]

websupporter:
## Description
Adds the order status change trigger.

## Code review notes
This PR introduces subject transformation. Therefore integrations can now register `SubjectTransformer` which `accept()` a certain type of subject and `return()` a certain type using `transform()`.

### The new `SubjectTransformerHandler` can 
1. Generate all possible subjects for an automation. It takes the subjects which are provided by the trigger and searches for possible transformations. This is used to be able to add steps to an automation, which do require different subjects than the trigger provides but are available thanks for transformations. In the ValidStepOrderRule this functionality replaces now the `collectSubjectKeys()` method.
2. Generate a transformation chain, which then transforms a subject A into a subject B. This transformation takes place, when we search for the required subjects in an automation in order to execute a specific step.

Why a chain? Imagine a trigger provides a `subject_a` and we need a `subject_c`, but there is no straight transformation. What we have is an `AtoB` and a `BtoC` transformer. The chain enables us to transform now `subject_a` to `subject_c` by using both transformers in an orderly fashion.

### The new WooCommerce integration
Now we can have a order status change trigger, which does not have a subscriber subject. With the new transformer from order to subscriber, we can transform these subjects when needed.

The review remarks of the [old PR](https://github.com/mailpoet/mailpoet/pull/4748) have been taken care of.

## QA notes

Besides checking the trigger if it works as expected, please check also if the "Run this automation only once per subscriber" works as expected.

## Linked PRs
https://github.com/mailpoet/mailpoet-premium/pull/686

## Linked tickets

 [MAILPOET-4935]

## After-merge notes

Do not merge before transactional emails have been merged. Ping @websupporter when in doubt.

[MAILPOET-4935]: https://mailpoet.atlassian.net/browse/MAILPOET-4935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


websupporter:
Just a quick note as this popped in my mind:

One problem I just see is with this logic:
https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Automation/Integrations/MailPoet/Hooks/CreateAutomationRunHook.php#L42-L47

Run only once relies on a subscriber subject being in the database. The transformation does not persist the subjects.

Even if we _would_ persist it: The checkbox does only appear in conjunction with triggers capable of providing `SubscriberSubject`.

We could either advance both logics: Persist the subjects on run creation and when `$requiredSubject` is missing (plugin with transformer could have been activated later) and extend the list of provided subjects by a trigger depending on the existing transformers.

Another idea would be to provide a "Run only once per customer" checkbox for woocommerce triggers :thinking:

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -0,0 +1,142 @@
...
+        return $step->getType() === StepData::TYPE_TRIGGER;
+      }
+    );
+
+    $triggers = array_filter(array_map(
+      function(StepData $step): ?Step {
+        return $this->registry->getStep($step->getKey());
+      },
+      $configuredTriggers
+    ));
```
JanJakes, line 35: Should the results be wrapped in `array_values` or are we safe here?

I guess it could also be merged to a single `array_filter/array_map`, but not important.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -0,0 +1,142 @@
...
+
+    $triggerKeys = [];
+    foreach ($triggers as $trigger) {
+      $triggerKeys[$trigger->getKey()] = $trigger->getSubjectKeys();
+    }
+    if (!$triggerKeys) {
+      return [];
+    }
+
+    $triggerKeys = count($triggerKeys) === 1 ? current($triggerKeys) : array_intersect(...array_values($triggerKeys));
```
JanJakes: Intersection... is it to take into account only those subject keys that all triggers provide together?

It kind of reminds me of [ValidStepOrderRule](https://github.com/mailpoet/mailpoet/blob/695c387210361d8f40c321fe15a3d76ae7890da2/mailpoet/lib/Automation/Engine/Validation/AutomationRules/ValidStepOrderRule.php#L56), but now I'm thinking — don't I have a "bug" there, doing just `array_unique(array_merge(...))`? 🤔 Certainly something to investigate.

That said, for the transformations, shouldn't we first transform and then look for common subjects (intersection), including the transformed ones? At least for validation, that would make sense to me. At runtime, we'll probably work with only one trigger at a time.

I just started going through this, so maybe I'll understand more later on.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -0,0 +1,142 @@
...
+  public function __construct(
+    Registry $registry,
+    AutomationStorage $automationStorage
+  ) {
+    $this->registry = $registry;
+    $this->automationStorage = $automationStorage;
+  }
+
+  /** @return string[] */
+  public function subjectKeysForAutomation(Automation $automation): array {
```
JanJakes: I think a better naming can be, e.g., `getSubjectKeysForAutomation`.

EDIT: + Similary `getSubjectKeysForTrigger`.

mailpoet/lib/Automation/Engine/Validation/AutomationRules/ValidStepOrderRule.php
```diff
@@ -39,7 +45,7 @@ public function visitNode(Automation $automation, AutomationNode $node): void {
       return;
     }
 
-    $subjectKeys = $this->collectSubjectKeys($automation, $node->getParents());
+    $subjectKeys = $this->subjectTransformerHandler->subjectKeysForAutomation($automation);
```
JanJakes: Ahh, I wrote about this rule above, but I guess you fixed it here 😄

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -0,0 +1,142 @@
...
+        $this->getPossibleTransformations($possibleKey, $stopKey)
+      );
+    }
+    return $possibleKeys;
+  }
+
+  /**
+   * @return SubjectData[]|null
+   */
+  public function transformSubjectData(string $target, AutomationRun $automationRun): ?array {
```
JanJakes: I think `$target` is the subject key so maybe something like `$targetSubjectKey` would be more descriptive.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -0,0 +1,142 @@
...
+    $configuredTriggers = array_filter(
+      $automation->getSteps(),
+      function (StepData $step): bool {
+        return $step->getType() === StepData::TYPE_TRIGGER;
+      }
+    );
+
+    $triggers = array_filter(array_map(
+      function(StepData $step): ?Step {
+        return $this->registry->getStep($step->getKey());
```
JanJakes: Should we filter out the triggers that are not in the registry or throw an error here? Maybe this is OK because on activation the validator would take care of this?

mailpoet/lib/Automation/Integrations/WooCommerce/Subjects/OrderSubject.php
```diff
@@ -0,0 +1,43 @@
...
+
+  public function getArgsSchema(): ObjectSchema {
+    return Builder::object([
+      'order_id' => Builder::integer()->required(),
+    ]);
+  }
+
+  public function getPayload(SubjectData $subjectData): Payload {
+    $id = $subjectData->getArgs()['order_id'];
+    $order = wc_get_order($id);
```
JanJakes: We could probably use the Woo helper...

mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/OrderStatusChangedTrigger.php
```diff
@@ -0,0 +1,102 @@
...
+
+    $this->wp->doAction(Hooks::TRIGGER, $this, [
+      new Subject(OrderStatusChangeSubject::KEY, ['from' => $oldStatus, 'to' => $newStatus]),
+      new Subject(OrderSubject::KEY, ['order_id' => $order->get_id()]),
+      new Subject(CustomerSubject::KEY, ['customer_id' => $order->get_customer_id()]),
+    ]);
+  }
+
+  public function isTriggeredBy(StepRunArgs $args): bool {
+    $orderPayload = $args->getSinglePayloadByClass(OrderStatusChangePayload::class);
```
JanJakes, line 88: The above method will throw an exception if there is no single payload, so the check below is not needed. Unless you want to catch the exception and return `false` but I guess that's not necessary.

mailpoet/lib/Automation/Integrations/MailPoet/SubjectTransformers/OrderSubjectToSubscriberSubjectTransformer.php
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  private function findSubscriber(Subject $order): ?SubscriberEntity {
+    $orderId = $order->getArgs()['order_id'] ?? null;
+    if (!$orderId) {
+      return null;
+    }
+    $wcOrder = $this->woocommerceHelper->wcGetOrder($orderId);
+    $billingEmail = $wcOrder->get_billing_email();
+    return $billingEmail ? $this->subscribersRepository->findOneBy(['email' => $billingEmail]) : $this->subscribersRepository->findOneBy(['wpUserId' => $wcOrder->get_user_id()]);
```
JanJakes: I think I was already asking about this, but, should email or WP user ID have priority?
Also, let's maybe split the ternary or multiple lines, it's a bit long.

mailpoet/lib/Automation/Integrations/MailPoet/SubjectTransformers/OrderSubjectToSubscriberSubjectTransformer.php
```diff
@@ -0,0 +1,80 @@
...
+      return new Subject(SubscriberSubject::KEY, ['subscriber_id' => $subscriber->getId()]);
+  }
+
+  public function accepts(): string {
+      return OrderSubject::KEY;
+  }
+
+  public function returns(): string {
+        return SubscriberSubject::KEY;
+  }
```
JanJakes, line 54: On about 20 lines above, the indents are somewhat strange. Maybe some IDE autoformat or something. Otherwise, this looks great.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -97,7 +97,42 @@ function (SubjectTransformer $transformer) use ($key): bool {
   }
 
   /**
```
JanJakes: I guess from here down it is for runtime data, and maybe it would be nice to extract it to another class, so the distinction is clear. The class would need to reuse the current class to call subjectKeysForAutomation, I guess. What do you think?

It kind of seems to me that what's above this comment doesn't really belong to the Control namespace, as it's static and used also for validation, but I'm not sure where else to put it 😄

JanJakes (commented):
Subject transformers, awesome! 👏

It's quite a complex logic (and I may need another look at `SubjectTransformerHandler`), I left only a few small comments.

JanJakes:
> One problem I just see is with this logic: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Automation/Integrations/MailPoet/Hooks/CreateAutomationRunHook.php#L42-L47

Back to this, is it resolved now? It seems to me that the hook will need to consider transformed subjects as well (which, I guess, we can do now).

JanJakes:
@websupporter One more thing I've been thinking about is the subject transformation API. What do you think about adding a method on the `Subject` interface? E.g., something like:
```php
class SubscriberSubject {
  public function derive(): array {
    return [
      'woocommerce:order' => function(SubjectData $data): SubjectData {
        return new SubjectData(...);
      },
    ];
  }
}
```

I like the simplicity here, but at the same time I'm not 100% sure that it's better.

_Pros:_ Simple, concise, collocated static definition (the subject key) with the transformation (the callback), near all the other subject functionality, similar to integration contexts.

_Cons:_ "Weak" typing (but we can validate at runtime), bloating the Subject interface (does it even belong there?).

I mean, I'm just thinking out loud, not sure what I prefer, in fact. The interfaces are nice in terms of strong typing and single responsibility. Would like to know your thoughts.

JanJakes:
@websupporter Looking a bit deeper at the logic, I feel like `SubjectTransformerHandler` has evolved into something more complex than it needs to be (and I understand why, as the requirements kind of changed). Especially, the transformer chain is probably not needed anymore, given that we'll **always instantiate all subjects** anyway. In such a scenario, the optimal path will emerge naturally, if the transformations proceed a "top-down" path. I'll try to explain with some (pseudo)code.

Let's split it to the two parts (that I suggested to maybe split even in the code):

**1. Static (getting all possible subject keys)**
This seems like it should be doable with a single loop of the following logic — `For each not yet seen subject key, add all subject keys that can be directly derived from it, repeating this until nothing new is added`. If I'm not making a mistake somewhere, that would result in:

```php
  public function subjectKeysForTrigger(Trigger $trigger): array {
    // Map of subject source key to all possible transformers.
    // Maybe this way could be the transformers indexed in the registry itself so we have a helper method for this?
    $transformerMap = [];
    foreach ($this->registry->getSubjectTransformer() as $transformer) {
      $transformerMap[$transformer->accepts()] = array_merge($transformerMap[$transformer->accepts()] ?? [], [$transformer]);
    }

    $all = $trigger->getSubjectKeys();
    $queue = $all;
    while ($key = array_shift($queue)) {
      foreach ($transformerMap[$key] ?? [] as $transformer) {
        $newKey = $transformer->returns();
        if (!in_array($newKey, $all, true)) {
          $all[] = $newKey;
          $queue[] = $newKey;
        }
      }
    }
    return $all;
  }
```

Another method of the "static" nature is `subjectKeysForAutomation`. This one doesn't seem to reuse the method above, although that seems to be the natural thing to do to me. I think it can be as simple as this:

```php
public function subjectKeysForAutomation(Automation $automation): array {
    $triggers = array_filter(
      $automation->getSteps(),
      function (StepData $step): bool {
        return $step->getType() === StepData::TYPE_TRIGGER;
      }
    );

    $keys = [];
    foreach ($triggers as $trigger) {
      $keys = array_merge($keys, $this->getSubjectKeysForTrigger($trigger));
    }
    return array_unique($keys);
  }
```

This seems shorter and simpler than the current `subjectKeysForTrigger` + `getPossibleTransformations` combo (no need for the recursion and the `$stopKey` part).

**2. Dynamic (getting actual subject data)**
Here, I think we only want to `get all the subject keys from point 1. and instantiate them`. That would be one way to look at that. But even simpler, we could say `do exactly the same as in point 1., only instead of the target subject key, get the transformed data`.  So I guess that would be sth like:

```php
  public function provideAllSubjects(array $subjects): array {
    // Map of subject source key to all possible transformers.
    // Maybe this way could be the transformers indexed in the registry itself so we have a helper method for this?
    $transformerMap = [];
    foreach ($this->registry->getSubjectTransformer() as $transformer) {
      $transformerMap[$transformer->accepts()] = array_merge($transformerMap[$transformer->accepts()] ?? [], [$transformer]);
    }

    // in this case, we'll use $all as a map of key to subject
    $all = [];
    foreach ($subjects as $subject) {
      $all[$subject->getKey()] = $subject;
    }

    $queue = array_keys($subjects);
    while ($key = array_shift($queue)) {
      foreach ($transformerMap[$key] ?? [] as $transformer) {
        $newKey = $transformer->returns();
        if (!isset($all[$newKey])) {
          $all[$newKey] = $transformer->transform($all[$key]);
          $queue[] = $newKey;
        }
      }
    }
    return array_values($all);

    // at this point, we may also consider calling "subjectKeysForTrigger" and checking that all subjets were created
  }
```
---
**Unified?**
Now, since `subjectKeysForTrigger` and `provideAllSubjects` are almost the same, we could even consider merging them to a single reusable code:

```php
  public function subjectKeysForTrigger(Trigger $trigger): array {
    $keys = $trigger->getSubjectKeys();
    $initialItems = array_combine($keys, $keys);
    return $this->collectTransformers($initialItems, function (SubjectTransformer $transformer) {
      return $transformer->returns(); // collecting the keys
    });
  }

  public function provideAllSubjects(Trigger $trigger, array $subjects): array {
    $initalItems = [];
    foreach ($subjects as $subject) {
      $initalItems[$subject->getKey()] = $subject;
    }

    $all = $this->collectTransformers($initalItems, function (SubjectTransformer $transformer, Subject $subject) {
      return $transformer->transform($subject); // collecting the transformed subjects
    });

    // we can also check whether we got all the subjects we expected here
    $keys = $this->subjectKeysForTrigger($trigger);
    ...
    return $all;
  }

  /**
   * $initialItems must be $key (subject key) => $value (subject data or nothing)
   * $callback will receive $transformer and the given $value
   */
  private function collectTransformers(array $initialItems, callable $callback): array {
    // ...get the $transformerMap somehow

    $all = $initialItems;
    $queue = array_keys($initialItems);
    while ($key = array_shift($queue)) {
      foreach ($transformerMap[$key] ?? [] as $transformer) {
        $newKey = $transformer->returns();
        if (!isset($all[$newKey])) {
          $all[$newKey] = $callback($transformer, $all[$key]);
          $queue[] = $newKey;
        }
      }
    }
    return array_values($all);
  }
```

Now, the question is whether it is worth it reusing. It certainly is nice to reuse the same loop logic for the two, but it also adds some complexity, which may lower the readability of the code a bit.

Anyway, I think gradually adding subject keys or transforming subjects until you have nothing more to add/transform should be all that's needed. No explicit infinite loop checking and stuff would be required in that case. It seems to cut the length and the complexity of the class quite a bit.

In all of the examples above, I used a `$queue` with `array_shift` and push (`$arr[] = ...`) because it avoids the need of nesting the loops even more. Alternatively, the same should be doable with one more nesting and some `$new` helper array for newly discovered items in each iteration. I think that would only be more complex...

Btw. **I didn't try to run any of the code** I wrote 😀 So yep, there will probably be bugs, but I hope it makes the concepts understandable. If I'm wrong or something, just ping me 😀

---
Finally, I think we could make the naming a bit more unified/consistent:
```bash
subjectKeysForTrigger    -> getSubjectKeysForTrigger
subjectKeysForAutomation -> getSubjectKeysForAutomation
provideAllSubjects       -> getAllSubjects (or maybe sth like transformAllSubjects, expandAllSubject, etc.)
```

mailpoet/lib/Automation/Integrations/MailPoet/SubjectTransformers/OrderSubjectToSubscriberSubjectTransformer.php
websupporter: Imo the billing email.

Afaik a registered user can change the billing email without changing the user email. In such a case, I would expect the billing email to be the one on which we act.

websupporter:
Thank you very much for your review @janjakes!!! :tada:

> I mean, I'm just thinking out loud, not sure what I prefer, in fact. The interfaces are nice in terms of strong typing and single responsibility. Would like to know your thoughts.

I prefer the `SubjectTransformer` for another reason. Vendor A provides `Subject A`, vendor B `Subject B`. No one of them provides a transformation. This transformation could be provided by yet another vendor, e.g. an agency tasked with a specific task.

> SubjectTransformerHandler has evolved into something more complex than it needs to be

Thanks so much for catching that. I followed your example and it works nicely. 

> Now, the question is whether it is worth it reusing. It certainly is nice to reuse the same loop logic for the two, but it also adds some complexity, which may lower the readability of the code a bit.

I would vote for readability in this case.

With regards to `subjectKeysForAutomation()`, I think we need to keep the `array_intersect()` logic. From the naming it might be not super clear, but essentially, when you look where the method is used: It is used in the `ValidStepOrderRule`, where the subjects which an automation provides meet the required subjects of a step. So basically, we need to ensure which ever entry a user takes (what ever trigger he triggers) all required subjects are provided, which we can do with array_intersect. But I agree, we should do first retrieve all possible keys for each trigger.

I rewrote the class here: [89b9030](https://github.com/mailpoet/mailpoet/pull/4794/commits/89b9030c0292102679389e25c8ffe0c33003138d)

mailpoet/lib/Automation/Integrations/WooCommerce/Subjects/OrderSubject.php
websupporter: Done in [4c60a89](https://github.com/mailpoet/mailpoet/pull/4794/commits/4c60a890060ec4bef9d79e9b2982c0e2cc9fdc38)

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: I don't have a strong opinion here. I think its ok as the validator would take care of it.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: Not applicable anymore.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: Thanks! Done in the rewrite.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: Imo save as $triggerData was wrapped in array_values, if I understand you correctly.

mailpoet/lib/Automation/Integrations/MailPoet/SubjectTransformers/OrderSubjectToSubscriberSubjectTransformer.php
websupporter: Ha!? Done in the readability commit https://github.com/mailpoet/mailpoet/pull/4794/commits/e407f43496fe35516258fd6f8411d014201fcca1

mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/OrderStatusChangedTrigger.php
websupporter: [1296e82](https://github.com/mailpoet/mailpoet/pull/4794/commits/1296e826c8167241e1615e88eba9e37d238de9b2)

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: Hmm... Yeah, not really sure. It "transforms" subjects, is it maybe a Builder or Mapper? Not sure if I would want to split the "static" and "dynamic" parts. I see where you are coming from :thinking: At least, I would like them to be in the same namespace. It is quite related and I hate it when I have to search for hours for "where was the other part of that again, was it now a validator or a helper or a mapper...."

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -0,0 +1,103 @@
...
+      $all[] = $this->getSubjectKeysForTrigger($trigger);
+    }
+    $all = count($all) > 1 ? array_intersect(...$all) : $all[0] ?? [];
+    return array_values(array_unique($all));
+  }
+
+  public function getSubjectKeysForTrigger(Trigger $trigger): array {
+    $all = [];
+    foreach ($trigger->getSubjectKeys() as $key) {
+      $all = array_merge($all, $this->getSubjectKeysForSingleKey($key));
```
JanJakes: I think now that you've extracted `getSubjectKeysForSingleKey` to a private method, it can produce duplicates in the `$all` array. We can surely make it unique, but I guess we may as well inline it here, since `getAllSubjects` also doesn't extract it out and it seems just slower to create the transformer map again and again, etc.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return $all;
+  }
+
+  public function getAllSubjects(Subject ...$subjects): array {
```
JanJakes: I noticed usage of this pattern in a few places. What's the reason, is it the typing? The thing is, it's not very consistent with the rest of the codebase. It's also not extensible (adding new params is hard, having multiple arrays this way is impossible), and it doesn't very nicely represent what we're passing in (an array of subjects).

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
JanJakes: The `$triggers` array can become associative due to `array_filter`, but in the current code, it's not a problem. Now I'm actually more thinking about what if a trigger disappears due to some plugin being temporarily deactivated, but I hope that's OK because in the UI, we'd already display an "unknown" trigger, in that case.

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
```diff
@@ -21,181 +19,85 @@ public function __construct(
...
-  /**
-   * @param Trigger $trigger
-   * @param Subject ...$subjects
-   * @return Subject[]
-   */
-  public function provideAllSubjects(Trigger $trigger, Subject ...$subjects): array {
-    $allSubjectsKeys = $this->subjectKeysForTrigger($trigger);
-    $allSubjectKeyTargets = array_filter(
-      array_diff($allSubjectsKeys, array_map(
+    $queue = array_map(
```
JanJakes: I wonder why you didn't keep $queue = array_keys($all), which seems to be much shorter.

mailpoet/lib/Automation/Engine/Registry.php
```diff
@@ -16,6 +17,9 @@ class Registry {
   /** @var array<string, Subject<Payload>> */
   private $subjects = [];
 
+  /** @var SubjectTransformer[] */
+  private $subjectTransformer = [];
```
JanJakes: Shouldn't this be called `$subjectTransformers` (and the getter `getSubjectTransformers`) in accordance with other naming?

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
JanJakes: Right, let's keep it in one now. It's a bit of a special case. We can always split it, if we figure we need it.

JanJakes (dismissed):
Almost there 👏  Added just a few comments.

mailpoet/lib/Automation/Integrations/MailPoet/SubjectTransformers/OrderSubjectToSubscriberSubjectTransformer.php
JanJakes: 👍  It probably makes sense in the context of sending an email after an order.

(I was looking at it like "we know who made the order so why we would potentially instantiate someone else as a subscriber".)

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: Good catch! Thank you.

In an earlier iteration, I used `getSubjectKeysForSingleKey()` also in the `...forAutomation()` method, this is when I created the private method. It survived until 79f5892d8

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: Yes. Big fan. But I see your points. 731d7a87a

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: Right! 1db8d564a

mailpoet/lib/Automation/Engine/Registry.php
websupporter: c4cde16a9

mailpoet/lib/Automation/Engine/Control/SubjectTransformerHandler.php
websupporter: `public function getSubjectKeysForAutomation()` -> Used in the validator: The result could be the Automation is invalid when you try to update it, but imo that should be already the case as the trigger is missing. Not sure, but rather an effect we would want to have.

`public function getSubjectKeysForTrigger(Trigger $trigger)` -> Used to give the information, which subjects are supported to the React applicaton. No effect as `Trigger $trigger` does not exist.

`public function getAllSubjects(array $subjects)` -> Used when collecting all subjects when the trigger was triggered. We return all subjects of this (not deactivated) trigger and what we can derive from this trigger. Thanks to `array_intersect` when validating the automation the returns subjects will be sufficient for all following steps.

TLDR;
> I hope that's OK because

I do not see a problem.

websupporter:
> Back to this, is it resolved now? It seems to me that the hook will need to consider transformed subjects as well (which, I guess, we can do now).

Yes, we persist now all possible subjects (by overwriting the default `$subjects` [here](https://github.com/mailpoet/mailpoet/pull/4794/files#diff-8ce8be17695037a60fcafd8c6d3f50fb549700f55f0407dcb26e71d1ee795a82R62)) when the AutomationRun is created. When the run is constructed, all subjects are pulled from the database and the check works, since we have the transformer.

image

The checkbox does also appear because all possible subject keys for a trigger are handed over to the UI here: https://github.com/mailpoet/mailpoet/pull/4794/files#diff-442ca25cb43862fbf33744041c2923cc4b894f919cba60589b3d45953ac3f509R104

JanJakes (dismissed):
Awesome, thanks for the fixes! Let's move this forward 🎉

websupporter:
Send email step needed a `SegmentSubject`. Problem was catched by @Aschepikov here: https://github.com/mailpoet/mailpoet-premium/pull/686#issuecomment-1488611610

Aschepikov (approved):
GJ!

PR #4795, opened 2023-03-16
Add new k6 test Newsletter Statistics

veljkho:
## Description

Adding a new k6 test Newsletter Statistics
Added also new helper method `waitForSelectorToBeVisible`
Added new selector `data-automation-id="engagement-tab"`

## Code review notes

This test requires Premium plugin active, but since we don't have it in the free plugin, I would like to keep this test in the suite base and run it later against an online site with active Premium plugin and in nightly tests.

## Linked PRs

Wait for #4791 to be merged first

## Linked tickets

[MAILPOET-4954]

[MAILPOET-4954]: https://mailpoet.atlassian.net/browse/MAILPOET-4954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/scenarios.js
+++ b/mailpoet/tests/performance/scenarios.js
@@ -13,8 +13,6 @@ import { formsAdding } from './tests/forms-adding.js';
 import { newsletterSearching } from './tests/newsletter-searching.js';
 import { newsletterSending } from './tests/newsletter-sending.js';
 import { listsViewSubscribers } from './tests/lists-view-subscribers.js';
-import { listsComplexSegment } from './tests/lists-complex-segment.js';
-import { newsletterStatistics } from './tests/newsletter-statistics.js';
 
 // Scenarios, Thresholds and Tags
 export let options = {
@@ -72,8 +70,6 @@ export async function pullRequests() {
   await newsletterSearching();
   await newsletterSending();
   await listsViewSubscribers();
-  await listsComplexSegment();
-  await newsletterStatistics();
 }
 
 // All the tests ran for a nightly testing
--- a/mailpoet/tests/performance/scenarios.js
+++ b/mailpoet/tests/performance/scenarios.js
@@ -14,6 +14,7 @@ import { newsletterSearching } from './tests/newsletter-searching.js';
 import { newsletterSending } from './tests/newsletter-sending.js';
 import { listsViewSubscribers } from './tests/lists-view-subscribers.js';
 import { listsComplexSegment } from './tests/lists-complex-segment.js';
+import { newsletterStatistics } from './tests/newsletter-statistics.js';
 
 // Scenarios, Thresholds and Tags
 export let options = {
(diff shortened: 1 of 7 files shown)


mailpoet/tests/performance/tests/newsletter-statistics.js
```diff
@@ -0,0 +1,85 @@
...
+    check(page, {
+      'no products present as sold text is present':
+        page.locator('.mailpoet-tab-content > p:nth-child(1)').textContent() ===
+        'Unfortunately, no products were sold as a result of this email!',
+    });
+
+    // Click the subscribers engagement tab
+    page.locator('[data-automation-id="engagement-tab"]');
+    page.waitForSelector('.mailpoet-listing-table');
+    waitForSelectorToBeVisible(
```
JanJakes, line 66: All promises will need to be awaited to prevent some race conditions. Aren't all the `waitFor...` functions promises?

mailpoet/tests/performance/scenarios.js
```diff
@@ -72,9 +74,19 @@ export async function pullRequests() {
...
+  await subscribersListing();
+  await settingsBasic();
+  await subscribersFiltering();
+  await subscribersAdding();
+  await formsAdding();
+  await newsletterSearching();
+  await newsletterSending();
+  await listsViewSubscribers();
+  await listsComplexSegment();
+  await newsletterStatistics();
```
JanJakes, line 89: So what's the difference between regular and nightly tests here?

veljkho: The difference is that in a nightly cycle, we will execute also those tests that require Premium plugin:
- Asserting newsletter stats
- Creating complex segments
- Sending newsletter
Also, worth noting that nightly tests are longer in duration. They ran for 3 iterations per each test case so we get more reliable stats rather then a single iteration. Also, they ran against real online site.

mailpoet/tests/performance/tests/newsletter-statistics.js
veljkho: I was thinking the same, maybe I could add `await` before all `page.` ... What do you think @JanJakes ?

JanJakes: @veljkho I guess that could work. Probably TypeScript would help here (knowing which ones are promises), but I don't know how much work that would be, we can think about it for a hacking topic, maybe.

veljkho: I'm planning to make a separate after-project PR to optimize the test suite with all necessary small changes and will include this. Please approve the PR if you don't have any other comments.

JanJakes (approved):

mailpoet/tests/performance/tests/newsletter-statistics.js
veljkho: Hi @JanJakes, I have included this in my latest PR #4820 . Thanks for suggesting it.

PR #4796, opened 2023-03-16
Transactional emails [MAILPOET-4999]

websupporter:
## Description

_N/A_

### UI
To be able to render the title and the description differently depending on the state, the `Step` definition needed to be extended. Like `subtitle()` they are now methods and can receive the current step data as well as the context in which they are supposed to be rendered. This enables us to render the title in the sidebar differently from the title in the automation view.

Since the title is no longer a string, I introduced `keywords[]`, which can be used for filtering.

To be able to send transactional emails, this PR goes with the meta data approach. The `queue` will get a `is_transactional` value set and based on this the safe guards in the queue will be disabled.

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/696

## Linked tickets

[MAILPOET-4999]

[MAILPOET-4999]: https://mailpoet.atlassian.net/browse/MAILPOET-4999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Engine/Data/Automation.php
```diff
@@ -160,6 +160,30 @@ public function needsFullValidation(): bool {
     return in_array($this->status, [Automation::STATUS_ACTIVE, Automation::STATUS_DEACTIVATING], true);
   }
 
+  public function getPositionOfStep(Step $step): int {
+    return $this->calculatePositionOfStep($step);
+  }
+
+  private function calculatePositionOfStep(Step $step, string $currentStepId = 'root', int $position = 0): int {
```
JanJakes: This might be a note to self as I may discover this later in the PR when I see how this is used, but are we looking for the lowest position or the highest one? What I mean by that is that with multiple triggers in the future, a step could have multiple positions.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -112,11 +116,14 @@ public function run(StepRunArgs $args): void {
...
+    $subscriberSegmentArgs = [
       'subscriber' => $subscriberId,
       'segment' => $segmentId,
-      'status' => SubscriberEntity::STATUS_SUBSCRIBED,
-    ]);
+    ];
+    if (!$this->isTransactional($args)) {
+      $subscriberSegmentArgs['status'] = SubscriberEntity::STATUS_SUBSCRIBED;
+    }
+    $subscriberSegment = $this->subscriberSegmentRepository->findOneBy($subscriberSegmentArgs);
```
JanJakes: Should we send transactional emails also to `bounced` contacts? I guess this may be a question for someone from the Sonnet team.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -128,17 +135,26 @@ public function run(StepRunArgs $args): void {
...
     }
   }
 
+  private function isTransactional(StepRunArgs $args): bool {
+    if (!in_array($args->getAutomationRun()->getTriggerKey(), self::TRANSACTIONAL_TRIGGER_KEYS, true)) {
+      return false;
+    }
+
+    //@ToDo: Position 0 = root, Position 1 = triggers, Position 2 is first possible action; validate assumption
+    return $args->getAutomation()->getPositionOfStep($args->getStep()) === 2;
```
JanJakes: Here's a `@ToDo`. But actually, now I see how this is used! Maybe a bit simpler than checking the position would be to find all parents of the step and check if all of them are `TRANSACTIONAL_TRIGGER_KEYS`. Otherwise, we'd be risking something like this:

```bash
transactional-trigger   non-transactional-trigger
                  \     /
                   \   /
                    \ /
                   email
```

We'd detect this "email" as transactional and then a non-transactional trigger could send it.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -279,6 +282,11 @@ public function getBatchSize(): int {
     return $this->throttlingHandler->getBatchSize();
   }
 
+  private function isTransactional(Sending $queue): bool {
+    $meta = $queue->getMeta();
+    return isset($meta[Sending::META_IS_TRANSACTIONAL]) && $meta[Sending::META_IS_TRANSACTIONAL];
```
JanJakes: The same code was in the previous commit (verify subscriber), so I guess we could add it as a helper method: `$queue->isTransactional()`

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -215,7 +216,9 @@ private function processSending(SendingTask $queue, int $timer): void {
       } else {
         // No segments = Welcome emails or some Automatic emails.
         // Welcome emails or some Automatic emails use segments only for scheduling and store them as a newsletter option
-        $foundSubscribers = SubscriberModel::whereIn('id', $subscribersToProcessIds)
+        $foundSubscribers = $this->isTransactional($queue) ? SubscriberModel::whereIn('id', $subscribersToProcessIds)
+          ->whereNull('deleted_at')
+          ->findMany() : SubscriberModel::whereIn('id', $subscribersToProcessIds)
```
JanJakes: The question if we want to include also `bounced` contacts applies also here.

mailpoet/assets/js/src/automation/integrations/core/steps/delay/index.tsx
```diff
@@ -12,16 +12,13 @@ const getDelayInformation = (delayTypeValue: string, value: number): string =>
     return current.subtitle(value);
   }, '');
 
+const keywords = ['wait', 'pause', 'delay', 'time'];
```
JanJakes: I guess we should make them translatable (applies to all other cases as well)?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/helper/is_transactional.ts
```diff
@@ -0,0 +1,40 @@
+import { select } from '@wordpress/data';
+import { Step } from '../../../../../editor/components/automation/types';
+import { storeName } from '../../../../../editor/store';
+
+const transactionalTriggers = ['woocommerce:order-status-changed'];
```
JanJakes, line 5: Let's pass those from PHP via a context? I mean... we can also do it any time later.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/helper/is_transactional.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  let stepPositionIsUnderNeathTransactionalTrigger = false;
+  triggers.forEach((trigger) => {
+    if (trigger.next_steps.map((next) => next.id).includes(step.id)) {
+      stepPositionIsUnderNeathTransactionalTrigger = true;
+    }
+  });
+  return stepPositionIsUnderNeathTransactionalTrigger;
+}
```
JanJakes, line 40: This is almost what I proposed in the comment in PHP. The only issue is that when there are multiple triggers as parents of the send email step, this will pass even if only one of them is transactional.

I think rather than filtering triggers first, it may be more straightforward to first get all the parents of the step and then check if all are transactional triggers.

JanJakes (commented):
Good stuff, just a few comments + the tiny icon fixes as discussed in Slack.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send_email/helper/is_transactional.ts
JanJakes: Just to link those two places: https://github.com/mailpoet/mailpoet/pull/4796#discussion_r1151591758

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
JanJakes: Linking it with the JS part, that almost works the way I described: https://github.com/mailpoet/mailpoet/pull/4796#discussion_r1151653810

websupporter: > Otherwise, we'd be risking something like this ...

If I understand you correctly, you fear the automation run might have been started with an non-transactional trigger, but we check for this just a few lines up

```
    if (!in_array($args->getAutomationRun()->getTriggerKey(), self::TRANSACTIONAL_TRIGGER_KEYS, true)) {
      return false;
    }
```

mailpoet/lib/Automation/Engine/Data/Automation.php
websupporter: I do not see how it could have multiple positions. My idea was something like that:
```

	[POS 0: root]
		|
	       / \
	      /   \
	     /     \
            /       \
     [POS 1: T1]  [POS 1: T2]
           \        /
            \      /
             \    /
              \  /
               \/
               |
        [POS 2: Action]
```

Do you mean a second trigger could maybe first link to another action before coming to the send email step? :thinking:

JanJakes: > Do you mean a second trigger could maybe first link to another action before coming to the send email step? 🤔

It surely could. In any way, it seems we don't really need the positions at all, do we?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
JanJakes: @websupporter Right, thinking further, you're "deciding at runtime if you send a specific email as transactional or not". I think we shouldn't do that, because we already make the decision in the UI. I think we should:
1. Save `is_transactional` into the step args.
2. Validate the structure (if all parents are transactional triggers) on save.
3. Here in the `SendEmailAction` verify it can be indeed sent as such.

websupporter:
Restart:

In order to be a transactional email: A send email step must directly follow from a trigger and each trigger of an automation must be a "transactional trigger" (currently `woocommerce:order-status-changed`).

Evaluation is done when we save the email. Therefore the hook got extended and the current `Automation` is a second parameter. This way, we can evaluate the conditions. Once the condition is met the type of the email is `transactional`.

Transactional emails are send to all `Subscriber` except for those with the status `bounced`.

Thank you for your informed questions and thorough review @JanJakes

JanJakes:
@websupporter I don't have a time to finish the review today, but one thing I noticed is you've opted to detect (and possibly adjust) the email type on save rather than saving `is_transactional` in step `args` and validating that it can be transactional indeed. Is there any specific reason for that?

I mean, It's probably OK, as long as the JS-side and PHP-side detections never divert (which they shouldn't even in case we just validate), but it closes doors to having the user choose the type of the email, where applicable. If it's not a lot of work, I think it would be safer to save `is_transactional` to step args, and then only use this https://github.com/mailpoet/mailpoet/pull/4796/commits/58b2ec11596a674d00223c2e775e9201bf805f48 to validate if it indeed can be transactional. That may also allow using the `validate` method and dropping https://github.com/mailpoet/mailpoet/pull/4796/commits/82c42ded54acf23ed092cdec3e1d5928bc50f5f4 (since `StepValidationArgs` already provide `$automation`, which would also make it more idiomatic. What do you think?

websupporter:
@JanJakes with regards to `is_transactional` as a argument of the `SendEmailAction` step, I created a maintenance ticket to revisit this issue: https://mailpoet.atlassian.net/browse/MAILPOET-5209

JanJakes (dismissed):
@websupporter Thanks for the changes, hitting approve, but it needs a rebase due to conflicts. You can now remove all table truncations from the tests.

JanJakes (approved):

Aschepikov:
Hi @websupporter as I see in the ticket we should use `transactional emails` just with the `order status changed` trigger. But on my site, every `send email` action is a transactional email. Can you check it please
image

Aschepikov (approved):
GJ!

PR #4797, opened 2023-03-16
Sending gets stuck if after scheduling the susbcribers unsubscribe or get deleted [MAILPOET-5134]

samnajian:
## Description

This PR invalidates scheduled tasks when in sending time determines they have no more subscriptions.

## QA notes

- Schedule a newsletter with one subscriber
- Unsubscribe the subscribers of the list selected in previous step
- (optional) Change the schedule time via in DB to reduce waiting
- Verify this task is not in `Running tasks` table under Help > System Status.

## Linked tickets

 [MAILPOET-5134]

[MAILPOET-5134]: https://mailpoet.atlassian.net/browse/MAILPOET-5134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -200,6 +200,17 @@ class SendingQueue {
 
     // get subscribers
     $subscriberBatches = new BatchIterator($queue->taskId, $this->getBatchSize());
+    if ($subscriberBatches->count() === 0) {
+      $this->loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->info(
+        'no subscribers to process',
+        ['task_id' => $queue->taskId]
+      );
+      $task = $queue->getSendingQueueEntity()->getTask();
+      if ($task) {
+        $this->scheduledTasksRepository->invalidateTask($task);
+      }
+      return;
+    }
     /** @var int[] $subscribersToProcessIds - it's required for PHPStan */
     foreach ($subscriberBatches as $subscribersToProcessIds) {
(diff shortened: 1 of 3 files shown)


mailpoet/tests/integration/Cron/Workers/SendingQueue/SendingQueueTest.php
```diff
@@ -1389,4 +1411,21 @@ private function getSendingQueueWorker($mailerMock = null): SendingQueueWorker {
       $this->sendingQueuesRepository
     );
   }
+
+  private function createListWithSubscriber(): array {
+    $segmentFactory = new SegmentFactory();
+    $segmentName = 'List ' . Security::generateRandomString();
```
costasovo, line 1417: We do reset DB before every single test, so there is no need to generate random names, and if you don't need specific names, you can go with the default. But using random is perfectly fine. No need to change it.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -200,6 +200,17 @@ private function processSending(SendingTask $queue, int $timer): void {
 
     // get subscribers
     $subscriberBatches = new BatchIterator($queue->taskId, $this->getBatchSize());
+    if ($subscriberBatches->count() == 0) {
```
costasovo: We prefer using === to check also the type. Could you please update it?

costasovo (changes requested):
Thank you for addressing this issue. 👍 I found only a small problem that I think needs to be addressed.

costasovo (approved):
LGTM! 👍

Aschepikov (approved):
GJ!

PR #4799, opened 2023-03-17
New WooCommerce segment: single order value [MAILPOET-4988]

rodrigoprimo:
## Description

This PR adds a new WooCommerce segment called "single order value". It allows users to create segments based on the value of a single order made by customers. For example, customers that placed an order with a total value equal to or higher than $1000 in the last seven days.

## Code review notes

I based this segment on the WooCommere total spent segment.

## QA notes

To test this new segment:

- Create a few WooCommerce others with different customers. This can be done via the web interface or using [wc-smooth-generator](https://github.com/woocommerce/wc-smooth-generator).
- Go to `/wp-admin/admin.php?page=mailpoet-segments#/new-segment` and create a new segment of the type `single order value`.
- Send an email using this segment and confirm that the recipients of the email are the expected WooCommerce customers that placed at least one order that matched the criteria chosen when creating the segment.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4988]

## After-merge notes

_N/A_

[MAILPOET-4988]: https://mailpoet.atlassian.net/browse/MAILPOET-4988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo:
@triple0t, right after I created this PR, I noticed a conflict due to a change that landed in `trunk` after I created this branch. I fixed the conflict and force-pushed the changes now.

triple0t (approved):
LGTM

Aschepikov:
Hi, @rodrigoprimo I have one issue with this segment, always when I am making a new order I see just the last subscriber in the segment. So there is always one last subscriber even if I have a few subscribers with correct orders for the segment requirements. 
Example 1, I have here 3 orders 45 (1 order with subscriber), 15 (2 order with subscriber), and 60 (3 order with unsubscribed) in total
image
I am creating a segment with more than 30 in total in the last 7 days, and it shows just 1 subscriber from the last order for 60 in total
image
And this email is not a subscriber. 
image
But it is the last order
image
The correct subscriber was 45 in total from the 1 order
image
And I had it in the segment, but after the last order, this subscriber disappeared from the segment list. If I will create a new one there will be just 1 last subscriber. 

I thought that problem could be related to hpos enabled on this site, so I tried it on a new one.

Example 2. I have two orders 72 and 110 in total
image
But in the list with more than 30 in the last 10 days I have just one last subscriber
image
It is a second subscriber from the second order and this subscriber replaced the first one in the segment
image
Now I can create a new order with the third subscriber 
image
I will see it in the orders
image
But in the segment, I still have just one subscriber 
image
And it is a last subscriber from the third order 
image

Aschepikov:
I checked it with the "total spent" segment too and it works the same just the last one subscriber
image
image
And segment is in country too 
image


So I guess it is an issue of all woocommerce segments and it is not connected with single order value

rodrigoprimo:
Thanks for sharing the credentials to your site, @Aschepikov.

I believe that the problem is that all the orders are associated with the same user. They do have different email addresses, but the customer is always "Test Test" (customer ID 1):

Screenshot from 2023-03-28 10-26-56

Aschepikov (approved):
OMG My fault, sorry. How I can forget about it?! GJ anyway!

PR #4800, opened 2023-03-18
Filter for license provisioning

brezocordero:
This commit adds a filter that will provision the API key on WordPress.com

## Description

To test the filter you need:
- A site that can connect to the Bridge (not localhost) and that has wp-cli
- A valid API key for a production subscription that belongs to you
- MailPoet installed and active
- Better if you taste with a fresh site (or click on reinstall from scratch)
- To run the filter: you can unzip the file attached, insert a good key, place the file where you can access it with `wp-cli` 
- run:`wp eval-file runProvisioningFilter.php` (or copy paste line by line inside of `wp shell`)
- Check that MSS is set up, the key added and verified.
- With a bad key, missing key or missing payload it should return a WP_Error.
- If it can't activate set the settings or refresh the key it should return a WP_Error

[runProvisioningFilter.php.zip](https://github.com/mailpoet/mailpoet/files/11007330/runProvisioningFilter.php.zip)

## Code review notes

Please check carefully and verify that I haven't missed anything to activate the MSS and validate the key. 
It's been a while since I haven't worked in the plugin. Please nitpick :)

## QA notes

_N/A_

## Linked PRs

[MAILPOET-5131]

## Linked tickets

_N/A_

## After-merge notes

_N/A_

[MAILPOET-5131]: https://mailpoet.atlassian.net/browse/MAILPOET-5131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Settings.php
+++ b/mailpoet/lib/API/JSON/v1/Settings.php
@@ -4,7 +4,9 @@ namespace MailPoet\API\JSON\v1;
 
 use MailPoet\API\JSON\Endpoint as APIEndpoint;
 use MailPoet\API\JSON\Error as APIError;
+use MailPoet\API\JSON\ErrorResponse;
 use MailPoet\API\JSON\Response;
+use MailPoet\API\JSON\SuccessResponse;
 use MailPoet\Config\AccessControl;
 use MailPoet\Config\ServicesChecker;
 use MailPoet\Cron\Workers\SubscribersEngagementScore;
@@ -492,5 +494,25 @@ class Settings extends APIEndpoint {
       'showNotice' => !!$draftReEngagementEmails,
       'action' => 'reactivate',
     ];
+  }
+
+  /**
+   * Prepares the settings to set up MSS with the given key and calls the set method.
+   *
+   * @param string $apiKey
+   * @return ErrorResponse|SuccessResponse
+   */
+  public function setupMSS(string $apiKey) {
+    $new_settings = [
+      'mta_group' => 'mailpoet',
+      'mta' => [
+        'method' => 'MailPoet',
+        'mailpoet_api_key' => $apiKey,
+      ],
+      'signup_confirmation' => [
+        'enabled' => '1',
+      ],
+    ];
+    return $this->set($new_settings);
   }
 }
--- a/mailpoet/lib/API/JSON/v1/Settings.php
+++ b/mailpoet/lib/API/JSON/v1/Settings.php
@@ -13,6 +13,7 @@ use MailPoet\Cron\Workers\SubscribersEngagementScore;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Form\FormMessageController;
+use MailPoet\Mailer\Mailer;
 use MailPoet\Mailer\MailerLog;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
@@ -502,11 +503,11 @@ class Settings extends APIEndpoint {
    * @param string $apiKey
    * @return ErrorResponse|SuccessResponse
    */
-  public function setupMSS(string $apiKey) {
+  public function setKeyAndSetupMss(string $apiKey) {
     $new_settings = [
       'mta_group' => 'mailpoet',
       'mta' => [
-        'method' => 'MailPoet',
+        'method' => Mailer::METHOD_MAILPOET,
         'mailpoet_api_key' => $apiKey,
       ],
       'signup_confirmation' => [
--- a/mailpoet/lib/API/JSON/v1/Settings.php
+++ b/mailpoet/lib/API/JSON/v1/Settings.php
@@ -512,6 +512,7 @@ class Settings extends APIEndpoint {
       'signup_confirmation' => [
         'enabled' => '1',
       ],
+      'premium.premium_key' => $apiKey,
     ];
     return $this->set($new_settings);
   }
--- a/mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
+++ b/mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
@@ -38,7 +38,7 @@ class DotcomLicenseProvisioner {
    *
    * @return bool
    */
-  private function isAtomicPlatform(): bool {
+  public function isAtomicPlatform(): bool {
     return defined('IS_ATOMIC') && IS_ATOMIC && defined('ATOMIC_CLIENT_ID') && (ATOMIC_CLIENT_ID === '2');
   }
 
@@ -83,7 +83,7 @@ class DotcomLicenseProvisioner {
    * @param string $apiKey
    * @return true|WP_Error
    */
-  private function activateMSS(string $apiKey) {
+  public function activateMSS(string $apiKey) {
     $response = $this->settings->setupMSS($apiKey);
     if ($response instanceof ErrorResponse) {
       $this->loggerFactory->getLogger(LoggerFactory::TOPIC_PROVISIONING)->error(
--- a/mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
+++ b/mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
@@ -1,0 +1,126 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\WPCOM;
+
+use MailPoet\API\JSON\ErrorResponse;
+use MailPoet\API\JSON\v1\Services;
+use MailPoet\API\JSON\v1\Settings;
+use MailPoet\Logging\LoggerFactory;
+use WP_Error;
+
+/**
+ * This class is responsible for receiving and activating the license purchased from WP.com Marketplace.
+ */
+class DotcomLicenseProvisioner {
+  const EVENT_TYPE_PROVISION_LICENSE = 'provision_license';
+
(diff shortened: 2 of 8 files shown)


costasovo (commented):
@brezocordero If I understand the task correctly in this filter, we want to replicate what happens when a user inserts a new API key in settings. Am I correct? For further comments, I'll assume I am :)

I like the idea of using API endpoints.

I noticed that in the current implementation, we store the API as an MSS key, which is correct, but in the plugin, we store the API key in two places (historically) as the key for MSS and as the key for the premium plugin. I assume we want to use this provisioning to set up access to both MSS and the premium plugin (if it is included in the plan). So I think we need also to set and verify the premium key.

In JS we call two actions: [verifyMssKey](https://github.com/mailpoet/mailpoet/blob/cf5718122f9183aa4963a1d088c731ca352c555c/mailpoet/assets/js/src/settings/store/actions/mss_and_premium.ts#L16-L70) and the [verifyPremiumKey](https://github.com/mailpoet/mailpoet/blob/cf5718122f9183aa4963a1d088c731ca352c555c/mailpoet/assets/js/src/settings/store/actions/mss_and_premium.ts#L73-L89)

I see that verifyMssKey is covered so we need to add also the functionality from verifyPremiunKey.

brezocordero:
> we want to replicate what happens when a user inserts a new API key in settings.

Yes, correct!

> in the plugin, we store the API key in two places (historically) as the key for MSS and as the key for the premium plugin
> we need to add also the functionality from verifyPremiunKey.

Thanks, excellent review! I will work on the part I missed.

brezocordero:
@costasovo Premium key saved and verified in 426f0dff75c2fab8c08e6cc84ee4344dcfb1037b

Could you please take another look?

mailpoet/lib/API/JSON/v1/Settings.php
```diff
@@ -493,4 +495,25 @@ public function reactivateReEngagementEmails(): array {
...
+   * Prepares the settings to set up MSS with the given key and calls the set method.
+   *
+   * @param string $apiKey
+   * @return ErrorResponse|SuccessResponse
+   */
+  public function setupMSS(string $apiKey) {
+    $new_settings = [
+      'mta_group' => 'mailpoet',
+      'mta' => [
+        'method' => 'MailPoet',
```
costasovo: We have a PHP constant for that `MailPoet\Mailer\Mailer::METHOD_MAILPOET` so please use the constant.

mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
```diff
@@ -0,0 +1,134 @@
...
+    $this->services = $services;
+  }
+
+  /**
+   * Returns true if in the context of WordPress.com Atomic platform.
+   *
+   * @return bool
+   */
+  public function isAtomicPlatform(): bool {
+    return defined('IS_ATOMIC') && IS_ATOMIC && defined('ATOMIC_CLIENT_ID') && (ATOMIC_CLIENT_ID === '2');
```
costasovo, line 43: I think that a little explanation at least why we need to check `ATOMIC_CLIENT_ID === '2'` would be helpful. If I remember correctly ATOMIC_CLIENT_ID 2 is client for all customer sites and there are other clients for internal sites. Am I correct?

mailpoet/lib/API/JSON/v1/Settings.php
```diff
@@ -493,4 +495,25 @@ public function reactivateReEngagementEmails(): array {
...
     ];
   }
+
+  /**
+   * Prepares the settings to set up MSS with the given key and calls the set method.
+   *
+   * @param string $apiKey
+   * @return ErrorResponse|SuccessResponse
+   */
+  public function setupMSS(string $apiKey) {
```
costasovo: Now that the method saves also the premium key maybe we could use a different name. Maybe something like `setKeyAndSetupMss` ? Hmm I'm not sure maybe I'm overthinking this a bit. If you think it is fine you can ignore this comment.

mailpoet/tests/integration/API/JSON/v1/SettingsTest.php
```diff
@@ -313,6 +313,35 @@ public function testDeleteReturnErrorIfSettingDoesntExist() {
...
+      $this->diContainer->get(SubscribersCountsController::class),
+      $this->diContainer->get(TrackingConfig::class),
+      $this->diContainer->get(ConfirmationEmailCustomizer::class)
+    );
+
+    expect($this->endpoint->setupMSS($newKey))->isInstanceOf(SuccessResponse::class);
+    expect($this->settings->get('mta.mailpoet_api_key'))->equals($newKey);
+    expect($this->settings->get('mta_group'))->equals('mailpoet');
+    expect($this->settings->get('mta.method'))->equals('MailPoet');
+    expect($this->settings->get('signup_confirmation.enabled'))->equals(1);
```
costasovo, line 342: We should also test that it sets the premium key.
`expect($this->settings->get('premium.premium_key'))->equals($newKey);`

mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    // This is necessary if the key changed but the sending method was already set to MailPoet
+    $response = $this->services->refreshMSSKeyStatus();
+    if ($response instanceof ErrorResponse) {
+      $this->loggerFactory->getLogger(LoggerFactory::TOPIC_PROVISIONING)->error(
+        'Refreshing the key failed',
+        ['$response' => $response]
+      );
+      return new WP_Error('Provisioning failed activating the data', $this->concatMessages($response));
```
costasovo, line 104: The error message is the same for MSS and Premium key. I think it would be nice to distinguish these two. Maybe `Provisioning failed to verify api key access for MSS/premium` ?

mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    $response = $this->services->refreshPremiumKeyStatus();
+    if ($response instanceof ErrorResponse) {
+      $this->loggerFactory->getLogger(LoggerFactory::TOPIC_PROVISIONING)->error(
+        'Refreshing Premium key failed',
+        ['$response' => $response]
+      );
+      return new WP_Error('Provisioning failed activating the data', $this->concatMessages($response));
```
costasovo: Hmm I've just realized that some plans that we offer in the shop has access to MSS but not to premium. Basically all Starter plans. Also Creator plans have only Premium and but not MSS.

I'm not sure if we offer those plans on WPCOM, but if we do we should make sure that the provisioning doesn't fail (end with WP_Error) for those. 

I see that we fixed this issue for the front end validation in https://github.com/mailpoet/mailpoet/pull/4399 and if I'm not mistaken the fix should also work for both `refreshPremiumKeyStatus` and `refreshMSSKeyStatus` because they internally reuse checkMSSKey respectively `checkPremiumKey` and if I read the code correctly they should retun Success response for `Bridge::KEY_VALID_UNDERPRIVILEGED`.

I rather keep this comment. I'm not sure if you are aware/tested the provisioning for such plan types.

costasovo (changes requested):
@brezocordero Nice job! Thank you for the detailed steps for testing. TIL about `wp eval-file`. It is quite useful. I remember creating a few wp cli commands in the past that could be replaced by this.

I tested the hook with valid and invalid keys and it worked as expected. You asked me to nitpick so I tried my best and left a couple of comments.

mailpoet/lib/API/JSON/v1/Settings.php
brezocordero: Thanks and done.

mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
brezocordero: Good point! Added a comment.

mailpoet/lib/API/JSON/v1/Settings.php
brezocordero: Changed to the more descriptive name

mailpoet/tests/integration/API/JSON/v1/SettingsTest.php
brezocordero: Done

mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
brezocordero: Done

mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
brezocordero: Thanks so much for taking the different cases into account!
We will have Starter and Business but not the other plans.
After your comment, I have also tested with the Starter plan, with and without the premium plugin.
The fix works as you said.

brezocordero:
Assigning it back to you @costasovo and thanks again for all the comments.

mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php
costasovo: Thanks for testing the started plan!

costasovo (approved):
Great job! 👍

Aschepikov (approved):
Gj!

PR #4802, opened 2023-03-20
Update TypeScript to version 5 [MAILPOET-5015]

JanJakes:
## Description

This PR updates TypeScript to version 5, enforces only that version using PNPM overrides, and fixes some reported issues.

## Code review notes

We can merge this when approved. QA can be skipped as TypeScript is only static type analysis.

## QA notes

QA can be skipped as TypeScript is only static type analysis.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/697

## Linked tickets

[MAILPOET-5015]

[MAILPOET-5015]: https://mailpoet.atlassian.net/browse/MAILPOET-5015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
@@ -5,7 +5,7 @@ import { MoreControlType, StepMoreControlsType } from '../../../types/filters';
 import { Step } from '../../../editor/components/automation/types';
 
 const emailStatisticsControl = (step: Step): MoreControlType => {
-  const hasEmail = step.args?.email_id > 0;
+  const hasEmail = (step.args?.email_id as number) > 0;
   return {
     key: 'statistics',
     control: {
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
@@ -15,7 +15,7 @@ const emailStatisticsControl = (step: Step): MoreControlType => {
       onClick: () => {
         window.open(
           `admin.php?page=mailpoet-newsletters#/stats/${
-            step.args.email_id as string
+            step.args.email_id as number
           }`,
           '_blank',
         );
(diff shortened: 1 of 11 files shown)


mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
```diff
@@ -5,7 +5,7 @@ import { MoreControlType, StepMoreControlsType } from '../../../types/filters';
 import { Step } from '../../../editor/components/automation/types';
 
 const emailStatisticsControl = (step: Step): MoreControlType => {
-  const hasEmail = step.args?.email_id > 0;
+  const hasEmail = (step.args?.email_id as number) > 0;
```
johnolek: Is this not somewhat inconsistent with L18, where we reference the same `email_id` using `as string`?

Just to double check my understanding of TS, using `as` means we're telling TS that we're confident something is the thing we're saying it is, right?

johnolek (dismissed):
Looks good, thanks for working on this @JanJakes! I had one tiny comment, mostly to verify my own understanding of TS, but I don't think it's actually important. If it does happen to warrant a change I'm leaving this unmerged for now, but feel free to merge it if no changes are needed.

mailpoet/assets/js/src/automation/integrations/mailpoet/step-controls/index.tsx
JanJakes: Good catch! L18 looks like a "bug" — [this is indeed a number](https://github.com/mailpoet/mailpoet/blob/2d0228e4e485b4e7ad07becfff0d19f211ccd009/mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php#L67). That said, the condition is a bit strange in the first place. I fixed it to a more descriptive `Number.isInteger(step.args?.email_id)` and I fixed the type assertion on L18.

> Just to double check my understanding of TS, using as means we're telling TS that we're confident something is the thing we're saying it is, right?

Yeah, and we're telling it to TS quite strongly so `as` should be avoided. But in cases like these (the const has type `unknown`), it probably makes sense.

johnolek: Thanks for making the tweak and for the explanation!

johnolek (approved):

PR #4803, opened 2023-03-20
Fix font sizes settings in the form editor [MAILPOET-5139]

costasovo:
## Description

After upgrading the Gutenberg JS packages, the font size pickers started to support also unit selection. During the upgrade, we forgot to add support for other units to the front-end renderer, so all font-sizes rendered in the front end are still in pixels no matter what unit was saved in the editor.

## Code review notes

_N/A_

### Replication
1) Create a form and add a paragraph, and set font size using em or rem unit
2) Go to form preview and check the value of the paragraph.

Note: The issue can also be replicated for global form styles. You can replicate/check it by setting the global font size and checking a label or text input font-size value using dev tools.

Note 2: When using relative units `em` or `rem` the font size might differ in the form editor in admin and on the front end. This is expected since a theme may set a different base size.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5139]

## After-merge notes

_N/A_

[MAILPOET-5139]: https://mailpoet.atlassian.net/browse/MAILPOET-5139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/form_editor/store/map_form_data_after_loading.jsx
+++ b/mailpoet/assets/js/src/form_editor/store/map_form_data_after_loading.jsx
@@ -179,7 +179,7 @@ export function mapFormDataAfterLoading(data) {
           : defaults.formStyles.inputPadding,
       borderColor: data.settings.border_color,
       fontFamily: data.settings.font_family,
-      fontSize: data.settings.fontSize && asNum(data.settings.fontSize),
+      fontSize: data.settings.fontSize,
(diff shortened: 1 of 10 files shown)


JanJakes (approved):

Aschepikov (approved):
GJ!

PR #4811, opened 2023-03-24
Refactor k6 tests and include chai.js

veljkho:
## Description

Now with chai.js library, we could use `expect` and `describe` in tests which will make our tests more stable and reliable. This can be seen also in the output of the test result, here is how it looks now when executed scenario.js:

https://d.pr/i/iUJZZP - The parent page - should do something - expect pass or fail

I removed `try` `finally` from tests so we can see when the test failed in execution (like E2E tests) and if it fails the performance testing is not reliable. Before, even if the E2E test has failed, the performance test will continue in execution and measuring, but then we wouldn't know what we're measuring with perf tests.

## Code review notes

This is about refactoring existing tests so the change is almost the same across all tests.

## Linked PRs

Wait for #4795 before merging

## Linked tickets

[MAILPOET-5156]

[MAILPOET-5156]: https://mailpoet.atlassian.net/browse/MAILPOET-5156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/tests/lists-complex-segment.js
+++ b/mailpoet/tests/performance/tests/lists-complex-segment.js
@@ -3,9 +3,13 @@
 /**
  * External dependencies
  */
-import { sleep, check } from 'k6';
+import { sleep } from 'k6';
 import { chromium } from 'k6/experimental/browser';
 import { randomIntBetween } from 'https://jslib.k6.io/k6-utils/1.1.0/index.js';
+import {
+  expect,
+  describe,
+} from 'https://jslib.k6.io/k6chaijs/4.3.4.2/index.js';
 
 /**
  * Internal dependencies
@@ -17,6 +21,7 @@ import {
   headlessSet,
   timeoutSet,
   defaultListName,
+  listsPageTitle,
 } from '../config.js';
 import { authenticate, selectInReact } from '../utils/helpers.js';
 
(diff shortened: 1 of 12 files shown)


mailpoet/tests/performance/tests/lists-complex-segment.js
```diff
@@ -27,100 +32,108 @@ export async function listsComplexSegment() {
...
+      '#segments_container > form > div > div.mailpoet-segments-segments-section > button',
+    )
+    .click();
 
-    // Select "Subscribed date" action
-    selectInReact(page, '#react-select-5-input', 'subscribed date');
+  // Select "Subscribed date" action
+  selectInReact(page, '#react-select-5-input', 'subscribed date');
+  page.waitForSelector('.mailpoet-form-notice-message');
+  describe(listsPageTitle, () => {
```
websupporter, line 82: I wonder if checking for "Calculating segment size" makes the test `flaky` as it is time constrained. I mean, it appears and disappears.

Imo it shouldn't be flaky at this point.

websupporter (dismissed):
LGTM. Thank you @veljkho, super cool project

mailpoet/tests/performance/tests/lists-complex-segment.js
veljkho: I understand your worry. Haven't seen it as flaky in multiple test executions, however, if I see it one time as flaky I will remove it.

websupporter (approved):
lgtm

PR #4813, opened 2023-03-24
ESLint config fixes [MAILPOET-5015]

JanJakes:
## Description

This PR fixes our ESLint configuration, so IDEs understand it, and adds some refactoring. It also seems to speed up the full linter check in our CircleCI job by about 20 to 30 seconds. It implements the following changes:
1. ESLint and all associated configs and plugins were updated to their latest versions.
2. Errors reported by the updated ESLint packages were addressed.
3. All ESLint configs were rewritten to the [new flat configuration format](https://eslint.org/docs/latest/use/configure/configuration-files-new), which unlocks future compatibility with ESLint 9. This step required using some ESLint back compatibility helpers and a few small fixes.
4. Merges all of our ESLint commands into one and makes sure the process has enough memory.
5. Merges `premium` config to `es6`, `premium-ts` config to `ts`, and `tests_newsletter_editor` config to `es5`, reducing the number of ESLint configs from 6 to 3.

It doesn't further restructure the configs, as this was where I was able to get, and the changeset was getting rather large. I think that in the future, we should follow up by:
1. Trying to consolidate ES5 and ES6 configs (btw, the naming is not correct anymore, both configs work with ES6, in fact) — ideally by fixing the code targeted by ES5 to pass ES6 config rules. If that's too difficult to do, we may consider having ES5 extend from ES6, or keeping both.
2. Making TS config extend the ES one, if possible.
3. Splitting the configuration into "base" and "full" configs (such as `airbnb/base` and `airbnb`) to differentiate between a universal non-React code (e.g., for Node.js apps), and React apps.

## Code review notes

To verify the changes work for you in your IDE, you can use the following steps:
1. Checkout this branch and install deps (`git switch eslint-config-fix` and `pnpm install --frozen-lockfile`).
2. Restart your IDE (just to be sure) and ensure that you have ESLint checking activated and set to automatic configuration (in PHPStorm this means `Settings > Languages & Frameworks > Code Quality Tools > ESLint` should be set to `Automatic ESLint configuration`).
3. Choose any file (e.g., `from-scratch.tsx`) and insert anything that would make ESLint unhappy (e.g., `/**test*/`).
4. You should see the issues reported by ESLint:
Screenshot 2023-03-25 at 22 29 27

This PR touches linter configuration, so you can skip QA and merge this when approved.

## QA notes

QA can be skipped.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/700

## Linked tickets

[MAILPOET-5015]

[MAILPOET-5015]: https://mailpoet.atlassian.net/browse/MAILPOET-5015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/js/eslint-config/package.json
```diff
@@ -4,6 +4,9 @@
   "version": "0.1.0",
   "description": "",
   "main": "index.js",
+  "bin": {
+    "eslint": "./node_modules/eslint/bin/eslint.js"
```
costasovo, line 8: Could you please add a note (maybe into the commit message) about why we need to export explicitly?

costasovo (dismissed):
Awesome job! 👏 Finally, we can don't need to set specific lint config files in IDE 🎉 
Also, TIL about the new flat format for eslint. 

I'm approving the PR. I added just one comment asking for some additional details about a change.

mailpoet/eslint.config.js
```diff
@@ -26,4 +26,27 @@ module.exports = [
     ...config,
     files: ['tests/javascript_newsletter_editor/**/*.js'],
   })),
+
+  // ES6 config overrides
+  {
+    files: ['assets/js/src/**/*.jsx', 'tests/javascript/**/*.js'],
```
costasovo, line 44: I see we also have a couple of .jsx files in the premium plugin also in the assets folder. So I guess these rules will be applied to those as well. Correct?

JanJakes: No, this config is not loaded in the premium plugin. In there it's handled by this one: https://github.com/mailpoet/mailpoet-premium/blob/22ba8630b34579ac782175173c1a434c9ff9bc53/eslint.config.js

costasovo:
@JanJakes, I added a couple of questions. Just so that I better understand some details. I approved both PRs but assigned them back to you. I assumed you might want to ping devs about the merge because they will need to update the IDE configuration. Please feel free to merge.

packages/js/eslint-config/package.json
JanJakes: @costasovo Hmm, we don't necessarily need it, actually, it "somehow" works without it (in the free plugin), but I was thinking it's the "correct" thing to do — a package should declare its `bin`s, if they're expected to be used. But in the premium plugin, this is, in fact, needed... otherwise the `../node_modules/.bin/eslint` paths need to be used. So yeah, I should (and will) add a comment to the commit 😀

JanJakes:
@costasovo Answered & added the commit info.

> I assumed you might want to ping devs about the merge because they will need to update the IDE configuration.

Makes sense, although I think until now it didn't work at all, so can't really make it much worse for anyone, I guess 😆

packages/js/eslint-config/package.json
costasovo: Thank you for the explanation. I was guessing it has something to do with the premium.

costasovo (approved):

PR #4818, opened 2023-03-28
MAILPOET-5145 Simplify integration test cleanup && MAILPOET-5185

johnolek:
## Description

This PR is an attempt to create more of a clean slate before each integration test is run. This is achieved primarily by deleting all the data in the tables that are part of the plugin before each test is run.

This proved to be much more work than I expected. It turns out many of our tests have assumptions about the IDs of created records, which was the main reason some tests would fail in a full run but would pass when run individually. This made it take a significant amount of time to test some changes because I would have to run the entire suite again. I also ran into some intermittent failures that were very difficult to debug.

### In-memory mount for mysql data
I also added a `tmpfs` mount in 644dff22109ccd452ab7fe9dc79f6f1cffd7a011. It does seem to have sped up test execution times, though not by a huge margin.

Before:
image

After:
image

For reference, here are some run times from a branch that doesn't include any of the changes in this PR:
image

Of course times can vary quite a bit from run to run, so take these with a grain of salt.

### [MAILPOET-5185](https://mailpoet.atlassian.net/browse/MAILPOET-5185) bug discovered and fixed
One of the tests that ended up failing was a legitimate failure and uncovered a bug, [MAILPOET-5185](https://mailpoet.atlassian.net/browse/MAILPOET-5185). I fixed that as part of this PR, otherwise the tests wouldn't pass. This fix ensures that the personal data exporter always includes all custom field data for subscribers (see ticket for more details).

## Code review notes

Although there are many changes, the vast majority are simply removing calls to truncating entities. I didn't remove the helper entirely because a few tests rely on truncating tables within the tests.

In my testing, I found that simply deleting all the rows in all the tables was much faster than truncating the tables. I suspect this may be because we never create all that much data to begin with for any given test. Run locally, the version where I truncated the tables before every test used 320 seconds just for the truncating operations, or about 0.14 seconds per test. I couldn't figure out a way to truncate the tables in a single command.

Simply running a `DELETE FROM table` command instead took a total of around 9 seconds total when executing all 2259 tests. 

## QA notes

The only thing that should require testing is the personal data exporting, MAILPOET-5185.

## Linked PRs

_N/A_

## Linked tickets
https://mailpoet.atlassian.net/browse/MAILPOET-5145
https://mailpoet.atlassian.net/browse/MAILPOET-5185

## After-merge notes

This includes some rather large changes, so I plan on publishing a P2 post summarizing what's changed and what engineers should do differently going forward when it gets merged.

[MAILPOET-5185]: https://mailpoet.atlassian.net/browse/MAILPOET-5185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/DI/ContainerConfigurator.php
+++ b/mailpoet/lib/DI/ContainerConfigurator.php
@@ -436,7 +436,7 @@ class ContainerConfigurator implements IContainerConfigurator {
     $container->autowire(\MailPoet\Subscription\Captcha\CaptchaConstants::class)->setPublic(true);
     $container->autowire(\MailPoet\Subscription\CaptchaFormRenderer::class)->setPublic(true);
     $container->autowire(\MailPoet\Subscription\Captcha\CaptchaSession::class);
-    $container->autowire(\MailPoet\Subscription\Captcha\CaptchaRenderer::class);
+    $container->autowire(\MailPoet\Subscription\Captcha\CaptchaRenderer::class)->setPublic(true);
     $container->autowire(\MailPoet\Subscription\Captcha\CaptchaPhrase::class);
     $container->autowire(\MailPoet\Subscription\Captcha\Validator\BuiltInCaptchaValidator::class)->setPublic(true);
     $container->autowire(\MailPoet\Subscription\Captcha\Validator\RecaptchaValidator::class)->setPublic(true);
--- a/mailpoet/tests/_support/IntegrationCleanupExtension.php
+++ b/mailpoet/tests/_support/IntegrationCleanupExtension.php
@@ -24,9 +24,26 @@ class IntegrationCleanupExtension extends Extension {
   private $deleteStatement;
 
   public function beforeSuite(SuiteEvent $event) {
+    global $wpdb;
+
     $this->entityManager = ContainerWrapper::getInstance()->get(EntityManager::class);
 
     $this->deleteStatement = 'SET FOREIGN_KEY_CHECKS=0;';
+
+    $automationTables = [
+      'mailpoet_automation_run_logs',
+      'mailpoet_automation_run_subjects',
+      'mailpoet_automation_runs',
+      'mailpoet_automation_triggers',
+      'mailpoet_automation_versions',
+      'mailpoet_automations',
+    ];
+
+    foreach ($automationTables as $automationTable) {
+      $fullTable = sprintf('%s%s', $wpdb->prefix, $automationTable);
+      $this->deleteStatement .= "DELETE FROM $fullTable;";
+    }
+
     foreach ($this->entityManager->getMetadataFactory()->getAllMetadata() as $metadata) {
       $class = $metadata->getName();
       $table = $metadata->getTableName();
--- a/mailpoet/tests/_support/IntegrationCleanupExtension.php
+++ b/mailpoet/tests/_support/IntegrationCleanupExtension.php
@@ -21,7 +21,7 @@ class IntegrationCleanupExtension extends Extension {
   ];
 
   /** @var string */
-  private $deleteStatement;
+  private $cleanupStatements;
 
   public function beforeSuite(SuiteEvent $event) {
     $this->entityManager = ContainerWrapper::getInstance()->get(EntityManager::class);
@@ -34,22 +34,22 @@ class IntegrationCleanupExtension extends Extension {
       AND table_name != '{$mpPrefix}migrations'
     ");
 
-    $this->deleteStatement = 'SET FOREIGN_KEY_CHECKS=0;';
+    $this->cleanupStatements = 'SET FOREIGN_KEY_CHECKS=0;';
     foreach ($tables as $table) {
-      $this->deleteStatement .= "DELETE FROM $table;";
+      $this->cleanupStatements .= "DELETE FROM $table;";
     }
 
     // save plugin version to avoid triggering migrator and populator
     $settingsTable = $this->entityManager->getMetadataFactory()->getMetadataFor(SettingEntity::class)->getTableName();
     $dbVersion = Env::$version;
-    $this->deleteStatement .= "
+    $this->cleanupStatements .= "
       INSERT INTO $settingsTable (name, value, created_at, updated_at)
       VALUES ('db_version', '$dbVersion', NOW(), NOW());
     ";
-    $this->deleteStatement .= 'SET FOREIGN_KEY_CHECKS=1';
+    $this->cleanupStatements .= 'SET FOREIGN_KEY_CHECKS=1';
   }
 
   public function beforeTest(TestEvent $event) {
-    $this->entityManager->getConnection()->executeStatement($this->deleteStatement);
+    $this->entityManager->getConnection()->executeStatement($this->cleanupStatements);
   }
 }
--- a/mailpoet/tests/_support/IntegrationCleanupExtension.php
+++ b/mailpoet/tests/_support/IntegrationCleanupExtension.php
@@ -1,0 +1,50 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\TestsSupport;
+
+use Codeception\Event\SuiteEvent;
+use Codeception\Event\TestEvent;
+use Codeception\Events;
+use Codeception\Extension;
+use MailPoet\Config\Env;
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\Entities\SettingEntity;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
+
+class IntegrationCleanupExtension extends Extension {
+  /** @var EntityManager */
+  private $entityManager;
+
+  public static $events = [
+    Events::TEST_BEFORE => 'beforeTest',
+    Events::SUITE_BEFORE => 'beforeSuite',
+  ];
+
+  /** @var string */
+  private $deleteStatement;
+
+  public function beforeSuite(SuiteEvent $event) {
+    $this->entityManager = ContainerWrapper::getInstance()->get(EntityManager::class);
+
+    $this->deleteStatement = 'SET FOREIGN_KEY_CHECKS=0;';
+    foreach ($this->entityManager->getMetadataFactory()->getAllMetadata() as $metadata) {
+      $class = $metadata->getName();
+      $table = $metadata->getTableName();
+      $this->deleteStatement .= "DELETE FROM $table;";
+
+      // save plugin version to avoid triggering migrator and populator
+      if ($class === SettingEntity::class) {
+        $dbVersion = Env::$version;
+        $this->deleteStatement .= "
+          INSERT INTO $table (name, value, created_at, updated_at)
+          VALUES ('db_version', '$dbVersion', NOW(), NOW());
+        ";
+      }
+    }
+    $this->deleteStatement .= 'SET FOREIGN_KEY_CHECKS=1';
+  }
+
+  public function beforeTest(TestEvent $event) {
+    $this->entityManager->getConnection()->executeStatement($this->deleteStatement);
+  }
+}
--- a/mailpoet/tests/_support/IntegrationCleanupExtension.php
+++ b/mailpoet/tests/_support/IntegrationCleanupExtension.php
@@ -24,40 +24,28 @@ class IntegrationCleanupExtension extends Extension {
   private $deleteStatement;
 
   public function beforeSuite(SuiteEvent $event) {
-    global $wpdb;
-
     $this->entityManager = ContainerWrapper::getInstance()->get(EntityManager::class);
 
+    $mpPrefix = Env::$dbPrefix;
+    $tables = $this->entityManager->getConnection()->fetchFirstColumn("
+      SELECT table_name
+      FROM information_schema.tables
+      WHERE table_name LIKE '{$mpPrefix}%'
+      AND table_name != '{$mpPrefix}migrations'
+    ");
+
     $this->deleteStatement = 'SET FOREIGN_KEY_CHECKS=0;';
-
-    $automationTables = [
-      'mailpoet_automation_run_logs',
-      'mailpoet_automation_run_subjects',
-      'mailpoet_automation_runs',
-      'mailpoet_automation_triggers',
-      'mailpoet_automation_versions',
-      'mailpoet_automations',
-    ];
-
-    foreach ($automationTables as $automationTable) {
-      $fullTable = sprintf('%s%s', $wpdb->prefix, $automationTable);
-      $this->deleteStatement .= "DELETE FROM $fullTable;";
-    }
-
-    foreach ($this->entityManager->getMetadataFactory()->getAllMetadata() as $metadata) {
-      $class = $metadata->getName();
-      $table = $metadata->getTableName();
+    foreach ($tables as $table) {
       $this->deleteStatement .= "DELETE FROM $table;";
-
-      // save plugin version to avoid triggering migrator and populator
-      if ($class === SettingEntity::class) {
-        $dbVersion = Env::$version;
-        $this->deleteStatement .= "
-          INSERT INTO $table (name, value, created_at, updated_at)
-          VALUES ('db_version', '$dbVersion', NOW(), NOW());
-        ";
-      }
     }
+
+    // save plugin version to avoid triggering migrator and populator
+    $settingsTable = $this->entityManager->getMetadataFactory()->getMetadataFor(SettingEntity::class)->getTableName();
+    $dbVersion = Env::$version;
+    $this->deleteStatement .= "
+      INSERT INTO $settingsTable (name, value, created_at, updated_at)
+      VALUES ('db_version', '$dbVersion', NOW(), NOW());
+    ";
     $this->deleteStatement .= 'SET FOREIGN_KEY_CHECKS=1';
   }
 
--- a/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/PurchasedProductTest.php
+++ b/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/PurchasedProductTest.php
@@ -10,8 +10,6 @@ use MailPoet\AutomaticEmails\WooCommerce\WooCommerceStubs\OrderDetails;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\NewsletterOptionEntity;
 use MailPoet\Entities\NewsletterOptionFieldEntity;
-use MailPoet\Entities\ScheduledTaskEntity;
-use MailPoet\Entities\ScheduledTaskSubscriberEntity;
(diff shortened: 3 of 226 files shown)


JanJakes:
@johnolek This is awesome! A tiny test extension and so much cleanup as a result. Really looking forward to having this merged!

> In my testing, I found that simply deleting all the rows in all the tables was much faster than truncating the tables.

Interesting! Maybe it has something to do with resetting auto-increments, indexes, and stuff? Anyway, I think deletes are fine, at least they will force us not to rely on constant IDs in the tests 😊

> I couldn't figure out a way to truncate the tables in a single command.

You could try something like the following, but based on your measurements, I think it will still be slower than deletes:

```sql
SELECT 'TRUNCATE TABLE ' + table_name + ';'
FROM information_schema.tables
WHERE table_name LIKE '<our-table-prefix-here>%'
```

And you could try the same for the deletes to compare with your approach, but I doubt there will be any improvements, as what you are doing now is also just a single roundtrip to the DB.

How much work do you think is left? I see that `integration_test_base` is passing, so you might be close?

johnolek:
@JanJakes I do think I'm close, but it's often the last 5% that takes the most amount of time. I'm going to try to power through and finish it today if I can, but some of the tests seem rather tricky.

johnolek:
@JanJakes I believe this should be ready for final review. Thank you so much for all of your help with it! If you think the commits are too messy let me know and I can try to rebase/squash, but with the number of changes I'm worried that might be more of a pain than it's worth.

I would like to explore other low hanging fruit for speeding up these tests. The big thing that stands out to me is the amount of time we spend pulling docker images. It's about a minute for every single job. I tried enabling circleci's docker layer caching in bb5ae12a1f31564d3221f8f249da0a06083ff680, but it didn't seem to work. I'll file a followup ticket for this unless you know of an easy fix, or a reason why we shouldn't do this at all (I know we've had naming conflicts in the past, but I think we might be able to fix that by including more unique data in the container names like in the commit above).

Even though this has taken way longer than I had ever hoped/planned, please don't hesitate to bring up any issues/ideas for improvements, even small ones!

mailpoet/lib/DI/ContainerConfigurator.php
```diff
@@ -436,7 +436,7 @@ public function configure(ContainerBuilder $container) {
     $container->autowire(\MailPoet\Subscription\Captcha\CaptchaConstants::class)->setPublic(true);
     $container->autowire(\MailPoet\Subscription\CaptchaFormRenderer::class)->setPublic(true);
     $container->autowire(\MailPoet\Subscription\Captcha\CaptchaSession::class);
-    $container->autowire(\MailPoet\Subscription\Captcha\CaptchaRenderer::class);
+    $container->autowire(\MailPoet\Subscription\Captcha\CaptchaRenderer::class)->setPublic(true);
```
JanJakes, line 439: This looks a bit random, does it belong here?

mailpoet/tests/_support/IntegrationCleanupExtension.php
```diff
@@ -0,0 +1,67 @@
...
+    $this->deleteStatement = 'SET FOREIGN_KEY_CHECKS=0;';
+
+    $automationTables = [
+      'mailpoet_automation_run_logs',
+      'mailpoet_automation_run_subjects',
+      'mailpoet_automation_runs',
+      'mailpoet_automation_triggers',
+      'mailpoet_automation_versions',
+      'mailpoet_automations',
+    ];
```
JanJakes: When we add a new automation table, we'll certainly forget to add it here. Could we match everything starting with `mailpoet_automation`, maybe? But if we do that... then why not matching all that starts with `mailpoet_` (excluding `mailpoet_migrations`).

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/PurchasedProductTest.php
```diff
@@ -284,14 +282,7 @@ private function createWooSubscriber(): SubscriberEntity {
...
-    $this->truncateEntity(NewsletterEntity::class);
-    $this->truncateEntity(NewsletterOptionEntity::class);
-    $this->truncateEntity(NewsletterOptionFieldEntity::class);
-    $this->truncateEntity(SubscriberEntity::class);
-    $this->truncateEntity(SubscriberSegmentEntity::class);
-    $this->truncateEntity(SendingQueueEntity::class);
-    $this->truncateEntity(ScheduledTaskEntity::class);
-    $this->truncateEntity(ScheduledTaskSubscriberEntity::class);
+    parent::_after();
     WPFunctions::set(new WPFunctions);
```
JanJakes: Maybe we could add this logic to the bootstrap or cleanup extension? I see it repeated and the runtime cost of this will probably be close to nothing...

mailpoet/tests/integration/Automation/Integrations/MailPoet/Triggers/SomeoneSubscribesTriggerTest.php
```diff
@@ -71,6 +71,7 @@ public function dataForTestTriggeredByAutomationRun(): array {
   }
 
   public function _after() {
+    parent::_after();
     $segmentIds = $this->getSegmentIds(array_keys($this->segments));
     $this->segmentRepository->bulkDelete($segmentIds);
```
JanJakes: Is this needed now, when we delete everything?

mailpoet/tests/integration/Cron/ActionScheduler/Actions/DaemonRunTest.php
```diff
@@ -127,7 +126,5 @@ private function cleanup(): void {
     $wpdb->query('TRUNCATE ' . $actionsTable);
     $claimsTable = $wpdb->prefix . 'actionscheduler_claims';
     $wpdb->query('TRUNCATE ' . $claimsTable);
```
JanJakes, line 128: I wonder if we could try to include Action Scheduler tables in the cleanup extension 🤔

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
```diff
@@ -981,17 +974,7 @@ private function getScheduler(?SubscribersFinder $subscribersFinder = null): Sch
   }
 
   public function _after() {
+    parent::_after();
     Carbon::setTestNow();
```
JanJakes: Another candidate to be also extracted to bootstrap/cleanup extension?

mailpoet/tests/integration/Models/CustomFieldTest.php
```diff
@@ -145,6 +145,7 @@ public function testItCanHaveAValue() {
   }
 
   public function _after() {
+    parent::_after();
     CustomField::deleteMany();
     Subscriber::deleteMany();
     SubscriberCustomField::deleteMany();
```
JanJakes: I think these deletes are also not needed anymore.

mailpoet/tests/integration/Models/SubscriberCustomFieldTest.php
```diff
@@ -77,6 +77,7 @@ public function testItCanDeleteSubscriberRelations() {
   }
 
   public function _after() {
+    parent::_after();
     SubscriberCustomField::deleteMany();
```
JanJakes: This one doesn't seem to be needed anymore.

mailpoet/tests/integration/Models/SubscriberSegmentTest.php
```diff
@@ -336,6 +336,7 @@ public function testItCannotDeleteSubscriptionToWPAndWooCommerceSegments() {
   }
 
   public function _after() {
+    parent::_after();
     Segment::deleteMany();
     Subscriber::deleteMany();
     SubscriberSegment::deleteMany();
```
JanJakes: And these, I think, are not needed anymore.

mailpoet/tests/integration/Newsletter/ShortcodesTest.php
```diff
@@ -482,15 +482,11 @@ public function _createQueue(): SendingQueueEntity {
...
   }
 
   public function cleanup() {
-    $this->truncateEntity(NewsletterEntity::class);
-    $this->truncateEntity(SubscriberEntity::class);
-    $this->truncateEntity(SendingQueueEntity::class);
-    $this->truncateEntity(CustomFieldEntity::class);
-    $this->truncateEntity(SubscriberCustomFieldEntity::class);
     if ($this->wPPost) wp_delete_post($this->wPPost, true);
     if ($this->wPUser) wp_delete_user($this->wPUser->ID);
```
JanJakes, line 491: What if we also delete all posts and related metadata in the extension? As long as we append the deletes to the same long multi-query, I think it would be just a tiny cost...?

mailpoet/tests/integration/Newsletter/ViewInBrowser/ViewInBrowserControllerTest.php
```diff
@@ -237,10 +237,7 @@ public function testItResetsQueueForAutomaticEmailsInPreview() {
...
 
   public function _after() {
-    $this->truncateEntity(NewsletterEntity::class);
-    $this->truncateEntity(SubscriberEntity::class);
-    $this->truncateEntity(ScheduledTaskEntity::class);
-    $this->truncateEntity(SendingQueueEntity::class);
+    parent::_after();
     // reset WP user role
     $wpUser = wp_get_current_user();
     $wpUser->add_role('administrator');
```
JanJakes, line 243: And when at it, maybe even resetting the WP users table to some defaults can be useful. I'm sharing these thoughts as I see them, but I think we can extract all of that into a follow-up ticket.

mailpoet/tests/integration/Segments/SegmentSubscribersRepositoryTest.php
```diff
@@ -247,10 +247,6 @@ private function createDynamicSegmentEntity(string $role = 'editor'): SegmentEnt
   }
 
   private function cleanup() {
-    $this->truncateEntity(SegmentEntity::class);
-    $this->truncateEntity(SubscriberEntity::class);
-    $this->truncateEntity(SubscriberSegmentEntity::class);
-    $this->truncateEntity(DynamicSegmentFilterEntity::class);
     $this->clearSubscribersCountCache();
```
JanJakes, line 250: Where is the cache stored, is it a WP transient? Just thinking if we should add transients to our list of additional things to clean up. On the other hand, they may also stay in memory 🤔

mailpoet/tests/integration/Subscribers/ImportExport/Import/ImportTest.php
```diff
@@ -880,12 +879,7 @@ private function createSubscriber(
   }
 
   public function _after(): void {
-    $this->truncateEntity(SubscriberEntity::class);
-    $this->truncateEntity(SegmentEntity::class);
-    $this->truncateEntity(SubscriberSegmentEntity::class);
+    parent::_after();
     $this->truncateEntity(CustomFieldEntity::class);
```
JanJakes: Did the custom field entity truncation stay here by mistake?

JanJakes (commented):
Great job, this is really awesome! 👏 

> If you think the commits are too messy let me know and I can try to rebase/squash, but with the number of changes I'm worried that might be more of a pain than it's worth.

Yeah, no need to do something overly complex. There are some try-revert sequences, so we could maybe drop those, if reasonably easy.

> I would like to explore other low hanging fruit for speeding up these tests. The big thing that stands out to me is the amount of time we spend pulling docker images. It's about a minute for every single job. I tried enabling circleci's docker layer caching in https://github.com/mailpoet/mailpoet/commit/bb5ae12a1f31564d3221f8f249da0a06083ff680, but it didn't seem to work. I'll file a followup ticket for this unless you know of an easy fix, or a reason why we shouldn't do this at all (I know we've had naming conflicts in the past, but I think we might be able to fix that by including more unique data in the container names like in the commit above).

As for Docker layer caching, [there was a thread](https://a8c.slack.com/archives/C01H71ZLBGQ/p1667985836388249) pretty much deciding against it, as it's an expensive feature that brings less speed-up than one may think. Docker layer caching makes sense only when `docker build` or `docker compose build` is actually run on the CI, which I think, is something we don't do. For prebuilt images, I wouldn't expect any reasonable speed ups (the cache is AWS S3, the Docker hub image probably as well).

We can certainly try to optimize this in some ways, so a ticket may be useful in any case.

> Even though this has taken way longer than I had ever hoped/planned, please don't hesitate to bring up any issues/ideas for improvements, even small ones!

I'm dropping all the ideas that pop up when looking at the PR. Of course, we can extract most of them (if not all) to follow-up tickets. If my support load is not high, I may try to look at some of those today. If I touch anything, I'll let you know here and/or on Slack.

mailpoet/tests/integration/_bootstrap.php
```diff
@@ -98,8 +98,9 @@ public function setUp(): void {
     $this->diContainer = ContainerWrapper::getInstance(WP_DEBUG);
     $this->connection = $this->diContainer->get(Connection::class);
     $this->entityManager = $this->diContainer->get(EntityManager::class);
-    // Cleanup scheduled tasks from previous tests and switch cron to Linux method
+    // Cleanup scheduled tasks from previous tests
     $this->truncateEntity(ScheduledTaskEntity::class);
```
JanJakes: This was removed and then added back. Is there a reason to keep this? I think it should be safe to remove.

mailpoet/tests/_support/IntegrationCleanupExtension.php
JanJakes: Done in: https://github.com/mailpoet/mailpoet/pull/4818/commits/03396e48d5e8cc4c16c408ae9852e85c6ddda86f

mailpoet/tests/integration/Automation/Integrations/MailPoet/Triggers/SomeoneSubscribesTriggerTest.php
JanJakes: Done in: https://github.com/mailpoet/mailpoet/pull/4818/commits/557577a432821dcc8238308bb4ff7b349b3907a4

mailpoet/tests/integration/Models/CustomFieldTest.php
JanJakes: Done in: https://github.com/mailpoet/mailpoet/pull/4818/commits/557577a432821dcc8238308bb4ff7b349b3907a4

mailpoet/tests/integration/Models/SubscriberCustomFieldTest.php
JanJakes: Done in: https://github.com/mailpoet/mailpoet/pull/4818/commits/557577a432821dcc8238308bb4ff7b349b3907a4

mailpoet/tests/integration/Models/SubscriberSegmentTest.php
JanJakes: Done in: https://github.com/mailpoet/mailpoet/pull/4818/commits/557577a432821dcc8238308bb4ff7b349b3907a4

mailpoet/tests/integration/Segments/SegmentSubscribersRepositoryTest.php
JanJakes: Ahh, we call that already in `_bootstrap` so removed: https://github.com/mailpoet/mailpoet/pull/4818/commits/557577a432821dcc8238308bb4ff7b349b3907a4

mailpoet/tests/integration/Subscribers/ImportExport/Import/ImportTest.php
JanJakes: Removed (https://github.com/mailpoet/mailpoet/pull/4818/commits/557577a432821dcc8238308bb4ff7b349b3907a4) and fixed a related bug: https://github.com/mailpoet/mailpoet/pull/4818/commits/550ac60c5e650fe496d1a8084987df28d81e9d65

mailpoet/tests/integration/_bootstrap.php
JanJakes: Removed: https://github.com/mailpoet/mailpoet/pull/4818/commits/557577a432821dcc8238308bb4ff7b349b3907a4

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/PurchasedProductTest.php
JanJakes: Done in: https://github.com/mailpoet/mailpoet/pull/4818/commits/0ef10c0348e845a09939afe795299897ae790897

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
JanJakes: Done in: https://github.com/mailpoet/mailpoet/pull/4818/commits/5bffd812189393f85ebf2985aaf32c8fe1134286

JanJakes:
@johnolek I had some time this morning, so I did the following:

1. I dropped all the try-revert commits and did a fixup of some like fix of Prettier formatting. I repushed the branch without any other changes. All my additions are added as new commits.
2. Cleaning up all MP tables (except for migrations) by prefix to avoid forgetting this for new automation tables.
3. Removed some forgotten test cleanup, and fixed a test bug that the removal of one more `TRUNCATE` revealed.
4. Extracted `WPFunctions` and `Carbon` reset to `_bootstrap.php`.

What I suggested but _didn't do_ is:
1. Clean up also WP tables (and related in-memory data like transients, if needed).
2. Clean up also Woo tables.
3. Clean up also Action Scheduler tables.

I think we can extract these to a new ticket. We need to be careful not to introduce slowdowns when adding more cleanups. The current speedup looks just great!

I'm hitting approve from my side, but assigning it back to you, @johnolek, so you can verify my changes. Feel free to pass this on to QA, if it looks good to you.

JanJakes (dismissed):

johnolek:
@JanJakes Thank you so much for your help with this! I think your changes all look great, and I definitely appreciate you taking care of some of the git cleanup.

I pushed one minor naming change (b3fc97fec814bbe423d59840caca18e1c97ef23e) now that the SQL we're executing isn't just delete statements, otherwise I think this looks good to go! I'll write a followup ticket for the other items you mentioned.

Aschepikov (approved):
GJ! @johnolek I made a regression for export personal data. I see there are a lot of changes in the pr. Can I merge it, or do you need to do something before it?

johnolek:
@Aschepikov Thanks for the review! I would like to publish a P2 before merging, so I can handle the merge.

PR #4824, opened 2023-03-30
Revert "Add order status change trigger [MAILPOET-4935]"

Aschepikov:
Reverts mailpoet/mailpoet#4794


websupporter:
Dont revert @Aschepikov. Will provide a PR which will hide the trigger.

PR #4828, opened 2023-03-31
Automation filters [MAILPOET-4946] [MAILPOET-4419] [MAILPOET-5002]

JanJakes:
## Description

This PR:
1. Finalizes subject `fields`.
2. Adds support for field `filters` on triggers (definition, saving, validation, UI, filtering logic).
3. Implements two filters (`string` and `enum_array`).
4. Adds `Subscriber segments` filter.
5. Adds test coverage.
6. Adds a few small fixes for issues encountered on the way.

Note that filters will work in the free plugin itself, but it's only premium that allows adding, removing, or modifying them. This is consistent with how we treat automation steps.

## Code review notes

I opted for adding a new `filters` step field. Does that make sense? At the moment, we use the filters only for triggers. I'm guessing we could reuse the same `filters` field as well for conditional steps, for instance.

## QA notes

Please test with and without the premium plugin and make sure that the subscriber segment filter works with all condition types. Please also check that updating a site with already existing and active automations works.

**Please, ping @JanJakes before merging.**

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/704

## Linked tickets

[MAILPOET-4946] (Filters, Subscriber segments filter)
[MAILPOET-4419] (String filter)
[MAILPOET-5002] (Enum array filter)

## After-merge notes

_N/A_

[MAILPOET-4946]: https://mailpoet.atlassian.net/browse/MAILPOET-4946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4419]: https://mailpoet.atlassian.net/browse/MAILPOET-4419
[MAILPOET-5002]: https://mailpoet.atlassian.net/browse/MAILPOET-5002

### DIFF
--- a/mailpoet/lib/Automation/Engine/Control/FilterHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/FilterHandler.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Control;
+
+use MailPoet\Automation\Engine\Data\Filter as FilterData;
+use MailPoet\Automation\Engine\Data\StepRunArgs;
+use MailPoet\Automation\Engine\Exceptions;
+use MailPoet\Automation\Engine\Registry;
+
+class FilterHandler {
+  /** @var Registry */
+  private $registry;
+
+  public function __construct(
+    Registry $registry
+  ) {
+    $this->registry = $registry;
+  }
+
+  public function matchesFilters(StepRunArgs $args): bool {
+    $filters = $args->getStep()->getFilters();
+    foreach ($filters as $filter) {
+      $value = $args->getFieldValue($filter->getFieldKey());
+      if (!$this->matchesFilter($filter, $value)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /** @param mixed $value */
+  private function matchesFilter(FilterData $data, $value): bool {
+    $filter = $this->registry->getFilter($data->getFieldType());
+    if (!$filter) {
+      throw Exceptions::filterNotFound($data->getFieldType());
+    }
+    return $filter->matches($data, $value);
+  }
+}
--- a/mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
@@ -28,6 +28,9 @@ class TriggerHandler {
   /** @var SubjectTransformerHandler */
   private $subjectTransformerHandler;
 
+  /** @var FilterHandler */
+  private $filterHandler;
+
   /** @var WordPress */
   private $wordPress;
 
@@ -37,6 +40,7 @@ class TriggerHandler {
     AutomationRunStorage $automationRunStorage,
     SubjectLoader $subjectLoader,
     SubjectTransformerHandler $subjectTransformerHandler,
+    FilterHandler $filterHandler,
     WordPress $wordPress
   ) {
     $this->actionScheduler = $actionScheduler;
@@ -44,6 +48,7 @@ class TriggerHandler {
     $this->automationRunStorage = $automationRunStorage;
     $this->subjectLoader = $subjectLoader;
     $this->subjectTransformerHandler = $subjectTransformerHandler;
+    $this->filterHandler = $filterHandler;
     $this->wordPress = $wordPress;
   }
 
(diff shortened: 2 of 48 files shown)


mailpoet/lib/Automation/Engine/Control/FilterHandler.php
```diff
@@ -0,0 +1,39 @@
...
+  ) {
+    $this->registry = $registry;
+  }
+
+  public function matchesFilters(StepRunArgs $args): bool {
+    $filters = $args->getStep()->getFilters();
+    foreach ($filters as $filter) {
+      $value = $args->getFieldValue($filter->getFieldKey());
+      if (!$this->matchesFilter($filter, $value)) {
+        return false;
```
websupporter, line 25: So its basically an `AND` relation between the different filters, right? If one filter does not match, we do not have a Match.

mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
```diff
@@ -11,46 +11,45 @@
 use MailPoet\Automation\Engine\Storage\AutomationRunStorage;
 use MailPoet\Automation\Engine\Storage\AutomationStorage;
 use MailPoet\Automation\Engine\WordPress;
```
websupporter, line 13: Ouch! Was not aware of this WordPress wrapper. Makes total sense. I will go at some point through the code base and check where I used `WP\Functions` instead.

mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
```diff
@@ -59,24 +58,30 @@ public function initialize(): void {
 
   /** @param Subject[] $subjects */
   public function processTrigger(Trigger $trigger, array $subjects): void {
-    $subjects = $this->subjectTransformerHandler->getAllSubjects($subjects);
     $automations = $this->automationStorage->getActiveAutomationsByTrigger($trigger);
+    if (!$automations) {
```
websupporter, line 62: :clap: That's a nice change! thanks for the clean up!

mailpoet/lib/Automation/Engine/Data/StepRunArgs.php
```diff
@@ -39,6 +46,11 @@ public function __construct(
       $key = $subject->getKey();
       $this->subjectEntries[$key] = array_merge($this->subjectEntries[$key] ?? [], [$entry]);
       $this->subjectKeyClassMap[get_class($subject)] = $key;
+
+      foreach ($subject->getFields() as $field) {
+        $this->fields[$field->getKey()] = $field;
+        $this->fieldToSubjectMap[$field->getKey()] = $key;
```
websupporter, line 52: should we account for the possibility that different subjects could provide the same field?

also, what i am just thinking: currently our code base can't provide for that (and maybe it shouldnt) but how would we handle `SubscriberSubject[]` where one subject might match and the other does not?

Just as a thought. Personally I am leaning towards not allowing `SubscriberSubject[]`, although sometimes it would be nice like in the case of `ProductSubject[]`

websupporter (dismissed):
lgtm. Some remarks, but only for better understanding. Thank you for this PR

mailpoet/lib/Automation/Engine/Control/FilterHandler.php
JanJakes: Yeah, but we already have [a ticket to allow choosing between AND/OR](https://mailpoet.atlassian.net/browse/MAILPOET-5218) and we were also discussing allowing for multiple filter "groups" at some point, to enable any combinations.

mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
JanJakes: This one was added in the first prototypes when we were very careful about the "independence" of the automations on the current MP codebase. Now, I'm not sure if it makes sense to duplicate this. Maybe we can reuse the MP one. If the codebase ever needs to somehow leave MailPoet, that would be easy to tackle.

mailpoet/lib/Automation/Engine/Data/StepRunArgs.php
JanJakes: > should we account for the possibility that different subjects could provide the same field?

That's currently [disabled by the registry](https://github.com/mailpoet/mailpoet/pull/4828/files#diff-7130f3edb9bfb25c69a217eda9f942040914ebd4b54b837f7f9f1f1bbb47c35dR83). I guess it's just the easiest way to do it now, and if we ever run into the need of having the same field from multiple sources, then we'll need to work on it. Now I'm not sure what would be all the implications (when the values are inconsistent, etc.).

JanJakes: > also, what i am just thinking: currently our code base can't provide for that (and maybe it shouldnt) but how would we handle SubscriberSubject[] where one subject might match and the other does not?

We don't allow multiple subjects of the same type at all currently. I feel like enabling that would be quite a challenge on multiple levels. The automation format doesn't limit that (subjects field is an array, not key-value), but I'm not sure if we should ever allow that. Maybe it will be better to define some "multi-subjects" instead. Something like `SubscriberGroupSubject` or a more generic `SubjectGroup`.

Aschepikov (approved):
All main flows checked gj!

PR #4820, opened 2023-03-29
Import sql to local Docker environment

veljkho:
## Description

In this PR, we're going to add SQL with `100k subscribers`, `5k newsletters` to a local test env.

Note: This PR is final PR for the project k6 implementation. There are some small changes not related to the importing SQL to Docker env, but small changes such as:

Additionally, added a new test just to bypass the onboarding wizard for the first time local execution: `onboardingWizard.js`

I also added taking screenshots in tests. Before, after, or in the middle. Now the screenshots will be included in the artifacts on CircleCI or `_screenshots` folder locally.

Refactored all codes to include `await`. This is very normal in Playwright testing of the E2E suite (on Solaris' side).

There's 1min to import data from the `data.sql` file to local env. I think this is fine for now, until we figure out how to better optimize it. The test run took 3min to complete.

I set the thresholds to pass always, but it's not literally always... If something's wrong it will always fail in the PR checks. The values set are very high but not too much.

## Code review notes

Sorry for the messy commit listing, but please focus on the `file changed`.

If you'd like to test it locally, feel free to execute:

`./do test:performance --scenario=pullrequests`

## Linked tickets

[MAILPOET-4950]

[MAILPOET-4950]: https://mailpoet.atlassian.net/browse/MAILPOET-4950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4835, opened 2023-04-05
Integrate k6 suite with CircleCI and stream to k6 cloud

veljkho:
## Description

In this PR, I'm going to connect local k6 execution along with CircleCI pull requests/nightlytests execution and streamline it to k6 cloud account for further better visualisation and analysis, trends and etc.

## Code review notes

If you'd like to access k6 cloud, please let me know so I can invite you to sign up there and be part of MailPoet project on k6 cloud, where you could see the k6 test run from this PR's build on CircleCI.

## Linked PRs

Wait for #4820 before you merge this one

## Linked tickets

[MAILPOET-5196]

[MAILPOET-5196]: https://mailpoet.atlassian.net/browse/MAILPOET-5196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -300,9 +300,22 @@ class RoboFile extends \Robo\Tasks {
     return $this->runTestsInContainer($opts);
   }
 
-  public function testPerformance($path = null, $opts = ['url' => null, 'pw' => null, 'head' => false, 'scenario' => null, 'name' => null]) {
+  public function testPerformance($path = null, $opts = ['url' => null, 'pw' => null, 'head' => false, 'scenario' => null]) {
     $dir = __DIR__;
-    return $this->collectionBuilder()
+    if ((getenv('K6_CLOUD_TOKEN')) === false) {
+      return $this->collectionBuilder()
+      ->addCode([$this, 'testPerformanceSetup'])
+      ->taskExec("php $dir/tools/k6.php")
+      ->arg('run')
+      ->option('env', 'K6_BROWSER_ENABLED=1')
+      ->option('env', 'URL=' . $opts['url'])
+      ->option('env', 'PW=' . $opts['pw'])
+      ->option('env', 'HEADLESS=' . ($opts['head'] ? 'false' : 'true'))
+      ->option('env', 'SCENARIO=' . $opts['scenario'])
+      ->arg($path ?? "$dir/tests/performance/scenarios.js")
+      ->dir($dir)->run();
+    } else {
+      return $this->collectionBuilder()
       ->addCode([$this, 'testPerformanceSetup'])
       ->taskExec("php $dir/tools/k6.php")
       ->arg('run')
@@ -313,10 +326,11 @@ class RoboFile extends \Robo\Tasks {
       ->option('env', 'SCENARIO=' . $opts['scenario'])
       ->option('env', 'K6_CLOUD_TOKEN=' . getenv('K6_CLOUD_TOKEN'))
       ->option('env', 'K6_CLOUD_ID=' . getenv('K6_CLOUD_ID'))
-      ->option('env', 'K6_PROJECT_NAME=' . $opts['name'])
+      ->option('env', 'K6_PROJECT_NAME=' . $opts['scenario'])
       ->option('out', 'cloud')
       ->arg($path ?? "$dir/tests/performance/scenarios.js")
       ->dir($dir)->run();
+    }
   }
 
   public function testPerformanceSetup() {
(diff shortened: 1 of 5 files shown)


mailpoet/RoboFile.php
```diff
@@ -311,6 +311,10 @@ public function testPerformance($path = null, $opts = ['url' => null, 'pw' => nu
       ->option('env', 'PW=' . $opts['pw'])
       ->option('env', 'HEADLESS=' . ($opts['head'] ? 'false' : 'true'))
       ->option('env', 'SCENARIO=' . $opts['scenario'])
+      ->option('env', 'K6_CLOUD_TOKEN=' . getenv('K6_CLOUD_TOKEN'))
+      ->option('env', 'K6_CLOUD_ID=' . getenv('K6_CLOUD_ID'))
+      ->option('env', 'K6_PROJECT_NAME=' . $opts['name'])
```
JanJakes: Can we just use `$opts['scenario']` here and remove the `name` parameter?

mailpoet/RoboFile.php
```diff
@@ -311,6 +311,10 @@ public function testPerformance($path = null, $opts = ['url' => null, 'pw' => nu
...
+      ->option('out', 'cloud')
```
JanJakes, line 330: What will happen when we run this locally? In that case, will it also try to log to the cloud? Or if I have no credentials, will that just skip it or something?

JanJakes (commented):
Could you pls fetch the latest `trunk` and rebase this on `trunk`? It contains many commits that are probably from a different PR. Thanks!

mailpoet/RoboFile.php
veljkho: I will try that

veljkho:
> Could you pls fetch the latest `trunk` and rebase this on `trunk`? It contains many commits that are probably from a different PR. Thanks!

I already did that but nothing to push after that, 0 changes. @JanJakes

mailpoet/RoboFile.php
veljkho: There will be error in the output that you need to have authorization token. The token should be placed in the `.env`, currently present in the secret store along with all other vars.

veljkho: And, if you want to run locally you should probably want to have it under different project name. So, you would add --name=localjanjakes for example.. otherwise, it will be ran under pullrequests. I will add this info in ReadMe file in a separate PR to update ReadMe with all other required info.

mailpoet/RoboFile.php
veljkho: I just tried it locally and it won't work.. especially if we want to add some custom naming, like I mentioned below `--name=localjanjakes` for example.. Let's leave it then as-is so it's customizable. @JanJakes [Here's mine test](https://d.pr/i/Bx0qUq)

mailpoet/RoboFile.php
JanJakes: @veljkho This is not good. The defaults should not be difficult to use, and I think there's no reason to log local runs to the cloud.

Let's maybe try something like "if `K6_CLOUD_ID` env var doesn't exist, don't log anything`, or similar setup, where by default, it would not use the cloud at all.

mailpoet/RoboFile.php
JanJakes: Having to use a custom `--name` locally would be very unintuitive. See also my comment https://github.com/mailpoet/mailpoet/pull/4835#discussion_r1160590017. I still think we should remove this option to keep things as simple as possible.

JanJakes:
> Could you pls fetch the latest trunk and rebase this on trunk? It contains many commits that are probably from a different PR. Thanks!

@veljkho I think that might be due to the previous `Merge branch 'trunk' into k6/output-test-results-in-k6-cloud`. Pls always use rebase instead, as in MP we don't use merge commits.

mailpoet/RoboFile.php
veljkho: You were right, thanks for suggestion!

veljkho:
> > Could you pls fetch the latest trunk and rebase this on trunk? It contains many commits that are probably from a different PR. Thanks!
> 
> @veljkho I think that might be due to the previous `Merge branch 'trunk' into k6/output-test-results-in-k6-cloud`. Pls always use rebase instead, as in MP we don't use merge commits.

I created a new branch from trunk and included changes from this PR, now it should be good. @JanJakes

mailpoet/RoboFile.php
veljkho: Sure, removed it 👍

JanJakes (approved):

JanJakes:
@veljkho Thanks for the changes! 👏

PR #4836, opened 2023-04-06
Abandoned cart trigger

websupporter:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

You might need to change the version [here](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/mailpoet.php#L23) manually to a higher version so that some new options can be set, which will be set in the database when you update (= the db version !== the current version)

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/708

## Linked tickets

_N/A_

## Merge notes

Should not be merged before #4796


mailpoet/lib/AutomaticEmails/WooCommerce/Events/AbandonedCart.php
```diff
@@ -165,11 +172,13 @@ private function scheduleAbandonedCartEmail(array $cartProductIds = []) {
       return;
     }
 
+    $this->wp->doAction(self::HOOK_SCHEDULE, $subscriber, $cartProductIds);
```
JanJakes, line 175: The updates to this class confused me for a while. It says `scheduleAbandonedCartEmail`, so I'm like `Do we need an old abandoned cart email set up to make this work?`. But it turns out this rather "maybe" schedules an email, if there is one.

I think it may deserve a small refactoring — we could move all the subscriber fetch & fire hook logic somewhere to `lib/WooCommerce` and then use those hooks here and in automations.

mailpoet/lib/Automation/Integrations/WooCommerce/Subjects/ProductsSubject.php
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public function getArgsSchema(): ObjectSchema {
+    return Builder::object([
+      'product_ids' => Builder::array(Builder::integer())->required(),
+    ]);
+  }
+
+  public function getPayload(SubjectData $subjectData): Payload {
+    $products = array_filter(array_map(
```
JanJakes: Since the array goes somewhere outside (to the payload), it's safer to wrap this with `array_values` or use a foreach loop.

mailpoet/lib/Automation/Integrations/WooCommerce/Subjects/ProductsSubject.php
```diff
@@ -0,0 +1,59 @@
...
+      function ($productId) {
+        return $this->woocommerceHelper->wcGetProduct($productId);
+      },
+      $subjectData->getArgs()['product_ids']
+    ));
+
+    // Question: What to do when the product was not found?
```
JanJakes: You mean when _none of the products_ was found... that's an edge case in which sending an abandoned cart email wouldn't make much sense 🤔 Hmm, so either we can keep the exception here, or we can let it pass and check that we have some products before sending the abandoned cart email.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -167,18 +181,51 @@ public function saveEmailSettings(Step $step, Automation $automation): void {
...
-    );
+    }
+
+    $optionField = $this->newsletterOptionFieldsRepository->findOneBy([
+      'name' => $optionName,
+      'newsletterType' => $newsletter->getType(),
+    ]);
+
+    if (!$optionField) {
+      $optionField = new NewsletterOptionFieldEntity();
```
JanJakes: Newsletter option fields are not meant to be used like this. The fields tell you "newsletter of type X has options A, B, C", they shouldn't be created dynamically. In fact, they only ever update [in Populator](https://github.com/mailpoet/mailpoet/blob/a9d89ff514ade08f1115b26fe0647549ae085f80/mailpoet/lib/Config/Populator.php#L375).

But what now... should `automation` emails have all the possible option fields that may be needed, depending on how the send email action is used? I'm not sure. I think it would make sense to rethink this for automations, but that could also get too complex. What are the particular fields you're adding used for? The editor somehow adjusts based on them?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -158,13 +160,30 @@ public function run(StepRunArgs $args): void {
...
+
+    $productSubject = $args->getSinglePayloadByClass(ProductsPayload::class);
+    $cartProductIds = array_map(
+      function(\WC_Product $product) {
+        return $product->get_id();
+      },
+      $productSubject->getProducts()
+    );
+
+    return [AbandonedCart::TASK_META_NAME => $cartProductIds];
```
JanJakes: I suppose this is how we're passing the product IDs today in the "legacy" abandoned cart emails, right? Then it makes sense, I guess. I think, ideally, automation emails could be aware of subjects/fields. But that sounds like a big refactor for the future.

mailpoet/lib/Automation/Integrations/WooCommerce/Subjects/ProductsSubject.php
```diff
@@ -0,0 +1,59 @@
...
+use MailPoet\Automation\Engine\Integration\Subject;
+use MailPoet\Automation\Integrations\WooCommerce\Payloads\ProductsPayload;
+use MailPoet\Validator\Builder;
+use MailPoet\Validator\Schema\ObjectSchema;
+use MailPoet\WooCommerce\Helper as WooCommerceHelper;
+
+/**
+ * @implements Subject<ProductsPayload>
+ */
+class ProductsSubject implements Subject {
```
JanJakes: Do you think the `product_ids` can be in the abandoned cart subject? Is there any reason to have this in two different subjects?

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
websupporter: Yes. It's is the condition on which the abandoned cart block shows up in the editor.

mailpoet/lib/Automation/Integrations/WooCommerce/Subjects/ProductsSubject.php
websupporter: consolidating discussion here: https://github.com/mailpoet/mailpoet-premium/pull/708#discussion_r1165032856

mailpoet/lib/AutomaticEmails/WooCommerce/Events/AbandonedCart.php
websupporter: Will open a follow up ticket.

websupporter:
thanks for your review! let me know what you think. once its okay, i will rebase @JanJakes

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/abandoned-cart/index.tsx
```diff
@@ -17,7 +17,7 @@ export const step: StepType = {
   title: () => __('Subscriber abandons cart', 'mailpoet-premium'),
   description: () =>
     __(
-      'Start the automation when logged-in subscribers who have items in their shopping carts leave your website for more than 30 minutes without checking out.',
+      'Start the automation when a subscriber who has items in the shopping cart leave your website without checking out.',
```
JanJakes: In this case, we need to also change `leave` -> `leaves`.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
JanJakes: @websupporter I think we can use them, but we should set the new fields [in the Populator](https://github.com/mailpoet/mailpoet/blob/a9d89ff514ade08f1115b26fe0647549ae085f80/mailpoet/lib/Config/Populator.php#L375) and here in the subject set only the option values. Otherwise, sending an email after an abandoned cart trigger will have the side effect that suddenly, automation emails will have these new fields. So it's better to define them all in the same place upfront.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/abandoned-cart/index.tsx
websupporter: [e2e86a4](https://github.com/mailpoet/mailpoet/pull/4836/commits/e2e86a4d215eaf7c8357270681ac35ffd1fdbdd5)

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
websupporter: [1322790](https://github.com/mailpoet/mailpoet/pull/4836/commits/1322790356ed5684cceb57b05161cb15be42830a)

Not sure about the change.
1. I added them to the populator
2. The newsletter gets created via XHR request in edit_newsletter.tsx. I need to create those options there already and decided to "introduce" `group:automation` and `event:automation`. Empty strings did not work
3. The SendEmailAction toggles now between those and the woocommerce specific ones.

websupporter:
@JanJakes thanks for the review and clarifying how options are supposed to work. Please have another look. Thank you :)

JanJakes (commented):
Passing this back as we discussed the options/option fields on Slack.

JanJakes (approved):

Aschepikov (approved):
GJ!

PR #4840, opened 2023-04-07
Fix multisite failure [MAILPOET-5145]

johnolek:
## Description

This PR fixes an issue caused by #4818 where a test started failing in multisite. It also updates a couple of tests that weren't included in the original PR.

## Code review notes

_N/A_

## QA notes

No QA needed, this only affects tests

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_


brezocordero (approved):
I hadn't run the whole integration suite locally for a long time, it is so much better now!
I tested that the failure is fixed and the code looks good.

PR #4842, opened 2023-04-10
Hide the brand settings if the site theme is not a block theme [MAILPOET-5194]

samnajian:
## Description
This PR disables the brand settings on the sidebar when the theme is not a block theme.

## Linked tickets
[MAILPOET-5194]

[MAILPOET-5194]: https://mailpoet.atlassian.net/browse/MAILPOET-5194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/views/newsletter/editor.html
+++ b/mailpoet/views/newsletter/editor.html
@@ -376,6 +376,14 @@
   var mailpoet_installed_at = '<%= settings.installed_at %>';
 </script>
 
+<% if brand_styles.available %>
+<script type="text/javascript">
+  var mailpoet_brand_styles = {
+    available: true
+  };
+</script>
+<% endif %>
+
 <% endblock %>
 
 <% block after_css %>
--- a/mailpoet/assets/js/src/global.d.ts
+++ b/mailpoet/assets/js/src/global.d.ts
@@ -236,4 +236,7 @@ interface Window {
     id: string;
     subject: string;
   }>;
+  mailpoet_brand_styles?: {
+    available: boolean;
+  };
 }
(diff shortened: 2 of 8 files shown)


mailpoet/views/newsletter/editor.html
```diff
@@ -376,6 +376,14 @@ <h1 style="display:none"><%= __('Newsletter Editor') %></h1>
...
 </script>
 
+<% if brand_styles.available %>
+<script type="text/javascript">
+  var mailpoet_brand_styles = {
+    available: true
+  };
+</script>
+<% endif %>
+
```
costasovo, line 386: Maybe you could always pass the brand_styles object to JS. If you do so you don't have to later in JS/TS handle situations when global `mailpoet_brand_styles` is not defined.

`var mailpoet_brand_style = <%= json_encode(homepage) %>;`

costasovo (commented):
Looks good. I have just one suggestion that might make the TS code a bit simpler.

mailpoet/views/newsletter/editor.html
samnajian: You're right about the TS part, but since we'll probably add more settings, thought not adding the object to the DOM at all when the brand styles are not available would be a better approach. But I'm also fine with your approach as we can optimize later.

costasovo: Aha. I see. It makes sense. With my approach, all the other settings would need to be optional or in an optional nested subobject. Feel free to keep it as is or change it as you decide.

costasovo (dismissed):

Aschepikov:
Hi @samnajian  there are some conflicts can you resolve them pls

samnajian:
@Aschepikov latest trunk is merged in.

Aschepikov (approved):
GJ!

PR #4846, opened 2023-04-11
Change 'score' label to 'engagement score' [MAILPOET-4995]

johnolek:
[MAILPOET-4995](https://mailpoet.atlassian.net/browse/MAILPOET-4995)

## Description

This changes the 'score' label to 'engagement score' when creating a segment.

Before:
image

After:
image

## Code review notes
I decided not to update the placeholder text because "engagement score" would get cut off with the input at its current size and I didn't want to change the size just for the placeholder text. It also seems fine to use "score" as a shorthand since the label itself has the full string. Please let me know if you disagree with this approach or can think of any potential issues.

I updated the existing translation instead of adding a new one. Could that cause any problems with sites running in other languages?

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

[MAILPOET-4995]: https://mailpoet.atlassian.net/browse/MAILPOET-4995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_opitions.ts
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber_opitions.ts
@@ -3,13 +3,13 @@ import { SegmentTypes, SubscriberActionTypes } from '../types';
 
 export const SubscriberSegmentOptions = [
   {
-    value: SubscriberActionTypes.MAILPOET_CUSTOM_FIELD,
-    label: MailPoet.I18n.t('mailpoetCustomField'),
+    value: SubscriberActionTypes.SUBSCRIBER_SCORE,
+    label: MailPoet.I18n.t('subscriberScore'),
     group: SegmentTypes.WordPressRole,
   },
   {
-    value: SubscriberActionTypes.SUBSCRIBER_SCORE,
-    label: MailPoet.I18n.t('subscriberScore'),
+    value: SubscriberActionTypes.MAILPOET_CUSTOM_FIELD,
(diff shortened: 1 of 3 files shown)


websupporter (commented):
> It also seems fine to use "score" as a shorthand since the label itself has the full string.

Yes, I like that.

> I updated the existing translation instead of adding a new one. Could that cause any problems with sites running in other languages?

No I don't think so. I think of it rather as "You replaced the old with a new one" instead of updated.


From the Jira ticket:
> Change the position within the “Subscriber” group to be sorted alphabetically.

Would that mean, we want to sort `SubscriberSegmentOptions` dynamically, as the translations might change the order?

Something like that:

```
[{"label":"b"}, {"label":"a"}].sort( (a,b) => a.label.localeCompare(b.label) )
```

johnolek:
@websupporter Great point about sorting in different languages! I think maybe we could make that into a separate ticket and apply it to dynamically sorting all the different sections? Or if you prefer I'm happy to make the change as part of this ticket.

websupporter:
Different ticket to catch all sections makes sense to me.

websupporter (approved):
lgtm

johnolek:
FYI @websupporter I created https://mailpoet.atlassian.net/browse/MAILPOET-5229 for dynamically sorting filter names in all sections. Thanks for the review!

Aschepikov (approved):
GJ!

PR #4838, opened 2023-04-06
Consistent statuses for WooCommerce purchase filters [MAILPOET-5083]

johnolek:
## Description

This PR updates all of the dynamic segment filters that look at WooCommerce orders to use the same order statuses, specifically only including completed and processing orders.

I used this as an opportunity to clean up some of the test code. There was a lot of duplication in how we were checking for emails that matched a filter before, so I extracted that logic to a helper.

## Code review notes

_N/A_

## QA notes

Please carefully test dynamic segments related to Woo. Test also combinations of multiple filters.
## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5083](https://mailpoet.atlassian.net/browse/MAILPOET-5083)

## After-merge notes

_N/A_

[MAILPOET-5083]: https://mailpoet.atlassian.net/browse/MAILPOET-5083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4848, opened 2023-04-12
Fix issues related to purchase date filter [MAILPOET-5220]

johnolek:
## Description

This PR fixes a few issues related to the new purchase date dynamic filter:

1. There should no longer be an error when switching from a purchase date filter to another woo commerce filter
2. The "single order value" filter will no longer be considered valid when it doesn't have all of its values filled in, which is currently possible when switching from "subscribed date" to "single order value".

## Code review notes

I ended up needing to do some refactoring to be able to re-use the shared `DateFields`. The original issue was coming from the fact that `DateFields` calls its own React hooks, and it was being included inside a conditional, causing `WooCommerceFields` to call a variable number of hooks inside the same function on subsequent calls. `WooCommerceFields` now operates more like `SubscriberFields` (see [here](https://github.com/mailpoet/mailpoet/blob/1ca06b615f41b9f7f95ccfd3c57c46a52dcd2677/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber.tsx#L54-L61)), which I think is better in general.

## QA notes

Please test all of the WooCommerce fields, as the refactoring could have affected any of them.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/date_fields.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/date_fields.tsx
@@ -144,7 +144,7 @@ export function DateFields({ filterIndex }: Props): JSX.Element {
   );
 }
 
-export function dateFieldValidator(formItems: DateFormItem): boolean {
+export function validateDateField(formItems: DateFormItem): boolean {
   if (!formItems.operator || !formItems.value) {
     return false;
   }
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/subscriber.tsx
@@ -7,7 +7,7 @@ import {
   SubscriberScoreFields,
   validateSubscriberScore,
 } from './subscriber_score';
-import { DateFields, dateFieldValidator, DateOperator } from './date_fields';
+import { DateFields, DateOperator, validateDateField } from './date_fields';
 import {
   MailPoetCustomFields,
   validateMailPoetCustomField,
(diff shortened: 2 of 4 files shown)


johnolek:
@costasovo I requested your review of this since you were just looking at the changes in #4838. Please let me know if you can think of any possible improvements, or more idiomatic ways of doing things in React.

costasovo (approved):
@johnolek Great job debugging and finding the issue!

> Please let me know if you can think of any possible improvements, or more idiomatic ways of doing things in React.

Refactoring the Woo-related filter forms to separate functions was a great idea. It improves the readability of the code, and we got rid of a bunch of if statements that could lead to a similar issue. I think we could even move the forms for specific filters to separate files.
If you think it is a good idea to move them, feel free to do so.
I'm approving the PR and will assign it back to you. So you can decide to move it to QA or split the woocommerce.tsx file.

johnolek:
Thanks for the review, @costasovo! 

> If you think it is a good idea to move them, feel free to do so.

I do think it's a good idea, but not something that needs to be done right away, and I would kind of like to group the work together with some other reorganization/refactoring of this part of the codebase.

With that in mind, I'm going to create a separate ticket for the extraction work and to keep track of other refactoring ideas, and I'll pass this off to QA as is.

Aschepikov (approved):
GJ!

Aschepikov (approved):
GJ!

costasovo:
> I do think it's a good idea, but not something that needs to be done right away, and I would kind of like to group the work together with some other reorganization/refactoring of this part of the codebase.

Sure! It makes perfect sense! 👍 

> With that in mind, I'm going to create a separate ticket for the extraction work and to keep track of other refactoring ideas, and I'll pass this off to QA as is.

Thanks for preparing the ticket!

PR #4855, opened 2023-04-14
Fix formatting in form

brezocordero:
[MAILPOET-5233]

[MAILPOET-5233]: https://mailpoet.atlassian.net/browse/MAILPOET-5233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Form/Block/Select.php
+++ b/mailpoet/lib/Form/Block/Select.php
@@ -51,7 +51,7 @@ class Select {
       if (!empty($block['params']['required'])) {
         $label .= ' *';
       }
-      $html .= '<option value="" disabled selected hidden>' . $this->wp->escAttr($label) . '</option>';
+      $html .= '<option value="" disabled selected hidden>' . $this->wp->escHtml($label) . '</option>';
     } else {
       if (empty($block['params']['required'])) {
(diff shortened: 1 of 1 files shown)


mailpoet/lib/Form/Block/Select.php
```diff
@@ -51,7 +51,7 @@ class="mailpoet_select"
       if (!empty($block['params']['required'])) {
         $label .= ' *';
       }
-      $html .= '<option value="" disabled selected hidden>' . $label . '</option>';
+      $html .= '<option value="" disabled selected hidden>' . $this->wp->escAttr($label) . '</option>';
```
costasovo: I think that for this case, we should use escHtml as this is not printed inside an attribute.

costasovo (approved):
@brezocordero Thanks for fixing this so quickly! I noticed that in one case, there was escAttr used for escaping inner Html instead of an attribute. I went ahead and added one commit fixing this.

Aschepikov (approved):
GJ!

PR #4857, opened 2023-04-17
Make sure generated empty index files contain PHP open tag [MAILPOET-5236]

JanJakes:
## Description

This error was detected by a rule "Internal.NoCodeFound" with the following message:

No PHP code was found in this file and short open tags are not allowed by this install of PHP. This file may be using short open tags but PHP does not allow them.

## QA notes

Check that the `index.php` files in any random folders in the plugin ZIP file contain "<?php" (and nothing else).

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/711

## Linked tickets

[MAILPOET-5236]

[MAILPOET-5236]: https://mailpoet.atlassian.net/browse/MAILPOET-5236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/build.sh
+++ b/mailpoet/build.sh
@@ -145,7 +145,7 @@ php "$(dirname "$0")"/tasks/fix-full-path-disclosure.php $plugin_name
 # Add index.php files if they don't exist to all folders
 echo '[BUILD] Adding index.php files to all project folders (to avoid directory listing disclosure)'
 find $plugin_name -type d -print0 | while read -d $'\0' dir; do
-  if [ ! -f "$dir/Index.php" ]; then
+  if [ ! -f "$dir/index.php" ] && [ ! -f "$dir/Index.php" ]; then
     echo "<?php" > "$dir/index.php"
   fi
 done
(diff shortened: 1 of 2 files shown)


costasovo (commented):
@JanJakes Thanks for the quick fix.
Just want to make sure I understand the change. This PR addresses only the short open tags part of the error message (
`No PHP code was found in this file and short open tags are not allowed by this install of PHP. `)
Shouldn't we add also a comment or so to fix also the "No PHP code" part of the warning or it is not necessary?

JanJakes:
@costasovo Now it has PHP code, doesn't it? Or what else do you think I should do?

costasovo:
@JanJakes Aha! So because of the short open tag this was not recognized as PHP code. The WP example https://d.pr/i/eG3QhR has also a comment so I thought that we need to add at least one line with a comment or so. But that might not be necessary.

costasovo (dismissed):

JanJakes:
@costasovo Ahh, I didn't understand the message that way. I'd try this fix, and add any additional content only if needed. I'm not sure what the sniff does, but I would be surprised if a comment would change much.

costasovo:
> Ahh, I didn't understand the message that way. I'd try this fix, and add any additional content only if needed. I'm not sure what the sniff does, but I would be surprised if a comment would change much.

@JanJakes I think your understanding of the issue/message is correct. Let's go with this fix.

costasovo (approved):

Aschepikov (approved):
GJ!

PR #4859, opened 2023-04-18
Fix formatting of custom checkbox values [MAILPOET-5241]

costasovo:
## Description

See [MAILPOET-5241]

[MAILPOET-5241]: https://mailpoet.atlassian.net/browse/MAILPOET-5241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/JSON/v1/CustomFields.php
+++ b/mailpoet/lib/API/JSON/v1/CustomFields.php
@@ -9,6 +9,7 @@ use MailPoet\API\JSON\ResponseBuilders\CustomFieldsResponseBuilder;
 use MailPoet\Config\AccessControl;
 use MailPoet\CustomFields\CustomFieldsRepository;
 use MailPoet\Entities\CustomFieldEntity;
+use MailPoet\Form\ApiDataSanitizer;
 
 class CustomFields extends APIEndpoint {
   public $permissions = [
@@ -21,12 +22,17 @@ class CustomFields extends APIEndpoint {
   /** @var CustomFieldsResponseBuilder */
   private $customFieldsResponseBuilder;
 
+  /** @var ApiDataSanitizer */
+  private $dataSanitizer;
+
   public function __construct(
     CustomFieldsRepository $customFieldsRepository,
-    CustomFieldsResponseBuilder $customFieldsResponseBuilder
+    CustomFieldsResponseBuilder $customFieldsResponseBuilder,
+    ApiDataSanitizer $dataSanitizer
   ) {
     $this->customFieldsRepository = $customFieldsRepository;
     $this->customFieldsResponseBuilder = $customFieldsResponseBuilder;
+    $this->dataSanitizer = $dataSanitizer;
   }
 
   public function getAll() {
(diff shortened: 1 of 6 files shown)


mailpoet/lib/Subscription/ManageSubscriptionFormRenderer.php
```diff
@@ -134,6 +140,9 @@ private function getCustomFields(SubscriberEntity $subscriber): array {
 
       $customField['params']['value'] = $customFieldValues[$customFieldEntity->getId()] ?? null;
 
+      // Historically, custom fields were not sanitized so we need to sanitize them here
+      $customField = $this->dataSanitizer->sanitizeBlock($customField);
```
rodrigoprimo: It is not clear to me why we are sanitizing on output. Shouldn't we escape instead of sanitize?

As a rule of thumb, I think it is preferable to escape as late as possible. With that in mind, have you considered escaping the checkbox value inside the checkbox renderer (https://github.com/mailpoet/mailpoet/blob/f9356281b9aa0563e36e0c765c1f9c8c12f0a713/mailpoet/lib/Form/Block/Checkbox.php#L69
https://github.com/Automattic/mailpoet-security-vulnerabilities/blob/f9356281b9aa0563e36e0c765c1f9c8c12f0a713/mailpoet/lib/Form/Block/Checkbox.php#L69)?

rodrigoprimo (commented):
I can confirm that it works, but I left a question for us to discuss the implementation further.

mailpoet/lib/Subscription/ManageSubscriptionFormRenderer.php
costasovo: > It is not clear to me why we are sanitizing on output. Shouldn't we escape instead of sanitize?

@rodrigoprimo Good question. We want to allow some HTML in the text displayed next to the checkbox. The common use-case is a link to a privacy policy page. So I can't just use esc_hmtl. So I wanted to achieve this using wp_kses. This is what the sanitizer does. I reused the sanitizer because it carries the array with the tags for the checkbox. Now I see it is confusing, and also, the sanitizer may change and start doing more in the future.
I'm thinking about calling wp_kses in the renderer and somehow sharing the array with the config for wp_kses. Do you find this a good solution or do you have a different idea?

rodrigoprimo: Thanks for the additional information, @costasovo. I had missed that we want to allow some HTML in the text displayed next to the checkbox. I like your solution. When you say calling wp_kses in the renderer, do you mean in \MailPoet\Form\Block\Checkbox::render()? Just checking as this is where I believe this code should go.

costasovo: Yes I want to use it in \MailPoet\Form\Block\Checkbox::render()! Thanks for the feedback. I'll update the PR.

costasovo:
@rodrigoprimo I replaced the second commit with a new one that implements the escaping, as discussed in the previous comments.

rodrigoprimo (approved):
Looks good to me. Great job!

Aschepikov (approved):
GJ!

PR #4783, opened 2023-03-09
Add email scheduling with minute resolution [MAILPOET-4602]

lysyjan:
## Description

This PR extends scheduling all email types with minute resolution.
In the scheduling for the abandoned cart, the item with `30 minutes` was removed, although according to the specification, it should have been retained. The reason was that the item in the selection had the same value.

## Code review notes

I was unsure if some acceptance tests should be added, but because there are only new items in the selections, I decided to skip this step.

## Linked tickets

[MAILPOET-4602]

[MAILPOET-4602]: https://mailpoet.atlassian.net/browse/MAILPOET-4602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4860, opened 2023-04-18
Fix scheduling post notifications by minute resolution [MAILPOET-5244]

lysyjan:
## Description

I recently allowed scheduling emails with the minute resolution, but I overlooked a logic in the PostNotificationScheduler class. This PR extends integration tests and also adds support for the minute resolution in the mentioned class.

## Code review notes

The third commit adds a migration that fixes the invalid value of the scheduled time for already existing post notifications. Nevertheless, it does not create a new record in the scheduled tasks.

## QA notes

1. Create a post notification on the trunk
2. Install the fixed build
3. Verify that migrations were executed (The migrations table contains a record with the name `Migration_20230419_080000`)
4. Update or create a new post and check that newsletter was scheduled properly

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/4783

## Linked tickets

[MAILPOET-5244]

## After-merge notes

_N/A_

[MAILPOET-5244]: https://mailpoet.atlassian.net/browse/MAILPOET-5244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Migrations/Migration_20230419_080000.php
+++ b/mailpoet/lib/Migrations/Migration_20230419_080000.php
@@ -1,0 +1,38 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Migrations;
+
+use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Migrator\Migration;
+use MailPoet\Newsletter\NewslettersRepository;
+use MailPoet\Newsletter\Scheduler\PostNotificationScheduler;
+
+class Migration_20230419_080000 extends Migration {
+  /** @var NewslettersRepository */
+  private $newslettersRepository;
+
+  /** @var PostNotificationScheduler */
+  private $postNotificationScheduler;
+
+  public function run(): void {
+    $this->newslettersRepository = $this->container->get(NewslettersRepository::class);
+    $this->postNotificationScheduler = $this->container->get(PostNotificationScheduler::class);
+    $this->fixPostNotificationScheduleTime();
+  }
+
+  /**
+   * Because we released PostNotificationScheduler that didn't schedule notifications with the minute resolution,
+   * which was added in version 4.10.0, we need to fix the scheduled time for all notifications.
+   *
+   * Ticket with bug: https://mailpoet.atlassian.net/browse/MAILPOET-5244
+   * Ticket with adding minute resolution: https://mailpoet.atlassian.net/browse/MAILPOET-4602
+   *
+   * @return void
+   */
+  private function fixPostNotificationScheduleTime() {
+    $newsletters = $this->newslettersRepository->findBy(['type' => NewsletterEntity::TYPE_NOTIFICATION]);
+    foreach ($newsletters as $newsletter) {
+      $this->postNotificationScheduler->processPostNotificationSchedule($newsletter);
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


mailpoet/lib/Migrations/Migration_20230419_080000.php
```diff
@@ -0,0 +1,29 @@
...
+  /** @var PostNotificationScheduler */
+  private $postNotificationScheduler;
+
+  public function run(): void {
+    $this->newslettersRepository = $this->container->get(NewslettersRepository::class);
+    $this->postNotificationScheduler = $this->container->get(PostNotificationScheduler::class);
+    $this->fixPostNotificationScheduleTime();
+  }
+
+  private function fixPostNotificationScheduleTime() {
```
costasovo, line 32: Could you please add a comment here describing what is this migration fixing? I think it would be helpful for the future.

mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
```diff
@@ -162,7 +163,8 @@ public function processPostNotificationSchedule(NewsletterEntity $newsletter) {
     $intervalType = $intervalTypeOption ? $intervalTypeOption->getValue() : null;
 
     $timeOfDayOption = $newsletter->getOption(NewsletterOptionFieldEntity::NAME_TIME_OF_DAY);
-    $hour = $timeOfDayOption ? (int)$timeOfDayOption->getValue() / self::SECONDS_IN_HOUR : null;
+    $hour = $timeOfDayOption ? (int)floor((int)$timeOfDayOption->getValue() / self::SECONDS_IN_HOUR) : null;
+    $minute = $timeOfDayOption && $hour ? ((int)$timeOfDayOption->getValue() - (int)($hour * self::SECONDS_IN_HOUR)) / self::SECONDS_IN_MINUTE : null;
```
costasovo: Will this code also work for the time between midnight and 1 am (e.g., 00:45)? If I understand the code correctly in such case, the `$hour` will be 0, so `$timeOfDayOption && $hour ` will evaluate as `false` and `$minute` will be `null`. Maybe we don't need the `&& $hour` here.

costasovo: Also with the `null` value in `$minute` and `$hour = 0` the  `sprintf('%s %s * * *', $minute, $hour);` will evaluate to `' 0 * * *'` which seems to be an incorrect value for cron timing. I tried checking in at `https://crontab.guru/#_0_*_*_*`

costasovo (commented):
Thanks for fixing the issue and adding a rich set of tests. I think I found a potential issue for times short after midnight. Please see my comments.

mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
lysyjan: Great catch 👍 Thank you

costasovo (approved):
Thanks! 👍

Aschepikov (approved):
GJ!

PR #4861, opened 2023-04-18
Replace some of the usages of the raw filter in the Twig templates

rodrigoprimo:
## Description

This PR replaces some of the usages of the raw filter in the Twig templates with more secure alternatives. See the commit messages for more details on each of the approaches.

It seems to me that there is not much we can do about the other uses of `| raw` that are still left in the code base. The variables that are being passed to the `raw` filter contain complex HTML that I'm not sure if we could escape or filter somehow. But please let me know if I missed something and you have a suggestion.

## Code review notes

I'm not an expert on Twig templates, so I'm not sure if the approach that I took here is the best or if there is something else that we should consider. I did search a bit, and I could not find anything better.

## QA notes

This PR should have no user facing changes. But it makes some modifications that could break things in the forms, the pages to import and export subscribers, the upgrade, and the manage where subscribers can manage their subscriptions. If possible, please check those pages are still working as expected.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5238]

## After-merge notes

_N/A_

[MAILPOET-5238]: https://mailpoet.atlassian.net/browse/MAILPOET-5238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Twig/Filters.php
```diff
@@ -20,6 +21,16 @@ public function getFilters() {
...
+        [$this, 'wpKses'],
+        ['is_safe' => ['html']]
+      ),
     ];
   }
+
+  public function wpKses($content, $allowedHtml) {
+    $wp = WPFunctions::get();
+    return $wp->wpKses($content, $allowedHtml);
+  }
```
costasovo, line 35: @rodrigoprimo The new filter is an excellent idea 👍 
On thing that came to my mind when I saw these variable definitions in templates (`<% set allowedHtml = {'link': {'href': [], 'rel': []}} %>` ) that we could merge all these allowedHtml into one array that would be hard coded in a constant for this filter. 
If I'm not mistaken, all those allowed tags and attributes are harmless and I don't see any problem allowing a bit more that is actually needed for some cases. We wouldn't need to define specific restrictions for each usage. If we do the change maybe we could then rename the filter to `allowedHtml`.
What do you think?

costasovo (commented):
Good job! I have an idea about the new filter that maybe we could consider. Please see my comment.

mailpoet/lib/Twig/Filters.php
rodrigoprimo: @costasovo, I considered this approach but I opted to sacrifice the developer experience in favor of being more specific in the list of allowed HTML tags. I'm not sure if this is the best approach or not in this case. My fear is that, down the road, we might end up wanting to add more HTML tags to the list of allowed HTML that would apply to all calls of this filter. Instead of just one specific call where it might be needed. Some of those need tags could be potentially problematic.

As an alternative to having to set the list of allowed HTML tags, developers can call the filter with one of the context names like `post` supported by `wp_kses()`. This should work in a way very similar to what you are suggesting. Would this be enough in your opinion, or would you still prefer that we implement a list of allowed HTML tags in the filter?

costasovo: @rodrigoprimo Ok. Makes sense. Let's keep it as is. Maybe in the future, we will add more, and it will be easier to decide what is the best approach. We could also define our own context or contexts.

rodrigoprimo: Sounds good to me. Thanks.

Aschepikov (approved):
GJ!

PR #4862, opened 2023-04-19
Add Semgrep security scan [MAILPOET-5230]

alex-mailpoet:
## Description

This PR does two things:
1. Adds a Semgrep security check to our CI build (and a `./do qa:semgrep` command for local use). Requires Docker to be running. [Sample run](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/14045/workflows/e32df982-5229-4425-990a-600a43285ee8/jobs/238445);
2. Ignores two issues found by Semgrep as false positives:
    - `Potential Unsafe usage (non escaped) of remove_query_arg() or
  add_query_arg()`. It was complaining about a wrapped WP function which is not doing anything unsafe by itself. I manually checked all use cases of `->addQueryArg()` and found them to be escaped or used in a safe context. This points to a larger issue with Semgrep checking our plugin: since we use wrapped WP functions, they do not get scanned all over the codebase. Should we keep it as-is, add and maintain a MailPoet-specific rule set with camel cased WP functions, unwrap WP functions (replace CamelCase back to snake_case) specifically for this job as suggested by @JanJakes, or do something else?
    - `Potential unsafe usage of Plugin_Upgrader()->install()`. We are not using it to download arbitrary plugins, only `mailpoet-premum`.
    
P.S.: It seems ignoring Semgrep errors instead of fixing will be common, because it doesn't understand usage context. It sees a function being called with a variable argument and result not escaped on the same line - it throws a complaint.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/715

## Linked tickets

[MAILPOET-5230]

## After-merge notes

_N/A_

[MAILPOET-5230]: https://mailpoet.atlassian.net/browse/MAILPOET-5230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,11 @@ anchors:
       - static_analysis_php8
     <<: *only_trunk_and_release
 
+  command_add_github_key: &command_add_github_key
+    command: |
+      # Add GitHub to known_hosts (for jobs without a checkout step that need to clone repos)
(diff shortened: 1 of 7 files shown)


JanJakes (commented):
Could we use a download link such as `https://github.com/Automattic/wpscan-semgrep-rules/archive/99721b33c00d309a15bd9508ff1c97ade0df5dbe.zip` to simplify the setup, or would that fail on some permission problems?

I think even if we keep using `git clone`, we could add a particular commit for stability and maybe a `--depth 0` for performance.

alex-mailpoet:
@JanJakes 
> Could we use a download link such as https://github.com/Automattic/wpscan-semgrep-rules/archive/99721b33c00d309a15bd9508ff1c97ade0df5dbe.zip to simplify the setup, or would that fail on some permission problems?

I did not try it. Since it's a private repository, this will be subject to the same permission issues, but with worse error handling: it just returns 404 Not Found whether a link is broken or permission is denied, leaving us guessing. The `git clone` command is more informative in that regard.

> we could add a particular commit for stability

I'm not sure it's a good idea because I think we should always be up-to-date in our security checks. Otherwise we may forget to update the check and discover that the marketplace rejects our build because of new security rules only during release time.

> and maybe a --depth 0 for performance.

`--depth 0` is not allowed by Git, you meant `--depth 1`? I can add it.

JanJakes (approved):

PR #4866, opened 2023-04-19
Improve escaping of textarea value [MAILPOET-5243]

costasovo:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5243]

## After-merge notes

_N/A_

[MAILPOET-5243]: https://mailpoet.atlassian.net/browse/MAILPOET-5243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Form/Block/Textarea.php
+++ b/mailpoet/lib/Form/Block/Textarea.php
@@ -55,7 +55,7 @@ class Textarea {
       $html .= 'style="' . $this->wp->escAttr($styles) . '" ';
     }
 
-    $html .= '>' . $this->rendererHelper->getFieldValue($block) . '</textarea>';
+    $html .= '>' . $this->rendererHelper->escapeShortCodes($this->rendererHelper->getFieldValue($block)) . '</textarea>';
 
     return $this->wrapper->render($block, $html);
   }
--- a/mailpoet/lib/Form/Block/BlockRendererHelper.php
+++ b/mailpoet/lib/Form/Block/BlockRendererHelper.php
@@ -258,6 +258,15 @@ class BlockRendererHelper {
     return join(' ', $modifiers);
   }
 
+  public function escapeShortCodes(?string $value): ?string {
+    if ($value === null) {
+      return null;
+    }
+    return preg_replace_callback('/' . $this->wp->getShortcodeRegex() . '/s', function ($matches) {
+      return str_replace(['[', ']'], ['&#91;', '&#93;'], $matches[0]);
+    }, $value);
+  }
+
   private function translateValidationErrorMessage(string $validate): string {
     switch ($validate) {
       case 'email':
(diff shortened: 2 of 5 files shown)


mailpoet/lib/Form/Block/Textarea.php
```diff
@@ -55,7 +55,7 @@ public function render(array $block, array $formSettings): string {
       $html .= 'style="' . $this->wp->escAttr($styles) . '" ';
     }
 
-    $html .= '>' . $this->rendererHelper->getFieldValue($block) . '</textarea>';
+    $html .= '>' . $this->rendererHelper->escapeShortCodes($this->rendererHelper->getFieldValue($block)) . '</textarea>';
```
rodrigoprimo, line 58: Instead of creating a custom field to escape the shortcodes, can we use strip_shortcodes() (https://developer.wordpress.org/reference/functions/strip_shortcodes/)? Or is there a reason we want to keep the shortcodes?

costasovo: @rodrigoprimo Thanks for pointing me to the strip_shorcodes function. I was not aware of it. I was searching for escaping because I didn't want to remove content that may possibly be formatted as a shortcode. But if I understand correctly strip_shorcodes removes only known shortcodes so I'll use it instead of the custom escape method.

costasovo: @rodrigoprimo I tried using `strip_shortcodes`, but I noticed a problem when using the function.
A shortcode can be escaped [also by adding additional brackets](https://codex.wordpress.org/Shortcode) (e.g. [[gallery]]), and WP should not process it.

Here is what `strip_shortcodes` does:
```php
strip_shortcodes('[mailpoet_manage]'); // Outputs ""
strip_shortcodes('[[mailpoet_manage]]'); // Outputs [mailpoet_manage]
strip_shortcodes('[[[mailpoet_manage]]]'); // Outputs [[mailpoet_manage]]
```
So if I run the function only once, there may remain a valid shortcode that could be interpreted. Actually, I would need to run it in a loop and stop after verifying the content doesn't contain a valid short code. So in the end the solution would be similarly complex. 

Now I'm more inclined to go with the current implementation as proposed in the PR. What do you think? Any suggestions?

rodrigoprimo: Moved the conversation to Slack.

https://a8c.slack.com/archives/C01GH835U90/p1681990755903259

rodrigoprimo (dismissed):
Looks good to me!

Aschepikov (dismissed):
GJ!

Aschepikov:
HI @costasovo there are some conflicts, can you resolve them pls

costasovo:
@Aschepikov I fixed the conflict!

Aschepikov (approved):
GJ!

PR #4870, opened 2023-04-20
Remove models that were deprecated over six months ago [MAILPOET-5252]

rodrigoprimo:
## Description

The removal date of all these models passed a couple of months ago. This PR also remove a few deprecated method from the Newsletter model that are related to two of the removed models.

## Code review notes

_N/A_

## QA notes

No QA is needed.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5252]

## After-merge notes

_N/A_

[MAILPOET-5252]: https://mailpoet.atlassian.net/browse/MAILPOET-5252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Models/NewsletterOption.php
triple0t, line 1: I found usage of this model in https://github.com/mailpoet/mailpoet/blob/a62fca9c80fc989196367ec90eaee098e03bac15/mailpoet/lib/Models/Newsletter.php#L110

mailpoet/lib/Models/NewsletterOptionField.php
triple0t, line 1: I found usage of this model in https://github.com/mailpoet/mailpoet/blob/a62fca9c80fc989196367ec90eaee098e03bac15/mailpoet/lib/Models/Newsletter.php#L109

triple0t (commented):
I found usage of these two models in `/Models/Newsletter`
here: https://github.com/mailpoet/mailpoet/blob/a62fca9c80fc989196367ec90eaee098e03bac15/mailpoet/lib/Models/Newsletter.php#L107-L114

Can you confirm it's okay to delete them?

rodrigoprimo:
> Can you confirm it's okay to delete them?

Good catch, @triple0t. When creating this PR, I failed to check if the models were still being used. I just blindly removed them when I saw that their removal date had passed and the tests did not fail without them. I should have checked as you did.

The method that is still using them doesn't seem to be used anymore (it is actually only used by a Newsletter method model that is not used). Despite this, to be on the safe side, I opted not to remove NewsletterOption and NewsletterOptionField. Instead, I deprecated the two Newsletter methods that still use those two models and bumped their removal date.

Let me know what you think of this solution.

triple0t:
> Good catch, @triple0t. When creating this PR, I failed to check if the models were still being used. I just blindly removed them when I saw that their removal date had passed and the tests did not fail without them. I should have checked as you did.
> 
> The method that is still using them doesn't seem to be used anymore (it is actually only used by a Newsletter method model that is not used). Despite this, to be on the safe side, I opted not to remove NewsletterOption and NewsletterOptionField. Instead, I deprecated the two Newsletter methods that still use those two models and bumped their removal date.
> 
> Let me know what you think of this solution.

Thank you for providing extra information.

I'm okay with the solution

triple0t (approved):
LGTM

PR #4880, opened 2023-04-21
Limit rows deleted from mailpoet_logs  [MAILPOET-5071]

costasovo:
## Description
We want to make sure that we don't lock the log table by deleting too many rows in one query.
If the probability of running the delete routine is 5%  for each log writing, the limit of 1000 should be fine.

## Code review notes

_N/A_

## QA notes

You can test this doesn't break logging.
Go to MailPoet > Settings > Advanced and set Logging to `everything`
Then you can play with the plugin, send some emails and check `wp-admin/admin.php?page=mailpoet-logs` that logs are written.

## Linked PRs

_N/A_

## Linked tickets

 [MAILPOET-5071]

## After-merge notes

_N/A_

[MAILPOET-5071]: https://mailpoet.atlassian.net/browse/MAILPOET-5071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Logging/LogRepository.php
+++ b/mailpoet/lib/Logging/LogRepository.php
@@ -70,7 +70,8 @@ class LogRepository extends Repository {
     $logsTable = $this->entityManager->getClassMetadata(LogEntity::class)->getTableName();
     $this->entityManager->getConnection()->executeStatement("
       DELETE FROM $logsTable
-      WHERE `created_at` < :date LIMIT :limit
+      WHERE `created_at` < :date
+      ORDER BY `id` ASC LIMIT :limit
     ", [
       'date' => Carbon::now()->subDays($daysToKeepLogs)->toDateTimeString(),
       'limit' => $limit,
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Logging/LogRepository.php
```diff
@@ -65,12 +66,19 @@ public function getLogs(
...
-  public function purgeOldLogs(int $daysToKeepLogs) {
-    $queryBuilder = $this->entityManager->createQueryBuilder();
-    return $queryBuilder->delete(LogEntity::class, 'l')
-      ->where('l.createdAt < :days')
-      ->setParameter('days', Carbon::now()->subDays($daysToKeepLogs)->toDateTimeString())
-      ->getQuery()->execute();
+  public function purgeOldLogs(int $daysToKeepLogs, int $limit = 1000) {
+    $logsTable = $this->entityManager->getClassMetadata(LogEntity::class)->getTableName();
+    $this->entityManager->getConnection()->executeStatement("
+      DELETE FROM $logsTable
```
lysyjan, line 72: I'm considering if we should use here a subquery with ordering to avoid an unexpected gap in logs. Did you think about it? Or what is your opinion here?
It's only an idea that I wanted to mention.

costasovo: I'll add the order by.

costasovo:
@lysyjan Thanks for pointing out the issue with ordering! I added one commit to address it.

lysyjan (approved):

Aschepikov (approved):
GJ!

PR #4882, opened 2023-04-21
Add average order value filter [MAILPOET-4987]

johnolek:
## Description

This adds a new WooCommerce dynamic segment filter to filter subscribers based on the average value of their orders within the last `n` days. It works very similarly to the "single order value" filter.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4987](https://mailpoet.atlassian.net/browse/MAILPOET-4987)

## After-merge notes

_N/A_

[MAILPOET-4987]: https://mailpoet.atlassian.net/browse/MAILPOET-4987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
@@ -79,6 +79,9 @@ export function validateWooCommerce(formItems: WooCommerceFormItem): boolean {
   ) {
     return false;
   }
+  if (formItems.action === WooCommerceActionTypes.AVERAGE_SPENT) {
+    return !!formItems.amount && !!formItems.operator && !!formItems.days;
+  }
   if (formItems.action === WooCommerceActionTypes.PURCHASE_DATE) {
     return validateDateField(formItems);
   }
@@ -390,6 +393,71 @@ function TotalSpentFields({ filterIndex }: Props): JSX.Element {
   );
 }
 
+function AverageSpentFields({ filterIndex }: Props): JSX.Element {
+  const segment: WooCommerceFormItem = useSelect(
+    (select) => select(storeName).getSegmentFilter(filterIndex),
+    [filterIndex],
+  );
+  const { updateSegmentFilter, updateSegmentFilterFromEvent } =
+    useDispatch(storeName);
+  const wooCurrencySymbol: string = useSelect(
+    (select) => select(storeName).getWooCommerceCurrencySymbol(),
+    [],
+  );
+  useEffect(() => {
+    const allowedOperators = ['=', '!=', '>', '<'];
+    if (!allowedOperators.includes(segment.operator)) {
+      void updateSegmentFilter({ operator: '>' }, filterIndex);
+    }
+  }, [updateSegmentFilter, segment, filterIndex]);
+  return (
+    <>
+      <Grid.CenteredRow>
+        <Select
+          key="select"
+          value={segment.operator}
+          onChange={(e): void => {
+            void updateSegmentFilterFromEvent('operator', filterIndex, e);
+          }}
+          automationId="select-average-spent-operator"
+        >
+          <option value="=">{MailPoet.I18n.t('equals')}</option>
+          <option value="!=">{MailPoet.I18n.t('notEquals')}</option>
+          <option value=">">{MailPoet.I18n.t('moreThan')}</option>
+          <option value="<">{MailPoet.I18n.t('lessThan')}</option>
+        </Select>
+        <Input
+          data-automation-id="input-average-spent-amount"
+          type="number"
+          min={0}
+          step={0.01}
+          value={segment.amount || ''}
+          placeholder={MailPoet.I18n.t('wooSpentAmount')}
+          onChange={(e): void => {
+            void updateSegmentFilterFromEvent('amount', filterIndex, e);
+          }}
+        />
+        <div>{wooCurrencySymbol}</div>
+      </Grid.CenteredRow>
+      <Grid.CenteredRow>
+        <div>{MailPoet.I18n.t('inTheLast')}</div>
+        <Input
+          data-automation-id="input-average-spent-days"
+          type="number"
+          min={1}
+          step={1}
+          value={segment.days || ''}
+          placeholder={MailPoet.I18n.t('daysPlaceholder')}
+          onChange={(e): void => {
+            void updateSegmentFilterFromEvent('days', filterIndex, e);
+          }}
+        />
+        <div>{MailPoet.I18n.t('days')}</div>
+      </Grid.CenteredRow>
+    </>
+  );
+}
+
 function SingleOrderValueFields({ filterIndex }: Props): JSX.Element {
   const segment: WooCommerceFormItem = useSelect(
     (select) => select(storeName).getSegmentFilter(filterIndex),
@@ -542,6 +610,7 @@ const componentsMap = {
   [WooCommerceActionTypes.PURCHASED_CATEGORY]: PurchasedCategoryFields,
   [WooCommerceActionTypes.SINGLE_ORDER_VALUE]: SingleOrderValueFields,
   [WooCommerceActionTypes.TOTAL_SPENT]: TotalSpentFields,
+  [WooCommerceActionTypes.AVERAGE_SPENT]: AverageSpentFields,
 };
 
 export function WooCommerceFields({ filterIndex }: Props): JSX.Element {
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
@@ -81,7 +81,9 @@ export function validateWooCommerce(formItems: WooCommerceFormItem): boolean {
   }
   if (
     formItems.action === WooCommerceActionTypes.AVERAGE_SPENT &&
-    (!formItems.amount || !formItems.operator || !formItems.days)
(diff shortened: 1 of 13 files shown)


mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -79,6 +79,9 @@ export function validateWooCommerce(formItems: WooCommerceFormItem): boolean {
   ) {
     return false;
   }
+  if (formItems.action === WooCommerceActionTypes.AVERAGE_SPENT) {
```
rodrigoprimo: I don't think it is necessary to change anything in this block, but I just wanted to mention that since it uses a slightly different pattern than the other checks in this method, it took me a little longer to understand it and confirm that it makes sense.

mailpoet/assets/js/src/segments/dynamic/types.ts
```diff
@@ -91,6 +91,8 @@ export interface WooCommerceFormItem extends FormItem {
   single_order_value_type?: string;
   single_order_value_amount?: number;
   single_order_value_days?: number;
+  amount?: number;
```
rodrigoprimo: I'd be curious to hear your reasoning for not including the filter type as a prefix in the name of the two form items that were added to this PR. I haven't thought too much about this, but I wonder if adding the prefix makes it easier to identify which filter type a given form item corresponds to (maybe there are also disadvantages that I'm missing).

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceAverageSpentTest.php
```diff
@@ -0,0 +1,132 @@
...
+      'days' => $days,
+    ]);
+
+    return $this->tester->getSubscriberEmailsMatchingDynamicFilter($filterData, $this->averageSpentFilter);
+  }
+
+  private function createCustomerWithOrderValues(string $customerEmail, array $values): void {
+    $customerId = $this->tester->createCustomer($customerEmail);
+    foreach ($values as $value) {
+      $this->createOrder($customerId, $value );
```
rodrigoprimo, line 147: Can we use \MailPoet\WooCommerce\Helper::wcCreateOrder() here instead of creating another `createOrder()` method? I think it would be ok to use it and change it to call `\IntegrationTester::updateWooOrderStats()`.

mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
```diff
@@ -421,9 +421,11 @@ function AverageSpentFields({ filterIndex }: Props): JSX.Element {
           }}
           automationId="select-average-spent-operator"
         >
+          <option value=">">{MailPoet.I18n.t('moreThan')}</option>
+          <option value=">=">{MailPoet.I18n.t('moreThanOrEqual')}</option>
           <option value="=">{MailPoet.I18n.t('equals')}</option>
           <option value="!=">{MailPoet.I18n.t('notEquals')}</option>
-          <option value=">">{MailPoet.I18n.t('moreThan')}</option>
+          <option value="<">{MailPoet.I18n.t('lessThanOrEqual')}</option>
```
rodrigoprimo: I think there is a typo here and the value should be `<=` instead

mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceTotalSpent.php
```diff
@@ -3,52 +3,30 @@
...
-    global $wpdb;
-    $subscribersTable = $this->entityManager->getClassMetadata(SubscriberEntity::class)->getTableName();
     $filterData = $filter->getFilterData();
     $type = $filterData->getParam('total_spent_type');
     $amount = $filterData->getParam('total_spent_amount');
     $days = $filterData->getParam('total_spent_days');
 
     $date = Carbon::now()->subDays($days);
     $parameterSuffix = $filter->getId() ?? Security::generateRandomString();
-    $collation = $this->collationChecker->getCollateIfNeeded(
```
rodrigoprimo, line 45: Just checking the commit message and the commit changes, it is not clear to me which helper will be used instead and why the removed code is not necessary anymore. Could you please explain?

rodrigoprimo (commented):
Overall the PR looks good to me. I left a few comments with some points for us to discuss and a typo that I found.

mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
johnolek: Thank you for this comment! I think it is worth updating this to be more consistent.

mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/woocommerce.tsx
johnolek: Good catch! I copied these options from the single order value fields, so I fixed it in both places.

mailpoet/assets/js/src/segments/dynamic/types.ts
johnolek: I was thinking I would like to see slightly DRYer code in this area, and if we wanted to have generic `OrderValueFields` or even `ValueFields` it probably wouldn't make sense to have those prefixes. That being said, if I wanted to do that then I should have done it! I'm going to change this to be consistent with the single order value options for now and maybe I'll look at refactoring sometime later.

mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceTotalSpent.php
johnolek: I apologize for the confusing commit! This indeed doesn't make sense. These are changes I should have made in 30c8ce08395d5e51e4e68a656ee02b9460e470cd, when I actually refactored this filter to use the `wooFilterHelper`, at which point `$collation` became an unused variable. 

Would it be best to rewrite the commit message?

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceAverageSpentTest.php
johnolek: I'm a bit torn on how to approach this. Can we easily do everything we need to using just `\MailPoet\WooCommerce\Helper::wcCreateOrder()`? That function is just a wrapper for `wc_create_order`, which as best as I can tell doesn't accept arguments for creation date or total, so at some point we would need to make changes to the returned `WC_Order` object. 

We're [already doing that](https://github.com/mailpoet/mailpoet/blob/fc6ab08ccdbf6422c9a947a30086fda3abfc1e8c/mailpoet/tests/_support/IntegrationTester.php#L104-L110) for `date_created` and `billing_email` in our helper method. I've updated the helper to also accept a `total`, and to pass the other arguments through to `wcCreateOrder` so we can pass default arguments like `status` and `customer_id`. This means not having to call `updateWooOrderStats` in this `createOrder`. Does that address your primary concern?

I still like having a `createOrder` helper here because I call it in a few places and wouldn't want to have to repeat everything each time.

Please do let me know if you can think of a more elegant solution, which I'm sure exists.

johnolek:
Thank you for your careful review, @rodrigoprimo! This is ready for a final review, and I'm happy to make more changes if you can think of any other improvements. I did decide, by the way, to add some error checking in fc6ab08ccdbf6422c9a947a30086fda3abfc1e8c that I had previously neglected.

mailpoet/assets/js/src/segments/dynamic/types.ts
rodrigoprimo: I considered this as well and I think it is a good idea to, at some point, refactor this part of the code and have generic names that are used by different filters. For now, the change that you implemented adding the prefix looks good to me. Thanks.

mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommerceAverageSpentTest.php
rodrigoprimo: > I still like having a createOrder helper here because I call it in a few places and wouldn't want to have to repeat everything each time.

:+1:

mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceTotalSpent.php
rodrigoprimo: Thanks for the clarification, @johnolek. I don't have a strong opinion here, but I'm inclined to say it is worth rewriting the commit message to make this more clear.

johnolek: I reworded in ef654550ffc9e0191e3f2ebc6c59d5444aaefa1e

rodrigoprimo (dismissed):
Good job!

veljkho (changes requested):
[I don't see calculation](https://d.pr/i/QiVM7P) when I enter values for the very first time. So, it requires me to actually change `more than` to something else to see the calculation, otherwise I don't see it. This is not happening with other segment.

johnolek:
Thank you so much for catching that, @veljkho! It should be fixed in 4c1e0a6a64b9e5d5dd1087a47299c284bfda61be

veljkho (approved):

PR #4868, opened 2023-04-19
Automation filter components [MAILPOET-4624] [MAILPOET-5001] [MAILPOET-5187]

JanJakes:
## Description

Implements number, integer, enum, and boolean filters.

## Code review notes

The pull request is against https://github.com/mailpoet/mailpoet/pull/4828, which needs to be merged first. Please, if approved, assign back to me, I'll take care of the rebase.

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/714

## Linked tickets

[MAILPOET-4624]
[MAILPOET-5001]
[MAILPOET-5187]

[MAILPOET-4624]: https://mailpoet.atlassian.net/browse/MAILPOET-4624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5001]: https://mailpoet.atlassian.net/browse/MAILPOET-5001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5187]: https://mailpoet.atlassian.net/browse/MAILPOET-5187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4884, opened 2023-04-25
Automation filter datetime [MAILPOET-5000]

JanJakes:
## Description

This is now opened against https://github.com/mailpoet/mailpoet/pull/4868 which must be merged first.

Apart from the date/time value, I added value cleanup on filter condition change for now. Thinking of better ways to handle it more generically (maybe another ticket).

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/719

## Linked tickets

[MAILPOET-5000]

## After-merge notes

_N/A_

[MAILPOET-5000]: https://mailpoet.atlassian.net/browse/MAILPOET-5000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Integrations/Core/Filters/DateTimeFilter.php
```diff
@@ -0,0 +1,173 @@
...
+    return in_array($day, $filterValue, true);
+  }
+
+  private function convertToWpTimezone(DateTimeInterface $datetime): DateTimeImmutable {
+    $value = DateTimeImmutable::createFromFormat('U', (string)$datetime->getTimestamp(), $this->timezone);
+    if (!$value) {
+      throw new InvalidStateException('Failed to convert datetime to WP timezone');
+    }
+    return $value;
+  }
```
websupporter: Hmm... If we call it `convertToWpTimezone` and bind it to the WpTimezone, I would at least name the `private $timezone` `private $wpTimezone` I think.

I like that it's a `\DateTimeZone` in the constructor and not the `WordPress`-wrapper, but it might make sense to use `$this->wp->wpTimezone()` instead here, if you want to tie it to `WpTimezone`.

Another thought would be `convertToLocalTimezone()` maybe, then I think `private $localTimezone` would be appropriate :thinking:

websupporter:
Thanks for tackling filters @JanJakes! Some remarks and a question

> I added value cleanup on filter condition change for now

I do not understand what you mean :)

mailpoet/lib/Automation/Integrations/Core/Filters/DateTimeFilter.php
JanJakes: Great point! Renamed to `localTimezone`.

JanJakes:
Thanks for the review!

> I do not understand what you mean :)

I actually mean this commit — https://github.com/mailpoet/mailpoet-premium/pull/719/commits/42c6d6b38780e1dba9458266c2b7a477e4f76537, in premium. When you change `before` to, say, `in-the-last`, I wanted to reset the values because they are incompatible. But changing `before` to `after` wouldn't necessarily need that. For now, though, I decided to reset always. I don't like that commit too much and think of better ways to solve it in the future.

websupporter (approved):
lgtm

Aschepikov (approved):
GJ!

PR #4888, opened 2023-04-25
Fix missing yesAddMe translation [MAILPOET-5261]

johnolek:
## Description

This prevents the text `TRANSLATION "yesAddMe" NOT FOUND` from getting saved to the settings table if a user activates the MSS as part of the welcome wizard. It also includes a migration to automatically fix any sites where this has already happened.

## Code review notes

I know we're going to refactor all of our translations soon to use `@wordpress/i18n`, but this seemed like something we should fix sooner rather than later and is not very complex. Maybe it's overkill to extract all the settings translations, but I wouldn't want something like this to happen again before we do the refactoring.

## QA notes

See the ticket for steps to reproduce.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5261](https://mailpoet.atlassian.net/browse/MAILPOET-5261)

## After-merge notes

_N/A_

[MAILPOET-5261]: https://mailpoet.atlassian.net/browse/MAILPOET-5261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoprimo (approved):
Looks good to me. Great detective work to find what was causing this issue.

veljkho (approved):

PR #4889, opened 2023-04-26
Cast step array keys to string [MAILPOET-5267]

websupporter:
## Description

_N/A_

## Code review notes

This PR just casts the key to a string. The disadvantage is that the keys might still be `integers`. I searched the code base and did not find any other instance where we use the keys of the array.

Another possibility would be to force the keys to be a string e.g. by adding a preceding `a`, but I am not sure what this would mean in terms of already existing automations and migrations :thinking:

Imo, the type cast is enough. I do not see a lot of other scenarios, where we will use the array key instead of `$step->getId()`.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5267]

## After-merge notes

_N/A_

[MAILPOET-5267]: https://mailpoet.atlassian.net/browse/MAILPOET-5267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Data/Automation.php
+++ b/mailpoet/lib/Automation/Engine/Data/Automation.php
@@ -42,7 +42,7 @@ class Automation {
   /** @var ?DateTimeImmutable */
   private $activatedAt = null;
 
-  /** @var array<string, Step> */
+  /** @var array<string|int, Step> */
   private $steps;
 
   /** @var array<string, mixed> */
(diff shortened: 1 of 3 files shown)


JanJakes:
@websupporter Thanks for catching and fixing this! Pure craziness. Adding our detailed thread for reference: https://a8c.slack.com/archives/C0313447YBC/p1682465140559079

JanJakes (approved):

PR #4895, opened 2023-04-28
Update notifications for exceeded limits [MAILPOET-5191]

costasovo:
## Description

This PR adds caching of the plan's limit data so that we can display better notification messages (including limit details) even when the key loses access to premium or MSS and can no longer retrieve those data. Because there still might be cases when the plugin doesn't have the info about the plan's limits, I added variants of the notifications that don't contain plan details and are used as a fallback.

When working on the ticket made a small audit of how the messages work. I found that some messages were not displayed properly. I fixed that and refactored the display logic a bit.

#### Notice when sending fails (displayed on the homepage and emails page)
Screenshot 2023-04-28 at 17 05 48

Note: we hide this message if the key check fails with the subscriber limit error and we display the other 👇  message instead

#### Notice when the API key check fails and we have the plan's data
Screenshot 2023-04-28 at 16 54 31

#### Notice when the API key check fails and doesn't have the plan's data (fallback)
Screenshot 2023-04-28 at 16 57 00

#### Notice when sending fails (displayed on the homepage and emails page)

Screenshot 2023-04-28 at 16 56 01

Note: we hide this message if the key check fails with the error volume limit error and we display the other 👇  message instead

#### Notice when the API key check fails and we have the plan's data
Screenshot 2023-04-28 at 17 07 18

#### Notice when the API key check fails and doesn't have the plan's data (fallback)
Screenshot 2023-04-28 at 17 08 02

## Code review notes

Sorry for the big PR! Please review commit by commit. 

Initially, I wanted to store the plan data per API key so that the plugin would cache the plan's limit data even when users swap keys, but I found that the complexity of the change was not worth it compared to the added value. So I added caching only for one key, and I added the fallback variants of messages for cases when we know the key is blocked, but we don't have the plan's details. When working on the initial idea, I made a couple of refactors around the Bridge service. These are not really needed for the simpler approach, but I decided to keep them because I believe they improve our codebase. They are in the first couple of commits.

## QA notes

Related changes in the shop are not released yet, because we first we need to release plugins that would show proper messages.
Because of that, we need to mock API responses to be able to display the message. 
You need a valid API key for testing.

### 2) Mock the key check response. This will simulate that the key was blocked in the shop. 
You can mock the API key check response in mailpoet/lib/Services/Bridge/API.php

Screenshot 2023-04-28 at 17 35 56

add `return ['code' => 403, 'data' => null, 'error_message' => self::ERROR_MESSAGE_EMAIL_VOLUME_LIMIT_REACHED];`
to get email volume limit reached notice

add `return ['code' => 403, 'data' => null, 'error_message' => self::ERROR_MESSAGE_SUBSCRIBERS_LIMIT_REACHED];`
to get subscriber limit reached notice

### 3) Go to settings, re-enter the key and verify that the correct notice is displayed across the plugin

Note: To test the fallback message variants, you need to start with a fresh site (no API key), and skip step one. This will cause the data from the valid key will not be cached. It simulates a situation when a user tries using a blocked key on a new site.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5191]

## After-merge notes

_N/A_

[MAILPOET-5191]: https://mailpoet.atlassian.net/browse/MAILPOET-5191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Settings/SettingsChangeHandler.php
+++ b/mailpoet/lib/Settings/SettingsChangeHandler.php
@@ -5,7 +5,10 @@ namespace MailPoet\Settings;
 use MailPoet\Cron\Workers\InactiveSubscribers;
 use MailPoet\Cron\Workers\WooCommerceSync;
 use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Mailer\Mailer;
 use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
+use MailPoet\Services\Bridge;
+use MailPoet\Services\SubscribersCountReporter;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
 
@@ -17,12 +20,22 @@ class SettingsChangeHandler {
   /** @var SettingsController */
   private $settingsController;
 
+  /** @var Bridge */
+  private $bridge;
+
+  /** @var SubscribersCountReporter */
+  private $subscribersCountReporter;
+
   public function __construct(
     ScheduledTasksRepository $scheduledTasksRepository,
-    SettingsController $settingsController
+    SettingsController $settingsController,
+    Bridge $bridge,
+    SubscribersCountReporter $subscribersCountReporter
   ) {
     $this->scheduledTasksRepository = $scheduledTasksRepository;
     $this->settingsController = $settingsController;
+    $this->bridge = $bridge;
+    $this->subscribersCountReporter = $subscribersCountReporter;
   }
 
   public function onSubscribeOldWoocommerceCustomersChange(): void {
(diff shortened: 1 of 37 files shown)


mailpoet/lib/Settings/SettingsChangeHandler.php
```diff
@@ -77,4 +90,23 @@ private function createScheduledTask(string $type): ScheduledTaskEntity {
...
+    }
+    if (!empty($premiumKey)) {
+      $premiumState = $this->bridge->checkPremiumKey($premiumKey);
+      $this->bridge->storePremiumKeyAndState($premiumKey, $premiumState);
+    }
+    if ($apiKey && !empty($apiKeyState) && in_array($apiKeyState['state'], [Bridge::KEY_VALID, Bridge::KEY_VALID_UNDERPRIVILEGED], true)) {
+      return $this->subscribersCountReporter->report($apiKey);
+    }
+    if ($premiumKey && !empty($premiumState) && in_array($premiumState['state'], [Bridge::KEY_VALID, Bridge::KEY_VALID_UNDERPRIVILEGED], true)) {
+      return $this->subscribersCountReporter->report($apiKey);
```
lysyjan: I know this line of code was only moved, but it looks suspicious to me. Should the report method use the variable `$premiumKey` instead of `$apiKey`? But maybe I miss something.

mailpoet/lib/Settings/SettingsChangeHandler.php
```diff
@@ -77,4 +90,23 @@ private function createScheduledTask(string $type): ScheduledTaskEntity {
     $task->setStatus(ScheduledTaskEntity::STATUS_SCHEDULED);
     return $task;
   }
+
+  public function updateBridge($settings) {
```
lysyjan: This name looks a bit confusing to me, but I don't know about something better. Maybe `updateApiKeysState`? What do you think?

lysyjan (commented):
@costasovo Amazing job! A lot of nice refactoring and descriptive comments.
I have noticed only one small detail.

mailpoet/lib/Settings/SettingsChangeHandler.php
costasovo: Aha. Good catch! In reality, it has no impact because we use one key, but on the backend, we "pretend" there are two different keys, but I'll update it.

costasovo:
@lysyjan Thanks for the review! I added two new commits to address your notes and rebased the branch on top of the current trunk.

lysyjan (approved):

veljkho (approved):

PR #4901, opened 2023-05-02
Install older version of Tracy based on PHP version [MAILPOET-5277]

costasovo:
## Description

This PR fixes failing nightly builds and is also needed for fixing failing builds of the premium plugin.

I decided to use the approach that we install the version based on the PHP version of the environment.
Locally when developers need to switch to PHP 7.4, they only need to call `./do install` to get the correct version of the Tracy.

This will also allow us to upgrade to the newest Tracy version for PHP8.x

## Code review notes

There is also a PR in the premium plugin repository https://github.com/mailpoet/mailpoet-premium/pull/722

I created a branch `use-proper-tracy-version-delme` where I allowed the "oldest" tests to demonstrate the fix works.
Here is [a passing build](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/14178/workflows/eefd6022-004f-4d93-a260-949364916bac).

## QA notes
As this is related only to the test environment, this doesn't require QA.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/722

## Linked tickets

[MAILPOET-5277]

## After-merge notes

Please delete also `use-proper-tracy-version-delme` branch

[MAILPOET-5277]: https://mailpoet.atlassian.net/browse/MAILPOET-5277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -395,6 +395,18 @@ jobs:
           name: 'Pull test docker images'
           # Pull docker images with 3 retries
           command: i='0';while ! docker-compose -f tests/docker/docker-compose.yml pull && ((i < 3)); do sleep 3 && i=$[$i+1]; done
+      - run:
+          name: Create docker containers for test
+          # We experienced some failures when creating containers so we do it explicitly with one retry
+          command: |
+            cd tests/docker
+            docker-compose create || docker-compose create
+      - run:
+          # Some tools we use may need different version based on PHP version used in docker
+          name: Ensure correct versions of tools
+          command: |
+            cd tests/docker
+            docker-compose run --rm -w /project -e COMPOSER_DEV_MODE=1 --entrypoint "php tools/install.php" codeception_acceptance
       - when:
           condition: << parameters.woo_core_version >>
           steps:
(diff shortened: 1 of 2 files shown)


rodrigoprimo (approved):
The changes look good to me. Thanks for creating a branch to run the acceptance_oldest build job.

One thing that was not clear to me (feel free to merge the PR as it is not a blocker), is why it was necessary to change the place where we call `docker-composer create` and then call `tools/install.php` setting COMPOSER_DEV_MODE to 1.

Without knowing these details, my likely wrong assumption is that just the changes to `tools/install.php` are enough to fix the build jobs.

costasovo:
@rodrigoprimo Thanks for the quick review!

> One thing that was not clear to me (feel free to merge the PR as it is not a blocker), is why it was necessary to change the place where we call docker-composer create and then call tools/install.php setting COMPOSER_DEV_MODE to 1.

I did move the `docker-composer create` command so that the containers are created before we run any command on them. 
It also unified that we call it in the same place in both acceptance and integration tests. Next time I'll add such a change into a separate commit with an explanation.

The COMPOSER_DEV_MODE is needed because `tools/install.php` checks the env variable and skips installation if it is not present. I'm not sure why we need to check for it.

rodrigoprimo:
Thanks for the additional explanation.

PR #4904, opened 2023-05-03
Dynamic segment filter for first/last name and email [MAILPOET-4997]

johnolek:
## Description

This PR adds a new dynamic segment filter for searching 3 different text columns on the subscribers table:
- `first_name`
- `last_name`
- `email`

### New indexes
Because first and last name values can now be used in complex queries, I've added indexes for those columns in a migration.

I'm not sure if it's really warranted just for this filter since I suspect only a small percentage of users will use the name options. However, I believe the standard subscribers search searches based on first name, last name and email address. That search might not be able to make use of the indexes because it appears to be using a `LIKE %searchString%` search, and I don't believe MySQL can use indexes for searches that begin with a wildcard.

Anyway, I'd like a second opinion about if it's worthwhile to add these indexes.

### Translation context
Does the context I added for the translations make sense?

### Tests
I started writing tests for all cases for all columns, but after adding 5 new operators to the specs that seemed like more work than it would be worth, especially since the code works the same regardless of column. Let me know if you think I should go back to add more tests.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4997](https://mailpoet.atlassian.net/browse/MAILPOET-4997)

## After-merge notes

_N/A_

[MAILPOET-4997]: https://mailpoet.atlassian.net/browse/MAILPOET-4997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/views/segments.html
```diff
@@ -152,12 +152,22 @@
...
+    'notStartsWith': _x('does not start with', 'Dynamic segment creation: operator used with subscriber fields. Should make sense in a sentence like "subscriber email does not start with bill"'),
+    'endsWith': _x('ends with', 'Dynamic segment creation: operator used with subscriber fields. Should make sense in a sentence like "subscriber email ends with yahoo.com"'),
+    'notEndsWith': _x('does not end with', 'Dynamic segment creation: operator used with subscriber fields. Should make sense in a sentence like "subscriber email does not end with yahoo.com"'),
+    'firstName': _x('first name', 'Dynamic segment creation: option in the segment select. Should make sense in a sentence like "subscriber first name does not contain Q"'),
+    'lastName': _x('last name', 'Dynamic segment creation: option in the segment select. Should make sense in a sentence like "subscriber last name does not contain Q"'),
+    'email': _x('email', 'Dynamic segment creation: option in the segment select. Should make sense in a sentence like "subscriber email does not contain Q"'),
```
JanJakes: Context is not meant to be used like that. Each context creates a new string, and it serves to differentiate strings, such as `trigger` (a noun) vs. `trigger` (a verb). If you want to add more "context" in terms of documentation, then a `// translators: ...` comment serves that purpose (it's included with extracted strings and displayed in translation UIs).

mailpoet/lib/Migrations/Migration_20230503_210945.php
```diff
@@ -0,0 +1,17 @@
...
+use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Migrator\Migration;
+
+class Migration_20230503_210945 extends Migration {
+  public function run(): void {
+    $subscribersTable = $this->getTableName(SubscriberEntity::class);
+    $this->connection->executeQuery(
+      "ALTER TABLE `{$subscribersTable}`
+        ADD INDEX `first_name` (`first_name`),
+        ADD INDEX `last_name` (`last_name`)"
```
JanJakes: These columns seem to be `varchar(255)`. We had some issues indexing those, right? I think with `utf8mb4` the maximum indexed column length is 191 (that's why we use it everywhere).

That said, we could just index a much smaller prefix anyway, what do you think? I mean, it could make the indexes huge, if we indexed those 191 maximum chars, and the benefit would probably be small. If most users will type a prefix of just a few chars, then indexing like first 10-30 chars could be enough.

JanJakes (commented):
LGTM, just two small (but important) notes. Thanks!

mailpoet/lib/Migrations/Migration_20230503_210945.php
johnolek: Thanks so much for catching this! I think an index length of 10 should be plenty for name columns so I'm going to go with that.

Edit: updated in e07240c5d7be431e1363de60dc59b3de2ef693bb

mailpoet/views/segments.html
johnolek: Thank you so much for clarifying. I've been following along with the conversations about translations but evidently not closely enough!

I feel like this should be clear from other conversations, but just to confirm, are we not using `translators: ...` comments in our views/templates/twig files because they wouldn't be included in the output the way we process them, is that right? And it is possible to include those comments in typescript files as long as we're using `@wordpress/i18n` functions directly instead of the `MailPoet.I18n` wrapper?

For many of these strings I should be able to re-use existing translations and just make sure they're lowercase. Are translations case sensitive? Would `__('word')` and `__('Word')` result in two translations?

The only translation that stood out to me as a potential issue was `email` since it could be both a verb and a noun, but I think we only ever use it as a noun in the codebase right now so there shouldn't be a need to distinguish. Does that sound right to you?

Edit: I made fixes for the translated strings in e78be61d962d37fcf8ee09e6b48da8c887ca2ec7

johnolek:
Thanks very much for catching those issues, @JanJakes! I think I've addressed them but please let me know if I've missed anything.

JanJakes (dismissed):

mailpoet/views/segments.html
JanJakes: @johnolek

> I feel like this should be clear from other conversations, but just to confirm, are we not using translators: ... comments in our views/templates/twig files because they wouldn't be included in the output the way we process them, is that right? And it is possible to include those comments in typescript files as long as we're using @wordpress/i18n functions directly instead of the MailPoet.I18n wrapper?

With `@wordpress/i18n` the comments should be extracted, but we need to fix our configuration to do that first. For the views and Twig templates, I think it's not worth fixing, we could consider them deprecated.

> For many of these strings I should be able to re-use existing translations and just make sure they're lowercase. Are translations case sensitive? Would __('word') and __('Word') result in two translations?

I'm not sure, but I think they are case sensitive.

> The only translation that stood out to me as a potential issue was email since it could be both a verb and a noun, but I think we only ever use it as a noun in the codebase right now so there shouldn't be a need to distinguish. Does that sound right to you?

Probably no need trying to fix that now, there will be some tickets trying to address these issues.

Aschepikov (approved):
GJ!

PR #4906, opened 2023-05-04
Synchronize Opt out with AutomateWoo

brezocordero:
## Description

This PR covers 2 cases:

- Customer unsubscribes on checkout
- Customer globally unsubscribes to any marketing emails

For those cases, we opt_out the customer in AutomateWoo.

This PR does not cover:
- Changes to WooCommerce segment subscription on bulk updates
- Changes to WooCommerce segment subscription on WooCommerce customer list synchronization

## Code review notes

I have added 2 new action hooks:

1. mailpoet_subscriber_status_changed
Triggered when Subscriber Entity Status changes, using a Doctrine EventListener on postUpdate.
There are limitations to this approach that we have accepted:
- It is not executed when subscriber is updated with a query
- It is not executed on bulk updates

3. mailpoet_woocommerce_segment_unsubscribed
Triggered on checkout when customer is unsubscribed to WooCommerce segment.
For now, this is only added on checkout/order-pay but we might decide later to add it in other places, like the customer list synchronization.
I have added a hook instead of calling directly the AtomateWoo functions. This is a more generic approach that we could use with other integrations.

There is an existing ticket to review the behavior of the opt-in checkbox on checkout. The action on checkout might need to be revisited then.

## QA notes

1. With WooCommerce and AutomateWoo active.
2. Go to AutomateWoo > Settings. Make sure Opt-in is selected and Show on Checkout page
3. Purchase a product and opt-in on checkout. The email should be added to AutomateWoo > Opt-in
4. Purchase a product again with same customer. The opt-in checkbox on checkout needs to be unchecked.
5. Verify the email is removed from AutomateWoo > Opt-ins

6. Repeat steps 1-3
7. On MailPoet> Subscribers, edit the subscriber, set the status to 'unsubscribed'
8. Verify the email is removed from AutomateWoo > Opt-ins

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5219]

## After-merge notes

_N/A_

[MAILPOET-5219]: https://mailpoet.atlassian.net/browse/MAILPOET-5219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/WooCommerce/Integrations/AutomateWooHooks.php
+++ b/mailpoet/lib/WooCommerce/Integrations/AutomateWooHooks.php
@@ -1,0 +1,65 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\WooCommerce\Integrations;
+
+use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Subscribers\SubscribersRepository;
+use MailPoet\WP\Functions as WPFunctions;
+
+class AutomateWooHooks {
+  const AUTOMATE_WOO_PLUGIN_SLUG = 'automatewoo/automatewoo.php';
+
+  /** @var SubscribersRepository */
+  private $subscribersRepository;
+
+  /** @var WPFunctions */
+  private $wp;
+
+  public function __construct(
+    SubscribersRepository $subscribersRepository,
+    WPFunctions $wp
+  ) {
+    $this->subscribersRepository = $subscribersRepository;
+    $this->wp = $wp;
+  }
+
+  public function isAutomateWooActive(): bool {
+    return $this->wp->isPluginActive(self::AUTOMATE_WOO_PLUGIN_SLUG);
+  }
+
+  public function areMethodsAvailable(): bool {
+    return class_exists('AutomateWoo\Customer_Factory') && method_exists('AutomateWoo\Customer_Factory', 'get_by_email') &&
+      class_exists('AutomateWoo\Customer') && method_exists('AutomateWoo\Customer', 'opt_out');
+  }
+
+  public function getAutomateWooCustomer(string $email): ?\AutomateWoo\Customer {
+    // AutomateWoo\Customer_Factory::get_by_email() returns false if customer is not found
+    // Second parameter is set to false to prevent creating new customer if not found
+    return \AutomateWoo\Customer_Factory::get_by_email($email, false);
+  }
+
+  public function setup(): void {
+    if (!$this->isAutomateWooActive() || !$this->areMethodsAvailable()) {
+      return;
+    }
+    $this->wp->addAction(SubscriberEntity::HOOK_SUBSCRIBER_STATUS_CHANGED, [$this, 'maybeOptOutSubscriber'], 10, 1);
+  }
+
+  public function maybeOptOutSubscriber(int $subscriberId): void {
+    if (!$this->isAutomateWooActive() || !$this->areMethodsAvailable()) {
+      return;
+    }
+
+    $subscriber = $this->subscribersRepository->findOneById($subscriberId);
+    if (!$subscriber || !$subscriber->getEmail() || $subscriber->getStatus() !== SubscriberEntity::STATUS_UNSUBSCRIBED) {
+      return;
+    }
+
+    $automateWooCustomer = $this->getAutomateWooCustomer($subscriber->getEmail());
+    if (!$automateWooCustomer) {
+      return;
+    }
+
+    $automateWooCustomer->opt_out();
+  }
+}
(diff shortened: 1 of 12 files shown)


mailpoet/lib/WooCommerce/Integrations/AutomateWooHooks.php
```diff
@@ -0,0 +1,73 @@
...
+    if (!$automateWooCustomer) {
+      return;
+    }
+
+    $automateWooCustomer->opt_out();
+  }
+
+  public function maybeOptOutSubscriber(int $subscriberId): void {
+    $subscriber = $this->subscribersRepository->findOneById($subscriberId);
+    if (!$subscriber || !$subscriber->getEmail() || $subscriber->getStatus() !== SubscriberEntity::STATUS_UNSUBSCRIBED) {
```
rodrigoprimo, line 67: I remember we chatted about supporting opting out of Automate Woo when the subscriber is deleted, and we were wondering what happens when a subscriber is deleted (if it is removed from the database or if we just set the deleted_at field). Since this use case was not mentioned in the PR description, I'm assuming we intentionally decided not to support it for now. I'm mentioning it here just to confirm that it was a conscious decision.

mailpoet/lib/WooCommerce/Subscription.php
```diff
@@ -223,6 +223,7 @@ public function handleSubscriberOptin(SubscriberEntity $subscriber, bool $checko
     $signupConfirmation = $this->settings->get('signup_confirmation');
 
     if (!$checkoutOptin) {
+      $this->wp->doAction('mailpoet_woocommerce_segment_unsubscribed', $subscriber);
```
rodrigoprimo, line 226: As you mentioned in the PR description, when https://mailpoet.atlassian.net/browse/MAILPOET-4178 is implemented, subscribers won't be subscribed anymore during the checkout if the checkbox is unchecked. I'm not sure if we need to implement support for this use case since it is going to be removed soon. Curious to hear what you think. I'm ok to leave it in place if you prefer.

rodrigoprimo (commented):
Code changes look good to me. I left to comments with some minor questions for us to discuss.

mailpoet/lib/WooCommerce/Subscription.php
brezocordero: Thanks for mentioning it, I was not sure when it would be implemented.
If the other issue is going to be ready soon, I can remove this commit.
@NeosinneR thoughts?

mailpoet/lib/WooCommerce/Integrations/AutomateWooHooks.php
brezocordero: Thanks for asking.
Yes, it was intentional. The PR only handles the changes in global status and the checkbox on checkout.
When a subscriber is deleted, we only set the deleted_at field. The global status does not change.

rodrigoprimo: :+1:

mailpoet/lib/WooCommerce/Subscription.php
rodrigoprimo: It is not a trivial ticket but unless we hit a blocker, it should be ready next week I guess. But happy to proceed with this PR as is. Let's wait for Ján's input.

NeosinneR: @rodrigoprimo if you feel your ticket will be released in the next week or two, @brezocordero can remove that case from this PR, which would only leave "Customer globally unsubscribes to any marketing emails" - that's the major one anyway.

Even in the optimistic scenario, MailPoet won't be bundled in Woo Express this month, so it's safe to assume that by the time MailPoet will, you won't be able to unsubscribe on the checkout - no point in adding new code around that.

rodrigoprimo: I think it will, but at this point, I'm not 100% sure.

I am unsure regarding the best approach. If it is easier, we can merge this PR as is, and I can remove the code related to handling subscriptions during checkout.

Alternatively, we can mark the ticket related to this PR as blocked by https://mailpoet.atlassian.net/browse/MAILPOET-4178. If I manage to finish MAILPOET-4178 in the next week or two, we merge this PR without the code to handle subscriptions during checkout.

Happy to proceed however you prefer, @brezocordero.

brezocordero: @rodrigoprimo 
I think it might be easier to merge it and then remove the action hook when you work on your ticket.
You might find that there is a case when we do have to unsubscribe a customer to the WooCommerce segment and in that case we already have the structure for the integration.

brezocordero: There is also an acceptance test for when they opt-in/opt-out with AutomateWoo.
It will be good to keep the test and change the outcome to reflect the changes on your ticket.

rodrigoprimo: Sounds good to me. Moving this PR to QA.

rodrigoprimo (approved):

Aschepikov (approved):
GJ!

PR #4908, opened 2023-05-05
Refactor/remove some methods from ModelValidator [MAILPOET-4343]

rodrigoprimo:
## Description

This PR refactors/removes some of the methods from the old ModelValidator class. Now this class is only used in the context of the remaining Paris models.

## Code review notes

@lysyjan, there are two things that I did in this PR that I'm not sure if they match what you had in mind when you created the ticket:

- Two public methods were left in the ModelValidator class (validateEmail() and validateRenderedNewsletterBody()) as they are still used in the Paris validator system. My suggestion is that we remove them only when we finish removing the remaining models. I assumed it was not worth the effort of checking how to use them in the models without using the Paris validator system.
- The ticket description suggests when possible to move functionality from ModelValidator to a service. I created a new \MailPoet\Services\Validator class. I'm not sure if this is the service that you had in mind or if it was something else.

Let me know if you prefer a different approach.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4343]

## After-merge notes

_N/A_

[MAILPOET-4343]: https://mailpoet.atlassian.net/browse/MAILPOET-4343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Services/Validator.php
+++ b/mailpoet/lib/Services/Validator.php
@@ -1,0 +1,60 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Services;
+
+use MailPoet\WP\Functions as WPFunctions;
+
+/**
+ * This class contains validation methods that were extracted from the \MailPoet\Models\ModelValidator class.
+ * It is used only in a few places and there is a chance in the future we can remove it.
+ */
+class Validator {
+  const EMAIL_MIN_LENGTH = 6;
+  const EMAIL_MAX_LENGTH = 150;
+  const ROLE_EMAILS = [
+    'abuse',
+    'compliance',
+    'devnull',
+    'dns',
+    'ftp',
+    'hostmaster',
+    'inoc',
+    'ispfeedback',
+    'ispsupport',
+    'list-request',
+    'list',
+    'maildaemon',
+    'noc',
+    'no-reply',
+    'noreply',
+    'nospam',
+    'null',
+    'phish',
+    'phishing',
+    'postmaster',
+    'privacy',
+    'registrar',
+    'root',
+    'security',
+    'spam',
+    'sysadmin',
+    'undisclosed-recipients',
+    'unsubscribe',
+    'usenet',
+    'uucp',
+    'webmaster',
+    'www',
+  ];
+
+  public function validateEmail($email) {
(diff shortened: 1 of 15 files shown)


mailpoet/lib/Services/Validator.php
```diff
@@ -0,0 +1,60 @@
...
+    'sysadmin',
+    'undisclosed-recipients',
+    'unsubscribe',
+    'usenet',
+    'uucp',
+    'webmaster',
+    'www',
+  ];
+
+  public function validateEmail($email) {
```
lysyjan: I know this code is moved from an old class, but maybe we could add return types here.

lysyjan (dismissed):
@rodrigoprimo Good job 👍 It looks good.
I have noticed one detail in the moved code from the ModelValidator, but it's up to you if you want to address it. Feel free to move it to QA if you consider it irrelevant.

> Two public methods were left in the ModelValidator class (validateEmail() and validateRenderedNewsletterBody()) as they are still used in the Paris validator system. My suggestion is that we remove them only when we finish removing the remaining models. I assumed it was not worth the effort of checking how to use them in the models without using the Paris validator system.

I like this idea 👍 The main goal of this task was to move used methods into our codebase and make this class independent on Doctrine.

> The ticket description suggests when possible to move functionality from ModelValidator to a service. I created a new \MailPoet\Services\Validator class. I'm not sure if this is the service that you had in mind or if it was something else.

Yes, your service is something that I expected 👍

mailpoet/lib/Services/Validator.php
rodrigoprimo: I'm ok with adding return types here. Just pushed a commit with that. Could you please take a second look?

I considered adding types to the parameters as well, but it was not straightforward due to the methods being called in places where the type of the parameters is not defined, so I decided not to do it.

lysyjan (dismissed):

Aschepikov:
Hi @rodrigoprimo There are some conflicts, can you resolve them pls

rodrigoprimo:
@Aschepikov, I just fixed the conflict. Can you please check again? Thanks.

Aschepikov (dismissed):
GJ!

Aschepikov (approved):
GJ!

PR #4917, opened 2023-05-11
Modify MailPoet when bundled and using AutomateWoo

brezocordero:
## Description

This PR removes the following MailPoet Automations when the plugin is using an API key that belongs to a subscription bundled with a Dotcom eCommerce plan and AutomateWoo is active:
- Automations Menu entry from MailPoet Menu
- WooCommerce tab on emails
- Automated WooCommerce emails (First purchase etc)

WooCommerce email customizer is still available.

The PR also adds a new box on emails > new, linking to AutomateWoo

When there is no API key or the API key does not belong to a bundle, there are no changes.

## Code review notes

- @costasovo I have assigned it to you as you have recently worked on the API key data storage. Please check I am not missing anything on the changes in the Bridge and ServiceChecker. 🙌 
- The API key will be automatically provisioned using [DotcomLicenseProvisioner](https://github.com/mailpoet/mailpoet/blob/dddd116b34121d3c3bfa77b533e403cdeafb82f8/mailpoet/lib/WPCOM/DotcomLicenseProvisioner.php#L54)
- The subscription type is not yet available, it will be added on SS-355

## QA notes

Changes can not be manually tested.

Please test there are no regressions in:
- MailPoet Menu
- MailPoet Emails tabs
- MailPoet Emails > New

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5223]

## After-merge notes

_N/A_

[MAILPOET-5223]: https://mailpoet.atlassian.net/browse/MAILPOET-5223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/newsletters/types.tsx
+++ b/mailpoet/assets/js/src/newsletters/types.tsx
@@ -103,12 +103,48 @@ function NewsletterTypesComponent({
     );
   };
 
+  const getRedirectToAutomateWooType = () => {
+    const redirectToAutomateWoo = (): void => {
+      MailPoet.trackEvent('Emails > Type selected', {
+        'Email type': 'woocommerce_automatewoo',
+      });
+      window.location.href = `edit.php?post_type=aw_workflow#presets`;
+      return null;
+    };
+    return {
+      slug: 'woocommerce_automatewoo',
+      title: __('Automations', 'mailpoet'),
+      description: __(
+        'Convert and retain customers with automated marketing that does the hard work for you. AutomateWoo has the tools you need to grow your store and make more money.',
+        'mailpoet',
+      ),
+      action: (
+        <Button
+          automationId="woocommerce_automatewoo"
+          onClick={redirectToAutomateWoo}
+          tabIndex={0}
+          onKeyDown={(event): void => {
+            if (
+              ['keydown', 'keypress'].includes(event.type) &&
+              ['Enter', ' '].includes(event.key)
+            ) {
+              event.preventDefault();
+              redirectToAutomateWoo();
+            }
+          }}
+        >
+          {__('Set up', 'mailpoet')}
+        </Button>
+      ),
+    };
+  };
+
   const getAdditionalTypes = (): Record<string, unknown>[] => {
     const show = MailPoet.isWoocommerceActive;
     if (!show) {
       return [];
     }
-    return [
+    const additionalTypes = [
       {
         slug: 'wc_transactional',
         title: __('WooCommerce Emails Customizer', 'mailpoet'),
@@ -136,6 +172,10 @@ function NewsletterTypesComponent({
         ),
       },
     ];
+    if (MailPoet.hideAutomations) {
+      additionalTypes.push(getRedirectToAutomateWooType());
+    }
+    return additionalTypes;
   };
 
   const getAutomaticEmails = (): JSX.Element[] => {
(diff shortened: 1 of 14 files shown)


mailpoet/assets/js/src/newsletters/types.tsx
```diff
@@ -103,12 +103,48 @@ function NewsletterTypesComponent({
...
   };
 
+  const getRedirectToAutomateWooType = () => {
+    const redirectToAutomateWoo = (): void => {
+      MailPoet.trackEvent('Emails > Type selected', {
+        'Email type': 'woocommerce_automatewoo',
+      });
+      window.location.href = `edit.php?post_type=aw_workflow#presets`;
+      return null;
+    };
```
costasovo: Event tracking is async, so with this implementation, the redirect may happen before the event is sent to Mixpanel. So I think not all events would be sent.

Here is the code that should ensure sending the event before the redirect
```JS
MailPoet.trackEvent(
    'Emails > Type selected',
    {
        'Email type': 'woocommerce_automatewoo',
     },
    { send_immediately: true }, // This tell Mixpanel client to send events from buffer immediately
    () => { // Anonymous callback which does the redirect
      window.location.href = `edit.php?post_type=aw_workflow#presets`;
    },
  );
```

mailpoet/lib/Config/ServicesChecker.php
```diff
@@ -133,6 +133,14 @@ public function isPremiumKeyValid($displayErrorNotice = true) {
     return false;
   }
 
+  public function isBundledSubscription(): bool {
+    if ($this->isMailPoetAPIKeyValid(false, true) || $this->isPremiumKeyValid(false)) {
```
costasovo: If I understand correctly, in case we disable the key on the shop `isBundledSubscription` will start returning `false`. The shop may block the key due to overreaching limits or sending spam. With the current implementation on the Bridge, the `/me` and `/premium` endpoints will start returning 403 or another code and keys will not be marked as valid.

So for users, this could lead to suddenly seeing non-premium variants of the automation and Woo features.
Maybe we could drop the check here? That would, IMO mean asking for the last known state from when the key was valid.

Or another option could be using `$this->getValidAccountKey() !== null)`. The getValidAccountKey method checks if the key is valid or valid but underprivileged. Such a key doesn't grant access to MSS or Premium but still can be used e.g. for reporting stats, and Bridge knows it exists. So the `isBundledSubscription` would also work for blocked keys but not for spam issues because the shop currently blocks spammers on the account level and removes all keys from the bridge (this will hopefully change in the future). But still its slightly better than checking only valid states.

costasovo (changes requested):
Thanks for the nice PR. The changes are nicely organized into commits. That makes it easier to review. Thanks also for adding acceptance tests.

I left two comments. I'm selecting request changes because I believe the comment about tracking needs to be addressed.

mailpoet/lib/Config/ServicesChecker.php
brezocordero: Thank you so much for explaining the behavior of spam and blocked cases.

> Maybe we could drop the check here? That would, IMO mean asking for the last known state from when the key was valid.

This is a good idea, it will provide a more consistent experience and work better for spam and blocked keys.
I have removed the check.

mailpoet/assets/js/src/newsletters/types.tsx
brezocordero: 🐛 Thanks for catching this. My mistake.
Fixed using your suggestion.

costasovo (approved):
Thank you! 👍

Aschepikov (approved):
GJ!

PR #4920, opened 2023-05-11
Add ability to duplicate dynamic segments [MAILPOET-4635]

johnolek:
## Description

This PR adds the ability to duplicate dynamic segments in the UI.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4635](https://mailpoet.atlassian.net/browse/MAILPOET-4635)

## After-merge notes

_N/A_

[MAILPOET-4635]: https://mailpoet.atlassian.net/browse/MAILPOET-4635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/SegmentSaveController.php
+++ b/mailpoet/lib/Segments/DynamicSegments/SegmentSaveController.php
@@ -6,6 +6,7 @@ use MailPoet\ConflictException;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\NotFoundException;
 use MailPoet\Segments\SegmentsRepository;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
 use MailPoetVendor\Doctrine\ORM\ORMException;
 
 class SegmentSaveController {
@@ -15,12 +16,17 @@ class SegmentSaveController {
   /** @var FilterDataMapper */
   private $filterDataMapper;
 
+  /** @var EntityManager */
+  private $entityManager;
+
   public function __construct(
     SegmentsRepository $segmentsRepository,
-    FilterDataMapper $filterDataMapper
+    FilterDataMapper $filterDataMapper,
+    EntityManager $entityManager
   ) {
     $this->segmentsRepository = $segmentsRepository;
     $this->filterDataMapper = $filterDataMapper;
+    $this->entityManager = $entityManager;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


johnolek:
@costasovo I think you've worked on related parts of the codebase which is why I'd like to get your input on this PR, but if you're busy with other reviews I'm happy to reassign this!

mailpoet/lib/Segments/DynamicSegments/SegmentSaveController.php
```diff
@@ -37,4 +43,23 @@ public function save(array $data = []): SegmentEntity {
 
     return $this->segmentsRepository->createOrUpdate($name, $description, SegmentEntity::TYPE_DYNAMIC, $filtersData, $id);
   }
+
+  public function duplicate(SegmentEntity $segmentEntity): SegmentEntity {
+    $duplicate = clone $segmentEntity;
+    // translators: %s is the name of the segment
+    $duplicate->setName(sprintf(__('Copy of %s', 'mailpoet'), $segmentEntity->getName()));
+    $this->segmentsRepository->verifyNameIsUnique($duplicate->getName(), $duplicate->getId());
+    $this->entityManager->wrapInTransaction(function(EntityManager $entityManager) use ($duplicate, $segmentEntity) {
```
costasovo, line 52: Good idea to use a transaction here 👍  We should do it more often.

costasovo (approved):
Great job!

mailpoet/lib/Segments/DynamicSegments/SegmentSaveController.php
johnolek: Thanks! But I can't take credit for the idea. We do it in a few other places, including duplicating lists/non-dynamic segments (see [here](https://github.com/mailpoet/mailpoet/blob/38e4b4d2faf3a36271c0f70d330073ed1ebfb484/mailpoet/lib/Segments/SegmentSaveController.php#L51-L51)). Those other places are using the deprecated-as-of-doctrine-orm-2.10 `transactional` instead of `wrapInTransaction`.

I agree that more transactions could be a great idea for related operations like these. The only potential downside I could see to using transactions more frequently is that they're not supported by MyISAM. There are still plenty of people who have MyISAM tables, so we could end up with different behavior for different configurations.

Aschepikov (approved):
GJ!

PR #4921, opened 2023-05-11
Change the behavior of the checkbox to subscribe on the checkout [MAILPOET-4178]

rodrigoprimo:
## Description

MailPoet adds a checkbox to the WooCommerce checkout to let users subscribe to mailing lists. This PR changes its behavior. It affects both the regular and the block checkout.

Now the checkbox is not checked by default for logged in users who already subscribed to a list. Also, users won't be unsubscribed anymore and have their global status change to unsubscribed if they uncheck the checkbox.

After the changes proposed here, the only action that is performed during checkout is to subscribe users if they check the checkbox.

It was necessary to change the behavior of
\MailPoet\Segments\WooCommerce::synchronizeRegisteredCustomer(). This method is called when a new WooCommerce customer is created. Before, it would always add the customer to the Woo segment expecting \MailPoet\WooCommerce\Subscription::subscribeOnCheckout() to unsubscribe the customer if the checkbox was not checked. Since the behavior of the latter method changed, the former method was updated to only subscriber the customer if not in checkout context or if in checkout context and optin checkbox is enabled and checked.

## Code review notes

I left a few GitHub comments in some parts of the code.

I checked the discussion about the complexities and inconsistencies in the MailPoet integration with WooCommerce (pcD9cT-1Ym-p2), but I'm afraid I was not able to help much in terms of simplifying this part of the code.

## QA notes

This PR touches a core part of the MailPoet integration with WooCommerce. The optin checkbox in the WooCommerce checkout. It seems to have extensive automated test coverage, but it might still be worth checking all possible paths during checkout with different configurations (guest checkout, logged in checkout, create account during checkout, optin checkbox enabled and disabled, confirmation email enabled and disabled) just to make sure nothing was missed. Pay special attention to the lists and the status of the subscriber after checkout.

Maybe it i also worth checking other ways to create a subscriber like when a new WP user is registered, creating a subscriber via wp-admin and creating subscribers for all existing WooCommerce customers.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4178]

## After-merge notes

_N/A_

[MAILPOET-4178]: https://mailpoet.atlassian.net/browse/MAILPOET-4178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
+++ b/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
@@ -95,7 +95,7 @@ class WooCommerceBlocksIntegration {
       [
       'defaultText' => $this->settings->get('woocommerce.optin_on_checkout.message', ''),
       'optinEnabled' => $this->settings->get('woocommerce.optin_on_checkout.enabled', false),
-      'defaultStatus' => $this->woocommerceSubscription->isCurrentUserSubscribed(),
+      'defaultStatus' => false,
       ],
       $this->wp
     ));
@@ -129,7 +129,6 @@ class WooCommerceBlocksIntegration {
   }
 
   public function processCheckoutBlockOptin(\WC_Order $order, $request) {
-    $checkoutOptinEnabled = (bool)$this->settings->get(WooCommerceSubscription::OPTIN_ENABLED_SETTING_NAME);
     $checkoutOptin = isset($request['extensions']['mailpoet']['optin']) ? (bool)$request['extensions']['mailpoet']['optin'] : false;
 
     if (!$order->get_billing_email()) {
@@ -149,6 +148,6 @@ class WooCommerceBlocksIntegration {
       return null;
     }
 
-    $this->woocommerceSubscription->handleSubscriberOptin($subscriber, $checkoutOptinEnabled, $checkoutOptin);
+    $this->woocommerceSubscription->handleSubscriberOptin($subscriber, $checkoutOptin);
   }
 }
--- a/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
+++ b/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
@@ -95,7 +95,6 @@ class WooCommerceBlocksIntegration {
       [
       'defaultText' => $this->settings->get('woocommerce.optin_on_checkout.message', ''),
       'optinEnabled' => $this->settings->get('woocommerce.optin_on_checkout.enabled', false),
-      'defaultStatus' => false,
       ],
       $this->wp
     ));
--- a/mailpoet/lib/WooCommerce/Helper.php
+++ b/mailpoet/lib/WooCommerce/Helper.php
@@ -227,4 +227,23 @@ class Helper {
   public function getPaymentGateways() {
     return $this->WC()->payment_gateways();
   }
+
+  /**
+   * Check whether the current request is processing a WooCommerce checkout.
+   * Works for both the normal checkout and the block checkout.
+   *
+   * This solution is not ideal, but I cheched with a few WooCommerce developers,
+   * and it is what they suggested. There is no helper function provided by Woo
+   * for this.
+   *
+   * @return bool
+   */
+  public function isCheckoutRequest(): bool {
+    $requestUri = !empty($_SERVER['REQUEST_URI']) ? sanitize_text_field(wp_unslash($_SERVER['REQUEST_URI'])) : '';
+    $isRegularCheckout = is_checkout();
+    $isBlockCheckout = WC()->is_rest_api_request()
+      && (strpos($requestUri, 'wc/store/checkout') !== false || strpos($requestUri, 'wc/store/v1/checkout') !== false);
+
+    return $isRegularCheckout || $isBlockCheckout;
+  }
 }
--- a/mailpoet/lib/WooCommerce/Helper.php
+++ b/mailpoet/lib/WooCommerce/Helper.php
@@ -232,7 +232,7 @@ class Helper {
    * Check whether the current request is processing a WooCommerce checkout.
    * Works for both the normal checkout and the block checkout.
    *
-   * This solution is not ideal, but I cheched with a few WooCommerce developers,
+   * This solution is not ideal, but I checked with a few WooCommerce developers,
    * and it is what they suggested. There is no helper function provided by Woo
    * for this.
    *
--- a/mailpoet/tests/acceptance/Subscriptions/WooCheckoutGuestSubscriptionsCest.php
+++ b/mailpoet/tests/acceptance/Subscriptions/WooCheckoutGuestSubscriptionsCest.php
@@ -55,7 +55,7 @@ class WooCheckoutGuestSubscriptionsCest {
     $customerEmail = 'woo_guest_uncheck@example.com';
     $i->orderProductWithoutRegistration($this->product, $customerEmail, false);
     $i->login();
-    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_UNSUBSCRIBED, null, ['WooCommerce Customers']);
+    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_UNCONFIRMED, null, ['WooCommerce Customers']);
     $i->seeConfirmationEmailWasNotReceived();
   }
 
@@ -85,7 +85,7 @@ class WooCheckoutGuestSubscriptionsCest {
     $customerEmail = 'woo_guest_uncheck@example.com';
     $i->orderProductWithoutRegistration($this->product, $customerEmail, false);
     $i->login();
-    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_UNSUBSCRIBED, null, ['WooCommerce Customers']);
+    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_SUBSCRIBED, null, ['WooCommerce Customers']);
     $i->seeConfirmationEmailWasNotReceived();
   }
 
--- a/mailpoet/assets/js/src/global.d.ts
+++ b/mailpoet/assets/js/src/global.d.ts
@@ -60,7 +60,6 @@ declare module '@woocommerce/settings' {
   interface MailPoetSettings {
     optinEnabled: boolean;
     defaultText: string;
-    defaultStatus: boolean;
   }
 
   function getSetting(name: 'mailpoet_data'): MailPoetSettings;
--- a/mailpoet/assets/js/src/marketing_optin_block/block.tsx
+++ b/mailpoet/assets/js/src/marketing_optin_block/block.tsx
@@ -6,8 +6,7 @@ import { CheckboxControl } from '@woocommerce/blocks-checkout';
 import { RawHTML, useEffect, useState } from '@wordpress/element';
 import { getSetting } from '@woocommerce/settings';
 
-const { optinEnabled, defaultText, defaultStatus } =
-  getSetting('mailpoet_data');
+const { optinEnabled, defaultText } = getSetting('mailpoet_data');
 
 export function FrontendBlock({
   text,
@@ -18,7 +17,7 @@ export function FrontendBlock({
     setExtensionData: (namespace: string, key: string, value: unknown) => void;
   };
 }): JSX.Element {
-  const [checked, setChecked] = useState(defaultStatus);
+  const [checked, setChecked] = useState(false);
   const { setExtensionData } = checkoutExtensionData || {};
   useEffect(() => {
     if (optinEnabled && setExtensionData) {
--- a/mailpoet/lib/Segments/WP.php
+++ b/mailpoet/lib/Segments/WP.php
@@ -17,7 +17,6 @@ use MailPoet\Subscribers\Source;
 use MailPoet\Subscribers\SubscriberSegmentRepository;
 use MailPoet\Subscribers\SubscribersRepository;
 use MailPoet\WooCommerce\Helper as WooCommerceHelper;
-use MailPoet\WooCommerce\Subscription as WooCommerceSubscription;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
 use MailPoetVendor\Idiorm\ORM;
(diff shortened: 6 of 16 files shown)


mailpoet/tests/acceptance/Subscriptions/WooCheckoutGuestSubscriptionsCest.php
```diff
@@ -85,7 +85,7 @@ public function checkoutOptInUncheckedNoConfirmation(\AcceptanceTester $i) {
     $customerEmail = 'woo_guest_uncheck@example.com';
     $i->orderProductWithoutRegistration($this->product, $customerEmail, false);
     $i->login();
-    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_UNSUBSCRIBED, null, ['WooCommerce Customers']);
+    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_SUBSCRIBED, null, ['WooCommerce Customers']);
```
rodrigoprimo, line 88: This is a change that I'm not sure about. Before the code would create the subscriber as `subscribed` in `\MailPoet\Segments\WooCommerce::synchronizeGuestCustomer()` and then the status would be changed to `unsubscribed` in `\MailPoet\WooCommerce\Subscription::handleSubscriberOptin()` since the optin checkbox is unchecked. Now that we changed the behavior of `handleSubscriberOptin()` and it doesn't unsubscribe subscribers anymore, in the case of this test, we are left with a subscriber with the status `subscribed` but without any list which seems a bit odd, at the same time, I'm not sure if there is a better status.

mailpoet/lib/WooCommerce/Helper.php
```diff
@@ -223,4 +223,23 @@ public function getLatestCoupon(): ?string {
...
+   * This solution is not ideal, but I cheched with a few WooCommerce developers,
+   * and it is what they suggested. There is no helper function provided by Woo
+   * for this.
+   *
+   * @return bool
+   */
+  public function isCheckoutRequest(): bool {
+    $requestUri = !empty($_SERVER['REQUEST_URI']) ? sanitize_text_field(wp_unslash($_SERVER['REQUEST_URI'])) : '';
+    $isRegularCheckout = is_checkout();
+    $isBlockCheckout = WC()->is_rest_api_request()
```
rodrigoprimo, line 244: The way that I'm using to check if it is the regular checkout and the block checkout is a bit hacky but it is the best I could find. I also discussed a bit with Woo core developers and they didn't have a better suggestion.

`is_checkout()` is meant to return true when the page being displayed is the checkout page. It is not meant to return true when processing the checkout request. But since the request is made to the checkout page, it works.

Regarding the block checkout, I could not find any WP or Woo method that would work, so I ended up with this approach that relies on the path of the checkout endpoint, which is a brittle solution. At least, a test is likely to break if the endpoint changes in the future (for example, a new version of the store API).

mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
```diff
@@ -95,7 +95,7 @@ public function registerCheckoutFrontendBlocks($integration_registry) {
       [
       'defaultText' => $this->settings->get('woocommerce.optin_on_checkout.message', ''),
       'optinEnabled' => $this->settings->get('woocommerce.optin_on_checkout.enabled', false),
-      'defaultStatus' => $this->woocommerceSubscription->isCurrentUserSubscribed(),
+      'defaultStatus' => false,
```
johnolek: Just a small thing: since this will always be false, is there a reason we wouldn't want to remove this entirely and make a change to `mailpoet/assets/js/src/marketing_optin_block/block.tsx` instead?

mailpoet/lib/WooCommerce/Helper.php
```diff
@@ -223,4 +223,23 @@ public function getLatestCoupon(): ?string {
 
     return $coupon ? $coupon->post_title : null; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
   }
+
+  /**
+   * Check whether the current request is processing a WooCommerce checkout.
+   * Works for both the normal checkout and the block checkout.
+   *
+   * This solution is not ideal, but I cheched with a few WooCommerce developers,
```
johnolek: No need to change but a tiny typo

```suggestion
   * This solution is not ideal, but I checked with a few WooCommerce developers,
```

mailpoet/lib/WooCommerce/Helper.php
johnolek: Did you explore checking to see if certain actions have fired? I was poking around at this a bit and was wondering if we could check for `did_action('woocommerce_store_api_checkout_order_processed')`. In my testing it returns `1` at this point in the code when checking out via the block checkout, and here's a comment from the [source code](https://github.com/woocommerce/woocommerce-blocks/blob/cde9117c9d57fe9a4c81501214d837e984a6c76a/src/StoreApi/Routes/V1/Checkout.php#L250-L266) (the second point specifically):

> 		 * This is similar to existing core hook woocommerce_checkout_order_processed. We're using a new action:
>		 * - To keep the interface focused (only pass $order, not passing request data).
>		 * - This also explicitly indicates these orders are from checkout block/StoreAPI.

mailpoet/tests/acceptance/Subscriptions/WooCheckoutGuestSubscriptionsCest.php
johnolek: This does seem a bit odd. I'm not sure what the best solution is. One behavioral change is that these users no longer immediately appear when you look at the WooCommerce Customers list. However, if you run the "force sync" operation, they do get added to the list with a status of unsubscribed.

There are other instances where we have inconsistencies between normal subscribing behavior vs running syncs but I wouldn't want to introduce another if we can avoid it. That being said, I think I remember running into this exact issue when I was looking into this flow many months ago and I remember it not seeming like there was a good solution.

johnolek (changes requested):
Nice work on this, @rodrigoprimo! This is not always an easy part of the codebase to modify.

I have a couple of minor comments, an idea for checking for block checkout that might not work, and a comment about the change to the opt-in disabled checkbox unchecked flow.

I definitely agree with your comment about this requiring some extensive QA. Depending on how you decide to go with the optin-disabled checkbox unchecked flow, it might be good to include a note to QA about expected changes there.

mailpoet/tests/acceptance/Subscriptions/WooCheckoutGuestSubscriptionsCest.php
rodrigoprimo: I'm not sure what is the best solution here. cc @NeosinneR for your input from the product perspective.

mailpoet/lib/WooCommerce/Helper.php
rodrigoprimo: I haven't considered using `did_action('woocommerce_store_api_checkout_order_processed')`. Thanks for mentioning it.

Initially, I thought that I preferred the approach that you suggested, and I implemented it. I ended up using `woocommerce_store_api_checkout_update_order_from_request` instead of `woocommerce_store_api_checkout_order_processed` as the latter runs after the MailPoet code to update the subscriber.

But then I realized that if we decide to use this action, `isCheckoutRequest()` will only work if called after a specific WooCommerce action is executed, and it might cause some unexpected behavior if, in the future, we decide to call it before the action is executed in a different part of our code. The current solution seems more generic to me and not significantly more brittle as WooCommerce could deprecate the actions that we are considering using.

So, I'm torn in terms of which approach is best. What do you think?

mailpoet/lib/WooCommerce/Helper.php
rodrigoprimo: Thanks, I committed your suggestion.

mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
rodrigoprimo: Good point, I missed that. I pushed a new commit implementing this change. Could you please check? I might have missed something, as it has been a while since the last time that I touched React code.

rodrigoprimo:
Thanks for the review, @johnolek. I believe I address all your points. Could you please take a second look?

mailpoet/lib/WooCommerce/Helper.php
johnolek: > it might cause some unexpected behavior if, in the future, we decide to call it before the action is executed in a different part of our code.

This is a very good point, and I appreciate the amount of thought you've put into this! I agree that it's better to go with the current solution for the reason you mention.

When I was looking at this I thought it might be nice if we could access the current `WP_REST_Request` and check for the route it's executing instead of having to rely on `$_SERVER` values, but as far as I can tell there's not a great way to achieve this, certainly not any way that would be an improvement over your current solution.

mailpoet/tests/acceptance/Subscriptions/WooCheckoutGuestSubscriptionsCest.php
NeosinneR: What about leaving it `unconfirmed` (globally) and letting `\MailPoet\WooCommerce\Subscription::handleSubscriberOptin()` set it to subscribed if the checkbox was checked?

mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
johnolek: The change looks good to me. The only thing I noticed is that there's still a `defaultStatus` in [this typescript interface](https://github.com/mailpoet/mailpoet/blob/2b6b04e31116ce45edfed532005aa1ca3e3fc769/mailpoet/assets/js/src/global.d.ts#L63-L63). Unless that's somehow referring to something else I imagine it would be best to remove that as well so typescript won't think something exists that doesn't.

johnolek:
@rodrigoprimo The only last thing I would want to confirm is that we're ok with the minor behavioral change I mentioned in [this comment](https://github.com/mailpoet/mailpoet/pull/4921#discussion_r1192641867), unless you made a change to that behavior that I didn't notice?

mailpoet/lib/WooCommerce/Helper.php
rodrigoprimo: Yeah, I also considered using the current `WP_Rest_Request` but came to the same conclusion that you did. Anyway, thanks for considering and suggesting other approaches.

mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
rodrigoprimo: I missed that. Since it was the last commit and it is a small one, I amended it to remove `defaultStatus` from the typescript interface and force pushed.

mailpoet/tests/acceptance/Subscriptions/WooCheckoutGuestSubscriptionsCest.php
rodrigoprimo: > What about leaving it unconfirmed (globally) and letting \MailPoet\WooCommerce\Subscription::handleSubscriberOptin() set it to subscribed if the checkbox was checked?

The status is set to `subscribed` when the email confirmation is disabled (I should have mentioned this in the initial comment). So I'm not sure if we want to leave the user with the status `unconfirmed` since email confirmation is disabled. This is what the code currently does:

https://github.com/mailpoet/mailpoet/blob/b80b7d9b1c19156b742354d8cb1a66e412e0bb59/mailpoet/lib/Segments/WooCommerce.php#L165-L168

> One behavioral change is that these users no longer immediately appear when you look at the WooCommerce Customers list. However, if you run the "force sync" operation, they do get added to the list with a status of unsubscribed.

@NeosinneR, what about the behavioral change that @johnolek mentioned? Are we ok with that?

NeosinneR: > what about the behavioral change that @johnolek mentioned? Are we ok with that?

But in the end, nothing changes functionally, no? Both before and after this PR, you can't contact these contacts, and they are not counted towards limits. The difference is only that they are not immediately seen. If that's the case, I think that's an acceptable change.

johnolek (dismissed):
LGTM! It looks like there may be some conflicts to resolve FYI.

mailpoet/tests/acceptance/Subscriptions/WooCheckoutGuestSubscriptionsCest.php
rodrigoprimo: Yeah, my understanding is that nothing changes in terms of functionality. Thanks for your input.

@johnolek, are you ok if we proceed with this PR the way it is? Please let me know if you prefer that we explore different options.

rodrigoprimo: @johnolek, ignore my previous comment. I just saw that you approved the PR. I will check the conflicts.

rodrigoprimo:
@brezocordero, when you get a chance, could you please take a look at this PR? As we discussed in https://github.com/mailpoet/mailpoet/pull/4906/, there was a conflict here related to the integration code that you added when users unsubscribed during checkout.

I had to remove the action that you added to Subscription.php, the code that used that action and a related integration test. I also had to update another integration test. All the changes are in commit 6ec1cf559c4f199bd943978361ba6692a0ab1981.

I just want to make sure I removed everything that I should and I didn't remove anything that I should have left.

brezocordero:
@rodrigoprimo thanks for checking.

> I just want to make sure I removed everything that I should and I didn't remove anything that I should have left.

Yes, it looks good, as it is not possible to unsubscribe to the WooCommerce segment on checkout anymore, the hook that integrates with AutomateWoo can be removed and your changes look good.

rodrigoprimo:
@johnolek, since I rebased this PR to fix the conflict, your review became stale. I will need another review to move this PR to QA. Can you please check it again?

johnolek (dismissed):

Aschepikov:
Hi, @rodrigoprimo there are some conflicts, can you resolve them pls

rodrigoprimo:
@Aschepikov, I have fixed the conflict. The CircleCI build is currently failing, but it is not related to the changes here. Trunk is also failing. See related discussion: p1685021410730029-slack-C01H71ZLBGQ

Aschepikov:
Hi @rodrigoprimo When I have unable opt-in checkbox I still can see the checkbox on the checkout page if I am not logged in. 
image
image


If I am logged in I don't see the checkbox.
image
image

I think it is not expected behavior but if I will check this checkbox it won't work as subscribing to the woocommerce list

Aschepikov:
On the current prod, I don't see the checkbox on the checkout page in both cases. Can you take a look pls

rodrigoprimo:
@Aschepikov, if I understood correctly, I'm not able to reproduce the behavior that you are describing. I don't see the opt-in checkbox on the checkout page for anonymous users when the corresponding option is disabled. Please see the video below. Am I missing some steps to reproduce the problem?


https://github.com/mailpoet/mailpoet/assets/77215/fcc0b929-6f06-4867-b327-063341e50a6f

NeosinneR:
@rodrigoprimo @Aschepikov is the checkbox maybe from another plugin, like AutomateWoo?

rodrigoprimo:
Good catch, @NeosinneR. I assumed that @Aschepikov had changed the default text of the MailPoet checkbox, but it seems, indeed, this checkbox is actually generated by AutomateWoo and not MailPoet: https://github.com/woocommerce/automatewoo/blob/4614e36d5164a98be2a9c17a94efcd1f615f4384/includes/Options.php#L55. At least the string is exactly the same.

@Aschepikov, can you please confirm?

Aschepikov:
@rodrigoprimo  Yep, I didn't deactivate automateWoo. In that case, everything is ok.

Aschepikov (approved):
GJ!

PR #4922, opened 2023-05-12
Analytics page [MAILPOET-5088] [MAILPOET-5090]

websupporter:
## Description
Adds the analytics page with the overview metrics

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/725

## Linked tickets
[MAILPOET-5088] [MAILPOET-5090]

[MAILPOET-5088]: https://mailpoet.atlassian.net/browse/MAILPOET-5088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5090]: https://mailpoet.atlassian.net/browse/MAILPOET-5090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/OverviewStatisticsController.php
```diff
@@ -0,0 +1,137 @@
...
+use MailPoet\Automation\Integrations\MailPoet\Actions\SendEmailAction;
+use MailPoet\Automation\Integrations\MailPoet\Analytics\Entities\Query;
+use MailPoet\Entities\StatisticsClickEntity;
+use MailPoet\Entities\StatisticsOpenEntity;
+use MailPoet\Newsletter\NewslettersRepository;
+use MailPoet\Newsletter\Statistics\NewsletterStatisticsRepository;
+use MailPoet\Newsletter\Statistics\WooCommerceRevenue;
+use MailPoet\WooCommerce\Helper;
+
+class OverviewStatisticsController {
```
websupporter, line 16: Naming of this (and structure of the response) is likely to change, as the Emails Tab could use all this information as well, when I structure it a bit differently. But this would happen in the follow up PRs

mailpoet/lib/Newsletter/Statistics/NewsletterStatisticsRepository.php
```diff
@@ -183,8 +196,15 @@ private function getTotalSentCounts(array $newsletters): array {
...
       ->setParameter('newsletters', $newsletters)
-      ->groupBy('n.id')
-      ->getQuery()
+      ->groupBy('n.id');
+
+    if ($from && $to) {
+      $query->andWhere('q.createdAt BETWEEN :from AND :to')
+        ->setParameter('from', $from)
+        ->setParameter('to', $to);
+    }
```
JanJakes, line 211: In this and the other cases, would it make more sense to allow also for filtering just one of these? I suppose it's not needed at the moment, but maybe ignoring `$from` when `$to` is not set could seem like an unexpected behavior from the outside of this class.

mailpoet/assets/css/src/components-automation-analytics/tabs.scss
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  .mailpoet-tab {
+    border: none;
+    font-size: 13px;
+    font-weight: bold;
+    padding: 16px 24px;
+
+    &.mailpoet-tab-active {
+      border-bottom: 5px solid $color-primary;
```
JanJakes: I was wondering what looks strange about the tabs, and it's the centering due to the bottom border:

Screenshot 2023-06-01 at 16 40 00

On the listing page this is done using inset `box-shadow`, particular `inset 0 -4px 0 0 var(--wp-admin-theme-color)`. I guess we can unify that. We could also maybe keep either 4px or use 5px for both cases?

mailpoet/assets/css/src/components-automation-analytics/tabs.scss
```diff
@@ -0,0 +1,31 @@
...
+  border: 1px solid $color-grey;
+
+  .mailpoet-tabs-wrapper {
+    border-bottom: 1px solid $color-grey;
+  }
+
+  .mailpoet-tab {
+    border: none;
+    font-size: 13px;
+    font-weight: bold;
```
JanJakes: This should be `600`, I think `bold` is `700`, and on my retina display the high-boldness is visible.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/api/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  };
+};
+
+export const initializeApi = () => {
+  const apiUrl = `${api.root}/mailpoet/v1/`;
+  apiFetch.use(apiFetch.createRootURLMiddleware(apiUrl));
+  apiFetch.use(apiFetch.createNonceMiddleware(api.nonce));
+};
+
+export async function updateSection(
```
JanJakes: You seem to be using a Redux store. Why not make this method a Redux action?

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/header/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <header className="mailpoet-analytics-header">
+      <Filter />
+      <Button
+        href={addQueryArgs(MailPoet.urls.automationEditor, {
+          id: automation.id,
+        })}
+        isPrimary
+      >
+        {__('Edit Automation', 'mailpoet')}
```
JanJakes: `Edit Automation` -> `Edit automation`

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  );
+}
+
+function App(): JSX.Element {
+  return (
+    <HashRouter>
+      <TopBarWithBeamer />
+      <Notices />
+      <Analytics />
+      <Tabs />
```
JanJakes: It feels somewhat strange that these `<Tabs>` are not within the `Analytics` component. They also seem to be analytics-specific tabs, so maybe the generic `Tabs` naming is also not ideal.

mailpoet/assets/js/src/webpack_admin_expose.js
```diff
@@ -33,7 +33,12 @@ export * as WordPressUrl from '@wordpress/url';
 export * as WordPressI18n from '@wordpress/i18n';
 export * as WordPressIcons from '@wordpress/icons';
 export * as WordPressDataControls from '@wordpress/data-controls';
+export * as WooCommerceDate from '@woocommerce/date';
+import { DateRangeFilterPicker } from '@woocommerce/components/build';
 
+export const WooCommerceComponents = {
+  DateRangeFilterPicker,
+};
```
JanJakes, line 41: This is an interesting idea, but at the same time it wouldn't tree-shake well, if Premium had multiple JS chunks or something. What's more noticeable now is that it looks inconsistent with how we export WP components. That said, I think the current method is not ideal, and sometimes I'm wondering how we could improve it. So I guess I don't know, should we unify it? Or... meh?

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/helpers/boot.ts
```diff
@@ -0,0 +1,12 @@
...
+import { initializeApi, updateSection } from '../api';
+import { Section, storeName } from '../store';
+
+export function boot() {
+  initializeApi();
+  select(storeName)
+    .getSections()
+    .forEach((section: Section) => {
+      void updateSection(section);
+    });
```
JanJakes: I already commented on this within the `updateSection`. I think it should be an action and in the "booting" process we should just call `dispatch`. Then, I guess, it doesn't even need its own file and method and could go directly to `index`.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+export function Tabs(): JSX.Element {
+  const tabs = [
+    <Tab
+      key="automation-flow"
+      route="automation-flow/(.*)?"
+      title={__('Automation flow', 'mailpoet')}
+      automationId="tab-automation-flow"
+    >
+      <AutomationFlow />
+    </Tab>,
```
JanJakes: This seems to create components inside the body of the parent. On each render, you'd get a completely new `Tab` component for each tab, which could get bad when there's a lot of content. If we need an array, for some reason, you can use `useMemo`, for instance.

Is the following not usable? (E.g., do those tabs forbid a `<>` as a child?)

```ts
<Tab ...>
{automationHasEmails() && (
   <>
     <Tab ...>
     <Tab ...>
     <Tab ...>
  </>
)}
```

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/helpers/automation.ts
```diff
@@ -0,0 +1,10 @@
+import { select } from '@wordpress/data';
+import { storeName } from '../store';
+
+export function automationHasEmails(): boolean {
+  const automation = select(storeName).getAutomation();
+  const emailSteps = Object.values(automation.steps).filter(
+    (step) => step.key === 'mailpoet:send-email',
+  );
+  return emailSteps.length > 0;
+}
```
JanJakes: This seems like it could (should?) got to Redux selectors.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/index.tsx
JanJakes: Ahh, an update — I see only now that this is using our old custom tabs. Could we use WP tabs instead? You can check the automation listing page, John implemented even some routing, if I'm not mistaken.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const { overview } = useSelect((s) => ({
+    overview: s(storeName).getSection('overview'),
+  })) as { overview: OverviewSection };
+
+  const percentageFormatter = new Intl.NumberFormat(locale.toString(), {
+    style: 'percent',
+    maximumFractionDigits: 2,
+  });
+  const numberFormatter = new Intl.NumberFormat(locale.toString());
+  const items: JSX.Element[] = [];
```
JanJakes, line 89: You're constructing an array of components in the component body. They would be regenerated on each re-render. You'll need to use `useMemo` or conditions within JSX.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+    return undefined;
+  }
+  const newValue = current > previous ? current - previous : previous - current;
+  return (newValue / previous) * 100;
+}
+
+export function Overview(): JSX.Element | null {
+  const { overview } = useSelect((s) => ({
+    overview: s(storeName).getSection('overview'),
+  })) as { overview: OverviewSection };
```
JanJakes: Why `as { overview: OverviewSection }`, is the auto-typing not working for your store?

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const percentageFormatter = new Intl.NumberFormat(locale.toString(), {
+    style: 'percent',
+    maximumFractionDigits: 2,
+  });
+  const numberFormatter = new Intl.NumberFormat(locale.toString());
+  const items: JSX.Element[] = [];
+  if (overview.data !== undefined) {
+    items.push(
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
JanJakes: We need to add a comment explaining the reason for each ignore.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+      <SummaryNumber
+        key="overview-revenue"
+        label={__('Revenue', 'mailpoet')}
+        delta={getWooCommerceDelta('revenue').toFixed(2) as unknown as number}
+        value={overview.data?.revenue_formatted?.current}
+      />,
+    );
```
JanJakes, line 124: I see kind of strange percentages here. It shows `%f%%` for orders and revenue. It also looks different from Figma, but I guess maybe that was somewhere discussed that it is OK?

Screenshot 2023-06-01 at 17 23 55

mailpoet/lib/AdminPages/Pages/AutomationAnalytics.php
```diff
@@ -0,0 +1,102 @@
...
+      $notice->displayWPNotice();
+      $this->pageRenderer->displayPage('blank.html');
+      return;
+    }
+
+    $this->pageRenderer->displayPage('automation/analytics.html', [
+      'registry' => $this->buildRegistry(),
+      'context' => $this->buildContext(),
+      'automation' => $this->automationMapper->buildAutomation($automation),
+      'sub_menu' => 'mailpoet-automation',
```
JanJakes: When you rebase on `trunk`, please remove `sub_menu`. This is no longer needed since https://github.com/mailpoet/mailpoet/pull/4933/files.

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/OverviewStatisticsController.php
```diff
@@ -0,0 +1,137 @@
...
+    /** @var NewslettersRepository */
+    private $newslettersRepository;
+
+    /** @var NewsletterStatisticsRepository */
+    private $newsletterStatisticsRepository;
+
+    /** @var Helper */
+    private $wooCommerceHelper;
+
+  public function __construct(
```
JanJakes, line 26: Something wrong with the indents again, above and below, in class properties and in function bodies. Does your IDE setting maybe not understand some of our indentation? I know I always say a sniff should watch this, but for some reason it doesn't.

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/OverviewStatisticsController.php
```diff
@@ -0,0 +1,137 @@
...
+      );
+
+    foreach ($previousStatistics as $statistic) {
+      $data['total']['previous'] += $statistic->getTotalSentCount();
+      $data['opened']['previous'] += $statistic->getOpenCount();
+      $data['clicked']['previous'] += $statistic->getClickCount();
+      $data['orders']['previous'] += $statistic->getWooCommerceRevenue() ? $statistic->getWooCommerceRevenue()->getOrdersCount() : 0;
+      $data['revenue']['previous'] += $statistic->getWooCommerceRevenue() ? $statistic->getWooCommerceRevenue()->getValue() : 0;
+    }
+
```
JanJakes, line 94: This is a so-called "zigzag" indentation 😂

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Endpoints/OverviewEndpoint.php
```diff
@@ -0,0 +1,67 @@
...
+  ) {
+    $this->automationStorage = $automationStorage;
+    $this->overviewStatisticsController = $overviewStatisticsController;
+  }
+
+  public function handle(Request $request): Response {
+    $automationid = $request->getParam('id');
+    if (!is_int($automationid)) {
+      throw new RuntimeException('Invalid automation id');
+    }
```
JanJakes: You can just cast this to `int`, the request schema is validated.

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Entities/Query.php
```diff
@@ -0,0 +1,79 @@
...
+    $primaryBefore = $primary['before'] ?? null;
+    $secondaryAfter = $secondary['after'] ?? null;
+    $secondaryBefore = $secondary['before'] ?? null;
+    if (
+      !is_string($primaryAfter) ||
+      !is_string($primaryBefore) ||
+      !is_string($secondaryAfter) ||
+      !is_string($secondaryBefore)
+    ) {
+      throw new \InvalidArgumentException('Invalid query parameters');
```
JanJakes: We should probably throw some of our exceptions. Not sure if we need to check everything with regards to params validation, but no harm.

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/RegisterAnalytics.php
```diff
@@ -0,0 +1,26 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Integrations\MailPoet\Analytics;
+
+use MailPoet\API\REST\API;
+use MailPoet\Automation\Engine\Hooks;
+use MailPoet\Automation\Engine\WordPress;
+use MailPoet\Automation\Integrations\MailPoet\Analytics\Endpoints\OverviewEndpoint;
+
+class RegisterAnalytics {
```
JanJakes: This sounds like a method name. Can we use just `Analytics`?

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/RegisterAnalytics.php
```diff
@@ -0,0 +1,26 @@
...
+
+  public function __construct(
+    WordPress $wordPress
+  ) {
+    $this->wordPress = $wordPress;
+  }
+
+  public function register(): void {
+    $this->wordPress->addAction(Hooks::API_INITIALIZE, function (API $api) {
+      $api->registerPostRoute('automation/analytics/overview', OverviewEndpoint::class);
```
JanJakes: Using `POST` doesn't fit very well with the REST API. We're only fetching data, so `GET` would get the correct semantics would (not just on paper, but also regarding how browsers treat caching, etc.). I don't know if WP suggests keeping the RESTful semantics, but I think in this case we can.

mailpoet/lib/Config/Menu.php
```diff
@@ -502,6 +504,16 @@ private function registerAutomationMenu(bool $showEntries) {
       [$this, 'automationEditor']
     );
 
+    // Automation analytics
+    $this->wp->addSubmenuPage(
+      true,
```
JanJakes: When you rebase on trunk, please use `self::AUTOMATIONS_PAGE_SLUG`. The JS method is no longer used since https://github.com/mailpoet/mailpoet/pull/4933/files.

JanJakes (commented):
Sorry it took me such a long time to get to this 😬

Please, rebase this on `trunk`, I left some comments related to the rebase. One more thing to note is that you'll need to discard the `pnpm-lock.yaml` changes and maybe just run the install commands again (so we also get the latest versions, if there are).

mailpoet/lib/Newsletter/Statistics/NewsletterStatisticsRepository.php
websupporter: 6e999ed7f

mailpoet/assets/css/src/components-automation-analytics/tabs.scss
websupporter: Outdated since we use now `@wordpress/components`.

mailpoet/assets/css/src/components-automation-analytics/tabs.scss
websupporter: Outdated since we use now `@wordpress/components`.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/api/index.ts
websupporter: I am not sure. Not really sure about best practice when it comes to Redux but it felt "bloated" to me. Currently what `updateSection` does:

1. Fetch the current query and automation from the store
2. Transform it to an API query
3. Empty the current section data in the store
4. Run the API query
5. Populate the current section data in the store

So it felt to me like it should be it's own function interacting with the store API (through dispatch etc) instead of being an specialized "action", which to me feels like it should only be concerned with the store itself. Moving this into an action feels to me like

```
class DBLayer {
  update() {
    //fetch data from somewhere
    //transform data
    mysqli_query($this->connection, "UPDATE ...");
  }
}
```

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/header/index.tsx
websupporter: Addressed in 18533dc81

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/index.tsx
websupporter: Thanks! Moved.

mailpoet/assets/js/src/webpack_admin_expose.js
websupporter: Atm for me "meh" tbh.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/helpers/automation.ts
websupporter: :100: fbc7fec8c

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
websupporter: Should be solved now by moving to the `@wordpress/component` tabs

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
websupporter: `getSection()` returns `Section`, but as those sections have different data structures, they get special types, in this case 

```
export type OverviewSection = Section & {
  data: undefined | OverviewSectionData;
};
```

So the Section type is good to access `id` or `name` or `endpoint`. If you want to access the data you would need to specify the specific Section.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
websupporter: Checked why I did it: Turns out, it works now, removed ts-ignore in f7106085a

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
websupporter: Should be solved here now: https://github.com/mailpoet/mailpoet/blob/MAILPOET-5088-analytics-page/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx#L70-L72

mailpoet/lib/AdminPages/Pages/AutomationAnalytics.php
websupporter: Addressed in a183e254f

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/OverviewStatisticsController.php
websupporter: :face_exhaling: done

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/OverviewStatisticsController.php
websupporter: :face_exhaling: done

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Endpoints/OverviewEndpoint.php
websupporter: 62937c83c

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Entities/Query.php
websupporter: 5c670f07d throws now `UnexpectedValueException`

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/RegisterAnalytics.php
websupporter: dd87bb888

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/RegisterAnalytics.php
websupporter: Thanks for catching! Forgot to change that. `GET` does not allow for a request body somehow, so I skipped that and forgot to change it later. 6dc044548 makes the endpoint now `GET`.

mailpoet/lib/Config/Menu.php
websupporter: addressed in a183e254f

mailpoet/assets/css/src/components-automation-analytics/tabs.scss
JanJakes: @websupporter This part is not outdated, right:

> On the listing page this is done using inset box-shadow, particular inset 0 -4px 0 0 var(--wp-admin-theme-color). I guess we can unify that. We could also maybe keep either 4px or use 5px for both cases?

The thing is that the listing tabs have different designs than sidebar tabs in Gutenberg.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/api/index.ts
JanJakes: @websupporter I think calling API and processing the responses is what's done also in other MailPoet and WP data stores. [WP even has utilities for that](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data-controls/), and if you want to extract the logic from the action, then WP has [controls](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#controls) for that (you can see them used in the form editor). Although they can be a bit more writing (apart from an action and a reducer, you need to implement a control as well), they can separate some concerns, if you want to separate them.

I kind of get what you mean, but maybe with actions and controls, "everything belongs to Redux". I'd prefer not to invent an additional custom data layer when `@wordpress/data` is likely meant to handle that, at least to my understanding.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/overview/index.tsx
JanJakes: Right, I see 👍 There are ways to tell TS to understand based on the params, but I guess not needed now.

JanJakes (dismissed):
Thanks!

websupporter:
There was a small bug in displaying the percentages: `Intl.NumberFormat` treats the value `1` as `100%`. [1514a54](https://github.com/mailpoet/mailpoet/pull/4922/commits/1514a54a8ea921de0a6fbfc8e9cdc258fdd5201d) fixes this problem. I also added a ticket [MAILPOET-5405] to write a test to cover the correct display of values @JanJakes

[MAILPOET-5405]: https://mailpoet.atlassian.net/browse/MAILPOET-5405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Aschepikov (approved):
GJ!

PR #4923, opened 2023-05-12
Update PHP version used for integration_tests - MAILPOET-5302

triple0t:
## Description

Update the PHP version used for Nightly integration_tests 

Note: The Nightly test is still falling due to https://mailpoet.atlassian.net/browse/MAILPOET-5304

## Code review notes

_N/A_

## QA notes

We can skip QA for this PR

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5302](https://mailpoet.atlassian.net/browse/MAILPOET-5302)

## After-merge notes

_N/A_

[MAILPOET-5302]: https://mailpoet.atlassian.net/browse/MAILPOET-5302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -626,6 +626,9 @@ jobs:
       mysql_image:
         type: string
         default: ''
+      woo_core_version:
+        type: string
+        default: ''
     steps:
       - attach_workspace:
           at: /home/circleci
@@ -645,6 +648,14 @@ jobs:
           command: |
             cd tests/docker
             docker-compose run --rm -w /project -e COMPOSER_DEV_MODE=1 --entrypoint "php tools/install.php" codeception_integration
+      - when:
+          condition: << parameters.woo_core_version >>
+          steps:
+            - run:
+                name: Download WooCommerce Core
+                command: |
+                  cd tests/docker
+                  docker-compose run --rm -w /project --entrypoint "./do download:woo-commerce-zip << parameters.woo_core_version >>" --no-deps -e WP_GITHUB_USERNAME=${WP_GITHUB_USERNAME} -e WP_GITHUB_TOKEN=${WP_GITHUB_TOKEN} codeception_integration
       - run:
           name: 'PHP Integration tests'
           command: |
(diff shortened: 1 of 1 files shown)


costasovo:
@triple0t I'm afraid we can't just simply switch to PHP 7.4 in this case. This job is configured to use the oldest supported versions on purpose. We want to check if tests are passing for the oldest supported PHP version. The current supported version is PHP 7.2. If we just switch it to a newer version, the job doesn't make sense anymore. I think there are other options for how we could solve the issue:

1. We can use an older Woo version for this job. This is a build for the oldest supported configuration, so having the oldest version supported versions of plugins makes sense. We do the same for acceptance tests.
2. We can switch to PHP 7.3. Woo still supports PHP 7.3, so we don't need to jump from 7.2 to 7.4 here. Also, we[ are about to drop support for PHP7.2 soon](https://mailpoet.atlassian.net/browse/MAILPOET-4939) and we will move to PHP 7.3.
3. We could disable Woo-related tests for the "oldest" job. 

In my opinion, ideally, we could combine 1) and 2). Or do 1) and do 2) when we will be dropping support for PHP 7.2.

costasovo (changes requested):
Please check my comment.

triple0t:

> @triple0t I'm afraid we can't just simply switch to PHP 7.4 in this case. This job is configured to use the oldest supported versions on purpose. We want to check if tests are passing for the oldest supported PHP version. The current supported version is PHP 7.2. If we just switch it to a newer version, the job doesn't make sense anymore. I think there are other options for how we could solve the issue:
> 
> 1. We can use an older Woo version for this job. This is a build for the oldest supported configuration, so having the oldest version supported versions of plugins makes sense. We do the same for acceptance tests.
> 2. We can switch to PHP 7.3. Woo still supports PHP 7.3, so we don't need to jump from 7.2 to 7.4 here. Also, we[ are about to drop support for PHP7.2 soon](https://mailpoet.atlassian.net/browse/MAILPOET-4939) and we will move to PHP 7.3.
> 3. We could disable Woo-related tests for the "oldest" job.
> 
> In my opinion, ideally, we could combine 1) and 2). Or do 1) and do 2) when we will be dropping support for PHP 7.2.


I implemented option 1.

I can't implement option 2 as that would require building a new Docker image for PHP7.3-cli.

As you noted, we will be [dropping support for PHP7.2 soon](https://mailpoet.atlassian.net/browse/MAILPOET-4939), and building the new image should be implemented as part of the ticket. We will need to switch all usage of PHP7.2 to PHP7.3 on CircleCI as well, which has been noted on the ticket.

Thank you

costasovo (approved):
LGTM! Thank you!

Issue #3129, opened 2020-10-25
Cannot translate Woocommerce checkbox opt-in message string in WPML


poehaldur:
In the file /wp-content/plugins/mailpoet/lib/WooCommerce/Subscription.php on line 59 is the following code:
Can this be made into snippet as a functions.php to overwrite the plugin label? 

`  private function getSubscriptionField($inputName, $checked, $labelString) {
    return '<p class="form-row">
      <label class="woocommerce-form__label woocommerce-form__label-for-checkbox checkbox" data-automation-id="woo-commerce-subscription-opt-in">
      <input type="checkbox" class="woocommerce-form__input woocommerce-form__input-checkbox input-checkbox" name="' . $this->wp->escAttr($inputName) . '" id="' . $this->wp->escAttr($inputName) . '" ' . ($checked ? 'checked' : '') . ' />
        <span class="woocommerce-terms-and-conditions-checkbox-text">' . $this->wp->escHtml($labelString) . '</label>
    </p>';
  }`

It seems to be a common problem in WPML ([https://wpml.org/forums/topic/cannot-translate-mailpoet-string/](https://wpml.org/forums/topic/cannot-translate-mailpoet-string/)) 
Really need to have it fixed, as it cannot be used on translated websites in checkout.

NeosinneR:
Hi @poehaldur, we are aware of this issue and are tracking it also in our internal system. I don't have any specific timeframe when it'll be fixed, but we'll let you know once it's done. 

Duplicate of #2795

lysyjan:
Hi @poehaldur,
your issue is fixed in our latest version.

Thanks again for reporting the issue!

Issue #2795, opened 2020-05-15
WooCommerce checkout newsletter field


Erehr:
In mailopet\lib\WooCommerce\Subscription.php on line 58 there's 'getSubscriptionField' function that define newsletter checkbox on checkout field. 

```
  private function getSubscriptionField($inputName, $checked, $labelString) {
    return '<p class="form-row">
      <label class="woocommerce-form__label woocommerce-form__label-for-checkbox checkbox">
      <input type="checkbox" class="woocommerce-form__input woocommerce-form__input-checkbox input-checkbox" name="' . $this->wp->escAttr($inputName) . '" id="' . $this->wp->escAttr($inputName) . '" ' . ($checked ? 'checked' : '') . ' />
        <span class="woocommerce-terms-and-conditions-checkbox-text">' . $this->wp->escHtml($labelString) . '</label>
    </p>';
  }
```

WooCommerce allow to hook into these fields and modify them with will not change this particullar one. Can you replace it with default WooCommerce form field function? Something like this:

```
  private function getSubscriptionField($inputName, $checked, $labelString) {  
    return woocommerce_form_field( $this->wp->escAttr($inputName), array(
        'type'          => 'checkbox',
  	'class' 	=> array('woocommerce-terms-and-conditions'), 
        'label'         => $this->wp->escHtml($labelString),
    ), ($checked ? '1' : '0' ));	  
  }
```

pavel-mailpoet:
Hello Erehr, 

thank you for the report. We will take a look if the function `woocommerce_form_field` can really be used here and if it does we will use it. Thanks again.

lysyjan:
Hi @Erehr ,
your issue is fixed in our latest version.

Thanks again for reporting the issue!

Erehr:
I know this issue is already closed but don't want to open another one as I found another place where the same thing could be changed. Checkbox on registration is also hardcoded and not using woocommerce_form_field instead.

```
 public function extendForm() {
    $label = $this->settings->get(
      'subscribe.on_register.label',
      WPFunctions::get()->__('Yes, please add me to your mailing list.', 'mailpoet')
    );

    print '<p class="registration-form-mailpoet">
      <label for="mailpoet_subscribe_on_register">
        <input
          type="hidden"
          id="mailpoet_subscribe_on_register_active"
          value="1"
          name="mailpoet[subscribe_on_register_active]"
        />
        <input
          type="checkbox"
          id="mailpoet_subscribe_on_register"
[...]

PR #4924, opened 2023-05-12
Fix placement of the opt-in checkbox in WooCommerce checkout when using Twenty Twenty-Three and Twenty Twenty-Two themes [MAILPOET-4133]

rodrigoprimo:
## Description

This PR fix a layout error in the optin checkbox that MailPoet adds to the WooCommerce checkout when the site is using the themes Twenty Twenty-Two and Twenty Twenty-Three.

This problem was happening because WooCommerce add CSS to make its own checkboxes work with those two themes, but it broke checkboxes the layout of checkboxes created with woocommerce_form_field().

The Woo checkboxes in the checkout display the text inside an <span> element and that doesn't happen for checkboxes created with woocommerce_form_field(). The problem is in those CSS rules:

https://github.com/woocommerce/woocommerce/blob/0ea5205672ea785ffa8ef3ad2a8b0048dd2b501e/plugins/woocommerce/client/legacy/css/twenty-twenty-three.scss#L731-L757

I opted to use HTML directly instead of woocommerce_form_field() as this is what WooCommerce itself uses for the checkboxes it adds to the checkout page.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4133]

## After-merge notes

_N/A_

[MAILPOET-4133]: https://mailpoet.atlassian.net/browse/MAILPOET-4133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/WooCommerce/Subscription.php
+++ b/mailpoet/lib/WooCommerce/Subscription.php
@@ -119,18 +119,12 @@ class Subscription {
   }
 
   private function getSubscriptionField($inputName, $checked, $labelString) {
-    return $this->wcHelper->woocommerceFormField(
-      $this->wp->escAttr($inputName),
-      [
-        'type' => 'checkbox',
-        'label' => $this->wp->escHtml($labelString),
-        'input_class' => ['woocommerce-form__input', 'woocommerce-form__input-checkbox', 'input-checkbox'],
-        'label_class' => ['woocommerce-form__label', 'woocommerce-form__label-for-checkbox', 'checkbox'],
-        'custom_attributes' => ['data-automation-id' => 'woo-commerce-subscription-opt-in'],
-        'return' => true,
-      ],
-      $checked ? '1' : '0'
-    );
+    $checked = checked($checked, true, false);
+
+    return '<label class="woocommerce-form__label woocommerce-form__label-for-checkbox checkbox" data-automation-id="woo-commerce-subscription-opt-in">
+      <input id="mailpoet_woocommerce_checkout_optin" class="woocommerce-form__input woocommerce-form__input-checkbox input-checkbox" ' . $checked . ' type="checkbox" name="' . $this->wp->escAttr($inputName) . '" value="1" />
+      <span>' . $this->wp->escHtml($labelString) . '</span>
+    </label>';
   }
 
   private function getSubscriptionPresenceCheckField() {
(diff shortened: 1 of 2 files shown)


brezocordero (commented):
@rodrigoprimo Your changes work to fix the alignment but I found that using `woocommerce_form_field_...` had been requested by customers to be able to customize the checkbox.

https://github.com/mailpoet/mailpoet/issues/2795
https://github.com/mailpoet/mailpoet/issues/3129

If we revert this back to fix the alignment with these two themes, we will break custom layouts.

This is not a new issue, because it is also broken for 2022 theme. Do you know if we have had customer reports about it (I can't tell from the JIRA issue, as this was reported by us.)?

If we don't have customer reports about it, what do you think of submitting an issue to WooCommerce repo instead?

rodrigoprimo:
Good find, @brezocordero!

> This is not a new issue, because it is also broken for 2022 theme. Do you know if we have had customer reports about it (I can't tell from the JIRA issue, as this was reported by us.)?

I believe @veljkho found this issue and reported it, but I could be wrong.

> If we don't have customer reports about it, what do you think of submitting an issue to WooCommerce repo instead?

I think it is worth trying and seeing what they reply.

I created an issue here: https://github.com/woocommerce/woocommerce/issues/38297

rodrigoprimo:
@brezocordero, I got a reply from the Woo team (https://github.com/woocommerce/woocommerce/issues/38297#issuecomment-1557037493). They acknowledged the problem, but fixing it is not a priority for them. They suggested we create a PR if we want to see this fixed sooner, but I'm not sure if we should dedicate more time to this issue. cc @NeosinneR for guidance on the next steps. Should we proceed with the solution in this PR or work on a PR to fix the root problem on Woo core?

It is not clear to me at this point how complex it would be to fix this problem on Woo core, so I'm more inclined to suggest that we merge this PR as is.

NeosinneR:
@rodrigoprimo, with uncertainty about how long it would take to fix the core, I agree with merging this PR and not pursuing the core fix, as the core itself is not using it to add checkboxes.

brezocordero (approved):
LGTM.

Aschepikov (approved):
GJ!

PR #4933, opened 2023-05-16
Fix opening and highlighting of MailPoet menu [MAILPOET-5296] [MAILPOET-5154]

JanJakes:
## Description

Fixes opening and highlighting of MailPoet menu, adds test coverage.

## Linked tickets

[MAILPOET-5296]
[MAILPOET-5154]

[MAILPOET-5296]: https://mailpoet.atlassian.net/browse/MAILPOET-5296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-5154]: https://mailpoet.atlassian.net/browse/MAILPOET-5154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


johnolek (changes requested):
Thanks very much for working on this! I've definitely found this annoying in the past.

These changes all look good to me, but I did a little poking around and noticed a few things I wanted to run by you.

### Sub pages don't highlight top level MailPoet menu
When you're on a primary MailPoet page like Automations, the top level MailPoet link has styling to show it's active:

image

Once you go into a sub page like automation template selection, we lose this styling:
image

I don't know if this is intentional, but it is inconsistent with other WordPress menus.

I think the issue is that we're replacing `wp-not-current-submenu` with `wp-has-current-submenu` for the top level `li` but not for the top level `a`. This JS produced what I would expect, though there's probably a better way to do it if we want to make this change:

```js
<script type="text/javascript">
jQuery('#adminmenu #toplevel_page_mailpoet-homepage')
  .addClass('wp-has-current-submenu')
  .removeClass('wp-not-current-submenu')
  .find('a[href$="mailpoet-homepage"]')
  .first()
  .addClass('wp-has-current-submenu')
  .removeClass('wp-not-current-submenu');
jQuery('#adminmenu #toplevel_page_mailpoet-homepage')
  .find('a[href$="<%= sub_menu %>"]')
  .addClass('current')
  .parent()
  .addClass('current');
</script>
```

### Welcome wizard does not highlight top level menu item
I think it would be really nice if we showed the MailPoet menu as being active when people are first interacting with the plugin.
image

### Automation email template selector and editor
This one I don't care about as the previous two, and might not even be a problem.

Now that we display submenus during template selection and newsletter editing, we're highlighting the Emails submenu even if we're working on an automation email. I think it would be nicer to have the active submenu be Automations to really drive home the idea that you're still in the automation context.

### Menu constants vs strings
This one I seriously don't care about at all, but I noticed from the [newsletter editor's sub_menu definition](https://github.com/mailpoet/mailpoet/blob/a783739b09a60efc1e69718cd06df05cec483107/mailpoet/lib/AdminPages/Pages/NewsletterEditor.php#L155-L155) that we have all these page slugs available as constants on the `Menu` class. Any reason to use strings instead of the constants? Again I really don't think this matters. Not like we're going to change slugs any time soon.

JanJakes:
@johnolek Thanks for your very detailed review! It made me challenge myself to find a better solution, see the new commits. The current implementation works without JS, fixes the main menu active styles, highlights the main menu also for wizard, landing page, error page, etc., and I also fixed the icon flicker on page reload (unified with Woo).

The only thing I didn't fix is this:

> I think it would be nicer to have the active submenu be Automations to really drive home the idea that you're still in the automation context.

Do you have any ideas how to do it? I think for the newsletter editor it's doable if I fetch the email and check its type (then I could change the page with a hook), but what about templates? It's just a React-routed page on top of newsletters 😬

JanJakes:
@johnolek Actually, I think it's good enough to use a URL param for that, so I added that in https://github.com/mailpoet/mailpoet/pull/4933/commits/8b8e76b2a5be2219f91de9f4969635d86c71662b.

johnolek (changes requested):
@JanJakes Great job on these changes! The PHP approach is much better. I love how there's no longer the delay in applying the styles.

There's one last case that I think you missed, which is re-entering the newsletter editor using the "Edit content" button in the automation editor. It's not a big deal, but with everything else being perfect I would hate to let that slip through 🙂

I somewhat prefer the idea of fetching the newsletter's type instead of using a URL param for the editor since it seems easier to maintain going forward, but I have no idea what kind of effort it would require, and simply adding the URL param to the edit button's URL is totally fine by me.

JanJakes:
@johnolek Ahh, thanks for the catch! Fixed.

> I somewhat prefer the idea of fetching the newsletter's type instead of using a URL param

Me too, but it's rather hard:
1. On the newsletters page we know the newsletter ID, so we could fetch the newsletter, but it seems to be too late to change the plugin page.
2. On the newsletter templates page we don't know the ID. It's actually just the newsletter listing page with React router displaying a different view (and fetching data via API).

Maybe we can look into it again with some UI changes or the new email editor. For instance... should we render the email editor in a modal overlay or iframe rather than changing the page? We'll see...

johnolek (dismissed):
@JanJakes Thanks for the fix and the explanation! I didn't realize it was too late in the process to use the PHP method for the editor. The fix you pushed is very simple though, and we can always revisit this as you said when working on the new editor.

This all looks good to me now! Once again, really nice work on this!

Aschepikov:
Hi @JanJakes there are some conflicts can you resolve them pls

JanJakes:
@Aschepikov Thanks, rebased.

Aschepikov (approved):
GJ!

PR #4942, opened 2023-05-18
Refactor MetaInfo to use Doctrine instead of Paris [MAILPOET-5344]

rodrigoprimo:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

QA is probably not needed.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5344]

## After-merge notes

_N/A_

[MAILPOET-5344]: https://mailpoet.atlassian.net/browse/MAILPOET-5344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -336,7 +336,7 @@ class SendingQueue {
       $unsubscribeUrls[] = $this->links->getUnsubscribeUrl($queue->id, $subscriberEntity);
       $oneClickUnsubscribeUrls[] = $this->links->getOneClickUnsubscribeUrl($queue->id, $subscriberEntity);
 
-      $metasForSubscriber = $this->mailerMetaInfo->getNewsletterMetaInfo($newsletter, $subscriberEntity);
+      $metasForSubscriber = $this->mailerMetaInfo->getNewsletterMetaInfo($newsletterEntity, $subscriberEntity);
       if ($campaignId) {
         $metasForSubscriber['campaign_id'] = $campaignId;
       }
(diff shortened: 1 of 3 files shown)


johnolek (dismissed):
Very straightforward changes, looks good to me!

Is there a reason we can't/shouldn't update the signature for the `getNewsletterMetaInfo` method to accept a `NewsletterEntity`?

I realize right now it's expecting an object version of a paris model newsletter (from [here](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Cron/Workers/SendingQueueSendingQueue.php#L177-L183) I believe?). Maybe we're changing the cloned version in a way that wouldn't be represented in the entity before it arrives at this method?

In general I can understand saving this for after we've refactored the sending queue code, which is why I'm happy to approve the PR as is, though I think this method is only used in [one place](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L339-L339), and it looks like we should definitely have a newsletter entity we could pass in at that point so I just wanted to check.

rodrigoprimo:
Thanks for the review!

> Is there a reason we can't/shouldn't update the signature for the getNewsletterMetaInfo method to accept a NewsletterEntity?

I was looking for classes to refactor searching for occurrences of `MailPoet\Models\`. When I found this class, I focused only on occurrences of `Newsletter::` until the import of `MailPoet\Models\Newsletter` was marked as unused. I was moving too fast and it didn't even occur to me that there could be more "hidden" uses of the model.

I pushed a new commit changing the signature of `getNewsletterMetaInfo()`.

johnolek (approved):
Thanks for the second commit, @rodrigoprimo, this looks good to me!

PR #4943, opened 2023-05-18
Remove Subscribers\Source::setSource() [MAILPOET-5345]

rodrigoprimo:
## Description

I found this method in the context of a ticket to refactor Paris code to Doctrine. As far as I can tell, it is not used anymore and it is
safe to delete it.

It was added in https://github.com/mailpoet/mailpoet/commit/1af5802030a809ca94d8235fc8cb40a764c5f955 and was used only inside Models\Subscriber::subscribe(). Then subscribe() was moved to Subscriber\SubscriberActions in https://github.com/mailpoet/mailpoet/commit/7528f0f52be28dddc2e19f96bae1ece458000fa9. Finally, commit https://github.com/mailpoet/mailpoet/commit/7db23849f7dbc78e886c854663870d0ea55cfdcc refactored SubscriberActions removing the only two calls to Source::setSource().

The test class SourceTest is removed as well as its only purpose was to test Source::setSource().

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5345]

## After-merge notes

_N/A_

[MAILPOET-5345]: https://mailpoet.atlassian.net/browse/MAILPOET-5345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


websupporter (commented):
Overall: LGTM.

As far as I understand the source is now either given `raw` to "createOrUpdate" or `SubscriberEntity::setSource()` is used.

While browsing the code and context I found that `SubscriberEntity::setSource()` validates against a set of allowed strings:
https://github.com/mailpoet/mailpoet/blob/a976006a2c145a50121938f5b577cdc256d351c8/mailpoet/lib/Entities/SubscriberEntity.php#L385-L401

Should we maybe make a `public const ALLOWED_SOURCES` in the `Source` class and use this array in the entity in order to unify the "Source"-source?

rodrigoprimo:
@websupporter, I think this is a good idea, but IMO it is outside the scope of this ticket. Maybe worth creating a separate maintenance ticket for us to address this improvement in the future? Let me know if you think it is better to address it in this PR for some reason that I'm missing.

websupporter (approved):
lgtm

Aschepikov (approved):
GJ!

PR #4949, opened 2023-05-23
Updated 3 tests and added new methods to tester

veljkho:
## Description

Updated 3 tests with similar scenarios, added new methods to `AcceptanceTester.php`.

These are the 3 tests improved:
`SendFirstPurchaseEmailCest`
`SendProductPurchaseEmailCest`
`SendCategoryPurchaseEmailCest`

These are the added methods to the tester:
`getBackToSite()`
`emptyMailbox()`
`checkEmailWasNotReceived($subject)`

I think these methods are worth having even for the other or new tests.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5312] & [MAILPOET-5313]

## After-merge notes

_N/A_

[MAILPOET-5312]: https://mailpoet.atlassian.net/browse/MAILPOET-5312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5313]: https://mailpoet.atlassian.net/browse/MAILPOET-5313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/_support/AcceptanceTester.php
+++ b/mailpoet/tests/_support/AcceptanceTester.php
@@ -58,14 +58,6 @@ class AcceptanceTester extends \Codeception\Actor {
   /**
    * Define custom actions here
    */
-  public function getBackToSite() {
-    $i = $this;
-    $i->amOnUrl(self::WP_URL);
-  }
-
-  /**
-   * Define custom actions here
-   */
   public function login() {
     $i = $this;
     $i->amOnPage('/wp-login.php');
@@ -124,6 +116,14 @@ class AcceptanceTester extends \Codeception\Actor {
     $i->amOnUrl(self::MAIL_URL);
     // ensure that angular is loaded by checking angular specific class
     $i->waitForElement('.messages.ng-scope');
+  }
+
+  /**
+   * Define custom actions here
+   */
+  public function amOnSiteHomepage() {
+    $i = $this;
+    $i->amOnUrl(self::WP_URL);
   }
 
   /**
--- a/mailpoet/tests/_support/AcceptanceTester.php
+++ b/mailpoet/tests/_support/AcceptanceTester.php
@@ -58,6 +58,14 @@ class AcceptanceTester extends \Codeception\Actor {
   /**
    * Define custom actions here
    */
+  public function getBackToSite() {
+    $i = $this;
+    $i->amOnUrl(self::WP_URL);
+  }
+
+  /**
+   * Define custom actions here
+   */
   public function login() {
     $i = $this;
     $i->amOnPage('/wp-login.php');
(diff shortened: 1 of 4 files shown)


mailpoet/tests/_support/AcceptanceTester.php
```diff
@@ -55,6 +55,14 @@ class AcceptanceTester extends \Codeception\Actor {
   const WOO_COMMERCE_SUBSCRIPTIONS_PLUGIN = 'woocommerce-subscriptions';
   const AUTOMATE_WOO_PLUGIN = 'automatewoo';
 
+  /**
+   * Define custom actions here
+   */
+  public function getBackToSite() {
```
JanJakes: The naming seems to be a bit confusing to me — when I was not on the site homepage previously, then it doesn't make much sense. I'd suggest we name it `amOnSiteHomepage`, for instance.

mailpoet/tests/_support/AcceptanceTester.php
```diff
@@ -118,6 +126,18 @@ public function amOnMailboxAppPage() {
...
+  /**
+   * Navigate to Mailhog page and empty the mailbox
+   */
+  public function emptyMailbox() {
+    $i = $this;
+    $i->amOnMailboxAppPage();
+    // click to empty the mailbox
+    $i->click('.glyphicon-remove-circle');
+    $i->waitForElementVisible('.btn-danger');
+    $i->click('.btn-danger');
```
JanJakes: If you want to perform this faster, then just call this command that we execute before every test:

https://github.com/mailpoet/mailpoet/blob/a8b04a8dcaca4258974d25a0c9770f6f07492be4/mailpoet/tests/_support/CleanupExtension.php#L78

mailpoet/tests/_support/AcceptanceTester.php
```diff
@@ -650,6 +670,19 @@ public function checkEmailWasReceived($subject) {
...
+   * @param string $subject
+   */
+  public function checkEmailWasNotReceived($subject) {
+    $i = $this;
+    $i->amOnMailboxAppPage();
+    $i->dontSee($subject);
+    // click refresh button to seek for new emails once again
+    $i->click('.glyphicon-refresh');
+    $i->dontSee($subject);
+  }
```
JanJakes, line 681: This won't be a very reliable check unless we make sure the cron was running, or add some kind of wait (easy, but slow). Email sending and delivery takes some time, and in most cases this will pass because the emails was not **yet** received. I think force-triggering Action Scheduler could work here. You could do something like this:

```php
    $i = $this;
    $i->triggerMailPoetActionScheduler();
    $i->amOnMailboxAppPage();
    $i->dontSee($subject);
```

Maybe in this case, the refresh isn't even needed? I let you choose...

JanJakes (commented):
Nice improvements! I added a some comments with a few suggestions.

mailpoet/tests/_support/AcceptanceTester.php
veljkho: This makes sense, updating!

mailpoet/tests/_support/AcceptanceTester.php
veljkho: I think this works! Great suggestion, thanks!

mailpoet/tests/_support/AcceptanceTester.php
veljkho: This makes sense! Updating.

JanJakes (approved):
Thanks! 👏

PR #4952, opened 2023-05-25
Improve the test to include checks for the blank form

veljkho:
## Description

I updated the test `EditorCreateBlankFormCest` to include checks of the blank form. Now, we also check for the 2 vital blocks inside but also form placement and default toggles. We want to ensure the blank form is properly created.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5316]

## After-merge notes

_N/A_

[MAILPOET-5316]: https://mailpoet.atlassian.net/browse/MAILPOET-5316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Forms/EditorCreateBlankFormCest.php
+++ b/mailpoet/tests/acceptance/Forms/EditorCreateBlankFormCest.php
@@ -6,13 +6,16 @@ use MailPoet\Test\DataFactories\Segment;
 
 class EditorCreateBlankFormCest {
   public function createBlankForm(\AcceptanceTester $i) {
-    $i->wantTo('Create a blamk form');
+    $i->wantTo('Create a blank form');
+
     $segmentFactory = new Segment();
     $segmentName = 'Fancy List';
     $segmentFactory->withName($segmentName)->create();
 
     $i->login();
+
     $i->amOnMailPoetPage('Forms');
+
     // Create a new form
     $formName = 'My awesome form';
     $i->click('[data-automation-id="create_new_form"]');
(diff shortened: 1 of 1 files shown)


johnolek (approved):
This looks good to me! 

Out of curiosity, is this addressing a specific issue we've seen with the editor not loading, or not being created with the correct settings?

Some of this seems like overkill. For example, if we updated our templates in the future to not include an `editor_spacer_block`, would this test failing be meaningful to us or just something annoying we have to update?

veljkho:
> This looks good to me!
> 
> Out of curiosity, is this addressing a specific issue we've seen with the editor not loading, or not being created with the correct settings?
> 
> Some of this seems like overkill. For example, if we updated our templates in the future to not include an `editor_spacer_block`, would this test failing be meaningful to us or just something annoying we have to update?

Hi @johnolek thanks for a review!

I don't recall if we ever had an issue. One of the points is to have checks so we don't need to manually test it.
We don't need to update this test if we exclude `editor_spacer_block` since the test says `dontSee` for such block.

Maintenance of the test suite is normal as we further develop our product. Anything related to a blank form would require a developer to take a look for possible updates of the test.

johnolek:
@veljkho Sorry, I totally missed that those were `dontSee` assertions! I think the point is still mostly the same, though. If we added a spacer block to the template then this test would start failing, right? How meaningful would that failure be to us? Certainly not as meaningful as it would be if the email input or submit button were suddenly missing.

You're right that it's normal for the test suite to require changes as we update related code. I just think it's interesting to think about if potential future test failures would be providing us with an important signal about something not working correctly.

In any case, this looks good to me as is! It's not like we're going to be changing these templates anytime soon in a way that would cause the test to break. I just wanted to bring this up as a general idea to get your thoughts on what kinds of assertions are most important.

PR #4960, opened 2023-05-30
Fix: subscriber segmentation can now target contacts who did not open any emails [MAILPOET-5347]

rodrigoprimo:
## Description

This PR fixes a bug in the query that is used to get the subscribers that opened emails that was preventing it from working when trying to get subscribers that opened zero emails.

The problem was that we were adding the user_agent_type	as a condition on the WHERE part of the clause. Meaning that it was being used to filter the overall results and thus excluding all subscribers that had no entries in the wp_mailpoet_statistics_opens table. By moving this condition to the ON clause in the LEFT JOIN, we use the user_agent_type field to match rows between tables. The value of this field will still be taken into consideration when counting the number of opens as we are counting the number of entries in the wp_mailpoet_statistics_opens table.

## Code review notes

_N/A_

## QA notes

In a MailPoet install with a few emails sent and a few emails opened by different subscribers, create a "number of opens" segment and make sure that the number of matched subscribers is equal to what is expected. Especially when setting the segment criteria to match zero opens or less than one open.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5347]

## After-merge notes

_N/A_

[MAILPOET-5347]: https://mailpoet.atlassian.net/browse/MAILPOET-5347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/EmailOpensAbsoluteCountAction.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/EmailOpensAbsoluteCountAction.php
@@ -36,7 +36,7 @@ class EmailOpensAbsoluteCountAction implements Filter {
       $subscribersTable,
       $statsTable,
       'opens',
-      "$subscribersTable.id = opens.subscriber_id AND opens.created_at > :newer" . $parameterSuffix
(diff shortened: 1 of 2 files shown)


johnolek (approved):
Nice work, looks good to me! And thank you for the detailed description of the issue and fix.

Aschepikov (approved):
GJ!

PR #4967, opened 2023-06-02
Add new "used shipping method" segment [MAILPOET-4992]

rodrigoprimo:
## Description

This PR adds a new WooCommerce segment called "used shipping method".

## Code review notes

I based this new segment on the "used payment method" segment.

I noticed that when the operator "none of" is used for the "used payment method" it matches not only WooCommerce customers but all types of subscribers. I followed the same logic for this new segment but I wonder if we should limit the scope of both segments only to WooCommerce customers? (cc @NeosinneR)

## QA notes

How to test this PR:

- Place a few WooCommerce orders using different shipping methods.
- Go to the page to create a new MailPoet segment (/wp-admin/admin.php?page=mailpoet-segments#/new-segment) and try different values for the parameters of the new "used shipping method" segment.
- Verify that the number of matched subscribers matches what is expected.
- Send an email to this segment and verify that the recipients match what is expected.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4992]

## After-merge notes

_N/A_

[MAILPOET-4992]: https://mailpoet.atlassian.net/browse/MAILPOET-4992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedShippingMethod.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedShippingMethod.php
@@ -1,0 +1,122 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Segments\DynamicSegments\Filters;
+
+use MailPoet\Entities\DynamicSegmentFilterData;
+use MailPoet\Entities\DynamicSegmentFilterEntity;
+use MailPoet\Segments\DynamicSegments\Exceptions\InvalidFilterException;
+use MailPoet\WooCommerce\Helper;
+use MailPoetVendor\Carbon\Carbon;
+use MailPoetVendor\Doctrine\DBAL\Connection;
+use MailPoetVendor\Doctrine\DBAL\Query\QueryBuilder;
+
+class WooCommerceUsedShippingMethod implements Filter {
+  const ACTION = 'usedShippingMethod';
+
+  const VALID_OPERATORS = [
+    DynamicSegmentFilterData::OPERATOR_NONE,
+    DynamicSegmentFilterData::OPERATOR_ANY,
+    DynamicSegmentFilterData::OPERATOR_ALL,
+  ];
+
+  /** @var WooFilterHelper */
+  private $wooFilterHelper;
+
+  /** @var Helper */
+  private $wooHelper;
+
+  /** @var FilterHelper */
+  private $filterHelper;
+
+  public function __construct(
+    FilterHelper $filterHelper,
+    WooFilterHelper $wooFilterHelper,
+    Helper $wooHelper
+  ) {
+    $this->wooFilterHelper = $wooFilterHelper;
+    $this->wooHelper = $wooHelper;
+    $this->filterHelper = $filterHelper;
+  }
+
+  public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $filter): QueryBuilder {
+    $filterData = $filter->getFilterData();
+    $operator = $filterData->getParam('operator');
+    $shippingMethods = $filterData->getParam('shipping_methods');
+    $days = $filterData->getParam('used_shipping_method_days');
+
+    if (!is_string($operator) || !in_array($operator, self::VALID_OPERATORS, true)) {
+      throw new InvalidFilterException('Invalid operator', InvalidFilterException::MISSING_OPERATOR);
+    }
+
+    if (!is_array($shippingMethods) || count($shippingMethods) < 1) {
+      throw new InvalidFilterException('Missing payment methods', InvalidFilterException::MISSING_VALUE);
+    }
+
+    if (!is_int($days) || $days < 1) {
+      throw new InvalidFilterException('Missing days', InvalidFilterException::MISSING_VALUE);
+    }
+
+    $includedStatuses = array_keys($this->wooHelper->getOrderStatuses());
+    $failedKey = array_search('wc-failed', $includedStatuses, true);
+    if ($failedKey !== false) {
+      unset($includedStatuses[$failedKey]);
+    }
+    $date = Carbon::now()->subDays($days);
+
+    switch ($operator) {
+      case DynamicSegmentFilterData::OPERATOR_ANY:
+        $this->applyForAnyOperator($queryBuilder, $includedStatuses, $shippingMethods, $date);
+        break;
+      case DynamicSegmentFilterData::OPERATOR_ALL:
+        $this->applyForAllOperator($queryBuilder, $includedStatuses, $shippingMethods, $date);
+        break;
+      case DynamicSegmentFilterData::OPERATOR_NONE:
+        $this->applyForNoneOperator($queryBuilder, $includedStatuses, $shippingMethods, $date);
+        break;
+    }
+
+    return $queryBuilder;
+  }
+
+  private function applyForAnyOperator(QueryBuilder $queryBuilder, array $includedStatuses, array $shippingMethods, Carbon $date): void {
+    $dateParam = $this->filterHelper->getUniqueParameterName('date');
+    $shippingMethodParam = $this->filterHelper->getUniqueParameterName('shippingMethod');
+
+    $orderItemsTable = $this->filterHelper->getPrefixedTable('woocommerce_order_items');
+    $orderItemsTableAlias = 'orderItems';
+    $orderStatsAlias = $this->wooFilterHelper->applyOrderStatusFilter($queryBuilder, $includedStatuses);
+    $queryBuilder
+      ->innerJoin($orderStatsAlias, $orderItemsTable, $orderItemsTableAlias, "$orderStatsAlias.order_id = $orderItemsTableAlias.order_id")
+      ->andWhere("$orderStatsAlias.date_created >= :$dateParam")
+      ->andWhere("$orderItemsTableAlias.order_item_name IN (:$shippingMethodParam)")
+      ->andWhere("$orderItemsTableAlias.order_item_type = 'shipping'")
+      ->setParameter($dateParam, $date->toDateTimeString())
+      ->setParameter($shippingMethodParam, $shippingMethods, Connection::PARAM_STR_ARRAY);
+  }
+
+  private function applyForAllOperator(QueryBuilder $queryBuilder, array $includedStatuses, array $shippingMethods, Carbon $date): void {
+    $dateParam = $this->filterHelper->getUniqueParameterName('date');
+    $orderItemTypeParam = $this->filterHelper->getUniqueParameterName('orderItemType');
+    $shippingMethodsParam = $this->filterHelper->getUniqueParameterName('shippingMethod');
+
+    $orderItemsTable = $this->filterHelper->getPrefixedTable('woocommerce_order_items');
+    $orderItemsAlias = 'orderItems';
+    $orderStatsAlias = $this->wooFilterHelper->applyOrderStatusFilter($queryBuilder, $includedStatuses);
+    $queryBuilder
+      ->innerJoin($orderStatsAlias, $orderItemsTable, $orderItemsAlias, "$orderStatsAlias.order_id = $orderItemsAlias.order_id")
+      ->andWhere("$orderStatsAlias.date_created >= :$dateParam")
+      ->andWhere("$orderItemsAlias.order_item_type = :$orderItemTypeParam")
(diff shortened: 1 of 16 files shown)


johnolek (changes requested):
@rodrigoprimo Nice work so far! I'm requesting changes because I'm not sure this gets to the level of specificity we need or includes all the actual shipping methods we want.

I'm going to include lots of notes here, but I think this will ultimately be a lot easier to discuss more synchronously so don't feel like you have to respond to any/all of this unless you really want to. 

Also I don't intend this like "here's what you have to do," rather it's, "here's the best I can come up with." I would welcome with open arms a simpler suggestion than what I describe below.

### Shipping method options
I don't think `WC()->shipping()->get_shipping_methods();` is doing everything we want. In my testing it doesn't return the actual shipping methods that a user has saved to a shipping zone, rather the general type/template/blueprint. If I recall correctly it's based on the registered PHP shipping classes. I think the current code might only work if a user doesn't rename the shipping method from the default.

For example, if I create a shipping method in the US zone based on the Flat Rate shipping method and then edit its name to be `Flat Rate US`, I don't see that in what `WC()->shipping()->get_shipping_methods();` returns, just the base `Flat Rate` template.

I think we'll need to use multiple approaches to construct our list of shipping methods if we want to include everything in an unambiguous way. First would be iterating over the shipping zones and getting each zone's shipping methods. In my testing these corresponded to the actual methods a user has saved to their database. [This code](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/src/Shipping/ShippingController.php#L314-L316) in WooCommerce is I think what runs to populate the list of zones shown in the admin.

Since method names aren't required to be unique, we'll need to include extra data about each method. I think the `instance_id` is probably the most unique property. Having the zone name would also be nice so we can differentiate in the UI, e.g. `Flat Rate (Minnesota)` vs `Flat Rate (USA)`.

We'll also need to include shipping methods that don't support zones, and I think we can retrieve those from `WC()->shipping()->get_shipping_methods();`, but I haven't done much testing with this.

### The Query
Relying on an `order_item_name` that matches the shipping method's title has a couple of major potential problems:
1. An admin can change the title of a shipping method, which will change this value for subsequent orders but won't update past orders (not that I would expect it to)
2. Shipping method titles are not unique (not even within the same shipping zone apparently), though I'm more concerned about distinguishing between two methods in different zones with the same  name. If a user creates two shipping methods with the same name in the same zone that's up to them to figure out.

If we want to let users properly target these ambiguous or renamed shipping methods, I think we'll need to expand the query, which may require more data for each method. The best idea I can come up with is to join the `wp_woocommerce_order_itemmeta` table. I think we would then be looking for a matching`method_id` as well as an `instance_id` (which I'm pretty sure corresponds to the `instance_id` in the `wp_woocommerce_shipping_zone_methods` table) if the shipping method supports zones.

Unfortunately we might need some extra logic to handle shipping methods that don't support/belong to any zones, which could make the `all` option interesting. Based on some past research I believe there are some very popular shipping plugins/extensions that don't support zones. Maybe for these we actually could rely on just the `order_item_name` since the shipping method title would presumably be hardcoded in the extension. Using the `method_id` might still be the safest in case the title ends up translated or something though.

### QA
If it's not too onerous, I think it might be good to try installing some of the most popular shipping plugins and ensuring that this filter is compatible.

NeosinneR:
> I followed the same logic for this new segment but I wonder if we should limit the scope of both segments only to WooCommerce customers?

We had this discussion in the past, but I can't find it 😑 But IIRC, the reason why we have it like this, is that we want opposing conditions (`any of` and `none of`) to always return 100% of subscribers. And if you want just WC customers, you have multiple options to do that (e.g. `# of orders > 0`).

rodrigoprimo:
> I don't think WC()->shipping()->get_shipping_methods(); is doing everything we want. In my testing it doesn't return the actual shipping methods that a user has saved to a shipping zone, rather the general type/template/blueprint.

Good catch, @johnolek. I missed that WooCommerce didn't store one of the values returned from `WC()->shipping()->get_shipping_methods()` in `wp_woocommerce_order_items`. I will check again, having your suggestions in mind.

> We had this discussion in the past, but I can't find it expressionless But IIRC, the reason why we have it like this, is that we want opposing conditions (any of and none of) to always return 100% of subscribers. And if you want just WC customers, you have multiple options to do that (e.g. # of orders > 0).

Thanks for confirming that this is the intended behavior, @NeosinneR.

rodrigoprimo:
Thanks again for your review, @johnolek!

I will share here some early results of what I did after your comments to get feedback before finishing this PR. I pushed a commit that might still need some adjustments to change the code that gets the list of shipping methods. I still need to update the query that we use to get the subscribers that match this segment and I might need to make some further adjustments to the commit that I just pushed.

> The best idea I can come up with is to join the wp_woocommerce_order_itemmeta table. I think we would then be looking for a matching method_id as well as an instance_id (which I'm pretty sure corresponds to the instance_id in the wp_woocommerce_shipping_zone_methods table) if the shipping method supports zones.

I tried to see if there were alternatives but I did not find anything, so I ended up following your suggestion. Now the code uses two different methods to get the shipping methods. One is to get shipping methods associated with all the zones, and another one is to get the shipping methods associated with the special zone that matches everything that is out of the specified zones. I followed [this](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/src/Shipping/ShippingController.php#L314-L316) code snippet that you shared.

The option value for each shipping method consists of a combination of the instance id and the method id.

> Unfortunately we might need some extra logic to handle shipping methods that don't support/belong to any zones, which could make the all option interesting.

Yesterday I was pairing with Brezo and she did some research and was not able to find an example of a shipping method that does not support zones. Checking the code, it seems that WooCommerce indeed supports those, but without a real example, I'm not sure how to properly account for this case. I'm unsure if I should investigate this more or if we should proceed without supporting those shipping methods. As a next step, I suggest we ask Woo folks about this to confirm if we need to do something about it or not. What do you think?

johnolek:
@rodrigoprimo Thanks for the update! The changes in your latest commit look good to me.

As for shipping methods that don't support zones, I also haven't been able to find any examples myself. I'm beginning to suspect that this is an edge case that isn't worth supporting, but I do think it would be worthwhile to reach out to some Woo people to confirm. The most compelling argument I could see for supporting these methods would be if there's an official Woo shipping extension that doesn't support zones and doesn't plan on supporting them in the future.

rodrigoprimo:
@johnolek, I just pushed the remaining commits. The resulting query is quite complex and is unlikely to perform well for larger stores, but it is the best that I could come up with.

Could you please take a second look at this PR?

rodrigoprimo:
Now that #4973 has been merged, I rebased this branch and force pushed the changes here.

mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedShippingMethod.php
```diff
@@ -65,58 +66,106 @@ public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $fi
...
+    foreach ($shippingMethodStrings as $shippingMethodString) {
+      if (preg_match('/^\w+:\d+$/', $shippingMethodString)) {
+        $parts = preg_split('/:/', $shippingMethodString);
+
+        if (is_array($parts) && is_string($parts[0]) && is_string($parts[1])) {
+          $shippingMethodIds[] = $parts[0];
+          $instanceIds[] = $parts[1];
+        }
+      }
+    }
```
johnolek: This question might be moot based on my general comment, but is there a reason we need to use regular expressions here instead of a simple string split? I think `\w` matches `[a-zA-Z0-9_]`, so if there was a shipping method with an id of `some-method` it wouldn't match and wouldn't get extracted.

I looked at `WC_Shipping_Method` and can't see anywhere that the `$id` is getting validated, so I guess it could include any characters.

My main concern with this is that we could potentially be allowing the user to select shipping methods in the UI that wouldn't make it through this extraction, producing unexpected results.

johnolek (commented):
@rodrigoprimo Great work on these updates!

This looks great to me, but something occurred to me as I was looking at the new queries. Do we even need the shipping method IDs? 

I think we only would if we were supporting shipping methods that don't support zones, since those presumably wouldn't have an `instance_id`, or at least wouldn't have a unique one. The `instance_id` is the primary key for the `woocommerce_shipping_zone_methods` table so it seems safe to assume that we can use it as a totally unique identifier for a method in a zone.

I hate to ask you to make additional significant changes to this PR, but I think it would simplify things a lot if this could be changed to only use `instance_id`s, potentially making a significant difference with query performance (for example, I'm wondering if using `CONCAT` for the `all` operator might cause inefficiencies). 

Maybe there's something I'm missing and the method IDs are actually needed. Let me know what you think!

mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedShippingMethod.php
rodrigoprimo: Marking as resolved based on your general comment. Thanks for raising this topic anyway.

rodrigoprimo:
Thanks again for the careful review, @johnolek! I appreciate your attention to detail.

> This looks great to me, but something occurred to me as I was looking at the new queries. Do we even need the shipping method IDs?

Good point. I agree that since we won't be supporting shipping methods that don't work with shipping zones, we don't need the shipping method IDs. I just pushed a new commit to simplify the code by removing the shipping method IDs.

Could you please check this PR again?

Please note that the build failed due to the performance tests which should not be a issue related to the changes proposed here. I'm reruning the build to see if it passes.

johnolek (dismissed):
Looks great! Thanks again for all the discussion and changes to make this as good as possible.

Aschepikov:
image
Hi @rodrigoprimo everything works ok, but can we change the display of the shipping methods? The length of the method depends on the country and the longest one is when the location is not covered
image

rodrigoprimo:
Good catch, @Aschepikov.

I implemented an workaround in a new commit using `overflow: auto`. Now if the text is longer than the existing space a scrollbar is added to the element and the text is not displayed anymore on top of the `X` used to remove the element.

This should work but it is not great as well. I'm not sure if we should get a designer involved here or if it is ok to proceed with the solution that I created (cc @NeosinneR and @johnolek for your thoughts).

Here is a screenshot:

Screenshot from 2023-06-22 11-07-14

NeosinneR:
> Now if the text is longer than the existing space a scrollbar is added to the element and the text is not displayed anymore on top of the X used to remove the element.

@rodrigoprimo have you or could you check this on Windows, or on macOS with scrollbars always visible (you can change that in System Settings)? I worry that with always visible scrollbars, this UX can be rather subpar. 

Screenshot 2023-06-22 at 16 21 32

NeosinneR:
Alternatively, I think it might be okay to truncate them, instead of letting them overflow. When choosing, you should still see long enough text before truncating to make sense of what you're picking. But I'm not sure if adding this globally won't mess with other usages of this component, e.g., when selecting emails, as they might have the same subject or at least the first portion of the subject.

rodrigoprimo:
> @rodrigoprimo have you or could you check this on Windows, or on macOS with scrollbars always visible (you can change that in System Settings)? I worry that with always visible scrollbars, this UX can be rather subpar.

@NeosinneR, I don't have easy access to a Windows or Mac machine to test this. Would you be able to test it?

> Alternatively, I think it might be okay to truncate them, instead of letting them overflow.

That is the first option that I tried. It works, but I was worried that some users might have shipping methods with longer names, and truncating them might make it impossible to differentiate. That is why I opted to show the scrollbar when needed. I was not aware that always displaying the scrollbar was an option. Maybe we decide once we know if changing the system setting that you mentioned affects the shipping method select box or not?

NeosinneR:
@rodrigoprimo This is how it looks with the always-on scrollbar, which is not the worst outcome. I think it's okay to ship it like this - I can't think of an easy way to fix it differently. 

Screenshot 2023-06-30 at 14 07 27

rodrigoprimo:
Thanks, @NeosinneR! Let's proceed with this solution then.

@Aschepikov, I'm assigning this PR back to you. I had to do a force push to fix a few conflicts.

Aschepikov (approved):
GJ!

PR #4969, opened 2023-06-05
Change 26 to 25 hours in the future when scheduling

veljkho:
## Description

Added missing + sign of the failing test. Changed back to 25 hours in the future, instead 26.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5374]

## After-merge notes

_N/A_

[MAILPOET-5374]: https://mailpoet.atlassian.net/browse/MAILPOET-5374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Newsletters/ScheduleNewsletterCest.php
+++ b/mailpoet/tests/acceptance/Newsletters/ScheduleNewsletterCest.php
@@ -79,7 +79,7 @@ class ScheduleNewsletterCest {
 
     $i->wantTo('Pick tomorrow‘s date');
     $i->click('select[name=time]');
-    $i->selectOption('form select[name=time]', $currentDateTime->modify("26 hour")->format('g:00 a'));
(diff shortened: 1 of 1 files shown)


brezocordero (changes requested):
@veljko could it be possible to change the test so that it always passes no matter when it runs?

veljkho:
> @veljko could it be possible to change the test so that it always passes no matter when it runs?

Hi @brezocordero 
I think I see now where was the problem and I updated the test to include the + sign before the value. I missed to add that when I first updated the test. It worked before normally with -1 and +25, so always working test.

brezocordero (approved):
LGTM!

PR #4972, opened 2023-06-05
WP user & WC customer fields [MAILPOET-5168] [MAILPOET-5171]

JanJakes:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5168]
[MAILPOET-5171]

## After-merge notes

_N/A_

[MAILPOET-5168]: https://mailpoet.atlassian.net/browse/MAILPOET-5168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5171]: https://mailpoet.atlassian.net/browse/MAILPOET-5171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
@@ -5,17 +5,25 @@ namespace MailPoet\Automation\Integrations\WooCommerce\Fields;
 use DateTimeImmutable;
 use DateTimeZone;
 use MailPoet\Automation\Engine\Data\Field;
+use MailPoet\Automation\Engine\WordPress;
 use MailPoet\Automation\Integrations\WooCommerce\Payloads\CustomerPayload;
 use MailPoet\Automation\Integrations\WooCommerce\WooCommerce;
 use WC_Customer;
+use WP_Error;
+use WP_Term;
 
 class CustomerOrderFieldsFactory {
   /** @var WooCommerce */
   private $wooCommerce;
 
+  /** @var WordPress */
+  private $wordPress;
+
   public function __construct(
+    WordPress $wordPress,
     WooCommerce $wooCommerce
   ) {
+    $this->wordPress = $wordPress;
     $this->wooCommerce = $wooCommerce;
   }
 
@@ -75,6 +83,30 @@ class CustomerOrderFieldsFactory {
           return $this->getPaidOrderDate($customer, false);
         }
       ),
+      new Field(
+        'woocommerce:customer:purchased-categories',
+        Field::TYPE_ENUM_ARRAY,
+        __('Purchased categories', 'mailpoet'),
+        function (CustomerPayload $payload) {
+          $customer = $payload->getCustomer();
+          return $customer ? $this->getOrderProductTermIds($customer, 'product_cat') : [];
+        },
+        [
+          'options' => $this->getTermsList('product_cat'),
+        ]
+      ),
+      new Field(
+        'woocommerce:customer:purchased-tags',
+        Field::TYPE_ENUM_ARRAY,
+        __('Purchased tags', 'mailpoet'),
+        function (CustomerPayload $payload) {
+          $customer = $payload->getCustomer();
+          return $customer ? $this->getOrderProductTermIds($customer, 'product_tag') : [];
+        },
+        [
+          'options' => $this->getTermsList('product_tag'),
+        ]
+      ),
     ];
   }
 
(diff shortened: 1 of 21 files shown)


mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
```diff
@@ -0,0 +1,217 @@
...
+    $sorting = $fetchFirst ? 'ASC' : 'DESC';
+    $statuses = array_map(function (string $status) {
+      return "wc-$status";
+    }, $this->wooCommerce->wcGetIsPaidStatuses());
+    $statusesPlaceholder = implode(',', array_fill(0, count($statuses), '%s'));
+
+    if ($this->wooCommerce->isWooCommerceCustomOrdersTableEnabled()) {
+      $statement = (string)$wpdb->prepare("
+        SELECT o.date_created_gmt
+        FROM {$wpdb->prefix}wc_orders o
```
websupporter, line 124: Not sure if HPOS plays into this. Orders used to be a post type, I remember we had something like a switch whether HPOS is active or not?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
```diff
@@ -0,0 +1,217 @@
...
+    } else {
+      $statement = (string)$wpdb->prepare("
+        SELECT p.post_date_gmt
+        FROM {$wpdb->prefix}posts p
+        LEFT JOIN {$wpdb->prefix}postmeta pm_total ON p.ID = pm_total.post_id AND pm_total.meta_key = '_order_total'
+        LEFT JOIN {$wpdb->prefix}postmeta pm_user ON p.ID = pm_user.post_id AND pm_user.meta_key = '_customer_user'
+        WHERE p.post_type = 'shop_order'
+        AND p.post_status IN ($statusesPlaceholder)
+        AND pm_user.meta_value = %d
+        AND pm_total.meta_value > 0
```
websupporter, line 140: Is this meta_value query really necessary? Maybe we can query also the items with `<= 0` and remove them in PHP. That should be faster imo.

websupporter: Actually, why not using
```
    wc_get_orders([
      'customer_id' => $customer->get_id(),
      'status' => $statuses,
      'limit' => 1,
      'orderby' => 'date',
      'order' => $sorting,
    ]);
```

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
```diff
@@ -0,0 +1,217 @@
...
+        ORDER BY p.post_date_gmt {$sorting}
+        LIMIT 1
+      ", array_merge($statuses, [$customer->get_id()]));
+    }
+
+    $date = $wpdb->get_var($statement);
+    return $date ? new DateTimeImmutable($date, new DateTimeZone('GMT')) : null;
+  }
+
+  private function getOrderProductTermIds(WC_Customer $customer, string $taxonomy): array {
```
websupporter, line 150: Here again, I am not sure, if we should directly query the database. There are reasons to do so sometimes (performance), but not using the WooCommerce and WP Api comes at a cost: Skipping several caching layers and ignoring the `applyFilters()` which might integrate custom logic.

Tbh at the moment I am not sure which side I lean towards

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
JanJakes: > Is this meta_value query really necessary? Maybe we can query also the items with <= 0 and remove them in PHP. That should be faster imo.

Fetching all orders to PHP would almost certainly be slower (fetching all order data, hydrating all of them, etc.), and could cause out-of-memory errors, so we'd need to iterate the result paginated in batches, and look for first/last order with `total > 0`. It's doable, but probably becomes more complex than the query.

> Actually, why not using

@websupporter I had that at first, but you can't do `total > 0` this way, so there really is not much of a choice.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
JanJakes: @websupporter Similarly as in the previous case, we'd need to iterate all orders of the customer (paginated), then iterate all products withing them, and then fetch all tag/category IDs from those products. That is a huge amount of data and CPU time to fetch just an array of IDs. At some point HPOS will be the default, and the complex non-HPOS branches will be deleted.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
JanJakes: @websupporter I checked AW, and they seem to be doing this: https://github.com/woocommerce/automatewoo/blob/6264b07f637bd17ff44f33cba303515e8a0d1230/includes/Customer.php#L738

So, it's quite similar, although they fetch and hydrate products and then iterate those. I went a step further — directly for the data I need.

mailpoet/lib/Automation/Integrations/WordPress/Subjects/UserSubject.php
```diff
@@ -0,0 +1,97 @@
...
+      new Field(
+        'wordpress:user:email',
+        Field::TYPE_STRING,
+        __('Email', 'mailpoet'),
+        function (UserPayload $payload) {
+          return $payload->getEmail();
+        }
+      ),
+      new Field(
+        'wordpress:user:is-guest',
```
websupporter, line 68: What is a WordPress User Guest? I understand the `exist()` boolean, but there is no such concept in WordPress at the end, right? Not sure if `wordpress:user:exists` makes sense for a field. But `guest` makes to me only sense in WooCommerce atm.

websupporter:
Thank you for working on this @JanJakes, some remarks, overall lgtm.

mailpoet/lib/Automation/Integrations/WordPress/Subjects/UserSubject.php
JanJakes: @websupporter Isn't it more the other way round — WP has guest users, but Woo doesn't? What I mean is that in WP you have a user with ID `0` and the `exists()` method, but Woo doesn't seem to have such a concept for customers.

I guess you're asking about the naming of this field? In any way, this is a question for @NeosinneR, as this is how it was specified.

NeosinneR: The idea behind is `Is guest` field is that you can have special automation for WooCommerce guest customers. E.g., if a guest finishes order, you can start an automation to let them know they can register and automatically get 5% discount or similar. Or share what benefits registering offers. 

So yeah, this is targeted toward WC rather than WP, but I'll let you two decide on what makes sense in the code.

JanJakes: @NeosinneR @websupporter Ahh, so the question was whether this belongs to WP or WC subject? Technically, it doesn't require WC at all, so it could be WP. But if we want to have it under WC (UX-wise), we could move it there. Or we could have it in both (as two "different" fields).

Note that it does make sense even without WC — for instance, I think someone may comment on a post, then we'll pass a WP user subject, and then `is-guest` can make sense for some automations.

websupporter (approved):
lgtm

Aschepikov (approved):
GJ!

PR #4975, opened 2023-06-06
Add form placement options for homepage, archive pages MAILPOET-4897

johnolek:
## Description

This PR adds new options for which pages a form will appear on:
- Display on the homepage
- Display on all post/product tag archives, or specifically selected tags
- Display on all post/product category archives, or specifically selected categories

Due to variability in theme types (block vs not) and themes themselves (no way to use an action/filter that would work the same for all themes), these options are not available for the "Below Pages" placement type.

This also includes a fix that prevents forms from appearing on the page selected as the "Posts page" in WordPress's `Settings > Reading` page, if one is selected. Now a form should correctly appear on this page if the `Display on all pages` options is set or if that page is specifically selected in the page selector. For the same reason mentioned above, this fix cannot apply to the `Below pages` placement.

## Code review notes

This required some refactoring for generating the form markup because forms can now be output in the footer instead of the content, which requires targeting an action (`wp_footer`) instead of a filter (`the_content`). On any page that calls both of these hooks, like any single page/post, this greatly increases the chance of duplicate forms getting rendered. In order to avoid overly complex logic to make these hooks mutually exclusive, I added state to `DisplayFormInWPContent` to track any time a form with a specific placement type was rendered and to subsequently skip forms of the same placement type.

## QA notes

This PR touches some of the primary form rendering logic, so in addition to testing that the new placement options work, please also confirm that the existing form placements work the same way as before (other than desired changes, like being able to target the selected "Posts page" as described above).

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4897](https://mailpoet.atlassian.net/browse/MAILPOET-4897)

## After-merge notes

_N/A_

[MAILPOET-4897]: https://mailpoet.atlassian.net/browse/MAILPOET-4897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Form/DisplayFormInWPContent.php
```diff
@@ -313,4 +357,28 @@ private function shouldDisplayFormOnTag(array $setup): bool {
...
+  private function shouldDisplayFormOnFrontPage(array $setup): bool {
+    return ($setup['homepage'] ?? false) === '1';
+  }
+
+  private function shouldDisplayFormOnCategoryArchive($setup): bool {
+    if (!isset($setup['categoryArchives'])) return false;
+    if (($setup['categoryArchives']['all'] ?? false) === '1') return true;
+    $selectedCategories = $setup['categoryArchives']['selected'] ?? [];
+    if ($selectedCategories === []) return false;
+    return $this->wp->hasCategory($selectedCategories) || $this->wp->hasTerm($selectedCategories, 'product_cat');
```
costasovo, line 392: Just to confirm my understanding: `$this->wp->hasCategory($selectedCategories)` covers posts categories and `$this->wp->hasTerm($selectedCategories, 'product_cat')` product categories, correct?

costasovo (dismissed):
@johnolek Nice work! Thank you for the thorough approach to this task and for adding unit tests for all the cases.
I'm approving the PR. I added one comment with a question just to confirm my understanding.

mailpoet/lib/Form/DisplayFormInWPContent.php
johnolek: Exactly. I followed the example of `shouldDisplayFormOnCategory` and `shouldDisplayFormOnTag`, which work the same way, including how categories/tags are selected.

Aschepikov:
Hi @johnolek I have a question about "Display on the homepage", should it be included in the 
"Display on all pages"? When the "Display on all pages" is enabled it is included on the homepage too, but there is no homepage in the "all pages" list, and when I enable "Display on all pages" I am thinking about pages that are in the list. With no homepage in the list and a separate toggle for a homepage, I can expect that it shouldn't include the homepage

johnolek:
> When the "Display on all pages" is enabled it is included on the homepage too, but there is no homepage in the "all pages" list, and when I enable "Display on all pages" I am thinking about pages that are in the list. With no homepage in the list and a separate toggle for a homepage, I can expect that it shouldn't include the homepage

Great point, @Aschepikov! I agree that the "all pages" option should only apply to the pages that would appear in the list. I've updated the behavior in 21aad26e99fb2aa8de221583eaa01794e938848c.

Aschepikov:
Hi @johnolek there are some checks that failed, can you look at it?

johnolek:
@Aschepikov I rebased on `trunk` and the tests are now passing. The failures should not have been related to any of the changes in this PR.

Aschepikov (approved):
GJ!

PR #4978, opened 2023-06-07
Improve the test Create Custom Field in Form

veljkho:
## Description

Improved the test to add additional checks.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5317]

## After-merge notes

_N/A_

[MAILPOET-5317]: https://mailpoet.atlassian.net/browse/MAILPOET-5317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Forms/EditorCreateCustomFieldCest.php
+++ b/mailpoet/tests/acceptance/Forms/EditorCreateCustomFieldCest.php
@@ -37,16 +37,22 @@ class EditorCreateCustomFieldCest {
   public function createCustomSelect(\AcceptanceTester $i) {
     $i->wantTo('Create custom field: select');
     $i->wantTo('Configure, check and save the custom field block');
+    $customFieldName = 'My custom select';
     $i->waitForElement('[data-automation-id="create_custom_field_form"]');
     $i->selectOption('[data-automation-id="create_custom_field_type_select"]', 'Select');
-    $i->fillField('[data-automation-id="create_custom_field_name_input"]', 'My custom select');
+    $i->fillField('[data-automation-id="create_custom_field_name_input"]', $customFieldName);
+    $i->fillField('[data-automation-id="settings_custom_text_label_input"]', 'Custom label');
+    $i->click('.components-form-toggle__input'); // Toggle as mandatory field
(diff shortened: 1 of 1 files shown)


mailpoet/tests/acceptance/Forms/EditorCreateCustomFieldCest.php
```diff
@@ -74,32 +92,44 @@ public function createCustomTextInput(\AcceptanceTester $i) {
     $this->checkCustomTextInputInForm($i);
 
     $i->wantTo('Change text input validation');
-    $i->selectOption('[data-automation-id="settings_custom_text_input_validation_type"]', 'Alphanumerical');
+    $i->selectOption('[data-automation-id="settings_custom_text_input_validation_type"]', 'Numbers only');
 
     $i->wantTo('Update label and save the form');
-    $i->fillField('[data-automation-id="settings_custom_text_label_input"]', 'My updated custom text input');
+    $i->fillField('[data-automation-id="settings_custom_text_label_input"]', $customFieldName . ' updated');
```
rodrigoprimo, line 98: This is just a detail and feel free to ignore it, but maybe since you already refactored this test to use `$customFieldName` instead of hard coding `My custom select` in several places, I wonder if we should do the same here and create a variable like `$updatedCustomFieldName` and used it instead of hard coding `$customFieldName . ' updated'` in several places. What do you think?

rodrigoprimo (approved):
Looks good to me! I left just one nitpick comment. Feel free to implement my suggestion or ignore it and merge this PR.

mailpoet/tests/acceptance/Forms/EditorCreateCustomFieldCest.php
veljkho: Thanks for quick review! I've been thinking about this and decided to make it this way just because there is also additional check on the front-end with * sign, in that case it would be `$updatedCustomFieldName . ' *'` so anyway I would need to use that . plus thing.

PR #4984, opened 2023-06-08
Order fields [MAILPOET-5169]

JanJakes:
## Description

Adds WooCommerce Order fields to automations.

## Code review notes

Woo order fields + some additional small fixes.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5169]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5169]: https://mailpoet.atlassian.net/browse/MAILPOET-5169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
@@ -12,6 +12,9 @@ use WC_Payment_Gateway;
 use WP_Post;
 
 class OrderFieldsFactory {
+  /** @var TermOptionsBuilder */
+  private $termOptionsBuilder;
+
   /** @var WordPress */
   private $wordPress;
 
@@ -19,9 +22,11 @@ class OrderFieldsFactory {
   private $wooCommerce;
 
   public function __construct(
+    TermOptionsBuilder $termOptionsBuilder,
     WordPress $wordPress,
     WooCommerce $wooCommerce
   ) {
+    $this->termOptionsBuilder = $termOptionsBuilder;
     $this->wordPress = $wordPress;
     $this->wooCommerce = $wooCommerce;
   }
@@ -199,6 +204,23 @@ class OrderFieldsFactory {
             $order = $payload->getOrder();
             return !$this->previousOrderExists($order);
           }
+        ),
+        new Field(
+          'woocommerce:order:categories',
+          Field::TYPE_ENUM_ARRAY,
+          __('Categories', 'mailpoet'),
+          function (OrderPayload $payload) {
+            $products = $this->getProducts($payload->getOrder());
+            $categoryIds = [];
+            foreach ($products as $product) {
+              $categoryIds = array_merge($categoryIds, $product->get_category_ids());
+            }
+            sort($categoryIds);
+            return array_unique($categoryIds);
+          },
+          [
+            'options' => $this->termOptionsBuilder->getCategoryOptions(),
+          ]
         ),
       ]
     );
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
@@ -296,7 +296,10 @@ class OrderFieldsFactory {
     $statuses = $this->wooCommerce->wcGetOrderStatuses();
     $options = [];
     foreach ($statuses as $id => $name) {
-      $options[] = ['id' => $id, 'name' => $name];
+      $options[] = [
+        'id' => substr($id, 0, 3) === 'wc-' ? substr($id, 3) : $id,
+        'name' => $name,
+      ];
     }
     return $options;
   }
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
@@ -227,7 +227,7 @@ class OrderFieldsFactory {
             return array_unique($categoryIds);
           },
           [
-            'options' => $this->termOptionsBuilder->getCategoryOptions(),
+            'options' => $this->termOptionsBuilder->getTermOptions('product_cat'),
           ]
         ),
         new Field(
@@ -244,7 +244,7 @@ class OrderFieldsFactory {
             return array_unique($tagIds);
           },
           [
-            'options' => $this->termOptionsBuilder->getTagOptions(),
+            'options' => $this->termOptionsBuilder->getTermOptions('product_tag'),
           ]
         ),
         new Field(
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
@@ -58,6 +58,54 @@ class OrderFieldsFactory {
             return $payload->getOrder()->get_billing_country();
           }
         ),
+        new Field(
+          'woocommerce:order:shipping-company',
+          Field::TYPE_STRING,
+          __('Shipping company', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_shipping_company();
+          }
+        ),
+        new Field(
+          'woocommerce:order:shipping-phone',
+          Field::TYPE_STRING,
+          __('Shipping phone', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_shipping_phone();
+          }
+        ),
+        new Field(
+          'woocommerce:order:shipping-city',
+          Field::TYPE_STRING,
+          __('Shipping city', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_shipping_city();
+          }
+        ),
+        new Field(
+          'woocommerce:order:shipping-postcode',
+          Field::TYPE_STRING,
+          __('Shipping postcode', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_shipping_postcode();
+          }
+        ),
+        new Field(
+          'woocommerce:order:shipping-state',
+          Field::TYPE_STRING,
+          __('Shipping state/county', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_shipping_state();
+          }
+        ),
+        new Field(
+          'woocommerce:order:shipping-country',
+          Field::TYPE_STRING,
+          __('Shipping country', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_shipping_country();
+          }
+        ),
       ]
     );
   }
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
@@ -2,10 +2,12 @@
 
 namespace MailPoet\Automation\Integrations\WooCommerce\Fields;
 
+use DateTimeImmutable;
 use MailPoet\Automation\Engine\Data\Field;
 use MailPoet\Automation\Engine\WordPress;
 use MailPoet\Automation\Integrations\WooCommerce\Payloads\OrderPayload;
 use MailPoet\Automation\Integrations\WooCommerce\WooCommerce;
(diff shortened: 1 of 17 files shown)


mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
```diff
@@ -0,0 +1,347 @@
...
+          'woocommerce:order:billing-state',
+          Field::TYPE_STRING,
+          __('Billing state/county', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_billing_state();
+          }
+        ),
+        new Field(
+          'woocommerce:order:billing-country',
+          Field::TYPE_STRING,
```
websupporter: Would `ENUM´ make sense here?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
```diff
@@ -0,0 +1,347 @@
...
+          'woocommerce:order:shipping-state',
+          Field::TYPE_STRING,
+          __('Shipping state/county', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_shipping_state();
+          }
+        ),
+        new Field(
+          'woocommerce:order:shipping-country',
+          Field::TYPE_STRING,
```
websupporter: `ENUM`?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
```diff
@@ -0,0 +1,347 @@
...
+      'posts_per_page' => -1,
+      'orderby' => 'name',
+      'order' => 'asc',
+    ]);
+
+    $options = [];
+    foreach ($coupons as $coupon) {
+      if ($coupon instanceof WP_Post) {
+        // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+        $options[] = ['id' => $coupon->post_title, 'name' => $coupon->post_name];
```
websupporter: I assume post_title is the ID because this is the code stored in the order item? But I am not sure why the `post_name` is the title and not again `post_title`?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
```diff
@@ -0,0 +1,347 @@
...
+          'woocommerce:order:status',
+          Field::TYPE_ENUM,
+          __('Status', 'mailpoet'),
+          function (OrderPayload $payload) {
+            return $payload->getOrder()->get_status();
+          },
+          [
+            'options' => $this->getOrderStatusOptions(),
+          ]
+        ),
```
websupporter, line 187: I am a bit afraid of the `wc-` prefix and if `get_status()` and `wc_get_order_statuses()` do match in the end. WooCommerce sometimes uses the prefix, sometimes it doesn't. Internally, it needs sometimes to do things like this: https://github.com/woocommerce/woocommerce/blob/1e8bbb2800c2c446eb663e2c802acc39a27e5322/plugins/woocommerce/includes/wc-order-functions.php#L147-L152

`wc_get_order_statuses()` returns with the prefix. Looking into the data base the prefix is stored with the order. So I assume `get_status()` returns the status with the prefix. But did you had a closer look into it?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
```diff
@@ -0,0 +1,347 @@
...
+        new Field(
+          'woocommerce:order:is-first-order',
+          Field::TYPE_BOOLEAN,
+          __('Is first order', 'mailpoet'),
+          function (OrderPayload $payload) {
+            $order = $payload->getOrder();
+            return !$this->previousOrderExists($order);
+          }
+        ),
```
websupporter, line 215: Maybe instead of the query we could utilize `$customer->get_order_count()`

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
```diff
@@ -0,0 +1,58 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Integrations\WooCommerce\Fields;
+
+use MailPoet\Automation\Engine\WordPress;
+use WP_Error;
+use WP_Term;
+
+class TermOptionsBuilder {
```
websupporter, line 9: It is in the WooCommerce namespace, but `TermOptionBuilder` still sounds a bit too generic to me. As if I could build term options for all WordPress terms.

Not sure if it makes sense to have a generic `TermOptionBuilder` already/at all and a `WooCommerceTermOptionBuilder` which would use the generic one as dependency :thinking:

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
```diff
@@ -0,0 +1,58 @@
...
+  private $wordPress;
+
+  public function __construct(
+    WordPress $wordPress
+  ) {
+    $this->wordPress = $wordPress;
+  }
+
+  /** @return array<array{id: int, name: string}> */
+  public function getCategoryOptions(): array {
```
websupporter: I would rename it to getProductCategoryOptions I think. When reading the function somewhere where it's used it would be immediately clear what you get imo

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
```diff
@@ -0,0 +1,58 @@
...
+    return $this->getTermOptions('product_cat');
+  }
+
+  /** @return array<array{id: int, name: string}> */
+  public function getTagOptions(): array {
```
websupporter: Same as above

websupporter:
Thank you @JanJakes, some small questions and remarks

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
JanJakes: @websupporter You mean using something like `WC_Countries`, right? Sounds like a good idea 👍

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
JanJakes: @websupporter Good catch, doesn't seem to make any sense to use `post_name` 👍

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
websupporter: Yep. You might want to restrict it further to "only the countries we sell to" 
image

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
JanJakes: @websupporter That's not equivalent, right? Imagine the field is used in a condition after some delays and stuff. What if more orders were created in the meantime? Then the first order is still the first one, I guess.

websupporter: Right! Thanks for clarifying.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
JanJakes: @websupporter The generic solution is [the private method](https://github.com/mailpoet/mailpoet/pull/4984/files#diff-f83d08b262f774ec342be404b44009815724dada8d48f61d3a2abc86577ae52fR30). Should I make it public and remove the helper methods for categories and tags? Would solve the naming issues as well, I guess.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
JanJakes: @websupporter Very good point.

So:
- `wc_get_order_statuses()` includes the `wc-` prefix
- `$order->get_status()` doesn't (even `$order->set_status()` followed by `$order->get_status()` already strips it)

We do one of:
1. Append `wc-` when fetching status (`'wc-' . $payload->getOrder()->get_status()`).
2. Strip `wc-` from the list of all statuses, when the prefix exists.

These are similar, but not fully equivalent. If WC saves the status with `wc-`, then maybe I'd lean to doing the same. On the other hand, do all WC statuses, including some custom ones, have the `wc-` prefix? If there's a risk of "no, they don't", then 2. is safer, I guess.

What do you think?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
JanJakes: https://github.com/mailpoet/mailpoet/pull/4984/commits/e0938bd6bb7a1452f3875602733c3ecea05a850b + https://github.com/mailpoet/mailpoet/pull/4984/commits/30b7eb6c57a91dd6a6804a2180824a99c744fa87

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
JanJakes: https://github.com/mailpoet/mailpoet/pull/4984/commits/e0938bd6bb7a1452f3875602733c3ecea05a850b + https://github.com/mailpoet/mailpoet/pull/4984/commits/30b7eb6c57a91dd6a6804a2180824a99c744fa87

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
JanJakes: https://github.com/mailpoet/mailpoet/pull/4984/commits/359e47e38669971d964eb38dd5e9f4d1fc867421

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/OrderFieldsFactory.php
JanJakes: As discussed on Slack, option 2 seems safer: https://github.com/mailpoet/mailpoet/pull/4984/commits/6666e9e10c26911c35cb7abe9162099eeb11ac2b

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
JanJakes: https://github.com/mailpoet/mailpoet/pull/4984/commits/9bebd41eaea85282facc243f5bff6133e24db1a3

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
JanJakes: https://github.com/mailpoet/mailpoet/pull/4984/commits/9bebd41eaea85282facc243f5bff6133e24db1a3

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
JanJakes: https://github.com/mailpoet/mailpoet/pull/4984/commits/9bebd41eaea85282facc243f5bff6133e24db1a3

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
websupporter: The only thing I would remark is that this could move in a more generic namespace or a `WordPress` namespace. But since we do not have a `WordPress` Integration atm, we can do that later imo.

websupporter (approved):
lgtm. thank you.

Aschepikov (approved):
GJ!

PR #4982, opened 2023-06-08
Fix for #4981 [MAILPOET-5383]

Drivingralle:
## Description

replace Null with empty string

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

Fixes: #4981

[MAILPOET-5383]

## After-merge notes

_N/A_

## Tasks

- [x] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] I added sufficient test coverage
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5383]: https://mailpoet.atlassian.net/browse/MAILPOET-5383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4985, opened 2023-06-08
Prevent PHP notices [MAILPOET-5383]

### LINKED ISSUE #4981: PHP 8: Deprication notice due to wrong parameter type
**Describe the bug**

On this file: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Config/Menu.php
The addSubmenuPage method uses `null` instead of an empty String and this causes a Depreciation notice on PHP >= 8.0 

Here is an example notice
Bildschirmfoto 2023-06-08 um 11 53 40

**To reproduce**

1. Enable all logging
2. Enable Mailpoet
3. Visit any backend view

**Actual behavior**
An Depreciation notice is given

**Expected behavior**
No notices

**Screenshots**

- WordPress version: 6.2.2
- PHP version: 8.1.9
- MailPoet version: 4.18.0 

**Additional context**

JanJakes:
## Description

Fix deprecation notices in Menu, from: https://github.com/mailpoet/mailpoet/pull/4982

Thanks, [Drivingralle](https://github.com/Drivingralle)!

Fixes: https://github.com/mailpoet/mailpoet/issues/4981
Closes: https://github.com/mailpoet/mailpoet/pull/4982

## Linked tickets

[MAILPOET-5383](https://mailpoet.atlassian.net/browse/MAILPOET-5383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

## Tasks

[MAILPOET-5383]: https://mailpoet.atlassian.net/browse/MAILPOET-5383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/WP/Functions.php
+++ b/mailpoet/lib/WP/Functions.php
@@ -72,12 +72,23 @@ class Functions {
     return add_image_size($name, $width, $height, $crop);
   }
 
-  public function addMenuPage($pageTitle, $menuTitle, $capability, $menuSlug, callable $function = null, $iconUrl = '', $position = null) {
-    if (is_null($function)) {
-      $function = function () {
+  /**
+   * @param string $pageTitle
+   * @param string $menuTitle
+   * @param string $capability
+   * @param string $menuSlug
+   * @param callable $callback|null
+   * @param string $iconUrl
+   * @param int $position
+   * @return string The resulting page's hook_suffix.
+   * @return string
+   */
+  public function addMenuPage($pageTitle, $menuTitle, $capability, $menuSlug, callable $callback = null, $iconUrl = '', $position = null) {
+    if (is_null($callback)) {
+      $callback = function () {
       };
     }
-    return add_menu_page($pageTitle, $menuTitle, $capability, $menuSlug, $function, $iconUrl, $position);
+    return add_menu_page($pageTitle, $menuTitle, $capability, $menuSlug, $callback, $iconUrl, $position);
   }
 
   public function addQueryArg($key, $value = false, $url = false) {
@@ -92,8 +103,18 @@ class Functions {
     return add_shortcode($tag, $callback);
   }
 
-  public function addSubmenuPage($parentSlug, $pageTitle, $menuTitle, $capability, $menuSlug, callable $function) {
-    return add_submenu_page($parentSlug, $pageTitle, $menuTitle, $capability, $menuSlug, $function);
+  /**
+   * @param string $parentSlug
+   * @param string $pageTitle
+   * @param string $menuTitle
+   * @param string $capability
+   * @param string $menuSlug
+   * @param callable $function
+   * @param int $position
+   * @return string|false
+   */
+  public function addSubmenuPage($parentSlug, $pageTitle, $menuTitle, $capability, $menuSlug, callable $function, $position = null) {
+    return add_submenu_page($parentSlug, $pageTitle, $menuTitle, $capability, $menuSlug, $function, $position);
   }
 
   public function adminUrl($path = '', $scheme = 'admin') {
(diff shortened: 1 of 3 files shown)


NeosinneR:
@JanJakes is no one assigned on purpose?

JanJakes:
@NeosinneR Nope, I think I pushed it to wait for the build and didn't get back. Fixed.

mailpoet/lib/WP/Functions.php
```diff
@@ -72,12 +72,23 @@ public function addImageSize($name, $width = 0, $height = 0, $crop = false) {
...
+   * @param string $menuTitle
+   * @param string $capability
+   * @param string $menuSlug
+   * @param callable $callback|null
+   * @param string $iconUrl
+   * @param int $position
+   * @return string The resulting page's hook_suffix.
+   * @return string
+   */
+  public function addMenuPage($pageTitle, $menuTitle, $capability, $menuSlug, callable $callback = null, $iconUrl = '', $position = null) {
```
brezocordero: Not a blocker but as we are adding the dockBlock, I would suggest matching the WordPress function signature: https://github.com/WordPress/wordpress-develop/blob/f0a27d908b41368b568189c48b65d873610f5e99/src/wp-admin/includes/plugin.php#L1397
`callback` has a default value of `''` and position can be `int|float`

mailpoet/lib/WP/Functions.php
```diff
@@ -92,8 +103,18 @@ public function addShortcode($tag, callable $callback) {
...
+   * @param string $parentSlug
+   * @param string $pageTitle
+   * @param string $menuTitle
+   * @param string $capability
+   * @param string $menuSlug
+   * @param callable $function
+   * @param int $position
+   * @return string|false
+   */
+  public function addSubmenuPage($parentSlug, $pageTitle, $menuTitle, $capability, $menuSlug, callable $function, $position = null) {
```
brezocordero: Same suggestion as above, to match the WordPress function:
https://github.com/WordPress/wordpress-develop/blob/f0a27d908b41368b568189c48b65d873610f5e99/src/wp-admin/includes/plugin.php#L1387

brezocordero (commented):
It all works well. Thanks @Drivingralle !

@JanJakes I have only a couple of small suggestions but they are not blocking.
Let me know what you think.

mailpoet/lib/WP/Functions.php
JanJakes: @brezocordero Thanks for noticing. Both of these cases are a bit problematic:

> position can be int|float

This is a [relatively recent](https://github.com/WordPress/wordpress-develop/commit/4f290ecd64e0e30bb52bbaf63446254e68e577f3) addition to WP and to make PHPStan understand, we'd need to install `php-stubs/wordpress-stubs` in version `^6.0.0`, which requires updating `szepeviktor/phpstan-wordpress`, which in turn requires updating PHPStan itself 😬

At this point, I'd suggest leaving it at `int`, until we update all of those packages.

> callback has a default value of ''

WP annotates that as a `callback`, but passes an empty string, that PHPStan will complain about. The only solution I could think of was to add an ignore rule to PHPStan config.

brezocordero: Thanks for looking into it @JanJakes.
Problematic indeed. We can leave it as is.

brezocordero (approved):

PR #4990, opened 2023-06-09
Segment filters for entered/exited automations [MAILPOET-5003]

johnolek:
## Description

This PR adds two new dynamic segment filters for subscriber entered/exited automation(s).

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs
_N/A_

## Linked tickets

[MAILPOET-5003](https://mailpoet.atlassian.net/browse/MAILPOET-5003)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5003]: https://mailpoet.atlassian.net/browse/MAILPOET-5003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
+++ b/mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
@@ -4,6 +4,7 @@ namespace MailPoet\Segments\DynamicSegments;
 
 use MailPoet\Entities\DynamicSegmentFilterData;
 use MailPoet\Segments\DynamicSegments\Exceptions\InvalidFilterException;
+use MailPoet\Segments\DynamicSegments\Filters\AutomationsEvents;
 use MailPoet\Segments\DynamicSegments\Filters\DateFilterHelper;
 use MailPoet\Segments\DynamicSegments\Filters\EmailAction;
 use MailPoet\Segments\DynamicSegments\Filters\EmailActionClickAny;
@@ -66,6 +67,8 @@ class FilterDataMapper {
 
   private function createFilter(array $filterData): DynamicSegmentFilterData {
     switch ($this->getSegmentType($filterData)) {
+      case DynamicSegmentFilterData::TYPE_AUTOMATIONS:
+        return $this->createAutomations($filterData);
       case DynamicSegmentFilterData::TYPE_USER_ROLE:
         return $this->createSubscriber($filterData);
       case DynamicSegmentFilterData::TYPE_EMAIL:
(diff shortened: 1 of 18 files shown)


johnolek:
@JanJakes and/or @websupporter, I would appreciate your review of one very specific part of this PR, which is the actual query that's constructed in `mailpoet/lib/Segments/DynamicSegments/Filters/AutomationsEvents.php`.

The solution is far from ideal. It relies on the idea that subscribers will only ever be associated with an automation run via the `mailpoet:subscriber` subject, and that the `args` for that subject will look exactly like `{"subscriber_id":14}`.

I don't think that adding an index to `args` would even help in this case because of the reliance on `CONCAT`. It would of course be nice if we could use a JSON column instead of LONGTEXT, but I assume supporting MySQL 5.5 prevents that.

Here are some ideas I had:

1. Performance could probably be improved a bit if the join is only based on the `key` of `mailpoet_automation_run_subjects` being `mailpoet:subscriber`, since the `key` is indexed, and then using a WHERE for the `args`. Very much a partial solution.
2. We could add Yet Another Automations Table for subject args, one row for each, with indexes for the keys and values. This wouldn't even necessarily have to be a refactor to the automations code -- it could exist strictly as a lookup table, and I suppose we could even limit it to specific `key`s in that case. This might not be perfect since we would probably lose the type data we're getting with JSON.
3. We could add Yet Another Subscribers Column to store a hash that would correspond to the hash of a `mailpoet:subscriber` subject. I don't like this idea because it would require a migration that could potentially take a long time for very large subscribers tables to backfill data, and it seems like an overly specific solution to a problem that I suspect we'll see come up again in a different form (i.e. needing to look up automation runs based on other subject data).

I'm hoping that I'm missing something more obvious that wouldn't require major changes to the automations code.

johnolek:
After more discussion in slack, we decided to leave the query logic as-is for now, but we will revisit it down the road.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
```diff
@@ -89,6 +92,39 @@ private function getSegmentType(array $data): string {
     return $data['segmentType'];
   }
 
+  private function createAutomations(array $data): DynamicSegmentFilterData {
```
costasovo, line 97: The method contains a bunch of conditions, so it might be useful to have it covered by a test. I believe this will make it easier for our future selves to feel more confident when making changes here.
We already have a [unit test](https://github.com/mailpoet/mailpoet/blob/974cd3295065e4c50855c985d50384aa07acfa23/mailpoet/tests/unit/Segments/DynamicSegments/FilterDataMapperTest.php) you can extend.

costasovo (commented):
Great job! Thanks for adding the extensive test for the filter 👍 
I added one suggestion about test coverage.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
johnolek: Thanks for pointing this out! I've updated the tests in 8964fb6aa2290a83f70ae34d995f195e2df65cfe. Please let me know if you think I missed any important cases.

costasovo (dismissed):
Thank you for adding the unit tests! 👍

Aschepikov:
Hi @johnolek When I opened the already-created segment with automation filters I couldn't see automations used in the segment. Every time when I open an already-created segment I don't see any selected automation but I can see the count of the subscribers. 
image
image
Can we change this behavior and show selected automations as usual in the segment filters?

johnolek:
Good catch @Aschepikov, thank you so much for finding that bug! I believe that should be fixed in 575399d7e08b139b86c194a9a970793b4ae837e6. Will you please re-test when you have a chance?

Aschepikov (approved):
GJ!

PR #4994, opened 2023-06-14
New filters for city and postal code [MAILPOET-4990]

johnolek:
## Description

This PR adds new dynamic segment filters for WooCommerce customers' city and postal codes. It also renames the `is in country` filter to simply `country`.

These filters work very similarly to the subscriber filters for first name, last name, and email address.

## Code review notes

_N/A_

## QA notes

I did some minor refactoring to allow re-use of some of the subscriber text field code, so it would be great if you could double check the first name, last name, and email filters.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4990](https://mailpoet.atlassian.net/browse/MAILPOET-4990)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4990]: https://mailpoet.atlassian.net/browse/MAILPOET-4990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
+++ b/mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
@@ -95,7 +95,9 @@ class FilterDataMapper {
       $data['action'] = DynamicSegmentFilterData::TYPE_USER_ROLE;
     }
     if ($data['action'] === SubscriberSubscribedDate::TYPE) {
-      if (empty($data['value'])) throw new InvalidFilterException('Missing number of days', InvalidFilterException::MISSING_VALUE);
+      if (empty($data['value'])) {
+        throw new InvalidFilterException('Missing number of days', InvalidFilterException::MISSING_VALUE);
+      }
       return new DynamicSegmentFilterData(DynamicSegmentFilterData::TYPE_USER_ROLE, $data['action'], [
         'value' => $data['value'],
         'operator' => $data['operator'] ?? DateFilterHelper::BEFORE,
@@ -103,7 +105,9 @@ class FilterDataMapper {
       ]);
     }
     if ($data['action'] === SubscriberScore::TYPE) {
-      if (!isset($data['value'])) throw new InvalidFilterException('Missing engagement score value', InvalidFilterException::MISSING_VALUE);
+      if (!isset($data['value'])) {
+        throw new InvalidFilterException('Missing engagement score value', InvalidFilterException::MISSING_VALUE);
+      }
       return new DynamicSegmentFilterData(DynamicSegmentFilterData::TYPE_USER_ROLE, $data['action'], [
         'value' => $data['value'],
         'operator' => $data['operator'] ?? SubscriberScore::HIGHER_THAN,
@@ -111,7 +115,9 @@ class FilterDataMapper {
       ]);
     }
     if ($data['action'] === SubscriberSegment::TYPE) {
-      if (empty($data['segments'])) throw new InvalidFilterException('Missing segments', InvalidFilterException::MISSING_VALUE);
+      if (empty($data['segments'])) {
+        throw new InvalidFilterException('Missing segments', InvalidFilterException::MISSING_VALUE);
+      }
       return new DynamicSegmentFilterData(DynamicSegmentFilterData::TYPE_USER_ROLE, $data['action'], [
         'segments' => array_map(function ($segmentId) {
           return intval($segmentId);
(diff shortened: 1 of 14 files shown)


mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
```diff
@@ -323,6 +324,15 @@ private function createWooCommerce(array $data): DynamicSegmentFilterData {
       $filterData['operator'] = $data['operator'];
       $filterData['payment_methods'] = $data['payment_methods'];
       $filterData['used_payment_method_days'] = intval($data['used_payment_method_days']);
+    } elseif (in_array($data['action'], WooCommerceCustomerTextField::ACTIONS)) {
+      if (empty($data['value'])) throw new InvalidFilterException('Missing value', InvalidFilterException::MISSING_VALUE);
+      if (empty($data['operator'])) throw new InvalidFilterException('Missing operator', InvalidFilterException::MISSING_VALUE);
```
rodrigoprimo: I guess the exception here should be `InvalidFilterException::MISSING_OPERATOR` instead of `InvalidFilterException::MISSING_VALUE`?

rodrigoprimo: And just one more nitpick that maybe it is just a personal preference of mine but I try to avoid using ifs without brackets. If my memory is not failing me, I believe the WP coding standard doesn't allow them (I know we don't use it). I tend to find ifs without brackets a little harder to read and a little easier to introduce errors (adding a new line below the if wrongly assuming that it will be inside the if block).

Since it seems our coding standard is ok with it, feel free to ignore this comment.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
```diff
@@ -323,6 +324,15 @@ private function createWooCommerce(array $data): DynamicSegmentFilterData {
...
+      if (!in_array($data['operator'], DynamicSegmentFilterData::TEXT_FIELD_OPERATORS)) {
+        throw new InvalidFilterException('Invalid operator', InvalidFilterException::MISSING_VALUE);
```
rodrigoprimo: Similar to the comment above, I belive the exception here is incorrect as well.

mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceCustomerTextField.php
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $filter): QueryBuilder {
+    $filterData = $filter->getFilterData();
+    $action = $filterData->getParam('action');
+    $value = $filterData->getParam('value');
+    $operator = $filterData->getParam('operator');
+
+    if (!is_string($action)) {
+      throw new InvalidFilterException('Missing action', InvalidFilterException::MISSING_VALUE);
```
rodrigoprimo: I think the exception used is incorrect in this line and on line 46.

rodrigoprimo (commented):
Great job! Thanks for taking the time to refactor and improve the code!

I left a few comments with some small things to double-check with you.

Besides that, one thing that I noticed that maybe was already discussed is that those two new filters and other filters like `country` behave in a way that is different from filters like `used shipping method` and `used payment method`. When using `none of` the former includes only WooCommerce customers, while the latter includes MailPoet subscribers that are not WooCommerce customers. Is this something that was already discussed? Do we want to do something about it? (cc @NeosinneR)

rodrigoprimo:
I remember we discussed including subscribers that are not WooCommerce customers in the `used shipping method` and `used payment method` filters but I don't recall if we took into consideration that other filters behave differently and that is why I'm raising this topic again.

NeosinneR:
> When using `none of` the former includes only WooCommerce customers, while the latter includes MailPoet subscribers that are not WooCommerce customers. Is this something that was already discussed? Do we want to do something about it?

Yeah, in general, these opposite conditions (`none of` and `any of`) should, combined, return 100% of _subscribers_ not just customers. The reasoning was, that you can easily segment just customers but adding one more condition (e.g., `# of orders > 0`).

johnolek:
@rodrigoprimo thanks for raising the question about the negative cases! I remember discussing this with @NeosinneR a while ago, specifically asking if `is in country` should behave the same way. I think the conclusion was that `is in country` is different from shipping/payment methods, but I could be misremembering.

In any case, I do think it makes sense to treat these filters differently. With shipping/payment methods, we can confidently say a subscriber hasn't used one because we have all the relevant data. With location data, we _can't_ confidently say that someone isn't in a particular country, city, or postal code just because they haven't placed an order.

Presumably a user would have a good reason for choosing a filter for subscribers who aren't in a particular city or zipcode, and it could be a problem if their email ends up going to people in those locations just because we don't have their location data yet.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
johnolek: I'm very glad you brought this up! I strongly dislike if statements that don't have braces -- I also find them more difficult to read. I wrote it this way to match the surrounding code, but ultimately decided to add braces to all the if statements in this file in 6ea2bf893979e1a82b6f0428c33a65b431ccb511.

johnolek:
@rodrigoprimo Thank you for your attention to detail! I think I fixed all of the incorrect exception codes, including for the `SubscriberTextField` code that I used as a starting point for this filter.

@NeosinneR based on my previous comment, can you please confirm if it's ok to leave the negative operators working they way they are?

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
rodrigoprimo: Thanks!

rodrigoprimo (dismissed):
The latest code changes look good to me. Thanks, John!

Feel free to move this to QA if @NeosinneR agrees with the current behavior.

NeosinneR:
@johnolek as part of this project, we'll be adding `is blank` and `is not blank` to some of the existing conditions. I think this could also solve the issue you raised:

> With location data, we can't confidently say that someone isn't in a particular country, city, or postal code just because they haven't placed an order.

When we add `is blank` and `is not blank` to these multi select conditions, then you could confidently use `country is not blank AND none of X, Y, Z`. 

But that's for another ticket. I'm okay to proceed with this one as is.

Aschepikov:
Hi @johnolek there are some conflicts, can you resolve them pls

johnolek:
@Aschepikov I resolved the conflicts.

Aschepikov (approved):
GJ!

PR #4998, opened 2023-06-15
Fix counts cache update for too few subscribers [MAILPOET-5390]

alex-mailpoet:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5390]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5390]: https://mailpoet.atlassian.net/browse/MAILPOET-5390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/AdminPages/PageRenderer.php
+++ b/mailpoet/lib/AdminPages/PageRenderer.php
@@ -125,7 +125,11 @@ class PageRenderer {
     $wpSegmentState = ($wpSegment instanceof SegmentEntity) && $wpSegment->getDeletedAt() === null ?
       SegmentEntity::SEGMENT_ENABLED : SegmentEntity::SEGMENT_DISABLED;
     $installedAtDiff = (new \DateTime($this->settings->get('installed_at')))->diff(new \DateTime());
-    $subscribersCacheCreatedAt = $this->transientCache->getOldestCreatedAt(TransientCache::SUBSCRIBERS_STATISTICS_COUNT_KEY) ?: Carbon::now();
+    $subscriberCount = $this->subscribersFeature->getSubscribersCount();
+    $subscribersCacheCreatedAt = Carbon::now();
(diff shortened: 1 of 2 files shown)


mailpoet/lib/Subscribers/SubscribersCountsController.php
```diff
@@ -169,9 +169,7 @@ private function getCacheItem(string $key, int $id): ?array {
   }
 
   private function setCacheItem(string $key, array $item, int $id): void {
-    if ($this->subscribersFeature->isSubscribersCountEnoughForCache()) {
-      $this->transientCache->setItem($key, $item, $id);
-    }
+    $this->transientCache->setItem($key, $item, $id);
```
costasovo: @alex-mailpoet  I see how it helps with the issue. But if I'm not mistaken, this change's side effect is that if the site has a low number of subscribers, the plugin will write to cache (so to DB) on every page load when subscriber count data are needed. It is probably not a heavy query, but also not a nice behavior.

Maybe we could fix it in UI or in the place where we send the data about the cache age to UI, and if `isSubscribersCountEnoughForCache` is `false` we say the cache is up to date and send 0.

What do you think?

costasovo (commented):
I added a question. Please see my other comment.

mailpoet/lib/Subscribers/SubscribersCountsController.php
alex-mailpoet: @costasovo Good idea, Rosta! I have made a change according to your proposition.

costasovo (approved):
LGTM! 👍

PR #4999, opened 2023-06-15
Convert Segments code to TSX [MAILPOET-5407]

rodrigoprimo:
## Description

This PR converts JSX code in `assets/js/src/segments/` to TSX. This is part of https://mailpoet.atlassian.net/browse/MAILPOET-5392. I decided to do it in a separate ticket to avoid creating a big PR.

## Code review notes

Please note that I'm still not very great with TS, so I might be missing something here. Any feedback is welcome.

The performance tests are failing but it doesn't seem to be related to the changes proposed here. I tried rerunning a few times but they continued failing.

## QA notes

This PR should not affect the behavior of the plugin, but it changes several files related to the lists and dynamic segments interface. If possible, please test anything related to those features that are not tested by our test suite.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5407]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5407]: https://mailpoet.atlassian.net/browse/MAILPOET-5407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/segments.tsx
+++ b/mailpoet/assets/js/src/segments/segments.tsx
@@ -4,18 +4,18 @@ import { HashRouter, Redirect, Route, Switch } from 'react-router-dom';
 import { MailPoet } from 'mailpoet';
 import { RoutedTabs } from 'common/tabs/routed_tabs';
 import { Tab } from 'common/tabs/tab';
-import { SegmentList } from 'segments/list.tsx';
-import { SegmentForm } from 'segments/form.tsx';
-import { ListHeading } from 'segments/heading.tsx';
+import { SegmentList } from 'segments/list';
+import { SegmentForm } from 'segments/form';
+import { ListHeading } from 'segments/heading';
 import { GlobalContext, useGlobalContextValue } from 'context';
 import { Notices } from 'notices/notices.jsx';
 import { registerTranslations, withBoundary } from 'common';
 import { Editor } from './dynamic/editor';
-import { DynamicSegmentList } from './dynamic/list.tsx';
+import { DynamicSegmentList } from './dynamic/list';
 
 const container = document.getElementById('segments_container');
 
-function Tabs() {
+function Tabs(): JSX.Element {
   return (
     <>
       <ListHeading />
@@ -42,19 +42,19 @@ function Tabs() {
 
 Tabs.displayName = 'SegmentTabs';
 
-function App() {
+function App(): JSX.Element {
   return (
     <GlobalContext.Provider value={useGlobalContextValue(window)}>
       <HashRouter>
         <Notices />
         <Switch>
           <Route exact path="/" render={() => <Redirect to="/lists" />} />
-          <Route path="/new" render={withBoundary(SegmentForm)} />
-          <Route path="/edit/:id" render={withBoundary(SegmentForm)} />
-          <Route path="/new-segment" render={withBoundary(Editor)} />
-          <Route path="/edit-segment/:id" render={withBoundary(Editor)} />
-          <Route path="/segments/(.*)?" render={withBoundary(Tabs)} />
-          <Route path="/lists/(.*)?" render={withBoundary(Tabs)} />
+          <Route path="/new" component={withBoundary(SegmentForm)} />
+          <Route path="/edit/:id" component={withBoundary(SegmentForm)} />
+          <Route path="/new-segment" component={withBoundary(Editor)} />
+          <Route path="/edit-segment/:id" component={withBoundary(Editor)} />
+          <Route path="/segments/(.*)?" component={withBoundary(Tabs)} />
+          <Route path="/lists/(.*)?" component={withBoundary(Tabs)} />
         </Switch>
       </HashRouter>
     </GlobalContext.Provider>
--- a/mailpoet/lib/API/JSON/ResponseBuilders/DynamicSegmentsResponseBuilder.php
+++ b/mailpoet/lib/API/JSON/ResponseBuilders/DynamicSegmentsResponseBuilder.php
@@ -82,13 +82,19 @@ class DynamicSegmentsResponseBuilder {
     if ($missingPlugins) {
       $missingPlugin = reset($missingPlugins);
       $data['is_plugin_missing'] = true;
-      $data['missing_plugin_message'] = $this->segmentDependencyValidator->getCustomErrorMessage($missingPlugin)
-        ?:
-        sprintf(
-          // translators: %s is the name of the missing plugin.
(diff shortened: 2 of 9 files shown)


mailpoet/lib/API/JSON/ResponseBuilders/DynamicSegmentsResponseBuilder.php
```diff
@@ -82,13 +82,19 @@ private function addMissingPluginProperties(SegmentEntity $segment, array $data)
...
       $data['is_plugin_missing'] = true;
-      $data['missing_plugin_message'] = $this->segmentDependencyValidator->getCustomErrorMessage($missingPlugin)
-        ?:
-        sprintf(
-          // translators: %s is the name of the missing plugin.
-          __('Activate the %s plugin to see the number of subscribers and enable the editing of this segment.', 'mailpoet'),
-          $missingPlugin
-        );
+
+      $missingPluginMessage = $this->segmentDependencyValidator->getCustomErrorMessage($missingPlugin);
```
rodrigoprimo, line 86: I wasn't sure how to tell TypeScript that `missing_plugin_message` (https://github.com/mailpoet/mailpoet/pull/4999/files#diff-6d0f885b48ca9cb6143374ead90096ca7ed547142b27db8480cccd6607746e7bR24) can be a string or an object and that is why I changed the structure on the PHP side.

JanJakes: In any case, it's a nice improvement to unify the response format :+1:

mailpoet/assets/js/src/segments/segments.tsx
```diff
@@ -42,19 +42,19 @@ function Tabs() {
...
-          <Route path="/new-segment" render={withBoundary(Editor)} />
-          <Route path="/edit-segment/:id" render={withBoundary(Editor)} />
-          <Route path="/segments/(.*)?" render={withBoundary(Tabs)} />
-          <Route path="/lists/(.*)?" render={withBoundary(Tabs)} />
+          <Route path="/new" component={withBoundary(SegmentForm)} />
+          <Route path="/edit/:id" component={withBoundary(SegmentForm)} />
+          <Route path="/new-segment" component={withBoundary(Editor)} />
+          <Route path="/edit-segment/:id" component={withBoundary(Editor)} />
+          <Route path="/segments/(.*)?" component={withBoundary(Tabs)} />
+          <Route path="/lists/(.*)?" component={withBoundary(Tabs)} />
```
JanJakes, line 57: Was the change from `render` to `component` necessary? With the `component` prop, a new component is instantiated on every render, which may not be ideal. From https://v5.reactrouter.com/web/api/Route/component:

> When you use component (instead of render or children, below) the router uses [React.createElement](https://facebook.github.io/react/docs/react-api.html#createelement) to create a new [React element](https://facebook.github.io/react/docs/rendering-elements.html) from the given component. That means if you provide an inline function to the component prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component.

JanJakes (commented):
Sorry, for the delay. I was the Porter and in a hurry for the release of new automation features. LGTM, just minor questions.

mailpoet/assets/js/src/segments/segments.tsx
rodrigoprimo: Without this change, I get the following error after converting the file to TSX:

```
TS2769: No overload matches this call.
Overload 1 of 2, '(props: (RouteProps<string, { [x: string]: string; }> & OmitNative<{}, keyof RouteProps<string, { [x: string]: string; }>>) | 
Readonly<RouteProps<string, { ...; }> & OmitNative<...>>): Route<...>', gave the following error.     Type 'ComponentType<SegmentFormPropType>' is not assignable to type '(props: RouteComponentProps<{ [x: string]: string; }, StaticContext, unknown>) => ReactNode'.
Type 'ComponentClass<SegmentFormPropType, any>' is not assignable to type '(props: RouteComponentProps<{ [x: string]: string; }, StaticContext, unknown>) => ReactNode'.
Type 'ComponentClass<SegmentFormPropType, any>' provides no match for the signature '(props: RouteComponentProps<{ [x: string]: string; }, StaticContext, unknown>): ReactNode'.
Overload 2 of 2, '(props: RouteProps<string, { [x: string]: string; }> & OmitNative<{}, keyof RouteProps<string, { [x: string]: string; }>>, context: any): Route<...>', gave the following error.
Type 'ComponentType<SegmentFormPropType>' is not assignable to type '(props: RouteComponentProps<{ [x: string]: string; }, StaticContext, unknown>) => ReactNode'.
Type 'ComponentClass<SegmentFormPropType, any>' is not assignable to type '(props: RouteComponentProps<{ [x: string]: string; }, StaticContext, unknown>) => ReactNode'.

index.d.ts(92, 5): The expected type comes from property 'render' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<Route<{}, string>> & Readonly<RouteProps<string, { [x: string]: string; }> & OmitNative<...>>'
index.d.ts(92, 5): The expected type comes from property 'render' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<Route<{}, string>> & Readonly<RouteProps<string, { [x: string]: string; }> & OmitNative<...>>'
```

I'm not sure if there is a better way to fix this error. I followed what I found in https://github.com/mailpoet/mailpoet/blob/91e333fa6cfffd9a16b3482b55ace103608c8152/mailpoet/assets/js/src/wizard/wizard.tsx#L23

Do you happy to know if there is a better approach? Happy to further investigate this if necessary. I would need to do some research as I don't know off the top of my head.

rodrigoprimo:
> Sorry, for the delay. I was the Porter and in a hurry for the release of new automation features. LGTM, just minor questions.

Thanks for the review, and no need to apologize. I completely forgot that you were the porter last week when I assigned this PR to you.

mailpoet/assets/js/src/segments/segments.tsx
JanJakes: @rodrigoprimo Ahh, `render` expects a function, so I guess what you have now is actually correct, as long as the HOC (`withBoundary`) behaves well. So seems fine 👍

JanJakes (approved):

rodrigoprimo:
Thanks for the additional details, JJ.

Aschepikov (approved):
GJ!

PR #5016, opened 2023-06-23
Automation filter improvements [MAILPOET-5415][MAILPOET-5418][MAILPOET-5426]

JanJakes:
## Description

This fixes and improves various automation issues:
1. Fix can't save `is-blank` and `is-not-blank` filters.
2. Refactor filter validation to work per-condition (`is-blank` expects a different value than `matches regex`, etc.).
3. Make filters editable so that we can prefill them only partly in some templates (e.g. "purchased product ...").
4. Mark filter errors, so users know which ones are not fully set.
5. Avoid calling automation run query on every request (fetch fields lazily).

## Code review notes

The solution for 5 may not be the ultimate one, but it's a quick one. We can think through a more robust approach to executing minimum automation code on every request. I was also considering creating field args via a callback (we can always add that as well).

## QA notes

Please, do a smoke test of automation filters.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/739

## Linked tickets

[MAILPOET-5415]
[MAILPOET-5418]
[MAILPOET-5426]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5415]: https://mailpoet.atlassian.net/browse/MAILPOET-5415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5418]: https://mailpoet.atlassian.net/browse/MAILPOET-5418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5426]: https://mailpoet.atlassian.net/browse/MAILPOET-5426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-automation-editor/_filters.scss
+++ b/mailpoet/assets/css/src/components-automation-editor/_filters.scss
@@ -41,7 +41,13 @@ button.components-button.mailpoet-automation-filters-panel-add-filter {
   display: grid;
   grid-template-columns: 1fr auto;
 
+  &.mailpoet-automation-filters-list-item-has-error {
+    background: #fcf0f1;
+    color: #8a2424;
+  }
+
   .mailpoet-automation-filters-list-item-content {
+    color: inherit;
     display: block;
     height: auto;
     line-height: inherit;
@@ -58,6 +64,12 @@ button.components-button.mailpoet-automation-filters-panel-add-filter {
     margin: 4px;
     padding: 0;
   }
+}
+
+.mailpoet-automation-filters-list-item-error {
+  color: #8a2424;
+  font-size: 12px;
+  margin-top: 4px;
 }
 
 .mailpoet-automation-filters-list-item-field,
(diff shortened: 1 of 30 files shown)


websupporter:
> The solution for 5 may not be the ultimate one, but it's a quick one. We can think through a more robust approach to executing minimum automation code on every request. I was also considering creating field args via a callback (we can always add that as well).

I like the solution for now, as it solves the immediate problem, but still, the current solution will slow down the Automations editor for some users. I agree: Lets get this out and think about a "more robust approach" further down the line :thinking: 

LGTM and thank you for taking this on.

websupporter (approved):
lgtm

Aschepikov (approved):
GJ!

PR #5018, opened 2023-06-23
Refactor Sending::getSubscribers() and Sending::setSubscribers() to use Doctrine instead of Paris [MAILPOET-4368]

rodrigoprimo:
## Description

This PR refactors the methods Sending::getSubscribers() and Sending::setSubscribers() to use Doctrine instead of Paris. The remaining subscriber related methods that exist in Sending will be refactored in another PR. 

## Code review notes

See one inline comment for a specific part that I wanted to highlight. 

## QA notes

This PR should not change any functionality, but it touches sending emails which is a core part of MailPoet. It would be nice to test anything sending related that is not covered by automated tests. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4368]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4368]: https://mailpoet.atlassian.net/browse/MAILPOET-4368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Scheduler/WelcomeScheduler.php
+++ b/mailpoet/lib/Newsletter/Scheduler/WelcomeScheduler.php
@@ -4,6 +4,7 @@ namespace MailPoet\Newsletter\Scheduler;
 
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\NewsletterOptionFieldEntity;
+use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\SubscriberEntity;
@@ -126,6 +127,19 @@ class WelcomeScheduler {
       $newsletter->getOptionValue(NewsletterOptionFieldEntity::NAME_AFTER_TIME_TYPE),
       $newsletter->getOptionValue(NewsletterOptionFieldEntity::NAME_AFTER_TIME_NUMBER)
     );
-    return $sendingTask->save();
+
+    $savedSendingTask = $sendingTask->save();
+
+    // Refreshing this entity here is needed while we are still using Paris to create the scheduled tasks and queues
+    // in the code above using \MailPoet\Tasks\Sending class. Doing this should avoid bugs where the loaded entity contain
+    // stale data after the corresponding entry in the database is updated using Paris. This code can be removed once
+    // https://mailpoet.atlassian.net/browse/MAILPOET-4375 is finished. Currently, if this code is removed a few integration
+    // tests fail (see https://app.circleci.com/pipelines/github/mailpoet/mailpoet/14806/workflows/0d441848-16db-461a-88ec-87bed101fe36/jobs/251385/tests#failed-test-0).
+    $scheduledTaskEntity = $this->scheduledTasksRepository->findOneScheduledByNewsletterAndSubscriber($newsletter, $subscriber);
+    if ($scheduledTaskEntity instanceof ScheduledTaskEntity) {
+      $this->scheduledTasksRepository->refresh($scheduledTaskEntity);
+    }
+
+    return $savedSendingTask;
   }
 }
--- a/mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
+++ b/mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
@@ -125,6 +125,17 @@ class ScheduledTaskSubscribersRepository extends Repository {
       ->execute();
   }
 
+  public function setSubscribers(ScheduledTaskEntity $task, array $subscriberIds): void {
+    $this->deleteByScheduledTask($task);
+
+    foreach ($subscriberIds as $subscriberId) {
+      $this->createOrUpdate([
+        'task_id' => $task->getId(),
+        'subscriber_id' => $subscriberId,
+      ]);
+    }
+  }
+
   private function checkCompleted(ScheduledTaskEntity $task): void {
     $count = $this->countBy(['task' => $task, 'processed' => ScheduledTaskSubscriberEntity::STATUS_UNPROCESSED]);
     if ($count === 0) {
--- a/mailpoet/lib/Tasks/Sending.php
+++ b/mailpoet/lib/Tasks/Sending.php
@@ -4,12 +4,14 @@ namespace MailPoet\Tasks;
 
 use MailPoet\Cron\Workers\SendingQueue\SendingQueue as SendingQueueAlias;
 use MailPoet\DI\ContainerWrapper;
+use MailPoet\Entities\ScheduledTaskSubscriberEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\InvalidStateException;
 use MailPoet\Logging\LoggerFactory;
 use MailPoet\Models\ScheduledTask;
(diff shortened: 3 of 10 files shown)


mailpoet/lib/Tasks/Sending.php
```diff
@@ -208,18 +221,37 @@ public function taskSubscribers() {
...
     }
-    $subscribers = $subscribers->findArray();
-    return array_column($subscribers, 'subscriber_id');
+
+    return array_map(
+      function(ScheduledTaskSubscriberEntity $scheduledTaskSubscriber) {
+        $subscriber = $scheduledTaskSubscriber->getSubscriber();
+
+        if ($subscriber instanceof SubscriberEntity) {
+          return (string)$subscriber->getId();
```
rodrigoprimo: Calling `$subscriber->getId()` to my surprise means that Doctrine is performing an additional query against `wp_mailpoet_subscribers` for every subscriber. This should not be necessary as the subscriber ID is available in `wp_mailpoet_scheduled_task_subscribers`. I was not able to find an easy way to get the subscriber ID using doctrine without making it fetch the full record from `wp_mailpoet_subscribers`. Any ideas on how to achieve that?

I haven't checked but maybe this is happening because of `SafeToOneAssociationLoadTrait` that is used in `ScheduledTaskSubscriberEntity`?

lysyjan: We discussed this when we met in London, and if I remember correctly, the mentioned trait causes this loading of each subscriber entity. The proposed solution can be a getter that doesn't use this trait or avoid using the trait because it's not needed anymore. Is it correct @rodrigoprimo?

rodrigoprimo: @lysyjan, indeed we discussed this in London. I implemented a getter that doesn't use the trait here: https://github.com/mailpoet/mailpoet/pull/5018/commits/4703a106a0fa83b5c596b5b1f4576e202542e91a#diff-27ecf3a634794e8f2882e6d154694bd3dde08a5ba96882ba0d7dc5e82452c576R130. I don't recall anyone mentioning that stop using the trait as it is not needed anymore was a possibility so I didn't consider this option.

I'm just waiting for the CircleCI build to finish to assign this PR for a final review.

mailpoet/lib/Newsletter/Scheduler/WelcomeScheduler.php
```diff
@@ -126,6 +127,19 @@ public function createWelcomeNotificationSendingTask(NewsletterEntity $newslette
...
+    $savedSendingTask = $sendingTask->save();
+
+    // Refreshing this entity here is needed while we are still using Paris to create the scheduled tasks and queues
+    // in the code above using \MailPoet\Tasks\Sending class. Doing this should avoid bugs where the loaded entity contain
+    // stale data after the corresponding entry in the database is updated using Paris. This code can be removed once
+    // https://mailpoet.atlassian.net/browse/MAILPOET-4375 is finished. Currently, if this code is removed a few integration
+    // tests fail (see https://app.circleci.com/pipelines/github/mailpoet/mailpoet/14806/workflows/0d441848-16db-461a-88ec-87bed101fe36/jobs/251385/tests#failed-test-0).
+    $scheduledTaskEntity = $this->scheduledTasksRepository->findOneScheduledByNewsletterAndSubscriber($newsletter, $subscriber);
+    if ($scheduledTaskEntity instanceof ScheduledTaskEntity) {
+      $this->scheduledTasksRepository->refresh($scheduledTaskEntity);
```
rodrigoprimo, line 140: This is my biggest concern with this PR. Since we are in a mixed state where some parts of the code are using Doctrine and others are using Paris, we could run into bugs where Paris models or Doctrine entities contain stale data. As far as I could check, we are only using `setSubscribers()` here (and in two integration tests), so we should be good, but still this is a concern of mine.

rodrigoprimo:
Just documenting here that I had to force push to this PR due to some conflicts with a PR that was merged yesterday (https://github.com/mailpoet/mailpoet/pull/5028).

lysyjan:
@rodrigoprimo Sorry for the late reaction here!

I checked the ticket description in Jira and noticed that the current PR doesn't match it. However, you have a better idea of the future state because this PR should only be the first for this task.

I will explain my expectations here to you. When I planned to replace Paris by Doctrine with @costasovo, we considered removing the whole Sending class or refactoring it to a stateless service as described in [a follow-up task](https://mailpoet.atlassian.net/browse/MAILPOET-4304). The main reason why I wanted to mention it here is to avoid misunderstanding in the future of the Sending class.

For example, a good candidate for avoiding a specific method in the Sending class is `setSubscribers()`, where no complex logic exists. I think creating a method in a repository could also be possible for `getSubscribers()`. But as I mentioned, in the beginning, it all depends on your vision for the whole solution.

rodrigoprimo:
@lysyjan, sorry for the long delay in replying, I'm getting back to this only this week as I'm the janitor.

Thanks for explaining what you and Rosta had in mind when creating the ticket. I had misunderstood the description and I agree that it makes sense to remove `getSubscribers()` and `setSubscribers()` from the `Sending` class.

That being said, I believe this PR is already as step in the right direction as it replaces Paris code with Doctrine and it will likely make it easier to move the methods somewhere. So, my suggestion is that we merge this PR as is and address the pending point in another PR. What do you think?

I rebased this PR since it was fairly old just in case. There were no conflicts so I made no manual changes after rebasing.

rodrigoprimo:
One more thing to note is that I ended up taking a different path than what you probably anticipated but I guess in the end we will reach the same result. In https://github.com/mailpoet/mailpoet/pull/5040/, I removed `MailPoet\Tasks\Subscribers`. Then, as I said above, it will be a matter of deciding what to do with the subscribers methods in the Sending class. I will think more about it.

rodrigoprimo:
Just documenting that I had to force push when rebasing to fix a conflict.

mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
```diff
@@ -125,6 +125,17 @@ public function deleteByScheduledTask(ScheduledTaskEntity $scheduledTask): void
       ->execute();
   }
 
+  public function setSubscribers(ScheduledTaskEntity $task, array $subscriberIds): void {
+    $this->deleteByScheduledTask($task);
```
JanJakes, line 129: Do you expect the delete to be mostly a noop or can it sometimes delete many records that would be then recreated again below? (In which case, it might be better to delete after with `WHERE id NOT IN (...)`.)

JanJakes (approved):
LGTM, just one question, but we can move it forward.

mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
rodrigoprimo: That is a good question and something that I haven't considered. Here, I basically implemented the same behavior that we had in the Paris code: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Models/ScheduledTaskSubscriber.php#L44

So, I guess we can keep it as is and improve in the future if needed.

rodrigoprimo:
Moving to QA

Aschepikov (approved):
GJ! Everything that is sending was tested

PR #5019, opened 2023-06-23
Split Lists page into Lists and Segments [MAILPOET-5392]

rodrigoprimo:
## Description

This PR splits the lists page into two different pages with two separate menu entries: Lists and Segments. Despite this change, the functionality of both pages should remain the same.

## Code review notes

There are a few things that I'm unsure about this PR that I would love to have a second opinion:

- Naming is a bit of a mess regarding this functionality. While in the UI we use lists and segments, in the code we use segments for lists and dynamic segments for segments. In this PR, when refactoring the JS code, I went one step ahead and introduced the notion of static segments to refer to lists. I did this as all the code lives in `mailpoet/assets/js/src/segments/` and now we have `mailpoet/assets/js/src/segments/dynamic` and `mailpoet/assets/js/src/segments/static`.
- I called the main function and file for each new page `app`. I'm not sure if there is a better name. Also, I'm not sure if using a different React app per page is the best approach, but I did this here as this is what we seem to be doing for other pages.

## QA notes

Please test everything in the lists and segments pages that are not covered by automatic tests.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/748

## Linked tickets

[MAILPOET-5392]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5392]: https://mailpoet.atlassian.net/browse/MAILPOET-5392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NeosinneR:
@rodrigoprimo as Jan is AFK for two weeks, I reassigned this PR to @johnolek

rodrigoprimo:
Thanks, @NeosinneR. I completely missed that he is AFK. I saw his AFK status on Slack but since there was no description, I wrongly assumed that he was away just for a few hours or for the day.

mailpoet/lib/AdminPages/Pages/Segments.php
```diff
@@ -78,8 +78,13 @@ public function __construct(
     $this->formsRepository = $formsRepository;
   }
 
-  public function render() {
+  /**
+   * @param string $type 'static' or 'dynamic'
+   * @return void
+   */
+  public function render(string $type) {
```
johnolek: Did you consider creating separate `render` methods for static vs dynamic segments? It seems like we would be loading a lot of extra stuff that the static segments don't require at all.

mailpoet/assets/js/src/webpack_admin_index.tsx
```diff
@@ -7,7 +7,8 @@
 import 'homepage/homepage'; // side effect - renders ReactDOM to document
 import 'subscribers/subscribers.jsx'; // side effect - renders ReactDOM to document
 import 'newsletters/newsletters.jsx'; // side effect - renders ReactDOM to window
-import 'segments/segments'; // side effect - renders ReactDOM to document
+import 'segments/static/app';
```
johnolek: For consistency this should probably include the side effect comment too

johnolek (changes requested):
Overall this looks good to me -- nice work!

The one thing I would prefer to have changed is the render method. I think we should avoid loading all the data we need for segments if we know we're rendering lists.

Regarding the naming, although I like the idea of using "static segments," I don't like the idea of introducing another term without getting rid of the old one. It seems like it could just make it more confusing for someone who wasn't familiar with the naming. Would it be too painful to split it into `dynamic_segments` and `segments` for now (or something similar that uses the current naming), and hopefully we can refactor everything together sometime later?

I think the `app` naming is very clear, but the convention seems to be using the same name as the directory as the filename, like `homepage/homepage` and wizard/wizard`, so I think if we could stick to that convention it would be nice, but not necessary.

mailpoet/lib/AdminPages/Pages/Segments.php
rodrigoprimo: At first, I didn't want to touch this part as we were already loading everything for both pages, and I was afraid splitting was going to be hard and could potentially break things.

I went ahead and created separate `render` methods. One for each page. The only part that I kept together is the translations as splitting them would be more involved and probably not worth it.

That being said, I doubled down on using `static` and `dynamic` segments. More on that in a separate comment.

mailpoet/assets/js/src/webpack_admin_index.tsx
rodrigoprimo: :+1:

rodrigoprimo:
Thanks for the review, @johnolek!

> Regarding the naming, although I like the idea of using "static segments," I don't like the idea of introducing another term without getting rid of the old one. It seems like it could just make it more confusing for someone who wasn't familiar with the naming. Would it be too painful to split it into dynamic_segments and segments for now (or something similar that uses the current naming), and hopefully we can refactor everything together sometime later?

I'm inclined to think that using `static` and `dynamic` to differentiate the types of segments is a bit better than `dynamic_segments` and `segments` and reduces a bit the confusion that is already present instead of making it worst. I say this because at least when you see `static` or `dynamic`, there is no confusion as to which is which. The same can't be said about `dynamic_segments` and `segments`. When seeing just `segments` it is not easy to know if it refers only to the static segments or if it refers to both types of segments.

We are already using `static` and `dynamic` in a few parts of the codebase (but I agree that it is not widely adopted). Here are a few examples:

https://github.com/mailpoet/mailpoet/blob/acbd7fe7235edd143d4f09a47069419f0df3a6a1/mailpoet/lib/AdminPages/Pages/Segments.php#L98
https://github.com/mailpoet/mailpoet/blob/acbd7fe7235edd143d4f09a47069419f0df3a6a1/mailpoet/lib/Segments/SubscribersFinder.php#L92
https://github.com/mailpoet/mailpoet/blob/acbd7fe7235edd143d4f09a47069419f0df3a6a1/mailpoet/assets/js/src/segments/dynamic/store/selectors.ts#L47
https://github.com/mailpoet/mailpoet/blob/acbd7fe7235edd143d4f09a47069419f0df3a6a1/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/subscriber/subscriber_subscribed_to_list.tsx#L14

Happy to chat further about this. I don't see an easy solution. My hope is that using `static` and `dynamic` in this PR makes the code a bit better and not worst, but I can definitely be missing a few things.

johnolek:
@rodrigoprimo Thanks for elaborating about the static terminology, and I'm happy to leave it the way you have it. I didn't even realize/remember that there are other places where we're already using the static term.

I definitely agree that `static` and `dynamic` is clearer than `segments` and `dynamic_segments`, and I wish that we were using those terms everywhere. I think it can be confusing when you have 2 potential ways of referring to the same thing in the codebase, but you're right that in this case it's probably more confusing to continue using an ambiguous term like `segment`.

Maybe it would make for a good maintenance ticket to try to standardize the terminology and make sure we're using the static terminology everywhere.

rodrigoprimo:
@johnolek, I assigned this PR back to you and I believe it is ready for a final review.

I agree that we should create a maintenance ticket to look into the naming problem with segments.

johnolek (dismissed):
Looks good to me. Nice work!

By the way, I updated the description to point to the new premium PR with the matching branch name.

rodrigoprimo:
> By the way, I updated the description to point to the new premium PR with the matching branch name.

Thanks!

Aschepikov:
HI @rodrigoprimo there are some conflicts can you resolve them pls

rodrigoprimo:
@Aschepikov, I just fixed the conflicts. I had to also update a few of the acceptance tests when rebasing as new acceptance tests were introduced to `trunk` using the old page structure.

Also, I should have made it more clear that the idea was to merge https://github.com/mailpoet/mailpoet-premium/pull/748 together with this PR (@veljkho). Since #748 was already merged, the acceptance tests for premium are failing. This shouldn't be a problem if we merge this PR soon, but please let me know if that is not possible and I will probably revert #748,

Aschepikov (approved):
GJ!

PR #5015, opened 2023-06-23
Analytics Orders Tab [PREMIUM-224]

websupporter:
## Description

Adds the orders tab to the analytics page. The filter will be added in a separate ticket

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/738

## Linked tickets

[PREMIUM-224]

## After-merge notes

_N/A_

## Tasks

- [x] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- NA I added sufficient test coverage (extra ticket for acc test)
- [x] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[PREMIUM-224]: https://mailpoet.atlassian.net/browse/PREMIUM-224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5024, opened 2023-06-27
Load analytics data by version [MAILPOET-5433]

websupporter:
## Description
Introduces AutomationEmailController which ensures that the correct emails are returned for an Automation in a given timeframe.

This PR is based of #5015 

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/742

## Linked tickets

[MAILPOET-5433]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5433]: https://mailpoet.atlassian.net/browse/MAILPOET-5433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/emails/rows.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/emails/rows.tsx
@@ -56,7 +56,10 @@ export function transformEmailsToRows(emails: EmailStats[]) {
 
   return emails.map((email) => {
     // Shows the percentage of clicked emails compared to the number of sent emails
-    const clickedPercentage = calculatePercentage(email.clicked, email.sent);
+    const clickedPercentage = calculatePercentage(
+      email.clicked,
+      email.sent.current,
+    );
     const clickedBadge = percentageBadgeCalculation(clickedPercentage);
 
     return [
@@ -79,19 +82,23 @@ export function transformEmailsToRows(emails: EmailStats[]) {
                 <a
                   href={`?page=mailpoet-newsletters#/sending-status/${email.id}`}
                 >
-                  {`${email.sent}`}
+                  {`${email.sent.current}`}
                 </a>
               </Tooltip>
             }
             subValue={
               // Shows the percentage of sent emails compared to the previous email
               percentageFormatter.format(
-                calculatePercentage(email.sent, email.sent, true) / 100,
+                calculatePercentage(
+                  email.sent.current,
+                  email.sent.previous,
+                  true,
+                ) / 100,
               )
             }
           />
         ),
-        value: email.sent,
+        value: email.sent.current,
       },
       {
         display: (
@@ -100,7 +107,7 @@ export function transformEmailsToRows(emails: EmailStats[]) {
             subValue={
               // Shows the percentage of opened emails compared to the number of sent emails
               percentageFormatter.format(
-                calculatePercentage(email.opened, email.sent) / 100,
+                calculatePercentage(email.opened, email.sent.current) / 100,
               )
             }
           />
@@ -112,13 +119,15 @@ export function transformEmailsToRows(emails: EmailStats[]) {
           <Cell
             value={email.clicked}
             className={
-              email.sent > 0
+              email.sent.current > 0
                 ? 'mailpoet-automation-analytics-email-clicked'
                 : ''
             }
             subValue={percentageFormatter.format(clickedPercentage / 100)}
-            badge={email.sent > 0 ? clickedBadge.badge : undefined}
-            badgeType={email.sent > 0 ? clickedBadge.badgeType : undefined}
+            badge={email.sent.current > 0 ? clickedBadge.badge : undefined}
+            badgeType={
+              email.sent.current > 0 ? clickedBadge.badgeType : undefined
+            }
           />
         ),
         value: email.clicked,
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/emails/rows.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/emails/rows.tsx
@@ -56,10 +56,7 @@ export function transformEmailsToRows(emails: EmailStats[]) {
 
   return emails.map((email) => {
     // Shows the percentage of clicked emails compared to the number of sent emails
-    const clickedPercentage = calculatePercentage(
-      email.clicked.current,
-      email.sent.current,
-    );
+    const clickedPercentage = calculatePercentage(email.clicked, email.sent);
     const clickedBadge = percentageBadgeCalculation(clickedPercentage);
 
     return [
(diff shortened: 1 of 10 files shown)


mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/AutomationEmailController.php
```diff
@@ -0,0 +1,93 @@
...
+        break;
+      }
+      if (!$versionIds || $automationVersion['created_at'] < $after) {
+        $versionIds = [(int)$automationVersion['id']];
+        continue;
+      }
+      $versionIds[] = (int)$automationVersion['id'];
+    }
+
+    $automations = $this->automationStorage->getAutomationWithDifferentVersions($versionIds);
```
websupporter: This will become its own function and the name of this class will be renamed to `AutomationTimeSpanController` in the Automation Flow Tab PR.

Not super happy with the name and location of the class.

mailpoet/lib/Automation/Engine/Storage/AutomationStorage.php
```diff
@@ -65,6 +68,34 @@ public function updateAutomation(Automation $automation): void {
     $this->insertAutomationTriggers($automation->getId(), $automation);
   }
 
+  /**
+   * @param int $automationId
+   * @return VersionDate[]
+   * @throws \Exception
+   */
+  public function getAutomationVersionDates(int $automationId): array {
+    $versionsTable = esc_sql($this->versionsTable);
```
alex-mailpoet, line 77: Maybe you need to wrap this WP function and not use it directly? Same applies to a few other `esc_sql` calls in this PR.

mailpoet/lib/Automation/Engine/Storage/AutomationStorage.php
```diff
@@ -96,6 +96,35 @@ function($row): array {
     ) : [];
   }
 
+  /**
+   * @param int[] $versionIds
+   * @return Automation[]
+   */
+  public function getAutomationWithDifferentVersions(array $versionIds): array {
```
alex-mailpoet, line 103: Note: this method allows getting versions of different automations as it doesn't check all versions belong to the same automation (it does only in a test).

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/emails/rows.tsx
```diff
@@ -82,56 +79,49 @@ export function transformEmailsToRows(emails: EmailStats[]) {
...
             }
             subValue={
               // Shows the percentage of sent emails compared to the previous email
               percentageFormatter.format(
-                calculatePercentage(
-                  email.sent.current,
-                  email.sent.previous,
-                  true,
-                ) / 100,
+                calculatePercentage(email.sent, email.sent, true) / 100,
```
alex-mailpoet: Since both values passed to the function are the same, it will always return 100%. Is it intended? Is it meaningful?

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/AutomationEmailController.php
```diff
@@ -0,0 +1,92 @@
...
+      }
+    );
+
+    // filter automations that were created before the after date
+    $versionIds = [];
+    foreach ($automationVersions as $automationVersion) {
+      if ($automationVersion['created_at'] > $before) {
+        break;
+      }
+      if (!$versionIds || $automationVersion['created_at'] < $after) {
```
alex-mailpoet: Could you please explain why do you add a version to the result if it doesn't match the `$after` constraint?

alex-mailpoet (commented):
Looks good overall. I have a few minor notes and questions.

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/AutomationEmailController.php
websupporter: Lets say
```
version_id | created_at
         1 | 2022-01-01
         2 | 2022-01-02
         3 | 2022-01-03
         4 | 2022-02-01
```

I want to get all versions which where active `$after` `2022-01-04`. In that case I want to catch version 3 as it was active the 4th of January although created before.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/emails/rows.tsx
websupporter: Great catch! Thank you. f74056e38 was a bit hasty. In case of `sent` we want to actually compare with the amount of emails sent in the previous period. Addressed in 401ca71fd

mailpoet/lib/Automation/Engine/Storage/AutomationStorage.php
websupporter: This would result in a phpcs error like the following:
```
----------------------------------------------------------------------
 339 | WARNING | Unescaped parameter $automationsTable used in
     |         | $wpdb->query("TRUNCATE {$automationsTable}")
     |         | $automationsTable assigned unsafely at line 338:
     |         |  $automationsTable =
     |         | $this->wordPress->escSql($this->automationsTable)
     |         | (WordPressDotOrg.sniffs.DirectDB.UnescapedDBParameter)
----------------------------------------------------------------------

```

mailpoet/lib/Automation/Engine/Storage/AutomationStorage.php
websupporter: Yes, that's true. Not sure, if that is a problem. But I am currently thinking if `getAutomationsByVersionIds()` is maybe a better name for this method :thinking: 

If you want, I can extend the method like 
```
getAutomationWithDifferentVersions(int $automationId, array $versionIds)
```

mailpoet/lib/Automation/Integrations/MailPoet/Analytics/Controller/AutomationEmailController.php
alex-mailpoet: But the `$automationVersion['created_at'] < $after` condition would also apply to versions 1 and 2 in your example. So it will initialize the `$versionIds` array several times until the correct version is put there. It works, but is a bit obfuscated.

websupporter: Hmm... if the `$automationVersions` array would be ordered reverse, maybe this looks better?
```
foreach ($automationVersions as $automationVersion) {
      if ($automationVersion['created_at'] > $before) {
        continue;
      }
      $versionIds[] = (int)$automationVersion['id'];
      if ($automationVersion['created_at'] < $after) {
      	break;
      }
}
```
What do you think @alex-mailpoet

websupporter: discussed via slack and fixed in 3fc69d05e8608f7eb83ed571fe5d45612bb18b7f

mailpoet/lib/Automation/Engine/Storage/AutomationStorage.php
alex-mailpoet: The new method name you proposed would more accurately describe what the function does, but we can leave it as is, not big of a problem.

alex-mailpoet (dismissed):

Aschepikov:
@websupporter Hi there are some conflicts can you resolve them pls

Aschepikov (approved):
GJ!

PR #5031, opened 2023-06-30
Backfill engagement data [MAILPOET-5404]

johnolek:
## Description

This PR backfills the recently created columns for specific subscriber engagement data.

## Code review notes

I decided to create both a cron worker and a service, because I thought it would be more straightforward to test the service directly, but if you think it would be better to keep everything contained to the cron worker I would be happy to refactor this.

## QA notes

It would be great if you could test this on a site that has a large amount of data, including WooCommerce data, to ensure it doesn't cause issues and also that it completes in a reasonable amount of time.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5404](https://mailpoet.atlassian.net/browse/MAILPOET-5404)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5404]: https://mailpoet.atlassian.net/browse/MAILPOET-5404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Subscribers/EngagementDataBackfiller.php
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * @return SubscriberEntity[]
+   */
+  public function getBatch(int $lastProcessedId = 0, int $batchSize = 100): array {
+    $subscribers = $this->entityManager->createQueryBuilder()
+      ->select('PARTIAL s.{id, email, lastPurchaseAt, lastClickAt, lastOpenAt, lastSendingAt}')
+      ->from(SubscriberEntity::class, 's')
+      ->where('s.id > :lastProcessedId')
+      ->setMaxResults($batchSize)
```
websupporter, line 54: Does it make sense to also order by `s.id`? Or can we rely on "by default it is always ordered by id?

mailpoet/lib/Cron/Workers/BackfillEngagementData.php
```diff
@@ -0,0 +1,48 @@
...
+        break;
+      }
+      $this->engagementDataBackfiller->updateBatch($batch);
+      $lastSubscriberId = $this->engagementDataBackfiller->getLastProcessedSubscriberId();
+    } while (count($batch) === self::BATCH_SIZE);
+
+    $meta['last_subscriber_id'] = $this->engagementDataBackfiller->getLastProcessedSubscriberId();
+    $task->setMeta($meta);
+    $this->scheduledTasksRepository->persist($task);
+    $this->scheduledTasksRepository->flush();
```
websupporter: Should we reschedule even when the backfill has been completed? To me it seems like we would restart with subscriber ID `0` when all have been processed:

The `emty($batch)` breaks the `do {} while` and then `getLastProcessedSubscriberId` will return 0 which will be stored as meta value and we start again.

mailpoet/lib/Subscribers/EngagementDataBackfiller.php
```diff
@@ -0,0 +1,200 @@
...
+      return [];
+    }
+
+    $subscribersTable = $this->filterHelper->getSubscribersTable();
+
+    $query = $this
+      ->entityManager
+      ->getConnection()
+      ->createQueryBuilder()
+      ->select("$subscribersTable.id, MAX(orderStats.date_created) as last_purchase_at")
```
websupporter, line 138: Is `orderStats` the correct alias? I am not too deep into the code but to me it looks like `wooFilterHelper` would use `customer` as default alias.

`wooFilterHelper->applyOrderStatusFilter` seems to return the alias, so it might make sense to use `applyOrderStatusFilter` and use it's return value instead of hardcoding `orderStats`

websupporter: Ah! Now I see `orderStats` is the default alias!

websupporter:
> Is there a better way to handle the initial scheduling of the job than using a migration?

I guess another way would be the initializer? I understand that the Migrator usually does the DB migration by itself and now it just starts a task, but somehow I think I prefer that over using the `Initializer` :thinkingface:

> if you think it would be better to keep everything contained to the cron worker I would be happy to refactor this.

I prefer your approach very much.

mailpoet/lib/Subscribers/EngagementDataBackfiller.php
johnolek: Thanks for catching this! I'll include an explicit orderBy.

mailpoet/lib/Cron/Workers/BackfillEngagementData.php
johnolek: You're right that we could end up saving the last processed ID as `0`, but I don't think that would cause the task to restart, and I think it might be a bit of an edge case.

The `empty` check would indeed break the `do` loop, but `getLastProcessedSubscriberId` would only return `0` if this happens during the first batch it attempts to process during a cron run. If it processed a subscriber from the previous batch, it would still be set in the backfiller's state.

I think what would need to happen is the execution limit would have to be hit immediately after processing the last batch. The next time the task runs, the batch would be empty and `getLastProcessedSubscriberId` would return the default of `0`.

I updated the code in b3a47865e6cabca0c3499cee29727a2b3761f2c7 to account for this case.

As for restarting, I don't think I'm doing any scheduling based off the `last_subscriber_id` in the meta. As long as the `processTaskStrategy` code completes before the execution limit, the task should get marked as complete. If it exceeds the limit, then I think it will [resume during the next run ](https://github.com/mailpoet/mailpoet/blob/MAILPOET-5404%2Fbackfill_engagement_data/mailpoet/lib/Cron/CronHelper.php#L227-L227).

I could definitely be wrong!

mailpoet/lib/Subscribers/EngagementDataBackfiller.php
johnolek: I agree this is confusing, so I added a comment in 33c987db339cc2e5de0184e058f789afa569e834. I think the wooFilterHelper could be better in general, but I didn't feel it was necessary to change it as part of this work.

mailpoet/lib/Cron/Workers/BackfillEngagementData.php
johnolek: You can disregard a lot of my previous comment. I realized I needed to save the meta value during the loop anyway, so I simplified everything in a248d485d381eb3b18b3f69734b478f0786a9943.

johnolek:
@websupporter I ended up switching to using the Populator, which seems to be responsible for other similar things like this: https://github.com/mailpoet/mailpoet/blob/04c75908ae7c1d342dd2b3a8c72ce92670298e70/mailpoet/lib/Config/Populator.php#L629-L634 

I believe this should be ready to re-review. Thanks very much for your initial review!

websupporter (approved):
lgtm. thank you @johnolek

veljkho (approved):
Tested  on 2 sites, one is staging mailpoet com where we have a lot of MP data. Tested also on site with a lot of Woo data.. Played around segmentation and it worked normally.

johnolek:
Thanks @veljkho! Did you happen to see how long it took for the task to complete on the site that had the most subscribers? You should be able to tell by looking at the started/completed times for the `backfill_engagement_data` task in the database.

PR #5034, opened 2023-07-04
Add Automation Flow Tab to Analytics [MAILPOET-5091] 

websupporter:
## Description
This PR adds the Automation flow tab as described in [MAILPOET-5091] .

This PR is based of #5024 

## Code review notes
We utilize now the Automation Editor store in the Analytics app to be able to render the tree by using the `<Automation />` component. This component and the `<Step />` component has now an added property `context`, which can be `edit` and `view`. If in the `view` context, we will not add a callback for the onClick in a `Step`. That's the main reason I added it. It serves also in the filters, where we can now decide whether we want to filter a value based on the context.

I did some house cleaning and moved the `user_roles` from `window` to `getContext()`

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/744

## Linked tickets

[MAILPOET-5091] 

## After-merge notes

Please wait for https://github.com/mailpoet/mailpoet/pull/5024 before you merge

## Tasks

[MAILPOET-5091]: https://mailpoet.atlassian.net/browse/MAILPOET-5091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5091]: https://mailpoet.atlassian.net/browse/MAILPOET-5091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/automation_flow/index.tsx
```diff
@@ -10,6 +10,7 @@ Hooks.addFilter(
     function statisticSeperatorWrapper(previousStepData: StepData) {
       return <StatisticSeparator previousStepId={previousStepData.id} />;
     },
+  20,
```
johnolek: Just curious, why is this needed? Would it make sense to include a comment about this?

mailpoet/lib/Automation/Engine/Storage/AutomationStatisticsStorage.php
```diff
@@ -76,17 +76,19 @@ private function getStatistics(array $automationIds, int $versionId = null): arr
     return array_combine(array_column($results, 'id'), $results) ?: [];
   }
 
-  private function getStatsQuery(array $automationIds, int $versionId = null, string $status = null): string {
+  private function getStatsQuery(array $automationIds, int $versionId = null, \DateTimeImmutable $after = null, \DateTimeImmutable $before = null, string $status = null): string {
     $table = esc_sql($this->table);
     $placeholders = implode(',', array_fill(0, count($automationIds), '%d'));
     $versionCondition = strval($versionId ? $this->wpdb->prepare('AND version_id = %d', $versionId) : '');
     $statusCondition = strval($status ? $this->wpdb->prepare('AND status = %s', $status) : '');
+    $dateCondition = $after !== null && $before !== null ? strval($this->wpdb->prepare('AND created_at BETWEEN %s AND %s', $after->format('Y-m-d H:i:s'), $before->format('Y-m-d H:i:s'))) : '';
```
johnolek, line 84: Minor question: do we not care about queries that are only for `after` or `before`?

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/automation_flow/index.tsx
```diff
@@ -59,5 +61,29 @@ export function AutomationFlow(): JSX.Element {
...
+      {section.data.tree_is_inconsistent && (
+        <div className="mailpoet-automation-editor-automation-notices">
+          <Notice
+            status="warning"
+            isDismissible={false}
+            className="mailpoet-automation-flow-notice"
+          >
+            <p>
+              {__(
+                'In this time period, the automation structure did change and therefore some numbers in the flow chart might not be accurate.',
```
johnolek, line 36: This message doesn't tell me much about what the inaccuracies might be, and doesn't provide any guidance for how to view stats that are accurate. Can we explain what the inaccuracies would be? Maybe we could include in this message a list of the dates of each version, so a user could limit their search to when just a single version was active.

johnolek (changes requested):
@websupporter I left a few minor comments, but this looks good to me as is. I'm afraid I can't give as good of a review as I would like to since it's been so long since I worked in this code, so if there's anything specific you have questions/concerns about please let me know and I can take a second look, or maybe someone else can help out.

I'm only requesting changes because there are some conflicts that need to be resolved

mailpoet/lib/Automation/Engine/Storage/AutomationStatisticsStorage.php
websupporter: In the current implementation, we have always both. What do you think, should I prepare for a situation where we only have `$before` or `$after`?

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/automation_flow/index.tsx
websupporter: Not sure if I understand what you mean. Do you mean `<StatisticSeparator />`? It is this element: 
image

Or do you mean the callback? I think it is a callback so that we can apply the filter once and then produce the separator to connect/"separate" different steps with each other.

Do you have an idea, how to comment that meaningful?

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/automation_flow/index.tsx
websupporter: Created a ticket to discuss this further: [MAILPOET-5461]

websupporter:
> requesting changes because there are some conflicts that need to be resolved

Thank you for your review @johnolek, I solved the conflicts.

mailpoet/lib/Automation/Engine/Storage/AutomationStatisticsStorage.php
johnolek: No need to change it now. That probably wasn't a helpful comment by me, really. It's not like it would be difficult to change this in the future if we want open ended date conditions.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/automation_flow/index.tsx
johnolek: Sorry I wasn't clearer! I was specifically wondering about the priority of the callback. Looking at it again I see why you would want this to come after other hooks have had a chance to fire since you wouldn't want the view version to get replaced by a different callback. Sometimes it can become tricky keeping track of callback priorities, but in this case it's actually pretty straightforward (unless I'm actually misinterpreting it!).

I don't think there's a need to comment this on second glance.

johnolek (dismissed):
Thanks for the replies to my comments and for resolving the conflicts! Unless you have anything else you'd like to change this looks good to me.

Aschepikov:
@websupporter There are some conflicts can you resolve them pls

Aschepikov:
Hi @websupporter when I am checking the flow of the new automation I can see NaN% maybe we should change it to 0, what do you think?
image

Aschepikov:
And when I am setting up the date filter in the past when there was no automation I have a 404 error, I think we should add it to the front end with an error message for the user. 
image

Aschepikov:
Fyi I see the error message on the clean install, it was a cache problem, sorry, just a question about NaN%

websupporter:
@Aschepikov thanks! Updated the code so it shows `0`

Aschepikov (approved):
GJ!

PR #5041, opened 2023-07-07
App and Db level migrations [MAILPOET-5416]

costasovo:
## Description
This PR addresses the problem that migrations that use higher-level services may fail when the DB structure is not up to date. Introducing new DB structure changes can easily break older migrations.

The PR fixes the issue by introducing two migration levels, the `Db` level, and the `App` level. 

A Db-level migration doesn't have access to the DI container or EntityManager. They can use a DBAL connection or `$wpdb`. They run before the App-level migrations.

An App-level migration can access the DI container and use all high-level services from the app. They run after the Db-level migrations.

I also have splat existing migrations into App-level and Dd-level.

## Code review notes
I was thinking about two possible approaches:
1) Implementing the App/Db logic on the Migrator level
2) Implementing the logic on the Repository and Runner level. The repository would ensure that migrations are loaded in the correct order and the runner looks for migrations in both subdirectories.

I started with 1). The idea was that the repository and runner would remain universal and wouldn't know about the division to App and Db levels. But after I learned that we need to use a different template (which is pretty low level) for App and Db migration and that doing the change on the Migrator level also means changes in Logger when running via WP CLI, I tried the second approach, which turned out to be quite easy to implement. The drawback is that it hardcodes that there have to be `App` and `Db` subdirectories in the migrations directory. I'm ok with the solution, but I would like to get another opinion.

## QA notes

Please test updating the plugin from older plugin versions:

1. a version prior to 3.103.0 (in 3.103.0, we introduced the new migration system so I think we should check with a slightly older plugin version that had still the old migration system)
2. a couple of random versions in between recent 4.22.0 and 3.103.0

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5416]

## After-merge notes

Publish a short P2 post announcing the change and informing about the new argument for `migrations:new` command (`./do migrations:new db`or `./do migrations:new app`) – ping @costasovo 

## Tasks

[MAILPOET-5416]: https://mailpoet.atlassian.net/browse/MAILPOET-5416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo:
@JanJakes I rebased the changes and moved newly added migrations to the correct subdirectories, and I also did some manual checks I didn't have time to do in the hacking week. The PR is now ready for review.

mailpoet/lib/Migrator/AppMigrationTemplate.php
```diff
@@ -0,0 +1,22 @@
...
+    /*
+     * TODO: Implement the migration logic here and remove this comment.
+     *
+     * App Level migration are intended for data migrations that use application level services.
+     * The application level services require the DB structure to be up to date so they run after all DB migrations.
+     *
+     * Do not make changes in the DB structure in App Level migrations!
+     *
+     * You can use:
+     *   $this->connection      For SQL queries using Doctrine DBAL.
```
JanJakes: The connection doesn't seem to be used in the template. It's the container instead, so I guess we need to update the comment.

mailpoet/tests/integration/Migrations/Migration_20221028_105818_Test.php
```diff
@@ -7,6 +7,7 @@
 use MailPoet\Entities\DynamicSegmentFilterData;
 use MailPoet\Entities\DynamicSegmentFilterEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Migrations\Db\Migration_20221028_105818;
```
JanJakes: Maybe we could nest (in `Db` and `App`) also the test files...

mailpoet/lib/Migrator/Runner.php
```diff
@@ -26,18 +26,13 @@ public function __construct(
   }
 
   public function runMigration(string $name): void {
```
JanJakes: I wonder if the method should take `$level` argument now as well, or if it's better to rely on the naming.

mailpoet/lib/Migrator/Runner.php
```diff
@@ -26,18 +26,13 @@ public function __construct(
...
-    if (!is_subclass_of($className, DbMigration::class)) {
-      throw MigratorException::migrationClassIsNotASubclassOf($className, DbMigration::class);
-    }
+    $className = $this->getClassName($name);
 
     try {
       $migration = new $className($this->container);
       $this->store->startMigration($name);
+      // PHPStan does not recognize that $migration is instance of DbMigration or AppMigration
+      /** @phpstan-ignore-next-line */
```
JanJakes: Probably slightly better to annotate that with `@var` than the PHPStan ignore.

mailpoet/RoboFile.php
```diff
@@ -447,9 +447,13 @@ public function doctrineGenerateCache() {
...
 
-  public function migrationsNew() {
+  /**
+   * Creates a new migration file. Creates Db level migration. Use --level=App to create an application level migration.
+   * @param $opts array{level: string}
+   */
+  public function migrationsNew($opts = ['level' => \MailPoet\Migrator\Repository::MIGRATIONS_LEVEL_DB]) {
     $generator = new \MailPoet\Migrator\Repository();
-    $result = $generator->create();
+    $result = $generator->create($opts['level']);
```
JanJakes: Maybe we could make that argument case-insensitive.
Also, should we validate it's either `db` or `app`?.

mailpoet/RoboFile.php
```diff
@@ -447,9 +447,13 @@ public function doctrineGenerateCache() {
     $this->say("Validator metadata generated to: $validatorMetadataDir");
   }
 
-  public function migrationsNew() {
+  /**
+   * Creates a new migration file. Creates Db level migration. Use --level=App to create an application level migration.
+   * @param $opts array{level: string}
+   */
+  public function migrationsNew($opts = ['level' => \MailPoet\Migrator\Repository::MIGRATIONS_LEVEL_DB]) {
```
JanJakes: > Creates Db level migration.

I'd rather say that `db` is the default. We could also consider `--db` and `--app` for a simpler CLI, but it's a bit harder to verify.

mailpoet/lib/Migrator/Repository.php
```diff
@@ -78,7 +79,7 @@ private function loadForLevel(string $level) {
     return $migrations;
   }
 
-  private function generateName(): string {
-    return 'Migration_' . gmdate('Ymd_His');
+  private function generateName(string $level): string {
+    return 'Migration_' . gmdate('Ymd_His') . '_' . $level;
```
JanJakes: Is this desirable? I saw migrations being moved from one to another in this PR, while preserving the name... Do we want to preserve that capability? Maybe not important, migrations shouldn't be changed frequently.

mailpoet/lib/Migrator/Repository.php
```diff
@@ -18,23 +18,24 @@ class Repository {
...
+      throw MigratorException::templateFileReadFailed($templateFile);
     }
-
-    $name = $this->generateName();
-    $migration = str_replace('class DbMigrationTemplate ', "class $name ", $template);
-    $path = $this->migrationsDir . "/$name.php";
+    $name = $this->generateName($level);
+    $classDeclaration = str_replace('{level}', $level, 'class {level}MigrationTemplate ');
+    $migration = str_replace($classDeclaration, "class $name ", $template);
+    $path = $this->migrationsDir . "/$level" . "/$name.php";
```
JanJakes: Given that you're using double quotes, I guess it all can be a single string without the two concatenations.

mailpoet/lib/Migrator/Repository.php
```diff
@@ -18,23 +18,24 @@ class Repository {
...
     if (!$template) {
-      throw MigratorException::templateFileReadFailed($this->templateFile);
+      throw MigratorException::templateFileReadFailed($templateFile);
     }
-
-    $name = $this->generateName();
-    $migration = str_replace('class DbMigrationTemplate ', "class $name ", $template);
-    $path = $this->migrationsDir . "/$name.php";
+    $name = $this->generateName($level);
+    $classDeclaration = str_replace('{level}', $level, 'class {level}MigrationTemplate ');
```
JanJakes: Maybe this variable can simply have the same name as the one below? (e.g., the `$migration`?)

mailpoet/tests/unit/Migrator/RepositoryTest.php
```diff
@@ -6,30 +6,22 @@
...
-    $migration = pathinfo($files[0], PATHINFO_FILENAME);
-    $this->assertSame(
-      str_replace("class DbMigrationTemplate", "class $migration", file_get_contents(self::TEMPLATE_FILE) ?: ''),
-      file_get_contents($files[0])
-    );
+  public function testItCreatesAppLevelMigrationFiles(): void {
+    $this->checkItCreatesMigrationFiles(Repository::MIGRATIONS_LEVEL_APP);
   }
 
   public function testItFailsCreatingWithInvalidTemplateFile(): void {
```
JanJakes, line 27: Seeing this, maybe we could add `testItFailsWithInvalidLevel`, but only if you think it's useful...

JanJakes (commented):
Looks great, thanks for working on this! 👏

mailpoet/lib/Migrator/Repository.php
costasovo: I don't think it is possible to move the migration between levels as it is determined by the service used in the migration. I added the name suffix to help us distinguish levels in the DB or in the CLI output.

mailpoet/RoboFile.php
costasovo: I changed it so that now it is not an optional argument but required parameter `./do migrations:new app` or `./do migrations:new db`

costasovo:
@JanJakes Thank you for the detailed review and your suggestions! I added seven new commits (starting [this one](https://github.com/mailpoet/mailpoet/pull/5041/commits/c7df9bdab0cc6c34df7409fb165e45d9aef2f534)) and I'm passing the PR back to you.

mailpoet/lib/Migrator/Runner.php
```diff
@@ -25,8 +25,8 @@ public function __construct(
     $this->namespace = $this->getMigrationsNamespace();
   }
 
-  public function runMigration(string $name): void {
-    $className = $this->getClassName($name);
+  public function runMigration(string $name, ?string $level): void {
```
JanJakes: It doesn't seem like `$level` should be optional... should it? Same for `getClassName`.

mailpoet/lib/Migrator/Migrator.php
```diff
@@ -62,25 +62,27 @@ public function run(Logger $logger = null): void {
    * 1. Db migrations ordered by filename
    * 2. App migrations ordered by filename
    * 3. Unknown migrations (saved in store but not in repository e.g., renamed or deleted)
-   * @return array{name: string, status: string, started_at: string|null, completed_at: string|null, retries: int|null, error: string|null, unknown: bool}[]
+   * @return array{name: string, level: string|null, status: string, started_at: string|null, completed_at: string|null, retries: int|null, error: string|null, unknown: bool}[]
```
JanJakes: Changing this probably also affects types in the [Logger](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Migrator/Logger.php). I wonder if it can be better reused, but not sure. [PHPStan type aliases](https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases) could work for PHPStan, but I don't know if the IDEs support that too, so maybe it's safer to keep it copy-pasted.

mailpoet/RoboFile.php
```diff
@@ -448,12 +448,13 @@ public function doctrineGenerateCache() {
...
-   * Creates a new migration file. Creates Db level migration. Use --level=App to create an application level migration.
-   * @param $opts array{level: string}
+   * Creates a new migration file. Use `migrations:new db` for a Db level migration or `migrations:new app` for App level migration.
+   * @param $level string - db or app
    */
-  public function migrationsNew($opts = ['level' => \MailPoet\Migrator\Repository::MIGRATIONS_LEVEL_DB]) {
+  public function migrationsNew($level) {
     $generator = new \MailPoet\Migrator\Repository();
-    $result = $generator->create($opts['level']);
+    $level = ucfirst(strtolower($level));
```
JanJakes: This is just a super tiny unimportant idea, but maybe I'd internally work with the lowercased slugs (`app` and `db`) and put the `ucfirst` only to the namespace and filename constructing methods. It would then display in outputs and logs also as `db` and `app`, which are the easiest to use on the CLI and it would me more consistent with the `status` field:

Screenshot 2023-07-25 at 10 06 13

mailpoet/lib/Migrator/Repository.php
```diff
@@ -55,10 +55,15 @@ public function create(string $level): array {
    * @return array<array{level: string, name: string}>
    */
   public function loadAll(): array {
-    return array_merge(
+    $migrations = array_merge(
       $this->loadForLevel(self::MIGRATIONS_LEVEL_DB),
       $this->loadForLevel(self::MIGRATIONS_LEVEL_APP)
     );
+    $duplicateNames = array_diff_assoc(array_column($migrations, 'name'), array_unique(array_column($migrations, 'name')));
```
JanJakes: Not important, but `array_column($migrations, 'name')` could be saved to a `$names` var first, since it's called twice.

mailpoet/lib/Migrator/Repository.php
```diff
@@ -58,7 +58,7 @@ public function loadAll(): array {
     );
   }
 
-  private function loadForLevel(string $level) {
+  private function loadForLevel(string $level): array {
```
JanJakes, line 74: This should probably get the same return type annotation as `loadAll`.

JanJakes (commented):
Thanks for the fixes! I have only a few last small ideas.

mailpoet/RoboFile.php
costasovo: I like the idea. I went with the first capital letter so that the constants match directory names, and we name directories with CamelCase. I'll use small letters in the constants and ucfirst for the namespace and directory.

costasovo:
@JanJakes Thanks for the review! I addressed your comments in the last four commits.

mailpoet/RoboFile.php
```diff
@@ -448,12 +448,12 @@ public function doctrineGenerateCache() {
   }
 
   /**
-   * Creates a new migration file. Use `migrations:new db` for a Db level migration or `migrations:new app` for App level migration.
+   * Creates a new migration file. Use `migrations:new db` for a bb level migration or `migrations:new app` for app level migration.
```
JanJakes: Typo `bb` -> `db`

mailpoet/lib/Migrator/MigratorException.php
```diff
@@ -14,7 +14,7 @@ public static function templateFileReadFailed(string $path): self {
 
   public static function invalidMigrationLevel(string $level): self {
     return self::create()->withMessage(
-      sprintf('Migration level "%s" is not supported! Use "App" of "Db".', $level)
+      sprintf('Migration level "%s" is not supported! Use "app" of "db".', $level)
```
JanJakes: Ahh, seeing only now, there's a typo `of` -> `or`.

JanJakes (dismissed):

JanJakes (dismissed):
Thanks for handling this! Great job 👏

Aschepikov (approved):
GJ!

PR #5045, opened 2023-07-10
Refactor code to create a component for the `in the last X days` part of some dynamic segments

rodrigoprimo:
## Description

As a first step to implementing changes to the `in the last X days` part of some dynamic segments (see [MAILPOET-4991]), in this PR, I refactored the code so that all the applicable dynamic segments use the same component for `in the last X days` part.

The affected dynamic segments are:

- `# of opens`
- `# of machine opens`
- `average order value`
- `# of orders`
- `single order value`
- `total spent`

## Code review notes

_N/A_

## QA notes

This PR should not change any user-facing behavior. It is probably a good idea to check that the functionality not covered by automated tests of the dynamic segments mentioned above remains the same.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4991]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4991]: https://mailpoet.atlassian.net/browse/MAILPOET-4991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-4991]: https://mailpoet.atlassian.net/browse/MAILPOET-4991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Migrations/Migration_20230712_180341.php
+++ b/mailpoet/lib/Migrations/Migration_20230712_180341.php
@@ -1,0 +1,53 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Migrations;
+
+use MailPoet\Entities\DynamicSegmentFilterData;
+use MailPoet\Migrator\Migration;
+use MailPoet\Segments\DynamicSegments\DynamicSegmentFilterRepository;
+use MailPoet\Segments\DynamicSegments\Filters\WooCommerceAverageSpent;
+use MailPoet\Segments\DynamicSegments\Filters\WooCommerceNumberOfOrders;
+use MailPoet\Segments\DynamicSegments\Filters\WooCommerceSingleOrderValue;
+use MailPoet\Segments\DynamicSegments\Filters\WooCommerceTotalSpent;
+
+class Migration_20230712_180341 extends Migration {
+  public function run(): void {
+    $dynamicSegmentFilterRepository = $this->container->get(DynamicSegmentFilterRepository::class);
+    $filters = $dynamicSegmentFilterRepository->findBy(
+      [
+        'filterData.action' => [
+          WooCommerceNumberOfOrders::ACTION_NUMBER_OF_ORDERS,
+          WooCommerceTotalSpent::ACTION_TOTAL_SPENT,
+          WooCommerceSingleOrderValue::ACTION_SINGLE_ORDER_VALUE,
+          WooCommerceAverageSpent::ACTION,
+        ],
+      ]
+    );
+
+    foreach ($filters as $filter) {
+      $filterData = $filter->getFilterData();
+      $data = $filter->getFilterData()->getData();
+
+      if (isset($data['number_of_orders_days'])) {
+        $days = $data['number_of_orders_days'];
+      } else if (isset($data['total_spent_days'])) {
+        $days = $data['total_spent_days'];
+      } else if (isset($data['single_order_value_days'])) {
+        $days = $data['single_order_value_days'];
+      } else if (isset($data['average_spent_days'])) {
+        $days = $data['average_spent_days'];
+      }
+
+      $filterType = $filterData->getFilterType();
+      $filterAction = $filterData->getAction();
+
+      if (isset($days) && is_string($filterType) && is_string($filterAction)) {
+        $data['days'] = $days;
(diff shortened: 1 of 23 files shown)


johnolek (changes requested):
This looks good to me for filters going forward, but I think it would break existing filters that have data saved under the old keys like `single_order_value_days`. I think we'll either need to include a migration to update existing filters or have the code still check for the existence of the old key and use it if there isn't a value present for the `days` key.

rodrigoprimo:
Good point! I completely forgot to consider existing filters. I will work on it. Thanks.

rodrigoprimo:
@johnolek, I ended up deciding to implement a migration to update the name of the parameters instead of maintaining the old keys working. I decided to do this when I saw that we had already migrations in the past that changed other parts of the dynamic segments (for example, https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Migrations/Migration_20221028_105818.php#L804 and https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Migrations/Migration_20221028_105818.php#L845).

I opted to preserve the old parameter and copy its value to the new one. This seemed a bit safer as it allows users to rollback to previous versions of MailPoet. The old parameter will be deleted if the user updates the segment.

Let me know what you think of this approach. This PR is ready for another review.

I had to force push to this branch to fix a conflict with trunk.

mailpoet/lib/Migrations/Migration_20230712_180341.php
```diff
@@ -0,0 +1,53 @@
...
+      ]
+    );
+
+    foreach ($filters as $filter) {
+      $filterData = $filter->getFilterData();
+      $data = $filter->getFilterData()->getData();
+
+      if (isset($data['number_of_orders_days'])) {
+        $days = $data['number_of_orders_days'];
+      } else if (isset($data['total_spent_days'])) {
```
johnolek, line 33: Just curious, is there a reason you're preferring `else if` to `elseif` here? It looks like the [WP standard](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#use-elseif-not-else-if) is to use `elseif`, but of course we aren't following WP standards in general.

mailpoet/tests/integration/Migrations/Migration_20230712_180341_Test.php
```diff
@@ -0,0 +1,95 @@
+<?php declare(strict_types = 1);
+
+namespace integration\Migrations;
```
johnolek: As far as I can tell it doesn't matter, but the one other migration test we have uses the `MailPoet\Migrations` namespace. Is there a reason you went with this instead?

johnolek (dismissed):
Thank you for the explanation and for digging up those examples! I really like this approach, especially allowing for rollbacks.

I have a couple of very minor questions, but nothing here needs changing that I can see. Nice work!

Feel free to pass along to QA if you don't plan on making any tweaks.

mailpoet/lib/Migrations/Migration_20230712_180341.php
rodrigoprimo: No reason other than muscle memory. Despite the WP coding standard, I'm more used to `else if` than `elseif`. Happy to change it if you prefer. I'm not sure if it is worth it or not.

mailpoet/tests/integration/Migrations/Migration_20230712_180341_Test.php
rodrigoprimo: My best guess is that this was autogenerated by PHPStorm and I didn't notice. I don't I added this. I just pushed a commit changing the namespace.

mailpoet/lib/Migrations/Migration_20230712_180341.php
johnolek: No preference here, just curious!

johnolek (approved):

Aschepikov (approved):
GJ!

PR #5047, opened 2023-07-11
Upgrade @woocommerce/components [PREMIUM-231]

websupporter:
## Description
Updates the `@woocommerce/components` package from 10.0.0 to 12.0.0

## Code review notes
This PR is a bit away from my comfort zone. Not sure what of this is a hack where a nicer solution is possible.

Upgrading this component involved adding new loaders for SASS and CSS and SVG to our webpack configuration. 8edb488edc8fbe36caa2ca28c296297c25820eaf

Afterwards it would still not compile as it could not resolve some references to `@woocommerce/settings`. Therefore I added the package.

The next problem is that the available `@woocommerce/settings` (the package seems deprecated) does not have `getAdminUrl()` as export, so I prepared a patch, which would export the function again. b9fd331a0df1d2ce28f8610498cad7feb46b1641

I also needed to regenerate the existing patch for `@woocommerce/components` as some CSS files contain SASS code. 47c5bd60b1c5b494ddf4e108971626cde8c7c356

Additionally, I was not able to use `@woocommerce/components/build` anymore, which then resulted in getting the problem again that without using `/build` it would try to get a `Text` component which does not exist and thus produce errors. Therefore I extended the patch. 47c5bd60b1c5b494ddf4e108971626cde8c7c356

Lastly:
I added the package using `pnpm add @woocommerce/components -f` to overwrite the `engines.pnpm` constraint which requires a specific `pnpm` version to install the component (see https://github.com/woocommerce/woocommerce/blob/7fbb12b2743dd63f5b05ce3c72153405102162a4/packages/js/components/package.json#L7-L10). I tried to overwrite this constrain with a `.pnpmfile.cjs` file, but that did not work for me, but simply forcing it worked and in the beginning there was no problem for the CI to install the dependency.

But in our plugin build process it failed for the existing `node` constrain. So I decided to force the installation there as well. 9fd79b80066207822a0f07c76671658121d5b742

## QA notes
WooCommerce components is mainly used in the automation listing page and analytics. This PR needed to include `@woocommerce/settings`, and there might be side effects in the "MailPoet Marketing Opt-in" block 

It might be useful to test also the premium plugin if "everything" still works. I have especially the Automations UI in mind.
## Linked tickets
[PREMIUM-231]

## After-merge notes

_N/A_

## Tasks

[PREMIUM-231]: https://mailpoet.atlassian.net/browse/PREMIUM-231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


websupporter:
@JanJakes No need to review for merge, as we need to update to 12.1.0 (which was released a few days ago), but still, have a look, so we can discuss internally, how to update and problems with the update. [Slack](p1689321134815389-slack-C0313447YBC)

JanJakes:
Closing in favor of https://github.com/mailpoet/mailpoet/pull/5060.

PR #5049, opened 2023-07-12
On or (after|before) segment filters [MAILPOET-5012]

johnolek:
## Description

This PR adds two new options, `on or before` and `on or after`, to the existing date based filters, Subscribed Date and Purchase Date.

## Code review notes

For the new translations, I didn't use `_x` like the other similar translations because my understanding is that isn't a best practice. I'm wondering if I should convert the existing `_x` translations to `__` or if I should just leave them as is (which is what I did) to avoid having to re-translate them.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5012

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/views/segments/translations.html
+++ b/mailpoet/views/segments/translations.html
@@ -137,7 +137,8 @@
   'after': _x('after', 'Meaning: "Subscriber subscribed after April'),
   'onOrBefore': __('on or before'),
   'on': _x('on', 'Meaning: "Subscriber subscribed on a given date"'),
-  'onOrAfter': __('on or after'),  'notOn': _x('not on', 'Meaning: "Subscriber subscribed on a date other than the given date"'),
+  'onOrAfter': __('on or after'),
+  'notOn': _x('not on', 'Meaning: "Subscriber subscribed on a date other than the given date"'),
   'inTheLast': _x('in the last', 'Meaning: "Subscriber subscribed in the last 3 days"'),
   'notInTheLast': _x('not in the last', 'Meaning: "Subscriber subscribed not in the last 3 days"'),
 
--- a/mailpoet/views/segments/translations.html
+++ b/mailpoet/views/segments/translations.html
@@ -135,8 +135,9 @@
 
   'before': _x('before', 'Meaning: "Subscriber subscribed before April"'),
   'after': _x('after', 'Meaning: "Subscriber subscribed after April'),
+  'onOrBefore': __('on or before'),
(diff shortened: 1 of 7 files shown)


mailpoet/views/segments/translations.html
```diff
@@ -135,8 +135,9 @@
 
   'before': _x('before', 'Meaning: "Subscriber subscribed before April"'),
   'after': _x('after', 'Meaning: "Subscriber subscribed after April'),
+  'onOrBefore': __('on or before'),
   'on': _x('on', 'Meaning: "Subscriber subscribed on a given date"'),
-  'notOn': _x('not on', 'Meaning: "Subscriber subscribed on a date other than the given date"'),
+  'onOrAfter': __('on or after'),  'notOn': _x('not on', 'Meaning: "Subscriber subscribed on a date other than the given date"'),
```
rodrigoprimo: Just a detail, but we are missing a line break here.

rodrigoprimo (commented):
Code changes look good to me. I just left a comment with a small formatting detail.

> For the new translations, I didn't use _x like the other similar translations because my understanding is that isn't a best practice. I'm wondering if I should convert the existing _x translations to __ or if I should just leave them as is (which is what I did) to avoid having to re-translate them.

I agree that according to the i18n best practices, we should not use `_x()` for the translations that you added. I'm not sure if converting the existing translations would force the translators to translate the strings again. Maybe something to check with them?

mailpoet/views/segments/translations.html
johnolek: Thanks for catching this!

johnolek:
@rodrigoprimo Thanks for the review!

> I'm not sure if converting the existing translations would force the translators to translate the strings again. Maybe something to check with them?

Looking at a recent discussion in #mailpoet-translation in slack, it sounds like it would require some human interaction to handle translations without context and I think they would like a heads up when we're going to update a lot of them at once. So I'm going to keep things as is for now.

I fixed the formatting issue so this should be ready to go.

rodrigoprimo:
> Looking at a recent discussion in #mailpoet-translation in slack, it sounds like it would require some human interaction to handle translations without context and I think they would like a heads up when we're going to update a lot of them at once. So I'm going to keep things as is for now.

:+1:

rodrigoprimo (approved):

Aschepikov (approved):
GJ!

PR #5050, opened 2023-07-12
New engagement date filters [MAILPOET-4989]

johnolek:
## Description

These filters rely on the new operators added in https://github.com/mailpoet/mailpoet/pull/5049, so that should be merged first, and then I will update the base branch for this PR. Feel free to ignore this PR until that one is merged.

This PR adds new dynamic segment filters for the engagement data we're now tracking on the subscriber table. This includes:

- last engagement date
- last purchase date
- last open date
- last click date
- last page view date
- last sending date

## Code review notes

Since these filters should all operate the same as `SubscriberSubscribedDate`, I removed that class in favor of a more general `SubscriberDateField`.

For the tests, I opted for using different fields for the different operators. I could have written tests for every combination of action/operator, but that seemed excessive, and I couldn't immediately think of an elegant way to handle it with a `dataProvider` method. I would welcome other ideas for how to handle the tests.

## QA notes

Please make sure that you have a segment with a  "subscribed date" filter before you start testing and ensure it continues to work the same after installing this version.

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/5049 - Must be merged before this one
https://github.com/mailpoet/mailpoet-premium/pull/749 - should be merged after this one

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-4989

## After-merge notes

_N/A_

## Tasks


johnolek:
@rodrigoprimo This PR is dependent on #5049 (based on its branch), so if you'd prefer to wait until that's merged to review this feel free to unassign yourself, but I don't suspect there will be changes to the other PR so I think this should be ok to review before the merge if you have time.

mailpoet/lib/Segments/DynamicSegments/Filters/SubscriberDateField.php
```diff
@@ -12,13 +12,15 @@ class SubscriberDateField implements Filter {
   const LAST_PURCHASE_DATE = 'lastPurchaseDate';
   const LAST_OPEN_DATE = 'lastOpenDate';
   const LAST_PAGE_VIEW_DATE = 'lastPageViewDate';
+  const SUBSCRIBED_DATE = 'subscribedDate';
```
lysyjan, line 18: Maybe we can comment on this constant that it has a different pattern name due to backward compatibility. What do you think

mailpoet/lib/Segments/DynamicSegments/Filters/DateFilterHelper.php
```diff
@@ -16,14 +16,14 @@ class DateFilterHelper {
   const IN_THE_LAST = 'inTheLast';
   const NOT_IN_THE_LAST = 'notInTheLast';
 
-  public function getValidOperators(): array {
+  public static function getValidOperators(): array {
```
lysyjan: Can I ask why did you change the non-static public method to static?
I'm asking because `DateFilterHelper` is a service that should be injected via DI, and in my opinion, those services shouldn't contain static methods.

mailpoet/tests/integration/Segments/DynamicSegments/Filters/SubscriberDateFieldTest.php
```diff
@@ -0,0 +1,205 @@
...
+      ->withLastPageViewAt(new CarbonImmutable('2023-07-14'))
+      ->create();
+    (new Subscriber())
+      ->withEmail('5@example.com')
+      ->withLastPageViewAt(new CarbonImmutable('2023-07-15'))
+      ->create();
+    $this->assertFilterReturnsEmails('lastPageViewDate', 'onOrAfter', '2023-07-13', ['3@example.com','4@example.com', '5@example.com']);
+  }
+
+  public function testItWorksForNotOn(): void {
```
lysyjan, line 132: > For the tests, I opted for using different fields for the different operators. I could have written tests for every combination of action/operator, but that seemed excessive, and I couldn't immediately think of an elegant way to handle it with a dataProvider method. I would welcome other ideas for how to handle the tests.

I thought about it and your approach is something I would do. I was also considering something more complex, but it could be very complex, which means easier bugs in the code.

lysyjan (commented):
Great job! I like your solution.

I have a question about the change to the static method and a small comment.

mailpoet/lib/Segments/DynamicSegments/Filters/DateFilterHelper.php
johnolek: Thank you for bringing this up. In the back of my mind I've been thinking about updating a lot of the validation code in `FilterDataMapper` to rely on the filter classes themselves, because we end up having to perform most of the same checks in both places. I imagined the constructor growing very large if we needed to inject every single filter that performs any validation, which is why the static method seemed nicer to me.

That being said, I totally agree with you and I will change this back to a public method and inject it into FilterDataMapper.

mailpoet/lib/Segments/DynamicSegments/Filters/SubscriberDateField.php
johnolek: Good idea! I've added it in 169bf537c3ef2044366cc324e846f44df12d7e15

mailpoet/lib/Segments/DynamicSegments/Filters/DateFilterHelper.php
johnolek: I updated the code in 823fdb331d31a7d6aec7635e144fdd7a14cd61ac

For the constructor of `FilterDataMapper`, would it be better to inject the container wrapper instead, or is it ok how I did it?

lysyjan: @johnolek I guess we should avoid using `ContainerWrapper::getInstance()` when it's not necessary or is a good reason for that. Do you think that there is a reason for it? I remember we use it sometimes in temporary code, especially in places that should be removed after finishing Doctrine refactoring.

johnolek: No good reason. I’ve seen it in enough places that I thought it might be idiomatic in some cases. I’ll update this to use injection as well.

johnolek:
@lysyjan I pushed some cleanup commits and I think this should be ready to go whenever you have a chance

lysyjan (approved):
Great job!
Thank you for your patience and cooperation 👍

Aschepikov (approved):
GJ!

PR #5052, opened 2023-07-13
Fix failing integration tests

johnolek:
## Description

Failing tests example: https://app.circleci.com/pipelines/github/mailpoet/mailpoet/14947/workflows/1b6f063d-f38e-4db2-9961-f1f64b8f4e86/jobs/253570

The tests seemed to be failing because orders were being created without
 a billing ID, so a random md5 email address was being inserted into the
  customer lookup table instead of the actual email address, causing the
   queries to return nothing. In addition to fixing that, I added some
   cleanup to ensure that these tests don't leave behind any WC data.

It's unclear to me why these tests were only failing on multisite. In order to check this fix, I used an SSH session on circleCI, checkout out the branch, and ran the full integration suite with multisite enabled, and those tests passed, so I'm pretty confident this will fix the issue.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_


rodrigoprimo (approved):
Looks good to me. I did not investigate why the tests were only failing on multisite.

PR #5057, opened 2023-07-18
Add lifetime option to last days component [MAILPOET-4991]

johnolek:
## Description

This PR changes the dynamic segment filter text  `in the last` for all filters that use the 'in the last X days' component to a select that allows a user to select either `in the last` or `over all time`. If `over all time` is selected, the days input disappears, and the filter should include subscribers regardless of the dates involved.

This affects the following filters:

- number of opens
- number of machine opens
- average order value
- number of orders
- single order value
- total spent

## Code review notes

_N/A_

## QA notes

It's very important that this change doesn't affect any current filters. It should be fully backwards compatible, so please make sure that filters created prior to installing this version of the plugin continue to work exactly as they did before.

## Linked PRs

~This is based on the branch https://github.com/mailpoet/mailpoet/pull/5045 and should not be merged until after that one.~

Update: the blocker PR has been merged.

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-4991

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/days_period_field.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/days_period_field.tsx
@@ -8,6 +8,7 @@ import {
   Timeframes,
 } from 'segments/dynamic/types';
 import { storeName } from 'segments/dynamic/store';
+import { useEffect } from 'react';
 import { isInEnum } from '../../../../utils';
 
 function replaceElementsInDaysSentence(
@@ -26,12 +27,14 @@ export function DaysPeriodField({ filterIndex }: FilterProps): JSX.Element {
   const { updateSegmentFilterFromEvent, updateSegmentFilter } =
     useDispatch(storeName);
 
-  if (!isInEnum(segment.timeframe, Timeframes)) {
-    void updateSegmentFilter(
-      { timeframe: Timeframes.IN_THE_LAST },
-      filterIndex,
-    );
-  }
+  useEffect(() => {
+    if (!isInEnum(segment.timeframe, Timeframes)) {
+      void updateSegmentFilter(
+        { timeframe: Timeframes.IN_THE_LAST },
+        filterIndex,
+      );
+    }
+  }, [segment, updateSegmentFilter, filterIndex]);
 
   const isInTheLast = segment.timeframe === Timeframes.IN_THE_LAST;
 
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/days_period_field.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/days_period_field.tsx
@@ -1,6 +1,7 @@
 import { useDispatch, useSelect } from '@wordpress/data';
 import { Input } from 'common';
 import { MailPoet } from 'mailpoet';
+import { Select } from 'common/form/select/select';
 import { DaysPeriodItem, FilterProps } from 'segments/dynamic/types';
 import { storeName } from 'segments/dynamic/store';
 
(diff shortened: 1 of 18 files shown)


mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/days_period_field.tsx
```diff
@@ -17,12 +18,19 @@ export function DaysPeriodField({ filterIndex }: FilterProps): JSX.Element {
     (select) => select(storeName).getSegmentFilter(filterIndex),
     [filterIndex],
   );
-  const { updateSegmentFilterFromEvent } = useDispatch(storeName);
+  const { updateSegmentFilterFromEvent, updateSegmentFilter } =
+    useDispatch(storeName);
+
+  if (!['inTheLast', 'allTime'].includes(segment.timeframe)) {
```
costasovo: Maybe we could define constants for these values ('inTheLast', 'allTime') the same way as we [do for email action types](https://github.com/mailpoet/mailpoet/blob/c0df921998a91d647b17c85bcc7476dbdef38f43/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/email.tsx#L21-L32) so that we lower a chance of an error due a typo.

costasovo (changes requested):
Nice job! I added one comment with a small suggestion.

When I tested it locally, I noticed a JS warning in the console. The warning is not present in the trunk and appears in the new segment form after selecting one of the types updated in this PR. I didn't dig into it, but I think we should rather address it.

Screenshot 2023-07-24 at 10 32 39

mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/days_period_field.tsx
johnolek: Thank you for this comment! I knew I should have done this from the get go. I extracted these to a type in
c61d454118b5e6019ff193701f3282b957843ed5.

I also ended up creating a helper function to check if a value is included in an enum. Without this I think I would have had to say `Object.values(TImeframes).includes(segment.timeframe as Timeframe)` which doesn't seem meaningful because we don't know if it is a Timeframe. Let me know if that looks ok to you and if there's a better place for it.

johnolek:
Thanks for the review, @costasovo! And good catch about the JS error. It looks like we should have been using `useEffect`. I've updated that in 64740123f379d1d96e0be4e1cbbeba2482fdd712.

costasovo (dismissed):
LGTM! Thank you for the helper function!

Aschepikov:
Hi @johnolek I tested it on one of my sites but when I resaved the old segment I saw the critical error, I will recheck it on another site with old segments and without them, if you have time I will give you creds from this site to check it.
image
image
image

Aschepikov:
Tested without the premium plugin and it works ok. I will try to reproduce this error

johnolek:
@Aschepikov I would be happy to check out the site myself, or if you can tell me how you had the segment configured previously I bet I could reproduce the issue locally, which would be easier for me to troubleshoot.

johnolek:
@Aschepikov I haven't been able to reproduce the error locally. I created 6 different segments on `trunk`, one for each of the affected filters, and when I switch to this branch and try to re-save they all worked ok. So please do let me know if you can figure out more specific steps to reproduce.

And by the way, I force pushed to resolve a merge conflict.

Aschepikov:
@johnolek  Thanks for resolving the conflicts. A few qawp sites were down today and alex too, so maybe it was a coincidence that the site was down at the moment of testing, now the backup was restored so I will retest it on this site

Aschepikov (approved):
GJ!

PR #5059, opened 2023-07-19
Add notice when premium features for key are available [MAILPOET-5429]

lysyjan:
## Description

To improve our user experience, this pull request adds a new notice that should be displayed only when the customer's key supports the premium feature and the premium plugin is not installed or downloaded.

## Code review notes

_N/A_

## QA notes

1. Set a key supporting premium features
  a. Check if the new notice is visible when the premium plugin is not installed. The message should contain a download link. As a part of it, the exact text should be visible in the premium banner, which is visible on the newsletter stats page.
  b. Check if the new notice is visible when the premium plugin is not activated. The message should contain an activation link. As a part of it, the exact text should be visible in the premium banner, which is visible on the newsletter stats page.
2. Use a key without premium features; the new notices shouldn't be visible

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5429]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5429]: https://mailpoet.atlassian.net/browse/MAILPOET-5429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/newsletters/campaign_stats/premium_banner.jsx
+++ b/mailpoet/assets/js/src/newsletters/campaign_stats/premium_banner.jsx
@@ -5,33 +5,69 @@ import { PremiumRequired } from 'common/premium_required/premium_required';
 import { withBoundary } from '../../common';
 
 function SkipDisplayingDetailedStats() {
-  const ctaButton = (
-    <Button
-      href={MailPoet.MailPoetComUrlFactory.getPurchasePlanUrl(
-        MailPoet.subscribersCount,
-        MailPoet.currentWpUserEmail,
-        'starter',
-        { utm_medium: 'stats', utm_campaign: 'signup' },
-      )}
-      target="_blank"
-      rel="noopener noreferrer"
-    >
-      {__('Upgrade', 'mailpoet')}
-    </Button>
-  );
+  let ctaButton;
+  let description;
 
-  const description = (
-    <p>
-      {__(
-        'Learn more about your subscribers and optimize your campaigns. See who opened your emails, which links they clicked, and then use the data to make your emails even better. And if you run a WooCommerce store, you’ll also see the revenue earned per email. All starting $10 per month.',
-        'mailpoet',
-      )}{' '}
-      <a href="admin.php?page=mailpoet-upgrade">
-        {__('Learn more', 'mailpoet')}
-      </a>
-      .
-    </p>
-  );
+  if (
+    MailPoet.hasValidPremiumKey &&
+    (!MailPoet.isPremiumPluginInstalled || !MailPoet.premiumActive)
+  ) {
+    description = (
+      <p>
+        {__(
+          'Your current MailPoet plan includes advanced features, but they require the MailPoet Premium plugin to be installed and activated.',
+          'mailpoet',
+        )}
+      </p>
+    );
+    ctaButton = (
+      <Button
+        href={MailPoet.premiumPluginActivationUrl}
+        rel="noopener noreferrer"
+      >
+        {__('Activate MailPoet Premium plugin', 'mailpoet')}
+      </Button>
+    );
+    // If the premium plugin is not installed, we need to provide a download link
+    if (!MailPoet.isPremiumPluginInstalled) {
+      ctaButton = (
+        <Button
+          href={MailPoet.premiumPluginDownloadUrl}
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          {__('Download MailPoet Premium plugin', 'mailpoet')}
+        </Button>
+      );
+    }
+  } else {
+    description = (
+      <p>
+        {__(
+          'Learn more about your subscribers and optimize your campaigns. See who opened your emails, which links they clicked, and then use the data to make your emails even better. And if you run a WooCommerce store, you’ll also see the revenue earned per email. All starting $10 per month.',
+          'mailpoet',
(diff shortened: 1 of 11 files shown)


mailpoet/lib/Util/Notices/PremiumFeaturesAvailableNotice.php
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      $noticeString .= ' [link]' . __('Activate MailPoet Premium plugin', 'mailpoet') . '[/link]';
+      $link = $this->premiumInstaller->generatePluginActivationUrl(Installer::PREMIUM_PLUGIN_PATH);
+      $attributes = [];
+    }
+
+    $noticeString = Helpers::replaceLinkTags($noticeString, $link, $attributes);
+    $extraClasses = 'mailpoet-dismissible-notice is-dismissible';
+
+    return Notice::displayInfo($noticeString, $extraClasses, self::OPTION_NAME);
```
johnolek: The specs say to `display a green notice`, so should we use `Notice::displaySuccess` here?

mailpoet/assets/js/src/newsletters/campaign_stats/premium_banner.jsx
```diff
@@ -5,33 +5,69 @@ import { PremiumRequired } from 'common/premium_required/premium_required';
...
+      ctaButton = (
+        <Button
+          href={MailPoet.premiumPluginDownloadUrl}
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          {__('Download MailPoet Premium plugin', 'mailpoet')}
+        </Button>
+      );
+    }
```
johnolek: This looks good to me as is, but I was wondering what you think about turning it into an ajax request instead so the user doesn't have to hit the back button themselves. ChatGPT helped me come up with this solution, which could definitely be improved but seems to work:

```js
function SkipDisplayingDetailedStats() {
  let ctaButton;
  let description;

  const [loading, setLoading] = useState(false);

  if (
    MailPoet.hasValidPremiumKey &&
    (!MailPoet.isPremiumPluginInstalled || !MailPoet.premiumActive)
  ) {
    description = (
      <p>
        {__(
          'Your current MailPoet plan includes advanced features, but they require the MailPoet Premium plugin to be installed and activated.',
          'mailpoet',
        )}
      </p>
    );
    ctaButton = (
      <Button
        withSpinner={loading}
        href={MailPoet.premiumPluginActivationUrl}
        rel="noopener noreferrer"
        onClick={(e) => {
          e.preventDefault();
          setLoading(true);

          jQuery.get(MailPoet.premiumPluginActivationUrl, (response) => {
            if (response.includes('Plugin activated')) {
              window.location.reload();
            } else {
              setLoading(false);
              console.error('Error activating plugin:', response);
            }
          });
        }}
        disabled={loading}
      >
        {loading
          ? __('Activating MailPoet premium...', 'mailpoet')
          : __('Activate MailPoet Premium plugin', 'mailpoet')}
      </Button>
    );
```

I'm totally happy to leave it as is though!

johnolek (dismissed):
Looks good to me! I had one idea for your consideration and also a question about the notice style. Neither should be considered required changes.

mailpoet/assets/js/src/newsletters/campaign_stats/premium_banner.jsx
lysyjan: @johnolek I really like your suggestion 👍 I will change it.

lysyjan:
@johnolek I addressed your notes in the last two commits. Also, I rebased this branch on the latest trunk, I you are okay with that.
Because `console.error` is not allowed by our ESLint configuration, the error state of the Ajax call is not handled, and I didn't have any idea how can I do it without significant changes. Of course, I'm all ears if you have some ideas.

johnolek:
@lysyjan Thanks for the updates!

For the error case I would be happy if we just include some kind of user facing message, the easiest way probably being `MailPoet.Notice.error`. Something along the lines of `We were unable to activate the premium plugin, please try visiting the [plugin page link] to activate it manually.`

johnolek (dismissed):
Looks good to me! And thanks for being open to the different idea

Aschepikov:
HI @lysyjan with a free key on the segment page when I am adding the second filter I can see the message "download mailpoet premium plugin" 
image
on other premium features, there are upgrade messages 
image
Can you check it please

lysyjan:
@Aschepikov Nice catch. This issue was separate from my changes. Anyway, I fixed this behavior on my PR.

@johnolek Please, take a look at the two latest commits.
- https://github.com/mailpoet/mailpoet/pull/5059/commits/3987ebb05aef69cc3d527b930c8ff337d0a176f3 This commit fixes an issue found by @Aschepikov. I decided to revert my changes and moved them to a higher level, where a part of this logic already was. The code that we discussed in this PR (activation with Ajax call).
- https://github.com/mailpoet/mailpoet/pull/5059/commits/8d93f218a46a9f201ee9c0720b7698fc0f5e006b The second commit unifies the behavior with the Segment page.

johnolek (approved):
The latest changes look good to me 👍

Aschepikov (approved):
GJ!

PR #5064, opened 2023-07-20
Number of reviews filter [MAILPOET-5413]

johnolek:
## Description

WooCommerce number of reviews filter for dynamic segments.

## Code review notes

_N/A_

## QA notes

This includes a fix for the validation for the number of days field, which is used in a number of filters. I mistakenly broke that logic in this commit: https://github.com/mailpoet/mailpoet/commit/8ab1df5ffa56199a1f4380dcbed9655710ef8235
Without this fix, filters will appear valid if they're set to "in the last" even if there isn't a number of days specified.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5413](https://mailpoet.atlassian.net/browse/MAILPOET-5413)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5413]: https://mailpoet.atlassian.net/browse/MAILPOET-5413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/days_period_field.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/days_period_field.tsx
@@ -2,11 +2,7 @@ import { useDispatch, useSelect } from '@wordpress/data';
 import { Input } from 'common';
 import { MailPoet } from 'mailpoet';
 import { Select } from 'common/form/select/select';
-import {
-  DaysPeriodItem,
-  FilterProps,
-  Timeframes,
-} from 'segments/dynamic/types';
+import { DaysPeriodItem, FilterProps, Timeframe } from 'segments/dynamic/types';
 import { storeName } from 'segments/dynamic/store';
 import { useEffect } from 'react';
 import { isInEnum } from '../../../../utils';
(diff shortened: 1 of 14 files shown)


triple0t (changes requested):
Hi @johnolek 
I'm getting this error in the console.
Screenshot 2023-08-03 at 7 44 43 AM

Full log
[mp3.localhost-1691045187137.log](https://github.com/mailpoet/mailpoet/files/12247708/mp3.localhost-1691045187137.log)

The filter wasn’t working... probably due to the error.

Can you take a look?

johnolek:
@triple0t I apologize for not testing this better myself before requesting your review!

I'm guessing that the main issue might have actually been an oversight on my part where I wasn't setting the default `rating` when adding a new filter in the builder. 

The warning shouldn't have been preventing the filter from working, but you would have needed to switch from the `Any` rating to something else before validation would have worked, because as far as React was concerned the filter had no `rating` value.

I fixed these issues in 89df118a2b3dcd55c3dbea6171eb1d79b1036a99, and I also decided to extract some of the strings into enums just as a general best practice.

I think this should be truly ready for review now. Sorry again!

triple0t:
> I apologize for not testing this better myself before requesting your review!
> 
> I'm guessing that the main issue might have actually been an oversight on my part where I wasn't setting the default `rating` when adding a new filter in the builder.
> 
> The warning shouldn't have been preventing the filter from working, but you would have needed to switch from the `Any` rating to something else before validation would have worked, because as far as React was concerned the filter had no `rating` value.
> 
> I fixed these issues in [89df118](https://github.com/mailpoet/mailpoet/commit/89df118a2b3dcd55c3dbea6171eb1d79b1036a99), and I also decided to extract some of the strings into enums just as a general best practice.
> 
> I think this should be truly ready for review now. Sorry again!

Not a problem, John. 

Thank you for fixing the issue.

My local environmental might have issues. I'm unable to Add ratings to products.

I'm approving this PR; the code looks good. 👍

triple0t (approved):

johnolek:
Thanks for the review, @triple0t.

FYI QA team, I just added the `priority in QA` label because there are a couple other PRs now that depend on some changes made in this one.

Aschepikov (approved):
GJ!

PR #5077, opened 2023-07-27
Update NPM packages [MAILPOET-4391]

johnolek:
## Description

This PR updates all of the NPM dependencies that were able to be updated without causing any obvious issues or requiring refactoring.

The last time we bulk updated NPM dependencies was, I believe, https://github.com/mailpoet/mailpoet/pull/4033

This does not include any `@wordpress/*` packages, as we want to update those separately.

These packages will be done later, as they either caused a problem or would require some refactoring:

@babel/preset-env 
babel-loader 
backbone 
backbone.marionette 
codemirror 
history 
react-router-dom 
react-string-replace 
react-tooltip 
stylelint 
tinymce 

## Code review notes
I performed updates in 2 passes, the first time respecting the `package.json` constraints and the second time with `pnpm up -L`, attempting to update to the latest version. This is why some packages were updated twice.

I committed each update separately because I wanted to be able to find the source of problems relatively quickly using `git bisect`.

I ended up wanting to run a handful of commands every time I updated a package to make sure nothing broke, so I ended up writing a script (thanks ChatGPT) to automate the process. The script isn't perfect, but I think it or something like it could be helpful going forward. Maybe if we were using dependabot or something it would take care of all that for us. I included the script here just in case.

I know we handle wordpress packages separately, but I wasn't sure if it would be a problem to update the `@types` packages, so those may have been improperly updated. Should those always stay in sync?

## QA notes

I'm not 100% sure what the standard procedure is for updates like this. Anything that uses JavaScript could potentially be affected. We should check that there are no new console warnings/errors or unexpected behavior with any JS related functionality.

## Linked PRs

_N/A_

## Linked tickets
https://mailpoet.atlassian.net/browse/MAILPOET-4391

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -28,6 +28,13 @@ class RoboFile extends \Robo\Tasks {
       ->run();
   }
 
+  public function installJs() {
+    return $this->taskExecStack()
+      ->stopOnFail()
+      ->exec('cd .. && pnpm install --frozen-lockfile --prefer-offline')
+      ->run();
+  }
+
   public function cleanupCachedFiles() {
     $this->say('Cleaning up generated folder.');
     $this->_exec('rm -rf ' . __DIR__ . '/generated/*');
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/index.tsx
@@ -17,7 +17,7 @@ export function Tabs(): JSX.Element {
     hasEmails: s(editorStoreName).automationHasStep('mailpoet:send-email'),
   }));
   const pageParams = useMemo(
-    () => new URLSearchParams(location.search),
+    () => new URLSearchParams(location.search as string),
     [location],
   );
   const currentTab = pageParams.get('tab') ?? 'automation-flow';
--- a/mailpoet/assets/js/src/subscribers/importExport/import/step_clean_list.tsx
+++ b/mailpoet/assets/js/src/subscribers/importExport/import/step_clean_list.tsx
@@ -3,7 +3,11 @@ import { CleanList } from './clean_list';
 
 function StepCleanList({ history }: RouteComponentProps): JSX.Element {
   return (
-    <CleanList onProceed={(): void => history.push('step_method_selection')} />
+    <CleanList
+      onProceed={(): void => {
+        history.push('step_method_selection');
+      }}
+    />
   );
 }
 
--- a/mailpoet/package.json
+++ b/mailpoet/package.json
@@ -125,7 +125,7 @@
     "@babel/register": "^7.22.5",
     "@mailpoet/eslint-config": "workspace:^",
     "@storybook/addon-actions": "^7.1.1",
-    "@storybook/addon-links": "^6.5.16",
+    "@storybook/addon-links": "^7.1.1",
     "@storybook/addon-storysource": "^6.5.16",
     "@storybook/addons": "^6.5.16",
     "@storybook/builder-webpack5": "^6.5.16",
--- a/mailpoet/package.json
+++ b/mailpoet/package.json
@@ -137,7 +137,7 @@
     "@types/js-cookie": "^3.0.3",
     "@types/lodash": "^4.14.196",
     "@types/mocha": "^9.1.1",
-    "@types/node": "^18.6.1",
+    "@types/node": "^18.17.1",
     "@types/react": "^18.0.28",
     "@types/react-datepicker": "^4.3.4",
     "@types/react-dom": "^18.0.11",
--- a/mailpoet/package.json
+++ b/mailpoet/package.json
@@ -192,7 +192,7 @@
     "terser-webpack-plugin": "^5.3.9",
     "ts-loader": "^9.4.4",
     "url": "^0.11.0",
-    "webpack": "^5.70.0",
+    "webpack": "^5.88.2",
     "webpack-cli": "^4.9.2",
     "webpack-manifest-plugin": "^5.0.0",
     "wp-types": "^3.59.1"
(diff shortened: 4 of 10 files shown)


mailpoet/RoboFile.php
```diff
@@ -28,6 +28,13 @@ public function install() {
       ->run();
   }
 
+  public function installJs() {
+    return $this->taskExecStack()
+      ->stopOnFail()
+      ->exec('cd .. && pnpm install --frozen-lockfile --prefer-offline')
+      ->run();
+  }
```
JanJakes, line 36: This is a minor observation, but I guess now this could be reused in the install function above (e.g., `->addCode(...)`).

mailpoet/package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "@babel/runtime": "^7.22.6",
     "@babel/runtime-corejs3": "^7.22.6",
-    "@emotion/react": "^11.8.2",
+    "@emotion/react": "^11.11.1",
```
JanJakes, line 32: Note to self to check later — why do we even install `@emotion` packages explicitly? 🤔

mailpoet/package.json
```diff
@@ -137,7 +137,7 @@
     "@types/js-cookie": "^3.0.3",
     "@types/lodash": "^4.14.196",
     "@types/mocha": "^10.0.1",
-    "@types/node": "^18.17.1",
+    "@types/node": "^20.4.5",
```
JanJakes, line 140: It seems that currently, we're using Node.js version 19.7.0. We should either upgrade to version 20 or keep the types in sync.

mailpoet/assets/js/src/subscribers/importExport/import/step_clean_list.tsx
```diff
@@ -3,7 +3,11 @@ import { CleanList } from './clean_list';
 
 function StepCleanList({ history }: RouteComponentProps): JSX.Element {
   return (
-    <CleanList onProceed={(): void => history.push('step_method_selection')} />
+    <CleanList
+      onProceed={(): void => {
+        history.push('step_method_selection');
+      }}
+    />
```
JanJakes, line 10: Seems like the `void` was misplaced here. The following should work:

```ts
<CleanList onProceed={() => void history.push('step_method_selection')} />
```

But your solution is also 👍

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/index.tsx
```diff
@@ -17,7 +17,7 @@ export function Tabs(): JSX.Element {
     hasEmails: s(editorStoreName).automationHasStep('mailpoet:send-email'),
   }));
   const pageParams = useMemo(
-    () => new URLSearchParams(location.search),
+    () => new URLSearchParams(location.search as string),
```
JanJakes, line 20: This looks like the type was completely lost. My guess is that it's because `history` is in version `4`, while `@types/history` was updated to version 5. Also, it seems that `@types/history` is deprecated now: https://www.npmjs.com/package/@types/history

mailpoet/webpack.config.js
```diff
@@ -9,6 +9,7 @@ const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 
 const globalPrefix = 'MailPoetLib';
 const PRODUCTION_ENV = process.env.NODE_ENV === 'production';
+const CIRCLECI_ENV = process.env.CIRCLECI === 'true';
```
JanJakes, line 12: There is also a `CI` global that most CI platforms use. Should we check for that one instead? Would be [future-proof with GitHub Actions](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables), in case we want to go that way one day.

JanJakes: Now I'm also wondering if the CI should compile JS in dev mode 🤔  Currently, [we have that only in the build script](https://github.com/mailpoet/mailpoet/blob/a5a79666635830cca9f837005f3919239c3c8ff2/mailpoet/build.sh#L18).

JanJakes (commented):
This is great! Nice approach with the script! 👏 

Unfortunately, there are some conflicts now, as it was waiting for me throughout the whole AFK. I hope it won't be too hard to resolve them.

We surely need to look into Dependabot/Renovate Bot, but before we get there, we can keep the script around.

> so I ended up writing a script (thanks ChatGPT) to automate the process

I find bash very hard to both write and read. Did you consider writing the script in PHP or JS?

> I know we handle wordpress packages separately, but I wasn't sure if it would be a problem to update the @types packages, so those may have been improperly updated. Should those always stay in sync?

I think a good rule of thumb is to keep the major versions the same. Otherwise, if it works, I guess it's fine. This will always be suboptimal until all the packages bundle TS defs themselves (`@wordpress` packages are heading in that direction).

mailpoet/webpack.config.js
```diff
@@ -107,7 +108,8 @@ const baseConfig = {
       ),
     },
   },
-  plugins: PRODUCTION_ENV ? [] : [new ForkTsCheckerWebpackPlugin()],
+  plugins:
+    PRODUCTION_ENV || CIRCLECI_ENV ? [] : [new ForkTsCheckerWebpackPlugin()],
```
JanJakes, line 112: An update — I just run into this and realized this shouldn't be needed at all as [we run the builds with the production env on CI anyway](https://github.com/mailpoet/mailpoet/blob/dcdd1c8cdb9773672b015c48fb5a814119bf6dbb/.circleci/config.yml#L151).

johnolek:
@JanJakes Thanks for your review of this and sorry I haven't responded to your comments and have just left this sitting! I totally agree that this script would be better in PHP, and I want to rewrite it as a robo command if I can. I'll try to convert it and address your very helpful comments soon.

mailpoet/webpack.config.js
johnolek: I'm a bit confused about this. Would that affect the JS tests too? For example this pipeline: https://app.circleci.com/pipelines/github/mailpoet/mailpoet/15351/workflows/8a8d463a-53db-498b-9251-f14eb7cf1f55/jobs/260417

johnolek:
I'm closing this in favor of #5127

PR #5081, opened 2023-07-31
Inactive subscriber calculation tweaks [MAILPOET-5410]

johnolek:
## Description

This PR makes some significant changes to the way engagement scores are calculated and updated.
- Sending a newsletter to a subscriber marks them as being due for an engagement score recalculation the next time the cron runs.
- if a subscriber currently has an engagement score but hasn't received a newsletter in the last year, their engagement score will be reset to "unknown" instead of simply remaining the same.
- We're no longer counting machine opens as opens for the engagement score calculation.
- All of the individual stats on a subscriber's statistics page are now based on data from the last year, not all time. The UI has been updated with text to make this clear.
- The WC revenues stats also now include data just for the last year.
- Individual stats pages now include a new section displaying dates of last engagement of different types (click, open, page view, etc).

## Code review notes

_N/A_

## QA notes

If possible, could you please benchmark a larger send before and after this change? It would be good to know if the additional query is going to have a noticeable affect on performance.

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5410

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/assets/js/src/subscribers/stats.tsx
+++ b/mailpoet/assets/js/src/subscribers/stats.tsx
@@ -4,7 +4,7 @@ import { MailPoet } from 'mailpoet';
 import { Loading } from 'common/loading';
 import { useGlobalContextValue } from 'context';
 
-import { Heading } from 'common/typography/heading/heading';
+import { Heading } from 'common';
 import { StatsHeading } from './stats/heading';
 import { Summary } from './stats/summary';
 import { WoocommerceRevenues } from './stats/woocommerce_revenues';
--- a/mailpoet/assets/js/src/subscribers/stats.tsx
+++ b/mailpoet/assets/js/src/subscribers/stats.tsx
@@ -4,7 +4,8 @@ import { MailPoet } from 'mailpoet';
 import { Loading } from 'common/loading';
 import { useGlobalContextValue } from 'context';
 
-import { StatsHeading as Heading } from './stats/heading';
+import { Heading } from 'common/typography/heading/heading';
+import { StatsHeading } from './stats/heading';
 import { Summary } from './stats/summary';
 import { WoocommerceRevenues } from './stats/woocommerce_revenues';
 import { OpenedEmailsStats } from './stats/opened_email_stats';
@@ -68,7 +69,8 @@ export function SubscriberStats(): JSX.Element {
 
   return (
     <div className="mailpoet-subscriber-stats">
-      <Heading email={stats.email} />
+      <StatsHeading email={stats.email} />
+      <Heading level={4}>{MailPoet.I18n.t('engagementPeriodHeading')}</Heading>
       <p>
         {MailPoet.I18n.t('lastEngagement')}
         {': '}
--- a/mailpoet/assets/js/src/subscribers/stats.tsx
+++ b/mailpoet/assets/js/src/subscribers/stats.tsx
@@ -9,6 +9,7 @@ import { StatsHeading } from './stats/heading';
 import { Summary } from './stats/summary';
 import { WoocommerceRevenues } from './stats/woocommerce_revenues';
 import { OpenedEmailsStats } from './stats/opened_email_stats';
+import { EngagementSummary } from './stats/engagement_summary';
 
 export type StatsType = {
   email: string;
@@ -18,6 +19,11 @@ export type StatsType = {
   click: number;
   engagement_score: number;
   last_engagement?: string;
+  last_click?: string;
+  last_open?: string;
+  last_sending?: string;
+  last_page_view?: string;
+  last_purchase?: string;
   woocommerce: {
     currency: string;
     value: number;
@@ -71,13 +77,6 @@ export function SubscriberStats(): JSX.Element {
     <div className="mailpoet-subscriber-stats">
       <StatsHeading email={stats.email} />
       <Heading level={4}>{MailPoet.I18n.t('engagementPeriodHeading')}</Heading>
-      <p>
-        {MailPoet.I18n.t('lastEngagement')}
-        {': '}
-        {stats.last_engagement
-          ? MailPoet.Date.format(stats.last_engagement)
-          : MailPoet.I18n.t('never')}
-      </p>
       <div className="mailpoet-subscriber-stats-summary-grid">
         <Summary
           click={stats.click}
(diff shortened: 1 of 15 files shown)


mailpoet/lib/Subscribers/SubscribersRepository.php
```diff
@@ -304,6 +304,21 @@ public function bulkUpdateLastSendingAt(array $ids, DateTimeInterface $dateTime)
...
+    }
+    $this->entityManager->createQueryBuilder()
+      ->update(SubscriberEntity::class, 's')
+      ->set('s.engagementScoreUpdatedAt', ':dateTime')
+      ->where('s.id IN (:ids)')
+      ->setParameter('dateTime', $dateTime)
+      ->setParameter('ids', $ids)
+      ->getQuery()
+      ->execute();
+    return count($ids);
```
brezocordero: What do you think about removing this return value?
It does not look like we use it. And I know we return this same count of ids in other update functions (might have done it myself) but it does not provide any information about how many ids where actually upated.

mailpoet/assets/js/src/subscribers/stats.tsx
```diff
@@ -4,10 +4,12 @@ import { MailPoet } from 'mailpoet';
 import { Loading } from 'common/loading';
 import { useGlobalContextValue } from 'context';
 
-import { StatsHeading as Heading } from './stats/heading';
+import { Heading } from 'common/typography/heading/heading';
```
brezocordero: It looks like the imports were simplified [here](https://github.com/mailpoet/mailpoet/commit/c4ce4fd10b8b55b3c1f536ba01a04ee11caaa7b1), in case you want to do the same.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -274,6 +274,8 @@ private function processSending(SendingTask $queue, int $timer): void {
         if (!$isTransactional) {
           $now = Carbon::createFromTimestamp((int)current_time('timestamp'));
           $this->subscribersRepository->bulkUpdateLastSendingAt($foundSubscribersIds, $now);
+          // We're nullifying this value so these subscribers' engagement score will be recalculated the next time the cron runs
+          $this->subscribersRepository->bulkUpdateEngagementScoreUpdatedAt($foundSubscribersIds, null);
```
brezocordero: You have probably thought about this but why not modify the bulkUpdateLastSendingAt and make this update in the same query?

mailpoet/assets/js/src/subscribers/stats.tsx
```diff
@@ -87,6 +88,15 @@ export function SubscriberStats(): JSX.Element {
             engagement_score: stats.engagement_score,
           }}
         />
+        <EngagementSummary
+          lastClick={stats.last_click}
+          lastEngagement={stats.last_engagement}
+          lastOpen={stats.last_open}
+          lastPageView={stats.last_page_view}
+          lastPurchase={stats.last_purchase}
+          lastSending={stats.last_sending}
```
brezocordero: These properties are optional on the StatsType, should we check for existence or make them required?

brezocordero (dismissed):
@johnolek This is great work.
I have some small questions/comments but I have approved the PR because they are not required.
Feel free to assign to QA if you prefer to ignore them.

mailpoet/lib/Subscribers/SubscribersRepository.php
johnolek: Very good point! I'll remove this.

mailpoet/assets/js/src/subscribers/stats.tsx
johnolek: Thanks for pointing this out! Updated in 59f5ff3d9a54304d01757fe8bf3736079db25c20

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
johnolek: I did think it would be nice to execute both of these updates in the same query, but I wanted to keep the methods separate in case we ever wanted to use them separately in other contexts, and to keep them easier/clearer to test.

My guess was that this wouldn't cause much of a performance hit, but it's probably worthwhile to do everything we can to minimize the query count during the sending process.

I decided to wrap the updates in a transaction in 0783c6ec0db9e47298f503957ff7b682b6ee326, which should address this issue for anyone using InnoDB. Does this seem like a reasonable approach?

mailpoet/assets/js/src/subscribers/stats.tsx
johnolek: It's ok for them to be `undefined` but I agree that it would be nicer to be more explicit about the fact that these are optional. I've refactored the `EngagementStats` component in 3dceb386e0157b189a2bf61b221eb3c9960e446 to hopefully do a better job of that.

johnolek:
Thanks for the review, @brezocordero ! I think I've addressed all your comments in my last few commits. Since you're AFK for the next couple of weeks I'll ask someone else on the team to take a second look.

johnolek:
@costasovo Brezo did an initial review of this PR and left for a couple weeks of AFK before I had a chance to complete the revisions. Would you have time to take a second look?

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
costasovo: @johnolek My knowledge of InnoDB isn't very profound. So just to confirm I understand. Using the transaction still means DB needs to run two queries, but the expected performance gain comes from the data being written only once. Correct?

costasovo (approved):
Thanks @johnolek for looking into this. This will make the engagement score more useful.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
johnolek: @costasovo All I was thinking was that since InnoDB supports transactions and MyISAM doesn't, if a user is using InnoDB this might result in only a single round trip to the database. I think I may have been mistaken though, and using a transaction doesn't necessarily mean the operations would only require a single round trip.

Do you think it would be worthwhile to create a more generic bulk update method, one that accepts key/value pairs and can bulk update multiple columns in a single query? 

My suspicion is that the code should be ok as is -- processing a sending queue already requires lots of separate DB queries and this one shouldn't be very slow. I might ask if QA can do a before/after test to see if this change affects sending time for very large sends. We can always optimize this later if we discover it's causing performance issues.

Aschepikov (approved):
GJ!

PR #5090, opened 2023-08-03
MAILPOET-5496/update-used-shipping-and-payment-methods-timespan

johnolek:
## Description

This adds the "over all time" option to the used shipping method and used payment method dynamic segment filters, which we missed when working on MAILPOET-4991 (https://github.com/mailpoet/mailpoet/pull/5057).

## Code review notes

_N/A_

## QA notes

This update changes how the data for the number of days is stored for these two filters, and it includes a migration to handle that. Please make sure that filters created before this version continue to work exactly as they did before after the update, and please also make sure that the migration ran successfully before testing (I think it should just require deactivating and reactivating the plugin).

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/5057 completed this work for other fliters.
I used some of the changes introduced in https://github.com/mailpoet/mailpoet/pull/5064, so that will need to be merged before this will be ready to review (edit: that PR has been merged and this PR has been rebased)

## Linked tickets
[MAILPOET-5496](https://mailpoet.atlassian.net/browse/MAILPOET-5496)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5496]: https://mailpoet.atlassian.net/browse/MAILPOET-5496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedPaymentMethod.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedPaymentMethod.php
@@ -54,10 +54,6 @@ class WooCommerceUsedPaymentMethod implements Filter {
     }
 
     $data = $filterData->getData();
-    // Backwards compatibility for filters saved before timeframe was added
-    if (!isset($data['timeframe'])) {
-      $data['timeframe'] = DynamicSegmentFilterData::TIMEFRAME_IN_THE_LAST;
-    }
     $this->filterHelper->validateDaysPeriodData((array)$data);
 
     $includedStatuses = array_keys($this->wooHelper->getOrderStatuses());
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedPaymentMethod.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedPaymentMethod.php
@@ -42,7 +42,8 @@ class WooCommerceUsedPaymentMethod implements Filter {
     $filterData = $filter->getFilterData();
     $operator = $filterData->getParam('operator');
     $paymentMethods = $filterData->getParam('payment_methods');
-    $days = $filterData->getParam('used_payment_method_days');
+    $days = $filterData->getParam('days');
+    $isAllTime = $filterData->getParam('timeframe') === DynamicSegmentFilterData::TIMEFRAME_ALL_TIME;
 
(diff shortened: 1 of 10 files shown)


mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedPaymentMethod.php
```diff
@@ -52,27 +53,30 @@ public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $fi
       throw new InvalidFilterException('Missing payment methods', InvalidFilterException::MISSING_VALUE);
     }
 
-    if (!is_int($days) || $days < 1) {
-      throw new InvalidFilterException('Missing days', InvalidFilterException::MISSING_VALUE);
+    $data = $filterData->getData();
+    // Backwards compatibility for filters saved before timeframe was added
+    if (!isset($data['timeframe'])) {
```
rodrigoprimo: Have you considered adding the `timeframe` in the migration instead of having this backward compatibility code? Maybe it is better to have it here in case this code runs before the migration is executed?

rodrigoprimo (commented):
Good job! Looks good to me. I just left a question for us to discuss and to confirm something that I'm assuming. Probably no change will be necessary.

mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceUsedPaymentMethod.php
johnolek: Thanks for this suggestion! I think adding it to the migrationand removing the backwards compatibility code makes a lot more sense. I've updated the migration in 7b12046ceec7eaefa9ddb071a833398e70ca48b8. Can you think of any way that the new migration could potentially fail? The update is wrapped in some conditionals, but they should always be true for any saved filters.

rodrigoprimo: Not really, I just assumed that was why you opted to add the BC code. Creating the `timeframe` in the migration seems like a good approach to me.

rodrigoprimo (dismissed):

johnolek:
FYI the last force push was just rebasing this branch on `trunk`, it didn't include any changes since the review.

Aschepikov (approved):
GJ! Thanks for the fixing

PR #5097, opened 2023-08-10
New automation templates [MAILPOET-5384]

JanJakes:
## Description

Adds new automation templates and makes them configurable from the premium plugin.

## Code review notes

_N/A_

## QA notes

Test all the three new "Purchased..." automation templates both with and without the premium plugin (there's a small difference — in premium there is no delay step and thus the emails are transactional).

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/764

## Linked tickets

[MAILPOET-5384]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5384]: https://mailpoet.atlassian.net/browse/MAILPOET-5384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Registry.php
+++ b/mailpoet/lib/Automation/Engine/Registry.php
@@ -54,6 +54,20 @@ class Registry {
 
   public function addTemplate(AutomationTemplate $template): void {
     $this->templates[$template->getSlug()] = $template;
+
+    // keep coming soon templates at the end
+    uasort(
+      $this->templates,
+      function (AutomationTemplate $a, AutomationTemplate $b): int {
+        if ($a->getType() === AutomationTemplate::TYPE_COMING_SOON) {
+          return 1;
+        }
+        if ($b->getType() === AutomationTemplate::TYPE_COMING_SOON) {
+          return -1;
+        }
+        return 0;
+      }
+    );
   }
 
   public function getTemplate(string $slug): ?AutomationTemplate {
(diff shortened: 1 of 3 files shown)


johnolek (changes requested):
@JanJakes This code looks 100% fine to me, but there are a couple of general issues I'm seeing:

1. Unless I'm doing something wrong, it's not possible to make any changes to filters without premium, and these all need modified filter values to work. Basically it doesn't seem possible to activate these automations as a free user.
2. This applies to some of the existing Woo templates as well: I don't think we should display Woo templates unless WC is active. The triggers aren't available without WC being active, so automations created from these templates with WC inactive will never be able to function.
3. This isn't a show stopper, but I was wondering if there might be a way to show users exactly what needs to be filled in/updated after creating a template. Here it doesn't feel immediately intuitive that I need to go into `Order status changed` to update a filter value. I thought about immediately running validation after loading the editor, but that doesn't seem quite right because it makes it look like the user forgot something. Again this is really minor but maybe something to think about for the future if you haven't already.

JanJakes:
@johnolek Ahh, thanks for so many great catches!

1. We don't want to make filters editable in the free plugin, so we'll need to wait until we have specific triggers for these. At the moment, I moved it all to premium. (Discussed [here](https://a8c.slack.com/archives/C0313447YBC/p1691748694442829).)
2. Thanks! Fixed. (Discussed [here](https://a8c.slack.com/archives/C0313447YBC/p1691756684927729).)
3. This is pretty much resolved by 1.

johnolek (approved):
Thanks for discussing this and finding a good solution! This looks good to me.

Aschepikov (approved):
GJ!

PR #5099, opened 2023-08-10
Remove fallback to fetch any sending queue for a preview when no queue ID set [MAILPOET-5206]

JanJakes:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

Before updating to this branch, try to reproduce the issue (with a `trunk` build or latest release build):
1. Create an automation with a send email step (e.g., "someone-subscribed" -> "send-email").
2. Trigger the automation and receive one emai.
3. Change the contents of the email, and save it.
4. Return to the email editor (from the automation step), and check the email preview. Now it's showing the old version without the edits in step 3 (= the bug).

Now update MP with the build from this branch, and the email preview should be fixed.

Please also ensure that email previews and view-in-browser links work in all other contexts.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5206]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5206]: https://mailpoet.atlassian.net/browse/MAILPOET-5206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserController.php
+++ b/mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserController.php
@@ -115,8 +115,6 @@ class ViewInBrowserController {
       return null;
     }
 
-    return !empty($data['queue_id'])
-      ? $this->sendingQueuesRepository->findOneById($data['queue_id'])
-      : $this->sendingQueuesRepository->findOneBy(['newsletter' => $newsletter->getId()]);
+    return isset($data['queue_id']) ? $this->sendingQueuesRepository->findOneById($data['queue_id']) : null;
   }
 }
--- a/mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserController.php
+++ b/mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserController.php
@@ -2,8 +2,6 @@
 
 namespace MailPoet\Newsletter\ViewInBrowser;
 
-use MailPoet\Entities\NewsletterEntity;
-use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Sending\SendingQueuesRepository;
@@ -53,7 +51,7 @@ class ViewInBrowserController {
     $subscriber = $this->getSubscriber($data);
 
     // if queue and subscriber exist, subscriber must have received the newsletter
-    $queue = $this->getQueue($newsletter, $data);
+    $queue = isset($data['queue_id']) ? $this->sendingQueuesRepository->findOneById($data['queue_id']) : null;
     if (!$isPreview && $queue && $subscriber->getId() && !$this->sendingQueuesRepository->isSubscriberProcessed($queue, $subscriber)) {
       throw new \InvalidArgumentException("Subscriber did not receive the newsletter yet");
     }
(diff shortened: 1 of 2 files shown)


rodrigoprimo:
I believe the commit below added this logic that we are removing in this PR. From the commit message, it is not clear to me why the logic was added.

https://github.com/mailpoet/mailpoet/commit/9698cf2d2e6b0e07d7a65cd1812ebe5f076212db#diff-40bd229f32573f36d7cb97ac3da1352c58883dc5c8c3d48fc872645583eadeb4L37
https://github.com/mailpoet/mailpoet/commit/9698cf2d2e6b0e07d7a65cd1812ebe5f076212db#diff-40bd229f32573f36d7cb97ac3da1352c58883dc5c8c3d48fc872645583eadeb4R43

Just sharing here for future reference if needed.

mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserController.php
```diff
@@ -103,20 +101,4 @@ private function getSubscriber(array $data): SubscriberEntity {
...
-    if ($newsletter->getType() === NewsletterEntity::TYPE_WELCOME) {
-      return null;
-    }
-
-    // reset queue when automatic email is being previewed
-    if ($newsletter->getType() === NewsletterEntity::TYPE_AUTOMATIC && !empty($data['preview'])) {
-      return null;
-    }
-
-    return !empty($data['queue_id'])
```
rodrigoprimo, line 118: Reading the Jira ticket and the commit messages, it is clear to me why it was necessary to change this part of the function and return `null` if `queue_id` is not set. But it is not clear to me why the second commit is necessary and why we don't need any more special handling for welcome and automatic emails. Could you please elaborate?

rodrigoprimo (commented):
Overall the changes look good to me. I left a comment linking to what I believe is the commit that introduced the behavior that we are changing here. It is unclear why it was introduced in the first place, even after checking the commit.

I left a comment with a clarifying question.

Should we add some testing instructions for QA on how to reproduce the problem that is being fixed by this PR? Or is there a reason you chose not to include the instructions?

mailpoet/lib/Newsletter/ViewInBrowser/ViewInBrowserController.php
JanJakes: @rodrigoprimo The second commit is not necessary, it's more like a cleanup. I think those IF's weren't necessary in the first place. They are just hotfixes on the original code, which used to be like this:

```php
if($data->newsletter->type !== Newsletter::TYPE_WELCOME) {
  $data->queue = ...; // fetch by queue ID or fall back to any queue
}

// now here we already have a fallback queue so we have to remove it
if($data->newsletter->type === Newsletter::TYPE_AUTOMATIC && !empty($data->preview)) {
  $data->queue = false;
}
```

Then, it was refactored to a form where the `fetch by queue ID or fall back to any queue` is actually **after** the IFs, and when we remove that fallback, those IFs make no more sense.

The second commit has in the message links to the original changes:

> Originally, this was done in an IF: https://github.com/mailpoet/mailpoet/commit/32f7d7771f2dcbd6e24a1dcc7912d7aacbd124c3
> Then another hotfix added here: https://github.com/mailpoet/mailpoet/commit/116ee51497ae33456dbb5f804bee0f0754587d61

rodrigoprimo: Thanks for the additional details. I agree that it seems those ifs we not necessary in the first place.

rodrigoprimo:
> Should we add some testing instructions for QA on how to reproduce the problem that is being fixed by this PR? Or is there a reason you chose not to include the instructions?

Maybe you missed the question above? Mentioning it for visibility.

JanJakes:
@rodrigoprimo Ahh, sorry, I totally missed that. Makes total sense. Updated!

rodrigoprimo (approved):
Looks good to me. Thanks for adding the testing instructions.

Aschepikov (approved):
GJ!

PR #5102, opened 2023-08-11
Create segment templates page [MAILPOET-5393]

rodrigoprimo:
## Description

This PR implements the design of the segment templates page. The functionality is not implemented yet.

I'm creating this PR as draft for early feedback. Checking first with a dev, and then I plan to check with the designers before finalizing it.

The new page can be accessed in the URL:

`/wp-admin/admin.php?page=mailpoet-segments#/segment-templates`

Things that I still need to do to finish this PR:

~~- Check why the elements are not behaving well on mobile. Maybe it is the custom CSS modifications I had to make, as I mostly use WP components for the parts that are broken on mobile.~~ I spoke with @NeosinneR about this, and he ask me to just spend a bit of time fixing those issues as other parts of MailPoet don't work well on mobile. I made a few adjustments. The only thing pending that I'm not planning to address is the categories tab and currently at least it is usable.
~~- Check what needs to be done for the WP native i18n functions to work on the new page.~~ (actually, just using the native WP i18n functions seems to be enough, I confirmed that the strings were added to the .pot file)
- ~~Get the link for the new segment page dynamically instead of hard coding it.~~ done

I tried to reuse as much as possible WP and Woo components. The result is similar to the Figma file, but some differences exist. In some cases, I opted to leave the default design of some Woo and WP components instead of customizing it (SearchControl and tabs). I may have missed one or another component that I could have used or opted to use a suboptimal component where there is a better alternative.

Also, my knowledge of CSS is minimal, so there might be better ways to achieve the things that I'm achieving here.

## Code review notes

_N/A_

## QA notes

- To view the page go to `/wp-admin/admin.php?page=mailpoet-segments#/segment-templates`
- There are some differences between the Figma file and what was created here. Those differences were approved by Shiva here: p1692390511991199-slack-C05BG9S6B96

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5393]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5393]: https://mailpoet.atlassian.net/browse/MAILPOET-5393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/mailpoet-templates.scss
+++ b/mailpoet/assets/css/src/mailpoet-templates.scss
@@ -1,0 +1,113 @@
+@import '../../../node_modules/@woocommerce/components/build-style/style';
+@import 'settings/colors';
+
+.mailpoet-templates-container {
+  display: flex;
+  flex-direction: column;
+  gap: 26px;
+  margin: auto;
+  max-width: 1072px;
+}
+
+.mailpoet-templates-heading {
+  align-items: center;
+  align-self: stretch;
+  display: flex;
+  justify-content: space-between;
+  padding: 16px 0;
+
+  h1.wp-heading-inline {
+    align-items: center;
+    display: flex;
+    padding: 4px 0;
+  }
+
+  .components-search-control {
+    width: 180px;
+  }
+
+  .components-base-control__field {
+    margin: 0;
+  }
+
+  .components-search-control input[type=search].components-search-control__input {
+    background: $color-white;
+    height: 36px;
+  }
+}
+
+.components-tab-panel__tabs {
+  border-bottom: 1px solid $color-poet-gray-dividers;
+
+  .woocommerce-badge {
+    background-color: #e5e5e8;
+    font-size: 11px;
+    font-weight: normal;
+    margin: 2px 0 2px 10px;
+    width: 26px;
+  }
+}
+
+.mailpoet-templates-card-grid {
+  border-bottom: 1px solid $color-poet-gray-dividers;
+  display: grid;
+  grid-gap: 32px;
+  grid-template-columns: repeat(auto-fill, minmax(336px, 1fr));
+  padding: 26px 0 32px;
+}
+
+.mailpoet-templates-card {
+  align-items: flex-end;
+  display: flex;
+  flex-direction: column;
+  flex-shrink: 0;
+  height: 148px;
+  width: 336px;
+}
+
+.mailpoet-templates-card-header {
+  border-bottom-width: 0;
+  padding: 24px 24px 8px;
+
+  .mailpoet-templates-badge-essential {
+    align-items: center;
+    background: $color-primary;
+    border-radius: 0 0 2px 2px;
+    display: flex;
+    height: 24px;
+    justify-content: center;
+    position: absolute;
+    right: 24px;
+    top: 0;
+    width: 24px;
+
+    svg {
+      fill: $color-white;
+    }
+  }
+
+  a.is-link {
+    font-size: 14px;
+    font-style: normal;
+    font-weight: 600;
+    line-height: 20px;
+    text-decoration: none;
+  }
+}
+
+.mailpoet-templates-card-body {
+  padding: 0 24px 24px;
+
+  p {
+    margin: 0;
+  }
+
+  .woocommerce-tag {
+    margin: 0;
+    padding-top: 10px;
+  }
+}
+
+.mailpoet-templates-footer {
+  text-align: center;
+}
--- a/mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
@@ -1,0 +1,46 @@
+import {
+  Button,
+  Card,
+  CardBody,
+  CardHeader,
+  Icon,
+  Tooltip,
+} from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { starFilled } from '@wordpress/icons';
+import { Tag } from '@woocommerce/components';
+
+type TemplateListItemProps = {
+  template: {
+    id: number;
+    name: string;
+    description: string;
+    category: string;
+    isEssential: boolean;
+  };
+};
+
+export function TemplateListItem({
+  template,
+}: TemplateListItemProps): JSX.Element {
+  return (
+    <Card className="mailpoet-templates-card">
+      <CardHeader className="mailpoet-templates-card-header">
+        {template.isEssential && (
+          <Tooltip text={__('Essential segment', 'mailpoet')}>
+            <div className="mailpoet-templates-badge-essential">
+              <Icon icon={starFilled} size={18} />
+            </div>
+          </Tooltip>
+        )}
+        <Button variant="link" href="#/">
+          {template.name}
+        </Button>
+      </CardHeader>
+      <CardBody className="mailpoet-templates-card-body">
+        <p>{template.description}</p>
+        <Tag label={template.category} />
+      </CardBody>
+    </Card>
+  );
+}
--- a/mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
@@ -12,6 +12,7 @@ import { Badge } from '@woocommerce/components';
 import { HideScreenOptions } from 'common/hide_screen_options/hide_screen_options';
 import { TopBarWithBeamer } from 'common/top_bar/top_bar';
 import { TemplateListItem } from 'segments/dynamic/templates/components/template_list_item';
+import * as ROUTES from 'segments/routes';
 
 const tabConfig = [
   {
@@ -172,7 +173,7 @@ export function SegmentTemplates(): JSX.Element {
         </FlexItem>
 
         <FlexItem>
-          <Button variant="secondary" href="#/new-segment">
+          <Button variant="secondary" href={`#${ROUTES.NEW_DYNAMIC_SEGMENT}`}>
             {__('Create custom segment', 'mailpoet')}
           </Button>
         </FlexItem>
@@ -190,7 +191,7 @@ export function SegmentTemplates(): JSX.Element {
 
       <div className="mailpoet-templates-footer">
         <p>{__('Want to set your own conditions?', 'mailpoet')}</p>
-        <Button variant="link" href="#/new-segment">
+        <Button variant="link" href={`#${ROUTES.NEW_DYNAMIC_SEGMENT}`}>
           {__('Create custom segment', 'mailpoet')}
         </Button>
       </div>
--- a/mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
@@ -1,0 +1,199 @@
+import { __ } from '@wordpress/i18n';
+import { chevronLeft } from '@wordpress/icons';
+import {
+  Button,
+  Flex,
+  FlexItem,
+  FlexBlock,
+  SearchControl,
+  TabPanel,
+} from '@wordpress/components';
+import { Badge } from '@woocommerce/components';
+import { HideScreenOptions } from 'common/hide_screen_options/hide_screen_options';
+import { TopBarWithBeamer } from 'common/top_bar/top_bar';
+import { TemplateListItem } from 'segments/dynamic/templates/components/template_list_item';
+
+const tabConfig = [
+  {
+    name: 'all',
+    title: (
+      <>
+        <span>All</span>
+        <Badge count={21} />
+      </>
+    ) as any, // eslint-disable-line @typescript-eslint/no-explicit-any -- typed as string but supports JSX
+  },
+  {
+    name: 'essentials',
+    title: (
+      <>
+        <span>Essentials</span>
+        <Badge count={10} />
+      </>
+    ) as any, // eslint-disable-line @typescript-eslint/no-explicit-any -- typed as string but supports JSX
+  },
+  {
+    name: 'engagement',
+    title: (
+      <>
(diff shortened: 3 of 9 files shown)


mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+            {__('Create custom segment', 'mailpoet')}
+          </Button>
+        </FlexItem>
+      </Flex>
+
+      <TabPanel tabs={tabConfig}>
+        {() => (
+          <div className="mailpoet-templates-card-grid">
+            {templates.map((template) => (
+              <TemplateListItem template={template} />
```
johnolek: I think these might need a `key` prop to avoid a React warning

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
```diff
@@ -0,0 +1,46 @@
...
+            </div>
+          </Tooltip>
+        )}
+        <Button variant="link" href="#/">
+          {template.name}
+        </Button>
+      </CardHeader>
+      <CardBody className="mailpoet-templates-card-body">
+        {template.description}
+        <Tag label={template.isEngaged ? 'Engagement' : 'Disengaged'} />
```
johnolek: I'm not seeing a line break between these elements locally
image

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+import { HideScreenOptions } from 'common/hide_screen_options/hide_screen_options';
+import { TopBarWithBeamer } from 'common/top_bar/top_bar';
+import { TemplateListItem } from 'segments/dynamic/templates/components/template_list_item';
+
+const tabConfig = [
+  {
+    name: 'all',
+    title: (
+      <>
+        <span>All</span>
```
johnolek, line 22: I believe these titles still need to be translated

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
johnolek: I'm assuming we're going to have a lot more tags than just these two. Are you thinking we would have booleans for all tags we'll have in the future? It seems like we might want a more flexible option, like passing in tag text (or an array of tags if we want to show multiple). These will also need to be translated.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export function TemplateListItem({
+  template,
+}: TemplateListItemProps): JSX.Element {
+  return (
+    <Card className="mailpoet-templates-card">
+      <CardHeader className="mailpoet-templates-card-header">
+        {template.isEssential && (
+          <Tooltip text={__('Essential segment', 'mailpoet')}>
```
johnolek, line 30: My preference would be to include a `delay` prop of 0 to make these appear immediately. I'm not sure what design had in mind.

mailpoet/assets/css/src/mailpoet-templates.scss
```diff
@@ -0,0 +1,107 @@
...
+  .components-search-control {
+    width: 180px;
+  }
+
+  .components-base-control__field {
+    margin: 0;
+  }
+
+  .components-search-control input[type=search].components-search-control__input {
+    background: #fff;
```
johnolek: I don't know if it makes much of a difference, but I know we have variables for a lot of colors in `mailpoet/assets/css/src/settings/_colors.scss`, and it seems more idiomatic in the codebase to use those. Is there any reason you went with the hex values for colors that we have variables for?

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
```diff
@@ -0,0 +1,46 @@
...
+  Icon,
+  Tooltip,
+} from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { starFilled } from '@wordpress/icons';
+import { Tag } from '@woocommerce/components';
+
+type TemplateListItemProps = {
+  template: {
+    id: number;
```
johnolek, line 15: Is the ID being used for anything?

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+          </h1>
+        </FlexBlock>
+
+        <FlexItem>
+          <SearchControl label={__('Search segment templates', 'mailpoet')} />
+        </FlexItem>
+
+        <FlexItem>
+          <Button variant="secondary" href="#/new-segment">
+            {__('Create custom segment', 'mailpoet')}
```
johnolek, line 177: The "back" button on the new segment page always goes to the segment listing page, so we might want to update it so a user returns to the templates page if they get there from here.

johnolek (changes requested):
@rodrigoprimo I think this is looking pretty good so far! I left a fair number of comments but not all of them are important.

I will freely admit that CSS is also not my strong suit, so will you please let me know if you have any specific concerns you'd like a second pair of eyes on?

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
rodrigoprimo: I'm not sure why you do not see the line break. I see it both in Firefox and Chrome on Linux.

Could you please check different browsers and see if you reproduce the same problem?

Here is what I'm seeing on a Jurassic Ninja site (to make sure there are no changes in the local environment that are affecting this):

Screenshot from 2023-08-14 15-11-33

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
rodrigoprimo: I'm ok with not handling this at this point as per the ticket description the categories won't be hardcoded. Instead, they will be generated automatically from the templates. But your comment made me wonder how we are going to support translation for the categories if they are not hardcoded. I asked Ján about this in the Jira ticket.

There is a separate ticket to make the categories work.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
rodrigoprimo: I don't have a strong preference here, but I'd incline to use the default value of 700 milliseconds for `delay` instead of changing it to `0`. Or is there an indication in the Figma file or somewhere else that I missed that we should change this to `0`?

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
johnolek: Thanks for the explanation. I should have taken a closer look at the ticket!

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
rodrigoprimo: It was not being used. Now it is being used for the `key` parameter that is passed to `<TemplateListItem>`. Since the hard coded templates will be replaced in a subsequent ticket, I haven't put much thought into the data structure of the templates in this ticket.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
johnolek: Strictly a personal preference, so feel free to leave it as is, especially since this would be the default behavior for tooltips in general.

I didn't think it was working at first because I wasn't seeing it, but I just wasn't being patient enough. I don't think it's really obvious that the star is a tooltip trigger, so I was thinking people might miss it.

I think there may be some accessibility issues with the tooltip itself, but that's just the tooltip component itself and not something we can/should address in this PR. The only thing I think we could maybe change is I believe tooltip triggers should be focusable with the keyboard, but that wouldn't be the greatest UX since you would have to tab through the same tooltip multiple times in a listing with multiple essential segments.

There's other stuff too, like how I don't think the container for the tooltip content should have `aria-hidden=true`, there's no `role="tooltip` or `aria-describedby`, and it can't be dismissed with the keyboard other than by focusing something else (I read that you should be able to dismiss them with something like the esc key), but again these are all (potential) issues with the core Tooltip component itself, and I could be wrong! I'm no accessibility expert.

All that being said, this isn't exactly vital information for a user, so I don't think it's very important that these tooltips be perfectly accessible.

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
rodrigoprimo: My assumption is that the segments page will work the same way that the automations page work and that the "New segment" button in the segment listing page will always direct users to the templates page. So, what I believe we need to do is, in a subsequent PR, when we launch the templates page, we need to update the back button on the new segment page. Let me know if that is not the case, and I'm missing something.

mailpoet/assets/css/src/mailpoet-templates.scss
rodrigoprimo: Good point. I failed to consider this. I updated the commit to use colors from `_colors.scss` whenever possible.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
rodrigoprimo: That is a good point. I assumed, maybe mistakenly, that we had only two options (`Engagement` and `Disengaged`). I agree that if there are more, I need to change this part of the code (cc @NeosinneR for confirmation). Meanwhile, I wrapped the two strings with `__()`.

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
rodrigoprimo: Done

rodrigoprimo:
Thanks for the initial review, John. I believe I addressed all your points. Can you take a second look when you have a chance?

Since this PR is still in draft state, to avoid adding too many commits, I ended up making the changes directly to the already existing commits and force pushing.

> I will freely admit that CSS is also not my strong suit, so will you please let me know if you have any specific concerns you'd like a second pair of eyes on?

There is nothing specific, but thanks for checking.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
johnolek: It looks fine to me in safari and chrome, but I'm seeing this very consistently in Firefox, both developer edition (117.0b5) and a fresh install of the normal Firefox (116.0.2) with no extensions/changes on MacOS 13.1. This is also using jurassic.ninja.

image

If I wrap the `template.description` in a `<p>` (would this be more semantic than having the description just in the main body div?) then I get consistency between firefox and other browsers, but it would need to be cleaned up:

image

I'm really not sure what's causing the difference, and it's interesting that it works in Firefox for you on linux. Maybe there's something very specific about my system that's off.

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
johnolek: That makes perfect sense, thanks for the explanation!

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
johnolek: Just wanted to make sure I wasn't missing anything, thanks!

johnolek:
Thanks for your detailed responses to my comments, @rodrigoprimo, and thanks for making those updates! I think I've responded to everything but let me know if I missed anything. And feel free to ignore my mini-rant about tooltips.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
rodrigoprimo: Thanks for sharing all your thoughts about the potential accessibility issues for this component. I agree that this is not something we should address in this PR.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
rodrigoprimo: I implemented your suggestion of wrapping the `template.description` in a `<p>`. Could you please check if the issue is now fixed?

Please note that adding the `<p>` changed the font size of the description from 14px to 13px. I checked the Figma file and 13px is actually the correct size.

rodrigoprimo: After checking with Ján M., I changed this part of the code to display the template category. It is still hardcoded and will need to be changed in the other ticket where we will actually implement the templates.

johnolek: Looks good to me now! Thanks for adjusting.

johnolek:
@rodrigoprimo Sorry it's taken me a little while to take a look at your latest changes. Everything looks good to me now. I'm assigning this back to you since this is still a draft.

johnolek (dismissed):

rodrigoprimo:
@johnolek, this PR is ready for a final review. I believe nothing changed since the last time you checked it.

johnolek (approved):
Sorry @rodrigoprimo once again for a delay. Indeed my local version of this branch looks identical to what's on GitHub.

Aschepikov (approved):
GJ!

PR #5107, opened 2023-08-14
Show subscriber stats for 3 time periods MAILPOET-5508

johnolek:
## Description

In a previous PR #5081 we changed all stats on the subscriber page to only show data for the past year (previously all the stats were lifetime data other than the engagement score). Since lifetime stats are still relevant, especially for WooCommerce data, this PR updates the stats page to include data for 3 different time periods: 30 days, 12 months, and lifetime. 

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

#5081 

## Linked tickets

[MAILPOET-5508](https://mailpoet.atlassian.net/browse/MAILPOET-5508)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5508]: https://mailpoet.atlassian.net/browse/MAILPOET-5508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
+++ b/mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
@@ -58,12 +58,9 @@ class SubscriberStatisticsRepository extends Repository {
   }
 
   public function getStatisticsOpenCountQuery(SubscriberEntity $subscriber, ?Carbon $startTime = null): QueryBuilder {
-    $queryBuilder = $this->getStatisticsCountQuery(StatisticsOpenEntity::class, $subscriber)
-      ->join('stats.newsletter', 'newsletter');
+    $queryBuilder = $this->getStatisticsCountQuery(StatisticsOpenEntity::class, $subscriber);
     if ($startTime) {
-      $queryBuilder
-        ->andWhere('(newsletter.sentAt >= :dateTime OR newsletter.sentAt IS NULL)')
-        ->andWhere('stats.createdAt >= :dateTime')
+      $queryBuilder->join(StatisticsNewsletterEntity::class, 'sent_stats', 'WITH', 'stats.newsletter = sent_stats.newsletter AND sent_stats.sentAt >= :dateTime')
         ->setParameter('dateTime', $startTime);
     }
     return $queryBuilder;
--- a/mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
+++ b/mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
@@ -48,9 +48,7 @@ class SubscriberStatisticsRepository extends Repository {
   public function getStatisticsClickCount(SubscriberEntity $subscriber, ?Carbon $startTime = null): int {
     $queryBuilder = $this->getStatisticsCountQuery(StatisticsClickEntity::class, $subscriber);
     if ($startTime) {
-      $queryBuilder
-        ->andWhere('stats.createdAt >= :dateTime')
-        ->setParameter('dateTime', $startTime);
+      $this->applyDateConstraint($queryBuilder, $startTime);
     }
     return (int)$queryBuilder
       ->getQuery()
@@ -60,8 +58,7 @@ class SubscriberStatisticsRepository extends Repository {
   public function getStatisticsOpenCountQuery(SubscriberEntity $subscriber, ?Carbon $startTime = null): QueryBuilder {
     $queryBuilder = $this->getStatisticsCountQuery(StatisticsOpenEntity::class, $subscriber);
     if ($startTime) {
-      $queryBuilder->join(StatisticsNewsletterEntity::class, 'sent_stats', 'WITH', 'stats.newsletter = sent_stats.newsletter AND sent_stats.sentAt >= :dateTime')
-        ->setParameter('dateTime', $startTime);
+      $this->applyDateConstraint($queryBuilder, $startTime);
     }
     return $queryBuilder;
   }
@@ -132,5 +129,12 @@ class SubscriberStatisticsRepository extends Repository {
       count($purchases),
       $this->wcHelper
     );
+  }
+
+  private function applyDateConstraint(QueryBuilder $queryBuilder, Carbon $startTime): QueryBuilder {
+    $queryBuilder->join(StatisticsNewsletterEntity::class, 'sent_stats', 'WITH', 'stats.newsletter = sent_stats.newsletter AND stats.subscriber = sent_stats.subscriber AND sent_stats.sentAt >= :dateTime')
+      ->setParameter('dateTime', $startTime);
+
+    return $queryBuilder;
   }
 }
--- a/mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
+++ b/mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
@@ -35,55 +35,64 @@ class SubscriberStatisticsRepository extends Repository {
     return SubscriberEntity::class;
   }
 
-  public function getStatistics(SubscriberEntity $subscriber) {
+  public function getStatistics(SubscriberEntity $subscriber, ?Carbon $startTime = null) {
     return new SubscriberStatistics(
-      $this->getStatisticsClickCount($subscriber),
-      $this->getStatisticsOpenCount($subscriber),
-      $this->getStatisticsMachineOpenCount($subscriber),
-      $this->getTotalSentCount($subscriber),
-      $this->getWooCommerceRevenue($subscriber)
+      $this->getStatisticsClickCount($subscriber, $startTime),
+      $this->getStatisticsOpenCount($subscriber, $startTime),
+      $this->getStatisticsMachineOpenCount($subscriber, $startTime),
+      $this->getTotalSentCount($subscriber, $startTime),
+      $this->getWooCommerceRevenue($subscriber, $startTime)
     );
   }
 
-  public function getStatisticsClickCount(SubscriberEntity $subscriber): int {
-    $dateTime = (new Carbon())->subYear();
-    return (int)$this->getStatisticsCountQuery(StatisticsClickEntity::class, $subscriber)
-      ->andWhere('stats.createdAt > :dateTime')
-      ->setParameter('dateTime', $dateTime)
+  public function getStatisticsClickCount(SubscriberEntity $subscriber, ?Carbon $startTime = null): int {
+    $queryBuilder = $this->getStatisticsCountQuery(StatisticsClickEntity::class, $subscriber);
+    if ($startTime) {
+      $queryBuilder
+        ->andWhere('stats.createdAt >= :dateTime')
+        ->setParameter('dateTime', $startTime);
+    }
+    return (int)$queryBuilder
       ->getQuery()
       ->getSingleScalarResult();
   }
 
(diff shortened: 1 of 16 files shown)


mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
```diff
@@ -35,55 +35,64 @@ protected function getEntityClassName() {
...
-      ->join('stats.newsletter', 'newsletter')
-      ->andWhere('(newsletter.sentAt > :dateTime OR newsletter.sentAt IS NULL)')
-      ->andWhere('stats.createdAt > :dateTime')
-      ->setParameter('dateTime', $dateTime);
+  public function getStatisticsOpenCountQuery(SubscriberEntity $subscriber, ?Carbon $startTime = null): QueryBuilder {
+    $queryBuilder = $this->getStatisticsCountQuery(StatisticsOpenEntity::class, $subscriber)
+      ->join('stats.newsletter', 'newsletter');
+    if ($startTime) {
+      $queryBuilder
+        ->andWhere('(newsletter.sentAt >= :dateTime OR newsletter.sentAt IS NULL)')
```
costasovo: I see that for click we check only `stats.createdAt`. Shouldn't we unify that also for opens? I see it was like that in the past, but maybe for the last 30 days, it may make sense to include emails that someone opens a couple of days after they received them.

costasovo (changes requested):
When I tried it locally, I got an error.
After I checkout the branch, I ran `./do compile:all`.
I have the premium plugin disabled, and I opened stats for a subscriber who hasn't received anything.

Screenshot 2023-08-18 at 11 30 13
Screenshot 2023-08-18 at 11 30 43

johnolek:
> When I tried it locally, I got an error. 
> After I checkout the branch, I ran ./do compile:all.
> I have the premium plugin disabled, and I opened stats for a subscriber who hasn't received anything.

@costasovo I tried to reproduce this locally and was only able to get the error when I had premium enabled. I think that this issue might have been fixed elsewhere, because after rebasing this branch on `trunk` I'm not able to reproduce the error at all, either with premium enabled or disabled, and looking at subscribers with and without stats. Could you please test again and let me know if you're able to reproduce with the latest version I pushed?

mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
johnolek: Thank you for commenting on this, because this line has stood out to me before and I wasn't quite sure what to make of it.

I agree it makes sense to unify clicks and opens, but does it make more sense to only check for `createdAt` for opens, or to add a `newsletter.sentAt` check for clicks?

It really seems like the numbers are intended to describe clicks/opens related to just the newsletters that were sent during the described period, especially considering how we calculate percentages.

If we eliminate `sentAt` for opens, if someone was sent 2 emails 31 days ago, then clicked both of them 29 days ago, and then didn't receive any other emails, we would see a 30 day column of 
> Sent email: 0
> Opened: 2
> Machine-opened: 0
> Clicked: 2

and I think Not opened would be `-2` because we calculate it as `total_sent - opens`.

Now let's say they were actually sent 20 emails 31 days ago and clicked on 2 of them, then they were sent 2 additional emails 29 days ago that they didn't open, we would see

> Sent email: 2
> Opened: 2 (100%)
> Machine-opened: 0
> Clicked: 2 (100%)
> Not opened: 0 (0%)

Then there's the case of someone clicking say 5 links from older newsletters, then they get sent 1 email in the last 30 days, and we have something like 5 clicks for 1 sent email and end up showing a click rate of 500%.

I realize these are edge cases that almost certainly wouldn't come up in real life very often, but there might be more common cases I'm not thinking of where the stats could end up not looking right/adding up as expected.

Does that make sense to you, or do you think it would be best to just look at `createdAt`?

costasovo: @johnolek Thank you for elaborating!

> I agree it makes sense to unify clicks and opens, but does it make more sense to only check for createdAt for opens, or to add a newsletter.sentAt check for clicks?

Based on what you described, I think eliminating `sentAt` would create confusion. 

Now I'm also thinking about `OR newsletter.sentAt IS NULL` part. I assume we need this because `automatic` and `welcome` etc. emails (let's call all these automatic). They are never set as sent. So it seems that we already kind of ignore `sentAt` for the automatic emails which could lead to the issues you've described.

Maybe the ideal solution would be that we pick all sent emails in the given period and then count only clicks and opens associated with those emails. AFAIK we store the association a sent email – subscriber in `wp_mailpoet_statistics_newsletters` so maybe we could load the stats primarily based on this table, filter by sentAt, and join clicks/opens. What do you think?

costasovo:
> I'm not able to reproduce the error at all, either with premium enabled or disabled, and looking at subscribers with and without stats. Could you please test again and let me know if you're able to reproduce with the latest version I pushed?

@johnolek After checking out the rebased branch I'm not able to reproduce the issue anymore 👍

costasovo (commented):
We have to decide if we want to change the SQL for fetching stats.

costasovo:
I didn't want to request changes 🤦 I just wanted to submit a comment.

mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
johnolek: Based on the commit history, I believe you're 100% right about the reason for `sentAt IS NULL` being for automatic emails, and I very much like your idea of basing the opens on the actual sent newsletters for the period. Would that mean ignoring automatic email stats though? 

It looks like we don't store it for emails from automations either based on a quick test. Do you think it might make sense to remove the `OR nwesletter.sentAt IS NULL` and look at the newsletter's `created_at`, or something else, for any automatic/automation emails?

Do you happen to know why we don't store the `sent_at` for automatic emails?

costasovo: > Based on the commit history, I believe you're 100% right about the reason for sentAt IS NULL being for automatic emails, and I very much like your idea of basing the opens on the actual sent newsletters for the period. Would that mean ignoring automatic email stats though?

@johnolek The idea is to look into `wp_mailpoet_statistics_newsletters` for the `sent_at` (instead of `wp_mailpoet_newsletters`). We add there [a row for a subscriber and newsletter for each email that is sent via SendingQueue](https://github.com/mailpoet/mailpoet/blob/021d8774e74c10ab6224b16aeb783222c4207c9e/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L494).
The table contains the exact time of sentAt for each subscriber. It has records for all types of newsletters that go through SendingQueue (they have a scheduled task and sending queue record). It covers also welcome and automatic and automation emails.

> It looks like we don't store it for emails from automations either based on a quick test. Do you think it might make sense to remove the OR nwesletter.sentAt IS NULL and look at the newsletter's created_at, or something else, for any automatic/automation emails?
> 
> Do you happen to know why we don't store the sent_at for automatic emails?

We don't store sentAt for automatic emails in `wp_mailpoet_newsletters` because they are activated and sent continuously. We save `sentAt` to the `wp_mailpoet_newsletters`  only for one-time batch emails (newsletters, post_notifications) when they are sent to a full batch. So to me the `sentAt` in `wp_mailpoet_newsletters` feels like a cached value for displaying in listings. It should be same as the latest `sent_at` in `wp_mailpoet_statistics_newsletters`. Note that I don't have historical context for this. Maybe we just added `wp_mailpoet_statistics_newsletters` later...

So I think that in the queries for single subscriber stats, we could switch from `wp_mailpoet_newsletters` to `wp_mailpoet_statistics_newsletters` and filter by `sent_at` in `wp_mailpoet_statistics_newsletters`. I guess we still might need to join `wp_mailpoet_newsletters` to get subjects, though. 

Overall the `wp_mailpoet_statistics_newsletters` seems to be a much better fit for fetching single subscriber stats. I wonder why we didn't use it from the beginning. 🤔

johnolek: Thank you so much for the detailed explanation! I'm not sure why it didn't quite click for me yesterday, but in retrospect it should have been clear what you meant. I'll move forward with using `mailpoet_statistics_newsletters` as the primary source for stats. It really does make the most sense.

johnolek: Sorry for the delay in updating this. I think I've addressed it in e4153e765b26825374eaa86a832f75978850d2db. Let me know if you see any issues!

costasovo: @johnolek Thanks for updating the query for opens 👍 

Didn't we also want to update the query for clicks to unify opens and clicks? If I understand correctly now, we have fixed the issue when an email is sent before the interval and opened in the interval  (`Sent email: 0
Opened: 2`) for opens but it still may occur for clicks (`Sent email: 0
Clicked: 2)` am I right?

I looked at the query that Doctrine generated and I have a small suggestion:
**Current query:**
```SQL
SELECT COUNT(DISTINCT w0_.newsletter_id) AS sclr_0
FROM wp_mailpoet_statistics_opens w0_ 
INNER JOIN wp_mailpoet_statistics_newsletters w1_ 
ON (w0_.newsletter_id = w1_.newsletter_id AND  w1_.sent_at >= '2022-08-29T08:06:05.252426Z')
WHERE w0_.subscriber_id = 1094 AND (w0_.user_agent_type = 0)
```
I think we need to improve the JOIN condition to also include `subscriber_id`. That way we get the exact date for the specific subscriber. I think it is important for automatic emails where sending is spread through multiple months or even years. If we ignore subsriber_id we may join a record about an email sent to a different subscriber and count an open into multiple intervals.

**Modified query:**
```SQL
SELECT COUNT(DISTINCT w0_.newsletter_id) AS sclr_0
FROM wp_mailpoet_statistics_opens w0_ 
INNER JOIN wp_mailpoet_statistics_newsletters w1_ 
ON (w0_.newsletter_id = w1_.newsletter_id AND  w1_.subscriber_id = 1094 AND w1_.sent_at >= '2022-08-29T08:06:05.252426Z')
WHERE w0_.subscriber_id = 1094 AND (w0_.user_agent_type = 0)
```

costasovo (changes requested):
I hope I didn't miss anything but I think we need to improve the condition of the query and apply the same sentAt check also for clicks.
Please see [my comment in the thread](https://github.com/mailpoet/mailpoet/pull/5107#discussion_r1308418090).

mailpoet/lib/Subscribers/Statistics/SubscriberStatisticsRepository.php
johnolek: Sorry for completely overlooking the click counts. Of course we want to unify that behavior as well.

Regarding the query, perhaps a "small suggestion", but a very important one! Thank you very much for pointing that out!

Both of these issues should be addressed in 2d6808402a88d9b39277cd3848b437f20f5ff99c.

johnolek:
Thank you as always for your very careful review and detailed comments @costasovo! I believe this should be ready for one final review.

costasovo (approved):
@johnolek LGTM! Thank you for starting the discussion that led to improving the stats queries!

Aschepikov (approved):
GJ!

PR #5108, opened 2023-08-14
Backend changes to support apply segment on send [MAILPOET-5509]

johnolek:
## Description

This PR adds the backend changes necessary to support applying a single dynamic filter to a newsletter at send time.

## Code review notes

I refactored one query to use a QueryBuilder instead of building a string because I thought this would make it cleaner to modify dynamically. I would appreciate it if you could double check that I didn't alter the existing behavior in any way.

## QA notes

Make sure to deactivate and reactivate the plugin before testing. This is necessary to create new newsletter option field rows.

You should see 3 new rows in the `mailpoet_newsletter_option_fields` table, each one with a `name` of `filterSegmentId`.

image

Note these IDs and newsletter types for later.

Since there is currently no UI for adding a "filter segment", this will require manual database changes to test.

1. Create a dynamic segment and note its `id`. Make sure there is some overlap between this segment and the list(s) you plan on sending to, otherwise there won't be any matching subscribers.
2. Create a newsletter and schedule it (you can't send right away because we need to manually add a newsletter option to the DB). Note the newsletter id.
3. In the database using a tool like adminer from the mailpoet debug plugin, add a new row to the `mailpoet_newsletter_option` table:
    `id`: auto increment
    `newsletter_id`: newsletter id from step 2
    `value`: the `id` of the segment you want to use as the filter
    `option_field_id`: One of the three IDs from the `mailpoet_newsletter_option_fields` table mentioned above. Use the id corresponding to the newsletter type you're sending. 
4. After saving that row, you can go back into the newsletter editor and send it right away and the filter segment should be applied, i.e. it should only be sent to users from the list(s) you selected AND the filter segment. It should also save data about the filter segment to the sending queue's `meta` column in the database.

Note: if premium is disabled and you choose a filter segment that has more than one filter, you will see an error message about there not being any users to send to. More graceful handling of this state will be added in a future PR.

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5509

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Config/Populator.php
+++ b/mailpoet/lib/Config/Populator.php
@@ -490,6 +490,18 @@ class Populator {
         'name' => 'automationStepId',
         'newsletter_type' => NewsletterEntity::TYPE_AUTOMATION,
       ],
+      [
+        'name' => 'filterSegmentId',
+        'newsletter_type' => NewsletterEntity::TYPE_STANDARD,
+      ],
+      [
+        'name' => 'filterSegmentId',
+        'newsletter_type' => NewsletterEntity::TYPE_RE_ENGAGEMENT,
+      ],
+      [
+        'name' => 'filterSegmentId',
+        'newsletter_type' => NewsletterEntity::TYPE_NOTIFICATION,
+      ],
     ];
 
     return [
--- a/mailpoet/lib/Config/Populator.php
+++ b/mailpoet/lib/Config/Populator.php
@@ -13,6 +13,7 @@ use MailPoet\Cron\Workers\SubscriberLinkTokens;
 use MailPoet\Cron\Workers\SubscribersLastEngagement;
 use MailPoet\Cron\Workers\UnsubscribeTokens;
 use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Entities\NewsletterOptionFieldEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\StatisticsFormEntity;
@@ -491,15 +492,15 @@ class Populator {
         'newsletter_type' => NewsletterEntity::TYPE_AUTOMATION,
       ],
       [
-        'name' => 'filterSegmentId',
+        'name' => NewsletterOptionFieldEntity::NAME_FILTER_SEGMENT_ID,
         'newsletter_type' => NewsletterEntity::TYPE_STANDARD,
       ],
       [
-        'name' => 'filterSegmentId',
+        'name' => NewsletterOptionFieldEntity::NAME_FILTER_SEGMENT_ID,
         'newsletter_type' => NewsletterEntity::TYPE_RE_ENGAGEMENT,
       ],
       [
-        'name' => 'filterSegmentId',
+        'name' => NewsletterOptionFieldEntity::NAME_FILTER_SEGMENT_ID,
         'newsletter_type' => NewsletterEntity::TYPE_NOTIFICATION,
       ],
     ];
--- a/mailpoet/lib/Segments/SegmentsRepository.php
+++ b/mailpoet/lib/Segments/SegmentsRepository.php
@@ -61,7 +61,7 @@ class SegmentsRepository extends Repository {
    * @return SegmentEntity[]
    */
   public function findByTypeNotIn(array $types): array {
-      return $this->doctrineRepository->createQueryBuilder('s')
+    return $this->doctrineRepository->createQueryBuilder('s')
       ->select('s')
       ->where('s.type NOT IN (:types)')
       ->setParameter('types', $types)
--- a/mailpoet/lib/Segments/SegmentsRepository.php
+++ b/mailpoet/lib/Segments/SegmentsRepository.php
@@ -11,6 +11,8 @@ use MailPoet\Entities\NewsletterSegmentEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SubscriberSegmentEntity;
 use MailPoet\Form\FormsRepository;
+use MailPoet\InvalidStateException;
+use MailPoet\Logging\LoggerFactory;
 use MailPoet\Newsletter\Segment\NewsletterSegmentRepository;
 use MailPoet\NotFoundException;
 use MailPoet\WP\Functions as WPFunctions;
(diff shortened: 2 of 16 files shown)


mailpoet/lib/Config/Populator.php
```diff
@@ -490,6 +490,18 @@ protected function newsletterOptionFields() {
         'name' => 'automationStepId',
         'newsletter_type' => NewsletterEntity::TYPE_AUTOMATION,
       ],
+      [
+        'name' => 'filterSegmentId',
```
rodrigoprimo: Have you considered using `NewsletterOptionFieldEntity::NAME_FILTER_SEGMENT_ID` here instead of hard-coding `filterSegmentId`?

mailpoet/lib/Segments/SubscribersFinder.php
```diff
@@ -42,6 +43,18 @@ public function findSubscribersInSegments($subscribersToProcessIds, $newsletterS
       }
       $result = array_merge($result, $this->findSubscribersInSegment($segment, $subscribersToProcessIds));
     }
+
+    if (is_int($filterSegmentId)) {
+      $filterSegment = $this->segmentsRepository->findOneById($filterSegmentId);
+      if ($filterSegment instanceof SegmentEntity) {
+        if ($filterSegment->getType() !== SegmentEntity::TYPE_DYNAMIC) {
+          throw new UnexpectedValueException("Filter segment ID must be a dynamic segment. Type of filter with ID {$filterSegment->getId()} is {$filterSegment->getType()}.");
```
rodrigoprimo: I did a quick test, and it seems to me that if we happen to have a non-dynamic segment, the interface acts as if the newsletter is being sent, but actually, it is stuck with the error above. The user is not able to send other emails. Should we maybe display this error when the user tries to send the newsletter instead of acting as if the newsletter is being sent?

Currently, this error does appear as a cron error in the system status page.

I'm not sure if we want to implement what I'm suggesting and if it is easy to do it. Just mentioning to make sure we are aware of the current behavior.

rodrigoprimo (commented):
Looks good to me. I left two comments for us to talk about.

> I refactored one query to use a QueryBuilder instead of building a string because I thought this would make it cleaner to modify dynamically. I would appreciate it if you could double check that I didn't alter the existing behavior in any way.

I compared both queries, and they seem to do the exactly same thing as far as I can see.me.

mailpoet/lib/Segments/SubscribersFinder.php
johnolek: Thanks for pointing this out! It definitely seems like something we shouldn't allow. I'll try to come up with a better solution.

mailpoet/lib/Config/Populator.php
johnolek: Thanks for bringing this up. I considered it but ended up doing it this way becanse it was consistent with the surrounding code. It definitely seems better to use the constant so I updated this in `12b90c48c100c0ede37eaa310d93ca3f40dfa3ab`.

(sorry for inconsistent commits, I rebased on `trunk`).

rodrigoprimo:
@johnolek, just checking if I should check this PR already or not? Asking as it is still assigned to you so I'm not sure if you still plan to do something before a final review.

johnolek:
@rodrigoprimo Sorry for the delay on this one and thanks for checking in. I'm still finalizing the changes and expect to push a final version today.

rodrigoprimo:
No worries, and there is no rush. I was just checking in case you were waiting for a review and I was waiting for you to assign the PR back to me.

mailpoet/lib/Segments/SubscribersFinder.php
johnolek: I addressed this issue in 39c1ac1c449cfce3e1372828da0ba64495c738f4. It's still possible to end up in a kind of strange state when a newsletter has a filter segment that belongs to a static list, where it completes but sends to nobody. The same state seems possible without this change by scheduling an email to send to a list and then removing all subscribers from that list. Since this very much an edge case and we'll do our best to prevent saving a list ID as a filter segment ID, I don't know that this warrants a more elegant solution.

Please let me know if you can think of a better way of handling this, or if you're still able to end up in a broken sending state.

mailpoet/lib/Segments/SegmentsRepository.php
```diff
@@ -54,7 +61,7 @@ protected function getEntityClassName() {
    * @return SegmentEntity[]
    */
   public function findByTypeNotIn(array $types): array {
-    return $this->doctrineRepository->createQueryBuilder('s')
+      return $this->doctrineRepository->createQueryBuilder('s')
```
rodrigoprimo: I guess we need to fix the indentation here. No idea why PHPCS didn't catch this. Maybe because of our custom ruleset. 

```suggestion
    return $this->doctrineRepository->createQueryBuilder('s')
```

mailpoet/lib/Segments/SegmentsRepository.php
```diff
@@ -136,6 +143,28 @@ public function verifyNameIsUnique(string $name, ?int $id): void {
     }
   }
 
+  /**
+   * @param int $id
+   *
+   * @return SegmentEntity
+   * @throws InvalidStateException
+   */
+  public function verifyFilterSegmentExists(int $id): SegmentEntity {
```
rodrigoprimo: I wonder if we should change the name of this method to indicate that it also checks if the segment is dynamic since it throws an exception if another type of segment is passed?

rodrigoprimo (commented):
Thanks for the changes, @johnolek. I left a few final comments before we move this PR to QA.

mailpoet/lib/Segments/SubscribersFinder.php
rodrigoprimo: The approach that you implemented looks good to me and I agree that there is no need to cover this other edge case that you described.

mailpoet/lib/Segments/SegmentsRepository.php
johnolek: Fixed in a7fed3f26d5252f6f5cd19762f8169c7aee87824

mailpoet/lib/Segments/SegmentsRepository.php
johnolek: Very good point! I updated this in 1a5d6d1b5d5977e3a3a57c6fa53ad4ba5b66ff79

johnolek:
Thanks for the attention to detail, @rodrigoprimo! I think this is ready for one last review.

rodrigoprimo (approved):

Aschepikov (approved):
GJ!

johnolek:
commenting to see if this updates the status in Jira

PR #5109, opened 2023-08-15
Calculate subscribers who are subscribed to a list correcty [MAILPOET-4487]

websupporter:
## Description
This PR takes into account that a user can be globally subscribed but unsubscribed to a certain segment. If you go to the subscriber listings and you selected e.g. to view only the subscribed users of a certain segment, you would see users who are unsubscribed to the specific list while globally subscribed. These users are now moved to the "Unsubscribed" tab and the numbers are calculated accordingly.

## Code review notes
Instead of changing the signature of the methods, I opted for setting a `private $definition` which gets populated on the public methods and can be used subsequently. My thinking here was that this contains the change and leaves the interface intact. I am open to changing the signatures of the methods instead. I think there are also good reasons to do so.

The main change happens in `filterGroup()`. If a static segment is used to filter the data and your group (status of the subscriber) is either `subscribed` or `unsubscribed` this PR takes now the local status of the subscription into account as well. From now on in the listing `subscribed` means you are globally subscribed _and_ subscribed to the specific segment. Unsubscribed means you are either unsubscribed from the specific segment or globally unsubscribed.

This PR gets also rid of the method `getSegmentGlobalStatusStatisticsCount()`. This was used to calculate number of bounces, subscribed, unsubscribed etc. users for a specific list and took only the global status into account. The method was solely used to calculate those numbers for the select fields in the subscriber list view, where you are able to switch between segments. The existing `getSegmentStatisticsCount()` can be used instead and produces the right numbers.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4487]

## Tasks

[MAILPOET-4487]: https://mailpoet.atlassian.net/browse/MAILPOET-4487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Settings.php
+++ b/mailpoet/lib/API/JSON/v1/Settings.php
@@ -436,9 +436,6 @@ class Settings extends APIEndpoint {
     $segments = $this->segmentsRepository->findAll();
     foreach ($segments as $segment) {
       $this->subscribersCountsController->recalculateSegmentStatisticsCache($segment);
-      if ($segment->isStatic()) {
-        $this->subscribersCountsController->recalculateSegmentGlobalStatusStatisticsCache($segment);
-      }
     }
     $this->subscribersCountsController->recalculateSubscribersWithoutSegmentStatisticsCache();
     // remove redundancies from cache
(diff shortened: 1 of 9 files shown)


rodrigoprimo (approved):
Good job!

Adding the `$definition` as a private property of the class seems like a good approach to me. Thanks for cleaning up the cache generation as well.

Aschepikov (approved):
GJ!

PR #5119, opened 2023-08-22
Fix flaky FilterHandlerTest [MAILPOET-5544]

JanJakes:
## Description

This test fails especially in `integration_oldest` tests on nightly builds, like [this one](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/15295/workflows/28d6343f-6ba7-4be9-9e7e-a272d4e7a397/jobs/259474).

One thing to maybe verify further, would be to check all usages of `FilterHandler::apply(...)` and check whether the `$queryBuilder` passed into them has some sort, so we don't end up with unpredictable order in the UI as well. As you're working on segments now, @johnolek, would you please help me verify these? I think the usages are:
1. [SegmentsFinder](https://github.com/mailpoet/mailpoet/blob/27a1602645e0dea545017080b6aa8c4671d8c668/mailpoet/lib/Segments/SegmentsFinder.php#L59) — seems to be missing `orderBy`, but fetches only one subscriber at a time, so I think OK ✅ 
2. [ImportExportRepository](https://github.com/mailpoet/mailpoet/blob/27a1602645e0dea545017080b6aa8c4671d8c668/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php#L211) — [looks OK](https://github.com/mailpoet/mailpoet/blob/27a1602645e0dea545017080b6aa8c4671d8c668/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php#L238) to me ✅ 
3. [SegmentSubscribersRepository](https://github.com/mailpoet/mailpoet/blob/27a1602645e0dea545017080b6aa8c4671d8c668/mailpoet/lib/Segments/SegmentSubscribersRepository.php#L399) — the call stack is a bit more complex, but I see no ordering ❗ 
4. [SubscriberListingRepository](https://github.com/mailpoet/mailpoet/blob/27a1602645e0dea545017080b6aa8c4671d8c668/mailpoet/lib/Subscribers/SubscriberListingRepository.php#L397) — this does seem to [apply some sorting](https://github.com/mailpoet/mailpoet/blob/27a1602645e0dea545017080b6aa8c4671d8c668/mailpoet/lib/Subscribers/SubscriberListingRepository.php#L187), but not sure if in all paths ❓ 

## Code review notes

No QA needed, just tests.

## QA notes

No QA needed, just tests.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5544]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5544]: https://mailpoet.atlassian.net/browse/MAILPOET-5544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/integration/Segments/DynamicSegments/FilterHandlerTest.php
+++ b/mailpoet/tests/integration/Segments/DynamicSegments/FilterHandlerTest.php
@@ -77,7 +77,8 @@ class FilterHandlerTest extends \MailPoetTest {
       ->getConnection()
       ->createQueryBuilder()
       ->select("$subscribersTable.id")
-      ->from($subscribersTable);
+      ->from($subscribersTable)
+      ->orderBy("$subscribersTable.id", 'ASC');
   }
 
   public function _after(): void {


johnolek (approved):
Approving and merging this to get tests working again. I'll reply to your other comment separately. Thanks for the fix!

johnolek:
@JanJakes Thanks for bringing up the topic of ordering!

I agree about the first two cases.

> [SegmentSubscribersRepository](https://github.com/mailpoet/mailpoet/blob/27a1602645e0dea545017080b6aa8c4671d8c668/mailpoet/lib/Segments/SegmentSubscribersRepository.php#L399) — the call stack is a bit more complex, but I see no ordering

I don't see any ordering either. I'm not concerned about this one because I don't see any usages where order would make a difference, unless it matters the order that we [add subscribers to a sending task](https://github.com/mailpoet/mailpoet/blob/00da2bd848be4f18b3ee5a4b51f24cb60f40f962/mailpoet/lib/Segments/SubscribersFinder.php#L145-L145). I would be ok with applying an order by here just to be on the safe side.

For [SubscriberListingRepository](https://github.com/mailpoet/mailpoet/blob/27a1602645e0dea545017080b6aa8c4671d8c668/mailpoet/lib/Subscribers/SubscriberListingRepository.php#L397), the two places where we don't have ordering is for a [count](https://github.com/mailpoet/mailpoet/blob/00da2bd848be4f18b3ee5a4b51f24cb60f40f962/mailpoet/lib/Subscribers/SubscriberListingRepository.php#L81-L81), where order definitely shouldn't matter, and for [retrieving all of the subscriber IDs](https://github.com/mailpoet/mailpoet/blob/00da2bd848be4f18b3ee5a4b51f24cb60f40f962/mailpoet/lib/Subscribers/SubscriberListingRepository.php#L96-L100) for a segment when someone selects the "all" option for a bulk operation when filtered by segment (that's a mouthful), which I think also shouldn't need ordering. We could apply a general order by and then let other code [override it](https://github.com/mailpoet/mailpoet/blob/00da2bd848be4f18b3ee5a4b51f24cb60f40f962/mailpoet/lib/Subscribers/SubscriberListingRepository.php#L369-L369) as needed, but it doesn't feel necessary to me. 

The only other usages we could consider changing are [in the premium plugin](https://github.com/mailpoet/mailpoet-premium/blob/fa546fb5a7c14d83354c7517cc9a5bdcc41ce0ae/tests/unit/Segments/DynamicSegments/FilterHandlerTest.php#L165-L165), but all of those tests are only looking at counts, not specific emails in a specific order, and the querybuilder instance comes from a private method in the test.

I'm guessing we saw these differences in the `_oldest` tests because of the older version of MySQL, or using MyISAM as the storage engine, and that newer versions/InnoDB tend to be more predictable, even though there's still no guarantee about order. Does that sound right to you?

In any case, what do you think about adding a check to `FilterHandler::apply` itself? Right before we [return the queryBuilder](https://github.com/mailpoet/mailpoet/blob/00da2bd848be4f18b3ee5a4b51f24cb60f40f962/mailpoet/lib/Segments/DynamicSegments/FilterHandler.php#L63-L63) we could check for `$queryBuilder->getQueryParts()['orderBy']`, and if it's empty add a default to order by subscriber ID ascending. I'm guessing this is what's happening in most cases anyway when we leave out the order by, we'll just be ensuring it, especially for older versions. This way we wouldn't have to worry about checking complex call stacks in the future. There's probably a reason it's not a good idea though!

PR #5123, opened 2023-08-24
Woocommerce revenue tracking improvements [MAILPOET-5485]

websupporter:
## Description

_N/A_

## Code review notes
Instead of hooking into special status changes, this PR suggest to hook into every status change. It extends the purchase entity and adds a `status` field. As soon as an order is created a purchase entity is created with the according status. When the status gets updated, the purchase entity gets updated as well.

We count as revenue only `completed`. So instead of storing just `completed` revenues, we filter the output for returning only `completed` purchase entities and counting them as "revenue". While this creates more data we store, it also opens up possibilities like tracking how many orders `failed` are `on-hold` etc. which might be interesting metrics as well.

## QA notes

We touched the revenue tracking code. Please verify order revenue is tracked and displayed in newsletter stats as expected. 
1) Create a newsletter and send it to a list
2) Make sure to click on links from a couple of received emails
3) Make a couple of orders and use emails from subscribers who clicked the links
4) Check the newsletter stats and observe that revenue was added
5) Try refunding an order and see change in stats

NOTE: For testing you will need premium from the branch same name (https://github.com/mailpoet/mailpoet-premium/pull/777)

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/777

## Linked tickets

[MAILPOET-5485]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5485]: https://mailpoet.atlassian.net/browse/MAILPOET-5485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Migrations/Db/Migration_20230824_054259_Db.php
+++ b/mailpoet/lib/Migrations/Db/Migration_20230824_054259_Db.php
@@ -1,0 +1,65 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Migrations\Db;
+
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\Entities\StatisticsWooCommercePurchaseEntity;
+use MailPoet\Migrator\DbMigration;
+use MailPoet\WooCommerce\Helper;
+
+class Migration_20230824_054259_Db extends DbMigration {
+
+  /** @var Helper */
+  private $wooCommerceHelper;
+
+  const DEFAULT_STATUS = 'unknown';
+
+  public function __construct(
+    ContainerWrapper $container
+  ) {
+    parent::__construct($container);
+    $this->wooCommerceHelper = $container->get(Helper::class);
+  }
+
+  public function run(): void {
+    $this->createStatusColumn();
+    $this->populateStatusColumn();
+  }
+
+  private function createStatusColumn(): void {
+    $revenueTable = $this->getTableName(StatisticsWooCommercePurchaseEntity::class);
+
+    if ($this->columnExists($revenueTable, 'status')) {
+      return;
+    }
+
+    $this->connection->executeQuery(
+      "ALTER TABLE `" . $revenueTable . "`
+        ADD COLUMN `status` VARCHAR(40) NOT NULL DEFAULT '" . self::DEFAULT_STATUS . "',
+        ADD INDEX `status` (`status`)"
+    );
+  }
+
+  private function populateStatusColumn(): void {
+
+    $this->wooCommerceHelper->isWooCommerceCustomOrdersTableEnabled() ?
+      $this->populateStatusColumnUsingHpos() : $this->populateStatusColumnUsingPost();
+  }
+
+  private function populateStatusColumnUsingHpos(): void {
+    global $wpdb;
+    $revenueTable = $this->getTableName(StatisticsWooCommercePurchaseEntity::class);
+
+    $sql = 'update ' . $revenueTable . ' as rev, ' . $wpdb->prefix . 'wc_orders as wc set rev.status=TRIM(Leading "wc-" FROM wc.status) where wc.id = rev.order_id AND rev.status="' . self::DEFAULT_STATUS . '"';
+    $this->connection->executeQuery($sql);
+  }
+
+  private function populateStatusColumnUsingPost(): void {
+
+    global $wpdb;
+    $revenueTable = $this->getTableName(StatisticsWooCommercePurchaseEntity::class);
+
+    $sql = 'update ' . $revenueTable . ' as rev, ' . $wpdb->posts . ' as wc set rev.status=TRIM(Leading "wc-" FROM wc.post_status) where wc.id = rev.order_id AND rev.status="' . self::DEFAULT_STATUS . '"';
+    $this->connection->executeQuery($sql);
+  }
+}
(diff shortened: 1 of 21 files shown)


mailpoet/lib/Migrations/Db/Migration_20230824_054259_Db.php
```diff
@@ -0,0 +1,65 @@
...
+  private $wooCommerceHelper;
+
+  const DEFAULT_STATUS = 'unknown';
+
+  public function __construct(
+    ContainerWrapper $container
+  ) {
+    parent::__construct($container);
+    $this->wooCommerceHelper = $container->get(Helper::class);
+  }
```
costasovo: We should avoid using containers in DB-level migrations. WooCommerce helper doesn't work with Doctrine so it safe for now but may change.
I would suggest splitting the migration into two:
1) DB Level migration that creates the new column
2) App Level migration that fills the column.
The App-level migrations always run after the DB level so the column will be there and in the App-level migration, the container is available by default.

costasovo (changes requested):
@websupporter I like the idea of storing the state 👍 Thank you for thinking beyond the specification! Also thank you for the nice commits and improving test coverage! Great job!

When testing the tracking locally, I noticed that when I add a refund to an order and I don't change its status, the record in our table is not updated. It seems to be related to hooking only into status changes. Maybe we need to listen to general order updates or add listening for refunds.

mailpoet/lib/Migrations/Db/Migration_20230824_054259_Db.php
websupporter: Ah! That makes sense. Didn't fully grasp the App/Db difference and how I should apply that. Thank you: 80c96c4

websupporter:
Thank you for your review @costasovo. Much appreciated! 

> When testing the tracking locally, I noticed that when I add a refund to an order and I don't change its status, the record in our table is not updated. It seems to be related to hooking only into status changes. Maybe we need to listen to general order updates or add listening for refunds.

It looks like you did a partial refund, which does not change the status. A fully refund should change the status:
https://github.com/woocommerce/woocommerce/blob/879136404d3b5445c1f1fb81c5e179d5333c96f4/plugins/woocommerce/includes/wc-order-functions.php#L650-L659

Nonetheless, we did not had a logic to adjust the revenue we generate from an order, when it got partially refunded, and actually not even when it got fully refunded as `$order->get_total()` will not change with a refund. `$order->get_remaining_refund_amount()` returns the amount you can refund, which is basically a calculation `$order->get_total() - $order->get_refunded_amount()`. I did not see any filters or some logic which could get into the way and think, if you really want to get how many revenue an order made, you need to use `get_remaining_refund_amount()`.

45fe6b3 adds a test and 66b666f passes this test.

costasovo (dismissed):
@websupporter Thanks for the fix and explanation! 👍 

When testing I missed that value didn't change even though I switched the status.

Aschepikov:
Hi @websupporter some checks have failed, can you take a look pls.

websupporter:
@costasovo I needed to do a small code change. For some reason where `rev.status="unknown"` produces an SQL error. When I switch to `rev.status='unknown'` it works. So I added 9f15b4916182b209a88f3b3623bbf8657d0f3b2f and switched the quote types.

costasovo (dismissed):
@websupporter Thanks for the fix! The `'` vs `"` issue could be related to [ANSI_QUOTES](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes) mode.
[WPDB disables a bunch of incompatible modes](https://github.com/WordPress/WordPress/blob/21011a6ebb3080024d867dfdd6721b6e87fce6ae/wp-includes/class-wpdb.php#L645-L652) but it doesn't touch `ANSI_QUOTES`.
Good that you caught this issue. It could cause failing migrations for multiple users.
We must remember to use single quotes when working with WPDB. Could we catch this by a linter?

Aschepikov:
Hi @websupporter  I tested revenue stats it is working well, but I don't see my orders in the email stats. I tested on the 2 sites because I thought it could be connected to some site´s problem
image
image
But on both sites, I see that the order is connected to the email but not in the email stats. 
image
In this case, these orders should be in the email stats yes?

websupporter:
@Aschepikov good catch! Thank you!

Basically, it appeared to early in the subscriber statistics. It should appear, once the order has the status "completed". I updated the code.

Aschepikov:
@websupporter Thank you! Some checks are failing. I restarted it just in case they are flaky.

websupporter:
Thank you @Aschepikov. Build is green now.

Aschepikov:
fui on the second site, everything is ok

Aschepikov:
Oh on the second site, the completed status is working on the stats!

Aschepikov:
On this site completed status isn't tracked in the statistics, and it was completed before the last changes. I will retest everything on the 3 site to check that the old and new orders are displayed correctly 
image

Aschepikov:
Also on the 1 site, new orders are tracking incorrectly. The order was made from the link in the 5123 email but is tracking as an order from another email. I will comment it here maybe it is connected with something else. I will try to reproduce it on another site
image

Aschepikov:
Hi @websupporter can you take a look at this error from the third site? If it is not connected with the pr I will approve it as it worked okay on one of my sites.

The third site was broken after the installation of the pr. On the clean site, I installed the 4.26 prod version, prepared a few emails and orders.
On this site, the order is displayed incorrectly too, it should be connected just with the 5123 email 3.
image
After the installation of plugins prs, I have these errors. I had some errors on the previous sites but they were gone after the cache cleaning. 
image
image

Aschepikov (approved):
It will be tested one more time on a new site with the release build

PR #5127, opened 2023-08-25
Update NPM packages  [MAILPOET-4391]

johnolek:
## Description

I made this PR in favor of #5077, which had a less than ideal commit history and used a bash script instead of a robo script, but had good comments that I didn't want to get lost.

This PR updates all of the NPM dependencies that were able to be updated without causing any obvious issues or requiring refactoring.

The last time we bulk updated NPM dependencies was, I believe, https://github.com/mailpoet/mailpoet/pull/4033

This does not include any @wordpress/* packages, as we want to update those separately.

These packages will be done later, as they either caused a problem or would require some refactoring:

@babel/preset-env
babel-loader
backbone
backbone.marionette
codemirror
history
react-router-dom
react-string-replace
react-tooltip
stylelint
tinymce
fork-ts-checker-webpack-plugin (not actually sure about this one)

## Code review notes

Running `pnpm dedupe` makes the JS newsletter tests ( `./do t:newsletter-editor test-results/mocha/newsletter_editor_junit.xml`) start failing. 

```
SyntaxError: /mailpoet/mailpoet/assets/js/src/newsletter_editor/components/save.js: Unexpected token (50:46)
  48 |       delete json.body.blockDefaults.woocommerceContent;
  49 |       if (json.body.blockDefaults && json.body.blockDefaults.coupon) {
> 50 |         delete json.body.blockDefaults.coupon?.couponId;
     |                                               ^
  51 |         delete json.body.blockDefaults.coupon?.code;
  52 |       }
  53 |     }
```

This could easily be fixed by changing the code, but I wasn't able to figure out why running dedupe would cause this to suddenly start happening, so this is still in draft status because I haven't run dedupe.

## QA notes

I'm not 100% sure what the standard procedure is for updates like this. Anything that uses JavaScript could potentially be affected. We should check that there are no new console warnings/errors or unexpected behavior with any JS related functionality.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4391]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4391]: https://mailpoet.atlassian.net/browse/MAILPOET-4391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


johnolek:
@JanJakes Sorry if it's confusing to open a different PR for this, but I didn't want to lose the history from the original PR.

I like this solution better than what I had before with the bash script but there still seem to be a few things that aren't quite working as expected, and I would really appreciate your help. If you would like to go ahead and push any changes to this branch I wouldn't be offended in the slightest, otherwise if you have any advice on how to make this as good as possible for now I'd appreciate it.

mailpoet/RoboFile.php
```diff
@@ -43,6 +49,119 @@ public function update() {
...
+      exit(1);
+    }
+
+    $outdatedPackagesOutput = $this->taskExec('pnpm outdated --no-table')
+      ->printOutput(false)
+      ->run()
+      ->getMessage();
+    $lines = explode("\n", $outdatedPackagesOutput);
+
+    // The package names themselves are every third line
```
JanJakes, line 70: Maybe the format was meant to be split on `\n\n`? That said, your approach seems to work well...

JanJakes (commented):
@johnolek I somehow missed that this was already for review, sorry about that.

Looks great, I don't have any more suggestions, but we need to solve the failures.

> Running pnpm dedupe makes the JS newsletter tests ( ./do t:newsletter-editor test-results/mocha/newsletter_editor_junit.xml) start failing.

This is in the old JS code, where I'm not sure if we ever had the support for the `?.` operator configured. These are the only two places in our `.js` files, where we're using that, in `.ts`, `.tsx`, and `.jsx` files, this seems to cause no issues.

Thinking further, this could be related to the `testConfig` in our `webpack.config.js`. It should inherit the `baseConfig` with `babel-loader`, though, so I'm not sure here, how the error happens...

Although... is that a runtime error? Does that mean that it's running on such a version of Node.js that doesn't support optional chaining?

johnolek:
@JanJakes Thanks for taking another look, @JanJakes. Unfortunately I'm at a loss as to what's causing the problems and how to move forward. Removing the `?.` leaves these out of memory problems that I'm not sure how to debug: https://app.circleci.com/pipelines/github/mailpoet/mailpoet/15389/workflows/4dce7f1f-b63e-4f9a-a536-baf37ca3f7bb/jobs/261098

In case it's helpful to see the full error, here's a link to a build that complained about the operator: https://app.circleci.com/pipelines/github/mailpoet/mailpoet/15388/workflows/f9a0fc52-ec2d-4e71-a5f8-d6f5a65d3b81/jobs/261078

Could it have something to do with these custom injectors we're using, like this one in the save spec? https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/javascript_newsletter_editor/newsletter_editor/components/save.spec.js#L7-L7

I also just noticed there are a few tests in the `test:newsletters` that simply fail when I am able to run them with the latest dependencies:

```
 420 passing (2s)
  31 pending
  3 failing

  1) Divider
       settings view
         once rendered
           changes color of available divider styles when actual divider color changes:

      #889912
      + expected - actual

      -rgb(136, 153, 18)
      +#889912
      
  

  2) Spacer
       block view
         renders:

      AssertionError: expected 'rgba(0, 0, 0, 0)' to equal 'transparent'
      + expected - actual

      -rgba(0, 0, 0, 0)
      +transparent
      
      at Context.eval (webpack-internal:///./tests/javascript_newsletter_editor/newsletter_editor/blocks/spacer.spec.js:89:69)
      at process.processImmediate (node:internal/timers:478:21)

  3) Spacer
       block view
         rerenders if model attributes change:

      AssertionError: expected '40px' to equal '71px'
      + expected - actual

      -40px
      +71px
      
      at Context.eval (webpack-internal:///./tests/javascript_newsletter_editor/newsletter_editor/blocks/spacer.spec.js:95:59)
      at process.processImmediate (node:internal/timers:478:21)



 [Exec]  Exit code 3  Time 2.786s
 2/2 [============================] 100%
```

any idea what could be causing those problems? If you don't have any ideas I can try doing some `git bisect`ing to pinpoint the problem update.

costasovo:
@johnolek , @JanJakes The out of memory problems were possibly fixed by this PR https://github.com/mailpoet/mailpoet/pull/5143

I've rebased this PR and tried to look into the JS Test errors in the last 30 minutes of my Janitor week, but I wasn't able to fix them. I think that the two hex vs rgba colors failures might be caused by the updated `jsdom` library. My theory is that it started returning computed CSS colors in RGB format so I guess we could just alter asserts to check rgb value.
For the third test, I have no idea. I tested the behavior in the editor app and setting height for spacer seems to work ok.

@JanJakes Maybe as the next Janitor you could finish this PR. I also wanted to add stylelint update, remove deprecated stylelint rules and enable prettier for SCSS.

costasovo (approved):
Good job! Great to see we have the dependencies up to date! 👍

Aschepikov (approved):
GJ!

PR #5132, opened 2023-08-31
Reset subscribers without resubscribing [MAILPOET-5555]

websupporter:
## Description
When you edit a subscriber and click save, we would resubscribe the subscriber to all selected lists and thus the subscriber would re-enter "Someone subscribed"-Automations. This PR stops actively resubscribing subscribers.

## Code review notes

In my tests I used `SubscriberEntity::getSubscriberSegments` to get the segments. This method accepted a status parameter, but did not use it's value. I think this was a bug, which I fixed in 8c2ce511faef2bc14b1e9a538090ce55c22644c9

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5555]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5555]: https://mailpoet.atlassian.net/browse/MAILPOET-5555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscribers/SubscriberSegmentRepository.php
+++ b/mailpoet/lib/Subscribers/SubscriberSegmentRepository.php
@@ -81,6 +81,7 @@ class SubscriberSegmentRepository extends Repository {
   }
 
   public function resetSubscriptions(SubscriberEntity $subscriber, array $segments): void {
+    // Already existing subscriptions are stored in $existingSegments. Their IDs in $existingSegmentIds.
     $existingSegments = array_values(array_filter(array_map(
       function(SubscriberSegmentEntity $subscriberSegmentEntity): ?SegmentEntity {
         return $subscriberSegmentEntity->getSegment();
@@ -93,23 +94,33 @@ class SubscriberSegmentRepository extends Repository {
       },
       $existingSegments
     );
+
+    // $segmentIds are the IDs of the segments we want the user to be subscribed to.
     $segmentIds = array_map(
       function(SegmentEntity $segment): int {
         return $segment->getId() ?? 0;
       },
       $segments
     );
+
+    // $unsubscribedSegments are the segment IDs to which we need to unsubscribe.
     $unsubscribedSegments = array_diff($existingSegmentIds, $segmentIds);
+
+    // $newlySubscribedSegments are the segment IDs to which we need to newly subscribe.
     $newlySubscribedSegments = array_diff($segmentIds, $existingSegmentIds);
     if (!$newlySubscribedSegments && !$unsubscribedSegments) {
       return;
     }
+
+    // The segments we need to unsubscribe.
     $unsubscribe = array_filter(
       $existingSegments,
       function(SegmentEntity $segment) use ($unsubscribedSegments): bool {
         return in_array($segment->getId(), $unsubscribedSegments);
       }
     );
+
+    // The segments we need to newly subscribe.
     $subscribe = array_filter(
       $segments,
       function(SegmentEntity $segment) use ($newlySubscribedSegments): bool {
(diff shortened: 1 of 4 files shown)


mailpoet/tests/integration/Subscribers/SubscriberSegmentRepositoryTest.php
```diff
@@ -0,0 +1,82 @@
...
+    $segment4 = (new Segment())->create();
+
+    $this->testee->resetSubscriptions($subscriber, [$segment1, $segment2, $segment3, $segment4]);
+    $subscribedSegments = $subscriber->getSubscriberSegments(SubscriberEntity::STATUS_SUBSCRIBED);
+    $unsubscribedSegments = $subscriber->getSubscriberSegments(SubscriberEntity::STATUS_UNSUBSCRIBED);
+    $subscribedSegmentIds = array_map(
+      function(SubscriberSegmentEntity $entity): ?int {
+        return $entity->getSegment() ? $entity->getSegment()->getId() : null;
+      },
+      $subscribedSegments->toArray()
```
lysyjan: This could be simplified by calling [the map method](https://www.doctrine-project.org/projects/doctrine-collections/en/stable/index.html#map) on the collection. Nevertheless, it is not a big deal, and I consider it optional.

What I would like to change here is to avoid code repetition because there is identical code three times. We could create a private method in this test class or a public method in the SubscriberEntity class.

mailpoet/lib/Subscribers/SubscriberSegmentRepository.php
```diff
@@ -81,8 +81,34 @@ public function unsubscribeFromSegments(SubscriberEntity $subscriber, array $seg
...
 
   public function resetSubscriptions(SubscriberEntity $subscriber, array $segments): void {
-    $this->unsubscribeFromSegments($subscriber);
-    $this->subscribeToSegments($subscriber, $segments);
+    $existingSegments = array_values(array_filter(array_map(
+      function(SubscriberSegmentEntity $subscriberSegmentEntity): ?SegmentEntity { return $subscriberSegmentEntity->getSegment(); 
+      },
+      $this->findBy(['subscriber' => $subscriber, 'status' => SubscriberEntity::STATUS_SUBSCRIBED])
+    )));
+    $existingSegmentIds = array_map(function(SegmentEntity $segment): int { return $segment->getId() ?? 0;
```
lysyjan: The code formatting here is weird. Can you check it, please? There is an empty line below, so I think the return statement should be at the line. You can find more similar cases in this method.

mailpoet/lib/Subscribers/SubscriberSegmentRepository.php
```diff
@@ -81,8 +81,34 @@ public function unsubscribeFromSegments(SubscriberEntity $subscriber, array $seg
   }
 
   public function resetSubscriptions(SubscriberEntity $subscriber, array $segments): void {
-    $this->unsubscribeFromSegments($subscriber);
-    $this->subscribeToSegments($subscriber, $segments);
+    $existingSegments = array_values(array_filter(array_map(
```
lysyjan, line 85: Please, could you add some comments here? It could help to better and faster understand how it works.

lysyjan (commented):
I have minor remarks about those changes. It's nothing big, I hope.
Thanks for quickly fixing the found bug.

mailpoet/tests/integration/Subscribers/SubscriberSegmentRepositoryTest.php
websupporter: 3cce6bca1

mailpoet/lib/Subscribers/SubscriberSegmentRepository.php
websupporter: 54872c01b

mailpoet/lib/Subscribers/SubscriberSegmentRepository.php
websupporter: a19774de7

websupporter:
Thanks for your review @lysyjan. I addressed the noted issues.

lysyjan (approved):
👍

Aschepikov (approved):
GJ!

PR #5141, opened 2023-09-04
Add filter segment select to send step [MAILPOET-5510] && [MAILPOET-5512]

johnolek:
## Description

This PR adds the ability to select a filter segment on the send step for standard and post notification emails. I also included MAILPOET-5512 to allow the creation of new segments from the send step.

## Code review notes

It might be nice to update some of the existing class components to functional components and definitely update them to use typescript, but considering some of their complexity I think this would be better left to a separate task.

Please let me know if you can think of simpler ways of handling things. For example, using a context for `saveDraftNewsletter` to accommodate the redirect to the new segment page seems complex, but I couldn't come up with an alternate method that wouldn't require fundamental changes to the forms code that might have unintended side effects elsewhere in the app.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5510] and [MAILPOET-5512]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5510]: https://mailpoet.atlassian.net/browse/MAILPOET-5510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5512]: https://mailpoet.atlassian.net/browse/MAILPOET-5512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


johnolek:
@rodrigoprimo I'm AFK all next week so no rush in taking a look at this.

rodrigoprimo (commented):
A few acceptance tests are failing in this PR. Could you please take a look?

johnolek:
@rodrigoprimo I fixed the issue with the tests and I believe this is ready to review.

mailpoet/lib/API/JSON/v1/Segments.php
```diff
@@ -294,6 +300,19 @@ public function bulkAction($data = []) {
     return $this->successResponse(null, ['count' => $count]);
   }
 
+  public function subscriberCount($data = []) {
+    $segmentIds = $data['segmentIds'] ?? [];
+    $response = ['count' => 0];
+    if (empty($segmentIds)) {
+      return $this->successResponse($response);
```
rodrigoprimo, line 307: I wonder if it is better to return a success response or an error if `$segmentIds` is empty. Is there a case for returning a success response here?

mailpoet/assets/js/src/common/loader/loading_spinner.tsx
```diff
@@ -0,0 +1,14 @@
+type LoadingSpinnerProps = {
+  alt: string;
+  className?: string;
+};
+
+export function LoadingSpinner({ alt, ...restProps }: LoadingSpinnerProps) {
```
rodrigoprimo, line 6: I haven't checked the details, but was adding a new loading spinner part of the specifications of this ticket? I'm asking because I'm wondering if we don't have already available some other way to indication loading that we could reuse in this PR instead of implementing a new one.

rodrigoprimo: Or if we need a new one, have we checked if WooCommerce or WordPress doesn't provide something that we can reuse?

mailpoet/assets/js/src/newsletters/send.tsx
```diff
@@ -227,6 +230,12 @@ class NewsletterSendComponent extends Component<
 
   getThumbnailPromise = (url) => this.state?.thumbnailPromise ?? fromUrl(url);
 
+  get premiumFeaturesAvailable() {
```
rodrigoprimo, line 234: TIL about the `get` keyword in JS. Thanks.

mailpoet/assets/js/src/newsletters/send/filter_segment.tsx
```diff
@@ -0,0 +1,167 @@
...
+          {__(
+            `Subscribers selected in 'Send to' will only receive an email if they also belong to this segment.`,
+            'mailpoet',
+          )}
+        </div>
+      </Tooltip>
+      <div className="mailpoet-gap" />
+      {filterSegmentSelect}
+      <p>
+        {ReactStringReplace(
```
rodrigoprimo: Have you checked if there is a WP recommend way to localize strings with links in JS? I'm wondering if something similar to [this](https://codex.wordpress.org/I18n_for_WordPress_Developers#HTML) can be used in JS or if it only applies to PHP.

rodrigoprimo (commented):
I started reviewing this PR but them chatted with John and we decided that it would be better to split this PR and create separate PRs for each of the tickets. So I'm publish this partial review.

mailpoet/lib/API/JSON/v1/Segments.php
johnolek: Good point! I'll update this.

I wrote this thinking that the send step might trigger an API call when it had a filter segment selected but no segment IDs, but we can ensure that doesn't happen, and it doesn't make sense to return a success response if no segment IDs are provided.

mailpoet/assets/js/src/common/loader/loading_spinner.tsx
johnolek: Thank you for calling this out. tldr I'm going to update this to use the WP components [Spinner](https://wordpress.github.io/gutenberg/?path=/docs/components-spinner--docs), which we're already using elsewhere in the codebase, unless you happen to have a better suggestion.

It's not part of the specs, but the method we do have to indicate loading on the send step blurs out the entire form and prevents editing. I was thinking that, especially if a user has very large/complex lists that take some time to calculate, it would be very annoying to have to wait until the API request completes before making other changes.

The spinner image is something we already use in a few other places ([example1](https://github.com/mailpoet/mailpoet/blob/cf1e89d39f285465177c4a5db32b7f848f6afabc/mailpoet/assets/css/src/generic/_helpers.scss#L86-L86), [example2](https://github.com/mailpoet/mailpoet/blob/cf1e89d39f285465177c4a5db32b7f848f6afabc/mailpoet/assets/css/src/generic/_buttons-spinners.scss#L2-L2)), but those didn't seem to be easily reusable and I thought having a component might be nice for the future.

rodrigoprimo: > tldr I'm going to update this to use the WP components [Spinner](https://wordpress.github.io/gutenberg/?path=/docs/components-spinner--docs)

Sounds like a good plan to me. Thanks.

PR #5143, opened 2023-09-05
Remove JS compilation from JS newsletter tests [MAILPOET-5571]

lysyjan:
## Description

_N/A_

## Code review notes

JS is compiled in the build job, so this step seems redundant.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -252,11 +252,7 @@ class RoboFile extends \Robo\Tasks {
       );
     }
 
-    return $this->collectionBuilder()
-      ->addCode(function () {
-        $this->compileJs();
-      })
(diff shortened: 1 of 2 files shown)


JanJakes (commented):
LGTM, but I think we should remove [TS fork pluging](https://github.com/mailpoet/mailpoet/blob/34a97a0b342170e8548dc79cdd9f22f95e709de8/mailpoet/webpack.config.js#L301) as well, when we're at it. After all, the `test` config doesn't inherit from the `base` config, so there is no reason to inherit the plugins only.

JanJakes (approved):

PR #5144, opened 2023-09-05
Added/Removed from tag trigger [PREMIUM-242]

websupporter:
## Description

_N/A_

## Code review notes

This PR adds also a small `./do` command to quickly set up triggers and actions. It was just a quick write up, as I will write some triggers in the near future and didn't want to go through the whole copy-paste process all the time.

Let me know, what you think. If you don't like it or it's a big rewrite, it can be excluded from this PR.

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/780

## Linked tickets

[PREMIUM-242]

## After-merge notes

_N/A_

## Tasks

[PREMIUM-242]: https://mailpoet.atlassian.net/browse/PREMIUM-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


websupporter:
icons will change, but rest can be reviewed

mailpoet/lib/Subscribers/SubscriberSaveController.php
```diff
@@ -317,6 +318,7 @@ private function updateTags(array $data, SubscriberEntity $subscriber): void {
         $subscriberTag = new SubscriberTagEntity($tag, $subscriber);
         $subscriber->getSubscriberTags()->add($subscriberTag);
         $this->subscriberTagRepository->persist($subscriberTag);
+        $this->wp->doAction('mailpoet_subscriber_tag_added', $subscriberTag);
```
JanJakes: I'm thinking... maybe we should call these hooks after `flush`? I've ran into a similar problem in the past — firing hook that a change has occurred, but not having the changes in the DB yet. It would mean collecting the tags to an array and iterating after flush. What do you think?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/tag_added/index.tsx
```diff
@@ -0,0 +1,32 @@
+import { __ } from '@wordpress/i18n';
+import { StepType } from '../../../../editor/store';
+import { LockedBadge } from '../../../../../common/premium_modal/locked_badge';
+import { PremiumModalForStepEdit } from '../../../../../common/premium_modal';
+
+const keywords = ['tag', 'tags', 'tag added', 'tags added'];
```
JanJakes: I think in other places we've already made these translatable.

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/tag_removed/index.tsx
```diff
@@ -0,0 +1,32 @@
+import { __ } from '@wordpress/i18n';
+import { StepType } from '../../../../editor/store';
+import { LockedBadge } from '../../../../../common/premium_modal/locked_badge';
+import { PremiumModalForStepEdit } from '../../../../../common/premium_modal';
+
+const keywords = ['tag', 'tags', 'tag removed', 'tags removed'];
```
JanJakes: Same here (translations).

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/tag_removed/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  icon: () => null,
+  edit: () => (
+    <PremiumModalForStepEdit
+      tracking={{
+        utm_medium: 'upsell_modal',
+        utm_campaign: 'create_automation_editor_tag_removed',
+      }}
+    >
+      {__(
+        'Starting an automatoin when a tag has been removed from a subscriber is a premium feature.',
```
JanJakes: Typo in `automatoin`

JanJakes (commented):
> This PR adds also a small ./do command to quickly set up triggers and actions. It was just a quick write up, as I will write some triggers in the near future and didn't want to go through the whole copy-paste process all the time.

👍 Fine with me. I suppose we won't maintain that long-term, so we can consider removing it in a couple of months.

mailpoet/lib/Subscribers/SubscriberSaveController.php
websupporter: c78d1d0ee

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/tag_added/index.tsx
websupporter: 35f516309

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/tag_removed/index.tsx
websupporter: 35f516309

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/tag_removed/index.tsx
websupporter: 608b7cfca

websupporter:
Thanks for the review @JanJakes. Issues adressed.

JanJakes (dismissed):
Thanks!

websupporter:
@lysyjan would like to have your input on ec79e28

I found out that when I create a new tag the `$data['tags']` is an array of tag names. When I do not create a new tag (e.g. I go to edit and change the status of the subscription) `$data['tags']` contains arrays of tags like `["name", "id", "subscriber_id", ...]`.

In the `updateTags` method I normalize this to be an array of names, but wanted to ask for your input what you think here.

Thank you.

lysyjan:
@websupporter Nice catch 👍 
I would prefer a follow-up ticket that would fix an incorrect API call. In my opinion, we shouldn't overlook this inconsistency.
Also, I guess we could consider this fix as temporary and remove it after removing the cause.

What do you think about it?

The method for updating tags is designed for work with tag names only because there is a unique index for the tag name.

websupporter:
> I would prefer a follow-up ticket that would fix an incorrect API call

I like that @lysyjan.

websupporter:
rebased to solve a merge conflict and created the ticket: https://mailpoet.atlassian.net/browse/MAILPOET-5606

Aschepikov:
GJ!

Aschepikov (approved):

PR #5145, opened 2023-09-06
Finish implementation of pre-built segments [MAILPOET-5394]

rodrigoprimo:
## Description

This PR finishes the implementation of the pre-built segments page that was started in #5102 when the page was added with just the design and no functionality.

IMPORTANT: this PR should be released together with https://github.com/mailpoet/mailpoet-premium/pull/781 or things will break.

## Code review notes

- I opted to reuse the dynamic segments form store and make it a store for the whole dynamic segments app. I did this because I vaguely remember someone sharing with me that it would be preferable to have a single store per app, but I don't know much about this. Please let me know if what I did makes sense or not.
- More generally, I used some React patterns that I'm not super familiar with, any feedback in this area is very welcome. There is a good chance I used some anti-patterns in this PR. 

Besides the code review notes above, I have a few comments/questions for @NeosinneR :

- Since we create a new segment before showing the segment editor when the user clicks on a segment template card, I opted to add a random string to the end of the segment name if there is already a segment that matches the name of the segment template. I had to do this to work around an error since segment names are unique. Please let me know if that is fine or if you prefer another approach.
- There is a "Back" button on the segment editor that always takes the user back to the segment listing page. But now, the users can get to the editor via the segment templates page as well. I believe we should still take the user to the listing page, but maybe we should change the name of the button? In my mind, I expect "Back" to return to the previous page. WP users "View Pages" or "View Posts". "View Segments" would be better than "Back" in my opinion, but not necessarily great. I also considered "Return to listing" or "Return to segments listing". Thoughts?
- The specs for this ticket says to auto-generate the list of categories based on the templates. I opted not to do that since I'm assuming we want to translate the names of the categories. There is a file where I harder all the categories so that their names can be translated.

## QA notes

This PR makes some significant changes to the code behind dynamic segments. Please test everything related to dynamic segments that is not covered by the automated tests.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5394]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5395]: https://mailpoet.atlassian.net/browse/MAILPOET-5395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-5394]: https://mailpoet.atlassian.net/browse/MAILPOET-5394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NeosinneR:
@rodrigoprimo 
> I opted to add a random string to the end of the segment name if there is already a segment that matches the name of the segment template

👍  

> There is a "Back" button on the segment editor that always takes the user back to the segment listing page.

I asked in Slack (p1694072087642829-slack-C040TSG288Y).

> The specs for this ticket says to auto-generate the list of categories based on the templates. I opted not to do that since I'm assuming we want to translate the names of the categories. There is a file where I harder all the categories so that their names can be translated.

👍

NeosinneR:
@rodrigoprimo an update, the back button should return to the previous page (p1694072710455989/1694072087.642829-slack-C040TSG288Y), so in this case back to the templates page, not the listings.

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
```diff
@@ -105,12 +79,17 @@ export function SegmentTemplates(): JSX.Element {
         </FlexItem>
       </Flex>
 
-      <TabPanel tabs={tabConfig}>
```
JanJakes, line 182: Unrelated to this PR, but out of curiosity — a few lines above (can't put the comment there, unfortunately), there is the following snippet:

```
direction={['column', 'row'] as any} // eslint-disable-line @typescript-eslint/no-explicit-any -- typed as string but supports string[] and this is needed to make the component responsive
```

What does that mean? CSS flex doesn't seem to support anything like that, and the resulting HTML contains only the `row` value. Looking at the [original component source codes](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/flex/flex), I also can't find the support of such a setting 🤔

mailpoet/assets/js/src/segments/dynamic/dynamic.tsx
```diff
@@ -8,10 +8,13 @@ import { Editor } from 'segments/dynamic/editor';
 import { DynamicSegmentList } from 'segments/dynamic/list';
 import { SegmentTemplates } from 'segments/dynamic/templates';
 import * as ROUTES from 'segments/routes';
+import { createStore } from 'segments/dynamic/store';
 
 const container = document.getElementById('dynamic_segments_container');
 
 function App(): JSX.Element {
+  createStore();
```
JanJakes: This place is actually a risky place to put it (and so was the original one), because it's in the rendering body of a component. The function can be called multiple times, if multiple re-renders occur for any reason.

In most scenarios, the solution is to put the code in `useEffect(..., [])`, but for a store, I think the best solution is to instantiate it before creating the React app. I mean at the end of this file, something like this:

```ts
if (container) {
  registerTranslations();
  createStore(); // added here
  ReactDOM.render(<App />, container);
}
```

mailpoet/assets/css/src/mailpoet-templates.scss
```diff
@@ -59,15 +59,30 @@
...
     display: grid;
     grid-template-rows: auto 1fr;
   }
+
+  &:hover {
+    background: #fff;
+    border: 1px solid #dcdcde;
+    box-shadow: 0 3px 6px rgba(0, 0, 0, .15);
+    color: inherit;
+  }
```
JanJakes, line 77: I'm not sure how this happens, but with the hover state, the border seems to jump 1px inside and get a different radius.

At the same time, I just noticed, that while the cards react to a click, they don't seem to be accessible (= they are not buttons with `tabindex` and stuff). I suppose this is rather OK, because TAB selects the headings (but it also likely means a screen reader wouldn't read the subtitle when selected). If it's a component we should reuse across the plugin, we could consider creating a ticket to improve this.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
```diff
@@ -19,8 +21,36 @@ type TemplateListItemProps = {
...
+    segment.filters = segmentTemplate.filters;
+
+    if (segmentTemplate.filtersConnect) {
+      segment.filters_connect = segmentTemplate.filtersConnect;
+    }
+
+    updateSegment({
+      ...segment,
+    });
+    createFromTemplate();
```
JanJakes: At first, I was a bit confused why we update a segment here and then call `createFromTemplate()`, but as I understand it, `createFromTemplate()` is pretty much just the final "save". In that case, we could also consider passing the `segmentTemplate` to `createFromTemplate(segmentTemplate)` and move the logic there. Not too important.

mailpoet/lib/Segments/DynamicSegments/SegmentSaveController.php
```diff
@@ -38,6 +38,11 @@ public function __construct(
   public function save(array $data = []): SegmentEntity {
     $id = isset($data['id']) ? (int)$data['id'] : null;
     $name = $data['name'] ?? '';
+
+    if (!$this->segmentsRepository->isNameUnique($name, null) && isset($data['force_creation']) && $data['force_creation'] === 'true') {
+      $name = $name . ' (' . wp_generate_password(12, false) . ')';
```
JanJakes: Since it's only for the name, I guess we could consider making the random hash shorter...

JanJakes (commented):
LGTM! Just small remarks and one question about an older line (not touched in this PR).

> I opted to reuse the dynamic segments form store and make it a store for the whole dynamic segments app. I did this because I vaguely remember someone sharing with me that it would be preferable to have a single store per app, but I don't know much about this. Please let me know if what I did makes sense or not.

I think it's a good idea, representing the full app state with a single store makes sense, and here it's not a huge store, so I'd say it looks good. This approach fails when we try to reuse some component subtree that relies on the store — then it needs its own — but it's not this case.

> More generally, I used some React patterns that I'm not super familiar with, any feedback in this area is very welcome. There is a good chance I used some anti-patterns in this PR.

LGTM!

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
rodrigoprimo: I'm not sure exactly how this works internally, but passing `{['column', 'row'] as any}` means that `flex-direction` will be automatically updated to `column` when displaying the component on smaller screens. This is needed especially when the search box is visible otherwise one element is stacked on top of the other.

mailpoet/assets/js/src/segments/dynamic/dynamic.tsx
rodrigoprimo: Good point! I added a new commit moving `createStore()` to the place you suggested.

mailpoet/assets/js/src/segments/dynamic/templates/components/template_list_item.tsx
rodrigoprimo: I like this suggestion and I added a commit implementing it. Thanks.

mailpoet/lib/Segments/DynamicSegments/SegmentSaveController.php
rodrigoprimo: I went with 12 just because it is the default value for `wp_generate_password()`. Making it shorter sounds good to me. Do you have a suggestion of what length to use?

mailpoet/assets/css/src/mailpoet-templates.scss
rodrigoprimo: Good catch! This problem was happening because the <Card> component uses `box-shadow` instead of `border` to create its border. I failed to notice this and I was using `border` in the CSS rules that I created. Should be fixed now.

rodrigoprimo: > At the same time, I just noticed, that while the cards react to a click, they don't seem to be accessible (= they are not buttons with tabindex and stuff). I suppose this is rather OK, because TAB selects the headings (but it also likely means a screen reader wouldn't read the subtitle when selected). If it's a component we should reuse across the plugin, we could consider creating a ticket to improve this.

I'm using the <Card> WordPress component, but maybe the problem is that this component is not meant to be clickable and thus, this is causing the accessibility problem that you mentioned. I think that creating a separate ticket to look into this is a good approach.

rodrigoprimo:
> @rodrigoprimo an update, the back button should return to the previous page (p1694072710455989/1694072087.642829-slack-C040TSG288Y), so in this case back to the templates page, not the listings.

Thanks for checking, @NeosinneR. I updated the behavior of the button. It will still return the browser to the listing page if the previous page is not set. This can happen if the user opens the page to edit a segment directly.

@JanJakes, implementing this behavior was not as straightforward as I thought it would be. Let me know if you can think of a better approach.

rodrigoprimo:
Thanks for the review, @JanJakes. I believe I addressed all your comments. Can you please check this PR again when you have a moment?

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
JanJakes: @rodrigoprimo Oh, wow, interesting. Did you find that in some docs/code or somewhere?

mailpoet/lib/Segments/DynamicSegments/SegmentSaveController.php
JanJakes: @rodrigoprimo Anything that "looks fine" in the listing, I guess. Like 5-6, maybe?

mailpoet/assets/js/src/segments/dynamic/dynamic.tsx
```diff
@@ -8,14 +8,36 @@ import { Editor } from 'segments/dynamic/editor';
...
+      history.listen((location) => {
+        setPreviousPage(previousPageRef.current);
+
+        previousPageRef.current = location.pathname;
+      }),
+    [history, setPreviousPage],
+  );
+
+  return null;
+}
```
JanJakes, line 34: If I understand this correctly, we look for _any_ previous page. In that case, could we just use `history.back()` or look at the browser history directly (`document.referrer`), when needed?

On the other hand, thinking further, this listener will save only pages withing this React app, which is probably what we want and that's why we need this hook, right?

JanJakes (commented):
LGTM! I'm approving this, but leaving assigned to you @rodrigoprimo, in case you want to add any more changes.

JanJakes (dismissed):

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
rodrigoprimo: @JanJakes, I found it in the docs: https://wordpress.github.io/gutenberg/?path=/docs/components-flex--docs#responsive-direction

mailpoet/assets/css/src/mailpoet-templates.scss
rodrigoprimo: Ticket here: https://mailpoet.atlassian.net/browse/MAILPOET-5585

mailpoet/lib/Segments/DynamicSegments/SegmentSaveController.php
rodrigoprimo: I changed it to 5

mailpoet/assets/js/src/segments/dynamic/dynamic.tsx
rodrigoprimo: I opted to use this listener to save only pages within this React app and also because this was the way that I found to have a default value and point the users to the listing page when they opened the segment edit page directly.

rodrigoprimo:
Just documenting that I had to force push after rebasing this branch with trunk to fix a failing CI build.

Aschepikov (approved):
GJ!

PR #5148, opened 2023-09-07
Custom Trigger and Custom Action [MAILPOET-5514]

websupporter:
## Description
Creates the stubs for the custom trigger and action. It also removes `mailpoet:segment` as a required subject for sending emails via the `mailpoet:send-email` action.

Additionally: I changed the behavior of `AutomationRunStorage::getCountForAutomation()`. While writing tests for the premium plugin, I used this method to see if runs had been created. Earlier behavior: If you did not specify the status of the runs you want to count, you would get `0`. Took me a while to figure that out. I think it makes much more sense to query runs of all statuses when no status has been defined.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/782

## Linked tickets

[MAILPOET-5514]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5514]: https://mailpoet.atlassian.net/browse/MAILPOET-5514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/custom_trigger/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/custom_trigger/index.tsx
@@ -16,7 +16,7 @@ export const step: StepType = {
   title: () => __('Custom trigger', 'mailpoet'),
   description: () =>
     __(
-      'Starts an automation when a certain action is fired. This action needs to hand over the email address of the subscriber.',
+      "This is an advanced feature for developers. Triggers an automation when a certain action is fired. The action's first argument must be the email address of a subscriber.",
       'mailpoet',
     ),
   subtitle: () => <LockedBadge text={__('Premium', 'mailpoet')} />,
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/custom_trigger/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/custom_trigger/index.tsx
@@ -1,0 +1,44 @@
+import { __ } from '@wordpress/i18n';
+import { plugins } from '@wordpress/icons';
+import { StepType } from '../../../../editor/store';
+import { LockedBadge } from '../../../../../common/premium_modal/locked_badge';
+import { PremiumModalForStepEdit } from '../../../../../common/premium_modal';
+
+const keywords = [
+  __('custom', 'mailpoet'),
+  __('hook', 'mailpoet'),
+  __('code', 'mailpoet'),
+];
+
+export const step: StepType = {
+  key: 'mailpoet:custom-trigger',
+  group: 'triggers',
+  title: () => __('Custom trigger', 'mailpoet'),
+  description: () =>
+    __(
+      'Starts an automation when a certain action is fired. This action needs to hand over the email address of the subscriber.',
+      'mailpoet',
+    ),
+  subtitle: () => <LockedBadge text={__('Premium', 'mailpoet')} />,
+  keywords,
+  foreground: '#2271b1',
+  background: '#f0f6fc',
+  icon: () => (
+    <div style={{ width: '100%', height: '100%', scale: '1.12' }}>
+      {plugins}
+    </div>
+  ),
+  edit: () => (
+    <PremiumModalForStepEdit
+      tracking={{
+        utm_medium: 'upsell_modal',
+        utm_campaign: 'create_automation_editor_custom_trigger',
+      }}
+    >
+      {__(
+        'Triggering an automation with a custom hook is a premium feature.',
+        'mailpoet',
+      )}
+    </PremiumModalForStepEdit>
+  ),
+} as const;
--- a/mailpoet/lib/Analytics/Reporter.php
+++ b/mailpoet/lib/Analytics/Reporter.php
@@ -303,14 +303,14 @@ class Reporter {
     }
     $averageSteps = $activeAutomationCount > 0 ? $totalSteps / $activeAutomationCount : 0;
 
-    $customTriggerHooks = array_values(array_map(
+    $customTriggerHooks = array_unique(array_values(array_map(
       function(Automation $automation): string {
         $trigger = $automation->getTrigger('mailpoet:custom-trigger');
         return $trigger ? (string)$trigger->getArgs()['hook'] : '';
       },
       $automationsWithCustomTrigger
-    ));
-    $customActionHooks = array_values(array_map(
+    )));
+    $customActionHooks = array_unique(array_values(array_map(
       function(Automation $automation): array {
         $customActionSteps = array_filter(
           $automation->getSteps(),
@@ -331,7 +331,7 @@ class Reporter {
 
       },
       $activeAutomations
-    ));
+    )));
     $customActionHooks = array_values(array_filter(array_merge(...$customActionHooks)));
     return [
       'Automation > Number of active automations' => $activeAutomationCount,
--- a/mailpoet/lib/Analytics/Reporter.php
+++ b/mailpoet/lib/Analytics/Reporter.php
@@ -255,6 +255,12 @@ class Reporter {
         return $automation->getStatus() === Automation::STATUS_DRAFT;
       }
     );
+    $automationsWithCustomTrigger = array_filter(
+      $activeAutomations,
+      function(Automation $automation): bool {
+        return $automation->getTrigger('mailpoet:custom-trigger') !== null;
+      }
+    );
     $automationsWithWordPressUserSubscribesTrigger = array_filter(
       $activeAutomations,
       function(Automation $automation): bool {
@@ -297,6 +303,36 @@ class Reporter {
     }
     $averageSteps = $activeAutomationCount > 0 ? $totalSteps / $activeAutomationCount : 0;
 
+    $customTriggerHooks = array_values(array_map(
+      function(Automation $automation): string {
+        $trigger = $automation->getTrigger('mailpoet:custom-trigger');
+        return $trigger ? (string)$trigger->getArgs()['hook'] : '';
+      },
+      $automationsWithCustomTrigger
+    ));
+    $customActionHooks = array_values(array_map(
+      function(Automation $automation): array {
+        $customActionSteps = array_filter(
+          $automation->getSteps(),
+          function(Step $step): bool {
+            return $step->getKey() === 'mailpoet:custom-action';
+          }
+        );
+        if (!$customActionSteps) {
+          return [];
+        }
+
+        return array_map(
+          function(Step $step): string {
+            return (string)$step->getArgs()['hook'];
+          },
+          $customActionSteps
+        );
+
+      },
+      $activeAutomations
+    ));
+    $customActionHooks = array_values(array_filter(array_merge(...$customActionHooks)));
     return [
       'Automation > Number of active automations' => $activeAutomationCount,
       'Automation > Number of draft automations' => count($draftAutomations),
@@ -307,6 +343,8 @@ class Reporter {
       'Automation > Number of steps in shortest active automation' => $minSteps,
       'Automation > Number of steps in longest active automation' => $maxSteps,
       'Automation > Average number of steps in active automations' => $averageSteps,
+      'Automation > Custom Trigger Hooks' => $customTriggerHooks,
+      'Automation > Custom Action Hooks' => $customActionHooks,
     ];
   }
 
--- a/mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
+++ b/mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
@@ -9,6 +9,7 @@ use MailPoet\Automation\Engine\Data\NextStep;
 use MailPoet\Automation\Engine\Data\Step;
 use MailPoet\Automation\Engine\Data\StepRunArgs;
 use MailPoet\Automation\Engine\Data\StepValidationArgs;
+use MailPoet\Automation\Engine\Exceptions\NotFoundException;
 use MailPoet\Automation\Engine\Integration\Action;
 use MailPoet\Automation\Engine\Integration\ValidationException;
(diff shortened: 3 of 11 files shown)


mailpoet/assets/js/src/automation/integrations/mailpoet/steps/custom_trigger/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  __('code', 'mailpoet'),
+];
+
+export const step: StepType = {
+  key: 'mailpoet:custom-trigger',
+  group: 'triggers',
+  title: () => __('Custom trigger', 'mailpoet'),
+  description: () =>
+    __(
+      'Starts an automation when a certain action is fired. This action needs to hand over the email address of the subscriber.',
```
johnolek: > This action needs to hand over the email address of the subscriber

I think this phrasing could be a little clearer and/or easier to translate if we used language more specific to actions. Maybe something like "The action's first argument must be the email address of a subscriber." since the [WP documentation](https://developer.wordpress.org/reference/functions/do_action/) refers to them as arguments.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -177,6 +162,34 @@ private function getNewsletterMeta(StepRunArgs $args): array {
...
+      $subscriber = $subscriberSegment->getSubscriber();
+      if (!$subscriber) {
+        throw InvalidStateException::create();
+      }
+    } catch (NotFoundException $e) {
+      $subscriber = $this->subscribersRepository->findOneById($subscriberId);
+      if (!$subscriber) {
+        throw InvalidStateException::create();
+      }
+    }
```
johnolek, line 190: This could be more intuitive if instead of relying on a `NotFoundException` from the payload not existing we simply checked for the existence of the `mailpoet:segment` payload. This works fine, but it's not immediately clear to me what the `NotFoundException` is, especially since it's technically possible that the exception [could be thrown due to multiple payloads existing](https://github.com/mailpoet/mailpoet/blob/491591254fb2e2b0fe1d4a94efad19f5f7413f0d/mailpoet/lib/Automation/Engine/Data/StepRunArgs.php#L126-L126).

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -108,7 +109,6 @@ public function getArgsSchema(): ObjectSchema {
 
   public function getSubjectKeys(): array {
     return [
-      'mailpoet:segment',
```
johnolek, line 112: I don't think I understand this action well enough to understand why it's necessary to make the segment optional. Would you mind providing a bit more context about this change?

mailpoet/lib/Analytics/Reporter.php
```diff
@@ -307,6 +343,8 @@ function(Step $step): bool {
       'Automation > Number of steps in shortest active automation' => $minSteps,
       'Automation > Number of steps in longest active automation' => $maxSteps,
       'Automation > Average number of steps in active automations' => $averageSteps,
+      'Automation > Custom Trigger Hooks' => $customTriggerHooks,
+      'Automation > Custom Action Hooks' => $customActionHooks,
```
johnolek, line 347: Is it helpful to see duplicates or should we apply an array_unique to these?

mailpoet/tasks/phpstan/phpstan.neon
```diff
@@ -44,7 +44,7 @@ parameters:
     - '/Call to method getName\(\) on an unknown class _generated\\([a-zA-Z])*Cookie/' # codeception generate incorrect return type in ../../tests/_support/_generated
     -
       message: "#^Cannot cast string|void to string\\.$#"
-      count: 10
```
johnolek, line 47: Is there a simple way to fix this in the code instead of incrementing this count?

johnolek (dismissed):
Just a few minor things, none of which should require changes. Feel free to move on to QA if you don't think changes are needed (and after the premium PR is ready)

mailpoet/tasks/phpstan/phpstan.neon
websupporter: I am not aware of a good way. You could do something like 
```
$query = $this->wpdb->prepare(<query>);
if (!is_string($query)) {
  // some error handling
}
```

But this seems quite a overhead.

mailpoet/lib/Analytics/Reporter.php
websupporter: Unique makes sense. Thank you.

a4f3d4819

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
websupporter: The `SubjectKeys` tells you what subjects a trigger needs to provide for the step to be able to function. The Custom Trigger is not tied to a specific list (e.g. Subscribe To List is tied to a specific segment). Still, we want to be able to send emails. So, we remove the `mailpoet:segment` as a necessary subject and reorganize how we try to get the subscriber. In case the segment subject is present, we get the subscriber as usual (so the subscriber needs to be subscribed to the segment. If no segment is given, we query the subscriber by it's ID.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
websupporter: I do not quite understand what you mean. How to check for the existence of `mailpoet:segment` payload without catching the `NotFoundException`?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/custom_trigger/index.tsx
websupporter: I like that. Thank you. 

3108723db

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
johnolek: Sorry if this doesn't make sense. I might be showing off my lack of expertise in this part of the codebase.

It looks like you might be able to look at the keys in the array returned by `$args->getSubjectEntries()`. Maybe we could add a helper like `hasSubjectArg($key)` to `StepRunArgs`, but you would probably know better than me if that makes any sense!

It's probably fine to leave this as is, but the `NotFoundException` is a potentially ambiguous (though probably not in practice -- I don't know what would actually lead to there being multiple payloads).

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
johnolek: Thanks very much for the explanation!

mailpoet/tasks/phpstan/phpstan.neon
johnolek: Agreed, that would be very annoying to have to do every time.

I'm far from a PHPStan expert, but it seems like one option would be to register a custom return type extension. As far as I can tell, `prepare` will only return `void` if it receives a `null` value for `$query`, so something like this should work:

```php
<?php declare(strict_types = 1);

namespace MailPoet\PHPStan\Extensions\CodeceptionExtension\Type;

use PhpParser\Node\Expr\MethodCall;
use PHPStan\Analyser\Scope;
use PHPStan\Reflection\MethodReflection;
use PHPStan\Type\DynamicMethodReturnTypeExtension;
use PHPStan\Type\NullType;
use PHPStan\Type\StringType;
use PHPStan\Type\Type;
use PHPStan\Type\VoidType;

class WPDBPrepareMethodReturnTypeExtension implements DynamicMethodReturnTypeExtension {
  public function getClass(): string {
    return 'wpdb';
  }

  public function isMethodSupported(MethodReflection $methodReflection): bool {
    return $methodReflection->getName() === 'prepare';
  }

  public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type {
    if (count($methodCall->args) > 0) {
      $argumentType = $scope->getType($methodCall->args[0]->value);

      if ($argumentType instanceof NullType) {
        return new VoidType();
      }

      return new StringType();
    }

    return new VoidType();
  }
}
```

Then we register it in `extensions.neon`:
```
services:
  # resolves return type from Stub::make(SomeMockedClass::class) and other Stub methods
  -
    class: MailPoet\PHPStan\Extensions\CodeceptionExtension\Type\StubDynamicReturnTypeExtension
    tags:
      - phpstan.broker.dynamicStaticMethodReturnTypeExtension

  # resolves return type from $this->make(SomeMockedClass::class) and other test case methods
  -
    class: MailPoet\PHPStan\Extensions\CodeceptionExtension\Type\TestCaseDynamicReturnTypeExtension
    tags:
      - phpstan.broker.dynamicMethodReturnTypeExtension
  -
    class: MailPoet\PHPStan\Extensions\CodeceptionExtension\Type\WPDBPrepareMethodReturnTypeExtension
    tags:
        - phpstan.broker.dynamicMethodReturnTypeExtension
```

When I tried this out locally, I got the results I was hoping for:

```
 1557/1557 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ------------------------------------------------------------------------------- 
  Line   lib/Automation/Engine/Storage/AutomationRunLogStorage.php                      
 ------ ------------------------------------------------------------------------------- 
  49     Else branch is unreachable because ternary operator condition is always true.  
 ------ ------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Automation/Engine/Storage/AutomationRunStorage.php                                                                       
 ------ ----------------------------------------------------------------------------------------------------------------------------- 
         Ignored error pattern #^Cannot cast string|void to string\.$# in path                                                        
         /Users/johnoleksowicz/Workspace/mailpoet/mailpoet/lib/Automation/Engine/Storage/AutomationRunStorage.php was not matched in  
         reported errors.                                                                                                             
  207    Else branch is unreachable because ternary operator condition is always true.                                                
 ------ ----------------------------------------------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Automation/Engine/Storage/AutomationStorage.php                                                                       
 ------ -------------------------------------------------------------------------------------------------------------------------- 
         Ignored error pattern #^Cannot cast string|void to string\.$# in path                                                     
         /Users/johnoleksowicz/Workspace/mailpoet/mailpoet/lib/Automation/Engine/Storage/AutomationStorage.php was not matched in  
         reported errors.                                                                                                          
 ------ -------------------------------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php                                                  
 ------ ------------------------------------------------------------------------------------------------------------------------------- 
         Ignored error pattern #^Cannot cast string|void to string\.$# in path                                                          
         /Users/johnoleksowicz/Workspace/mailpoet/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.p  
         hp was not matched in reported errors.                                                                                         
 ------ ------------------------------------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Automation/Integrations/WooCommerce/Fields/CustomerReviewFieldsFactory.php                                                 
 ------ ------------------------------------------------------------------------------------------------------------------------------- 
         Ignored error pattern #^Cannot cast string|void to string\.$# in path                                                          
         /Users/johnoleksowicz/Workspace/mailpoet/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerReviewFieldsFactory.  
         php was not matched in reported errors.                                                                                        
 ------ ------------------------------------------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Automation/Integrations/WooCommerce/Fields/TermParentsLoader.php                                                        
 ------ ---------------------------------------------------------------------------------------------------------------------------- 
         Ignored error pattern #^Cannot cast string|void to string\.$# in path                                                       
         /Users/johnoleksowicz/Workspace/mailpoet/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermParentsLoader.php was  
         not matched in reported errors.                                                                                             
 ------ ---------------------------------------------------------------------------------------------------------------------------- 

                                                                                                                        
 [ERROR] Found 7 errors                                                                                                 
                                                                                                                        

 [Exec]  Exit code 1  Time 52.19s
 1/1 [============================] 100%
```

Just as a test I tried adding some code that explicitly called `$wpdb->prepare(null)`, and PHPStan caught it: `Result of method wpdb::prepare() (void) is used.`

I think this would also mean not having to perform the cast to string in the first place since PHPStan should see it as a string.

If we want to make this change it would probably be better to do it as a separate PR. I do think it would make our lives a tiny bit easier when working directly with `wpdb`. Curious to hear your thoughts.

johnolek (dismissed):
Thanks for the changes! This looks good to me now. I'll leave it up to you if you want to make any final tweaks.

mailpoet/tasks/phpstan/phpstan.neon
websupporter: I love it! Thank you @johnolek. I prepared a ticket [MAILPOET-5612]

Aschepikov:
@websupporter hi here are some conflicts can you resolve them pls and in the premium too

websupporter:
Done @Aschepikov

Aschepikov (approved):
GJ!

PR #5150, opened 2023-09-08
Automation async steps [MAILPOET-5568] [MAILPOET-5569]

JanJakes:
## Description

This implements changes to the automation engine that are necessary for async steps (e.g., waiting for an email to be sent), conditions, etc. It also adds multiple of fixes that I ran into while following a variety of 🐇 holes.

## Code review notes

_N/A_

## QA notes

Please, test this on a site with some existing automations and their history (existing logs and data in analytics).
After update to this build, all should still work, analytics should look the same, and migrations shouldn't fail.
Please, also include a general smoke test of executing automations.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/784

## Linked tickets

[MAILPOET-5568]
[MAILPOET-5569]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5568]: https://mailpoet.atlassian.net/browse/MAILPOET-5568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5569]: https://mailpoet.atlassian.net/browse/MAILPOET-5569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Engine/Data/StepRunArgs.php
```diff
@@ -150,4 +155,12 @@ public function getFieldValue(string $key) {
...
     return $value;
   }
+
+  public function getRunNumber(): int {
+    return $this->runNumber;
+  }
+
+  public function isFirstRun(): bool {
+    return $this->runNumber === 1;
+  }
```
websupporter, line 165: nitpick: not sure how useful this method is. We only need it in delay action and atm to me this seems more like extracting delay action logic than providing a general useful StepRunArgs method. Not sure though yet and not a big deal as well :thinking:

mailpoet/lib/Automation/Engine/Control/StepHandler.php
```diff
@@ -78,55 +78,70 @@ public function initialize(): void {
...
       throw new InvalidStateException();
     }
 
-    // Action Scheduler catches only Exception instances, not other errors.
-    // We need to convert them to exceptions to be processed and logged.
+    $runId = (int)$args['automation_run_id'];
+    $stepId = (string)$args['step_id'];
+    $runNumber = (int)($args['run_number'] ?? 1);
+
+    // BC — complete automation run if "step_id" is empty (was nullable in the past)
```
websupporter, line 84: For being backwards compatible. Do you have an idea on how long we want to maintain this?

mailpoet/lib/Automation/Engine/Control/StepHandler.php
```diff
@@ -78,55 +78,70 @@ public function initialize(): void {
...
+      $log->markFailed();
+      $log->setError($e);
+
+      // Action Scheduler catches only Exception instances, not other errors.
+      // We need to convert them to exceptions to be processed and logged.
       if (!$e instanceof Exception) {
         throw new Exception($e->getMessage(), intval($e->getCode()), $e);
       }
       throw $e;
+    } finally {
```
websupporter, line 107: Just, because I was not sure and needed to checked myself: Throwing an error in the catch block still means, `finally` gets executed (https://3v4l.org/OCS4F). You use `finally` because you want to execute `doAutomationStepAfterRun()` always, right?

mailpoet/lib/Automation/Engine/Control/StepRunLogger.php
```diff
@@ -0,0 +1,85 @@
...
+      throw new InvalidStateException('Failed to create automation run log');
+    }
+    return $this->log;
+  }
+
+  private function triggerAfterRunHook(AutomationRunLog $log): void {
+    try {
+      $this->hooks->doAutomationStepAfterRun($log);
+    } catch (Throwable $e) {
+      // ignore integration errors
```
websupporter: Hmmm.... Why do we `ignore integration errors`?

I see we want to somewhat protect our runs and ensure we finish each step and log properly etc.pp.

But I am just wondering, if this might be risky too when it comes to debugging :thinking: 

So, lets say I build some 

```
add_action('<step_run_after>', function(){
 //oups
 throw Error();
});
```

this will be silenced and i might overlook it during development. I mean, the only place I would see those errors would be in php logs, `debug.log` or something (as its async), but there i would see them otherwise usually.

Maybe we could do something like

```
try {
// some integration
} catch( Throwable $e) {
 if (defined(WP_DEBUG)) {
 throw $e;
 }
}
```

which makes me think (I think i saw this "ignore integration errors" somewhere else) if we should handle this generally in `Hooks` instead. Even if we conclude in the end to just ignore, it looks to me like this is something which would be better handled there :thinking:

websupporter (changes requested):
Thank you @JanJakes. That's a cool PR. Some questions, see above. Otherwise lgtm

mailpoet/lib/Automation/Engine/Data/StepRunArgs.php
JanJakes: @websupporter My thinking here was that usually actions that need to schedule "progress" will differentiate between "first" and "other" run. For instance, the send email step will schedule sending on the first run, then schedule progress, and on any other run, it will just check if the email was already sent, and if not, it can mark it as failed, for instance (we will also inform the automation immediately after email sending to avoid "polling", but I discussed with JM scheduling a "progress" to a week or a month to catch when no email was sent at all).

mailpoet/lib/Automation/Engine/Control/StepHandler.php
JanJakes: Most users that already use automations will probably update the plugin sooner than later, and then very quickly all of the steps with `null` should disappear. Usually, those were scheduled to "immediately", I think. My guess is that a couple of months should be fine, but it's not a huge piece of code, so I guess it can stay for a while.

mailpoet/lib/Automation/Engine/Data/StepRunArgs.php
websupporter: That makes sense. Thank you for clarifying.

mailpoet/lib/Automation/Engine/Control/StepHandler.php
JanJakes: @websupporter Yes, and I think it comes from the `finally` that was originally used deeper in the code for logging, and I extracted it higher up, but I think it really fits here.

mailpoet/lib/Automation/Engine/Control/StepRunLogger.php
JanJakes: @websupporter I kept this as it was in the original code, only moved it up. This is only a case when an integration tries to add some logging data and fails to do so, although the hook should probably be renamed to reflect that. I can also add the `WP_DEBUG` option, makes sense 👍

JanJakes:
@websupporter Added 2 commits based on the discussion.

websupporter (approved):
lgtm.

I think we can leave the exception handling for now where it is and discuss in some maintenance issue if this should be handled in `Hooks` instead.

Thank you @JanJakes

Aschepikov (approved):
GJ!

PR #5151, opened 2023-09-08
Add email settings sidebar to the new email editor [MAILPOET-5562]

lysyjan:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

1. Enable Gutenberg feature if it's not active yet. You do it at /wp-admin/admin.php?page=mailpoet-experimental
2. Create a newsletter in the new editor - you should see a modal window when you click on the new email button
3. Check the new sidebar containing subject and pre-header store data properly
4. The subject and pre-header can be checked in the next step

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5562]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5562]: https://mailpoet.atlassian.net/browse/MAILPOET-5562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/email_editor/integrations/MailPoet/components/details_panel.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  // Render email details panel using PluginDocumentSettingPanel component
+  return (
+    <PluginDocumentSettingPanel
+      className="mailpoet-email-editor__settings-panel"
+      title={__('Details', 'mailpoet')}
+      name="mailpoet-email-editor-setting-panel"
+    >
+      <TextareaControl
+        className="settings-panel__subject"
```
costasovo: I would suggest we keep all classNames prefixed `mailpoet-` to minimize the chance of conflict with other plugins.

costasovo (commented):
@lysyjan Nice job! I have just a tiny suggestion about the class name see the other comment.

As we added some CSS for specific components, I would like to discuss where to place those.
I'm thinking about two approaches:

1) We may assume that we won't need to write many CSS rules, because we reuse CSS from WP/ so we could keep adding them to email_editor.scss (as done in this PR)
2) We can distribute CSS files similarly to the Gutenberg repository and place them next to [components](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/button)
I guess that in this case, we could create a subdirectory for the settings panel and add one SCSS file to it.

Maybe we could try to start with 1) and refactor to 2) later. Or do you think we should start using the second approach straight away?

costasovo:
@lysyjan One more thought. This change adds functionality that will most probably not change much in the MVP we will release, so I would suggest that we ask QA to check it.
And we could also extend the [existing acceptance test](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/acceptance/Newsletters/CreateAndSendEmailUsingGutenbergCest.php) by setting a subject and verifying it was used

lysyjan:
@costasovo Thank your feedback. I will answer your questions below.
> Maybe we could try to start with 1) and refactor to 2) later. Or do you think we should start using the second approach straight away?

The second option sounds better to me because it allows us to maintain our components easily. If I understand correctly, we would need to have a directory for each component containing an index file. If you agree with this, I would prefer refactoring sooner than later because it will be easier.

> And we could also extend the [existing acceptance test](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/acceptance/Newsletters/CreateAndSendEmailUsingGutenbergCest.php) by setting a subject and verifying it was used

Thanks for this suggestion. I was looking for the acceptance test but didn't find it. Sorry for that. I should try looking better.

costasovo:
> The second option sounds better to me because it allows us to maintain our components easily. If I understand correctly, we would need to have a directory for each component containing an index file. If you agree with this, I would prefer refactoring sooner than later because it will be easier.

@lysyjan Yep. Let's do it sooner 👍 We don't need strictly a directory per component. We can also group a few closely related components into one directory. Here is [an example in the Gutenberg repo](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/block-controls). So in this case I feel it would be okay to put email settings panel stuff together.

> Thanks for this suggestion. I was looking for the acceptance test but didn't find it. Sorry for that. I should try looking better.

I think the test is pretty hidden. I should have placed it elsewhere. What about creating EmailEditor directory in acceptance tests and moving it there?

costasovo (dismissed):
Great job. Thank you!

Aschepikov:
Hi @lysyjan I wanted to ask about the font size of this part. it is 12, but everything else is 13. Should it be 12?
image
And there are some conflicts can you resolve them pls

lysyjan:
> Hi @lysyjan I wanted to ask about the font size of this part. it is 12, but everything else is 13. Should it be 12?

@Aschepikov Good point. The reason is that I used a font size from the Figma file, where it's 12px. Nevertheless, a Gutenberg component I was inspired by contains 13px, so I changed it in the last commit.

Also, this branch is rebased, and the conflict is solved.

Aschepikov (approved):
GJ!

PR #5161, opened 2023-09-15
Check table exists before altering it

brezocordero:
## Description

Table `mailpoet_statistics_woocommerce_purchases`  name is long and there are sites where it was never created because the prefix + name > 64.
This was not an issue on sites that were not using Woo but the latest migration alters the table and fails for those sites.

We will have another ticket to shorten the table name, in the meantime, this PR modifies the migrations so that it does not break those sites. 

If the table does not exist and they have Woo, the migration to populate the new column will still fail. This is on purpose because when Woo is active, that table and column need to be present.

## Code review notes

_N/A_

## QA notes

To test:
Check it still works:
With the correct table name, delete the column `status` if present
Set the latest too migrations to completed = NULL and error =NULL
Deactivate-reactivate the plugin and check there are no errors
Check status is created and populated if there was data on the table

Delete the table or truncate the name of `mailpoet_statistics_woocommerce_purchases` 
Deactivate Woo
Set the latest too migrations to completed = NULL and error =NULL
Deactivate-reactivate the plugin and check there are no errors

Reactivate Woo
Set the latest too migrations to completed = NULL and error =NULL
Deactivate-reactivate the plugin and check there are errros

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5590]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5590]: https://mailpoet.atlassian.net/browse/MAILPOET-5590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Migrations/App/Migration_20230825_093531_App.php
+++ b/mailpoet/lib/Migrations/App/Migration_20230825_093531_App.php
@@ -13,6 +13,12 @@ class Migration_20230825_093531_App extends AppMigration {
   public function run(): void {
 
     $wooCommerceHelper = $this->container->get(Helper::class);
+
+    // If Woo is not active and the table doesn't exist, we can skip this migration
+    if (!$wooCommerceHelper->isWooCommerceActive() && !$this->tableExists()) {
+      return;
+    }
+
     $wooCommerceHelper->isWooCommerceCustomOrdersTableEnabled() ?
       $this->populateStatusColumnUsingHpos() : $this->populateStatusColumnUsingPost();
   }
@@ -35,5 +41,14 @@ class Migration_20230825_093531_App extends AppMigration {
 
   private function getTableName(): string {
     return $this->entityManager->getClassMetadata(StatisticsWooCommercePurchaseEntity::class)->getTableName();
+  }
+
+  private function tableExists(): bool {
+    global $wpdb;
+
+    $revenueTable = $this->getTableName();
+    $query = $wpdb->prepare('SHOW TABLES LIKE %s', $wpdb->esc_like($revenueTable));
+
+    return $wpdb->get_var($query) === $revenueTable;
   }
 }
--- a/mailpoet/lib/Migrator/DbMigration.php
+++ b/mailpoet/lib/Migrator/DbMigration.php
@@ -50,6 +50,15 @@ abstract class DbMigration {
     ", [Env::$dbName, $tableName, $columnName])->fetchOne() !== false;
   }
 
+  protected function tableExists(string $tableName): bool {
+    return $this->connection->executeQuery("
+        SELECT 1
+        FROM information_schema.columns
+        WHERE table_schema = COALESCE(DATABASE(), ?)
+        AND table_name = ?
+      ", [Env::$dbName, $tableName])->fetchOne() !== false;
+  }
+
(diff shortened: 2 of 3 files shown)


mailpoet/lib/Migrations/App/Migration_20230825_093531_App.php
```diff
@@ -13,6 +13,12 @@ class Migration_20230825_093531_App extends AppMigration {
   public function run(): void {
 
     $wooCommerceHelper = $this->container->get(Helper::class);
+
+    // If Woo is not active and the table doesn't exist, we can skip this migration
+    if (!$wooCommerceHelper->isWooCommerceActive() && !$this->tableExists()) {
```
websupporter, line 18: if the table does not exist but woocommerce is active, we should try to populate the column? i think `||` is the better operator here?

mailpoet/lib/Migrator/DbMigration.php
```diff
@@ -50,6 +50,15 @@ protected function columnExists(string $tableName, string $columnName): bool {
     ", [Env::$dbName, $tableName, $columnName])->fetchOne() !== false;
   }
 
+  protected function tableExists(string $tableName): bool {
```
websupporter, line 53: why do you use a different sql statement here to check if the table exists?

websupporter (approved):
overall lgtm. thank you @brezocordero. two small remarks

mailpoet/lib/Migrator/DbMigration.php
brezocordero: Different from the one in the other migration file?
Because this class is only using a Doctrine connection instead of $wpdb.
On the other class I used the same query as in the [WP upgrader](https://github.com/WordPress/wordpress-develop/blob/6a264d2de04ca9ccc1fdd13791925f7323c3c61d/src/wp-admin/includes/upgrade.php#L2496).
You are more familiar than me with the migrations, how would you have done it?

mailpoet/lib/Migrations/App/Migration_20230825_093531_App.php
brezocordero: > if the table does not exist but woocommerce is active, we should try to populate the column?

The idea was to only skip the migration when Woo is not active and the table does not exist.
If Woo is active, the migration needs to run because the table and column are needed. It will fail if the table does not exist. 
I could throw an error instead of continuing if it helps with clarity. Let me know.

mailpoet/lib/Migrator/DbMigration.php
websupporter: not sure tbh. but i did not understand it :smile:

mailpoet/lib/Migrations/App/Migration_20230825_093531_App.php
websupporter: Ah! Yes. That makes sense. Thank you for clarifying.

websupporter (approved):
lgtm. thank you for clarifying.

Aschepikov (approved):
GJ!

PR #5164, opened 2023-09-18
Refactor loading coupons to Ajax requests [MAILPOET-5123]

lysyjan:
## Description

This PR removes loading all existing coupons into a template variable and starts using Ajax requests for loading them continually as it needed.

## Code review notes

- Because I haven't found a way how to use an existing WP or WC API for getting coupons via Ajax requests, I created our own solution.
- I implemented our own solution of an infinite scroll to the select with coupons in the segment form.
- I added loading indication to our react select component.

## QA notes

Please test that loading coupons works properly in:
- email editor: the coupon block
- segments: adding new segment of a type `used coupon code` and editing an existing one

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5123]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5123]: https://mailpoet.atlassian.net/browse/MAILPOET-5123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/woocommerce/used_coupon_code.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/woocommerce/used_coupon_code.tsx
@@ -1,10 +1,12 @@
 import { useDispatch, useSelect } from '@wordpress/data';
-import { useEffect } from 'react';
+import { useCallback, useEffect, useState } from 'react';
+import { __ } from '@wordpress/i18n';
 import { Grid } from 'common/grid';
 import { Select } from 'common';
 import { ReactSelect } from 'common/form/react_select/react_select';
 import { MailPoet } from 'mailpoet';
-import { filter } from 'lodash/fp';
+import { filter, uniqBy } from 'lodash/fp';
+import { APIErrorsNotice } from '../../../../../notices/api_errors_notice';
 import {
   WooCommerceFormItem,
   FilterProps,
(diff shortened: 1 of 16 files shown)


mailpoet/assets/js/src/segments/dynamic/dynamic_segments_filters/fields/woocommerce/used_coupon_code.tsx
```diff
@@ -36,6 +38,111 @@ export function UsedCouponCodeFields({
...
+          const nextPage = newPage + 1;
+          if (loadedCoupons.length === 0) {
+            setHasMore(false);
+          } else {
+            setCoupons((prevOptions: SelectOption[]) =>
+              uniqBy(
+                (coupon) => coupon.value,
+                [...prevOptions, ...loadedCoupons],
+              ),
+            );
```
johnolek, line 93: Very minor, but when I tested this out (with a lower page size since I don't have thousands of coupons), I noticed the select jumps back to the top after loading the new coupons. Probably no easy way to prevent that, and if anyone has >1000 coupons to go through they'll probably be using the search instead.

johnolek (approved):
Overall this looks good to me, nice work!

I can think of potential optimizations, like some client side caching, e.g. so you don't have to reload coupons every time you switch to the "all coupons" tab in the editor block, but none of the ideas I had seemed actually necessary or worth the added complexity. Just bringing it up in case you think differently.

johnolek:
@lysyjan I assigned back to you just in case you wanted to make any optimizations like I mentioned, but please feel free to move along to QA.

lysyjan:
@johnolek Thank for the review and notes. Because this task was a part of my Janitor week, I don't want to spend more time with it because I should focus on my current project.

Aschepikov (approved):
GJ!

PR #5166, opened 2023-09-19
Fix Redirect to the landing page after activation [MAILPOET-5221]

costasovo:
## Description

This PR fixes unwanted redirects to the MP landing page in case the plugin was not activated via the WP plugins page.

## Code review notes
I tested the page redirect after the change, and it works when I activate the plugin on the WP plugins page.
I'm not 100% sure if there are some activation cases/flows that might be affected by this change.

## QA notes
Please follow the replication steps in the ticket.
Please test also that the redirect works correctly when we want it to happen.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5221]

## After-merge notes

_N/A_

## Tasks

- [] I added sufficient test coverage - This could be covered by an acceptance test, but as this is rather an edge case, I chose to skip the test in this case.

[MAILPOET-5221]: https://mailpoet.atlassian.net/browse/MAILPOET-5221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Initializer.php
+++ b/mailpoet/lib/Config/Initializer.php
@@ -369,12 +369,15 @@ class Initializer {
 
     // wp automatically redirect to `wp-admin/plugins.php?activate=true&...` after plugin activation
     $activatedByWpAdmin = !empty(strpos($currentUrl, 'plugins.php')) && isset($_GET['activate']) && (bool)$_GET['activate'];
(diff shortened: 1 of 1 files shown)


triple0t (approved):
LGTM

Aschepikov (approved):
GJ!

PR #5149, opened 2023-09-08
Move abandoned cart to free [MAILPOET-5576]

websupporter:
## Description
Moves the abandoned cart trigger to the free plugin

## Code review notes
A test for the new trigger was missing. I added one.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/783

## Linked tickets

[MAILPOET-5576]

## Tasks

- [x] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] I added sufficient test coverage
- [x] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5576]: https://mailpoet.atlassian.net/browse/MAILPOET-5576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5168, opened 2023-09-20
Fix automation factory method calls

johnolek:
## Description

addStep was renamed to withStep but the calls in this helper weren't updated, and now pipelines are failing at the static analysis step.

It looks like this was caused by the merge of https://github.com/mailpoet/mailpoet/pull/5150, which must not have included the new test from https://github.com/mailpoet/mailpoet/pull/5149.

## Code review notes

_N/A_

## QA notes

No QA should be necessary

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks


lysyjan (approved):
Thanks for the fix!

PR #5169, opened 2023-09-21
Show filter segment used in listings/stats [MAILPOET-5511]

johnolek:
## Description

This PR adds a tag on listing pages/stats pages for newsletters that support filter segments if the email is currently configured to use a filter segment or if the email was sent using a filter segment.

Currently the UI for adding a filter segment to an email is not merged, but you can use the testing directions from https://github.com/mailpoet/mailpoet/pull/5108 to test this immediately.

The tooltip is going to be updated to include specific conditions in a separate ticket.

## Code review notes

I realized during my own testing that the filter segment data wasn't getting saved to the queue for post notification emails, and this PR fixes that issue by having the filter segment getter check the parent newsletter. This seemed cleaner to me than creating a new option in the database when creating the child newsletter.

## QA notes

For sent emails, the filter segment is stored on the sending queue itself, including the name. This means that if the segment name is changed we will continue to see the old segment name for sent emails. This also ensures we can continue to display the filter segment name if the segment gets deleted. Please make sure to test these cases, though it will change in the future to include more data.

I wasn't sure of a good way to test re-engagement emails locally, so please make sure to test that they get tested.

While working on this I noticed a bug where having premium installed but not active makes it impossible to view a newsletter's stats page. The plugin still tries to hit a premium-only endpoint, resulting in an `Invalid API Endpoint` error. This PR includes a fix for this and I would appreciate if you could test this as well.

image

## Linked PRs

You can use #5175 to add filter segments to emails.

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5512

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/AdminPages/Pages/Newsletters.php
+++ b/mailpoet/lib/AdminPages/Pages/Newsletters.php
@@ -6,6 +6,7 @@ use MailPoet\AdminPages\PageRenderer;
 use MailPoet\AutomaticEmails\AutomaticEmails;
 use MailPoet\Config\Env;
 use MailPoet\Config\Menu;
+use MailPoet\Config\ServicesChecker;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Listing\PageLimit;
 use MailPoet\Newsletter\NewslettersRepository;
@@ -57,6 +58,9 @@ class Newsletters {
   /** @var SubscribersFeature */
   private $subscribersFeature;
 
+  /** @var ServicesChecker */
+  private $servicesChecker;
+
   public function __construct(
     PageRenderer $pageRenderer,
     PageLimit $listingPageLimit,
@@ -69,7 +73,8 @@ class Newsletters {
     NewslettersRepository $newslettersRepository,
     Bridge $bridge,
     AuthorizedSenderDomainController $senderDomainController,
-    SubscribersFeature $subscribersFeature
+    SubscribersFeature $subscribersFeature,
+    ServicesChecker $servicesChecker
   ) {
     $this->pageRenderer = $pageRenderer;
     $this->listingPageLimit = $listingPageLimit;
@@ -83,6 +88,7 @@ class Newsletters {
     $this->bridge = $bridge;
     $this->senderDomainController = $senderDomainController;
     $this->subscribersFeature = $subscribersFeature;
+    $this->servicesChecker = $servicesChecker;
   }
 
   public function render() {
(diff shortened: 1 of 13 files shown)


johnolek:
@rodrigoprimo I believe the tests are only failing because of the current issue with dockerhub. I'll try re-running them again tomorrow if the one running now doesn't finish. I don't expect there to be any true failures but feel free to wait until all is green before starting to review.

rodrigoprimo:
@johnolek, just FYI, this PR needs to be rebased and some conflicts fixed after https://github.com/mailpoet/mailpoet/pull/5184 was merged.

mailpoet/lib/Entities/NewsletterEntity.php
```diff
@@ -495,7 +495,14 @@ public function getOptionValue(string $name) {
 
   public function getFilterSegmentId(): ?int {
     $optionValue = $this->getOptionValue(NewsletterOptionFieldEntity::NAME_FILTER_SEGMENT_ID);
-    return $optionValue ? (int)$optionValue : null;
+    if ($optionValue) {
+      return (int)$optionValue;
+    }
+    $parentNewsletter = $this->getParent();
+    if ($parentNewsletter instanceof NewsletterEntity && $this->getId() !== $parentNewsletter->getId()) {
```
rodrigoprimo, line 502: Would it make sense to check the newsletter type to make sure this is only applied for post notifications? Maybe it doesn't. I'm not sure.

Also, I'm curious as to why we need to check if `$this->getId() !== $parentNewsletter->getId()`? Is there a scenario where `$this->getId() === $parentNewsletter->getId()`?

mailpoet/lib/AdminPages/Pages/Newsletters.php
```diff
@@ -121,7 +127,7 @@ public function render() {
 
     $data['sent_newsletters_count'] = $this->newslettersRepository->countBy(['status' => NewsletterEntity::STATUS_SENT]);
     $data['woocommerce_transactional_email_id'] = $this->settings->get(TransactionalEmails::SETTING_EMAIL_ID);
-    $data['display_detailed_stats'] = $this->subscribersFeature->hasValidPremiumKey() && !$this->subscribersFeature->check();
+    $data['display_detailed_stats'] = $this->subscribersFeature->hasValidPremiumKey() && !$this->subscribersFeature->check() && $this->servicesChecker->isPremiumPluginActive();
```
rodrigoprimo, line 130: Good catch! Would it make sense to highlight this commit in the QA notes so that QA is aware of this fix and can test it if they consider necessary?

rodrigoprimo (approved):
Looks good to me! I just left two small comments, but they are not blockers. Feel free to move this directly to QA if you prefer.

mailpoet/lib/AdminPages/Pages/Newsletters.php
johnolek: Good call! I'll update the description.

mailpoet/lib/Entities/NewsletterEntity.php
johnolek: > Would it make sense to check the newsletter type to make sure this is only applied for post notifications?

My inclination is to keep this as is. I'm pretty sure parent newsletters are only used for post notifications, but if we had others or add others in the future it seems like it would make sense to defer to the parent's filter segment if the child doesn't have one of its own. I don't think it would make much of a difference right now though.

> Also, I'm curious as to why we need to check if $this->getId() !== $parentNewsletter->getId()? Is there a scenario where $this->getId() === $parentNewsletter->getId()?

Thanks for calling this out. There are a couple of tests that set the parent newsletter to itself ([example1](https://github.com/mailpoet/mailpoet/blob/360d9410f915b97c3adb053948167e2552b55fbc/mailpoet/tests/integration/Cron/Workers/SendingQueue/Tasks/NewsletterTest.php#L221-L221), [example2](https://github.com/mailpoet/mailpoet/blob/360d9410f915b97c3adb053948167e2552b55fbc/mailpoet/tests/integration/Cron/Workers/SendingQueue/Tasks/NewsletterTest.php#L240-L240)) that also preprocess the newsletter, resulting in an infinite loop. I'm almost positive we would never see this in the wild, but I would rather play it on the safe side.

Aschepikov (approved):
GJ!

PR #5174, opened 2023-09-22
Update new segment page design [MAILPOET-5398]

rodrigoprimo:
## Description

This PR updates the design of the page used to create new segments and edit existing ones.

## Code review notes

I tried to implement the new design by creating components in the `common` directory that can be reused in the future as it is likely that we will have other pages using the same structure. I'm not super happy with the names that I used for those components. Let me know if you have any suggestions.

Per the ticket description, I'm using the `PremiumModal` component instead of the premium banner. The design of this component is different than what we have in the Figma files. @NeosinneR, maybe we need to create a separate ticket to update the design of the modal if one doesn't exist already?

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5398]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5398]: https://mailpoet.atlassian.net/browse/MAILPOET-5398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/css/src/components-plugin/_common.scss
```diff
@@ -141,3 +141,25 @@ body .components-modal__screen-overlay {
...
+.mailpoet-heading {
+  align-items: center;
+  align-self: stretch;
+  display: flex;
+  justify-content: space-between;
+  padding: 16px 0;
+
+  h1.wp-heading-inline {
+    align-items: center !important;
+    display: flex !important;
```
johnolek: Is there a way to do this without `!important`?

mailpoet/assets/js/src/segments/dynamic/editor.tsx
```diff
@@ -28,22 +29,32 @@ export function Editor(): JSX.Element {
...
+        gap="16px"
+      >
+        <FlexBlock>
+          <h1 className="wp-heading-inline">
+            <Button
+              id="mailpoet-segments-back-button"
+              icon={chevronLeft}
+              href={`#${returnPage}`}
+              label={__('Return to previous page', 'mailpoet')}
+            />
```
johnolek, line 51: Is there no way to use a `<Link>` or something else that works within the router? My local ESLint is seeing an issue with this, but maybe it's not an issue if our QA script doesn't see it: `ESLint: Invalid type "any" of template literal expression.(@typescript-eslint/restrict-template-expressions)`.

mailpoet/tests/acceptance/Segments/ManageSegmentsCest.php
```diff
@@ -425,7 +425,7 @@ public function createSegmentFromTemplate(\AcceptanceTester $i) {
     $i->waitForElement('.mailpoet-templates-card-grid div:first-child');
     $i->click('.mailpoet-templates-card-grid div:first-child');
     $i->waitForText('Recently Subscribed');
-    $i->click('Back');
+    $i->click('#mailpoet-segments-back-button');
```
johnolek, line 428: Does this work if you use the label text instead, `Return to previous page`? No big deal if not, but I like the practice of using text that users would actually see when possible instead of IDs.

mailpoet/assets/js/src/segments/dynamic/form.tsx
```diff
@@ -25,6 +22,8 @@ import {
   Segment,
   SubscriberCount,
 } from './types';
+import { FieldsSection } from '../../common/fields_section/fields_section';
+import { FieldWrapper } from '../../common/fields_section/field_wrapper';
```
johnolek: I don't think you should need the `../../`

mailpoet/assets/js/src/segments/dynamic/form.tsx
```diff
@@ -82,130 +81,136 @@ export function Form({ segmentId }: Props): JSX.Element {
...
+        {errors.length > 0 && (
+          <APIErrorsNotice
+            errors={errors.map((error) => ({ message: error }))}
+          />
+        )}
+        <FieldsSection
+          title={__('Details', 'mailpoet')}
+          description={__('What do you want to call this segment?', 'mailpoet')}
+        >
+          <FieldWrapper slug="name" title={__('Name', 'mailpoet')}>
```
johnolek, line 101: It's not intuitive to me that the `slug` here would control the label's `for`, and that the `Input`'s `id` needs to be `field_{slug}`. I'm not sure what a better solution would be though. Maybe having a component that renders both a field wrapper and an input?

mailpoet/assets/js/src/segments/dynamic/form.tsx
```diff
@@ -82,130 +81,136 @@ export function Form({ segmentId }: Props): JSX.Element {
...
-              void handleSave(segmentId);
-            }}
-            isDisabled={
-              !isFormValid(segment.filters) ||
-              subscriberCount.count === undefined
-            }
-          >
-            {__('Save', 'mailpoet')}
-          </Button>
+        <div className="mailpoet-admin-fields-bottom">
```
johnolek: Maybe `mailpoet-admin-fields-footer`?

mailpoet/assets/js/src/segments/dynamic/form.tsx
```diff
@@ -169,25 +176,37 @@ export function Form({ segmentId }: Props): JSX.Element {
...
+                {(!MailPoet.premiumActive ||
+                  !MailPoet.hasValidPremiumKey ||
+                  MailPoet.subscribersLimitReached) && (
+                  <LockedBadge text={__('UPGRADE', 'mailpoet')} />
+                )}
+
+                {showPremiumModal && (
+                  <PremiumModal onRequestClose={closePremiumModal}>
+                    {__(
+                      'Multiple filters per segment are not available in the free version of the MailPoet plugin.',
```
johnolek: We should probably refer to them as `conditions` instead of `filters` to match the rest of the UI.

johnolek (changes requested):
Looking good to me overall! I left a few specific comments/questions, most of which shouldn't require any changes.

I noticed a few general things:

### Validation errors not clearing
If you get an API error when trying to create a segment, e.g. no name or duplicate name, the error can reappear after it disappears when you navigate away.

1. Create a new custom segment
2. Add valid filter but no name
3. Click save
4. Observe the error notice and wait for it to disappear
5. Hit the back button to get back to the templates page
6. Observe the error notice has reappeared
7. It will reappear again if you click back into creating a new custom segment

### No default filter when creating from scratch
The previous behavior was to include a WordPress user role filter as the default, but now I'm seeing just a blank when navigating to the page using the "create from scratch" buttons on the template page.
image
I think it would be nice to have some kind of message like "add a condition to get started" if we want to have this state, but we might want to just keep the old behavior because free users can't use the "Add a condition" button at all, even to add the first.

Interestingly, I still get the default filter if I visit the new segment URL directly instead of using one of the buttons on the template page.

### Some selects start too small
Selects that don't start with a default value start just big enough to contain placeholder text, and the items have long scrollbars. This ends up being way too small in certain cases, the most notable that I found being lists and emails:

image

Once a longer value has been selected it's not a problem since the select gets resized.

NeosinneR:
> The design of this component is different than what we have in the Figma files. @NeosinneR, maybe we need to create a separate ticket to update the design of the modal if one doesn't exist already?

@rodrigoprimo yes, the design differs as during initial automations phases we decided to go with the existing modal component from `@wordpress/components` instead of customizing it or building a new one with a custom design. I don't think there is a ticket to update it, but if you could please create one, that would be good to keep track of.

rodrigoprimo:
Thanks, @NeosinneR. Ticket here: https://mailpoet.atlassian.net/browse/MAILPOET-5604

mailpoet/tests/acceptance/Segments/ManageSegmentsCest.php
rodrigoprimo: I'd be curious to hear more about why do you prefer to use the text instead of IDs? Is it to make sure that the interface is displaying what we expect it to display? Maybe it is even a QA best practice?

Here, I opted to use the ID instead of the text to avoid a scenario where we change the text, push the commits to GitHub only to realize that we broke an acceptance test. The test would be broken, but the interface would actually be working. At the end of the day, it is a trade-off, and I understand if the recommendation is to use the text instead of the ID.

mailpoet/assets/css/src/components-plugin/_common.scss
rodrigoprimo: I just noticed that `!important` is needed only for `display` and not for `align-items`. So I will push a commit changing that.

But regarding your question, I don't know if there is a way to do this without `!important`. If there is, I'm missing it. http://mp3.localhost/wp-admin/css/common.css sets `display: inline-block` for this class and it is enqueued after MP's CSS.

mailpoet/assets/js/src/segments/dynamic/editor.tsx
rodrigoprimo: I'm failing to see the relation between using `<Link>` and the ESLint error you are seeing. Could you please elaborate? I have a vague memory that I tried to use `<Link>` but was not able to, but unfortunately, I don't remember the details.

Anyway, I will push a commit to declare the type of `returnPage` as string and that should remove the error that you mentioned. PHPStorm is able to infer the type of this variable as `string` so I don't know if declaring it is needed, but I guess it doesn't hurt.

Let me know if I'm missing something.

mailpoet/assets/js/src/segments/dynamic/form.tsx
rodrigoprimo: PHPStorm adds the imports with `../` automatically and sometimes I forget to change it. This should be fixed now. I need to check if there is a way to change PHPStorm's default behavior.

mailpoet/assets/js/src/segments/dynamic/form.tsx
rodrigoprimo: Updated

mailpoet/assets/js/src/segments/dynamic/form.tsx
rodrigoprimo: Good point, I just pushed a commit changing this

mailpoet/assets/js/src/segments/dynamic/form.tsx
rodrigoprimo: I agree that this solution is not great precisely because of the reason that you pointed out. That being said, if I understand your suggestion, I'm not sure if it would be much better because then we would need to create one component for each of the calls to `<FieldWrapper>`, no? Or at least, for each of the calls that have different components as children. This would remove the flexibility and reusability of `<FieldWrapper>`. Maybe we can ask other developers if they have another suggestion? I can't think of anything with my limited React knowledge.

rodrigoprimo:
> Validation errors not clearing

Good catch! I'm able to reproduce this on trunk, so I believe we can address it as a separate ticket. I went ahead and created it, but please let me know if you believe we should address this in this PR.

https://mailpoet.atlassian.net/browse/MAILPOET-5605

> No default filter when creating from scratch

I'm unable to reproduce this behavior. I'm seeing the default filter when creating from scratch. See the screencast below. Is there another step that I should take to reproduce this problem? I tested it on my local environment and also on Jurassic Ninja.

[Screencast from 2023-09-25 15-04-00.webm](https://github.com/mailpoet/mailpoet/assets/77215/95f117db-b9e4-4042-9fa8-c70241739e0c)

> Some selects start too small

Good catch, I had missed this. I'm not sure what is the best solution to this problem. I believe I fixed it for the <EmailClickStatisticsFields> in 3e7399c440f6bd5c6d84b558f1954bc364e7bb2e. Could you please take a look and tell me what you think? If you like this approach, I will implement it for the rest of the segments where this is needed.

rodrigoprimo:
@johnolek, I believe I addressed all of your comments. Can you please take a look again when you have a moment?

mailpoet/assets/css/src/components-plugin/_common.scss
johnolek: Not a problem!

mailpoet/assets/js/src/segments/dynamic/editor.tsx
johnolek: Sorry for the confusion! I meant for these to be two separate ideas. Thanks for making the type super clear, even though it's probably not needed, and it's fine to keep this as a button if a `<Link>` wasn't going to work for whatever reason.

mailpoet/assets/js/src/segments/dynamic/form.tsx
johnolek: If you figure this out please let me know! This bugs me too and I just spent a few minutes trying to figure this out but didn't get anywhere.

mailpoet/assets/js/src/segments/dynamic/form.tsx
johnolek: Yes, my idea would require creating a new component for each configuration of children, which I agree wouldn't be much better.

I may not be fully understanding the idea behind `FieldWrapper`. Right now it's being used for the name, description, and conditions on the segment editor page. In the case of the [conditions](https://github.com/mailpoet/mailpoet/blob/3e7399c440f6bd5c6d84b558f1954bc364e7bb2e/mailpoet/assets/js/src/segments/dynamic/form.tsx#L141-L141), it doesn't seem like it's doing much. Removing it doesn't have a visual effect that I can see, and although you're passing it a `slug` I don't think it gets used because it's not receiving a `title`. 

Some combinations also might not make sense. For example, `FieldWrapper` will render multiple children, but it would never make sense to give it multiple `Input`s because it only ever renders a single label/description.

Right now it seems like the wrapper is mostly useful as a wrapper around text inputs, which is why I think it feels simpler to create something more specific like a `TextInputField` that would include the wrapper and the input, but again I think I don't understand the idea for how it should be used more generally.

This could definitely be a good thing to bring up with the group, especially if this is a component we'll be using a lot in newly designed pages.

johnolek:
Thanks for all the changes!

> I went ahead and created it, but please let me know if you believe we should address this in this PR.

Keeping it as a separate ticket sounds just fine to me, and thanks for creating the ticket! It's a bit of an edge case anyway.

> I'm unable to reproduce this behavior. I'm seeing the default filter when creating from scratch. Is there another step that > I should take to reproduce this problem?

Yes, although I didn't realize it at the time. It requires visiting the editor, going back, and then returning to the editor. This is from a jurassic ninja site:
https://github.com/mailpoet/mailpoet/assets/8656640/c32dbec3-bce8-49b5-9102-1e7c270f70dc

This doesn't feel like a blocker so if it would be time consuming to fix I think it would be fine to create a separate ticket.

The last thing is deciding if we want to do anything different with the `FieldWrapper`. It definitely works fine as is and we can always refactor in the future, so I'll leave that decision up to you.

johnolek (approved):

johnolek:
@rodrigoprimo Sorry, I forgot to address your last comment. The changes in https://github.com/mailpoet/mailpoet/commit/3e7399c440f6bd5c6d84b558f1954bc364e7bb2e look good to me and I'm ok with applying the same to other components.

johnolek (changes requested):
Sorry for the back and forth. Requesting changes just for applying the select size changes to the other components, I don't think anything else is a blocker.

mailpoet/assets/js/src/segments/dynamic/form.tsx
rodrigoprimo: > I may not be fully understanding the idea behind FieldWrapper.

My goal with creating <FieldWrapper> was to create a component that, together with <FieldsSection>, would make it easier to recreate the same design for similar pages.

> In the case of the conditions, it doesn't seem like it's doing much.

I agree, and this is something that I hadn't considered.

> Removing it doesn't have a visual effect that I can see, and although you're passing it a slug I don't think it gets used because it's not receiving a title.

Good point, I failed to notice this when I made title optional. I updated the component to make slug optional as well.

> For example, FieldWrapper will render multiple children, but it would never make sense to give it multiple Inputs because it only ever renders a single label/description.

I'm not sure if there is a way to limit the number of children that can be passed to this component. In the future, I think we can improve it to accept a type of input field and then the component itself will create the field.

> Right now it seems like the wrapper is mostly useful as a wrapper around text inputs, which is why I think it feels simpler to create something more specific like a TextInputField that would include the wrapper and the input, but again I think I don't understand the idea for how it should be used more generally.

I assumed it could be used as well for other types of inputs, like a single select box. But I'm not sure as we currently don't have such a field.

I agree that the `<FieldWrapper>` abstraction needs to be improved and maybe even completely changed. My suggestion is that we leave it as is for now and reconsider it when it is time to implement another page with the same design. I feel like then we will have more data to judge what is the best path for this component.

rodrigoprimo:
> Yes, although I didn't realize it at the time. It requires visiting the editor, going back, and then returning to the editor. This is from a jurassic ninja site:
https://github.com/mailpoet/mailpoet/assets/8656640/c32dbec3-bce8-49b5-9102-1e7c270f70dc
> This doesn't feel like a blocker so if it would be time consuming to fix I think it would be fine to create a separate ticket.

Thanks for the additional details. I was able to reproduce this on trunk as well, so I went ahead and created a separate ticket: https://mailpoet.atlassian.net/browse/MAILPOET-5614

> Requesting changes just for applying the select size changes to the other components, I don't think anything else is a blocker.

Sounds good to me. I will just wait for @shivanarrthine's reply to the question you asked on Slack (p1695849671497739/1695238625.419619-slack-C05BG9S6B96) as this could impact how we are going to solve the issue with the size of the selects before implementing this change for the rest of the conditions.

mailpoet/assets/js/src/segments/dynamic/form.tsx
johnolek: Thanks for your thoughtful reply! I agree with leaving it as is for now and modifying it as needed as we need to implement different designs.

rodrigoprimo:
Just documenting that I had to force push to fix a conflict with trunk.

rodrigoprimo:
@johnolek, I believe I addressed all the pending items. I fixed the layout of the remaining segments that have select boxes with longer text and fixed the contrast issue we discussed on Slack. I added two new commits. The rest I did not change (besides having to force push due to the conflict with trunk). Can you please check this PR again?

johnolek (dismissed):
Thanks for the updates, this looks good to me!

rodrigoprimo:
@johnolek, due to https://github.com/mailpoet/mailpoet/pull/5184 being merged, I had to fix a few conflicts and add a new commit. Just to be on the safe side, could you please take a final look at this PR before we move it to QA?

johnolek:
@rodrigoprimo Looks good to me! I'll assign to QA.

johnolek (dismissed):

Aschepikov:
Hi @rodrigoprimo I found some issues can you take a look at them
In the **WordPress user role** filter, the user role field is too wide when there is a second and more line
image
The **does not contain, does not start with, does not end with** cant be visible completely 
image
The **not in the last** cant be visible completely 
image
The **less/more than or equal** cant be visible completely 
image
In the **Number of reviews** filter, the 3 and the 4 fields are not wide enough
image

rodrigoprimo:
Thanks for the detailed testing, @Aschepikov. I pushed a few new commits to fix the issues that you found. Can you please test again?

Aschepikov (approved):
GJ!

PR #5175, opened 2023-09-22
Add filter segment select to send step [MAILPOET-5510]

johnolek:
## Description

This PR adds the ability to select a filter segment on the send step for standard, post notification, and re-engagement emails.

This was originally included in https://github.com/mailpoet/mailpoet/pull/5141, but I'm closing that one since it included two separate tickets and we decided we should keep them separate.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5510](https://mailpoet.atlassian.net/browse/MAILPOET-5510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5510]: https://mailpoet.atlassian.net/browse/MAILPOET-5510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/newsletters/send/filter_segment.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+export function FilterSegment({
+  item,
+  onValueChange,
+  field,
+}: FilterSegmentProps) {
+  const currentFilterSegmentId = item?.options.filterSegmentId;
+  const premiumFeaturesAvailable: boolean = useMemo(
+    () =>
+      window.mailpoet_premium_active && window.mailpoet_has_valid_premium_key,
```
rodrigoprimo: Do we also need to check here if the site reached the subscribers limit?

rodrigoprimo: It would be nice to have a constant or a help function that we could use to check that. We do have a few attempts at that that might need consolidation:

https://github.com/mailpoet/mailpoet-premium/blob/bdb626b23e37940251b2e084cb5c65f37c5d46ac/assets/js/src/premium.ts#L6
https://github.com/mailpoet/mailpoet/blob/f8c7adf5a6cc697c4b84307bb4a2cc8e269906cc/mailpoet/assets/js/src/common/premium_modal/index.tsx#L29
https://github.com/mailpoet/mailpoet/blob/9930f2915c5fca19a4d03433f964432f2c6b655c/mailpoet/assets/js/src/common/premium_modal/upgrade_info.ts#L21

I believe there is even one more that I tried to use when creating https://github.com/mailpoet/mailpoet/pull/5174, but it didn't work straightaway and I didn't want to spend more time on that PR.

I don't think you necessarily need to use a helper function or constant here since this seems to be a bigger problem. But mentioning it in case you want to give it a go.

rodrigoprimo: I don't recall the details, but I believe it is recommended that we use the MailPoet object instead of window to check if those flags are true or false.

mailpoet/assets/js/src/newsletters/send/filter_segment.tsx
```diff
@@ -0,0 +1,147 @@
...
+  field: Field;
+};
+
+export function FilterSegment({
+  item,
+  onValueChange,
+  field,
+}: FilterSegmentProps) {
+  const currentFilterSegmentId = item?.options.filterSegmentId;
+  const premiumFeaturesAvailable: boolean = useMemo(
```
rodrigoprimo: I'm asking this more to learn React than anything else, but why do we need `useMemo()` here? My limited understanding is that it should be used for performance optimization when whatever is calculated inside it is an expensive calculation, which doesn't seem to be the case here, but I'm probably missing something.

My concern without checking the details is that this could add more overhead than benefits. I don't know if that is the case here or not.

mailpoet/assets/js/src/newsletters/send.tsx
```diff
@@ -227,6 +228,12 @@ class NewsletterSendComponent extends Component<
 
   getThumbnailPromise = (url) => this.state?.thumbnailPromise ?? fromUrl(url);
 
+  get premiumFeaturesAvailable() {
+    return (
+      window.mailpoet_premium_active && window.mailpoet_has_valid_premium_key
```
rodrigoprimo: Same as in a previous comment where we are using the same two booleans.

rodrigoprimo (commented):
Looks good to me! Thanks for splitting the PRs and for creating more granular commits in this PR. It made reviewing the code changes much easier. I appreciate the extra effort you took.

I left just a few small comments for us to discuss before moving this PR forward.

mailpoet/assets/js/src/newsletters/send/filter_segment.tsx
johnolek: Thanks for catching this! I'm going to use the existing `premiumValidAndActive` https://github.com/mailpoet/mailpoet/blob/MAILPOET-5510%2Fupdate-sending-step-only/mailpoet/assets/js/src/common/premium_modal/index.tsx#L28-L28, which I think captures what we want.

Part of what I was doing here is I needed this to match what we have in `send.tsx` (get premiumFeaturesAvailable), so I updated that as well.

This change is in commit bf6d858b28680b00c06ab9769ba282a82c2cda87

mailpoet/assets/js/src/newsletters/send/filter_segment.tsx
johnolek: Thank you for bringing this up! You're right, it's not necessary and probably does overcomplicate things unnecessarily. Sometimes I reach for hooks reflexively and this was one of those cases.

In any case, this is simplified now that I'm pulling in the get `premiumFeaturesAvailable` value.

johnolek:
@rodrigoprimo Thank you for the review! I think everything should be addressed in commit bf6d858b28680b00c06ab9769ba282a82c2cda87. I may not have consolidated as much as I could have, but relying on this existing value seems like a step in the right direction. Let me know if you think of anything else!

rodrigoprimo (dismissed):

rodrigoprimo:
@NeosinneR, John and I are assuming that it was due to an oversight that we were not checking if the site reached the subscriber limit when deciding whether or not to display the Google Analytics fields. Is that the case?

See https://github.com/mailpoet/mailpoet/pull/5175/commits/bf6d858b28680b00c06ab9769ba282a82c2cda87 and https://github.com/mailpoet/mailpoet/pull/5175#discussion_r1335885642

NeosinneR:
> John and I are assuming that it was due to an oversight that we were not checking if the site reached the subscriber limit when deciding whether or not to display the Google Analytics fields. Is that the case?
> 
> See [bf6d858](https://github.com/mailpoet/mailpoet/commit/bf6d858b28680b00c06ab9769ba282a82c2cda87) and https://github.com/mailpoet/mailpoet/pull/5175#discussion_r1335885642

@rodrigoprimo @johnolek Nice catch and thanks for fixing it. It was indeed an oversight.

rodrigoprimo:
@johnolek, just FYI, this PR needs to be rebased and some conflicts fixed after https://github.com/mailpoet/mailpoet/pull/5184 was merged.

johnolek:
Thanks for the ping, @rodrigoprimo! I've rebased, updated and force pushed. No significant changes were necessary.

Aschepikov (approved):
GJ!

PR #5176, opened 2023-09-22
Add link to create new segment on send step [MAILPOET-5512]

johnolek:
## Description

This adds a link to create a new filter segment from the sending step. After creating a new segment, the user is returned to the send step with the filter segment pre-filled.

Originally included in https://github.com/mailpoet/mailpoet/pull/5141 but I closed that PR to create 2 instead.

I included a quick fix in this PR (f4752d3e20e4b2914407a98a97d954682c20c2d9) to ensure the filter segment tags show up in the listing for standard newsletters.

## Code review notes

If you can think of a simpler way of handling saving the email before the redirect I'm all ears.

I recommend hiding whitespace changes when reviewing because I wrapped some existing code in a React context and the diff with whitespace doesn't make sense.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5512

## After-merge notes

_N/A_

## Tasks


mailpoet/assets/js/src/newsletters/send.tsx
```diff
@@ -793,84 +810,92 @@
...
+                {__('or simply', 'mailpoet')}
+                &nbsp;
+                <a
+                  className="mailpoet-link"
+                  href={
+                    MailPoet.FeaturesController.isSupported(
+                      'gutenberg_email_editor',
+                    ) && wpPostId
+                      ? `post.php?post=${wpPostId}&action=edit`
+                      : `?page=mailpoet-newsletter-editor&id=${this.props.match.params.id}`
```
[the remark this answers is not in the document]
johnolek: This isn't a change introduced by this PR (I think it's only being flagged because this part of the code changed indentation level) and I'm not entirely sure what the concern is. If it's about `wpPostId`, that value should be coming straight from the database and it's an integer column, so I don't think it's a risk.

johnolek:
@rodrigoprimo I know I've just sent a lot of review requests your way. Please let me know if you'd like me to reassign any of them, though I'm hoping they shouldn't take too much time, especially with your level of familiarity with this part of the codebase.

mailpoet/assets/js/src/newsletters/send/filter-segment.tsx
```diff
@@ -138,6 +148,33 @@ export function FilterSegment({
...
+                className="mailpoet-link"
+                key={i}
+                rel="noopener noreferrer"
+                onClick={(event) => {
+                  event.preventDefault();
+                  context.saveDraftNewsletter(() => {
+                    window.location.href = `admin.php?page=mailpoet-segments#/new-segment?newsletterId=${item.id}`;
+                  });
+                }}
+                href={`admin.php?page=mailpoet-segments#/new-segment?newsletterId=${item.id}`}
```
rodrigoprimo: Is this attribute needed? If so, maybe we can use a variable for the URL, as it is the same as in line 167? It is just a detail, but I'm just trying to avoid a situation in the future where we update the URL in one place but don't update in the other creating a minor inconsistency.

rodrigoprimo: I wonder if there is a way to [`NEW_DYNAMIC_SEGMENT`](https://github.com/mailpoet/mailpoet/blob/2b5cf32b12226953b8ca2e04b1ffcc4e4eb76ea5/mailpoet/assets/js/src/segments/routes.ts#L1) here. Probably not, since it is a different React app.

mailpoet/assets/js/src/segments/dynamic/editor.tsx
```diff
@@ -48,6 +52,12 @@ export function Editor(): JSX.Element {
               icon={chevronLeft}
               href={`#${returnPage}`}
               label={__('Return to previous page', 'mailpoet')}
+              onClick={(event) => {
```
rodrigoprimo, line 55: Instead of using the `onClick` event to return to the newsletter page, would it make sense to combine this logic with the logic above to define the value of `returnPage`? My first reaction without much thought is that returning to the newsletter page should be another possible value of `returnPage`.

rodrigoprimo (commented):
Looks good to me! I left a few comments with small things for us to discuss.

> If you can think of a simpler way of handling saving the email before the redirect I'm all ears.

The way you implemented it makes sense to me. That being said, I don't think I'm the best person to tell whether or not there is a better approach. So my plan is to ask for a second review specifically about this after we discuss the points that I raised.

mailpoet/assets/js/src/segments/dynamic/editor.tsx
johnolek: That's what I wanted to do as well, but I don't think it's possible because of how this component is wrapped in a `HashRouter`. Since the root of this react app is essentially `admin.php?page=mailpoet-segments#`, I don't think there's a way to specify a `returnPage` that would navigate to a different page.

mailpoet/assets/js/src/newsletters/send/filter-segment.tsx
johnolek: Details are important! Thank you for this comment. I extracted it to a variable in f6ac44a0c1690ec10d3d00d1daad330593baed31.

It wouldn't be strictly necessary to have the `href` since the click event should always get intercepted, but without the `href` the browser doesn't see it as an interactive element, so it doesn't exist in the tab order among other issues.

mailpoet/assets/js/src/segments/dynamic/editor.tsx
rodrigoprimo: Fair point, thanks for looking into it.

rodrigoprimo (dismissed):

rodrigoprimo:
I'm assigning this PR to @JanJakes for a second look, specially around the part below as I'm not sure if there is a better approach.

> If you can think of a simpler way of handling saving the email before the redirect I'm all ears.

mailpoet/assets/js/src/newsletters/send.tsx
```diff
@@ -793,84 +810,92 @@ class NewsletterSendComponent extends Component<
...
-            item={this.state.item}
-            loading={this.state.loading}
-            onChange={this.handleFormChange}
-            onSubmit={this.handleSave}
+          <SendContext.Provider
+            /* not sure how to resolve this in a class component, the suggestion is to use useMemo */
+            /* eslint-disable-next-line react/jsx-no-constructed-context-values */
+            value={{
+              saveDraftNewsletter: this.saveDraftNewsletter,
+            }}
```
JanJakes: The issue is that you're creating a new object for each render. Any solution that can prevent that will be fine. I don't know much about class components, but probably sth like `this.value`, where `value = { saveDraftNewsletter: this.saveDraftNewsletter };` could do it? At least in case it's only `render` that's called again and again.

JanJakes:
@johnolek @rodrigoprimo Could you please point me to what is the problem, exactly? I found a few places with a redirect but without additional context, I'm not sure what is the issue. Is it the fact that React context is used? Or the "what to do in class component" when `useMemo` is recommended (added a comment)?

mailpoet/assets/js/src/newsletters/send.tsx
```diff
@@ -764,6 +774,13 @@ class NewsletterSendComponent extends Component<
   updatePendingApprovalState = (mssKeyPendingApproval: boolean): void =>
     this.setState({ mssKeyPendingApproval });
 
+  saveDraftNewsletter = (afterSaveCallback: () => void) => {
```
JanJakes: Could this be an async function instead? Passing callback seems a bit old-school, but maybe we're using old code here, that is not easy to wrap with a promise...?

johnolek:
> @johnolek @rodrigoprimo Could you please point me to what is the problem, exactly? I found a few places with a redirect but without additional context, I'm not sure what is the issue. Is it the fact that React context is used? Or the "what to do in class component" when `useMemo` is recommended (added a comment)?

@JanJakes Thanks for taking a look at this! There's not necessarily a problem. My main question was about the React context, specifically if there's a more straightforward or more idiomatic way to achieve the same behavior, basically if any of this would be violating best practices. I was wondering because it doesn't seem like we use contexts in many other places in the codebase and I haven't used them before myself.

mailpoet/assets/js/src/newsletters/send.tsx
johnolek: In 58978ec921db86709636ff4cc03ebc4c6c7cdd49 I changed this so the value exists in the component's state. Does that seem reasonable to you?

mailpoet/assets/js/src/newsletters/send.tsx
johnolek: Thanks for pointing this out! I updated this in ee7e9fee7a7bd971c888cb57d70aafdff438c7a8. Let me know if it looks off in any way.

JanJakes:
@johnolek Ahh, the question "whether to use context" here is probably related to options of solving it without it. I quickly checked, and given that it's needed in a field that's initialized using `component: withBoundary(FilterSegment)`, props are not a good way here. As for a function that could be just imported, that doesn't work easily either, considering that `this.something` calls are needed. Then context seems fine to me 👍 LGTM, feel free to move it forward.

mailpoet/assets/js/src/newsletters/send.tsx
JanJakes: @johnolek Yeah, that looks good. Another good idea how to make the object identity stable 👍

mailpoet/assets/js/src/newsletters/send/filter-segment.tsx
```diff
@@ -164,9 +164,9 @@ export function FilterSegment({
                 className="mailpoet-link"
                 key={i}
                 rel="noopener noreferrer"
-                onClick={(event) => {
+                onClick={async (event) => {
                   event.preventDefault();
-                  context.saveDraftNewsletter(() => {
+                  await context.saveDraftNewsletter().then(() => {
```
JanJakes: The only tiny nitpick is I'd personally convert `await something().then(...)` to:

```ts
await something();
// here do the "then"
```

JanJakes (dismissed):

mailpoet/assets/js/src/newsletters/send/filter-segment.tsx
johnolek: I always appreciate "nitpick"s! I modified the commit so now 2a1eff63eb56a4419d8f63ec0352410086043f82 has it like you describe.

johnolek:
Thanks @JanJakes for your input/confirmation! I made one small change to address your [comment](https://github.com/mailpoet/mailpoet/pull/5176#discussion_r1356934124). Could you (or @rodrigoprimo) please approve again when you have a chance?

JanJakes (approved):

Aschepikov:
HI @johnolek can you rebase it pls to the current master

johnolek:
@Aschepikov Rebased and pushed, let me know if you need anything else!

johnolek:
Sorry @Aschepikov, I just pushed again because I didn't have all the latest changes from `trunk` locally the last time I rebased. This branch should now be fully up to date.

Aschepikov (approved):
GJ!

PR #5185, opened 2023-09-28
Fix message displayed when a new segment is created [MAILPOET-5615]

rodrigoprimo:
## Description

There was a bug and the code was displaying `Segment succesfully updated!` instead of `Segment succesfull added!` when the user created a new segment.

This PR fixes this problem. The issue was that Number(match.params.id) returns `NaN` instead of `undefined` when there is no segment ID. So the check inside handleSave() was always calling messages.onUpdate().

Since the segment ID was never used. I replaced it with a boolean and renamed the variable from `segmentId` to `isNewSegment` to better indicate how it is used.

@NeosinneR, just FYI, because of this bug, we were never triggering [this event](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/segments/dynamic/store/actions.ts#L127) when a segment was created. Now, we will trigger it, but `type` and `subType` will be unknown as the structure of `data` changed (probably when we added support for multiple conditions).

## Code review notes

_N/A_

## QA notes

Check that the proper messages are displayed when creating and updating a segment. It might not be necessary to manually check this, as we have acceptance tests covering it.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5615]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5615]: https://mailpoet.atlassian.net/browse/MAILPOET-5615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/editor.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/editor.tsx
@@ -27,6 +27,9 @@ export function Editor(): JSX.Element {
     };
   }, [match.params.id, pageLoaded, pageUnloaded]);
 
+  const isNewSegment =
+    match.params.id === undefined || Number.isNaN(Number(match.params.id));
+
   return (
     <>
       <Background color="#fff" />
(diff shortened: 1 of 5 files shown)


alex-mpoet (approved):
Looks fine to me. Since it's covered by acceptance tests, let's skip QA.

PR #5186, opened 2023-09-28
Fix displaying default condition when adding a segment [MAILPOET-5614]

rodrigoprimo:
## Description

Whenever the user goes to add a new segment, MP displays `WordPress user role` as the default condition. But there was a
bug that was preventing the default condition from being displayed, if the user left the page by clicking in the back button, and then returned by clicking again on the `Create custom segment` button.

This bug was introduced in dba4ba4dfc4ef6f5ce8120eef014710f966292ae. It added code to unset the segment when the user left the segment editing page. This was necessary as, in that PR, we started using the same store for all the dynamic segment pages. The problem is that the added code sets the segment to an empty array and actually we need to restore the initial state which is a segment with the `WordPress user role` selected as its single filter. This is what is done in this PR.

## Code review notes

_N/A_

## QA notes

Check the Jira ticket for steps on how to reproduce this issue. Ensure it is not an issue anymore when testing this PR.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5614]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5614]: https://mailpoet.atlassian.net/browse/MAILPOET-5614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/store/initial-state.ts
+++ b/mailpoet/assets/js/src/segments/dynamic/store/initial-state.ts
@@ -9,6 +9,18 @@ import {
 
 declare let window: SegmentFormDataWindow;
 
+export function getSegmentInitialState() {
+  return {
+    filters_connect: SegmentConnectTypes.AND,
+    filters: [
+      {
+        segmentType: SegmentTypes.WordPressRole,
+        action: SubscriberActionTypes.WORDPRESS_ROLE,
+      },
+    ],
+  };
+}
+
 export const getInitialState = (): StateType => ({
   automations: window.mailpoet_automations,
   products: window.mailpoet_products,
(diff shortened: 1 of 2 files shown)


brezocordero (dismissed):
Thanks for the fix.
The code looks good. 
I have tested it fixes the bug only on current editor. From the code, it should also work with the new one.

Aschepikov:
Hi @rodrigoprimo  there are some conflicts can you resolve them pls

rodrigoprimo:
@Aschepikov, done!

Aschepikov (approved):
GJ!

PR #5187, opened 2023-09-28
Fix: stop persisting segment error messages across different pages [MAILPOET-5605]

rodrigoprimo:
## Description

In #5145, we started using the same Redux store for all segment pages. This created an unintended side effect that the error messages displayed when the user is creating a segment are persisted across the other different segment pages.

So, for example, if a user tries to create a segment with a name that already exists and then clicks the back button to go back to the segment templates or segment listing page, the error message will still be displayed.

This PR fixes this problem by emptying the error messages before unloading the page to edit or create segments.

## Code review notes

I'm not sure if this is the best way to approach this when we have the same Redux store for multiple pages. This means that we will have to remember to manually reset the state of everything that is used on more than one page. I wonder if there is another more standard way of doing this? Please let me know if you have any thoughts.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5605]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5605]: https://mailpoet.atlassian.net/browse/MAILPOET-5605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JanJakes (approved):
LGTM.

> I'm not sure if this is the best way to approach this when we have the same Redux store for multiple pages. This means that we will have to remember to manually reset the state of everything that is used on more than one page. I wonder if there is another more standard way of doing this? Please let me know if you have any thoughts.

@rodrigoprimo Also not sure. One thing that comes to mind is — could we reset some part of the store on every page (route) change? Would it make sense to have a part of the store per-page only? Something like `page_state: { ... }` property on the Redux state, where all of the per-page data would be saved 🤔

rodrigoprimo:
> @rodrigoprimo Also not sure. One thing that comes to mind is — could we reset some part of the store on every page (route) change? Would it make sense to have a part of the store per-page only? Something like page_state: { ... } property on the Redux state, where all of the per-page data would be saved 🤔

That looks like a good approach to me. I will keep it in mind if I have to make more modifications to this store in the future. In this particular case, we can maybe even create a `editor_page_state: { ... }` property to hold everything that relates to the editor page, and it is not used anywhere else.

Aschepikov (approved):
GJ!

PR #5197, opened 2023-10-03
Add Comment trigger [PREMIUM-248]

websupporter:
## Description
Adds the comment trigger as described in [PREMIUM-248]

## Code review notes
So far, the subject transformer method would always need to return a subject or throw an error. For the CommentToSubscriber transformer, I wanted the possibility to not return a subscriber, as we discussed with @NeosinneR not to "create" subscribers on the fly. Therefore the interface changed to allow for `null`.

In its current implementation (e31d914cad89c3ce696cfeacbf122e7e50ccfc8c), that means the automation would start and fail at the first step where a subscriber subject is required. Another idea would be to check in the TriggerHandler if we have all required subjects for all existing steps in an automation before starting the automation. This would mean not to start the run at all. I am undecided between these two approaches.

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/796

## Linked tickets

[PREMIUM-248]

## After-merge notes

_N/A_

## Tasks

[PREMIUM-248]: https://mailpoet.atlassian.net/browse/PREMIUM-248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PREMIUM-248]: https://mailpoet.atlassian.net/browse/PREMIUM-248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Engine/WordPress.php
```diff
@@ -96,4 +118,38 @@ public function getAvatarUrl($idOrEmail, $args = null) {
...
+    return get_post_types($args, $output, $operator);
+  }
+
+  public function postTypeSupports(string $type, string $feature): bool {
+    return post_type_supports($type, $feature);
+  }
+
+  public function getTaxonomies(array $args = [], string $output = 'names', string $operator = 'AND'): array {
+    return get_taxonomies($args, $output, $operator);
+  }
```
JanJakes, line 164: Some of the new methods have PHPDocs, some don't. Do they have PHPDoc in WP? For IDE experience and static analysis, it would be best to copy all the types.

JanJakes (commented):
LGTM.

> In its current implementation (https://github.com/mailpoet/mailpoet/commit/e31d914cad89c3ce696cfeacbf122e7e50ccfc8c), that means the automation would start and fail at the first step where a subscriber subject is required.

I wonder _how_ excatly the automation would fail. Do we have some handling for that? Or will it be a "random" failure based on other checks? I'm not sure what happens when a Send Email step is used, or what happens when a field filter on subsriber is used, etc.

I was also thinking whether the `SubjectTransformer` interface should somehow denote whether it always returns something or not, but I guess it may not be that important, once optional transformaiton is allowed.

Also, should we add a test for that?

mailpoet/assets/js/src/automation/integrations/wordpress/steps/made-a-comment/index.tsx
```diff
@@ -0,0 +1,33 @@
+import { __ } from '@wordpress/i18n';
+import { StepType } from '../../../../editor/store';
+import { LockedBadge } from '../../../../../common/premium-modal/locked-badge';
+import { PremiumModalForStepEdit } from '../../../../../common/premium-modal';
+import { Icon } from './icon';
+
+const keywords = ['comment'];
```
JanJakes: Translations 🙏

websupporter:
> I wonder how excatly the automation would fail. Do we have some handling for that?

For an action it will fail at this action with the error message `Subject data for subject with key 'mailpoet:subscriber' not found for automation run with ID '1'`. This is how it will appear in the error logs database.

But for filter (thanks for pointing to those) it will atm throw a fatal error during comment creation!
image

In https://github.com/mailpoet/mailpoet/pull/5197/commits/fdfdea1420ac32d516da3660186ecc27756443cf I catch the error and do not start the automation at all. As opposed to the action, this fails silently and does not add something to the logs.

mailpoet/lib/Automation/Engine/WordPress.php
websupporter: https://github.com/mailpoet/mailpoet/pull/5197/commits/90945687db2419bb5325cea2e81f279e9f4e2f03 adds the return types for two methods to specify it is `string[]` instead of just `array`. I think, we do not need to add doctypes where we have them defined in the method definition.

E.g. I don't think we need it in this case:
```
public function postTypeSupports(string $type, string $feature): bool
```

But I agree that `: array` should be specified.

mailpoet/assets/js/src/automation/integrations/wordpress/steps/made-a-comment/index.tsx
websupporter: https://github.com/mailpoet/mailpoet/pull/5197/commits/1cfbb9c74488de30403a8333f453164ff5ab76e3

mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
```diff
@@ -82,7 +82,13 @@ public function processTrigger(Trigger $trigger, array $subjects): void {
       $automationRun = new AutomationRun($automation->getId(), $automation->getVersionId(), $trigger->getKey(), $subjects);
       $stepRunArgs = new StepRunArgs($automation, $automationRun, $step, $subjectEntries, 1);
 
-      if (!$this->filterHandler->matchesFilters($stepRunArgs)) {
+      $match = false;
+      try {
+        $match = $this->filterHandler->matchesFilters($stepRunArgs);
+      } catch (Exceptions\Exception $e) {
+        ;
```
JanJakes, line 90: Let's maybe put a comment here? Sth like `// failed filter evaluation won't match`. Should we catch `Throwable` or do you think it's better to be more specific?

mailpoet/lib/Automation/Integrations/WordPress/Fields/PostFieldsFactory.php
```diff
@@ -0,0 +1,241 @@
...
+        }
+      ),
+      new Field(
+        'wordpress:post:type',
+        Field::TYPE_ENUM,
+        __('Post type', 'mailpoet'),
+        function (PostPayload $payload) {
+          $post = $payload->getPost();
+          //phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+          return $post ? $post->post_type : '';
```
JanJakes: In many cases, the empty value here is `''`. Is that just for PHPStan, or can it really be empty (in which case I think `null` could be more correct)?

mailpoet/lib/Automation/Integrations/WordPress/Fields/CommentFieldsFactory.php
```diff
@@ -0,0 +1,170 @@
...
+        }
+      ),
+      new Field(
+        'wordpress:comment:author-name',
+        Field::TYPE_STRING,
+        __('Comment author name', 'mailpoet'),
+        function (CommentPayload $payload) {
+          $comment = $payload->getComment();
+          //phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+          return $comment ? $comment->comment_author : '';
```
JanJakes: Same question here — are the empty strings just for PHPStan, or should they be `null`?

mailpoet/lib/Automation/Integrations/MailPoet/SubjectTransformers/CommentSubjectToSubscriberSubjectTransformer.php
```diff
@@ -0,0 +1,64 @@
...
+    if (!$comment instanceof \WP_Comment) {
+      return null;
+    }
+    //phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+    $email = $comment->comment_author_email;
+    if (!$this->wp->isEmail($email)) {
+      return null;
+    }
+
+    $subscriber = $this->subscribersRepository->findOneBy(['email' => $email]);
```
JanJakes, line 44: In WP, can it happen that I would be logged out and comment under your email without logging in? Just wondering if it can be misuses, or if in that case, WP would force me to log in.

JanJakes (commented):
Nice, thanks! 👏 Just a few questions...

mailpoet/lib/Automation/Integrations/MailPoet/SubjectTransformers/CommentSubjectToSubscriberSubjectTransformer.php
websupporter: I just tested, you do not need to be logged in to use e.g. the admin email address. The comment will be stored, depending on the settings even auto-approved, but the user_id will not be associated with the admin :thinking:

mailpoet/lib/Automation/Engine/Control/TriggerHandler.php
websupporter: https://github.com/mailpoet/mailpoet/pull/5197/commits/df75a23569e929c3f7c76b66e961c9a2ec15dc78 for the comment. Regarding the scope, I do not have a strong opinion. My thinking was to use a specific error class as the ones under `Exceptions\Exception` are basically `Exceptions::fieldLoadFailed` and `Exceptions::fieldNotFound` for which we defined: In this context that means `false`. Other throwables could be bugs which would hide in the dark with this `catch()`.

mailpoet/lib/Automation/Integrations/WordPress/Fields/PostFieldsFactory.php
websupporter: https://github.com/mailpoet/mailpoet/pull/5197/commits/a20cc87ee8322182010d9eb7f38cee78b7814352

mailpoet/lib/Automation/Integrations/WordPress/Fields/CommentFieldsFactory.php
websupporter: https://github.com/mailpoet/mailpoet/pull/5197/commits/a20cc87ee8322182010d9eb7f38cee78b7814352

websupporter:
Added tests for Post and Comment fields after catching some errors in the comment status in https://github.com/mailpoet/mailpoet/pull/5197/commits/45815f388a741afc06ab15e24b115a5a346bfe62

JanJakes (dismissed):

Aschepikov (approved):
GJ!

PR #5200, opened 2023-10-04
Catch activation in progress exception when it runs on DB version check [MAILPOET-5627]

costasovo:
## Description

I think the issue was introduced when started running the activator earlier (https://github.com/mailpoet/mailpoet/pull/5134). 
With the change, we moved the activation out of the try/catch block.

## Code review notes

_N/A_

## QA notes

Please check that upgrading the plugin works as expected. I recommend testing upgrading from a couple of months old version and test upgrading via WP admin and also upgrading via uploading the new version via FTP.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5627]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5627]: https://mailpoet.atlassian.net/browse/MAILPOET-5627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Initializer.php
+++ b/mailpoet/lib/Config/Initializer.php
@@ -411,9 +411,15 @@ class Initializer {
       $currentDbVersion = null;
     }
 
+    if (version_compare((string)$currentDbVersion, Env::$version) === 0) {
+      return;
+    }
+
     // if current db version and plugin version differ
-    if (version_compare((string)$currentDbVersion, Env::$version) !== 0) {
+    try {
       $this->activator->activate();
+    } catch (InvalidStateException $e) {
+      $this->handleRunningMigration($e);
     }
   }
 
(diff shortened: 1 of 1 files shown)


rodrigoprimo (approved):
Looks good to me!

Despite the fact, as you mentioned, that this error is hard to replicate, I believe it is still worth moving this PR to QA in case they want to check that upgrading MailPoet still works without a problem (I'm not sure if this is covered by our test suite or not).

costasovo:
> Despite the fact, as you mentioned, that this error is hard to replicate, I believe it is still worth moving this PR to QA in case they want to check that upgrading MailPoet still works without a problem (I'm not sure if this is covered by our test suite or not).

@rodrigoprimo Good point! 👍  I added instructions for QA.

Aschepikov (approved):
GJ!

PR #5204, opened 2023-10-06
Add email preview to the Gutenberg email editor [MAILPOET-5563]

lysyjan:
## Description

_N/A_

## Code review notes

- As a part of this PR is removing an old fix for setting a newsletter hash
- I added a store for the new email editor
- Some changes may be a bit confusing after rebasing on trunk because the trunk branch contains a new rule for naming files and folder

## QA notes

1. Active new email editor feature if it's disabled
2. Create a new email
3. Try if all preview options work properly

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5563]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5563]: https://mailpoet.atlassian.net/browse/MAILPOET-5563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/integrations/mailpoet/components/send-preview-email.tsx
+++ b/mailpoet/assets/js/src/email-editor/integrations/mailpoet/components/send-preview-email.tsx
@@ -1,0 +1,72 @@
+import { Button, Modal, TextControl } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import ReactStringReplace from 'react-string-replace';
+
+type SendPreviewEmailProps = {
+  isOpen: boolean;
+  closeCallback: () => void;
+};
+
+export function SendPreviewEmail({
+  isOpen,
+  closeCallback,
+}: SendPreviewEmailProps) {
+  let description = ReactStringReplace(
+    __(
+      'Send yourself a test email to test how your email would look like in different email apps. You could also enter your [link1]Mail Tester[/link1] email below to test your spam score. [link2]Learn more[/link2].',
+      'mailpoet',
+    ),
+    /\[link1\](.*?)\[\/link1\]/g,
+    (match, i) => (
+      <a
+        key={i}
+        href="https://www.mail-tester.com/"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        {match}
+      </a>
+    ),
+  );
+  description = ReactStringReplace(
+    description,
+    /\[link2\](.*?)\[\/link2\]/g,
+    (match, i) => (
+      <a
+        key={i}
+        href="https://kb.mailpoet.com/article/147-test-your-spam-score-with-mail-tester"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        {match}
+      </a>
+    ),
+  );
+
+  return (
+    <>
+      {isOpen ? (
+        <Modal
+          className="mailpoet-send-preview-email"
+          title={__('Send a test email', 'mailpoet')}
+          onRequestClose={closeCallback}
+        >
+          <p>{description}</p>
+          <TextControl
+            label={__('Send to', 'mailpoet')}
+            onChange={() => {}}
+            value=""
+          />
+          <div className="mailpoet-send-preview-modal-footer">
+            <Button variant="tertiary" onClick={closeCallback}>
+              {__('Close', 'mailpoet')}
+            </Button>
+            <Button variant="primary" onClick={() => {}}>
+              {__('Send test email', 'mailpoet')}
+            </Button>
+          </div>
+        </Modal>
+      ) : null}
+    </>
+  );
+}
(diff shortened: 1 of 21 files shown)


mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
```diff
@@ -52,6 +59,7 @@ public function getEmailDataSchema(): array {
       'id' => Builder::integer()->nullable(),
       'subject' => Builder::string(),
       'preheader' => Builder::string(),
+      'preview_url' => Builder::string()->nullable(),
```
costasovo: Is it nullable? It looks like `getViewInBrowserUrl` always returns a string even when the $newsletter is null.

mailpoet/assets/js/src/email-editor/integrations/mailpoet/components/send-preview-email.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <a
+        key={i}
+        href="https://kb.mailpoet.com/article/147-test-your-spam-score-with-mail-tester"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        {match}
+      </a>
+    ),
+  );
```
costasovo: Please, let's not use ReactStringReplace. We can use `createInterpolateElement` please see https://mailpoetdev.wordpress.com/2023/10/04/plugins-now-use-wp-to-enqueue-js-and-css/

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
```diff
@@ -112,6 +113,25 @@ public function enqueueAssets(): void {
...
+    // Enqueue inline script with MailPoet specific editor settings
+    $jsonAPIRoot = rtrim($this->wp->escUrlRaw(admin_url('admin-ajax.php')), '/');
+    $token = $this->wp->wpCreateNonce('mailpoet_token');
+    $apiVersion = API::CURRENT_VERSION;
+    $inlineScript = <<<EOL
+var mailpoet_json_api_root = '%s';
+var mailpoet_token = '%s';
+var mailpoet_api_version = '%s';
+EOL;
+    $this->wp->wpAddInlineScript(
```
costasovo: As these are only static variables it would be nicer to use https://developer.wordpress.org/reference/functions/wp_localize_script/

Example useage in MailPoet: PHP: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Form/AssetsController.php#L101-L105
Types: https://github.com/mailpoet/mailpoet/blob/d2d465165dfb91d50854754fffcc789b98f23f89/mailpoet/assets/js/src/global.d.ts#L155C1-L158C5

costasovo (changes requested):
@lysyjan Thank you! Nice job! I have just a few suggestions. Please see comments.

Also, one idea for the future. Because of the reusability in other projects, I think we could move the preview popup to the core code and add a slot for adding custom functionality and then render send preview emails and preview link in that slot from the MailPoet integration code. I wouldn't change that in this PR. Just want to know your opinion.

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
lysyjan: That's right. 👍 Thank you

mailpoet/assets/js/src/email-editor/integrations/mailpoet/components/send-preview-email.tsx
lysyjan: The power of habit. I'll fix it. Thank you
@costasovo Should I also fix it in [the details panel](https://github.com/mailpoet/mailpoet/blob/c8fc6bf49aa8add3f989530c792ae1cf5fac31ae/mailpoet/assets/js/src/email-editor/integrations/mailpoet/components/settings-panel/details-panel.tsx#L38-L68)? Or will it be a part of the custom editor bootstrap?

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
lysyjan: Nice suggestion. I will change it.

mailpoet/assets/js/src/email-editor/integrations/mailpoet/components/send-preview-email.tsx
costasovo: We can do it in the custom editor bootstrap. Good idea! 👍

costasovo (approved):
LGTM! Thank you!

PR #5206, opened 2023-10-06
Remove double log entries before adding unique index [MAILPOET-5630]

websupporter:
## Description
Fixes [MAILPOET-5630]

## Code review notes

_N/A_

## QA notes

Install a MailPoet version not younger than version [4.26.0](https://github.com/mailpoet/mailpoet/releases/tag/4.26.0). Produce some log entries (run one or two automations). Open phpmyadmin or something similar and duplicate some of the produced rows.

Update to the MP version with the fix. This PR works when:
1. The duplicate entries disappeared
2. There is no error in the ` wp_mailpoet_migrations` table for the `Migration_20230831_143755_Db` entry, and - actually - for no entry

## Linked PRs

_N/A_

## Linked tickets
[MAILPOET-5630]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5630]: https://mailpoet.atlassian.net/browse/MAILPOET-5630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5630]: https://mailpoet.atlassian.net/browse/MAILPOET-5630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Migrations/Db/Migration_20230831_143755_Db.php
+++ b/mailpoet/lib/Migrations/Db/Migration_20230831_143755_Db.php
@@ -52,6 +52,9 @@ class Migration_20230831_143755_Db extends DbMigration {
 
     // add unique index, remove no longer needed index
     if (!$this->indexExists($logsTable, 'automation_run_id_step_id')) {
+      $this->connection->executeStatement(
+        "DELETE t1 FROM $logsTable as t1, $logsTable as t2 WHERE t1.id < t2.id AND t1.automation_run_id = t2.automation_run_id AND t1.step_id=t2.step_id"
+      );
       $this->connection->executeStatement("ALTER TABLE $logsTable ADD UNIQUE automation_run_id_step_id (automation_run_id, step_id)");
     }
     if ($this->indexExists($logsTable, 'automation_run_id')) {


mailpoet/lib/Migrations/Db/Migration_20230831_143755_Db.php
```diff
@@ -52,6 +52,9 @@ public function run(): void {
 
     // add unique index, remove no longer needed index
     if (!$this->indexExists($logsTable, 'automation_run_id_step_id')) {
+      $this->connection->executeStatement(
+        "DELETE t1 FROM $logsTable as t1, $logsTable as t2 WHERE t1.id < t2.id AND CONCAT(t1.automation_run_id,t1.step_id) = CONCAT(t2.automation_run_id,t2.step_id); "
+      );
```
costasovo: @websupporter @JanJakes pointed out to me that concat can be problematic. E.g., run id1 and step id1 are 1, 11, and run id2 and step id2 are 11, 1.  
It can be replaced by multiple conditions
```sql
DELETE t1 FROM $logsTable as t1, $logsTable as t2 WHERE t1.id < t2.id AND t1.automation_run_id = t2.automation_run_id AND t1.step_id = t2.step_id
```

costasovo (changes requested):
Please replace the concats with multiple conditions. See my comment.

mailpoet/lib/Migrations/Db/Migration_20230831_143755_Db.php
websupporter: :facepalm: when you forget to do `git add` again after @JanJakes points out to me that concat can be problematic

costasovo (approved):
Thank you for addressing it so quickly!

Aschepikov (approved):
GJ!

PR #5208, opened 2023-10-06
[MAILPOET-5625] Store lookup data for filter segments

johnolek:
## Description

This PR updates how we store a snapshot of filter segments to include "lookup data" for any entities that we currently only have IDs for. Currently, I believe this includes:

- entered automation (automations)
- exited automation (automations)
- was sent (emails)
- opened (emails)
- machine-opened (emails)
- clicked (emails, links)
- MailPoet custom field (custom field name)
- subscribed to list (list name)
- subscribed via form (form name)
- WordPress user role (role names)
- Purchased in category (category names)
- Purchased product (product names)
- Used coupon code (coupon code)
- Used payment method (payment method name)
- Used shipping method (shipping method name)
- tag (tag name) - handled in the premium PR

This ensures we will have the correct data available any time in the future, even if those entities change or are deleted.

## Code review notes

The main idea with this code is that we want to be able to display tooltips for filter segments on listing pages that show the exact configuration they had at send time. My first attempt involved storing way more data, including strings for operators and the names of the filters, but I realized it makes more sense to store just the lookup information that's missing and leave the tooltip generation for the frontend code. Please let me know if you can think of anything this PR leaves out that would be necessary for constructing those tooltips (see MAILPOET-5600 for more details about the tooltips).

If we ever fail to retrieve the value for a particular ID, I'm leaving it out of the return value with the assumption that the frontend code will need to assume the data might not exist.

## QA notes

This is only a change to the backend, meaning that to test it you will need to verify that lookup information is stored in the sending queue's `meta` column under the `filterSegment` key. 

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/799

## Linked tickets
[MAILPOET-5625](https://mailpoet.atlassian.net/browse/MAILPOET-5625)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5625]: https://mailpoet.atlassian.net/browse/MAILPOET-5625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Entities/DynamicSegmentFilterData.php
+++ b/mailpoet/lib/Entities/DynamicSegmentFilterData.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Entities;
 
+use MailPoet\InvalidStateException;
 use MailPoet\Segments\DynamicSegments\Filters\UserRole;
 use MailPoet\Segments\DynamicSegments\Filters\WooCommerceProduct;
 use MailPoetVendor\Doctrine\ORM\Mapping as ORM;
@@ -85,6 +86,35 @@ class DynamicSegmentFilterData {
    */
   public function getParam(string $name) {
     return $this->filterData[$name] ?? null;
+  }
+
+  public function getStringParam(string $name): string {
+    $value = $this->filterData[$name] ?? null;
+    if (!is_string($value)) {
+      throw new InvalidStateException("No string value found in filter data for param $name.");
+    }
+    return $value;
+  }
+
+  public function getIntParam(string $name): int {
+    $value = $this->filterData[$name] ?? null;
+    if (is_int($value)) {
+      return $value;
+    }
+
+    if (is_string($value)) {
+      return (int)($value);
+    }
+
+    throw new InvalidStateException("No compatible integer value found in filter data for param $name.");
+  }
+
+  public function getArrayParam(string $name): array {
+    $value = $this->getParam($name);
+    if (!is_array($value)) {
+      throw new InvalidStateException("No array value found in filter data for param $name.");
+    }
+    return $value;
   }
 
   public function getFilterType(): ?string {
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/SubscriberTag.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/SubscriberTag.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Segments\DynamicSegments\Filters;
 
+use MailPoet\Entities\DynamicSegmentFilterData;
 use MailPoet\Entities\DynamicSegmentFilterEntity;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Doctrine\DBAL\Query\QueryBuilder;
@@ -21,5 +22,9 @@ class SubscriberTag implements Filter {
   public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $filter): QueryBuilder {
     $this->wp->applyFilters('mailpoet_dynamic_segments_filter_subscriber_tag_apply', $queryBuilder, $filter);
     return $queryBuilder;
+  }
+
+  public function getLookupData(DynamicSegmentFilterData $filterData): array {
+    return [];
   }
 }
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/SubscriberTag.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/SubscriberTag.php
@@ -7,6 +7,9 @@ use MailPoet\Entities\DynamicSegmentFilterEntity;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Doctrine\DBAL\Query\QueryBuilder;
 
+/**
+ * The filters in this class are primarily intended for the premium plugin
+ */
 class SubscriberTag implements Filter {
   const TYPE = 'subscriberTag';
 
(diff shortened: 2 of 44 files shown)


mailpoet/lib/Segments/DynamicSegments/Filters/SubscriberTag.php
```diff
@@ -25,6 +25,11 @@ public function apply(QueryBuilder $queryBuilder, DynamicSegmentFilterEntity $fi
   }
 
   public function getLookupData(DynamicSegmentFilterData $filterData): array {
-    return [];
+    $default = ['tags' => []];
+    $filteredLookupData = $this->wp->applyFilters('mailpoet_dynamic_segments_filter_subscriber_tag_getLookupData', $default, $filterData);
```
rodrigoprimo, line 32: Maybe it is worth adding a comment here mentioning that this filter is used in mailpoet-premium? It took me a while to figure that out, but that could have been because my copy of premium was in the `trunk` branch and not `MAILPOET-5625/store-lookup-data-for-filters`.

mailpoet/lib/Entities/DynamicSegmentFilterData.php
```diff
@@ -87,6 +88,35 @@ public function getParam(string $name) {
     return $this->filterData[$name] ?? null;
   }
 
+  public function getStringParam(string $name): string {
```
rodrigoprimo, line 91: I was not able to find where this method is being used. Do we still need it?

rodrigoprimo (commented):
Great job! I left two comments.

Besides that, I'm noticing an unexpected behavior that I'm not sure if it is something that I'm doing wrong or if it is a problem. I won't have time to debug it today, and I didn't want to delay publishing this review. So I will describe it here and ask you to test it.

Populating the `lookupData` key seems to be working fine when I tested with the user role segment (output below from the wp_mailpoet_sending_queues table):

```
|  6 |             6 | {"campaignId":"5dc75e285fe30a67","filterSegment":{"id":32,"name":"New segment created via newslettter","updatedAt":{"date":"2023-10-09 13:54:31.000000","timezone_type":3,"timezone":"UTC"},"filters":[{"filterType":"userRole","action":"wordpressRole","data":{"wordpressRole":["administrator"],"operator":"any","connect":"and"},"lookupData":{"roles":{"administrator":"Administrator"}}}]}}
```

 But I got an empty `lookupData` keys for all filters when I created a segment using the `Engaged Subscribers (3 months)` segment template and use it as the filter for sending a newsletter.

```
|  8 |             8 | {"campaignId":"2ec0d2d2d1e8ba46","filterSegment":{"id":34,"name":"Engaged Subscribers (3 months)","updatedAt":{"date":"2023-10-09 18:26:01.000000","timezone_type":3,"timezone":"UTC"},"filters":[{"filterType":"userRole","action":"lastClickDate","data":{"value":"90","operator":"inTheLast","connect":"or"},"lookupData":[]},{"filterType":"userRole","action":"lastOpenDate","data":{"value":"90","operator":"inTheLast","connect":"or"},"lookupData":[]},{"filterType":"userRole","action":"lastPurchaseDate","data":{"value":"90","operator":"inTheLast","connect":"or"},"lookupData":[]},{"filterType":"userRole","action":"lastSendingDate","data":{"value":"90","operator":"inTheLast","connect":"or"},"lookupData":[]}]}} |
```

Are you able to reproduce the same problem?

johnolek:
@rodrigoprimo Thank you for the review and testing!

> But I got an empty lookupData keys for all filters when I created a segment using the Engaged Subscribers (3 months) segment template and use it as the filter for sending a newsletter.

This is actually expected, and I realize now that I forgot to update the ticket description before assigning this to you. Sorry for the confusion!

I originally thought we would need to store the filter name, operator and value for every filter segment, and I started working on an implementation that did just that. I realized halfway through that storing the data that way would pose an i18n problem (i.e. we wouldn't be able to show the tooltip details in different languages, and it wasn't as trivial as I thought to generate the full strings for some filters). That's when I decided to change approaches to one where I'm only storing lookup for things where we currently only have the IDs.

Since the engaged subscribers segment doesn't depend on any entities for its config, we don't need to store any lookup data.

If you think it would be cleaner or clearer to leave out empty arrays in cases like these I could definitely make that change. It could be nice being able to assume that `lookupData` will exist and be an array, but it shouldn't make much difference one way or another.

edit: I updated the Jira ticket to indicate that we only want to store names of entities.

mailpoet/lib/Entities/DynamicSegmentFilterData.php
johnolek: This is left over from my initial approach, and you're right it's not currently used anywhere.

I'm inclined to keep it because I think it would be helpful in getting rid of things like this in our PHPStan config: https://github.com/mailpoet/mailpoet/blob/d55742aebdac050e1af943798ab9d63f8ee3189f/mailpoet/tasks/phpstan/phpstan-7-baseline.neon#L539-L541

Maybe it would be better to wait and handle that all in a separate ticket though. I'm happy to remove this from this PR if you prefer!

mailpoet/lib/Segments/DynamicSegments/Filters/SubscriberTag.php
johnolek: I agree this could be clearer. I added a brief note in e683e85d580aaed9012cf5e834f974dcb6af88f0. Let me know if you think that's sufficient!

mailpoet/lib/Entities/DynamicSegmentFilterData.php
rodrigoprimo: Thanks for the additional details. We can keep it.

rodrigoprimo:
> Since the engaged subscribers segment doesn't depend on any entities for its config, we don't need to store any lookup data.

I completely missed this when testing.

Moving this PR to QA.

rodrigoprimo (approved):

Aschepikov (approved):
GJ!

PR #5212, opened 2023-10-10
Use toString instead of html [MAILPOET-5632]

johnolek:
## Description

The pQuery `html` method decodes all HTML entities. This isn't what we want because the editor automatically encodes entities that should be encoded. By using `toString` instead we should see more consistency between the editor and previews/rendered emails.

## Code review notes

_N/A_

## QA notes

This change could affect many different elements in a newsletter. Please be sure to test with a wide variety of content, including nested HTML content, to ensure this doesn't have any unintended side effects.

## Linked PRs

_N/A_

## Linked tickets
https://mailpoet.atlassian.net/browse/MAILPOET-5632

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Newsletter/Editor/PostTransformerContentsExtractor.php
+++ b/mailpoet/lib/Newsletter/Editor/PostTransformerContentsExtractor.php
@@ -151,6 +151,7 @@ class PostTransformerContentsExtractor {
 
     $title = '<' . $tag . ' data-post-id="' . $post->ID . '" style="text-align: ' . $alignment . ';">' . $title . '</' . $tag . '>';
 
+    // The allowed HTML is based on all the possible ways we might construct a $title above
     $commonAttributes = [
       'data-post-id' => [],
       'style' => [],
--- a/mailpoet/lib/Newsletter/Editor/PostTransformerContentsExtractor.php
+++ b/mailpoet/lib/Newsletter/Editor/PostTransformerContentsExtractor.php
@@ -150,9 +150,25 @@ class PostTransformerContentsExtractor {
     $alignment = (in_array($this->args['titleAlignment'], ['left', 'right', 'center'])) ? $this->args['titleAlignment'] : 'left';
 
     $title = '<' . $tag . ' data-post-id="' . $post->ID . '" style="text-align: ' . $alignment . ';">' . $title . '</' . $tag . '>';
+
+    $commonAttributes = [
+      'data-post-id' => [],
+      'style' => [],
+    ];
+
+    $allowedTitleHtml = [
+      'a' => [
+        'href' => [],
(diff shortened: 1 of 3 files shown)


costasovo (approved):
LGTM! 👍

Aschepikov:
Hi @johnolek I tried to include the nested html and it looks like there is an issue, can you take a look pls
image

johnolek:
@Aschepikov When you say nested HTML I'm not sure what exactly you mean. Does that just mean manually entering HTML into the editor?

If so I believe what you're seeing is what we're expecting, and I believe it also mirrors what we see in WordPress itself. For example, if I create a post that contains `<strong>strong</strong>` in a standard text block I see it like this on the frontend:

image

Part of the idea with this change is we're allowing users to show HTML markup in their messages.

@costasovo please correct me if I'm wrong about any of this.

costasovo:
> Part of the idea with this change is we're allowing users to show HTML markup in their messages.

@johnolek, @Aschepikov  Yes. I think it is correct, and the behavior of the production editor is IMO buggy. We want WYSIWYG behavior so if you type HTML code directly into the text block, we should render it as a code (not interpret it). If you switch the text block to HTML editor mode and enter HTML there, it should render as interpreted HTML.

Here is what it looks like directly in the editor (It should look exactly the same in the preview). At the top the strong and list tags are entered directly as content and below via HTML edit mode. 
Screenshot 2023-10-17 at 13 48 58

Editing text in HTML mode:
Screenshot 2023-10-17 at 13 49 09

I would rather confirm this behavior with @NeosinneR. @NeosinneR Is the above behavior correct?

@Aschepikov Did you use HTML mode for entering the HTML code or entered it directly?

@johnolek I think the behavior introduced in this PR is correct. But seeing @Aschepikov's comment, I realized that at the same time, the PR kind of removes a hidden "feature" (using HTML directly in the editor and expecting it will be interpreted in email).

So, users who were using (or rather misusing) this bug may experience a BC break. If we wanted to prevent it, the only thing that comes to my mind is a pretty complex migration where we would remove HTML encoding in all text blocks in all saved emails. I have a concern that such a migration has the potential to cause even more issues.

Aschepikov:
@costasovo I entered it directly because usually, it was working too. As we have an HTML editor there is no need to maintain this functionality. There is a chance that we have clients that use HTML directly because we allow it, but we have a special instrument for it. If @NeosinneR agrees we can proceed with the current changes

NeosinneR:
Sounds good to me. The new behavior is what I'd also expect when I enter HTML code in the visual editor - it should continue to render as text and not interpret it. So 👍 from me.

Aschepikov (approved):
GJ!

PR #5213, opened 2023-10-11
Fix: make sure filters are ordered alphabetically [MAILPOET-5229]

rodrigoprimo:
## Description

This PR makes sure filters are ordered alphabetically per group in the page to create/edit segments. It uses .sort() to ensure the ordering happens after the label is translated if the site is using a language other than English.

It also changes the order of the filters in the .ts files where they are defined to match the order in which they are displayed when the site is in English to make it easier for developers to add new filters and to find filters when coding.

## Code review notes

Note that ordering the segments alphabetically uncovered what I believe is a bug in `AcceptanceTester::selectOptionInReactSelect()`. I was not able to find a quick solution for this bug and didn't want to spend more time on this ticket. So, in f1e359ad99d7326dbe0753e40b3b3a696b647625, I implemented a workaround. Please let me know if you can find a better approach. If not, I plan to create a ticket for us to fix `AcceptanceTester::selectOptionInReactSelect()`.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5229]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5229]: https://mailpoet.atlassian.net/browse/MAILPOET-5229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/automation-options.ts
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/automation-options.ts
@@ -1,4 +1,5 @@
 import { MailPoet } from 'mailpoet';
+import { sortFilters } from 'segments/dynamic/dynamic-segments-filters/sort-filters';
 import { SegmentTypes } from '../types';
 
 export enum AutomationsActionTypes {
@@ -17,4 +18,4 @@ export const AutomationsOptions = [
     label: MailPoet.I18n.t('automationsExitedAutomation'),
     group: SegmentTypes.Automations,
   },
-];
+].sort(sortFilters);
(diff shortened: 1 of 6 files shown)


johnolek (approved):
Looks good to me! 👍 

I think it's best to create a separate ticket for the react select issue.

Thanks for reordering the options in the code, though I suppose now it won't matter too much what the order is. Something that occurred to me is it would be even easier to get a sense for what's what if we updated these files to use `__` instead. I'll leave it up to you if you think that's a good idea or if you would prefer to just move this to QA.

rodrigoprimo:
I think it is a good idea, but I don't think I will have time to work on it this week, so I will move this PR to QA.

Aschepikov (approved):
GJ!

PR #5214, opened 2023-10-11
Use mixpanel PHP instead of JS for user properties [MAILPOET-5161]

johnolek:
## Description

This PR moves some of our (opt-in) tracking to use the mixpanel PHP library instead of JavaScript. This is because the JS only runs when a user visits a MailPoet page, and there's no guarantee users will do that on a weekly basis.

## Code review notes

Unfortunately the PHP library does not play nicely with our prefixer. For now I'm including it unprefixed, but if you foresee that being a problem I'm happy to add a fix.

At one point I ran into a problem where a test was failing because it was running [this code](https://github.com/mailpoet/mailpoet/blob/6e705fb316e8b2099036707ba37444928addfa70/mailpoet/lib/Newsletter/NewslettersRepository.php#L176-L176), which was throwing an error because we delete all newsletter option field rows between tests. I decided to change our integration test cleanup code to leave the option field records alone because most of our code reasonably assumes these records will exist in a MailPoet installation and I thought this would get our test environment closer to real life.

## QA notes

To test, you'll need to have "Share anonymous data" enabled in MailPoet settings. You may need to deactivate and reactivate the plugin in order to schedule the very first scheduled task. Please reach out to me if you don't have access to mixpanel and need me to verify when certain data was sent.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5161](https://mailpoet.atlassian.net/browse/MAILPOET-5161)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5161]: https://mailpoet.atlassian.net/browse/MAILPOET-5161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/Mixpanel.php
+++ b/mailpoet/lib/Cron/Workers/Mixpanel.php
@@ -57,4 +57,8 @@ class Mixpanel extends SimpleWorker {
 
     return true;
   }
+
+  public function getNextRunDate() {
+    return $this->analytics->getNextSendDate()->addMinutes(rand(0, 59));
+  }
 }
--- a/mailpoet/lib/Cron/Workers/Mixpanel.php
+++ b/mailpoet/lib/Cron/Workers/Mixpanel.php
@@ -1,0 +1,60 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Cron\Workers;
+
+use MailPoet\Analytics\Analytics;
+use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\WP\Functions;
+use Mixpanel as MixpanelLibrary;
+
+class Mixpanel extends SimpleWorker {
+
+  const PRODUCTION_PROJECT_ID = '8cce373b255e5a76fb22d57b85db0c92';
+
+  /** @var Analytics */
+  private $analytics;
+
+  const TASK_TYPE = 'mixpanel';
+
+  /** @var MixpanelLibrary */
+  private $mixpanel;
+
+  public function __construct(
+    Analytics $analytics,
+    Functions $wp
+  ) {
+    parent::__construct($wp);
+    $this->analytics = $analytics;
+    $this->mixpanel = MixpanelLibrary::getInstance(self::PRODUCTION_PROJECT_ID);
+    $this->mixpanel->register('Platform', 'Plugin');
+  }
+
+  public function processTaskStrategy(ScheduledTaskEntity $task, $timer) {
+    return $this->maybeReportAnalyticsToMixpanel();
+  }
+
+  public function maybeReportAnalyticsToMixpanel(): bool {
+    if (!$this->analytics->shouldSend()) {
+      return true;
+    }
+    return $this->reportAnalyticsToMixpanel();
+  }
+
+  public function reportAnalyticsToMixpanel(): bool {
+    $publicId = $this->analytics->getPublicId();
+
+    if (strlen($publicId) < 1) {
+      return true;
+    }
+
+    $data = $this->analytics->getAnalyticsData();
+
+    $this->mixpanel->identify($publicId);
+    $this->mixpanel->people->set($publicId, $data);
+    $this->mixpanel->track('User Properties', $data);
+
+    $this->analytics->recordDataSent();
+
+    return true;
+  }
+}
--- a/mailpoet/views/analytics.html
+++ b/mailpoet/views/analytics.html
@@ -5,12 +5,7 @@
 
 window.mixpanelTrackingId = "8cce373b255e5a76fb22d57b85db0c92";
 
-mixpanel.init(window.mixpanelTrackingId, {
-  loaded: function(mixpanel) {
-    // used in lib/Analytics/Analytics.php
-    document.cookie = "mixpanel_distinct_id=" + mixpanel.get_distinct_id();
-  }
-});
+mixpanel.init(window.mixpanelTrackingId);
 
 mixpanel.register({'Platform': 'Plugin'});
 
--- a/mailpoet/views/analytics.html
+++ b/mailpoet/views/analytics.html
@@ -5,24 +5,21 @@
 
 window.mixpanelTrackingId = "8cce373b255e5a76fb22d57b85db0c92";
 
-if (mailpoet_analytics_enabled) {
-
-  mixpanel.init(window.mixpanelTrackingId, {
-    loaded: function(mixpanel) {
-      // used in lib/Analytics/Analytics.php
-      document.cookie = "mixpanel_distinct_id=" + mixpanel.get_distinct_id();
-    }
-  });
-
-  mixpanel.register({'Platform': 'Plugin'});
-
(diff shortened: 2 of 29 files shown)


johnolek:
@websupporter I didn't remember you were porter this week. I can assign this to someone else if you haven't started looking at it already. I also need to fix the pipeline failures.

mailpoet/lib/Cron/Workers/Mixpanel.php
```diff
@@ -0,0 +1,64 @@
...
+
+  public function maybeReportAnalyticsToMixpanel(): bool {
+    if (!$this->analytics->shouldSend()) {
+      return true;
+    }
+    return $this->reportAnalyticsToMixpanel();
+  }
+
+  public function reportAnalyticsToMixpanel(): bool {
+    $publicId = $this->analytics->getPublicId();
```
websupporter, line 44: If I see this correctly we rely on cookies for the public id. Maybe this comes later, but where do we `set` the public ID and have access to cookies (as this is a cron job here)?

websupporter: https://github.com/mailpoet/mailpoet/blob/8fbac3e2c25328a5b967e2f9c074e24caf0f0381/mailpoet/lib/Analytics/Analytics.php#L60-L71

websupporter (commented):
Overall lgtm. I am a bit worried about getting and setting the public ID.

Also the comment
> You may need to deactivate and reactivate the plugin in order to schedule the very first scheduled task.

But older plugins at some point schedule the task right? I guess this is just because we do not jump versions or something why this is necessary at this moment?

mailpoet/lib/Cron/Workers/Mixpanel.php
johnolek: Thanks for flagging this, this is a valid concern!

I think it's unlikely we would end up in a situation where would need to rely on the cookie in the cron task. If a user is opted in to analytics and 3rd party libraries, we load `analytics.html` on every MailPoet page (https://github.com/mailpoet/mailpoet/blob/6e705fb316e8b2099036707ba37444928addfa70/mailpoet/views/layout.html#L48-L48) so we would end up with a value in the settings table.

The problem would be if the user isn't opted in to 3rd party libraries, in which case this never loads and we never get the public id.

I believe the only other time we explicitly set public id is if we get one back from the MSS:
https://github.com/mailpoet/mailpoet/blob/6e705fb316e8b2099036707ba37444928addfa70/mailpoet/lib/API/JSON/v1/Services.php#L131-L131
https://github.com/mailpoet/mailpoet/blob/6e705fb316e8b2099036707ba37444928addfa70/mailpoet/lib/API/JSON/v1/Services.php#L197-L197

We could end up with empty public IDs now, but I don't think we would ever try to use them if since we wouldn't load `analytics.html`.

It doesn't seem like the actual ID matters at all if we're using an md5, so my inclination is to simply set a random value instead of relying on the cookie. How does that sound to you? I'm going to check with the team just to make sure I'm not missing anything.

johnolek:
Thanks for the review, @websupporter!

> But older plugins at some point schedule the task right? I guess this is just because we do not jump versions or something why this is necessary at this moment?

Deactivating and reactivating should mainly be necessary if you're just swapping out the plugin files. When people upgrade they will trigger the populator and the initial task should get scheduled.

Thanks again for calling out the public_id thing. I'll plan on defaulting to a random string instead of relying on the cookie unless someone can tell me why that's a bad idea, or there's a way to get a distinct ID from the PHP library.

I also think we might be able to move the aliasing to the PHP library (currently here: https://github.com/mailpoet/mailpoet/blob/6e705fb316e8b2099036707ba37444928addfa70/mailpoet/views/analytics.html#L21-L21), but I'm not sure if the aliasing is even working right now because it should take 2 parameters (see https://github.com/mixpanel/mixpanel-js/blob/8ac526e5cb8563d11e2206046ab986c6491ac6d7/src/mixpanel-core.js#L1376-L1412) so I might create a separate ticket for that.

mailpoet/lib/Cron/Workers/Mixpanel.php
johnolek: 943e1a64ef9c01cf251978bd21c51e9a341fd516 modifies this to use a random string instead of relying on the cookie. Under the hood it looks like this is basically what mixpanel does anyway: https://github.com/mixpanel/mixpanel-js/blob/8ac526e5cb8563d11e2206046ab986c6491ac6d7/src/mixpanel-core.js#L323C1-L333C6

mailpoet/views/analytics.html
```diff
@@ -5,24 +5,21 @@
...
-      // used in lib/Analytics/Analytics.php
-      document.cookie = "mixpanel_distinct_id=" + mixpanel.get_distinct_id();
-    }
-  });
+mixpanel.init(window.mixpanelTrackingId, {
+  loaded: function(mixpanel) {
+    // used in lib/Analytics/Analytics.php
+    document.cookie = "mixpanel_distinct_id=" + mixpanel.get_distinct_id();
+  }
+});
```
websupporter: Why do we set the cookie still and reference Analytics.php. I think this is not necessary anymore? Or we could change the behavior in a way that we generate a random id only when the cookie does not exist?

mailpoet/lib/Cron/Workers/Mixpanel.php
websupporter: Yeah, in general I agree, it seems we can just set an ID ourselves.

mailpoet/views/analytics.html
websupporter: Do we use mixpanel in JavaScript anywhere? I assume otherwise we could get rid of the analytics.html? I am worried we could get out of sync between JS events and PHP events, but probably not, since we set the public Id in the twig function, right? So I assume we should just remove this setting of the cookie to avoid confusion :thinking:

johnolek: @websupporter I believe you're right that we're no longer using the cookie. I wasn't 100% sure that was the case, but since I can't find `mixpanel_distinct_id` anywhere else in the codebase this feels safe to remove, which I've done in 82c49bb81f9b24bbaefbe16fb537d3dbaa804086.

As for mixpanel in JavaScript, we do use it extensively to track specific actions. It looks like we wrap it in a way where we will initialize mixpanel if it hasn't already been initialized (https://github.com/mailpoet/mailpoet/blob/e1454b3d03d153e803ae452abace2c83a54b61e9/mailpoet/assets/js/src/analytics-event.js#L30-L30), but that still depends on `window.mixpanel` and `window.mixpanelTrackingId` existing, which I assume only come from this file. My inclination is to not remove too much from here right now, but I suspect we may be able to simplify this in the future ticket.

johnolek:
@websupporter FYI I had to force push because of conflicts with `composer.lock`.

mailpoet/views/analytics.html
websupporter: Sounds good to me. Thank you @johnolek

Aschepikov (approved):
GJ!

PR #5215, opened 2023-10-12
Skip tracking step for dotcom [MAILPOET-5549]

johnolek:
## Description

This PR changes the welcome wizard to skip the `Confirm privacy and data settings` step if a user is on wordpress.com. If the user doesn't have any values already saved for analytics and 3rd party libraries, they should be opted in automatically. If values already exist in the database for those settings, the wizard should not change them. 

## Code review notes

My initial implementation didn't show the correct number of steps at the top of the wizard. It would still show the number of steps that would exist with the tracking step, but it would simply skip over that step. I fixed this in 4fbd262cf2b588ece71e0c63fb875eccf2364b3a. I thought it risky how the step counter and the step mapper needed to have the same logic, so I did some refactoring to consolidate the logic to one place. tldr, please take a look to make sure this refactor didn't introduce any unintended side effects.

## QA notes

Please make sure to test the welcome wizard in all possible configurations (wpcom and not, with/without woocommerce, with/without a valid MSS key).

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5549](https://mailpoet.atlassian.net/browse/MAILPOET-5549)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5549]: https://mailpoet.atlassian.net/browse/MAILPOET-5549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/global.d.ts
+++ b/mailpoet/assets/js/src/global.d.ts
@@ -262,4 +262,5 @@ interface Window {
     type: 'default' | 'wp_users' | 'woocommerce_users' | 'dynamic';
   }>;
   mailpoet_admin_plugins_url: string;
+  mailpoet_is_dotcom: boolean;
 }
--- a/mailpoet/assets/js/src/global.d.ts
+++ b/mailpoet/assets/js/src/global.d.ts
@@ -178,6 +178,7 @@ interface Window {
   wizard_tracking_illustration_url: string;
   wizard_MSS_pitch_illustration_url: string;
   wizard_woocommerce_illustration_url: string;
+  wizard_has_tracking_settings: boolean;
   mailpoet_account_url: string;
   mailpoet_show_customers_import: boolean;
   mailpoet_installed_days_ago: number;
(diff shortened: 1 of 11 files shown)


lysyjan (dismissed):
LGTM 👍

Aschepikov:
Hi @johnolek I see in the pr description that the **Subsequent visits to the welcome wizard should always show the second step.** and in the ticket that **If the wizard is loaded next and the values have already been set, don’t change them.** and it should be still skipped. How should it work?

johnolek:
@Aschepikov Good question! My interpretation was that we should only skip the step if there aren't already values saved in the database for both analytics and 3rd party libraries, and if there are values saved that we should show the second step and allow users to change them. 

But you're right, the ticket as written looks like we should continue to skip the step and simply not make changes if we already have values stored in the database. @NeosinneR could you please confirm the desired behavior? I'm assuming I should change the behavior to always skip this step, in which case I'll need to revise this PR.

NeosinneR:
@Aschepikov good catch. 

> we should continue to skip the step and simply not make changes if we already have values stored in the database

@johnolek Yes, this is the intended behavior. On Dotcom, this step should never be shown, but if the user has already opted out in settings, visiting the onboarding should not opt in in the background.

johnolek:
@lysyjan I ended up doing some refactoring (see updated PR comment). Could you please take a second look at this when you have a chance?

lysyjan (approved):
@johnolek Looks good to me. I like the refactoring that you did.

Aschepikov (approved):
GJ!

PR #5216, opened 2023-10-13
Conditions [MAILPOET-5586] [MAILPOET-5587] [MAILPOET-5667]

JanJakes:
## Description

Implements automation conditions backend and frontend.

## Code review notes

_N/A_

## QA notes

Please, test conditions, and also that existing automations and templates work.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/802

## Linked tickets

[MAILPOET-5586]
[MAILPOET-5587]
[MAILPOET-5667]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5586]: https://mailpoet.atlassian.net/browse/MAILPOET-5586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-5587]: https://mailpoet.atlassian.net/browse/MAILPOET-5587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-5667]: https://mailpoet.atlassian.net/browse/MAILPOET-5667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/editor/components/filters/chip.tsx
```diff
@@ -39,15 +42,15 @@ export function StepFilters({ step }: Props): JSX.Element | null {
           onClick={onToggle}
           ariaExpanded={isOpen}
         >
-          {__(`Filters: ${filterCount}`, 'mailpoet')}
+          {__(`${strings.label}: ${filterCount}`, 'mailpoet')}
```
websupporter: Shouldn't it be `sprintf(__('Filters: %d', 'mailpoet'), filterCount)`?

mailpoet/assets/js/src/automation/integrations/core/steps/if-else/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { storeName, StepType } from '../../../../editor/store';
+import { BranchBadge } from './branch-badge';
+import { Footer } from './footer';
+
+const keywords = [
+  __('wait', 'mailpoet'),
+  __('pause', 'mailpoet'),
+  __('delay', 'mailpoet'),
+  __('time', 'mailpoet'),
+];
```
websupporter: probably copy-pasted from the delay block.

websupporter:
Thanks for working on this @JanJakes! Huge step for automations :clap: 

Some remarks:

websupporter:
I can move horizontally only when i scroll to the bottom.
image

websupporter:
Deleting a step in one branch leads to strange behavior, see these two screenshots, the second is the end result:
image
image

mailpoet/lib/Automation/Engine/Data/NextStep.php
```diff
@@ -3,16 +3,16 @@
 namespace MailPoet\Automation\Engine\Data;
 
 class NextStep {
-  /** @var string */
+  /** @var string|null */
   protected $id;
 
   public function __construct(
-    string $id
```
websupporter, line 10: I just saw this, when I switched to another branch. In support, we sometimes suggest to revert back to an older version. This change means, we can not revert back to an older version. It seems to break the whole automation system. Maybe we could use empty string instead?
image

JanJakes: @websupporter 🤔 An automation with conditions won't work after a downgrade... so I think the bug is rather about where and why do we load the full automation body, outside the editor.

JanJakes:
@websupporter I can move horizontally only when i scroll to the bottom.

I suppose that's Linux, and which browser is that? I can't replicate that.
And is that about the actual scrolling on a trackpad or the visibility of the toolbar?

mailpoet/lib/Automation/Engine/Data/NextStep.php
JanJakes: @websupporter Ohh, but we can't undo the "bug" in older versions... 😬 I'm hesitant to change the API only to cover this edge case. I mean, it really is just the edge case of 1) using IF/ELSE already, and 2) needing to downgrade after that. If that downgrade fails, it's not the end of the World, they can upgrade back again...

mailpoet/assets/js/src/automation/editor/components/filters/chip.tsx
JanJakes: Good catch! The translation is actually not needed here at all, the strings should come in translated. Fix: https://github.com/mailpoet/mailpoet/pull/5216/commits/5799461827702d7aac8c1f6394adc0050f300e30

mailpoet/assets/js/src/automation/integrations/core/steps/if-else/index.tsx
JanJakes: Thx. Fix: https://github.com/mailpoet/mailpoet/pull/5216/commits/8c6d27742f6f2693c6202ceb2765cfc4ff50b8cf

JanJakes:
> I can move horizontally only when i scroll to the bottom.

I implemented a new scrolling behavior that can also handle Figma-like scrolling in both directions on trackpads: https://github.com/mailpoet/mailpoet/pull/5216/commits/40fde04af58ec82d081f5eb341c427dc1a385607

I also already added drag-to-scroll: https://github.com/mailpoet/mailpoet/pull/5216/commits/34a19edaf9f3390747182d42da9ac5d8e6ecfa15

JanJakes:
Thanks for the catch, the fix is in premium: https://github.com/mailpoet/mailpoet-premium/pull/802/commits/2f299d9461ba18ce4f60fdeae1cefd4a8cd9fc55

mailpoet/assets/js/src/automation/editor/components/filters/chip.tsx
websupporter: Not sure if there are edge cases in translations where something like `%d: TranslatedLabel` would be the correct order of the elements. I don't think it's in LTR languages, but not sure if there are other cases maybe. This is why I would suggest something like `sprintf(strings.labelCount, filterCount)` and adding to the strings `labelCount: __('Filters: %d', 'mailpoet')`. But up to you, not super nice as well.

websupporter:
Thank you @JanJakes for the fixes!

I found another problem.

1. Add if/else step
2. Add step for yes
3. Add step for no
4. Delete if/else
5. Click Save
6. See: "Invalid automation structure: Unreachable steps found in automation graph"

It appears from the payload that the no step is not removed from the tree and now orphaned:

```
{
  "id": 2,
  "name": "New automation",
  "status": "draft",
  "created_at": "2023-10-20T14:02:41+00:00",
  "updated_at": "2023-10-20T08:06:45+00:00",
  "activated_at": null,
  "author": {
    "id": 1,
    "name": "admin"
  },
  "stats": {
    "automation_id": 2,
    "totals": {
      "entered": 0,
      "in_progress": 0,
      "exited": 0
    }
  },
  "steps": {
    "root": {
      "id": "root",
      "type": "root",
      "key": "core:root",
      "args": [],
      "next_steps": [
        {
          "id": "820b1f113f44b692"
        }
      ],
      "filters": null
    },
    "820b1f113f44b692": {
      "id": "820b1f113f44b692",
      "type": "trigger",
      "key": "mailpoet:custom-trigger",
      "args": {
        "hook": "my_custom_hook"
      },
      "next_steps": [
        {
          "id": "d26201d68a0620b1"
        }
      ],
      "filters": null
    },
    "d26201d68a0620b1": {
      "id": "d26201d68a0620b1",
      "type": "action",
      "key": "mailpoet:add-tag",
      "args": {
        "skip_triggers": false
      },
      "next_steps": [
        {
          "id": "dc43731660a32d54"
        }
      ],
      "filters": null
    },
    "dc43731660a32d54": {
      "id": "dc43731660a32d54",
      "type": "action",
      "key": "mailpoet:add-to-list",
      "args": {
        "segment_ids": [],
        "skip_triggers": false
      },
      "next_steps": [
        {
          "id": null
        }
      ],
      "filters": null
    },
    "f369332949c99113": {
      "id": "f369332949c99113",
      "type": "action",
      "key": "mailpoet:remove-from-list",
      "args": {
        "segment_ids": []
      },
      "next_steps": [
        {
          "id": null
        }
      ],
      "filters": null
    }
  },
  "meta": {}
}
```

See the step `f369332949c99113` in the payload

mailpoet/assets/js/src/automation/editor/components/automation/step.tsx
```diff
@@ -55,16 +55,21 @@ export function Step({ step, isSelected, context }: Props): JSX.Element {
...
   );
   const { openSidebar, selectStep } = useDispatch(storeName);
+  const { context } = useContext(AutomationContext);
   const compositeState = useContext(AutomationCompositeContext);
   const { batch } = useRegistry();
 
   const compositeItemId = `step-${step.id}`;
   const stepTypeData = stepType ?? getUnknownStepType(step);
 
+  const Footer = stepType.footer;
```
websupporter: `stepType` can be undefined.

just had an automation with a trigger from another branch and opened it in your built.

Expected: To see the "Unknown trigger" panel
Current behavior: Empty screen and console error `Uncaught TypeError: stepType is undefined`

JanJakes:
> Thank you @JanJakes for the fixes!
> I found another problem.

Great catch, @websupporter, thanks! I forgot to actually delete the subtree that we don't keep, when deleting an if/else step. Fix is in premium: https://github.com/mailpoet/mailpoet-premium/pull/802/commits/93e5d5efd10f53de7bbe7af67442e628101c5162

(Btw. I already prepared a ticket to actually ask the user which subtree to keep.)

mailpoet/assets/js/src/automation/editor/components/automation/step.tsx
JanJakes: 👍 Fix: https://github.com/mailpoet/mailpoet/pull/5216/commits/f15d82a3b1a8d2ec75dbfda37d73d79775a07633

mailpoet/assets/js/src/automation/editor/components/filters/chip.tsx
JanJakes: :+1: Makes sense, https://github.com/mailpoet/mailpoet/pull/5216/commits/e5dacda93c4354c1e4daf62147fd5f7cb3239e12 + https://github.com/mailpoet/mailpoet/pull/5216/commits/569f83a80295d63a09d9daa8cc219aeb7c86bc8b.

websupporter (dismissed):
lgtm. Thank you @JanJakes

Aschepikov:
Hi @JanJakes 
1. I see the issue with the "if else" step when I am trying to save the automation without setting up any condition there will be a "not set" label after I set up the condition the label is still here even after the updating/activating the automation the label will be here
image

2. Also, I saw the issue when the changes were not saving after setting up the email design. 
Steps to reproduce: create an automation with some steps and activate it, then I am adding a "send email" step to both paths yes/no, trying to update the active email. see the error with not set up
image
Then I am going to design one of them it 
image
After pressing the "save and continue" button I don't see the " send email" steps because it was downgraded to the last active saved version
image
Conditions to replicate this are at least two send email steps in both paths with a "not set" label and active automation. 

3. When I open the analytics for the automation with the "if/else" step I see these errors and a blank page
image

Aschepikov:
@JanJakes  Also, I have a question if we are deleting the "if/else" step should we delete everything that was connected with it?
For example here If I delete the first "if/else" step 
image
I can expect everything above this step will be deleted, but one of the branches will be saved
image
If I delete this step from the saved branch there will be some saved steps under it
image

JanJakes:
@Aschepikov Thanks!

1. Fixed.
2. This seems to be an unrelated bug, I'll create a ticket. The problem is that we're trying to save a draft before the redirect, but we can't save a draft when an automation is active.
3. Fixed.

> Also, I have a question if we are deleting the "if/else" step should we delete everything that was connected with it?

For now, it keeps the left branch ("Yes") and removes the right branch. We will add a feature to ask the user what to do.

JanJakes:
@websupporter Added a few fixes, you may want to have a quick look.

Aschepikov (approved):
GJ!

PR #5229, opened 2023-10-19
Add missing segment tracking [MAILPOET-5503]

johnolek:
## Description

This PR updates the tracking code to include usage of segments that have been added recently.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/assets/js/src/analytics.js
+++ b/mailpoet/assets/js/src/analytics.js
@@ -35,8 +35,6 @@ export function mapFilterType(filter) {
         return 'number of machine-opens';
       case 'opensAbsoluteCount':
         return 'number of opens';
-      case 'numberOfOrders':
-        return 'number of orders';
       case 'clicked':
         return 'clicked';
       case 'clickedAny':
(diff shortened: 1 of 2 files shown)


rodrigoprimo (approved):
Code changes look good to me.

I wonder if you checked if there is a way to dynamically add all existing segments to tracking instead of keeping this manual list of segments? I'm sure we will forget to update this list again in the future. I don't think we necessarily have to implement this in this PR. I'm just wondering if you checked if it is possible. Maybe we can create a separate ticket to look into this.

Feel free to move this PR to QA.

johnolek:
Thanks for the review, @rodrigoprimo!

> I wonder if you checked if there is a way to dynamically add all existing segments to tracking instead of keeping this manual list of segments?

This is immediately where my mind went when I started working on this ticket as well. Unfortunately I couldn't think of a straightforward way to do it, especially considering there's both JS and PHP that needs to be updated. I do think it's worth looking into, perhaps as part of a larger refactoring effort. At least for the time being we've become aware of this requirement and will be on the lookout for it when reviewing PRs.

I've added a subtask to ticket MAILPOET 5227, which I created a while ago to track general refactoring/improvement tasks.

Aschepikov (approved):
GJ!

rodrigoprimo:
Sounds good to me, @johnolek. Thanks.

PR #5231, opened 2023-10-20
Update codeception/verify package and fix tests [MAILPOET-5664]

rodrigoprimo:
## Description

This PR updates the package codeception/verify from version 1.5.0 to 2.1.1. Version 2 of this package introduced several breaking changes so it was necessary to update a huge number of tests. Luckily, it was possible to perform the majority of the updates using search and replace.

I tried to make this PR easier to review by creating a single commit per method that was changed. Only tests had to be updated so the changes proposed here should not affect the users.

codeception/verify doesn't seem to be a very popular package and I was not able to find a lot of information about the changes in the most recent versions. There is a comment in the release notes mentioning that the "expect function now works with expectations instead of verifiers" (https://github.com/Codeception/Verify/releases/tag/2.1.0) which is one of the changes that affect us the most. But also, some verifiers had their names changed: https://github.com/Codeception/Verify/blob/master/UPGRADE.md

I ended up replacing all the calls to `expect()` with `verify()` because at first, I was under the impression that this was the only option. So for example, I replaced `expect()->equals()` with `verify()->equals()`. Only after a while, I realized that it also could have been possible to replace it with `expect()->toEqual()`. I don't have a preference and when I realized that, I had already created several commits.

A bit more context here: p1697654234037389-slack-C01H71ZLBGQ

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5664]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5664]: https://mailpoet.atlassian.net/browse/MAILPOET-5664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/composer.json
```diff
@@ -18,7 +18,7 @@
     "codeception/module-cli": "^1.0",
     "codeception/module-db": "^1.1",
     "codeception/module-webdriver": "^1.4",
-    "codeception/verify": "1.5.0",
+    "codeception/verify": "^2.1",
```
websupporter, line 22: Why did you choose `2.1` and not `3.0`? https://github.com/Codeception/Verify/releases

websupporter (commented):
overall LGTM @rodrigoprimo. Thank you for that effort.

Besides the question about choosing this specific version of `codeception/verify` I was wondering, why there is no PR for premium. If I understand our setup correctly, we rely on the mailpoet `codeception` in premium (e.g.: [here](https://github.com/mailpoet/mailpoet-premium/blob/76352e4cda04887ac4d4b8ca4d3d99a69d9dde5f/tests/docker/codeception/docker-entrypoint.sh#L86) and [here](https://github.com/mailpoet/mailpoet-premium/blob/76352e4cda04887ac4d4b8ca4d3d99a69d9dde5f/RoboFile.php#L239)) and we have some usage of `expect()` which would then seem outdated to me ([example](https://github.com/mailpoet/mailpoet-premium/blob/76352e4cda04887ac4d4b8ca4d3d99a69d9dde5f/tests/unit/Config/HooksTest.php#L31-L36)).

mailpoet/composer.json
rodrigoprimo: We declare support for PHP 7.3 in our composer.json, and version 3.0 requires PHP >= 7.4.

rodrigoprimo:
> Besides the question about choosing this specific version of codeception/verify I was wondering, why there is no PR for premium. If I understand our setup correctly, we rely on the mailpoet codeception in premium (e.g.: [here](https://github.com/mailpoet/mailpoet-premium/blob/76352e4cda04887ac4d4b8ca4d3d99a69d9dde5f/tests/docker/codeception/docker-entrypoint.sh#L86) and [here](https://github.com/mailpoet/mailpoet-premium/blob/76352e4cda04887ac4d4b8ca4d3d99a69d9dde5f/RoboFile.php#L239)) and we have some usage of expect() which would then seem outdated to me ([example](https://github.com/mailpoet/mailpoet-premium/blob/76352e4cda04887ac4d4b8ca4d3d99a69d9dde5f/tests/unit/Config/HooksTest.php#L31-L36)).

@websupporter, I forgot about the premium plugin and that is why there was no PR for it.

websupporter (approved):
approved. will merge as it only concerns composer update and test env

PR #5238, opened 2023-10-24
Finish refactoring subscriber related methods in the Sending class [MAILPOET-4368]

rodrigoprimo:
## Description

In https://github.com/mailpoet/mailpoet/pull/5018, we refactored two subscriber related methods in the `Sending` class to use Doctrine instead of Paris. This refactors the remaining subscriber related methods. It also removes `mailpoet/lib/Tasks/Subscribers.php`.

Historically, we have used an incremental approach to refactor Paris to Doctrine. I'm a bit worried about it in the case of the Sending class as I had to fix some bugs that would have been easy to miss if not for our automated tests. All the bugs were related to Doctrine entities having stale data due to changes made using Paris. There are still several parts of the Sending class that use Paris so this PR might introduce more bugs that were not caught by our automation tests. I tried to add as many tests as possible, but the problem is that those bugs don't manifest themselves directly in the parts of the code that were changed.

That being said, I'm not sure if refactoring the whole Sending class in one push is feasible, but maybe we can build upon this PR and try it?

## Code review notes

As @lysyjan pointed out in this comment https://github.com/mailpoet/mailpoet/pull/5018#issuecomment-1638302275, I ended up taking an approach that is different than what was initially thought when the ticket was created. The idea was to move the subscriber related methods out of the Sending class maybe to repository classes. I refactored the methods to use Doctrine but kept them in the Sending class. We need to decide if we want to create a separate ticket to move the method out of the Sending class or if we tackle this in https://mailpoet.atlassian.net/browse/MAILPOET-4375. I'm inclined to think the first approach is the best one but I'm open to suggestions.

This PR was created on top of https://github.com/mailpoet/mailpoet/pull/5018, so the first six commits belong to https://github.com/mailpoet/mailpoet/pull/5018.

## QA notes

This PR should not change any functionality, but it touches sending emails which is a core part of MailPoet. It would be nice to test anything sending related that is not covered by automated tests.

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/5018

## Linked tickets

[MAILPOET-4368]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4368]: https://mailpoet.atlassian.net/browse/MAILPOET-4368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Newsletter/Sending/ScheduledTaskSubscribersRepository.php
```diff
@@ -140,6 +141,7 @@ private function checkCompleted(ScheduledTaskEntity $task): void {
     $count = $this->countBy(['task' => $task, 'processed' => ScheduledTaskSubscriberEntity::STATUS_UNPROCESSED]);
     if ($count === 0) {
       $task->setStatus(ScheduledTaskEntity::STATUS_COMPLETED);
+      $task->setProcessedAt(new Carbon());
```
JanJakes, line 179: Nice catch!

mailpoet/lib/Tasks/Sending.php
```diff
@@ -276,8 +276,10 @@ public function removeSubscribers(array $subscriberIds) {
     $this->updateCount();
   }
 
-  public function updateProcessedSubscribers(array $processedSubscribers) {
-    $this->taskSubscribers->updateProcessedSubscribers($processedSubscribers);
+  public function updateProcessedSubscribers(array $processedSubscribers): bool {
+    $this->scheduledTaskSubscribersRepository->updateProcessedSubscribers($this->scheduledTaskEntity, $processedSubscribers);
+    $this->scheduledTasksRepository->refresh($this->scheduledTaskEntity); // needed while Sending still uses Paris
```
JanJakes, line 276: I wonder if that comment is accurate. The change was done using Doctrine, so the refresh is probably not needed for Paris, but rather due to the fact that some query didn't reflect changes to a Doctrine entity?

That said, is the refresh needed at all? It seems that `updateProcessedSubscribers` and `checkCompleted` don't use DBAL, so the changes on `$this->scheduledTaskEntity` could simply be up-to-date.

mailpoet/lib/Tasks/Sending.php
```diff
@@ -209,22 +213,30 @@ public function save() {
   }
 
   public function delete() {
-    $this->taskSubscribers->removeAllSubscribers();
-    $this->task->delete();
-    $this->queue->delete();
+    $this->scheduledTaskSubscribersRepository->deleteByScheduledTask($this->scheduledTaskEntity);
+    $this->scheduledTasksRepository->remove($this->scheduledTaskEntity);
+
+    $sendingQueueEntity = $this->sendingQueuesRepository->findOneById($this->queue->id);
```
JanJakes: You can also do `$this->scheduledTasksRepository->getSendingQueue()`. It's a tiny improvement, but it eliminates the usage of `$this->queue`.

mailpoet/lib/Tasks/Sending.php
```diff
@@ -209,22 +213,30 @@ public function save() {
...
+
+    if (!$sendingQueueEntity) {
+      throw new InvalidStateException();
+    }
```
JanJakes, line 218: Rather than throwing an exception here, I think we could do this:

```php
if ($sendingQueueEntity) {
  $this->sendingQueuesRepository->remove($sendingQueueEntity);
}
```

That way, if the data is for any reason incorrect, it will still pass.

mailpoet/lib/Entities/SendingQueueEntity.php
```diff
@@ -223,4 +223,36 @@ public function getNewsletter() {
   public function setNewsletter(NewsletterEntity $newsletter) {
     $this->newsletter = $newsletter;
   }
+
+  public function toArray(): array {
```
JanJakes: I think the specific API response format belongs rather to [ResponseBuilders](https://github.com/mailpoet/mailpoet/tree/trunk/mailpoet/lib/API/JSON/ResponseBuilders), e.g. a new `SendingQueueResponseBuilder` class.

mailpoet/lib/Entities/SendingQueueEntity.php
```diff
@@ -223,4 +223,36 @@ public function getNewsletter() {
...
+    if (!$this->getTask() instanceof ScheduledTaskEntity) {
+      throw new \RuntimeException('Invalid state. SendingQueue has ScheduledTask associated.');
```
JanJakes: I think the message is missing a negation, such as "SendingQueue has **no** ScheduledTask associated."

mailpoet/lib/Tasks/Sending.php
rodrigoprimo: > I wonder if that comment is accurate. The change was done using Doctrine, so the refresh is probably not needed for Paris, but rather due to the fact that some query didn't reflect changes to a Doctrine entity?

Maybe I could rephrase the comment (please let me know if you think that is necessary and if you have a suggestion), but I'm not trying to say that the refresh is needed for Paris, but rather, while we still use Paris in the Sending class.

> That said, is the refresh needed at all? It seems that updateProcessedSubscribers and checkCompleted don't use DBAL, so the changes on $this->scheduledTaskEntity could simply be up-to-date.

This refresh is still needed. Without it, the test `\MailPoet\Test\Cron\Workers\SendingQueue\SendingQueueTest::testItDoesNotEnforceExecutionLimitsAfterSendingWhenQueueStatusIsSetToComplete()` and maybe others (I haven't checked while writing this comment) fail. This is the tricky thing about the changes introduced in this PR, they have the potential to break seemingly unrelated parts of the code. Speaking off the top of my head, using Doctrine here means that we are loading the Doctrine entity earlier. Later in the code execution, Paris code changes the rows associated with the loaded Doctrine entities and they became outdated. Calling refresh here, forces Doctrine to query the database again the next time the entity is used.

rodrigoprimo:
Just documenting that I had to rebase this branch and force push to fix a conflict. No code changes were needed to fix the conflicts and git was able to fix everything on its own.

mailpoet/lib/Entities/SendingQueueEntity.php
rodrigoprimo: Done in e7ba64c26f2705c9b57023939750b3cbb2684b86

mailpoet/lib/Entities/SendingQueueEntity.php
rodrigoprimo: I saw a few entities with a `toArray()` method, and that threw me off. I have moved the code to a ResponseBuilder in bdbfdbc7626afdf5cdf2691fc553cf9d640bb06f

mailpoet/lib/Tasks/Sending.php
rodrigoprimo: I'm ok with this change: cf54db89500a8580b77f3ffaed0999e2ad950194

mailpoet/lib/Tasks/Sending.php
rodrigoprimo: Thanks for your attention to the details: 4fcdd9a7f0b1bca4e421009a2ef4955fea628374

rodrigoprimo:
@JanJakes, I believe I addressed all of your comments. Could you please check this PR again?

mailpoet/lib/Tasks/Sending.php
JanJakes: @rodrigoprimo Ahh, I understand now, thanks for the clarification!

JanJakes (approved):
LGTM, thanks for the fixes!

Aschepikov (approved):
GJ!

PR #5241, opened 2023-10-25
Remove usages of MailPoet\Tasks\Sending — Part 1

JanJakes:
## Description

This PR removes the first batch of usages of `MailPoet\Tasks\Sending`.

## Code review notes

In the first part, this PR removes some fairly simple usages of `MailPoet\Tasks\Sending`. The second part removes the usages from `MailPoet\Cron\Workers\Scheduler` and does so gradually, method by method — the commits tell the story.

## QA notes

Please, verify that all types of email sending in the plugin work and that the emails render correcdtly:

1. Newsletters (immediate as well as scheduling).
2. Post notifications.
3. Welcome emails.
4. Transactional emails.
5. Automation emails (including transactional).
6. Re-engagement emails.
7. Stat notifications.

We have a fairly good test coverage for email sending, but it is worth verifying.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4375]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4375]: https://mailpoet.atlassian.net/browse/MAILPOET-4375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/StatsNotifications/Worker.php
+++ b/mailpoet/lib/Cron/Workers/StatsNotifications/Worker.php
@@ -15,7 +15,6 @@ use MailPoet\Mailer\MetaInfo;
 use MailPoet\Newsletter\Statistics\NewsletterStatisticsRepository;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Subscribers\SubscribersRepository;
-use MailPoet\Tasks\Sending;
 use MailPoet\Util\License\Features\Subscribers as SubscribersFeature;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
@@ -25,6 +24,7 @@ class Worker {
 
   const TASK_TYPE = 'stats_notification';
   const SETTINGS_KEY = 'stats_notifications';
+  const BATCH_SIZE = 5;
 
   /** @var Renderer */
   private $renderer;
@@ -96,7 +96,7 @@ class Worker {
     $settings = $this->settings->get(self::SETTINGS_KEY);
     // Cleanup potential orphaned task created due bug MAILPOET-3015
     $this->repository->deleteOrphanedScheduledTasks();
-    foreach ($this->repository->findScheduled(Sending::RESULT_BATCH_SIZE) as $statsNotificationEntity) {
+    foreach ($this->repository->findScheduled(self::BATCH_SIZE) as $statsNotificationEntity) {
       try {
         $extraParams = [
           'meta' => $this->mailerMetaInfo->getStatsNotificationMetaInfo(),
--- a/mailpoet/lib/Newsletter/Scheduler/WelcomeScheduler.php
+++ b/mailpoet/lib/Newsletter/Scheduler/WelcomeScheduler.php
@@ -2,9 +2,11 @@
 
 namespace MailPoet\Newsletter\Scheduler;
 
+use MailPoet\Cron\Workers\SendingQueue\SendingQueue;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\NewsletterOptionFieldEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
+use MailPoet\Entities\ScheduledTaskSubscriberEntity;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Entities\SubscriberEntity;
@@ -12,11 +14,14 @@ use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
 use MailPoet\Segments\SegmentsRepository;
 use MailPoet\Subscribers\SubscribersRepository;
-use MailPoet\Tasks\Sending as SendingTask;
+use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class WelcomeScheduler {
 
   const WORDPRESS_ALL_ROLES = 'mailpoet_all';
+
+  /** @var EntityManager */
+  private $entityManager;
 
   /** @var SubscribersRepository */
   private $subscribersRepository;
@@ -34,12 +39,14 @@ class WelcomeScheduler {
   private $scheduler;
 
   public function __construct(
+    EntityManager $entityManager,
     SubscribersRepository $subscribersRepository,
     SegmentsRepository $segmentsRepository,
     NewslettersRepository $newslettersRepository,
     ScheduledTasksRepository $scheduledTasksRepository,
     Scheduler $scheduler
   ) {
+    $this->entityManager = $entityManager;
     $this->subscribersRepository = $subscribersRepository;
     $this->segmentsRepository = $segmentsRepository;
     $this->newslettersRepository = $newslettersRepository;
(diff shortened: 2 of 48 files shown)


mailpoet/lib/Newsletter/Scheduler/WelcomeScheduler.php
```diff
@@ -118,28 +119,31 @@ public function createWelcomeNotificationSendingTask(NewsletterEntity $newslette
...
-
-    return $savedSendingTask;
+    ));
+    $this->entityManager->persist($task);
+
+    // queue
+    $queue = new SendingQueueEntity();
+    $queue->setTask($task);
+    $queue->setNewsletter($newsletter);
+    $queue->setSubscribers((string)$subscriberId);
```
JanJakes: Do you happen to know what's this column for? Seems to be a string (comma separated IDs?), but couldn't find any usages of `getSubscribers()` on the queue entity. It's the task entity where we keep the relationships, but I'm not sure what the queue subscribers column is for, and whether I should set it here or not.

mailpoet/lib/Cron/Workers/StatsNotifications/Worker.php
```diff
@@ -96,7 +96,7 @@ public function process($timer = false) {
     $settings = $this->settings->get(self::SETTINGS_KEY);
     // Cleanup potential orphaned task created due bug MAILPOET-3015
     $this->repository->deleteOrphanedScheduledTasks();
-    foreach ($this->repository->findScheduled(Sending::RESULT_BATCH_SIZE) as $statsNotificationEntity) {
+    foreach ($this->repository->findScheduled(self::BATCH_SIZE) as $statsNotificationEntity) {
```
rodrigoprimo, line 99: Can we remove `Sending::RESULT_BATCH_SIZE` since it doesn't seem to be used anymore?

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
```diff
@@ -582,15 +584,19 @@ public function testItCanProcessPostNotificationNewsletter() {
...
+    $task = $this->createTaskWithQueue($newsletter);
+    $task->setScheduledAt(Carbon::createFromTimestamp(WPFunctions::get()->currentTime('timestamp')));
+    $this->entityManager->flush();
+
+    // remove newsletter, but not any related data
+    $this->entityManager->getConnection()->delete(
+      $this->entityManager->getClassMetadata(NewsletterEntity::class)->getTableName(),
+      ['id' => $newsletter->getId()]
+    );
+    $this->entityManager->clear();
```
rodrigoprimo, line 543: It is not necessarily a problem, but it is not obvious to me why we need to call `clear()` here. The test passes without it.

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
```diff
@@ -768,149 +782,112 @@ public function testItProcessesScheduledJobsWhenNewsletterIsScheduled() {
...
-    verify($task->newsletterId)->equals($newsletter->getId());
-    verify($task->getSubscribers())->equals([$subscriber->getId()]);
+    $this->createTaskWithQueue($newsletter);
+
+    // remove subscriber, but not scheduled task subscriber
+    $this->entityManager->getConnection()->delete(
+      $this->entityManager->getClassMetadata(SubscriberEntity::class)->getTableName(),
+      ['id' => $subscriber->getId()]
+    );
+    $this->entityManager->clear();
```
rodrigoprimo, line 707: Same here, it doesn't seem like we need to call `clear()` here.

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
```diff
@@ -768,149 +782,112 @@ public function testItProcessesScheduledJobsWhenNewsletterIsScheduled() {
...
```
rodrigoprimo, line 772: One more

mailpoet/lib/Newsletter/Scheduler/WelcomeScheduler.php
rodrigoprimo: I don't know why we have a subscribers column in wp_mailpoet_sending_queues. Maybe it is a field that is not used anymore?

> whether I should set it here or not

Do you have any indication that we should set it? I'm inclined to think we don't need to set it. On a quick check, it seems that the code, even before the refactor that I did in the subscriber related methods of the Sending class, does not set the `subscribers` column in `createWelcomeNotificationSendingTask()`.

mailpoet/lib/Newsletter/Sending/SendingQueuesRepository.php
```diff
@@ -205,4 +205,20 @@ public function saveFilterSegmentMeta(SendingQueueEntity $queue, SegmentEntity $
     $queue->setMeta($meta);
     $this->flush();
   }
+
+  public function updateCounts(SendingQueueEntity $queue, ?int $count = null): void {
```
rodrigoprimo, line 209: Is it on purpose that this method doesn't call `flush()`? I'm not sure if there is a convention for that or not, but I would expect it to persist the changes to the database. That is what the other methods in this repository class seem to be doing.

rodrigoprimo (commented):
Great work! The changes look good to me. I left a few minor comments for us to discuss.

mailpoet/lib/Newsletter/Sending/SendingQueuesRepository.php
JanJakes: From the current usages it didn't seem necessary, but on the other hand, I agree it's safer to flush here.

Added: https://github.com/mailpoet/mailpoet/pull/5241/commits/08d95841d5596566e095a24a66306c0e5d453212

mailpoet/lib/Cron/Workers/StatsNotifications/Worker.php
JanJakes: Removed: https://github.com/mailpoet/mailpoet/pull/5241/commits/45efeda4b7e97d829e8453bc74a250597edd9078

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
JanJakes: Oh, good catch! It was a bug, this test (and the two others) passed even without the delete! It was because there was neither a scheduled task subscriber, not a matching segment, so it deleted the queue in any way. I fixed all of them, and now the deletes and `->clear()` calls are needed.

Fix: https://github.com/mailpoet/mailpoet/pull/5241/commits/fcbd3c81023885952dc189966779335441702a89

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
JanJakes: Fix: https://github.com/mailpoet/mailpoet/pull/5241/commits/810e28f645b5b45eaf25b96ec3999b60ae98f0dd

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
JanJakes: Fix: https://github.com/mailpoet/mailpoet/pull/5241/commits/6b2120a9ce8bedd9eb894f282b8be34ae4ba7391

mailpoet/lib/Newsletter/Scheduler/WelcomeScheduler.php
JanJakes: You're right, I think I added that by mistake instead of setting scheduled task subscribers to the `$task`, then noticed those were missing, and then added those too. This indeed seems to do nothing.

I checked all usages and removed them: https://github.com/mailpoet/mailpoet/pull/5241/commits/388b078725a12f93aa95aa0bb736862926640bf7

JanJakes:
@rodrigoprimo Thanks for the review and your observations! It made me double-check all of the scheduler tests one by one and reveal more issues and unnecessary code. I fixed, improved, and simplified more of the tests, starting from https://github.com/mailpoet/mailpoet/pull/5241/commits/fcbd3c81023885952dc189966779335441702a89.

rodrigoprimo (approved):

Aschepikov (approved):
GJ!

PR #5245, opened 2023-10-25
git hooks for automatically installing dependencies [MAILPOET-4237]

johnolek:
## Description

This PR adds a few husky hooks for automatically installing JS/PHP dependencies when an update is detected (a change in a lock file).

After a P2 discussion, it was also decided that we should make all git hooks configurable via environment variables, so that is now included as well.

## Code review notes

The ticket only mentions a `post-merge` hook for installing dependencies, but I think this would be useful behavior for all additional hooks I included.

## QA notes

QA shouldn't be necessary.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/840

## Linked tickets

[MAILPOET-4237](https://mailpoet.atlassian.net/browse/MAILPOET-4237)

## After-merge notes

We will need to publish a P2 announcing the changes.

## Tasks

[MAILPOET-4237]: https://mailpoet.atlassian.net/browse/MAILPOET-4237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


johnolek:
@JanJakes I know you have other PRs on your plate right now, but I remember you saying you don't always use the docker environment so I'm especially interested to get your opinion on the approach I took. This is also a relatively small PR so I hope it doesn't take much time.

Happy to reassign if you're swamped with other things!

JanJakes (commented):
> Is it ok to assume people are using the docker environment and will be able to run do commands? Or should I make it more generic?

For a "docker environment", I would say "no" (having to rely on Docker to only switch a Git branch feels unpractical), but from the code, I see no dependency on Docker, so I suppose what you mean is our "dev" environment, the `./do` commands and such, and that I find absolutely fine.

> I remember you saying you don't always use the docker environment so I'm especially interested to get your opinion on the approach I took.

I'm always using our dev environment (`./do` commands), but wasn't using Docker for some time (was fed up with Docker for Mac), but since I have [OrbStack](https://orbstack.dev/), I'm back to Docker in full, as it resolved all my issues and performance.

---

I would also add a rather personal note/request. It may be that not everyone is a fan of these hooks. I am one of these people — I personally find it very disturbing, if the Git commands take more than a 1 second or so.

I'm using the `git` command pretty extensively, switching, rebasing, amending, creating temporary branches, backup branches, etc., and even with the current commit lints, it can completely throw me out of focus (a command takes multiple seconds, I scroll Twitter, loose context), so I'm almost always skipping the hooks.

What if we added an ENV variable for each of our hooks and made them configurable? Such as `MP_GIT_HOOKS_ESLINT=true`, `MP_GIT_HOOKS_INSTALL_JS=false`, etc.? I'm not sure how much work that is, so we can also create a separate ticket.

I also know that I can (and probably should) use `HUSKY_SKIP_HOOKS=true` to just skip all hooks, but maybe it would be nice to be able to configure them granularly. I'm also curious to know if I'm the only one who's not much of a fan of these hooks, or what's the general opinion (we could write a short P2), but configurability can solve this for all of us, I guess.

EDIT: Made this into a [P2 post](https://mailpoetdev.wordpress.com/2023/10/26/git-hooks-in-our-repository/).

johnolek:
@JanJakes Thank you very much for sharing your perspective, and for writing the P2!

For some reason I still associate the `do` commands as things that are run through containers. Thank you for reminding me that these particular commands don't require a running container.

I'll share my thoughts about these hooks in general in the P2 so we can have all of that discussion in one place.

johnolek:
@JanJakes Sorry it's taken me so long to revisit this. I just pushed some changes that make the commands configurable via environment variables. This was a bit trickier than I expected because of how `lint-staged` runs commands, and I thought it would be best to extract those commands into separate scripts for better readability and to have more control.

I'm sure there's a lot of room for improvement but I'd like to get your input on where things are right now, whenever you have time (I expect you'll be busy next week).

mailpoet/package.json
```diff
@@ -17,14 +17,9 @@
...
-    "*.{js,jsx,ts,tsx}": "eslint --max-warnings 0",
-    "*.php": [
-      "phplint",
-      "./do qa:code-sniffer",
-      "./do qa:minimal-plugin-standard",
-      "bash -c './do qa:phpstan'"
-    ]
+    "*.{scss,css}": "tasks/lint-staged-styles.sh",
+    "*.{js,jsx,ts,tsx}": "tasks/lint-staged-scripts.sh",
+    "*.php": "tasks/lint-staged-php.sh"
```
JanJakes, line 22: Maybe, to unify with `lint-staged-php.sh`, we could name the other two `lint-staged-js.sh` and `lint-staged-css.sh`.

mailpoet/tasks/lint-staged-php.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
JanJakes, line 1: I think for all the `sh` scripts, we may need to set `set -e` so that the script exits on the first failed command.

mailpoet/tasks/lint-staged-php.sh
```diff
@@ -0,0 +1,33 @@
...
+
+source $PWD/.env
+
+if [ "$MP_GIT_HOOKS_ENABLE" != "true" ]; then
+    echo "MP_GIT_HOOKS_ENABLE is not set to 'true'. Skipping lint-staged-php."
+    exit 0
```
JanJakes: Sometimes the indents are 2 spaces, sometimes 4. I've seen this also in the other scripts.

.husky/common.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
JanJakes, line 1: Maybe here on in the particular scripts, we should also set some defaults for cases when the variables are not defined. Something like this should work:

```sh
export MP_GIT_HOOKS_ENABLE="${MP_GIT_HOOKS_ENABLE:-true}"
```

As for the defaults, I think we could do more or less these:
```sh
MP_GIT_HOOKS_ENABLE=true
MP_GIT_HOOKS_ESLINT=true
MP_GIT_HOOKS_STYLELINT=true
MP_GIT_HOOKS_PHPLINT=true
MP_GIT_HOOKS_CODE_SNIFFER=true
MP_GIT_HOOKS_MINIMAL_PLUGIN_STANDARDS=true
MP_GIT_HOOKS_PHPSTAN=false
MP_GIT_HOOKS_INSTALL_JS=false
MP_GIT_HOOKS_INSTALL_PHP=false
```

Or do you think we should set different defaults? To me, it seems that these would sort of preserve BC, while disabling PHPStan, that is problematic and slow.

mailpoet/.env.sample
```diff
@@ -44,3 +44,19 @@ WP_TEST_PERFORMANCE_DATA_URL=
...
+MP_GIT_HOOKS_STYLELINT=true
+
+MP_GIT_HOOKS_PHPLINT=true
+MP_GIT_HOOKS_CODE_SNIFFER=true
+MP_GIT_HOOKS_MINIMAL_PLUGIN_STANDARDS=true
+MP_GIT_HOOKS_PHPSTAN=true
+
+MP_GIT_HOOKS_INSTALL_JS=true
+MP_GIT_HOOKS_INSTALL_PHP=true
+# End git hooks configuration
```
JanJakes: This is just a tiny stylistic note, but the begin/end comments seem to be rather inconsistent with the rest of the file. I think a single block with a single heading above is fine:

```sh
# Git hooks configuration
MP_GIT_HOOKS_ENABLE=true
MP_GIT_HOOKS_ESLINT=true
MP_GIT_HOOKS_STYLELINT=true
MP_GIT_HOOKS_PHPLINT=true
MP_GIT_HOOKS_CODE_SNIFFER=true
MP_GIT_HOOKS_MINIMAL_PLUGIN_STANDARDS=true
MP_GIT_HOOKS_PHPSTAN=true
MP_GIT_HOOKS_INSTALL_JS=true
MP_GIT_HOOKS_INSTALL_PHP=true
```

JanJakes (commented):
Thanks for working on this for all the hooks!

mailpoet/package.json
johnolek: Good call, updated in c64335398a0479b210fc97a665b24b2755a87e2a

mailpoet/tasks/lint-staged-php.sh
johnolek: Thanks for catching this, updated in e27a5def34eeb013eb7b60e5c2a58309ebb1fc04

mailpoet/tasks/lint-staged-php.sh
johnolek: Fixed indentation in b9a0d3cb934c77a1f9279d335d555e8d2cd05385

.husky/common.sh
johnolek: I was on the fence about including defaults. I was thinking I would just publish a P2 and everyone could update their own `.env` based on their preferences, but maintaining backwards compatibility makes sense to me.

I added these defaults in 9ac1b84287385aec2e6c97c91eba80d63addd56c. I decided to leave phpstan enabled for true BC, but if you feel strongly about defaulting to `false` I'm ok changing it.

On the topic of PhpStan, I think the command could be improved to optionally only run on staged files (basically make the command accept a list of files to scan), which would speed things up if the cache gets cleared. I started playing around with this but ran into a problem when the robofile itself failed phpstan. I would need to make sure it doesn't run on files we don't normally include in the scan. Does this seem like a worthwhile idea to pursue? I could write a janitor ticket.

mailpoet/.env.sample
johnolek: I agree, updated in 244c9ab73f839fa955f9288679926d851b9c0d64

johnolek:
Thanks for the review @JanJakes! I think I've addressed everything. Will you please take one more look?

.husky/common.sh
JanJakes: > On the topic of PhpStan, I think the command could be improved to optionally only run on staged files

@johnolek Is that possible with PHPStan? I think it can't work. Unlike linters, PHPStan can't view files independently, meaning that scanning only some files is likely impossible, or, more precisely, the type inference would then work only from the specific files, leading to incomplete or unknown types, and if you cause type incompatibilities elsewhere, it wouldn't catch them.

It actually seems to be explained here: https://phpstan.org/blog/why-you-should-always-analyse-whole-project

What we could do, though, is to stash all the new unstaged files, and after running the command we could pop the stash.

JanJakes (approved):
@johnolek Thanks for the changes and for working on this!

Feel free to merge this anytime (no QA needed). I'm passing it back to you only in case you want to try to handle PHPStan as per https://github.com/mailpoet/mailpoet/pull/5245/commits/9ac1b84287385aec2e6c97c91eba80d63addd56c#r1474484351.

.husky/common.sh
johnolek: That makes sense, thanks for the explanation and link. I think it's probably best to just leave it alone.

johnolek:
@JanJakes I'm going to leave this PR as is, but I'm going to hold off on merging until the premium PR is also ready to merge (I requested your review on that as well). Thanks for all your help with this!

PR #5247, opened 2023-10-26
Order created/cancelled/completed triggers [MAILPOET-5661]

websupporter:
## Description
Adds the order created, cancelled and completed triggers.

## Code review notes
While testing I realized that `wc_create_order()` has no way of setting the billing email address for guest users. This is a problem, because when no customer is set `wc_create_order()` would not trigger our automation. My workaround is in such cases to start to listen to `woocommerce_after_order_object_save` until the email address is set.

The `wc_create_order()` description says:

> Create a new order programmatically.Returns a new order object on success which can then be used to add additional data.

The current solution works when the billing email address is added subsequently in the same PHP process (see the test). Still, our automation would not catch orders if the email address would be added in another PHP process. It looks to me that this is an edge case we can look past. In my tests the default order creation through the interface worked fine.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5661]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5661]: https://mailpoet.atlassian.net/browse/MAILPOET-5661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Automation/Engine/WordPress.php
```diff
@@ -21,6 +21,10 @@ public function addAction(string $hookName, callable $callback, int $priority =
     return add_action($hookName, $callback, $priority, $acceptedArgs);
   }
 
+  public function removeAction(string $hookName, callable $callback, int $priority = 10): bool {
+    return remove_action($hookName, $callback, $priority);
+  }
```
johnolek, line 26: Is this being used anywhere? No objection to keeping it, just want to make sure I'm not missing anything.

mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/Orders/OrderCancelledTrigger.php
```diff
@@ -0,0 +1,28 @@
...
+  public function getKey(): string {
+    return 'woocommerce:order-cancelled';
+  }
+
+  public function getName(): string {
+    return __('Order cancelled', 'mailpoet');
+  }
+
+  public function isTriggeredBy(StepRunArgs $args): bool {
+    /** @var OrderStatusChangePayload $orderPayload */
```
johnolek, line 20: Not complaining, but does this type not come through automatically from this?

```php
  /**
   * @template P of Payload
   * @param class-string<P> $class
   * @return P
   */
  public function getSinglePayloadByClass(string $class): Payload {
```

My IDE infers the type correctly, but maybe this is for PHPStan?

mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/Orders/OrderCreatedTrigger.php
```diff
@@ -0,0 +1,100 @@
...
+    $this->processedOrders[] = $orderId;
+    $this->wp->doAction(Hooks::TRIGGER, $this, [
+      new Subject(OrderSubject::KEY, ['order_id' => $order->get_id()]),
+      new Subject(CustomerSubject::KEY, ['customer_id' => $order->get_customer_id()]),
+    ]);
+  }
+
+  public function isTriggeredBy(StepRunArgs $args): bool {
+    return true;
+  }
```
johnolek: This trigger is always triggered? If this is necessary it might be good to leave a comment.

mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/Orders/OrderStatusChangedTrigger.php
```diff
@@ -19,10 +19,10 @@
 class OrderStatusChangedTrigger implements Trigger {
 
   /** @var Functions */
-  private $wp;
+  protected $wp;
```
johnolek, line 22: Is there a reason we sometimes use `WPFunctions` and sometimes use `WordPress` in the automations codebase?

johnolek (changes requested):
Nothing major in the comments other than my question about `isTriggeredBy`.

Is there no way to use the existing order status changed trigger for new orders? I can see how it might be tricky if there isn't a specific status transition we can always rely on occurring for every single order. I'm guessing you looked into this and am just curious what you found.

One other general question is about how the triggers work. If I understand it correctly, the OrderStatusChangedTrigger fires an action that cancelled/completed triggers listen for. Since triggers only register their hooks if there is an active automation (https://github.com/mailpoet/mailpoet/blob/dd26234d14147460a51bf7092cb0a2c271261312/mailpoet/lib/Automation/Engine/Engine.php#L90-L91), will this work if a user doesn't have an active OrderStatusChanged automation?

mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/Orders/OrderStatusChangedTrigger.php
websupporter: Good catch! Thank you. Replaced all `Functions` with the `Engine\WordPress` in 022bf27a1

mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/Orders/OrderCreatedTrigger.php
websupporter: Added a comment in 3f822c13a

mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/Orders/OrderCancelledTrigger.php
websupporter: Hmmm... Works for me now in my IDE. Sometimes it does not (maybe some indexing process while trying), this is why I added it. Removed it in 022bf27a1

mailpoet/lib/Automation/Engine/WordPress.php
websupporter: No actually not. I needed it in an earlier iteration and forgot to remove it. When I saw I did not remove it I was also like "No objection to keeping it"

websupporter:
Thank you for your review @johnolek 

> Is there no way to use the existing order status changed trigger for new orders?

Not in a reliable way. E.g. `wp_create_order(['status' => 'on-hold'])` counts for me still as an order created. Additionally, there is not necessarily a status transition happening. The default is just `pending` but there is not from `new to processing` transition happening. [here](https://github.com/woocommerce/woocommerce/blob/9570483f2e73799a14233e6c0b80d6b387868656/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L627-L652) you can see the mapping. If from and to do not differ no transition hook will be fired. So, if we want to catch the order creation, I concluded to go with the hook Woocommerce provides.

> If I understand it correctly, the OrderStatusChangedTrigger fires an action that cancelled/completed triggers listen for

That is not the case. The cancelled and completed triggers listen to native WooCommerce hooks. Canceled and completed extend the status trigger hook and use the same `registerHook()` method. This way they are listing to the same hooks the status trigger is listening to. So they basically all share the initiation (listening and handling) [here](https://github.com/mailpoet/mailpoet/blob/893f96f180ec5cece9d29e0339f2bbf36761af07/mailpoet/lib/Automation/Integrations/WooCommerce/Triggers/OrderStatusChangedTrigger.php#L61-L84) but then differ in `isTriggeredBy()`.

> will this work if a user doesn't have an active OrderStatusChanged automation

Yes :)

johnolek:
Thanks for the explanation about the necessity of the order created logic!

> That is not the case.

I'm glad to have been wrong! Thanks for the explanation. I still don't have a very intuitive handle on a lot of the automations codebase (but this is something I should've been able to figure out myself).

I noticed one minor issue with the `OrderCreatedTriggerTest` that only seems to be an issue if you run tests with the docker container (like how I run them in PhpStorm). The test passes once but fails on subsequent runs unless you manually delete the test automation. 

The issue is that the [database doesn't get cleaned up until after WordPress loads](https://github.com/mailpoet/mailpoet/blob/3f822c13ad3a054ad4583208ae70a13d9040a5c5/mailpoet/tests/integration/_bootstrap.php#L23-L49). When WP loads it initializes the Automation engine, which registers hooks for active automations, and the automation from the previous run is still there and active. Since you're [calling](https://github.com/mailpoet/mailpoet/blob/3f822c13ad3a054ad4583208ae70a13d9040a5c5/mailpoet/tests/integration/Automation/Integrations/WooCommerce/Triggers/Orders/OrderCreatedTriggerTest.php#L41-L41) `registerHooks` manually in the test and because the trigger is [not configured to be shared](https://github.com/mailpoet/mailpoet/blob/3f822c13ad3a054ad4583208ae70a13d9040a5c5/mailpoet/lib/DI/ContainerConfigurator.php#L203-L203), this ends up registering two separate instances of the trigger which don't share state (i.e. knowing which order IDs have been processed) and therefore create 2 runs when we expect 1.

If anything this seems like something we could/should fix in the tests, perhaps adding an after suite callback to clean the database: https://github.com/mailpoet/mailpoet/blob/3f822c13ad3a054ad4583208ae70a13d9040a5c5/mailpoet/tests/_support/IntegrationCleanupExtension.php#L20-L20. 

No need to do that as part of this PR (the tests always run fine when using the `do` commands) but I'd be interested to hear your opinion.

mailpoet/assets/js/src/automation/integrations/woocommerce/steps/order-cancelled/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { __, _x } from '@wordpress/i18n';
+import { StepType } from '../../../../editor/store';
+import { Icon } from './icon';
+
+const keywords = [__('woocommerce', 'mailpoet'), __('order', 'mailpoet')];
```
johnolek: @websupporter While testing I noticed I couldn't find the triggers by searching for `cancelled`, `completed` or `created` which seems like something we should fix. Do you know if there's a reason why the search doesn't also search the titles of triggers?

johnolek:
@websupporter Thanks again for the explanations and the updates. The only thing I'd still like to see changed is something I didn't notice before, namely the keywords for the new triggers (see [this comment](https://github.com/mailpoet/mailpoet/pull/5247/files#r1374877956))

mailpoet/assets/js/src/automation/integrations/woocommerce/steps/order-cancelled/index.tsx
websupporter: Ah! Thanks. Fixed it in https://github.com/mailpoet/mailpoet/pull/5247/commits/3135b8c6836f86c330164e746875d9a931733870

> Do you know if there's a reason why the search doesn't also search the titles of triggers

We first used only the title but then at some point the title needed to be able to return `JSX.Element` and we introduced keywords.

johnolek: @websupporter Thanks for making that tweak!

Could we update filterItems to include `title(null, 'inserter')` if we detect that it's a string and not JSX? Or if we wanted to include the JSX in the search too we could use [`renderToStaticMarkup`](https://react.dev/reference/react-dom/server/renderToStaticMarkup) or similar.

https://github.com/mailpoet/mailpoet/blob/3f822c13ad3a054ad4583208ae70a13d9040a5c5/mailpoet/assets/js/src/automation/editor/components/inserter/index.tsx#L16-L22

No need to change it for this PR, just curious if it seems like it would be helpful. This way the keywords would only ever have to be words not included in the title, which seems nicer to me, especially if there is a difference between how a keyword is translated individually vs its translation in the full string.

johnolek (dismissed):

Aschepikov (dismissed):
GJ!

Aschepikov:
Hi @websupporter  there are some conflicts can you resolve them pls

websupporter:
rebased @Aschepikov

Aschepikov (approved):
GJ!

PR #5250, opened 2023-10-26
Refactor UnsubscribeTokens to use Doctrine instead of Paris [MAILPOET-5680]

rodrigoprimo:
## Description

This PR refactors UnsubscribeTokens and its test class to use Doctrine instead of Paris.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5680]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5680]: https://mailpoet.atlassian.net/browse/MAILPOET-5680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/UnsubscribeTokens.php
+++ b/mailpoet/lib/Cron/Workers/UnsubscribeTokens.php
@@ -2,11 +2,14 @@
 
 namespace MailPoet\Cron\Workers;
 
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
-use MailPoet\Models\Newsletter;
-use MailPoet\Models\Subscriber;
+use MailPoet\Entities\SubscriberEntity;
+use MailPoet\InvalidStateException;
 use MailPoet\Util\Security;
 use MailPoetVendor\Carbon\Carbon;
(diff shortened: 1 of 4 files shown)


johnolek (dismissed):
This all looks good to me, nice work @rodrigoprimo!

The only idea I had for a minor improvement was that the query could be changed to only fetch the `id` since it's the only piece of data we need to retrieve

i.e. instead of 

`->select('e')`

we could have 

`->select('PARTIAL e.{id}')`

but after testing it locally with large amounts of test data (10,000 subscribers) it didn't seem to make much, if any, difference, I'm guessing because we're processing them in batches.

I'm only bringing this up in case you think it might be beneficial to make this change anyway, but as-is the query this is generating is essentially the same as in the old version, which feels safer anyway.

rodrigoprimo:
@johnolek, good point. I agree that it is not going to make a big difference in this case, but I think it is a good practice to get only the data we actually need. These things can pile up. I added another commit implementing the change that you suggested. Could you please check this PR again?

johnolek (approved):
@rodrigoprimo Thanks for the update. Looks good to me!

Aschepikov (approved):
GJ!

PR #5260, opened 2023-11-01
Add email editor columns rendering [MAILPOET-5591]

lysyjan:
## Description

This PR implements email rendering of Gutenberg `core/columns` and `core/column` blocks. As a part of this solution is also refactoring of the rendering process to be more aligned with WP.

## Code review notes

- Because I needed to rebase this branch twice, the review can be more difficult.
- There are also several follow-up tickets.
- I added preprocessing of blocks for calculating the width, typography inheritance, etc.
- I could not test the final output in Litmus because we overreached the rendering limit.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5591]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5591]: https://mailpoet.atlassian.net/browse/MAILPOET-5591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/CleanupPreprocessor.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Engine\Renderer\Preprocessors;
+
+class CleanupPreprocessor implements Preprocessor {
+  public function preprocess(array $parsedBlocks, array $layoutStyles): array {
+    foreach ($parsedBlocks as $key => $block) {
+      // https://core.trac.wordpress.org/ticket/45312
+      // \WP_Block_Parser::parse_blocks() sometimes add a block with name null that can cause unexpected spaces in rendered content
+      // This behavior was reported as an issue, but it was closed as won't fix
```
costasovo, line 10: Thanks for adding this comment and link to the track ticket!

mailpoet/lib/EmailEditor/Engine/Renderer/template.html
```diff
@@ -5,16 +5,18 @@
   xmlns:v="urn:schemas-microsoft-com:vml"
   xmlns:o="urn:schemas-microsoft-com:office:office"
 >
-  <title>{{newsletter_subject}}</title>
-  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
-  <meta name="viewport" content="width=device-width, initial-scale=1" />
-  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-  <meta name="format-detection" content="telephone=no" />
-  {{email_meta_robots}}
+  <head>
```
costasovo, line 8: Could you please add to the commit message what was fixed by this addition?
I remember it was because GMAIL email apps were stripping CSS from the head, but I will forget soon. It would be nice to document this knowledge, at least in the commit message.

mailpoet/assets/js/src/email-editor/engine/editor.tsx
```diff
@@ -5,6 +5,7 @@ import { registerCoreBlocks } from '@wordpress/block-library';
 import { Popover, SlotFillProvider } from '@wordpress/components';
 import { ShortcutProvider } from '@wordpress/keyboard-shortcuts';
 import { EntityProvider } from '@wordpress/core-data';
+import { Hooks } from '../../hooks';
```
costasovo: Please use `import { addFilter } from '@wordpress/hooks';`. We use `@wordpress/dependency-extraction-webpack-plugin` plugin for fetching modules directly from WP

mailpoet/assets/js/src/email-editor/engine/editor.tsx
```diff
@@ -33,12 +34,41 @@ function Editor() {
...
+            },
+          },
+        };
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      return settings;
+    },
+  );
+}
```
costasovo: I suggest moving this code somewhere else. 
My idea: We could add `blocks` folder on top level and add a file per block.
So this code would live in `blocks\core\column`.
Then in `blocks/index.ts` we could have this code:
```js
export function initBlocks() {
   disableNestedColumns();
   registerCoreBlocks();	
}
```
What do you think?

mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
```diff
@@ -7,11 +7,16 @@
...
   public function preprocess(array $parsedBlocks, array $layoutStyles): array {
-    $layoutWidth = $layoutStyles['width'];
+    $layoutWidth = $this->parseNumberFromStringWithPixels($layoutStyles['width']);
     // Subtract padding from the width of the layout element
-    $layoutWidth -= $layoutStyles['padding']['left'] ?? 0;
-    $layoutWidth -= $layoutStyles['padding']['right'] ?? 0;
+    $layoutWidth -= $this->parseNumberFromStringWithPixels($layoutStyles['padding']['left'] ?? '0px');
+    $layoutWidth -= $this->parseNumberFromStringWithPixels($layoutStyles['padding']['right'] ?? '0px');
     foreach ($parsedBlocks as $key => $block) {
       $width = $this->convertWithToPixels($block['attrs']['width'] ?? '100%', $layoutWidth);
```
costasovo: A typo `convertWithToPixels` -> `convertWidthToPixels`.

mailpoet/lib/EmailEditor/Engine/StylesController.php
```diff
@@ -5,9 +5,9 @@
 class StylesController {
   /**
    * Width of the email in pixels.
-   * @var int
+   * @var string
    */
-  const EMAIL_WIDTH = 660;
+  const EMAIL_WIDTH = '660px';
```
costasovo: Could you please explain a bit why we need this change? To me, it seems that this forces us to use `parseNumberFromStringWithPixels` to get back the int value. Maybe we could return values with units only from the `BlocksWidthPreprocessor` and keep int?

mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
```diff
@@ -7,11 +7,16 @@
  * The final width in pixels is stored in the email_attrs array because we would like to avoid changing the original attributes.
  */
 class BlocksWidthPreprocessor implements Preprocessor {
+
+  const BLOCKS_WITHOUT_WIDTH = [
+    'core/paragraph',
+  ];
```
costasovo, line 13: I'm afraid in the future when we add third-party blocks or universal renderers, we won't be able to have such constants, or we will have to require a third party to give us the information if the block has or hasn't width.
Could we make it work without such a list?

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
```diff
@@ -30,19 +30,19 @@ private function prepareColumnTemplate(array $parsedBlock): string {
...
-              <td align="left" style="font-size:0px;padding-left:' . $paddingLeft . ';padding-right:' . $paddingRight . ';padding-bottom:' . $paddingBottom . ';padding-top:' . $paddingTop . ';">
-                <div style="line-height:1;text-align:left;">{column_content}</div>
-              </td>
-            </tr>
-          </tbody>
-        </table>
-      </div>
-      <!--[if mso | IE]></td><![endif]-->
+      <td class="block" style="vertical-align:top;width:' . $width . ';background:' . $backgroundColor . ';background-color:' . $backgroundColor . ';">
+        <div class="email_column" style="background:' . $backgroundColor . ';background-color:' . $backgroundColor . ';width:100%;max-width:' . $width . ';font-size:0px;text-align:left;display:inline-block;vertical-align:top;">
```
costasovo, line 35: Could you please add more details about why this change was needed? Maybe also to the commit message.
 I wonder if we need to enable the table cell for all email clients. What if we move te `block` class to the div inside?

mailpoet/lib/EmailEditor/Engine/Renderer/BlocksRegistry.php
```diff
@@ -9,6 +9,12 @@ class BlocksRegistry {
 
   public function addBlockRenderer(string $blockName, BlockRenderer $renderer): void {
     $this->blockRenderersMap[$blockName] = $renderer;
+    add_filter('render_block_' . $blockName, [$renderer, 'render'], 10, 2);
+  }
```
costasovo: Just an idea. What if instead of registering ` [$renderer, 'render']` directly, we could register 
```php
 add_filter('render_block_' . $blockName, [$this, 'renderBlock'], 10, 2);`
 ```
  and then in render block  we could do:
 ```php
 public function renderBlock($blockContent, $parsedBlock) {
   $renderer = $this->blockRendererMap[$parsedBlock['name']]->render($blockContent, $parsedBlock, $potencialyAdditionalParameter);
 }
 ```
 I think this way we could allow passing `$settingsController` as a parameter to the render method rather than adding it as a dependency in the render parent class (done in follow up commit).
 
 I think it is more convenient because when some third-party block renderer wants to prepare their own renderer and wants to use data from `$settingsController` we will not force them to figure out where to get the settings controller to pass it to the constructor of the renderer.
 What do you say?

File: mailpoet/lib/EmailEditor/Engine/Renderer/BlockRenderer.php [hunk 10]
```diff
@@ -2,6 +2,28 @@
 
 namespace MailPoet\EmailEditor\Engine\Renderer;
 
-interface BlockRenderer {
-  public function render(string $blockContent, array $parsedBlock): string;
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\EmailEditor\Engine\SettingsController;
+
+abstract class BlockRenderer {
```
costasovo: I would rather avoid adding abstract classes if possible. It makes it harder for extenders to construct renderer because they need to somehow get the `$settingsController`. Please see my other comment in RendrererRegistry

mailpoet/lib/EmailEditor/Engine/Renderer/BlockRenderer.php
```diff
@@ -2,6 +2,28 @@
...
+    $layoutStyles = $this->settingsController->getEmailLayoutStyles();
+    $width = $this->parseNumberFromStringWithPixels($layoutStyles['width']);
+    $width -= $this->parseNumberFromStringWithPixels($layoutStyles['padding']['left']);
+    $width -= $this->parseNumberFromStringWithPixels($layoutStyles['padding']['right']);
+    return "{$width}px";
+  }
+
+  private function parseNumberFromStringWithPixels(string $string): float {
+    return (float)str_replace('px', '', $string);
+  }
```
costasovo: This works only with data from the settings controller so I guess we could move it into the controller.

costasovo (commented):
Great job! The PR is huge but also a huge improvement!

I added a couple of comments with ideas, questions and suggestions.
Thanks again for looking into this difficult task!

mailpoet/lib/EmailEditor/Engine/Renderer/template.html
lysyjan: Good point 👍 I will do it.

mailpoet/assets/js/src/email-editor/engine/editor.tsx
lysyjan: I like your idea. Nevertheless, I thought something like this could be done later.
Do you agree we can create a follow-up ticket?

mailpoet/lib/EmailEditor/Engine/StylesController.php
lysyjan: When we work with the email width in the preprocessor, it was for me more clear to work with a value containing units. Also, we build email layout styles in `SettingsController`, and there should not be width without units.
Of course, it's not necessary and I can change it or we can change it later when it is too annoying for us.

mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
lysyjan: That is a good point. I guess we can manage this in the paragraph block renderer.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
lysyjan: I'm sorry for your confusion. I added another table wrapper to unify the height of the `core/column` blocks inside a `core/columns`. For example, when you set a different background for each column, the height of the block depends on the content inside. This change fixes this issue.
[Here](https://d.pr/i/eu7Dz3) is an image containing a comparison.

mailpoet/lib/EmailEditor/Engine/Renderer/BlocksRegistry.php
lysyjan: Thank you for pointing this out.
I will try to refactor this behavior for better extensibility, as you recommended.

mailpoet/assets/js/src/email-editor/engine/editor.tsx
costasovo: Yes sure. Could you please prepare the ticket? I think it would be nice to have this before we jump on the other blocks so that we have a place for adding hooks for other blocks.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
costasovo: Thank you for the explanation.

mailpoet/lib/EmailEditor/Engine/StylesController.php
costasovo: Ok. Let's continue with the units in value. I guess we will see soon enough what's working better for us.

lysyjan:
@costasovo I hope I addressed all your remarks. New changes are in the last four commits. Just so you know, I used interactive rebase because I added some descriptions to commit messages.

costasovo (approved):
Thank you for adding the changes!

PR #5263, opened 2023-11-02
Force @wordpress/rich-text to use version 6.5.0 [MAILPOET-5665]

johnolek:
## Description

This should address a regression where the paragraph toolbar in the form  editor (and maybe elsewhere) wasn't showing all of the expected options  and wasn't working correctly when setting the font family.

The bug was introduced after running `pnpm dedupe` after updating dependencies. I believe that changed the dependency of some packages like woocommerce/components to use @wordpress/rich-text version
 6.16.0 instead of 6.5.0.

My first attempt at a fix involved updating all `@wordpress/*` packages to use the version tagged `wp-6.3` (as discussed during a team meeting), and while that resolved this bug, it introduced many TS errors (see this build: https://app.circleci.com/pipelines/github/mailpoet/mailpoet/16096/workflows/5020cf66-7489-4dab-a11f-ebcef6b51733/jobs/272991), I think too many to handle as part of this ticket.

## Code review notes

I'm not sure why this update changed so many of the `react` lines in the lock file. Please let me know if you think this could cause an issue and if I should run the update commands differently.

## QA notes

If you know of any other places we have this editor toolbar other than the form editor, please test those places as well.

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5665

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/package.json
+++ b/mailpoet/package.json
@@ -64,7 +64,7 @@
     "@wordpress/notices": "^3.28.0",
     "@wordpress/plugins": "6.7.0",
     "@wordpress/preferences": "^3.5.0",
-    "@wordpress/rich-text": "^6.5.0",
+    "@wordpress/rich-text": "6.5.0",
     "@wordpress/url": "^3.29.0",
     "backbone": "1.3.3",
     "backbone.marionette": "3.2.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: false
+  excludeLinksFromLockfile: false
+
 overrides:
   react: 18.2.0
   react-dom: 18.2.0
@@ -153,7 +157,7 @@ importers:
         specifier: ^3.5.0
         version: 3.5.0(@types/react@18.0.28)(react-dom@18.2.0)(react@18.2.0)
       '@wordpress/rich-text':
-        specifier: ^6.5.0
+        specifier: 6.5.0
         version: 6.5.0(react@18.2.0)
       '@wordpress/url':
         specifier: ^3.29.0
@@ -530,7 +534,7 @@ importers:
         version: 5.1.0(stylelint@15.10.3)
       terser-webpack-plugin:
         specifier: ^5.3.9
-        version: 5.3.9(@swc/core@1.3.80)(webpack@5.88.2)
+        version: 5.3.9(webpack@5.88.2)
       ts-loader:
         specifier: ^9.4.4
         version: 9.4.4(typescript@5.0.2)(webpack@5.88.2)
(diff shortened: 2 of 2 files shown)


costasovo (commented):
@johnolek Great job debugging this issue! 👍 

There is another change in the pnpm-lock.yaml that caught my eye.
This addition:
```yaml
settings:
  autoInstallPeers: false
  excludeLinksFromLockfile: false
 ```
 
[autoInstallPeers is by default true ](https://pnpm.io/next/npmrc#auto-install-peers) and we switched it to false. The description from the doc (see the link) sounds dangerous. On the other hand if non-optional peer deps were really missing we would probably won't be able to compile the project 🤔 
I don't know if this is a problem or not. @johnolek Did you set it yourself? For the `excludeLinksFromLockfile` I wasn't even able to find documentation.

johnolek:
@costasovo Thanks for taking a look! I noticed that as well, and I should have mentioned that it's not an actual change, it's just moved from the bottom of the file to the top. Here's where it is currently: https://github.com/mailpoet/mailpoet/blob/8ccff6e3e659629e1d39caef64ec4eb90074b116/pnpm-lock.yaml#L22022-L22024

That being said, it's a relatively recent change. I think I introduced it in [this commit](https://github.com/mailpoet/mailpoet/commit/20b0cab1f0abe81331545a4d6e593e45c2caf734) when updating dependencies, so it's been around since August. I did not make the change intentionally or manually, it just happened when I was running installation commands. 

Perhaps it's coming from our .npmrc?: https://github.com/mailpoet/mailpoet/blob/8ccff6e3e659629e1d39caef64ec4eb90074b116/.npmrc#L1-L1

It looks like these settings were added in version [8.6.0](https://github.com/pnpm/pnpm/releases/tag/v8.6.0) of pnpm, released May 28th, 2023, so it makes sense that we only started seeing them recently.

Should we always run `pnpm` commands in the docker container to ensure we're all using the same version? I made these changes with version 8.10.1, and it looks like my docker container has version 8.6.12.

costasovo:
@johnolek Thanks for the explanation!

> I noticed that as well, and I should have mentioned that it's not an actual change, it's just moved from the bottom of the file to the top. 

👍 I totally missed that.

> It looks like these settings were added in version [8.6.0](https://github.com/pnpm/pnpm/releases/tag/v8.6.0) of pnpm, released May 28th, 2023, so it makes sense that we only started seeing them recently.

I see JJ added these settings when [updating to 8.5.1](https://github.com/mailpoet/mailpoet/commit/355cf5dc057bc118d294292c83d67f02be8ff978). So maybe the settings were older but started to be written in the lock file only recently. Anyway it seems that the setting comes from .npmrc and is not a new addition so I think we are fine.

> Should we always run pnpm commands in the docker container to ensure we're all using the same version? I made these changes with version 8.10.1, and it looks like my docker container has version 8.6.12.

I followed [JJ's instructions](https://mailpoetdev.wordpress.com/2022/07/26/mailpoet-adopts-pnpm/#what-to-do) and used `corepack` so now node should use the proper version of pnpm (when I run `pnpm --version` in MP directory I get 8.5.1 and out of the directory it is different). I thought you should get an error when you run it with a different version 🤔

costasovo (approved):

johnolek:
Thanks very much for pointing out those instructions, @costasovo! Somehow I never set up corepack correctly when I followed those instructions back in the day. I'm now (finally) using the correct version of pnpm.

Aschepikov (approved):
GJ!

PR #5268, opened 2023-11-03
Update react rendering method [MAILPOET-5120]

johnolek:
## Description

This PR converts all of our React Apps from using ReactDOM.render to createRoot. This switches the rendering version from React 17 to React 18. There are more details in the ticket.

## Code review notes

_N/A_

## QA notes

This PR affects many different parts of the plugin and will need to be tested carefully. We basically want to be sure that nothing has changed noticeably. I didn't see any issues when I tested locally but there are likely things that I missed.

You can see the different parts of the plugin that were affected by viewing the commit messages. There is a separate commit for every render method that was updated.

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5120

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/assets/js/src/automation/automation.tsx
+++ b/mailpoet/assets/js/src/automation/automation.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react';
-import ReactDOM from 'react-dom';
+import { createRoot } from 'react-dom/client';
 import { BrowserRouter } from 'react-router-dom';
 import { TopBarWithBeamer } from 'common/top-bar/top-bar';
 import { Popover, SlotFillProvider } from '@wordpress/components';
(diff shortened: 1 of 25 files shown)


JanJakes (approved):
Nice, thanks for working on this!
I had no idea we have so many React entrypoints 🙀

johnolek:
I was surprised by the number as well, @JanJakes! Does that seem like a problem to you at all, something where we might want to consider consolidating at some point? 

Thanks for the review!

Aschepikov (approved):
GJ!

JanJakes:
@johnolek Probably not a problem, but interesting to know. It relates to bundling, routing, etc., but I wonder how the setup will work with `@wordpress/scripts`, if we switch our Webpack config to that. The easiest would be to just write routes with dynamic imports and let the bundler sort it out, but that would probably be only something like Next.js.

PR #5274, opened 2023-11-08
Update UnsubscribeTokens to use dependency injection [MAILPOET-5710]

johnolek:
## Description

This PR updates the `UnsubscribeTokens` worker to use DI instead of retrieving services directly from the container. One of the services, `MailPoet\Util\Security`, is configured (by default) to be private and shouldn't be accessed this way. We are seeing this error: `UnsubscribeTokens: The "MailPoet\Util\Security" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.`

Currently I don't believe the UnsubscribeTokens worker is able to create tokens for users. I confirmed this by installing a fresh copy of mailpoet, observing the same error noted in the ticket, and seeing that the default subscriber doesn't have an unsubscribe token. I'm not sure precisely what effect this has. It didn't seem to affect my ability to unsubscribe using the unsubscribe link in a test email. It might also affect newsletters.

This task doesn't auto reschedule itself, but it does get scheduled when the populator runs (i.e. on plugin activation/upgrade), so any missing tokens should be filled in after the plugin updates.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5710]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5710]: https://mailpoet.atlassian.net/browse/MAILPOET-5710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/UnsubscribeTokens.php
+++ b/mailpoet/lib/Cron/Workers/UnsubscribeTokens.php
@@ -1,13 +1,13 @@
-<?php // phpcs:ignore SlevomatCodingStandard.TypeHints.DeclareStrictTypes.DeclareStrictTypesMissing
+<?php declare(strict_types = 1);
 
 namespace MailPoet\Cron\Workers;
 
-use MailPoet\DI\ContainerWrapper;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\InvalidStateException;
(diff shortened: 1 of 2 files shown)


rodrigoprimo (approved):

Aschepikov (approved):
GJ!

PR #5275, opened 2023-11-09
Automation template improvements [MAILPOET-4676]

JanJakes:
## Description

This PR extracts segment template components to more reusable ones and reuses them in automations.

## Code review notes

_N/A_

## QA notes

Please, apart from automation templates, test also segment templates. I refactored and reused the same components.
Please test both with and without Premium activated.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/818

## Linked tickets

[MAILPOET-4676]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4676]: https://mailpoet.atlassian.net/browse/MAILPOET-4676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
```diff
@@ -64,72 +55,63 @@ export function SegmentTemplates(): JSX.Element {
...
               )
               .map((template) => (
-                <TemplateListItem key={template.name} template={template} />
+                <Item
+                  name={template.name}
+                  description={template.description}
+                  category={getCategoryNameBySlug(template.category)}
+                  badge={template.isEssential ? 'essential' : undefined}
+                  onClick={() => void createFromTemplate(template)}
+                />
```
johnolek: On the segments template page I'm getting

```
On segments page:
Warning: Each child in a list should have a unique "key" prop.
Check the render method of `TabPanel`. See https://reactjs.org/link/warning-keys for more information.
...
```

I think these `Item`s need a `key` to prevent it.

johnolek (changes requested):
These changes look great to me overall!

One minor thing is I think you could use imports directly from `common/` instead of relative `../../common/`, but there's probably no need to change this.

A very minor design idea: for disabled cards it might be nice to change the `cursor`, maybe to `not-allowed` to really drive home that they can't be interacted with.

The final design/layout here is different than in the Figma linked in the ticket (e.g. the Figma has tags above the description and a totally different style for coming soon). Is there a newer version of the design that matches what's in this branch, or will there be a separate PR to update some of the other styles/layout? Or maybe it's that we're matching the segments templates styles instead of the styles in the linked figma?

mailpoet/assets/js/src/segments/dynamic/templates/index.tsx
JanJakes: Thanks, fixed in https://github.com/mailpoet/mailpoet/pull/5275/commits/84bc0e543c22cfae40433ef3188e37688a226bf8.

JanJakes (commented):
@johnolek Thanks for the review, and sorry for assigning it to you during your annual support rotation, I didn't notice it was your rotation week.

> One minor thing is I think you could use imports directly from common/ instead of relative ../../common/, but there's probably no need to change this.

This is what the IDE is autocompleting, and we have a mix of both. In the maintenance sheet we have a task to unify this somehow. I'm not sure if it's worth trying manually by then.

> A very minor design idea: for disabled cards it might be nice to change the cursor, maybe to not-allowed to really drive home that they can't be interacted with.

Although I like that, I'm not sure if we should do it, as it seems that WP components don't apply these cursor styles to disabled buttons. What do you think?

> The final design/layout here is different than in the Figma linked in the ticket...

Ahh, sorry, I should have mentioned that. The designs are outdated, the requirement is to follow segment templates as closely as possible. In the meantime, the designs were updated (see [here](https://www.figma.com/file/OMalS6vybv1zCw06RE1kVg/%5BNew%5D-MailPoet-Design-Library?type=design&node-id=31633-539&mode=design&t=qOiPzEK1Y1SJFCxL-0) and [here](https://www.figma.com/file/veTL847n7hs7kU8ZXcvVvD/MailPoet---Segments?type=design&node-id=1482-73744&mode=design&t=J5AAGiefXOv7T6wV-0)).

johnolek:
@JanJakes Thanks for posting the updated Figmas. I figured that must be the case. It might be nice to update the ticket and/or PR description with the new links so it's clear for QA what the desired state is.

> Although I like [the idea of a disabled cursor state], I'm not sure if we should do it, as it seems that WP components don't apply these cursor styles to disabled buttons. What do you think?

That's a good point. Do you think it would be worthwhile to open a PR/issue to update the WP components themselves?

> This is what the IDE is autocompleting, and we have a mix of both. In the maintenance sheet we have a task to unify this somehow. I'm not sure if it's worth trying manually by then.

Agreed, not a big deal. I would love to know how to get PhpStorm to autocomplete using the non-relative path myself.

johnolek (dismissed):

JanJakes:
@johnolek

> That's a good point. Do you think it would be worthwhile to open a PR/issue to update the WP components themselves?

I'm not sure if I can just do that, maybe it's better to ask design (so [I did](https://a8c.slack.com/archives/C040TSG288Y/p1700236137828679)). I also only now realized that we use cursor `pointer`, even when the cards are disabled. So [I fixed that](https://github.com/mailpoet/mailpoet/pull/5275/commits/7c8fc86967d5279c70bbc94d1cea1222cf23af5b).

> Agreed, not a big deal. I would love to know how to get PhpStorm to autocomplete using the non-relative path myself.

I think the only reliable way is to use path aliases, e.g., `@mailpoet/commons` or `@/commons`, etc. I guess we should look into that.

johnolek (approved):

Aschepikov:
DIscussed in slack (not an issue)

Aschepikov (approved):
GJ!

PR #5279, opened 2023-11-11
Add support for Core heading block - MAILPOET-5643

triple0t:
## Description

Add support for Core heading block rendering

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5643](https://mailpoet.atlassian.net/browse/MAILPOET-5643)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5643]: https://mailpoet.atlassian.net/browse/MAILPOET-5643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lysyjan:
@triple0t I have a few remarks/questions here:
- I have tested the text alignment (left, right, center), and it works in the editor but doesn't work in the renderer. Please, can you take a look?
- The heading toolbar contains some configuration, which we don't want to support (at least not in this phase, I guess), such as `Language` and `Inline image`. Could you take a look at how we can deactivate them?
- Because you changed a lot of logic here, maybe you should also add an integration test for this block. What do you think about that?

triple0t:
Thank you for the in-depth review, @lysyjan 

> I have tested the text alignment (left, right, center), and it works in the editor but doesn't work in the renderer. Please, can you take a look?

Done in https://github.com/mailpoet/mailpoet/pull/5279/commits/debfdc4348073a611406b3c40989a6d206c4d7fa

> The heading toolbar contains some configuration, which we don't want to support (at least not in this phase, I guess), such as Language and Inline image. Could you take a look at how we can deactivate them?

Done in https://github.com/mailpoet/mailpoet/pull/5279/commits/62b79818fb2533f705ca74dccbae19f976d81ed7

> Because you changed a lot of logic here, maybe you should also add an integration test for this block. What do you think about that?

Done in https://github.com/mailpoet/mailpoet/pull/5279/commits/d3d4eca6c68d54fc9fbe8f1aac0ad7834e78dbe1

lysyjan (approved):

PR #5283, opened 2023-11-13
Add email campaign name [MAILPOET-5646]

costasovo:
## Description

This PR adds campaign name input into the new email editor, and it also adds support for the campaign name in other parts of the plugin, where we refer to an email by name.

#### Campaign Name in the editor

Screenshot 2023-12-13 at 15 47 02

#### Campaign in other parts of the plugin

1. In emails listing, when an email has a campaign name we show the campaign name and a subject is displayed on the other line in smaller font.
Screenshot 2023-11-13 at 10 05 04

2. List of emails in dynamic segments UI. For segments where a user needs to pick segment we display only the campaign name if available.
3. The page title in email stats page – we display both name and subject.

## Code review notes

@JanJakes I chose to store the campaign name as a title in `wp_post` table. I think this makes sense for the future reusability of the editor. In order to support the campaign name in listings and other places in the plugin where we use Doctrine I added a minimalistic Doctrine entity for the WPPost. To support the entity I needed to do a few changes in the metadata factory for table prefixes. I put some logic directly to the factory class which makes this class aware that some tables are from WordPress. We may want to add an abstraction, but as this is a project (not a library) I thought it is ok.  I want to ask you for review of the related commits: c5c4e47f5094e8a02fcbb5b1585425a737c33844, fc2103661e2c83d4de7ab9ef9594629987c3ccf5 

## QA notes

1. Go to the experimental page and enable the new email editor (`wp-admin/admin.php?page=mailpoet-experimental`)
2. Create a standard newsletter in the new editor and set the campaign name and title
3. Verify that the campaign name is displayed in all places as expected
4. Are there some places where we should email name I missed?

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/820

## Linked tickets

[MAILPOET-5646]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5646]: https://mailpoet.atlassian.net/browse/MAILPOET-5646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/components/header/document-actions.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/header/document-actions.tsx
@@ -73,6 +73,7 @@ export function DocumentActions() {
               onChange={(newTitle) => {
                 setTitle(newTitle);
               }}
+              name="campaign_name"
               help={__(
                 `Name your email campaign to indicate its purpose. This would only be visible to you and not shown to your subscribers.`,
                 'mailpoet',
--- a/mailpoet/assets/js/src/email-editor/engine/components/header/document-actions.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/header/document-actions.tsx
@@ -1,0 +1,86 @@
+import { useRef } from '@wordpress/element';
+import {
+  Button,
+  Dropdown,
+  VisuallyHidden,
+  __experimentalText as Text,
+  TextControl,
+} from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { chevronDown } from '@wordpress/icons';
+import { useSelect } from '@wordpress/data';
+import { useEntityProp } from '@wordpress/core-data';
+import { storeName } from '../../store';
+
+// @see https://github.com/WordPress/gutenberg/blob/5e0ffdbc36cb2e967dfa6a6b812a10a2e56a598f/packages/edit-post/src/components/header/document-actions/index.js
+
+export function DocumentActions() {
+  const { showIconLabels } = useSelect(
+    (select) => ({
+      showIconLabels: select(storeName).isFeatureActive('showIconLabels'),
+      postId: select(storeName).getEmailPostId(),
+    }),
+    [],
+  );
+
+  const [emailTitle = '', setTitle] = useEntityProp(
+    'postType',
+    'mailpoet_email',
+    'title',
+  );
+
+  const titleRef = useRef(null);
+  return (
+    <div ref={titleRef} className="mailpoet-email-editor-document-actions">
+      <Dropdown
+        popoverProps={{
+          placement: 'bottom',
+          anchor: titleRef.current,
+        }}
+        contentClassName="mailpoet-email-editor-document-actions__dropdown"
+        renderToggle={({ isOpen, onToggle }) => (
+          <>
+            <Button
+              onClick={onToggle}
+              className="mailpoet-email-document-actions__link"
+            >
+              <Text size="body" as="h1">
+                <VisuallyHidden as="span">
+                  {__('Editing email:', 'mailpoet')}
+                </VisuallyHidden>
+                {emailTitle}
+              </Text>
+            </Button>
+            <Button
+              className="mailpoet-email-document-actions__toggle"
+              icon={chevronDown}
+              aria-expanded={isOpen}
+              aria-haspopup="true"
+              onClick={onToggle}
+              label={__('Change campaign name', 'mailpoet')}
+            >
+              {showIconLabels && __('Rename', 'mailpoet')}
+            </Button>
+          </>
+        )}
+        renderContent={() => (
+          <div className="mailpoet-email-editor-email-title-edit">
+            <div className="mailpoet-email-editor-dropdown-name-edit-title">
+              {__('Campaign name', 'mailpoet')}
+            </div>
+            <TextControl
+              value={emailTitle}
+              onChange={(newTitle) => {
+                setTitle(newTitle);
+              }}
+              help={__(
+                `Name your email campaign to indicate its purpose. This would only be visible to you and not shown to your subscribers.`,
+                'mailpoet',
+              )}
+            />
+          </div>
+        )}
+      />
+    </div>
+  );
+}
--- a/mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
+++ b/mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
@@ -26,7 +26,7 @@
 // Document actions - Popup for editing email/campaign title
 .mailpoet-email-editor-campaign-name__dropdown {
   .components-popover__content {
-    min-width: 240px;
+    min-width: 280px;
     padding: 0;
   }
 
--- a/mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
+++ b/mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
@@ -1,5 +1,5 @@
 // Document actions - Component in header for displaying email/campaign title edit popup
-.mailpoet-email-editor-document-actions {
+.mailpoet-email-editor-campaign-name {
   align-items: center;
   display: flex;
   flex-direction: row;
@@ -15,7 +15,7 @@
     padding: 0;
   }
 
-  .mailpoet-email-document-actions__link {
+  .mailpoet-email-campaign-name__link {
     display: inline-flex;
     height: fit-content;
     margin-right: 10px;
@@ -24,7 +24,7 @@
 }
 
 // Document actions - Popup for editing email/campaign title
-.mailpoet-email-editor-document-actions__dropdown {
+.mailpoet-email-editor-campaign-name__dropdown {
   .components-popover__content {
     min-width: 240px;
     padding: 0;
--- a/mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
+++ b/mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
@@ -33,8 +33,4 @@
   .mailpoet-email-editor-email-title-edit {
     padding: 16px;
   }
-
-  .mailpoet-email-editor-dropdown-name-edit-title {
-    margin-bottom: 10px;
-  }
 }
--- a/mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
+++ b/mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
@@ -1,0 +1,40 @@
+// Document actions - Component in header for displaying email/campaign title edit popup
+.mailpoet-email-editor-document-actions {
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  min-width: 0;
+
+  .components-dropdown {
+    display: inline-flex;
+  }
+
+  .components-button {
+    min-width: 0;
+    padding: 0;
+  }
+
+  .mailpoet-email-document-actions__link {
+    display: inline-flex;
+    height: fit-content;
+    margin-right: 10px;
+    margin-top: 10px;
+  }
+}
+
+// Document actions - Popup for editing email/campaign title
+.mailpoet-email-editor-document-actions__dropdown {
+  .components-popover__content {
+    min-width: 240px;
+    padding: 0;
+  }
+
+  .mailpoet-email-editor-email-title-edit {
+    padding: 16px;
+  }
+
+  .mailpoet-email-editor-dropdown-name-edit-title {
+    margin-bottom: 10px;
+  }
+}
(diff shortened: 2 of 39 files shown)


JanJakes:
> @JanJakes I chose to store the campaign name as a title in wp_post table. I think this makes sense for the future reusability of the editor. In order to support the campaign name in listings and other places in the plugin where we use Doctrine I added a minimalistic Doctrine entity for the WPPost. To support the entity I needed to do a few changes in the metadata factory for table prefixes. I put some logic directly to the factory class which makes this class aware that some tables are from WordPress. We may want to add an abstraction, but as this is a project (not a library) I thought it is ok.

@costasovo Interesting solution, I've never thought of that (adding entities for some or all WP tables), but it seems that so far, we haven't really had a strong need. Now, I wonder — should we start using them in our codebase? And if so, in which cases yes, and which cases no?

Maybe it comes down to why do you need the entity in this case, and, more broadly, what's the plan for storing newsletters in the future. Currently, old emails have the `NewsletterEntity`, while emails for the new editor are saved in WP posts, I suppose. But why do we need to join them? Do you also create a `NewsletterEntity` record for each email that is stored in WP posts?

Without a deeper understanding, I tend to think it would be better to avoid having a Doctrine entity for WP posts, so we don't use it in more and more places, and rather stick to WP-native functionality. For instance, to fetch multiple post titles, could we just use `get_posts` or `WP_Query` with some arguments?

> I want to ask you for review of the related commits: https://github.com/mailpoet/mailpoet/commit/c5c4e47f5094e8a02fcbb5b1585425a737c33844, https://github.com/mailpoet/mailpoet/commit/fc2103661e2c83d4de7ab9ef9594629987c3ccf5

Should I only check the WP post entity part, or do the full PR review?

costasovo:
@JanJakes Thank you for all these questions. They helped to challenge this idea. Maybe there could be a better approach...

> @costasovo Interesting solution, I've never thought of that (adding entities for some or all WP tables), but it seems that so far, we haven't really had a strong need. 

The need comes from the idea of storing the campaign name as the title of the post in combination with the fact that we want to present the campaign name as the name of an email (replacing the subject) in a couple of places in the plugin. Now why do I want to store the campaign name as a title? 
The editor will be used in more projects, and I want to provide a way to set the title of the email document, and saving in the post's title column feels like a natural way of doing this. In the editor, there is no problem with this. The main issue comes from the listing repository, where we build on top of the doctrine's DQL, and we need to cover sorting and searching. Without an entity, I can't refer to/fetch the value using DQL.

> Now, I wonder — should we start using them in our codebase? And if so, in which cases yes, and which cases no?

I would avoid using them as much as possible and limit the usage only to reading the data. Usage for updates would cause skipping various WP hooks. Could mark the entity as [Read Only be sufficient](https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/improving-performance.html#read-only-entities)?

> Maybe it comes down to why do you need the entity in this case, and, more broadly, what's the plan for storing newsletters in the future. Currently, old emails have the NewsletterEntity, while emails for the new editor are saved in WP posts, I suppose. 

I'm trying to distinguish between common email data vs MailPoet specific email data. I would like to store the common data (content, title, subject?, preheader?) in the wp posts infrastructure so that the editor can be used without MailPoet tables but still want to use MailPoet tables for MailPoet-specific stuff and also for email processing logic (sending queue, scheduled tasks, etc.). 

> But why do we need to join them? Do you also create a NewsletterEntity record for each email that is stored in WP posts?

I'm joining them (in the sense of linking them with a foreign key) because of the sending processing as it works with the NewsletterEtity, yes I create a NewsletterEntity for every post that has mailpoet_email type. As for joining on the query level, I used that in the listing query and one other query used for fetching lists of emails for the dynamic segment builder.

> Without a deeper understanding, I tend to think it would be better to avoid having a Doctrine entity for WP posts, so we don't use it in more and more places, and rather stick to WP-native functionality. For instance, to fetch multiple post titles, could we just use get_posts or WP_Query with some arguments?

I was trying to hold to the approach of relying only on WP-native functionality, but then I ran into the issue with the listing repository. So I was thinking we need an exception for this case. I agree that we should rather avoid doing this. This makes me think that we could make an exception the other way around. I mean, we could add a column for the campaign name to the newsletters table and sync the title into it (e.g. in a post's update hook). This would also help with fetching the names for lists and might be less "dangerous" than introducing an entity for a WP post. @JanJakes , @lysyjan What do you think about the alternative?

lysyjan:
> I would avoid using them as much as possible and limit the usage only to reading the data. Usage for updates would cause skipping various WP hooks. Could mark the entity as [Read Only be sufficient](https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/improving-performance.html#read-only-entities)?

This approach sounds good to me, but if we want to avoid having entities for WP posts, we can choose the approach with a copied campaign name in our newsletter table. I am not a fan of duplicated data, but I can understand that if needed.

JanJakes:
@costasovo Thanks for the details.

> still want to use MailPoet tables for MailPoet-specific stuff and also for email processing logic (sending queue, scheduled tasks, etc.).

I understand we'll need that now, so we don't have to rewrite all the existing logic, but in the final state, shouldn't the `NewsletterEntity` no longer exist? I don't know what data you consider to be MailPoet-specific, but WP posts allow you to save pretty much anything as post metadata, and things like tasks and queues are not stored in the `NewsletterEntity` (in other words, they can reference the WP post ID directly.)

In any case, in the meantime, I think it's fine to use the entity and mark it readonly, if that makes the code significantly simpler. Rather than a query hint, we can [annotate that way the entity itself](https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/attributes-reference.html#entity). I still don't fully understand why it's really required, as using `get_posts` or `WP_Query` for the titles seems to be simple as well (it's not a big deal to have two queries for the listing data, I suppose), but ultimately, I'm fine with any of these solutions.

Last food for thought — WP posts can be extended by 3rd party plugins for fine-grained access control, etc. If we want to preserve that functionality, we'd need to stick to `get_posts` or `WP_Query` to avoid bypassing 3rd party hooks. In that case, you can do one WP query to fetch all IDs, titles, timestamps, and such, and then JOIN our entity on that.

costasovo:
@lysyjan I discussed the issue with @JanJakes, and we will go with the Read-only entity to avoid duplicate data.

mailpoet/assets/js/src/email-editor/engine/components/header/document-actions.tsx
```diff
@@ -0,0 +1,86 @@
...
+} from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { chevronDown } from '@wordpress/icons';
+import { useSelect } from '@wordpress/data';
+import { useEntityProp } from '@wordpress/core-data';
+import { storeName } from '../../store';
+
+// @see https://github.com/WordPress/gutenberg/blob/5e0ffdbc36cb2e967dfa6a6b812a10a2e56a598f/packages/edit-post/src/components/header/document-actions/index.js
+
+export function DocumentActions() {
```
lysyjan: The name of this component is a bit confusing to me. Why did you choose it? I'm expecting it's based on the Gutenberg component but I think we could use something more related to the functionality of the component.

mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
```diff
@@ -0,0 +1,40 @@
...
+    height: fit-content;
+    margin-right: 10px;
+    margin-top: 10px;
+  }
+}
+
+// Document actions - Popup for editing email/campaign title
+.mailpoet-email-editor-document-actions__dropdown {
+  .components-popover__content {
+    min-width: 240px;
```
lysyjan: It's only a detail, but I have checked the Figma design, and it contains a higher width. Do you have some reason for using 240px instead of 280px?

mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
```diff
@@ -0,0 +1,40 @@
...
+    padding: 0;
+  }
+
+  .mailpoet-email-editor-email-title-edit {
+    padding: 16px;
+  }
+
+  .mailpoet-email-editor-dropdown-name-edit-title {
```
lysyjan: From what I could see in the Figma, the title should be upper case and bold. Nevertheless, I checked the Automation, and there is a similar component, so I am not sure about that. In my opinion, at least a bold font would be fine here. What do you think?

mailpoet/lib/AdminPages/Pages/DynamicSegments.php
```diff
@@ -186,6 +186,7 @@ private function getNewslettersList(): array {
       $result[] = [
         'id' => (string)$newsletter->getId(),
         'subject' => $newsletter->getSubject(),
+        'name' => $newsletter->getPresentableName(),
```
lysyjan: NewsletterResponseBuilder uses `campaign_name` and here is `name`. Is there a reason for that? I think we should unify the used name for all places. It can be confusing to mix those two.

lysyjan (commented):
I'm sorry for the late review. Both of us have different duties in the last weeks.
I know how complicated it was to solve the problem with WP posts.
I have only a few minor remarks, I guess.

mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
costasovo: I copied the style from automations. 280px looks nicer. I'll update it.

mailpoet/assets/js/src/email-editor/engine/components/header/index.scss
costasovo: Nice catch! It doesn't even have to be separately. Moving the text into the TextControl label will make it look as in the Figma design.

costasovo:
@lysyjan I've rebased the branch and addressed all the comments in the new commits.

lysyjan (dismissed):
👍

Aschepikov:
Hi, @costasovo I found some issues with the gutenberg editor, some of them are on the prod maybe after the refactoring, Pls check it and tell me what can be fixed here and for what we need a separate ticket.

1. The sender is always empty,  there is no auto-filling like in the current newsletter editor. It is present on the prod too.
image

2. We are implementing the campaign name and the logic is that there is a campaign name and underneath there is a subject text. If I am creating a campaign name on the email editor step
image
So in the next step, I will see the send page 
image
Without a campaign name but with the subject and preview text
image
With the filled campaign name, subject, and preview I will see this name in the list with the default campaign name (the new one wasn't saved) and the subject so we need to add the campaign name on the send page with the subject and without preview text.
image

So the campaign name is not saving it looks like there is no autosaving at all, but last time I checked it was ok. And the campaign name should be on the send page too.

3. The sending of the Gutenberg emails isn't working, I am checking the last versions to understand where it was implemented, the last time sending was ok
image
image
Can it be connected that the email wasn't saved and is empty?

Aschepikov:
For everything else it is possible to check when the campaign name will be saved and the email will be sent

costasovo:
@Aschepikov Thank you for the thorough review 👍 
ad 1) This seems like a bug. Probably not related to the PR but I'll look into it.
ad 2) We will get rid of the send page in later phases of the new email editor, but I see it could be confusing. I suggest we add the campaign name to the send page as a no-editable heading. That way it would be present in all steps. What do you think? @pavel-mailpoet Any other ideas?
ad 3) I'll look into it.

pavel-mailpoet:
> ad 2) We will get rid of the send page in later phases of the new email editor, but I see it could be confusing. I suggest we add the campaign name to the send page as a no-editable heading. That way it would be present in all steps. What do you think? @pavel-mailpoet Any other ideas?

Hmm, we will remove the send page. However, we want to release the editor as a beta with the send page. So some users will still using it for a bit and so we should make it usable. The solution you suggested sounds easy enough to do  and could improve this feature for the users.

costasovo:
@lysyjan @Aschepikov I added five commits addressing the issues.
ad 1) Fixed but only for newly created emails. We are still behind a feature flag so I doesn't make sense adding a migration.
ad 2) I added a heading at the top of the send page. The heading is displayed only if the email has a campaign name.
ad 3) The problem with sending is related to the empty email body. I added a check that send button is disabled in case the **saved body** is empty. Please note this is probably a temporary fix because we plan to add autosave and also email content validation when user clicks on the send button.

lysyjan (approved):

Aschepikov:
HI @costasovo thank you for the fixes. Now when I am duplicating the email there are creating two copies of it, one with the copy of campaign name and one with the copy of campaign name and subject 
image

pavel-mailpoet:
@Aschepikov I don't think this problem is related to this branch. We have a ticket for it already: https://mailpoet.atlassian.net/browse/MAILPOET-5810

Aschepikov (approved):
GJ!

PR #5289, opened 2023-11-16
Feat/create mailpoet marketing channel - MAILPOET-5692

triple0t:
## Description

Add base support for WooCommerce Multichannel integration.

## Code review notes

Please check commits

## QA notes

1. Ensure **WooCommerce** is installed and **activated**
2. Enable `mailpoet_woocommerce_multichannel_integration` at `/wp-admin/admin.php?page=mailpoet-experimental`

3. Visit the WooCommerce Marketing tab at `/wp-admin/admin.php?page=wc-admin&path=%2Fmarketing`
4. You should see **MailPoet** on the list of Channels

**Important Note**: 
We haven’t pushed the icon to the WP.org repository.
If you can't see the channel Icon, kindly add `define( 'MAILPOET_USE_CDN', false );` to your `wp-config.php` file.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5692](https://mailpoet.atlassian.net/browse/MAILPOET-5692)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5692]: https://mailpoet.atlassian.net/browse/MAILPOET-5692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
```diff
@@ -0,0 +1,101 @@
...
+    return __( 'MailPoet', 'mailpoet' );
+  }
+
+  /**
+   * Returns the description of the marketing channel.
+   *
+   * @return string
+   */
+  public function get_description(): string { // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    return __( 'Create and send newsletters, post notifications and welcome emails from your WordPress.', 'mailpoet' );
```
johnolek, line 46: Is there a benefit to hardcoding this string instead of dynamically fetching the plugin description? I think that's what WooCommerce is doing for displaying the description of installed extensions here: https://github.com/woocommerce/woocommerce/blob/6f6b5f5b4aa6fb636ad75b7f9853a1644b66fdcc/plugins/woocommerce/src/Admin/Marketing/InstalledExtensions.php#L616-L631

mailpoet/tasks/phpstan/custom-stubs.php
```diff
@@ -37,6 +37,64 @@ function change_feature_enable(string $feature_id, bool $enable) {}
   }
 }
 
+// Temporary stubs for marketing channels
+// I have no idea why PhpStorm is complaining, but the woocommerce-stubs library has the latest updates
```
johnolek, line 88: What issue are you seeing with PhpStorm if you don't have these stubs? I would be happy to help troubleshoot, but when I removed these stubs it didn't seem to cause any issues in my PhpStorm.

mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
```diff
@@ -41,8 +52,7 @@ public function get_description(): string { // phpcs:ignore PSR1.Methods.CamelCa
    * @return string
    */
   public function get_icon_url(): string { // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
-    // TODO: use the correct image
-    return 'https://ps.w.org/mailpoet/assets/icon-256x256.png';
+    return $this->cdnAssetUrl->generateCdnUrl('icon-white-123x128.png');
```
johnolek, line 55: This comment is about MPMarketingChannel itself. It seems kind of strange to me that it requires a URL while the installed extensions section can use an SVG. Any idea if we could somehow re-use that SVG instead?

One thing that's a little strange to me is that the design is different for the SVG and the icon, so the logo looks different when MailPoet shows up as a channel vs. an installed extension.

image
image

johnolek (approved):
A few questions in the comments but nothing that needs to be changed.

I have a couple of general questions. It seems that when I enable this feature, all of these default Channels disappear and I only see MailPoet, is that expected?

image

I also noticed that MailPoet disappears from the Installed Extensions section when this feature, which makes sense to me since otherwise there would be duplication, but I wanted to double check that that's expected.

mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
triple0t: To be honest, I'm not sure if the information here should be the same as the information on the plugin description. I'm using the hardcoded string so we can easily change it later.

mailpoet/tasks/phpstan/custom-stubs.php
triple0t: I get this error `Undefined class 'MarketingChannelInterface' `, etc.

Note the yellow lines

> I would be happy to help troubleshoot

Thank you. I'll send you a DM

mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
triple0t: > This comment is about MPMarketingChannel itself. It seems kind of strange to me that it requires a URL while the installed extensions section can use an SVG. Any idea if we could somehow re-use that SVG instead?

It requires a [URL](https://github.com/woocommerce/woocommerce/blob/6fb787e954af29605d793c6cbb35c558d6e1d737/plugins/woocommerce/src/Admin/Marketing/MarketingChannelInterface.php#L42-L47C41), unfortunately.

I tested with an SVG, but it refused to load the image.

> One thing that's a little strange to me is that the design is different for the SVG and the icon, so the logo looks different when MailPoet shows up as a channel vs. an installed extension.

I noticed this as well, but since the design logo looks closer to our normal MP logo, I decided to go with that.

e.g.



From: footer of https://www.mailpoet.com/

triple0t:
Thank you for the in-depth review, @johnolek 

> It seems that when I enable this feature, all of these default Channels disappear and I only see MailPoet, is that expected?

Yes.

It's hidden here 

You can see the other suggestions when you click on the "Add channels" text with icon


> I also noticed that MailPoet disappears from the Installed Extensions section when this feature, which makes sense to me since otherwise there would be duplication, but I wanted to double check that that's expected.

Yes, this is expected.

I addressed your questions in the comments. Do let me know if you have any more follow-up questions.

mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
johnolek: Makes sense!

johnolek:
Thanks for the replies, @triple0t, this all looks good to me!

Aschepikov (approved):
GJ!

PR #5291, opened 2023-11-17
Feat/add setup information for the mailpoet marketing channel - MAILPOET-5695

triple0t:
## Description

Add setup information for the MailPoet Marketing Channel

This PR handles the implementation of `is_setup_completed` and `get_setup_url`

## Code review notes

Please check commits

## QA notes

Start with the QA notes here: https://github.com/mailpoet/mailpoet/pull/5289

* You might want to set `version` on the `wp_mailpoet_settings` table to `null`
* Visit the WooCommerce Marketing tab at `/wp-admin/admin.php?page=wc-admin&path=%2Fmarketing`
* Click on "Finish Setup"

* Complete the welcome wizard

## Linked PRs

Child of https://github.com/mailpoet/mailpoet/pull/5289

## Linked tickets

[MAILPOET-5695](https://mailpoet.atlassian.net/browse/MAILPOET-5695)

## After-merge notes

Please, **only** merge after https://github.com/mailpoet/mailpoet/pull/5289 

## Tasks

[MAILPOET-5695]: https://mailpoet.atlassian.net/browse/MAILPOET-5695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
+++ b/mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
@@ -6,6 +6,7 @@ use Automattic\WooCommerce\Admin\Marketing\MarketingCampaign;
 use Automattic\WooCommerce\Admin\Marketing\MarketingCampaignType;
 use Automattic\WooCommerce\Admin\Marketing\MarketingChannelInterface;
 use MailPoet\Config\Menu;
+use MailPoet\Settings\SettingsController;
 use MailPoet\Util\CdnAssetUrl;
 
 class MPMarketingChannel implements MarketingChannelInterface {
@@ -13,10 +14,17 @@ class MPMarketingChannel implements MarketingChannelInterface {
   /** @var CdnAssetUrl */
   private $cdnAssetUrl;
 
+  /**
+   * @var SettingsController
+   */
+  private $settings;
+
   public function __construct(
-    CdnAssetUrl $cdnAssetUrl
+    CdnAssetUrl $cdnAssetUrl,
+    SettingsController $settings
   ) {
     $this->cdnAssetUrl = $cdnAssetUrl;
+    $this->settings = $settings;
   }
 
   /**
@@ -61,7 +69,7 @@ class MPMarketingChannel implements MarketingChannelInterface {
    * @return bool
    */
   public function is_setup_completed(): bool { // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
-    return true; // will be updated in MAILPOET-5695
+    return $this->isMPSetupComplete();
   }
 
   /**
@@ -70,7 +78,11 @@ class MPMarketingChannel implements MarketingChannelInterface {
    * @return string
    */
   public function get_setup_url(): string { // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
-    return admin_url('admin.php?page=' . Menu::MAIN_PAGE_SLUG); // will be updated in MAILPOET-5695
+    if ($this->isMPSetupComplete()) {
+      return admin_url('admin.php?page=' . Menu::MAIN_PAGE_SLUG);
+    }
+
+    return admin_url('admin.php?page=' . Menu::WELCOME_WIZARD_PAGE_SLUG . '&mailpoet_wizard_loaded_via_woocommerce_marketing_dashboard');
   }
 
   /**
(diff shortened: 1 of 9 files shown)


mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
```diff
@@ -108,4 +120,15 @@ public function get_supported_campaign_types(): array { // phpcs:ignore PSR1.Met
   public function get_campaigns(): array { // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
     return []; // will be updated in MAILPOET-5698
   }
+
+  /**
+   * Whether the task is completed.
+   * If the setting 'version' is not null it means the welcome wizard
+   * was already completed so we mark this task as completed as well.
+   */
+  protected function isMPSetupComplete(): bool {
```
rodrigoprimo, line 129: I wonder if we should reuse (or create if one doesn't exist already) a more generic method to check if the MP setup is completed? My concern here is that we probably need to check this in other places and things might break if, in the future, we change the logic of the setup and use a criteria other than whether or not `version` exists to determine if the setup is completed or not. What do you think?

rodrigoprimo (commented):
Good job! Overall, it looks good to me. I have two comments. One below and another one as an inline comment.

> Please, only merge after https://github.com/mailpoet/mailpoet/pull/5289

I'm not sure if I'm missing something, but this this PR was created against `feat/create-mailpoet-marketing-channel` and not `trunk`, maybe we need to merge it before #5289 is merged? Or was the intention to create it against `trunk`?

mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
triple0t: Thank you for the suggestion, and I agree with you.
For the two places ([1](https://github.com/mailpoet/mailpoet/blob/1a0d3ece134e2a4079a9842147befff253524bb7/mailpoet/lib/WooCommerce/MailPoetTask.php), [2](https://github.com/mailpoet/mailpoet/blob/1a0d3ece134e2a4079a9842147befff253524bb7/mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php)) we need to run these checks, the parent classes do not pass through the MailPoet DI.
How would you suggest we go about implementing it?

triple0t:
Thank you for the review, @rodrigoprimo 

> I'm not sure if I'm missing something, but this this PR was created against feat/create-mailpoet-marketing-channel and not trunk, maybe we need to merge it before https://github.com/mailpoet/mailpoet/pull/5289 is merged? Or was the intention to create it against trunk?

I created the PR against `feat/create-mailpoet-marketing-channel` to assist with reviewing.

Technically, we will start with merging `feat/create-mailpoet-marketing-channel` into `trunk`
Rebase this PR on the trunk branch and merge in.

The PR base would change automatically from `feat/create-mailpoet-marketing-channel` to `trunk` once the initial PR is merged.

mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
rodrigoprimo: Good point. In this case, maybe the best approach is to use `ContainerWrapper::getInstance()->get()`? What do you think?

rodrigoprimo:
> The PR base would change automatically from feat/create-mailpoet-marketing-channel to trunk once the initial PR is merged.

Thanks for the clarification. I was not aware that GitHub would change the base branch automatically to `trunk` once the initial PR is merged. That is good to know.

mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
triple0t: The scope of this might be a bit larger than I expected. I've created another ticket to handle the implementation: https://mailpoet.atlassian.net/browse/MAILPOET-5736.

rodrigoprimo: Sounds good to me! Thanks!

Is there anything else you would like to do before a final review? Asking as the PR is still assigned to you.

I will be AFK tomorrow, so I will leave the PR approved in case you want to move it to QA tomorrow. Have to further check anything when I'm back.

rodrigoprimo (approved):

mailpoet/lib/WooCommerce/MultichannelMarketing/MPMarketingChannel.php
triple0t: > Is there anything else you would like to do before a final review? Asking as the PR is still assigned to you.

Nope.

> I will be AFK tomorrow, so I will leave the PR approved in case you want to move it to QA tomorrow. Have to further check anything when I'm back.

Thank you, @rodrigoprimo 

I'll move the PR to QA.

Aschepikov (approved):
GJ!

PR #5294, opened 2023-11-18
Check if post is set when setting the title

brezocordero:
## Description

We got a report of the following warning:
`PHP Warning:  Attempt to read property "post_title" on null in .../plugins/mailpoet/lib/Subscription/Pages.php on line 287`

I have not been able to reproduce it but the added check should remove the warning.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5716]

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Subscription/Pages.php
+++ b/mailpoet/lib/Subscription/Pages.php
@@ -281,6 +281,8 @@ class Pages {
     global $post;
 
     if (
+      (!isset($post))
+      ||
       ($post->post_title !== __('MailPoet Page', 'mailpoet')) // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
       ||
       ($pageTitle !== $this->wp->singlePostTitle('', false))


rodrigoprimo:
I have rebased this branch with trunk to hopefully make the CircleCI build pass.

mailpoet/lib/Subscription/Pages.php
```diff
@@ -281,6 +281,8 @@ public function setPageTitle($pageTitle = '') {
     global $post;
 
     if (
+      (!isset($post))
```
rodrigoprimo, line 284: I'm not entirely sure what this method does, so I could be missing something, but is there a scenario where `$post` is not set, but `$pageTitle` is equal to `$this->wp->singlePostTitle('', false)`? If so, I believe this change will alter the behavior of the method. Again, I don't know if what I'm asking could happen or not, but I'm asking just to be sure.

rodrigoprimo (commented):
I like the suggested approach to protect against the PHP warning. I left a comment as I'm not sure if the current implementation could change the behavior of the method.

mailpoet/lib/Subscription/Pages.php
brezocordero: If `$post` is not set, the title check would be true, it would stop at `($test->post_title !== __('MailPoet Page', 'mailpoet')`. https://3v4l.org/FMIld
and it would not reach `($pageTitle !== $this->wp->singlePostTitle('', false)`
Or is there a way to reach that last condition when `$post` is not set?

rodrigoprimo: You are right. I missed that the title check evaluates to true when `$post` is not set.

rodrigoprimo (approved):

Aschepikov (approved):
GJ!

PR #5295, opened 2023-11-18
Use different method to hide menus

brezocordero:
## Description

`add_submenu_page` with empty `parent_slug` generates a deprecation error.
This PR uses a non-existing parent slug instead of an empty string.

## Code review notes

We could also wait for the trac ticket but not sure if it will be implemented or when.

## QA notes

2 ways to test it:
- On Jurassic.ninja, click on 'Select More Options', select 'PHP 8.1', WooCommerce, and create a new site.
- Install plugin Query Monitor.
- Upload the build
- Check there are no errors in Query Monitor when you go to MailPoet landing page

And locally:
- On your local site, if running PHP 8.1, you can test there is no log by going to:
- ?page=mailpoet-welcome-wizard
- ?page=mailpoet-landingpage
- ?page=mailpoet-woocommerce-setup

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5718]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5718]: https://mailpoet.atlassian.net/browse/MAILPOET-5718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Menu.php
+++ b/mailpoet/lib/Config/Menu.php
@@ -33,6 +33,7 @@ use MailPoet\WP\Functions as WPFunctions;
 
 class Menu {
   const MAIN_PAGE_SLUG = self::HOMEPAGE_PAGE_SLUG;
+  const NO_PARENT_PAGE_SLUG = 'mailpoet-no-parent';
 
   const EMAILS_PAGE_SLUG = 'mailpoet-newsletters';
   const FORMS_PAGE_SLUG = 'mailpoet-forms';
@@ -190,7 +191,7 @@ class Menu {
 
     // Welcome wizard page
     $this->wp->addSubmenuPage(
-      '',
+      self::NO_PARENT_PAGE_SLUG,
       $this->setPageTitle(__('Welcome Wizard', 'mailpoet')),
       esc_html__('Welcome Wizard', 'mailpoet'),
       AccessControl::PERMISSION_ACCESS_PLUGIN_ADMIN,
@@ -203,7 +204,7 @@ class Menu {
 
     // Landingpage
     $this->wp->addSubmenuPage(
-      '',
+      self::NO_PARENT_PAGE_SLUG,
       $this->setPageTitle(__('MailPoet', 'mailpoet')),
       esc_html__('MailPoet', 'mailpoet'),
       AccessControl::PERMISSION_ACCESS_PLUGIN_ADMIN,
@@ -487,7 +488,7 @@ class Menu {
 
     // WooCommerce Setup
     $this->wp->addSubmenuPage(
-      '',
+      self::NO_PARENT_PAGE_SLUG,
       $this->setPageTitle(__('WooCommerce Setup', 'mailpoet')),
       esc_html__('WooCommerce Setup', 'mailpoet'),
       AccessControl::PERMISSION_ACCESS_PLUGIN_ADMIN,
@@ -527,7 +528,7 @@ class Menu {
     if (
       !$this->wp->applyFilters('mailpoet_show_automations', $showAutomations)
     ) {
-      $parentSlug = '';
+      $parentSlug = self::NO_PARENT_PAGE_SLUG;
     }
 
     $automationPage = $this->wp->addSubmenuPage(
@@ -723,7 +724,7 @@ class Menu {
       }
     }
 
-    if ($parentSlug) {
+    if ($parentSlug && $parentSlug !== self::NO_PARENT_PAGE_SLUG) {
       // highlight parent submenu item
       $plugin_page = $parentSlug;
     } else {
@@ -783,7 +784,7 @@ class Menu {
       return false;
     }
     WPFunctions::get()->addSubmenuPage(
-      '',
+      self::NO_PARENT_PAGE_SLUG,
       'MailPoet',
       'MailPoet',
       AccessControl::PERMISSION_ACCESS_PLUGIN_ADMIN,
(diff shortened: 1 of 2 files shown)


mailpoet/lib/Config/Menu.php
```diff
@@ -213,8 +221,58 @@ private function registerMailPoetMenu() {
         'landingPage',
       ]
     );
+    $this->hideSubmenuPage(self::MAIN_PAGE_SLUG, self::LANDINGPAGE_PAGE_SLUG);
+  }
 
-    $this->registerMailPoetSubMenuEntries();
+  /**
+   * Hide the submenu page based on slug and return the item that was hidden.
```
JanJakes: The method doesn't seem to be returning anything, even though it is mentioned here (and again a few lines below).

mailpoet/lib/Config/Menu.php
```diff
@@ -213,8 +221,58 @@ private function registerMailPoetMenu() {
...
+   *
+   * Borrowed from Jetpack's Base_Admin_Menu class.
+   *
+   * Instead of actually removing the submenu item, a safer approach is to hide it and filter it in the API response.
+   * In this manner we'll avoid breaking third-party plugins depending on items that no longer exist.
```
JanJakes: From this description, it seems that this is a workaround for calling `remove_submenu_page`. Since we're not hiding any WP or 3rd party pages, but only our own, and we're not doing that conditionally, but always, could we simply call `remove_submenu_page` at a correct moment?

mailpoet/lib/Config/Menu.php
```diff
@@ -213,8 +221,58 @@ private function registerMailPoetMenu() {
...
+   *
+   * A false|array value is returned to be consistent with remove_submenu_page() function
+   *
+   * @param string $menu_slug The parent menu slug.
+   * @param string $submenu_slug The submenu slug that should be hidden.
+   * @return void
+   */
+  public function hideSubmenuPage(string $menu_slug, string $submenu_slug): void {
```
JanJakes: The naming of the variables (`$menu_slug`, etc.) doesn't follow our coding standards, but, interestingly, the lint passes? 🤔

mailpoet/lib/Config/Menu.php
```diff
@@ -61,6 +61,7 @@ class Menu {
...
+  const HIDE_CSS_CLASS = 'hide-if-js';
```
JanJakes: What does the class name mean? Is it a native class from WP?

JanJakes (commented):
It took me some time to realize what is the issue here, but I think I get it now. We're talking only about pages, that should have no parent, and that's only a few pages now, is that correct?

What if we added them under a menu item, e.g. the MailPoet homepage? It would highlight the homepage in the menu, though, not sure if that's desired.

Alternatively, what if we use `unknown-parent-slug` or something like that? Would that fail with a warning as well?

I'm adding other remarks and questions in the code.

mailpoet/lib/Config/Menu.php
brezocordero: Thanks for noticing. Changed.

mailpoet/lib/Config/Menu.php
brezocordero: Modified

mailpoet/lib/Config/Menu.php
brezocordero: With `remove_submenu_page` you can not access the page but we still want the pages to be available.

mailpoet/lib/Config/Menu.php
brezocordero: Yes,[ it is in core](https://github.com/WordPress/wordpress-develop/blob/30954f7ac0840cfdad464928021d7f380940c347/src/wp-admin/css/common.css#L176). It is used by [WPCOM API to hide the menus from the JS navigation](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-admin-menu.php#L237).
It is also [the approach that is used in Woo Express to hide the menus](https://github.com/Automattic/wc-calypso-bridge/blob/20af7c76771c7bbe70b825e17a25a46c2d5d3111/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php).

brezocordero:
@JanJakes I have fixed the variable names and the return of the function
I have explained a bit more about this approach in the comments.
To sum up: 
We hide the menus with CSS using a specific class `.hide-if-js` that is present in WP core. 
The class is also used to filter which menus are displayed on the WPCOM API/JS navigation.
The pages can still be accessed using the direct link.

JanJakes:
@brezocordero Thanks, for the context, I think I understand what's happening now. The issue seems to occur only when you access one of the hidden pages, not when the menu is rendered. And it seems to me that the empty string really is the only problem, so I wonder, why not keeping the current method, and using just a simple fix such as [this](https://github.com/mailpoet/mailpoet/commit/2a938fcf4c22efdd1b9563a4d626336fa86a945a)?

What confuses me in the PR is that the current hiding method is still somehow kept — the `unset($submenu[self::MAIN_PAGE_SLUG]);` and `$plugin_page = self::MAIN_PAGE_SLUG;` (actually even added to one more place) — while we also add hiding with CSS. The code then becomes pretty hard to understand (and it seems we don't need the CSS workaround at all).

As I was a little bit struggling with understanding and replicating the issue, I'd also suggest extending the QA notes with something like this:

> Test this under PHP 8.1 or higher (the error won't be visible on Jurassic Ninja, which runs PHP 8.0).
> Check that this PR fixes the error on the following pages:
>   1. `?page=mailpoet-welcome-wizard`
>   2. `?page=mailpoet-landingpage`
>   3. `?page=mailpoet-woocommerce-setup`
>   4. MailPoet error page — there's no easy way to test this, so no need.

brezocordero:
> why not keeping the current method, and using just a simple fix such as [this](https://github.com/mailpoet/mailpoet/commit/2a938fcf4c22efdd1b9563a4d626336fa86a945a)?

Ok, makes sense. Thanks a lot, I also like it better like this.
I have taken your commit, added a bit that was missing and updated the test.
Updated the PR description and the QA instructions.

JanJakes (approved):
@brezocordero Thanks! It's good to know about the CSS solution, in case we'd run into any issues with our current solution again.

Aschepikov (approved):
GJ!

PR #5308, opened 2023-11-24
Fix send page layout [MAILPOET-5734]

NeosinneR:
## Description

_N/A_

## Code review notes

I decided not to use "mailpoet-main-container" because it's too wide, for two 368px columns.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5734]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5734]: https://mailpoet.atlassian.net/browse/MAILPOET-5734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-plugin/_forms.scss
+++ b/mailpoet/assets/css/src/components-plugin/_forms.scss
@@ -89,15 +89,8 @@ progress::-moz-progress-bar {
     margin: $grid-gap 0;
   }
 
-  .mailpoet-form-send-email & {
-    grid-column: 2 / -1;
-    padding-top: $grid-gap + ($heading-font-size-h4 * $heading-line-height) +
-      ($font-size * $line-height * 2);
-    text-align: right;
-  }
-
   @include respond-to(small-screen) {
-    grid-column: 1 !important;
+    grid-column: 1;
   }
 }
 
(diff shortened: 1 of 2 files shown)


mailpoet/assets/css/src/components-plugin/_forms.scss
```diff
@@ -75,8 +75,13 @@ progress::-moz-progress-bar {
   justify-content: space-between;
   margin: $grid-gap auto;
 
+  .mailpoet-form-send-email & {
+    grid-template-columns: $grid-column $grid-column;
+    max-width: 976px;
+  }
+
```
JanJakes: > I decided not to use "mailpoet-main-container" because it's too wide, for two 368px columns.

:+1:, but do you think this is a good place to fix it (plus the `!important`)? I would argue that a form from "components" doesn't need to know about a specific email form. What I mean is that rather than `.mailpoet-form-grid { .mailpoet-form-send-email & { ... } }` here we could do `.mailpoet-form-send-email { .mailpoet-form-grid { ... } }` in another place.

Or maybe even easier, add a new `.mailpoet-form-send-email-container`...

NeosinneR: That's fair. I've done it as you suggested and also moved out similarly component-specific style and `!important` with it.

JanJakes (approved):
Thanks!

Aschepikov (approved):
GJ!

PR #5315, opened 2023-11-28
DQL delete & sync fixes [MAILPOET-5745]

### LINKED ISSUE #5281: Action scheduler/cron error
**Describe the bug**

I have been experiencing cron problems targeting Mailpoet for a few days now.
here is an example

mailpoet/cron/daemon-runFailedmailpoet-cronNon-repeating2023-11-08 11:23:10 +0000

action failed via WP Cron: A new entity was found through the relationship 'MailPoet\Entities\ScheduledTaskSubscriberEntity#task' that was not configured to cascade persist operations for entity: MailPoet\Entities\ScheduledTaskEntity@22447. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'MailPoet\Entities\ScheduledTaskEntity#__toString()' to get a clue.

Similar to https://github.com/mailpoet/mailpoet/issues/4451
**To reproduce**

1. Not clear...apparently have emails scheduled.

**Actual behavior**
A clear and concise description of the incorrect behavior.

Error and presumably the email wasn't sent.

**Expected behavior**

Error free email sending.
**Screenshots**

- WordPress version: 6.4.1
- PHP version: [e.g: 7.4.2]
- MailPoet version: 4.35.1

**Additional context**

7291841-zen
https://wordpress.org/support/topic/fatal-error-and-scheduler-is-not-creating-for-some-post-2/

JanJakes:
## Description

When deleting entities via DQL, the changes aren't reflected in the Doctrine identity map, and we need to synchronize (detach) them manually. This adds a helper to do so, without the need of fetching them to memory before.

The way it works is we iterate all entities of the type already loaded in memory, and figure out which ones to detach.

I also added a fix for calling `flush()` in the logger, which can have unintended side effects (explained in the commit and as a code comment).

I also fixed a `SettingsController::set()` side effect of clearing all settings entities, and a bug causing `SettingsController::fetch()` to return cached entity data instead of fresh data (as it was originally intended).

Fixes https://github.com/mailpoet/mailpoet/issues/5281.

## Code review notes

The DQL/DBAL vs. identity map problem is not universally solvable, and we need to manually sync such changes. For most cases, `$entityManager->refresh()` is the method to use, but for bulk deletes of large collections where we don't want to load all entities to memory, we need a way to detach them without fetching them first.

It's very much an edge case, and likely calling `$entityManager->clear()` between cron runs within the same request could prevent this (and other possible bugs) as well. But it's better to be on the safe side, and ultimately, implement both.

## QA notes

Please, test sending, especially post notification, plus some automations, and check `?page=mailpoet-logs` as well as [Action Scheduler logs](https://actionscheduler.org/admin/) to make sure no errors were logged.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5745]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5745]: https://mailpoet.atlassian.net/browse/MAILPOET-5745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Doctrine/Repository.php
+++ b/mailpoet/lib/Doctrine/Repository.php
@@ -127,6 +127,24 @@ abstract class Repository {
   }
 
   /**
+   * @param callable(T): bool|null $filter
+   */
+  public function detachAll(callable $filter = null): void {
+    $className = $this->getEntityClassName();
+    $rootClassName = $this->entityManager->getClassMetadata($className)->rootEntityName;
+    $entities = $this->entityManager->getUnitOfWork()->getIdentityMap()[$rootClassName] ?? [];
+    foreach ($entities as $entity) {
+      if (!($entity instanceof $className)) {
+        continue;
+      }
+      if ($filter && !$filter($entity)) {
+        continue;
+      }
+      $this->entityManager->detach($entity);
+    }
+  }
+
+  /**
    * @return class-string<T>
    */
   abstract protected function getEntityClassName();
--- a/mailpoet/lib/Form/FormsRepository.php
+++ b/mailpoet/lib/Form/FormsRepository.php
@@ -96,10 +96,17 @@ class FormsRepository extends Repository {
       return 0;
     }
 
-    return $this->entityManager->createQueryBuilder()
+    $result = $this->entityManager->createQueryBuilder()
       ->delete(FormEntity::class, 'f')
       ->where('f.id IN (:ids)')
       ->setParameter('ids', $ids)
       ->getQuery()->execute();
+
+    // delete was done via DQL, make sure the entities are also detached from the entity manager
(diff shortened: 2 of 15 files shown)


costasovo (approved):
@JanJakes Thank you for the thorough investigation of the issues! This will improve stability and reduce the risk of weird issues with data. 👏

JanJakes:
I'm adding priority as I would like to release this soon to monitor the errors before we merge in more Doctrine changes.

Aschepikov (approved):
GJ!

PR #5028, opened 2023-06-29
Optimize sending code [MAILPOET-5178]

alex-mailpoet:
## Description

These sending code optimizations were described in the [sending performance report](https://mailpoet.atlassian.net/browse/MAILPOET-5178?focusedCommentId=24113).

## Code review notes

_N/A_

## QA notes

Prerequisites for testing sending speed:
* Very big list (100K-1M);
* Blackhole addresses;
* MSS enabled;
* Hosting: WPCOM, Hetzner etc.
Write me on Slack for access to a test host where everything is set up.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5178]

## After-merge notes

_N/A_

## Tasks

- [x] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] I added sufficient test coverage
- [x] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5178]: https://mailpoet.atlassian.net/browse/MAILPOET-5178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5316, opened 2023-11-28
Fix WooCommerce email counters [MAILPOET-5719]

lysyjan:
## Description

This PR fixes the reported issue with displaying the count of sent WooCommerce emails, such as Abandoned cart, First purchase, etc.

## Code review notes

The fix is in two parts:
1.  Because we changed the logic of updating counts in sending queues after a successful finish, the method `Sending::updateCounts()` stopped working for WC emails correctly. The reason was [this commit](https://github.com/mailpoet/mailpoet/pull/5028/commits/bcea5357dc416f24fca6a1443c51c8ad072ba45b#diff-ae01c9cf1eb5e11c90936b70c6f47711f8c92e05d705cdbe3813dc2a025e654dR248) when the first part of the if statement restricts the min and max values for `countProcessed`. The counts in the sending queue were not initialized properly, and all were in the default value of 0. My solution is based on setting correct values after entity creation.
2. For the reason that the count of sent WC emails is calculated from the `countProcessed` in the sending_queue table, we need to repair the affected rows. I prepared a migration that should guarantee.

## QA notes

- In the beginning, I recommend preparing all WC email types.
- As the next step, replicate the incorrect behavior on each new email.
- After replication, install the new version containing the fix and ensure the DB migration was executed.
- Invoke sending all WC emails again and observe if the updated count of sent emails is correct.

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/5028

## Linked tickets

[MAILPOET-5719]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5719]: https://mailpoet.atlassian.net/browse/MAILPOET-5719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Migrations/App/Migration_20231128_120355_App.php
+++ b/mailpoet/lib/Migrations/App/Migration_20231128_120355_App.php
@@ -29,7 +29,7 @@ class Migration_20231128_120355_App extends AppMigration {
     $sendingQueuesTable = $this->getTableName(SendingQueueEntity::class);
     $scheduledTasksTable = $this->getTableName(ScheduledTaskEntity::class);
     $newslettersTable = $this->getTableName(NewsletterEntity::class);
-    $typeAutomatic = NewsletterEntity::TYPE_AUTOMATIC;
+    $newsletterTypes = [NewsletterEntity::TYPE_AUTOMATIC, NewsletterEntity::TYPE_WELCOME];
     $statusCompleted = ScheduledTaskEntity::STATUS_COMPLETED;
     $connection->executeStatement("
       UPDATE {$sendingQueuesTable}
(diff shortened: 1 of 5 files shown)


alex-mpoet (dismissed):
Looks good to me! Thanks for fixing it. Went unreported for quite a long time.

Aschepikov:
Hi @lysyjan as I see welcome emails are affected too, can you add the fix for them or will it be in a separate ticket?

lysyjan:
@Aschepikov Nice catch! I added two commits to address this issue. @alex-mpoet Please, can you look at those new commits that fix welcome emails.

alex-mpoet (approved):
New changes approved. @Aschepikov Could you please check that older emails get fixed with a migration?

Aschepikov:
GJ!

Aschepikov (approved):

PR #5319, opened 2023-11-30
Fix displaying columns full width in the gmail app  [MAILPOET-5753]

costasovo:
## Description

The main content wrapping table was not set with a width 100%, and it was causing the Gmail app to stretch the content area width by content size. So, it was not related to columns but rather to the main template wrapper.

We found that the table is not needed since columns add their own table wrappers.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

 [MAILPOET-5753]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5753]: https://mailpoet.atlassian.net/browse/MAILPOET-5753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/template.html
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/template.html
@@ -57,16 +57,7 @@
                   text-align: center;
                 "
               >
-                <table
-                  role="presentation"
-                  border="0"
-                  cellpadding="0"
-                  cellspacing="0"
-                >
-                  <tbody>
-                    {{email_body}}
-                  </tbody>
-                </table>
+                {{email_body}}
               </td>
             </tr>
           </tbody>


mailpoet/lib/EmailEditor/Engine/Renderer/template.html
```diff
@@ -62,6 +62,7 @@
                   border="0"
                   cellpadding="0"
                   cellspacing="0"
+                  style="width: 100%"
```
lysyjan: 👍 Good catch!
During the review and testing, I noticed that this table does not contain table tags the row and the column.
Will we create a ticket, or will you take a look?

costasovo: @lysyjan  That is an even better catch! Maybe we could remove the table. I'll take a look.

lysyjan (approved):
👍 Thanks for the fix

PR #5322, opened 2023-11-30
Refactor \MailPoet\Segments\WP to use Doctrine instead of Paris [MAILPOET-5752]

rodrigoprimo:
## Description

Rewrites the WordPress user segment to use Doctrine.

## Code review notes

_N/A_

## QA notes

Please ensure that WordPress user synchronization to the MailPoet segment works, welcome emails are scheduled, and test subscriber import and export.

**When QA passes, you can merge this PR (no need to assign back to @JanJakes).**

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5752]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5752]: https://mailpoet.atlassian.net/browse/MAILPOET-5752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
+++ b/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
@@ -183,6 +183,7 @@ class ImportExportRepository {
       " . implode(' AND ', $keyColumnsConditions) . "
     ", $parameters, $parameterTypes);
     $this->notifyUpdates($className, $columns, $data);
+    $this->entityManager->clear($className);
     return $count;
   }
 
--- a/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
+++ b/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
@@ -195,8 +195,12 @@ class ImportExportRepository {
       " . implode(' AND ', $keyColumnsConditions) . "
     ", $parameters, $parameterTypes);
     $this->notifyUpdates($className, $columns, $data);
-    $this->subscribersRepository->detachAll();
-    $this->subscriberCustomFieldRepository->detachAll();
+    if ($className === SubscriberEntity::class) {
+      $this->subscribersRepository->refreshAll();
+    }
+    if ($className === SubscriberCustomFieldEntity::class) {
+      $this->subscriberCustomFieldRepository->refreshAll();
+    }
     return $count;
   }
 
--- a/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
+++ b/mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
@@ -10,6 +10,8 @@ use MailPoet\Entities\SubscriberCustomFieldEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Entities\SubscriberSegmentEntity;
 use MailPoet\Segments\DynamicSegments\FilterHandler;
+use MailPoet\Subscribers\SubscriberCustomFieldRepository;
+use MailPoet\Subscribers\SubscribersRepository;
 use MailPoetVendor\Doctrine\DBAL\Connection;
 use MailPoetVendor\Doctrine\DBAL\Driver\Statement;
 use MailPoetVendor\Doctrine\DBAL\Query\QueryBuilder;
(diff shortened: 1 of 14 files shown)


mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
```diff
@@ -170,6 +170,7 @@ public function updateMultiple(
       " . implode(' AND ', $keyColumnsConditions) . "
     ", $parameters, $parameterTypes);
     $this->notifyUpdates($className, $columns, $data);
+    $this->entityManager->clear($className);
```
costasovo: @rodrigoprimo `EntityManage::clear with entity name is deprecated. The issue is pretty much related to the investigation and changes @JanJakes did in https://github.com/mailpoet/mailpoet/pull/5315. We should use [the helper](https://github.com/mailpoet/mailpoet/pull/5315/commits/8457976c5f5c0933a255bb5a60db037e39ede3bc) introduced in Jan's PR.

costasovo (changes requested):
@rodrigoprimo I think we need to replace the entity manager flush in export/impoer repository by the new helper introduced in https://github.com/mailpoet/mailpoet/pull/5315.

mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
rodrigoprimo: Thanks for the review. It makes sense to me to use the helper introduced by JJ in #5315. I guess we will have to wait for his PR to be merged. I will ask @websupporter how to proceed here since I will move to another team next week.

websupporter:
Assigned @alex-mpoet, as he will continue this task

mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
alex-mpoet: @costasovo There is a complication for using the helper. The `ImportExportRepository` class doesn't inherit from `Repository` because it works with multiple entities and can't implement the `getEntityClassName` abstract method needed by the helper (confirmed by @lysyjan who made this class). 

What do you think would be the best workaround? Implementing a private version of the helper for this class which would work (we know the `className` and constraints)?

costasovo: @alex-mpoet Hmm, I see that the method is called for SubscriberCustomFieldEntity and SubscriberEntity, so I guess we could also inject repositories for these two entities and call `detachAll` on those. In my opinion, better than duplicating the code somewhere else. Would such a solution work?

alex-mpoet: Yes, that would work, thanks! I made the change.

costasovo (commented):
Thank you, @alex-mpoet, for adding the repository cleanup!

I went through the PR again, and I realized that in `MailPoet\Segments\WP::synchronizeUsers`, we do multiple changes to subscribers' data using SQL, and these will get Entities in EntityMap out of sync. The sync happens in an extra requests (API Call for sync users or plugin activation), but I think that it would be safer to add `$this->subscribersRepository->detachAll();` into the [synchronizeUsers](https://github.com/mailpoet/mailpoet/blob/55480fcf5317d52acd3b379acfd7e6dfe8f2fbfe/mailpoet/lib/Segments/WP.php#L248-L268) method. What do you think?

alex-mpoet:
> it would be safer to add $this->subscribersRepository->detachAll(); into the [synchronizeUsers](https://github.com/mailpoet/mailpoet/blob/55480fcf5317d52acd3b379acfd7e6dfe8f2fbfe/mailpoet/lib/Segments/WP.php#L248-L268) method. What do you think?

Sure, extra safety won't hurt. Done!

costasovo (dismissed):
LGTM! 👍 

@JanJakes You mentioned you want to coordinate passing the rest of Doctrine PRs to QA so I'm assigning the PR to you.

JanJakes:
@costasovo I see you were discussing the `detachAll` logic here, do you remember the context? As far as I understand, it was to replace `clear`. It indeed would behave the same as `clear` here, but stepping back, I'm not sure if it's the right helper to use.

The `detachAll` helper should be used in scenarios like "I removed something in the DB, remove it also in memory, in case it exists". Now we have also a `refreshAll` helper that can be used like "I updated something, update it also in memory, in case it's there".

I feel like `refreshAll` is the proper helper to use for updates, the only downside is that it will eagerly refresh entities that are in memory. On the other hand, if no entities are loaded in memory, it will be a noop.

I'm unsure what happens with `detachAll` with updates. I guess in case you try to access the entity via Entity Manager, it would result in a "refresh", which would be done "lazily", which is nice. But... in case you already have a reference to an entity, then accessing it's properties would result in usage of outdated data. With `refreshAll` this wouldn't happen.

costasovo:
@JanJakes Thanks for the ping!

> I see you were discussing the detachAll logic here, do you remember the context?

I was assigned to the PR at the same time you were working on the first helper, and using the new helper sounded like a nice replacement for the deprecated `clear`. I didn't step back and think about the full context as you did.

> The detachAll helper should be used in scenarios like "I removed something in the DB, remove it also in memory, in case it exists". Now we have also a refreshAll helper that can be used like "I updated something, update it also in memory, in case it's there".

> I feel like refreshAll is the proper helper to use for updates, the only downside is that it will eagerly refresh entities that are in memory. On the other hand, if no entities are loaded in memory, it will be a noop.

Thank you for stepping back and thinking about it. I agree that for this "update scenario" using refresh is much more appropriate.

I can't find the `refreshAll` to see how it works. Isn't it still in a branch?  I assume it fetches everything. I've briefly checked the `MailPoet\Subscribers\ImportExport\Import\Import` to see what entities could be loaded. I see we fetch all custom fields and subscriber tags (during the processing, we fetch existing and create new ones based on the import data). I also see that for rendering the export page we fetch all custom fields and all tags (`ImportExportFactory::bootstrap`), so we are already assuming that there will not be many rows. So I think we should switch to the refresh helper.

Aschepikov (approved):
GJ!

PR #5326, opened 2023-12-01
Update email stats page - MAILPOET-5699

triple0t:
## Description

We updated the email stats page in this PR.

## Code review notes

Please check commits

## QA notes

Note: Please test with MailPoet Premium activated and deactivated.
Associated premium branch here: https://github.com/mailpoet/mailpoet-premium/pull/829

Start with the QA notes here: https://github.com/mailpoet/mailpoet/pull/5289
* Check email stats here: `/wp-admin/admin.php?page=mailpoet-newsletters%2F#/stats/newsletter_id`
* Please test with different types of newsletters e.g. standard, welcome, post notifications, post notifications history
* You should be able to edit, duplicate (if possible), and delete newsletters.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/829

## Linked tickets

[MAILPOET-5699](https://mailpoet.atlassian.net/browse/MAILPOET-5699)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5699]: https://mailpoet.atlassian.net/browse/MAILPOET-5699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
+++ b/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
@@ -120,7 +120,9 @@ type Props = {
 
 function NewsletterStatsInfo({ newsletter }: Props) {
   const newsletterDate =
-    newsletter.queue.scheduled_at || newsletter.queue.created_at;
+    newsletter?.queue?.scheduled_at ||
+    newsletter?.queue?.created_at ||
+    newsletter?.created_at;
   return (
     <Grid.ThreeColumns className="mailpoet-stats-info">
       <div>
--- a/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
+++ b/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
@@ -224,7 +224,7 @@ function NewsletterStatsInfo({ newsletter }: Props) {
                   aria-expanded={isOpen}
                   variant="primary"
                 >
-                  <br />
+                  &nbsp;
                   <Icon icon={chevronDown} size={18} />
                 </Button>
               </ButtonGroup>
--- a/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
+++ b/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
@@ -181,7 +181,7 @@ function NewsletterStatsInfo({ newsletter }: Props) {
             {__('Preview', 'mailpoet')}
           </Button>
           <Dropdown
-            className="mailpoet-stats-buttons-header"
+            className="mailpoet-stats-has-margin-left"
             focusOnMount={false}
             popoverProps={{ placement: 'bottom-end' }}
             renderToggle={({ isOpen, onToggle }) => (
--- a/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
+++ b/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
@@ -1,5 +1,11 @@
 import { __ } from '@wordpress/i18n';
-import { Button, ButtonGroup, Dropdown, MenuItem } from '@wordpress/components';
+import {
+  Button,
+  ButtonGroup,
+  Dropdown,
+  MenuItem,
+  MenuGroup,
+} from '@wordpress/components';
 import { chevronDown, Icon } from '@wordpress/icons';
 import { MailPoet } from 'mailpoet';
 import { Heading } from 'common/typography/heading/heading';
@@ -61,7 +67,7 @@ function NewsletterStatsInfo({ newsletter }: Props) {
           </div>
         </div>
       </div>
-      <div className="mailpoet-stats-info-sender-preview">
+      <div className="mailpoet-stats-button-group">
         <ButtonGroup>
           <Button
             href={newsletter.preview_url}
@@ -72,6 +78,7 @@ function NewsletterStatsInfo({ newsletter }: Props) {
             {__('Preview', 'mailpoet')}
           </Button>
           <Dropdown
+            className="mailpoet-stats-buttons-header"
             focusOnMount={false}
             popoverProps={{ placement: 'bottom-end' }}
             renderToggle={({ isOpen, onToggle }) => (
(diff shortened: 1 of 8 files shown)


mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
```diff
@@ -10,9 +10,110 @@ import { chevronDown, Icon } from '@wordpress/icons';
...
+  let editorHref = `?page=mailpoet-newsletter-editor&id=${newsletter.id}`;
+  if (
+    MailPoet.FeaturesController.isSupported('gutenberg_email_editor') &&
+    newsletter.wp_post_id
+  ) {
+    editorHref = `admin.php?page=mailpoet-email-editor&postId=${newsletter.wp_post_id}`;
+  }
+
+  return editorHref;
+};
```
costasovo, line 37: Just an idea. I see that we already have `preview_url`[ in NewsletterResponseBuilder](https://github.com/mailpoet/mailpoet/blob/53bc5a25059b700aca87b951edab83e8f9f06e17/mailpoet/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php#L154) in API. So maybe we could add `editor_url` also to the standard newsletter response from API. I guess we could use it in multiple places. We don't need to do it within this task. Just got the idea as I'm reading the code. What do you think?

mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
```diff
@@ -128,6 +134,10 @@ function NewsletterStatsInfo({ newsletter }: Props) {
       <div>
         <Heading level={1}>{newsletter.subject}</Heading>
         <div>
+          <Tag isInverted={false}>
+            {capitalizeFirstLetter(newsletter.status)}
```
costasovo: This way, the status won't be translated. AFAIK we display the page only for a couple of statues so maybe we can have a simple logic here to display a proper translation. There is also [a component for displaying newsletter status](https://github.com/mailpoet/mailpoet/blob/c8fc6bf49aa8add3f989530c792ae1cf5fac31ae/mailpoet/assets/js/src/common/listings/newsletter-status.tsx) but I'm not sure if it is good fit for this case.

costasovo (changes requested):
Good job! I think we should address an issue with translations of the status. Otherwise LGTM.

mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
triple0t: Nice catch. Fixed in https://github.com/mailpoet/mailpoet/pull/5326/commits/e7b4d3df9a5126313bf232c763ae5acb254b7f48

mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
triple0t: Hmm 🤔, this is a great suggestion.
I made a quick attempt at implementing the suggestion in this PR, but I realized the scope might be larger than I expected. 

We do not return `preview_url` in [in NewsletterResponseBuilder](https://github.com/mailpoet/mailpoet/blob/53bc5a25059b700aca87b951edab83e8f9f06e17/mailpoet/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php#L154) but instead in [Newsletters API endpoint](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/API/JSON/v1/Newsletters.php#L151). I don't know why yet.

It occurred to me we might need to build a service to handle this and also update this section on [the Premium plugin](https://github.com/mailpoet/mailpoet-premium/blob/trunk/lib/API/JSON/v1/ResponseBuilders/StatsResponseBuilder.php#L32-L47) to reuse the service.

triple0t:
Thank you for the quick review, @costasovo 

I've updated the PR

mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
costasovo: @triple0t Thank you for exploring the scope of the change.

> We do not return preview_url in [in NewsletterResponseBuilder](https://github.com/mailpoet/mailpoet/blob/53bc5a25059b700aca87b951edab83e8f9f06e17/mailpoet/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php#L154) but instead in [Newsletters API endpoint](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/API/JSON/v1/Newsletters.php#L151). I don't know why yet.

Aha! I [saw it in the NewsletterResponseBuilder](https://github.com/mailpoet/mailpoet/blob/553fcfa1b4d0007acd470470301a21bbff333580/mailpoet/lib/API/JSON/ResponseBuilders/NewslettersResponseBuilder.php#L154) but I missed it is added only when generating response for listing. So we have it in multiple places. It could be nice to refactor as well.

> I made a quick attempt at implementing the suggestion in this PR, but I realized the scope might be larger than I expected.

Let's leave it for an extra ticket. I created [a Jira ticket for that](https://mailpoet.atlassian.net/browse/MAILPOET-5776). Please feel free to add more details in the ticket.

costasovo (dismissed):
LGTM! Thank you for the fixes!

Aschepikov:
Hi @triple0t  There should be a "sending status" like on the linked page?
image

triple0t:
> There should be a "sending status" like on the linked page?

Fixed in https://github.com/mailpoet/mailpoet/pull/5326/commits/ff3b3fb90fa42436d55ca7004fb04c78d378785a

Thanks, @Aschepikov

Aschepikov:
image
@triple0t  I don't see the deactivate option for the post notifications, I can get the post notification statistics from the history, and in this case, it is not the active email so it is logical, but I see in the ticket that it should work for the post notifications. Is it expected?
image

triple0t:
> I don't see the deactivate option for the post notifications, I can get the post notification statistics from the history, and in this case, it is not the active email so it is logical, but I see in the ticket that it should work for the post notifications. Is it expected?


Yes.

Post notifications newsletter don’t have statistics (we don’t send them)
Post notifications history newsletter do.

We will be linking directly to the Post notifications history items from the multichannel dashboard so we don’t need to support the post notification item itself.

Aschepikov (approved):
GJ!

PR #5330, opened 2023-12-05
Update PHPStan version used in static analysis job [MAILPOET-5751]

lysyjan:
## Description

_N/A_

## Code review notes

- I tried to avoid extending baseline files as much as I could. Nevertheless, I added a few rules when I didn't know how to fix them easily.
- I experimented with failing child process on CircleCI, and after several tests with different configurations of parallelism via SSH I found out the best way is to change the instance type from medium to medium+. I think it's because the newer version of PHPStan contains more tests. It could be possible to run the static analysis job on the medium, but it would take 5 minutes. In my opinion, it's too much.
- Removed methods were unused.
- Major changes are based on two new rules:
  - `strval()` and `intval()`, etc., functions don't accept a value of type `mixed`
  - `$wpdb->prepare()` expects a variable of type `literal-string`

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/832

## Linked tickets

[MAILPOET-5751]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5751]: https://mailpoet.atlassian.net/browse/MAILPOET-5751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/tasks/phpstan/composer.json
```diff
@@ -1,10 +1,10 @@
...
-    "php-stubs/woocommerce-stubs": "8.2.2",
-    "phpstan/phpstan": "1.4.10",
-    "phpstan/phpstan-doctrine": "1.2.11",
-    "phpstan/phpstan-phpunit": "1.0.0",
-    "szepeviktor/phpstan-wordpress": "1.0.3"
+    "php-stubs/woocommerce-stubs": "v8.3.1",
+    "phpstan/phpstan": "1.10.46",
+    "phpstan/phpstan-doctrine": "1.3.53",
+    "phpstan/phpstan-phpunit": "1.3.15",
+    "szepeviktor/phpstan-wordpress": "v1.3.2"
```
johnolek: Is there a reason two of these are now prefixed with a `v`? As far as I know it doesn't make a difference but it could be a tiny bit confusing when it comes to making updates later.

johnolek (dismissed):
This looks good to me. Thank you for making all of these updates! I only had one small but probably inconsequential question about version strings in composer.json.

I agree it's better to use medium+ and have the jobs run in 2 minutes instead of using medium without parallelization and taking 5 minutes, since it seems to be 10 credits/minute for medium and 15 credits/minute for medium+. Just to double check, you don't think this will have a significant impact on the number of credits we're using monthly, right?

mailpoet/tasks/phpstan/composer.json
lysyjan: Good point @johnolek. I addressed it in the last commit.

lysyjan:
> I agree it's better to use medium+ and have the jobs run in 2 minutes instead of using medium without parallelization and taking 5 minutes, since it seems to be 10 credits/minute for medium and 15 credits/minute for medium+. Just to double check, you don't think this will have a significant impact on the number of credits we're using monthly, right?

@johnolek I have briefly checked the static analysis time for the trunk branch, and it takes approximately 2 minutes and 30 seconds. For updated PHPStan and instance medium+, it takes approximately 2 minutes. So, I don't expect a significant impact.

johnolek (approved):
@lysyjan Thanks for the update and for confirming the lack of expected CircleCI impact!

Aschepikov (approved):
The major plugin functionality tested

PR #5333, opened 2023-12-06
Fix purchased in category filter bug [MAILPOET-5264]

johnolek:
## Description

This fixes a bug with the "purchased in category" dynamic filter, where it was not correctly returning subscribers for the "all of" condition. Previously, if a subscriber made two separate purchases, one that included category1 and the other that included category2, a filter for "all of" both category1 and category2 would not return that subscriber.

## Code review notes

I'm open to ideas if you can think of a good way to restructure the query without having to rely on so many subqueries/joins, but it seems like it could get very complex due to the hierarchical nature of categories.

## QA notes

Please test the "purchased in category" filter carefully, especially the "all of" option, and please also test to ensure that the filter works as expected with hierarchical categories. For example, if you have `term1` and `term1child`, and a filter that is looking for anyone who has purchased `term1`, it should include users who have only purchased items with `term1child`.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/WooCommerceCategory.php
@@ -64,10 +64,9 @@ class WooCommerceCategory implements Filter {
       $this->applyProductJoin($queryBuilder, $orderStatsAlias);
       $this->applyTermRelationshipsJoin($queryBuilder);
       $this->applyTermTaxonomyJoin($queryBuilder, $parameterSuffix)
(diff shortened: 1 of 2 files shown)


johnolek:
@triple0t Please ignore my review request for now, I see that there are some things I need to fix

johnolek:
@triple0t This is now ready to review. Thanks in advance for taking a look!

triple0t (approved):
LGTM

Aschepikov (approved):
GJ!

PR #5334, opened 2023-12-06
Number of orders with coupon code dynamic segment filter [MAILPOET-5373]

johnolek:
## Description

This PR adds a new dynamic segment filter for number of orders with a coupon code.

## Code review notes

Fortunately it was possible to rely almost entirely on the existing "number of purchases" filter, simply adding an inner join when we want to restrict it to orders that used coupon codes.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/days-period-field.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/days-period-field.tsx
@@ -84,5 +84,6 @@ export function validateDaysPeriod(formItems: DaysPeriodItem): boolean {
   if (formItems.timeframe === Timeframe.ALL_TIME) {
     return true;
   }
-  return !!formItems.days;
+  const days = parseInt(formItems.days, 10);
+  return days >= 1;
 }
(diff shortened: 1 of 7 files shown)


brezocordero (dismissed):
@johnolek Nice implementation using the existing filter.
Small thing and not related to your change, so I am approving the PR. Maybe create a separate ticket with low priority? 
You can change the value of the input field [days-period-field](https://github.com/mailpoet/mailpoet/blob/91b8ed9ddeb3b6ba251a8b46756995f157394c33/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/days-period-field.tsx#L48) to 0 and the filter will [throw an error about missing values.](https://github.com/mailpoet/mailpoet/blob/91b8ed9ddeb3b6ba251a8b46756995f157394c33/mailpoet/lib/Segments/DynamicSegments/Filters/FilterHelper.php#L78)
Screenshot 2023-12-07 at 9 17 51 PM

johnolek:
@brezocordero Thanks for the review and for pointing out the issue with the days period field! I thought it was a simple enough fix that a separate ticket seemed unnecessary. I fixed it in 2cceaf02492b4a5a120139595761b0e6e84adf06. This will also apply to all the other filters that use this days period component.

brezocordero (approved):
Nice! Thanks for fixing it.

Aschepikov:
Hi @johnolek there are some failed tests can you take a look pls

johnolek:
@Aschepikov I think those failures were flaky tests. Just to be on the safe side, since this PR is a bit old now, I rebased on the latest trunk and re-pushed. Looks like everything passed this time!

Aschepikov (approved):
GJ!

PR #5335, opened 2023-12-07
Segment list design [MAILPOET-5395]

websupporter:
## Description
This fixes [MAILPOET-5395]. Some elements are not completely like in the design. E.g. you can't sort by contacts or subscribers and the magnifier is missing in the search box. That is to be expected as for now.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5395]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5395]: https://mailpoet.atlassian.net/browse/MAILPOET-5395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5395]: https://mailpoet.atlassian.net/browse/MAILPOET-5395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


johnolek:
@websupporter It looks like there are some pipeline failures that will need to be addressed. Do you have time to look at this again or would you like me to take it over?

johnolek:
@websupporter Do you think you would have time to take a look at this? I think I have it working and all the tests passing finally. I ask because you were working on it before and are already familiar with it, but if you don't have time that's totally ok and I can reassign it.

websupporter:
Hi @johnolek, I would like other engineers to get familiar with the context. I reassigned the PR therefore.

mailpoet/assets/css/src/mailpoet-dynamic-segments.scss
```diff
@@ -0,0 +1,35 @@
+@import 'settings/colors';
+
+#dynamic_segments_container {
+  padding: 0 12px;
```
JanJakes: This seems to cause a padding that is inconsistent with other pages. It may have been designed like that, but I don't think we should do it. Look at other pages vs. this one, especially automations vs. segments. If we were to make larger paddings, then it should be done on all pages.

mailpoet/assets/css/src/mailpoet-dynamic-segments.scss
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.mailpoet-segment-listing-heading {
+  padding: 12px 0 32px;
```
JanJakes: Same here, I don't think we should add inconsistent spacing here. As for headings, I also think we should use the existing [PageHeader](https://github.com/mailpoet/mailpoet/tree/trunk/mailpoet/assets/js/src/common/page-header) component rather than introducing a new one.

mailpoet/assets/js/src/segments/dynamic/list/listing-header.tsx
```diff
@@ -0,0 +1,42 @@
+import { Button, Flex } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import * as ROUTES from 'segments/routes';
+import { plusIcon } from 'common/button/icon/plus';
+import { SubscribersInPlan } from 'common/subscribers-in-plan';
+import { MailPoet } from 'mailpoet';
+import { SubscribersCacheMessage } from 'common/subscribers-cache-message';
+import { MssAccessNotices } from 'notices/mss-access-notices';
+
+export function ListingHeader(): JSX.Element {
```
JanJakes: Could we use the existing [PageHeader](https://github.com/mailpoet/mailpoet/tree/trunk/mailpoet/assets/js/src/common/page-header) component?

mailpoet/assets/js/src/segments/static/heading.tsx
```diff
@@ -5,32 +5,19 @@ import { plusIcon } from 'common/button/icon/plus';
 import { SubscribersInPlan } from 'common/subscribers-in-plan';
 import { MssAccessNotices } from 'notices/mss-access-notices';
 import { SubscribersCacheMessage } from 'common/subscribers-cache-message';
-import * as ROUTES from 'segments/routes';
 
-function ListHeading({ segmentType }): JSX.Element {
+function ListHeading(): JSX.Element {
```
JanJakes, line 9: Hmm, now I'm a bit confused what is `ListingHeader` vs. `ListHeading`, so I'm not sure if the [PageHeader](https://github.com/mailpoet/mailpoet/tree/trunk/mailpoet/assets/js/src/common/page-header) argument applies here...?

mailpoet/assets/js/src/segments/dynamic/store/actions.ts
```diff
@@ -222,3 +223,34 @@ export function setPreviousPage(data: string): SetPreviousPageActionType {
...
+      sort_by: 'name',
+      sort_order: 'desc',
+    },
+  })
+    .done((response) => {
+      data = response.data || [];
+    })
+    .fail((response) => {
+      if (response.errors.length > 0) {
+        // TODO: handle errors
```
JanJakes: There is a `TODO` here, but I also wonder if `done`/`fail` is needed, or if with `await` we can just look at the returned values...

mailpoet/assets/css/src/components-automation-listing/_listing.scss
```diff
@@ -29,45 +31,6 @@
   }
 }
 
-.mailpoet-filter-tab-panel {
-  background-color: #fff;
-  border: 1px solid #dcdcde;
-  border-radius: 2px;
-  margin-top: 16px;
```
JanJakes, line 36: I think this commit seems to cause a missing spacing on the automation listing page:

Screenshot 2024-02-21 at 09 18 13

mailpoet/assets/css/src/components-automation-listing/_listing.scss
```diff
@@ -29,45 +31,6 @@
   }
 }
 
-.mailpoet-filter-tab-panel {
-  background-color: #fff;
-  border: 1px solid #dcdcde;
```
JanJakes, line 34: The listing in automations used to have an outer border that now disappeared. This may have been updated in Figma, but we should probably double-check if that's desired. Borderless tables or cards seem rather unusual in the WP/Woo admin design.

mailpoet/assets/js/src/segments/dynamic/list/listing-table.tsx
```diff
@@ -0,0 +1,157 @@
...
+  );
+
+  const dynamicSegments = useSelect((select) =>
+    select(storeName).getDynamicSegments(),
+  );
+  const { loadDynamicSegments } = useDispatch(storeName);
+
+  useEffect(() => {
+    loadDynamicSegments().catch(() => {
+      // TODO: check if this is the right away to use loadDynamicSegments and handle errors if it is
```
JanJakes: There is a `TODO` here, and a small thing to note is we're generally preferring async/await over then/catch.

mailpoet/assets/js/src/segments/dynamic/list/listing-table.tsx
```diff
@@ -0,0 +1,157 @@
...
+              typeof TableCard
+            >['headers']
+          }
+          rows={rows}
+          rowKey={(_, i) => dynamicSegments[i].id}
+          rowsPerPage={rowsPerPage}
+          totalRows={dynamicSegments.length}
+          query={Object.fromEntries(pageSearch)}
+          showMenu={false}
+        />
```
JanJakes: Do you think we could also do [this](https://github.com/mailpoet/mailpoet/pull/5404/commits/b1586403cd60372b1389fef4de6c5decdec3f94e) to have the placeholder use the real number of segments?

mailpoet/assets/js/src/segments/dynamic/list/listing-table.tsx
```diff
@@ -1,13 +1,74 @@
...
+  }
+  const defaultQuery = {
+    offset: 0,
+    limit: 2,
+    filter: {},
+    search: '',
+    sort_by: 'updated_at',
+    sort_order: 'desc',
+    group: 'all',
+  };
```
JanJakes: The default query seems to be defined exactly the same way a few lines above. Maybe we could extract & reuse it.

mailpoet/assets/js/src/segments/dynamic/list/listing-table.tsx
```diff
@@ -1,13 +1,74 @@
...
+          `${queryKeys[queryKeysIndex]}[`,
+        )
+      ) {
+        query[queryKeys[queryKeysIndex]] = pathElements[pathElementsIndex]
+          .replace(`${queryKeys[queryKeysIndex]}[`, '')
+          .replace(']', '');
+      }
+    }
+  }
+  updateDynamicQuery(query);
```
JanJakes: This code looks a bit complex, and I'm wondering what's happening here in general. We seem to have:
1. `updateDynamicQuery` that syncs the query to the store,
2. `updateDynamicQueryFromLocation` that does "the same", but from the URL.

It seems to me that the state is unnecessarily duplicated, and that it may be hard to sync those two (store and URL). Why don't we store the dynamic query state in URL only? It's a common pattern to use URL as the only state in those cases.

The sort of unusually looking URL format comes from the Woo component? In any case, as for the parsing, I think it would be better to parse the full path like:

```ts
const values = str.split('/')
  .map((part) => part.replace(/]$/, '').split('['))
  .reduce((map, [k, v]) => ({ ...map, [k]: v }), {});
```

To get an object like:

```ts
{
  group: "all",
  limit: "25",
  other: undefined,
  search: "asadf",
  sort_by: "updated_at",
  sort_order: "desc"
}
```

And then use that.

mailpoet/assets/js/src/segments/dynamic/store/actions.ts
```diff
@@ -224,33 +226,58 @@ export function setPreviousPage(data: string): SetPreviousPageActionType {
...
-    .fail((response) => {
-      if (response.errors.length > 0) {
-        // TODO: handle errors
+      const keys = Object.keys(response);
+      if (keys.includes('data') && keys.includes('meta')) {
+        data = response as DynamicSegmentsList;
       }
+    })
+    .catch(() => {
+      // @ToDo: show notice
```
JanJakes: A `TODO` here

mailpoet/assets/js/src/segments/dynamic/list/listing-table.tsx
```diff
@@ -108,48 +168,108 @@ export function ListingTable(): JSX.Element {
...
+                sort_by: param,
+              } as Partial<DynamicSegmentQuery>;
+              if (dynamicSegmentQuery.sort_by === param) {
+                newParams.sort_order =
+                  dynamicSegmentQuery.sort_order === 'asc' ? 'desc' : 'asc';
+              }
+              updateDynamicQuery(newParams);
+            }
+          }}
+          query={tableQueryParams}
```
JanJakes: One thing I noticed when typing into the search field is that there is no debounce and/or request aborting. Apart from firing many requests at the same time, this can probably lead to out-of-order responses displaying wrong data.

mailpoet/assets/js/src/segments/dynamic/list/get-row.tsx
```diff
@@ -2,13 +2,31 @@ import { Link } from 'react-router-dom';
...
+import { storeName } from '../store';
 
 export function getRow(dynamicSegment: DynamicSegment): object[] {
+  const toggleSelect = (): void => {
+    if (dynamicSegment?.selected) {
+      dispatch(storeName).unselectDynamicSection(dynamicSegment);
+      return;
+    }
+    dispatch(storeName).selectDynamicSection(dynamicSegment);
+  };
```
JanJakes, line 50: I'm not fully sure if selection belongs to the store, but let's see later commits...

mailpoet/assets/js/src/segments/dynamic/list/bulk-actions.tsx
```diff
@@ -0,0 +1,202 @@
...
+    action: 'bulk_action',
+    data: {
+      action,
+      listing: {
+        selection: segments.map((segment) => segment.id),
+      },
+    },
+  }).then(() => {
+    void dispatch(storeName).loadDynamicSegments();
+  });
```
JanJakes: We should probably prefer `async`/`await`. Here it would have the side effect of revealing a potential request error here and not somewhere else. And yeah, errors are probably not handled here, but maybe they don't need to be, not sure.

mailpoet/assets/js/src/segments/dynamic/list/bulk-actions.tsx
```diff
@@ -0,0 +1,202 @@
...
+      <p>{message}</p>
+      <Button variant="primary" onClick={onConfirm}>
+        {confirmText}
+      </Button>
+      <Button variant="tertiary" onClick={onClose}>
+        {__('Cancel', 'mailpoet')}
+      </Button>
+    </Modal>
+  );
+}
```
JanJakes: In some places, we're using `import { __experimentalConfirmDialog as ConfirmDialog } from '@wordpress/components';`, but I guess no big deal to keep it as it is now.

mailpoet/assets/js/src/segments/dynamic/list/bulk-actions.tsx
```diff
@@ -0,0 +1,202 @@
...
+      title = _n(
+        'Trash selected segment',
+        'Trash selected segments',
+        allSelected.length,
+        'mailpoet',
+      );
+      message = sprintf(
+        _n(
+          'Are you sure you want to trash the selected segment %s?',
+          'Are you sure you want to trash the selected segments %s and %s?',
```
JanJakes: A small improvement: `%s and %s` => `%s, and %s` (+ for all other messages below)

mailpoet/assets/js/src/segments/dynamic/list/bulk-actions.tsx
```diff
@@ -0,0 +1,202 @@
...
+    allSelected.length > 0
+      ? `"${allSelected[allSelected.length - 1].name}"`
+      : '';
+  const firstSelected =
+    allSelected.length > 1
+      ? allSelected
+          .slice(0, -1)
+          .map((segment) => `"${segment.name}"`)
+          .join(', ')
+      : lastSelected;
```
JanJakes: Generally, for these things there is the [Intl.ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat), but I guess no need to necessarily change now.

```ts
const list = new Intl.ListFormat(language, { style: 'long', type: 'conjunction'})
  .format(allSelected.map((value) => `"${value}"`));
```

mailpoet/assets/js/src/segments/dynamic/list/get-row.tsx
JanJakes: Update: I found the confirmation modal using this, I guess it makes sense.

mailpoet/assets/js/src/segments/dynamic/list/action-confirm.tsx
```diff
@@ -0,0 +1,194 @@
...
+  })
+    .then(() => {
+      void dispatch(noticesStore).createSuccessNotice(successMessage);
+      void dispatch(storeName).loadDynamicSegments();
+    })
+    .fail(() => {
+      void dispatch(noticesStore).createErrorNotice(errorMessage, {
+        explicitDismiss: true,
+      });
+    });
```
JanJakes: Could we use `async`/`await`?

mailpoet/assets/js/src/segments/dynamic/list/get-row.tsx
```diff
@@ -79,7 +114,13 @@ export function getRow(dynamicSegment: DynamicSegment): object[] {
           >
             {__('Edit', 'mailpoet')}
           </Button>
-          <Icon icon={moreVertical} />
+          <DropdownMenu
+            className="mailpoet-automation-listing-more-button"
```
JanJakes: `mailpoet-automation-listing-...` — "automation" :)

mailpoet/assets/js/src/segments/dynamic/list/get-row.tsx
```diff
@@ -3,11 +3,37 @@ import { DynamicSegment, DynamicSegmentAction } from 'segments/types';
...
+        newSegment.name,
+      );
+      void dispatch(noticesStore).createSuccessNotice(successMessage);
+      void dispatch(storeName).loadDynamicSegments();
+    })
+    .catch((error: { errors: { error: string; message: string }[] }) => {
+      error.errors.map(
+        (e) => void dispatch(noticesStore).createErrorNotice(e.message),
+      );
+    });
```
JanJakes, line 22: Just noticing the then/catch vs. async/await again.

JanJakes (commented):
Sorry for possibly a bit messy/outdated comments. The PR contains quite a bit of refactoring of logic written earlier, so some of my comments might be a bit outdated, before I got to the end of the review.

mailpoet/assets/js/src/segments/dynamic/list/listing-helpers.ts
```diff
@@ -25,33 +25,18 @@ export function updateDynamicQuery(values: Partial<DynamicSegmentQuery>): void {
...
-      }
+
+  pathElements.forEach((element) => {
+    const match = element.match(/(\w+)\[(.*?)]/);
+    if (match) {
+      const [, key, value] = match;
+      query[key] = integerKeys.includes(key) ? parseInt(value, 10) : value;
     }
-  }
+  });
```
JanJakes: I see this was simplified. I think I suggested even a simpler approach [in my comment](https://github.com/mailpoet/mailpoet/pull/5335#discussion_r1497114432), but not sure if it does the same as this one.

mailpoet/assets/js/src/segments/dynamic/list/listing-table.tsx
JanJakes: Update: I think this was done later.

mailpoet/assets/js/src/segments/dynamic/list/bulk-actions.tsx
JanJakes: Update: I think this was done later.

johnolek:
Thanks as always for your thorough review @JanJakes! Sorry it's taken me some time to get back around to this PR. I've been working on implementing some fixes and I'm hoping to have everything wrapped up tomorrow.

JanJakes:
I added a series of commits starting from https://github.com/mailpoet/mailpoet/pull/5335/commits/09cf50d7d7e5094a2a2a451baa02b09f5864e15b, including:
- reuse existing header component
- unify designs of automation and segment listing
- reuse more styles between automation and segment listing
- use `Intl.ListFormat`
- do not store listing parameters in both URL and Redux store (use URL only)
- debounce search queries
- improve display on smaller screens
- abort listing requests when quickly changing listing query
- fix flaky tests

JanJakes:
@websupporter You're actually the original author, so I can't make you a reviewer. So I only assigned it to you. If you're too busy, don't hesitate to reassign.

websupporter:
LGTM. Thank you @JanJakes

Aschepikov:
Hi @JanJakes @websupporter  I am not sure who will continue with this pr
1. there is an issue with the raws per page, if we choose another number like 50 75 100 there will be an issue that the raws per page still will be 25 but the list of segments will be longer
image
image
image

2. Modified column sorting doesn't work 
image
image

3. There is a 0  with an add a condition button
image

JanJakes:
@Aschepikov I'll check it out and will fix the issues.

JanJakes:
@Aschepikov

1. Fixed: https://github.com/mailpoet/mailpoet/pull/5335/commits/f7cee95295cd7d409d3ef1bc65e6f216edb2a1dd
2. Works fine for me, how would I replicate the issue?
3. Good catch, this is unrelated, also on `trunk`, I will create a ticket so we can move on with this.

Screenshot 2024-04-22 at 14 10 39
Screenshot 2024-04-22 at 14 10 43

Aschepikov (approved):
GJ! @JanJakes Yep I checked it on a new install and looks like it is connected with the template data

PR #5346, opened 2023-12-18
Update email block core/paragraph renderer [MAILPOET-5642]

lysyjan:
## Description

We had a simple core/paragraph block render that didn't support all possible configurations. In this renderer iteration, we add the possibility of text alignment and rendering compatibility across email clients.

## Code review notes

_N/A_

## QA notes

QA could be skipped here, I would let the decision on the reviewer.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5642]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5642]: https://mailpoet.atlassian.net/browse/MAILPOET-5642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
@@ -39,10 +39,7 @@ class BlocksWidthPreprocessor implements Preprocessor {
       $modifiedLayoutStyles['padding']['left'] = $block['attrs']['style']['spacing']['padding']['left'] ?? '0px';
       $modifiedLayoutStyles['padding']['right'] = $block['attrs']['style']['spacing']['padding']['right'] ?? '0px';
 
-      // Set current block values and reassign it to $parsedBlocks, but don't set width for blocks that are not supposed to have it
-      if (!in_array($block['blockName'], self::BLOCKS_WITHOUT_WIDTH, true)) {
-        $block['email_attrs']['width'] = "{$width}px";
-      }
+      $block['email_attrs']['width'] = "{$width}px";
       $block['innerBlocks'] = $this->preprocess($block['innerBlocks'], $modifiedLayoutStyles);
       $parsedBlocks[$key] = $block;
     }
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
@@ -7,11 +7,6 @@ namespace MailPoet\EmailEditor\Engine\Renderer\Preprocessors;
  * The final width in pixels is stored in the email_attrs array because we would like to avoid changing the original attributes.
  */
 class BlocksWidthPreprocessor implements Preprocessor {
-
-  const BLOCKS_WITHOUT_WIDTH = [
-    'core/paragraph',
-  ];
-
   public function preprocess(array $parsedBlocks, array $layoutStyles): array {
     foreach ($parsedBlocks as $key => $block) {
       // Layout width is recalculated for each block because full-width blocks don't exclude padding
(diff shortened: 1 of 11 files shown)


mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Paragraph.php
```diff
@@ -28,20 +44,78 @@ private function getBlockWrapper(array $parsedBlock, SettingsController $setting
...
+
+    // using custom rules because colors do not automatically resolve to hex value
+    $supportedColorValues = ['backgroundColor', 'textColor'];
+    foreach ($supportedColorValues as $supportedColorValue) {
+      if (array_key_exists($supportedColorValue, $attrs)) {
+        $colorKey = $attrs[$supportedColorValue];
+
+        $cssString = $availableStylesheets ?? '';
+
+        $colorRegex = "/--wp--preset--color--$colorKey: (#[0-9a-fA-F]{6});/";
```
costasovo: Why are we adding `--wp--preset--color--` here? Is that a convention for specifying colors in block themes? Is it documented somewhere? I think it would be helpful to have a comment explaining this here. Or maybe a comment describing a bit higher level how we get the color values. 🙏

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Paragraph.php
```diff
@@ -28,20 +44,78 @@ private function getBlockWrapper(array $parsedBlock, SettingsController $setting
...
+      }
+    }
+
+    return $styles;
+  }
+
+  /**
+   * @param array{tag_name: string, class_name?: string} $tag
+   */
+  private function removePaddingFromElement($blockContent, array $tag) {
```
costasovo: I guess we could add a return typehit here

mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
```diff
@@ -34,10 +34,7 @@ public function preprocess(array $parsedBlocks, array $layoutStyles): array {
       $modifiedLayoutStyles['padding']['left'] = $block['attrs']['style']['spacing']['padding']['left'] ?? '0px';
       $modifiedLayoutStyles['padding']['right'] = $block['attrs']['style']['spacing']['padding']['right'] ?? '0px';
 
-      // Set current block values and reassign it to $parsedBlocks, but don't set width for blocks that are not supposed to have it
-      if (!in_array($block['blockName'], self::BLOCKS_WITHOUT_WIDTH, true)) {
```
costasovo, line 43: It seems that this change made BLOCKS_WITHOUT_WIDTH constant unused. It would be better to remove it.

costasovo (changes requested):
@lysyjan Thank you for adding the support for the paragraph block!

I added a couple of comments. Just details.

Btw. I noticed that the formatting options for the paragraph in the email editor are limited compared to the formatting options in the post editor. I think it makes sense that we can't support all the options for emails. But maybe we could add support at least for `Highlight` because I find this one quite usable. What do you think? It would be an extra ticket...
Screenshot 2024-01-03 at 16 07 14

lysyjan:
@costasovo Thank you for your review. I addressed all your remarks in the three new commits.

> Btw. I noticed that the formatting options for the paragraph in the email editor are limited compared to the formatting options in the post editor. I think it makes sense that we can't support all the options for emails. But maybe we could add support at least for Highlight because I find this one quite usable. What do you think? It would be an extra ticket...

I agree with you. [By the comment in the file](https://github.com/mailpoet/mailpoet/blob/4567c6d3c1a1564e6de355d634576ac40cbe7dcb/mailpoet/assets/js/src/email-editor/engine/blocks/core/rich-text.ts#L17-L18) it looks we want to add the support later. So, creating a follow-up ticket makes sense to me.

costasovo (approved):
Thanks for the changes and the comment! The comment will help me to recall why we use the prefix in the future.

PR #5351, opened 2023-12-20
Update Gutenberg packages [MAILPOET-5714]

costasovo:
## Description
This PR updates Gutenberg packages to the versions tagged `wp-6.4` as well as updates of a few packages mentioned in the Dependabot alerts.

## Code review notes
- most of the commits are related to fixing Typescript errors. I'm not sure I addressed all the issues in the best possible way. I'm grateful for any feedback.
- after I fixed all the issues, I faced failing `build_release_zip` jobs. To avoid using a larger instance, I configured build.sh to skip building tests.
- please check also the [PR in the premium plugin](https://github.com/mailpoet/mailpoet-premium/pull/834)

## QA notes

This update affects multiple parts of the plugin as we are using wordpress/components more and more.
The parts that use the Gutenberg packages the most heavily are:

- Form editor
- Automations (all parts)
- Dynamic segments editor
- New email editor
- Coupons block in the current email editor

Please test these parts thoroughly.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/834

## Linked tickets

[MAILPOET-5714]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5714]: https://mailpoet.atlassian.net/browse/MAILPOET-5714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/editor/components/header/inserter-toggle.tsx
```diff
@@ -1,9 +1,17 @@
...
+import { Button, ToolbarItem as WpToolbarItem } from '@wordpress/components';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
 import { plus } from '@wordpress/icons';
 import { storeName } from '../../store';
 
+// Build type for ToolbarItem contains only "as" and "children" properties but it takes all props from
+// component passed to "as" property (in this case Button). So as fix for TS errors we need to pass all props from Button to ToolbarItem.
+// We should be able to remove this fix when ToolbarItem will be fixed in Gutenberg.
+const ToolbarItem = WpToolbarItem as React.ForwardRefExoticComponent<
```
JanJakes, line 10: I only wonder what exactly is `ForwardRefExoticComponent` and where it comes from. Did you use it because it was in the original type? Or is it the only variant that works?

mailpoet/assets/js/src/automation/editor/components/actions/trash-button.tsx
```diff
@@ -1,12 +1,16 @@
...
-  Button,
-} from '@wordpress/components';
+import { __experimentalConfirmDialog, Button } from '@wordpress/components';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { __, sprintf } from '@wordpress/i18n';
 import { storeName } from '../../store';
 
+// With __experimentalConfirmDialog's type from build-types Typescript complains:
+// JSX element type __experimentalConfirmDialog does not have any construct or call signatures
+// Wrapping the type to React.FC fixes the issue
```
JanJakes, line 9: I wonder if this was actually useful in some form: https://github.com/mailpoet/mailpoet/pull/5351/commits/d66769e80ed93ac5570c223961682e225e5fa8d9#diff-f121ab46250da530ff0e99594c24c398d2c4527aeab72d5b9595b8ed7e3fc28bL39

mailpoet/assets/css/src/components-automation-editor/_dropdown.scss
```diff
@@ -1,3 +1,39 @@
+.mailpoet-automation-editor-edit-site-document-actions {
```
JanJakes, line 1: Here I'm wondering:

1. Should we use the new look to align with Gutenberg, or it doesn't work in this case?
2. Should we create our custom component for this?

mailpoet/webpack.config.js
```diff
@@ -452,7 +452,9 @@ const emailEditorCustom = Object.assign({}, wpScriptConfig, {
     ...wpScriptConfig.resolve,
     modules: ['node_modules', 'assets/js/src'],
   },
-  plugins: [...wpScriptConfig.plugins, ...[new ForkTsCheckerWebpackPlugin()]],
+  plugins: PRODUCTION_ENV
+    ? wpScriptConfig.plugins
+    : [...wpScriptConfig.plugins, ...[new ForkTsCheckerWebpackPlugin()]],
```
JanJakes: Very unimportant, but this is the same: `[...wpScriptConfig.plugins, new ForkTsCheckerWebpackPlugin()]`

Or maybe it was meant to be like this?

```ts
plugins: [...wpScriptConfig.plugins, ...(PRODUCTION_ENV ? [new ForkTsCheckerWebpackPlugin()] : [])],
```

Then the spread would make sense...

mailpoet/build.sh
```diff
@@ -15,7 +15,7 @@ mkdir $plugin_name
 echo '[BUILD] Generating production CSS and JS assets'
 rm -rf node_modules
 pnpm install --frozen-lockfile --prefer-offline
-./do compile:all --env production
+./do compile:all --env production --tests skip
```
JanJakes: Should we maybe use `--skip-tests` or `--skip-tests-build` to make it more unified with `--skip-deps` for tests?

JanJakes: I also wonder if it could work based on `env`. Do we ever build tests with `production` env?

mailpoet/assets/js/src/types/index.ts
```diff
@@ -1,9 +1,7 @@
-import { ComponentPropsWithoutRef } from 'react';
```
JanJakes: Related to this commit, https://github.com/mailpoet/mailpoet/pull/5351/commits/6d842ec0f0342bde89d01de78bacb345b7bdfd64 doesn't remove `@types/wordpress__components`. Maybe we discussed this, but are now both needed?

mailpoet/assets/css/src/components-automation-editor/_dropdown.scss
JanJakes: I think previously, it looked a bit different.

Now:
Screenshot 2024-01-17 at 12 34 07

Before:
Screenshot 2024-01-17 at 12 35 54

mailpoet/RoboFile.php
```diff
@@ -202,7 +202,7 @@ public function compileJs($opts = ['env' => null]) {
     $env = ($opts['env']) ?
       sprintf('./node_modules/.bin/cross-env NODE_ENV="%s"', $opts['env']) :
       null;
-    return $this->_exec($env . ' ./node_modules/webpack/bin/webpack.js');
+    return $this->_exec($env . ' ./node_modules/webpack/bin/webpack.js --env BUILD_TESTS=' . ($opts['tests'] === 'build' ? 'build' : 'skip'));
```
JanJakes: On my machine, `./do compile:js` now fails (`./do compile:all` works):

Screenshot 2024-01-17 at 12 37 45

mailpoet/assets/css/src/mailpoet-plugin.scss
```diff
@@ -18,6 +18,11 @@
 @import 'mixins/clearfix';
 @import 'mixins/full-width-background';
 
+// WordPress Admin Theme Color Variables
+// Needed for the WordPress Admin color scheme support for @wordpress/components.
+@import '../../../node_modules/@wordpress/base-styles/mixins';
+@include wordpress-admin-schemes();
+
```
JanJakes: I'm still seeing strange colors on the homepage:

Screenshot 2024-01-17 at 12 39 04

I installed all deps and compiled all assets, caching in browser disabled (dev mode).

mailpoet/assets/js/src/automation/integrations/mailpoet/components/button.tsx
```diff
@@ -1,13 +1,27 @@
...
+  | WordPressComponentProps<
+      Omit<ButtonAsButtonProps, keyof ExtendedProps> & ExtendedProps,
+      'button',
+      false
+    >
+  | WordPressComponentProps<
+      Omit<ButtonAsAnchorProps, keyof ExtendedProps> & ExtendedProps,
+      'a',
+      false
+    >;
```
JanJakes: The issue here is that the Button props are a union of multiple types. We need to apply the modifications (omit + add new props) to all the union members. That's more universally doable using [distributive conditional types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types):

```ts
type UpdateUnionProps<T, U> = T extends unknown ? Omit<T, keyof U> & U : never;

type Props = UpdateUnionProps<
  React.ComponentProps<typeof WpButton>,
  ExtendedProps
>;
```

JanJakes (commented):
Great job! I have only a few comments.

mailpoet/assets/js/src/automation/editor/components/header/inserter-toggle.tsx
costasovo: @JanJakes I had quite a hard time with some of those Typescript errors. I'm not sure I remember exactly, but in this case, I think yes I used it because it was in the original build types. From that, I saw that I could add another component, and it would accept props. In this place, I did it only for Button.

`/node_modules/@wordpress/components/build-types/toolbar/toolbar-item/index.d.ts`

```ts
/// <reference types="react" />
declare const _default: import("react").ForwardRefExoticComponent<Omit<import("react").HTMLAttributes<any>, "children"> & {
    as?: import("react").ElementType | undefined;
    children?: import("react").ReactNode | ((props: import("react").HTMLAttributes<any> & import("react").RefAttributes<any>) => import("react").ReactElement<any, string | import("react").JSXElementConstructor<any>> | null);
} & import("react").RefAttributes<any>>;
export default _default;
//# sourceMappingURL=index.d.ts.map
```

mailpoet/assets/js/src/automation/editor/components/actions/trash-button.tsx
costasovo: I think that after the update, these were ignored because TS took the definition from build types.

mailpoet/assets/css/src/components-automation-editor/_dropdown.scss
costasovo: @JanJakes, I'll update CSS  so that the result is closer to the original.

> Here I'm wondering:
> 
> Should we use the new look to align with Gutenberg, or it doesn't work in this case?
> Should we create our custom component for this?

Good questions. 

The new Gutenberg design is something completely different. Click on the component triggers the command palette.

Screenshot 2024-01-19 at 15 26 08

As I see it, what we currently have is not aligned with Gutenberg (design or purpose), so we should create a custom component. Or even discuss with the design team if we still want to use that area (center of the header) for this component as it is now not aligned with Gutenberg.

Do you want me to create the component within this PR? I would suggest going with the CSS fix for now so that we no longer block this PR and I would prepare a ticket about refactoring the component. Would you be ok with that?

mailpoet/build.sh
costasovo: I like the idea of unifying the naming! 

As for making it dependent on production I saw that [on CI for the JS tests we use the production env](https://github.com/mailpoet/mailpoet/blob/db9813a3aac5e1ebfd4fd3d692ab54f0ed95bb5f/.circleci/config.yml#L157). When I searched node_modules for `process.env.NODE_ENV` I saw that it is used in many cases to check if the value is `production` so I assume the production environment has additional effects. So I didn't want to change that we run the tests build with the production settings.

mailpoet/assets/js/src/types/index.ts
costasovo: @JanJakes Ha. Thanks for mentioning the `@types/wordpress__components` I see I haven't removed the package. I skipped removing it because it completely broke the premium build. Somehow, the imports from `@wordpress/components `that are exposed from the free plugin don't work if I remove the package. I had a lot of issues to fix and completely forgot about this one. I'll try to look into it. 

Also, I don't understand what you are reffering to in "Maybe we discussed this, but are now both needed?" What is the second thing?

mailpoet/assets/js/src/automation/integrations/mailpoet/components/button.tsx
costasovo: I used your definition. I need to take a look at it again with a fresh head to fully understand how that works.

JanJakes: @costasovo It basically means "for each member of union T, do `Omit<T, keyof U>`".

costasovo:
@JanJakes I believe I addressed all your comments. I rebased the PR and added 7 new commits (beginning with https://github.com/mailpoet/mailpoet/pull/5351/commits/acab2ab80d04b06c72a2be0cfde210dbc6f96fb0). 

After I removed the `@types/wordpress__components` I needed to do additional changes in the [related premium-plugin PR](https://github.com/mailpoet/mailpoet-premium/pull/834)

JanJakes (dismissed):

Aschepikov:
Hi @costasovo I found some small UI issues, can you take a look pls
1 In the new email editor, there is no slider for the padding and no background for the button, the current prod is on the right side
image

2 In the segments, some of the conditions are too wide and overlap with the remove icon, on the prod everything is ok
image

3 In the automation analytics, there is some overlap in date range setting
image

4 Any click on the date or the month scroll closes the date popover (automation analytics)
image

5 The gradient isn't saving and after reloading the page there is no gradient (form editor)
image
image
image

costasovo:
Thank you, @Aschepikov, for the thorough review! I believe I addressed all the issues.

> 1 In the new email editor, there is no slider for the padding and no background for the button, the current prod is on the right side

I fixed that by configuring the editor to provide steps for setting padding. This will make it more comfortable for users and they still have an option to insert exact units.

> 2 In the segments, some of the conditions are too wide and overlap with the remove icon, on the prod everything is ok

Fixed by adding a CSS for the trash bin icon. The CSS was added to the premium plugin because we allow multiple conditions only with premium. So please test with the build from the premium plugin PR https://github.com/mailpoet/mailpoet-premium/pull/834

> 3,4,5 

Should be fixed.

@JanJakes I rebased the PR please check the last 6 commits.

JanJakes (dismissed):

Aschepikov:
Hi @costasovo  I am rechecking everything and I see in the automation analytics that there is a problem with step icons on the subscriber tab
image
and on the automation flow too
image
Also, I have a problem with the edit automation button. I have this on all browsers but there was a case when it wasn't reproducible on the dev PC so I am not sure about it
image

costasovo:
@Aschepikov Nice catch! It was caused by my fix of filters in Automation analytics 🤦 
@JanJakes I added one new commit with a fix. Could you please take a look? 🙏

JanJakes (dismissed):

Aschepikov:
Ho @costasovo  Great, everything looks good but I see again the 4th issue "Any click on the date or the month scroll closes the date popover (automation analytics)". So with the month scroll buttons, it is impossible to change the month as it closes after every click. The same with the date range but the changes are saving. On the prod, I don't have it. Sorry, I hope it is the last call 
image

costasovo:
@Aschepikov My bad! I missed the issue and thought all in the picker was related to the broken visual appearance.  It should be fixed now.

@JanJakes One more review, please 🙏  I added three more commits. I noticed a couple of warnings in the console on the automation stats page. There are still few warnings left. These were [fixed in woocommerce/components](https://github.com/woocommerce/woocommerce/pull/41845) but we need to update woocommerce/components.

JanJakes (approved):

Aschepikov:
HI @costasovo on the subscriber's tab in the automation analytics the icons are non-centrical in the circle
image

costasovo:
> HI @costasovo on the subscriber's tab in the automation analytics the icons are non-centrical in the circle

@Aschepikov Is this happening in a specific browser? I tested it in Chrome, Firefox, and Safari. I also tried using the zip from Circle CI, but in all cases, the icon was centered correctly.

Aschepikov:
I have it in explorer and chrome but it can be connected with the site after a lot of installs, it should be fixed with a release build as you don't have this, thank you

Aschepikov (approved):
GJ!

PR #5352, opened 2023-12-20
Shared sender domain status logic [MAILPOET-5783]

johnolek:
## Description

This PR contains logic necessary for MAILPOET-5783 as well as some of the other sender domain tickets. Since it's just the shared logic it doesn't actually complete ticket 5783 (i.e. this doesn't add any new notices). That will be handled in a separate PR.

## Code review notes

Is there a more appropriate place to put the logic for rewriting emails?

I thought about adding other helper methods but figured we could add/change as needed in other PRs. Let me know if you can think of any others we would definitely want right away though.

## QA notes

No QA should be necessary since this isn't changing any real application logic yet.

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5783

## After-merge notes

_N/A_

## Tasks

- [ ] ~I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations~
- [ ] ~I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes~

### DIFF
--- a/mailpoet/lib/Services/AuthorizedSenderDomainController.php
+++ b/mailpoet/lib/Services/AuthorizedSenderDomainController.php
@@ -104,6 +104,10 @@ class AuthorizedSenderDomainController {
     return $response;
   }
 
+  public function getRewrittenEmailAddress(string $email): string {
+    return sprintf('%s@replies.sendingservice.net', str_replace('@', '=', $email));
+  }
+
   /**
    * Verify Sender Domain
    *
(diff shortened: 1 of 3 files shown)


brezocordero (approved):
@johnolek LGTM!

Could you please check if there could be cached API responses (in the plugin or the MSS?) and if so how to invalidate it? It can be added on the second PR.
Thanks!

johnolek:
@brezocordero 

> Could you please check if there could be cached API responses (in the plugin or the MSS?) and if so how to invalidate it? 

Other than the per-request in-memory cache that stores the API response (https://github.com/mailpoet/mailpoet/pull/5352/files#diff-bb3ed7eebb1f2d8b7eb64282f525dca9707564422133e3e37b1d1813c6e0ad19R231-R237), I don't see any caching in the plugin. I also looked at the bridge code and didn't see any caching there, but I'm no expert, so we might want to double check with someone on Sonnet if this is a major concern.

Do you think this would be ok to merge in right away or is there anything else you'd like to see first? Or have tested?

PR #5353, opened 2023-12-20
Add banner notices for unverified sender domains [MAILPOET-5783]

johnolek:
## Description

This PR adds banner notices based on the user's default from address (from the settings page) and the number of contacts they have. The banners will only show if the user is using MSS for sending and meet certain conditions (see ticket for details).

## Code review notes
_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5783

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
+++ b/mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
@@ -1,0 +1,167 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Services\AuthorizedSenderDomainController;
+use MailPoet\Services\Bridge;
+use MailPoet\Settings\SettingsController;
+use MailPoet\Util\FreeDomains;
+use MailPoet\Util\Helpers;
+use MailPoet\Util\License\Features\Subscribers;
+use MailPoet\WP\Notice;
+
+class SenderDomainAuthenticationNotices {
+
+  const LOWER_LIMIT = 500;
+  const UPPER_LIMIT = 1000;
+
+  const FREE_MAIL_KB_URL = 'https://kb.mailpoet.com/article/259-your-from-address-cannot-be-yahoo-com-gmail-com-outlook-com';
+  const SPF_DKIM_DMARC_KB_URL = 'https://kb.mailpoet.com/article/295-spf-dkim-dmarc';
+
+  private SettingsController $settingsController;
+
+  private Subscribers $subscribersFeatures;
+
+  private FreeDomains $freeDomains;
+
+  private AuthorizedSenderDomainController $authorizedSenderDomainController;
+
+  private Bridge $bridge;
+
+  public function __construct(
+    SettingsController $settingsController,
+    Subscribers $subscribersFeatures,
+    FreeDomains $freeDomains,
+    AuthorizedSenderDomainController $authorizedEmailsController,
+    Bridge $bridge
+  ) {
+    $this->settingsController = $settingsController;
+    $this->subscribersFeatures = $subscribersFeatures;
+    $this->freeDomains = $freeDomains;
+    $this->authorizedSenderDomainController = $authorizedEmailsController;
+    $this->bridge = $bridge;
+  }
+
+  public function getDefaultFromAddress(): string {
+    return $this->settingsController->get('sender.address', '');
+  }
+
+  public function getDefaultFromDomain(): string {
+    return Helpers::extractEmailDomain($this->getDefaultFromAddress());
+  }
+
+  public function isFreeMailUser(): bool {
+    return $this->freeDomains->isEmailOnFreeDomain($this->getDefaultFromDomain());
+  }
+
+  public function init($shouldDisplay): ?Notice {
+    if (
+      !$shouldDisplay
+      || !$this->bridge->isMailpoetSendingServiceEnabled()
+      || in_array($this->getDefaultFromDomain(), $this->authorizedSenderDomainController->getFullyVerifiedSenderDomains(true))
+      || $this->isFreeMailUser() && $this->subscribersFeatures->getSubscribersCount() <= self::LOWER_LIMIT
+    ) {
+      return null;
+    }
+
+    return $this->display();
+  }
+
+  public function display(): Notice {
+    $contactCount = $this->subscribersFeatures->getSubscribersCount();
+    $isFreeMailUser = $this->isFreeMailUser();
+
+    $isErrorStyle = $contactCount > self::UPPER_LIMIT;
+
+    $noticeContent = $isFreeMailUser
+      ? $this->getNoticeContentForFreeMailUsers($contactCount)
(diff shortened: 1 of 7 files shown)


mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    return $this->display();
+  }
+
+  public function display(): Notice {
+    $contactCount = $this->subscribersFeatures->getSubscribersCount();
+    $isFreeMailUser = $this->isFreeMailUser();
+
+    $isErrorStyle = $contactCount > self::UPPER_LIMIT;
```
brezocordero: From the Figma file, I think that the error style does not apply to branded domains that are partially verified, independently of the subscribers count.

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
```diff
@@ -0,0 +1,167 @@
...
+      $this->getLearnMoreAboutFreeMailButton()
+    );
+  }
+
+  public function getNoticeContentForBrandedDomainUsers(bool $isPartiallyVerified, int $contactCount): string {
+    if ($isPartiallyVerified || $contactCount <= self::LOWER_LIMIT) {
+      // translators: %1$s is HTML for an 'authenticate domain' button, %2$s is HTML for a Learn More button
+      return sprintf(__("<strong>Authenticate your sender domain to improve email delivery rates.</strong>
+<span>Major mailbox providers require you to authenticate your sender domain to confirm you sent the emails, and may place unauthenticated emails in the \"Spam\" folder. Please authenticate your sender domain to ensure your marketing campaigns are compliant and will reach your contacts.</span><p>%1\$s &nbsp; %2\$s</p>", 'mailpoet'),
+        $this->getAuthenticateDomainButton(),
```
brezocordero, line 154: From the Figma file, for partially verified domains, the button would always be secondary.
You might want to pass `$isPartiallyVerified` to this function to generate the class.

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
```diff
@@ -0,0 +1,167 @@
...
+  public function display(): Notice {
+    $contactCount = $this->subscribersFeatures->getSubscribersCount();
+    $isFreeMailUser = $this->isFreeMailUser();
+
+    $isErrorStyle = $contactCount > self::UPPER_LIMIT;
+
+    $noticeContent = $isFreeMailUser
+      ? $this->getNoticeContentForFreeMailUsers($contactCount)
+      : $this->getNoticeContentForBrandedDomainUsers(in_array($this->getDefaultFromDomain(), $this->authorizedSenderDomainController->getPartiallyVerifiedSenderDomains(true)), $contactCount);
+
```
brezocordero, line 102: This is my mistake because the specs say that the notices are non dismissible but there is a close icon on the designs.
Could you please add the `$extraClasses` for that?

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  public function isFreeMailUser(): bool {
+    return $this->freeDomains->isEmailOnFreeDomain($this->getDefaultFromDomain());
+  }
+
+  public function init($shouldDisplay): ?Notice {
+    if (
+      !$shouldDisplay
+      || !$this->bridge->isMailpoetSendingServiceEnabled()
```
brezocordero, line 84: The banners should not be displayed for new users.
I might have missed it but I don't seem to find where the condition is implemented?

brezocordero (changes requested):
@johnolek Nice job!

I don't think we need to extract the common strings, I agree with you that it could make the code harder to read and also make it harder for the translators.

Let me know if any of the requested changes are not clear or if I missed something.

Thanks!

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
johnolek: How are we defining new users?  `$shouldDisplay` is defined here, similar to many other notices: https://github.com/mailpoet/mailpoet/blob/MAILPOET-5783%2Fsender-domain-banner-notices/mailpoet/lib/Util/Notices/PermanentNotices.php#L147-L147

It should prevent the notice from being displayed during the welcome wizard. Is there a different check we should include?

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
johnolek: Do we want them to stay dismissed? I'm guessing not but just want to confirm. 

I added the classes in c8e21b4d9ecb314e792c548237a170dd8d9d4f74

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
johnolek: Good catch! I think this should be resolved in 14bb974f1b387c898bd1eaa02cca1c1ae8ed1a53

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
johnolek: Another good catch! This should also be resolved in 14bb974f1b387c898bd1eaa02cca1c1ae8ed1a53.

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
brezocordero: > How are we defining new users?

The idea is to show the banners only to users that were already sending emails before the new requirements were applied.

New users will see the onboarding task and the inline notices when the start creating emails but existing users might not go inside the editor or to the settings page, so we need to make them aware of the changes.

New users are users that have not sent an email yet.

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
brezocordero: > I'm guessing not

Same. Maybe after a few months we could make them dismissible but for now we want to be sure that users know how their emails are being impacted.

brezocordero (changes requested):
@johnolek 
Thanks for making the changes! 

The only thing left is showing the banners only to users already sending emails.

mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
johnolek: @brezocordero 3db27727fe21715d323c8e1fb08d6224206f1814 should ensure that we don't display these notices for new users.

johnolek:
@brezocordero Sorry for the delay. This should be ready to review again.

brezocordero (approved):
Thanks for the update.
The code looks good, I like how you implemented the new user condition.

veljkho (approved):

PR #5355, opened 2024-01-03
Email editor - button and buttons blocks [MAILPOET-5644]

costasovo:
## Description
This PR adds support for buttons and button blocks to the email editor.

## Code review notes
- colors set via CSS class don't work in renderer (we have an extra ticket for that 5741)

## QA notes
I would suggest skipping QA for now and doing QA after we finish bug fixes and refactoring the renderer.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5644]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5644]: https://mailpoet.atlassian.net/browse/MAILPOET-5644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
@@ -19,6 +19,8 @@ class BlocksWidthPreprocessor implements Preprocessor {
       }
 
       $widthInput = $block['attrs']['width'] ?? '100%';
+      // Currently we support only % and px units in case only the number is provided we assume it's %
+      // because editor saves percent values as a number.
       $widthInput = is_numeric($widthInput) ? "$widthInput%" : $widthInput;
       $width = $this->convertWidthToPixels($widthInput, $layoutWidth);
 
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
@@ -18,7 +18,9 @@ class BlocksWidthPreprocessor implements Preprocessor {
         $layoutWidth -= $this->parseNumberFromStringWithPixels($layoutStyles['padding']['right'] ?? '0px');
       }
 
-      $width = $this->convertWidthToPixels($block['attrs']['width'] ?? '100%', $layoutWidth);
+      $widthInput = $block['attrs']['width'] ?? '100%';
+      $widthInput = is_numeric($widthInput) ? "$widthInput%" : $widthInput;
+      $width = $this->convertWidthToPixels($widthInput, $layoutWidth);
 
       if ($block['blockName'] === 'core/columns') {
         // Calculate width of the columns based on the layout width and padding
--- a/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
+++ b/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
@@ -46,41 +46,41 @@ class Button implements BlockRenderer {
 
     // Styles attributes
     $wrapperStyles = [
-      "background: $bgColor",
-      'cursor: auto',
-      'word-break: break-word',
-      'box-sizing: border-box',
+      'background' => $bgColor,
+      'cursor' => 'auto',
+      'word-break' => 'break-word',
+      'box-sizing' => 'border-box',
     ];
     $linkStyles = [
-      'display: block',
-      'line-height: 120%',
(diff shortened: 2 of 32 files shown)


mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
```diff
@@ -23,7 +23,9 @@ public function preprocess(array $parsedBlocks, array $layoutStyles): array {
         $layoutWidth -= $this->parseNumberFromStringWithPixels($layoutStyles['padding']['right'] ?? '0px');
       }
 
-      $width = $this->convertWidthToPixels($block['attrs']['width'] ?? '100%', $layoutWidth);
+      $widthInput = $block['attrs']['width'] ?? '100%';
+      $widthInput = is_numeric($widthInput) ? "$widthInput%" : $widthInput;
```
lysyjan, line 24: I guess I understand that this is a necessary check, but maybe a comment about the reasons would be more transparent for other developers. If I am not mistaken, the reason is the different width values of some blocks, such as buttons.
What do you think about a comment here?

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
```diff
@@ -28,7 +28,13 @@ public function render($blockContent, array $parsedBlock, SettingsController $se
     $markup = str_replace('{linkUrl}', $buttonLink->getAttribute('href') ?: '#', $markup);
 
     // Width
-    $markup = str_replace('{width}', $parsedBlock['email_attrs']['width'] ?? '100%', $markup);
+    // Parent block prepares container with proper width. If the width is set let's use full width of the container
+    // otherwise let's use auto width.
+    $width = 'auto';
+    if (($parsedBlock['attrs']['width'] ?? null)) {
```
lysyjan: Using the null coalescing operator in this condition doesn't make sense to me. I would say `isset` could work better here. But maybe I'm missing something. What do you think about it? There are more similar cases.

mailpoet/tests/integration/EmailEditor/Integration/Core/Renderer/Blocks/ButtonTest.php
```diff
@@ -0,0 +1,164 @@
...
+    verify($output)->stringContainsString('border-radius:10px;');
+  }
+
+  public function testItRendersCornerSpecificBorderRadius() {
+    $this->parsedButton['attrs']['style']['border']['radius'] = [
+      'topLeft' => '1px',
+      'topRight' => '2px',
+      'bottomLeft' => '3px',
+      'bottomRight' => '4px',
+      ];
```
lysyjan: This is just a small remark because the indent is incorrect here.
Also, I have noticed that all methods don't have a return type. Personally, I try to use void in tests, but maybe it's only a personal preference.

lysyjan (commented):
A huge piece of work! Thanks for many fixes and improvements 👍 

During my brief testing, I noticed that the browser preview contains an unexpected gap when there are other blocks around.

Editor:
Screenshot 2024-01-04 at 15 50 26

Preview:
Screenshot 2024-01-04 at 15 50 50

I would say it's caused by setting font-size on a different HTML element than is needed.

mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/BlocksWidthPreprocessor.php
costasovo: 💯 I was thinking about adding a comment there but forgot. Thanks for the heads up!

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
costasovo: Oh yes. I'll use isset. I don't know why I used the operator.

lysyjan:
@costasovo I almost forgot about another idea here. When I was trying to use the button block, a missing feature was the vertical spacing configuration. Have you considered something like that? The post editor allows setting vertical margins. I know about a possible workaround when I wrap buttons by columns, but something like that would be more straightforward.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
lysyjan: I would say gradual code modifications and looking for a working solution caused this.

costasovo:
@lysyjan Thank you for the review! I addressed all your suggestions in the new five commits.

lysyjan (approved):
Thank you! Great job

PR #5362, opened 2024-01-09
Mailpoet 5782 sender inline notices

brezocordero:
## Description

Display Sender Domain notices in Settings and Send screens
Automation notices will be done in a separate ticket

## Code review notes

_N/A_

## QA notes

For new installations:
Please test the different notices from the Figma file linked in the ticket.
For old installations:
Please see the update 2024-01-10 in the ticket

Note: You can change the LOWER_LIMIT on file `mailpoet/lib/Services/AuthorizedSenderDomainController.php` to test the different cases.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5782]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5769]: https://mailpoet.atlassian.net/browse/MAILPOET-5769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-5782]: https://mailpoet.atlassian.net/browse/MAILPOET-5782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brezocordero:
@johnolek reassigning this back to me to implement the new changes
I will reassign it back when done.
Thanks!

mailpoet/assets/js/src/common/sender-domain-notice/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+          rewrittenFromEmail: <strong>{escapeHTML(rewrittenEmail)}</strong>,
+        },
+      )}
+    </p>
+  );
+}
+
+function SenderDomainInlineNotice({
+  emailAddress,
+  emailAddressDomain,
```
johnolek: Not a big deal, but would it be simpler to have this component extract the domain itself? Or is there actually a case where the emailAddress and emailAddressDomain would have different domains?

mailpoet/assets/js/src/common/sender-domain-notice/sender-domain-notice-body.tsx
```diff
@@ -0,0 +1,62 @@
...
+  emailAddressDomain: string;
+  isFreeDomain: boolean;
+  isPartiallyVerifiedDomain: boolean;
+  isSmallSender: boolean;
+  onlyShowWarnings?: boolean;
+}) {
+  const renderMessage = (messageKey: string) => {
+    const messages: { [key: string]: string } = {
+      freeSmall:
+        "Shared 3rd-party domains like <emailDomain/> will send from MailPoet's shared domain. We recommend you to use your site's branded domain instead.",
```
johnolek: The wording is just a little awkward to my ear. I would recommend this small change:
```suggestion
        "Shared 3rd-party domains like <emailDomain/> will send from MailPoet's shared domain. We recommend that you use your site's branded domain instead.",
```

mailpoet/assets/js/src/newsletters/send/sender-address-field.jsx
```diff
@@ -80,12 +80,16 @@ class SenderField extends Component {
 
     if (!isFieldValid(this.domElementSelector)) {
       validateField(this.domElementSelector);
+      // Hide domain error to not stack error messages
+      this.showSenderDomainError(false);
```
johnolek: I haven't tested it, but just looking at the code, is this doing what you want? It looks like `showSenderDomainError` can only ever set the state to true.

```js
  showSenderDomainError = (status) => {
    if (!status) return;

    this.setState({ showSenderDomainWarning: true });
  };
```

johnolek (changes requested):
Nice work on this, @brezocordero!

I left a few comments, but the only thing I actually care about changing is tweaking the wording for one of the messages: https://github.com/mailpoet/mailpoet/pull/5362/files#r1449117147

There also appear to be some conflicts, probably due to merging/rebasing https://github.com/mailpoet/mailpoet/pull/5365. 

I'm not suggesting we do this as part of this PR, but it looks like we might be able to remove some now-unused code related to direct checking of dmarc policies, i.e. I don't think [checkSenderEmailDomainDmarcPolicy](https://github.com/mailpoet/mailpoet/blob/efe4c475684df0067e4ce3d94b10cf88166113e3/mailpoet/assets/js/src/common/check-sender-domain-dmarc-policy.ts#L25-L25) is being used anymore. Maybe we can write a cleanup ticket for that.

Thank you by the way for extracting the sender domain notice logic into the controller. I should have put it there from the beginning!

mailpoet/assets/js/src/newsletters/send/sender-address-field.jsx
brezocordero: Fixed! Thanks for catching that 
image

mailpoet/assets/js/src/common/sender-domain-notice/sender-domain-notice-body.tsx
brezocordero: Changed

brezocordero:
> I don't think [checkSenderEmailDomainDmarcPolicy](https://github.com/mailpoet/mailpoet/blob/efe4c475684df0067e4ce3d94b10cf88166113e3/mailpoet/assets/js/src/common/check-sender-domain-dmarc-policy.ts#L25-L25) is being used anymore. Maybe we can write a cleanup ticket for that.

I removed it, thanks for noticing.

mailpoet/assets/js/src/common/sender-domain-notice/index.tsx
brezocordero: Done

mailpoet/assets/js/src/newsletters/send/sender-address-field.jsx
```diff
@@ -95,7 +95,7 @@ class SenderField extends Component {
   showSenderDomainError = (status) => {
     if (!status) return;
 
-    this.setState({ showSenderDomainWarning: true });
+    this.setState({ showSenderDomainWarning: status });
```
johnolek: I think this will still only work when `status` is `true` due to the early return above.

johnolek (dismissed):
Thanks for all those updates, especially the cleanup of the dmarc code!

Is there a reason you left out this commit from the final PR https://github.com/mailpoet/mailpoet/commit/908b99e8f358b742bef2a0d05141db61133b70a3? It was actually something that I thought about bringing up myself.

I left [one last comment](https://github.com/mailpoet/mailpoet/pull/5362/commits/ef00c215b53d05b7215f9211156a27039e5a13d3#r1449469229) about the `showSenderDomainError` method, but I'm assuming it doesn't really matter, or I'm misunderstanding it somehow, so I'm approving the PR but feel free to modify if you like.

brezocordero:
> Is there a reason you left out this commit from the final PR https://github.com/mailpoet/mailpoet/commit/908b99e8f358b742bef2a0d05141db61133b70a3? It was actually something that I thought about bringing up myself.

I brought it back. I removed it because I thought it was too much to add here.

> I'm approving the PR but feel free to modify if you like

I made both modifications. Thanks again for the review.

mailpoet/assets/js/src/newsletters/send/sender-address-field.jsx
brezocordero: I don't know what I was thinking...
I rewrote the commit.
I left this function as it was, and removed the `showSenderDomainError(false)` that was actually not needed.

johnolek (approved):

Aschepikov (approved):
GJ!

PR #5364, opened 2024-01-10
Move filter adding MailPoet image resolution [MAILPOET-5749]

lysyjan:
## Description

_N/A_

## Code review notes

- From my understanding, moving the filter that adds MailPoet image size to the NewsletterEditor page shouldn't break anything.

## QA notes

1. Create a newsletter containing at least one image and Post with the featured image. All photos should have a width higher than 1320px.
2. Send a preview and check that all images in the email are with MailPoet resolution (width 1320px)
3. Send the newsletter and check again the image sizes.
4. Test it for all email types

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5749]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5749]: https://mailpoet.atlassian.net/browse/MAILPOET-5749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dev/php82/Dockerfile
+++ b/dev/php82/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:8.2.0RC6-apache
+FROM wordpress:php8.2-apache
 
 ARG UID=1000
 ARG GID=1000
--- a/mailpoet/lib/AdminPages/Pages/NewsletterEditor.php
+++ b/mailpoet/lib/AdminPages/Pages/NewsletterEditor.php
@@ -92,6 +92,7 @@ class NewsletterEditor {
   }
 
   public function render() {
+    $this->setupImageSize();
     $this->assetsController->setupNewsletterEditorDependencies();
     $newsletterId = (isset($_GET['id']) ? (int)$_GET['id'] : 0);
     $woocommerceTemplateId = (int)$this->settings->get(TransactionalEmails::SETTING_EMAIL_ID, null);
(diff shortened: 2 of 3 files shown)


dev/php82/Dockerfile
```diff
@@ -37,6 +37,11 @@ RUN cat /tmp/xdebug.ini >> $XDEBUGINI_PATH
 # php extensions
 RUN docker-php-ext-install pdo_mysql
 RUN docker-php-ext-install mysqli
+# imagick is needed for generating image thumbnails
+RUN apt-get update; \
```
costasovo: Hmm, I see we still use PHP:8.2 RC image. I think we should switch to `wordpress:php8.2-apache`. WordPress image contains imagemagic. 

@lysyjan  Have you tried switching the image? Apart from the image name/tag, the docker file is very similar to the one for php8.1 that uses the WordPress image. I would suggest trying to switch the image instead of installing image magic as an extra step. If the switching image doesn't work or would lead to more changes we could go with this code and create an extra ticket for image update. What do you think?

costasovo (commented):
I would suggest trying to fix the imagemagick issue by switching the base image. Please see my comment.

dev/php82/Dockerfile
lysyjan: Great point @costasovo!
I was focused only on searching for a reason why thumbnails are not generated.

The switching docker image works. I will edit this commit and push an updated version.

lysyjan:
@costasovo Thank you for your suggestion. I dropped the previous commit and added a new one.

costasovo (approved):
@lysyjan Good job! Thanks for updating the dev Dockerfile for PHP8.2. Until now, we have been using PHP 8.2 RC in our dev environment.

veljkho (approved):
Tested via Litmus for preview and sending. The resolution of 1320px [was set only for post's images](https://d.pr/i/8PuScI) and [not added images](https://d.pr/i/tAZuAM). Is this expected @lysyjan ? In the two screenshots you can see 1320px and 980px for the same image. The above was part of `Posts` block and post's featured image, but below image was just an image block.

PR #5365, opened 2024-01-10
Display different notices before Feb 1st [MAILPOET-5819]

johnolek:
## Description

This PR changes sender domain authentication banner notices to have different content before February 1st, 2024, since we aren't actively enforcing new restrictions until then.

The main change involves the notice for free mail users. Both the title and the text reference the February 1st deadline. I also changed the last sentence to say "Your emails" instead of "Your existing scheduled and active emails", which is slightly different than what we discussed in slack.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5819

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
+++ b/mailpoet/lib/Util/Notices/SenderDomainAuthenticationNotices.php
@@ -10,6 +10,7 @@ use MailPoet\Util\FreeDomains;
 use MailPoet\Util\Helpers;
 use MailPoet\Util\License\Features\Subscribers;
 use MailPoet\WP\Notice;
+use MailPoetVendor\Carbon\Carbon;
 
 class SenderDomainAuthenticationNotices {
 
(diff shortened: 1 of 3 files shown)


brezocordero (approved):
LGTM!

Aschepikov:
Hi @johnolek we were trying to reproduce the notice when the domain is pending 
image
image
But there was no notice even with the sending or activating emails
image
Can you check it or is there a way to trigger those notices because we tried too much ways to reproduce them and there was no success

Aschepikov (approved):
GJ!

johnolek:
@Aschepikov Did you end up figuring out a way to trigger the notices? The intention is to only show these to older users, so if that's a brand new jurassic ninja site that might explain why it's not displaying. In that case, you might need to delete the `installed_after_new_domain_restrictions` setting row in the `wp_mailpoet_settings` table.

If this is still a problem I can certainly log in to the site and test some things out myself, just let me know.

PR #5367, opened 2024-01-11
Do not schedule a post notification if the old status was publish [MAILPOET-5822]

websupporter:
## Description
Solves [MAILPOET-5822]

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets
[MAILPOET-5822]

## After-merge notes

_N/A_

## Tasks

- [ ] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [ ] I added sufficient test coverage
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5822]: https://mailpoet.atlassian.net/browse/MAILPOET-5822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5822]: https://mailpoet.atlassian.net/browse/MAILPOET-5822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5374, opened 2024-01-12
Detach Entities after bulkdeleting newsletters [MAILPOET-5829]

websupporter:
## Description
This solves [MAILPOET-5829]

## Code review notes
This PR builds on top of https://github.com/mailpoet/mailpoet/pull/5315. It introduces a second helper method which allows for detaching other entity types as well.

## QA notes
Get a post notification scheduled and ensure that at the time of sending no post is to be send out. You should find an error message like this in your logs:

```
SendingQueue: A new entity was found through the relationship 'MailPoet\Entities\NewsletterLinkEntity#queue' that was not configured to cascade persist operations for entity: MailPoet\Entities\SendingQueueEntity@2448. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'MailPoet\Entities\SendingQueueEntity#__toString()' to get a clue.
```

After applying the PR you shouldn't see this error no more.

You can simulate this also with the following mu-plugin and MailPoet version 4.41.1:

```
<?php
/**
 * Plugin Name: Test
 **/

add_action('init', function() {
  $post = get_post(501);
  wp_transition_post_status('publish', 'publish', $post);
});
```

Replace the 501 in `get_post(501);` with the ID of an post which you have already send out and which is published. This works only as long as https://github.com/mailpoet/mailpoet/pull/5367 is not merged (e.g. MailPoet version 4.41.1)

Please also test other instances where newsletters get deleted (in bulk) to see that everything works as expected there as well.

## Linked tickets
[MAILPOET-5829]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5829]: https://mailpoet.atlassian.net/browse/MAILPOET-5829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5829]: https://mailpoet.atlassian.net/browse/MAILPOET-5829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/NewslettersRepository.php
+++ b/mailpoet/lib/Newsletter/NewslettersRepository.php
@@ -368,7 +368,15 @@ class NewslettersRepository extends Repository {
     $childrenIds = $this->fetchChildrenIds($ids);
     $ids = array_merge($ids, $childrenIds);
 
-    $this->entityManager->transactional(function (EntityManager $entityManager) use ($ids) {
+    $isRelatedNewsletterToBeDeleted = function($entity) use ($ids): bool {
+      if (is_string($entity) || !method_exists($entity, 'getNewsletter')) {
+        return false;
+      }
+      $newsletter = $entity->getNewsletter();
+      $newsletterId = $newsletter ? $newsletter->getId() : null;
+      return in_array($newsletterId, $ids, true);
+    };
+    $this->entityManager->transactional(function (EntityManager $entityManager) use ($ids, $isRelatedNewsletterToBeDeleted) {
       // Delete statistics data
       $newsletterStatisticsTable = $entityManager->getClassMetadata(StatisticsNewsletterEntity::class)->getTableName();
       $entityManager->getConnection()->executeStatement("
@@ -484,6 +492,27 @@ class NewslettersRepository extends Repository {
         ->where('n.id IN (:ids)')
         ->setParameter('ids', $ids)
         ->getQuery()->execute();
+
+
+      $entityTypesToBeDetached = [
+        StatisticsNewsletterEntity::class,
+        StatisticsOpenEntity::class,
+        StatisticsClickEntity::class,
+        NewsletterPostEntity::class,
+        NewsletterOptionEntity::class,
+        NewsletterLinkEntity::class,
+        StatsNotificationEntity::class,
+        SendingQueueEntity::class,
+        ScheduledTaskSubscriberEntity::class,
+        NewsletterSegmentEntity::class,
+      ];
+      foreach ($entityTypesToBeDetached as $entityType) {
+        $this->detachEntitiesOfType($entityType, $isRelatedNewsletterToBeDeleted);
+      }
+
+      $this->detachEntitiesOfType(ScheduledTaskEntity::class, function($entity) use ($taskIds): bool {
+        return !is_string($entity) && method_exists($entity, 'getId') && in_array($entity->getId(), $taskIds, true);
+      });
     });
     return count($ids);
   }
--- a/mailpoet/lib/Doctrine/Repository.php
+++ b/mailpoet/lib/Doctrine/Repository.php
@@ -145,6 +145,24 @@ abstract class Repository {
   }
 
   /**
+   * @param class-string<object> $className
+   * @param callable|null $filter
+   */
+  public function detachEntitiesOfType($className, callable $filter = null): void {
+    $rootClassName = $this->entityManager->getClassMetadata($className)->rootEntityName;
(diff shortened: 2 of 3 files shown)


JanJakes:
@websupporter Thanks for debugging this and finding the cause of these issues!

I'm trying to understand the problem better because while it may sometimes make sense to call the detach manually, I wouldn’t like to overuse it. So, I'm just thinking...

1. Why does that happen? It seems that in some cases, we call the `bulk` methods not only via the JSON API (where the request usually exits immediately), but also from cron workers, where we're using some data even after it. Usually, we are actually not doing a bulk request (like [here](https://github.com/mailpoet/mailpoet/blob/5d006c87e7ffbd6b89feb5eafae2c8619562d77d/mailpoet/lib/Cron/Workers/SendingQueue/Tasks/Newsletter.php#L194)), but at the same time, I understand it would be great to have these repository methods universally usable from anywhere in the code (not just the API).
2. Thinking about simplifications, it seems that we can't just call `$entityManager->clear()`, because then we may end up with valid and existing entities that the entity manager would no longer know. That said, sometimes we have _multiple cron runs in a single request_, and in that case, I think we could call the `clear()` between the runs (we already discussed this in the past).
3. The fix you did adds a bit of an inconsistent API that will allow a repository of one entity detach entities of all other entities. Especially together with the existing `detachAll`, I'm not sure if it's the best API. Sure, otherwise we'd need to inject many other repositories, but maybe it's worth it? We could also consider by adding `cascade={"detach"}` to the `OneToMany` associations on the `NewsletterEntity`, and then detaching one entity would correctly do the same for any related entities that no longer make sense, but that doesn't guarantee the SQL delete is called correctly, in fact.

Ideally, it would probably make sense to extract each of the sub-deletes into its correct repository (such as `StatisticsClicksRepository::deleteByNewsletter(NewsletterEntity $newsletter)`), and placing the simple detach code into that method. Then we could reuse all of these methods from the `NewslettersRepository::bulkDelete()` method. But it has a catch as well — we're wrapping it in a transaction, and that would no longer work (maybe it would with [the explicit syntax](https://www.doctrine-project.org/projects/doctrine-orm/en/2.17/reference/transactions-and-concurrency.html#approach-2-explicitly)).

So then each of the repositories would have a method like this:

```php
public function deleteByNewsletter(NewsletterEntity $newsletter): void {
    $this->entityManager->getConnection()->executeStatement("
       DELETE s FROM $newsletterStatisticsTable s
       WHERE s.`newsletter_id` IN (:ids)
    ", ['ids' => $ids], ['ids' => Connection::PARAM_INT_ARRAY]);

    // delete was done via DQL, make sure the entities are also detached from the entity manager
    $this->detachAll(function (StatisticsNewsletterEntity $entity) use ($newsletter) {
      return $entity->getNewsletter() === $newsletter;
    });
}
```

And this way we'd split it into small units with single responsibility and just reuse those. It would also be more consistent with what we've done in https://github.com/mailpoet/mailpoet/pull/5315.

---

Anyway, that's just me thinking out loud about what we can do about this problem in general. I'm not against merging this and improving it later, but `detachEntitiesOfType()` is a bit strange in context of the existing repository methods.

mailpoet/lib/Newsletter/NewslettersRepository.php
```diff
@@ -484,6 +492,27 @@ public function bulkDelete(array $ids) {
...
+        StatsNotificationEntity::class,
+        SendingQueueEntity::class,
+        ScheduledTaskSubscriberEntity::class,
+        NewsletterSegmentEntity::class,
+      ];
+      foreach ($entityTypesToBeDetached as $entityType) {
+        $this->detachEntitiesOfType($entityType, $isRelatedNewsletterToBeDeleted);
+      }
+
+      $this->detachEntitiesOfType(ScheduledTaskEntity::class, function($entity) use ($taskIds): bool {
```
JanJakes, line 513: I'm checking this deeper and the `$taskIds` seems to come from stats notification tasks (https://github.com/mailpoet/mailpoet/pull/5374/files#diff-365816efcb88f989e0039dfbd79719a9d5ec37961bae01b7744f25a674d907b8R430), which might not be what's expected (= all task IDs related to the deleted newsletters).

JanJakes (approved):
Let's merge this and I will prepare a follow-up ticket/PR with the suggested changes.

Aschepikov (approved):
GJ!

PR #5376, opened 2024-01-13
Hide AutomateWoo checkout opt-in block so only one checkbox is visible [MAILPOET-5721]

alex-mpoet:
## Description

Notes:
1. I used the code which removes the frontend block, but leaves a block stub in the editor:
image
The latter if deactivated leaves another stub:
image
I personally like the first one better. Removing the last one would mean editing the checkout page content (potentially in multiple page versions) which is too intrusive, IMO. So I'd leave the first one.
2. This error was only reproducible in acceptance tests: 
Screenshot on 2023-11-21 at 14 00 58
I could not reproduce it in my dev environment, but I solved it by unregistering the Store API extension by AW.

## Code review notes

_N/A_

## QA notes

Testing instructions are the same as in https://github.com/mailpoet/mailpoet/pull/4740, but the new block-based Woo checkout page should be tested, not the old checkout page. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5721]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5721]: https://mailpoet.atlassian.net/browse/MAILPOET-5721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
+++ b/mailpoet/lib/PostEditorBlocks/WooCommerceBlocksIntegration.php
@@ -52,7 +52,8 @@ class WooCommerceBlocksIntegration {
   public function init() {
     $this->wp->addAction(
       'woocommerce_blocks_checkout_block_registration',
-      [$this, 'registerCheckoutFrontendBlocks']
+      [$this, 'registerCheckoutFrontendBlocks'],
+      15 // Run after AutomateWoo hooks
     );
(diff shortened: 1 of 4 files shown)


johnolek (dismissed):
Looks good to me, thanks for the fix!

veljkho (approved):
This is great fix, I waited for this to enable AutomateWoo acceptance tests back and I committed small change which is enabling them back (moving about of skipped tests). The tests are passing and in my manual test all scenarios worked fine.

PR #5389, opened 2024-01-18
Add sender domain authentication task for new users [MAILPOET-5781]

alex-mpoet:
## Description

Notes:
1. I've edited the task description string as per Figma file comment.
2. After the task is clicked, the domain authorization modal is shown on the Settings page after a redirect.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5781]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5781]: https://mailpoet.atlassian.net/browse/MAILPOET-5781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/settings/pages/basics/basics.tsx
```diff
@@ -11,9 +12,10 @@ import { GdprCompliant } from './gdpr-compliant';
 import { ReEngagementPage } from './re-engagement-page';
 
 export function Basics() {
+  const { showModal } = useParams<{ showModal: string }>();
```
johnolek, line 15: I think the typing for showModal is slightly off because it's optional in the route, so it could be `undefined` instead of a string. This might be a moot point.

mailpoet/views/homepage.html
```diff
@@ -24,6 +24,9 @@
   'subscribersAddedTask': _x('Import existing subscribers', 'Link in a task list'),
   'subscribersAddedTaskDoneByImport': _x('Subscribers imported!', 'Link in a task list'),
   'subscribersAddedTaskDoneByForm': _x('Subscription form created!', 'Link in a task list'),
+  'senderDomainAuthenticatedTask': _x('Authenticate your sender domain', 'Link in a task list'),
+  'senderDomainAuthenticatedTaskDone': _x('Sender domain authenticated!', 'Link in a task list'),
+  'improveDeliveryRates': __('Improve email delivery dates'),
```
johnolek:
```suggestion
  'improveDeliveryRates': __('Improve email delivery rates'),
```

johnolek (changes requested):
There's one small typo to fix.

The ticket says `Clicking on the task list will open the modal or take them to admin.php?page=mailpoet-settings#/basics if they are using a free email address.`

My interpretation is that we want to open the modal on the homepage, but in this implementation it looks like we always navigate to the settings page and we'll open the modal there if the user isn't using a known free email.

The ticket says implementation can be modified and it definitely works fine how you have it, but I wanted to confirm it's what we want.

mailpoet/views/homepage.html
alex-mpoet: Good catch, fixed!

mailpoet/assets/js/src/settings/pages/basics/basics.tsx
alex-mpoet: Technically you're correct. But it's done the same way in the "Send with..." tab and things seem to work fine. We only use strict comparison for these optional vars, and I've seen no console errors or broken logic when it's undefined.

alex-mpoet:
@johnolek 

> My interpretation is that we want to open the modal on the homepage, but in this implementation it looks like we always navigate to the settings page and we'll open the modal there if the user isn't using a known free email.

Yes, I have mentioned this in my note number 2. I've did it this way because it was easier to implement (`Task` component doesn't support modals, only links), but if this is critical, it can be reworked. @brezocordero Do you think navigating to the settings page and opening the domain auth modal there after clicking on the task would work from the UX standpoint?

brezocordero:
> Do you think navigating to the settings page and opening the domain auth modal there after clicking on the task would work from the UX standpoint?

If it is very complex to open the modal from the task list, I would go ahead with this solution.
From my point of view, this alternative also works. Opening the modal on the settings page gives context to the action and they could also check the reply-to address.

I would like to have @shivanarrthine input as this is different from the [design](https://www.figma.com/file/woLQC5fKEGnyxllTudBEHE/Sender-domain-authentication?type=design&node-id=351-33008&mode=dev).

shivanarrthine:
> If it is very complex to open the modal from the task list, I would go ahead with this solution.
From my point of view, this alternative also works. Opening the modal on the settings page gives context to the action and they could also check the reply-to address.

Agreeing with @brezocordero. I think it's fine to go to the settings page with the modal open.

johnolek (approved):

johnolek:
Thanks for your input, @brezocordero and @shivanarrthine!

Aschepikov (approved):
GJ!

PR #5390, opened 2024-01-18
Add default spacing between blocks [MAILPOET-5816]

lysyjan:
## Description

_N/A_

## Code review notes

- I decided to use the `FLEX_GAP` constant as a default padding in the email because the unified spaces look better.
- As a part of this PR, it is removing some reset CSS styles because it makes the rendered HTML not transparent.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5816]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5816]: https://mailpoet.atlassian.net/browse/MAILPOET-5816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/SpacingPreprocessor.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/SpacingPreprocessor.php
@@ -16,8 +16,8 @@ class SpacingPreprocessor implements Preprocessor {
 
   private function addMarginTopToBlocks(array $parsedBlocks): array {
     foreach ($parsedBlocks as $key => $block) {
-      // We don't want to add margin-top to column block
-      if ($key !== 0 && $block['blockName'] !== 'core/column') {
+      // We don't want to add margin-top to the first block in the email or to the first block in the columns block
+      if ($key !== 0) {
         $block['email_attrs']['margin-top'] = SettingsController::FLEX_GAP;
       }
 
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/SpacingPreprocessor.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/SpacingPreprocessor.php
@@ -1,0 +1,31 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Engine\Renderer\Preprocessors;
+
+use MailPoet\EmailEditor\Engine\SettingsController;
+
+/**
+ * This preprocessor is responsible for setting default spacing values for blocks.
+ * In the early development phase, we are setting only margin-top for blocks that are not first or last in the columns block.
+ */
+class SpacingPreprocessor implements Preprocessor {
+  public function preprocess(array $parsedBlocks, array $layoutStyles): array {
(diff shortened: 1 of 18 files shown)


mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/SpacingPreprocessor.php
```diff
@@ -0,0 +1,31 @@
...
+    $parsedBlocks = $this->addMarginTopToBlocks($parsedBlocks);
+    return $parsedBlocks;
+  }
+
+  private function addMarginTopToBlocks(array $parsedBlocks): array {
+    foreach ($parsedBlocks as $key => $block) {
+      // We don't want to add margin-top to column block
+      if ($key !== 0 && $block['blockName'] !== 'core/column') {
+        $block['email_attrs']['margin-top'] = SettingsController::FLEX_GAP;
+      }
```
costasovo: Why do we want to skip columns? Is it because of some specific behavior? Would it be possible to detect it differently than checking the block name? 

The thing is that here we are in the `Engine` namespace, so we should not use block names from the core integration in the code.

Maybe we could find some other common attributes that would help us detect when to skip adding the margin and it would work also for other blocks we will add later or alternatively we could use a filter (something like `$blocksWithoutMargin = apply_filters('mailpoet_xyz', [])`) and we could pass the columns there in `Integration/Core/Initializer`. What do you think?

costasovo (changes requested):
Great job! This improves the look of the final email!

I found one problem with using a core block name in the engine code. Please see my other comment.

mailpoet/lib/EmailEditor/Engine/Renderer/Preprocessors/SpacingPreprocessor.php
lysyjan: > Why do we want to skip columns? Is it because of some specific behavior? Would it be possible to detect it differently than checking the block name?

The main reason is that the `core/column` block is rendered horizontally compared to the rest of the blocks.
So the top margin could cause something like this:
Screenshot 2024-01-23 at 17 18 44

Nevertheless, I realized now the top margin is not used in the renderer for this block, so I think we can remove this part of the condition, and we don't need a filter here. Does it make sense to you?

costasovo: @lysyjan Thank you for the explanation!

> Nevertheless, I realized now the top margin is not used in the renderer for this block, so I think we can remove this part of the condition, and we don't need a filter here. Does it make sense to you?

👍 If it is not needed, let's remove it. My concern was about mixing the engine and integration logic so if we remove the check for the specific block we are good.

lysyjan: @costasovo I updated the condition in the last commit and rebased the branch due to conflicts with trunk.

costasovo (approved):
LGTM!

PR #5391, opened 2024-01-19
Block sending existing campaigns for big senders    if sender domain is not authorized

brezocordero:
## Description

Implements the remaining sending rules regarding sender domain authentication.

If using MSS and is a big sender (>1000 subscribers), existing campaigns with unauthenticated domains will be paused during sending.

This rule will be applied to existing users on February 1st.

## Code review notes

I have added a transient to reduce the number of API calls.
The function `checkAuthorizedEmailAddresses` uses `getVerifiedSenderDomainsIgnoringCache` that resets the transient. I believe this function is used in all the places where we would need to update the verified sender domains but please double check I did not forget anything.

## QA notes

We can't test yet the errors from the API but you can test the check before sending:

Reduce the lower_limit and upper_limit so that your subscribers number > upper_limit
Activate a post notification that sends immediately and that uses a verified domain as sender address.
Remove the verified domain from the shop (make sure the address is still authenticated)
Publish a post
Check that the post notification history is not sent and it is paused.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5832]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5832]: https://mailpoet.atlassian.net/browse/MAILPOET-5832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingErrorHandler.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingErrorHandler.php
@@ -4,6 +4,7 @@ namespace MailPoet\Cron\Workers\SendingQueue;
 
 use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SendingQueueEntity;
+use MailPoet\Logging\LoggerFactory;
 use MailPoet\Mailer\MailerError;
 use MailPoet\Mailer\MailerLog;
 use MailPoet\Newsletter\Sending\ScheduledTaskSubscribersRepository;
@@ -19,14 +20,19 @@ class SendingErrorHandler {
   /** @var SendingQueuesRepository */
   private $sendingQueuesRepository;
 
+  /** @var LoggerFactory */
+  private $loggerFactory;
+
   public function __construct(
     ScheduledTaskSubscribersRepository $scheduledTaskSubscribersRepository,
     SendingThrottlingHandler $throttlingHandler,
-    SendingQueuesRepository $sendingQueuesRepository
+    SendingQueuesRepository $sendingQueuesRepository,
+    LoggerFactory $loggerFactory
   ) {
     $this->scheduledTaskSubscribersRepository = $scheduledTaskSubscribersRepository;
     $this->throttlingHandler = $throttlingHandler;
     $this->sendingQueuesRepository = $sendingQueuesRepository;
+    $this->loggerFactory = $loggerFactory;
   }
 
   public function processError(
@@ -63,6 +69,14 @@ class SendingErrorHandler {
     $queue = $task->getSendingQueue();
 
     if ($queue instanceof SendingQueueEntity) {
+      if ($error->getOperation() === MailerError::OPERATION_DOMAIN_AUTHORIZATION) {
+        $this->loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->info(
+          'Paused task in sending queue due to sender domain authorization error',
+          ['task_id' => $task->getId()]
+        );
+        $this->sendingQueuesRepository->pause($queue);
+        return;
+      }
       $this->sendingQueuesRepository->updateCounts($queue);
     }
   }
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -308,6 +308,8 @@ class SendingQueue {
       $this->reScheduleBounceTask();
 
       // Check task has not been paused before continue processing
+      // This is needed because the task can be paused in the middle of the batch processing,
+      // for example on API error ERROR_MESSAGE_BULK_EMAIL_FORBIDDEN
       if ($task->getStatus() === ScheduledTaskEntity::STATUS_PAUSED) {
         return;
       }
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -307,6 +307,11 @@ class SendingQueue {
       // reschedule bounce task to run sooner, if needed
       $this->reScheduleBounceTask();
 
+      // Check task has not been paused before continue processing
+      if ($task->getStatus() === ScheduledTaskEntity::STATUS_PAUSED) {
+        return;
+      }
+
       if ($newsletter->getStatus() !== NewsletterEntity::STATUS_CORRUPT) {
         $this->processQueue(
           $task,
--- a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
+++ b/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
@@ -20,6 +20,7 @@ use MailPoet\Newsletter\Sending\ScheduledTaskSubscribersRepository;
 use MailPoet\Newsletter\Sending\SendingQueuesRepository;
 use MailPoet\Segments\SegmentsRepository;
 use MailPoet\Segments\SubscribersFinder;
+use MailPoet\Services\AuthorizedEmailsController;
 use MailPoet\Statistics\StatisticsNewslettersRepository;
 use MailPoet\Subscribers\SubscribersRepository;
 use MailPoet\Tasks\Subscribers\BatchIterator;
@@ -87,6 +88,9 @@ class SendingQueue {
   /** @var StatisticsNewslettersRepository */
   private $statisticsNewslettersRepository;
 
+  /** @var AuthorizedEmailsController */
+  private $authorizedEmailsController;
+
   public function __construct(
     SendingErrorHandler $errorHandler,
     SendingThrottlingHandler $throttlingHandler,
@@ -104,6 +108,7 @@ class SendingQueue {
     SendingQueuesRepository $sendingQueuesRepository,
     EntityManager $entityManager,
     StatisticsNewslettersRepository $statisticsNewslettersRepository,
+    AuthorizedEmailsController $authorizedEmailsController,
     $newsletterTask = false
   ) {
     $this->errorHandler = $errorHandler;
@@ -124,6 +129,7 @@ class SendingQueue {
     $this->sendingQueuesRepository = $sendingQueuesRepository;
     $this->entityManager = $entityManager;
     $this->statisticsNewslettersRepository = $statisticsNewslettersRepository;
+    $this->authorizedEmailsController = $authorizedEmailsController;
   }
 
   public function process($timer = false) {
@@ -205,6 +211,17 @@ class SendingQueue {
       $task->setStatus(ScheduledTaskEntity::STATUS_PAUSED);
       $this->scheduledTasksRepository->flush();
       $this->wp->setTransient(self::EMAIL_WITH_INVALID_SEGMENT_OPTION, $newsletter->getSubject());
+      return;
+    }
+
+    // Pause task if sender domain requirements are not met
+    if (!$this->authorizedEmailsController->isSenderAddressValid($newsletter, 'sending')) {
+      $this->loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->info(
+        'pause task in sending queue due to sender domain requirements',
+        ['task_id' => $task->getId()]
+      );
+      $task->setStatus(ScheduledTaskEntity::STATUS_PAUSED);
+      $this->scheduledTasksRepository->flush();
       return;
     }
 
--- a/mailpoet/lib/Mailer/Methods/MailPoet.php
+++ b/mailpoet/lib/Mailer/Methods/MailPoet.php
@@ -82,21 +82,28 @@ class MailPoet implements MailerMethod {
   }
 
   public function processSendError($result, $subscriber, $newsletter) {
-    if (!empty($result['code'])) {
-      if ($result['code'] === API::RESPONSE_CODE_KEY_INVALID) {
-        $this->bridge->invalidateMssKey();
-      } elseif (
-        ($result['code'] === API::RESPONSE_CODE_CAN_NOT_SEND
-          && $result['error'] === API::ERROR_MESSAGE_INVALID_FROM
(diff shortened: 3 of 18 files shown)


mailpoet/lib/Services/AuthorizedSenderDomainController.php
```diff
@@ -288,7 +296,11 @@ public function isSmallSender(): bool {
     return $this->subscribers->getSubscribersCount() <= self::LOWER_LIMIT;
   }
 
-  public function isAuthorizedDomainRequiredForNewCampaigns(): bool {
+  public function isBigSender(): bool {
+    return $this->subscribers->getSubscribersCount() > self::UPPER_LIMIT;
+  }
+
+  private function restrictionsApply() {
```
johnolek:
```suggestion
  private function restrictionsApply(): bool {
```

mailpoet/lib/Services/AuthorizedEmailsController.php
```diff
@@ -153,16 +153,16 @@ public function onNewsletterSenderAddressUpdate(NewsletterEntity $newsletter, st
...
-
+  public function isSenderAddressValid(NewsletterEntity $newsletter, string $context = 'activation'): bool {
     if (!in_array($newsletter->getType(), NewsletterEntity::CAMPAIGN_TYPES)) {
       return true;
     }
 
-    if (!$this->senderDomainController->isAuthorizedDomainRequiredForNewCampaigns()) {
+    $isAuthorizedDomainRequired = $context === 'activation' ?
+      $this->senderDomainController->isAuthorizedDomainRequiredForNewCampaigns() :
+      $this->senderDomainController->isAuthorizedDomainRequiredForExistingCampaigns();
```
johnolek: Minor style thing, I'm more used to seeing the `?` and `:` of ternaries at the beginning of the lines, and a quick scan through the codebase makes it look like it's much more common.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
```diff
@@ -208,6 +214,17 @@ private function processSending(ScheduledTaskEntity $task, int $timer): void {
...
+    // Pause task if sender domain requirements are not met
+    if (!$this->authorizedEmailsController->isSenderAddressValid($newsletter, 'sending')) {
+      $this->loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->info(
+        'pause task in sending queue due to sender domain requirements',
+        ['task_id' => $task->getId()]
+      );
+      $task->setStatus(ScheduledTaskEntity::STATUS_PAUSED);
+      $this->scheduledTasksRepository->flush();
+      return;
+    }
```
johnolek, line 226: This looks totally fine to me, but I'm curious to know what you think about the placement. We run this check before we start working on batches. Do we ever run into issues where the state changes in between batches and we should pause the sending but don't?

I think this would actually be much more likely to apply to the existing check above looking for deleted/trashed segments, and it's probably not relevant for this code since we're caching results for 24 hours with the transient.

On that note, what does that caching mean for users who have their campaigns paused and then validate their email address? Is there any chance they would have to wait, or can we always rely on [resumeSending](https://github.com/mailpoet/mailpoet/blob/994f46762d141387a06a205b933f5f96557cc887/mailpoet/lib/API/JSON/v1/Mailer.php#L86-L86) to break the cache through [checkAuthorizedEmailAddresses](https://github.com/mailpoet/mailpoet/blob/994f46762d141387a06a205b933f5f96557cc887/mailpoet/lib/Services/AuthorizedEmailsController.php#L125-L125)?

I'm mostly just curious if you think there's an argument for moving checks like these into `enforceSendingAndExecutionLimits` which runs on every batch.

mailpoet/lib/Cron/Workers/SendingQueue/SendingErrorHandler.php
```diff
@@ -63,6 +69,14 @@ private function processSoftError(MailerError $error, ScheduledTaskEntity $task,
     $queue = $task->getSendingQueue();
 
     if ($queue instanceof SendingQueueEntity) {
+      if ($error->getOperation() === MailerError::OPERATION_DOMAIN_AUTHORIZATION) {
+        $this->loggerFactory->getLogger(LoggerFactory::TOPIC_NEWSLETTERS)->info(
+          'Paused task in sending queue due to sender domain authorization error',
+          ['task_id' => $task->getId()]
```
johnolek, line 75: It's probably not even common for people to use multiple sending domains, but for those that do, what do you think about trying to get the actual sender domain or from address in the context? I was thinking it could make debugging easier but I could also see it being redundant if it's clear from other logs or in the UI.

mailpoet/lib/Services/AuthorizedSenderDomainController.php
```diff
@@ -235,9 +244,21 @@ private function returnAllDomains(array $records): array {
     return $domains;
   }
 
+  private function reloadCache() {
+    $this->currentRecords = null;
+    $this->currentRawData = $this->bridge->getRawSenderDomainData();
+    $this->wp->setTransient(self::SENDER_DOMAINS_KEY, $this->currentRawData, 60 * 60 * 24);
+  }
```
johnolek: Extremely minor comment since this looks fine as is.

This method looks slightly inconsistent. It's simply setting `currentRecords` to `null` and doesn't reload them. I think this might be simpler if instead it were called `clearCache`, and you could set `currentRawData` to `null` as well, and also delete the transient.

It would work exactly the same. The only real benefit would be not having to worry about updating the transient time in multiple places, but that could easily be fixed with a class constant... so again, no big deal.

mailpoet/lib/Mailer/Methods/MailPoet.php
```diff
@@ -82,14 +82,20 @@ public function send($newsletter, $subscriber, $extraParams = []): array {
...
+      } elseif (
+        ($result['code'] === API::RESPONSE_CODE_CAN_NOT_SEND
+          && $result['error'] === API::ERROR_MESSAGE_INVALID_FROM
+        )
+        ||
+        ($result['code'] === API::RESPONSE_CODE_PAYLOAD_ERROR
+          && !empty($result['error']) && $result['error'] === API::ERROR_MESSAGE_BULK_EMAIL_FORBIDDEN
+        )
+      ) {
+          $this->authorizedEmailsController->checkAuthorizedEmailAddresses();
```
johnolek, line 96: Looks like the indentation is slightly off here

mailpoet/lib/Mailer/Methods/MailPoet.php
```diff
@@ -82,14 +82,20 @@ public function send($newsletter, $subscriber, $extraParams = []): array {
...
   public function processSendError($result, $subscriber, $newsletter) {
-    if (!empty($result['code']) && $result['code'] === API::RESPONSE_CODE_KEY_INVALID) {
-      $this->bridge->invalidateMssKey();
-    } elseif (
-      !empty($result['code'])
-      && $result['code'] === API::RESPONSE_CODE_CAN_NOT_SEND
-      && $result['error'] === API::ERROR_MESSAGE_INVALID_FROM
-    ) {
-      $this->authorizedEmailsController->checkAuthorizedEmailAddresses();
+    if (!empty($result['code'])) {
```
johnolek: There's something about the formatting of the conditionals that makes this a little hard for me to read. Do you have any guess about if we'll be adding to this in the future? If so (i.e. it might get even more complex) I might personally prefer something more like this (no guarantee that it's functionally the same!):

```php
public function processSendError($result, $subscriber, $newsletter) {
    if (empty($result['code'])) {
        return $this->errorMapper->getErrorForResult($result, $subscriber, $this->sender, $newsletter);
    }

    switch ($result['code']) {
        case API::RESPONSE_CODE_KEY_INVALID:
            $this->bridge->invalidateMssKey();
            break;

        case API::RESPONSE_CODE_CAN_NOT_SEND:
            if ($result['error'] === API::ERROR_MESSAGE_INVALID_FROM) {
                $this->authorizedEmailsController->checkAuthorizedEmailAddresses();
            }
            break;

        case API::RESPONSE_CODE_PAYLOAD_ERROR:
            if (!empty($result['error']) && $result['error'] === API::ERROR_MESSAGE_BULK_EMAIL_FORBIDDEN) {
                $this->authorizedEmailsController->checkAuthorizedEmailAddresses();
            }
            break;
    }

    return $this->errorMapper->getErrorForResult($result, $subscriber, $this->sender, $newsletter);
}
```

johnolek (dismissed):
I left a bunch of comments but none of them seem important, so I'm approving this and letting you decide if anything seems worth changing.

One note I couldn't add as a comment is `getAllSenderDomainsIgnoringCache` doesn't appear to be used anymore so it would be good to remove that method unless we might need it in the future?

I think you could remove the old constants like `DOMAIN_VERIFICATION_STATUS_VALID`, or maybe change the values and use them instead of the ones like `OVERALL_STATUS_VERIFIED`. I don't think the `OVERALL_` distinction matters anymore after these changes but I could be mistaken.

I didn't see anywhere else we would need to reset the transient.

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
brezocordero: > Do we ever run into issues where the state changes in between batches and we should pause the sending but don't?

I don't think it would be common but it can happen. In that case the API would throw the error and we would pause the task. That's why I added the check for Paused tasks inside of the subscribers loop.
I can add a comment about it.

> On that note, what does that caching mean for users who have their campaigns paused and then validate their email address?

Validating the email address will reset the transient for Sender Domains because it calls `checkAuthorizedEmailAddresses`. 

> I'm mostly just curious if you think there's an argument for moving checks like these into enforceSendingAndExecutionLimits which runs on every batch.

I think it is a good idea and it would have made the work on this ticket much easier. 
I would do it on a separate ticket. We might not need to run all the checks every time `enforceSendingAndExecutionLimits` runs and we would need to be careful with the changes and the performance. This is also a critical part of the application, so taking the time for the refactor and adding tests would be the best.

mailpoet/lib/Cron/Workers/SendingQueue/SendingErrorHandler.php
brezocordero: > I could also see it being redundant if it's clear from other logs or in the UI.

Thanks for the comment. There is another log with the sender domain from the function getDmarcMessage. I would skip it adding it here too, for now, and add it later if we find we need it for debugging.

mailpoet/lib/Services/AuthorizedSenderDomainController.php
brezocordero: I wanted to avoid another API call. 
I am not fully sure when `$this->currentRecords` would need to be set to `null` or even if we still need it.
I didn't have time to look at it in detail but I can open a Janitorial ticket to do some cleanup here.

mailpoet/lib/Mailer/Methods/MailPoet.php
brezocordero: Thanks! The linter missed this but you have the eye.

mailpoet/lib/Mailer/Methods/MailPoet.php
brezocordero: Thanks for the suggestion. I will use the `switch`.

mailpoet/lib/Services/AuthorizedEmailsController.php
brezocordero: I can change this :)

mailpoet/lib/Mailer/Methods/MailPoet.php
brezocordero: Done in https://github.com/mailpoet/mailpoet/pull/5391/commits/0882a36c7bebb612b77833979953aa094038eebd

mailpoet/lib/Mailer/Methods/MailPoet.php
brezocordero: I used the `switch` instead

mailpoet/lib/Services/AuthorizedSenderDomainController.php
brezocordero: I added some cleanup to the ticket to remove the enforcement date, including checking `$this->currentRecords` and the function `reloadCache`

mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php
brezocordero: I added a comment when checking if the task has been paused when processing a batch in https://github.com/mailpoet/mailpoet/pull/5391/commits/c071b7f9b6c26a91b2905be6a42cfe58b7e39442

To move the checks to `enforceSendingAndExecutionLimits`, could you please create a maintenance ticket?

mailpoet/lib/Services/AuthorizedEmailsController.php
brezocordero: Done in https://github.com/mailpoet/mailpoet/pull/5391/commits/a16a6e9d7dfbc28a890cee42d220c9a71f75765d

mailpoet/lib/Services/AuthorizedSenderDomainController.php
brezocordero: Done in https://github.com/mailpoet/mailpoet/pull/5391/commits/40f5c52c637c78f71fd013ce58b35c3378e21069

brezocordero:
@johnolek Thanks for the review.

I addressed all the comments here or added them to a ticket.

> One note I couldn't add as a comment is getAllSenderDomainsIgnoringCache doesn't appear to be used anymore so it would be good to remove that method unless we might need it in the future?

I removed it on https://github.com/mailpoet/mailpoet/pull/5391/commits/6ac8e2b6fcfffe2b6ba48407dc8dfcb6da306487

Aschepikov (approved):
GJ!

PR #5392, opened 2024-01-19
Email editor – font styles + default button block  [MAILPOET-5740][MAILPOET-5814]

costasovo:
## Description
This PR enables font family and font-size settings for all supported blocks and adds default button styles.

It also changes how we work with theme.json a bit. Instead of loading core theme.json and manually inserting some stuff from the editor's theme.json, we now load the core theme and merge the editor's theme into it. The benefit is that we no longer have to manually change individual settings and we can use WP_Theme_JSON::get_settings and other methods. I have also moved a few configurations (which were easy to move) from SettingsController to theme.json.

## Code review notes
I tried to explain changes in the commit messages.

I included the default button styles in this PR because when doing small refactors around theme.json processing, the default styles started somewhat to work in the editor, so I decided to fix it and add the support also to the renderer. It makes the PR a bit bigger, though. Sorry about that!

## QA notes
I would skip QA for now and pass it to QA after we fix/implement the rest of the tickets related to the rendering.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5740][MAILPOET-5814]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5740]: https://mailpoet.atlassian.net/browse/MAILPOET-5740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Engine/SettingsController.php
+++ b/mailpoet/lib/EmailEditor/Engine/SettingsController.php
@@ -56,45 +56,6 @@ class SettingsController {
    */
   const FLEX_GAP = '16px';
 
-  /**
-   * Default styles applied to the email. These are going to be replaced by style settings.
-   * This is currently more af a proof of concept that we can apply styles to the email.
-   * We will gradually replace these hardcoded values with styles saved as global styles or styles saved with the email.
-   * @var array
-   */
-  const DEFAULT_EMAIL_CONTENT_STYLES = [
-    'h1' => [
-      'typography' => [
-        'fontSize' => '32px',
-      ],
-    ],
-    'h2' => [
-      'typography' => [
-        'fontSize' => '24px',
-      ],
-    ],
-    'h3' => [
-      'typography' => [
-        'fontSize' => '18px',
-      ],
-    ],
-    'h4' => [
-      'typography' => [
-        'fontSize' => '16px',
-      ],
-    ],
-    'h5' => [
-      'typography' => [
-        'fontSize' => '14px',
-      ],
-    ],
-    'h6' => [
-      'typography' => [
-        'fontSize' => '12px',
-      ],
-    ],
-  ];
-
   private $availableStylesheets = '';
 
   public function getSettings(): array {
@@ -162,10 +123,6 @@ class SettingsController {
       'contentSize' => self::EMAIL_WIDTH,
       'layout' => 'constrained',
     ];
-  }
-
-  public function getEmailContentStyles(): array {
-    return self::DEFAULT_EMAIL_CONTENT_STYLES;
   }
 
   public function getAvailableStylesheets(): string {
--- a/mailpoet/lib/EmailEditor/Engine/SettingsController.php
+++ b/mailpoet/lib/EmailEditor/Engine/SettingsController.php
@@ -106,14 +106,16 @@ class SettingsController {
     $coreThemeData = \WP_Theme_JSON_Resolver::get_core_data();
     $coreSettings = $coreThemeData->get_settings();
 
+    // Email Editor Theme
+    $editorTheme = $this->getTheme();
+
     // Enable custom spacing
     $coreSettings['spacing']['units'] = ['px'];
     $coreSettings['spacing']['padding'] = true;
     // Typography
     $coreSettings['typography']['dropCap'] = false; // Showing large initial letter cannot be implemented in emails
     $coreSettings['typography']['fontWeight'] = false; // Font weight will be handled by the font family later
-
-    $theme = $this->getTheme();
+    $coreSettings['typography']['fontFamilies']['default'] = $editorTheme->get_data()['settings']['typography']['fontFamilies'];
 
     // body selector is later transformed to .editor-styles-wrapper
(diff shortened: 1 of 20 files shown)


mailpoet/lib/EmailEditor/Engine/theme.json
```diff
@@ -149,6 +149,23 @@
...
+    "blocks": {
+      "core/button": {
+        "variations": {},
+        "color": {
+          "background": "#32373c",
+          "text": "#ffffff"
+        },
+        "spacing": {
+          "padding": {
+            "bottom": "0.7em",
```
lysyjan: I thought we only wanted to use `px` in the early phase. Font size and other values are in pixels now. Have you considered converting those padding to them?

mailpoet/lib/EmailEditor/Engine/theme.json
```diff
@@ -5,7 +5,14 @@
     "layout": {
       "contentSize": "660px"
     },
+    "spacing": {
+      "units": ["px"],
+      "padding": true,
+      "margin": false
+    },
     "typography": {
+      "dropCap": false,
```
lysyjan, line 25: Moving all those configurations here simplifies the code, and I like it. Nevertheless, we lost short comments about some details. Maybe we should keep them somewhere. I think it's better to start earlier than later.

mailpoet/lib/EmailEditor/Engine/SettingsController.php
```diff
@@ -191,12 +191,17 @@ public function getTheme(): \WP_Theme_JSON {
...
-    $settings = $this->getTheme()->get_settings();
+    $coreThemeSettings = \WP_Theme_JSON_Resolver::get_core_data()->get_settings();
+    $emailThemeSettings = $this->getTheme()->get_settings();
     $css = '';
     // Font family classes
-    foreach ($settings['typography']['fontFamilies']['theme'] as $fontFamily) {
+    foreach ($emailThemeSettings['typography']['fontFamilies']['theme'] as $fontFamily) {
       $css .= ".has-{$fontFamily['slug']}-font-family { font-family: {$fontFamily['fontFamily']}; } \n";
     }
+    // Font size classes
```
lysyjan, line 168: > This is needed for cases where we don't touch the HTML output
to insert font size from the preprocessor (e.g., list item)

Your comment in the commit message mentions that this change is needed for the original HTML.
Do you think that we could use those classes in our updated/generated HTML instead of inlining extracted values?
It's only an idea and question because I'm a bit curious here.

lysyjan (dismissed):
A lot of improvements, fixes, and new features.
I'm very happy about those changes. Great work!
@costasovo I had some questions, but from my side, this PR is ready for merge.

mailpoet/lib/EmailEditor/Engine/theme.json
costasovo: Good point. I was thinking about creating a class for working with the theme and adding those comments there. I don't have the class. So for now I'll add a section to readme

mailpoet/lib/EmailEditor/Engine/SettingsController.php
costasovo: > Do you think that we could use those classes in our updated/generated HTML instead of inlining extracted values?

@lysyjan  This is a really good question. 👍 

I considered this option, but in the end, I think it was not worth it. For two reasons:

1) We allow custom font size values, we need to render those as CSS rules anyway. With classes we would still keep support also for pixel values. Supporting both classes and values at the block level is more complex than just values.
2) Because we reset the font size to 0 at a couple of wrappers, we need to bubble the values down. For custom font-size we send pixel value, if we wanted to support also classes we would have to bubble down both. So, supporting both classes and values at the block level is more complex than just values.

So this inlining currently covers only list items blocks. In general this could cover all blocks which are leaf blocks and which keep the original HTML from Gutenberg.

Does that make sense to you?

mailpoet/lib/EmailEditor/Engine/theme.json
costasovo: I started with pixels, but learned that em works much nicer. When you change font-size em keeps a nice ratio for all font sizes. We have a Jira ticket for fixing the buttons for Outlook so maybe we will have to switch to pixels anyway. 

If you don't mind I would keep the em unit for now and we will see what will be needed for the Outlook fixes.

lysyjan: > If you don't mind I would keep the em unit for now and we will see what will be needed for the Outlook fixes.

Of course. Thank you for the explanation.

costasovo:
@lysyjan I added [a commit](https://github.com/mailpoet/mailpoet/pull/5392/commits/5f571f092ac2c785237e92880eee3887589124c6) with a documentation for theme.json. Do you think it is sufficient, or do you have an idea of what to add? Also, please feel free to edit it.

lysyjan (approved):

PR #5393, opened 2024-01-19
Prevent stale/orphaned Doctrine data when using DQL/SQL [MAILPOET-5845]

JanJakes:
## Description

This is a more proper fix and refactor of https://github.com/mailpoet/mailpoet/pull/5374, plus a few more improvements and entity manager clearing between runs.

See also: https://mailpoetdev.wordpress.com/2024/01/19/issue-with-stuck-post-notifications/

## Code review notes

_N/A_

## QA notes

Please, test that post notifications get scheduled and sent (and don't get stuck). Please, try a larger list, if possible.

It would also be good to smoke test other types of sending, and background jobs.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5845]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5845]: https://mailpoet.atlassian.net/browse/MAILPOET-5845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Newsletters.php
+++ b/mailpoet/lib/API/JSON/v1/Newsletters.php
@@ -15,6 +15,7 @@ use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\InvalidStateException;
 use MailPoet\Listing;
 use MailPoet\Newsletter\Listing\NewsletterListingRepository;
+use MailPoet\Newsletter\NewsletterDeleteController;
 use MailPoet\Newsletter\NewsletterSaveController;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\Newsletter\NewsletterValidator;
@@ -73,6 +74,8 @@ class Newsletters extends APIEndpoint {
   /** @var NewsletterSaveController */
   private $newsletterSaveController;
 
+  private NewsletterDeleteController $newsletterDeleteController;
+
   /** @var NewsletterUrl */
   private $newsletterUrl;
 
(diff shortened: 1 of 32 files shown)


alex-mpoet (commented):
@JanJakes Looks good overall!

But I'm seeing all these `deleteByNewsletterIds()` (and variations like `removeNewsletterDataByNewsletterIds()`, `newsletterRepository->deleteByIds()`) methods duplicated in so many repositories, they differ only in class name passed to them, can they be unified and extracted to a trait?

JanJakes:
@alex-mpoet That's a good idea. It makes me think, if I can add some sort of more generic helper to the parent repository class for these bulk actions. I'll check it out.

JanJakes:
@alex-mpoet I tried to use [Doctrine Criteria](https://www.doctrine-project.org/projects/doctrine-collections/en/stable/expression-builder.html) for this to automatically do both deletes and the in-memory cleanup, such as:

```php
// in any repository
$this->deleteAll(new Criteria(Criteria::expr()->in('id', $ids)));
```

The idea is that I could apply criteria to the SQL query (query builder) as well as to the in memory data (collection). I [implemented it](https://github.com/mailpoet/mailpoet/tree/doctrine-bulk-actions-criteria), but during testing, I realized the criteria evaluation for collections was poorly implemented and I couldn't reliably use it. There were issues, such as the inability to use IDs for relations (I needed to fetch entity references), and even worse, their `IN` operator compares objects non-strictly (like `==`), which recurses the whole object tree and can cause infinite depth with circular references.

So, I threw that away and wanted to try a trait, but that's not easy either, because the `deleteByNewsletterIds` calls parent repository function (`detachAll`), and needs private properties (`$this->entityManager`).

So, I think I would leave it this way for now. We'll see what more we'll need to do, regarding the Doctrine bulk actions.

alex-mpoet (approved):
Thanks for looking into it and trying to fix it! I agree, it's worth looking into Doctrine bulk actions later to see if any improvements are possible.

Aschepikov (approved):
GJ!

PR #5400, opened 2024-01-25
Ensure invalid tasks don't prevent new post notifications [MAILPOET-5853]

johnolek:
## Description

If a sending task associated with a post notification ever gets into an `invalid` status, it will prevent new post notifications from being scheduled if the associated newsletter is not marked as sent.

I'm not sure if this is a good approach or not, but I thought it would at least be worth considering.

## Code review notes

_N/A_

## QA notes

Make sure sending works (post notifications).

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
+++ b/mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
@@ -111,7 +111,7 @@ class PostNotificationScheduler {
   }
 
   public function createPostNotificationSendingTask(NewsletterEntity $newsletter): ?ScheduledTaskEntity {
-    $notificationHistory = $this->newslettersRepository->findSendingNotificationHistoryWithoutPausedTask($newsletter);
+    $notificationHistory = $this->newslettersRepository->findSendingNotificationHistoryWithoutPausedOrInvalidTask($newsletter);
     if (count($notificationHistory) > 0) {
       return null;
     }
--- a/mailpoet/lib/Newsletter/NewslettersRepository.php
+++ b/mailpoet/lib/Newsletter/NewslettersRepository.php
@@ -520,7 +520,7 @@ class NewslettersRepository extends Repository {
   /**
    * @return NewsletterEntity[]
    */
-  public function findSendingNotificationHistoryWithoutPausedTask(NewsletterEntity $newsletter): array {
+  public function findSendingNotificationHistoryWithoutPausedOrInvalidTask(NewsletterEntity $newsletter): array {
     return $this->entityManager->createQueryBuilder()
       ->select('n')
       ->from(NewsletterEntity::class, 'n')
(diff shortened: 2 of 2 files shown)


mailpoet/lib/Newsletter/Scheduler/PostNotificationScheduler.php
```diff
@@ -111,7 +111,7 @@ public function schedulePostNotification($postId) {
   }
 
   public function createPostNotificationSendingTask(NewsletterEntity $newsletter): ?ScheduledTaskEntity {
-    $notificationHistory = $this->newslettersRepository->findSendingNotificationHistoryWithoutPausedTask($newsletter);
+    $notificationHistory = $this->newslettersRepository->findSendingNotificationHistoryWithoutPausedOrInvalidTask($newsletter);
```
JanJakes, line 114: I'm trying to understand what was originally happening here. There is `count($notificationHistory) > 0` just below, but that would mean that if there are, for instance, any history records that are `completed`, it would never schedule anything again?

Or does this code mean something else?

johnolek: The main thing is that it's looking just for notification histories that have a status of `sending`. The point seems to be that we don't want to have multiple post notification histories being sent at the same time for the same parent post notification.

Since we're running into an issue where the sending queue never gets a chance to update that status from `sending` to `sent`, this code is now preventing new sending tasks from being created because there's the post notification history that's supposedly `sending` but the task is `invalid`, and currently only `paused` tasks are allowed.

The way I tested this locally was to update [this conditional](https://github.com/mailpoet/mailpoet/blob/c12a075b33c45dae8e0e5bba058c282e7569ce3a/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L230-L230) to be `true`, which ended up causing that doctrine entity by the way:

```
MailPoet.ERROR: Multiple non-persisted new entities were found through the given association graph:   * A new entity was found through the relationship 'MailPoet\Entities\ScheduledTaskEntity#sendingQueue' that was not configured to cascade persist operations for entity: MailPoet\Entities\SendingQueueEntity@5089. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'MailPoet\Entities\SendingQueueEntity#__toString()' to get a clue.  * A new entity was found through the relationship 'MailPoet\Entities\NewsletterEntity#newsletterSegments' that was not configured to cascade persist operations for entity: MailPoet\Entities\NewsletterSegmentEntity@7446. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'MailPoet\Entities\NewsletterSegmentEntity#__toString()' to get a clue. {"error":"[object] (MailPoetVendor\\Doctrine\\ORM\
```

After this error, it was this code that was preventing new post notification histories from getting created after publishing a new post.

I hope that all makes sense, and it's very probable that there's a better fix.

JanJakes: @johnolek

> The main thing is that it's looking just for notification histories that have a status of sending.

This. I missed that. Thanks for the explanation. I think we should merge this fix, and then consider if we should do something more robust.

We're sort of between two directions here:

1.  Really select only tasks that are running (in that case, we could check for that state rather than excluding all other states).
2. Do not exclude any state (or maybe just `paused`), and then the cron run must make sure all is fine and consistent. What I mean by that is that, having a newsletter history record with status `sending`, but the related task being in `invalid` state, is probably not a great state to leave things for good.

I think maybe the original idea was something like 2. — until the notification history record is completed, it should be queued and somehow completed by the cron.

JanJakes (approved):

Aschepikov (approved):
GJ!

PR #5401, opened 2024-01-26
Mailpoet 5796 inline notices

brezocordero:
## Description

This PR adds the sender domain notices to Automations.
These notices are only warnings as we don't block sending for Automations.

## Code review notes

_N/A_

## QA notes

To test:

Site needs to be connected to the shop with a valid key
Edit an automation that has a couple of email steps
Click on the email step
Check that you see the notices under the from email as the screenshot in the ticket
Authenticate a domain, check the notice disappears
Click on another email step
Check that if you add the same domain that you authenticated in the other step, you don't see the notice.

You can also check that the sender notices in Settings or inside of a newsletter/post notification have not changed.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5843]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5843]: https://mailpoet.atlassian.net/browse/MAILPOET-5843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/email-panel.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/email-panel.tsx
@@ -2,7 +2,9 @@ import { ComponentProps } from 'react';
 import { PanelBody, TextareaControl, TextControl } from '@wordpress/components';
 import { dispatch, useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
+import { isEmail } from 'common/functions';
 import { ShortcodeHelpText } from './shortcode-help-text';
+import { SenderDomainNotice } from './sender-domain-notice';
 import { PlainBodyTitle } from '../../../../../editor/components';
 import { storeName } from '../../../../../editor/store';
 import { StepName } from '../../../../../editor/components/panel/step-name';
@@ -95,6 +97,12 @@ export function EmailPanel(): JSX.Element {
           )
         }
       />
+      {window.mailpoet_mss_active &&
+        isEmail((selectedStep.args.sender_address as string) ?? '') && (
+          <SenderDomainNotice
+            email={(selectedStep.args.sender_address as string) ?? ''}
+          />
+        )}
       <SingleLineTextareaControl
         className={
           subjectErrorMessage ? 'mailpoet-automation-field__error' : ''
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/sender-domain-notice.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/sender-domain-notice.tsx
@@ -1,0 +1,87 @@
+import { useState, useMemo, useEffect } from 'react';
+import { dispatch } from '@wordpress/data';
+import { extractEmailDomain } from 'common/functions';
+import { SenderEmailAddressWarning } from '../../../../../../common/sender-email-address-warning';
+import { MailPoet } from '../../../../../../mailpoet';
+import { getContext } from '../../../context';
+import { storeName } from '../../../../../editor/store';
+
+type SenderDomainInlineNoticeProps = {
+  email: string;
+};
+
+/**
+ * SenderDomainStatusNotice is a wrapper for SenderEmailAddressWarning to use with Automations.
+ * It populates the props (mostly) with data from the context, instead of the Window object.
+ * It still uses the window object for values that are present in all MailPoet pages
+ * like mailpoet_mss_active or subscribers_count further down the tree.
+ *
+ * It updates the context when the user authorizes the email or domain
+ * so that other email steps in the automation have an updated configuration.
+ *
+ * @param email - FROM address email
+ * @returns JSX.Element
+ */
+function SenderDomainStatusNotice({
+  email,
+}: SenderDomainInlineNoticeProps): JSX.Element {
+  const [senderDomainsConfig, setSendersDomainConfig] = useState(
+    getContext().senderDomainsConfig,
+  );
+
+  const domain = extractEmailDomain(email);
+  const isPartiallyVerifiedDomain =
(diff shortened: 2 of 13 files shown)


mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/sender-domain-notice.tsx
```diff
@@ -0,0 +1,87 @@
...
+    senderDomainsConfig.authorizedEmails.includes(email) ||
+    senderDomainsConfig.verifiedSenderDomains.includes(domain);
+  const showSenderDomainWarning =
+    !senderDomainsConfig.verifiedSenderDomains.includes(domain);
+
+  useEffect(() => {
+    dispatch(storeName).alterContext('mailpoet', 'senderDomainsConfig', {
+      ...senderDomainsConfig,
+    });
+  }, [senderDomainsConfig]);
```
JanJakes: I noticed this component has a state that takes defaults from the context, and on change, it syncs them back to the context. Is it then necessary to have the state? Maybe a `useSelect()` on the context would work to read the value and the `dispatch` to update it? It seems like the state might be just a "copy".

brezocordero: Thanks for the early feedback @JanJakes I really appreciate it.
I will follow your suggestion.

brezocordero: Done in https://github.com/mailpoet/mailpoet/pull/5401/commits/52ca5a21194e500f80117e660bd61dcdff071303

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/email-panel.tsx
```diff
@@ -95,6 +97,12 @@ export function EmailPanel(): JSX.Element {
           )
         }
       />
+      {window.mailpoet_mss_active &&
```
websupporter, line 100: Not a blocker, but why is this information in `window` and not in the context?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/sender-domain-notice.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * It updates the context when the user authorizes the email or domain
+ * so that other email steps in the automation have an updated configuration.
+ *
+ * @param email - FROM address email
+ * @returns JSX.Element
+ */
+function SenderDomainStatusNotice({
+  email,
+}: SenderDomainInlineNoticeProps): JSX.Element {
+  const [senderDomainsConfig, setSendersDomainConfig] = useState(
```
websupporter: I get a blank screen with an API key from the creator plan. As far as I understand, that is because senderDomainsConfig can be `null` when the key is not able to send via MSS or another sending method is used. I pushed dc8ec6f47 to account for that configuration.

mailpoet/lib/Services/AuthorizedSenderDomainController.php
```diff
@@ -311,4 +311,25 @@ public function isAuthorizedDomainRequiredForNewCampaigns(): bool {
...
   }
+
+  public function getContextData(): array {
+    return [
+      'verifiedSenderDomains' => $this->getFullyVerifiedSenderDomains(true),
+      'partiallyVerifiedSenderDomains' => $this->getPartiallyVerifiedSenderDomains(true),
+      'allSenderDomains' => $this->getAllSenderDomains(),
+      'senderRestrictions' => [
+        'lowerLimit' => self::LOWER_LIMIT,
+        'upperLimit' => self::UPPER_LIMIT,
```
websupporter, line 322: `upperLimit` seems not to be part of the TSX definition and not used anywhere. I wonder what this value is good for.

websupporter (approved):
I found one error, which I fixed in dc8ec6f476238594fff3711087dfe90c3333061b. And I have two questions, but I think this can move forward to QA, as to me it seems this is nothing major and we want to release this change with the next release. @brezocordero can you check if my fix is correct?

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/email-panel.tsx
brezocordero: It is always loaded in the window object in every page. Coming from [PageRenderer](https://github.com/mailpoet/mailpoet/blob/5cd7dad6268f0735082ee856b368dfe5e4afe615/mailpoet/lib/AdminPages/PageRenderer.php#L183).

mailpoet/assets/js/src/automation/integrations/mailpoet/steps/send-email/edit/sender-domain-notice.tsx
brezocordero: Thanks for the fix

mailpoet/lib/Services/AuthorizedSenderDomainController.php
brezocordero: It is a leftover from a first iteration. I have added it to the cleanup ticket.

brezocordero:
> can you check if my fix is correct?

It works thanks. I didn't know you could activate MSS with a creator key. Maybe a different issue.

Aschepikov:
Hi @brezocordero  I don't see the notice for the branded domain with no auth, in the settings I see it but in the automations no, can you take a look pls 
image

brezocordero:
> I don't see the notice for the branded domain with no auth, in the settings I see it but in the automations no, can you take a look pls

I can see it locally @Aschepikov . Is your key valid and for a plan with MSS?

Aschepikov:
Looks like a cache issue

Aschepikov (approved):
GJ!

PR #5405, opened 2024-01-29
Add a warning for opening a sent email [MAILPOET-5851]

pavel-mailpoet:
## Description

We have the same warning in the old email editor. If a user opens a previously sent email, we add a warning and prevent moving forward by disabling the “Send” button.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

[MAILPOET-5851]

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5851]: https://mailpoet.atlassian.net/browse/MAILPOET-5851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/components/notices/sent-email-notice.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/notices/sent-email-notice.tsx
@@ -1,0 +1,31 @@
+import { dispatch, useSelect } from '@wordpress/data';
+import { useEffect } from '@wordpress/element';
+import { store as noticesStore } from '@wordpress/notices';
+import { __ } from '@wordpress/i18n';
+
+import { storeName } from '../../store';
+
+export function SentEmailNotice() {
+  const { isEmailSent } = useSelect(
+    (select) => ({
+      isEmailSent: select(storeName).isEmailSent(),
+    }),
+    [],
+  );
+
+  useEffect(() => {
+    if (isEmailSent) {
+      dispatch(noticesStore).createErrorNotice(
+        __(
+          'This email has already been sent. It can be edited, but not sent again. Duplicate this email if you want to send it again.',
+          'mailpoet',
+        ),
+        {
+          isDismissible: false,
+        },
+      );
+    }
+  }, [isEmailSent]);
+
+  return null;
+}
--- a/mailpoet/assets/js/src/email-editor/engine/components/notices/sent-email-notice.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/notices/sent-email-notice.tsx
@@ -15,7 +15,8 @@ export function SentEmailNotice() {
 
   useEffect(() => {
     if (isEmailSent) {
-      dispatch(noticesStore).createErrorNotice(
+      dispatch(noticesStore).createNotice(
+        'warning',
         __(
           'This email has already been sent. It can be edited, but not sent again. Duplicate this email if you want to send it again.',
           'mailpoet',
--- a/mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
+++ b/mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
@@ -16,10 +16,6 @@ class NewsletterListener {
     $this->wp = $wp;
   }
 
-  public function prePersist(): void {
-    // the tests fail without this method
-  }
-
   public function preUpdate(LifecycleEventArgs $eventArgs): void {
     $entity = $eventArgs->getEntity();
     if (!$entity instanceof NewsletterEntity) {
--- a/mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
+++ b/mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
@@ -8,8 +8,7 @@ use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Doctrine\ORM\Event\LifecycleEventArgs;
 
 class NewsletterListener {
-  /** @var WPFunctions */
-  private $wp;
+  private WPFunctions $wp;
 
   public function __construct(
     WPFunctions $wp
--- a/mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
+++ b/mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
@@ -1,0 +1,50 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Doctrine\EventListeners;
+
+use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Entities\WpPostEntity;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoetVendor\Doctrine\ORM\Event\LifecycleEventArgs;
+
+class NewsletterListener {
+  /** @var WPFunctions */
+  private $wp;
+
+  public function __construct(
+    WPFunctions $wp
+  ) {
+    $this->wp = $wp;
+  }
+
+  public function prePersist(): void {
+    // the tests fail without this method
+  }
+
+  public function preUpdate(LifecycleEventArgs $eventArgs): void {
+    $entity = $eventArgs->getEntity();
+    if (!$entity instanceof NewsletterEntity) {
+      return;
+    }
+
+    $unitOfWork = $eventArgs->getEntityManager()->getUnitOfWork();
+
+    /** @var array{status: array{0: string, 1: string}} $changeSet */
+    $changeSet = $unitOfWork->getEntityChangeSet($entity);
+    if (!isset($changeSet['status'])) {
+      return;
+    }
+
+    [$oldStatus, $newStatus] = $changeSet['status'];
+
+    if ($oldStatus !== NewsletterEntity::STATUS_SENT && $newStatus === NewsletterEntity::STATUS_SENT) {
+      $post = $entity->getWpPost();
+      if ($post instanceof WpPostEntity) {
+        $this->wp->wpUpdatePost([
+          'ID' => $post->getId(),
+          'post_status' => NewsletterEntity::STATUS_SENT,
+        ]);
+      }
+    }
+  }
+}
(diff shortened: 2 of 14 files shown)


mailpoet/assets/js/src/email-editor/engine/components/notices/sent-email-notice.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { isEmailSent } = useSelect(
+    (select) => ({
+      isEmailSent: select(storeName).isEmailSent(),
+    }),
+    [],
+  );
+
+  useEffect(() => {
+    if (isEmailSent) {
+      dispatch(noticesStore).createErrorNotice(
```
lysyjan: Have you considered the use of a warning instead of an error?
The warning notice seems better to me because it looks better and is more accurate to this type of message.

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
```diff
@@ -0,0 +1,50 @@
...
+namespace MailPoet\Doctrine\EventListeners;
+
+use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Entities\WpPostEntity;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoetVendor\Doctrine\ORM\Event\LifecycleEventArgs;
+
+class NewsletterListener {
+  /** @var WPFunctions */
+  private $wp;
```
lysyjan: I noticed you used a typed property in the class EntityManagerFactory.php for the new property, which is pretty nice 👍. Maybe we could also use typed properties in this class and its test. What do you think?

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
```diff
@@ -0,0 +1,50 @@
...
+
+  public function __construct(
+    WPFunctions $wp
+  ) {
+    $this->wp = $wp;
+  }
+
+  public function prePersist(): void {
```
lysyjan: I don't think this is necessary. We could update `EventListenersBaseTest::replaceListeners()` by adding an optional parameter with listener events that should be replaced. When the new parameter is optional with the default value `[Events::prePersist, Events::preUpdate]`, then we don't need to touch the other tests extending this class. What do you think about this approach?

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
```diff
@@ -0,0 +1,50 @@
...
+
+  public function preUpdate(LifecycleEventArgs $eventArgs): void {
+    $entity = $eventArgs->getEntity();
+    if (!$entity instanceof NewsletterEntity) {
+      return;
+    }
+
+    $unitOfWork = $eventArgs->getEntityManager()->getUnitOfWork();
+
+    /** @var array{status: array{0: string, 1: string}} $changeSet */
```
lysyjan, line 27: FYI, I was digging a bit more about this line required to avoid an error from PHPStan, and I found [this line in our PHPStan baseline](https://github.com/mailpoet/mailpoet/blob/d4d4db7cc5e796a3dbb8cc4e988bbfa0393fc4ca/mailpoet/tasks/phpstan/phpstan-8-baseline.neon#L209-L211). 😬

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    [$oldStatus, $newStatus] = $changeSet['status'];
+
+    if ($oldStatus !== NewsletterEntity::STATUS_SENT && $newStatus === NewsletterEntity::STATUS_SENT) {
+      $post = $entity->getWpPost();
+      if ($post instanceof WpPostEntity) {
+        $this->wp->wpUpdatePost([
+          'ID' => $post->getId(),
+          'post_status' => NewsletterEntity::STATUS_SENT,
```
lysyjan, line 40: We set here a post status that is not a part of supported status by default. Maybe we should add this statuses as [a custom status](https://wordpress.org/documentation/article/post-status/#custom-status). What do you think about it? Perhaps this can be solved in another ticket. I just wanted to mention it here to avoid forgetting.

lysyjan (commented):
Good job! I have a few little remarks and will be happy to discuss them.

mailpoet/assets/js/src/email-editor/engine/components/notices/sent-email-notice.tsx
pavel-mailpoet: Hmm, the older email editor displays it as an error, so I made it the same: https://d.pr/i/yrQ6KB

warning is also ok: https://d.pr/i/MuLO7y

I will change it.

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
pavel-mailpoet: Hmm, interesting. It didn't occur to me to check the baseline file. What do you suggest? Should we keep this line here, add this file to the base line as well or something else?

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
pavel-mailpoet: > What do you think about it?

I am completely not sure. What are the problems or benefits? I mean, this code works, so I am not sure what happens if we don't do it.

And where do you recommend calling the `register_post_status` function?

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
lysyjan: I would keep it as it is. If I'm not wrong, we have a ticket for removing baseline files.
I wanted to let you know only the result of my finding.

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
lysyjan: >What are the problems or benefits?

I would say we could find some issues when we use our custom post types somewhere in WP UI. So, maybe we can wait with that. I would like to know @costasovo opinion here.

>And where do you recommend calling the register_post_status function?

I think [the place](https://github.com/mailpoet/mailpoet/blob/f9a7ffc337d5fd31c1ff997497ae9aff298c450d/mailpoet/lib/EmailEditor/Engine/EmailEditor.php) when we register custom post type should be fine.

costasovo: > What are the problems or benefits?

@pavel-mailpoet  I share the same concern as @lysyjan. In addition, I haven't searched if there are any existing issues, but I think there is a risk that a post with an unknown status might be treated improperly. For example, some post cleanup plugins may remove it, etc.
Woo [registers the order statuses](https://github.com/woocommerce/woocommerce/blob/0e1239867ad9080ba25f6ba1ce5391db5be95d78/plugins/woocommerce/includes/class-wc-post-types.php#L559).

pavel-mailpoet: I am still trying to figure out the implications. What happens (and changes) when we make the call and register the status? What is the difference? Do I understand correctly that we would call the `register_post_status` only when a user visits the editor? So, if the editor is not active, is the status still unknown?

lysyjan: > What happens (and changes) when we make the call and register the status?

- I'm afraid that our post without registered status could be unexpectedly deleted or updated. For example, the status could be changed to the status `draft`.
- Another aspect could be querying posts via WP_Query because posts with an unknown status would be ignored.
- Integrations with other plugins should be easier for developers when the status is registered.

lysyjan (approved):
Thanks!
LGTM

mailpoet/lib/Doctrine/EventListeners/NewsletterListener.php
costasovo: >  Do I understand correctly that we would call the register_post_status only when a user visits the editor? So, if the editor is not active, is the status still unknown?

@pavel-mailpoet We need to have it registered similarly to the custom post type, as Jan suggested. The [custom post type is registered on every page](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Config/Initializer.php#L557-L560). We also need to consider if we want to put this to Engine namespace or Integration. Currently, we allow integrations to register and configure their own post types, so I would suggest adding it to the integration code.
I think we need to add the custom status after the post type is registered ([here](https://github.com/mailpoet/mailpoet/blob/f9a7ffc337d5fd31c1ff997497ae9aff298c450d/mailpoet/lib/EmailEditor/Engine/EmailEditor.php#L21)) so we will probably need another action in the engine where integration could do something (register custom statuses) after the engine registers all post types.

costasovo: I asked Chat GPT about the consequences of not registering the status. 

Here is what I got:

In WordPress, custom post statuses can be a powerful feature for managing content in a way that goes beyond the default statuses such as "Publish," "Draft," "Pending," etc. To create a custom post status, you indeed use the register_post_status function, which is separate from registering a custom post type (register_post_type). Here's an overview of why and how you should use register_post_status:


#### Why Register
Registering a custom post status using register_post_status integrates your custom status into the WordPress ecosystem properly. This allows you to:

1. Ensure compatibility with WordPress core functions and hooks.
2. Make your custom status available in the WordPress Admin UI, depending on how you configure it.
3. Utilize WordPress functions to set or change the post status, ensuring that your actions are compatible with WordPress's way of handling data.

#### Consequences of Not Registering

1. Lack of Integration: If you don't register the post status and instead directly manipulate the database via SQL to set a custom status, you might face issues with WordPress core functions not recognizing your custom status. This can lead to unexpected behavior in both the admin area and your theme.
2. Upgrade Issues: Directly manipulating the database can lead to compatibility issues when WordPress updates. WordPress updates can change database structures, behaviors, or how core functions work, potentially breaking your custom implementation.
3. Security and Stability: WordPress provides a structured way to interact with its database through functions and APIs, which includes security measures like data sanitization and validation. Bypassing these by directly using SQL queries can introduce security vulnerabilities or data integrity issues.

#### Conclusion
While directly setting a post status via SQL might seem like a shortcut, it's generally advisable to use the WordPress APIs and functions provided. This ensures compatibility, security, and ease of maintenance. Registering custom post statuses through register_post_status is the recommended approach for adding custom statuses to your WordPress site.

pavel-mailpoet: Thanks @costasovo I did the change yesterday and now I fully understand why we need it.

Aschepikov (approved):
GJ!

PR #5406, opened 2024-01-29
Reduce sender domain authentication limits [MAILPOET-5856]

brezocordero:
## Description

Reduce sender domain authentication limits: Small senders to 100 subscribers and big senders to 200 subscribers

Please ~do not~ merge.

## Code review notes

_N/A_

## QA notes

Please ~do not~ merge.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5856]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5856]: https://mailpoet.atlassian.net/browse/MAILPOET-5856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Services/AuthorizedSenderDomainController.php
+++ b/mailpoet/lib/Services/AuthorizedSenderDomainController.php
@@ -19,8 +19,8 @@ class AuthorizedSenderDomainController {
   const AUTHORIZED_SENDER_DOMAIN_ERROR_NOT_CREATED = 'Sender domain does not exist';
   const AUTHORIZED_SENDER_DOMAIN_ERROR_ALREADY_VERIFIED = 'Sender domain already verified';
 
-  const LOWER_LIMIT = 500;
-  const UPPER_LIMIT = 1000;
+  const LOWER_LIMIT = 100;
+  const UPPER_LIMIT = 200;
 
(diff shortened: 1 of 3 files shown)


johnolek (approved):
Looks good to me. My only concern is potentially needing to change these again in the future and needing to coordinate any such changes with the MSS. 

I like the idea of having the MSS include its currently configured lower/upper limits in some API endpoint, and having the hardcoded values in the plugin be fallbacks. It would allow immediate fine tuning without having to worry about users updating their plugin to see the correct notices.

Of course that could be handled in a separate PR if it's even something we want to consider.

brezocordero:
> I like the idea of having the MSS include its currently configured lower/upper limits in some API endpoint, and having the hardcoded values in the plugin be fallbacks. It would allow immediate fine tuning without having to worry about users updating their plugin to see the correct notices.

Thanks for the suggestion @johnolek 

We have not yet decided if we are merging this PR and I don't have enough visibility to know if we we are going to continue changing the values.
If we are, you have a good point because we would not depend on a plugin release.

Let me follow up on this once I have more information to understand if the change is worth it.

Aschepikov (approved):
GJ!

PR #5416, opened 2024-02-02
Fix "The EntityManager is closed" in sending [MAILPOET-5886]

costasovo:
## Description

This PR fixes an error happening when sending all types of emails on a site with MySQL 8. It causes that sending is very slow and you can see The EntityManager is closed errors in logs.

## Code review notes
I considered configuring the sentAt columns `insertable: false`, but using a default value seemed more common in Doctrine.
I also ran all integration tests with MySQL8 and fixed one failing test I found.

## QA notes

Please see the replication instructions in the Jira ticket.

To test the migration, please generate a reasonable number of emails sent in an invalid state (Not sent yet). This can be done by following the instructions and sending a couple of emails. After you run the migration the emails should display correct status.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5886]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5886]: https://mailpoet.atlassian.net/browse/MAILPOET-5886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Entities/StatisticsNewsletterEntity.php
+++ b/mailpoet/lib/Entities/StatisticsNewsletterEntity.php
@@ -36,19 +36,21 @@ class StatisticsNewsletterEntity {
   private $subscriber;
 
   /**
-   * @ORM\Column(type="datetimetz", nullable=true)
-   * @var \DateTimeInterface|null
+   * @ORM\Column(type="datetimetz", nullable=false)
+   * @var \DateTimeInterface
    */
   private $sentAt;
 
   public function __construct(
     NewsletterEntity $newsletter,
     SendingQueueEntity $queue,
-    SubscriberEntity $subscriber
+    SubscriberEntity $subscriber,
+    \DateTimeInterface $sentAt = null
   ) {
     $this->newsletter = $newsletter;
     $this->queue = $queue;
     $this->subscriber = $subscriber;
+    $this->sentAt = $sentAt ?: new \DateTimeImmutable();
   }
 
   /**
--- a/mailpoet/lib/Migrations/App/Migration_20240202_130053_App.php
+++ b/mailpoet/lib/Migrations/App/Migration_20240202_130053_App.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Migrations\App;
+
+use MailPoet\Entities\NewsletterEntity;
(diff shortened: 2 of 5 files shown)


mailpoet/lib/Entities/StatisticsNewsletterEntity.php
```diff
@@ -36,19 +36,21 @@ class StatisticsNewsletterEntity {
...
     NewsletterEntity $newsletter,
     SendingQueueEntity $queue,
-    SubscriberEntity $subscriber
+    SubscriberEntity $subscriber,
+    \DateTimeInterface $sentAt = null
   ) {
     $this->newsletter = $newsletter;
     $this->queue = $queue;
     $this->subscriber = $subscriber;
+    $this->sentAt = $sentAt ?: new \DateTimeImmutable();
```
JanJakes, line 53: Sometimes we use `Carbon::createFromTimestamp($this->wp->currentTime('timestamp'))`. But I'm not sure if that's needed at all. Is there a difference?

JanJakes (approved):
@costasovo LGTM, thanks. I added one small question, but approving, feel free to move this forward.

JanJakes:
@costasovo One more question:

> I also ran all integration tests with MySQL8 and fixed one failing test I found.

You told me we don't run them on version 8. Could we fix that in this PR?

mailpoet/lib/Entities/StatisticsNewsletterEntity.php
costasovo: I pass `Carbon::createFromTimestamp($this->wp->currentTime('timestamp'))` date into the constructor but as a default value (used now only in tests) I used the DateTimeImmutable because I didn't want to add a dependency on a WP function into the entity constructor.

costasovo:
@JanJakes Thanks for the quick review. I'm moving this forward.

> You told me we don't run them on version 8. Could we fix that in this PR?

I looked into that, but it seems that MySQL 8 requires additional changes in the command we pass to the container. I didn't add it to this PR because of time reasons. I created a maintenance ticket for that  - MAILPOET-5888

Aschepikov (approved):

PR #5418, opened 2024-02-07
Add migration to fix sending tasks stuck in "invalid" state [MAILPOET-5887]

JanJakes:
## Description

Due to the recent issues with refactoring sending code to Doctrine in combination with https://github.com/mailpoet/mailpoet/pull/4797, there might be some sending tasks and newsletter entities incorrectly marked as invalid respectively sending. This PR adds a migration to fix those.

## Code review notes

It's hard to test the invalid tasks, and I wrote tests for them.
That said, if you have a site with some scheduled emails, notifications, and historical data, you can try to update to this branch, deactivate & activate the plugin to ensure migrations run, and then make sure all sending data still looks fine.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5887]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5887]: https://mailpoet.atlassian.net/browse/MAILPOET-5887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
+++ b/mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
@@ -18,7 +18,11 @@ use MailPoet\Migrator\AppMigration;
  */
 class Migration_20240207_105912_App extends AppMigration {
   public function run(): void {
-    // pause invalid tasks with unprocessed subscribers
+    $this->pauseInvalidTasksWithUnprocessedSubscribers();
+    $this->completeInvalidTasksWithAllSubscribersProcessed();
+  }
+
+  private function pauseInvalidTasksWithUnprocessedSubscribers(): void {
     $ids = $this->entityManager->createQueryBuilder()
       ->select('DISTINCT t.id')
       ->from(ScheduledTaskEntity::class, 't')
@@ -45,8 +49,9 @@ class Migration_20240207_105912_App extends AppMigration {
       ->setParameter('ids', $ids)
       ->getQuery()
       ->execute();
+  }
 
-    // complete invalid tasks with all subscribers processed, mark newsletters as sent
+  private function completeInvalidTasksWithAllSubscribersProcessed(): void {
     $result = $this->entityManager->createQueryBuilder()
       ->select('DISTINCT t.id, n.id AS nid')
       ->from(ScheduledTaskEntity::class, 't')
--- a/mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
+++ b/mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
@@ -5,7 +5,9 @@ namespace MailPoet\Migrations\App;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\ScheduledTaskSubscriberEntity;
+use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\Migrator\AppMigration;
+use MailPoetVendor\Doctrine\DBAL\Connection;
 
 /**
  * We've had a set of bugs where campaign type newsletters (see NewsletterEntity::CAMPAIGN_TYPES),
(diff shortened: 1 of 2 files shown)


websupporter:
Hi @JanJakes,
thanks for working on this! Two things which are not happening in the migration, but are necessary:

`wp_mailpoet_newsletters.sent_at` needs to be updated
`wp_mailpoet_sending_queues.count_processed` needs to be updated
`wp_mailpoet_sending_queues.count_to_processed` needs to be updated

The counts and the sent_at are displayed in the UI. The counts render the circle how many emails have been sent and the sent_at gets rendered as well in the last column.

JanJakes:
@websupporter Great point, thanks for catching that! Just nothing here for myself that the `sent_at` status was not handled in https://github.com/mailpoet/mailpoet/pull/5416 either, and I'll try to fix it, also checking if we can backfill the failed newsletter stats records.

mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
```diff
@@ -0,0 +1,173 @@
...
+
+    // get affected newsletter IDs
+    $ids = $this->entityManager->createQueryBuilder()
+      ->select('n.id')
+      ->from(NewsletterEntity::class, 'n')
+      ->where('n.status = :sent')
+      ->andWhere('n.sentAt IS NULL')
+      ->setParameter('sent', NewsletterEntity::STATUS_SENT)
+      ->getQuery()
+      ->getSingleColumnResult();
```
JanJakes, line 138: Do you think we need any more filtering here (like only some newsletter types, etc.)?
I suppose that state is invalid, so shouldn't occur, apart from the migrated newsletters.
The original PR had no additional filtering: https://github.com/mailpoet/mailpoet/pull/5416/files

mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
```diff
@@ -0,0 +1,173 @@
...
+      ->join('q.task', 't')
+      ->join('t.subscribers', 's')
+      ->leftJoin(StatisticsNewsletterEntity::class, 'ns', 'WITH', 'ns.queue = q AND ns.subscriber = s.subscriber')
+      ->where('q.newsletter IN (:ids)')
+      ->andWhere('ns.id IS NULL')
+      ->andWhere('s.processed = :processed')
+      ->andWhere('s.failed = :ok')
+      ->setParameter('ids', $ids)
+      ->setParameter('processed', ScheduledTaskSubscriberEntity::STATUS_PROCESSED)
+      ->setParameter('ok', ScheduledTaskSubscriberEntity::FAIL_STATUS_OK)
```
JanJakes: I suppose we save newsletter statistics only for those, as the entity has a `sentAt` field.

mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
websupporter: No, I think that makes sense.

mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
websupporter: Actually, we don't, if I understand it correctly here: https://github.com/mailpoet/mailpoet/blob/bbe471e6531a1e821020097303d89f557c7325b2/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L512-L543

In L521 we check the `sendResult['response']` and process errors, but as we do not return, we will create the entries in L543.

websupporter:
Tested it in my installation with my test data and it worked :tada:

mailpoet/lib/Migrations/App/Migration_20240207_105912_App.php
JanJakes: @websupporter Ahh, right. So they must be `processed` (otherwise they wouldn't get to that stats saving logic), but `failed` status we seem to ignore.

websupporter (approved):
lgtm.

Checked it with some demo data and it works as expected.

PR #5420, opened 2024-02-08
Render forms on empty product pages [MAILPOET-5859]

johnolek:
## Description

The current logic for form rendering assumes that a single product page will trigger the filter `the_content`. This is not always true. In cases where the product has no description, WooCommerce does not render the description tab at all, which is where that filter usually fires (in my testing).

This change ensures that we still give these forms a chance to render on such pages.

## Code review notes

_N/A_

## QA notes

This does not make the "Below pages" placement option work for product pages without content because they don't render any content. This seems reasonable to me.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5859](https://mailpoet.atlassian.net/browse/MAILPOET-5859)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5859]: https://mailpoet.atlassian.net/browse/MAILPOET-5859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Form/DisplayFormInWPContent.php
+++ b/mailpoet/lib/Form/DisplayFormInWPContent.php
@@ -130,7 +130,7 @@ class DisplayFormInWPContent {
    * @return void
    */
   public function maybeRenderFormsInFooter(): void {
-    if ($this->wp->isArchive() || $this->wp->isFrontPage() || $this->wp->isHome()) {
+    if ($this->wp->isArchive() || $this->wp->isFrontPage() || $this->wp->isHome() || $this->isWooProductPageWithoutContent()) {
       $formMarkup = $this->getFormMarkup();
       if (!empty($formMarkup)) {
         $this->assetsController->setupFrontEndDependencies();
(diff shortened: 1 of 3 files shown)


brezocordero (dismissed):
Nice fix @johnolek 
I have checked if there could be a way to have `woocommerce_product_loop_end` display the form that would also be displayed with `wp_footer` with the new condition and I couldn't find any.
I also tested the form is displayed when viewing a product without a description using Storefront.

Aschepikov:
Hi @johnolek there are some failed tests can you rebase it pls

johnolek:
@Aschepikov Rebased and looks good to go

Aschepikov:
Hi @johnolek  I can't trigger the form on the product with an empty description. I am using all types of forms with different settings and couldn't trigger it on it. Maybe I should do something else or can you check it, I can provide you a creds from my site

johnolek:
@Aschepikov It's still working as expected for me in my local environment, so sure, I can take a look at your test site.

Aschepikov:
@johnolek Sent you a creds via slack and also check pls the editing of the forms with fixed bar and slide in as I can't edit them and there is a console error
image
image

brezocordero (approved):

Aschepikov (approved):
GJ!

PR #5421, opened 2024-02-08
Keep track of onboarding progress [MAILPOET-5216]

johnolek:
## Description

This introduces a new setting to track where a user is in the welcome wizard. If a user closes their browser before completing the welcome wizard and returns via the button on the landing page, they should be taken back to the last step they were looking at.

I also updated the button on the landing page so it will say "Continue setup" instead of "Begin setup" for these cases.

I didn't want it to be impossible to return to a previous step without changing the url manually, so I updated the steps component to allow clicking to return to previous steps.

## Code review notes

I ran into an annoying issue where including `useSetting` in the main welcome wizard app was causing a re-render every time state changed, e.g. typing a letter in an input field. This was causing the input to lose focus. More details in 0d434bb4e411133d5cfadd24e7ad665abd40d7fb, where I modified the behavior. I would really like to figure out if we can stop `useSetting` from triggering re-renders in general because I think it might be causing unnecessary renders wherever we're using it. I tried a few things but ultimately wasn't successful. 

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5216

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/assets/css/src/components/_steps.scss
+++ b/mailpoet/assets/css/src/components/_steps.scss
@@ -72,6 +72,10 @@
   .mailpoet-step-done & {
     background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='6' height='6' viewBox='0 0 6 6'%3E%3Cpath fill='%237ED321' fill-rule='nonzero' d='M4.647.319c.246-.35.715-.423 1.048-.165.334.257.404.75.158 1.098L3.084 5.181c-.28.397-.835.429-1.154.066L.199 3.283c-.281-.319-.262-.816.042-1.11.305-.295.78-.275 1.06.044l1.115 1.266L4.646.319z' /%3E%3C/svg%3E%0A");
   }
+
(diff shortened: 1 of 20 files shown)


JanJakes (approved):
LGTM!

JanJakes:
@johnolek For the `useSetting` issue, I think that's because we have a [custom hook for that](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/settings/store/hooks/use-setting.ts), and that hook returns a new array on each call.

I think it should be rather something like this:

```ts
  return useMemo<ReturnType<typeof useSetting>>(
    () => [getValue(path), (value) => setValue(path, value)],
    [getValue, path, setValue],
  );
```

(I didn't test it. thouth.)

Aschepikov:
Hi @johnolek there are some failed tests can you rebase it pls

johnolek:
@Aschepikov Rebased and should be good to go

Aschepikov (approved):
GJ!

PR #5428, opened 2024-02-12
Allow text highlighting in the email editor [MAILPOET-5809]

lysyjan:
## Description

This PR reactivates text highlighting in the new email editor.

## Code review notes

- I decided to use post-processing to replace `mark` tags with `span` because it solves replacing for all blocks.
- I fixed some basic line-height styles.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5809]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5809]: https://mailpoet.atlassian.net/browse/MAILPOET-5809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Postprocessors/HighlightingPostprocessor.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Postprocessors/HighlightingPostprocessor.php
@@ -1,0 +1,16 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Engine\Renderer\Postprocessors;
+
+/**
+ * This postprocessor replaces <mark> tags with <span> tags because mark tags are not supported across all email clients
+ */
+class HighlightingPostprocessor implements Postprocessor {
+  public function postprocess(string $html): string {
+    return str_replace(
+      ['<mark', '</mark>'],
+      ['<span', '</span>'],
+      $html
+    );
+  }
+}
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Renderer.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Renderer.php
@@ -15,7 +15,7 @@ class Renderer {
   private $blocksRegistry;
 
   /** @var ProcessManager */
-  private $preprocessManager;
+  private $processManager;
 
   /** @var SettingsController */
   private $settingsController;
@@ -33,7 +33,7 @@ class Renderer {
     SettingsController $settingsController
   ) {
     $this->cssInliner = $cssInliner;
-    $this->preprocessManager = $preprocessManager;
+    $this->processManager = $preprocessManager;
     $this->blocksRegistry = $blocksRegistry;
     $this->settingsController = $settingsController;
   }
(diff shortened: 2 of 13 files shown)


mailpoet/tests/unit/EmailEditor/Engine/Renderer/ProcessManagerTest.php
```diff
@@ -39,7 +39,7 @@ public function testItCallsPreprocessorsProperly(): void {
     $secondPreprocessor = $this->createMock(TopLevelPreprocessor::class);
     $secondPreprocessor->expects($this->once())->method('preprocess')->willReturn([]);
 
-    $preprocessManager = new PreprocessManager($cleanup, $topLevel, $blocksWidth, $typography, $spacing);
+    $preprocessManager = new ProcessManager($cleanup, $topLevel, $blocksWidth, $typography, $spacing);
     $preprocessManager->registerPreprocessor($secondPreprocessor);
```
costasovo: The variable name $preprocessManager could be a bit confusing after the renaming

mailpoet/lib/EmailEditor/Engine/Renderer/Renderer.php
```diff
@@ -85,6 +85,7 @@ public function render(\WP_Post $post, string $subject, string $preHeader, strin
 
     $templateWithContentsDom = $this->inlineCSSStyles($templateWithContents);
     $templateWithContents = $this->postProcessTemplate($templateWithContentsDom);
+    $templateWithContents = $this->preprocessManager->postprocess($templateWithContents);
```
costasovo: Also here `$this->preprocessManager` should probably change as it is no longer only preprocessor.

mailpoet/lib/EmailEditor/Engine/Renderer/Postprocessors/HighlightingPostprocessor.php
```diff
@@ -0,0 +1,16 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Engine\Renderer\Postprocessors;
+
+/**
+ * This postprocessor replaces <mark> tags with <span> tags because mark tags are not supported across all email clients
+ */
+class HighlightingPostprocessor implements Postprocessor {
```
costasovo, line 8: Just an idea. Maybe we could do this also in a preprocessor by changing the innerHtml and saved content of the individual blocks. With this approach we wouldn't need to introduce postprocessors. On the other hand I don't see any other significant advantages. Maybe we could add some specific conditions per block if needed? What do you tnink?

costasovo (commented):
Great job. I found a few variables that we could name better and added a comment about a possible alternative approach using a preprocessor.

mailpoet/lib/EmailEditor/Engine/Renderer/Postprocessors/HighlightingPostprocessor.php
lysyjan: This is a good idea, but I'm afraid of different rendering implementations of each block. In my opinion, this approach is safer for us.

mailpoet/tests/unit/EmailEditor/Engine/Renderer/ProcessManagerTest.php
lysyjan: This was changed when [I added a post-process class](https://github.com/mailpoet/mailpoet/pull/5428/commits/9d09f72bdf6e2d7817ef5030c2976a0c3c3d2373#diff-f538429edce390b87c33f415a2f4fe8ca5cec7c76bb6ac722a1d446c67422b5eR46-R47).

costasovo (approved):
LGTM!

PR #5437, opened 2024-02-16
New segment filter purchased with attribute [MAILPOET-5467]

johnolek:
## Description

This PR adds a new dynamic segment filter for "purchased with attribute".

Attributes in WooCommerce can be either global, which are saved as taxonomy/terms, or local, which are specific to a product. This filter accounts for both types, but with the limitation that it will only look at local attributes that were used to create product variations.

When local attributes are used for product variations, a new row is created in the `postmeta` table for each attribute/value pair, which makes it possible to write efficient queries to target them. If a simple product has multiple values for a local attribute, it's also saved in the `postmeta` table, but as a serialized string like this:

```
a:1:{s:15:"local-attribute";a:6:{s:4:"name";s:15:"Local Attribute";s:5:"value";s:31:"One | Two | Three | Four | Five";s:8:"position";i:0;s:10:"is_visible";i:1;s:12:"is_variation";i:0;s:11:"is_taxonomy";i:0;}}
```

Since the `postmeta` table is likely to be massive on any larger site and the `meta_value` column is not indexed, I don't see a way to query for this data in an efficient way.

I don't see this as a major issue because if a product has some unique local attributes, a user could simply use the "purchased product" filter.

## Code review notes

_N/A_

## QA notes

Please make sure to test both global (taxonomy based) and local attributes. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5467](https://mailpoet.atlassian.net/browse/MAILPOET-5467)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5467]: https://mailpoet.atlassian.net/browse/MAILPOET-5467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
@@ -57,7 +57,7 @@ export function PurchasedWithAttributeFields({
     productAttributeTermsOptionsRef.current = productAttributes[
       segment.attribute_taxonomy_slug
     ].terms.map((term) => ({
-      value: term.term_id,
+      value: term.term_id.toString(),
       label: term.name,
     }));
   }, [segment.attribute_taxonomy_slug, productAttributes]);
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
@@ -18,8 +18,9 @@ export function validatePurchasedWithAttribute(
 ): boolean {
   const purchasedProductWithAttributeIsInvalid =
     !formItems.operator ||
-    formItems.attribute_id === undefined ||
-    formItems.attribute_term_ids === undefined;
+    formItems.attribute_taxonomy_slug === undefined ||
+    !Array.isArray(formItems.attribute_term_ids) === undefined ||
+    formItems.attribute_term_ids.length === 0;
 
   return !purchasedProductWithAttributeIsInvalid;
 }
@@ -40,7 +41,7 @@ export function PurchasedWithAttributeFields({
 
   const productAttributesOptions = Object.values(productAttributes).map(
     (attribute) => ({
-      value: attribute.id,
+      value: attribute.taxonomy,
       label: attribute.label,
     }),
   );
@@ -48,18 +49,18 @@ export function PurchasedWithAttributeFields({
   const productAttributeTermsOptionsRef = useRef(null);
 
   useEffect(() => {
-    if (segment.attribute_id === undefined) {
+    if (segment.attribute_taxonomy_slug === undefined) {
       productAttributeTermsOptionsRef.current = null;
       return;
     }
 
     productAttributeTermsOptionsRef.current = productAttributes[
-      segment.attribute_id
+      segment.attribute_taxonomy_slug
     ].terms.map((term) => ({
       value: term.term_id,
       label: term.name,
     }));
-  }, [segment.attribute_id, productAttributes]);
+  }, [segment.attribute_taxonomy_slug, productAttributes]);
 
   useEffect(() => {
     if (
@@ -91,15 +92,17 @@ export function PurchasedWithAttributeFields({
         placeholder={__('Search attributes', 'mailpoet')}
         options={productAttributesOptions}
         value={filter((productAttributeOption) => {
-          if (segment.attribute_id === undefined) {
+          if (segment.attribute_taxonomy_slug === undefined) {
             return undefined;
           }
-          return segment.attribute_id === productAttributeOption.value;
+          return (
+            segment.attribute_taxonomy_slug === productAttributeOption.value
+          );
         }, productAttributesOptions)}
         onChange={(option: SelectOption): void => {
           void updateSegmentFilter(
             {
-              attribute_id: option.value,
+              attribute_taxonomy_slug: option.value,
               attribute_term_ids: [],
             },
             filterIndex,
--- a/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
+++ b/mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
@@ -9,6 +9,7 @@ import {
   AnyValueTypes,
   FilterProps,
   SelectOption,
+  WindowLocalProductAttributes,
   WindowProductAttributes,
   WooCommerceFormItem,
 } from '../../../types';
@@ -18,9 +19,15 @@ export function validatePurchasedWithAttribute(
 ): boolean {
   const purchasedProductWithAttributeIsInvalid =
     !formItems.operator ||
-    formItems.attribute_taxonomy_slug === undefined ||
-    !Array.isArray(formItems.attribute_term_ids) === undefined ||
-    formItems.attribute_term_ids.length === 0;
+    (formItems.attribute_type === 'taxonomy' &&
+      (formItems.attribute_taxonomy_slug === undefined ||
+        !Array.isArray(formItems.attribute_term_ids) ||
+        formItems.attribute_term_ids.length === 0)) ||
+    (formItems.attribute_type === 'local' &&
+      (!formItems.attribute_local_name ||
+        formItems.attribute_local_name.length === 0 ||
+        !Array.isArray(formItems.attribute_local_values) ||
+        formItems.attribute_local_values.length === 0));
 
   return !purchasedProductWithAttributeIsInvalid;
 }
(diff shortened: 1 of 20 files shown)


mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
```diff
@@ -0,0 +1,137 @@
...
+      return;
+    }
+
+    productAttributeTermsOptionsRef.current = productAttributes[
+      segment.attribute_id
+    ].terms.map((term) => ({
+      value: term.term_id,
+      label: term.name,
+    }));
+  }, [segment.attribute_id, productAttributes]);
```
costasovo: Maybe I'm missing something, but to me, it seems that here we want to "cache" the mapped value of product attributes between renders. Am I correct?
I would say that more natural and probably also more efficient way of doing this could using `useMemo` (https://react.dev/reference/react/useMemo)

Could look like this (not tested):
```js

const productAttributeTermsOptions = useMemo(() => {
  if (segment.attribute_id === undefined) {
     return null;
  }
  return productAttributes[
    segment.attribute_id
  ].terms.map((term) => ({
    value: term.term_id,
    label: term.name,
  })
}, [segment.attribute_id, productAttributes]);

File: mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+
+  useEffect(() => {
+    if (
+      segment.operator !== AnyValueTypes.ALL &&
+      segment.operator !== AnyValueTypes.ANY &&
+      segment.operator !== AnyValueTypes.NONE
+    ) {
+      void updateSegmentFilter({ operator: AnyValueTypes.ANY }, filterIndex);
+    }
+  }, [updateSegmentFilter, segment, filterIndex]);
```
costasovo, line 122: I've recently read a nice post about cases where `useEffect` might not be the best solution. I think this might be one of those cases ([Adjusting some state when props changes](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)). Using the advice from the post could save us a render. But it is not a big deal.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
```diff
@@ -509,6 +510,10 @@ private function createWooCommerce(array $data): DynamicSegmentFilterData {
       $filterData['coupon_code_ids'] = $data['coupon_code_ids'];
       $filterData['days'] = $data['days'];
       $filterData['timeframe'] = $data['timeframe'];
+    } elseif ($data['action'] === WooCommercePurchasedWithAttribute::ACTION) {
+      $filterData['operator'] = $data['operator'];
+      $filterData['attribute_taxonomy_slug'] = $data['attribute_taxonomy_slug'];
+      $filterData['attribute_term_ids'] = $data['attribute_term_ids'];
```
costasovo: Maybe we could add a couple of validations like we have for other filter types ([e.g.](https://github.com/mailpoet/mailpoet/pull/5437/commits/90e35c24325934cf629a34b46df5c8b55211edc2#diff-44dc56dc4354bb109ce8925b08581157b0d1a381f2e399d20204d4ea4fe40455R494-R503)).
This could prevent saving invalid data from the clients. What do you think?

Please note that there is also [a test](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/integration/Segments/DynamicSegments/FilterDataMapperTest.php) for this class so it would be nice to add cases for this filter type.

mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
```diff
@@ -90,23 +137,52 @@ export function PurchasedWithAttributeFields({
...
+            : filter((productAttributeOption) => {
+                if (segment.attribute_taxonomy_slug === undefined) {
+                  return undefined;
+                }
+                return (
+                  segment.attribute_taxonomy_slug ===
+                  productAttributeOption.value
+                );
+              }, productAttributesOptions)
+        }
```
costasovo: Maybe it could be slightly more readable to store the value into a variable before the JSX code. I think it would be easier to read (possibly also cache with useMemo) out of the JSX code, and also, the JSX would probably be more readable.

mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
```diff
@@ -90,23 +137,52 @@ export function PurchasedWithAttributeFields({
...
+              {
+                attribute_type: 'taxonomy',
+                attribute_local_name: null,
+                attribute_local_values: null,
+                attribute_taxonomy_slug: option.value,
+                attribute_term_ids: [],
+              },
+              filterIndex,
+            );
+          }
```
costasovo: Similarly to the previous comment we could move this into a function out of the JSX code.

costasovo (commented):
@johnolek Great job! I added a few comments. All just minor details.

As this code contains a lot of TS code please rebase it on the current trunk. We merged Gutenberg Update PR and since segments use wordpress/components there might be some TS errors after the update.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
johnolek: Good point about the missing tests for the class.  I added some here: https://github.com/mailpoet/mailpoet/tree/e030c017c6cd57be5a502e7c0d8dcfe902239f29

I did have validations before [here](https://github.com/mailpoet/mailpoet/blob/e030c017c6cd57be5a502e7c0d8dcfe902239f29/mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php#L518-L518). I've found it nicer to include the validation logic in the filter class itself instead of in the data mapper. I think it's [more straightforward to test](https://github.com/mailpoet/mailpoet/blob/e030c017c6cd57be5a502e7c0d8dcfe902239f29/mailpoet/tests/integration/Segments/DynamicSegments/Filters/WooCommercePurchasedWithAttributeTest.php#L236-L328) and it's useful to run the validations in the `apply` as well to be sure that we have all the necessary data before building the query. Let me know if you think there's a better way to handle this though.

mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
johnolek: Thanks for this comment. I wasn't sure what the intention was with the original `useRef` myself. I've refactored this to use `useMemo`.

mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
johnolek: Thanks for the link! I'm going to keep this as is for now since it's such a common pattern in this part of the codebase.

mailpoet/assets/js/src/segments/dynamic/dynamic-segments-filters/fields/woocommerce/purchased-with-attribute.tsx
johnolek: I ended up extracting most things and then memoized a lot of stuff to make sure we're avoiding re-renders. Let me know what you think.

johnolek:
Thanks very much for the quick review @costasovo and the great suggestions as always! I appreciate your eye for code quality. I think I've addressed all your points so please feel free to take another look whenever you have time.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
costasovo: Having it together with a filter definition makes sense, and I agree it is a better place than in the mapper. My concern is that we would skip validation on save and allow saving invalid combinations (maybe due to a change after the JS dependencies update, we may start sending wrong data). But I might be missing something. Are the data validated somehow before saving?

costasovo (commented):
@johnolek Thank you for addressing all the suggestions! I find it easier to read after the changes. Do you agree?

I tested it locally and it works as expected!

I noticed an issue worth fixing during segment creation.
1) Created a new segment
2) Selected "purchased with attribute" (nothing else)
3) Got an error message (see screenshot)

Screenshot 2024-02-20 at 11 07 16

It triggered the request for size calculation when the segment was not fully configured. The error in the XHR request was: `{"errors":[{"error":"bad_request","message":"The segment type is unknown."}]}`

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
johnolek: @costasovo The validation is running in the mapper as well, currently [here](https://github.com/mailpoet/mailpoet/blob/a870a2f3edec650b70fe3406e5bdcce6bf380c03/mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php#L518-L518). I don't know of anywhere else we would need to run the validation before saving.

johnolek:
Thanks for doing some local testing and for catching that issue, @costasovo! I pushed a fix in a870a2f3edec650b70fe3406e5bdcce6bf380c03. I think that should be everything.

mailpoet/lib/Segments/DynamicSegments/FilterDataMapper.php
costasovo: Ahh! Perfect! So I missed that. Thank you!

costasovo (dismissed):
@johnolek Great job! 👍

veljkho:
Hi @johnolek 
There is conflicting file and there is no artifact build to download and test this PR. Can you please take a look. Thanks!

johnolek:
@veljkho I rebased and I think this should be good to go now

Aschepikov (approved):
GJ!

PR #5438, opened 2024-02-19
Automation filters with "in the last X days" [PREMIUM-253]

JanJakes:
## Description

Adds "in the last X days" filter to these fields:

**WooCommerce customer**

1. Average spent
2. Order count
3. Total spent

**MailPoet subscriber**

1. Automations — entered
2. Automations — exited
3. Automations — processing
4. Email — clicked count
5. Email — machine opened count
6. Email — opened count
7. Email — sent count

**WooCommerce customer**

1. Purchased categories
2. Purchased tags
3. Review count

## Code review notes

This includes a refactor that adds a more generic support for organizing and rendering filters. I needed this to be able to properly validate the vales to allow/disable the save button.

As for filter parameters, there is a bit of a code repetition now. I didn't want to over-architect it and introduce new generic abstractions, while we have only a single parameter on 4 different filters for now.

## QA notes

I guess there are two ways to test this:

1. Apply filters with this parameter to a trigger, and see if it triggers an automation or not.
2. Use filters with this parameter in an if/else step, and see which branch the automation goes.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/843

## Linked tickets

[PREMIUM-253]

## After-merge notes

_N/A_

## Tasks

[PREMIUM-253]: https://mailpoet.atlassian.net/browse/PREMIUM-253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/filters/enum.ts
+++ b/mailpoet/assets/js/src/automation/editor/filters/enum.ts
@@ -1,5 +1,8 @@
 import { __ } from '@wordpress/i18n';
-import { validateInTheLastParam } from './params/in-the-last';
+import {
+  formatInTheLastParam,
+  validateInTheLastParam,
+} from './params/in-the-last';
 import { FilterType } from '../store/types';
 
 export const filter: FilterType = {
@@ -18,6 +21,7 @@ export const filter: FilterType = {
     const label = options.find(({ id }) => id === args.value)?.name;
     return label ?? __('Unknown value', 'mailpoet');
   },
+  formatParams: ({ args }) => formatInTheLastParam(args),
   validateArgs: (args, _, field) => {
     if (!validateInTheLastParam(args)) {
       return false;
--- a/mailpoet/assets/js/src/automation/editor/filters/enum.ts
+++ b/mailpoet/assets/js/src/automation/editor/filters/enum.ts
@@ -1,0 +1,21 @@
+import { __ } from '@wordpress/i18n';
+import { FilterType } from '../store/types';
+
+export const filter: FilterType = {
+  key: 'enum',
+  fieldType: 'enum',
+  formatValue: ({ args }, field) => {
+    if (args.value === undefined) {
+      return undefined;
+    }
+
+    const options = (field.args.options ?? []) as {
+      id: string;
+      name: string;
+    }[];
+
+    const label = options.find(({ id }) => id === args.value)?.name;
+    return label ?? __('Unknown value', 'mailpoet');
+  },
+  edit: undefined,
+};
--- a/mailpoet/assets/js/src/automation/editor/filters/enum.ts
+++ b/mailpoet/assets/js/src/automation/editor/filters/enum.ts
@@ -17,5 +17,13 @@ export const filter: FilterType = {
     const label = options.find(({ id }) => id === args.value)?.name;
     return label ?? __('Unknown value', 'mailpoet');
   },
+  validateArgs: (args, _, field) => {
+    const value = args.value;
+    const options = (field.args.options ?? []) as {
+      id: string;
+      name: string;
+    }[];
+    return options.some(({ id }) => id === value);
+  },
   edit: undefined,
 };
--- a/mailpoet/assets/js/src/automation/editor/filters/enum.ts
+++ b/mailpoet/assets/js/src/automation/editor/filters/enum.ts
@@ -1,4 +1,5 @@
 import { __ } from '@wordpress/i18n';
+import { validateInTheLastParam } from './params/in-the-last';
 import { FilterType } from '../store/types';
 
 export const filter: FilterType = {
@@ -18,6 +19,10 @@ export const filter: FilterType = {
     return label ?? __('Unknown value', 'mailpoet');
   },
   validateArgs: (args, _, field) => {
+    if (!validateInTheLastParam(args)) {
+      return false;
+    }
+
     const value = args.value;
     const options = (field.args.options ?? []) as {
       id: string;
--- a/mailpoet/assets/js/src/automation/editor/filters/enum.ts
+++ b/mailpoet/assets/js/src/automation/editor/filters/enum.ts
@@ -32,7 +32,10 @@ export const filter: FilterType = {
       id: string;
       name: string;
     }[];
-    return options.some(({ id }) => id === value);
+    return (
+      Array.isArray(value) &&
+      value.every((item) => options.some(({ id }) => id === item))
+    );
   },
   edit: undefined,
 };
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
@@ -69,9 +69,16 @@ class CustomerOrderFieldsFactory {
         'woocommerce:customer:order-count',
         Field::TYPE_INTEGER,
         __('Order count', 'mailpoet'),
-        function (CustomerPayload $payload) {
+        function (CustomerPayload $payload, array $params = []) {
           $customer = $payload->getCustomer();
-          return $customer ? $customer->get_order_count() : 0;
+          if (!$customer) {
+            return 0;
+          }
+
+          $inTheLastSeconds = isset($params['in_the_last_seconds']) ? (int)$params['in_the_last_seconds'] : null;
+          return $inTheLastSeconds === null
+            ? $customer->get_order_count()
+            : $this->getRecentOrderCount($customer, $inTheLastSeconds);
         }
       ),
       new Field(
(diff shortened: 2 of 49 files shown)


mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
```diff
@@ -122,6 +129,40 @@ function (CustomerPayload $payload) {
     ];
   }
 
+  private function getRecentSpentTotal(WC_Customer $customer, int $inTheLastSeconds): float {
+    $wpdb = $this->wordPress->getWpdb();
+    $statuses = array_map(function (string $status) {
+      return "wc-$status";
+    }, $this->wooCommerce->wcGetIsPaidStatuses());
+    $statusesPlaceholder = implode(',', array_fill(0, count($statuses), '%s'));
```
brezocordero, line 174: @JanJakes I am assuming you followed [this]
(https://github.com/woocommerce/woocommerce/blob/be4a36579a10da88657682c5f77134429fed0429/plugins/woocommerce/includes/data-stores/class-wc-customer-data-store.php#L452)
Maybe add a comment that the code is based on that function?
And you are doing the escaping differently but it also works. 
The WC function has a bit more escaping but it also does not seem needed here. I am pretty sure you checked that already. [$customer->get_id()](https://github.com/woocommerce/woocommerce/blob/be4a36579a10da88657682c5f77134429fed0429/plugins/woocommerce/includes/data-stores/class-wc-customer-data-store.php#L479)

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
```diff
@@ -163,6 +170,37 @@ private function getRecentSpentTotal(WC_Customer $customer, int $inTheLastSecond
     return (float)$wpdb->get_var($statement);
   }
 
+  private function getRecentOrderCount(WC_Customer $customer, int $inTheLastSeconds): int {
+    $wpdb = $this->wordPress->getWpdb();
+    $statuses = array_keys($this->wooCommerce->wcGetOrderStatuses());
```
brezocordero, line 205: Same as before, I think it would be useful to add a comment about the original function

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerReviewFieldsFactory.php
```diff
@@ -46,18 +50,27 @@ function (CustomerPayload $payload) {
...
+  private function getUniqueProductReviewCount(WC_Customer $customer, int $inTheLastSeconds = null): int {
     $wpdb = $this->wordPress->getWpdb();
+    $inTheLastFilter = isset($inTheLastSeconds) ? "AND comment_date >= DATE_SUB(current_timestamp, INTERVAL %d SECOND)" : '';
+
     /** @var literal-string $sql */
     $sql = "
       SELECT COUNT(DISTINCT comment_post_ID) FROM {$wpdb->comments}
       WHERE comment_parent = 0
       AND comment_approved = 1
       AND comment_type = 'review'
```
brezocordero: This is not related to your change and more of an edge case but there could be comments with type 'review' that are not for the post_type 'product', either from other plugins or from custom code. I would leave it as it is but what do you think of adding a comment that this could happen?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerReviewFieldsFactory.php
```diff
@@ -46,18 +50,27 @@ function (CustomerPayload $payload) {
    * Calculate the customer's review count excluding multiple reviews on the same product.
    * Inspired by AutomateWoo implementation.
    */
-  private function getUniqueProductReviewCount(WC_Customer $customer): int {
+  private function getUniqueProductReviewCount(WC_Customer $customer, int $inTheLastSeconds = null): int {
     $wpdb = $this->wordPress->getWpdb();
+    $inTheLastFilter = isset($inTheLastSeconds) ? "AND comment_date >= DATE_SUB(current_timestamp, INTERVAL %d SECOND)" : '';
```
brezocordero: `comment_date_gmt`? asking because it would be consistent with the other filters.

mailpoet/assets/js/src/automation/editor/filters/enum.ts
```diff
@@ -17,5 +17,13 @@ export const filter: FilterType = {
     const label = options.find(({ id }) => id === args.value)?.name;
     return label ?? __('Unknown value', 'mailpoet');
   },
+  validateArgs: (args, _, field) => {
+    const value = args.value;
```
brezocordero, line 30: Maybe it is something in my local env but `args.value` can be an array and then the validation fails.
For instance with `woocommerce:order:billing-country` and `is-any-of`

brezocordero (changes requested):
@JanJakes 
Great job and this is a lot of work!
I have small comments and one validation that is not working for me.
Could you please take a look?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
JanJakes: @brezocordero Actually, I used what we already had in the same file ([here](https://github.com/mailpoet/mailpoet/blob/eea5ca0b832e14db56b63ea789a2ab5c5f4729c9/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php#L131) and [here](https://github.com/mailpoet/mailpoet/blob/eea5ca0b832e14db56b63ea789a2ab5c5f4729c9/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php#L172)) and [elsewhere](https://github.com/search?q=repo%3Amailpoet%2Fmailpoet+%22implode%28%27%2C%27%2C+array_fill%280%2C+count%22&type=code), and I don't remember where those came from. When we're using prepared statements with placeholders, there is no need for escaping the parameters whatsoever, is there? Even if the status is named `DROP database mailpoet`, with `$wpdb->prepare` and the `%s` placeholder, it's just a string.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerReviewFieldsFactory.php
JanJakes: @brezocordero That's a great point. Originally, I just used the AutomateWoo approach, but it absolutely makes sense to check for the post type. Added in https://github.com/mailpoet/mailpoet/pull/5438/commits/8722ab95f7c20d2dce951cb3295aade9608fc3d6.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerReviewFieldsFactory.php
JanJakes: Ahh, thank you! This may explain why I had difficulties testing this manually, while the integration test was always passing. Fixed: https://github.com/mailpoet/mailpoet/pull/5438/commits/7113db18b4316b0266041b22e5d19a97ce78b732

mailpoet/assets/js/src/automation/editor/filters/enum.ts
JanJakes: @brezocordero Thanks for yet another great catch! You're right. In fact, the value of the _filter_ is always an array, while the value of the enum field is not, but for enum-array it is, and I got caught in that. Fix: https://github.com/mailpoet/mailpoet/pull/5438/commits/f3f910a78940104369d877cb5396429a4e5d33f2

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
brezocordero: > I used what we already had in the same file

Ok then :-) and my question about the escaping was more about the differences than about the correctness, if the code was not based on that function it makes sense that it is a bit different.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerReviewFieldsFactory.php
brezocordero: > I just used the AutomateWoo approach

Now you make me wonder if there is a reason for it :-D 
Thanks for making the change.

brezocordero (approved):
Thanks for the fixes and the explanations @JanJakes 
Looks good, passing it to QA

Aschepikov (approved):
GJ!

PR #5448, opened 2024-02-22
Email editor - color palette support [MAILPOET-5741]

costasovo:
## Description

This PR adds support for colors selected from the color palette. Prior to this change, values from the palette were ignored by the renderer.
As a bonus, I also added support for the site's theme palette (tested with TT4)

Screenshot 2024-02-22 at 15 45 17

## Code review notes

This doesn't cover the color palette support for border colors. I created an extra ticket for that.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5741]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5741]: https://mailpoet.atlassian.net/browse/MAILPOET-5741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
+++ b/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
@@ -51,17 +51,15 @@ class Column implements BlockRenderer {
       'width' => $width,
       'vertical-align' => $verticalAlign,
     ];
-    $mainCellClasses = 'email_column_wrap';
 
     // The default column alignment is `stretch to fill` which means that we need to set the background color to the main cell
     // to create a feeling of a stretched column
     if (!isset($parsedBlock['attrs']['verticalAlignment']) || $parsedBlock['attrs']['verticalAlignment'] === 'stretch') {
       $mainCellStyles = array_merge($mainCellStyles, $colorStyles);
-      $mainCellClasses .= ' ' . $classes;
     }
 
     return '
-      <td class="block ' . esc_attr($mainCellClasses) . '" style="' . esc_attr($settingsController->convertStylesToString($mainCellStyles)) . '">
+      <td class="block ' . esc_attr($classes) . '" style="' . esc_attr($settingsController->convertStylesToString($mainCellStyles)) . '">
         <div class="email_column" style="width:100%;max-width:' . esc_attr($width) . ';font-size:0px;text-align:left;display:inline-block;">
           <table class="email_column ' . esc_attr($classes) . '" border="0" cellpadding="0" cellspacing="0" role="presentation" style="' . esc_attr($settingsController->convertStylesToString($colorStyles)) . ';min-width:100%;width:100%;max-width:' . esc_attr($width) . ';vertical-align:top;" width="' . esc_attr($width) . '">
             <tbody>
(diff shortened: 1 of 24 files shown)


mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
```diff
@@ -40,19 +50,22 @@ private function getBlockWrapper(array $parsedBlock, SettingsController $setting
       'width' => $width,
       'vertical-align' => $verticalAlign,
     ];
+    $mainCellClasses = 'email_column_wrap';
```
lysyjan: What is the purpose of this class? I haven't found it. If it's still relevant, a comment would be nice here.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Columns.php
```diff
@@ -44,16 +54,17 @@ private function getBlockWrapper(array $parsedBlock, SettingsController $setting
...
           cellpadding="0"
           cellspacing="0"
           role="presentation"
-          style="background:' . $backgroundColor . ';background-color:' . $backgroundColor . ';max-width:' . $width . ';width:100%;"
+          style="' . esc_attr($settingsController->convertStylesToString($colorStyles)) . ';max-width:' . $width . ';width:100%;"
         >
           <tbody>
             <tr>
-              <td style="font-size:0px;background:' . $backgroundColor . ';background-color:' . $backgroundColor . ';padding-left:' . $paddingLeft . ';padding-right:' . $paddingRight . ';padding-bottom:' . $paddingBottom . ';padding-top:' . $paddingTop . ';text-align:left;">
+              <td style="font-size:0px;padding-left:' . $paddingLeft . ';padding-right:' . $paddingRight . ';padding-bottom:' . $paddingBottom . ';padding-top:' . $paddingTop . ';text-align:left;">
```
lysyjan, line 67: You used the `esc_attr` function in the previous class. I a bit wonder why it's not also used here. Maybe we can create a ticket to unify all renderers in this way. What do you think?

lysyjan (dismissed):
Good job! I have only a question about what I don't understand.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
costasovo: Aha. Good catch! It is a leftover after an idea I abandoned. I'll remove it. 👍

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Columns.php
costasovo: I addressed [this particular line in a follow-up PR](https://github.com/mailpoet/mailpoet/pull/5450/commits/caceefe0bac052c791c28dcc7210a8767adc06e8#diff-dcddd58f27adb701f13eac70e8f89fc681684e1d45d39df71e77f8d73ca7af23R66) but there are more unescaped places.
@lysyjan I'm glad you asked. I was thinking about adding a ticket for this so I'll prepare a ticket. IMO we should add proper escaping to all the places where we build the HTML.

costasovo:
@lysyjan Thank you for the detailed review! I added a commit that fixes the unwanted class. As for the escaping, I created an extra ticket and added you as a reviewer.

lysyjan (approved):
@costasovo I'm not sure about QA here. I'm assigning it back to you. Just so you know, you can decide about the next step.

costasovo:
@lysyjan Thanks. There are still a few known issues I would like to fix before we pass rendering to QA so I'm going to merge this.

PR #5454, opened 2024-02-26
Add styles panel to the new email editor [MAILPOET-5638]

lysyjan:
## Description

_N/A_

## Code review notes

- I  needed to refactor the settings sidebar slightly because it was not possible to use the `ComplementaryArea` component properly.
- I tried to prepare the data structure for the styles panel.
- The settings panel contains a few experimental components because I was inspired by the site editor.
- I am opened to discussion about each part of this PR.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5638]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5638]: https://mailpoet.atlassian.net/browse/MAILPOET-5638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/preview.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/preview.tsx
@@ -1,0 +1,218 @@
+import {
+  __unstableMotion as motion,
+  __experimentalHStack as HStack,
+  __experimentalVStack as VStack,
+} from '@wordpress/components';
+
+const firstFrame = {
+  start: {
+    scale: 1,
+    opacity: 1,
+  },
+  hover: {
+    scale: 0,
+    opacity: 0,
+  },
+};
+
+const midFrame = {
+  hover: {
+    opacity: 1,
+  },
+  start: {
+    opacity: 0.5,
+  },
+};
+
+const secondFrame = {
+  hover: {
+    scale: 1,
+    opacity: 1,
+  },
+  start: {
+    scale: 0,
+    opacity: 0,
+  },
+};
+
+const normalizedHeight = 152;
+const normalizedColorSwatchSize = 32;
+
+/**
+ * Component to render the styles preview based on the component from the site editor:
+ * https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-site/src/components/global-styles/preview.js
+ */
+export function Preview(): JSX.Element {
+  const style = {
+    backgroundColor: '#f3f3f3',
+    headingFontFamily: 'Arial',
+    headingColor: '#000000',
+    headingFontWeight: 'normal',
+    paletteColors: [
+      {
+        name: 'Sample Background',
+        slug: 'Sample-background',
+        color: '#f9f8f3',
+      },
+    ],
+    highlightedColors: [
+      {
+        name: 'Sample primary',
+        slug: 'sample-primary',
+        color: '#e5e2d3',
+      },
+      {
+        name: 'Sample Secondary',
+        slug: 'sample-secondary',
+        color: '#111111',
+      },
+    ],
+  };
+
+  const isFocused = false;
+  const isHovered = false;
+  const disableMotion = false;
+  const label = 'Some Label';
+  const ratio = 1;
+
+  const gradientValue = undefined;
+  const withHoverView = true;
+
+  return (
+    <motion.div
+      style={{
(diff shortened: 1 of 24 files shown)


mailpoet/lib/EmailEditor/Engine/SettingsController.php
```diff
@@ -94,26 +94,37 @@ public function getLayout(): array {
...
+        'width' => self::EMAIL_WIDTH,
+        'padding' => [
+          'bottom' => self::FLEX_GAP,
+          'left' => self::FLEX_GAP,
+          'right' => self::FLEX_GAP,
+          'top' => self::FLEX_GAP,
+        ],
+      ],
+      'colors' => [
+        'background' => '#000000',
```
costasovo: What is the difference between layout.background and colors.background?

mailpoet/lib/EmailEditor/Engine/SettingsController.php
```diff
@@ -113,10 +113,18 @@ public function getEmailStyles(): array {
         ],
       ],
       'colors' => [
-        'background' => '#000000',
+        'background' => '#ffffff',
       ],
       'typography' => [
       ],
+      // Value are only for purpose of displaying in the preview component in style sidebar
+      'elements' => [
```
costasovo, line 121: I assume these are temporary and will be replaced by some set values or defaults. Am I correct?

mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/preview.tsx
```diff
@@ -38,181 +41,191 @@ const secondFrame = {
...
+      slug: 'sample-secondary',
+      color: '#e5e2d3',
+    },
+    {
+      name: 'Sample Background',
+      slug: 'Sample-background',
+      color: '#a8c7a9',
+    },
+  ];
+  // https://github.com/WordPress/gutenberg/blob/7fa03fafeb421ab4c3604564211ce6007cc38e84/packages/edit-site/src/components/global-styles/hooks.js#L68-L73
```
costasovo: Where will we get these colors from?

I checked the link and if I understand correctly the primary is text color, the secondary is heading color and the background is document background. Am I right?

Shouldn't we set those in theme.json and apply them here as defaults? Later if a user sets their own, we should merge user-defined data with our theme.json data, and user's would win.

costasovo (commented):
@lysyjan Nice job! I was not digging much into the components I assume we copied some stuff from the site editor. I was rather focused on the values and aded a few questions.

I also noticed that the back button for returning to the main panel differs from the site editor.

Back in our editor:

Screenshot 2024-03-01 at 10 41 56

Back in site editor:
Screenshot 2024-03-01 at 10 45 21

lysyjan:
> I also noticed that the back button for returning to the main panel differs from the site editor.

@costasovo Those buttons are only for testing purposes. I'm expecting a replacement during the implementation of each section.

mailpoet/lib/EmailEditor/Engine/SettingsController.php
lysyjan: The `layout.background` is used in the email editor wrapper, and the `colors.background` is used in the email.
example

I would prefer storing the layout background color in a different place, but I haven't found the right one yet.

mailpoet/lib/EmailEditor/Engine/SettingsController.php
lysyjan: Yes, my expectation here is to merge the email editor default values and values set for an email post.

mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/preview.tsx
lysyjan: @costasovo Based on our call, I added a commit using colors from the settings in the preview.
I hope I addressed all your questions during our call.

costasovo (approved):
@lysyjan Thanks for updating the values from the styles preview component

mailpoet/lib/EmailEditor/Engine/SettingsController.php
costasovo: Ok. We will figure that out as we go.

PR #5456, opened 2024-02-26
Handle shop unavailable when checking the sender domain authentication during sending

brezocordero:
## Description

This PR:
- When retrieving the sender domains from the shop, if the data is not available, we do not overwrite the transient that stores the sender domains with an empty array
- Extends the life of the transient from 1 day to 1 week.
- If during sending, an authenticated domain is required and the shop does not return data, we allow sending. If the domain had already been verified, the sending service will have it stored and the email will be sent. If not, it will be rejected and the campaign will be paused.

## Code review notes

I chose the option to send anyways when the shop is not available. I think that the changes with the transient will cover most of the cases and the sending service has the additional storage of the domains.

## QA notes

Can't be tested on QA but regression testing would be good:
- authenticate a domain
- Import a list with >200 subscribers so that you are a big sender (or reduce the limits in the code)
- Send a campaign (newsletter or post notification)
- Check that sending works as expected

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5921]

## After-merge notes

_N/A_

## Tasks

- [ ] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations -> NA
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes -> NA

[MAILPOET-5921]: https://mailpoet.atlassian.net/browse/MAILPOET-5921?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Services/AuthorizedEmailsController.php
+++ b/mailpoet/lib/Services/AuthorizedEmailsController.php
@@ -170,6 +170,11 @@ class AuthorizedEmailsController {
       ? $this->senderDomainController->getVerifiedSenderDomainsIgnoringCache()
       : $this->senderDomainController->getVerifiedSenderDomains();
 
+    // The shop is not returning data, so we allow sending and let the Sending Service block the campaign if needed.
+    if ($context === 'sending' && empty($verifiedDomains) && !$this->senderDomainController->isCacheAvailable()) {
+      return true;
+    }
+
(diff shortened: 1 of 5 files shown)


alex-mpoet (approved):
Looks good!

Aschepikov (approved):
GJ!

PR #5458, opened 2024-02-27
Add new k6 test automation trash and restore

veljkho:
## Description

Adding new k6 test to cover moving automation to trash and restoring it.

Additionally updated WordPress image and fixed screenshot issue in the helper (missing .png).

## Code review notes

I tested this locally many times to make sure there is no flakiness. It worked well for both local and nightly test execution.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5866]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5866]: https://mailpoet.atlassian.net/browse/MAILPOET-5866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/tests/automation-trash-restore.js
+++ b/mailpoet/tests/performance/tests/automation-trash-restore.js
@@ -44,10 +44,9 @@ export async function automationTrashRestore() {
     await page.$$('.mailpoet-automation-listing-more-button')[0].click(); // click More
     await page.locator('.components-popover__content').click(); // click Trash
     await page.waitForLoadState('networkidle');
-    sleep(1); // this is to avoid false-positive
-    await page.$$('.is-primary')[1].focus();
-    await page.$$('.is-primary')[1].click(); // click Move to trash
-    await waitForSelectorToBeVisible(page, '.notice-success');
+    await page.locator('div.components-flex > button.is-primary').focus();
+    await page.locator('div.components-flex > button.is-primary').click(); // click Move to trash
+    await page.waitForLoadState('networkidle');
 
     // Wait for the success notice message and confirm it
     const movedToTrashNotice =
@@ -71,10 +70,9 @@ export async function automationTrashRestore() {
     await waitForSelectorToBeVisible(page, '.mailpoet-tab-trash.is-active');
     await page.$$('.mailpoet-automation-listing-more-button')[0].click(); // click More
     await page.waitForLoadState('networkidle');
-    sleep(1); // this is to avoid false-positive
     await page.$$('.components-dropdown-menu__menu-item')[0].focus();
     await page.$$('.components-dropdown-menu__menu-item')[0].click(); // click Restore
-    await waitForSelectorToBeVisible(page, '.notice-success');
+    await page.waitForLoadState('networkidle');
 
     // Wait for the success notice message and confirm it
     const restoredFromTrashNotice =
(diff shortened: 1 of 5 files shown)


mailpoet/tests/performance/tests/automation-trash-restore.js
```diff
@@ -0,0 +1,104 @@
...
+
+    await page.waitForLoadState('networkidle');
+    await page.screenshot({
+      path: screenshotPath + 'Automation_Trash_Restore_01.png',
+      fullPage: fullPageSet,
+    });
+
+    // Move to trash one of the existing automation listing
+    await page.$$('.mailpoet-automation-listing-more-button')[0].click(); // click More
+    await page.locator('.components-popover__content').click(); // click Trash
```
costasovo, line 45: Currently, the popover contains only the trash action, but we may add more in the future and that would make this test flaky because half of the content would still work ok. Could we maybe specify somehow we want to click on "Trash"?

mailpoet/tests/performance/tests/automation-trash-restore.js
```diff
@@ -0,0 +1,104 @@
...
+    // Restore from the trash one of trashed automation items
+    await waitForSelectorToBeVisible(page, '.mailpoet-tab-trash');
+    await page.locator('.mailpoet-tab-trash').click(); // click Trash tab
+    await page.waitForLoadState('networkidle');
+    await waitForSelectorToBeVisible(page, '.mailpoet-tab-trash.is-active');
+    await page.$$('.mailpoet-automation-listing-more-button')[0].click(); // click More
+    await page.waitForLoadState('networkidle');
+    sleep(1); // this is to avoid false-positive
+    await page.$$('.components-dropdown-menu__menu-item')[0].focus();
+    await page.$$('.components-dropdown-menu__menu-item')[0].click(); // click Restore
```
costasovo, line 74: Similar to the note about the click on Trash.

costasovo (commented):
Nice job @veljkho. I'm not familiar with the framework, so I can't check for the best practices. But I'm able to read an understand what we want to do. 

I added just a minor comment about a potential issue.

mailpoet/tests/performance/tests/automation-trash-restore.js
veljkho: Hi @costasovo 
The problem is that k6 is not NodeJS nor it is browser and they don't support text-based selectors as of yet.
I tried using XPath instead, as they suggested, but for some reason it fails to perform click saying "The node provided is '#document-fragment', which is not a valid context node type."

Attempted with:
`"//button[@type='button' and contains(., 'Trash')]"` and also other similarities but no luck.

I will see if they start supporting text-based selectors or if we change the UI here to add some specific selectors for Trash button. Please suggest if I should add some specific selector now or we are good to go until we change the UI later.

mailpoet/tests/performance/tests/automation-trash-restore.js
veljkho: Same response here as above.

mailpoet/tests/performance/tests/automation-trash-restore.js
costasovo: @veljkho Ahh. Ok. Makes sense! So in that case we can merge it.

costasovo (dismissed):

costasovo (approved):

costasovo (approved):

PR #5460, opened 2024-02-28
Add Email Editor Empty State [MAILPOET-5912]

mikejolley:
## Description

Adds default content for emails created in the new editor. The content includes a fake logo, heading, image block, paragraph and footer text.

Screenshot 2024-02-28 at 13 35 34

## Code review notes

To test:

1. Go to MailPoet > Emails > New Email
2. In the create dropdown, select the new editor
3. See content populated

## QA notes

Added a test to check content is not empty like before.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5912](https://mailpoet.atlassian.net/browse/MAILPOET-5912)

## After-merge notes

As a follow up we should investigate using patterns and update this template accordingly.

## Tasks

[MAILPOET-5912]: https://mailpoet.atlassian.net/browse/MAILPOET-5912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
```diff
@@ -56,4 +57,27 @@ public function extendEmailPostApi() {
       'schema' => $this->emailApiController->getEmailDataSchema(),
     ]);
   }
+
+  public static function getEmailDefaultContent(): string {
```
mikejolley: Was not sure about this. It needs to be a function because of localisation, but perhaps there is a better place for this method to live?

It needs access to `Env::$assetsUrl` for images.

costasovo: I think it is good 👍 . I see this starter content as a MailPoet-specific feature, so Integration/MailPoet namespace makes perfect sense. 
This is the kind of static stuff that could be a constant, but we need the localization so the static method is fine.

costasovo (commented):
@mikejolley Great stuff!

We just need to address [the failing acceptance test](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/17104/workflows/45f962a1-be3e-4586-9de1-4fecef9775b9/jobs/290706/parallel-runs/7) and it is ready to be merged.

To run the test locally you need to be in the plugin folder (./mailpoet) of the repo and run:

```
./do test:acceptance --skip-deps --file tests/acceptance/EmailEditor/CreateAndSendEmailUsingGutenbergCest.php:displayNewsletterPreview
```
The `--skip-deps` param speeds up the start of the test by skipping reinstalling dependencies. Without it the test installs production only composer packages.

To debug via remote desktop you can (after you start test) open Safari and go to `vnc://localhost:5900` and use the password `secret` when asked.

You can find more info about the tests we use [here](https://github.com/mailpoet/mailpoet/tree/trunk/mailpoet#acceptance-testing).

And a few more debuggin tips here: https://codeception.com/docs/Debugging
(`$i->pause();` is a cool feature).

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
lysyjan: I think we should not use `Env:$assetsUrl` here but use the same approach as it is in [the form templates](https://github.com/mailpoet/mailpoet/blob/926620e8f82ca691f1c11f6e973a2f86a6117522/mailpoet/lib/Form/Templates/FormTemplate.php#L147-L149). Those images are stored in [the SVN plugin repository](https://github.com/mailpoet/mailpoet/tree/926620e8f82ca691f1c11f6e973a2f86a6117522/mailpoet/plugin_repository/assets) and are not a part of the plugin package. So, the plugin build can be a bit smaller.
What do you think @costasovo and @mikejolley?

costasovo: Aha. Good point @lysyjan. Thank you! I missed/forgot that. Yes, we put the images from the form and email templates to the .org cdn. So to be aligned with the rest of the plugin we should probably do the same here.
So, to use it, we need to inject [the service for generating URLs](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Util/CdnAssetUrl.php) and probably change the static method to be an instance method.

mikejolley: When using the CDN I assume assets are pushed there on deploy, so this will 404 unless `MAILPOET_USE_CDN` is used? How do I add an asset there.

costasovo: For development I have `define( 'MAILPOET_USE_CDN', false );` in wp-config.php` but for the acceptance, I can't see it used anywhere, so until we push it we will get 404 for the image.

costasovo: I tried to add it to the acceptance tests and it seems that we might only need to add `wp config set MAILPOET_USE_CDN false --raw` into [this place](https://github.com/mailpoet/mailpoet/blob/eea5ca0b832e14db56b63ea789a2ab5c5f4729c9/mailpoet/tests/docker/codeception/docker-entrypoint.sh#L161-L165). So I guess we could add it. I see it as an improvement to the tests

mikejolley:
@costasovo Feedback has been addressed and tests are passing.

costasovo (approved):
Nice job! Thank you for adding the CND configuration for tests!

veljkho (changes requested):
Hi @mikejolley 
I [don't see logo visible](https://d.pr/i/9pTPqa), the image is not uploaded to https://ps.w.org/mailpoet/assets/email-editor/...

Also, I don't see MailPoet logo on the bottom of the email editor if I use Free Starter plan. I expected to see it when on a free plan, is this implemented?

mikejolley:
@veljkho I assume it won't be uploaded until its released? See comments here on MAILPOST_USE_CDN for development https://github.com/mailpoet/mailpoet/pull/5460#discussion_r1507490283

@costasovo Does the deploy script upload assets to .org for us?

> Also, I don't see MailPoet logo on the bottom of the email editor if I use Free Starter plan. I expected to see it when on a free plan, is this implemented?

This was not in the PR description. It has not been implemented yet.

veljkho:
> This was not in the PR description. It has not been implemented yet.
> I assume it won't be uploaded until its released?

In that case, I am approving the PR. 👍

veljkho (approved):

PR #5465, opened 2024-03-02
Product Capabilities

brezocordero:
## Description

With this PR, we introduce product capabilities in the plugin.
If a product has capabilities, we will use them to check access to certain features:
- mailpoet logo in emails
- number of automation steps in an automation (action steps)
- number of segment filters in a dynamic filter.
- detailed statistics for newsletters, subscribers and automations

If a product does not have capabilities, it should work as before.

What is not included:
- Updating the upgrade/premium modals and the plugin upsells. That will be done in a separate ticket.
- Applying the capabilities to detailed stats that are not newsletters

## Code review notes

@alex-mailpoet please check the premium PR for questions
You can start the review with this PR and finish with the premium one. 
The commits build on top of each other.

Please check that I didn't miss any checks for the features listed.

## QA notes

Please run regression testing and check:
- the logo is added/removed to the email as it should
- detailed statistics are displayed when the subscription gives access to them
- Automations can be edited when it applies
- Segments can have multiple conditions when it applies

It is not possible yet to test the capabilities but locally you can modify the class CapabilitiesManager and change the function getCapabilities to this:
```
  public function getCapabilities(): Capabilities {
    return new Capabilities(
      false, // logo will not be added
      true, // access to detailed stats
      1, // can add one automation step of type action
      1, // can add one condition to a dynamic segment
    );
  }
```
you also need a key that gives access to premium to test the capabilities

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/845

## Linked tickets

[MAILPOET-5898]

## After-merge notes

_N/A_

## Tasks

- [ ] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations -> NA
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes -> NA

[MAILPOET-5892]: https://mailpoet.atlassian.net/browse/MAILPOET-5892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-5889]: https://mailpoet.atlassian.net/browse/MAILPOET-5889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-5898]: https://mailpoet.atlassian.net/browse/MAILPOET-5898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Util/License/Features/CapabilitiesManager.php
+++ b/mailpoet/lib/Util/License/Features/CapabilitiesManager.php
@@ -56,7 +56,7 @@ class CapabilitiesManager {
       return false;
     }
 
-    return (isset($this->tier) && $this->tier < self::MIN_TIER_LOGO_NOT_REQUIRED);
+    return !isset($this->tier) || $this->tier < self::MIN_TIER_LOGO_NOT_REQUIRED;
   }
 
   private function isDetailedAnalyticsEnabled(): bool {
--- a/mailpoet/lib/Util/License/Features/CapabilitiesManager.php
+++ b/mailpoet/lib/Util/License/Features/CapabilitiesManager.php
@@ -1,0 +1,99 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Util\License\Features;
+
+use MailPoet\Settings\SettingsController;
+use MailPoet\Util\License\Features\Data\Capabilities;
+
+class CapabilitiesManager {
+  // Settings mapping
+  const MSS_TIER_SETTING_KEY = 'mta.mailpoet_api_key_state.data.tier';
+  const MSS_MAILPOET_LOGO_IN_EMAILS_SETTING_KEY = 'mta.mailpoet_api_key_state.data.mailpoet_logo_in_emails';
+  const MSS_DETAILED_ANALYTICS_SETTING_KEY = 'mta.mailpoet_api_key_state.data.detailed_analytics';
+  const MSS_AUTOMATION_STEPS_SETTING_KEY = 'mta.mailpoet_api_key_state.data.automation_steps';
+  const MSS_SEGMENT_FILTERS_SETTING_KEY = 'mta.mailpoet_api_key_state.data.segment_filters';
+  // Product capabilities mapping
+  const MIN_TIER_LOGO_NOT_REQUIRED = 1;
+  const MIN_TIER_ANALYTICS_ENABLED = 1;
+  const MIN_TIER_UNLIMITED_AUTOMATION_STEPS = 2;
+  const MIN_TIER_UNLIMITED_SEGMENT_FILTERS = 2;
+
+  private SettingsController $settings;
+  private ?int $tier;
+
+  public function __construct(
+    SettingsController $settings
+  ) {
+    $this->settings = $settings;
+    $this->tier = $this->getTier();
+  }
+
+  private function getTier(): ?int {
+    $tier = $this->settings->get(self::MSS_TIER_SETTING_KEY);
+    return isset($tier) ? (int)$tier : null;
+  }
+
+  private function isMailpoetLogoInEmailsRequired(): bool {
+    $mailpoetLogoInEmails = $this->settings->get(self::MSS_MAILPOET_LOGO_IN_EMAILS_SETTING_KEY);
+
+    if (!isset($this->tier) && !isset($mailpoetLogoInEmails)) {
+      return true;
+    }
+
+    // Allow for less restrictive individual capability to take precedence
+    if (isset($mailpoetLogoInEmails) && (bool)$mailpoetLogoInEmails === false) {
+      return false;
+    }
+
+    return (isset($this->tier) && $this->tier < self::MIN_TIER_LOGO_NOT_REQUIRED) || (isset($mailpoetLogoInEmails) && (bool)$mailpoetLogoInEmails === true);
+  }
+
+  private function isDetailedAnalyticsEnabled(): bool {
+    $detailedAnalytics = $this->settings->get(self::MSS_DETAILED_ANALYTICS_SETTING_KEY);
+
+    if (!isset($this->tier) && !isset($detailedAnalytics)) {
+      return false;
+    }
+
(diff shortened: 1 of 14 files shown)


mailpoet/lib/Util/License/Features/Data/Capabilities.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace MailPoet\Util\License\Features\Data;
+
+class Capabilities {
+  private bool $mailpoetLogoInEmails;
+  private bool $detailedAnalytics;
+  private int $automationSteps;
+  private int $segmentFilters;
+
+  public function __construct(
```
alex-mpoet, line 11: While correct in theory, enumerating capabilities in the constructor can become cumbersome if there will be a lot more capabilities, imagine having a couple dozen arguments. Passing them as an associative array could be an alternative way which is more flexible in terms of order and presence of keys. But for now the current way is fine, can be reworked later if a need arises.

mailpoet/lib/Util/License/Features/CapabilitiesManager.php
```diff
@@ -0,0 +1,118 @@
...
+    if (!$this->isKeyValid) {
+      return true;
+    }
+
+    // Allow for less restrictive individual capability to take precedence over tier
+    if (isset($mailpoetLogoInEmails) && (bool)$mailpoetLogoInEmails === false) {
+      return false;
+    }
+
+    return (isset($this->tier) && $this->tier < self::MIN_TIER_LOGO_NOT_REQUIRED) || (isset($mailpoetLogoInEmails) && (bool)$mailpoetLogoInEmails === true);
```
alex-mpoet: As implemented here, a more restrictive individual capability also takes precedence over tier. For example, if `tier` is `1` (no logo) and `mailpoetLogoInEmails` is `true` (show logo), the logo will be shown. I'm not sure such a combination is designed to happen. But if we consider the implemented logic to be correct, it can be simplified to:

```suggestion
    // Allow for individual capability to take precedence over tier
    if (isset($mailpoetLogoInEmails)) {
      return (bool)$mailpoetLogoInEmails;
    }

    return (isset($this->tier) && $this->tier < self::MIN_TIER_LOGO_NOT_REQUIRED);
```

mailpoet/lib/Util/License/Features/CapabilitiesManager.php
```diff
@@ -0,0 +1,118 @@
...
+    if (!isset($this->tier) && !isset($detailedAnalytics)) {
+      return true; // Backward compatibility is true when preconditions have been met
+    }
+
+    // Allow for less restrictive individual capability to take precedence
+    if (isset($detailedAnalytics) && (bool)$detailedAnalytics === true) {
+      return true;
+    }
+
+    return (isset($this->tier) && $this->tier >= self::MIN_TIER_ANALYTICS_ENABLED) || (isset($detailedAnalytics) && (bool)$detailedAnalytics === true);
```
alex-mpoet: Here the `isset($detailedAnalytics) && (bool)$detailedAnalytics === true` condition is duplicated. Is it intentional?

alex-mpoet (commented):
@brezocordero Good job overall! 

General feedback:
* The capability-checking functions in `CapabilitiesManager` feel a bit complicated. I left specific comments where I saw possible mistakes. I know some things aren't fully ironed out in specs yet, and the need to keep backwards compatibility adds more logic, so we might revisit them later.
* I wanted to make sure we did not forget to cache these capabilities in case the shop is not available. As I see, right now they are updated when a license key is checked, should be adequate for the time being. Let's keep that aspect in mind.

mailpoet/lib/Util/License/Features/CapabilitiesManager.php
brezocordero: > As implemented here, a more restrictive individual capability also takes precedence over tier. 

Thanks for the catch, that was wrong. Fixed.

mailpoet/lib/Util/License/Features/CapabilitiesManager.php
brezocordero: Not intentional. I have removed it.

brezocordero:
@alex-mpoet Thanks for the review.
I have fixed the bug and added tests for it, removed some duplication and done a bit of refactoring.

mailpoet/lib/Util/License/Features/CapabilitiesManager.php
```diff
@@ -56,7 +56,7 @@ private function isMailpoetLogoInEmailsRequired(): bool {
       return false;
     }
 
-    return (isset($this->tier) && $this->tier < self::MIN_TIER_LOGO_NOT_REQUIRED) || (isset($mailpoetLogoInEmails) && (bool)$mailpoetLogoInEmails === true);
+    return (isset($this->tier) && $this->tier < self::MIN_TIER_LOGO_NOT_REQUIRED);
```
alex-mpoet: If `tier` is not set and `mailpoetLogoInEmails` is `true`, this will return false. Maybe a way to fix this would be:

```suggestion
return (isset($this->tier)) ? $this->tier < self::MIN_TIER_LOGO_NOT_REQUIRED : true;
```

brezocordero: Sorry for that @alex-mpoet. It should be good now. 
I guess the difficulty here is that this capability is reversed, it removes a feature when it is true instead of giving it.
Anyways, I have modified the condition and added tests.

alex-mpoet (dismissed):

Aschepikov (approved):
GJ!

PR #5450, opened 2024-02-23
Email editor - padding rendering refactor [MAILPOET-5918]

costasovo:
## Description

In the Gutenberg update, we added a stepped padding control, but the renderer ignores values set in the control. This PR fixes that.

Screenshot 2024-03-01 at 8 48 18

This PR contains a small refactor of how we handle paddings in the renderers. We are moving from custom code to WP native function for generating block styles.

## Code review notes

_N/A_

## QA notes

1. Go to the email editor
2. Add column, paragraph, and button and set padding using stepped padding control and also by manual value
3. Open the email preview and check the padding was applied

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5918]

## After-merge notes

_N/A_

## Tasks

- [x] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] I added sufficient test coverage
- [x] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5918]: https://mailpoet.atlassian.net/browse/MAILPOET-5918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5466, opened 2024-03-04
Email Editor - WP 6.5 compatibility fixes

costasovo:
## Description

This PR fixes issues found when we tested the plugin with WP 6.5 beta 1. Some the issues in the ticket were addressed in different PRs, but I've also found a couple of new issues.

#### Issues addressed in this PR:

1. The appearance of the toolbar menu
2. Sidebar header (tabs)
3. The main area appearance
4. Inserter goes below the browser window
5. Blocks supporting borders allow shadow configuration - I disabled this for now.
6. Fixed Toolbar was not working

#### Issues I was not able to replicate (I was working with beta 2 so possibly some issues were already fixed)

1. Missing Image settings
2. Image missing in Preview
3. The Button's full-width setting is not working (possibly related to 3) The main area appearance)
4. Column block spacing is not working (possibly related to 3) The main area appearance)

#### Issues solved in different PR/ticket

1. Padding cannot be changed (https://github.com/mailpoet/mailpoet/pull/5450)
2. Button rendering is not working (https://github.com/mailpoet/mailpoet/pull/5450)

## Code review notes

I was not able to replicate all the issues (e.g. the problems with image blocks) and also found that some of those in the ticket were not related to WP 6.5 but they were just issues in the rederer (e.g. problems with padding in button were addressed in https://github.com/mailpoet/mailpoet/pull/5450). 

I fixed the issues with BC in mind so the editor should work in WP 6.4 and WP 6.5.

Two issues worth additional comments are:

1. Problem with Tabs in Sidebar Header - The post editor switched to [Tabs component](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/tabs) but this one is in private APIs and not available in WP 6.4. I was able to make it work for WP 6.5 but since it is still private and not supported in WP6.4 I chose to fix the issue by adding CSS for the current implemantation.
2. Fixed Toolbar Not Working - The fix for WP 6.5 caused that in WP 6.4 we rendered two toolbars in the header. I needed to add a condition based on WP/Gutenberg version. I'm not happy with such a condition, but I hope we can get rid of it as soon as we drop support of WP 6.4.

## QA notes

Please test the editor in WP 6.4 and WP 6.5 Beta. There might be some differences because WP6.5 provides newer Gutenberg packages, but we should not introduce any regression compared to WP 6.4

Veljko's note to Alex: We can use site cloning on InstaWP or creating a new site instance to test the difference.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5907]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5907]: https://mailpoet.atlassian.net/browse/MAILPOET-5907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/blocks/core/general-block-support.ts
+++ b/mailpoet/assets/js/src/email-editor/engine/blocks/core/general-block-support.ts
@@ -1,0 +1,33 @@
+import { addFilter } from '@wordpress/hooks';
+import {
+  Block as WPBlock,
+  BlockSupports as WPBlockSupports,
+} from '@wordpress/blocks';
+
+// Extend the BlockSupports type to include shadow
+// The shadow is not included in WP6.4 but it is in WP6.5
+// We can remove it once we upgrade packages to WP6.5
+type BlockSupports = WPBlockSupports & { shadow: boolean };
+type Block = WPBlock & { supports?: BlockSupports };
+
+/**
+ * Disables Shadow Support for all blocks
+ * Currently we are not able to read these styles in renderer
+ */
+function alterSupportConfiguration() {
+  addFilter(
+    'blocks.registerBlockType',
+    'mailpoet-email-editor/block-support',
+    (settings: Block) => {
+      if (settings.supports?.shadow) {
+        return {
+          ...settings,
+          supports: { ...settings.supports, shadow: false },
+        };
+      }
+      return settings;
+    },
+  );
+}
+
+export { alterSupportConfiguration };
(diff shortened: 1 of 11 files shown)


lysyjan (dismissed):
LGTM

Aschepikov:
Hi @costasovo 
With the 4. Inserter goes below the browser window
I see that it is fixed in cases when the border of the window is very close to the block like here
image
But in the cases when the border is not so close it is still going below the window (yep there is a scrollbar but it is hard to select anything)
image

The border in the preview and the emails is always black even when it is not selected, white, or with another color
image

Aschepikov:
Also there are Some conflicts can you resolve them pls @costasovo

costasovo:
@Aschepikov Thanks for the detailed checking. I changed the bottom padding of the canvas to mitigate the issue with the hidden inserter. The issue with the border colors was fixed in a different PR. I updated this one on top of the trunk, so border colors should work as well now.

@lysyjan I rebased the branch on the trunk and added one new commit. Can you please take a look?

lysyjan (approved):

veljkho (approved):
I see some issues still but will be testing against trunk and open a separate Jira ticket to solve them.

PR #5341, opened 2023-12-12
Add full alignment support to blocks in the new email editor [MAILPOET-5688]

lysyjan:
## Description

This PR enables full alignment for specific blocks such as `core/columns`, `core/heading`, `core/paragraph`. As part of the implementation is modifying the renderer and the email editor.

## Code review notes

- Adding `theme.json` and generating styles for the email editor is there for more straightforward implementation.
- Rendering full-width blocks is based on wrapping by the block `core/columns` on the top level because the editor doesn't allow configuration full alignment for the nested blocks.
- Left and right paddings were moved from the `layout.html` to `core/columns` block because it allows stretching blocks to full-width.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5688]

## After-merge notes

_N/A_

## Tasks

- [x] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] I added sufficient test coverage
- [x] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5688]: https://mailpoet.atlassian.net/browse/MAILPOET-5688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5468, opened 2024-03-04
Add background image support to Column/Columns Blocks [MAILPOET-5650]

mikejolley:
## Description

Enhances the `core/column` and `core/columns` block types with support for background images.

As well as adding support for images, this PR also fixes and refactors how styles are generated and applied to columns using `wp_style_engine_get_styles` and `WP_Style_Engine::compile_css`.

I used _Email on Acid_ to check the column layouts worked across clients. Background images do not render in older versions of Outlook. 

Screenshot 2024-03-04 at 16 56 02

## Code review notes

To test you need the new email editor enabled.

1. Go to Emails > New email > Create using beta editor
2. Insert a columns block layout. 
3. Add content to each column.
4. Select the columns block. In the inspector there is a background image section. Upload an image there.
5. Select a column block. In the inspector there is a background image section. Upload an image there.
6. Save and preview the email and check it renders correctly.
7. Select one of the columns and change the alignment from stretch to middle. check the background only covers the block and does not stretch the entire height.
8. Resize preview window to simulate mobile browser. The columns should wrap.

Test coverage was updated as some tests were checking for specific widths; in this PR I made it take the widths from the block itself. I also made it move some of the styles so that they apply to the wrapper with the correct classnames, so some tests were updated to target the correct element.

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/5450

## Linked tickets

[MAILPOET-5650](https://mailpoet.atlassian.net/browse/MAILPOET-5650)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5650]: https://mailpoet.atlassian.net/browse/MAILPOET-5650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/blocks/core/columns.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/blocks/core/columns.tsx
@@ -1,5 +1,6 @@
 import { InspectorControls } from '@wordpress/block-editor';
 import { createHigherOrderComponent } from '@wordpress/compose';
+import { Block } from '@wordpress/blocks';
 import { addFilter } from '@wordpress/hooks';
 
 const columnsEditCallback = createHigherOrderComponent(
@@ -62,4 +63,25 @@ function disableColumnsLayout() {
   );
 }
 
-export { deactivateStackOnMobile, disableColumnsLayout };
+function enhanceColumnsBlock() {
+  addFilter(
+    'blocks.registerBlockType',
+    'mailpoet-email-editor/change-column',
+    (settings: Block, name) => {
+      if (name === 'core/columns') {
+        return {
+          ...settings,
+          supports: {
+            ...settings.supports,
+            background: {
+              backgroundImage: true,
+            },
+          },
+        };
+      }
+      return settings;
+    },
+  );
+}
+
+export { deactivateStackOnMobile, disableColumnsLayout, enhanceColumnsBlock };
--- a/mailpoet/assets/js/src/email-editor/engine/blocks/core/columns.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/blocks/core/columns.tsx
@@ -66,7 +66,7 @@ function disableColumnsLayout() {
 function enhanceColumnsBlock() {
   addFilter(
     'blocks.registerBlockType',
-    'mailpoet-email-editor/change-column',
+    'mailpoet-email-editor/change-columns',
     (settings: Block, name) => {
       if (name === 'core/columns') {
         return {
--- a/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
+++ b/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
@@ -20,55 +20,68 @@ class Column implements BlockRenderer {
     );
   }
 
+  private function getStylesFromBlock(array $block_styles) {
+    $styles = wp_style_engine_get_styles( $block_styles );
+    return $styles['declarations'] ?? [];
+  }
+
   /**
    * Based on MJML <mj-column>
    */
   private function getBlockWrapper(string $blockContent, array $parsedBlock, SettingsController $settingsController): string {
-    // Getting individual border properties
-    $borderStyles = wp_style_engine_get_styles(['border' => $parsedBlock['attrs']['style']['border'] ?? []]);
-    $borderStyles = $borderStyles['declarations'] ?? [];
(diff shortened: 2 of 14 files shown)


mailpoet/assets/js/src/email-editor/engine/blocks/core/columns.tsx
```diff
@@ -62,4 +63,25 @@ function disableColumnsLayout() {
   );
 }
 
-export { deactivateStackOnMobile, disableColumnsLayout };
+function enhanceColumnsBlock() {
+  addFilter(
+    'blocks.registerBlockType',
+    'mailpoet-email-editor/change-column',
```
costasovo: I noticed the namespace here is the same as in the column block. Is that intentional? I'm not sure about the best practices with the namespace argument. After reading [this post](https://make.wordpress.org/core/2022/12/09/update-to-content-only-editing-filter-namespace/), I tend to use unique namespaces.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Column.php
```diff
@@ -20,55 +20,68 @@ public function render(string $blockContent, array $parsedBlock, SettingsControl
...
-    }
+    $block_attributes = isset( $parsedBlock['attrs'] ) && is_array( $parsedBlock['attrs'] ) ? $parsedBlock['attrs'] : [];
+    $block_styles = isset( $block_attributes['style'] ) && is_array( $block_attributes['style'] ) ? $block_attributes['style'] : [];
+    $classes = (new DomDocumentHelper($blockContent))->getAttributeValueByTagName('div', 'class') ?? '';
 
     $width = $parsedBlock['email_attrs']['width'] ?? $settingsController->getLayoutWidthWithoutPadding();
+    $paddingBottom = $block_attributes['style']['spacing']['padding']['bottom'] ?? '0px';
+    $paddingLeft = $block_attributes['style']['spacing']['padding']['left'] ?? '0px';
+    $paddingRight = $block_attributes['style']['spacing']['padding']['right'] ?? '0px';
+    $paddingTop = $block_attributes['style']['spacing']['padding']['top'] ?? '0px';
```
costasovo: I see those added but not used anywhere. I assume that's a temporary change during the development.

costasovo (commented):
@mikejolley Great job! I really like the refactored code and usage of the style engine in the column, columns renderers. I tested the background image and it works as expected.

There are a few issues I noticed when testing it. Probably related to the refactor.

#### 1) Column background color is not applied on full column in the renderer
 Editor:
Screenshot 2024-03-12 at 11 22 26

Renderer:
Screenshot 2024-03-12 at 11 22 35

When I checked on the trunk it renders correctly.

#### 2) Columns block set to render full-width has not full width
Editor:
Screenshot 2024-03-12 at 14 24 38

Renderer:
Screenshot 2024-03-12 at 14 24 49

Here is [the PR that added the functionality](https://github.com/mailpoet/mailpoet/pull/5341) maybe it could give clue.


#### 3) In a test on Email On Acid I see borders around columns in all Outlooks. 
Screenshot 2024-03-12 at 14 03 19

For Outlook we need to say border none. It seems that once it see `solid` it renders some border :( Maybe related to the last commit.

#### 4) Missing warning for users

[The jira ticket](https://mailpoet.atlassian.net/browse/MAILPOET-5650) mentions adding a warning for users "A few email clients do not show background images. Additionally, select a background color for email clients not supporting background images.”
I couldn't find it. We have something similar for borders:

Screenshot 2024-03-12 at 14 32 18

mailpoet/assets/js/src/email-editor/engine/blocks/core/columns.tsx
mikejolley: https://github.com/mailpoet/mailpoet/pull/5468/commits/31ff40379bb2f80e9958f02ee5eb0ee0dd0b3194

mikejolley:
@costasovo Issues should be taken care of. I also fixed widths of full width columns under Outlook.

costasovo (approved):
@mikejolley Great job! There is a failing test, but it is due to an unrelated issue. So I'm going to merge this one.

Note: When testing the rendering I noticed an issue with a column width, but this one is present also on trunk so I created an [extra ticket for that](https://mailpoet.atlassian.net/browse/MAILPOET-5945).

PR #5469, opened 2024-03-04
Track sent campaigns [MAILPOET-5629]

johnolek:
## Description

This PR adds a lot of data to the reporter about campaigns sent in the last 90 days.

## Code review notes

I intentionally left out the "sent to segment" and "filtered by segment" stats for automations because I don't think it's feasible to do it efficiently without making some bigger changes. I know we want to have more stats about automations (e.g. other trigger filter usage), which will run into a similar issue, so I think it's best to leave these stats for that work.

I'm not thrilled about the structure of the code and may update it before requesting review.

My one big question is if it's reasonable to fetch all the data at once or if I should handle it in batches. It seems like a site would need to have a ton of sending queues in a 3 month period to run into performance issues with this query, but it could happen.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Analytics/Reporter.php
+++ b/mailpoet/lib/Analytics/Reporter.php
@@ -440,14 +440,6 @@ class Reporter {
       'Number of automations campaigns sent in the last 30 days' => [],
       'Number of automations campaigns sent in the last 3 months' => [],
 
-      'Number of automations campaigns sent to segment in the last 7 days' => [],
-      'Number of automations campaigns sent to segment in the last 30 days' => [],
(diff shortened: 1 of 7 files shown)


websupporter (approved):
lgtm. I moved all campaign data into its own class and simplified the logic a little bit so that phpstan would not time out and fix the Newsletter in the data factory as other tests where failing when the meta-key was added to the queue creation.

Aschepikov (approved):
GJ!

PR #5478, opened 2024-03-07
Backfill WC customer fields for guests from order [MAILPOET-5378]

JanJakes:
## Description

Some customer fields in automations will now get values from order for `guest` users.

## Code review notes

_N/A_

## QA notes

Using automation filters (trigger or if/else condition) on the following fields should now work even for `guest` orders:

- Billing company
- Billing phone
- Billing city
- Billing postcode
- Billing state
- Billing country
- Shipping company
- Shipping phone
- Shipping city
- Shipping postcode
- Shipping state
- Shipping country
- Total spent
- Average spent
- Order count
- First paid order date
- Last paid order date
- Purchased categories
- Purchased tags

The fields will be backfilled as follows:

- Billing and shipping info will get the same values as filled in the order.
- Total and average spent will equal the order value, in case the order was already paid.
- Order count will be `1`.
- First and last paid order date will equal the order creation date, in case the order was already paid.
- Purchased categories and tags will get values from product in the order, in case the order was already paid.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5378]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5378]: https://mailpoet.atlassian.net/browse/MAILPOET-5378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
@@ -94,7 +94,8 @@ class CustomerOrderFieldsFactory {
         function (CustomerPayload $payload) {
           $customer = $payload->getCustomer();
           if (!$customer) {
-            return null;
+            $order = $payload->getOrder();
+            return $order && $order->is_paid() ? $order->get_date_created() : null;
           }
           return $this->getPaidOrderDate($customer, true);
         }
@@ -106,7 +107,8 @@ class CustomerOrderFieldsFactory {
         function (CustomerPayload $payload) {
           $customer = $payload->getCustomer();
           if (!$customer) {
-            return null;
+            $order = $payload->getOrder();
+            return $order && $order->is_paid() ? $order->get_date_created() : null;
           }
           return $this->getPaidOrderDate($customer, false);
         }
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
@@ -9,6 +9,8 @@ use MailPoet\Automation\Engine\WordPress;
 use MailPoet\Automation\Integrations\WooCommerce\Payloads\CustomerPayload;
 use MailPoet\Automation\Integrations\WooCommerce\WooCommerce;
 use WC_Customer;
+use WC_Order_Item_Product;
+use WC_Product;
 
 class CustomerOrderFieldsFactory {
   /** @var WooCommerce */
@@ -119,11 +121,19 @@ class CustomerOrderFieldsFactory {
         __('Purchased categories', 'mailpoet'),
         function (CustomerPayload $payload, array $params = []) {
           $customer = $payload->getCustomer();
-          if (!$customer) {
-            return [];
-          }
           $inTheLastSeconds = isset($params['in_the_last']) ? (int)$params['in_the_last'] : null;
-          $ids = $this->getOrderProductTermIds($customer, 'product_cat', $inTheLastSeconds);
+          if (!$customer) {
+            $order = $payload->getOrder();
+            $items = $order && $order->is_paid() ? $order->get_items() : [];
+            $ids = [];
+            foreach ($items as $item) {
+              $product = $item instanceof WC_Order_Item_Product ? $item->get_product() : null;
+              $ids = array_merge($ids, $product instanceof WC_Product ? $product->get_category_ids() : []);
+            }
+            $ids = array_unique($ids);
+          } else {
+            $ids = $this->getOrderProductTermIds($customer, 'product_cat', $inTheLastSeconds);
+          }
           $ids = array_merge($ids, $this->termParentsLoader->getParentIds($ids));
           sort($ids);
           return $ids;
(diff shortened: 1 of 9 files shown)


JanJakes:
@johnolek If you won't be able to get to this, let me know, and we can reassign it.

mailpoet/lib/Automation/Integrations/WooCommerce/Payloads/CustomerPayload.php
```diff
@@ -102,6 +102,26 @@ public function getShippingCountry(): ?string {
     return $this->customer->get_shipping_country();
   }
 
+  public function getTotalSpent(): float {
+    if ($this->isGuest()) {
+      return $this->order && $this->order->is_paid() ? (float)$this->order->get_total() : 0.0;
+    }
+    return (float)$this->customer->get_total_spent();
+  }
```
johnolek, line 110: Are the float typecasts necessary? Phpstorm is flagging them as redundant, but since the WC methods don't actually have explicit return types I can see why we might want to have these for safety. But in that case we might want to add casts for other things too, like `int` for `get_order_count` below.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
```diff
@@ -69,7 +69,8 @@ function (CustomerPayload $payload) {
         function (CustomerPayload $payload) {
           $customer = $payload->getCustomer();
           if (!$customer) {
-            return null;
+            $order = $payload->getOrder();
+            return $order && $order->is_paid() ? $order->get_date_created() : null;
```
johnolek, line 115: Is there a reason you prefer to keep this logic in the factory instead of the CustomerPayload? There's a tiny amount of duplication below, and also this code also depends on the idea that a guest customer payload will only have a single order, which I want to double check is a valid assumption.

To be clear I think it's totally fine for this code to live here, more just curious about how you decided where it should live.

johnolek (changes requested):
@JanJakes The code itself looks good to me, though I had a couple of minor questions.

My bigger question about all of this, however, is what seems to be an assumption that a guest customer will only ever have a single order (please let me know if I'm wrong that that's an assumption at all).

I'm having trouble seeing how this would work for a guest customer with multiple orders if I set up an automation where the trigger is something unrelated to an order (e.g. someone subscribes to X) with a filter for first paid order date.

Let me know if I'm misunderstanding anything. It wouldn't be the first time!

JanJakes:
@johnolek 

> My bigger question about all of this, however, is what seems to be an assumption that a guest customer will only ever have a single order (please let me know if I'm wrong that that's an assumption at all).

I wondered this too, but thankfully, [it is the desired behavior](https://a8c.slack.com/archives/C0313447YBC/p1709647370675159). Otherwise, the implementation would be much more complex, since there is no such thing as a guest `WC_Customer` in Woo and we would have to implement it ourselves.

mailpoet/lib/Automation/Integrations/WooCommerce/Payloads/CustomerPayload.php
JanJakes: @johnolek Ahh, I guess I got them copy-pasted or from a Copilot. I'll have a look to unify it one way or another.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
JanJakes: I can't move this to the payload unless I move there all of the `$this->getPaidOrderDate($customer, true);` method implementation, and I'm unsure whether it should go there or not. It would be rather inconsistent with other factories. Also, in the factory I have access to DI, in the payload not.

> the idea that a guest customer payload will only have a single order, which I want to double check is a valid assumption

That will be the case as there is no entity for a guest customer in Woo, so the fact that it's a guest is only derived from the order (the order has no customer attached).

johnolek:
> I wondered this too, but thankfully, [it is the desired behavior](https://a8c.slack.com/archives/C0313447YBC/p1709647370675159). Otherwise, the implementation would be much more complex, since there is no such thing as a guest WC_Customer in Woo and we would have to implement it ourselves.

Thanks for the clarification, @JanJakes! I can see how it would get very complicated having to implement our own guest customer abstraction, especially with potentially different values for the same fields across multiple orders. I suppose if WC core ever adds a guest customer concept then we can revisit this.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/CustomerOrderFieldsFactory.php
johnolek: Makes sense to me! Thanks for the explanation. I think I was a little thrown off by the fact that other logic was moved into the customer payload, but it's all pretty simple stuff, and I can see why you'd want to keep the payload logic very simple without needing access to DI or the database.

johnolek (approved):

JanJakes:
Rebased and added [a commit](https://github.com/mailpoet/mailpoet/pull/5478/commits/650ef4e18e9d107bf8ad32ba1189782d7887f0b6) to resolve conflicts with https://github.com/mailpoet/mailpoet/pull/5438.

Aschepikov (approved):
GJ!

PR #5480, opened 2024-03-10
Keep on-save email validation only in SubscriberSaveController [MAILPOET-5878]

alex-mpoet:
## Description

The email validation in question was added as part of the [refactoring of Subscribers::save() to SubscriberSaveController](https://github.com/mailpoet/mailpoet/pull/3408). I decided to disable it by default and only turn it on in the aforementioned controller for which it was originally introduced. I think this controller deals with all cases of saving emails from user input. Cron workers and other internal services that may work with `SubscriberEntity`  won't trip up on that validation anymore.

To implement this fix, I added the support of the Symfony `Validator` [validation groups](https://symfony.com/doc/current/validation/groups.html) to our `ValidationListener`.

## Code review notes

If you think on-save email validation needs to be kept somewhere else besides `SubscriberSaveController`, please let me know.

## QA notes

Please check that it didn't break email validation in the plugin forms (front-end and admin UI) and smoke-test that other types of form validation are working.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5878]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5878]: https://mailpoet.atlassian.net/browse/MAILPOET-5878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscribers/SubscriberSaveController.php
+++ b/mailpoet/lib/Subscribers/SubscriberSaveController.php
@@ -229,6 +229,9 @@ class SubscriberSaveController {
     // wipe any unconfirmed data at this point
     $subscriber->setUnconfirmedData(null);
 
+    // Validate the email (Saving group) + everything else (Default group)
+    $subscriber->setValidationGroups(['Saving', 'Default']);
+
     try {
       $this->subscribersRepository->persist($subscriber);
       $this->subscribersRepository->flush();
--- a/mailpoet/lib/Doctrine/EntityTraits/ValidationGroupsTrait.php
+++ b/mailpoet/lib/Doctrine/EntityTraits/ValidationGroupsTrait.php
@@ -1,0 +1,18 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Doctrine\EntityTraits;
+
+trait ValidationGroupsTrait {
+  /**
+   * @var array|null
+   */
+  private $validationGroups;
+
+  public function getValidationGroups(): ?array {
+    return $this->validationGroups;
+  }
+
+  public function setValidationGroups(?array $validationGroups): void {
+    $this->validationGroups = $validationGroups;
+  }
+}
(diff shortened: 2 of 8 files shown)


mailpoet/lib/Subscribers/SubscriberSaveController.php
```diff
@@ -229,6 +229,8 @@ public function createOrUpdate(array $data, ?SubscriberEntity $subscriber): Subs
     // wipe any unconfirmed data at this point
     $subscriber->setUnconfirmedData(null);
 
+    $subscriber->setValidationGroups(['Saving', 'Default']);
```
costasovo, line 233: Why do we need to pass both `Saving` and `Default` here? Looking at the integration test I would expect just `Saving` should be enough. If that's needed for some reason it might be a good idea to add a comment.

costasovo (commented):
@alex-mpoet Nice job! TIL about validation groups. I like the idea. I tried to think about the other places we may

> If you think on-save email validation needs to be kept somewhere else besides SubscriberSaveController, please let me know.

I tried to think about the places where we could use the validation, but I found that all of them use the controller (API for third party, WP users or Woo customer sync and the internal API). Batch functions like import/export validate emails explicitly.

Great job! 

I added just one question.

mailpoet/lib/Subscribers/SubscriberSaveController.php
alex-mpoet: If we just pass `Saving`, only the email check will be performed on an entity. But we need to validate everything else (`Default` group) plus the email (`Saving` group). I'll add a comment about it and update the test so the proper usage pattern will be there too.

alex-mpoet: Done.

costasovo: Aha! Thank you for the explanation and for adding the comment. I think it will also be helpful for the other developers reading the code.

costasovo (approved):

veljkho (approved):
Tested submitting forms with captchas, editing subscriber and email field.. form editor.

PR #5486, opened 2024-03-12
Add MailPoet logo to email footer [MAILPOET-5926]

mikejolley:
## Description

Adds the MailPoet logo to the email editor canvas and rendered email. 

The image differs slightly to the design because the background color could change (so we need a fixed background color). This could be revised in the future to include some user-selectable variations (this would require converting this component into a block, or including some styling options somewhere in the editor).

Screenshot 2024-03-12 at 12 42 57

## Code review notes

_N/A_

## QA notes

To test:

1. Go to MailPoet > Emails > New Email > Create using new editor
2. The logo should be shown on the canvase
3. Click the logo. You'll see modal window explaining its removed for premium users.
4. Preview the email. Check the logo is displayed there also.

Screenshot 2024-03-12 at 12 43 11

If using the premium plugin, this logo should be hidden in both the editor, and in the preview/rendered email.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5926](https://mailpoet.atlassian.net/browse/MAILPOET-5926)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5926]: https://mailpoet.atlassian.net/browse/MAILPOET-5926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/EmailEditor/Engine/SettingsController.php
```diff
@@ -30,7 +30,7 @@ class SettingsController {
    * Color of email layout background.
    * @var string
    */
-  const EMAIL_LAYOUT_BACKGROUND = '#cccccc';
+  const EMAIL_LAYOUT_BACKGROUND = '#f0f0f0';
```
mikejolley: This is the color used in the designs.

lysyjan (dismissed):
LGTM 👍

mailpoet/lib/EmailEditor/Engine/SettingsController.php
lysyjan: Thanks for the comment, I suppose we will change it soon with color styles.

Aschepikov:
Hi @mikejolley I think the logo image wasn't pushed yet so I don't see it, can you add it somehow to the pr? In the past, I remember that it could be added additionally before merging
Also, I have a question about where it should be exactly, in the preview I see it under the email. In the editor, it is from the side and it is not very comfortable as it takes place
image

Aschepikov:
@mikejolley  @pavel-mailpoet  Also I wanted to ask why we are not using it the same way as on automations. For example there we have just 2 cases with premium plugin and without. On the automations, we have with/without premium plugin and cases when we should activate or download it. I think it will be better to unify it in all places as we have now in the statistics and automations.
image

mikejolley:
@Aschepikov The logo comes from CDN so is not pushed to .org until the plugin is deployed. You can enable this to use the local image: https://github.com/mailpoet/mailpoet/pull/5460#discussion_r1507490283

In `wp-config.php`:

```
define( 'MAILPOET_USE_CDN', false );
```

> Also, I have a question about where it should be exactly, in the preview I see it under the email. In the editor, it is from the side and it is not very comfortable as it takes plac

Once you get the image loading, ensure you do the build because the styles will not be present otherwise. `./do compile:js`. It should match the screenshot in the PR description.

> Also I wanted to ask why we are not using it the same way as on automations. For example there we have just 2 cases with premium plugin and without. On the automations, we have with/without premium plugin and cases when we should activate or download it. I think it will be better to unify it in all places as we have now in the statistics and automations.

I'm using the `PremiumModal` component that was preexisting. Are you suggesting a different component?

mikejolley:
@lysyjan Just assigning this back due to the changes in supporting the modal. You can see those changes here https://github.com/mailpoet/mailpoet/pull/5486/files#diff-390fe13ff8bc655070bdc6a557e524aba6e5864eaec38867bfc1b63004992127R88-R107

mailpoet/lib/AdminPages/Pages/EmailEditor.php
```diff
@@ -64,6 +85,26 @@ public function render() {
...
+      'mailpoet_has_valid_premium_key' => $this->subscribersFeature->hasValidPremiumKey(),
+      'mailpoet_has_premium_support' => $this->subscribersFeature->hasPremiumSupport(),
+      'mailpoet_plugin_partial_key' => $this->servicesChecker->generatePartialApiKey(),
+      'mailpoet_subscribers_count' => $this->subscribersFeature->getSubscribersCount(),
+      'mailpoet_subscribers_limit' => $this->subscribersFeature->getSubscribersLimit(),
+      'mailpoet_subscribers_limit_reached' => $this->subscribersFeature->check(),
+    ];
+    $this->wp->wpAddInlineScript( 'mailpoet_email_editor', implode( '', array_map( function($key) use ($inline_script_data) {
+      return sprintf( "var %s=%s;", $key, json_encode( $inline_script_data[$key] ) );
+    }, array_keys( $inline_script_data ) ) ), 'before');
```
lysyjan: The implementation looks good. However, we don't use a space after the opening bracket or before the closing bracket. We use a space after the keyword `function'. I tried to prepare a suggestion because I know the code style is different from what you are used to.

```suggestion
    $this->wp->wpAddInlineScript('mailpoet_email_editor', implode('', array_map(function ($key) use ($inline_script_data) {
      return sprintf("var %s=%s;", $key, json_encode($inline_script_data[$key]));
    }, array_keys($inline_script_data))), 'before');
```

lysyjan (commented):
Thanks for the premium modal improvements. I noticed a bit of a different code style and prepared a code change suggestion for you.

mailpoet/lib/AdminPages/Pages/EmailEditor.php
mikejolley: phpcs didn't format this automatically -.-

lysyjan: Unfortunately, it doesn't work now, but Rosta created [a ticket](https://mailpoet.atlassian.net/browse/MAILPOET-5946) to fix that

mikejolley:
Applied the suggestion @lysyjan I'm working on improving PHPCS here: https://github.com/mailpoet/mailpoet/pull/5503

lysyjan (approved):
Thanks

Aschepikov (approved):
GJ!

PR #5488, opened 2024-03-14
Capabilities upsell

brezocordero:
## Description

This PR:
- Updates the modals and banners for specific capabilities
- Ads the capability and subscribers count to the upgrade link

Not on this PR:
- When using the starter plan or an invalid key, the banners display a Purchase link instead of an upgrade link. That link does not have the capabilities info and it has some issues for which I opened a ticket. 
@alex-mpoet I don't know why we don't show an upgrade for valid keys without Premium, maybe some legacy plans? It might be good to check if we should change it.

## Code review notes

I did a bit of refactoring for capabilities, I would have done some more but this was getting too big.
I consolidated some code for the Premium banners [on this commit ](https://github.com/mailpoet/mailpoet/pull/5488/commits/cccbc1e550002c8330c470cf8c98a2eee5b8c6ea) and removed [some code that was never reached](https://github.com/mailpoet/mailpoet/pull/5488/commits/b3d8fbf88db0ba55b5b992073c1163089b1edcd0).
There was [a request to change the link on the Subscribers stats banner](https://github.com/mailpoet/mailpoet-premium/pull/848#issuecomment-1997505426), it was an easy change so I did it on [this commit](https://github.com/mailpoet/mailpoet/pull/5488/commits/e7ca4d22b512fe00890d1dd88a5a88a2eb28e507)

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/851

## Linked tickets

[MAILPOET-5938]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5938]: https://mailpoet.atlassian.net/browse/MAILPOET-5938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brezocordero:
@alex-mailpoet I'm waiting for  https://github.com/mailpoet/mailpoet-premium/pull/848 to add some missing checks on the detailed stats but it is almost ready.

While waiting, would you mind taking a look at this [commit](https://github.com/mailpoet/mailpoet/pull/5488/commits/4ff03a433ce15d465c1ed5632a78890f9b9d6e06)? If we have time, now that I have a better view of the data structure, I would like to refactor it further with the array you suggested.

Instead of having the capabilities detailed in the Capabilities class:
- create a generic class Capability with $name, $type, $value, $isRestricted
- Have an associative array of capabilities as you suggested `array<string, Capability>`
- the associative array could be a property in the CapabilitiesManager with a function to addCapability and the Capabilities Class would disappear
- getCapabilities would return the associative array

Let me know what you think. If we don't have time, I can create a ticket to do it later, if/when we add more capabilities.

alex-mpoet:
@brezocordero I took a look, it's great that you've made the capability checks more flexible and encapsulated.

> Let me know what you think. If we don't have time, I can create a ticket to do it later, if/when we add more capabilities.

If you expect it to not take much time, you can refactor it. Otherwise, creating a ticket for future refactoring would work too.

mailpoet/assets/js/src/subscribers/stats/no-access-info.tsx
```diff
@@ -42,10 +46,25 @@ export function NoAccessInfo(): JSX.Element {
     >
       <thead>
         <tr>
-          <th>{MailPoet.I18n.t('email')}</th>
-          <th>{MailPoet.I18n.t('columnAction')}</th>
-          <th>{MailPoet.I18n.t('columnCount')}</th>
-          <th>{MailPoet.I18n.t('columnActionOn')}</th>
```
alex-mpoet, line 94: When removing placeholder-based translations, could you please remove them from the views too if they're not used anywhere else?

alex-mpoet (commented):
Looks good to me! Thank you for these meticulous changes and refactoring! I have a note about translations removal.

> I don't know why we don't show an upgrade for valid keys without Premium, maybe some legacy plans? It might be good to check if we should change it.

I think we should change it in the same vein as you did [here](https://github.com/mailpoet/mailpoet/pull/5508/commits/0c7251de607fadc9929450a1bcfb4e587c9c18c9).

mailpoet/assets/js/src/subscribers/stats/no-access-info.tsx
brezocordero: Good catch, I didn't think of that.
I was able to remove 2, the others were still in use in other places.

brezocordero:
>  think we should change it

@alex-mpoet  I changed it to show also the upgrade message and upgrade link, if they have a valid key that is not premium.
The modal and the upgrade page do the same, so should be good.

alex-mpoet (approved):

Aschepikov (approved):
GJ!

PR #5490, opened 2024-03-14
Email editor/renderer refactor - split main renderer [MAILPOET-5798]

costasovo:
## Description

In this PR, I split the Renderer class, which was doing a lot of stuff, into two classes: Renderer and Content Renderer.
I wanted to separate the logic for placing rendered posts into email HTML templates and rendering of the email post content.
The motivation for this change is to provide the ability to get rendered output for a single (or multiple) post separately from the email template. This is needed in potential usage of the render outside the plugin but may come in handy also when we render posts for post-notification emails.

**Content renderer** - renders the block's content of the email post and applies all post-processing and inline theme styles.
**Renderer** - uses Content renderer block content and places the content into the email HTML template and generates text version.

## Code review notes

This is just the first PR related to the ticket. I assume a few more to come. 
For now, the Renderer still applies the background color and the global typography settings. This may change in the future as we explore adding top-level blocks for the content.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5798]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5798]: https://mailpoet.atlassian.net/browse/MAILPOET-5798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/tests/integration/EmailEditor/Engine/Renderer/ContentRendererTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace MailPoet\EmailEditor\Engine\Renderer;
+
+use MailPoet\EmailEditor\Engine\EmailEditor;
+
+require_once __DIR__ . '/DummyBlockRenderer.php';
+
+class ContentRendererTest extends \MailPoetTest {
+  /** @var ContentRenderer */
+  private $renderer;
```
lysyjan: You could also use typed properties here.

mailpoet/lib/EmailEditor/Engine/Renderer/content.css
```diff
@@ -0,0 +1,50 @@
+/**
+ CSS reset for email clients for elements used in email content
+ StyleLint is disabled because some rules contain properties that linter marks as unknown (e.g. mso- prefix), but they are valid for email rendering
+ */
+/* stylelint-disable property-no-unknown */
+table,
```
lysyjan: After splitting styles into content and template, we slight duplicity here. But I don't know how we can avoid it.

mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer.php
```diff
@@ -0,0 +1,92 @@
...
+    $this->settingsController = $settingsController;
+    $this->themeController = $themeController;
+  }
+
+  public function render(\WP_Post $post): string {
+    $parser = new \WP_Block_Parser();
+    $parsedBlocks = $parser->parse($post->post_content); // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+
+    $layoutStyles = $this->settingsController->getEmailStyles()['layout'];
+    $parsedBlocks = $this->processManager->preprocess($parsedBlocks, $layoutStyles);
```
lysyjan: Currently, you have `ContentRenderer` and `Renderer` on the same namespace level. I would expect preprocessors and postprocessors to be used in `Renderer` because they are in the same namespace. Maybe we could move them to another level with `ContentRenderer`. Or what are your plans here?

mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer.php
```diff
@@ -0,0 +1,92 @@
...
+   */
+  private function inlineCSSStyles($template) {
+    return $this->cssInliner->inlineCSS($template);
+  }
+
+  /**
+   * @param DomNode $templateDom
+   * @return string
+   */
+  private function postProcessTemplate(DomNode $templateDom) {
```
lysyjan: We could completely remove PHPDocs here and add a return type only.

lysyjan (commented):
Overall, it looks good.
Because I don't know your plans, I had an idea about preprocessors.
I also noticed a few details.

mailpoet/lib/EmailEditor/Engine/Renderer/content.css
costasovo: Yep. I copied styles for tables because template.html contains a couple of table tags as well. Maybe we could inline these manually to the template.html. I would wait how we end up with the exploration of the content block. If we add a content block we will render some markup that is currently in template.html from the blocks and we get rid of the duplicity that way.

mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer.php
costasovo: I don't have detailed plans here.  Moving to the namespace makes sense. I'll do that.

lysyjan (approved):

PR #5491, opened 2024-03-14
Add layout screen with dimension panel [MAILPOET-5640]

lysyjan:
## Description

_N/A_

## Code review notes

This PR is the first of more PRs to add complex functionality to email-specific styles set in the new styles sidebar.
- There is only a new screen component with a dimension panel. Callbacks, actions, and selectors will be part of the next pull requests.

## QA notes

We can definitely skip QA here. Maybe a reviewer can check the UI components after branch checkout.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5640]

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/dimensions-panel.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/dimensions-panel.tsx
@@ -1,6 +1,6 @@
 import {
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
+  // We can remove the ts-expect-error comments once the types are available.
+  // @ts-expect-error TS7016: Could not find a declaration file for module '@wordpress/block-editor'.
   __experimentalSpacingSizesControl as SpacingSizesControl,
   useSetting,
 } from '@wordpress/block-editor';
@@ -12,9 +12,7 @@ import {
 import { __ } from '@wordpress/i18n';
 
 export function DimensionsPanel() {
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
-  const availableUnits: string[] = useSetting('spacing.units');
+  const availableUnits = useSetting('spacing.units') as string[];
   const units = useCustomUnits({
     availableUnits,
   });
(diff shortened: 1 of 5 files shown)


mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/screen-header.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <NavigatorToParentButton
+              style={{ minWidth: 24, padding: 0 }}
+              icon={chevronLeft}
+              size="small"
+              aria-label={__('Navigate to the previous view')}
+              onClick={onBack}
+            />
+            <Spacer>
+              {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
+              {/* @ts-ignore */}
```
costasovo: I've recently learned about `@ts-expect-error`. Here is [an example usage](https://github.com/mailpoet/mailpoet/commit/ca2ddfbbf99605a9e5339caa52a5c59d2e1adbaf).

The difference is that in case the TS error disappears (e.g. after updating packages) we get notified by TS type checker and we remove the comment. SO I would suggest using it instead of `ts-ignore`

mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/dimensions-panel.tsx
```diff
@@ -0,0 +1,75 @@
...
+        hasValue={() => true}
+        label={__('Padding')}
+        onDeselect={() => resetPadding()}
+        className="tools-panel-item-spacing"
+      >
+        <SpacingSizesControl
+          allowReset
+          values={paddingValues}
+          onChange={setPaddingValues}
+          label={__('Padding', 'mailpoet')}
```
costasovo: I think we should provide a comment for translators here. OR, I see a couple of lines above where `__('Padding')` is used without MailPoet's namespace. Hmm 'Padding' is commonly used in the block editor and we load the module from the WP so we should be able also to load it's translations (maybe we already do I'm not 100% sure) I guess that for `Padding` and also for `Block spacing` we could skip adding `mailpoet` namespace.

costasovo (commented):
@lysyjan Thank you for splitting the PR. Reviewing smaller chunks makes it easier for me   and hopefully I'm able to review the changes more thoroughly.

I added a couple of suggestions. Just minor details.

mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/dimensions-panel.tsx
lysyjan: Using those translations without a namespace makes sense to me here. I will update it.

mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/screen-header.tsx
lysyjan: Nice. Thanks for this suggestion. I will change it.

costasovo (approved):
Thanks for the fixes!

PR #5484, opened 2024-03-12
Order explicitly by name of taxonomy [MAILPOET-5942]

websupporter:
## Description
Improves the performance of the underlying taxonomy query.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5942]

## After-merge notes

_N/A_

## Tasks

- [ ] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [ ] I added sufficient test coverage
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5942]: https://mailpoet.atlassian.net/browse/MAILPOET-5942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5492, opened 2024-03-14
Improve term loading

websupporter:
## Description
This is a follow up for https://github.com/mailpoet/mailpoet/pull/5484 and further improves the loading of terms.

## Code review notes
This PR improves the performance of the term loading. Two major improvements are suggested
1. When we loaded the terms of a taxonomy already, we do not need to load it again
2. We can improve the list building with a.) a lookup table, which limits the arrays we need to walk through in the recursive calls and b.) by not attempting to build a hierarchical list for non-hierarchical taxonomies.

## QA notes
This PR is concerned with the automation filters, more particularly the filters with which you can limit the triggers to only trigger for certain product categories or product tags.
image

The goal is to improve performance when you have a lot of categories or tags. It is important that we do not add any regressions.

1. Can I select all product tags or are some missing?
2. Can I select all product categories or are some missing?
3. The product categories are organized hierarchically. A child category should be underneath its parent and its name should look like "Parentname | Childname"

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5942]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5942]: https://mailpoet.atlassian.net/browse/MAILPOET-5942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
@@ -24,17 +24,32 @@ class TermOptionsBuilder {
     if (isset($this->terms[$taxonomy])) {
       return $this->terms[$taxonomy];
     }
+
+    $taxonomyObject = $this->wordPress->getTaxonomy($taxonomy);
+    if (!$taxonomyObject) {
+      $this->terms[$taxonomy] = [];
+      return $this->terms[$taxonomy];
+    }
+
     $terms = $this->wordPress->getTerms(['taxonomy' => $taxonomy, 'hide_empty' => false, 'orderby' => 'name']);
     if ($terms instanceof WP_Error) {
       $this->terms[$taxonomy] = [];
       return $this->terms[$taxonomy];
     }
-    $this->terms[$taxonomy] = $this->buildTermsList((array)$terms);
+    $this->terms[$taxonomy] = $this->buildTermsList((array)$terms, $taxonomyObject);
     return $this->terms[$taxonomy];
   }
 
   /** @return array<array{id: int, name: string}> */
-  private function buildTermsList(array $terms, int $parentId = 0, array $lookup = []): array {
+  private function buildTermsList(array $terms, \WP_Taxonomy $taxonomy, int $parentId = 0, array $lookup = []): array {
+    if (!$taxonomy->hierarchical) {
+      return array_map(
+        function ($term) {
+          return ['id' => (int)$term->term_id, 'name' => $term->name]; //phpcs:ignore: Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+        },
+        $terms
+      );
+    }
     if ($lookup === []) {
       foreach ($terms as $term) {
         if (!$term instanceof WP_Term) {
@@ -57,7 +72,7 @@ class TermOptionsBuilder {
       if (empty($lookup[$id])) {
         continue;
       }
-      foreach ($this->buildTermsList($lookup[$id], $id, $lookup) as $child) {
+      foreach ($this->buildTermsList($lookup[$id], $taxonomy, $id, $lookup) as $child) {
         $list[] = ['id' => (int)$child['id'], 'name' => "$term->name | {$child['name']}"];
       }
     }
--- a/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
+++ b/mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
@@ -10,6 +10,9 @@ class TermOptionsBuilder {
   /** @var WordPress */
   private $wordPress;
 
+  /** @var array<string, array<array{id: int, name: string}>> */
+  private $terms = [];
+
   public function __construct(
     WordPress $wordPress
   ) {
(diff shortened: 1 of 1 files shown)


mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
```diff
@@ -18,28 +21,58 @@ public function __construct(
 
   /** @return array<array{id: int, name: string}> */
   public function getTermOptions(string $taxonomy): array {
+    if (isset($this->terms[$taxonomy])) {
```
JanJakes, line 24: Since this is a cache, let's call it that way so that it's obvious? Like `$this->cache`, for instance.

I think it's also more idiomatic and more robust with respect to future changes to do:

```
if (!isset($this->cache[$taxonomy])) {
  // do everything like before & save to cache
}
$this->cache[$taxonomy];
```

Now there seem to be multiple different places for handling the cache.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
```diff
@@ -18,28 +21,58 @@ public function __construct(
...
+      return $this->terms[$taxonomy];
+    }
+
+    $taxonomyObject = $this->wordPress->getTaxonomy($taxonomy);
+    if (!$taxonomyObject) {
+      $this->terms[$taxonomy] = [];
+      return $this->terms[$taxonomy];
+    }
```
JanJakes, line 32: Does this optimize something? Isn't it just a DB query, same as with `$this->wordPress->getTerms`?

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
```diff
@@ -18,28 +21,58 @@ public function __construct(
...
-    });
+  private function buildTermsList(array $terms, \WP_Taxonomy $taxonomy, int $parentId = 0, array $lookup = []): array {
+    if (!$taxonomy->hierarchical) {
+      return array_map(
+        function ($term) {
+          return ['id' => (int)$term->term_id, 'name' => $term->name]; //phpcs:ignore: Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+        },
+        $terms
+      );
+    }
```
JanJakes, line 52: This duplicates the code in the loop below + calling the function on each item is slower than the `foreach` loop. I think it would be nicer to just avoid calling the nested foreach if the taxonomy is not hierarchical.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
```diff
@@ -18,28 +21,58 @@ public function __construct(
...
     $list = [];
-    foreach ($parents as $term) {
+    foreach ($lookup[$parentId] as $term) {
+      if (!$term instanceof WP_Term || $term->parent !== $parentId) {
+        continue;
+      }
       $id = $term->term_id; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
       $list[] = ['id' => (int)$id, 'name' => $term->name];
-      foreach ($this->buildTermsList($terms, $id) as $child) {
+      if (empty($lookup[$id])) {
```
JanJakes, line 72: I think `empty()` is not pretty slow and wasn't recommended to be used within A8C, `isset` is probably better.

mailpoet/lib/Automation/Integrations/WooCommerce/Fields/TermOptionsBuilder.php
```diff
@@ -18,28 +21,58 @@ public function __construct(
...
+      return array_map(
+        function ($term) {
+          return ['id' => (int)$term->term_id, 'name' => $term->name]; //phpcs:ignore: Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+        },
+        $terms
+      );
+    }
+    if ($lookup === []) {
+      foreach ($terms as $term) {
+        if (!$term instanceof WP_Term) {
```
JanJakes, line 55: These `instanceof` checks are actually only to satisfy PHPStan, maybe we can just annotate the `$terms` array instead.

JanJakes (commented):
@websupporter Thanks for these optimizations! After our discussions, I went thought them, and I understand how are beneficial. When going through the code, I started to add comments with ideas to simplify the code, but in the end, I ended up outlining it (didn't test it), and will provide it all:

```php
<?php declare(strict_types = 1);

namespace MailPoet\Automation\Integrations\WooCommerce\Fields;

use MailPoet\Automation\Engine\WordPress;
use WP_Error;
use WP_Term;

class TermOptionsBuilder {
  /** @var WordPress */
  private $wordPress;

  /** @var array<string, array<array{id: int, name: string}>> */
  private $cache = [];

  public function __construct(
    WordPress $wordPress
  ) {
    $this->wordPress = $wordPress;
  }

  /** @return array<array{id: int, name: string}> */
  public function getTermOptions(string $taxonomy): array {
    if (!isset($this->cache[$taxonomy])) {
      $this->cache[$taxonomy] = $this->fetchTermOptions($taxonomy);
    }
    return $this->cache[$taxonomy];
  }

  /** @return array<array{id: int, name: string}> */
  private function fetchTermOptions(string $taxonomy): array {
    /** @var WP_Term[]|WP_Error $terms */
    $terms = $this->wordPress->getTerms(['taxonomy' => $taxonomy, 'hide_empty' => false, 'orderby' => 'name']);
    if ($terms instanceof WP_Error) {
      return [];
    }

    $termsMap = [];
    foreach ($terms as $term) {
      $termsMap[$term->parent][] = $term;
    }
    return $this->buildTermsList($termsMap);
  }

  /**
   * @param array<int, array<WP_Term>> $termsMap
   * @return array<array{id: int, name: string}>
   */
  private function buildTermsList(array $termsMap, int $parentId = 0): array {
    $list = [];
    foreach ($termsMap[$parentId] ?? [] as $term) {
      $id = $term->term_id; // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
      $list[] = ['id' => $id, 'name' => $term->name];
      if (isset($termsMap[$id])) {
        foreach ($this->buildTermsList($termsMap, $id) as $child) {
          $list[] = ['id' => $child['id'], 'name' => "$term->name | {$child['name']}"];
        }
      }
    }
    return $list;
  }
}
```

I think this version is easier to read, and it doesn't need extra checks like whether a taxonomy is hiearchical, because that naturally emerges from the lookup table (that I called `$termsMap`. If a taxonomy is not hierarchical, then `isset($termsMap[$id])` is always `false` on the first level, which hopefully is a very quick check.

websupporter:
closed in favor of https://github.com/mailpoet/mailpoet/pull/5495

PR #5493, opened 2024-03-14
Validate content before send in the new email editor [MAILPOET-5913]

mikejolley:
## Description

Implements validation to ensure that the email content contains certain elements, in this case the unsubscribe link.

If when the send button is pressed some required content is missing, sending is blocked with a notice:

Screenshot 2024-03-14 at 13 59 34

After correcting the error, the send button is re-enabled. You can correct manually, or click the "insert" button in the notice to resolve the issue.

## Code review notes

1. There is a short debounce to prevent excessive updates via the `subscribe` method
2. Validation messages have `email-validation` context. Other errors now use `email-editor` context. This is so they can be displayed separately.
3. The notice created by `SentEmailNotice` was lacking an ID so I fixed this to prevent duplicates.
4. There was a second task to change the text when an email was not saved; this error already existed so I just updated the text to read "The email could not be saved. Please, clear browser cache and reload the page. If the problem persists, duplicate the email and try again.".
5. `getEditedEmailContent` was borrowed from Gutenberg because that only handled data from the post editor. We needed a function to get data from our stores.

## QA notes

1. Ensure you build before testing from this branch `./do compile:js` otherwise styles etc may be missing.
2. Emails > New Email > Create using new editor
3. The default content includes an "unsubscribe" link. Remove this block.
4. Press "send"
5. The error notice will appear and "send" is disabled.
6. Use the "insert link" button in the notice. The "unsubscribe" block should be added to the content and the notice will be hidden. Send is active again.
7. Hit send and it should redirect to the send page.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5913](https://mailpoet.atlassian.net/browse/MAILPOET-5913)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5913]: https://mailpoet.atlassian.net/browse/MAILPOET-5913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/package.json
```diff
@@ -146,10 +146,12 @@
     "@types/wordpress__edit-post": "^7.5.6",
     "@types/wordpress__editor": "^13.6.7",
     "@types/wordpress__media-utils": "^4.14.4",
+    "@types/wordpress__notices": "^3.27.6",
```
costasovo: Do we need to add the package? I see that @wordpress/notices contain build-types. AFAIK if a package provides own build types the type definitions from @types and basically ignored. I also see you added a type for Notice. I tried removing the package locally and QA checks  for js `./do qa:lint-javascript` are still passing. Am I missing something or can we remove the package?

mailpoet/assets/js/src/email-editor/engine/components/header/send-button.tsx
```diff
@@ -3,21 +3,19 @@ import { Button } from '@wordpress/components';
...
+
+  const { validateContent, isValid } = useContentValidation();
+  const { hasEmptyContent, isEmailSent } = useSelect(
     (select) => ({
       hasEmptyContent: select(storeName).hasEmptyContent(),
-    }),
-    [],
-  );
-  const { isEmailSent } = useSelect(
-    (select) => ({
```
costasovo, line 20: Thanks for merging the useSelect calls into one!

costasovo (commented):
@mikejolley Great job! I really like how you designed the code so that we can easily add new validation rules and import and call the validation from different places. That's awesome!

I added one question about types/__notices package. Please see my other comment.

mailpoet/package.json
mikejolley: I added this trying to get the `Notice` type working but in the end I had to duplicate it. I can remove it, it wasn't useful.

costasovo (approved):
Thanks for removing the package!

Aschepikov (approved):
GJ!

PR #5497, opened 2024-03-15
Use the same format for email styles as it is in theme.json

lysyjan:
## Description

_N/A_

## Code review notes

This PR is the second in a row related to the settings styles panel.
- The goal here is to prepare the style data structure for merging and allowing easier manipulation on the front end.

## QA notes

We can skip QA here and ask for testing in the last PR.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5640]

## After-merge notes

_N/A_

## Tasks


mailpoet/assets/js/src/email-editor/engine/store/types.ts
```diff
@@ -27,24 +27,31 @@ export type ExperimentalSettings = {
...
-      color: string;
-      fontFamily: string;
-      fontWeight: string;
+      color: {
+        text: string;
+      };
+      typography: {
+        fontFamily: string;
+        fontWeight: string;
+      };
```
costasovo, line 54: The change was added earlier, but I struggle to understand why we define only h1 in this type definition. Shouldn't we have some common type for element styles?

costasovo: Oh I see now h1 is used in `email-editor/engine/components/styles-sidebar/preview.tsx`.
Ok. I guess we can keep it for now. We are aiming to align with theme.json, so I'm hoping in the future, we will be able to grab some types from Gutenberg.

costasovo (dismissed):
@lysyjan Thanks for unifying the format. It looks like it was a tedious job to find all the places. Great job!

mailpoet/assets/js/src/email-editor/engine/store/types.ts
lysyjan: @costasovo I think we will need to add more definitions when we implement the typography panel. So, our theme.json will be more robust in the future.

costasovo (approved):

PR #5499, opened 2024-03-18
Add saving email specific styles [MAILPOET-5640]

lysyjan:
## Description

_N/A_

## Code review notes

This is the final pull request related to the layout styles.
- It adds saving post metadata such as styles.
- It activates block-editor private API.
- It creates a new component that renders styles and applies changes via the styles panel.

## QA notes

_N/A_

## Linked PRs

- https://github.com/mailpoet/mailpoet/pull/5491
- https://github.com/mailpoet/mailpoet/pull/5497

## Linked tickets

[MAILPOET-5640]

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/components/block-editor/block-editor.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/block-editor/block-editor.tsx
@@ -14,6 +14,9 @@ import {
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore No types for this exist yet.
   __unstableEditorStyles as EditorStyles,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore No types for this exist yet.
+  BlockContextProvider,
 } from '@wordpress/block-editor';
 import { UnsavedChangesWarning } from '@wordpress/editor';
 import { uploadMedia } from '@wordpress/media-utils';
@@ -33,7 +36,7 @@ import { Header } from '../header';
 import { ListviewSidebar } from '../listview-sidebar/listview-sidebar';
 import { InserterSidebar } from '../inserter-sidebar/inserter-sidebar';
 import { EditorNotices, EditorSnackbars, SentEmailNotice } from '../notices';
-import { StylesSidebar } from '../styles-sidebar';
+import { StylesSidebar, ThemeStyles } from '../styles-sidebar';
 import { FooterCredit } from './footer-credit';
 
 export function BlockEditor() {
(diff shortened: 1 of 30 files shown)


mailpoet/assets/js/src/email-editor/engine/components/block-editor/block-editor.tsx
```diff
@@ -14,6 +14,9 @@ import {
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore No types for this exist yet.
   __unstableEditorStyles as EditorStyles,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore No types for this exist yet.
```
costasovo: Could we use ts-expect-error here?

mailpoet/assets/js/src/email-editor/engine/components/block-editor/block-editor.tsx
```diff
@@ -122,82 +125,92 @@ export function BlockEditor() {
...
-      onInput={onInput}
-      onChange={onChange}
-      settings={settings}
-      useSubRegistry={false}
+    // BlockContextProvider has set templateSlug because the hook `useGlobalStylesOutputWithConfig` doesn't render layout padding without it.
+    <BlockContextProvider
+      value={{
+        postId,
+        postType: 'mailpoet_email',
+        templateSlug: 'mailpoet-email',
```
costasovo: Just a detail. It looks a bit weird to see `mailpoet_email` and `mailpoet-email`. Could we maybe use `mailpoet_email_template`? If I understand correctly it doesn't matter what we use, right?

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
```diff
@@ -51,6 +55,17 @@ public function addEmailPostType(array $postTypes): array {
...
         'rewrite' => ['slug' => self::MAILPOET_EMAIL_POST_TYPE],
       ],
+      'meta' => [
+        [
+          'key' => self::MAILPOET_EMAIL_META_THEME_TYPE,
+          'args' => [
+            'show_in_rest' => true,
+            'single' => true,
+            'type' => 'object',
+            'default' => ['version' => 2],
```
costasovo: I wonder why we use `['version' => 2]`?

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
```diff
@@ -85,4 +100,12 @@ public function getEmailDefaultContent(): string {
...
     ';
   }
+
+  public function extendEmailThemeStyles(WP_Theme_JSON $theme, WP_Post $post): WP_Theme_JSON {
+    $emailTheme = get_post_meta($post->ID, self::MAILPOET_EMAIL_META_THEME_TYPE, true);
+    if ($emailTheme && is_array($emailTheme)) {
+      $theme->merge(new WP_Theme_JSON($emailTheme));
+    }
+    return $theme;
+  }
```
costasovo: I think that saving/fetching and the definition of the meta key for the styles settings should be part of the engine rather than the MailPoet integration. I mean the editor needs them so we shouldn't rely on an integration to provide those.

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
```diff
@@ -49,6 +50,8 @@ public function saveEmailData(array $data, \WP_Post $emailPost): void {
       throw new UnexpectedValueException('Newsletter ID does not match the post ID');
     }
 
+    update_post_meta($emailPost->ID, EmailEditor::MAILPOET_EMAIL_META_THEME_TYPE, $data['theme']);
+
```
costasovo: I think we are mixing two concepts here.

I like the idea of saving the theme in post-meta.  We are using [register_rest_field](https://developer.wordpress.org/reference/functions/register_rest_field/) for saving/fetching some data we store in custom tables.

But for post meta, I would expect the WP Rest API to have built-in handling so we don't have to pass them around via our extra field, but we should be able to fetch them and update them directly via core/store.
I googled a bit and found this post https://jeffreycarandang.com/how-to-enable-read-and-update-post-meta-in-the-wordpress-block-editor-gutenberg/

So could we use the built-in API for fetching/saving post meta fields. That way we can move it to the engine more easily.

costasovo (commented):
@lysyjan Good job!

I added a couple of comments with questions and suggestions. I hope they make sense.

mailpoet/assets/js/src/email-editor/engine/components/block-editor/block-editor.tsx
lysyjan: 👍 I tried to address this across the whole email editor: https://github.com/mailpoet/mailpoet/pull/5499/commits/ba5d61f6325193328c907070b2c138f4719bee89

mailpoet/assets/js/src/email-editor/engine/components/block-editor/block-editor.tsx
lysyjan: 👍 Yes, it doesn't matter about the name. Changed here: https://github.com/mailpoet/mailpoet/pull/5499/commits/c7ad438e5e703a286a60154eb0d081764459a242

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
lysyjan: If we don't use the same version as it is in our theme.json files, the merge doesn't work properly. I added a comment.

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
lysyjan: Good point. Moved to the engine: https://github.com/mailpoet/mailpoet/pull/5499/commits/642f5f5cf4361db2529620b0cea38327fa6704d1

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
lysyjan: @costasovo Great point here! I refactored this a bit, and the final code looks better. Please check all new commits.

costasovo (dismissed):
Thank you for fixing the usage of metadata! 👍 LGTM!

pavel-mailpoet (approved):

PR #5507, opened 2024-03-18
Use the value of lastSubscribedAt instead of createdAt in subscribers listing [MAILPOET-5894]

lysyjan:
## Description

Because the plugin uses the column `lastSubscribedAt` in the segment conditions, it was confusing to display values from the column `createdAt` as Subscribed on, so a user reported it as an issue. This PR replaces the rendering `createdAt` by `lastSubscribedAt`.
## Code review notes

_N/A_

## QA notes

1. Visit the subscribers' admin page
2. Check that the values in the column "Subscribed on" are equal to the values in the lastSubscribedAt in the DB
3. Try ordering by the column "Subscribed on"

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5894]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5894]: https://mailpoet.atlassian.net/browse/MAILPOET-5894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/api_methods/GetSubscriber.md
+++ b/doc/api_methods/GetSubscriber.md
@@ -31,6 +31,7 @@ This method throws an `\Exception` in the event a subscriber with a given email 
 | created_at               | string\|null | -         | UTC time of creation in 'Y-m-d H:i:s' format                                                                                   |
 | updated_at               | string       | -         | UTC time of last update in 'Y-m-d H:i:s' format                                                                                |
 | deleted_at               | string\|null | -         | This property in not null in case that list is in trash and contains UTC time in 'Y-m-d H:i:s' format.                         |
+| last_subscribed_at       | string\|null | -         | UTC time of last confirmed subscription in 'Y-m-d H:i:s' format.                                                               |
 | unconfirmed_data         | string\|null | 65K chars | May contain serialized subscriber data in case when there are pending changes waiting for a confirmation from a subscriber     |
 | source                   | string\|null | -         | Possible values: `form`,`imported`,`administrator`,`api`,`wordpress_user`,`woocommerce_user`,`woocommerce_checkout`,`unknown`) |
 | count_confirmations      | string       | 11 chars  | Counter for confirmation emails                                                                                                |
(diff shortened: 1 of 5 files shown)


alex-mpoet (commented):
@lysyjan Looks good! One question: do we want to include this field into the MP API [response](https://github.com/mailpoet/mailpoet/blob/trunk/doc/api_methods/GetSubscriber.md)?

lysyjan:
@alex-mpoet I'm not aware of any reason why the value `lastSubscribedAt` should not be part of the MP API response. Do you see an issue here?

alex-mpoet:
@lysyjan 

> I'm not aware of any reason why the value lastSubscribedAt should not be part of the MP API response. Do you see an issue here?

It's not in the MP API response, that's the issue. Could you please add it there and update the API docs?

lysyjan:
@alex-mpoet The property `lastSubscribedAt` is already a part of the MP API because the subscriber endpoints use the `SubscribersResponseBuilder::build()` method that returns more details. I added this property to the docs in the last commit.

alex-mpoet (dismissed):
@lysyjan Thank you! Now it's documented.

Aschepikov:
Hi @lysyjan  What we are expecting to see when there is no lastSubscribedAt? I see the invalid date if the user is unconfirmed and never was subscribed 
image
image

lysyjan:
@Aschepikov Nice catch. I added a commit that checks if the value is not empty. I guess this is an expected behavior in our listings when we render nothing for empty values.

@alex-mpoet can you please review the last added commit?

alex-mpoet (approved):

Aschepikov:
@lysyjan Maybe, in this case, we should leave it empty or we can show that it is/was not subscribed, now I see the createdAt for the empty lastSubscribedAt. Is it ok for now?

lysyjan:
@Aschepikov I think it's okay now.

Aschepikov (approved):
GJ!

PR #5508, opened 2024-03-18
Display upgrade page for Tier 1

brezocordero:
## Description

Display the upgrade page also if Tier 1

## Code review notes

_N/A_

## QA notes

Check that the upgrade page is displayed if:
- premium is not installed
- key is valid but they have a Starter plan or they don't have premium
- key is not valid

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5905]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5905]: https://mailpoet.atlassian.net/browse/MAILPOET-5905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Util/License/Features/CapabilitiesManager.php
+++ b/mailpoet/lib/Util/License/Features/CapabilitiesManager.php
@@ -16,6 +16,7 @@ class CapabilitiesManager {
   // Product capabilities mapping
   const MIN_TIER_LOGO_NOT_REQUIRED = 1;
   const MIN_TIER_ANALYTICS_ENABLED = 1;
+  const MIN_TIER_NO_UPGRADE_PAGE = 2;
   const MIN_TIER_UNLIMITED_AUTOMATION_STEPS = 2;
   const MIN_TIER_UNLIMITED_SEGMENT_FILTERS = 2;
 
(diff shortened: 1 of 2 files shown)


mailpoet/lib/Util/License/Features/CapabilitiesManager.php
```diff
@@ -115,4 +116,12 @@ public function getCapabilities(): Capabilities {
       $this->getSegmentFiltersLimit(),
     );
   }
+
+  public function showUpgradePage(): bool {
+    $tier = $this->getTier();
+    if ($this->subscribersFeature->hasValidPremiumKey() && isset($tier) && $tier < self::MIN_TIER_NO_UPGRADE_PAGE) {
```
brezocordero: We could also display the page if the Premium key is not valid but not sure current design works for that.

alex-mpoet: Could you please elaborate, why can't this page be shown when the Premium key is invalid?

brezocordero: > why can't this page be shown when the Premium key is invalid?

It was not sure if we wanted to change existing behavior. 
I've tested the page without a plan, with an invalid key and with a starter key and looks good, so added in this [commit](https://github.com/mailpoet/mailpoet/pull/5508/commits/0c7251de607fadc9929450a1bcfb4e587c9c18c9)

alex-mpoet: Thanks for making the change!

alex-mpoet (approved):

veljkho (approved):
This is verified. Thanks for the PR!

PR #5511, opened 2024-03-20
Email Editor: Fix button rendering in Outlook [MAILPOET-5836]

mikejolley:
## Description

Fixes button rendering in outlook and across other email clients using inline styles (globally and from block settings) using table based markup.

I've included a new `AbstractBlockRenderer` class to house shared logic across renderers, and I've simplified the render logic as much as I think possible.

## Code review notes

- `safe_style_css` lets us define some non-standard properties so WordPress will still compile them.
- `AbstractBlockRenderer` has some utility for getting block styles and compiling CSS to save repetition.
- The original class had a `getMarkup` method with placeholders, but it seemed to make more sense to inline this using `sprintf`.
- I attempted using the conditional padding method from https://www.litmus.com/blog/a-guide-to-bulletproof-buttons-in-email-design before going back to tables which worked across most clients except the newer Outlook versions where no padding was present at all. Using a table for all seemed like the simplest approach given the requirements.
- The previous code had `mso-padding-alt` usage to allow the links to have padding in modern clients but the wrapper to have padding in outlook. This did not work on latest Outlook 365 and resulted in no padding being applied at all. Because of this I've opted to use padding only on the wrapper. cc @costasovo 
- I removed the border fallbacks which didn't appear to be needed. Border style solid only gets forced when borders are enabled, otherwise no border styles are added. I didn't notice any problems when testing email clients with this approach. Test coverage was updated to reflect this change.

I tested the rendered output using Email On Acid. Some examples:

o365_w10_lm_dt
gmailw10_chr26_win

Full report:

[Button testing_jHf8MtQzvSnZ5nSpA2owmXns5w06nWwwLl0rq6kP2k8Rh_2024-03-20 07.21.14.AM.zip](https://github.com/mailpoet/mailpoet/files/14667892/Button.testing_jHf8MtQzvSnZ5nSpA2owmXns5w06nWwwLl0rq6kP2k8Rh_2024-03-20.07.21.14.AM.zip)

## QA notes

Go to MailPoet > Emails > New Email > Create using new editor.

Test adding buttons with various colors, background colors, padding, and typography options. You can check the rendering by previewing the email.

To test other clients you can use Email on Acid as I did above.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5836](https://mailpoet.atlassian.net/browse/MAILPOET-5836)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5836]: https://mailpoet.atlassian.net/browse/MAILPOET-5836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/AbstractBlockRenderer.php
+++ b/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/AbstractBlockRenderer.php
@@ -1,0 +1,37 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Integrations\Core\Renderer\Blocks;
+
+use MailPoet\EmailEditor\Engine\Renderer\ContentRenderer\BlockRenderer;
+use WP_Style_Engine;
+
+/**
+ * Shared functionality for block renderers.
+ */
+abstract class AbstractBlockRenderer implements BlockRenderer {
+  /**
+   * Wrapper for wp_style_engine_get_styles which ensures all values are returned.
+   *
+   * @param array $block_styles Array of block styles.
+   * @param bool $skip_convert_vars If true, --wp_preset--spacing--x type values will be left in the original var:preset:spacing:x format.
+   * @return array
+   */
+  protected function getStylesFromBlock(array $block_styles, $skip_convert_vars = false) {
+    $styles = wp_style_engine_get_styles( $block_styles, [ 'convert_vars_to_classnames' => $skip_convert_vars ] );
+    return wp_parse_args($styles, [
+      'css' => '',
+      'declarations' => [],
+      'classnames' => '',
+    ]);
+  }
+
+  /**
+   * Compile objects containing CSS properties to a string.
+   *
+   * @param array ...$styles Style arrays to compile.
+   * @return string
+   */
+  protected function compileCss(...$styles): string {
+    return WP_Style_Engine::compile_css( array_merge(...$styles), '' );
+  }
+}
--- a/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/AbstractBlockRenderer.php
+++ b/mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/AbstractBlockRenderer.php
@@ -17,7 +17,7 @@ abstract class AbstractBlockRenderer implements BlockRenderer {
    * @return array
    */
   protected function getStylesFromBlock(array $block_styles, $skip_convert_vars = false) {
-    $styles = wp_style_engine_get_styles( $block_styles, [ 'convert_vars_to_classnames' => $skip_convert_vars ] );
+    $styles = wp_style_engine_get_styles($block_styles, ['convert_vars_to_classnames' => $skip_convert_vars]);
     return wp_parse_args($styles, [
       'css' => '',
       'declarations' => [],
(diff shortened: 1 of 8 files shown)


lysyjan:
@mikejolley There are some conflicts and incorrect whitespaces. After merging your PHPCS improvement, it should be easy to fix. Can you please rebase it and fix it?
Thank you

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
```diff
@@ -2,137 +2,101 @@
 
 namespace MailPoet\EmailEditor\Integrations\Core\Renderer\Blocks;
 
-use MailPoet\EmailEditor\Engine\Renderer\ContentRenderer\BlockRenderer;
 use MailPoet\EmailEditor\Engine\SettingsController;
+use MailPoet\EmailEditor\Integrations\Core\Renderer\Blocks\AbstractBlockRenderer;
```
lysyjan: I think this line is redundant because `AbstractBlockRenderer` is from the same namespace. The situation is similar in the `Column` and `Columns` renderer.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/AbstractBlockRenderer.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+abstract class AbstractBlockRenderer implements BlockRenderer {
+  /**
+   * Wrapper for wp_style_engine_get_styles which ensures all values are returned.
+   *
+   * @param array $block_styles Array of block styles.
+   * @param bool $skip_convert_vars If true, --wp_preset--spacing--x type values will be left in the original var:preset:spacing:x format.
+   * @return array
+   */
+  protected function getStylesFromBlock(array $block_styles, $skip_convert_vars = false) {
```
lysyjan, line 19: This method originally returned an object in the Column and Columns classes. Why did you decide to don't convert it again and use an array instead? I'm just curious.

lysyjan (dismissed):
Good job @mikejolley. 
I have one question and noticed a detail, but I think we can move this PR forward.

lysyjan:
@mikejolley Do you want to ask QA for testing, or can we merge it? In my opinion, we can merge it because you used Email on Acid here.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/AbstractBlockRenderer.php
mikejolley: @lysyjan I think it was PHPStan, pre-commit hook, not liking the usage of stdClass. Rather than exclude everything it seemed easier to just leave in the original array format instead of casting.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
mikejolley: Resolved in https://github.com/mailpoet/mailpoet/pull/5511/commits/1104d196af9de50211c044f3b09a538d427377ff

mikejolley:
@lysyjan sounds good. I'll wait for CI to pass again then do it.

mikejolley:
@lysyjan Commit dismissed review so just need ✅

lysyjan (approved):

PR #5517, opened 2024-03-23
Mailpoet 5906 upgrade page

brezocordero:
## Description

- This PR adds the new upgrade page.
- It's based on the figma file + the latest copy (see links on ticket)
- The new upgrade page is only shown when there is a valid key for a new product but the logic could be modified to display the page also if there is no valid key.
- The links in the buttons on the hero section in the footer were not defined, I added links to the shop and FAQs in the kb but might need to be modified.

Not included:
- The footer

## Code review notes

- I separated the buttons to its own template because they had more logic and we might need to change the link structure
@alex-mailpoet for new purchases, how will the shop know this is for the new products and which group?
I guess we need to add tier or group to the purchase/upgrade links?

- The billing switch is the one from the shop adapted to the plugin design. We might need to add some client js if we want to add the billing period to the shop links. `new_plan_selector` in the shop does something similar.

- I think the __ function in Twig escapes the text but I'm not that familiar with it, so please check I didn't miss escaping anything, like [here](https://github.com/mailpoet/mailpoet/commit/eea7d3cd49954c639f16c8b38a7e6dca755ca9af).

- I've tested the layout with larger font in browser settings, the buttons get a bit misaligned on very large but reads well. - Also tested mobile display + keyboard tabbing.

- The contrast in the 'Most popular' label is not good but it is the specified color in the design. It's mainly decorative so I didn't change it.

## QA notes

Unless we have the new products, the only way to test is modifying the code.
Change `mailpoet/lib/AdminPages/Pages/Upgrade.php` to the following and then you can modify the tier to 0,1 or 2:

```
  public function render() {
    $data = [
      'current_mailpoet_plan_tier' => 0,
    ];

    $this->pageRenderer->displayPage('upgrade_tiers.html', $data);
  }
```

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5906]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5906]: https://mailpoet.atlassian.net/browse/MAILPOET-5906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brezocordero:
@websupporter I rebased, added the tooltips and changed the tracking event names.

Remaining questions:
- How do we build the purchase links for the new products?
- What are the links for the hero section actions in the footer?
- Do we need to change the billing period in the shop links when they change the selection?

websupporter:
> How do we build the purchase links for the new products?

*A. No API Key*

Current:
https://account.mailpoet.com/?s=28&g=business&billing=monthly&email=david.remer@localhost.com&utm_source=plugin&utm_medium=premium&utm_campaign=purchase&ref=plugin-upgrade-page

s = number of subscribers the user has
g = the group. we have now "free", "basic", "pro", "agency" and "creator"

the rest should remain with the same behavior.

Important: Additionally, we need a parameter like `&2024=1` to distinct the shown page.

*B. With API Key*
https://account.mailpoet.com/orders/upgrade/<partial-api-key>?utm_source=plugin&utm_medium=premium&utm_campaign=upgrade&ref=plugin-upgrade-page

This seems like it can stay for the most part the same except we would like to have the `2024=1` parameter. The shop will be able to select the correct plans to select from.

> What are the links for the hero section actions in the footer?

These?
image

Where they are pointing to atm lgtm.

> Do we need to change the billing period in the shop links when they change the selection?

For "without api key" there is the billing parameter which works, so I would suggest utilizing it. It seems on the upgrade such a parameter does not work. So for now, no need imo.

JanJakes:
@websupporter @brezocordero Do you guys need me to review this or anything? I'm the reviewer here, but it's assigned to @websupporter, so I'm not sure.

brezocordero:
> Do you guys need me to review this or anything? I'm the reviewer here, but it's assigned to @websupporter, so I'm not sure.

@JanJakes 

I have assigned this back to me and will send it back to you for review after updating the shop links.

It is a big PR, if you want to start reviewing already. I will be changing `mailpoet/views/upgrade/upgrade_button.html` and maybe adding some jQuery to update the links when the billing period changes.

The rest will not be modified. Thanks.

JanJakes:
@brezocordero From the history, it seems like @websupporter took it over when it was assigned to me. This week I'm a Porter, let's see what David says.

mailpoet/lib/AdminPages/Pages/Upgrade.php
```diff
@@ -12,18 +13,33 @@ class Upgrade {
...
     $this->wp = $wp;
+    $this->capabilitiesManager = $capabilitiesManager;
   }
 
   public function render() {
     $data = [
       'current_wp_user' => $this->wp->wpGetCurrentUser()->to_array(),
     ];
+
+    // @todo change this after a/b test to keep only one of the pages
```
websupporter, line 33: Would be good to mention this in [MAILPOET-5986] or mention the ticket in the comment.

mailpoet/lib/Util/License/Features/CapabilitiesManager.php
```diff
@@ -144,4 +144,17 @@ public function showUpgradePage(): bool {
     }
     return false;
   }
+
+  /**
+   * Returns true if there is a valid key for a tier that can be upgraded
+   * @todo remove after the a/b test is finished and keep only one page
```
websupporter, line 150: same [MAILPOET-5986]

mailpoet/views/upgrade_tiers.html
```diff
@@ -0,0 +1,273 @@
...
+    </div>
+  </a>
+</div>
+<div class="mailpoet-tiers-page">
+  <div class="mailpoet-tiers-page-hero-section narrow">
+    <h1 class="mailpoet-tiers-page-hero-section__title">
+      <%= __('Find the plan that fits your flock') %>
```
websupporter: Regarding escaping:

```
__('Find the plan that fits your flock<script>alert()</script>')|escape
```
prevents

image

mailpoet/lib/AdminPages/Pages/Upgrade.php
brezocordero: Added to the ticket

mailpoet/lib/Util/License/Features/CapabilitiesManager.php
brezocordero: Added to the ticket

mailpoet/views/upgrade_tiers.html
brezocordero: Done in https://github.com/mailpoet/mailpoet/pull/5517/commits/477c78a2a8ecd25b01ca42439f88511a6a9a3128

brezocordero:
@websupporter changes are done. Could you please also check if the prices are correct? Not sure if they have changed since.

websupporter (approved):
LGTM. Thank you @brezocordero

PR #5521, opened 2024-03-27
Email Editor: Add typography panel to manage email level styling [MAILPOET-5639]

mikejolley:
## Description

1. Adds typography panel to the new email editor. Here you can customise styles for elements such as text, headings, buttons, and links
2. Updates preview rendering to style elements using new styles
3. Refactors panels/screens for reuse
4. Introduces `useEmailStyles` hook for shared styling logic (getting and setting styles to post meta)
5. Updated dimensions panel to use new hooks

| Main Panel | Element Panel | Tools Panel |
| :----------------: | :------: | :----: |
| Screenshot 2024-03-27 at 16 46 15  |  Screenshot 2024-03-27 at 16 46 23 | Screenshot 2024-03-27 at 16 46 32 |

## Code review notes

I had to fix an issue  in the CSS Inliner which was breaking `calc()` rules for padding. It would cut off and break inline styles with something like `padding: calc(10em;`. This is likely due to shorthand optimisation which I disabled to fix the problem.

Also regarding CSS inliner, it does not support `:not()` selectors. So I could not do `a:not(.wp-block-button a)` to style only links. I worked around it with a different selector, but this might need improvement in a follow up.

## QA notes

Not yet passing.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5639](https://mailpoet.atlassian.net/browse/MAILPOET-5639)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5639]: https://mailpoet.atlassian.net/browse/MAILPOET-5639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/panels/typography-panel.tsx
```diff
@@ -1,34 +1,83 @@
...
+export function TypographyPanel() {
+  return (
+    <Card size="small" variant="primary" isBorderless>
+      <CardBody>
+        <VStack spacing={3}>
+          <Heading level={3} className="edit-site-global-styles-subtitle">
+            {__('Elements')}
+          </Heading>
+          <ItemGroup isBordered isSeparated size="small">
+            <ElementItem element="text" label={__('Text', 'mailpoet')} />
```
lysyjan: Maybe we don't need to use the `mailpoet` domain for those translations because they are not plugin-specific, but we could use here native WP translations. The main reason is that those words are very common. What do you think about it?

mailpoet/assets/js/src/email-editor/engine/store/types.ts
```diff
@@ -56,7 +56,13 @@ export type EmailStyles = {
     letterSpacing: string;
     lineHeight: string;
     textDecoration: string;
-    textTransform: string;
+    textTransform:
```
lysyjan: If I don't overlook something, the typography could be a type that is reused in more places to avoid code repetition. What do you think about it?

lysyjan (commented):
@mikejolley The code looks very good. You make a lot of improvements. Thanks for that.
Nevertheless, I found a few bugs during my testing:

1. If I create a new email, the styles sidebar panels don't work: https://d.pr/v/6UVgSE
    I installed all dependencies and compiled the latest editor version.

2. Heading font-size is not reflected in the email editor: example

3. Heading and button font-sizes are not reflected in rendering: example

4. Change an element's appearance affects other settings: https://d.pr/v/ZhJCow
    This was luck when I noticed, and I think it's unexpected behavior when changing one style resets the others.

mailpoet/assets/js/src/email-editor/engine/store/types.ts
mikejolley: Done in https://github.com/mailpoet/mailpoet/pull/5521/commits/76f4d9c332df6d6adcd8e078c99f8082000e9707

mailpoet/assets/js/src/email-editor/engine/components/styles-sidebar/panels/typography-panel.tsx
mikejolley: Done in https://github.com/mailpoet/mailpoet/pull/5521/commits/54c94abbcc5a5c7a48ee67d5f8d4dde6bc280fd9

mikejolley:
@lysyjan I think this is ready for another look. The button styles had an issue because the block level styles were taking priority. Should be ok now.

mikejolley:
@lysyjan I think this is sorted now. The font error was caused by the WP update.

lysyjan (dismissed):
@mikejolley Good job. I'm approving. I would pass it to QA if you agree because this is quite complex.

Aschepikov:
Hi @mikejolley some found issues
1. I see that this panel is shorter than the window and because of that when I open some settings it looks strange.
For example button typography, you need to enable all options
image
And then open the appearance settings. It looks strange with 2 scrollbars and so on
image
With a smaller window, it looks better but anyway 2 scrollbars are just for this panel and work the same
image

2. Heading settings don't affect a heading,
image

3. Text typography size affects a list/button/link text and doesn't affect a paragraph
image

Aschepikov:
Also, I noticed that email saved messages don't disappear, but I think it is from a separate ticket yes?
image

mikejolley:
> Text typography size affects a list/button/link text and doesn't affect a paragraph

If the paragraph has a size defined at block level it takes precedence @Aschepikov. If you insert a new paragraph and then style with the panel you'll see it size correctly based on that.

> Heading settings don't affect a heading,

In the screenshot you're showing a H2 heading but the H4 styles in the sidebar; was this intentional?

Aschepikov:
> If the paragraph has a size defined at block level it takes precedence @Aschepikov. If you insert a new paragraph and then style with the panel you'll see it size correctly based on that.

@mikejolley Oh I get it, yes it is working as expected

But with the headings no, even with the new heading block it is not changing from the headings typography

mikejolley:
@Aschepikov In your screenshot you have h1/h2 headings, but you're styling a h4 heading. 
Screenshot 2024-04-10 at 11 10 53

I just tested locally and the strikeout option worked for the h2.

Aschepikov:
@mikejolley Sorry I missed that it is a selection for the settings, I thought it should change all headings to it. Now I get it. Thank you for your explanation!

In this case, when I am selecting All and changing the letter case it should change them all but it changes just H3-H6
image

Appearance settings change H2-H6
image

Other settings work perfectly

Aschepikov:
@mikejolley  Hmm or it is connected that here some settings are more important too that's why it is not changing with All

Aschepikov:
Logically, some settings are for all of them, but if the same setting is different in the specific H typography it doesn't change it. So looks like it is expected.

Aschepikov:
@mikejolley Ok, so just a first issue that the panel is shorter than a window. Thank you

mikejolley:
@Aschepikov should be resolved now!

Aschepikov:
@mikejolley  Thank you, I see an excess scrollbar, can we fix it?
image

mikejolley:
@Aschepikov I'm not seeing this. What browser and OS is this on, and do you have excess scrollbars in the post or site editors as well?

Aschepikov:
@mikejolley  I am on Windows with Chrome and Edge, on both, I see this issue if there is an opened list for appearance settings for example
image

mikejolley:
Screenshot 2024-04-22 at 11 50 58

I updated styles to only include a scrollbar on the panel. Should all be resolved now. I will merge this was CI passes.

Aschepikov:
@mikejolley I still see the issue, I will recheck it on the release build tomorrow, it is by the feature flag so there is no problem, I will write to you if the issue will be presented

PR #5523, opened 2024-03-28
Email Editor: Alternative block rendering using block templates [MAILPOET-5978]

mikejolley:
## Description

This PR aims to make the new email editor render blocks more similarly to Gutenberg.

Instead of handling block parsing and rendering manually, we can use a simple block template to render the post content within, and functions such as `get_the_block_template_html` like WP core does. 

- `template-canvas.php` is the equivalent to this [WordPress version](https://github.com/WordPress/WordPress/blob/d5f0448e7835ac76993adc3d93ff21405b71d2e9/wp-includes/template-canvas.php), but with email specific content and no wp-head/footer tags. This is also where things like $preHeader are rendered as before.
- `email-general.html` is the block template and contains a single column and post content. Post content is rendered by a custom block (`mailpoet/email-content`) because the post content block requires a parent query block/context.
- The renderer gets `email-general.html` from the filesystem (it is not exposed in the UI) and uses this to render the blocks via `get_the_block_template_html()`.
- Custom block parsing is removed. We do override the core block parser so we can run a filter for preprocessing.
- I removed the top level prerenderer class which was wrapping content in a column block—we can just use a column block in the block template instead.
- To prevent issues with padding and `alignfull` this PR passes padding values down to children via `email_attrs`. Negative margins are not supported by some email clients

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

https://mailpoet.atlassian.net/browse/MAILPOET-5978

## After-merge notes

_N/A_

## Tasks


mikejolley:
@costasovo I may need some help getting tests to pass because I think the environment doesn't like the gutenberg functions. I've put this up though so you can look at the approach.

mailpoet/lib/EmailEditor/Engine/Renderer/Templates/Templates.php
```diff
@@ -0,0 +1,54 @@
...
+      $template->slug = $templateObject->slug;
+      $template->type = $templateObject->type;
+      $template->title = $templateObject->slug;
+      $template->description = $templateObject->slug;
+      $template->status = 'publish';
+      // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+      $template->has_theme_file = false;
+      $template->origin = $templateObject->source;
+       // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
+      $template->post_types = $templateObject->post_types;
```
costasovo, line 49: This is probably an unimportant detail for now, but shouldn't this contain post type of email custom posts?

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
```diff
@@ -95,7 +95,7 @@ public function render($blockContent, array $parsedBlock, SettingsController $se
       esc_attr($linkStyles->classname),
       esc_attr($linkStyles->css),
       esc_url($buttonUrl),
-      wp_kses_post($buttonText),
+      $buttonText,
```
costasovo, line 98: Why it is safe to remove it now? Does it have something to do with usage of `get_the_block_template_html` ?

costasovo (commented):
@mikejolley Thank you for looking into this! 

I think moving toward get_the_block_template_html` is the right direction because it allows us to render emails that are associated with a template and include template parts.

I would also like ot ask @lysyjan and @pavel-mailpoet to share their opinion.

> template-canvas.php is the equivalent to this [WordPress version](https://github.com/WordPress/WordPress/blob/d5f0448e7835ac76993adc3d93ff21405b71d2e9/wp-includes/template-canvas.php), but with email specific content and no wp-head/footer tags. This is also where things like $preHeader are rendered as before.

I see it is similar to what we did with replacing placeholders in a template where the renderer class needed to be aware of what placeholders are in the template. Now it needs to define the variables. I don't have a strong opinion about which approach is better.

> email-general.html is the block template and contains a single column and post content.

If I understand correctly, this would be our blank template, right?

> Postprocessing is unified to run on the entire template, rather than doing it twice (once for template, once for content)

This is something I was thinking we could push closer to the content or maybe even to blocks. One reason is usage in WPCOM where we need to render just a post, but I guess this could be worked around by a custom code on WPCOM side (I'll try that). Other potential reason I was thinking of is the API for new block renderers. WPCOM uses custom CSS that is inlined for certain blocks. We don't have such a usecase for now. I was thinking if we did the inlining and postprocessing per block we could allow integrators to alter inlined CSS per block or even add/remove post-processors per block and that would provide better isolation for block specific tweaks. But there is no need to change anything now I just wanted to share the idea. I think we will get to it when we will work on API for extending.

> I removed the top level prerenderer class which was wrapping content in a column block—we can just use a column block in the block template instead.

I'm not sure if that's related but when testing it locally I noticed that after this change full-width columns aren't rendered properly.

Screenshot 2024-04-03 at 11 38 16

pavel-mailpoet:
> I think moving toward get_the_block_template_html` is the right direction because it allows us to render emails that are associated with a template and include template parts.

> I would also like ot ask @lysyjan and @pavel-mailpoet to share their opinion.


It is difficult to answer this. It seems like a good idea now, but I don't think I can foresee all the implications. So I say let's go for it.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Button.php
mikejolley: KSES was breaking some inline styles only when using the new template renderer. I added the KSES in a different PR.

lysyjan:
> I would also like ot ask @lysyjan and @pavel-mailpoet to share their opinion.

@costasovo I wrote a comment to the Jira ticket last week that I like Mike's solution. So, I think we can continue with it.

costasovo (approved):
Great job! 👏 

I think that the `EmailContent` block should go to the `MailPoet\EmailEditor\Engine`, as it is a key part for the renderer to work. Now it is in the `MailPoet\EmailEditor\Integrations\MailPoet` The MailPoet integration should contain only stuff specific for the MailPoet plugin. But I think we can merge it and address it later on.

PR #5532, opened 2024-04-03
[Email Editor] Add move to trash button [MAILPOET-5947] 

mikejolley:
## Description

Adds trash functionality to the email editor via the options menu.

Screenshot 2024-04-03 at 15 23 42

## Code review notes

- When a WP Post is trashed, the newsletter is also trashed
- Bulk trash/bulk restore at newsletter level will trash/untrash the WP Post.

## QA notes

Testing instructions:

1. Go to Emails > Create new > Create using new email editor
2. Click the options (...) top right of the interface and choose "move to trash"
3. Confirm and the page should redirect to the listing. The email should not be listed; it should be under `trash` status.
4. Navigate to `trash`
5. Restore the email
6. Edit it again. Confirm you can edit, and move to trash again successfully. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5947](https://mailpoet.atlassian.net/browse/MAILPOET-5947)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5947]: https://mailpoet.atlassian.net/browse/MAILPOET-5947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/components/header/more-menu.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/header/more-menu.tsx
@@ -27,7 +27,8 @@ export function MoreMenu(): JSX.Element {
     'mailpoet_email',
     'status',
   );
-  const { saveEditedEmail } = useDispatch(storeName);
+  const { saveEditedEmail, updateEmailMailPoetProperty } =
+    useDispatch(storeName);
   const goToListings = () => {
     window.location.href = urls.listings;
   };
@@ -83,6 +84,7 @@ export function MoreMenu(): JSX.Element {
                 <MenuItem
                   onClick={async () => {
                     await setStatus('draft');
+                    await updateEmailMailPoetProperty('deleted_at', '');
                     await saveEditedEmail();
                   }}
                 >
--- a/mailpoet/assets/js/src/email-editor/engine/components/header/more-menu.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/header/more-menu.tsx
@@ -1,10 +1,11 @@
 import { MenuGroup, MenuItem } from '@wordpress/components';
 import { useState } from '@wordpress/element';
 import { displayShortcut } from '@wordpress/keycodes';
+import { useEntityProp } from '@wordpress/core-data';
 import { __, _x } from '@wordpress/i18n';
 import { MoreMenuDropdown } from '@wordpress/interface';
 import { PreferenceToggleMenuItem } from '@wordpress/preferences';
-import { useSelect } from '@wordpress/data';
+import { useSelect, useDispatch } from '@wordpress/data';
 import { storeName } from '../../store';
 import { TrashModal } from './trash-modal';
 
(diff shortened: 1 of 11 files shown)


lysyjan (dismissed):
@mikejolley The code looks good to me 👍 

I would like to ask about the behavior because when the email is already in the trash I can open it, and the action for moving to the trash is still available. Should we deactivate it? Or maybe replace it with the restore action?
It's only an idea and it can be addressed in a new ticket. cc @pavel-mailpoet

pavel-mailpoet:
> I would like to ask about the behavior because when the email is already in the trash I can open it, and the action for moving to the trash is still available. Should we deactivate it? Or maybe replace it with the restore action?

- it is fine that you can still open the email from the trash, the old email editor behaves the same. 
- if it is not too difficult, we should change the trash action with the restore action. Or deactivate it, whichever is easier. I am fine doing it as part of this ticket or as a new one.

mikejolley:
@lysyjan added the ability to restore.

mailpoet/assets/js/src/email-editor/engine/components/header/more-menu.tsx
```diff
@@ -72,9 +79,20 @@ export function MoreMenu(): JSX.Element {
               />
             </MenuGroup>
             <MenuGroup>
-              <MenuItem onClick={() => setShowTrashModal(true)} isDestructive>
-                {__('Move to trash', 'mailpoet')}
-              </MenuItem>
+              {status === 'trash' ? (
+                <MenuItem
+                  onClick={async () => {
+                    await setStatus('draft');
```
lysyjan, line 86: I think changing the post status is not enough because the newsletter entity stays marked as deleted. We also have to set the property `deletedAt` as null. This means that after restoring from the menu, the newsletter is still in the trash on the listing.

lysyjan (commented):
Thanks for the update @mikejolley. After the email restoring, I found one inconsistency between the newsletter entity and the WP post.

mailpoet/assets/js/src/email-editor/engine/components/header/more-menu.tsx
mikejolley: Should we remove it then? We don't have an endpoint in place to sync and update both entities at the same time.

lysyjan: When we extend the `mailpoet_data` in [EmailApiController](https://github.com/mailpoet/mailpoet/blob/55aae2c4a12dc5303249a6b9ee05bdaa56b5d331/mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php) it should work. After that, you can add updating mailpoet email property deleted_at: `await updateEmailMailPoetProperty('deleted_at', null);`. What do you think?

mikejolley: @lysyjan Done in https://github.com/mailpoet/mailpoet/pull/5532/commits/11f4f8f8c7ca0f0b87d24ca5e9fd7af2eada3e44

mikejolley:
CI fail is .org related.

lysyjan:
@mikejolley Rebasing the branch on the current trunk should solve it. See related [slack thread](https://a8c.slack.com/archives/C01GH835U90/p1713449019003419)

mikejolley:
@lysyjan passing now!

lysyjan (approved):
LGTM

Aschepikov (approved):
GJ!

PR #5533, opened 2024-04-04
Fix checking whether an email should contain an abandoned cart block [MAILPOET-5999]

JanJakes:
## Description

The problem with the current conditions is that when any of these options doesn't exist at all, the condition will not pass (= it will render the block, but it shouldn't).

## Code review notes

_N/A_

## QA notes

**Before this fix:**

1. Have WooCommerce set up with some existing products.
2. Create an automation, e.g. from Purchased a product template.
3. Add an email and use Abandoned Cart – Fitness template from the WooCommerce Emails templates tab.
4. You won’t see any Abandoned cart block in the email editor.
5. Preview the email, you will see an Abandoned cart block with example products in it — **this is wrong**.

**After the fix:**
In 5, no abandoned cart block should be rendered in the preview.

## Linked PRs

[MAILPOET-5999]

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5999]: https://mailpoet.atlassian.net/browse/MAILPOET-5999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Renderer/Blocks/AbandonedCartContent.php
+++ b/mailpoet/lib/Newsletter/Renderer/Blocks/AbandonedCartContent.php
@@ -49,8 +49,10 @@ class AbandonedCartContent {
       return $optionField && $optionField->getName() === 'event';
     })->first();
     if (
-      ($groupOption instanceof NewsletterOptionEntity && $groupOption->getValue() !== WooCommerceEmail::SLUG)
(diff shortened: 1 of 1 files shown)


brezocordero (approved):
LGTM
I tested the products are not displayed (in preview or else) if the block is not in an abandoned cart email created from an automation.
Also tested it still displays if it is.

Aschepikov (approved):
GJ!

PR #5524, opened 2024-03-28
Add skipping new email editor acceptance test on WP 6.3 [MAILPOET-5987]

lysyjan:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5987]

## After-merge notes

_N/A_

## Tasks

- [ ] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [ ] I added sufficient test coverage
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5987]: https://mailpoet.atlassian.net/browse/MAILPOET-5987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5535, opened 2024-04-05
Update all places with new WP 6.5 and Woo 8.8 rc1

veljkho:
## Description

Update all places where necessary after the new WordPress CMS version is released (in this case 6.5).

Just because we skipped testing against Woo 8.7 due to known bug they fixed in 8.8, we can still test against Woo 8.8 RC build until they release the final version. Better to test against Woo RC latest then 2 versions back.

I will make another PR updating to Woo latest when they release the next 8.8 version.

Additionally, reverted 2 skipped tests [as per this PR](https://github.com/mailpoet/mailpoet/pull/5524), they are now testable against WP 6.5

## Code review notes

2 tests were failing due to false-positive outcome. I refactored them and improved, previously debugged and confirmed what was wrong.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6003]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6003]: https://mailpoet.atlassian.net/browse/MAILPOET-6003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/readme.txt
+++ b/mailpoet/readme.txt
@@ -1,9 +1,9 @@
 === MailPoet - Newsletters, Email Marketing, and Automation ===
 Contributors: mailpoet, woocommerce, automattic
 Tags: email, email marketing, post notification, woocommerce emails, email automation, newsletter, newsletter builder, newsletter subscribers
-Requires at least: 6.3
-Tested up to: 6.4
-Stable tag: 4.48.2
+Requires at least: 6.4
+Tested up to: 6.5
+Stable tag: 4.48.1
 Requires PHP: 7.4
 License: GPLv3
 License URI: https://www.gnu.org/licenses/gpl-3.0.html
--- a/mailpoet/readme.txt
+++ b/mailpoet/readme.txt
@@ -3,7 +3,7 @@ Contributors: mailpoet, woocommerce, automattic
 Tags: email, email marketing, post notification, woocommerce emails, email automation, newsletter, newsletter builder, newsletter subscribers
 Requires at least: 6.4
 Tested up to: 6.5
-Stable tag: 4.48.1
+Stable tag: 4.48.2
 Requires PHP: 7.4
 License: GPLv3
 License URI: https://www.gnu.org/licenses/gpl-3.0.html
--- a/mailpoet/tests/docker/docker-compose.yml
+++ b/mailpoet/tests/docker/docker-compose.yml
@@ -71,7 +71,7 @@ services:
       - mailhog-data:/mailhog-data
 
   wordpress:
-    image: mailpoet/wordpress:${WORDPRESS_IMAGE_VERSION:-wp-6.4_php8.0_20231114.1}
+    image: mailpoet/wordpress:${WORDPRESS_IMAGE_VERSION:-wp-6.5_php8.1_20240405.1}
     container_name: wordpress_${CIRCLE_NODE_INDEX:-default}
     depends_on:
       smtp:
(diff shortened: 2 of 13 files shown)


mailpoet/readme.txt
```diff
@@ -1,9 +1,9 @@
 === MailPoet - Newsletters, Email Marketing, and Automation ===
 Contributors: mailpoet, woocommerce, automattic
 Tags: email, email marketing, post notification, woocommerce emails, email automation, newsletter, newsletter builder, newsletter subscribers
-Requires at least: 6.3
-Tested up to: 6.4
-Stable tag: 4.48.2
+Requires at least: 6.4
+Tested up to: 6.5
+Stable tag: 4.48.1
```
lysyjan: You should not change the stable tag here. The tag should stay the same as it is in `trunk` branch

mailpoet/tests/docker/docker-compose.yml
```diff
@@ -71,7 +71,7 @@ services:
       - mailhog-data:/mailhog-data
 
   wordpress:
-    image: mailpoet/wordpress:${WORDPRESS_IMAGE_VERSION:-wp-6.4_php8.0_20231114.1}
```
lysyjan, line 74: I think we don't want to change the used PHP version. The previous image for WP 6.4 was with PHP 8.0, and the new one is with PHP 8.1. Let me know if you need some assistance here.

mailpoet/tests/performance/docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
       WORDPRESS_DB_NAME: wordpress
 
   wordpress:
-    image: mailpoet/wordpress:wp-6.4_php8.2_20231120.1
+    image: mailpoet/wordpress:wp-6.5_php8.1_20240405.1
```
lysyjan: This is a similar case as above, also here with a newer version of WP you change the version of PHP used.

mailpoet/tests/docker/docker-compose.yml
veljkho: Yes, but as of WP update to 6.5, we had to bump also PHP version. @JanJakes has probably more information to this.

mailpoet/tests/performance/docker-compose.yml
veljkho: I can update here to [wp-6.5_php8.2_20240405.1](https://hub.docker.com/layers/mailpoet/wordpress/wp-6.5_php8.2_20240405.1/images/sha256-1aa2507aec7f1893bc47f2f2b7de23c18eda17c7c02469f39172d70ebb6ea7fc?context=explore), at the moment of committing the tag was not present so was only with 8.1.

mailpoet/readme.txt
veljkho: Thanks, I will update it.

mailpoet/tests/performance/docker-compose.yml
veljkho: Updated to [wp-6.5_php8.2_20240405.1](https://hub.docker.com/layers/mailpoet/wordpress/wp-6.5_php8.2_20240405.1/images/sha256-1aa2507aec7f1893bc47f2f2b7de23c18eda17c7c02469f39172d70ebb6ea7fc?context=explore)

lysyjan (approved):

PR #5539, opened 2024-04-08
Fix guest customer incorrect subscription status - MAILPOET-5963

triple0t:
## Description

This PR updates the behavior of guest `checkout` with OptIn Unchecked.

Previously, when a Woo customer checks out without registration and leaves OptIn Unchecked, MailPoet creates a subscriber with the status `UNCONFIRMED` without any list.

These subscribers cannot be sent newsletters and are only counted toward the user's plan limits.

This is corrected in this PR ensuring the other features are working as expected.

## Code review notes

_N/A_

#### Scenario 1 

* withWooCommerceCheckoutOptinEnabled
* withConfirmationEmailEnabled

**Test**: Check out a Woo product and leave optin **unchecked**.
-> A subscriber should be created with status `UNSUBSCRIBED`

**Test**: Check out a Woo product and leave optin **checked**.
-> A subscriber should be created with status `UNCONFIRMED`

#### Scenario 2
* withWooCommerceCheckoutOptinEnabled
* withConfirmationEmailDisabled

**Test**: Check out a Woo product and leave optin **unchecked**.
-> A subscriber should be created with status `UNSUBSCRIBED`

**Test**: Check out a Woo product and leave optin **checked**.
-> A subscriber should be created with status `SUBSCRIBED`

#### Scenario 3
* withWooCommerceCheckoutOptinDisabled
* withConfirmationEmailEnabled

**Test**: Check out a Woo product and leave optin **unchecked**.
-> A subscriber should be created with status `UNSUBSCRIBED`

#### Scenario 4
* withWooCommerceCheckoutOptinDisabled
* withConfirmationEmailDisabled

**Test**: Check out a Woo product and leave optin **unchecked**.
-> A subscriber should be created with status `UNSUBSCRIBED`
*** How was this even allowed?

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5963](https://mailpoet.atlassian.net/browse/MAILPOET-5963)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5963]: https://mailpoet.atlassian.net/browse/MAILPOET-5963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Subscriptions/WooCheckoutAutomateWooSubscriptionsCest.php
+++ b/mailpoet/tests/acceptance/Subscriptions/WooCheckoutAutomateWooSubscriptionsCest.php
@@ -61,7 +61,7 @@ class WooCheckoutAutomateWooSubscriptionsCest {
     $customerEmail = 'woo_guest_uncheck@example.com';
     $i->orderProductWithoutRegistration($this->product, $customerEmail, false);
     $i->login();
-    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_UNCONFIRMED, null, ['WooCommerce Customers']);
+    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_UNSUBSCRIBED, null, ['WooCommerce Customers']);
     $i->amOnPage('/wp-admin/admin.php?page=automatewoo-opt-ins');
     $i->dontSee($customerEmail, '.automatewoo-content');
     $i->seeConfirmationEmailWasNotReceived();
--- a/mailpoet/lib/Segments/WooCommerce.php
+++ b/mailpoet/lib/Segments/WooCommerce.php
@@ -167,7 +167,7 @@ class WooCommerce {
 
     if (!$wcOrder instanceof \WC_Order) return;
     $signupConfirmation = $this->settings->get('signup_confirmation');
-    $status = SubscriberEntity::STATUS_UNCONFIRMED;
+    $status = SubscriberEntity::STATUS_UNSUBSCRIBED;
     if ((bool)$signupConfirmation['enabled'] === false) {
       $status = SubscriberEntity::STATUS_SUBSCRIBED;
     }
--- a/mailpoet/tests/acceptance/Subscriptions/WooCheckoutBlocksCest.php
+++ b/mailpoet/tests/acceptance/Subscriptions/WooCheckoutBlocksCest.php
@@ -66,7 +66,7 @@ class WooCheckoutBlocksCest {
     $i->logOut();
     $this->orderProduct($i, $customerEmail, true, false);
     $i->login();
-    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_UNCONFIRMED, ['WordPress Users'], ['WooCommerce Customers']);
+    $i->checkSubscriberStatusAndLists($customerEmail, SubscriberEntity::STATUS_UNSUBSCRIBED, ['WordPress Users'], ['WooCommerce Customers']);
     $i->seeConfirmationEmailWasNotReceived();
   }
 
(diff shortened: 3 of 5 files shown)


mailpoet/tests/integration/PostEditorBlocks/WooCommerceBlocksIntegrationTest.php
```diff
@@ -60,21 +60,6 @@ public function testItHandlesOptInForGuestCustomer() {
...
-    $this->setupSyncGuestUserMock($email);
-    $request['extensions']['mailpoet']['optin'] = false;
-    $this->integration->processCheckoutBlockOptin($this->wcOrderMock, $request);
-
-    $subscriber = $this->entityManager->getRepository(SubscriberEntity::class)->findOneBy(['email' => $email]);
-    $this->assertInstanceOf(SubscriberEntity::class, $subscriber);
-    $this->entityManager->refresh($subscriber);
-    verify($subscriber->getStatus())->equals(SubscriberEntity::STATUS_UNCONFIRMED);
-  }
-
```
costasovo: I think it would be better rather to update the test to reflect the new behavior instead of removing it. We can test that the subscriber was not added. IMO, the test still has value. Some changes may cause us to start adding the subscriber, but this test will catch it.

mailpoet/tests/acceptance/Subscriptions/WooCheckoutAutomateWooSubscriptionsCest.php
```diff
@@ -55,18 +55,6 @@ public function checkoutOptInEnabled(\AcceptanceTester $i, $scenario) {
     $i->dontSee(self::AUTOMATE_WOO_OPTIN_TEXT);
   }
 
-  public function checkoutOptInUnchecked(\AcceptanceTester $i) {
```
costasovo: Same as for the integration test, I think we should rather check the subscriber was not added instead of removing the tests.

costasovo (changes requested):
@triple0t Thanks for jumping on fixing this issue!

This fixes the issue for sites that use blocks-based checkout pages, but the issue is still present for the classic checkout page, which is probably still used in many stores created before the blocks page was introduced.

### Replication of the issue on the classic checkout page

1) To setup the classic checkout page you can create a page and add shortcode `woocommerce_checkout`
2) After inserting the product to cart navigate to the page URL and you will see the classic checkout form
3) Fill in the form as a guest user and observe the issue

When I clicked `force sync` in the Lists > WooCommerce Customers the plugin synced the guest customer but this time with the expected `unsubscribed` status. I think this might be a bit confusing ("where they all suddenly came from?").

Considering the issue is also present on the classic checkout page and the force sync creates the subscriber anyway (I'm not sure we can avoid this) maybe we could change the logic so that the subscriber is still created but as as unsubscribed to get the same outcome as from the force sync. What do you think?

mailpoet/tests/acceptance/Subscriptions/WooCheckoutAutomateWooSubscriptionsCest.php
triple0t: Done

mailpoet/tests/integration/PostEditorBlocks/WooCommerceBlocksIntegrationTest.php
triple0t: Done.

triple0t:

The PR has been updated.

costasovo:
@triple0t Sorry for the late review. I missed the notification about your last comment, and I was not assigned to the PR, so I didn't notice I should re-review it. It is better to assign back to the reviewer after fixing so that the person who should act next is assigned.

Thank you for updating the tests. Did you address the issue with the classic checkout (see my comment about it: https://github.com/mailpoet/mailpoet/pull/5539#pullrequestreview-1988380227)? I don't see any changes related to this.

triple0t:
Hi @costasovo 👋🏾

> @triple0t Sorry for the late review. I missed the notification about your last comment, and I was not assigned to the PR, so I didn't notice I should re-review it. It is better to assign back to the reviewer after fixing so that the person who should act next is assigned.

Noted. Not a problem.
 
> Thank you for updating the tests. Did you address the issue with the classic checkout (see my comment about it: [#5539 (review)](https://github.com/mailpoet/mailpoet/pull/5539#pullrequestreview-1988380227))? I don't see any changes related to this.

Yup. 
Instead of creating the subscriber with status `UNCONFIRMED`, we now create them with the status `UNSUBSCRIBED`

This affects both Block and classic checkout.

When the user opts in for marketing emails, the status is updated back to `UNCONFIRMED`.

costasovo (approved):
> Instead of creating the subscriber with status UNCONFIRMED, we now create them with the status UNSUBSCRIBED

@triple0t Sorry I missed that both commit changed 🤦 It looks good! Moving it foeward!

Aschepikov (approved):
GJ!

PR #5544, opened 2024-04-15
Email editor - small refactors in block renderers [MAILPOET-5798]

costasovo:
## Description

As a part of "[Revision of blocks rendering codebase](https://mailpoet.atlassian.net/browse/MAILPOET-5798)" ticket I went through blocks and added a couple of enhancements that hopefully improve the code.

1. I replaced our custom CSS compile function with `WP_Style_Engine::compile_css` and removed the custom function
2. I tried to address that the heading and paragraph block code were 90% identical. At first, I wanted to extract the functionality, but in the end, I decided to merge these two renderers into one and introduced `Text` renderer. Paragraph and Heading are almost the same. There was a problematic style attribute `align`, which is in the heading used for full width but in a paragraph for text alignment (the heading has textAlign for that). This can be handled by an additional check so I don't see it a big issue.

## Code review notes

[Email on Acid test](https://app.emailonacid.com/app/editor/project/BdcGrpOiHMD7f41gRQ3Q5WloBcOXMHSCBJ8GN9AXgBYgw)

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5798]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5798]: https://mailpoet.atlassian.net/browse/MAILPOET-5798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Text.php
```diff
@@ -0,0 +1,90 @@
...
+
+use MailPoet\EmailEditor\Engine\SettingsController;
+
+/**
+ * This renderer covers both core/paragraph and core/heading blocks
+ */
+class Text extends AbstractBlockRenderer {
+  protected function renderContent(string $blockContent, array $parsedBlock, SettingsController $settingsController): string {
+    $blockContent = $this->adjustStyleAttribute($blockContent);
+    return str_replace('{heading_content}', $blockContent, $this->getBlockWrapper($blockContent, $parsedBlock));
```
mikejolley: `heading_content` isn't really correct now this is abstracted. 

I'd probably prefer this and `getBlockWrapper` being combined and using `sprintf` placeholders instead to form the final string. Just note `%` in the resulting string would need to be `%%` to prevent sprintf erroring!

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Text.php
```diff
@@ -0,0 +1,90 @@
...
+      <!--[if mso | IE]></td></tr></table><![endif]-->
+    ';
+  }
+
+  /**
+   * 1) We need to remove padding because we render padding on wrapping table cell
+   * 2) We also need to replace font-size to avoid clamp() because clamp() is not supported in many email clients.
+   * The font size values is automatically converted to clamp() when WP site theme is configured to use fluid layouts.
+   * Currently (WP 6.5), there is no way to disable this behavior.
+   */
```
mikejolley, line 64: Why don't we do this above? Generate the spacing rules separately and put padding where it belongs without manipulating HTML?

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Text.php
```diff
@@ -0,0 +1,90 @@
...
+    if (isset($parsedBlock['attrs']['textAlign'])) {
+      $styles['text-align'] = $parsedBlock['attrs']['textAlign'];
+    } elseif (in_array($parsedBlock['attrs']['align'] ?? null, ['left', 'center', 'right'])) {
+      $styles['text-align'] = $parsedBlock['attrs']['align'];
+    }
+
+    $compiledStyles = $this->compileCss($blockStyles['declarations'], $styles);
+
+    return '
+      <!--[if mso | IE]><table align="left" role="presentation" border="0" cellpadding="0" cellspacing="0" width="100%"><tr><td><![endif]-->
```
mikejolley: Given we have a table here anyway and the styles are the same, do we need the `mso` table? The table within doesn't have any special styling that would break.

mikejolley (commented):
Combining everything into a text renderer is a good idea, but I think we can make some other optimisations within. It could be a follow up since `adjustStyleAttribute` and the other markup exists now.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Text.php
costasovo: @mikejolley Actually, I don't know how to do that without manipulating the original HTML. If user sets custom values for padding and font size the input HTML we get to the renderer looks like this:
```html
<h4 class="wp-block-heading alignfull has-background" style="background-color:#8cb3cb;padding-top:29px;padding-right:19px;padding-bottom:29px;padding-left:19px;font-size:clamp(21.027px, 1.314rem + ((1vw - 3.2px) * 1.351), 34px);">Heading with custom values</h4>
```
The thing is, we need to apply padding and background CSS on the table wrapper to get it rendered in Outlook, but we can't keep it on the original HTML because clients that support styles for paragraphs and headings would render the padding on both.

Maybe I just can't see a better approach. Any idea?

costasovo:
@mikejolley I addressed two of your comments in the two latest commits. For the third one, I would like to ask for your feedback.

mailpoet/lib/EmailEditor/Integrations/Core/Renderer/Blocks/Text.php
mikejolley: I understand now and see why its done this way. I would have suggested re-rendering the element, but since this code is shared across headings and paragraphs the current solution is better. You can ignore my comment.

costasovo: > I would have suggested re-rendering the element, but since this code is shared across headings and paragraphs the current solution is better.

@mikejolley I was trying that approach. I could get the tag using the WP_HTML_Tag_Processor and then the inner HTML, but this doesn't seem to be supported by the WP HTML API, so I would probably have to use PHP Dom so it didn't look like a simpler/cleaner solution than what we have in place.

> You can ignore my comment.

In that case. Could you please take another look at the PR? I added two commits.

mikejolley (approved):

PR #5565, opened 2024-04-29
Wait for email sent in send automation step [MAILPOET-4977]

JanJakes:
## Description

Note that the large timeout (1 month) is for cases such as when solving a ban or sending issues takes more than a week.

## Code review notes

_N/A_

## QA notes

The "Send email" automation step should now be "running" until it is really sent. The automation will continue only after the email will be successfully sent and the step completed. If sending fails, the step should fail as well. If no sending is successful for a long time (1 month), the step will fail.

## Linked PRs

[MAILPOET-4977]

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4977]: https://mailpoet.atlassian.net/browse/MAILPOET-4977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Control/ActionScheduler.php
+++ b/mailpoet/lib/Automation/Engine/Control/ActionScheduler.php
@@ -2,6 +2,8 @@
 
 namespace MailPoet\Automation\Engine\Control;
 
+use ActionScheduler_Action;
+
 class ActionScheduler {
   private const GROUP_ID = 'mailpoet-automation';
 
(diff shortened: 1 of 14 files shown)


alex-mpoet:
@JanJakes Looks good overall. I just wanted to make sure I understand the whole flow correctly:
1. The `SendEmailAction` schedules a progress run to be executed in a month from now.
2. The run is kept in the running state until the email is sent, the sending fires the `mailpoet_automation_email_sent` hook.
3. Triggered by the hook, `SendEmailAction` enqueues an immediate progress run which replaces the run scheduled in point 1.
4. In the progress run it checks that the email was sent. If it wasn't sent, it throws an exception, and the run is failed, the automation halts. If the email was sent, it does nothing thus letting the step to be completed, the automation moves to the next step or completes.

JanJakes:
@alex-mpoet Your understanding is 100% correct. Sorry, I should've added such a description to the PR myself.

alex-mpoet (approved):

Aschepikov (approved):
GJ!

PR #5566, opened 2024-04-29
Refactor CI workflows and testing logic with WC HPOS and syncing

veljkho:
## Description

Please see [Jira ticket](https://mailpoet.atlassian.net/browse/MAILPOET-6040) for additional description.

This change is about renaming COT to HPOS (the cli commands will also be renamed to hpos from WC 8.9) but also refactoring how we test against HPOS enabled, disabled and with syncing as active.

The HPOS is enabled by default for new WC Installations (without syncing as ON).

We used plugin COT helper but that's very obsolete way to test against HPOS.

After the release of WC 8.9, I will make a follow-up PR to update CLI commands to use `hpos` instead `cot`.
Jira [ticket prepared](https://mailpoet.atlassian.net/browse/MAILPOET-6041).

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6040]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6040]: https://mailpoet.atlassian.net/browse/MAILPOET-6040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo (changes requested):
@veljkho Thank you for the update!

I noticed that this change breaks HPOS configuration for local development environment. We need to update env variables used in mailpoet/RoboFile.php. I guess we could also rename parameters there from enable-cot to enable-hpos etc.

Parameters for integration tests are defined here: https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/RoboFile.php#L354-L363

Parameters for acceptance tests here:
https://github.com/mailpoet/mailpoet/blob/e47ce06b4a7f59b92212c25c350e6fa76cc365e7/mailpoet/RoboFile.php#L410-L412

and the parameters are then used for setting the env variables here:
https://github.com/mailpoet/mailpoet/blob/e47ce06b4a7f59b92212c25c350e6fa76cc365e7/mailpoet/RoboFile.php#L1640-L1641

Could you please update those as well?

veljkho:
> Could you please update those as well?

Thanks for the review and catching what I missed to update @costasovo 

I updated them now. `enable-cot` is now `disable-hpos,` and I optionally added `enable-hpos` for any edge-case-specific testing needs (we know that hpos is enabled by default anyway).

I also deleted the cot helper plugin in my new commit. Please review it once again.

costasovo (approved):
I tested it locally and the commands work as expected.
Good job! Thank you for updating the test config.

PR #5568, opened 2024-04-30
Display when subscriber was created in listings [MAILPOET-6042]

pavel-mailpoet:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6042]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6042]: https://mailpoet.atlassian.net/browse/MAILPOET-6042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/subscribers/list.tsx
+++ b/mailpoet/assets/js/src/subscribers/list.tsx
@@ -82,12 +82,12 @@ const columns = [
   },
   {
     name: 'last_subscribed_at',
-    label: MailPoet.I18n.t('confirmedOn'),
+    label: MailPoet.I18n.t('subscribedOn'),
     sortable: true,
   },
   {
     name: 'created_at',
-    label: MailPoet.I18n.t('subscribedOn'),
+    label: MailPoet.I18n.t('createdOn'),
     sortable: true,
   },
 ];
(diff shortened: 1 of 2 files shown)


mailpoet/assets/js/src/subscribers/list.tsx
```diff
@@ -82,6 +82,11 @@ const columns = [
   },
   {
     name: 'last_subscribed_at',
+    label: MailPoet.I18n.t('confirmedOn'),
```
lysyjan: I'm not sure about changing the label for the column `last_subscribed_at` because with the name `Subscribed on`, there was better consistency between its value and the segment filter `subscribed date`.
Do you think we can come to a compromise here?

I have in my mind two ideas now:
1. change the filter name from `subscribed date` to `subscription confirmed at`
2. using label `Created on` for the values`created_at`

What do you think about it? Do you have more ideas?

pavel-mailpoet: Good point. I was thinking about it. I am not sure about changing the filter name, which could cause more confusion. 

I thought that the name "Subscribed on" refers to the time when the user actually subscribed – that means when they filled in the form. But I agree with your point. So, I will go with option 2. and change the label.

lysyjan (approved):
Thank you @pavel-mailpoet!

Aschepikov (approved):
GJ!

PR #5571, opened 2024-04-30
Fix scheduled task entity having the same association twice [MAILPOET-6047]

JanJakes:
## Description

Fix scheduled task entity having the same association twice
It also seems that `orphanRemoval` doesn't play well with our explicit `detachAll` logic.

I realized that it would be better to remove cascades and orphan removals anyway because they always need to load the whole collection into memory and then delete/update the entities one by one.

## Code review notes

I think this doesn't need QA.

## QA notes

I think this doesn't need QA.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6047]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6047]: https://mailpoet.atlassian.net/browse/MAILPOET-6047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Entities/ScheduledTaskEntity.php
+++ b/mailpoet/lib/Entities/ScheduledTaskEntity.php
@@ -95,15 +95,8 @@ class ScheduledTaskEntity {
    */
   private $sendingQueue;
 
-  /**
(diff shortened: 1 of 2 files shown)


costasovo (approved):
Ha, it was used only in the inversedBy. Nice that we caught it before we used it in more places.

PR #5572, opened 2024-05-01
Email Editor: Add color styles panel [MAILPOET-5641]

mikejolley:
## Description

Adds color options to the styles panel:

Screenshot 2024-05-01 at 15 00 53

Selections are saved with the current template globally. This PR also ensures that the frontend renderer loads these styles and merges them with post + theme styles, as well as fixing some issues that the renderer had replacing color presets. 

In the ticket ([MAILPOET-5641](https://mailpoet.atlassian.net/browse/MAILPOET-5641)) we also mention palettes, but there is a separate ticket to build the palette UI.

The designs also include 2 background, but we no longer need this because content background is at block level. We also should avoid using custom components where possible, so I've used the Color Panel from Guternberg—we should just use that design for consistency.

## Code review notes

To test:

1. Create a new email using the new editor
2. Open the styles panel, top right of the screen
3. Change some options, for example, layout background
4. Save
5. Preview. Ensure styles persist.

## QA notes

Please test using the instructions above. The changes may also affect other style settings, so please test briefly also other style sections.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5641](https://mailpoet.atlassian.net/browse/MAILPOET-5641)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5641]: https://mailpoet.atlassian.net/browse/MAILPOET-5641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5641]: https://mailpoet.atlassian.net/browse/MAILPOET-5641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/store/types.ts
+++ b/mailpoet/assets/js/src/email-editor/engine/store/types.ts
@@ -29,6 +29,88 @@ export type EmailEditorSettings = EditorSettings & ExperimentalSettings;
 export type EmailTheme = {
   version?: number;
   styles?: EmailStyles;
+  // Ref: https://github.com/WordPress/gutenberg/blob/38d0a4351105e6ba4b72c4dcb90985305aacf921/packages/block-editor/src/components/global-styles/hooks.js#L24C7-L24C21
+  settings?: {
+    appearanceTools?: boolean;
+    useRootPaddingAwareAlignments?: boolean;
+    background?: {
+      backgroundImage?: boolean;
+      backgroundRepeat?: boolean;
+      backgroundSize?: boolean;
+      backgroundPosition?: boolean;
+    };
+    border?: {
+      radius?: boolean;
+      width?: boolean;
+      style?: boolean;
+      color?: boolean;
+    };
+    shadow?: {
+      presets?: boolean;
+      defaultPresets?: boolean;
+    };
+    color?: {
+      background?: boolean;
+      button?: boolean;
+      caption?: boolean;
+      custom?: boolean;
+      customDuotone?: boolean;
+      customGradient?: boolean;
+      defaultDuotone?: boolean;
+      defaultGradients?: boolean;
+      defaultPalette?: boolean;
+      duotone?: boolean;
+      gradients?: {
+        default?: boolean;
+        theme?: boolean;
+        custom?: boolean;
+      };
+      heading?: boolean;
+      link?: boolean;
+      palette?: boolean;
+      text?: boolean;
+    };
+    dimensions?: {
+      aspectRatio?: boolean;
+      minHeight?: boolean;
+    };
+    layout?: {
+      contentSize?: string;
+      wideSize?: string;
+    };
+    spacing?: {
+      customSpacingSize?: number;
+      blockGap?: number;
+      margin?: boolean;
+      padding?: boolean;
+      spacingSizes?: number[];
+      spacingScale?: number;
+      units?: string[];
+    };
+    position?: {
+      fixed?: boolean;
+      sticky?: boolean;
+    };
+    typography?: {
+      customFontSize?: boolean;
+      defaultFontSizes?: boolean;
+      dropCap?: boolean;
+      fontFamilies?: boolean;
+      fontSizes?: boolean;
+      fontStyle?: boolean;
+      fontWeight?: boolean;
+      letterSpacing?: boolean;
+      lineHeight?: boolean;
+      textColumns?: boolean;
+      textDecoration?: boolean;
+      textTransform?: boolean;
+      writingMode?: boolean;
+    };
+    lightbox?: {
+      enabled?: boolean;
+      allowEditing?: boolean;
+    };
+  };
 };
 
 export interface TypographyProperties {
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/Renderer.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/Renderer.php
@@ -37,6 +37,20 @@ class Renderer {
     return self::$theme;
   }
 
+  public static function maybeConvertPreset($value, $variables) {
+    if (is_string($value) && str_contains($value, 'var:preset|color|')) {
+        $value = str_replace('var:preset|color|', '', $value);
+        $value = sprintf('var(--wp--preset--color--%s)', $value);
+    }
+
+    foreach ($variables as $varName => $varValue) {
+      $varPattern = '/var\(' . preg_quote($varName, '/') . '\)/i';
+      $value = preg_replace($varPattern, $varValue, $value);
+    }
+
+    return $value;
+  }
+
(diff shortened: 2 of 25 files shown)


mailpoet/assets/js/src/email-editor/engine/store/types.ts
```diff
@@ -29,6 +29,88 @@ export type EmailEditorSettings = EditorSettings & ExperimentalSettings;
 export type EmailTheme = {
   version?: number;
   styles?: EmailStyles;
+  // Ref: https://github.com/WordPress/gutenberg/blob/38d0a4351105e6ba4b72c4dcb90985305aacf921/packages/block-editor/src/components/global-styles/hooks.js#L24C7-L24C21
+  settings?: {
```
costasovo, line 33: I see a lot of booleans. I would expect many of those to be a string (e.g., `color. background`). Are those booleans correct?

mailpoet/lib/EmailEditor/Engine/Renderer/Renderer.php
```diff
@@ -45,6 +45,20 @@ public static function getTheme() {
     return self::$theme;
   }
 
+  public static function maybeConvertPreset($value, $variables) {
```
costasovo: Does this need to be `public static`? I see usage only in this class, so what about making it private? 

I also see it is needed because `$this->themeController->getStyles` may return some variables in the "compressed" format. The `ThemeController:getStyles` is also used in the ContentRenderer to pass styles to preprocessors. These are the only two usages and both are in the renderer. What about moving this `maybeConvertPreset` to the ThemeController? If we want to retain the `getStyles` to return exactly what is set we could introduce another `ThemeController:getStylesWithValidPresets` method that would return styles with valid variable names? What do you think?

costasovo (commented):
Thank you for adding the panel! It is nice that we don't need to add many new components 👍 

I have a few observations from testing: 

1) I noticed that there is no color setting for Links. I see it in Figma and see that it is available in global styles in the site editor. I find links pretty useful. Could we add those?

2) I added a button and set no styles, but the colors from the styles settings were not applied. We have some default styles for button block defined in [theme.json](https://github.com/mailpoet/mailpoet/blob/a8e9b167cb2d5454938cc0330e425a37cde2956f/mailpoet/lib/EmailEditor/Integrations/Core/theme.json#L8-L11) and it looks like these override the settings from the styles panel. Idealy the settings in the styles panel should replace those. I wonder if that could be related to element vs block styles settings on the theme.json level.

3) When I was changing background color the canvas was flickering a lot (see video).  It looks like that when we apply the styles there is a short period when some empty styles are applied or something like that. I'm not sure if that is related to this change.

https://github.com/mailpoet/mailpoet/assets/1082140/0a3bffbe-f37b-4e15-b7ff-2d55856d3719

mailpoet/assets/js/src/email-editor/engine/store/types.ts
mikejolley: Some may be wrong, but `settings` is not the same as `styles`. Most of these determine if a field is supported. Thats likely why "link" is not being shown.

costasovo: Oh! Makes sense! I thought these were for styles. Now I see styles reuse EmailStyles 👍

mailpoet/lib/EmailEditor/Engine/Renderer/Renderer.php
mikejolley: Handled in https://github.com/mailpoet/mailpoet/pull/5572/commits/24ce2a8fe30524ff2aaa83ff5b723f3ee8918d91

mikejolley:
1. I added link styles in https://github.com/mailpoet/mailpoet/pull/5572/commits/f628b264e6aebbd1978fc3a9c5428b1b2b9e4ad9 Sadly you cannot disable :hover styles without replacing the entire component. For now I added a work around to "support" hover, after inlining, but we might need to add something upstream so it can be turned off. e.g. setting for `linkHover: false`.

mikejolley:
@costasovo This can be looked at again. Reported issues should be resolved.

costasovo (dismissed):
@mikejolley Great job! I tested the panel and it worked as expected for me! 
Thank you for splitting the Template class 👍

Aschepikov:
Hi @mikejolley there are failed tests and conflicts can you check it pls

mikejolley:
@Aschepikov passing now, just needed a rebase.

Aschepikov:
@mikejolley Thank you!

Aschepikov:
Hi @mikejolley there are some conflicts can you resolve them pls
Also I can't select colors for the H1 there is always no color selected
image
The text color changes the link text color but not the text (on the preview it is working like it was changed)
image
And the background color doesn't change anything (on the preview it is working like it was changed)
image
On the preview, I can see changed background and text color
image

Aschepikov:
Two times I saw this error can you tell me when it was implemented and in what cases? As I didn't see it on the previous versions I can't figure out how I am reproducing it. Should I try more or is it ok 
image

SyntaxError: "undefined" is not valid JSON
    at JSON.parse (<anonymous>)
    at https://alex.qawp.net/wp-content/plugins/gutenberg/build/block-editor/index.min.js?ver=ed236a1b8017e31195bc:202:20262
    at Array.map (<anonymous>)
    at QP (https://alex.qawp.net/wp-content/plugins/gutenberg/build/block-editor/index.min.js?ver=ed236a1b8017e31195bc:202:20240)
    at https://alex.qawp.net/wp-content/plugins/gutenberg/build/block-editor/index.min.js?ver=ed236a1b8017e31195bc:202:21561
    at tM (https://alex.qawp.net/wp-content/plugins/gutenberg/build/block-editor/index.min.js?ver=ed236a1b8017e31195bc:202:22494)
    at https://alex.qawp.net/wp-content/plugins/gutenberg/build/block-editor/index.min.js?ver=ed236a1b8017e31195bc:202:28362
    at Object.Ht [as useMemo] (https://alex.qawp.net/wp-content/plugins/gutenberg/build/vendors/react-dom.min.js?ver=18:1:50054)
    at e.useMemo (https://alex.qawp.net/wp-content/plugins/gutenberg/build/vendors/react.min.js?ver=18:1:9894)
    at rM (https://alex.qawp.net/wp-content/plugins/gutenberg/build/block-editor/index.min.js?ver=ed236a1b8017e31195bc:202:27837)

mikejolley:
@Aschepikov could you check again after rebase? It seems to be working now. I enabled iframe mode which also fixed the background color.

Aschepikov (approved):
Looks great, thank you!

PR #5585, opened 2024-05-08
Fix performance test for saving segment after calculating segmentsCreateCustom

veljkho:
## Description

Fix performance test in the nightly tests.
There is a race condition before calculating segment and saving it.
Added wait for text before clicking save button.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6062]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6062]: https://mailpoet.atlassian.net/browse/MAILPOET-6062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/performance/tests/segments-create-custom.js
+++ b/mailpoet/tests/performance/tests/segments-create-custom.js
@@ -21,7 +21,7 @@ import {
   fullPageSet,
   screenshotPath,
 } from '../config.js';
-import { login, selectInReact } from '../utils/helpers.js';
+import { login, selectInReact, focusAndClick } from '../utils/helpers.js';
 
 export async function segmentsCreateCustom() {
   const page = browser.newPage();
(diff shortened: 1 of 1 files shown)


brezocordero (commented):
@veljkho ok for the change but which is the test that was failing?
I looked at the latest nightly runs of the performance tests and they are green.

veljkho:
Hi @brezocordero 
Thanks for the review, good question.

For performance tests we don't have failures in terms of when the test breaks (the E2E part of it). Instead, we have thresholds for reached performance metrics. Usually performance tests are like that, they aren't made to stop development or nightly runs due to broken test, instead we get notice that something's wrong on the Grafana or CI for that and make necessary change afterwards.

If you check [the nightly CI run](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/17789/workflows/0fcc91d0-0f64-4a81-930b-27169d2bc162/jobs/302032) and expand the `Run performance tests` tab, you will see that the number of passed checks is 35, but the maximum is 41 ( [as stated in the scenario file](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/performance/scenarios.js#L109) ).

I am monitoring Grafana dashboard for performance and I could see if something's wrong with the checks there as well. Would be good to have some notification alert in the future.

brezocordero:
> instead we get notice that something's wrong on the Grafana or CI for that and make necessary change afterwards.

Thanks for the explanation @veljkho !

PR #5586, opened 2024-05-09
Do not run cron when a plugin update is running - MAILPOET-6021

triple0t:
## Description

Do not run cron when a plugin update is running.

The ticket description mentioned this: `We need to notify a user if the cron is stuck due to ongoing update`, but that's not necessary again at this point since we are stopping all cron actions before plugin update.

## Code review notes

I have no clue how QA can test this, but Devs can test the PR by adding `error_log` or `var_dump` to the required functions.
You can imitate the plugin update process by switching the [version of the plugin ](https://github.com/mailpoet/mailpoet/blob/b440a791f954e4da34267f5bf998ebfc0f11a568/mailpoet/mailpoet.php#L5) to a lower version. WordPress should prompt you to update the plugin.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6021](https://mailpoet.atlassian.net/browse/MAILPOET-6021)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6021]: https://mailpoet.atlassian.net/browse/MAILPOET-6021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -554,6 +554,11 @@ class Hooks {
       10,
       2
     );
+
+    $this->wp->addAction(
+      'action_scheduler_before_process_queue',
+      [$this, 'deactivateCronWhenInMaintenanceMode']
+    );
   }
 
   /**
@@ -582,5 +587,19 @@ class Hooks {
     $this->cronTrigger->disable();
 
     return $response;
+  }
+
+  public function deactivateCronWhenInMaintenanceMode(): void {
+    if (!$this->wp->wpIsMaintenanceMode()) {
+      return;
+    }
+
+    $this->wp->addFilter('action_scheduler_queue_runner_batch_size', function () {
+      // return 0 batch sizes to prevent the queue runner from running;
+      // this is the fastest way to stop the current running cron
+      return 0;
+    });
+
+    $this->cronTrigger->disable();
   }
 }
--- a/mailpoet/lib/Config/Hooks.php
+++ b/mailpoet/lib/Config/Hooks.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Config;
 
+use MailPoet\Cron\CronTrigger;
 use MailPoet\Form\DisplayFormInWPContent;
 use MailPoet\Mailer\WordPress\WordpressMailerReplacer;
 use MailPoet\Newsletter\Scheduler\PostNotificationScheduler;
@@ -66,6 +67,9 @@ class Hooks {
   /** @var WooSystemInfoController */
   private $wooSystemInfoController;
 
+  /** @var CronTrigger */
+  private $cronTrigger;
+
   public function __construct(
     Form $subscriptionForm,
     Comment $subscriptionComment,
@@ -82,7 +86,8 @@ class Hooks {
     WP $wpSegment,
     DotcomLicenseProvisioner $dotcomLicenseProvisioner,
     AutomateWooHooks $automateWooHooks,
-    WooSystemInfoController $wooSystemInfoController
+    WooSystemInfoController $wooSystemInfoController,
+    CronTrigger $cronTrigger
   ) {
     $this->subscriptionForm = $subscriptionForm;
     $this->subscriptionComment = $subscriptionComment;
@@ -100,6 +105,7 @@ class Hooks {
     $this->dotcomLicenseProvisioner = $dotcomLicenseProvisioner;
     $this->automateWooHooks = $automateWooHooks;
     $this->wooSystemInfoController = $wooSystemInfoController;
+    $this->cronTrigger = $cronTrigger;
   }
 
   public function init() {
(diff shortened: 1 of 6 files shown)


costasovo (commented):
Thank you for adding the functionality! Nice idea using the maintenance mode check.
I tested it by adding logging to file `cron.log` which is created in plugins folder and tested on a jurrasic.ninja site. I logged when cron was disabled (both upgrade hook and maintanece mode check) but also logged when the cron was re-enabled in CronTrigger.

It worked as expected when updating the plugin via plugins page update link. But when I was using the file upload and reinstalled/updated the plugin by providing a zip file it didn't log that the cron was disabled. Perhaps WordPress doesn't trigger the hook for this case? I don't know. It would be nice to cover also this case. But IMO, it is not a must-have.

I would also suggest adding the test for the maintenance mode to one more place. When the cron daemon is [running it may run for 20+ seconds and it creates new services using DI container for every worker](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Cron/Daemon.php#L47-L72) (getWorkers is a generator). I think the risk here is that because this is a long running separate request (or background job) the plugin files might be deleted by update in between the individual worker runs and suddenly it may fail to instantiate the next worker. So I think it would be nice to add the check for the maintenance mode somwhere into [the loop](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Cron/Daemon.php#L47-L72) and break it in case the mode is active. Does that make sense?

> The ticket description mentioned this: We need to notify a user if the cron is stuck due to ongoing update, but that's not necessary again at this point since we are stopping all cron actions before plugin update.

I see that we rely on WP maintenance mode and every next request reactivates crons so I agree we don't need to notify.

Here is the zip I used:
[mailpoet.zip](https://github.com/mailpoet/mailpoet/files/15262521/mailpoet.zip)

triple0t:

> But when I was using the file upload and reinstalled/updated the plugin by providing a zip file it didn't log that the cron was disabled. Perhaps WordPress doesn't trigger the hook for this case? I don't know. It would be nice to cover also this case. But IMO, it is not a must-have.

WordPress sees this as an install event, not an upgrade event, so while this hook `upgrader_pre_install` is triggered, it doesn't provide enough information to differentiate and identify which plugin is being installed.


> I would also suggest adding the test for the maintenance mode to one more place. When the cron daemon is [running it may run for 20+ seconds and it creates new services using DI container for every worker](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Cron/Daemon.php#L47-L72) (getWorkers is a generator). I think the risk here is that because this is a long running separate request (or background job) the plugin files might be deleted by update in between the individual worker runs and suddenly it may fail to instantiate the next worker. So I think it would be nice to add the check for the maintenance mode somwhere into [the loop](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Cron/Daemon.php#L47-L72) and break it in case the mode is active. Does that make sense?

Thanks for the suggestion. Done here https://github.com/mailpoet/mailpoet/pull/5586/commits/79f0d48193b2ea1593ac7cf96b825d077d61e8d9

costasovo:
> > But when I was using the file upload and reinstalled/updated the plugin by providing a zip file it didn't log that the cron was disabled. Perhaps WordPress doesn't trigger the hook for this case? I don't know. It would be nice to cover also this case. But IMO, it is not a must-have.

> WordPress sees this as an install event, not an upgrade event, so while this hook upgrader_pre_install is triggered, it doesn't provide enough information to differentiate and identify which plugin is being installed.

@triple0t I see. Hmm, what about using the `upgrader_pre_install` and ignoring which plugin is being installed? I think we could still use the hook and simply disable cron no matter what plugin is being installed. The question is if WordPress triggers a callback coming from a plugin that is currently being installed. Do you know?

triple0t:
Hi @costasovo 

> Hmm, what about using the upgrader_pre_install and ignoring which plugin is being installed? I think we could still use the hook and simply disable cron no matter what plugin is being installed.

I don't think that's a good idea. The filter is called multiple times and at different action points (plugin install/upgrade, theme install/upgrade). Disabling MailPoet Cron all that time would definitely have a negative effect.

> The question is if WordPress triggers a callback coming from a plugin that is currently being installed. Do you know?

In this case, the callback is called and executed before the plugin is updated.

costasovo (approved):
@triple0t Thank you for adding the check to the workers loop 👍 

> I don't think that's a good idea. The filter is called multiple times and at different action points (plugin install/upgrade, theme install/upgrade). Disabling MailPoet Cron all that time would definitely have a negative effect.

Ok. Let's keep this as an option for the future. After we release this we should observe a decrease in number of errors we get after releases.

costasovo:
@triple0t Approved. It is hard to be tested by QA, so please feel free to merge it.

PR #5588, opened 2024-05-10
Email Editor: Add support for core/group block [MAILPOET-6057]

mikejolley:
## Description

Adds group to the list of supported blocks. Disables support for group variations because flexbox causes issues in email clients.

While doing this I adjusted how contentSize is applied so groups inherit it correcrly, and I standardised some block class naming.

@costasovo could use your feedback on this one once tests are passing.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6057](https://mailpoet.atlassian.net/browse/MAILPOET-6057)
[MAILPOET-6052](https://mailpoet.atlassian.net/browse/MAILPOET-6052)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6057]: https://mailpoet.atlassian.net/browse/MAILPOET-6057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-6052]: https://mailpoet.atlassian.net/browse/MAILPOET-6052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer/ContentRenderer.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer/ContentRenderer.php
@@ -92,7 +92,7 @@ class ContentRenderer {
     $layout = $this->settingsController->getLayout();
     $styles .= sprintf(
       '
-      .is-layout-constrained > *:not(.alignleft):not(.alignright):not(.alignfull):not(.alignwide) {
+      .is-layout-constrained > *:not(.alignleft):not(.alignright):not(.alignfull) {
         max-width: %1$s;
         margin-left: auto !important;
         margin-right: auto !important;
@@ -109,8 +109,7 @@ class ContentRenderer {
 
     // Get styles from theme.
     $styles .= $this->themeController->getStylesheetForRendering($post);
-
-    $blockSupportStyles = \wp_style_engine_get_stylesheet_from_context('block-supports', []);
+    $blockSupportStyles = $this->themeController->getStylesheetFromContext('block-supports', []);
     // Get styles from block-supports stylesheet. This includes rules such as layout (contentWidth) that some blocks use.
     // @see https://github.com/WordPress/WordPress/blob/3c5da9c74344aaf5bf8097f2e2c6a1a781600e03/wp-includes/script-loader.php#L3134
     // @internal :where is not supported by emogrifier, so we need to replace it with *.
--- a/mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer/ContentRenderer.php
+++ b/mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer/ContentRenderer.php
@@ -87,15 +87,55 @@ class ContentRenderer {
    */
   private function inlineStyles($html, WP_Post $post) {
     $styles = (string)file_get_contents(dirname(__FILE__) . '/' . self::CONTENT_STYLES_FILE);
+
+    // Apply default contentWidth to constrained blocks.
+    $layout = $this->settingsController->getLayout();
+    $styles .= sprintf(
+      '
+      .is-layout-constrained > *:not(.alignleft):not(.alignright):not(.alignfull):not(.alignwide) {
+        max-width: %1$s;
+        margin-left: auto !important;
+        margin-right: auto !important;
+      }
+      .is-layout-constrained > .alignwide {
+        max-width: %2$s;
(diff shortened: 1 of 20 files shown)


mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer/ContentRenderer.php
```diff
@@ -87,15 +87,55 @@ private function reset() {
    */
   private function inlineStyles($html, WP_Post $post) {
     $styles = (string)file_get_contents(dirname(__FILE__) . '/' . self::CONTENT_STYLES_FILE);
+
+    // Apply default contentWidth to constrained blocks.
+    $layout = $this->settingsController->getLayout();
+    $styles .= sprintf(
+      '
+      .is-layout-constrained > *:not(.alignleft):not(.alignright):not(.alignfull):not(.alignwide) {
```
mikejolley: I added these rules so our defined contentSize in theme.json was applied by default. Without this, everything was full width during render, but narrow in editor.

mikejolley:
@costasovo I will leave this as draft. It's done but I need some help with mocking in the tests. Will sort it out Monday.

mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer/ContentRenderer.php
```diff
@@ -87,15 +87,55 @@ private function reset() {
...
+      'group-is-layout-$1 > tbody tr td >',
+      $blockSupportStyles
+    );
+
+    $styles .= $blockSupportStyles;
+
+    // Debugging for content styles. Remember these get inlined.
+    //echo '<pre>';
+    //var_dump($styles);
+    //echo '</pre>';
```
costasovo, line 136: Isn't this a forgotten debug code?

costasovo:
@mikejolley I looked into the failing tests and updated the mock in a new commit. 

I played with the block, and when I inserted it into the email, it showed "Group blocks together. Select a layout:" which is probably the initial configuration, but I can't add a block. 
Screenshot 2024-05-12 at 20 45 54

The same in the Template mode. I found that in template mode, we have an additional bug with the inline iserter popup. Instead, after clicking on (+) template content, e.g., column block, the editor ends up blank, but no error in the console. This is not related to the PR and is very likely the same/similar error as we had with the post-locked popup.

mailpoet/lib/EmailEditor/Engine/Renderer/ContentRenderer/ContentRenderer.php
costasovo: 👍 I guess this might be related also to the error we observed with the latest Gutenberg version (https://mailpoet.atlassian.net/browse/MAILPOET-6052)

mikejolley:
@costasovo thanks for fixing the test. I had to put the default variation back in place to fix the UI. Upstream we should fix this or allow the default to be set that skips initial selection. For now its working though, and will work fine for default layouts and templates where we can set layout to 'constrained'.

costasovo:
@mikejolley Thanks for adding the variation select. I see the branch has some conflicts. Could you please rebase it?
~~Also, for some reason, I don't see an icon in the placeholder (see image). Is that expected?
Screenshot 2024-05-15 at 7 43 59~~

I see now it is a proper icon. It just looks weird because the other two variations are missing.

mikejolley:
@costasovo I've rebased and added a workaround for MAILPOET-6052. The issue was partially resolved here, but fixed completely by using the gutenberg prefixed function when present.

costasovo (approved):
@mikejolley Thank you for fixing it! I was curious where the side-effect causing that layout styles are added happens and it is here: https://github.com/WordPress/WordPress/blob/master/wp-includes/block-supports/layout.php#L524-L538

The styles are then stored in a store via `WP_Style_Engine::store_css_rule` but in Gutenberg plugin the class is WP_Style_Engine_Gutenberg and then also the store has the suffix so the styles generated by the side-effect are stored in a different place so the WP function doesn't see them.

FYI I created a follow-up task for the upstream change https://mailpoet.atlassian.net/browse/MAILPOET-6067

PR #5592, opened 2024-05-14
Invalid API endpoint exception improvement - MAILPOET-6022

triple0t:
## Description

Invalid API endpoint exception improvement

## Code review notes

_N/A_

## QA notes

Please check Jan's [note](https://github.com/mailpoet/mailpoet/pull/5592#issuecomment-2133383764) below
You can rename or delete one of the [JSON API endpoints](https://github.com/mailpoet/mailpoet/tree/605664228ee1bd337ac5024eec4526778a5ed7fa/mailpoet/lib/API/JSON/v1)

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6022](https://mailpoet.atlassian.net/browse/MAILPOET-6022)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6022]: https://mailpoet.atlassian.net/browse/MAILPOET-6022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/context/use-notices.jsx
+++ b/mailpoet/assets/js/src/context/use-notices.jsx
@@ -1,6 +1,6 @@
 import { useState, useCallback } from 'react';
 import { createInterpolateElement } from '@wordpress/element';
-import { sprintf, __ } from '@wordpress/i18n';
+import { __ } from '@wordpress/i18n';
 
 export const useNotices = () => {
   const [state, setState] = useState({
@@ -67,15 +67,12 @@ export const useNotices = () => {
             <p key={err.message}>
               {err.error === 'reinstall_plugin'
                 ? createInterpolateElement(
-                    sprintf(
-                      __(
-                        'The plugin has encountered an unexpected error. Please reload the page. If that does not help, re-install the MailPoet Plugin. See: %s for more information',
-                        'mailpoet',
-                      ),
-                      '<a> https://kb.mailpoet.com/article/258-re-installing-updating-the-plugin-via-ftp </a>',
+                    __(
+                      'The plugin has encountered an unexpected error. Please reload the page. If that does not help, <link>re-install the MailPoet Plugin.</link>',
+                      'mailpoet',
                     ),
                     {
-                      a: (
+                      link: (
                         <a
                           aria-label={err.error}
                           href="https://kb.mailpoet.com/article/258-re-installing-updating-the-plugin-via-ftp"
--- a/mailpoet/assets/js/src/context/use-notices.jsx
+++ b/mailpoet/assets/js/src/context/use-notices.jsx
@@ -75,13 +75,15 @@ export const useNotices = () => {
                       '<a> https://kb.mailpoet.com/article/258-re-installing-updating-the-plugin-via-ftp </a>',
                     ),
                     {
-                      // eslint-disable-next-line jsx-a11y/anchor-has-content, jsx-a11y/control-has-associated-label
                       a: (
                         <a
+                          aria-label={err.error}
                           href="https://kb.mailpoet.com/article/258-re-installing-updating-the-plugin-via-ftp"
                           target="_blank"
                           rel="noopener noreferrer"
-                        />
+                        >
+                          &nbsp;
+                        </a>
                       ),
                     },
                   )
--- a/mailpoet/assets/js/src/context/use-notices.jsx
+++ b/mailpoet/assets/js/src/context/use-notices.jsx
@@ -1,4 +1,6 @@
 import { useState, useCallback } from 'react';
+import { createInterpolateElement } from '@wordpress/element';
+import { sprintf, __ } from '@wordpress/i18n';
 
 export const useNotices = () => {
   const [state, setState] = useState({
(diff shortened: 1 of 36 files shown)


lysyjan (commented):
@triple0t Thank you for this change.
I tried to replicate this behavior briefly, but the message was dismissed after quite a short time, so the link was useless for me because I could not click or copy the link. Do you think we could find a way to make the link clickable? I am afraid that in this state, the change will not have the expected impact. What do you think about it?

Also, adding replication steps for QA would be great.

triple0t:

> I tried to replicate this behavior briefly, but the message was dismissed after quite a short time, so the link was useless for me because I could not click or copy the link.

Hmm, I'm not returning the message as an HTML because not all places on the Frontend accept such error messages.
There are probably a few I could find and update, but there are way more I'll miss (here is an [example](https://github.com/search?q=repo%3Amailpoet%2Fmailpoet%20.fail((response)%20%3D%3E&type=code)).
I decided on this approach to be backward compatible. 

We definitely don't want the message to be easily dismissed though.
Can you send me the link to the component you observed this on?

> Also, adding replication steps for QA would be great.

Thanks for the suggestion. 
I could replicate this error by intentionally throwing an error in the code.
Do you have an easy way to allow QA to test this?

Otherwise, I was planning on creating a .zip file with the modified code, which shows/throws the exception.

lysyjan:
> Hmm, I'm not returning the message as an HTML because not all places on the Frontend accept such error messages.
There are probably a few I could find and update, but there are way more I'll miss (here is an [example](https://github.com/search?q=repo%3Amailpoet%2Fmailpoet%20.fail((response)%20%3D%3E&type=code)).
I decided on this approach to be backward compatible.

I'm not sure about this approach because it is not user-friendly. Copying the URL from the error message offers more space for a user's mistake. What do you think @pavel-mailpoet about it? I'm asking you because you created this ticket.

> We definitely don't want the message to be easily dismissed though.
Can you send me the link to the component you observed this on?


@triple0t I tested this behavior by renaming one of the JSON API endpoints, for example, from `Newsletters.php` to `Newsletter.php`, and then I visited the newsletters listing page. So, I don't know which component it was.

pavel-mailpoet:
> I'm not sure about this approach because it is not user-friendly. Copying the URL from the error message offers more space for a user's mistake. What do you think @pavel-mailpoet about it? I'm asking you because you created this ticket.

If it is easy, we should definitely go for the user-friendly approach. That is what we are trying. Help users so that they don't need to contact us.

triple0t:
I've updated the PR. Kindly help take a look @lysyjan

lysyjan:
@triple0t When I tried to test the new error message, I saw a non-clickable link:
Screenshot 2024-05-29 at 11 27 50

Did I do something wrong? Also, the message was automatically dismissed after several seconds. Of course, I compiled the latest JS from your branch.

triple0t:
Updated @lysyjan 

Thanks for the feedback

mailpoet/assets/js/src/context/use-notices.jsx
```diff
@@ -75,13 +75,15 @@ export const useNotices = () => {
                       '<a> https://kb.mailpoet.com/article/258-re-installing-updating-the-plugin-via-ftp </a>',
                     ),
                     {
-                      // eslint-disable-next-line jsx-a11y/anchor-has-content, jsx-a11y/control-has-associated-label
```
lysyjan: Just a note that this comment didn't work because it should have been at the line below.

mailpoet/assets/js/src/context/use-notices.jsx
```diff
@@ -46,12 +48,69 @@ export const useNotices = () => {
...
+          content = response.errors.map((err) => (
+            <p key={err.message}>
+              {err.error === 'reinstall_plugin'
+                ? createInterpolateElement(
+                    sprintf(
+                      __(
+                        'The plugin has encountered an unexpected error. Please reload the page. If that does not help, re-install the MailPoet Plugin. See: %s for more information',
+                        'mailpoet',
+                      ),
+                      '<a> https://kb.mailpoet.com/article/258-re-installing-updating-the-plugin-via-ftp </a>',
```
lysyjan: - The text in the link element should be different from the ticket specification. Or is there a reason why you used this approach?
- Also, we usually use the <link> mark here, which can be a part of the translation string, and translators already know it. Maybe we should stay with this approach.
- Is there some reason for the redundant spaces around the text inside the tag?

lysyjan (commented):
@triple0t Thanks for the update.

Because this is very complex issue that we are trying to solve. I was testing it again and maybe more than is from a developer expected, but I found some incosistencies for some pages. For example, the Automations page has a different behavior:
Screenshot 2024-05-30 at 13 22 25

Error message on the Segments page is not rendered properly:
Screenshot 2024-05-30 at 13 23 46

The Subscribers and Lists pages end with an infinite loop of error messages, but here I'm not sure if we should create a new ticket for it or address it here.
Screenshot 2024-05-30 at 13 25 46

mailpoet/assets/js/src/context/use-notices.jsx
triple0t: This was fixed here https://github.com/mailpoet/mailpoet/pull/5592/commits/4a39e3b5ce1f27a931a31659367ed13e38f31069

mailpoet/assets/js/src/context/use-notices.jsx
triple0t: > The text in the link element should be different from the ticket specification. Or is there a reason why you used this approach?

Uh? Why should the text in the link element be different?

> Also, we usually use the mark here, which can be a part of the translation string, and translators already know it. Maybe we should stay with this approach.

I'm lost. Are you suggesting we use <mark> instead of <a> ?  We need the <a> for the interpolation.

> Is there some reason for the redundant spaces around the text inside the tag?

Not really. It can/will be removed.

triple0t:
Thanks for the in-depth review, @lysyjan 

I have addressed the issues you noted, except for the automation one.

Automation uses WP JSON directly and not MP REST endpoint. Configuring that to work with the use case in this PR will require some effort and a few refactors. I don't want to do something out of scope for this ticket, so I'll leave it as is.

mailpoet/assets/js/src/context/use-notices.jsx
lysyjan: > Uh? Why should the text in the link element be different?
> I'm lost. Are you suggesting we use instead of ? We need the for the interpolation.

@triple0t The ticket specification contains the following sentence:
`The plugin has encountered an unexpected error. Please reload the page. If that doesn't help [link]re-install the MailPoet Plugin.[/link]`,
but you used: `The plugin has encountered an unexpected error. Please reload the page. If that does not help, re-install the MailPoet Plugin. See: %s for more information`.
Displaying the target URL can make the text less readable. I don't think we use this approach anywhere.
Also, when we use string replacing via `sprintf` with the `%s` specifier, we should add a comment for translators. [Here](https://github.com/mailpoet/mailpoet/blob/c143da4ccfe266beb6091ce0ede742bad2ff639c/mailpoet/assets/js/src/automation/listing/components/menu/delete.tsx#L47-L51) is an example.

Nevertheless, If I add this message, I would try to follow the ticket specification exactly, and it could look like this:
```
createInterpolateElement(
  __(
    'The plugin has encountered an unexpected error. Please reload the page. If that does not help, <link>re-install the MailPoet Plugin.</link>,
    'mailpoet',
  ),
  {
    link: (
    <a
      href="https://kb.mailpoet.com/article/258-re-installing-updating-the-plugin-via-ftp"
      target="_blank"
      rel="noopener noreferrer"
    >
      &nbsp;
    </a>
    ),
  },
)
```

Maybe we can display the whole URL in the text, but in my opinion, it should be at the end of the sentence and not in the middle, and we would need to add a comment for translators.

lysyjan:
@triple0t Thanks for fixing the issues. I think we can create a ticket for better handling errors on the Automation page.

Please, take a look at the last comment related to the message format.

mailpoet/assets/js/src/context/use-notices.jsx
triple0t: Noted. 
Fixed here https://github.com/mailpoet/mailpoet/pull/5592/commits/656a66299bd1b2c85449b60b394983bdfab88455
Thanks

lysyjan (approved):

Aschepikov (approved):
GJ!

PR #5598, opened 2024-05-20
Release 4.50.1

lysyjan:


lysyjan:
I'm closing it because we want to use the translations build command fix.

PR #5601, opened 2024-05-21
Email editor - fix blank screen in template mode [MAILPOET-6065]

costasovo:
## Description

This PR fixes the issue with the strange WSOD happening in the editor's template mode. Plus a few tiny fixes.

## Code review notes

I found that it was related to the fact that we hide the editor when the template is not loaded. The problem with that is that in template mode, the main document is a template, and there is no template associated with it. So then, when something caused the rerendering of the Layout component, it turned blank. Such a re-render could be caused, for example, by the inserter popup, which is rendered out of the frame. I assume the issue is also related to the problem with PostLockedModal, which we deactivated. 

#### Replication

1. Open the email editor and switch to template mode
2. Insert columns block
3. Click the appender (+) in column block in the editor and observe issue

## QA notes

I think it doesn't require QA and can be tested and merged by devs.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6065]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6065]: https://mailpoet.atlassian.net/browse/MAILPOET-6065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/components/block-editor/editor.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/block-editor/editor.tsx
@@ -75,7 +75,7 @@ export function InnerEditor({
     [settings, onNavigateToEntityRecord, onNavigateToPreviousEntityRecord],
   );
 
-  if (!post) {
+  if (!post || (currentPost.postType !== 'wp_template' && !template)) {
     return null;
   }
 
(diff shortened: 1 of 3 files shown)


mikejolley (dismissed):
This fixed the issue, good catch. I didn't spot any other issues besides styling issues but this is due to requiring a rebase.

PR #5608, opened 2024-05-24
Add a knowledge base link to mysql server has gone away - MAILPOET-6020

triple0t:
## Description

This error can be thrown in multiple places. I  added an exception handler for MailPoet files.
We can't do anything about the error when it's thrown from within WordPress core or from other plugins or themes.

I tested the error on a non-English WP site, and it seems to work, but please reconfirm.

## Code review notes

`handleExceptionDuringQuery` is marked internal, but it's the method we need.
The alternative would be to wrap all the query methods on the `MailPoetVendor\Doctrine\DBAL\Connection` class.

## QA notes

I think the easiest way to test this PR is to trigger the error.

Get these values and save them somewhere
(If you get an error, login to the DB with root credentials)

```
SHOW VARIABLES LIKE 'net_buffer_length';
SHOW VARIABLES LIKE 'max_allowed_packet';
SHOW VARIABLES LIKE 'wait_timeout';
```

Set this value
```
SET GLOBAL wait_timeout = 1;
SET GLOBAL net_buffer_length = 1200;
SET GLOBAL max_allowed_packet = 3000;
```

Please experiment with the `max_allowed_packet` value. Examples can be 3000, 5000, etc

Visit `/wp-admin/admin.php?page=mailpoet-newsletters#/new` and try to create a new email
You should get the error
Also, check your error log file

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6020](https://mailpoet.atlassian.net/browse/MAILPOET-6020)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6020]: https://mailpoet.atlassian.net/browse/MAILPOET-6020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Cron/CronTrigger.php
+++ b/mailpoet/lib/Cron/CronTrigger.php
@@ -4,6 +4,7 @@ namespace MailPoet\Cron;
 
 use MailPoet\Cron\Triggers\WordPress;
 use MailPoet\Settings\SettingsController;
+use MailPoet\Util\Helpers;
 
 class CronTrigger {
   const METHOD_LINUX_CRON = 'Linux Cron';
@@ -51,6 +52,7 @@ class CronTrigger {
       return false;
     } catch (\Exception $e) {
       // cron exceptions should not prevent the rest of the site from loading
+      Helpers::mySqlGoneAwayExceptionHandler($e);
     }
   }
 
--- a/mailpoet/lib/Cron/Daemon.php
+++ b/mailpoet/lib/Cron/Daemon.php
@@ -4,6 +4,7 @@ namespace MailPoet\Cron;
 
 use MailPoet\Cron\Workers\WorkersFactory;
 use MailPoet\Logging\LoggerFactory;
+use MailPoet\Util\Helpers;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class Daemon {
@@ -61,6 +62,8 @@ class Daemon {
           $worker->process($this->timer); // BC for workers not implementing CronWorkerInterface
         }
       } catch (\Exception $e) {
+        Helpers::mySqlGoneAwayExceptionHandler($e);
+
         $workerClassNameParts = explode('\\', get_class($worker));
         $workerName = end($workerClassNameParts);
         $errors[] = [
--- a/mailpoet/lib/Cron/ActionScheduler/RemoteExecutorHandler.php
+++ b/mailpoet/lib/Cron/ActionScheduler/RemoteExecutorHandler.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Cron\ActionScheduler;
 
+use MailPoet\Util\Helpers;
 use MailPoet\WP\Functions as WPFunctions;
 
 class RemoteExecutorHandler {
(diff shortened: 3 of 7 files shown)


mailpoet/lib/Cron/CronTrigger.php
```diff
@@ -51,6 +52,7 @@ public function init(string $environment = '') {
       return false;
     } catch (\Exception $e) {
       // cron exceptions should not prevent the rest of the site from loading
+      Helpers::mySqlGoneAwayExceptionHandler($e);
```
costasovo, line 55: Just want to make sure I understand. Here, we intentionally ignore the returned message and call the helper just to log the error via error_log. Correct?

mailpoet/lib/Cron/Daemon.php
```diff
@@ -61,6 +62,8 @@ public function run($settingsDaemonData) {
           $worker->process($this->timer); // BC for workers not implementing CronWorkerInterface
         }
       } catch (\Exception $e) {
+        Helpers::mySqlGoneAwayExceptionHandler($e);
```
costasovo, line 65: Just want to make sure I understand. Here, we intentionally ignore the returned message and call the helper just to log the error via error_log. Correct?

costasovo (commented):
@triple0t Codewise looks good! I have just a question to verify my understanding.
I tried to replicate the mysql has gone away error using your instructions but I was not able to replicate it. 
Perhaps because I'm getting different errors:  (Got timeout reading communication packets) or (Got a packet bigger than 'max_allowed_packet' bytes). I'm using  10.5.8-MariaDB.
But I guess I could move it forward and let QA to test it.

mailpoet/lib/Cron/CronTrigger.php
triple0t: Yup, that's correct. @costasovo 
I have no idea why we aren't logging the error with our LoggerFactory class.
Do you think we should log it or leave it as it currently is?

mailpoet/lib/Cron/Daemon.php
triple0t: Yup, that's correct.

mailpoet/lib/Cron/CronTrigger.php
triple0t: Oh, if we have a Database error, we can't write to the Database, so it probably won't work.

triple0t:

> Perhaps because I'm getting different errors: (Got timeout reading communication packets) or (Got a packet bigger than 'max_allowed_packet' bytes). I'm using 10.5.8-MariaDB.
But I guess I could move it forward and let QA to test it.

Yeah, this happens.
You can experiment with the `max_allowed_packet` value. Examples can be 3000, 5000, etc

costasovo (approved):
@triple0t Looks good!
As for testing, I'm still getting different errors (different from "mysql server has gone away")from the DB so the specific handler is not triggered. I'm sending it to QA hoping they will have more luck with the replication.

Aschepikov (approved):
GJ!

PR #5612, opened 2024-05-24
Refactor and relogic acceptance tests to be capable of testing blockbased and non block themes

veljkho:
## Description

In this PR, I'd like to refactor and change logic of some tests to fulfil requirement to have acceptance tests testing both blockbased and non-block themes. This can be a go for testing more themes, but for now let's stick to these two types.

We will have additional job with all those frontend tests that are actually related to different theme use. We don't want to waste time and money on testing tests that are irrelevant of the theme in use. In that favor, I added additional group `frontend` to catch only those important tests for theme testing.

## Code review notes

Check the additional CI job `acceptance_tests_blockbased_theme` that will run over 100 tests on a block-based TT4 theme.

I had to rewrite some tests because the old tests were not compatible with a block-based theme, some of the tests.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5787]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5787]: https://mailpoet.atlassian.net/browse/MAILPOET-5787?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/acceptance/Forms/SubscribeToMultipleListsCest.php
+++ b/mailpoet/tests/acceptance/Forms/SubscribeToMultipleListsCest.php
@@ -3,6 +3,8 @@
 namespace MailPoet\Test\Acceptance;
 
 use Codeception\Util\Locator;
+use MailPoet\DI\ContainerWrapper;
+use MailPoet\Form\FormMessageController;
 use MailPoet\Subscription\Captcha\CaptchaConstants;
 use MailPoet\Test\DataFactories\Form;
 use MailPoet\Test\DataFactories\Segment;
@@ -24,6 +26,7 @@ class SubscribeToMultipleListsCest {
     $formName = 'Multiple Lists Form';
     $formFactory = new Form();
     $form = $formFactory->withName($formName)->withSegments([$segment1, $segment2, $segment3])->create();
+    $formId = $form->getId();
 
     $settings = new Settings();
     $settings
@@ -34,9 +37,30 @@ class SubscribeToMultipleListsCest {
 
     $formFactory->withDefaultSuccessMessage();
 
-    //Add this form to a widget
-    $i->createFormAndSubscribe();
-    //Subscribe via that form
+    $i->havePostInDatabase([
+      'post_author' => 1,
+      'post_type' => 'page',
+      'post_name' => 'form-test',
+      'post_title' => 'Form Test',
+      'post_content' => '
+        Regular form:
+          [mailpoet_form id="' . $formId . '"]
+        Iframe form:
+          <iframe class="mailpoet_form_iframe" id="mailpoet_form_iframe" tabindex="0" src="http://test.local?mailpoet_form_iframe=1" width="100%" height="100%" frameborder="0" marginwidth="0" marginheight="0" scrolling="no"></iframe>
+      ',
+      'post_status' => 'publish',
+    ]);
+  
+    $i->wantTo('Subscribe to the form with multiple lists attached');
+    $i->amOnPage('/form-test');
+    $i->waitForElement('[data-automation-id="form_email"]');
+    $i->fillField('[data-automation-id="form_email"]', 'subscriber@example.com');
+    $i->click('[data-automation-id="subscribe-submit-button"]');
+    $messageController = ContainerWrapper::getInstance()->get(FormMessageController::class);
+    $i->waitForText($messageController->getDefaultSuccessMessage(), 30, '.mailpoet_validate_success');
+    $i->seeNoJSErrors();
+
+    $i->wantTo('Confirm subscription to subscribed form');
     $i->amOnMailboxAppPage();
     $i->click(Locator::contains('span.subject', 'Subscribe to multiple test subject'));
     $i->switchToIframe('#preview-html');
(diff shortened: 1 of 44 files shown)


mailpoet/tests/acceptance/Forms/SubscribeToMultipleListsCest.php
```diff
@@ -32,7 +35,7 @@ public function subscribeToMultipleLists(\AcceptanceTester $i) {
     $formFactory->withDefaultSuccessMessage();
 
     //Add this form to a widget
```
brezocordero: Is this comment still relevant?

mailpoet/tests/acceptance/Settings/BuiltInCaptchaSubscriptionCest.php
```diff
@@ -26,17 +32,29 @@ public function _before(\AcceptanceTester $i) {
...
     $subscriberFactory = new Subscriber();
     $subscriberFactory->withEmail($this->subscriberEmail)->withCountConfirmations(1)->create();
 
-    $i->cli(['widget', 'add', 'mailpoet_form', 'sidebar-1', '2', "--form={$form->getId()}", '--title="Subscribe to Our Newsletter"']);
+    $i->havePostInDatabase([
+      'post_author' => 1,
+      'post_type' => 'page',
+      'post_name' => 'form-test',
+      'post_title' => 'Form Test',
+      'post_content' => '
```
brezocordero, line 45: The commit that added this code says '[Refactor acceptance tests](https://github.com/mailpoet/mailpoet/pull/5612/commits/d7d07385694f386024b493dfe9f8948ba6441049)'

Would you mind explaining why we are changing the form?

mailpoet/tests/docker/codeception/docker-entrypoint.sh
```diff
@@ -169,6 +169,9 @@ wp theme install twentytwentyone --activate
 if [[ $MULTISITE == "1" ]]; then
   wp theme install twentytwentyone --url=$HTTP_HOST/$WP_TEST_MULTISITE_SLUG --activate
```
brezocordero, line 170: I noticed that you tried upgrading to `twentytwentythree`. Why did you revert the change? Was there any blocker in upgrading?

brezocordero (dismissed):
@veljkho looks good, just minor questions on the comments.
Note: I am assuming you have tested the jobs that don't run on on `build_and_test` like `acceptance_oldest`

mailpoet/tests/acceptance/Forms/SubscribeToMultipleListsCest.php
veljkho: Yes, you are right; the comment should be obsolete. I will update it.

veljkho: I will update the test to reflect the scenario as the first attempt was not testing multi list form.

mailpoet/tests/acceptance/Settings/BuiltInCaptchaSubscriptionCest.php
veljkho: Maybe I was wrong with the terminology, but I had to update `createFormAndSubscribe` helper to support testing blockbased theme. There is no widgets in blockbased themes so modifying the helper method to create a form via normal post rather then widget (I believe the widget was just randomly chosen and not on purpose) solved issue with multiple themes blockbased and non-blockbased.

mailpoet/tests/docker/codeception/docker-entrypoint.sh
veljkho: I realized that TT3 is also block-based theme. We have TT4 as a block theme so testing on TT1 will be non block theme. We can stay with TT1 and extend with TT4 and other themes.

veljkho:
> Note: I am assuming you have tested the jobs that don't run on on build_and_test like acceptance_oldest

I haven't tested @brezocordero. How do you recommend testing that? Did you mean locally?
I don't think there will be any conflicts with it.

mailpoet/tests/docker/codeception/docker-entrypoint.sh
brezocordero: Thanks for the explanation!

mailpoet/tests/acceptance/Settings/BuiltInCaptchaSubscriptionCest.php
brezocordero: Thanks, that makes sense.

mailpoet/tests/acceptance/Forms/SubscribeToMultipleListsCest.php
brezocordero: Thanks for making the change.

brezocordero:
>  How do you recommend testing that? 

You could temporarily modify the `.circleci/config.yml` file and add the jobs to the workflow `build_and_test`. Then, push it to this branch, or a separate branch, and remove the commit after.

> I don't think there will be any conflicts with it.

If you are confident about it, I'm ok with it. Feel free to merge.

brezocordero (dismissed):
LGTM!

veljkho:
Tested and it worked. Now pushed with reverted commit.

brezocordero:
@veljkho  thanks so much for testing the nightly jobs.
There is a failing acceptance test. I have rerun it in case it was flaky but it still fails. 
It might not be related but would you mind taking a look?

veljkho:
Yes, @brezocordero , this is not related to this PR and it started failing just recently on trunk. It is not a flaky test, I commented on Slack -dev channel.

veljkho:
@brezocordero thanks for reviewing! I think the PR is now ready for merging.

PR #5614, opened 2024-05-27
Show notice when WP-cron is disabled [MAILPOET-6027]

alex-mpoet:
## Description

The notice is shown only when the "WordPress built-in cron" option for Newsletter task scheduler is selected in Settings > Advanced.

## Code review notes

_N/A_

## QA notes

Testing instructions:
1. Select the "WordPress built-in cron option" for Newsletter task scheduler in Settings > Advanced.
2. Disable WP cron by setting the `DISABLE_WP_CRON` constant to `true`, e.g. in theme functions.php.
3. Observe the error message. It can be dismissed.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6027]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6027]: https://mailpoet.atlassian.net/browse/MAILPOET-6027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Util/Notices/DisabledWPCronNotice.php
+++ b/mailpoet/lib/Util/Notices/DisabledWPCronNotice.php
@@ -1,0 +1,62 @@
+<?php // phpcs:ignore SlevomatCodingStandard.TypeHints.DeclareStrictTypes.DeclareStrictTypesMissing
+
+namespace MailPoet\Util\Notices;
+
+use MailPoet\Cron\CronTrigger;
+use MailPoet\Settings\SettingsController;
+use MailPoet\Util\Helpers;
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class DisabledWPCronNotice {
+
+  const DISMISS_NOTICE_TIMEOUT_SECONDS = YEAR_IN_SECONDS;
+  const OPTION_NAME = 'dismissed-wp-cron-disabled-notice';
+
+  /** @var WPFunctions */
+  private $wp;
+
+  /** @var SettingsController */
+  private $settings;
+
+  public function __construct(
+    WPFunctions $wp,
+    SettingsController $settings
+  ) {
+    $this->wp = $wp;
+    $this->settings = $settings;
+  }
+
+  public function init($shouldDisplay) {
+    if (!$shouldDisplay) {
+      return null;
+    }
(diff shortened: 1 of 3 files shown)


JanJakes (approved):
@wxa Implementation-wise, LGTM, so I'm approving this.

But I have one question about the ticket itself, rather about the specs, so we may loop in the reporter, @pavel-mailpoet, as well.

I think it is not uncommon for some hosts to somehow disable WP cron and execute the cron tasks/Action Scheduler using dedicated workers. Do we know this won't result in false-positive displays of this notice? I'm not sure how, technically, the hosts might do it, neither I know which hosts have such a setup, but I'm wondering if that could be an issue.

pavel-mailpoet:
> Do we know this won't result in false-positive displays of this notice?

Good question, and the short answer is: I don't know. I saw some tickets where this was an issue. But I am not sure if this could cause even more harm. The notice is dismissible so people can cancel it, but I am not sure how to proceed with this.

JanJakes:
@pavel-mailpoet Just throwing some ideas:

1. We could release it and see if we get any reports...
2. We could try to investigate how the hosts do this.
3. We could display the notice only if last cron run was some time ago (e.g., a day or a couple of hours).

What do you think? Would option 3 be easy to do? @alex-mpoet

pavel-mailpoet:
@JanJakes that sounds good to me; as you are the DRI for this project, I will leave it to you to decide which way to go. Investigation sounds good, but the other options are as well.

alex-mpoet:
@JanJakes 
> Would option 3 be easy to do? 

It's easy to do, but I'm not sure the message will cover all cases then. When do users usually encounter the disabled cron issue? During initial setup when cron was not run yet? Or when cron suddenly becomes disabled and the user is unaware of that? I think there are cases when a website was neglected for a while, and cron did not run recently, but the message could help when the user comes back. It can be dismissed in the worst case.

JanJakes:
@wxa I discussed this with David, and we decided to merge this as-is and address issues with false positives if they arise.

JanJakes (approved):

Aschepikov (approved):
GJ!

PR #5627, opened 2024-05-31
Update composer dependencies [MAILPOET-6094]

lysyjan:
## Description

_N/A_

## Code review notes

I updated all possible dependencies, including some of the developments. I skipped all the Codeception dependencies and will create a separate ticket for that. I didn't update some packages because I was limited by the supported PHP versions.

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/872

## Linked tickets

[MAILPOET-6094]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6094]: https://mailpoet.atlassian.net/browse/MAILPOET-6094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/composer.json
+++ b/mailpoet/composer.json
@@ -29,7 +29,7 @@
     "phpunit/phpunit": "9.6.13",
     "totten/lurkerlite": "^1.3",
     "vlucas/phpdotenv": "v5.6.0",
-    "wp-cli/wp-cli-bundle": "2.8.1"
+    "wp-cli/wp-cli-bundle": "2.10.0"
   },
   "autoload": {
     "classmap": [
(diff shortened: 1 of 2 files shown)


costasovo (approved):
@lysyjan Good job!

Note: I see that [Woo is using Action Scheduler 3.7.4](https://github.com/woocommerce/woocommerce/blob/b4efa72b07cd3ad1947b550215104d60569e4fc5/plugins/woocommerce/composer.json#L29) that means we will distribute the newest Action Scheduler so the AS from MailPoet will be used and for some time we may get more pings from users who may report errors that are not related to MP just because the happened withing AS run (they will see MailPoet in the path). But I guess that's ok I assume Woo will update soon.

PR #5629, opened 2024-05-31
Improve Sending Queue Worker [MAILPOET-5881]

costasovo:
## Description

This PR adds a couple of fixes and improvements for issues we identified during debugging sending issues after we released the Doctrine refactor.

## Code review notes

I followed the ideas from the ticket. Just a few notes:

1) I kept one call of the method for ending the sending also inside the batch processing loop because we have the execution limits check there, and in case the check throws an error, we would need to wait for another cron run to pick the task again. So it is more like an optimization.

2) I wanted to cover all the states in the method for handling the end of sending, but I was not sure what to do for the unexpected states (`PAUSED`, `INVALID`, `SCHEDULED`). These can, IMO, happen only because of some parallel run, or maybe the user clicking pause at the very exact moment. Paused and invalid tasks shouldn't block sending as these are not picked again. The schedule should be picked up again and hopefully processed. An alternative option I was thinking about was checking if there are unprocessed subscribers and, if not, marking a task as completed. Any thoughts/suggestions about this?

## QA notes

Please test sending thoroughly. 
Try sending different types of emails simultaneously to small and large lists.
Test also cases like welcome email and switching the related subscriber to unsubscribed before the email is trigerred.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5881]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5881]: https://mailpoet.atlassian.net/browse/MAILPOET-5881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/tests/integration/Cron/Workers/SendingQueue/SendingQueueTest.php
+++ b/mailpoet/tests/integration/Cron/Workers/SendingQueue/SendingQueueTest.php
@@ -840,10 +840,45 @@ class SendingQueueTest extends \MailPoetTest {
     verify($sendingQueue->getCountToProcess())->equals(0);
   }
 
-  public function testItCompletesEverythingProperlyWhenThereIsNoOneToSendTo() {
+  public function testItCompletesEverythingProperlyWhenThereIsNoSubscribedSubscriberToSendTo() {
     // Set the only scheduled task subscriber to unsubscribed
     $this->subscriber->setStatus(SubscriberEntity::STATUS_UNSUBSCRIBED);
     $this->entityManager->flush();
+
+    $sendingQueueWorker = $this->getSendingQueueWorker(
+      $this->construct(
+        MailerTask::class,
+        [$this->diContainer->get(MailerFactory::class)],
+        [
+          'send' => Expected::never(),
+        ]
+      )
+    );
+    $sendingQueueWorker->process();
+
+    // queue status is set to completed
+    $sendingQueue = $this->sendingQueuesRepository->findOneById($this->sendingQueue->getId());
+    $this->assertInstanceOf(SendingQueueEntity::class, $sendingQueue);
+    $scheduledTask = $this->scheduledTasksRepository->findOneBySendingQueue($sendingQueue);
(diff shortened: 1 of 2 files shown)


lysyjan (commented):
@costasovo A lot of improvements! Great job!
I noticed one issue, which could cause failing nightly tests.

mailpoet/tests/integration/Cron/Workers/SendingQueue/SendingQueueTest.php
```diff
@@ -1600,7 +1632,7 @@ private function getSendingQueueWorker($mailerMock = null, $authorizedEmailContr
       $this->sendingQueuesRepository,
       $this->entityManager,
       $this->statisticsNewslettersRepository,
-      $authorizedEmailControllerMock ?? $this->authorizedEmailsController
+      $authorizedEmailControllerMock ?? $this->authorizedEmailsController,
```
lysyjan, line 1635: Why did you add a comma here? I think it is not a big issue, but it can cause failing tests on lower PHP versions.

costasovo: @lysyjan Aha. Nice catch! It was accidental. I added one more argument, but then I decided it was not a good idea, so I deleted it and I forgot the comma. Trailing comma should be [fine for PHP7.3+](https://php.watch/versions/7.3/func-call-trailing-comma) and we use PHP7.4 as the oldest one so this should be fine even for nightly tests. So I guess it is not a big deal, right?

lysyjan: Ah, right! I thought it was from 8.0, but I got it mixed up with [trailing comma in function declarations](https://php.watch/versions/8.0/trailing-comma-parameter-use-list).

lysyjan (approved):

Aschepikov (approved):
GJ!

PR #5631, opened 2024-06-03
Show cron ping error globally [MAILPOET-6029]

alex-mpoet:
## Description

The specs weren't very detailed, so I did it using my own justification. The check is performed asynchronously after a page load to not cause delays. The results aren't cached. It only works for the "WordPress" cron trigger method. 

## Code review notes

_N/A_

## QA notes

Testing instructions:
1. Break cron, e.g. by replacing `self::PING_SUCCESS_RESPONSE` with `"test error"` [here](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Cron/DaemonHttpRunner.php#L54).
2. Select "Visitor to your website" in Settings > Advanced.
3. Clear the `mailpoet_cron_ping_last_success_at` key in your browser's localStorage.
4. Observe the error message about task scheduler on all menu pages.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6029]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6029]: https://mailpoet.atlassian.net/browse/MAILPOET-6029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo (commented):
@alex-mpoet Thank you for adding the notice! I tested it, and it works as expected.

We do a request on every page load, and the endpoint also makes an additional loopback request. 

We basically test if a loopback requests work. This is usually caused by server configuration. The condition causing the issue doesn't change dynamically, so I think we could add some caching to prevent these requests from being made on every page. What about remembering the successful response (e.g. in the browser's session storage) so when the response is OK we don't do more requests for some time. But in case of an error we keep trying on every page, so that the notice is gone immediately after admin fixes the issue.

alex-mpoet:
@costasovo 
> What about remembering the successful response

Sure, that makes sense. Implemented.

costasovo (commented):
@alex-mpoet Thank you for adding the cache! Works for me as expected.

I noticed that the GlobalNotices is written in .jsx file. We should add new JS code as Typescript. I missed that in the first review. Sorry! Could you please update it to .tsx? Or is there a reason why we need to use .jsx?

alex-mpoet:
@costasovo 

> Could you please update it to .tsx?

Good catch! Updated. 

> Or is there a reason why we need to use .jsx?

No reason, that was purely accidental.

costasovo (dismissed):
Good job! Thank you!

Aschepikov:
Hi @costasovo there are some conflicts can you resolve them pls

alex-mpoet:
@Aschepikov Conflicts resolved.

Aschepikov (approved):

PR #5635, opened 2024-06-04
Show "Bridge is not accessible" notice globally [MAILPOET-6030]

alex-mpoet:
## Description

The PR is very similar to https://github.com/mailpoet/mailpoet/pull/5631, but with a different endpoint and text.

## Code review notes

_N/A_

## QA notes

Testing instructions:
1. Break MSS connection, e.g. by replacing `self::BRIDGE_URL` with `'https://some-non-existent-domain-for-testing-mailpoet.com'` [here](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Services/Bridge.php#L107).
2. Select MailPoet Sending Service as the sending method in "Settings > Send with...".
3. Clear the `mailpoet_services_pingbridge_last_success_at` key in your browser's localStorage.
4. Observe the error message about MSS connection on all menu pages.
5. Please also retest the https://github.com/mailpoet/mailpoet/pull/5631 PR, it was refactored in this one.

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/5631

## Linked tickets

[MAILPOET-6030]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6030]: https://mailpoet.atlassian.net/browse/MAILPOET-6030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/notices/bridge-ping-error-notice.tsx
+++ b/mailpoet/assets/js/src/notices/bridge-ping-error-notice.tsx
@@ -1,0 +1,85 @@
+import { useState, useEffect } from 'react';
+import { extractPageNameFromUrl } from 'common/functions';
+import { Notice } from 'notices/notice';
+import { MailPoet } from 'mailpoet';
+import ReactStringReplace from 'react-string-replace';
+
+const LOCAL_STORAGE_KEY = 'mailpoet_mss_ping_last_success_at';
+const CACHE_TIME = 60 * 60 * 1000; // 1 hour
+
+const cacheSuccessfulResult = () => {
+  window.localStorage.setItem(LOCAL_STORAGE_KEY, Date.now().toString());
+};
+
+const isSuccessfulResultCached = () => {
+  const lastSuccessAt = window.localStorage.getItem(LOCAL_STORAGE_KEY);
+  return (
+    lastSuccessAt && Date.now() - parseInt(lastSuccessAt, 10) <= CACHE_TIME
+  );
+};
+
+function BridgePingErrorNotice() {
+  const [bridgeError, setBridgeError] = useState<string | null>(null);
+
+  useEffect(() => {
+    // Show only when sending method is MailPoet
+    if (MailPoet.mtaMethod !== 'MailPoet') return;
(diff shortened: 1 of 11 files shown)


mailpoet/assets/js/src/notices/bridge-ping-error-notice.tsx
```diff
@@ -0,0 +1,85 @@
...
+        const errObject = errorResponse?.errors?.[0];
+        if (errObject.error && errObject.message) {
+          // Display MSS error if server responded with error message
+          setBridgeError(errObject.message);
+        } else {
+          // Do not display the notice if there is no server response
+          setBridgeError(null);
+        }
+      });
+  }, []);
```
costasovo: This piece of code, as it is pretty similar to the code for the cron ping, is a nice candidate for a custom hook. Something like `const isError  = useServiceCheck('pingBridge')`. It could encapsulate also the caching code. I consider the change nice to have.

costasovo (dismissed):
@alex-mpoet Great job! I tested it ,and it works as expected.

I have one suggestion for an improvement (see the other comment), but it is nice to have. I approve the PR and in case you find the idea valuable please feel free to implement it or you can pass the PR to QA.

mailpoet/assets/js/src/notices/bridge-ping-error-notice.tsx
alex-mpoet: I agree that this code is very similar to the cron ping code and can be extracted to a separate hook, but I don't see any more similar notices coming in the future, so this hook will likely be used only in two places. Seems a bit like premature optimization. I can still do it, though, if you insist.

alex-mpoet: @costasovo I've decided to incorporate your suggestion and extracted the common logic into a custom hook. Please check.

costasovo (dismissed):
@alex-mpoet Thank you for adding the hook.

Aschepikov (approved):
GJ!

PR #5637, opened 2024-06-04
Add a limit to timeout in throttling function

brezocordero:
## Description

A customer reported the following error:
```
MailPoet\Subscribers\SubscriberIPsRepository::findOneByIPAndCreatedAtAfterTimeInSeconds(): Argument #2 ($seconds) must be of type int, float given, called in /home/site/public_html/wp-content/plugins/mailpoet/Subscription/Throttling.php on line 41
```
This can happen in the [throttling function](https://github.com/mailpoet/mailpoet/blob/db370dc96295a023457313106d6cace85b298f16/mailpoet/lib/Subscription/Throttling.php#L37) if `pow` returns a float (eg: `$subscriptionsCount` > 63).

To avoid that we can limit the value of the timeout to `$subscriptionLimitWindow`

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6103]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6103]: https://mailpoet.atlassian.net/browse/MAILPOET-6103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscription/Throttling.php
+++ b/mailpoet/lib/Subscription/Throttling.php
@@ -35,6 +35,8 @@ class Throttling {
         $subscriptionCount = $this->subscriberIPsRepository->getCountByIPAndCreatedAtAfterTimeInSeconds($subscriberIp, $subscriptionLimitWindow);
         if ($subscriptionCount > 0) {
           $timeout = $subscriptionLimitBase * pow(2, $subscriptionCount - 1);
+          // Cap timeout and avoid float numbers
+          $timeout = min($timeout, $subscriptionLimitWindow);
           $existingUser = $this->subscriberIPsRepository->findOneByIPAndCreatedAtAfterTimeInSeconds($subscriberIp, $timeout);
           if (!empty($existingUser)) {
             return $timeout;
--- a/mailpoet/tests/integration/Subscription/ThrottlingTest.php
+++ b/mailpoet/tests/integration/Subscription/ThrottlingTest.php
@@ -31,6 +31,16 @@ class ThrottlingTest extends \MailPoetTest {
     verify($this->throttling->throttle())->equals(MINUTE_IN_SECONDS * pow(2, 10));
   }
 
+  public function testItDoesNotThrowErrorForBigNumbersAndCapsTimeout() {
+    $_SERVER['REMOTE_ADDR'] = '127.0.0.1';
+    verify($this->throttling->throttle())->equals(false);
+    verify($this->throttling->throttle())->equals(60);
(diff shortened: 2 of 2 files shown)


mailpoet/lib/Subscription/Throttling.php
```diff
@@ -34,7 +34,9 @@ public function throttle() {
       if (!empty($subscriberIp)) {
         $subscriptionCount = $this->subscriberIPsRepository->getCountByIPAndCreatedAtAfterTimeInSeconds($subscriberIp, $subscriptionLimitWindow);
         if ($subscriptionCount > 0) {
-          $timeout = $subscriptionLimitBase * pow(2, $subscriptionCount - 1);
+          $timeout = $subscriptionLimitBase * (pow(2, $subscriptionCount - 1));
+          // Cap timeout and avoid float numbers that can happen if $subscriptionCount > 63
```
alex-mpoet: I wouldn't mention the number of subscriptions here. In my tests it was different, maybe this depends on a system.

mailpoet/lib/Subscription/Throttling.php
```diff
@@ -34,7 +34,9 @@ public function throttle() {
       if (!empty($subscriberIp)) {
         $subscriptionCount = $this->subscriberIPsRepository->getCountByIPAndCreatedAtAfterTimeInSeconds($subscriberIp, $subscriptionLimitWindow);
         if ($subscriptionCount > 0) {
-          $timeout = $subscriptionLimitBase * pow(2, $subscriptionCount - 1);
+          $timeout = $subscriptionLimitBase * (pow(2, $subscriptionCount - 1));
```
alex-mpoet: Are extra brackets necessary here?

mailpoet/lib/Subscription/Throttling.php
brezocordero: > In my tests it was different, maybe this depends on a system.

True, it depends on `PHP_INT_MAX`. Removed

mailpoet/lib/Subscription/Throttling.php
brezocordero: > Are extra brackets necessary here?

No, it is a leftover from a local change to test converting with `intval.`
Thanks for noticing, fixed.

alex-mpoet (approved):
Looks good! Thank you!

PR #5641, opened 2024-06-06
Update prefixed composer dependecies [MAILPOET-6101]

lysyjan:
## Description

This PR updates dependencies prefixed and packed as a part of the plugin.

## Code review notes

- I tried to use the latest possible dependencies, but sometimes, not possible due to conflicts.
- I also increased the PHP version because it was locked on 7.2, but we now support 7.4.

## QA notes

Please test all core functionality across both plugins.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/874

## Linked tickets

[MAILPOET-6101]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6101]: https://mailpoet.atlassian.net/browse/MAILPOET-6101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Doctrine/SerializableConnection.php
```diff
@@ -47,16 +47,4 @@ public function rollBack() {
       throw $e;
     }
   }
-
-  public function handleExceptionDuringQuery(Throwable $e, string $sql, array $params = [], array $types = []): void {
```
alex-mpoet, line 51: Why was this method removed? Does the updated package version handle the "MySQL Server has gone away" error now?

mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
```diff
@@ -118,7 +118,7 @@ public function insertMultiple(
       }, $columns);
 
       foreach ($item as $columnKey => $column) {
-        $parameters[$paramNames[$columnKey]] = $column;
+        $parameters[substr($paramNames[$columnKey], 1)] = $column;
```
alex-mpoet, line 122: Are you removing the leading colon here in the loop? It would be great to add a comment about it.

mailpoet/prefixer/fix-doctrine.php
```diff
@@ -25,23 +25,6 @@
 $php7CachedItem = str_replace('final class CacheItem', 'final class TypedCacheItem', $php7CachedItem);
 file_put_contents(__DIR__ . "/../vendor-prefixed/doctrine/cache/lib/Doctrine/Common/Cache/Psr6/TypedCacheItem.php", $php7CachedItem);
 
-// Replace PHP8 syntax in ReflectionReadonlyProperty.
```
alex-mpoet: Why is this code not necessary anymore?

mailpoet/lib/Segments/DynamicSegments/Filters/MailPoetCustomFields.php
```diff
@@ -88,30 +88,30 @@ private function applyForDate(QueryBuilder $queryBuilder, DynamicSegmentFilterEn
   }
 
   private function applyForDateMonth(QueryBuilder $queryBuilder, string $valueParam): QueryBuilder {
-    $queryBuilder->andWhere("month(subscribers_custom_field.value) = month($valueParam)");
+    $queryBuilder->andWhere("month(subscribers_custom_field.value) = month(:$valueParam)");
```
alex-mpoet, line 91: I see that in other places you're removing colons, but here and below you're adding them instead, I'd like to get some context about this.

alex-mpoet (commented):
I have a few notes, please see below.

mailpoet/lib/Doctrine/SerializableConnection.php
lysyjan: Thanks for pointing out it! I wanted to test, but in the end, I forgot.

This solution is no longer possible because we updated Doctrine DBAL from 2 to 3, but I added a new one in [this commit](https://github.com/mailpoet/mailpoet/pull/5641/commits/76f457eb2c087732a2d5e0987da4d1db5c22f9cd).

mailpoet/lib/Subscribers/ImportExport/ImportExportRepository.php
lysyjan: Good point. I added a comment [here](https://github.com/mailpoet/mailpoet/pull/5641/commits/2e8cb0a6e606b194cf708b7663afee6cf28fd4b7).

mailpoet/prefixer/fix-doctrine.php
lysyjan: Thanks. My bad, because I misunderstood the comment. I thought it is not needed with the new version.
I reverted this change [here](https://github.com/mailpoet/mailpoet/pull/5641/commits/bb5c8b15cb8a9ca1f7b287a04f9328b71331653d).

mailpoet/lib/Segments/DynamicSegments/Filters/MailPoetCustomFields.php
lysyjan: I had to change this in more places, and it depends on the context. Because we used a newer version of Doctrine DBAL, a parameter name with the colon in a name in the method `setParameter()` is no longer allowed. But when we use named parameters in SQL, we still need to use the colon here because it identifies the parameter.

For example, when we have a parameter name stored in a variable `$parameter = ':param'` in the past it was no issue when we used it in SQL `WHERE s.name = $parameter` and then used `$qb->setParameter($parameter)`. With the version DBAL 3 we cannot do this anymore, and I have to remove the colon from variables and add it before them in SQL. So, now it must be: `$parameter = 'param'`, `WHERE s.name = :$parameter`.

I hope this will help you to understand those changes better. 

I'm sorry I should add it to the PR description.

alex-mpoet: Thanks for the explanation. This is a bit of a gotcha, I wonder if it could be documented somewhere so there would be less risk of a parameter with an invalid name quietly slipping into code when other developers work on the codebase. What happens if you forget to put a colon in SQL before a variable?

mailpoet/lib/Doctrine/SerializableConnection.php
alex-mpoet: Thank you for adding this code!

mailpoet/lib/Segments/DynamicSegments/Filters/MailPoetCustomFields.php
lysyjan: > This is a bit of a gotcha, I wonder if it could be documented somewhere so there would be less risk of a parameter with an invalid name quietly slipping into code when other developers work on the codebase. What happens if you forget to put a colon in SQL before a variable?

It took me some time to find out why some of our queries stopped working. I tried to find a page about it in the documentation, but I was not successful.
When you forget to put the colon in a query, it will suspect that it is not a variable but the column's name. When you use the colon in a parameter name, it will throw an exception about a missing value for the parameter.

TLDR: Your query will not work.

It's interesting that a part of our code was without an issue, but somewhere, we needed to fix it. I think we don't need to mention anything special in our documentation.

alex-mpoet: > When you forget to put the colon in a query, it will suspect that it is not a variable but the column's name

I assume it will throw an error in this case. Is it clear enough so it can be quickly fixed? Or will it require lengthy debugging?

lysyjan: > I assume it will throw an error in this case. Is it clear enough so it can be quickly fixed? Or will it require lengthy debugging?

The exception contains `Named parameter "email_0" does not have a bound value.`. I think that's a pretty good guide. 
What do you think?

alex-mpoet: Yes, it seems descriptive enough to assume that current error handling allows such cases to be noticed early and fixed promptly. Thank you for sharing the details!

alex-mpoet (approved):
Good job!

Aschepikov (approved):
GJ!

PR #5643, opened 2024-06-07
Fix live preview when changing button font size in form editor [MAILPOET-5655]

NeosinneR:
## Description

This PR fixes live preview when changing the button's font size. 

## Code review notes

The issue was that [FontSizePicker](https://developer.wordpress.org/block-editor/reference-guides/components/font-size-picker/) can return a [number or string](https://developer.wordpress.org/block-editor/reference-guides/components/font-size-picker/#onchange-value-number-string-undefined-selecteditem-fontsize-void) when changing value. However, we were expecting a number and automatically suffixed `px`, which made the value invalid (`20pxpx`). 

Screenshot 2024-06-07 at 15 50 28

The fix checks the type of the value, and if it's already a string, it won't suffix it because the component could be set to `em` or `rem` instead of `px`.

## QA notes

In the form editor change the font size of the button. Try also using different units (`em` or `rem`). In all cases, you should see the font size changing also on the button. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5655]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5655]: https://mailpoet.atlassian.net/browse/MAILPOET-5655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/form-editor/blocks/submit/edit.tsx
+++ b/mailpoet/assets/js/src/form-editor/blocks/submit/edit.tsx
@@ -46,6 +46,11 @@ function SubmitEdit({ attributes, setAttributes }: Props): JSX.Element {
     </InspectorControls>
   );
 
+  const fontSize =
+    typeof attributes.styles.fontSize === 'number'
+      ? `${attributes.styles.fontSize}px`
+      : attributes.styles.fontSize;
+
   const styles: CSSProperties = !attributes.styles.inheritFromTheme
     ? {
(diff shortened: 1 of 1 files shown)


lysyjan (approved):
Good job! It looks good.

veljkho (approved):
Looks good, verified !

PR #5644, opened 2024-06-07
Don't allow importing to "WordPress Users" list [MAILPOET-3977]

NeosinneR:
## Description

This PR removes the option to select the "WordPress Users" list when importing subscribers. You can't really import them anyway because the next time a WP user sync happens, it would remove everyone who isn't a WP user, which would result in data loss. 

## Code review notes

I opted for the simplest solution of just hiding the "WordPress Users" list on the import page. 

I considered keeping it there and showing a message that you can only update subscribers in that list, but it has some downsides
- on the next page, you'd still see `X subscribers added to "WordPress Users"` or `X existing subscribers were updated and added to "WordPress Users"`.
- if I'd just hide those names from these notices, they would look broken, because they expect at least one list
- the backend would still import them, only to remove them the next time a sync occurs

In the end, I think it wouldn't really solve the issue we are trying to solve. 

Regarding supporters' use case (p1637742878412000/1637725149.408100-slack-C01GH6AED0S), that they use the list to update existing subscribers' status without creating new ones: a workaround is to create a new list for this import, and then bulk remove those new subscribers. At least it's clear what's happening, without misleading notices and potentially vanished users after sync. 

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-3977]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-3977]: https://mailpoet.atlassian.net/browse/MAILPOET-3977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscribers/ImportExport/ImportExportFactory.php
+++ b/mailpoet/lib/Subscribers/ImportExport/ImportExportFactory.php
@@ -37,7 +37,7 @@ class ImportExportFactory {
 
   public function getSegments() {
     if ($this->action === self::IMPORT_ACTION) {
-      $segments = $this->segmentsListRepository->getListWithSubscribedSubscribersCounts([SegmentEntity::TYPE_DEFAULT, SegmentEntity::TYPE_WP_USERS]);
+      $segments = $this->segmentsListRepository->getListWithSubscribedSubscribersCounts([SegmentEntity::TYPE_DEFAULT]);
     } else {
       $segments = $this->segmentsListRepository->getListWithAssociatedSubscribersCounts();
       $segments = $this->segmentsListRepository->addVirtualSubscribersWithoutListSegment($segments);
--- a/mailpoet/tests/integration/Subscribers/ImportExport/ImportExportFactoryTest.php
+++ b/mailpoet/tests/integration/Subscribers/ImportExport/ImportExportFactoryTest.php
@@ -3,6 +3,7 @@
 namespace MailPoet\Test\Subscribers\ImportExport;
 
 use MailPoet\Entities\CustomFieldEntity;
+use MailPoet\Entities\SegmentEntity;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Subscribers\ImportExport\ImportExportFactory;
 use MailPoet\Subscribers\SubscribersRepository;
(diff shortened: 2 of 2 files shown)


mailpoet/assets/js/src/subscribers/import-export/import/step-data-manipulation/select-segment.jsx
```diff
@@ -15,7 +15,13 @@ function SelectSegment({ setSelectedSegments }) {
   const { segments: segmentsContext } = useContext(GlobalContext);
   const { all: originalSegments, updateAll: updateContextSegments } =
     segmentsContext;
-  const [selectionSegments, setSelectionSegments] = useState(originalSegments);
+
+  // Remove WordPress Users list from options, because you can't import to it
+  const [selectionSegments, setSelectionSegments] = useState(() =>
+    originalSegments.filter(
+      (segment) => segment.name !== MailPoet.I18n.t('wordPressUsersListName'),
```
costasovo: I'm not a fan of using the name to filter here. AFAIK we store the name in the DB so it is store in the language which is detected when we run the migration. I didn't test it but I think that admins that use different language would still get the list.
I see that we also have `segment.type` (or we can easily added if not present on the import page) so I think it would be safer to use `segment.type === 'wp_users'` for filtering.

Screenshot 2024-06-10 at 13 26 03

costasovo (commented):
@NeosinneR  I suggest changing how we filter out the list. Please see my other comment.

mailpoet/assets/js/src/subscribers/import-export/import/step-data-manipulation/select-segment.jsx
NeosinneR: @costasovo great suggestion, thank you. There wasn't `type` on the import pages, so it didn't even occur to me it's an option. And as it turned out, there is a much simpler solution by not fetching the WP Users list in the first place :) I also added tests.

costasovo (approved):
@NeosinneR Thank you for the adjustment. I tested it locally and it works as expected.

1. The list is not present in the import section
2. The list is present in the export section
3. When attempting to create `WordPress Users` in the import section I got reasonable error.

Aschepikov (approved):
GJ!

PR #5649, opened 2024-06-11
Change time for the nightly Circle CI workflow and update used packages [MAILPOET-6095]

lysyjan:
## Description

This PR slightly updates our Circle CI jobs.
- it changes the time of the nightly job
- it updates used versions of the 3rd plugins
- it adds using the oldest PHP and MySQL in the nightly job with premium

## Code review notes

_N/A_

## QA notes

We can skip QA here.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6095]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6095]: https://mailpoet.atlassian.net/browse/MAILPOET-6095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -957,7 +957,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: '0 22 * * 1-5'
+          cron: '0 1 * * 1-5'
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -993,8 +993,8 @@ workflows:
           <<: *slack-fail-post-step
           name: acceptance_oldest
           woo_core_version: 8.8.3
-          woo_subscriptions_version: 5.6.0
-          woo_memberships_version: 1.23.1
+          woo_subscriptions_version: 5.9.1
+          woo_memberships_version: 1.25.2
           automate_woo_version: 5.8.5
           mysql_command: --max_allowed_packet=100M --default-storage-engine=MYISAM
           mysql_image: mysql:5.5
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1060,3 +1060,32 @@ workflows:
...
+          codeception_image_version: 7.4-cli_20220605.0
+          wordpress_image_version: wp-6.4_php8.0_20231114.1
+          requires:
+            - build_premium
+      - unit_tests:
+          <<: *slack-fail-post-step
+          name: unit_with_premium_oldest
+          executor: wpcli_php_mysql_oldest
+          requires:
+            - build_premium
```
costasovo: Do we need this job? I mean, unit tests should test only one class or function with all other stuff mocked, so by this principle, they shouldn't be dependent on premium being active. I think that we don't even load WordPress for the unit tests.

costasovo (commented):
@lysyjan Thanks for the update. The new schedule works better with the weekend gap because the results will always be from the same day. 👍 

I have a question about L-1 versions. I thought we wanted to automate the L-1 detection so that we don't have to hardcode and later manually update those. I see there is another ticket for detecting beta and RC versions but there is no for L-1. Shouldn't the automation of L-1 version download be the part of this ticket?

.circleci/config.yml
lysyjan: Good point. I will update it.

lysyjan:
> I have a question about L-1 versions. I thought we wanted to automate the L-1 detection so that we don't have to hardcode and later manually update those. I see there is another ticket for detecting beta and RC versions, but there is no ticket for L-1. Shouldn't the automation of L-1 version download be part of this ticket?

@costasovo We have [a different ticket](https://mailpoet.atlassian.net/browse/MAILPOET-6097) for automated updates, which contain WP, WC, and other plugins. It is in a separate ticket because it is a spike and should be time-boxed. We are not sure right now if it is possible to achieve our expectations. Does it make sense to you?

costasovo (approved):
@lysyjan Thanks for the explanation. I missed the ticket for research. It makes sense to do it timeboxed 👍

PR #5650, opened 2024-06-11
Show correct email's scheduled time when server and client are in different timezones [MAILPOET-5495]

NeosinneR:
## Description

This PR fixes the issue of showing the correct scheduled time for sending an email when the server and client time zones are different. When the email is scheduled, the date time in the database is saved without a time zone, practically meaning it's saved with the server's time zone in mind. However, on the front end, when the time zone is not provided, it assumes browser's time zone. 

Because of this, on the email listing page, the scheduled time is incorrectly rendered in the server timezone, which is misleading, and it also causes the UI to show `Not sent yet` for the duration of the time zone difference before the email is actually sent. 

**Example #1** - server time is 8:30 am, client time is 10:30 am. I schedule email to be sent in 15 minutes (8:45 am server, 10:45 am client).

Screenshot 2024-06-11 at 10 30 47

**Before this PR**, the client already thought that the email is past scheduled time, and because of that it showed `Not sent yet` message.

Screenshot 2024-06-11 at 10 28 53

**After this PR**, it correctly shows 10:45 am. 

Screenshot 2024-06-11 at 10 29 55

**Example #2** - server time is 8:30 am, client time is 10:30 am. I schedule email to be sent in 3 hours (11:30 am server, 1:30 pm client).

Screenshot 2024-06-11 at 10 32 34

**Before this PR**, the email listing page showed that the email would be sent at 11:30 am, which was misleading because it would be in fact sent at 1:30 pm client's time.

Screenshot 2024-06-11 at 10 33 10

**After this PR**, it correctly shows 1:30 pm. 

Screenshot 2024-06-11 at 10 32 21

## Code review notes

I haven't figured out how to write tests for this, as manipulating time zones on the front end is not very straightforward. Open to any suggestions.

## QA notes

You don't necessarily have to change the server time zone to test this, all it takes is to change the time zone of your computer. Please test using the same timezone, but also time zones that are ahead and behind the server (if the server is at +00:00, test in +00:00, +02:00, and -02:00). Also test scheduling the email within the timezone difference and outside of it (i.e., in the next 15 minutes, but also on the next day). 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5495]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5495]: https://mailpoet.atlassian.net/browse/MAILPOET-5495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/AdminPages/Pages/Newsletters.php
+++ b/mailpoet/lib/AdminPages/Pages/Newsletters.php
@@ -109,7 +109,6 @@ class Newsletters {
       ->format(DateTime::DEFAULT_DATE_FORMAT);
     $data['current_time'] = $dateTime->getCurrentTime();
     $data['current_date_time'] = $dateTime->getCurrentDateTime()->format(DateTime::DEFAULT_DATE_TIME_FORMAT);
-    $data['server_timezone_in_minutes'] = $dateTime->getServerTimezoneInMinutes();
     $data['schedule_time_of_day'] = $dateTime->getTimeInterval(
       '00:00:00',
       '+15 minutes',
--- a/mailpoet/lib/AdminPages/Pages/Newsletters.php
+++ b/mailpoet/lib/AdminPages/Pages/Newsletters.php
@@ -109,6 +109,7 @@ class Newsletters {
       ->format(DateTime::DEFAULT_DATE_FORMAT);
     $data['current_time'] = $dateTime->getCurrentTime();
     $data['current_date_time'] = $dateTime->getCurrentDateTime()->format(DateTime::DEFAULT_DATE_TIME_FORMAT);
+    $data['server_timezone_in_minutes'] = $dateTime->getServerTimezoneInMinutes();
     $data['schedule_time_of_day'] = $dateTime->getTimeInterval(
       '00:00:00',
       '+15 minutes',
--- a/mailpoet/assets/js/src/date.ts
+++ b/mailpoet/assets/js/src/date.ts
@@ -20,6 +20,7 @@ export const MailPoetDate: {
   time: (date: MomentInput) => string;
   convertFormat: (format: string) => string;
   isInFuture: (dateString: string, currentTime: MomentInput) => boolean;
+  adjustForTimezoneDifference: (date: Date) => Date;
 } = {
   version: 0.1,
   options: {},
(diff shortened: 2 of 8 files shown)


mailpoet/lib/AdminPages/Pages/Newsletters.php
```diff
@@ -109,6 +109,7 @@ public function render() {
       ->format(DateTime::DEFAULT_DATE_FORMAT);
     $data['current_time'] = $dateTime->getCurrentTime();
     $data['current_date_time'] = $dateTime->getCurrentDateTime()->format(DateTime::DEFAULT_DATE_TIME_FORMAT);
+    $data['server_timezone_in_minutes'] = $dateTime->getServerTimezoneInMinutes();
```
triple0t: Because `MailPoet.Date.adjustForTimezoneDifference()` can be used anywhere in the application, I would suggest we move this variable to a global place e.g [layout](https://github.com/mailpoet/mailpoet/blob/6e66bf1f4b77db32ae6fe260e1842303925f7d7d/mailpoet/views/layout.html) or anywhere else that works.

triple0t (commented):
Looks good. 
I added a tiny comment for review

mailpoet/lib/AdminPages/Pages/Newsletters.php
NeosinneR: @triple0t nice suggestion 😉 Fixed in https://github.com/mailpoet/mailpoet/pull/5650/commits/c8afc972a25316deea1cb330263c635f22de68bd

triple0t (dismissed):
LGTM

Aschepikov:
Hi @NeosinneR as I understand everything is ok on the send page, but on the list page the time is changing according **to the chosen UTC AND +2 as my real-time**, that's why I see some confusing things

UTC -8 shows my time -8 +2 on the email list page
image
image

UTC -2 shows not sent yet my current time -2 +2 looks like it is not in the past but I don't see the time
image
image


UTC +2  11:45 +2 +2 is 15:45 
image
image

UTC +8 and with this time everything is tomorrow because +8 +2 
image
On the email list, I see that it is tomorrow and there is no difference in what time I am choosing
image

Aschepikov:
With UTC 0 it works correctly, but I am not sure if is it a desired behavior or not when there is another UTC in the settings

NeosinneR:
@Aschepikov nice catch! Since I've been operating in server +00:00 timezone, I missed this issue. The thing is, that PHP calculates time zone offset with a positive number when it's ahead of UTC, while JS calculates it as negative when it's ahead of UTC.

E.g., UTC-2 in PHP is -120 minutes, but it's +120 minutes in JS. 

I fixed this issue in https://github.com/mailpoet/mailpoet/commit/f3442c441f76041eed830e5ce1e5ab5fcebb650b and also added some commentary in the code.

Aschepikov (approved):
GJ!

PR #5655, opened 2024-06-13
Fix post-notification history subject search - MAILPOET-6072

triple0t:
## Description

Fix post-notification history subject search. 

When searching for post-notification history items, we only check for the item in the Newsletter subject table column. Post notification titles may contain MailPoet shortcodes like [newsletter:post_title], which correspond to the Post title when the Newsletter is sent. Due to the limitation of only checking for the newsletter subject (which would be saved in the Database as "Post notification title [whatever shortcode]"), we were unable to fetch some post notification history items.

This PR fixes the issue by ensuring we also check for the Rendered Newsletter subject when searching for a post-notification history item.

## Code review notes

_N/A_

## QA notes

* Create a post notification email for testing, and make sure the subject line contains dynamic content, for example → “New Post Alert! [newsletter:post_title]”

* Create a few WordPress posts and ensure the post notification emails are sent so there is a history

* Under MailPoet > Emails, click the Post Notification tab

* Click “View History” next to this Post Notification email that has dynamic content in the subject line

* use the search box in the top left to search for a specific subject, i.e. the dynamic content filled into the subject/name of the post

* Check and confirm the items exist

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/876

## Linked tickets

[MAILPOET-6072](https://mailpoet.atlassian.net/browse/MAILPOET-6072)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6072]: https://mailpoet.atlassian.net/browse/MAILPOET-6072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Listing/NewsletterListingRepository.php
+++ b/mailpoet/lib/Newsletter/Listing/NewsletterListingRepository.php
@@ -211,8 +211,8 @@ class NewsletterListingRepository extends ListingRepository {
 
     if ($type && $type === NewsletterEntity::TYPE_NOTIFICATION_HISTORY) {
       $queryBuilder
-        ->join('n.queues', 'sq')
-        ->andWhere('sq.newsletterRenderedSubject LIKE :search')
+        ->leftJoin('n.queues', 'sq')
+        ->andWhere('sq.newsletterRenderedSubject LIKE :search or n.subject LIKE :search')
         ->setParameter('search', "%$search%");
     } else {
       $queryBuilder
--- a/mailpoet/lib/Newsletter/Listing/NewsletterListingRepository.php
+++ b/mailpoet/lib/Newsletter/Listing/NewsletterListingRepository.php
@@ -204,11 +204,21 @@ class NewsletterListingRepository extends ListingRepository {
       ->setParameter('status', $group);
   }
 
-  protected function applySearch(QueryBuilder $queryBuilder, string $search) {
+  protected function applySearch(QueryBuilder $queryBuilder, string $search, array $parameters = []) {
     $search = Helpers::escapeSearch($search);
-    $queryBuilder
-      ->andWhere('n.subject LIKE :search')
-      ->setParameter('search', "%$search%");
+
+    $type = $parameters['type'] ?? null;
+
+    if ($type && $type === NewsletterEntity::TYPE_NOTIFICATION_HISTORY) {
+      $queryBuilder
+        ->join('n.queues', 'sq')
+        ->andWhere('sq.newsletterRenderedSubject LIKE :search')
+        ->setParameter('search', "%$search%");
+    } else {
+      $queryBuilder
+        ->andWhere('n.subject LIKE :search')
+        ->setParameter('search', "%$search%");
+    }
   }
 
   protected function applyFilters(QueryBuilder $queryBuilder, array $filters) {
--- a/mailpoet/tests/_support/AcceptanceTester.php
+++ b/mailpoet/tests/_support/AcceptanceTester.php
@@ -876,7 +876,7 @@ class AcceptanceTester extends \Codeception\Actor {
     $i->cli(['action-scheduler', 'run', '--force']);
   }
 
-  public function triggerNewsletterScheduledTask($newsletterId) {
+  public function rescheduleNewsletterToBeSentImmediately($newsletterId) {
     $i = $this;
 
     // scheduler will create a task and schedule run in the next minute.
--- a/mailpoet/tests/_support/AcceptanceTester.php
+++ b/mailpoet/tests/_support/AcceptanceTester.php
@@ -876,6 +876,19 @@ class AcceptanceTester extends \Codeception\Actor {
     $i->cli(['action-scheduler', 'run', '--force']);
   }
 
+  public function triggerNewsletterScheduledTask($newsletterId) {
+    $i = $this;
+
+    // scheduler will create a task and schedule run in the next minute.
+    // setting to a date and time in the past to force sending
+    $sql = "UPDATE mp_mailpoet_scheduled_tasks t "
+      . " JOIN mp_mailpoet_sending_queues q ON t.id = q.task_id "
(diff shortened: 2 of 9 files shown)


mailpoet/tests/_support/AcceptanceTester.php
```diff
@@ -876,6 +876,19 @@ public function triggerAutomationActionScheduler(): void {
     $i->cli(['action-scheduler', 'run', '--force']);
   }
 
+  public function triggerNewsletterScheduledTask($newsletterId) {
```
costasovo: I would suggest renaming the method to `rescheduleNewsletterToBeSentImmediatelly`. I think it better describes what it does and could be easier to understand in the test context. Does that sound reasonable?

mailpoet/lib/Newsletter/Listing/NewsletterListingRepository.php
```diff
@@ -204,11 +204,21 @@ protected function applyGroup(QueryBuilder $queryBuilder, string $group) {
...
-      ->andWhere('n.subject LIKE :search')
-      ->setParameter('search', "%$search%");
+
+    $type = $parameters['type'] ?? null;
+
+    if ($type && $type === NewsletterEntity::TYPE_NOTIFICATION_HISTORY) {
+      $queryBuilder
+        ->join('n.queues', 'sq')
+        ->andWhere('sq.newsletterRenderedSubject LIKE :search')
+        ->setParameter('search', "%$search%");
```
costasovo: Currently [when we create a post notification](https://github.com/mailpoet/mailpoet/blob/c0ef898dc04949d0928b1e353c32be9ec31191ca/mailpoet/lib/Cron/Workers/Scheduler.php#L394-L427) it is set to sending status, but at this time there is no newsletterRenderedSubject or queue. The time depends on the background worker setup. So I would suggest to cover both in the search.

Something like this should work (I haven't tested it):
```php
$queryBuilder
        ->join('n.queues', 'sq')
        ->andWhere('sq.newsletterRenderedSubject LIKE :search OR n.subject LIKE :search')
        ->setParameter('search', "%$search%");
```

costasovo (commented):
@triple0t Overall looks good! Good job! I have two small suggestions. Please see my comments.

As for the tests, did you consider adding an integration test to the [NewsletterListingRepository.php](https://github.com/mailpoet/mailpoet/blob/9b56fdc688753c705d43ef93c7671db926922c3a/mailpoet/tests/integration/Newsletter/Listing/NewsletterListingRepositoryTest.php#L12)?
I would personally add an integration test. Acceptance tests are the most expensive (in terms of execution time and maintenance). The change is related only to the code of the listing repository so I think an integration test on the NewsletterListingRepository would be sufficient.

mailpoet/lib/Newsletter/Listing/NewsletterListingRepository.php
triple0t: This is already implemented here https://github.com/mailpoet/mailpoet/blob/d6c953aacec39b70f98735e9edafdcad04a016d7/mailpoet/lib/Newsletter/Listing/NewsletterListingRepository.php#L212-L216

mailpoet/tests/_support/AcceptanceTester.php
triple0t: Noted. This will be done in a follow up commit

triple0t:
> As for the tests, did you consider adding an integration test to the [NewsletterListingRepository.php](https://github.com/mailpoet/mailpoet/blob/9b56fdc688753c705d43ef93c7671db926922c3a/mailpoet/tests/integration/Newsletter/Listing/NewsletterListingRepositoryTest.php#L12)?
I would personally add an integration test. Acceptance tests are the most expensive (in terms of execution time and maintenance). The change is related only to the code of the listing repository so I think an integration test on the NewsletterListingRepository would be sufficient.

This is primarily a UI-driven change, so having an acceptance test instead of an integration test makes more sense. Plus, I needed the test case to cover both rendered and unrendered newsletters, hence the decision to go with the acceptance test.

---- 

PR updated. Thanks for the review, @costasovo

mailpoet/lib/Newsletter/Listing/NewsletterListingRepository.php
costasovo: Aha! Sorry, I missed that. It was changed later! 👍

costasovo:
> This is primarily a UI-driven change, so having an acceptance test instead of an integration test makes more sense. 

@triple0t I'm sorry, but I struggle to understand that. I mean, I understand that we noticed the issue in the UI, but it's because we take input from the user and present the results in the UI. I see that the whole logic is encapsulated in the NewsletterListingRepository, which accepts a filter definition and returns results based on the filter. In this PR, we changed the behavior for certain filter values. I don't see the benefit of using the acceptance test in this case. What am I missing?

The acceptance test [took 6.8 seconds on Circle CI](https://output.circle-artifacts.com/output/job/94c6553c-1c32-432a-ae67-325871aa8fae/artifacts/6/tests/_output/report.html) the current [integration tests for the NewsletterListingRepository take less than half a second](https://output.circle-artifacts.com/output/job/bba97d86-600a-440d-a760-8088db780471/artifacts/0/codeception/report.html). That's the main reason I still think it is worth considering testing this in an integration test.

> Plus, I needed the test case to cover both rendered and unrendered newsletters,

I think you can prepare the data for both cases using the factories we use in integration tests.

triple0t:
Thank you for taking the time to explain and share your reservation regarding the test case, @costasovo. I truly appreciate it.

I agree with you; we may be able to replicate this acceptance test with an integration test.

Unfortunately, I currently do not have the time or the capacity to carry out this action. (I would have loved to give it a try, though)

If the change in this PR matches your expectations, would you be okay with moving it forward? Alternatively, would you be open to converting the written test if you feel strongly about the test cases?

costasovo:
@seun I added the integration test and simplified the acceptance to make it faster. I decided to keep it in a simpler form because there was no coverage for that feature.

costasovo (approved):

triple0t:
>  I added the integration test and simplified the acceptance to make it faster. I decided to keep it in a simpler form because there was no coverage for that feature.

Thank you, @costasovo  👏🏾👏🏾

veljkho (approved):
Works as expected!

PR #5662, opened 2024-06-19
Allow to cancel scheduled and running sending tasks [MAILPOET-5755][MAILPOET-6113]

NeosinneR:
## Description

This PR introduces a couple of changes:

1. Ability to cancel scheduled or running tasks (e.g., if they're stuck).
2. Ability to reschedule cancelled tasks with the originally scheduled date.
3. Show subscriber email for sending tasks associated with a single subscriber.
4. Increase the number of tasks from 5 to 20 (in each group - sending, cancelled, running, completed).

Screenshot 2024-06-19 at 08 40 52

## Code review notes

In https://github.com/mailpoet/mailpoet/pull/5662/commits/07a21ebb2181745f3eb941c735f456e0f9ed3b9a I couldn't import TasksListDataRowProps (from `./tasks-list-data-row.tsx`) because of circular dependencies, so I retyped it again. 

As part of this PR, I also committed a fix for [MAILPOET-6113] as I've used the same limitation for canceling and rescheduling confirmation modals.

## QA notes

**Cancelling a scheduled task:**
1. Create an email campaign and schedule it in the future.
2. On the `Help > System Status` page, you should see the option to cancel the task.

**Cancelling a running email:**
1. Unless this is a long-running task (e.g., sending an email to thousands of people), a workaround is to change the `wp_mailpoet_scheduled_tasks.status` column to `NULL`. However, if you have working cron, by the time you reload the System Status page, the cron might pick up the task and process it. 

**Rescheduling a cancelled task**
1. When the task is cancelled, you should see it in the list of Cancelled tasks.
2. On the reschedule confirm dialog, depending on wether you cancelled scheduled or running task, you should see a different message - either that the task will be rescheduled immediately or on the originally scheduled date in the future.

Screenshot 2024-06-19 at 08 52 01

**Handling errors**
1. If between the time you click on `Cancel task` or `Reschedule task` and confirm the action in the dialog, the task status changes, you should see an error message that the task can't be cancelled/rescheduled.
2. You can simulate this by clicking the button, changing the status in database, and only then confirming the dialog.

Screenshot 2024-06-19 at 08 54 46

**Subscriber's email**
1. For any 1:1 emails (e.g., welcome emails), you should see the subscriber's email address associated with the task.
2. If you send a regular newsletter to many subscribers, you should see `Multiple subscribers`.
3. However, if you send a regular newsletter to just a single subscriber, you should still see their email.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5755]
[MAILPOET-6113]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5755]: https://mailpoet.atlassian.net/browse/MAILPOET-5755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAILPOET-6113]: https://mailpoet.atlassian.net/browse/MAILPOET-6113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/API/JSON/v1/Help.php
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  private function validateAndFindTask($data) {
+    if (!isset($data['id'])) {
+      return $this->handleException(__('Missing mandatory argument `id`.', 'mailpoet'));
+    }
+    $id = $data['id'];
+    $task = $this->scheduledTasksRepository->findOneById($id);
+    if (!$task) {
+      return $this->handleException(__('Task not found.', 'mailpoet'));
```
lysyjan: I think we should use return here an error of type `ApiError:NOT_FOUND`, but your logic here always returns `BAD_REQUEST`.

mailpoet/lib/API/JSON/v1/Help.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+    try {
+      $this->scheduledTasksRepository->rescheduleTask($task);
+      return $this->successResponse();
+    } catch (\Exception $e) {
+      return $this->handleException($e->getMessage());
+    }
+  }
+
+  private function validateAndFindTask($data) {
```
lysyjan: This method makes two different things, and I think it would be better to separate them.
For example, one method is for getting a task with an entity return type, and the second method is for validation when it returns a bool value. When the return value from the validation is false, you return a bad request error.
After that, you could remove the condition `!($task instanceof ScheduledTaskEntity)`.

What do you think about it?

mailpoet/lib/API/JSON/v1/Help.php
```diff
@@ -0,0 +1,69 @@
...
+use MailPoet\Newsletter\Sending\ScheduledTasksRepository;
+
+class Help extends APIEndpoint {
+
+  public $permissions = [
+    'global' => AccessControl::PERMISSION_MANAGE_HELP,
+  ];
+
+  /** @var ScheduledTasksRepository */
+  private $scheduledTasksRepository;
```
lysyjan: I think we are using typed properties in the new code now because we support PHP 7.4 and higher.

mailpoet/tests/integration/API/JSON/v1/HelpTest.php
```diff
@@ -0,0 +1,94 @@
...
+    $response = $this->endpoint->rescheduleTask([]);
+    verify($response)->instanceOf('\MailPoet\API\JSON\ErrorResponse');
+    verify($response->status)->equals(400);
+    verify($response->errors[0]['message'])->equals('Missing mandatory argument `id`.');
+  }
+
+  public function testItReturnsErrorWhenTaskDoesntExist() {
+    $response = $this->endpoint->cancelTask(['id' => 99999]);
+    verify($response)->instanceOf('\MailPoet\API\JSON\ErrorResponse');
+    verify($response->status)->equals(400);
```
lysyjan: I think that the HTML status code here should be 404 and not 400. I think it's related to comments in `Help.php `in this pull request.

lysyjan (commented):
It looks very good, I just noticed some minor inconsistencies in the new API endpoint.

NeosinneR:
Thanks @lysyjan. I pushed two commits, which should address all your notes (which I agree with) :)

lysyjan (dismissed):

veljkho (changes requested):
I just installed the build and experienced 500 error and then this error is everywhere `An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'i0_.cancelled_at' in 'field list'`

However, this is my old test site and as I have already done some changes in the database before it could be not valid, but I would like some debugging first to ensure this is my site failure or maybe real issue.

veljkho:
I see the same also with the new test instance (not fresh new site with never installed MailPoet, but empty MailPoet with previous version like). I don't see this issue only if I install it on a fresh WordPress site.

I am afraid there is some issue in the build that provokes this error and on existing sites (like our customers). Can you please take a look at the site, I can provide credentials on Slack? @NeosinneR

NeosinneR:
@veljkho, there is a new migration in this PR. I'm not entirely sure if the migrations are run if you just upload the new build, or only when it's auto-updated or installed from scratch. Could you please check the DB, if there is `Migration_20240617_122847_Db` in the `mailpoet_migrations` table?

veljkho:
I triggered migration in the database and now it is good, no issues.

veljkho (changes requested):
Hey @NeosinneR 
Now I see this error on a blank page but occasionally `An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'subscriber_id' cannot be null` Is there anything suspicious here? It is happening for a moment and then everything's good. I feel it is happening when some task is in progress.

On a side note, do you think the following scenario is issue?
1. I have newsletter as scheduled to send
2. I click Cancel Task in the help page
3. Then I see Pause button in the Emails page, clicked it and then it turned to Resume automatically
4. Click Resume button
5. The newsletter says 0/0 and no task in the help page
6. Check [this GIF](https://d.pr/i/lMORWr) for reproduction

NeosinneR:
Just for reference, I moved the discussion to p1719488463821379-slack-C01GPBNS1RS

veljkho:
@NeosinneR 
Im trying now on a fresh new site and I have set local own host to send emails, set to send 1 every minute ( to test sending ).
Then I went to see how many emails are sent for sending newsletter, it was 0 / 7 subscribers, then I went to Help page to click cancel for the sending task, then I clicked to reschedule it back to sending and [I see blank page in the Help page](https://d.pr/i/qxuihc).

I am afraid that this PR at its state is very risky to go to production as this was 3rd time to experience some blank/error issue while testing.

NeosinneR:
@veljkho I've added a couple of changes that should address your issues:
1. When cancelled, there is a `Cancelled` tag on the Emails page - you no longer can pause or resume cancelled emails. 
2. I've added a list of paused tasks in `Help > System Status` so you should always see an email there. 
3. When rescheduling cancelled task, it should correctly return to its previous state - scheduled, or running (also on the Emails page).

I've asked @lysyjan for additional code review as these were non-trivial changes, so either wait for that approval or feel free to retest it now.

mailpoet/lib/Newsletter/Sending/ScheduledTasksRepository.php
```diff
@@ -363,8 +365,15 @@ public function rescheduleTask(ScheduledTaskEntity $task): void {
     if ($task->getStatus() !== ScheduledTaskEntity::STATUS_CANCELLED) {
       throw new \Exception(__('Only cancelled tasks can be rescheduled', 'mailpoet'), 400);
     }
-    $task->setStatus(ScheduledTaskEntity::STATUS_SCHEDULED);
-    $task->setInProgress(null);
+    if ($task->getScheduledAt() <= Carbon::createFromTimestamp($this->wp->currentTime('timestamp'))) {
+      $task->setStatus(ScheduledTaskEntity::VIRTUAL_STATUS_RUNNING);
+      $queue = $task->getSendingQueue();
```
lysyjan, line 373: Do you think we can cover this behavior in tests? I mean resuming a sending queue.

lysyjan (dismissed):
@NeosinneR LGTM. Maybe we can extend the new API test, but I think we can move it to QA now.

mailpoet/lib/Newsletter/Sending/ScheduledTasksRepository.php
NeosinneR: Added a bunch of new tests, including checking that the sending queue is resumed.

Aschepikov:
Hi @NeosinneR  I am catching up after vacation. 
This is the scheduled email
image
And when we cancelled it it looks like it was sent with a nice job message and a count of clicks. Is it expected or should it be empty like the first image but with the canceled badge? 
image

NeosinneR:
@Aschepikov good catch. The stats should be shown only when at least 1 email was sent. I fixed it in https://github.com/mailpoet/mailpoet/pull/5662/commits/6f768185eb60a1a5bf2f689761769ced833a705e just now. 

Screenshot 2024-07-03 at 12 36 12

Aschepikov (approved):
GJ!

Aschepikov:
Hi @veljkho can you dismiss the review as I can't merge it with the changes requested

veljkho (approved):

PR #5668, opened 2024-06-19
Fix abandoned cart task from product discovery on homepage [MAILPOET-5847][MAILPOET-5848]

lysyjan:
## Description

This PR fixes two issues:
- The link for the `Set up an abandoned cart email` item on the homepage
- The condition when the item is marked as done

## Code review notes

- I was considering if the condition from the NewsletterRepository class should be changed to equal with the new type, but in the end, I decided to make it backward compatible. This means that we detect if the abandoned cart emails with old and new types.

## QA notes

There are two parts to the check:
1. Check that the link for the `Set up an abandoned cart email` item redirects to the page with automation templates
2. When the automation for an abandoned cart is created, the task should be marked as done

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5847]
[MAILPOET-5848]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5847]: https://mailpoet.atlassian.net/browse/MAILPOET-5847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5848]: https://mailpoet.atlassian.net/browse/MAILPOET-5848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/templates/index.tsx
+++ b/mailpoet/assets/js/src/automation/templates/index.tsx
@@ -35,7 +35,11 @@ const tabs = [
 ];
 
 function Templates(): JSX.Element {
-  if (window.location.search.includes('loadedvia=woo_multichannel_dashboard')) {
+  const pageParams = new URLSearchParams(window.location.search);
+  const loadedviaParam = pageParams.get('loadedvia');
+  const initialTabParam = pageParams.get('initialTab');
+
+  if (loadedviaParam === 'woo_multichannel_dashboard') {
     window.MailPoet.trackEvent(
       'MailPoet - WooCommerce Multichannel Marketing dashboard > Automation template selection page',
       {
@@ -59,7 +63,7 @@ function Templates(): JSX.Element {
         <FromScratchButton />
       </PageHeader>
 
-      <TabPanel tabs={tabs}>
+      <TabPanel tabs={tabs} initialTabName={initialTabParam}>
         {(tab) => (
           <TemplatesGrid
             templates={automationTemplates.filter(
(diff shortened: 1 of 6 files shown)


NeosinneR (changes requested):
@lysyjan the code looks good, but I'd like to propose a small change to make this more useful for users.

I know this wasn't mentioned specifically in the ticket, but previously the link led directly to setting up abandoned cart email (`mailpoet-newsletters#/new/woocommerce/woocommerce_abandoned_shopping_cart/conditions`). I think we should keep this functionality and link to `Abandoned cart` tab on the templates page. 

Screenshot 2024-06-20 at 09 17 27

Currently, there isn't a way to do that, but with a simple change, it'd be possible to set initial tab via URL (`admin.php?page=mailpoet-automation-templates&initialTab=abandoned-cart`).

Screenshot 2024-06-20 at 09 14 01

What do you think?

lysyjan:
@NeosinneR I agree. I was considering it but decided it was out of scope. Nevertheless, this pull request is not too big, so we can add it.

lysyjan:
@NeosinneR I added two new commits that address your suggestion.

NeosinneR (approved):
Thanks for the updates, @lysyjan. Works like a charm.

veljkho (approved):
Works as expected!

PR #5669, opened 2024-06-20
Trigger abandoned cart for non subscribed subscribers [MAILPOET-6115]

websupporter:
## Description

In Automations, we treat abandoned carts as transactional. Therefore we must start the abandoned cart process without checking the subscriber status.

This change does not apply to legacy abandoned cart emails because we check in the SendingQueue the status of the subscriber [here](https://github.com/mailpoet/mailpoet/blob/aea6eb3dbbfc88d891c86f52cad11bd05f2677be/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L275-L281) (and maybe in other places, would need to check). I think that is an okay trade off, but let me know if you think we should apply this change also to legacy abandoned cart emails.

## Code review notes

_N/A_

## QA notes

Check if the abandoned cart automation works with subscribers who do not have the status "subscribed". Please check also legacy abandoned cart emails. I did so by downloading an old plugin (e.g. [4.41.0](https://github.com/mailpoet/mailpoet/releases/tag/4.41.0)) to set the legacy abandoned cart email up and then updating to this version.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6115]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6115]: https://mailpoet.atlassian.net/browse/MAILPOET-6115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/AutomaticEmails/WooCommerce/Events/AbandonedCart.php
+++ b/mailpoet/lib/AutomaticEmails/WooCommerce/Events/AbandonedCart.php
@@ -168,7 +168,7 @@ class AbandonedCart {
 
   private function scheduleAbandonedCartEmail(array $cartProductIds = []) {
     $subscriber = $this->getSubscriber();
-    if (!$subscriber || $subscriber->getStatus() !== SubscriberEntity::STATUS_SUBSCRIBED) {
+    if (!$subscriber) {
       return;
     }
 
--- a/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
+++ b/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
@@ -162,6 +162,17 @@ class AbandonedCartTest extends \MailPoetTest {
     $this->assertCount(1, $this->scheduledTasksRepository->findAll());
   }
 
+  public function testItSchedulesAbandonedCartAlsoForNonSubscribedSubscribers() {
+    $this->createNewsletter();
+    $this->createSubscriberAsCurrentUser('unconfirmed');
+    $this->wooCommerceCartMock->method('is_empty')->willReturn(false);
+
+    $abandonedCartEmail = $this->createAbandonedCartEmail();
+    $abandonedCartEmail->init();
+    $abandonedCartEmail->handleCartChange();
+    $this->assertCount(1, $this->scheduledTasksRepository->findAll());
+  }
+
   public function testItFindsUserByCookie() {
     $this->createNewsletter();
     $subscriber = $this->createSubscriber();
@@ -348,14 +359,15 @@ class AbandonedCartTest extends \MailPoetTest {
     return $scheduledTask;
   }
 
-  private function createSubscriber(): SubscriberEntity {
+  private function createSubscriber(string $status = 'subscribed'): SubscriberEntity {
     return (new SubscriberFactory())
       ->withWpUserId(123)
+      ->withStatus($status)
       ->create();
   }
 
-  private function createSubscriberAsCurrentUser(): SubscriberEntity {
-    $subscriber = $this->createSubscriber();
+  private function createSubscriberAsCurrentUser(string $status = 'subscribed'): SubscriberEntity {
+    $subscriber = $this->createSubscriber($status);
     $this->wp->method('wpGetCurrentUser')->willReturn(
       $this->makeEmpty(WP_User::class, [
         'ID' => $subscriber->getWpUserId(),
--- a/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
+++ b/mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
@@ -164,7 +164,7 @@ class AbandonedCartTest extends \MailPoetTest {
 
   public function testItSchedulesAbandonedCartAlsoForNonSubscribedSubscribers() {
     $this->createNewsletter();
-    $this->createSubscriberAsCurrentUser('unconfirmed');
+    $this->createSubscriberAsCurrentUser(SubscriberEntity::STATUS_UNCONFIRMED);
     $this->wooCommerceCartMock->method('is_empty')->willReturn(false);
 
     $abandonedCartEmail = $this->createAbandonedCartEmail();
@@ -359,14 +359,14 @@ class AbandonedCartTest extends \MailPoetTest {
     return $scheduledTask;
   }
 
-  private function createSubscriber(string $status = 'subscribed'): SubscriberEntity {
+  private function createSubscriber(string $status = SubscriberEntity::STATUS_SUBSCRIBED): SubscriberEntity {
     return (new SubscriberFactory())
       ->withWpUserId(123)
       ->withStatus($status)
       ->create();
   }
 
-  private function createSubscriberAsCurrentUser(string $status = 'subscribed'): SubscriberEntity {
+  private function createSubscriberAsCurrentUser(string $status = SubscriberEntity::STATUS_SUBSCRIBED): SubscriberEntity {
     $subscriber = $this->createSubscriber($status);
     $this->wp->method('wpGetCurrentUser')->willReturn(
       $this->makeEmpty(WP_User::class, [
--- a/mailpoet/lib/Cron/Workers/Scheduler.php
+++ b/mailpoet/lib/Cron/Workers/Scheduler.php
@@ -363,7 +363,7 @@ class Scheduler {
   public function verifySubscriber(SubscriberEntity $subscriber, ScheduledTaskEntity $task): bool {
     $queue = $task->getSendingQueue();
     $newsletter = $queue ? $queue->getNewsletter() : null;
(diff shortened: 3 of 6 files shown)


mailpoet/lib/AutomaticEmails/WooCommerce/Events/AbandonedCart.php
```diff
@@ -168,7 +168,7 @@ private function getCartProductIds($cart) {
 
   private function scheduleAbandonedCartEmail(array $cartProductIds = []) {
     $subscriber = $this->getSubscriber();
-    if (!$subscriber || $subscriber->getStatus() !== SubscriberEntity::STATUS_SUBSCRIBED) {
+    if (!$subscriber) {
```
costasovo, line 171: I see this method "triggers" the scheduling of both the legacy abandoned cart email and the one from automation (via a hook). So, this means the scheduled tasks will be created also for legacy email. I think (I didn't test it) that such tasks will end up as `completed` and may potentially skew the statistics we show for the legacy abandoned emails. So, IMO, a plus point for also fixing the legacy emails, OR alternatively, we could change the code here and trigger the hook for automation but not schedule the legacy email.

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
```diff
@@ -348,14 +359,15 @@ private function createSendingTask(NewsletterEntity $newsletter, SubscriberEntit
     return $scheduledTask;
   }
 
-  private function createSubscriber(): SubscriberEntity {
+  private function createSubscriber(string $status = 'subscribed'): SubscriberEntity {
```
costasovo:
```suggestion
  private function createSubscriber(string $status = SubscriberEntity::STATUS_SUBSCRIBED): SubscriberEntity {
```

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
```diff
@@ -348,14 +359,15 @@ private function createSendingTask(NewsletterEntity $newsletter, SubscriberEntit
...
     return (new SubscriberFactory())
       ->withWpUserId(123)
+      ->withStatus($status)
       ->create();
   }
 
-  private function createSubscriberAsCurrentUser(): SubscriberEntity {
-    $subscriber = $this->createSubscriber();
+  private function createSubscriberAsCurrentUser(string $status = 'subscribed'): SubscriberEntity {
```
costasovo:
```suggestion
  private function createSubscriberAsCurrentUser(string $status = SubscriberEntity::STATUS_SUBSCRIBED): SubscriberEntity {
```

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
```diff
@@ -162,6 +162,17 @@ public function testItFindsUserByWordPressSession() {
     $this->assertCount(1, $this->scheduledTasksRepository->findAll());
   }
 
+  public function testItSchedulesAbandonedCartAlsoForNonSubscribedSubscribers() {
+    $this->createNewsletter();
+    $this->createSubscriberAsCurrentUser('unconfirmed');
```
costasovo: I would personally use a constant `SubscriberEntity::STATUS_UNCONFIRMED` so that it is easier to find all the places in our codebase where we use the status or eventually update it (very unlikely).

```suggestion
    $this->createSubscriberAsCurrentUser(SubscriberEntity::STATUS_UNCONFIRMED);
```

costasovo (commented):
@websupporter Thanks for looking into this!

> In Automations, we treat abandoned carts as transactional. Therefore we must start the abandoned cart process without checking the subscriber status.

TIL. I always thought abandoned cart are more marketing, but according to Google they could be both :)

> This change does not apply to legacy abandoned cart emails because we check in the SendingQueue the status of the subscriber [here](https://github.com/mailpoet/mailpoet/blob/aea6eb3dbbfc88d891c86f52cad11bd05f2677be/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L275-L281) (and maybe in other places, would need to check). I think that is an okay trade off, but let me know if you think we should apply this change also to legacy abandoned cart emails.

It already partially applies to the legacy emails (see my other [comment](https://github.com/mailpoet/mailpoet/pull/5669#discussion_r1647132303)). So I would suggest applying the change to the legacy emails. I think it should be enough to alter the condition [here](https://github.com/mailpoet/mailpoet/blob/aea6eb3dbbfc88d891c86f52cad11bd05f2677be/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L275) to match also automatic emails with proper event for woo abandoned email. Or perhaps update/reuse [this variable](https://github.com/mailpoet/mailpoet/blob/aea6eb3dbbfc88d891c86f52cad11bd05f2677be/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L190-L192).

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
websupporter: [2c966a7](https://github.com/mailpoet/mailpoet/pull/5669/commits/2c966a7cfcd7daf91da1eefb080bc7545bbcd2bd)

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
websupporter: [2c966a7](https://github.com/mailpoet/mailpoet/pull/5669/commits/2c966a7cfcd7daf91da1eefb080bc7545bbcd2bd)

mailpoet/tests/integration/AutomaticEmails/WooCommerce/Events/AbandonedCartTest.php
websupporter: [2c966a7](https://github.com/mailpoet/mailpoet/pull/5669/commits/2c966a7cfcd7daf91da1eefb080bc7545bbcd2bd)

websupporter:
Hello @costasovo,
thank you for your review and your suggestions.

I extended the NewsletterEntity with the method `isTransactional()`. To me it made sense to locate this knowledge there. I applied then this method (replaced e.g. `$isTransactional`) and extended the method to include legacy abandoned cart emails. While testing, I found another instance where this method needed to be applied in the `Scheduler`.

Please let me know what you think.

Thank you :)

costasovo (approved):
@websupporter Great job! I like the outcome of this PR. Cool that you unified the logic for checking that an email is transactional 👍 and that it is now in one place.

I tested it locally with both legacy and automation and it worked as expected. I added a link to version 4.41.0 to the description so that it is easier for QA to download the version that supports creating legacy automatic emails.

veljkho (changes requested):
Hi @websupporter 

I have created 2 WP users with `unconfirmed` and `unsubscribed` status and in a private window logged and added item to cart.. then checking scheduled newsletters or tasks in `Help` page, there is no anything scheduled for sending? I will wait for 30 mins, but is this proper scenario, should it send to [these users](https://d.pr/i/3BGWOr)?

veljkho:
I see it is as in the progress of sending.. @websupporter I think it works. It was my test site thing, but now when I unblocked the sending irrelevant to the PR, I see abandoned cart as scheduled.

veljkho (changes requested):
I tested this on 2 sites and can't get automated abandoned cart as sent. I see automation with entries but never with success. Also I see `automation_abandoned_cart` as scheduled and completed in the database, but no entries in automation and/or sent emails.

I'm testing with the above scenario of creating new WP user and adding to cart.

websupporter:
Hi @veljkho,
thank you for testing. I tested it now on a jurassic installation and it worked for me with an `unsubscribed` subscriber.
image
image

Can you give me more details on how to reproduce the issue?

veljkho:
@websupporter Im trying with any status, and I just cant get Abandoned cart. Im creating new WP user and then logging in in a private browser or even different browser, then adding product to cart and leaving the page/browser.
I could see `automation_abandoned_car` as triggered in the database and scheduled for 30 mins.. then I edit it and reschedule for 1 min in advance. Then nothing happens, no emails sent, no automation entries...

What am I doing wrong? I tried also on JN site.

websupporter:
> I could see automation_abandoned_car as triggered in the database and scheduled for 30 mins.. then I edit it and reschedule for 1 min in advance.

Ah! I see. You should not edit it. There is a second check if really 30minutes have passed before actually activating the automation. Reason being: Imagine two different active Abandoned Cart Automations. One fires after 30m, another after 1h. When the hook fires after 30m, we need to ensure that only the 30m automation starts and not the 1h automation.

veljkho (approved):
It worked for `unconfirmed` after waiting, but I will test this more thoroughly on the release tomorrow as I am not sure why the automation is slow.

PR #5659, opened 2024-06-18
Using official WordPress images and introducing new GitHub action [MAILPOET-6097]

lysyjan:
## Description

The goal of this PR is to start using official WordPress docker images in our integration and acceptance tests. This step allows us to create automated updating docker images, which is possible via a new GitHub action. I want to test it manually in the first phase, so I need to merge those changes to the trunk branch.

## Code review notes

- I want to add more comments and polish the scripts in the next pull request when I am sure that the scripts work as I expect.
- In the first phase the action should only commit changes to the specific branch if there are some
- The next pull request will add creating a pull request

## QA notes

We can skip QA here.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6097]

## After-merge notes

_N/A_

## Tasks

- [x] 🚫 I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] 🚫 I added sufficient test coverage
- [x] 🚫 I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-6097]: https://mailpoet.atlassian.net/browse/MAILPOET-6097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5670, opened 2024-06-20
Update GitHub action updating used WordPress and plugins [MAILPOET-6097]

lysyjan:
## Description

This PR fixes and improves GH action for updating used WordPress images and downloaded plugins:

- fixed using env variable for the step updating Automate Woo
- added saving versions to temporary files
- added step opening pull-request
- added comments
- used function for updating version based on tags from GH private repositories to avoid code repetition
- used a specific version for WooCommerce instead of the latest for unification with the rest of the plugins

## Code review notes

This is almost the final step. I would like to execute this action on the trunk, and if everything is fine, we can add a cron definition.

## QA notes

We don't need QA here.

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/5659
https://github.com/mailpoet/mailpoet/pull/5660
https://github.com/mailpoet/mailpoet/pull/5661

## Linked tickets

[MAILPOET-6097]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6097]: https://mailpoet.atlassian.net/browse/MAILPOET-6097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/check-wordpress-versions.yml
```diff
@@ -127,3 +127,13 @@ jobs:
...
           git push origin HEAD:refs/heads/update-plugins-and-wordpress
+
+      # Create a pull request if there are changes
+      - name: Create Pull Request
+        if: env.CHANGES_DETECTED == 'true'
+        uses: peter-evans/create-pull-request@v6
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: update-plugins-and-wordpress
+          title: 'Update WordPress Docker and WooCommerce Versions'
```
costasovo: Afaik, it is not only WooCommerce, but it could also be other plugins, right?

.github/workflows/scripts/check_wordpress_versions.php
```diff
@@ -46,15 +61,21 @@ function getLatestAndPreviousVersions($sortedVersions) {
   return [reset($latestVersionGroup), end($previousVersionGroup)];
 }
 
-function replaceLatestVersion($latestVersion) {
+/**
+ * We specify the latest WordPress version only in the docker-compose file for the tests.
+ */
+function replaceLatestVersion($latestVersion): void {
```
costasovo: I see that you don't use typehints for method parameters. Here but also for other functions. Is that on purpose? I think they can be helpful to catch error when something (e.g. API response) changes.

costasovo (commented):
@lysyjan Thanks for adding the comments to the functions! It helps to figure out what a function does. Overall looks good! Great job!

I added just a couple of comments. Small details.

.github/workflows/check-wordpress-versions.yml
lysyjan: This title is a bit tricky because it can be all plugins without WP or only WC, etc.
Maybe Porter can update it. We could generate it, but I am not sure if it is worth it.

lysyjan:
@costasovo Thank you for your feedback. I added two new commits addressing your suggestions.

.github/workflows/check-wordpress-versions.yml
costasovo: It is probly not worth it. I would keep it as is. In case we find it confusing we can use something more generic like "update versions in Circle CI configuration"

costasovo (approved):

PR #5671, opened 2024-06-20
Ensure not loaded Mixpanel is not blocking UI [MAILPOET-6116]

NeosinneR:
## Description

We initialize [basic Mixpanel object](https://github.com/mailpoet/mailpoet/blob/638fb7e56a2d5928dffc9cd1a61c6badf4cec480/mailpoet/views/analytics.html) within the plugin, and only load the full Mixpanel script from their CDN. However, some of UI components depends on calling a callback within the track function, which breaks, when the Mixpanel library is not loaded due to AdBlockers. This causes the track event to never call the callback and the original action is not performed (e.g., a redirect). 

## Code review notes

I tried to refactor `analytics-events.js` to TypeScript, but got stuck for almost two hours so rather decided to keep it `.js` in favor of shipping the fix. 

## QA notes

I think QA can be skipped. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6116]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6116]: https://mailpoet.atlassian.net/browse/MAILPOET-6116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/analytics-event.js
+++ b/mailpoet/assets/js/src/analytics-event.js
@@ -27,7 +27,7 @@ function track(name, data = [], options = {}, callback = null) {
   const optionsData = options === CacheEventOptionSaveInStorage ? {} : options;
 
   if (typeof window.mixpanel.track !== 'function') {
-    window.mixpanel.init(window.mixpanelTrackingId);
+    window.mixpanel.init(window.mixpanelTrackingId, window.mixpanelInitConfig);
   }
 
   if (typeof window.mailpoet_version !== 'undefined') {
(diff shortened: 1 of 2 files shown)


costasovo (approved):
@NeosinneR I tested it locally with and without 3rd party libs enabled and it works as expected. I don't have an adblocker so I didn't test with an add-blocker, but I assume you've tested it.

PR #5697, opened 2024-06-27
Move changelog from readme.txt to separate file [MAILPOET-6119]

pavel-mailpoet:
## Description

_N/A_

## Code review notes

See the Jira ticket for details. 

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks


mailpoet/tasks/release/ChangelogController.php
```diff
@@ -72,6 +72,14 @@ private function sanitizePunctuation($message, $fallback) {
   }
 
   private function updateReadme($heading, $changesList) {
+    if (file_exists(dirname($this->readmeFile) . DIRECTORY_SEPARATOR . 'CHANGELOG.md')) {
```
lysyjan, line 75: Please, can you explain why we need this condition? In previous commits, the changelog was added, so I would assume we don't need this condition, but maybe I'm missing something.

mailpoet/tasks/release/ChangelogController.php
```diff
@@ -84,4 +92,32 @@ private function updateReadme($heading, $changesList) {
     }
     file_put_contents($this->readmeFile, $readme);
   }
+
+  private function updateReadmeWitchChangelogFile($heading, $changesList) {
```
lysyjan: Is it a typo here? Should be the method name `updateReadmeWithChangelogFile`?

mailpoet/tasks/release/ChangelogController.php
```diff
@@ -72,6 +72,14 @@ private function sanitizePunctuation($message, $fallback) {
   }
 
   private function updateReadme($heading, $changesList) {
```
lysyjan, line 74: If we rely on that both files exist, we can change this method to `updateReadmeAndChangelog` and add the changelog as a second parameter. What do you think about it? It can also remove the generated changelog file path in the new code.

lysyjan (commented):
@pavel-mailpoet Thanks for those changes. I added some comments; please take a look and reply when you can.

mailpoet/tasks/release/ChangelogController.php
pavel-mailpoet: Yes, we need this condition here. The code is used in both plugins; the changelog doesn't exist in the Premium plugin. I didn't want to check specifically for free/premium, but I made it more general.

mailpoet/tasks/release/ChangelogController.php
pavel-mailpoet: Yep, a typo, I will fix that.

mailpoet/tasks/release/ChangelogController.php
pavel-mailpoet: Sorry, I don't think I understand what you mean. 

The name as is makes more sense to me. You are calling the function when you want to update readme. It may move the latest changelog to a new file if it exists or leave it there. However, that is an implementation detail of this function. 

Could you please elaborate?

mailpoet/tasks/release/ChangelogController.php
lysyjan: Thanks, I didn't realize that.

mailpoet/tasks/release/ChangelogController.php
lysyjan: My suggestion was based on the condition (first comment). It doesn't make sense anymore because of usage in the premium plugin.

mailpoet/tasks/release/ChangelogController.php
pavel-mailpoet: fixed

lysyjan (changes requested):
@pavel-mailpoet I have tested your solution, and the changelog and readme should contain the latest change notes. So, we should not copy the previous version from readme to the changelog, as it is in your current solution. Does this make sense to you?

If you want, you can check it as it looks on [the WooCommerce plugin page](https://wordpress.org/plugins/woocommerce/#developers).

lysyjan (approved):
Good job! Thank you

PR #5695, opened 2024-06-26
Normalize automation action icons [MAILPOET-6132][MAILPOET-5960]

NeosinneR:
## Description

This PR fixes 3 issues
1. Inconsistent padding in icons (i.e., some SVGs had padding, some were from edge to edge)
2. Blurred icons in Firefox ([MAILPOET-5960]) caused by scaling
3. Overflowing icons when in smaller boxes

The screenshots below are from a yet-to-be-merged branch, but the same format is also used on the `Analytics > Subscribers` list page.

**Before**
Screenshot 2024-06-26 at 16 58 45

**After**
Screenshot 2024-06-26 at 17 53 10

## Code review notes

_N/A_

## QA notes

Check if trigger and action icons are not missing and they are properly shown in:
- Automation editor
- Trigger/Action block popup
- Automation analytics

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/885

## Linked tickets

[MAILPOET-6132]
[MAILPOET-5960]

## After-merge notes

_N/A_

## Tasks

- [x] 🚫 I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] 🚫 I added sufficient test coverage
- [x] 🚫 I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5960]: https://mailpoet.atlassian.net/browse/MAILPOET-5960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-6132]: https://mailpoet.atlassian.net/browse/MAILPOET-6132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAILPOET-5960]: https://mailpoet.atlassian.net/browse/MAILPOET-5960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5700, opened 2024-06-27
Automation logs [MAILPOET-5436]

NeosinneR:
## Description

In automation analytics, the admin can now see every subscriber's full activity. This is helpful when debugging issues and seeing which subscribers took which branch (when the `if/else` step is used). 

Screenshot 2024-06-27 at 22 24 32

## Code review notes

_N/A_

## QA notes

To see automation logs, you need to have active automation with a `Send email` step and have at least one subscriber enter it. 

Then, on the `Analytics > Subscribers` page, you'll see the `View activity` link, which opens the modal.

Things to check:
1. The URL parameter controls modal. When the modal is opened, you can copy the URL and open it in a new tab, and you should see the same modal.  
2. When the step failed for some reason, you should see that reason written in the modal.
3. You should see an error notice when you try to open an invalid run ID by changing the URL. 

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/886

## Linked tickets

[MAILPOET-5436]

## After-merge notes

_N/A_

## Tasks

- [x] I added sufficient test coverage (they are in the premium branch)

[MAILPOET-5436]: https://mailpoet.atlassian.net/browse/MAILPOET-5436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/automation-flow/step-footer.tsx
```diff
@@ -103,7 +103,7 @@ export function StepFooter({ step }: { step: Step }): JSX.Element | null {
   return (
     <>
       <FailedStep step={step} />
-      <Tooltip text={__('View subscribers', 'mailpoet')}>
+      <Tooltip text={__('View subscribers activity', 'mailpoet')}>
```
JanJakes: I'm wondering if the label "View subscribers activity" is correct in English. It could mean sth like "the activity of multiple subscribers", but it may not be what we mean, and even in that case, "View subscriber activity" may be more natural. For a single subscriber, it can be "View subscriber's activity", but even there, simply "View subscriber activity" sounds better to me.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/activity-modal.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const pageParams = useMemo(
+    () => new URLSearchParams(location.search),
+    [location],
+  );
+  const [runId, setRunId] = useState(null);
+
+  // Open modal when activity param changes in the URL
+  useEffect(() => {
+    setRunId(pageParams.get('runId'));
+  }, [pageParams]);
```
JanJakes: Is there a reason to mirror the URL state to a local state? Using it directly from the URL could simplify the code and save a re-render.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/activity-modal.tsx
```diff
@@ -59,13 +60,29 @@ export function ActivityModal(): JSX.Element {
     };
   }, [runId]);
 
+  if (state === 'hidden') {
+    return <div />;
```
JanJakes: Can we just return `null`?

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/activity-modal.tsx
```diff
@@ -9,22 +11,58 @@ export function ActivityModal(): JSX.Element {
...
+    const controller = new AbortController();
+    const loadRunData = async () => {
+      if (!runId) {
+        return;
+      }
+      setState('loading');
+
+      try {
+        const data = await apiFetch<{ data: RunData }>({
+          path: `/automation/analytics/run-logs&id=${runId}`,
```
JanJakes: While the modal is under analytics, I'm not sure if the API for run logs should be. After all, seems like a rather universal automation run log API, so it could as well be sth like `/automation-runs/logs` or similar.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/modal/rows.tsx
```diff
@@ -0,0 +1,54 @@
...
+  {
+    key: 'status',
+    isSortable: false,
+    label: __('Status', 'mailpoet'),
+  },
+];
+
+export function transformLogsToRows(logs: Log[], steps: Steps) {
+  return logs
+    ? logs.map((log) => [
```
JanJakes: Seems like the ternary operator is not needed. Empty arrays are truthy, and even if they werent, an empty array would be correctly mapped to empty result array.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/cells/activity.tsx
```diff
@@ -0,0 +1,6 @@
+import { __ } from '@wordpress/i18n';
+import { Button } from '@wordpress/components';
+
+export function ActivityCell(): JSX.Element {
+  return <Button variant="link">{__('View activity', 'mailpoet')}</Button>;
```
JanJakes: With `variant="tertiary"`, this would have a nicer clicking area (see Automation listing or the new Segments listing, for instance). Just a nice to have...

JanJakes (commented):
Great stuff!

A few things I noticed:

1. The heading seems to differ from the design — email vs. subscriber name, font size, doesn't look really vertically aligned:

Screenshot 2024-06-28 at 15 48 02
<br/><br/><br/>

2. When there is a badge only in a cell, it is not vertically centered. Empty `.mailpoet-analytics-activity-modal-status-info` takes some space due to a margin:
Screenshot 2024-06-28 at 15 48 24
<br/><br/>

3. In the modal, I seem to always see only sample data, even with premium active and keys valid. No requests in the Network tab. The rest of analytics correctly switches between sample/real data. Also, all rows from the sample data show only one same person (probably not a big deal).

I also see some icons misaligned, but realized that was fixed in https://github.com/mailpoet/mailpoet/pull/5695.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/activity-modal.tsx
JanJakes: Actually, see also: https://github.com/mailpoet/mailpoet-premium/pull/886#discussion_r1658738492

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/automation-flow/step-footer.tsx
NeosinneR: This button should mean "the activity of multiple subscribers." It is not related to the activity log modal - it is a button on the automation step in the editor, which redirects to the Subscribers tab with that step filtered. But still, I think you are correct that as it is, it is not grammatically correct and should be `View subscribers' activity` (with the apostrophe). 

Screenshot 2024-07-01 at 10 02 38

JanJakes: @NeosinneR And I think that `View subscriber activity` should work for both singular and plural as well.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/activity-modal.tsx
NeosinneR: This was meant to prevent loading activity modal when any other URL parameter changes. This `useEffect` can be called many times when the URL changes, but the other one with calling an API will only be called if `runId` param changes. Do you know if there is a better way to handle this?

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/activity-modal.tsx
NeosinneR: Based on the discussion in p1719816749285269-slack-C0313447YBC, we'll keep it premium.

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/activity-modal.tsx
JanJakes: I think you don't need the state for `runId`, you could probably just do:

```ts
const runId = pageParams.get('runId');
```

Then you can remove one state and one `useEffect`. Wouldn't that work the same? You'd be still passing only `runId` as a dependency to the data fetching effect. Would that work? (Didn't try it.)

NeosinneR: It did work, thanks! https://github.com/mailpoet/mailpoet/pull/5700/commits/8a2e8a59c6cdf0114482a15976d325b730f9934d

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/activity-modal.tsx
NeosinneR: Yep, https://github.com/mailpoet/mailpoet/pull/5700/commits/8d576d77712eebb26a454b8d604aa50087037ae3

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/cells/activity.tsx
NeosinneR: Fixed in https://github.com/mailpoet/mailpoet/pull/5700/commits/a71fcd136cf19b7d5f647fa1a55d197b120f1dcd

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/subscribers/modal/rows.tsx
NeosinneR: Fixed in https://github.com/mailpoet/mailpoet/pull/5700/commits/40098963b63a7ea6ec76eb8bbd8482a5a28f7684

mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/automation-flow/step-footer.tsx
NeosinneR: Changed in https://github.com/mailpoet/mailpoet/pull/5700/commits/7fe1e7b80c432e153a3a20f3278c26dc1b7487df

NeosinneR:
@JanJakes thank you for the detailed review! To address those 3 points outside of inline comments

> The heading seems to differ from the design — email vs. subscriber name, font size, doesn't look really vertically aligned

It differs from the original design but follows the suggestion from Wagner in pcD9cT-8fb-p2#comment-7195. But I did update the font size and aligned it vertically, so it should look better now (https://github.com/mailpoet/mailpoet/pull/5700/commits/cada73f3b3e3a76d93acb7a2a5eadd94b7ac5c8d).

Screenshot 2024-07-01 at 14 17 13

> When there is a badge only in a cell, it is not vertically centered.

Fixed in https://github.com/mailpoet/mailpoet/pull/5700/commits/001e3a5a558132b9b15bbf9463db6e96d4991892

> all rows from the sample data show only one same person (probably not a big deal).

Improved in https://github.com/mailpoet/mailpoet/pull/5700/commits/7f200363b4b8e64f33191985fbf8bb67dede2d0c by adding sample data for each subscriber.

> I seem to always see only sample data, even with premium active and keys valid. No requests in the Network tab. The rest of analytics correctly switches between sample/real data.

Were you testing from ZIP or you pulled the code? If the latter, have you built premium assets? There is a change in https://github.com/mailpoet/mailpoet-premium/pull/886/commits/a2e0e5d463458a2ce9a5f2a9f646fe385caaf907 that allows running log sample data to be replaced with real ones but requires building JS assets.

JanJakes (approved):
Looks and works great, thanks!

Aschepikov (approved):
GJ!

PR #5702, opened 2024-06-28
Fix rendering of special characters using [site:name] shortcode [MAILPOET-5359]

### LINKED ISSUE #4947: Subscription confirmation email isn't html-unescaping site:title in the subject field
**Describe the bug**
The template for confirmation emails includes `[site:title]` by default, but this includes HTML encoded characters. My site name is "Jason O'Neil" and when someone tries to subscribe to my site they see an email with the subject "Confirm your subscription to Jason O&#039;Neil".

**To reproduce**

1. Create a wordpress site with "General Settings -> Site Title" including a special character, like "Jason O'Neil" (not the unicode apostrophe).
2. Set up MailPoet with a list and subscription form
3. In "MailPoet -> Settings -> Signup Confirmation -> Open Template Editor" set the subscription to include `[site:title]` (it should already have this by default)
4. Attempt to subscribe a new email address using one of the forms

**Actual behavior**
The email sent to that address should use unicode in the plain text subject. "Confirm your subscription to Jason O&#039;Neil"

Note: `[site: title]` works correctly in the template body

**Expected behavior**
The email sent to that address has HTML encoding in the plain text subject. "Confirm your subscription to Jason O'Neil"

**Screenshots**
image

image

- WordPress version: 6.2
- PHP version: 7.4.3
- MailPoet version: 4.11.1

JanJakes:
## Description

This fix is analogous to https://github.com/mailpoet/mailpoet/commit/aba54c51160b61eb53a8f640ccb1a438763aed0b.

I'm only not sure if `$this->wp->wpStripAllTags` should be used as well. The referenced commit uses it, but in the order of 1. strip all tags and 2. then decode HTML special chars, which, I believe, will have no effect (=  we strip tags that are still encoded).

I think either we should use `$this->wp->wpStripAllTags(...)` in both cases _after_ `htmlspecialchars_decode` or maybe not use it at all, at least for the site title, that is unlikely to contain HTML tags.

What do you think?

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5359], closes #4947

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5359]: https://mailpoet.atlassian.net/browse/MAILPOET-5359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Shortcodes/Categories/Site.php
+++ b/mailpoet/lib/Newsletter/Shortcodes/Categories/Site.php
@@ -27,7 +27,8 @@ class Site implements CategoryInterface {
   ): ?string {
     switch ($shortcodeDetails['action']) {
       case 'title':
-        return $this->wp->getBloginfo('name');
+        // Decoding special characters such as &amp; to &, etc.
+        return htmlspecialchars_decode($this->wp->getBloginfo('name'));
 
       case 'homepage_url':
         return $this->wp->getBloginfo('url');
--- a/mailpoet/tests/integration/Newsletter/ShortcodesTest.php
+++ b/mailpoet/tests/integration/Newsletter/ShortcodesTest.php
@@ -414,8 +414,8 @@ class ShortcodesTest extends \MailPoetTest {
   }
 
   public function testItCanProcessSiteTitleShortcode() {
-    $optionName = 'blogname';
-    $siteName = get_option($optionName);
+    $siteName = 'Test site name with characters like ’, <, >, &';
+    update_option('blogname', $siteName);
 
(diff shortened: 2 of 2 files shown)


lysyjan (approved):
@JanJakes I think we can remove striping tags, and we have [a ticket](https://mailpoet.atlassian.net/browse/MAILPOET-5920) which should dig more about it.

Aschepikov (approved):
GJ!

PR #5714, opened 2024-07-04
Fix forms privacy policy links [MAILPOET-4504]

alex-mpoet:
## Description

The replacement of link tags was failing if a privacy policy page wasn't set up. In this case incorrect arguments were passed to the `Helpers::replaceLinkTags()` method.

~~Existing saved forms need to be fixed too, I've made a migration for that. There is no easy way to tell a privacy policy link from another link (it can be translated, modified etc.), but looking at form templates I see that only privacy policy links had `[link]` tags that could have remained unprocessed, so we can replace all of them. If a form body is malformed, there will be a PHP warning about unserialization, but the migration should still be completed.~~

## Code review notes

_N/A_

## QA notes

Follow the reproduction steps from the Jira ticket. Check with the privacy page set up or not in WP Admin > Settings > Privacy.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4504]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-4504]: https://mailpoet.atlassian.net/browse/MAILPOET-4504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Form/Templates/FormTemplate.php
+++ b/mailpoet/lib/Form/Templates/FormTemplate.php
@@ -148,20 +148,18 @@ EOL;
     return $this->cdnAssetUrl->generateCdnUrl("form-templates/{$this->assetsDirectory}/$filename");
   }
 
-  protected function replaceLinkTags($source, $link, $attributes = [], $linkTag = false): string {
-    return Helpers::replaceLinkTags($source, $link, $attributes, $linkTag);
+  protected function replaceLinkTags($source, $link, $attributes = []): string {
+    return Helpers::replaceLinkTags($source, $link, $attributes);
   }
 
   protected function replacePrivacyLinkTags($source, $link = '#'): string {
     $privacyPolicyUrl = $this->wp->getPrivacyPolicyUrl();
     $attributes = [];
-    $linkTag = false;
 
     if (!empty($privacyPolicyUrl)) {
       $link = $this->wp->escUrl($privacyPolicyUrl);
       $attributes = ['target' => '_blank'];
-      $linkTag = 'link';
     }
-    return $this->replaceLinkTags($source, $link, $attributes, $linkTag);
+    return $this->replaceLinkTags($source, $link, $attributes);
   }
 }


mailpoet/lib/Migrations/App/Migration_20240704_104122_App.php
```diff
@@ -0,0 +1,62 @@
...
+      return;
+    }
+    $changed = false;
+    foreach ($body as $key => $block) {
+      // Skip blocks of irrelevant types and malformed blocks
+      if (!isset($block['type']) || $block['type'] !== 'paragraph' || !isset($block['params']['content'])) {
+        continue;
+      }
+      // Skip blocks without link tags
+      if (!strpos($block['params']['content'], '[link]')) {
```
costasovo: I think we need to check `strpos($block['params']['content'], '[link]') === false` here because if content starts with `[link]` strpos would return 0 and we would skip it.

mailpoet/lib/Migrations/App/Migration_20240704_104122_App.php
```diff
@@ -0,0 +1,62 @@
...
+        continue;
+      }
+      $block['params']['content'] = Helpers::replaceLinkTags($block['params']['content'], '#');
```
costasovo: Shouldn't we use `get_privacy_policy_url` to get the proper URL instead of replacing it with `#`? This way we remove `[link]` from the form, but will create a link that doesn't go to the privacy policy page. So actually this won't help much...

Thinking about it, I would hope that users would notice that a form they placed on their site contains a strange [link] and replace it with the correct link or remove it. So maybe we don't need the migration. What do you think?

costasovo (commented):
@alex-mpoet Thanks for the fix! I think we could improve the migration a bit or maybe remove it :) Please see my comments.

mailpoet/lib/Migrations/App/Migration_20240704_104122_App.php
alex-mpoet: > Shouldn't we use get_privacy_policy_url to get the proper URL instead of replacing it with #?

Fair point. As I mentioned in the description, we can't tell privacy policy link tags from other link tags, so I was a bit concerned that we can encounter unrelated link tags in the wild, and having a dummy link is better than a wrong link in this case, in my opinion.

> Thinking about it, I would hope that users would notice that a form they placed on their site contains a strange [link] and replace it with the correct link or remove it. So maybe we don't need the migration. What do you think?

Since we haven't seen many (if any) complaints about this, and this isn't a high-impact issue if these tags are kept in existing forms, we can remove this migration.

mailpoet/lib/Migrations/App/Migration_20240704_104122_App.php
alex-mpoet: Good catch! Spoiled by JS `String.prototype.includes()`, I guess! 😄

costasovo (approved):
Thanks for the fix!

Aschepikov (approved):
GJ!

PR #5715, opened 2024-07-04
Don't strip shortcode-like text in post notifications content [MAILPOET-5197]

alex-mpoet:
## Description

Here we have a tradeoff between allowing shortcode-like text and not allowing actual shortcodes to slip into the email content. I have used the `strip_shortcodes()` WP function to strip registered shortcodes from the content, but some shortcodes may not be registered in WP (although this is officially [discouraged](https://codex.wordpress.org/Shortcode_API#Unregistered_Names)), they are processed by plugins on their own terms and they might end up in the email. Should someone need to strip these non-typical codes, they will have to do it by themselves with a filter such as [mailpoet_rendering_post_process](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Newsletter/Renderer/Renderer.php#L228-L231) or `strip_shortcodes_tagnames`.

## Code review notes

_N/A_

## QA notes

Check that [text] in square brackets isn't stripped from post contents, but actual shortcodes still are ([mailpoet_form], [gallery] etc.).

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5197]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5197]: https://mailpoet.atlassian.net/browse/MAILPOET-5197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Editor/PostContentManager.php
+++ b/mailpoet/lib/Newsletter/Editor/PostContentManager.php
@@ -132,7 +132,7 @@ class PostContentManager {
     );
 
     // remove other shortcodes
-    $content = preg_replace('/\[[^\[\]]*\]/', '', $content);
+    $content = $this->wp->stripShortcodes($content);
 
     return $content;
   }
--- a/mailpoet/lib/WP/Functions.php
+++ b/mailpoet/lib/WP/Functions.php
@@ -498,6 +498,10 @@ class Functions {
     status_header($code, $description);
   }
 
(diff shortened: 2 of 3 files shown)


NeosinneR (approved):
LGTM

Aschepikov (approved):
GJ!

PR #5725, opened 2024-07-10
Fix loading external images for email template thumbnails [MAILPOET-3923]

alex-mpoet:
## Description

The proxy for html2canvas is added here to work around CORS when loading external images for email template thumbnails. It's access is restricted by authentication, it only works for MailPoet stock templates by default, image types are also restricted.

## Code review notes

Please let me know if you see any security issues with this approach. Or if there's a better way altogether.

## QA notes

Follow the reproduction steps from the Jira ticket.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-3923]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-3923]: https://mailpoet.atlassian.net/browse/MAILPOET-3923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/NewsletterTemplates/TemplateImageLoader.php
+++ b/mailpoet/lib/NewsletterTemplates/TemplateImageLoader.php
@@ -29,6 +29,7 @@ class TemplateImageLoader {
       // Failed to load the image
       return false;
     }
+    $mime = $this->wp->wpGetImageMime($image);
     if (!$this->isTypeAllowed($image, $mime)) {
       // Wrong file type
       @unlink($image);
@@ -46,11 +47,11 @@ class TemplateImageLoader {
   }
 
   private function isUrlAllowed($url) {
-    $urlParts = parse_url($url);
+    $urlParts = $this->wp->wpParseUrl($url);
     $allowedExtensions = ['gif', 'png', 'jpg', 'jpeg'];
     if (
       !isset($urlParts['path'])
-      || !preg_match('/(' . join('|', $allowedExtensions) . ')$/', $urlParts['path'])
+      || !preg_match('/\.(' . join('|', $allowedExtensions) . ')$/i', $urlParts['path'])
     ) {
       return false;
     }
@@ -59,10 +60,12 @@ class TemplateImageLoader {
       'https://ps.w.org/mailpoet/assets/newsletter-templates/',
     ]);
     foreach ($allowedUrls as $allowedUrl) {
-      $allowedUrlParts = parse_url($allowedUrl);
+      $allowedUrlParts = $this->wp->wpParseUrl($allowedUrl);
       if (
-        isset($urlParts['host'], $allowedUrlParts['host'], $allowedUrlParts['path'])
+        isset($urlParts['host'], $urlParts['scheme'])
+        && isset($allowedUrlParts['host'], $allowedUrlParts['scheme'], $allowedUrlParts['path'])
         && $urlParts['host'] === $allowedUrlParts['host']
+        && $urlParts['scheme'] === $allowedUrlParts['scheme']
         && strpos($urlParts['path'], $allowedUrlParts['path']) === 0
       ) {
         return true;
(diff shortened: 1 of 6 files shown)


mailpoet/lib/NewsletterTemplates/TemplateImageLoader.php
```diff
@@ -0,0 +1,83 @@
...
+        && $urlParts['host'] === $allowedUrlParts['host']
+        && strpos($urlParts['path'], $allowedUrlParts['path']) === 0
+      ) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private function isTypeAllowed($image, &$mime = null) {
```
NeosinneR: Is there a reason for using the referenced parameter instead of calling `$mime = $this->wp->wpGetImageMime($image);` before and passing the value as a parameter? 

Looking at the rest of the codebase, we don't use them as often, and I find them to worsen readability when looking at where the function is called (i.e., where does the `$mime` come from in `if (!$this->isTypeAllowed($image, $mime)) {`)

mailpoet/lib/NewsletterTemplates/TemplateImageLoader.php
```diff
@@ -0,0 +1,83 @@
...
+    require_once ABSPATH . '/wp-admin/includes/file.php';
+    return download_url($url, self::TIMEOUT);
+  }
+
+  private function isUrlAllowed($url) {
+    $urlParts = parse_url($url);
+    $allowedExtensions = ['gif', 'png', 'jpg', 'jpeg'];
+    if (
+      !isset($urlParts['path'])
+      || !preg_match('/(' . join('|', $allowedExtensions) . ')$/', $urlParts['path'])
```
NeosinneR: To make this a bit more secure, I'd also check for the `.` before the extension. Not sure if it's exploitable, but otherwise it would allow `.insecure-jpg`.
```suggestion
      || !preg_match('/\.(' . join('|', $allowedExtensions) . ')$/', $urlParts['path'])
```

And on the other hand, make it a bit more permissive by ignoring case, `.JPG`, .`PNG`...
```suggestion
      || !preg_match('/(' . join('|', $allowedExtensions) . ')$/i', $urlParts['path'])
```

mailpoet/lib/NewsletterTemplates/TemplateImageLoader.php
```diff
@@ -0,0 +1,83 @@
...
+    ) {
+      return false;
+    }
+    /** @var string[] */
+    $allowedUrls = (array)$this->wp->applyFilters('mailpoet_template_image_allowed_urls', [
+      'https://ps.w.org/mailpoet/assets/newsletter-templates/',
+    ]);
+    foreach ($allowedUrls as $allowedUrl) {
+      $allowedUrlParts = parse_url($allowedUrl);
```
NeosinneR: Have you considered using `$this->wp->wpParseUrl()`? It should [handle consistency](https://developer.wordpress.org/reference/functions/wp_parse_url/) in the return values across PHP versions.

NeosinneR (changes requested):
Nice job, Alex! Besides some small suggestions below, it looks good to me. 

And one more question, have you considered also checking schema? E.g., not to allow something like `javascript:alert('1')`, although I can't think of ways to exploit that in this context.

mailpoet/lib/NewsletterTemplates/TemplateImageLoader.php
alex-mpoet: No particular reason. I agree that readability is a priority, so I have changed this as per your comment.

mailpoet/lib/NewsletterTemplates/TemplateImageLoader.php
alex-mpoet: Thanks! Done.

mailpoet/lib/NewsletterTemplates/TemplateImageLoader.php
alex-mpoet: Sure, why not. Let's use it.

alex-mpoet:
> And one more question, have you considered also checking schema? E.g., not to allow something like javascript:alert('1'), although I can't think of ways to exploit that in this context.

Nice suggestion, let's compare the `scheme` part of URLs too.

NeosinneR (approved):
LGTM, thanks for the changes. I'm approving even though the build failed, because it's due to another issue (p1720730101267939-slack-C01H71ZLBGQ).

@alex-mpoet once the issue is fixed, rebase might be needed, but you can then assign QA and they can approve it themselves so it's mergeable.

Aschepikov (approved):
GJ!

PR #5726, opened 2024-07-11
Don't show empty WooCommerce list on send page when Woo is not active [MAILPOET-5224]

costasovo:
## Description

This PR fixes the issue of displaying the WooCommerce list on the Newsletter send page even though Woo is inactive.
I found that there is already logic for displaying/hiding the WooCommerce list in the list listing, so I reused the same logic.

The Woo list is shown if WooCommerce is active OR if there are some WooCommerce customers. This also covers cases when a Woo plugin was active Woo list was populated and then the Woo plugin was deactivated. In case the Woo was active and was deactivated the Woo list is empty it should be hidden.

## Code review notes

_N/A_

## QA notes

Please verify that the WooCommerce list is displayed according to the logic described above. The behavior on the List page and Newsletter sending page should be the same.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5224]

## After-merge notes

_N/A_

## Tasks

- [] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [] I added sufficient test coverage
- [] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-5224]: https://mailpoet.atlassian.net/browse/MAILPOET-5224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/AdminPages/Pages/Newsletters.php
+++ b/mailpoet/lib/AdminPages/Pages/Newsletters.php
@@ -7,10 +7,12 @@ use MailPoet\AutomaticEmails\AutomaticEmails;
 use MailPoet\Config\Env;
 use MailPoet\Config\Menu;
 use MailPoet\Entities\NewsletterEntity;
+use MailPoet\Entities\SegmentEntity;
 use MailPoet\Listing\PageLimit;
 use MailPoet\Newsletter\NewslettersRepository;
 use MailPoet\NewsletterTemplates\NewsletterTemplatesRepository;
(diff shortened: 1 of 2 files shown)


alex-mpoet (approved):
Looks good to me!

Aschepikov (approved):
GJ!

PR #5727, opened 2024-07-11
Fix captcha [MAILPOET-6038]

JanJakes:
## Description

Fixes CAPTCHA race condition and simplifies and refactors the logic.

Previously, a new CAPTCHA phrase was generated when the CAPTCHA image or audio was requested "again". For that, a counter needed to be stored in the session, and since image and audio were fetched in parallel, any of them could initialize the new CAPTCHA phrase. The problem appeared when they both generated a new phrase at the same time. This resulted in a race condition, where the image and audio had different CAPTCHA phrases and some users weren't able to pass the CAPTCHA at all. This was occuring especially in Safari (maybe somehow due the type dectection range "preflight" request).

Now, the phrase is initialized on the CAPTCHA page load (and reloaded via AJAX),
so we don't need to store any additional metadata, apart from the phrase itself.

I also compressed the audio CAPTCHA mp3s more aggressively to save space. The files now take about 3.5x less space (224KB to 63KB). With Opus audio, we could go even lower, but Safari doesn't have a good support yet.

The command used was:

```bash
for f in originals/*.mp3; do ffmpeg -i "$f" -c:a libmp3lame -af "highpass=f=200, lowpass=f=15000, afftdn=nf=-30, anlmdn=s=1, rubberband=pitch=1.05, volume=1.5" -b:a 8k -ac 1 -ar 8000 "${f##*/}"; done
```

Where:
- `-c:a libmp3lame` is the MP3 codec,
- `-b:a 8k` is the bitrate,
- `-ar 8000` is the sampling rate,
- `-ac 1` specifies single channel (mono).

I also used the `-af` parameter for some extra cleanup and enhancement, with the following settings:
- `highpass` and `lowpass` to exclude frequencies out of the voice range,
- `afftdn=nf=-30` to denoise the audio with FFT,
- `anlmdn=s=1` to reduce broadband noise using Non-Local Means algorithm,
- `rubberband=pitch=1.05` to slightly raise the voice pitch for easier understanding,
- `volume=1.5` to increase volume.

## Code review notes

_N/A_

## QA notes

Please, test that the CAPTCHA works in all browsers, both image and audio with reloading, failure, success, etc.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6038]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6038]: https://mailpoet.atlassian.net/browse/MAILPOET-6038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
```diff
@@ -23,9 +23,12 @@ public function isSupported() {
   }
 
   public function renderAudio($sessionId, $return = false) {
+    if (!$this->isSupported()) {
```
alex-mpoet: The `isSupported` method checks graphical captcha support, why is it used to check support for audio rendering? We don't seem to need any external libraries for that.

mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
```diff
@@ -21,37 +21,28 @@ public function isSupported(): bool {
...
+    header("Cache-Control: no-store, no-cache, must-revalidate");
+    header('Content-Type: audio/mpeg');
+
     $audioPath = Env::$assetsPath . '/audio/';
     $phrase = $this->getPhrase($sessionId);
-    $audio = null;
     foreach (str_split($phrase) as $character) {
       $file = $audioPath . strtolower($character) . '.mp3';
       if (!file_exists($file)) {
         throw new \RuntimeException("File not found.");
```
alex-mpoet, line 32: Throwing an exception after you have outputted an audio/mpeg header may result in a browser displaying broken audio instead of a meaningful error (if error_reporting is on).

alex-mpoet (dismissed):
Great job refactoring this! Now the captcha can be passed in Safari on macOS again. 

However, there is another issue now: captcha phrase is not changing when clicking the refresh icon (🔄). Please try a few times, it's sometimes updated the first time, but then it stops until you enter an incorrect value. Is that intended? I would expect the phrase to be changed each time I click to refresh. Imagine a person not hearing the audio captcha well, it will stay the same despite refreshing.

mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
JanJakes: @alex-mpoet Good point, I probably just unified the methods. I can drop this.

JanJakes:
@alex-mpoet 

> However, there is another issue now: captcha phrase is not changing when clicking the refresh icon (🔄). Please try a few times, it's sometimes updated the first time, but then it stops until you enter an incorrect value. Is that intended? I would expect the phrase to be changed each time I click to refresh. Imagine a person not hearing the audio captcha well, it will stay the same despite refreshing.

I'm trying to replicate this, but no luck. Clicking the 🔄 again and again. Do you have any idea how I could replicate this?

JanJakes:
@alex-mpoet I guess I could make the refresh request a POST request to be sure to bypass all caching and stuff. Let me try that...

mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
JanJakes: @alex-mpoet What could we do instead? I didn't touch the lines, but if there's a better way, I can update it. That said, those files should always exist.

alex-mpoet: You did not touch these lines, but you put HTTP headers before them. Originally headers were outputted after the possible exceptions. I guess a more correct way would be to first check the existence of all necessary files in a loop and throw an exception on failure, then print headers, then iterate over files again and print them.

alex-mpoet:
@JanJakes 
> I'm trying to replicate this, but no luck. Clicking the 🔄 again and again. Do you have any idea how I could replicate this?

Here is a video recording I made: https://mailpoet.atlassian.net/browse/MAILPOET-6038?focusedCommentId=25113

mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
JanJakes: Ah, I see, addressed in https://github.com/mailpoet/mailpoet/pull/5727/commits/36f76959d53c960ba85f412274e53efffc2e8cd6.

mailpoet/lib/Subscription/Captcha/CaptchaRenderer.php
JanJakes: Fixed in https://github.com/mailpoet/mailpoet/pull/5727/commits/84912a0e687275bafecb29cd299d0b6d6900deb0.

JanJakes:
@alex-mpoet

> Here is a video recording I made: https://mailpoet.atlassian.net/browse/MAILPOET-6038?focusedCommentId=25113

Thanks! I tried to address this in https://github.com/mailpoet/mailpoet/pull/5727/commits/49a85a614eefa426ee3e82032034ba8fffe4e424. Could you, please, try if the new build fixes that for you? 🙏

alex-mpoet (approved):
@JanJakes Now it seems to work, thank you for making the improvements! I'll approve it, let's have another pair of eyes from QA on it.

Aschepikov (approved):
GJ

PR #5729, opened 2024-07-12
Fix broken HPOS tests

pavel-mailpoet:
## Description

We complete the actions in the database. However, we don't delete the batch processes. Woo thinks those crashed. It tries to log an error but doesn't have sufficient rights for that.

This pull request deletes `wc_pending_batch_processes`, which will prevent the batch process controller from retrying those.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks


costasovo (approved):
Nice catch!

PR #5734, opened 2024-07-16
WPDB driver for Doctrine [MAILPOET-6142]

JanJakes:
## Description

This PR implements WPDB driver for Doctrine and replaces the current PDO with this new driver.

**It includes:**
1. Removing old models and Paris, Idiorm and Sudzy.
2. Replacing all PDO-specific expressions.
3. Using single quotes for strings in SQL queries (more standard and required by WPDB).
4. WPDB driver, connection, statement, result, and parameter handling.
6. Exception handling with SQL error codes and SQLSTATE for conversion to their respective DBAL exceptions.
7. Removing dependency on PDO altogether.
8. Better cleanup of unused Doctrine drivers.

**It doesn't include:**
1. WPDB cache integration. This is something we could consider in the future, but as we're not using any caching adapters in Doctrine at the moment, it makes sense to first replace implement the WPDB driver without additional caching.
2. SQLite-specific code. The WP Playground uses SQLite, but I'm not sure if any SQLite specifics are needed (such as configuring SQLite Platform for Doctrine), as the WP Playground may be emulating some MySQL-like behavior, which could even mean any SQLite-specific settings could be counter-productive.

## Code review notes

There are some fixes to the existing Doctrine code (like strings to single quotes and passing parameters). If you find a way to check I didn't miss any more of those, it would be great. I tried some full-text and pattern search, but a second pair of eyes is always useful.

## QA notes

We need a general plugin smoke check, trying to trigger as much functionality as possible. This replaces the database driver, so if any issues appear, it could be anywhere. That said, our test being green already gives us some confidence.

There was a significant change in the underlying code for working with times and time zones. Please check that the times we present to users (listings, logs, help page) are correct and that scheduling newsletters and post notifications works as expected. We store time in the DB in GMT, but we always present time in the site's timezone, and we also expect the user to enter time in the site's timezone. 

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/894

## Linked tickets

[MAILPOET-6142]

## After-merge notes

_N/A_

## Tasks


mailpoet/lib/Config/Database.php
```diff
@@ -1,53 +0,0 @@
...
-      define('MP_STATISTICS_OPENS_TABLE', Env::$dbPrefix . 'statistics_opens');
-      define('MP_STATISTICS_UNSUBSCRIBES_TABLE', Env::$dbPrefix . 'statistics_unsubscribes');
-      define('MP_STATISTICS_FORMS_TABLE', Env::$dbPrefix . 'statistics_forms');
-      define('MP_STATISTICS_WOOCOMMERCE_PURCHASES_TABLE', Env::$dbPrefix . 'statistics_woocommerce_purchases');
-      define('MP_MAPPING_TO_EXTERNAL_ENTITIES_TABLE', Env::$dbPrefix . 'mapping_to_external_entities');
-      define('MP_LOG_TABLE', Env::$dbPrefix . 'log');
-      define('MP_STATS_NOTIFICATIONS_TABLE', Env::$dbPrefix . 'stats_notifications');
-      define('MP_USER_FLAGS_TABLE', Env::$dbPrefix . 'user_flags');
-      define('MP_FEATURE_FLAGS_TABLE', Env::$dbPrefix . 'feature_flags');
-      define('MP_DYNAMIC_SEGMENTS_FILTERS_TABLE', Env::$dbPrefix . 'dynamic_segment_filters');
```
costasovo: I have a concern that some plugins that integrate with MailPoet may use these constants in raw SQL queries. So maybe we could keep the definitions and perhaps mark them as deprecated (if that's possible)

mailpoet/lib/Doctrine/WPDB/Result.php
```diff
@@ -4,40 +4,69 @@
...
+  ) {
+    foreach ($result as $value) {
+      $this->result[] = (array)$value;
+    }
+    $this->rowCount = $rowCount;
+  }
+
   public function fetchNumeric() {
-    // TODO: Implement fetchNumeric() method.
+    $value = $this->result[$this->cursor++] ?? null;
```
costasovo: Just a suggestion. If I understand correctly, instead of the `$value` variable, we could use `$row` because `$result` is an array of rows, and each row is an associative array, right?
So, I would suggest renaming `$value` to `$row` in the whole file. For example later in `fetchOne` is see `reset($value)` but `reset($row)` reads much better for me and it is easier to understand that it returns the value from first column.

mailpoet/lib/Doctrine/WPDB/ConvertParameters.php
```diff
@@ -0,0 +1,84 @@
...
+    if (!array_key_exists($key, $this->params)) {
+      throw new MissingParameterException(sprintf("Parameter '%s' was defined in the query, but not provided.", $key));
+    }
+    [, $value, $type] = $this->params[$key];
+
+    // WPDB doesn't support NULL values. We need to handle them explicitly.
+    if ($value === null) {
+      $this->buffer[] = 'NULL';
+      return;
+    }
```
costasovo, line 68: This replicates the current behavior, but I see in the test it generates queries with conditions `WHERE column = null`, which is always false. I wonder how many such cases we have. I guess we can try to log this in our tests.

costasovo (commented):
@JanJakes This is awesome! 🎉 I'm super excited to see this is happening!

I added a few comments. As you are on the rotation I'll look into this tomorrow.

mailpoet/lib/Config/Database.php
costasovo: I added a helper class which reintroduces the constants.

mailpoet/lib/Doctrine/WPDB/ConvertParameters.php
JanJakes: @costasovo Did you manage to figure out how many of such cases we have in our codebase? That said, I think we'll need to keep this behavior to satisfy Doctrine's behavior. What I mean is that if the ORM can generate such queries, we should probably support them on the DBAL level as well.

costasovo: @JanJakes I ran the tests and logged all cases of null arguments that appeared after WHERE or JOIN. In the end, there weren't many places (actually 5), but they were called repeatedly, so the log looked quite big. 

Here is the code I used to check which cases to log to skip null in values:
```php
$shouldLog = false;
foreach ($this->buffer as $bufferPart) {
  if ((stripos($bufferPart, 'WHERE') !== false) || (stripos($bufferPart, 'JOIN') !== false)) {
    $shouldLog = true;
    break;
  }
}
```

I added two commits to address a few cases. Most of the logged occurrences came from queries created by high-level "findBy" methods, but Doctrine seems to be smart enough to cover those.
I found that, for example:
```php
$task = $this->scheduledTasksRepository->findOneBy(
      [
        'status' => [ScheduledTaskEntity::STATUS_SCHEDULED, null],
      ]
    );
```
is converted to
 ```sql
 (status IN ('scheduled', null) OR status IS NULL)
 ```
 
 I went through rest of the cases and added two commits, but they were more like nice to have.
 
 So I think we are fine.

Aschepikov:
Hi @JanJakes there is some failed test, can you check it pls

Aschepikov:
Hi @JanJakes there are some conflicts can you check it pls

Aschepikov (approved):
GJ!

PR #5743, opened 2024-07-22
Woo email customizer - style settings improvements [MAILPOET-4180]

costasovo:
## Description

This PR addresses the issue that, when editing Woo transactional emails with the email editor, some style settings were not applied consistently to both Woo content and content added in the editor.

### Changes

1. It is now possible to set font size for H1, H2, H3
2. It is now possible to set font family and font color for headings and the rest of the content
4. The H1 font size setting is applied to the main heading in the header
5. All headings use branding color. Note: Headings added via Text block cab still have a color set via the TinyMCE
6. Heading font size, font family and font color are also applied to headings in Woo block (order details, new customer, etc)
7. Text font size is also applied to content in Woo block
8. The changes applied to the Woo content are not applied immediately after updating to the new plugin version, but the user needs to save the email template in the Woo email editor. This is done to make sure the user is aware of the changes.

#### Before
screenshot-2024-07-16-at-10 52 14

#### After
Screenshot 2024-07-23 at 12 35 02

## Code review notes

I recommend going through the commits one by one.

## QA notes

Please test the customizer with various Woo transactional emails. The changes were made in the editor and in the renderer, so we also need to check the emails that were sent.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-4180]

## After-merge notes

_N/A_

## Tasks

- [x] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes -- I edited two JS files in the current editor, but I chose not to convert those to TS

[MAILPOET-4180]: https://mailpoet.atlassian.net/browse/MAILPOET-4180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/WooCommerce/TransactionalEmails/Renderer.php
```diff
@@ -70,6 +70,13 @@ public function getHTMLAfterContent() {
    * - We update inlined font-family rules in the content block generated by Woo
    */
   public function updateRenderedContent(NewsletterEntity $newsletter, string $content): string {
+    $isSavedWithStyledWooBlock = $newsletter->getGlobalStyle('woocommerce', 'isSavedWithUpdatedStyles');
+    // For Backward compatibility do not apply styles for content unless the template was edited with the editor
+    // and user visually checked and is aware of updated styles feature.
+    // This check is save to remove after 2025-07-31
```
NeosinneR: Is there an assumption that everyone will re-save their current Woo templates within a year? As far as I understand, if we remove this check after a year, it can still break the styling of emails that were saved before now but haven't been re-saved since then. Unless there is a need to remove this check, I think it would be safer to keep it so we don't have to debug why something broke in a year's time.

NeosinneR (changes requested):
Overall, the changes look good to me, @costasovo, but I'd like to check with you regarding two things.

I only realized this while testing it locally, but a light brand color can make content headings hard (or impossible) to read. The main heading is fine as it adjusts by changing the font color from white to black automatically.

Screenshot 2024-07-22 at 15 05 15

What would you say about allowing color change for all headings? 

**Before:**
Screenshot 2024-07-22 at 15 10 38

**After:**
Screenshot 2024-07-22 at 15 11 34

Screenshot 2024-07-22 at 15 12 50

mailpoet/lib/WooCommerce/TransactionalEmails/Renderer.php
costasovo: > Is there an assumption that everyone will re-save their current Woo templates within a year?

Yep. That's how I was thinking about it. I don't like this type of condition, so I wanted to figure out a way to get rid of it. But you're right. It may cause unexpected issues after a year or so. I'll remove that part of comment.

costasovo:
@NeosinneR Good catch about the light branding colors. I'll add the heading color setting.

costasovo:
@NeosinneR I implemented the proposed changes. The heading colors are, by default, not set, and in this case the headings use the branding colors. I believe that this is what users will need in the majority of cases.

NeosinneR (approved):
LGTM, thanks for the changes 👍

Aschepikov (approved):
GJ!

PR #5745, opened 2024-07-22
Fix rendering buttons in Outlook 2023 [MAILPOET-5732]

lysyjan:
## Description

_N/A_

## Code review notes

The new email editor inspired the solution, where the buttons are compatible with the new Outlook. I kept the conditional rendering because I didn't want to break the old Outlooks.

## QA notes

Test the old email editor buttons across supported email clients.
You can use Email on Acid for this purpose.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5732]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5732]: https://mailpoet.atlassian.net/browse/MAILPOET-5732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Renderer/Blocks/Button.php
+++ b/mailpoet/lib/Newsletter/Renderer/Blocks/Button.php
@@ -8,7 +8,7 @@ use MailPoet\Newsletter\Renderer\StylesHelper;
 class Button {
   public function render($element, $columnBaseWidth) {
     $element['styles']['block']['width'] = $this->calculateWidth($element, $columnBaseWidth);
-    $styles = 'display:inline-block;-webkit-text-size-adjust:none;mso-hide:all;text-decoration:none;text-align:center;' . StylesHelper::getBlockStyles($element, $exclude = ['textAlign']);
+    $styles = 'display:block;text-decoration:none;text-align:center;' . StylesHelper::getBlockStyles($element, $exclude = ['textAlign']);
     $styles = EHelper::escapeHtmlStyleAttr($styles);
     $template = '
       <tr>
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Newsletter/Renderer/Blocks/Button.php
```diff
@@ -35,7 +35,29 @@ public function render($element, $columnBaseWidth) {
...
                   <![endif]-->
                   <!--[if !mso]><!-- -->
-                  <a class="mailpoet_button" href="' . EHelper::escapeHtmlLinkAttr($element['url']) . '" style="' . $styles . '"> ' . EHelper::escapeHtmlText($element['text']) . '</a>
+                  <table
+                    border="0"
+                    cellspacing="0"
+                    cellpadding="0"
+                    role="presentation"
+                    style="display:inline-block;border-collapse:separate;mso-table-lspace:0;mso-table-rspace:0;width:' . EHelper::escapeHtmlStyleAttr($element['styles']['block']['width']) . '"
+                    width="' . EHelper::escapeHtmlStyleAttr($element['styles']['block']['width']) . '"
```
costasovo: The wrapping table has a width calculated as button width - border, but it wraps the button with the border, so the inner button (td) overflows the wrapping table. It causes the whole button (if centered) to misalign a bit to the right. We could use the original width for the wrapping table.
Please see the screenshot.

Here, I have a centered button with a width set to 280px in the editor and a border of 10px.
Screenshot 2024-07-23 at 9 42 07

costasovo (changes requested):
@lysyjan Nice job!

I see that the changes are done in non `mso` branch of the code. So it looks like the new Outlook doesn't identify as "mso" `<!--[if mso]>`. It is surprising to me.

I noticed just a tiny detail about the wrapping table width. It looks like an easy fix so I think it is worth fixing it.

mailpoet/lib/Newsletter/Renderer/Blocks/Button.php
lysyjan: Good catch. I am going to fix it.

mailpoet/lib/Newsletter/Renderer/Blocks/Button.php
```diff
@@ -7,6 +7,7 @@
 
 class Button {
   public function render($element, $columnBaseWidth) {
+    $originalWidth = $element['styles']['block']['width'];
     $element['styles']['block']['width'] = $this->calculateWidth($element, $columnBaseWidth);
```
costasovo: Seeing this change, I noticed that the width is processed in `Button::calculateWidth` and it takes into account also column width. The method ensures that the resulting width is not bigger than the column.
So when we use the $originalWidth as is and place button in the three column design the wrapping table is wider than the column.

Screenshot 2024-07-23 at 16 04 55

I didn't observe any rendering issue with that, but I would prefer we fix that to be safe. What do you think?

costasovo (commented):
@lysyjan Thank you for the fix. I noticed a small issue with the width again. I'm not 100% it needs fixing. Please see my comment.

mailpoet/lib/Newsletter/Renderer/Blocks/Button.php
lysyjan: @costasovo Good catch. It's addressed in [the latest commit](https://github.com/mailpoet/mailpoet/pull/5745/commits/de0984ae2c1145a4b7b40d9f49f62e27577ceec6).

costasovo (approved):
Thank you for the fix!

Aschepikov (approved):
GJ

PR #5749, opened 2024-07-23
Fix empty else branch failure [MAILPOET-6154]

NeosinneR:
## Description

If one of the if/else branches is empty, the automation will result in a failed state. While the `StepScheduler` correctly [completes the automation run](https://github.com/mailpoet/mailpoet/blob/dec9a02ca48e4b9209b969b81aceffb994729610/mailpoet/lib/Automation/Engine/Control/StepScheduler.php#L59), `StepHandler` [will try to schedule next step](https://github.com/mailpoet/mailpoet/blob/dec9a02ca48e4b9209b969b81aceffb994729610/mailpoet/lib/Automation/Engine/Control/StepHandler.php#L154), which will fail, because if/else action has two next steps.

This PR fixes it by checking if the automation run isn't completed between running the step, and scheduling the next one. 

In addition, while implementing I realized we don't show  the `Subscribers` tab in Analytics, even though there is no technical or product reason for it (surfaced also in p1721676078611049-slack-C0313447YBC), so this PR also adds the tab to be always visible (I used it for debugging in automations without `Send email` step).

## Code review notes

_N/A_

## QA notes

To test:
1. Create automation with if/else step, but don't add any more steps below (can be as simple as tag added trigger + if/else action). 
2. Run the automation with a subscriber.
3. On Analytics > Subscribers, observe that the subscriber status is `completed` as well as the if/else step status is completed. 

In this screenshot, the first three rows are before this PR (wrong status), the other three are after this PR (correct status). 
Screenshot 2024-07-23 at 17 16 19

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6154]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6154]: https://mailpoet.atlassian.net/browse/MAILPOET-6154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/index.tsx
+++ b/mailpoet/assets/js/src/automation/integrations/mailpoet/analytics/components/tabs/index.tsx
@@ -29,6 +29,20 @@ export function Tabs(): JSX.Element {
       className: 'mailpoet-analytics-tab-flow',
       title: __('Automation flow', 'mailpoet'),
     },
+    {
+      name: 'automation-subscribers',
+      className: classNames(
+        'mailpoet-analytics-tab-subscribers',
+        !MailPoet.capabilities.detailedAnalytics.isRestricted
+          ? 'is-unlocked'
+          : '',
+      ),
+      title: (
+        <>
+          {__('Subscribers', 'mailpoet')} <Icon icon={lockSmall} />
+        </>
+      ) as unknown as string,
+    },
   ];
   if (hasEmails) {
     tabs.push({
@@ -54,20 +68,6 @@ export function Tabs(): JSX.Element {
         ) as unknown as string,
       });
     }
-    tabs.push({
-      name: 'automation-subscribers',
-      className: classNames(
-        'mailpoet-analytics-tab-subscribers',
-        !MailPoet.capabilities.detailedAnalytics.isRestricted
-          ? 'is-unlocked'
-          : '',
-      ),
-      title: (
-        <>
-          {__('Subscribers', 'mailpoet')} <Icon icon={lockSmall} />
-        </>
-      ) as unknown as string,
-    });
   }
 
   const updateUrlSearchString = useCallback(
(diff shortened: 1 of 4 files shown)


lysyjan (dismissed):
It looks like the fix works. I have a question here, @NeosinneR. Do you think it's possible to fix the existing data? A lot of users can see a wrong status for their runs.

NeosinneR:
> Do you think it's possible to fix the existing data? A lot of users can see a wrong status for their runs.

It should be possible. To fetch those runs logs,
```sql
SELECT * FROM `wp_mailpoet_automation_run_logs` WHERE 
`status` = 'failed' 
AND `error` LIKE '%nextStepNotScheduled%'
AND `step_key` = 'core:if-else'
```

I see two possible approaches to fix it:

1. A complex but safer approach would be to use app migration. Get those runs, then verify that it is indeed an `if/else` step, then that the branch was empty in that version of automation, and then update the status of automation run and automation run log. While this could be technically possible with database migration, parsing the automation JSON would make it more complicated.
2. Easier approach, with a slight chance of changing the status of the `failed` log that indeed should be failed (e.g., if the next step failed to be scheduled even though it should). In this approach, I'd just fetch those run logs, and change the status to `completed` as well as associated automation run. 

I'd go with second approach, as I don't think there is other way to trigger that error than in an empty if/else branch.

```sql
UPDATE wp_mailpoet_automation_runs
SET status = 'complete'
WHERE id IN (
    SELECT automation_run_id
    FROM wp_mailpoet_automation_run_logs
    WHERE status = 'failed'
      AND error LIKE '%nextStepNotScheduled%'
      AND step_key = 'core:if-else'
);

UPDATE wp_mailpoet_automation_run_logs
SET
  status = 'complete',
  error = NULL
WHERE status = 'failed'
  AND error LIKE '%nextStepNotScheduled%'
  AND step_key = 'core:if-else';
);
```

I tested these UPDATEs locally, and the results look correct as far as I can tell. What do you think, @lysyjan?

NeosinneR:
@lysyjan I was testing the migration a bit further (realized it also needs `updated_at = updated_at`, otherwise it would change the run log completion date), so I already committed the migration. But let me know if you'd rather take a different approach and I'll change it.

lysyjan (approved):
@NeosinneR Your solution with migration seems good to me. Thanks!

Aschepikov (approved):
GJ!

PR #5751, opened 2024-07-23
Fix automation action selector popup on small devices [MAILPOET-5184]

NeosinneR:
## Description

Add a `min-width` for automation action selector popup, otherwise it will be too narrow due to `width: min-content` property. 

## Code review notes

_N/A_

## QA notes

QA can be skipped. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5184]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5184]: https://mailpoet.atlassian.net/browse/MAILPOET-5184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/css/src/components-automation-editor/_automation.scss
+++ b/mailpoet/assets/css/src/components-automation-editor/_automation.scss
@@ -9,6 +9,11 @@
 
 .mailpoet-automation-editor-automation-wrapper {
   padding: 0 20px 50px;
+
+  // Fix for action selector popup on small devices
+  .components-popover__content {
+    min-width: 280px;
+  }
 }
 
 .mailpoet-automation-editor-automation-flow {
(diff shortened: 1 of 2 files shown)


costasovo (commented):
@NeosinneR  I tried it locally. It is way better. 
But on iPhone Max sized and smaller displays the popup doesn't fit the screen and it causes strange infinite horizontal scrolling. Please see the screenshot.

One way we could prevent it is setting `position={'middle center bottom'}` to the inserter popup component. There is [wordpress/viewport](https://github.com/WordPress/gutenberg/tree/0b94a19ca3307f98fc7124573e8be51cbe32cfd6/packages/viewport) package that allows to detect screen size in react. So with that we could center the popup only on small screens.

Screenshot 2024-07-24 at 13 46 01

NeosinneR:
@costasovo great suggestion. It turned out that the position is not the best even on a 1050px wide desktop, so I just centered it everywhere, not just on mobiles.

FYI, `position` is [deprecated](https://developer.wordpress.org/block-editor/reference-guides/components/popover/#position-yaxis-xaxis-optionalcorner), so I used `placement` instead.

Screenshot 2024-07-24 at 21 07 40

costasovo (approved):
Nicely done! Thank you!

PR #5752, opened 2024-07-24
Fix SendEmailAction's check to support multiple emails per subscriber [MAILPOET-6155]

costasovo:
## Description

It is possible that one email (e.g., purchase in category) is sent multiple times to the same subscriber. Due to an issue in a query that checks the status of sending, some sent email actions were marked as failed even though the email was successfully sent.

This PR fixes the query that was causing the issue.

## Code review notes

AutomationEmailScheduler::getScheduledTaskSubscriber was selecting the task based on subscriber and newsletter. In the case of multiple emails sent to one subscriber, the method failed to pick ScheduledTaskSubsrciberEntity because the query was fetching multiple results, but getOneOrNullResult expects only one result.

This commit fixes it by adding additional filtering by $runId to get the ScheduledTaskSubsriberEntity associated with the correct run.

I did the filtering in PHP because an alternative would be using LIKE %% in the query. The meta column is text. 

#### Steps to replication:

1. Create an automation for example Purchased in a category
3. Make sure that option Run this automation only once per subscriber is disabled (It should be by default)
5. Purchase a product from the category with the same subscriber at least twice
7. Make order as complete
9. Observe the issue on the automation analytics → tab Subscribers

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6155]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6155]: https://mailpoet.atlassian.net/browse/MAILPOET-6155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
+++ b/mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
@@ -76,8 +76,11 @@ class AutomationEmailScheduler {
       ->andWhere('st.createdAt >= :runCreatedAt')
       ->setParameter('newsletter', $email)
       ->setParameter('subscriber', $subscriber)
-      // Subtract one second to prevent rounding issue as $run->getCreatedAt() may contain milliseconds
-      ->setParameter('runCreatedAt', $run->getCreatedAt()->modify('-1 second'))
+      // Automation Run is fetched via WPDB and it ignores the gmt_offset. This query is processed via Doctrine.
+      // Doctrine uses PDO connection and uses offset. So the run's created_at is expected to be provided with the offset.
+      // By removing one day we make sure the offset is not a problem. It makes no harm as this is only performance optimization.
+      // After we switch to WPDB we could remove this modification and use the exact created_at.
+      ->setParameter('runCreatedAt', $run->getCreatedAt()->modify('-1 day'))
       ->getQuery()
       ->getResult();
     $result = null;
--- a/mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
+++ b/mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
@@ -76,7 +76,8 @@ class AutomationEmailScheduler {
       ->andWhere('st.createdAt >= :runCreatedAt')
       ->setParameter('newsletter', $email)
       ->setParameter('subscriber', $subscriber)
-      ->setParameter('runCreatedAt', $run->getCreatedAt())
+      // Subtract one second to prevent rounding issue as $run->getCreatedAt() may contain milliseconds
+      ->setParameter('runCreatedAt', $run->getCreatedAt()->modify('-1 second'))
       ->getQuery()
       ->getResult();
     $result = null;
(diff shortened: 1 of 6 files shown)


NeosinneR (dismissed):
LGTM!

mailpoet/lib/Newsletter/Scheduler/AutomationEmailScheduler.php
```diff
@@ -75,7 +75,19 @@ public function getScheduledTaskSubscriber(NewsletterEntity $email, SubscriberEn
...
+      $task = $scheduledTaskSubscriber->getTask();
+      if (!$task instanceof ScheduledTaskEntity) {
+        continue;
+      }
+      $meta = $task->getMeta();
+      if (($meta['automation']['run_id'] ?? null) === $runId) {
+        $result = $scheduledTaskSubscriber;
+        break;
+      }
+    }
```
JanJakes, line 97: @costasovo Maybe I'm overthinking this, but what if there are many of such records? Won't that be suboptimal in terms of always fetching all of them? I don't have a specific scenario, but an automation can be anything... I could imagine that you send an email after every order or something, and then, if a subscriber reaches hundreds of orders, we could run into performance issues or memory errors here.

I know that addressing this properly is not easy given the `meta` field usage (would need a more complex query matching the run ID on the DBAL level).

But I have another idea — what if we limit the scheduled tasks so that `$automationRun->getCreatedAt()` is `less or equal` than scheduled task's (or sending queue's) `createdAt`?

costasovo: @JanJakes Thanks for noticing the potential performance issue. I'll update the PR based on your suggestions.

costasovo:
@JanJakes I added three more commits. They improve the performance of the method that searches for proper scheduled task and fix an issue with TimestampListener I ran into.

JanJakes (dismissed):

Aschepikov:
HI, @costasovo I created automation to reproduce the issue and checked it before the fix and after installing. Before I had an issue as was mentioned here 
image
 and after I had another issue
image
With the new automation, the situation is the same everytime failed on this
image

Aschepikov:
15 mins later I received 2 emails, 1 from auto 1 and 1 from auto 2. But with the fix installed, there should be 3 emails from auto 1 and 2 from auto 2. Status is not renewed
image

costasovo:
@Aschepikov This is probably related to the last changes I made. I got similar errors in tests, and I thought I understood and fixed those. But apparently not :( What timezone does your site use?

costasovo:
@Aschepikov I was able to replicate the issue with the timezone set to Denver. Working on a fix.

costasovo:
@JanJakes I needed to add one more commit. I originally misunderstood how offset on DB connection works.
The problem is that AutomationRun is handled via WPDB, which stores and returns UTC values.
The PDO connection uses gmt_offset and it stores UTC, but returns and also expects values as UTC+offset.

The `created_at` comparison was failing for negative offset because the PDO connection expects UTC+offset as the parameter value, but the value from AutomationRun is UTC because it was fetched via WPDB. The exact fix would be adding offset to the AutomationRun's created_at before passing it as a parameter. The problem will be solved with the switch to WPDB. Here, we need to reduce the amount of fetched rows because of performance, so I went with a simple solution subtracting one day.

JanJakes (approved):
Awesome, thanks for tackling this!

Aschepikov (approved):
GJ!

PR #5756, opened 2024-07-26
Add stubs for deleted Idiorm models [MAILPOET-6142]

costasovo:
## Description

The motivation of this PR is the minimalization of the potential impact of the removal of Idorm models. This should prevent "Class not-found" errors triggered by the usage of these models in third-party extensions/integrations or custom code on sites.

The idea is to provide stubs for the public interface of the removed models that will throw deprecation warnings.
The stubs would return reasonable empty values. Methods for fluent interface return self, methods that should return array return empty arrays, and some (e.g., findOne) that may return a model of false returns false. So, the integration that checks returned values should be fine and log a bunch of deprecation warnings instead of fatal errors.

## Code review notes

I used the PHPDoc on the `\MailPoet\Model\Model` as the source for lists of methods I divided by return type. Then, I added methods from the model classes we currently have.

To make things simpler, I did not distinguish between individual models, so all methods work for all models. 

Another simplification is that the stubs support even non-existent method names. For these, they return null and also throw a deprecation warning. I'm not aware of any drawbacks, and that way, we cover methods I accidentally missed. If you can think of potential drawbacks to this approach, let me know, and we could list all the props and methods.

We may decide to merge it, or we can keep this branch so we can respond quickly in case we run into issues after the release.

## QA notes

Can be skipped.

## Linked PRs

This is related to this PR https://github.com/mailpoet/mailpoet/pull/5734

We may decide to merge it, or we can keep this branch so we can respond quickly in case we run into issues after the release.

## Linked tickets

[MAILPOET-6142]

## After-merge notes

_N/A_

## Tasks


costasovo:
Closing as we decided we won't add it to the primary release. I'll keep the branch for some time.

PR #5764, opened 2024-07-30
Allow changing opt-in checkbox position on the checkout page  [MAILPOET-5807]

NeosinneR:
## Description

This PR adds a new setting to change the position of the opt-in checkbox on the checkout page. 

Screenshot 2024-07-31 at 00 54 38

## Code review notes

I realized that it's not clear which checkout page we run tests against (block or shortcode), so I updated tests to explicitly use one or the other or both to increase test coverage. Especially with AutomateWoo, it turned out to be valuable because changing hooks caused the AutomateWoo opt-in checkbox to display, even if it shouldn't. 

## QA notes

1. Test that all opt-in checkbox positions are working as expected. 
2. Test that the opt-in checkbox on old installations remains positioned before terms and conditions.
3. Test that the opt-in checkbox on new installations is positioned after billing info. 
4. Test that the AutomateWoo checkbox is hidden when MailPoet's is displayed. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5807]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-5807]: https://mailpoet.atlassian.net/browse/MAILPOET-5807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Migrations/App/Migration_20240730_212419_App.php
```diff
@@ -0,0 +1,29 @@
...
+      return;
+    }
+    // Skip if the opt-in checkbox is not enabled. When enabling, the user chooses
+    // the position of the opt-in checkbox on the same settings page, so no default
+    // value is necessary.
+    if (!$this->settings->get(Subscription::OPTIN_ENABLED_SETTING_NAME, false)) {
+      return;
+    }
+    $this->settings->set(Subscription::OPTIN_POSITION_SETTING_NAME, Hooks::OPTIN_POSITION_BEFORE_TERMS_AND_CONDITIONS);
+  }
```
costasovo: Perhaps we could skip the version check and just check if the position setting is missing. 

```php
// value is not necessary or is set
if (
 !$this->settings->get(Subscription::OPTIN_ENABLED_SETTING_NAME, false)
 || $this->settings->get(Subscription::OPTIN_POSITION_SETTING_NAME, false)
 ) {
  return;
 }
 
 $this->settings->set(Subscription::OPTIN_POSITION_SETTING_NAME, Hooks::OPTIN_POSITION_BEFORE_TERMS_AND_CONDITIONS);
 ```
It is a tiny detail, but I like to avoid the version check conditions. They are always a bit risky as we may forget to update the version in case the PR merge is postponed. And we are already reading the settings.

costasovo (commented):
@NeosinneR Great job!
Thank you for improving the tests! The checkout functionality is very important so it is great we cover all the combinations.

I have just a tiny suggestion.

mailpoet/lib/Migrations/App/Migration_20240730_212419_App.php
NeosinneR: Good call; I've updated the migration. It also made me realize, that I forgot to set the default value in Popuplator, so I also added that.

costasovo (dismissed):
@NeosinneR Thank you for the updates!

Aschepikov:
@NeosinneR Hi everything works well but there is a tiny issue, reproducible with chrome and edge
image

NeosinneR:
@Aschepikov fixed in https://github.com/mailpoet/mailpoet/pull/5764/commits/a7cc7f34ed83db6fad7c83308cbedcbea7538470 :) I think it's small enough change not to ask for another code review, so once the build is green, feel free to approve and merge if it's okay by you

Aschepikov (approved):
GJ!

PR #5767, opened 2024-08-02
Minimize the risk of race conditions for "once per subscriber" automations [MAILPOET-6177]

JanJakes:
## Description

This minimizes the risk of race conditions for "once per subscriber" automations to almost zero. While it's technically not possible to avoid them with 100% certainty using WP transients that don't support atomic operations, the implemented method should avoid them effectively in all realistic use cases.

## Code review notes

Transients are used not only to acquire a lock for a given run ID & subscriber subject combination, but also to check against a unique value to avoid additional race conditions within the transients themselves. Code comments and tests explain more.

## QA notes

Testing the race conditions is hard; a quick test that "run once per subscriber" in automations works well is enough.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6177]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6177]: https://mailpoet.atlassian.net/browse/MAILPOET-6177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


costasovo (approved):
@JanJakes Great job! Thank you for adding the comments documenting the locking mechanism and checks into the code.

Aschepikov (approved):
GJ!

PR #5769, opened 2024-08-05
Add data inconsistencies checks [MAILPOET-1587]

### LINKED ISSUE #2651: Cleanup/prevention of orphaned scheduled tasks records
During the issue described in #2650, the script was terminated when it exceeded the configured memory limits of PHP.

The script termination happened after the record in the "scheduled_tasks" table was created but before any "scheduled_task_subscribers" were added. The scheduled task records were left in a weird state and the cron job scheduler would try to work on them even though they were empty.

This could be addressed by either a DB transaction that would prevent the record from being saved in the first place, or a cleanup routine that would detect empty pending tasks and complete them.

costasovo:
## Description

This PR adds the Data Inconsistencies section to the Help > System Status page. The goal is to surface data inconsistencies to users or HEs who check issues at user sites.

Normally, there shouldn't be any inconsistencies, but some might be created by manual changes in the DB or due bugs or unexpected PHP crashes, or execution timeouts.

Currently, we check for these inconsistencies:
- Orphaned sending tasks (count of sending tasks with missing queue)
- Queues without email  (count of queues with missing newsletter)
- Orphaned task subscribers (count task subscribers with missing tasks)
- Orphaned subscriptions (count of wp_mailpoet_subscriber_segment without segment or without subscriber)
- Orphaned links (count of wp_mailpoet_newsletter_links without newsletter or queue)
- Orphaned newsletter post (count of wp_mailpoet_newsletter_posts without newsletter)

If no inconsistencies are detected, the section is hidden. 

The fix button fixes the inconsistency. In all cases above, it means deleting the records from the DB.

Screenshot 2024-08-01 at 15 37 31

## Code review notes

Please review the commits one by one. For time reasons, I added only the checks listed on the ticket. The feature is built so that it is easy to add a new check. If you can think of any important check we could add please let me know. I can add it to this PR or create a follow-up ticket.

## QA notes

The easiest way to create inconsistencies is to manually delete some rows in the DB.

- Orphaned sending tasks - send a newsletter and delete the related row in `wp_mailpoet_sending_queues`
- Queues without email - send a newsletter and delete the related newsletter in `wp_mailpoet_newsletters`
- Orphaned task subscribers - send an email and delete related row in `wp_mailpoet_sending_tasks`
- Orphaned subscriptions - create subscribers add them to list and then delete the list in the `wp_mailpoet_segments_table`
- Orphaned links - send an email and delete it in `wp_mailpoet_newsletters`
- Orphaned newsletter post - send a post notification email and delete the post notification history in `wp_mailpoet_newsletters`

Note: It is possible that fixing and inconsistency may create a different one.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-1587], closes #2651

## After-merge notes

_N/A_

## Tasks

[MAILPOET-1587]: https://mailpoet.atlassian.net/browse/MAILPOET-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mailpoet/lib/Util/DataInconsistency/DataInconsistencyRepository.php
```diff
@@ -24,6 +24,17 @@ public function getOrphanedSendingTasksCount(): int {
     return (int)$this->buildOrphanedSendingTasksQuery($builder)->getSingleScalarResult();
   }
 
+  public function getOrphanedScheduledTasksSubscribersCount(): int {
+    $stTable = $this->entityManager->getClassMetadata(ScheduledTaskEntity::class)->getTableName();
+    $stsTable = $this->entityManager->getClassMetadata(ScheduledTaskSubscriberEntity::class)->getTableName();
+    $count = $this->entityManager->getConnection()->executeQuery("
+      SELECT count(*) FROM $stsTable sts
+      LEFT JOIN $stTable st ON st.`id` = sts.`task_id`
+      WHERE st.`id` IS NULL
```
lysyjan, line 45: I noticed some methods getting orphaned rows containing a condition with `OR`. Maybe we can add here a check that the subscriber exists. What do you think?

lysyjan (approved):
Good job!
I had only a small note, but we can keep it as it is.

> If you can think of any important check we could add please let me know. I can add it to this PR or create a follow-up ticket.

I was thinking about it, but I don't have anything special in my mind.

mailpoet/lib/Util/DataInconsistency/DataInconsistencyRepository.php
costasovo: I thought about it, but I see that when deleting a subscriber, we keep (I assume intentionally) the rows in task subscribers so that we display correct logs about sent emails.

Aschepikov (approved):
GJ!

PR #5775, opened 2024-08-06
Replace abandoned mtdowling/cron-expression package [MAILPOET-6167]

alex-mpoet:
## Description

The new library has a lot of fixes, I hope they don't break anything for us. I think we use correct expressions and don't rely on bugs and edge cases, so we should be good.

## Code review notes

I retained backward compatibility in two cases when switching the libraries:
1. The new `CronExpression` doesn't accept non-string values, so I added a type cast;
2. The new `CronExpression` [returns a `DateTime` object](https://github.com/dragonmantank/cron-expression/blob/master/src/Cron/CronExpression.php#L468), while the old one returned an object of the same type it was called with. We relied on the former behavior: passed `Carbon` objects and expected them to be returned. So I had to add conversion of the result from `DateTime` to `Carbon`.

## QA notes

Please ensure that scheduling of emails and scheduled tasks works properly, doesn't have time zone issues etc. This is a change of our core library so it fine to take time and test it thoroughly.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6167]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6167]: https://mailpoet.atlassian.net/browse/MAILPOET-6167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JanJakes (dismissed):
LGTM, thanks!

Aschepikov:
Hi @alex-mpoet  there are some conflicts can you resolve them pls

alex-mpoet:
@Aschepikov Rebased. The failing acceptance job fails on `master` too.

Aschepikov (approved):
gj!

PR #5778, opened 2024-08-07
Improve form validation [MAILPOET-6182]

NeosinneR:
## Description

1. Change the error message for checkbox and radio blocks from `Please select at least one option`  to `This field is required.`
    - The existing error is misleading when there is only a single checkbox to select (e.g., to agree to Terms and conditions) or for radio buttons because, in radio, you can select only one option, not multiple.
2. Add the required message (`This field is required.`) to the reCAPTCHA v2 Checkbox.
3. Fix errors message position on custom fields when rendering the same form on the page multiple times.

**Screenshot for the wrong behavior in point 3.**
Screenshot 2024-08-06 at 16 49 58

## Code review notes

There is a new randomly generated `validation_id`, so even the same block, when rendered multiple times, has a different ID for the error container. When for some reason it's not set, it will fall back to the previous implementation of using `{$blockId}_{$formId}`.

## QA notes

There are a couple of things to test:

1. Add two forms on the page (easiest is via a shortcode - just add two) that include some custom fields (checkbox, radio, or dates) and submit the invalid form to see, that error messages are shown in the submitted form and not in the other one.
2. When `reCAPTCHA v2 Checkbox` is enabled, check that the form can't be submitted and you see a required error below reCAPTCHA.
3. When any other CAPTCHA is active, check that the form can be submitted when valid. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6182](https://mailpoet.atlassian.net/browse/MAILPOET-6182) STOMAIL-7314

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6182]: https://mailpoet.atlassian.net/browse/MAILPOET-6182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/public.tsx
+++ b/mailpoet/assets/js/src/public.tsx
@@ -435,6 +435,24 @@ jQuery(($) => {
     // setup form validation
     $('form.mailpoet_form').each((_, element) => {
       const form = $(element);
+
+      form
+        .parsley()
+        .on('form:validate', (formInstance: { validationResult: boolean }) => {
+          const reCaptcha = form.find('.mailpoet_recaptcha');
+          const isReCatpchaVisible =
+            reCaptcha.length &&
+            reCaptcha.first().attr('data-size') !== 'invisible';
+          if (isReCatpchaVisible) {
+            if (window.grecaptcha.getResponse() === '') {
+              // eslint-disable-next-line no-param-reassign
+              formInstance.validationResult = false;
+              form.find('.mailpoet_error_recaptcha').addClass('filled');
+            } else {
+              form.find('.mailpoet_error_recaptcha').removeClass('filled');
+            }
+          }
+        });
       // Detect form is placed in tight container
       form.parsley().on('form:validated', () => {
         // clear messages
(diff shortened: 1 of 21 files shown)


mailpoet/assets/js/src/public.tsx
```diff
@@ -435,6 +435,24 @@ jQuery(($) => {
     // setup form validation
     $('form.mailpoet_form').each((_, element) => {
       const form = $(element);
+
+      form
+        .parsley()
+        .on('form:validate', (formInstance: { validationResult: boolean }) => {
+          const reCaptcha = form.find('.mailpoet_recaptcha');
+          const isReCatpchaVisible =
```
alex-mpoet, line 443: Typo: isReCatpchaVisible => isReCaptchaVisible.

alex-mpoet (approved):
Looks good! There is a typo, but it's in a local scope variable, so not a big deal. Moving to QA.

Aschepikov (approved):
GJ!

PR #5784, opened 2024-08-12
Doctrine to WPDB - timezone fixes [MAILPOET-6142]

costasovo:
## Description

This PR fixes issues related to timezones. With the WPDB refactor, we can't no longer rely on the timezone set on the DB connection. This change was causing a number of issues.

1. We started presenting times in listings in GMT
2. Scheduling was using incorrect times

During fixing I found that we use a couple of different approaches so I unified those as outlined [in the related P2 post](https://mailpoetdev.wordpress.com/2024/08/07/rfc-timezones-in-the-plugin-unification/).

The rules are simple.

- In PHP and DB we store times in GMT
- When presenting times to users or when we want them to enter time we expect site's timezone

## Code review notes

Please review commit by commit. The changes and motivations are described in the commit message or (I hope) are obvious.

## QA notes

This will be merged with the WPDB branch. I'll update testing instructions there. 

## Linked PRs

_N/A_

## Linked tickets

 [MAILPOET-6142]

## After-merge notes

_N/A_

## Tasks

- [x] ⛔ I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes - I was doing some batch refactors with IDE so I touched some jsx files but didn't rewrite them to Typescript.

### DIFF
--- a/mailpoet/assets/js/src/common/listings/_stories/newsletter-status.tsx
+++ b/mailpoet/assets/js/src/common/listings/_stories/newsletter-status.tsx
@@ -24,20 +24,24 @@ export function NewsletterStatuses() {
       <div className="mailpoet-gap" />
 
       <Heading level={3}>Scheduled in the future</Heading>
-      <NewsletterStatus scheduledFor={inOneHour} />
-      <NewsletterStatus scheduledFor={inSixHours} />
-      <NewsletterStatus scheduledFor={inTwelveHours} />
-      <NewsletterStatus scheduledFor={inOneDay} />
-      <NewsletterStatus scheduledFor={inOneDay} isPaused />
+      <NewsletterStatus scheduledFor={inOneHour.toISOString()} />
+      <NewsletterStatus scheduledFor={inSixHours.toISOString()} />
+      <NewsletterStatus scheduledFor={inTwelveHours.toISOString()} />
+      <NewsletterStatus scheduledFor={inOneDay.toISOString()} />
+      <NewsletterStatus scheduledFor={inOneDay.toISOString()} isPaused />
 
       <div className="mailpoet-gap" />
 
       <Heading level={3}>Sending in progress</Heading>
       <NewsletterStatus total={200} processed={0} />
       <NewsletterStatus total={400} processed={150} />
-      <NewsletterStatus scheduledFor={inPast} total={300} processed={270} />
       <NewsletterStatus
-        scheduledFor={inPast}
+        scheduledFor={inPast.toISOString()}
+        total={300}
+        processed={270}
+      />
+      <NewsletterStatus
+        scheduledFor={inPast.toISOString()}
         total={300}
         processed={270}
         isPaused
--- a/mailpoet/assets/js/src/date.ts
+++ b/mailpoet/assets/js/src/date.ts
@@ -27,6 +27,8 @@ export const MailPoetDate: {
   isInFutureGmt: (dateString: string, currentTime: MomentInput) => boolean;
   isInPast: (dateString: string, currentTime: MomentInput) => boolean;
   isInPastGmt: (dateString: string, currentTime: MomentInput) => boolean;
+  gmtToSiteTimestamp: (utcTimeStamp: string) => string;
+  siteToGmtTimestamp: (siteTimeStamp: string) => string;
   adjustForTimezoneDifference: (date: Date) => Date;
 } = {
   version: 0.1,
@@ -66,7 +68,6 @@ export const MailPoetDate: {
   format: function format(date: MomentInput, opts?: DateOptions): string {
     const options = opts || {};
     this.init(options);
-
     const momentDate = Moment(date);
     return momentDate.format(this.convertFormat(this.options.format) as string);
   },
@@ -118,6 +119,19 @@ export const MailPoetDate: {
   time: function time(date: MomentInput): string {
     return this.time(date, {
       format: window.mailpoet_time_format || 'H:i:s',
+    });
+  },
+  gmtToSiteTimestamp: function gmtToSiteTimestamp(
+    gmtTimeStamp: string,
+  ): string {
+    return this.formatFromGmt(gmtTimeStamp, { format: 'Y-m-d H:i:s' });
+  },
+  siteToGmtTimestamp: function siteToGmtTimestamp(
+    gmtTimeStamp: string,
+  ): string {
+    return this.formatFromGmt(gmtTimeStamp, {
+      format: 'Y-m-d H:i:s',
+      offset: parseFloat(window.mailpoet_date_offset || '0') * -1,
     });
   },
   convertFormat: function convertFormat(format: string): string {
--- a/mailpoet/assets/js/src/date.ts
+++ b/mailpoet/assets/js/src/date.ts
@@ -30,7 +30,6 @@ export const MailPoetDate: {
   isInPastGmt: (dateString: string, currentTime: MomentInput) => boolean;
   gmtToSiteTimestamp: (utcTimeStamp: string) => string;
   siteToGmtTimestamp: (siteTimeStamp: string) => string;
-  adjustForTimezoneDifference: (date: Date) => Date;
 } = {
   version: 0.1,
   options: {},
@@ -245,23 +244,4 @@ export const MailPoetDate: {
     Moment(dateString).isBefore(currentTime, 's'),
   isInPastGmt: (dateString: string, currentTime: MomentInput): boolean =>
     Moment.utc(dateString).isBefore(currentTime, 's'),
-  adjustForTimezoneDifference: function adjustForTimezoneDifference(
-    date: Date,
-  ): Date {
-    // PHP offset is the same as timezone, e.g., UTC-2 is -120 (minutes)
-    const serverOffsetMinutes = window.mailpoet_server_timezone_in_minutes || 0;
-    // JS offset is the opposite, e.g., UTC-2 is +120 (minutes)
-    const browserOffsetMinutes = new Date().getTimezoneOffset();
-    // Because of this different representation, we can just sum these two
-    // E.g., server UTC-2, browser UTC+2: -120 + -120 = -240 minuts difference
-    // E.g., server UTC+1, browser UTC+2: +60 + -120 = -60 minutes difference
-    const offsetDifference = serverOffsetMinutes + browserOffsetMinutes;
-    if (!offsetDifference) {
-      return date;
-    }
-    // Because the difference is calculated from browser to server, we need to subtract
-    // the difference to adjust server time to browser's time zone
-    date.setMinutes(date.getMinutes() - offsetDifference);
-    return date;
-  },
 } as const;
--- a/mailpoet/assets/js/src/date.ts
+++ b/mailpoet/assets/js/src/date.ts
@@ -21,8 +21,8 @@ export const MailPoetDate: {
   convertFormat: (format: string) => string;
   isInFuture: (dateString: string, currentTime: MomentInput) => boolean;
   isInPast: (dateString: string, currentTime: MomentInput) => boolean;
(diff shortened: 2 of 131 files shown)


mailpoet/lib/Newsletter/Sending/ScheduledTasksRepository.php
```diff
@@ -311,7 +307,7 @@ public function findByTypeAndSubscriber(string $type, SubscriberEntity $subscrib
   }
 
   public function touchAllByIds(array $ids): void {
-    $now = CarbonImmutable::createFromTimestamp((int)$this->wp->currentTime('timestamp'));
+    $now = CarbonImmutable::now();
```
JanJakes: This one doesn't have the `millisecond(0)` part. Should we add it, for consistency?

mailpoet/assets/js/src/date.ts
```diff
@@ -14,12 +14,19 @@ export const MailPoetDate: {
...
+  shortFromGmt: (date: MomentInput) => string;
   full: (date: MomentInput) => string;
+  fullFromGmt: (date: MomentInput) => string;
   time: (date: MomentInput) => string;
+  timeFromGmt: (date: MomentInput) => string;
   convertFormat: (format: string) => string;
   isInFuture: (dateString: string, currentTime: MomentInput) => boolean;
+  isInFutureGmt: (dateString: string, currentTime: MomentInput) => boolean;
+  isInPast: (dateString: string, currentTime: MomentInput) => boolean;
+  isInPastGmt: (dateString: string, currentTime: MomentInput) => boolean;
```
JanJakes: Let's keep only one variant here (the new one) to avoid confusion when using the Date object.

mailpoet/assets/js/src/date.ts
```diff
@@ -120,6 +121,19 @@ export const MailPoetDate: {
...
+    return this.formatFromGmt(gmtTimeStamp, { format: 'Y-m-d H:i:s' });
+  },
+  siteToGmtTimestamp: function siteToGmtTimestamp(
+    gmtTimeStamp: string,
+  ): string {
+    return this.formatFromGmt(gmtTimeStamp, {
+      format: 'Y-m-d H:i:s',
+      offset: parseFloat(window.mailpoet_date_offset || '0') * -1,
+    });
+  },
```
JanJakes, line 100: As both the input and output seem to be strings and I guess they both are `Y-m-d H:i:s`, I'm not sure if the `timestamp` in the name makes sense. Maybe we could use a `string` suffix, or sth like `datetime`.

Actually, I think the first method should follow the naming of the other methods like `fullFromGmt` (and after simplifying to only one variant, it would be only `full`), so I guess that format needs a name. We could name it `datetimeString`, maybe, to avoid confusion with datetime objects.

As for the second method, I think we first need a generic counterpart to `formatFromGmt` (~ `format`). Something like `formatToGmt`, maybe. Then, if the first method is named `datetimeString`, then the second one could be called `toGmtDatetimeString`, or something like that. And I suppose both could accept any date input, not just strings, to align with the rest of the API.

mailpoet/assets/js/src/newsletters/send/standard.tsx
```diff
@@ -85,7 +83,8 @@ class StandardScheduling extends Component<StandardSchedulingProps> {
       schedulingOptions = (
         <>
           <span className="mailpoet-form-schedule-time">
-            {__('Your website’s time is', 'mailpoet')} {currentTime}
+            {__('Your website’s time is', 'mailpoet')}{' '}
+            {MailPoet.Date.timeFromGmt(new Date().toISOString())}
```
JanJakes: We've tested that this should work correctly. Is the `toISOString()` part needed?

JanJakes: Oh, and seeing `Your website’s time is`, I wonder if it would make sense to add such information to somewhere to Help and maybe also the timezone offset. That might be out of the scope of this PR, however.

mailpoet/assets/js/src/newsletters/send/standard.tsx
```diff
@@ -239,13 +238,12 @@ export const StandardNewsletterFields = {
...
   ): SendButtonOptions => {
-    const currentDateTime = Moment(window.mailpoet_current_date_time);
     const isScheduled =
       typeof newsletter.options === 'object' &&
       newsletter.options?.isScheduled === '1' &&
-      MailPoet.Date.isInFuture(
+      MailPoet.Date.isInFutureGmt(
         newsletter.options?.scheduledAt,
-        currentDateTime,
+        new Date().toISOString(),
```
JanJakes: Is the `toISOString()` needed here?

mailpoet/assets/js/src/common/listings/_stories/newsletter-status.tsx
```diff
@@ -24,20 +24,24 @@ export function NewsletterStatuses() {
...
-      <NewsletterStatus scheduledFor={inOneHour} />
-      <NewsletterStatus scheduledFor={inSixHours} />
-      <NewsletterStatus scheduledFor={inTwelveHours} />
-      <NewsletterStatus scheduledFor={inOneDay} />
-      <NewsletterStatus scheduledFor={inOneDay} isPaused />
+      <NewsletterStatus scheduledFor={inOneHour.toISOString()} />
+      <NewsletterStatus scheduledFor={inSixHours.toISOString()} />
+      <NewsletterStatus scheduledFor={inTwelveHours.toISOString()} />
+      <NewsletterStatus scheduledFor={inOneDay.toISOString()} />
+      <NewsletterStatus scheduledFor={inOneDay.toISOString()} isPaused />
```
JanJakes, line 31: Actually, I'm realizing that my question applies more generally — is `.toISOString()` needed anywhere at all? If the new datetime methods accept also date objects and take their own timezone into consideration rather than stripping it, then this should not be necessary.

mailpoet/lib/Newsletter/Scheduler/Scheduler.php
```diff
@@ -33,11 +33,16 @@ public function getNextRunDate($schedule, $fromTimestamp = false) {
   }
 
   public function getPreviousRunDate($schedule, $fromTimestamp = false) {
-    $fromTimestamp = ($fromTimestamp) ?: $this->wp->currentTime('timestamp', true);
+    // User enters time in WordPress site timezone, but we need to calculate it in UTC before we save it to DB
+    // 1) As the initial tile we use time in site timezone (gmt is false)
```
JanJakes: `tile` is probably a typo

mailpoet/lib/Newsletter/Scheduler/Scheduler.php
```diff
@@ -33,11 +33,16 @@ public function getNextRunDate($schedule, $fromTimestamp = false) {
...
+    // 2) We use CronExpression to calculate previous run (still in site's timezone)
+    // 3) We convert the calculated time to UTC
+    $fromTimestamp = ($fromTimestamp) ?: $this->wp->currentTime('timestamp', false);
```
JanJakes: I suppose this whole expression could be replaced by a more understandable `$from = $this->wp->currentDatetime()` that would call WP's `current_datetime` function.

mailpoet/lib/Newsletter/Scheduler/Scheduler.php
```diff
@@ -82,10 +87,15 @@ public function formatDatetimeString($datetimeString) {
    * @return \DateTime|false
    */
   public function getNextRunDateTime($schedule, $fromTimestamp = false) {
-    $fromTimestamp = $fromTimestamp ?: $this->wp->currentTime('timestamp', true);
+    // User enters time in WordPress site timezone, but we need to calculate it in UTC before we save it to DB
+    // 1) As the initial tile we use time in site timezone (gmt is false)
+    // 2) We use CronExpression to calculate next run (still in site's timezone)
+    // 3) We convert the calculated time to UTC
+    $fromTimestamp = $fromTimestamp ?: $this->wp->currentTime('timestamp', false);
```
JanJakes: Same applies as on lines 37 & 40.

mailpoet/lib/Doctrine/WPDB/Connection.php
```diff
@@ -18,6 +18,11 @@ public function __construct() {
     if (!$wpdb instanceof wpdb) {
       throw new ConnectionException('WPDB is not initialized.');
     }
+    /**
+     * Ensure connection is in UTC timezone.
+     * We use timestamp columns and we expect to get UTC time
+     */
+    $wpdb->query("SET time_zone = '+00:00'");
```
JanJakes: I think that currently the WPDB driver doesn't include much MailPoet-specific functionality. This line seems more specific, so I'm just wondering if it could be handled [via a DBAL event](https://www.doctrine-project.org/projects/doctrine-dbal/en/3.1/reference/events.html#postconnect-event), just to keep it separated. It would also go through the proper `query()` error handling. Probably not a big deal, just an idea.

JanJakes (commented):
This is a great piece of work, thanks for handling this! 👏

mailpoet/assets/js/src/common/listings/_stories/newsletter-status.tsx
costasovo: In this case, this is just for the story book. The newsletter status component newly accepts only string value (as it used to in the past) and handles dates formating internally. The .toISOString was the simplest way how to fix typescript warnings comming from the stories.

mailpoet/assets/js/src/newsletters/send/standard.tsx
costasovo: Website time is visible in WP's settings, but we didn't have WP's timezone in the system report (also sent to the support), so I added it there.

costasovo:
@JanJakes Thank you for the detailed review and the suggestions; they improved the final outcome. I believe I addressed all of them. I addressed them in 8 new commits. Could you please take another look?

mailpoet/lib/Doctrine/Middlewares/PostConnectMiddleware.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace MailPoet\Doctrine\Middlewares;
+
+use MailPoetVendor\Doctrine\DBAL\Driver\Connection;
+use MailPoetVendor\Doctrine\DBAL\Driver\Middleware\AbstractDriverMiddleware;
+
+class PostConnectMiddleware extends AbstractDriverMiddleware {
+  public function connect(array $params): Connection {
+    $connection = parent::connect($params);
+    $res = $connection->exec('SET time_zone = "+00:00"');
```
JanJakes: `$res` seems to be unused (but no CI checks are complaining).

mailpoet/assets/js/src/date.ts
```diff
@@ -21,8 +21,8 @@ export const MailPoetDate: {
   convertFormat: (format: string) => string;
   isInFuture: (dateString: string, currentTime: MomentInput) => boolean;
   isInPast: (dateString: string, currentTime: MomentInput) => boolean;
-  gmtToSiteTimestamp: (utcTimeStamp: string) => string;
-  siteToGmtTimestamp: (siteTimeStamp: string) => string;
+  datetimeString: (utcTimeStamp: string) => string;
+  toGmtDatetimeString: (siteTimeStamp: string) => string;
```
JanJakes: Probably forgotten, I suppose here the parameters should be `date: MomentInput`.

JanJakes (commented):
Great, thanks! Just two tiny comments.

mailpoet/lib/Doctrine/Middlewares/PostConnectMiddleware.php
costasovo: You're right; I used it to debug the return value. But strange there are no warnings

PR #5785, opened 2024-08-12
Display a notice when some MailPoet tables don't use the InnoDB engine [MAILPOET-6184]

pavel-mailpoet:
## Description

Some users have been running into performance issues that turned out to be caused by MyISAM table engine. We can detect if this is happening and display a notice to the user.

## Code review notes

_N/A_

## QA notes

You can change the table engines using a plugin: https://wordpress.org/plugins/wp-phpmyadmin-extension/ and then test if you see the notice

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6184]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6184]: https://mailpoet.atlassian.net/browse/MAILPOET-6184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Util/Notices/DatabaseEngineNotice.php
+++ b/mailpoet/lib/Util/Notices/DatabaseEngineNotice.php
@@ -71,7 +71,7 @@ class DatabaseEngineNotice {
       fn($row) => $row['Name'],
       array_filter(
         $data,
-        fn($row) => is_string($row['Engine']) && (strtolower($row['Engine']) !== 'innodb')
+        fn($row) => isset($row['Engine']) && is_string($row['Engine']) && (strtolower($row['Engine']) !== 'innodb')
       )
     );
   }
--- a/mailpoet/lib/Util/Notices/DatabaseEngineNotice.php
+++ b/mailpoet/lib/Util/Notices/DatabaseEngineNotice.php
@@ -1,0 +1,36 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Util\Notices;
+
+use MailPoet\WP\Functions as WPFunctions;
+use MailPoet\WP\Notice;
+
+class DatabaseEngineNotice {
+  const OPTION_NAME = 'database-engine-notice';
+  const DISMISS_NOTICE_TIMEOUT_SECONDS = 15_552_000; // 6 months
+
+  /** @var WPFunctions */
+  private $wp;
+
+  public function __construct(
+    WPFunctions $wp
+  ) {
+    $this->wp = $wp;
+  }
+
+  public function init($shouldDisplay): ?Notice {
+    if (!$shouldDisplay || $this->wp->getTransient(self::OPTION_NAME)) {
+      return null;
+    }
+
+    return $this->display();
+  }
+
+  private function display(): ?Notice {
+    return null;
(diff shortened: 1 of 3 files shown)


mailpoet/lib/Util/Notices/DatabaseEngineNotice.php
```diff
@@ -54,12 +67,12 @@ private function checkTableEngines(): array {
       fn($row) => $row['Name'],
       array_filter(
         $data,
-        fn($row) => strtolower($row['Engine']) !== 'innodb'
+        fn($row) => is_string($row['Engine']) && (strtolower($row['Engine']) !== 'innodb')
```
alex-mpoet: Can there be cases when the `Engine` key is not defined? Then it will generate a warning.

mailpoet/lib/Util/Notices/DatabaseEngineNotice.php
```diff
@@ -54,12 +67,12 @@ private function checkTableEngines(): array {
       fn($row) => $row['Name'],
       array_filter(
         $data,
```
alex-mpoet, line 73: Should we ensure that data is an array before using it? Or will `fetchAllAssociative()` always return an array even if the query fails? By the way, what will happen if it fails?

alex-mpoet (commented):
Looks good to me overall! 

The only thing I'm not sure about is the cache timeout. The database engines for tables do not change every day, rather on a very rare occasion, so even checking once a week could suffice. One the other hand, most users will probably see the notice right away after it's included in the plugin, so if they decide to fix the engines, the notice will stay there until the cache timeout passes, so waiting a day for it to go away is better than waiting a week or more.

I also left a couple smaller notes.

mailpoet/lib/Util/Notices/DatabaseEngineNotice.php
pavel-mailpoet: `fetchAllAssociative()` always returns an array, so we don't need a check here. I added a try ... catch if the query fails, but if a db query fails here, there is a bigger problem.

mailpoet/lib/Util/Notices/DatabaseEngineNotice.php
pavel-mailpoet: I don't know what to do if the key is not defined. Because if the query is misbehaving, a warning could be a good option. However, I added a check and if the key is not defined the user will not see the warning.

pavel-mailpoet:
> The only thing I'm not sure about is the cache timeout. The database engines for tables do not change every day, rather on a very rare occasion, so even checking once a week could suffice. One the other hand, most users will probably see the notice right away after it's included in the plugin, so if they decide to fix the engines, the notice will stay there until the cache timeout passes, so waiting a day for it to go away is better than waiting a week or more.

Yeah, you are right. From those two options, I think one day is a good compromise and the query is not that heavy so we can keep it once a day.

alex-mpoet (approved):

Aschepikov (approved):
GJ!

PR #5788, opened 2024-08-14
Prevent adding subscriber with empty email [MAILPOET-6188]

costasovo:
## Description

This fixes an error caught by logs. We don't allow to save a subscriber with an empty email, and it seems that in some cases, there are attempts to save such subscriber when a new user registers on a site.

## Code review notes

_N/A_

## QA notes

We can skip QA for this one.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6188]

## After-merge notes

_N/A_

## Tasks

- [x] ⛔ I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] I added sufficient test coverage.
- [x] ⛔ I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-6188]: https://mailpoet.atlassian.net/browse/MAILPOET-6188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Subscription/Registration.php
+++ b/mailpoet/lib/Subscription/Registration.php
@@ -68,6 +68,7 @@ class Registration {
       if (
         isset($_POST['mailpoet']['subscribe_on_register'])
         && (bool)$_POST['mailpoet']['subscribe_on_register'] === true
+        && !empty($result['user_email'])
       ) {
         $this->subscribeNewUser(
           $result['user_name'],
--- a/mailpoet/lib/Subscription/Registration.php
+++ b/mailpoet/lib/Subscription/Registration.php
@@ -87,6 +87,7 @@ class Registration {
       empty($errors->errors)
       && isset($_POST['mailpoet']['subscribe_on_register'])
       && (bool)$_POST['mailpoet']['subscribe_on_register'] === true
+      && !empty($userEmail)
     ) {
       $this->subscribeNewUser(
         $userLogin,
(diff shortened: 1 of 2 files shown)


mailpoet/lib/Subscription/Registration.php
```diff
@@ -87,6 +87,7 @@ public function onRegister(
       empty($errors->errors)
       && isset($_POST['mailpoet']['subscribe_on_register'])
       && (bool)$_POST['mailpoet']['subscribe_on_register'] === true
+      && !empty($userEmail)
```
pavel-mailpoet, line 91: I am confused here about the method signature, it says that the email argument is optional. So, is there a valid scenario when the email here can be null?

Also, should we do this check in the previous method `onMultiSiteRegister` as well?

costasovo: > I am confused here about the method signature, it says that the email argument is optional. So, is there a valid scenario when the email here can be null?

@pavel-mailpoet I was not able to find why exactly this parameter can be skipped. It is a very old code. I see that we [register the callback on `registration_erros` filter](https://github.com/mailpoet/mailpoet/blob/59733db6deb51dbf9563af4bacea0f569e8ad474/mailpoet/lib/Config/Hooks.php#L208-L214) so perhaps in some cases in may come null when there are errors. But the Fatal error happens when there are no registration errors but the email was an empty string. That's what I see in the stack trace.

I added the check also to the onMultisiteRegister.

pavel-mailpoet (approved):

PR #5790, opened 2024-08-15
Fix dual buttons in Outlook [MAILPOET-6190]

costasovo:
## Description

This PR fixes an issue introduced recently in https://github.com/mailpoet/mailpoet/pull/5745.
The issue doesn't happen when sending a preview email, but it is reproducible when sending it to a list. 

## Code review notes

I found that the issue was caused by a comment in button output `<!--[if !mso]><!-- -->` was being replaced with `<!--[if !mso]><![endif]-->`

This was a known issue when parsing HTML via pQuery and we had a fix for that in the Renderer class. I found that that we use pQuery in more places in processing email during sending e.g. in code that adds GA tracking parameters.

I fixed it by moving the fix directly to pQuery's toString method.

#### Replication

1. Create a newsletter with a button
2. Make sure GA tracking is enabled
3. Send the email via send page (note sending preview is not enough for replication)
4. Observe the issue in Outlook (you can use Litmus for checking)

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6190]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6190]: https://mailpoet.atlassian.net/browse/MAILPOET-6190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib-3rd-party/pquery/gan_node_html.php
+++ b/mailpoet/lib-3rd-party/pquery/gan_node_html.php
@@ -271,7 +271,9 @@ class DomNode implements IQuery {
 	 * @access private
 	 */
 	function __toString() {
-		return (($this->tag === '~root~') ? $this->toString(true, true, 1) : $this->tag);
+		$string = (($this->tag === '~root~') ? $this->toString(true, true, 1) : $this->tag);
+		// because tburry/pquery contains a bug and replaces the opening non mso condition incorrectly we have to replace the opening tag with correct value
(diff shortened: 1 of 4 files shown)


pavel-mailpoet (approved):

Aschepikov (approved):
GJ!

PR #5716, opened 2024-07-05
Replace interact.js with interactjs [MAILPOET-6109]

NeosinneR:
## Description

`interact.js` package was renamed to `interactjs` (the same author in the [same repo](https://github.com/taye/interact.js)), but there were API changes. The most notable one for us is that `event.interaction.element` becomes a read-only, so it can't be assigned a clone. 

There is [no direct API to support clones](https://interactjs.io/docs/faq/#clone-target-draggable), and the [recommendation](https://interactjs.io/docs/faq/#clone-target-draggable) is not as straightforward to implement in our case, so I decided to do a simple workaround with `event.target.__clone`. 

## Code review notes

I decided not to convert those files to TypeScript to prevent introducing any new bugs into the old codebase, which we try not to touch. 

## QA notes

Please test that resizing, dragging, and dropping in the email editor works and that no console errors are logged. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6109]

## After-merge notes

_N/A_

## Tasks

- [x] 🚫 I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] 🚫 I added sufficient test coverage
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-6109]: https://mailpoet.atlassian.net/browse/MAILPOET-6109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5792, opened 2024-08-19
Update axios [MAILPOET-6191]

costasovo:
## Description

This PR addresses the Dependabot's report https://github.com/mailpoet/mailpoet/security/dependabot/107
The axios is used by `@wordpress/scripts` 

## Code review notes

- I updated the required pnpm version (`corepack up`). It looks like some of us have been already using a newer version than the one set in package.json. The newer version writes the version for overridden packages directly to the lock file. So, running even a little update in the older version was causing many changes in the lockfile ([example - this was actually from older to newer](https://github.com/mailpoet/mailpoet/pull/5716/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bb)). 
- I updated axios
- I ran `pnpm dedupe` after the update. The `pnpm dedupe` updated a couple of minor versions, and Typescript
- I downgraded Typescript back to the current version because the new version requires more updates and configuration changes in eslint. See more info in commit message.

## QA notes

We can skip QA

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6191]

## After-merge notes

Ping #mailpoetdev 

> Fyi, we updated the required pnpm version for plugins to 8.15.9. Note that it is recommended to use Corepack to manage the version automatically (run `corepack enable`). 
## Tasks

- [x] 🚫  I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] 🚫  I added sufficient test coverage
- [x] 🚫  I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-6191]: https://mailpoet.atlassian.net/browse/MAILPOET-6191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/segments/dynamic/types.ts
+++ b/mailpoet/assets/js/src/segments/dynamic/types.ts
@@ -183,6 +183,7 @@ export interface EmailFormItem extends FormItem {
   opens?: string;
   emails?: string;
   clicks?: string;
+  timeframe?: Timeframe;
 }
 
 export type DynamicSegment = {
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "prettier": "2.6.2",
     "typescript": "^5.0.2"
   },
-  "packageManager": "pnpm@8.5.1",
+  "packageManager": "pnpm@8.15.9",
   "volta": {
     "node": "19.7.0"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1017,6 +1017,11 @@ packages:
     engines: {node: '>=6.9.0'}
     dev: true
 
+  /@babel/helper-plugin-utils@7.24.8:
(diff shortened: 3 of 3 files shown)


mailpoet/assets/js/src/segments/dynamic/types.ts
```diff
@@ -183,6 +183,7 @@ export interface EmailFormItem extends FormItem {
   opens?: string;
   emails?: string;
   clicks?: string;
+  timeframe?: Timeframe;
```
JanJakes, line 186: Why was this change needed when we seem to keep the same TS version?

JanJakes (approved):
LGTM, thanks!

mailpoet/assets/js/src/segments/dynamic/types.ts
costasovo: I added it after the update because it was causing error with the updated TS version. I found that this prop was indeed missing in the definition but with the current TS version there is no error. I didn't revert the change as this is an improvement. So it was actually not needed it and I kept it kind of for the future. Sorry for the confusion.

PR #5801, opened 2024-08-22
QIT fixes [MAILPOET-6150]

lysyjan:
## Description

_N/A_

## Code review notes

- I used the placeholder `%i`, but I had to remove a check about [MinimalPluginStandard](https://github.com/WordPress/wporg-code-analysis/tree/trunk/MinimalPluginStandard) because it is outdated and doesn't know this new placeholder.
- Sometimes, I removed using a variable for a query to avoid more comments.
- To avoid errors in preparing queries, I had to extract `$wpdb` to a variable because when it was used directly from a property, the tool reported it as an error.
- I tried to minimize the number of needed ignore comments, but sometimes, I didn't find a way to do it. For example, when the query contains a variable that is already prepared.

## QA notes

If the review passes, we can skip QA here because the changes are cosmetic, and the automated tests pass.

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/904

## Linked tickets

[MAILPOET-6150]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6150]: https://mailpoet.atlassian.net/browse/MAILPOET-6150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JanJakes:
@lysyjan Before diving deeper into this, I think we actually can use table placeholders with WPDB. It's the `%i` placeholder, see also the docs: https://developer.wordpress.org/reference/classes/wpdb/prepare/

It would be probably nicer to use this rather than adding the ignore comments.

lysyjan:
@JanJakes 🤦 You are right. I didn't know this, and it was not mentioned in the sources that I checked. Of course, I will update it.

JanJakes:
@lysyjan It was actually added only in WP 6.2, that's why not so long ago, we weren't able to use that indeed.

JanJakes (commented):
> but I had to remove a check about [MinimalPluginStandard](https://github.com/WordPress/wporg-code-analysis/tree/trunk/MinimalPluginStandard) because it is outdated and doesn't know this new placeholder

@lysyjan Do you think it's fine to just remove it? Won't we lose all the .org checks this way? What was your reasoning regarding the removal? Do you think/know that QIT covers all of these? Checking but I had to remove a check about [MinimalPluginStandard](https://github.com/WordPress/wporg-code-analysis/tree/trunk/MinimalPluginStandard) because it is outdated and doesn't know this new placeholder, there seem to be some solid reasons for these checks.

JanJakes:
@lysyjan So I think understand what likely happened. WPCS was updated and WPORG removed, which hints that you were thinking that we're somehow using the WPCS rules even without WPORG. But we're not, it was the WPORG `MinimalPluginStandard` ruleset that was using some (mainly security-related) WPCS rules + some of its own. Removing the WPORG checks also removes all usage of WPCS rules, which is a large and unwanted side effect.

To address that, I aliased WPCS 3.1 to 2.99.99 for the WPORG ruleset to install and added a few patches for compatibility. Since the repo seems rather abandoned, we should probably incorporate WPCS rules into our ruleset directly in the future.

JanJakes:
Reviewing further, I noticed that only some cases were rewritten to inline queries in `prepare()`, while others were left out with comments. I pushed this further and removed as much ignore comments as possible.

I also updated this to introduce WPCS (as per P2) and properly remove MinimalPluginStandard. My changes start at https://github.com/mailpoet/mailpoet/pull/5801/commits/3e33bccb09058c98263e7279fe6b308c716c8303.

NeosinneR (approved):
Nice work, @lysyjan and @JanJakes. I find many methods and queries more readable now, after the fixes.

PR #5803, opened 2024-08-23
Fix schedule date in calendar on the send page [MAILPOET-6197]

costasovo:
## Description

This PR fixes an issue in scheduling the calendar on the send page. In timezones that are far from UTC we display incorrect date.

## Code review notes

I tried to explain the issue in the commit message.

## QA notes

# Replication

1. Set your site timezone to Los Angels timezone (UTC-7)
3. Set your browser timezone to the same timezone ([here is how to do it in dev tools](https://www.browserstack.com/guide/change-time-zone-in-chrome-for-testing))
5. Create a newsletter and schedule it close to midnight
7. Observe the issue in the calendar and in the listing

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6197]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6197]: https://mailpoet.atlassian.net/browse/MAILPOET-6197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/newsletters/send/date-text.tsx
+++ b/mailpoet/assets/js/src/newsletters/send/date-text.tsx
@@ -1,10 +1,10 @@
 import { Component, SyntheticEvent } from 'react';
+import Moment from 'moment';
 import { __, _x } from '@wordpress/i18n';
 import { registerLocale } from 'react-datepicker';
 import locale from 'date-fns/locale/en-US';
 import { Datepicker } from 'common/datepicker/datepicker';
 import { MailPoet } from 'mailpoet';
-import { DateOptions } from 'date';
 
 /**
  * This function is a copy of the buildLocalizeFn function from date-fns (date-fns/locale/_lib/buildLocalizeFn)
@@ -134,10 +134,10 @@ class DateText extends Component<DateTextProps> {
   onChange = (value: Date, event) => {
     const changeEvent: DateTextEvent = event;
     // Swap display format to storage format
-    const storageDate = this.getStorageDate(value);
+    const formattedDate = this.getAsStringInFormat(value);
 
     changeEvent.target.name = this.getFieldName();
-    changeEvent.target.value = storageDate;
+    changeEvent.target.value = formattedDate;
     this.props.onChange(changeEvent);
   };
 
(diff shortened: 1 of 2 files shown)


mailpoet/assets/js/src/newsletters/send/date-text.tsx
```diff
@@ -155,16 +155,12 @@ class DateText extends Component<DateTextProps> {
...
-    } as DateOptions;
-    return MailPoet.Date.toDate(date, formatting);
-  };
+  getDate = (date: string) => Moment(date).toDate();
 
-  getStorageDate = (date: Date) => {
+  getAsStringInFormat = (date: Date) => {
     const formatting = {
       format: this.props.storageFormat,
+      offset: 0, // set 0 to prevent applying offset. At this point we want to return value exactly as picked in UI.
```
JanJakes: Checking the implementation:

Screenshot 2024-08-23 at 14 17 59



And the [Moment.js docs](https://momentjscom.readthedocs.io/en/latest/moment/03-manipulating/09-utc-offset/):

Screenshot 2024-08-23 at 14 18 45

It seems to me that this will preserve the point in time (Universal Time), but the display time will change:

> Passing false (the default) will keep the same instant in Universal Time, but the local time will change.

However, the comment in the code seems to suggest that it's not what you want:

> // set 0 to prevent applying offset. At this point we want to return value exactly as picked in UI.

So, that's a bit confusing to me. Maybe it's just me interpreting the comment wrong? What you're actually doing here is converting the time to UTC and then formatting that. Is that the desired behavior?

JanJakes: Ahh, maybe in combination with `Moment.utc(date)` ([link](https://momentjscom.readthedocs.io/en/latest/moment/01-parsing/13-utc/)), this has no effect and only prevents some potential automatic conversion?

costasovo: > Ahh, maybe in combination with Moment.utc(date) ([link](https://momentjscom.readthedocs.io/en/latest/moment/01-parsing/13-utc/)), this has no effect and only prevents some potential automatic conversion?

Yes. I was thinking about it like that, but I will rather double check.

JanJakes (approved):

PR #5777, opened 2024-08-06
Initial WordPress Playground support

JanJakes:
## Description

This PR implements initial WordPress Playground support, such that the build from this branch activates on https://playground.wordpress.net/.

Tracking issue for [WordPress Playground](https://wordpress.org/playground/) support.

Reported issues (✔️ when verified that it works for us):

- [x] https://github.com/WordPress/sqlite-database-integration/pull/129
- [x] https://github.com/WordPress/sqlite-database-integration/pull/130
- [x] https://github.com/WordPress/sqlite-database-integration/pull/134
- [x] https://github.com/WordPress/sqlite-database-integration/pull/142
- [x] https://github.com/WordPress/sqlite-database-integration/pull/145
- [x] https://github.com/WordPress/sqlite-database-integration/pull/147
- [x] https://github.com/WordPress/sqlite-database-integration/pull/150
- [x] https://github.com/WordPress/sqlite-database-integration/pull/156
- [x] https://github.com/WordPress/sqlite-database-integration/pull/158
- [x] https://github.com/WordPress/sqlite-database-integration/pull/151
- [x] https://github.com/WordPress/sqlite-database-integration/pull/154

These issues may take a bit longer to handle properly, but we have workarounds:

- [ ] https://github.com/WordPress/sqlite-database-integration/pull/152
- [ ] https://github.com/WordPress/sqlite-database-integration/issues/146

## Code review notes

I implemented some functionality that may be generally useful to our codebase:

1. Handle errors from PDO in the WPDB driver, in case WPDB runs on PDO (like [SQLite](https://github.com/WordPress/sqlite-database-integration/)).
2. A helper for detecting that we're running on SQLite.
3. New detection whether a table/column/index exists that doesn't rely on `information_schema`.

Additionally, I added some temporary hotfixes when running on SQLite so that the plugin can activate.

## QA notes

QA can be skipped, but the code reviewer can verify that the MailPoet build from this branch activates on https://playground.wordpress.net/.

## Linked tickets

[MAILPOET-6185]

[MAILPOET-6185]: https://mailpoet.atlassian.net/browse/MAILPOET-6185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5813, opened 2024-08-30
WordPress Playground & Studio improvements [MAILPOET-6185]

JanJakes:
## Description

Just a few more fixes for WordPress Playground & Studio, and a notice about this being a plugin preview.

Screenshot 2024-08-30 at 16 57 47

## Code review notes

Please, just see the commit messages. QA can be skipped, I think.

## Linked PRs

https://github.com/mailpoet/mailpoet/pull/5777

## Linked tickets

[MAILPOET-6185]

[MAILPOET-6185]: https://mailpoet.atlassian.net/browse/MAILPOET-6185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:wp-playground-improvements)

_The latest successful build from `wp-playground-improvements` will be used. If none is available, the link won't work._

### DIFF
--- a/.github/workflows/wordpress-playground-preview.yml
+++ b/.github/workflows/wordpress-playground-preview.yml
@@ -1,0 +1,27 @@
+name: Add link to WordPress Playground preview
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  append-description:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Check and append description
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          PR_NUMBER="${{ github.event.pull_request.number }}"
+          BRANCH_NAME="${{ github.head_ref }}"
+          DESCRIPTION="$(gh pr view $PR_NUMBER --json body -q .body)"
+          HEADING="## Preview"
+          CONTENT="$(printf "\n\n${HEADING}\n\n[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:${BRANCH_NAME})\n\n_The latest successful build from \`${BRANCH_NAME}\` will be used. If none is available, the link won't work._")"
+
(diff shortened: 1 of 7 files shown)


NeosinneR (commented):
Overall looks good, but what I'm not 100% aligned with is using `Connection::isSQLite()` directly in conditions, as over time, this might get lost about why it's there. 

@JanJakes what do you think about a simple `isWordPressPlayground()` method, that would just call `Connection::isSQLite()`? It would improve the readability and in future it's easier to update if it's no longer needed or we have a better way to detect Playground (e.g. via a constant).

JanJakes:
@NeosinneR I agree that the  `Connection::isSQLite()` should be reviewed at some point, but at the moment, it's probably the best test because it really kind of is about https://github.com/WordPress/sqlite-database-integration/, which is used by both Playground and Studio, and the Studio support is also a priority at the moment (https://github.com/Automattic/studio/issues/132). In the future, there might be differences between the Playground and Studio usage, etc. It will need some more thought, I guess.

I would now take it as a quick and a bit hacky way to get it off the ground, and later we surely should improve that.

NeosinneR (dismissed):

NeosinneR:
@JanJakes ah, I approved this already, but noticed, that the Playground link should probably include `branch:`?

JanJakes:
@NeosinneR

> @JanJakes ah, I approved this already, but noticed, that the Playground link should probably include `branch:`? 

🤦‍♂️ Yes, for sure. Let me quickly fix that. Btw. is it possible to change a branch name in the same PR? Just wondering if the logic of the GitHub action should be more of "create or update" than "create or ignore". That said, we can improve that later.

JanJakes:
@NeosinneR Repushed.

NeosinneR (approved):
> is it possible to change a branch name in the same PR?

I don't think so, or at least I wasn't able to find a way to do so. And I don't remember us doing that in the past, so even if possible, probably a rare case we don't have to accommodate.

PR #5818, opened 2024-09-04
Reduce the number of queries on plugin update and activation [MAILPOET-6218]

JanJakes:
## Description

These changes reduce the number of queries executed on plugin activation and update by about 300.

See also: [Eliminating 300 database queries on MailPoet plugin activation and update](https://mailpoetdev.wordpress.com/2024/09/04/eliminating-300-database-queries-on-mailpoet-plugin-activation-and-update/)

## Code review notes

I added some tests, I think QA can be skipped.

## QA notes

I added some tests, I think QA can be skipped.

## Linked tickets

[MAILPOET-6218]

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:activation-queries)

_The latest successful build from `activation-queries` will be used. If none is available, the link won't work._

[MAILPOET-6218]: https://mailpoet.atlassian.net/browse/MAILPOET-6218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Setup.php
+++ b/mailpoet/lib/API/JSON/v1/Setup.php
@@ -5,6 +5,7 @@ namespace MailPoet\API\JSON\v1;
 use MailPoet\API\JSON\Endpoint as APIEndpoint;
 use MailPoet\Config\AccessControl;
 use MailPoet\Config\Activator;
+use MailPoet\Settings\SettingsController;
 use MailPoet\WP\Functions as WPFunctions;
 
 class Setup extends APIEndpoint {
@@ -18,17 +19,23 @@ class Setup extends APIEndpoint {
   /** @var Activator */
   private $activator;
 
+  /** @var SettingsController */
+  private $settings;
+
   public function __construct(
     WPFunctions $wp,
-    Activator $activator
+    Activator $activator,
+    SettingsController $settings
   ) {
     $this->wp = $wp;
     $this->activator = $activator;
+    $this->settings = $settings;
   }
 
   public function reset() {
     try {
       $this->activator->deactivate();
+      $this->settings->resetCache();
       $this->activator->activate();
       $this->wp->doAction('mailpoet_setup_reset');
       return $this->successResponse();
--- a/mailpoet/lib/Config/Populator.php
+++ b/mailpoet/lib/Config/Populator.php
@@ -35,14 +35,12 @@ use MailPoet\Subscribers\NewSubscriberNotificationMailer;
 use MailPoet\Subscribers\Source;
 use MailPoet\Subscription\Captcha\CaptchaConstants;
 use MailPoet\Subscription\Captcha\CaptchaRenderer;
-use MailPoet\Util\Helpers;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class Populator {
   public $prefix;
-  public $models;
   public $templates;
   /** @var SettingsController */
   private $settings;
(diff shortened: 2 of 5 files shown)


mailpoet/lib/Config/Populator.php
```diff
@@ -509,134 +505,116 @@ protected function newsletterOptionFields() {
...
+    $inserts = [];
+    $updates = [];
+    foreach ($templates as $template) {
+      $existing = $existingTemplates[$template['name']] ?? null;
+      if (
+        $existing
+        && $existing['categories'] === $template['categories']
+        && $existing['body'] === $template['body']
+        && $existing['thumbnail'] === $template['thumbnail']
+      ) {
```
mhadidg, line 566: There might be other considerations that I'm not aware of, but I'd rather compare the hash of the entire templates with `hash('sha256', serialize($var))`.

Still, this LGTM as is!

mailpoet/lib/API/JSON/v1/Setup.php
```diff
@@ -18,17 +19,23 @@ class Setup extends APIEndpoint {
...
   ) {
     $this->wp = $wp;
     $this->activator = $activator;
+    $this->settings = $settings;
   }
 
   public function reset() {
     try {
       $this->activator->deactivate();
+      $this->settings->resetCache();
```
mhadidg, line 38: I'm still naive to the codebase, and I don't quite understand the motive behind this change. Could you please provide some context? A commit description would be even better for puzzled readers :wink:

mhadidg (commented):
Great refactoring! It's much easier for me to review tested code :+1:

This LGTM, but I have a question before approving it.

mailpoet/lib/Config/Populator.php
JanJakes: @mhadidg That's a cool idea, I don't really like the long IF condition.

I'm looking into it, but there are also some issues:
1. It would also include the value of `readonly`. The PHP templates return it as in integer `1`, while from the DB it comes as a string `'1'`, so I would need to normalize it.
2. The row from the database and the array from PHP would both need to be sorted the same way.
3. It probably has a bit higher CPU cost, but I suppose it's maybe not even measurable.

So in the end, I'm more like let's maybe leave it as it is, it's just 3 compares, after all.

mailpoet/lib/API/JSON/v1/Setup.php
JanJakes: @mhadidg It's just that `$this->activator->deactivate()` wipes out the database tables, but since the following `$this->activator->activate()` is within the same request, the settings are not wiped out — it contains the old settings in memory. Do you think I should also amend the commit message?

mailpoet/lib/Config/Populator.php
JanJakes: Oh, and it's important to say that all of the three checked values are just strings.

mhadidg: Well, it doesn't seem like a good idea then. Thanks for looking into it!

mailpoet/lib/API/JSON/v1/Setup.php
mhadidg: Aha, I get it. It is needed after the changes [here](https://github.com/mailpoet/mailpoet/pull/5818/commits/1a6aa487aae9f26081575caa63f3a0001f44166a)!

> Do you think I should also amend the commit message?

Not really, it's more like nice-to-have.

mhadidg (approved):
Thanks for addressing the comments.

PR #5820, opened 2024-09-04
Release 5.1.1

lysyjan:
## Preview


veljkho (approved):
Approving. 
The ALC and WC Products blocks can't be moved inside columns blocks in both newsletter editor and WC customizer editor.

PR #5828, opened 2024-09-10
Fix data inconsistencies loading [MAILPOET-6219]

costasovo:
## Description
This PR fixes the issue preventing loading of the MailPoet > Help page for senders with long sending history (e.g. mailpoet.com).

#### Changes
- optimized check for orphaned scheduled tasks subscribers 
- data inconsistencies are now fetched via API so that fetching doesn't block page load

## Code review notes
Query improved using @JanJakes ideas https://github.com/mailpoet/mailpoet/pull/5828#discussion_r1753470931

## QA notes

Please follow the testing instructions from [the original PR](https://github.com/mailpoet/mailpoet/pull/5769), but check only `Orphaned task subscribers`.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6219]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6219]: https://mailpoet.atlassian.net/browse/MAILPOET-6219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:fix-data-inconsistencies)

_The latest successful build from `fix-data-inconsistencies` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/Util/DataInconsistency/DataInconsistencyRepository.php
+++ b/mailpoet/lib/Util/DataInconsistency/DataInconsistencyRepository.php
@@ -40,9 +40,21 @@ class DataInconsistencyRepository {
     $stsTable = $this->entityManager->getClassMetadata(ScheduledTaskSubscriberEntity::class)->getTableName();
     /** @var string $count */
     $count = $this->entityManager->getConnection()->executeQuery("
-      SELECT count(*) FROM $stsTable sts
-      LEFT JOIN $stTable st ON st.`id` = sts.`task_id`
-      WHERE st.`id` IS NULL
+      SELECT COUNT(*)
+        FROM $stsTable sts
+        WHERE sts.task_id IN (
+          -- SELECT ... FROM (subquery) forces the subquery to materialize first.
+          -- We're using this twice here to get the reduced data set before JOIN.
+          SELECT oprhaned_task_ids.task_id FROM
+          (
+            SELECT task_ids.task_id
+            FROM (
+              SELECT DISTINCT task_id FROM $stsTable
+            ) AS task_ids
+            LEFT JOIN $stTable st ON st.id = task_ids.task_id
+            WHERE st.id IS NULL
+          ) AS oprhaned_task_ids
+        );
     ")->fetchOne();
     return intval($count);
   }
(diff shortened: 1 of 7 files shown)


mailpoet/lib/Util/DataInconsistency/DataInconsistencyRepository.php
```diff
@@ -35,18 +35,6 @@ public function getOrphanedSendingTasksCount(): int {
...
-    $stsTable = $this->entityManager->getClassMetadata(ScheduledTaskSubscriberEntity::class)->getTableName();
-    /** @var string $count */
-    $count = $this->entityManager->getConnection()->executeQuery("
-      SELECT count(*) FROM $stsTable sts
-      LEFT JOIN $stTable st ON st.`id` = sts.`task_id`
-      WHERE st.`id` IS NULL
-    ")->fetchOne();
-    return intval($count);
-  }
-
```
JanJakes: I gave this a quick try and it seems that splitting it in two queries makes it very fast:

```sql
SELECT task_ids.task_id
FROM (
  SELECT DISTINCT task_id FROM wp_mailpoet_scheduled_task_subscribers
) AS task_ids
LEFT JOIN wp_mailpoet_scheduled_tasks st ON st.id = task_ids.task_id
WHERE st.id IS NULL;

SELECT COUNT(*)
FROM wp_mailpoet_scheduled_task_subscribers sts
WHERE sts.task_id IN (27106,30812,337919,337921);
```

The important piece is the `DISTINCT task_id`. The `wp_mailpoet_scheduled_task_subscribers` table is huge, so JOINing in any way _before_ having just the disctinct task IDs will be ineffective. However, if we first fetch the affected task IDs, we get subsecond query times even on mailpoet.com.

JanJakes (commented):
LGTM, but I think I also found a way to write the query efficiently.

mailpoet/lib/Util/DataInconsistency/DataInconsistencyRepository.php
JanJakes: @costasovo Here's how to write it with a single query, plus an explaining comment:

```sql
SELECT COUNT(*)
FROM wp_mailpoet_scheduled_task_subscribers sts
WHERE sts.task_id IN (
  -- SELECT ... FROM (subquery) forces the subquery to materialize first.
  -- We're using this twice here to get the reduced data set before JOIN.
  SELECT oprhaned_task_ids.task_id FROM
  (
    SELECT task_ids.task_id
    FROM (
      SELECT DISTINCT task_id FROM wp_mailpoet_scheduled_task_subscribers
    ) AS task_ids
    LEFT JOIN wp_mailpoet_scheduled_tasks st ON st.id = task_ids.task_id
    WHERE st.id IS NULL
    GROUP BY task_ids.task_id
  ) AS oprhaned_task_ids
);
```

This runs in about 0.5s on mailpoet.com.

JanJakes: As for `DELETE`, if we want to use it with `LIMIT`, we could probably wrap the whole query with `DELETE FROM (... LIMIT <limit>)` and replace the `COUNT` with IDs.

Aschepikov (approved):
GJ!

PR #5829, opened 2024-09-10
Add new acceptance tests for timezone change scenario

veljkho:
## Description

This is to cover the recent issue in production when changing timezone for the scheduled newsletter resulted in hotfix release.

I added tests to cover scheduled newsletter, subscribed subscriber and newsletter stats page.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6221]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6221]: https://mailpoet.atlassian.net/browse/MAILPOET-6221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:acceptance/new/timezone-tests)

_The latest successful build from `acceptance/new/timezone-tests` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/tests/acceptance/Subscribers/ManageSubscribersCest.php
+++ b/mailpoet/tests/acceptance/Subscribers/ManageSubscribersCest.php
@@ -6,6 +6,7 @@ use Codeception\Util\Locator;
 use MailPoet\Entities\SegmentEntity;
 use MailPoet\Test\DataFactories\Segment;
 use MailPoet\Test\DataFactories\Subscriber;
+use MailPoetVendor\Carbon\Carbon;
 
 class ManageSubscribersCest {
 
@@ -38,6 +39,7 @@ class ManageSubscribersCest {
   private function generateSingleSubscriber($newSubscriberEmail, $subscriberFirstName, $subscriberLastName, $segment) {
     $subscriberFactory = new Subscriber();
     return $subscriberFactory
+      ->withCreatedAt(new Carbon('2024-09-01 06:00:00'))
       ->withEmail($newSubscriberEmail)
       ->withFirstName($subscriberFirstName)
       ->withLastName($subscriberLastName)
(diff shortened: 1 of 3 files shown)


mailpoet/tests/acceptance/Subscribers/ManageSubscribersCest.php
```diff
@@ -70,6 +72,21 @@ public function viewSubscriberList(\AcceptanceTester $i) {
     $i->searchFor('Alec Saunders');
     $i->waitForText('Alec Saunders');
     $i->seeNoJSErrors();
+
+    $i->wantTo('see subscribed date changed upon switching timezone');
+    $subscriberFirstName = 'John ';
```
lysyjan: In my opinion, the space at the end of the first name reduces the readability of the test. I was a bit confused by the concatenation without a space. I would prefer removing the space from the first naem and using the space in the name concatenation. Maybe we can also add waiting on the subscriber's email or replace waiting on the name with waiting on the email.
What do you think about it?

lysyjan (commented):
Thanks for those improvements @veljkho. I noticed a small thing that can improve readability.

mailpoet/tests/acceptance/Subscribers/ManageSubscribersCest.php
veljkho: Thanks for review @lysyjan 
I updated the test as suggested. First I wanted to make like in JS `'${var} ${var}'` but the wanted to avoid doing like this  `$i->waitForText($subscriberFirstName . ' ' . $subscriberLastName);` and added that simple space in the name which worked fine. Now it's like suggested.

lysyjan (approved):

PR #5833, opened 2024-09-11
Improve WP-Cron notice by checking whether the cron is running [MAILPOET-6224]

JanJakes:
## Description

Some hosts set `DISABLE_WP_CRON` to `true`, but execute the cron in a different way. This is also the case of WordPress VIP.

In addition to checking the value of the constant, we will now also make sure that the cron is not running indeed (= last cron run was more than an hour ago).

## Code review notes

QA can be skipped.

## QA notes

QA can be skipped.

## Linked PRs

The original implementation was done in https://github.com/mailpoet/mailpoet/pull/5614.

## Linked tickets

[MAILPOET-6224]

[MAILPOET-6224]: https://mailpoet.atlassian.net/browse/MAILPOET-6224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:wp-cron-notice)

_The latest successful build from `wp-cron-notice` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/Util/Notices/DisabledWPCronNotice.php
+++ b/mailpoet/lib/Util/Notices/DisabledWPCronNotice.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\Util\Notices;
 
+use MailPoet\Cron\CronHelper;
 use MailPoet\Cron\CronTrigger;
 use MailPoet\Settings\SettingsController;
 use MailPoet\Util\Helpers;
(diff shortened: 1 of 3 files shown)


lysyjan (approved):
Good job 👍

PR #5838, opened 2024-09-12
Use WP-native `nocache_headers()` when bypassing cache

JanJakes:
## Description

Currently, we set some no-cache headers manually in a few places in the plugin. In WordPress, there is the `nocache_headers()` function for that. We should use it, and it should [ensure our compatibility with WordPress VIP](https://docs.wpvip.com/caching/page-cache/#h-cache-bypass-or-prevention).

## Code review notes

I decided to apply the no-cache headers on the whole public `mailpoet_router`. We seem to use it as an API, or for user-specific actions, but since it uses GET requests, they could get accidentally cached.

These actions are:
- Open and click tracking.
- Cron daemon (run and ping).
- Subscription management including CAPTCHA.
- View-in-browser.
- Form previews.
- Proxy for external template images when creating a thumbnail.

I believe none of these should allow caching.

QA can be skipped.

## QA notes

QA can be skipped.

## Linked tickets

[MAILPOET-6226]

[MAILPOET-6226]: https://mailpoet.atlassian.net/browse/MAILPOET-6226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:no-cache-headers)

_The latest successful build from `no-cache-headers` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/Cron/DaemonHttpRunner.php
+++ b/mailpoet/lib/Cron/DaemonHttpRunner.php
@@ -4,6 +4,7 @@ namespace MailPoet\Cron;
 
 use MailPoet\Cron\Triggers\WordPress;
 use MailPoet\Settings\SettingsController;
+use MailPoet\Util\Headers;
 use MailPoet\WP\Functions as WPFunctions;
 use Tracy\Debugger;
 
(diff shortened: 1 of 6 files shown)


NeosinneR (approved):
LGTM; nice improvement. I agree that the public MailPoet router shouldn't be cached.

PR #5846, opened 2024-09-19
Move Email Editor engine to the PHP package [MAILPOET-6216]

lysyjan:
## Description

This PR moves the email editor PHP code to the package, which is symlinked during the development and copied to the final build by Composer.

## Code review notes

- I decided to create a simple DI Container for injecting class instances into integration tests because the package does not have access to the plugin container in integration tests.
- Integration tests run on the common architecture, and we don't need to install Codeception packages again in the editor package.
- Unit and Integration tests were moved to [the GitHub Action workflow](https://github.com/mailpoet/mailpoet/actions/workflows/email-editor-package.yml). I added some caching here, and the script installs composer dependencies when the cache is not found.
- The CdnAssetUrl class and Validator classes were copied, and I would want to follow up here on one of the next tickets.
- Because I installed newer PHPUnit, I needed to remove usages of the `verify` function. So, the unit tests are slightly refactored but nothing critical.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6216]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6216]: https://mailpoet.atlassian.net/browse/MAILPOET-6216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/php-package)

_The latest successful build from `email-editor/php-package` will be used. If none is available, the link won't work._


packages/php/email-editor/tests/_support/IntegrationTester.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+    $this->wpTermIds = [];
+  }
+
+  private function deleteCreatedComments() {
+    foreach ($this->createdCommentIds as $commentId) {
+      wp_delete_comment($commentId, true);
+    }
+    $this->createdCommentIds = [];
+  }
```
costasovo: I see these two methods delete data based on local properties `wpTermIds` and `createdCommentIds`, and we don't write in them, so I think we should remove these cleanup.

packages/php/email-editor/src/Container.php
```diff
@@ -0,0 +1,28 @@
...
+
+class Container {
+  protected array $services = [];
+  protected array $instances = [];
+
+  public function set(string $name, callable $callable): void {
+    $this->services[$name] = $callable;
+  }
+
+  public function get(string $name) {
```
costasovo, line 18: In the free plugin's container we have this annotation for `get`:
```
/**
   * @template T
   * @param class-string<T> $id
   * @return T
   */
```
We use classnames as id and this helps PHPStan and also IDE to figure out type of service when we call `$service = $container->get(Service::class);`
If that works without this annotation please ignore this comment.

packages/php/email-editor/src/Utils/CdnAssetUrl.php
```diff
@@ -0,0 +1,20 @@
...
+    string $baseUrl
+  ) {
+    $this->baseUrl = $baseUrl;
+  }
+
+  public function generateCdnUrl($path) {
+    $useCdn = defined('MAILPOET_USE_CDN') ? MAILPOET_USE_CDN : true;
+    return ($useCdn ? self::CDN_URL : $this->baseUrl . '/plugin_repository/') . "assets/$path";
+  }
+}
```
costasovo, line 20: This is ok to copy now. But, IMO, it is something we should not use in this package as this is dependent on plugin assets. I'm not 100% sure where it is used. I assume it is in a pattern for basic content or perhaps a template. So later we need to move those to MailPoet plugin space.

.github/workflows/email-editor-package.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Cache Composer vendor-prefixed dependencies for MailPoet
+        id: vendor-prefixed-cache
+        uses: actions/cache@v4
+        with:
+          path: mailpoet/vendor-prefixed
+          key: ${{ runner.os }}-vendor-prefixed-${{ matrix.php-version }}-${{ hashFiles('mailpoet/prefixer/composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-prefixer-mailpoet-${{ matrix.php-version }}-
+            ${{ runner.os }}-prefixer-mailpoet-
```
costasovo, line 36: Why do we need to install these dependencies? If they are needed now, it's fine for this PR, but for the future, we should remove dependency on those or move them to monorepo root.  I'm asking basically to understand what needs to be done.

costasovo (commented):
@lysyjan Great job!
I really appreciate how the PR was structured to commits.

I left a couple of comments about tiny details we could improve before merge and a few questions about future steps.

packages/php/email-editor/src/Utils/CdnAssetUrl.php
lysyjan: This is something I would like to address in a follow-up ticket. I will create it before moving this PR forward.

.github/workflows/email-editor-package.yml
lysyjan: There are more reasons for it.
- Codeception package, which is reused to avoid installing multiple times.
- Plugin activation in the tests docker entry point.
I will create a follow-up ticket for this.

packages/php/email-editor/src/Container.php
lysyjan: That's a good point. PHPStan will be added to [the ticket focusing on WP Coding Standard](https://mailpoet.atlassian.net/browse/MAILPOET-6240) so I can add it to the next PR.

lysyjan:
@costasovo I added two new commits and also created two new Jira tickets. Both tickets are assigned to you in Jira.

packages/php/email-editor/tests/_support/IntegrationTester.php
lysyjan: It's addressed in one of the new commits.

packages/php/email-editor/src/Container.php
lysyjan: It's addressed in one of the new commits.

packages/php/email-editor/src/Container.php
```diff
@@ -0,0 +1,33 @@
...
+  protected array $services = [];
+  protected array $instances = [];
+
+  public function set(string $name, callable $callable): void {
+    $this->services[$name] = $callable;
+  }
+
+  /**
+   * @template T
+   * @param class-string<T> $id
```
costasovo: Oh! I noticed it should be `$name` in this case. Sorry! I copied the annotation from the current free container where we use $id as parameter name.

```suggestion
   * @param class-string<T> $name
```

.github/workflows/email-editor-package.yml
costasovo: Got it! Thank you for explaining and for preparing the tickets 👍

costasovo (commented):
@lysyjan Thanks for updating the PR and for creating the follow-up tickets. I noticed that a parameter name in the DI container get method doesn't match with the annotation. Please see the other comment!
Apart from that I think we can go ahead and merge it.

lysyjan:
@costasovo I force-pushed the last commit. Thanks for catching it.

costasovo (approved):
@lysyjan Thanks for fixes! I think we can go ahead and merge it.
Should we ping the other developers that they need to reinstall deps or something?

lysyjan:
> Should we ping the other developers that they need to reinstall deps or something?

@costasovo Yes. I will write a message on the Slack channel.

PR #5849, opened 2024-09-19
Fix calculating stats in automation analytics flow [MAILPOET-6230]

NeosinneR:
## Description

Fix automation flow analytics when the if/else branch is included. 

**Before:**
Annotation on 2024-09-19 at 22-01-47

**After:**
Screenshot 2024-09-19 at 22 26 08

## Code review notes

Stats are rendered in the step separator, which accepts the previous step. This was causing issues because both branches of the if/else step would receive the if/else step to calculate stats from—this results in both showing the same number (I don't know how this could ever work before). 

This PR changes this so the separator also receives the next step, which is primarily used to calculate stats. The previous step is only used if the next step is unavailable (end of automation or a branch). 

There is also a special case with an empty if/else branch, in which we need to calculate the stats by looking at the sibling. 

## QA notes

Please test that the numbers in the automation analytics flow are correctly calculated when if/else steps are used. Also, focus on empty branches and processing users (those who have yet to complete the automation). 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6230]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6230]: https://mailpoet.atlassian.net/browse/MAILPOET-6230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:if-else-stats)

_The latest successful build from `if-else-stats` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/components/automation/flow-ending.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/automation/flow-ending.tsx
@@ -3,14 +3,23 @@ import { FlowSeparator } from './flow-separator';
 import { Step as StepData } from './types';
 
 type Props = {
-  stepData: StepData;
+  previousStepData: StepData;
   index: number;
+  nextStepData?: StepData;
 };
 
-export function FlowEnding({ stepData, index }: Props): JSX.Element {
+export function FlowEnding({
+  previousStepData,
+  index,
+  nextStepData,
+}: Props): JSX.Element {
   return (
     <div className="mailpoet-automation-editor-step-wrapper">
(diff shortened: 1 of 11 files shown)


websupporter (approved):
lgtm. thank you @NeosinneR

Aschepikov (approved):
GJ!

websupporter:
jira did not pick up the merged status, checking if a comment will just help to sync the status

PR #5851, opened 2024-09-20
Email editor - fixes for WP 6.6

costasovo:
## Description

This PR fixes the issue of the email editor crashing immediately after starting. This PR fixes the crash, but there is still an issue related to layouts, so some functionalities, such as "Align/Full, Align/Wide," don't work as expected. To unlock the other work these issues will be addressed in a follow-up PR(s).

Note: With this change, we will support WP 6.6, but we are dropping support of WP 6.5.

### Changes
- I needed to update lock/unlock string to match one in WP6.6
- the ExperimantalEditorProvider from `wordpress/editor` is no longer exported (was private) so I went with the EditorProvider which works but I had to fetch patterns and pass them to the block editor store via settings.
- the EditorCanvas from `wordpress/editor` is gone (was exported private) I workaround that by copying visual-editor component and simplified it to support our usecases. Well, perhaps I oversimplified it because layouts don't work as expected. I'll address this in an extra PR.
- refactored how we work with preview state and we will read deviceType state from `wordpress/editor` store.
- a couple of className updates and small visual fixes

## QA notes

There are still some issues and we need to merge this to unblock further work, so we could perhaps skip QA. Or we could ask them to report issues and address them as a follow-up.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6105]

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/wp66-fix-v1)

_The latest successful build from `email-editor/wp66-fix-v1` will be used. If none is available, the link won't work._

[MAILPOET-6105]: https://mailpoet.atlassian.net/browse/MAILPOET-6105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/components/block-editor/editor.tsx
+++ b/mailpoet/assets/js/src/email-editor/engine/components/block-editor/editor.tsx
@@ -10,8 +10,9 @@ import {
 } from '@wordpress/editor';
 import { useMemo } from '@wordpress/element';
 import { SlotFillProvider, Spinner } from '@wordpress/components';
-import { store as coreStore } from '@wordpress/core-data';
+import { Post, store as coreStore } from '@wordpress/core-data';
 import { storeName } from '../../store';
+import { unlock } from '../../../lock-unlock';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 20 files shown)


mailpoet/lib/EmailEditor/Engine/theme.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://schemas.wp.org/trunk/theme.json",
-  "version": 2,
+  "version": 3,
```
lysyjan: @costasovo You increased the theme's version here. Should we also increase this version in [our Core theme](https://github.com/mailpoet/mailpoet/blob/b510e07942af573162ccaf241418b8e53c16a6e3/mailpoet/lib/EmailEditor/Integrations/Core/theme.json)?

lysyjan (dismissed):
I think you did a great job.
Maybe we can disable skipping our acceptance tests again?

I would suggest adding it to a new ticket and merging this. Do you agree?

mailpoet/lib/EmailEditor/Engine/theme.json
costasovo: Good point. I'll do it.

costasovo:
> Maybe we can disable skipping our acceptance tests again? 

@lysyjan  I totally forgot they are disabled. Good catch! I'll enable and update them.

costasovo:
@lysyjan, I added two new commits with an update of the schema version and acceptance tests.

lysyjan (approved):
Good job! Thanks

PR #5859, opened 2024-09-27
Use cached authorized emails when API is down

mhadidg:
## Description

If the shop API is down or broken, the sender-authorized email check will fail, and we will block sending. It seems to be the case even if the user has previously used the same authorized email address. See the linked Jira issue for more details.

I've implemented caching logic, where API data for authorized emails is cached. However, the logic will **only fall back to the cache if the service method returns `NULL`**, which may occur when:

- The shop API is down, or
- The shop API is broken (returns a non-2xx code), or
- The bridge API is down, in which case caching won't really help anyway!

## Code review notes

_N/A_

## QA notes

Here are a couple of suggestions, but feel free to expand with more scenarios:

- Verify that you can get a newly authorized email to be accepted by the plugin.
- After successful authorization, verify that you can send a newsletter.
- Then, attempt to send a newsletter while the shop is down.

The first two scenarios ensure there is no regression; only the last one verifies the changes introduced here!

> [!TIP]
> I'm not sure if this would work, but you should be able to simulate the "shop down" scenario by mocking the bridge API host in the Docker Compose file. If this introduces a blocker (e.g., preliminary step fails), it would render testing unfeasible, but it's worth trying!

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-5980]

## Tasks

[MAILPOET-5980]: https://mailpoet.atlassian.net/browse/MAILPOET-5980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:fix-shop-down-break-sending)

_The latest successful build from `fix-shop-down-break-sending` will be used. If none is available, the link won't work._

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -55,3 +55,4 @@ ij_php_space_after_for_semicolon = true
 ij_php_space_after_colon_in_return_type = true
 ij_php_space_before_else_keyword = true
 ij_php_for_statement_new_line_after_left_paren = true
+ij_php_class_brace_style = end_of_line
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,4 @@
 #!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
 . "$(dirname "$0")/common.sh"
 [ "$MP_GIT_HOOKS_ENABLE" != "true" ] && exit 0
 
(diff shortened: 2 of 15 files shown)


mhadidg:
David is the Janitor this week; I'll re-assign the reviewer to Jan M.

NeosinneR (approved):
Sorry for the delay in review, @mhadidg. This looks good, thanks for the improvement!

Aschepikov (approved):
GJ!

PR #5876, opened 2024-10-14
Ensure the newsletter has hash before preview [MAILPOET-6273]

pavel-mailpoet:
## Description

The preview of the Sign-Up Confirmation email is not working

There is no form hash being passed in the encoded URL param. The happiness team reported this. 
WP-Admin > MailPoet > Settings > Sign-up Confirmation > Open template editor > Preview
It works on some older sites but not on new ones.

## Code review notes

_N/A_

## QA notes

To replicate the issue, you need to find a newsletter with the type `confirmation_email` in the `mailpoet_newsletters` table and ensure that the `hash` column is empty. Then in MailPoet > Settings > Sign-up Confirmation > Open template editor and click Preview and observe that the preview is not displaying any content.

The issue is fixed in migration, so you need to run migrations, e.g., by activating and deactivating the plugin. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6273]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6273]: https://mailpoet.atlassian.net/browse/MAILPOET-6273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:preview-confirmation)

_The latest successful build from `preview-confirmation` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Newsletters.php
+++ b/mailpoet/lib/API/JSON/v1/Newsletters.php
@@ -28,6 +28,7 @@ use MailPoet\Services\AuthorizedEmailsController;
 use MailPoet\Settings\SettingsController;
 use MailPoet\UnexpectedValueException;
 use MailPoet\Util\License\Features\Subscribers as SubscribersFeature;
+use MailPoet\Util\Security;
 use MailPoet\WP\Emoji;
 use MailPoet\WP\Functions as WPFunctions;
 use MailPoetVendor\Carbon\Carbon;
@@ -327,6 +328,10 @@ class Newsletters extends APIEndpoint {
     $newsletter->setBody(
       json_decode($this->emoji->encodeForUTF8Column($newslettersTableName, 'body', $data['body']), true)
     );
+    // ensure newsletter has hash
+    if (!$newsletter->getHash()) {
+      $newsletter->setHash(Security::generateHash());
+    }
     $this->newslettersRepository->flush();
 
     $response = $this->newslettersResponseBuilder->build($newsletter);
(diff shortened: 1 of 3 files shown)


mailpoet/lib/API/JSON/v1/Newsletters.php
```diff
@@ -327,6 +328,10 @@ public function showPreview($data = []) {
     $newsletter->setBody(
       json_decode($this->emoji->encodeForUTF8Column($newslettersTableName, 'body', $data['body']), true)
     );
+    // ensure newsletter has hash
+    if (!$newsletter->getHash()) {
+      $newsletter->setHash(Security::generateHash());
+    }
```
costasovo: I see how this fixes the issue with the preview, but I have some concerns about this approach. 
It seems that the source of the problem is that the confirmation email doesn't have the hash in the DB. With this fix the missing hash is added only when a user attempts to get a preview of the email so the sites where a user doesn't click preview will still have the missing hash issue.

Why the hash is not created when the email is added to the DB?
Could the missing hash cause issues in different places in the admin now or in the future (e.g. we add the preview link somewhere else)?
Did you consider an approach in which we ensure that the hash always exists for the confirmation email so that new sites don't suffer from the issue? For the older sites, we could add a migration.

costasovo (commented):
@pavel-mailpoet Thanks for addressing the issue! I have some concerns about the approach. I shared more details in the other comment.

costasovo (commented):
@pavel-mailpoet I see you've added a commit that prevents creating the confirmation email without the hash for new installs. 👍 

Please correct me if I'm wrong. The sites where users don't click the preview button will still have the confirmation email without the hash, right? 
I think we should also handle the issue with the existing installations and instead of fixing the data only on the click on the preview link we should fix that in the migration to ensure all sites are in the correct state. 
My concern here is that leaving some sites with the missing hash may lead to mysterious and hard to replicate issues in the future. I think this could be addressed in an App level migration where you can inject any service from the DI. The migration can be created easily via `./do migration:new app`

costasovo (approved):

Aschepikov (approved):
GJ!

PR #5879, opened 2024-10-14
Update performance tests to be in line with the latest changes from k6 browser

veljkho:
## Description

This is to make the tests executable at least from the local (confirmed working).

There will be follow-up PR to fix checks visibility in the output, currently says 15 ✅ but in reality it is 41 ✅ due to some issue in the latest k6 browser, will be heading to community forum for help.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6274]

## After-merge notes

_N/A_

[MAILPOET-6274]: https://mailpoet.atlassian.net/browse/MAILPOET-6274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:performance/fix-tests)

_The latest successful build from `performance/fix-tests` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/tests/performance/config.js
+++ b/mailpoet/tests/performance/config.js
@@ -17,7 +17,7 @@ export const firstName = 'John';
 export const lastName = 'Doe';
 export const defaultListName = 'Newsletter mailing list';
 
-export const thinkTimeMin = '1';
+export const thinkTimeMin = '2';
 export const thinkTimeMax = '3';
 
 export const emailsPageTitle = 'Emails Page';
--- a/mailpoet/tests/performance/README.md
+++ b/mailpoet/tests/performance/README.md
@@ -93,7 +93,7 @@ Included in the `tests` folder is a set of test scripts which includes all scena
 
 Another aspect that affects the traffic pattern of the tests is the amount of “Think Time” in between requests. In real world usage of an application users will spend some amount of time before doing some action. For some situations there will be a need to simulate this as part of the test traffic and it is common to have this as part of load tests.
 
-To do this a sleep step is included between each request ``sleep(randomIntBetween(`${think_time_min}`, `${think_time_max}`))``.
+To do this a sleep step is included between each request `` sleep(randomIntBetween(`${think_time_min}`, `${think_time_max}`)) ``.
 The amount of think time can be controlled from `config.js`.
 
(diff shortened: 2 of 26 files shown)


mailpoet/tests/performance/config.js
```diff
@@ -17,7 +17,7 @@ export const firstName = 'John';
 export const lastName = 'Doe';
 export const defaultListName = 'Newsletter mailing list';
 
-export const thinkTimeMin = '1';
+export const thinkTimeMin = '2';
```
costasovo, line 20: I just want to ask if this is needed and make sure it isn't a leftover from some debugging. Is it needed?

costasovo (commented):
@veljkho Thank you for the update. I'm not familiar with the framework so I don't understand why we need to add asyc/await, but I assume that's needed because of some changes in an underlying library. Providing a little bit more context would be helpful for understanding. Otherwise it looks fine. I added just one question about the changed constant.

mailpoet/tests/performance/config.js
veljkho: Yes, this is added on purpose. I'd like to use it in tests instead of sleep(2), as it is [added here](https://github.com/mailpoet/mailpoet/pull/5879/files#diff-3d8a23534fcadfd2f81497f14f747ce497a217b59e13d638b08d9f7df0d5590bL75). K6 recommends placing sleeps in tests to slow down loading the pages to fully test the performance. Anyway I keep it very low and rarely use sleeps in tests but instead wait for other elements to load or the whole page.

To use `await` (and async somewhere in order to place `await`) is the new normal in k6 and is [required by the migration guide here](https://grafana.com/docs/k6/latest/using-k6-browser/migrating-to-k6-v0-52/).

Thanks for the review!

costasovo: @veljkho Thank you for providing the details and the link to the migration guide. Seeing the migration guide helped me to understand why the changes were done 👍

costasovo (approved):
@veljkho Approved!

PR #5881, opened 2024-10-15
Fix inactive notifications clog sending queue [MAILPOET-6259]

pavel-mailpoet:
## Description

From Support: a user has 80+ post notifications. In the database a lot of very old scheduled tasks that would not go out. The associated newsletters where not active. In the scheduler we pick up 5 sending tasks to start sending. If no queue or newsletter is found we delete the task. Otherwise we start to process the newsletter, except for when the newsletter is not active, then we continue; and the sending task will just return into the queue until the status of the newsletter changes and gets actually processed.

## Code review notes

Please let me know if you think this is the right approach. It was described in the ticket, but I am not sure if that is correct. 

## QA notes

Sending was changed. If there is a paused email with a scheduled task the task is now paused as well. If the email is unpaused the task is scheduled and should be sent.  

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6259]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6259]: https://mailpoet.atlassian.net/browse/MAILPOET-6259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:fix-inactive-notifications)

_The latest successful build from `fix-inactive-notifications` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Newsletters.php
+++ b/mailpoet/lib/API/JSON/v1/Newsletters.php
@@ -11,6 +11,7 @@ use MailPoet\Cron\CronHelper;
 use MailPoet\Doctrine\Validator\ValidationException;
 use MailPoet\Entities\NewsletterEntity;
 use MailPoet\Entities\NewsletterOptionFieldEntity;
+use MailPoet\Entities\ScheduledTaskEntity;
 use MailPoet\Entities\SendingQueueEntity;
 use MailPoet\InvalidStateException;
 use MailPoet\Listing;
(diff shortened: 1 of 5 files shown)


alex-mpoet (commented):
The changes look good. I just wanted to note that in the hotfix version @websupporter also added task deletion for inactive newsletters [here](https://github.com/mailpoet/mailpoet/blob/714d8da93afc5812571114f59b198986ea961e46/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L178) and [here](https://github.com/mailpoet/mailpoet/blob/714d8da93afc5812571114f59b198986ea961e46/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L187), probably to fix clogged tasks that are already running (not scheduled). Do we need it in this PR? It's not obligatory as these can be cleared manually via the System Status page in the worst case.

Also, as we discussed, post notifications can be clogged for active newsletters too. I know you're looking at it right now, so maybe, if it's closely related, we can wait a bit and include a fix for that case too.

pavel-mailpoet:
@alex-mpoet 

> I just wanted to note that in the hotfix version @websupporter also added task deletion for inactive newsletters [here](https://github.com/mailpoet/mailpoet/blob/714d8da93afc5812571114f59b198986ea961e46/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L178) and [here](https://github.com/mailpoet/mailpoet/blob/714d8da93afc5812571114f59b198986ea961e46/mailpoet/lib/Cron/Workers/SendingQueue/SendingQueue.php#L187), probably to fix clogged tasks that are already running (not scheduled).

Good idea. I added a commit with that.

> Also, we we discussed, post notifications can be clogged for active newsletters too. I know you're looking at it right now, so maybe, if it's closely related, we can wait a bit and include a fix for that case too.

I think there was a different reason for that and with the soft delete the code shouldn't have the same issue.

alex-mpoet (approved):

Aschepikov (approved):
GJ!

PR #5883, opened 2024-10-15
Use post_type parameter to filter email-related templates [MAILPOET-6279]

costasovo:
## Description

This PR improves the fetching of templates for the email editor by moving the filtering of the templates from the client to the REST API.

## Code review notes

Please verify that the template selector offers the correct templates.

## QA notes

We can skip QA.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6279]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6279]: https://mailpoet.atlassian.net/browse/MAILPOET-6279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/templates-fetch-by-post-type)

_The latest successful build from `email-editor/templates-fetch-by-post-type` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/store/selectors.ts
+++ b/mailpoet/assets/js/src/email-editor/engine/store/selectors.ts
@@ -196,10 +196,18 @@ export const getCurrentTemplate = createRegistrySelector((select) => () => {
  */
 export const getEmailTemplates = createRegistrySelector(
   (select) => () =>
-    select(coreDataStore).getEntityRecords('postType', 'wp_template', {
-      per_page: -1,
-      post_type: 'mailpoet_email',
-    }),
+    select(coreDataStore)
+      .getEntityRecords('postType', 'wp_template', {
+        per_page: -1,
+        post_type: 'mailpoet_email',
+      })
+      // We still need to filter the templates because, in some cases, the API also returns custom templates
+      // ignoring the post_type filter in the query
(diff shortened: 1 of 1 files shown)


lysyjan (commented):
When I tested it with the WP 6.7 Beta and active theme `Twenty Twenty Five` I could see five templates instead of four. Is it correct?

Screenshot on 2024-10-16 at 12-46-32

costasovo:
> When I tested it with the WP 6.7 Beta and active theme Twenty Twenty Five I could see five templates instead of four. Is it correct?

@lysyjan It is not expected :( It looks like the template is somehow added to the response, even though the API should filter it. Good catch! I'll dig into it. Maybe we will need to keep the client-side filtering in place to be sure.

costasovo:
@lysyjan It looks like there is a bug in the WP's REST API for templates. When a custom template is created in the site editor the API returns it even when the post_type filter is set to `mailpoet_emails`. I returned the client side filtering. I think we should still use the query filter as it still filters a bunch of templates.

lysyjan (approved):

PR #5884, opened 2024-10-17
Use a different port for testing in order

pavel-mailpoet:
## Description

We need to prevent a conflict with the dashboard

## Code review notes

_N/A_

## QA notes

We don't need QA for this

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:testing-port)

_The latest successful build from `testing-port` will be used. If none is available, the link won't work._

### DIFF
--- a/tests_env/docker/docker-compose.yml
+++ b/tests_env/docker/docker-compose.yml
@@ -89,8 +89,6 @@ services:
       - ../../mailpoet-premium:/var/www/html/wp-content/plugins/mailpoet-premium
     tmpfs:
       - /var/www/html/wp-content/uploads/mailpoet/
-    ports:
-      - 8889:80
     environment:
       WORDPRESS_DB_PASSWORD: wordpress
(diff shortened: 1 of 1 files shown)


mhadidg (changes requested):
Thanks for pushing a fix :pray: 

[Seems like](https://documentation.mamp.info/en/MAMP-Mac/Preferences/Ports/) MAMP already claimed the `8889` port; this app is pretty common among macOS folks! Let's pick another one, or we could remove port mapping from this container...

I think we should be fine without it! Containers use Docker's private network (with defined aliases) to communicate with each other. Mapping simplifies access to the WordPress instance from the host, but I don't see any reason for that; these are testing containers!

mhadidg (approved):
Looks better to me :+1:

PR #5899, opened 2024-10-23
Fix translation warning message for WP 6.7 - MAILPOET-6298

triple0t:
## Description

Please check the JIRA ticket for details.

[MAILPOET-6298](https://mailpoet.atlassian.net/browse/MAILPOET-6298)

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6298]: https://mailpoet.atlassian.net/browse/MAILPOET-6298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:adding-support-for-wp-6_7)

_The latest successful build from `adding-support-for-wp-6_7` will be used. If none is available, the link won't work._


mailpoet/lib/Automation/Engine/Registry.php
```diff
@@ -56,6 +56,13 @@ public function __construct(
     $this->wordPress = $wordPress;
     $this->steps[$rootStep->getKey()] = $rootStep;
 
+    $this->wordPress->addAction('init', [
+      $this,
+      'setupTemplateCategories',
+    ]);
+  }
```
costasovo: I have an idea for a small optimization. If I understand properly the code is called on every page load. This happens because of the dependency tree and probably some higher level service needs this service. I think that, in general, we should avoid adding logic to constructors.

I would suggest initiating the template categories lazily. There is a method `getTemplateCategories` that could be used to fill the Registry::$templateCategories lazily with the translated data. We just need to replace places where we access `$this->templateCategories` with `$this->getTemplateCategories`.

I see that we call `addTemplate` couple of times on every page load so technically we wouldn't save the calls of the translation functions, but I find it better from the design perspective as the class would take the resources when needed. What do you think?

mailpoet/lib/WooCommerce/TransactionalEmails.php
```diff
@@ -42,6 +42,13 @@ public function __construct(
     $this->template = $template;
     $this->woocommerceHelper = $woocommerceHelper;
     $this->newslettersRepository = $newslettersRepository;
+    $this->wp->addAction('init', [
+      $this,
+      'setupEmailHeadings',
+    ]);
+  }
```
costasovo: Similar to the previous comment. We could initiate the `emailHeadings` property lazily in `getEmailHeadings` method. In this case, moving the initialization of headings would mean we would not execute the translations functions on every page load, so we would save some CPU time.

costasovo (commented):
@triple0t Great job finding and addressing the issues!
I have a couple of suggestions I believe are small improvements we could do. Please see my other comments.

mailpoet/lib/Automation/Engine/Registry.php
triple0t: I agree with you. I'm not very familiar with the Automations codebase and Initially wanted to make very minimal changes. I understand your thought process and I belive we should implement it.

mailpoet/lib/WooCommerce/TransactionalEmails.php
triple0t: Thanks for the feedback. Implemented here https://github.com/mailpoet/mailpoet/pull/5899/commits/e56cc248ebd5f22a776ac37bb186f7ba5ffb5a9e

mailpoet/lib/Automation/Engine/Registry.php
triple0t: Implemented here https://github.com/mailpoet/mailpoet/pull/5899/commits/e56cc248ebd5f22a776ac37bb186f7ba5ffb5a9e

costasovo (approved):
@triple0t Thank you for implementing the improvements.

There are no instructions for QA or a note that we can skip the QA step in the description so I'm assigning the PR back to you. I think we could skip the QA in this case, but I let you decide.

PR #5900, opened 2024-10-23
Email editor - fixes of rendering inconsistencies [MAILPOET-6249][MAILPOET-6088]

costasovo:
## Description
This PR aims to fix inconsistencies between the email editor content rendering and the email renderer content rendering.
Some of the issues were introduced by us when we merged template prototypes too early and the last two WP updates (WP 6.5 and 6.6).

#### Content width restriction
With the template prototype, we allow users to use the full size of the canvas when editing a template. The renderer is currently not supporting this well. The renderer currently needs some width to compute column widths, etc. So, it was quite easy to edit the template in the template mode, so it looked different in the editor and in the rendered output. I fixed this by restricting editable canvas width to contentSize defined in theme.json (660px). I would like to revisit the idea of unlimited canvas size in a follow-up ticket. The current solution with restricted width limits the editor's capabilities but is good enough for the first iteration.

#### Removed full-width alignment option
The full-width alignment was problematic from the start. We can't use nice CSS, so to achieve the full alignment, we needed to recompute horizontal paddings for the whole document tree. This was causing issues with nested paddings. There is a workaround, and a user can achieve the full width by inserting a block at the root level of the content block. We could try to reintroduce this option, but I think we should reconsider it because of its complexity.

#### Change in Styles > Layout > Padding horizontal application
I've unified how the horizontal padding is applied with the vertical padding. The horizontal padding was applied inside the post-content block, and the header and footer blocks were ignored. Newly, it's applied on the root canvas block list so it affects the area left for the post content block. This is something we can discuss further. I would like to publish a post about that.

#### Removed demo templates
The demo templates we had were ugly. I created one new template from the default content. So currently, we have two templates, which look the same in the selector, but the one called "Simple Light" (not sure about the name) has header and footer as a part of the template VS the "General Email" that is basically the blank template and all is inserted into the content.

Screenshot 2024-10-23 at 15 46 35

#### Other fixes
- fixed columns rendering in Gmail (they were always side by side, even on mobile) and Outlook (were not side by side)
- incorrect widths because padding values were stored as compressed vars (`var:spacing|padding...`)
- mobile preview was not inconsistent with mobile render
- image which was aligned to left or right was breaking mobile version of the email
- unified block gap rendering for list items – I chose not to render block gap
- fixed that images with srcset attribute were not rendered in iOS clients

## Code review notes
I'm very sorry for such a big PR. It addresses multiple issues and contains a couple of smaller refactors (mostly removing duplicate code or logic). I did my best to keep commits small and explain the reasons in commit messages or code comments.

I recommend cleaning up all emails and modified templates before testing. To cleanup modified templates go to adminer wp_posts and look for wp_template post type and delete all that comes from MailPoet.

The new template contains texts that are not translatable. I think it is ok for now and we will move templates definition to PHP when we switch to the new API for registering templates.

## QA notes
I would still skip the QA as there are still some known issues.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6249] [MAILPOET-6088]

## After-merge notes

_N/A_

## Tasks

- [x] I added sufficient test coverage - I changed and updated a couple of tests but there is still a ticket for adding more tests related to templates

[MAILPOET-6249]: https://mailpoet.atlassian.net/browse/MAILPOET-6249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/layout)

_The latest successful build from `email-editor/layout` will be used. If none is available, the link won't work._


lysyjan (approved):
@costasovo Excellent work!
So many fixes and improvements.

PR #5902, opened 2024-10-25
Fix adding unsubscribe link to the email [MAILPOET-6254]

pavel-mailpoet:
## Description

To reproduce

1. Remove the Unsubscribe link
2. Click “Send”
3. See the error message
4. Click “Insert link”

What is expected: A paragraph block to be added to the content with an unsubscribe link and a manage subscription link.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6254]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6254]: https://mailpoet.atlassian.net/browse/MAILPOET-6254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:unsubscribe-link-action-fix)

_The latest successful build from `unsubscribe-link-action-fix` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/assets/js/src/email-editor/engine/hooks/use-content-validation.ts
+++ b/mailpoet/assets/js/src/email-editor/engine/hooks/use-content-validation.ts
@@ -3,7 +3,6 @@ import { useCallback, useMemo } from '@wordpress/element';
 import { dispatch, useSelect, subscribe } from '@wordpress/data';
 import { store as blockEditorStore } from '@wordpress/block-editor';
 import { createBlock } from '@wordpress/blocks';
-import { useDebounce } from '@wordpress/compose';
 import { storeName as emailEditorStore } from 'email-editor/engine/store';
 import { useShallowEqual } from './use-shallow-equal';
 import { useValidationNotices } from './use-validation-notices';
@@ -89,14 +88,12 @@ export const useContentValidation = (): ContentValidationData => {
     rules,
   ]);
 
-  const debouncedValidateContent = useDebounce(validateContent, 500);
-
   // Subscribe to updates so notices can be dismissed once resolved.
   subscribe(() => {
     if (!hasValidationNotice()) {
       return;
     }
-    debouncedValidateContent();
+    validateContent();
   }, emailEditorStore);
 
   return {
--- a/mailpoet/assets/js/src/email-editor/engine/hooks/use-content-validation.ts
+++ b/mailpoet/assets/js/src/email-editor/engine/hooks/use-content-validation.ts
@@ -4,6 +4,7 @@ import { dispatch, useSelect, subscribe } from '@wordpress/data';
 import { store as blockEditorStore } from '@wordpress/block-editor';
 import { createBlock } from '@wordpress/blocks';
 import { storeName as emailEditorStore } from 'email-editor/engine/store';
+import { store as coreDataStore } from '@wordpress/core-data';
 import { useShallowEqual } from './use-shallow-equal';
 import { useValidationNotices } from './use-validation-notices';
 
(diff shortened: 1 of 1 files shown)


pavel-mailpoet:
@costasovo, can you check the implementation and let me know what you think about it and if you see any issues with it? Also, I am not sure how to insert the link into the footer. Can you give me a hint on how to do that?

mailpoet/assets/js/src/email-editor/engine/hooks/use-content-validation.ts
```diff
@@ -89,14 +88,12 @@ export const useContentValidation = (): ContentValidationData => {
     rules,
   ]);
 
-  const debouncedValidateContent = useDebounce(validateContent, 500);
```
costasovo, line 74: Did you test the impact of removing the debounce? If the subscription itself is triggered reasonably perhaps we don't need the debounce, but I assume we added it because of performance reasons. We don't run the validation on every key press when writing in the editor.
There might be another workaround for the problem. Maybe we could run `select(emailEditorStore).getEditedEmailContent()` and `select(emailEditorStore).getTemplateContent()` directly in the validateContent function to get fresh data. 

> First with the invalid content and
then with the right content with the link present.
With debounce the test was not run for the content with the
link.

I would expect debounce to work the other way around. Maybe, it is related to the useCallback we use for validateContent.

costasovo (commented):
@pavel-mailpoet Thank you for working on the fix! I like the approach I'm just concerned about removing the debounce. Please see my other comment. 

> Also, I am not sure how to insert the link into the footer. Can you give me a hint on how to do that?

I see you detect the `core/post-content` and insert the link into it. For adding the link to the template I would assume that just skipping the rootBlockId in the insertBlock should work, but I remember that there was an issue with the template being locked when thje editor was in the `template-lock` mode when editing content. Am I right?

As a workaround we could:
1) Fetch the template content using `select(emailEditorStore).getTemplateContent()`
2) Add the paragraph directly to the content at the end.
3) Use core data actions to edit the tempalte record directly. [editEntityRecord](https://github.com/WordPress/gutenberg/tree/trunk/packages/core-data#editentityrecord) should do that.

I think we also need logic to decide when to add the link to a template and when to the post content. In case the email is using "General Template" which is a blank template we should put the link into the post content as you do now. In case the template has header and footer  we should add it into the footer. I think we should be able to determine that by checking the template blocks. If the template contains only `core/post-content` or `core/post-content` + the block with mailpoet logo for free users` we could treat the tempalte as "Blank template" and put the link directly into the content.

mailpoet/assets/js/src/email-editor/engine/hooks/use-content-validation.ts
pavel-mailpoet: I will test the performance again. It was not an issue when I tested it. There were some more renders, but nothing too bad. The issue with the workaround you suggested is that the function would be called “before” the store has the correct data. So even with directly pulling data we may get the old values. And with debounce the function would not be called again after it is populated.

pavel-mailpoet: @costasovo I tested that, and I think the performance is good enough. The `validateContent` function is not called when the notice is not displayed. If the notice is shown, it sometimes gets called too often. But I need to figure out what the alternative is. As far as I know, calling `select(emailEditorStore).getEditedEmailContent()` is impossible inside `validateContent`. For `select`, we would need to use a hook inside the function, which doesn't work, does it? We can use hooks only on the top level of components, can't we?

costasovo (approved):
Thanks for implementing the link addition for templates.

PR #5910, opened 2024-11-04
Add reCAPTCHA to WP & WC registration form [MAILPOET-6282]

mhadidg:
## Description

A partial implementation of the referenced ticket. This PR adds a Google's reCAPTCHA to WordPress and WooCommerce registration form. It supports both checkbox and invisible reCAPTCHA, based on the user settings. Read Jira description for more details.

I've opted for incremental implementation to make smaller PRs. After this, I'll push another one for built-in CAPTCHA. Once both are merged, I'll add a setting to make it accessible to end users.

## Rendering

In WordPress registration form:

wp-register-form

In WooCommerce registration form:

wc-register-form

## Code review notes

The setting for this feature, as per the Jira spec, has not been introduced here. Instead, I've added a `MP_ENV` env variable to activate the logic in testing and development ONLY.

I highly recommend reviewing by commit. I've added commit messages to explain some of the least obvious changes.

If you are not familiar with reCAPTCHA, I recommend reviewing the [official doc](https://developers.google.com/recaptcha/docs/display).

## QA notes

In the development environment, make sure of the following on **both WordPress & WooCommerce** registration forms:

- When "Google reCAPTCHA v2 Checkbox" option of "Protect your forms against spam signups" setting is selected,
  - The reCAPTCHA is rendered as checkbox, and
  - When unchecked, it breaks the registration, displaying CAPTCHA-specific error.
  - When checked, it should not break the registration.

- When "Google reCAPTCHA v2 Invisible" option is selected,
  - The reCAPTCHA should not be visible.
  - A reCAPTCHA badge in the bottom right [should be visible](https://wpblog.com/wp-content/uploads/2018/01/recaptcha-860x599.png).
  - It should not break registration on form submission.

- When any other option is selected, the reCAPTCHA should not be rendered in WordPress & WooCommerce registration forms. The reCAPTCHA root div has `g-recaptcha` class.

I've adjusted the reCAPTCHA validation logic on MP subscription forms. To ensure no regression, verify that checkbox and invisible reCAPTCHA still works there. 

## Linked tickets

[MAILPOET-6282]

## Tasks

- [x] I embraced TypeScript

[MAILPOET-6282]: https://mailpoet.atlassian.net/browse/MAILPOET-6282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:add-captcha-register-form)

_The latest successful build from `add-captcha-register-form` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/assets/css/src/components-public/captcha.scss
+++ b/mailpoet/assets/css/src/components-public/captcha.scss
@@ -10,3 +10,10 @@
     }
   }
 }
+
+// WP registration form
+form#registerform .g-recaptcha:not([data-size='invisible']) {
+  scale: 0.9;
+  -webkit-transform-origin: 0 0;
+  transform-origin: 0 0;
+}
--- a/mailpoet/assets/css/src/components-public/captcha.scss
+++ b/mailpoet/assets/css/src/components-public/captcha.scss
@@ -17,3 +17,8 @@ form#registerform .g-recaptcha:not([data-size='invisible']) {
   -webkit-transform-origin: 0 0;
   transform-origin: 0 0;
 }
+
+// WC registration form
+form.woocommerce-form-register .g-recaptcha {
+  padding-inline-start: 3px;
+}
--- a/mailpoet/assets/js/src/public.tsx
+++ b/mailpoet/assets/js/src/public.tsx
@@ -527,5 +527,21 @@ jQuery(($) => {
       }
       $('.mailpoet-manage-subscription .mailpoet-submit-success').hide();
     });
+
+    // reCAPTCHA on WP registration form
+    if ($('.g-recaptcha[data-size="invisible"]').length) {
+      const wpRegisterForm: JQuery<HTMLFormElement> = $('form#registerform');
+      if (wpRegisterForm.length) {
+        wpRegisterForm.parsley().on('form:submit', () => {
+          void grecaptcha.execute();
+          return false; // disable submission
+        });
+
+        // Triggered by invisible reCAPTCHA
+        window.onInvisibleReCaptchaSubmit = () => {
+          wpRegisterForm[0].submit();
+        };
+      }
+    }
   })();
 });
--- a/mailpoet/assets/js/src/public.tsx
+++ b/mailpoet/assets/js/src/public.tsx
@@ -536,10 +536,13 @@ jQuery(($) => {
       );
 
       let registerForm: JQuery<HTMLFormElement>;
+      let submitInput: object;
       if (wpRegisterForm.length) {
         registerForm = wpRegisterForm;
+        submitInput = { type: 'hidden', name: 'wp-submit', value: 'Register' };
       } else if (wcRegisterForm.length) {
         registerForm = wcRegisterForm;
+        submitInput = { type: 'hidden', name: 'register', value: 'Register' };
       }
 
       if (registerForm) {
@@ -550,6 +553,11 @@ jQuery(($) => {
 
         // Triggered by invisible reCAPTCHA
         window.onInvisibleReCaptchaSubmit = () => {
+          // A workaround to include the submit button's field in form data.
+          // This field is included when the form is submitted by the user.
+          // Yet, it is omitted when the form is submitted programmatically.
+          $('<input>').attr(submitInput).appendTo(registerForm);
+
           registerForm[0].submit();
         };
       }
--- a/mailpoet/assets/js/src/public.tsx
+++ b/mailpoet/assets/js/src/public.tsx
@@ -536,13 +536,13 @@ jQuery(($) => {
       );
 
       let registerForm: JQuery<HTMLFormElement>;
-      let submitInput: object;
+      let submitBtnSelector: string;
       if (wpRegisterForm.length) {
         registerForm = wpRegisterForm;
-        submitInput = { type: 'hidden', name: 'wp-submit', value: 'Register' };
+        submitBtnSelector = 'input[type="submit"]';
       } else if (wcRegisterForm.length) {
         registerForm = wcRegisterForm;
-        submitInput = { type: 'hidden', name: 'register', value: 'Register' };
+        submitBtnSelector = 'button[type="submit"]';
       }
 
       if (registerForm) {
@@ -556,7 +556,15 @@ jQuery(($) => {
           // A workaround to include the submit button's field in form data.
           // This field is included when the form is submitted by the user.
           // Yet, it is omitted when the form is submitted programmatically.
-          $('<input>').attr(submitInput).appendTo(registerForm);
+          const field = registerForm.find(
+            submitBtnSelector,
+          )[0] as HTMLButtonElement;
+          const attrs = {
+            type: 'hidden',
+            name: field.name,
+            value: field.value,
+          };
+          $('<input>').attr(attrs).appendTo(registerForm);
 
           registerForm[0].submit();
         };
--- a/mailpoet/assets/js/src/public.tsx
+++ b/mailpoet/assets/js/src/public.tsx
@@ -528,18 +528,29 @@ jQuery(($) => {
       $('.mailpoet-manage-subscription .mailpoet-submit-success').hide();
     });
 
-    // reCAPTCHA on WP registration form
+    // reCAPTCHA on WP & WC registration form
     if ($('.g-recaptcha[data-size="invisible"]').length) {
       const wpRegisterForm: JQuery<HTMLFormElement> = $('form#registerform');
+      const wcRegisterForm: JQuery<HTMLFormElement> = $(
+        'form.woocommerce-form-register',
+      );
+
+      let registerForm: JQuery<HTMLFormElement>;
       if (wpRegisterForm.length) {
-        wpRegisterForm.parsley().on('form:submit', () => {
+        registerForm = wpRegisterForm;
+      } else if (wcRegisterForm.length) {
+        registerForm = wcRegisterForm;
+      }
+
+      if (registerForm) {
+        registerForm.parsley().on('form:submit', () => {
           void grecaptcha.execute();
           return false; // disable submission
         });
 
         // Triggered by invisible reCAPTCHA
         window.onInvisibleReCaptchaSubmit = () => {
-          wpRegisterForm[0].submit();
+          registerForm[0].submit();
         };
       }
     }
--- a/mailpoet/lib/Captcha/ReCaptchaRenderer.php
+++ b/mailpoet/lib/Captcha/ReCaptchaRenderer.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Captcha;
+
+use MailPoet\Settings\SettingsController;
+use MailPoet\Subscription\Captcha\CaptchaConstants;
+use MailPoet\WP\Functions as WPFunctions;
+
+class ReCaptchaRenderer {
+
+  /** @var SettingsController */
+  private $settings;
+
+  /** @var WPFunctions */
+  private $wp;
+
+  public function __construct(
+    SettingsController $settings,
+    WPFunctions $wp
+  ) {
+    $this->settings = $settings;
+    $this->wp = $wp;
+  }
+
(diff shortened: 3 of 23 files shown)


mailpoet/assets/css/src/components-public/captcha.scss
```diff
@@ -10,3 +10,10 @@
     }
   }
 }
+
+// WP registration form
+form#registerform .g-recaptcha:not([data-size='invisible']) {
```
alex-mpoet, line 15: Can this affect other plugins that may put ReCaptcha on registration forms?

mailpoet/assets/js/src/public.tsx
```diff
@@ -527,5 +527,21 @@ jQuery(($) => {
       }
       $('.mailpoet-manage-subscription .mailpoet-submit-success').hide();
     });
+
+    // reCAPTCHA on WP registration form
+    if ($('.g-recaptcha[data-size="invisible"]').length) {
```
alex-mpoet, line 532: Can this conflict with other plugins that add ReCaptcha to registration forms?

mailpoet/lib/Captcha/ReCaptchaRenderer.php
```diff
@@ -0,0 +1,39 @@
...
+    SettingsController $settings,
+    WPFunctions $wp
+  ) {
+    $this->settings = $settings;
+    $this->wp = $wp;
+  }
+
+  public function render(): string {
+    $captchaSettings = $this->settings->get('captcha');
+    $isInvisible = $captchaSettings['type'] === CaptchaConstants::TYPE_RECAPTCHA_INVISIBLE;
```
alex-mpoet, line 27: I would suggest to check the existence of array keys before using them, e.g. `!empty($captchaSettings['type']) && $captchaSettings['type'] === ...` in order to avoid unnecessary PHP notices in case these settings are absent in a user's DB.

mailpoet/lib/Captcha/ReCaptchaValidator.php
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * @throws \Exception response token is missing or invalid.
+   */
+  public function validate(string $responseToken) {
+    $captchaSettings = $this->settings->get('captcha');
+    if (empty($responseToken)) {
+      throw new \Exception(__('Please check the CAPTCHA.', 'mailpoet'));
+    }
+
+    $secretToken = $captchaSettings['type'] === CaptchaConstants::TYPE_RECAPTCHA
```
alex-mpoet, line 36: Same as before, please check the `type` array key existence before using it.

mailpoet/tests/unit/Captcha/ReCaptchaRendererTest.php
```diff
@@ -0,0 +1,62 @@
...
+  /** @var ReCaptchaRenderer */
+  private $renderer;
+
+  public function _before() {
+    $this->htmlParser = new HtmlParser();
+    $this->settingsMock = $this->createMock(SettingsController::class);
+    $this->renderer = new ReCaptchaRenderer($this->settingsMock, new WPFunctions());
+  }
+
+  public function testRenderingCheckbox() {
```
alex-mpoet: I think we had a convention to name tests in present tense and in 3rd person, e.g. `testItRendersCheckbox`. But it's not a hard rule.

mailpoet/lib/Captcha/ReCaptchaValidator.php
```diff
@@ -0,0 +1,57 @@
...
+        'response' => $responseToken,
+      ],
+    ]);
+
+    if ($this->wp->isWpError($response)) {
+      throw new \Exception(__('Error while validating the CAPTCHA.', 'mailpoet'));
+    }
+
+    /** @var \stdClass $response */
+    $response = json_decode($this->wp->wpRemoteRetrieveBody($response));
```
alex-mpoet, line 52: Theoretically, there can be a JSON decoding error if a server response is invalid, and it's not handled here, so it could throw a warning. But such a case seems not very likely to happen, and it can be given a pass for now.

mailpoet/assets/css/src/components-public/captcha.scss
```diff
@@ -17,3 +17,8 @@ form#registerform .g-recaptcha:not([data-size='invisible']) {
   -webkit-transform-origin: 0 0;
   transform-origin: 0 0;
 }
+
+// WC registration form
+form.woocommerce-form-register .g-recaptcha {
```
alex-mpoet, line 22: Can there be a conflict of styles with other plugins that implement a captcha in the Woo registration form? It would be nice to use an identifier unique to MailPoet.

mailpoet/lib/Config/Hooks.php
```diff
@@ -250,6 +250,24 @@ public function setupSubscriptionEvents() {
         10,
         3
       );
+
+      // reCAPTCHA on WC registration form
+      if (class_exists('WooCommerce')) {
```
alex-mpoet: We have a [helper](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/WooCommerce/Helper.php#L20) for this which unifies WooCommerce detection and comes in handy for testing as it's easy to mock.

mailpoet/assets/js/src/public.tsx
```diff
@@ -536,10 +536,13 @@ jQuery(($) => {
       );
 
       let registerForm: JQuery<HTMLFormElement>;
+      let submitInput: object;
       if (wpRegisterForm.length) {
         registerForm = wpRegisterForm;
+        submitInput = { type: 'hidden', name: 'wp-submit', value: 'Register' };
```
alex-mpoet: Just curious, does the value matter in this case? Because if it does, it would require translation.

mailpoet/lib/Subscription/Pages.php
```diff
@@ -491,7 +491,7 @@ public function getManageLink($params) {
 
     // get label or display default label
     $text = (
-      isset($params['text'])
+    isset($params['text'])
```
alex-mpoet: I think all three lines of the ternary operator should be indented one level deeper, since it's an argument of a multi-line bracketed expression.

alex-mpoet (commented):
Great job @mhadidg! I only have a few minor notes, but nothing major.

mailpoet/assets/css/src/components-public/captcha.scss
mhadidg: Good point. I think so.

There’s a chance of conflict if both our reCAPTCHA and another solution are active at the same time. However, if the MP solution is turned off, it won’t load any CSS or JS on the WP registration form, so there shouldn’t be any issues for other plugins acting on the same form.

When both our solution and another one are active at the same time, that would likely break both rendering and validation. It's not normal to have two reCAPTCHA solution on the same form. I didn’t address the conflict issue intentionally because there are so many options out there. I’d consider it a user mistake—it’s easy to fix by disabling one of the solutions.

mailpoet/assets/js/src/public.tsx
mhadidg: Only when our reCAPTCHA solution is active. The risk of conflict is real, but we're not planning to address it as I explained in the other comment.

mailpoet/lib/Captcha/ReCaptchaRenderer.php
mhadidg: At this point, `captcha.type` is set to either the checkbox or invisible reCAPTCHA option. Otherwise, MP's reCAPTCHA hooks won’t be added, and this function will never called.

This is a precondition for the function. I’d rather avoid bloating the logic with redundant checks.

For the built-in CAPTCHA, though, if `captcha.type` is set to `disabled`, but CAPTCHA is enabled on WP/WC forms, we should default to the built-in solution. In that case, `captcha.type` would be `nil`, right?

mailpoet/lib/Captcha/ReCaptchaValidator.php
mhadidg: My comment above cover this note as well.

mailpoet/tests/unit/Captcha/ReCaptchaRendererTest.php
mhadidg: Love this comment! I was actually expecting to accidentally break some rules and have them fixed during review.

mailpoet/lib/Captcha/ReCaptchaValidator.php
mhadidg: Good point. I think it would render a technical error message in the registration form. I'll fix it.

mailpoet/assets/css/src/components-public/captcha.scss
mhadidg: Again. This only happens when multiple reCAPTCHA solutions are active at the same time, and ours was one of them. When such conflict occur, CSS issues are the least of the worries. They would likely make the registration form totally broken.

mailpoet/assets/js/src/public.tsx
mhadidg: No. This value is rendered by the WP/WC form itself and we should not tamper with it. The backed logic of WP/WC expect it as is.

mailpoet/lib/Subscription/Pages.php
mhadidg: Yeah. This looks worse and a bit odd. There might be a missing rule in `.editorconfig` :thinking:

mhadidg: In your PHPStorm setup, does auto-format indent the first line of the ternary operator statement? If so, could you please export your code style for PHP. There's a lot of settings. I couldn't locate the implicated one.

alex-mpoet: I'm not currently using PHPStorm, so I can't provide you with a config, unfortunately. The logic here is that anything within a multi-line bracketed expression should be indented, be it ternary or not. But here the first line is on the same level as the brackets. Regarding indentation within a ternary expression itself, I'm not sure we enforce this currently. I'd indent all but the first line so the final formatting would look like this:
```
$text = (
  isset($params['text'])
    ? htmlspecialchars($params['text'])
    : __('Manage your subscription', 'mailpoet')
);
```

mailpoet/lib/Captcha/ReCaptchaRenderer.php
alex-mpoet: > Otherwise, MP's reCAPTCHA hooks won’t be added, and this function will never called.

> This is a precondition for the function.

Seems like this is the case, indeed. Then it's good as it is. Thanks for reassuring!

> In that case, captcha.type would be nil, right?

Yes, if captcha is disabled, `type` would be `null`. My note was about a hypothetical case when the whole `captcha` setting is absent, but the `render()` method is called. However, as you explained, it won't be actually called.

mailpoet/assets/js/src/public.tsx
alex-mpoet: I've checked the WP source code. The value is actually [translated](https://core.trac.wordpress.org/browser/trunk/src/wp-login.php#L1191), but here we don't translate it, so it **is** being tampered with. Thankfully, this value is not checked on the backend. But some 3rd-party plugins could use it. It would be good to retain the original value here, if possible. If it's challenging to do, let's skip it for now.

alex-mpoet:
@mhadidg Thanks for the changes! They look good to me. I have only one [remark](https://github.com/mailpoet/mailpoet/pull/5910#discussion_r1836571210).

mailpoet/assets/js/src/public.tsx
mhadidg: Oh. That's true!

I'll adjust the JS code to dynamically extract the field's name and value and pass it as is to the backend.

alex-mpoet (approved):

Aschepikov (approved):
GJ!

PR #5913, opened 2024-11-05
Release 5.3.6

costasovo:
## Preview

### DIFF
--- a/mailpoet/CHANGELOG.md
+++ b/mailpoet/CHANGELOG.md
@@ -1,5 +1,10 @@
 == Changelog ==
 
+= 5.3.6 - 2024-11-05 =
+
+- Fixed: race condition in links tracking;
+- Fixed: WP 6.7 compatibility issues.
+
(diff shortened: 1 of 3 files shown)


websupporter:
https://github.com/mailpoet/mailpoet/pull/5913/files#diff-034b1ff937a5cf059fb10d88f73f8bcf8d836a7f51c98eaec8d0836b854e4fb1L5

reading the changelog: are we tested up to 6.6 or 6.7 now @costasovo

costasovo:
> reading the changelog: are we tested up to 6.6 or 6.7 now @costasovo

@websupporter Afaik, we fixed known issues. @veljkho and @Aschepikov will test the build against current WP 6.7 RC. According to WP's release [schedule](https://make.wordpress.org/core/6-7/) there will be one more RC (today) so I think we should increase tested up to after we test with the last RC.

Aschepikov (approved):

PR #5914, opened 2024-11-05
Extract email editor JS code into a package - MAILPOET-6215

triple0t:
## Description

Extract email editor JS code into a package.

In this PR, we move the email editor files into a JS package, which would make the project easier to consume and distribute.

We did the following in this PR
* Moved the JS files into a package
* Update webpack build process; we are still reusing the current format, i.e., dist files still live in the `mailpoet/assets/dist` folder. We don’t need to change this now until we are ready to start distributing the package
* Setup JS and CSS lint using the WP-scripts
* Migrate to WP JS coding style 
* Setup linters to run to pre-commit

Next 
1. ~Configure prettier for the email editor JS package. We disabled the general mailpoet prettier formatter because the version causes conflicts with WP coding standards. We can setup prettier using the wp-scripts format command~ Done
2. ~Setup tsc typecheck. TSC is currently throwing an error. We need to fix this~ Done
3. We need to rewrite the nps poll feature. It’s not currently compatible with the migrated email editor 
4. We can now use wp script to setup the test infrastructure 

I’m trying to minimize the massive file changes as much as possible. I don’t think there should be a need to move/edit a large number of files again after this PR, but lemme know if you think otherwise. 

**Question**
Translations: I notice we exclude [JS assets](https://github.com/mailpoet/mailpoet/blob/e9f153c7137b6d9632e4a6d2ab38f7e4dd4d83ea/mailpoet/RoboFile.php#L252) from make-pot. Does another process handle this?

**Note**
This PR will block any further changes for the email editor on the repository until it’s merged. 

## Code review notes

Please check each commit.
I added the massive file updates to the git-blame-ignore-revs file.

## QA notes

Please ensure the editor is fully functional and all blocks work as expected.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6215](https://mailpoet.atlassian.net/browse/MAILPOET-6215)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6215]: https://mailpoet.atlassian.net/browse/MAILPOET-6215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:part-one-extract-js-code-for-the-email-editor-into-a-package)

_The latest successful build from `part-one-extract-js-code-for-the-email-editor-into-a-package` will be used. If none is available, the link won't work._


.git-blame-ignore-revs
```diff
@@ -45,3 +45,6 @@ ab27eaee2df740c0add4331a7f8c115a87ecfa2b
 
 # Switch email editor js packages to WP coding style
 d8ba1af97864e373b81dd08bd884055a138e1306
+
+# Email editor Prettier autoformatting
+cdf89c46599e68648a4522d968a0d1948895060a
```
costasovo, line 50: I think that when we rebase the PR, which we do when we rebase and merge the PR here on GitHub, the hash will change because of the rebase. So, we will need a subsequent PR to update these two hashes to new values.

packages/js/email-editor/src/blocks/powered-by-mailpoet/block.tsx
```diff
@@ -4,7 +4,7 @@ import { useBlockProps, InspectorControls } from '@wordpress/block-editor';
 import { PanelBody, RadioControl, Icon } from '@wordpress/components';
```
costasovo: @triple0t  I'm commenting on this file, but it is related to the blocks/powered-by-mailpoet. Would it be possible to return the folder `blocks/powered-by-mailpoet` to the mailpoet/assets? The block is a special block from the MailPoet plugin to be used only in MailPoet.
It would be nice to do this early because with this change we could also get rid of the `engine` folder. So instead of
`packages/js/email-editor/engine/components` we would have only `packages/js/email-editor/components`, etc. Does that make sense to you? Do you agree? I think this would be ideal to do now to avoid the next change in directories that could cause more conflicts.

An alternative approach could be just moving files from the `engine` folder out of the folder to get the final directory structure and move the block back to the MailPoet plugin later.

lysyjan (dismissed):
@triple0t I'd probably merge the PR. I haven't found any major problem and moving forward with this pull request will unblock us from further work.

But I'd like to mention that the ignore hashes won't work because we use rebase and merge.

packages/js/email-editor/src/blocks/powered-by-mailpoet/block.tsx
costasovo: @triple0t Sorry for the late comment! I thought I posted it on Wednesday, but I had it as a pending comment 🤦

.git-blame-ignore-revs
triple0t: Noted. I will create a new PR to update the commit hashes

packages/js/email-editor/src/blocks/powered-by-mailpoet/block.tsx
triple0t: @costasovo Thank you for the suggestion. That would indeed simplify the folder structure.

I have a question about the `powered-by-mailpoet` block. It is currently using data from the [email editor](https://github.com/mailpoet/mailpoet/blob/9be3d76a2c31f9a18b790689eb14a53f362e955a/packages/js/email-editor/src/blocks/powered-by-mailpoet/block.tsx#L35-L40), how would you suggest we handle this?

costasovo: > I have a question about the powered-by-mailpoet block. It is currently using data from the [email editor](https://github.com/mailpoet/mailpoet/blob/9be3d76a2c31f9a18b790689eb14a53f362e955a/packages/js/email-editor/src/blocks/powered-by-mailpoet/block.tsx#L35-L40), how would you suggest we handle this?

@triple0t I see the CDN url and isPremiumPlugin active. I think we need to put those out of the store because both are related only to MailPoet, so either we could create a new store only for MailPoet stuff, or, as these two don't change dynamically, we could read them from a global space for now.

triple0t: True. Noted. I will make the update.
Thanks @costasovo

triple0t:
@lysyjan  The PR is ready. Can we merge it?

costasovo (approved):
@triple0t Thank you for moving the block with the logo back to plugin and for removing the engine folder.
After we merge this we will need to update hashes in .git-blame-ignore-revs because rebase on GH will change them.

PR #5924, opened 2024-11-11
Fix detection of an existing index in the table in migrations [MAILPOET-6314]

lysyjan:
## Description

This PR fixes the incorrect detection of an existing index in a table for MySQL 5.7. 

## Code review notes

The problem is that the detection of existing indexes was changed to catch an error when the plugin tries to create an index with the same name for a non-existing column. Unfortunately, this does not work for MySQL 5.7 and some MariaDB versions because it returns an error about no existing column.
This PR adds a fallback when the DB returns an error about the non-existing column, and the error code matches the expected value.

I also added a new integration test, and this test failed before adding a fix.

## QA notes

Because the tests are passing, I think we can skip QA here.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6314]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6314]: https://mailpoet.atlassian.net/browse/MAILPOET-6314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:migrations-old-mysql)

_The latest successful build from `migrations-old-mysql` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/Migrator/DbMigration.php
+++ b/mailpoet/lib/Migrator/DbMigration.php
@@ -74,6 +74,22 @@ abstract class DbMigration {
     try {
       $this->connection->executeStatement("ALTER TABLE $tableName ADD INDEX $indexName (__non__existent__column__name__)");
     } catch (Exception $e) {
+      // Index creating index failed on not existing column we use a fallback. This can happen on MySQL 5.7 and lower and on some MariaDB versions.
+      if ($e->getCode() === 1072) {
+        $database = $wpdb->dbname;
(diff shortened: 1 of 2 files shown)


triple0t (approved):
LGTM. Well done

PR #5929, opened 2024-11-13
Make send test email functionality independent of mailpoet - MAILPOET-6092

triple0t:
## Description

This PR ensures the send test email functionality is wholly independent of the MailPoet plugin.

I added the following improvements to the email editor PHP packages as part of my work on this PR. If you prefer, I can create a new PR for them.
* I added code-style-fix, which calls the phpcbf utility to automatically fix code style lint errors
* Since we have code-style linting support [here](https://github.com/mailpoet/mailpoet/blob/trunk/packages/php/email-editor/composer.json#L32), I added it as part of the pre-commit hook.

## Code review notes

I have no idea what's going on with GitHub action. I ran the test locally with the `composer integration-test,` and it passed.

Comment out [this line](https://github.com/mailpoet/mailpoet/blob/1b5cb0fe536506121be048718ac12c6240e68267/mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php#L44) to test the core editor send preview email functionality

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6092](https://mailpoet.atlassian.net/browse/MAILPOET-6092)

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:make-send-test-email-functionality-independent-on-mailpoet)

_The latest successful build from `make-send-test-email-functionality-independent-on-mailpoet` will be used. If none is available, the link won't work._

[MAILPOET-6092]: https://mailpoet.atlassian.net/browse/MAILPOET-6092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/php/email-editor/src/Integrations/Utils/class-send-preview-email.php
```diff
@@ -0,0 +1,141 @@
...
+
+	/**
+	 * Sends an email preview.
+	 *
+	 * @param string $to The recipient email address.
+	 * @param string $subject The subject of the email.
+	 * @param string $body The body content of the email.
+	 * @return bool Returns true if the email was sent successfully, false otherwise.
+	 */
+	public function send_email( string $to, string $subject, string $body ): bool {
```
costasovo: This method could be `private`.  I think it is safer to keep methods private if possible to prevent/minimize the chance of unwanted reuse.

packages/php/email-editor/src/Integrations/Utils/class-send-preview-email.php
```diff
@@ -0,0 +1,141 @@
+<?php
+/**
+ * This file is part of the MailPoet plugin.
+ *
+ * @package MailPoet\EmailEditor
+ */
+
+declare( strict_types = 1 );
+
+namespace MailPoet\EmailEditor\Integrations\Utils;
```
costasovo: I'm not sure about the namespace here. I personally think about sending preview as a core functionality, and I would add it as `MailPoet\EmailEditor\Engine\Send_Preview_Email`. I struggle with reasoning why it should be under Integration\Utils it somehow doesn't fit me here. But maybe I have just a different perspective. Do you feel strongly about using this namespace?

costasovo (commented):
@triple0t Excellent job! I tested sending via MailPoet and the editor core. Both worked fine! I added two comments/questions, but they are just details.

Could you please add the new filter to the table with filters in README.md (at the bottom)?

> I added the following improvements to the email editor PHP packages as part of my work on this PR. If you prefer, I can create a new PR for them.

👍 Kudos for adding those. I think it is a good idea. It is ok in this PR.

packages/php/email-editor/src/Integrations/Utils/class-send-preview-email.php
triple0t: Yes, this can be done.

packages/php/email-editor/src/Integrations/Utils/class-send-preview-email.php
triple0t: I agree with you. We can move the file to the `Engine` folder. I don't have a strong opinion on where the file lives.

packages/php/email-editor/src/Integrations/Utils/class-send-preview-email.php
triple0t: Update: I'm keeping this method public for now as we use and override the method in the integration test

packages/php/email-editor/src/Integrations/Utils/class-send-preview-email.php
triple0t: Done in https://github.com/mailpoet/mailpoet/pull/5929/commits/77b3ddef21aca8119ebd74f027689cfefa6e486b

triple0t:
Thank you for the review, @costasovo.

> Could you please add the new filter to the table with filters in README.md (at the bottom)?

Done in https://github.com/mailpoet/mailpoet/pull/5929/commits/77b3ddef21aca8119ebd74f027689cfefa6e486b

> 👍 Kudos for adding those. I think it is a good idea. It is ok in this PR

Noted.

costasovo (approved):
Thank you for the updates! Great job!

PR #5930, opened 2024-11-14
Email Editor - Move custom editor to the post editor page [MAILPOET-6090]

costasovo:
## Description

This PR moves the email editor to `wp-admin/post.php?post={postId}&action=edit` page and removes the custom page.

With this change, it might be easier to switch to a post editor in the future, and in case we use native WP's listing for posts, the edit link will go to the correct editor.

## Code review notes
I kept all the code in MailPoet plugin integration in `MailPoetmailEditor\Integrations\MailPoetditorPageRenderer`. I think it makes sense for now because the assets we enqueue are from the plugin. In the future, we should move some of that code to the PHP package to make things easier for an integrator. 

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6090]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6090]: https://mailpoet.atlassian.net/browse/MAILPOET-6090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/move-to-post-page)

_The latest successful build from `email-editor/move-to-post-page` will be used. If none is available, the link won't work._


mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
```diff
@@ -65,4 +77,14 @@ public function extendEmailPostApi() {
       'schema' => $this->emailApiController->getEmailDataSchema(),
     ]);
   }
+
+  public function replaceEditor($replace, $post) {
+    $currentScreen = get_current_screen();
+    if (use_block_editor_for_post($post) && $post->post_type === self::MAILPOET_EMAIL_POST_TYPE && $currentScreen) { // phpcs:ignore Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps
```
triple0t: [Nitpick]
Since we are already checking for the MailPoet custom post type. Do we still need to use the `use_block_editor_for_post`?  
On checking the [docs](https://developer.wordpress.org/reference/functions/use_block_editor_for_post/), the function has a filter that can be used to disable the block editor for certain (or all post type), is that a behavior we expect and want?

mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
```diff
@@ -38,9 +38,8 @@ const getEditorLink = (newsletter: NewsletterType) => {
     MailPoet.FeaturesController.isSupported('gutenberg_email_editor') &&
     newsletter.wp_post_id
   ) {
-    editorHref = `admin.php?page=mailpoet-email-editor&postId=${newsletter.wp_post_id}`;
+    editorHref = `post.php?post=${newsletter.wp_post_id}&action=edit`;
```
triple0t: We missed [this](https://github.com/mailpoet/mailpoet/blob/7bcfe93269b3532056cf5f6878aa70bff614af0d/mailpoet/assets/js/src/newsletters/send.tsx#L876-L879) check.  We might also use this opportunity to consider the initial option you raised of using a custom function for the Gutenberg email editor URL

mailpoet/lib/Config/Menu.php
```diff
@@ -257,27 +261,6 @@ private function registerMailPoetSubMenuEntries() {
...
-      ]
-    );
-
-    // Add body class for email editor page
-    // We need to mark the page as a block editor page so that some of the block editor styles are applied properly
-    $this->wp->addAction('load-' . $emailEditorPage, function() {
-      $this->wp->addFilter('admin_body_class', function ($classes) {
-        return ltrim($classes . ' block-editor-page');
-      });
-    });
```
triple0t, line 279: We removed the `block-editor-page` class here, but I discovered that the new email editor page does not contain this class.  
Screenshot 2024-11-15 at 11 28 54 AM

Noting it down in case it's important.

triple0t (commented):
This looks good overall. Thank you for working on this. I added some comments, kindly check them below.

Final question from me: Do you think we should consider adding a redirect or something to the new URL if a link like this `/wp-admin/admin.php?page=mailpoet-email-editor&postId=POST_ID` already exists in the wild?

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php
costasovo: > On checking the [docs](https://developer.wordpress.org/reference/functions/use_block_editor_for_post/), the function has a filter that can be used to disable the block editor for certain (or all post type), is that a behavior we expect and want?

Nice catch! We want to always use block editor so the `use_block_editor_for_post` makes no sense. Removed in https://github.com/mailpoet/mailpoet/pull/5930/commits/475ef6d9df01f18c046d6a5cbab8bc4d7560e696

mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
costasovo: I would bet I was searching codebase for `mailpoet-email-editor`. Somehow I missed one. Thanks a lot for double checking. I added a function for the Gutenberg Editor URL in https://github.com/mailpoet/mailpoet/pull/5930/commits/912fe668233c26082f3bca4eb027f5dbefa54271

mailpoet/lib/Config/Menu.php
costasovo: @triple0t  The class `block-editor-page` should be present on the body element. I can see it there. 
Screenshot 2024-11-15 at 15 37 58. It causes visual issues when missing.

I can see on your screenshot that you are focused in `editor-canvas` iframe. Please see just above the console command in your screenshot. This would explain why the query selector didn't find it. Could you please double check?

costasovo:
@triple0t Thank you for the thorough review! I'm happy we caught the issues. They should be addressed in new commits. Could you please take another look?

> Final question from me: Do you think we should consider adding a redirect or something to the new URL if a link like this /wp-admin/admin.php?page=mailpoet-email-editor&postId=POST_ID already exists in the wild?

I wouldn't do that. The email editor is behind a feature flag. Technically, it is possible for users to enable it, but even the page with feature flags is kind of secret and not linked from anywhere.  I don't expect many people to use it, and even if so, it is experimental functionality, so changes like that are expected. My thinking here is that we added the feature flags so that we are allowed to change things this way. Does that make sense to you?

mailpoet/lib/Config/Menu.php
triple0t: I just checked again, and I can confirm the presence of the CSS class. Thank you for pointing out the console.log context

triple0t:
> I wouldn't do that. The email editor is behind a feature flag. Technically, it is possible for users to enable it, but even the page with feature flags is kind of secret and not linked from anywhere. I don't expect many people to use it, and even if so, it is experimental functionality, so changes like that are expected. My thinking here is that we added the feature flags so that we are allowed to change things this way. Does that make sense to you?

Noted. That's fine with me. Well done, @costasovo

triple0t (approved):
LGTM

PR #5934, opened 2024-11-15
Email editor - Move patterns to plugin [MAILPOET-6243]

costasovo:
## Description

The PR moves patterns from the editor package to the plugin. The category registration remains in the editor package.

I think that the patterns and templates shouldn't be part of the engine (except for the blank template). They may carry specific content and assets and should be added by integrators.

## Code review notes

~~I don't feel strongly about adding the new filters for patterns and categories into the engine. As for now, the only thing the engine does with those patterns and categories is register them via WordPress's functions.
At this point, I'm thinking about completely removing the pattern-related functionality from the engine and doing the pattern registration completely in the MailPoet plugin. What do you think about that?~~

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6243]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6243]: https://mailpoet.atlassian.net/browse/MAILPOET-6243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/move-patterns)

_The latest successful build from `email-editor/move-patterns` will be used. If none is available, the link won't work._


triple0t:
@costasovo  Is this PR ready for review?

> I don't feel strongly about adding the new filters for patterns and categories into the engine. As for now, the only thing the engine does with those patterns and categories is register them via WordPress's functions.
At this point, I'm thinking about completely removing the pattern-related functionality from the engine and doing the pattern registration completely in the MailPoet plugin. What do you think about that?

Hmmmm. I think there might be value in keeping the patterns and pattern categories in the email editor package for now. To show a basic working concept of the editor (without MailPoet, i.e., using the package outside the MailPoet plugin), it would benefit the users if they have a starting point with the editor. To this end, we may add a basic email editor native pattern and pattern category. I believe this might also be useful for extenders, as it shows a working example of how to add custom patterns and pattern categories to their project.

For example, without MailPoet, you get a blank screen when selecting a template.
Screenshot 2024-11-15 at 5 06 12 PM

costasovo:
@triple0t Thanks a lot for this feedback! It provides valid points. I'll share what I think about this, and let's see where we get.

> Hmmmm. I think there might be value in keeping the patterns and pattern categories in the email editor package for now. To show a basic working concept of the editor (without MailPoet, i.e., using the package outside the MailPoet plugin), it would benefit the users if they have a starting point with the editor.

There is value in showing the basic working concept. I think we could include a simple demo integration (a demo plugin?) that would show how to set up the editor. You shared a screenshot of the editor running outside the MailPoet, so perhaps you already have such a demo. Do you? Such a demo would also be beneficial for testing and development and to ensure the editor works without MailPoet. What do you think about adding a demo of the email editor plugin to the repo?

> To this end, we may add a basic email editor native pattern and pattern category. I believe this might also be useful for extenders, as it shows a working example of how to add custom patterns and pattern categories to their project.

Initially, I thought that we should keep the editor framework free of any content. My reason was that, as an integrator, I wouldn't like to get demo patterns that might not fit my needs. We could use the demo to showcase adding patterns and categories. 

I was looking at WordPress Core. There are a [a couple of patterns, but primarily for query block](https://github.com/WordPress/wordpress-develop/tree/trunk/src/wp-includes/block-patterns), and I was not able to find any pattern definitions directly in the Gutenberg repository.

On the other hand, having an initial set of categories (e.g., email headers, email footers, email contents) could be helpful so that integrators could reuse those instead of creating their custom categories for everything and potentially creating a mess. As for adding native patterns, I'm not sure. I think that this should be the plugin or theme area. I also think the patterns we have now are probably not a good fit for native patterns as they contain MailPoet unsubscribe links shortcodes and also the image assets that we distribute with the MailPoet plugin. At this point we don't have a way how to distribute content assets in the editor package.

> For example, without MailPoet, you get a blank screen when selecting a template.

I see. In the case when there is nothing to display, we could just skip the modal and start with an empty email. We will focus on templates in the next project so we can add this behavior.

Here is an idea of how things related to patterns, categories, and templates could look like:

1. The editor package defines a set of pattern categories
2. Patterns are added by plugins and themes
3. The editor package contains one template, which is the blank template, that contains only the area for content (no header or footer). This could be used to write emails from scratch and would work as empty content when there are no options for the template select modal.
4. More templates are added by plugins or themes
5. There is a demo that showcases how to integrate the email editor and how to add patterns and templates.

@triple0t What is your opinion about this approach?

costasovo:
> @costasovo Is this PR ready for review?

@triple0t I forgot to answer this. I created it as a draft because I was unsure about the idea after I implemented it. So I would discuss it a bit before eventually making it ready.

triple0t:
Thanks for the in-depth response, @costasovo 

> You shared a screenshot of the editor running outside the MailPoet, so perhaps you already have such a demo. Do you? 

Not right now. I believe we can get something up and running before the end of the week.

> There is value in showing the basic working concept. I think we could include a simple demo integration (a demo plugin?) that would show how to set up the editor. Such a demo would also be beneficial for testing and development and to ensure the editor works without MailPoet.

I agree with you. It would be helpful to us as well.

>  What do you think about adding a demo of the email editor plugin to the repo?

I'm not clear on this question. Which repo do you mean? 

> Initially, I thought that we should keep the editor framework free of any content. My reason was that, as an integrator, I wouldn't like to get demo patterns that might not fit my needs. We could use the demo to showcase adding patterns and categories.
> 
> I was looking at WordPress Core. There are a [a couple of patterns, but primarily for query block](https://github.com/WordPress/wordpress-develop/tree/trunk/src/wp-includes/block-patterns), and I was not able to find any pattern definitions directly in the Gutenberg repository

I don't have a strong opinion on this. My initial thought behind the comment was to have a basic pattern support out of the box. If the integrator didn't like it, they could use filters (that would be provided) to remove it.
If you believe this is a better way to go, then let's go with it.


> On the other hand, having an initial set of categories (e.g., email headers, email footers, email contents) could be helpful so that integrators could reuse those instead of creating their custom categories for everything and potentially creating a mess. As for adding native patterns, I'm not sure. I think that this should be the plugin or theme area. I also think the patterns we have now are probably not a good fit for native patterns as they contain MailPoet unsubscribe links shortcodes and also the image assets that we distribute with the MailPoet plugin. At this point we don't have a way how to distribute content assets in the editor package.

Yes, this is accurate. 

> I see. In the case when there is nothing to display, we could just skip the modal and start with an empty email. We will focus on templates in the next project so we can add this behavior.

Noted. 


> Here is an idea of how things related to patterns, categories, and templates could look like:
> 
> The editor package defines a set of pattern categories
Patterns are added by plugins and themes
The editor package contains one template, which is the blank template, that contains only the area for content (no header or footer). This could be used to write emails from scratch and would work as empty content when there are no options for the template select modal.
More templates are added by plugins or themes
There is a demo that showcases how to integrate the email editor and how to add patterns and templates.
@triple0t What is your opinion about this approach?

I love the drilled-down and well-detailed approach.
Yes, please. Let's go with this suggestion.


> I created it as a draft because I was unsure about the idea after I implemented it. So I would discuss it a bit before eventually making it ready.

Okay, I'm fine with that.

costasovo:
@triple0t Thanks for the exhaustive response!

>> You shared a screenshot of the editor running outside the MailPoet, so perhaps you already have such a demo. Do you?
> 
> Not right now. I believe we can get something up and running before the end of the week.

We don't have to rush. I thought you might already have something, that we could just commit. It is important that we agree that this would be beneficial so we can propose it as a task. 

>> What do you think about adding a demo of the email editor plugin to the repo?
>
> I'm not clear on this question. Which repo do you mean?

I would add it to the MailPoet monorepo. I'm not sure where. Perhaps it could go to the root as email-editor-demo?

> I don't have a strong opinion on this. My initial thought behind the comment was to have a basic pattern support out of the box. If the integrator didn't like it, they could use filters (that would be provided) to remove it.
If you believe this is a better way to go, then let's go with it.

I'm not strictly against adding patterns to the editor core. I would start from the plugin side, and in the future, we could pick some patterns and create a pattern library for the core. I think it is safer to add them later.

> I love the drilled-down and well-detailed approach.
Yes, please. Let's go with this suggestion.

I'm happy you agree. I want to change this PR to follow the outlined approach. We are at the beginning and we will revisit patterns in the future, so at this phase, I will:
1) Keep pattern categories in the email editor package, but I'll add only a single category, "Email Contents".
2) I'll discard the filters for adding more categories and patterns I added in this PR.
3) I'll move registration for our two content patterns directly to the MailPoet plugin code.

I think this is aligned with the approach we agreed on. Do you agree?

triple0t:
> I think this is aligned with the approach we agreed on. Do you agree?

@costasovo Yes, I agree. Thank you

> I would add it to the MailPoet monorepo. I'm not sure where. Perhaps it could go to the root as email-editor-demo?

I don't think we should put it in the MailPoet monorepo, as it already has a bunch of stuff. Let me think about this again and get back to you.

costasovo:
> I don't think we should put it in the MailPoet monorepo, as it already has a bunch of stuff. Let me think about this again and get back to you.

@triple0t I think that's the feature of a monorepo that it could contain a bunch of stuff. One advantage I see is that it would allow to use the demo plugin also for development the same way we use MailPoet now. And we could add changes to the editor and demo in one PR.

triple0t:
> I think that's the feature of a monorepo that it could contain a bunch of stuff. One advantage I see is that it would allow to use the demo plugin also for development the same way we use MailPoet now. And we could add changes to the editor and demo in one PR.

@costasovo Okay. Let's go for it

costasovo:
@triple0t I updated the PR based on the discussion. The last two commits address the three points:

>1. Keep pattern categories in the email editor package, but I'll add only a single category, "Email Contents".
>2. I'll discard the filters for adding more categories and patterns I added in this PR.
>3. I'll move registration for our two content patterns directly to the MailPoet plugin code.

I'm marking it as ready to review.

triple0t (approved):
LGTM

PR #5939, opened 2024-11-19
Fix email editor js package nps poll feature - MAILPOET-6315

triple0t:
## Description

In this PR, we extend the MailPoet NpsPoll feature to the Email editor package using WordPress filters.

Due to the separation of concern we are currently trying to keep, i.e., MailPoet-related components within MailPoet and Email editor components within the email editor, we need a way for both projects to communicate. One way is using the WordPress filters. Another way is publishing shared components that can then be reused within both projects. We are sticking with the first option for now.

## Code review notes

If the `satismeter` poll is not showing up for you. You can edit this line https://github.com/mailpoet/mailpoet/blob/bc1bf9929a2ff4c849114496c0b89c8c4f42b549/mailpoet/assets/js/src/nps-poll.jsx#L48

Use any value (number) as the `userId`.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6315](https://mailpoet.atlassian.net/browse/MAILPOET-6315)

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:fix-email-editor-js-package-nps-poll-feature)

_The latest successful build from `fix-email-editor-js-package-nps-poll-feature` will be used. If none is available, the link won't work._

[MAILPOET-6315]: https://mailpoet.atlassian.net/browse/MAILPOET-6315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/js/email-editor/src/editor.tsx
```diff
@@ -32,6 +32,11 @@ function Editor() {
 	);
 }
 
+const WrappedEditor = applyFilters(
+	'mailpoet_email_editor_the_editor_component',
```
lysyjan: I think we could find a better name for the wrapping filter. What do you think?
From my experience, the determiner `the` is not usually used in filter names.
What do you think about the name `mailpoet_email_editor_wrap_editor_component `?

packages/js/email-editor/src/editor.tsx
```diff
@@ -42,5 +47,6 @@ export function initialize( elementId: string ) {
 	initBlocks();
 	initHooks();
 	const root = createRoot( container );
-	root.render( <Editor /> );
+	// @ts-ignore
```
lysyjan: It would be good to add a comment on why we need to use `@ts-ignore`. Please, can you add it?

lysyjan (commented):
Good job @triple0t. I have only two minor remarks.

packages/js/email-editor/src/editor.tsx
triple0t: Noted. Done in https://github.com/mailpoet/mailpoet/pull/5939/commits/3f74dacb51aef9724547a819098f911c4a58c502

packages/js/email-editor/src/editor.tsx
triple0t: Hmmm, that was because we passed the editor component to the filter. I have updated the filter name in https://github.com/mailpoet/mailpoet/pull/5939/commits/3f74dacb51aef9724547a819098f911c4a58c502

triple0t:
I have made the update

lysyjan (approved):

PR #5940, opened 2024-11-19
Add built-in CAPTCHA to WP & WC register forms [MAILPOET-6325] 

mhadidg:
## Description

I did a significant refactoring of the built-in CAPTCHA system to decouple it from MP's subscription forms, enabling broader use cases, specifically for integration with WordPress and WooCommerce registration forms.

Refer to the [parent ticket](https://mailpoet.atlassian.net/browse/MAILPOET-6282) for description.

### Refactoring

- Added a [dedicated endpoint](https://github.com/mailpoet/mailpoet/commit/1ba755273b9102e84ce859fa8170af66618236e4) router specifically for CAPTCHA handling. This was necessary to...

  - Decouple the built-in CAPTCHA system from the subscription namespace.
  - Adjust the URLs used for CAPTCHA rendering and validation to accommodate a wider range of forms beyond MP's subscription forms.

- Moved all CAPTCHA-related code to the `MailPoet/Captcha` directory.

### Integrating with WP/WC form

As the built-in CAPTCHA is handled in a dedicated page, the integration was a bit comptex. This is my attempt to explain, in high level, how the integration works.

In rendering, a couple of control fields has been added. These fields utilized later by the CAPTCHA form renderer. The original form action URL is [rewritten](https://github.com/mailpoet/mailpoet/commit/73a7a8f878a2f7769cd59c310eb90977ae22b9b2#diff-fcfc6b9f9e220bcdb069ac94b6928c3e42a04e8bbea22a90b565cddac7647299R45-R46) to submit the form data to a new AJAX endpoint. For example, in WP registration form:

```
# original form URL
/wp-login.php?action=register

# rewritten form URL
/wp-admin/admin-ajax.php
  action: mailpoet
  endpoint: captcha
  method: render
  api_version: v1
  // control fields for the renderer
  data[referrer_form]: wp_register_form
  data[referrer_form_url]: /wp-login.php?action=register
  // plus original form fields wrapped in data[] format
```

The [AJAX endpoint](https://github.com/mailpoet/mailpoet/commit/8a57b477041ba85a29e516386bfcc2f9077a59db) processes the incoming form data and redirects to the new [CAPTCHA router](https://github.com/mailpoet/mailpoet/commit/1ba755273b9102e84ce859fa8170af66618236e4):

```
/?mailpoet_page=template&mailpoet_router&endpoint=captcha&action=render&data=(...)
```

The `data` param contains the original form inputs (from the WP or WC registration form), along with MP-specific control fields, encoded in Base64:

```
{
  // original form fields
  "user_login": "test",
  "user_email": "test@example.com",
  "redirect_to": "",
  "wp-submit": "Register",
  // control fields attached to the original WP/WC form
  "referrer_form": "wp_register_form",
  "referrer_form_url": "/wp-login.php?action=register",
  // added by the AJAX endpoint
  "captcha_session_id": "5xkzsc5zlowsso0cc4gwsogwc4wkc8ok"
}
```

For the challenge page, the built-in CAPTCHA [renders](https://github.com/mailpoet/mailpoet/commit/73a7a8f878a2f7769cd59c310eb90977ae22b9b2#diff-108c5730a3439bf00062a3e0a864ed49fa5268cc479638c1b66ed0d96e8d6d52R112-R129) a form that submits to the original action URL. This form includes hidden fields containing the original form data, the CAPTCHA session ID, and the user's CAPTCHA input for validation:

```
/wp-login.php?action=register
  // CAPTCHA fields
  data[captcha_session_id]: <session-id>
  data[captcha]: <user-input>
  // original form fields
  user_login: test
  user_email: test@example.com
  ...
```

A [validation hook](https://github.com/mailpoet/mailpoet/commit/162585bf5f18cc9dbfc0b9469dc7ba39539dd343) is attached to the registration flow to verify the CAPTCHA input before processing the registration:

- If the CAPTCHA is valid, the registration proceeds as normal.
- If the CAPTCHA is invalid, an error is returned, and the user is prompted to retry.

## Code review notes

The setting for this feature has not been introduced here. Instead, I've added a `MP_ENV` env variable to activate the logic in testing and development only.

Reviewing by commit is recommended.

## QA notes

This built-in CAPTCHA should be active ONLY when "Built-in CAPTCHA" option of "Protect your forms against spam signups" setting is selected.

## Linked tickets

[MAILPOET-6325](https://mailpoet.atlassian.net/browse/MAILPOET-6325)

## After-merge notes

_N/A_

## Tasks

- [x] I embraced TypeScript

[MAILPOET-6325]: https://mailpoet.atlassian.net/browse/MAILPOET-6325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:add-builtin-captcha-register-form)

_The latest successful build from `add-builtin-captcha-register-form` will be used. If none is available, the link won't work._


mailpoet/lib/Captcha/CaptchaUrlFactory.php
```diff
@@ -0,0 +1,74 @@
...
+      CaptchaEndpoint::ACTION_AUDIO,
+      [
+        'cacheBust' => time(),
+        'captcha_session_id' => $sessionId,
+      ]
+    );
+  }
+
+  private function getUrl(string $action, array $data) {
+    $post = $this->wp->getPost($this->settings->get('subscription.pages.captcha'));
```
triple0t, line 61: Please use a constant for this string or move this string to a constant and use the constant here.

mailpoet/lib/Captcha/Validator/RecaptchaValidator.php
```diff
@@ -1,10 +1,10 @@
...
-namespace MailPoet\Subscription\Captcha\Validator;
+namespace MailPoet\Captcha\Validator;
 
 use MailPoet\Captcha\ReCaptchaValidator as Validator;
 
-class RecaptchaValidator implements CaptchaValidator {
+class RecaptchaValidator {
 
   /** @var Validator */
   private $validator;
```
triple0t, line 10: I don't really agree with this update.

You removed the CaptchaValidator interface here https://github.com/mailpoet/mailpoet/pull/5940/commits/744c59033b1e199c6df760925ead56853234154e and renamed the class here https://github.com/mailpoet/mailpoet/pull/5940/commits/47787ca2b6445e986594af1ad54d7628f7d0c3f2

I believe keeping the `CaptchaValidator` is helpful for code readability and ensures a consistent class interface and experience for other captcha subclasses.

Here, you also renamed `BuiltInCaptchaValidator` (which is more explicit about its intended purpose) with `CaptchaValidator`  - This makes it a bit confusing now that we have just `CaptchaValidator` and `RecaptchaValidator` (which is what? What does it do? etc.).

Additionally, you renamed the class but left both the unit and integrated test classes the same. May I ask why that is so? It can increase some of the confusion I previously highlighted.

If this change does not affect the overall feature, I would suggest reverting it.

mailpoet/lib/Captcha/CaptchaFormRenderer.php
```diff
@@ -21,8 +19,8 @@ class CaptchaFormRenderer {
   /** @var CaptchaPhrase */
   private $captchaPhrase;
 
-  /** @var SubscriptionUrlFactory */
-  private $subscriptionUrlFactory;
+  /** @var CaptchaUrlFactory */
+  private $urlFactory;
```
triple0t: Let's use a descriptive variable name here. Please switch to `$captchaUrlFactory`.
This will aid with readability and reduce a little mental gymnastic

mailpoet/lib/Router/Endpoints/Captcha.php
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public function refresh($data) {
+    $sessionId = $data['captcha_session_id'] ?? null;
+    if (!$sessionId) {
+      return;
+    }
+
+    $this->captchaRenderer->refreshPhrase($sessionId);
+    exit;
```
triple0t: Since this is meant to refresh the captcha, do we need this exit here?

mailpoet/lib/Captcha/CaptchaHooks.php
```diff
@@ -0,0 +1,85 @@
...
+    $this->settings = $settings;
+    $this->captchaValidator = $captchaValidator;
+  }
+
+  public function isEnabled(): bool {
+    // A transient code to enable incremental development of the feature.
+    // Later when a setting is introduced, this function will be adjusted.
+    if (!in_array(getenv('MP_ENV'), ['development', 'test'])) {
+      return false;
+    }
```
triple0t, line 27: I have a question on this workflow. Why not start the development of this feature with the settings page UI and hide the UI display behind an experimental feature flag?

triple0t (commented):
To start, I would like to commend you on the incredible amount of work done in this PR. 
This is amazing, and I'm excited about this feature. Well done.

Please ensure QA takes a look at this PR. They are better equipped to test and consider edge cases.

I dropped some general code comments below, but here are some quick observations.

### What is working well
**Built-in Captcha**
* WordPress account registration page: `/wp-login.php?action=register`
* WooCommerce account registration from my account page `/my-account/`
* Clicking Manage your subscription link on an old email --> ensure I am redirected to the appropriate page
* Ensure clicking on the unsubscribe link from an email newsletter goes to the correct page.

**ReCaptcha checkbox**
* Works on WordPress registration page: `/wp-login.php?action=register`

**ReCaptcha invisible**
* Works on WordPress registration page: `/wp-login.php?action=register`
* Works on WooCommerce account registration page `/my-account/`  (Not sure if it's working, but I didn't get an error. The Google recaptcha logo was showing)
Screenshot 2024-11-21 at 4 46 51 PM
*Image: WooCommerce my account page when using ReCaptcha invisible*



### Issues I found from my testing:
**Built-in Captcha**
* Not working when using the new Woo block: enable guest registration, attempt checkout, and select create a new account. See no captcha check (No errors)
* Same as using the Old checkout Shortcode

**ReCaptcha checkbox**
* Not working on WooCommerce account registration page `/my-account/` (Errored out -> Unable to register a new account)
Screenshot 2024-11-21 at 4 33 34 PM
*Image: WooCommerce my account page when using ReCaptcha checkbox*

* Not working on WooCommerce checkout page (create account process) both Shortcode and Block-based checkout (No errors were observed as well)
Screenshot 2024-11-21 at 4 40 12 PM
*Image: WooCommerce checkout page (Block-based checkout) when using ReCaptcha checkbox*


**ReCaptcha invisible**
* Not working on WooCommerce checkout page (create account process), both Shortcode and Block-based checkout (No errors were observed)
Screenshot 2024-11-21 at 4 48 37 PM
*Image: WooCommerce checkout page (Shortcode-based checkout) when using ReCaptcha invisible*


### General questions and thoughts
* Have we thought about how this feature will behave on WordPress multi-site? I know we don't officially support WP multi-site, but I'm just curious if we have tested and confirmed no errors.
* What about if a user has a plugin that already adds a captcha solution, e.g., Google Recaptcha?
* Would this feature work seamlessly with membership or other custom user registration plugins? For example, MailPoet offers support for membership (powered by WooCommerce Memberships)
* Would this feature work if a user uses a custom registration page?
* Would this feature work with block-based (block-powered) sites? I confirmed its useability with the WooCommerce checkout block above.
* Should we enable this generally based on only the indicated setting option or when the user sends all site emails with MailPoet?
* Do we need to be concerned about old links like `/?mailpoet_page=subscriptions&mailpoet_router&endpoint=subscription&action=manage&data=eyJ0b2t`. This would not work with this PR.

mailpoet/lib/Captcha/Validator/RecaptchaValidator.php
mhadidg: > You removed the CaptchaValidator interface here https://github.com/mailpoet/mailpoet/commit/744c59033b1e199c6df760925ead56853234154e and renamed the class here https://github.com/mailpoet/mailpoet/commit/47787ca2b6445e986594af1ad54d7628f7d0c3f2

Absolutely. 

I made the change in one commit initially, but during the rebase, I had to fix a conflict in the first commit. Not sure what went wrong, but some of the changes made it to the next commit :man_shrugging: 

> I believe keeping the CaptchaValidator is helpful for code readability and ensures a consistent class interface and experience for other captcha subclasses.

I see your point. I don't have a hard opinion here, happy to revert that :wink: 

I have a slight preference to keep the change though. Ensuring a "single consistent method name" between polymorphic classes doesn't sound like a good reason for an interface. To me, interfaces are meant for polymorphic references. Yet. If your argument aligns with similar instances elsewhere in the codebase, I'd favor consistency above all.

> Here, you also renamed BuiltInCaptchaValidator (which is more explicit about its intended purpose) with CaptchaValidator - This makes it a bit confusing now that we have just CaptchaValidator and RecaptchaValidator (which is what? What does it do? etc.).

I'd reach the same conclusion if I looked at it in isolation. Yet. Every other class related to the built-in CAPTCHA start with `Captcha`. I believe this change maintains consistency:

- CaptchaConstants.php
- CaptchaPhrase.php
- CaptchaRenderer.php
- CaptchaSession.php
- CaptchaFormRenderer.php

> Additionally, you renamed the class but left both the unit and integrated test classes the same. May I ask why that is so? It can increase some of the confusion I previously highlighted.

Oops, I missed that. I think I relied too much on the IDE's refactoring tool.

mailpoet/lib/Captcha/CaptchaUrlFactory.php
mhadidg: A constant is not really needed if the value is referenced only once. That's my go to. Constants add a layer of indirection. Why do you think it's a good idea in this particular case? :thinking:

mailpoet/lib/Router/Endpoints/Captcha.php
mhadidg: Not really. Good catch :+1:

mailpoet/lib/Captcha/CaptchaHooks.php
mhadidg: Didn't cross my mind. That's a much better approach actually.

Now that I have taken a different path, it's not super easy nor convenient to shift gears. I'll keep this in mind though for the future :rocket:

mailpoet/lib/Captcha/CaptchaUrlFactory.php
triple0t: Oh, I saw that string used multiple times. Searching again, and I could only find a single instance of it. It's okay in this case. 👍🏽

mailpoet/lib/Captcha/Validator/RecaptchaValidator.php
triple0t: > I have a slight preference to keep the change though. Ensuring a "single consistent method name" between polymorphic classes doesn't sound like a good reason for an interface. To me, interfaces are meant for polymorphic references. Yet. If your argument aligns with similar instances elsewhere in the codebase, I'd favor consistency above all.

I may not fully understand what you mean by `polymorphic classes` and `polymorphic references`. 
From my understanding of [Polymorphism](https://en.wikipedia.org/wiki/Polymorphism_(computer_science)), it appears we are using it correctly.

I believe it's nice to align with the rest of the use case in the codebase.

> I'd reach the same conclusion if I looked at it in isolation. Yet. Every other class related to the built-in CAPTCHA start with Captcha. I believe this change maintains consistency:
> 
> CaptchaConstants.php
CaptchaPhrase.php
CaptchaRenderer.php
CaptchaSession.php
CaptchaFormRenderer.php


Oh, this makes sense in that case. Are these classes not used by other captcha methods?


> Oops, I missed that. I think I relied too much on the IDE's refactoring tool.

Not a problem. This happens sometimes.

mailpoet/lib/Captcha/CaptchaHooks.php
triple0t: Here is my primary concern: Would QA be able to test this PR and its related sibling PRs?

They generally need to download and upload the mailpoet.zip file to their testing sites.

mhadidg:
That was a comprehensive review. Grateful for your time @triple0t :clap: 

For WooCommerce, **signup on "my account" page is the only support flow**. During checkout is not support for now. For the initial release, the only supported signup forms are:

- WP register form at  `/wp-login.php?action=register`, and
- WC register form at `/my-account` page.

ReCAPTCHA on "my account" page works just fine for me (screenshot below). I'm not sure what might be wrong in your test :man_shrugging: It might be a caching issue. Can you please hard refresh and check?

mp-recapcha-wc-signup

---

> What about if a user has a plugin that already adds a captcha solution, e.g., Google Recaptcha?

I didn’t address the conflict issue intentionally because there are so many options out there. I’d consider it a user mistake—it’s easy to fix by disabling one of the solutions.

Why would someone enable multiple CAPTCHA solutions anyway? Don't seem like a usable scenario to start with.

> Have we thought about how this feature will behave on WordPress multi-site? I know we don't officially support WP multi-site, but I'm just curious if we have tested and confirmed no errors.

Not really. Should I? I'm not sure if there is an easy way to test that.

> Would this feature work seamlessly with membership or other custom user registration plugins? For example, MailPoet offers support for membership (powered by WooCommerce Memberships)
> Would this feature work if a user uses a custom registration page?

Good point. It's not really feasible to ensure compatibility with all other plugins—that would be a huge time sink. 

I'd rather focus on core WP/WC and handle any plugin conflicts as they come up. The feature is for a tiny percentage of customers, so being proactive on this seems like a waste. Even if conflicts exist, do we really want to fix them? They might not even surface given the small user base. Also, if we're made aware of a conflict, we might intentionally decide not to fix it if the scale is not big enough.

Keep in mind, there are alternatives for this feature. If it doesn't work for someone, support may suggest a specialized plugin instead. Problem solved. This fact diminishes the value of fixing any potential conflict.

> Would this feature work with block-based (block-powered) sites? I confirmed its useability with the WooCommerce checkout block above.

This is only relevant for signup during checkout, right? Like I mentioned earlier, that's not a supported flow. At least not for the initial release of this feature.  

> Should we enable this generally based on only the indicated setting option or when the user sends all site emails with MailPoet?

It should be based only on the specific setting option, which will be disabled by default. Support will recommend enabling it only if a customer gets banned due to signup form attacks. There will still be some vulnerable forms, but we hope this would be good enough for most cases.

> Do we need to be concerned about old links like /?mailpoet_page=subscriptions&mailpoet_router&endpoint=subscription&action=manage&data=eyJ0b2t. This would not work with this PR.

Huge one. I think it will break old links. `mailpoet_page=subscriptions` will no longer be valid. It would be a missing page. I'll push a commit to fix that.

mailpoet/lib/Captcha/Validator/RecaptchaValidator.php
mhadidg: Let me explain my argument with an example:

```
<?php

interface Shape {
    public function draw();
}

class Circle implements Shape {
    public function draw() {
        echo "Drawing a Circle\n";
    }
}

class Rectangle implements Shape {
    public function draw() {
        echo "Drawing a Rectangle\n";
    }
}

// Here's a polymorphic reference.
function renderShape(Shape $shape) {
    $shape->draw();
}
```

In this example, `renderShape` expect an interface argument. That's a usage.

The `CaptchaValidator` interface is not referenced anywhere. Absolutely no usage. The relevant logic is not referencing the interface, but implementation classes :shrug:

```
private function validateCaptcha($captchaSettings, $data): array {
  if (empty($captchaSettings['type'])) {
    return [];
  }

  try {
    if ($captchaSettings['type'] === CaptchaConstants::TYPE_BUILTIN) {
      $this->builtInCaptchaValidator->validate($data);
    }
    if (CaptchaConstants::isReCaptcha($captchaSettings['type'])) {
      $this->recaptchaValidator->validate($data);
    }
  } catch (ValidationError $error) {
    return $error->getMeta();
  }
  return [];
}
```

Something like this would justify the interface:

```
private function validateCaptcha(CaptchaValidator $validator): array {
  $this->validate($data);
}
```

I checked the codebase. It seems like interfaces aligns with my argument here. For example, `MailerMethod` [interface](https://github.com/mailpoet/mailpoet/blob/13bf305aeb29bbadd0695ee02a3735e62cc4f21f/mailpoet/lib/Mailer/Methods/MailerMethod.php) is referenced [here](https://github.com/mailpoet/mailpoet/blob/13bf305aeb29bbadd0695ee02a3735e62cc4f21f/mailpoet/lib/Mailer/Mailer.php#L20-L24). That's a usage!

This interface might have been used properly before. This no longer holds.

mailpoet/lib/Captcha/CaptchaHooks.php
mhadidg: Great point. I think so. The same logic was used in a [sibling PR](https://github.com/mailpoet/mailpoet/pull/5910). QA was able to test things there.

Hopefully, that would also be the case here.

triple0t:
Thank you for the well-detailed response, @mhadidg.

Please assign the PR back to me when you're done with the update.

--- 

> For WooCommerce, signup on "my account" page is the only support flow. During checkout is not support for now. For the initial release, the only supported signup forms are:
> 
> WP register form at /wp-login.php?action=register, and
> WC register form at /my-account page.

Noted. 

> ReCAPTCHA on "my account" page works just fine for me (screenshot below). I'm not sure what might be wrong in your test 🤷‍♂️ It might be a caching issue. Can you please hard refresh and check?

I checked again using another test site instance, and I can confirm that this works. 


> I didn’t address the conflict issue intentionally because there are so many options out there. I’d consider it a user mistake—it’s easy to fix by disabling one of the solutions.

> Why would someone enable multiple CAPTCHA solutions anyway? Don't seem like a usable scenario to start with.

This is true. Noted. Not a problem.


> Not really. Should I? I'm not sure if there is an easy way to test that.

Hmmm, I would probably perform a smoke test: install the plugin on a multi-site, activate the feature, and confirm that the site doesn't crash—something like that.

> Good point. It's not really feasible to ensure compatibility with all other plugins—that would be a huge time sink.
> 
> I'd rather focus on core WP/WC and handle any plugin conflicts as they come up. The feature is for a tiny percentage of customers, so being proactive on this seems like a waste. Even if conflicts exist, do we really want to fix them? They might not even surface given the small user base. Also, if we're made aware of a conflict, we might intentionally decide not to fix it if the scale is not big enough.
> 
> Keep in mind, there are alternatives for this feature. If it doesn't work for someone, support may suggest a specialized plugin instead. Problem solved. This fact diminishes the value of fixing any potential conflict.

I totally agree with you. This is okay for the core WP/WC target.


> This is only relevant for signup during checkout, right? Like I mentioned earlier, that's not a supported flow. At least not for the initial release of this feature.

This is noted.


> It should be based only on the specific setting option, which will be disabled by default. Support will recommend enabling it only if a customer gets banned due to signup form attacks. There will still be some vulnerable forms, but we hope this would be good enough for most cases.

Okay. 
I guess my idea behind asking this question was if we wanted to showcase this option or point it to our users if they enable the `send all site email with MailPoet` option.

> Huge one. I think it will break old links. mailpoet_page=subscriptions will no longer be valid. It would be a missing page. I'll push a commit to fix that.

Okay 👌🏿

mailpoet/lib/Captcha/Validator/RecaptchaValidator.php
triple0t: Oh, I see your definition of `usage` now. I was a bit confused about it. Thanks for providing an example.

I disagree with it, though, as the interface is being used.

Please find some of the usages of `CaptchaValidator` here:
* https://github.com/mailpoet/mailpoet/blob/324e02f2cfa0f2d775236c16b41433f904f7f92c/mailpoet/lib/Subscription/Captcha/Validator/RecaptchaValidator.php#L7
*  https://github.com/mailpoet/mailpoet/blob/324e02f2cfa0f2d775236c16b41433f904f7f92c/mailpoet/lib/Subscription/Captcha/Validator/BuiltInCaptchaValidator.php#L12


By the way, `CaptchaValidator` is not a hard stop for me. I was mostly concerned about the `BuiltInCaptchaValidator` rename, which will cause confusion when paired with other captcha options.

mailpoet/lib/Captcha/CaptchaHooks.php
triple0t: Okay. This is noted.

mhadidg:
This is weird. Integration tests are passing locally, but they are mass failing in CI. 

I'll rebase and see if that fixes the issue.

mhadidg:
Hey @triple0t. I just pushed the fixes. 

Because I had to rebase, the new changes are not clearly distinguished. They start here https://github.com/mailpoet/mailpoet/pull/5940/commits/2a37cc70fb65cebaef232850a6a8d291a7ca47ed.

> Hmmm, I would probably perform a smoke test: install the plugin on a multi-site, activate the feature, and confirm that the site doesn't crash—something like that.

I just did using https://jurassic.ninja. It didn't crash. The built-in CAPTCHA on MP forms was working just fine. Couldn't check the logic on signup forms as user registration was disabled.

> [!NOTE]
> Because I reverted the changes made [here](https://github.com/mailpoet/mailpoet/pull/5940/commits/dd3ff1f80500910e1edb75fcbe850e04365c0db3), the MP page (in DB) might be corrupted. Check MP post in `wp_posts` table ([Adminer](http://localhost:8081/?server=db&username=wordpress&db=wordpress&select=wp_posts)). If it carries `template` in `post_name` field. Remove it. Then, deactivate and reactivate the plugin to run the populator, which should set the proper values.

mailpoet/lib/Captcha/Validator/RecaptchaValidator.php
mhadidg: That's cheating. Implementation references doesn't count, not for the purpose of this discussion :wink:

triple0t (dismissed):
LGTM

Aschepikov:
Hi @mhadidg there are some conflicts can you resolve them pls

PR #5943, opened 2024-11-20
Run PHPStan on Email Editor Github actions - MAILPOET-6318

triple0t:
## Description

Run PHPStan on Email Editor Github actions

This file does the magic 
https://github.com/mailpoet/mailpoet/blob/883c9b0a1f478bb323819a4fa098a22092a722e2/packages/php/email-editor/tasks/run-phpstan.php

From Email editor PHP package directory: 
Run
```bash
composer phpstan -- --php-version=70400
```
or
```bash
composer phpstan -- --php-version=80200
```

I also updated the [pre-commit configuration](https://github.com/mailpoet/mailpoet/blob/883c9b0a1f478bb323819a4fa098a22092a722e2/.husky/pre-commit) and made it more streamlined. You can find the project settings here: https://github.com/mailpoet/mailpoet/blob/883c9b0a1f478bb323819a4fa098a22092a722e2/packages/php/email-editor/.lintstagedrc.json

## Code review notes

_N/A_

## QA notes

We can skip QA

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6318](https://mailpoet.atlassian.net/browse/MAILPOET-6318)

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:add-github-action-with-phpstan-in-email-editor-package)

_The latest successful build from `add-github-action-with-phpstan-in-email-editor-package` will be used. If none is available, the link won't work._

[MAILPOET-6318]: https://mailpoet.atlassian.net/browse/MAILPOET-6318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/email-editor-package.yml
+++ b/.github/workflows/email-editor-package.yml
@@ -145,6 +145,13 @@ jobs:
           path: mailpoet/vendor-prefixed
           key: ${{ runner.os }}-vendor-prefixed-${{ matrix.php-version }}-${{ hashFiles('mailpoet/prefixer/composer.lock') }}-${{ hashFiles('mailpoet/composer.json') }}
 
+      - name: Cache Composer vendor for test environment
+        id: composer-tests-env-cache
+        uses: actions/cache@v4
+        with:
+          path: tests_env/vendor
+          key: ${{ runner.os }}-composer-mailpoet-${{ matrix.php-version }}-${{ hashFiles('tests_env/composer.lock') }}-${{ hashFiles('tests_env/composer.json') }}
+
       - name: Cache Composer dependencies for Email Editor
         id: composer-email-editor-cache
         uses: actions/cache@v4
@@ -163,6 +170,12 @@ jobs:
           COMPOSER_DEV_MODE=1 php tools/install.php
           touch .env
         working-directory: mailpoet
+
+      # Install Test Environment dependencies only if the cache was not hit
+      - name: Install test environment dependencies
+        if: steps.composer-tests-env-cache.outputs.cache-hit != 'true'
+        run: ../mailpoet/tools/vendor/composer.phar install
+        working-directory: tests_env
 
       # Install MailPoet dependencies only if the cache was not hit
       - name: Install mailpoet dependencies
--- a/.github/workflows/email-editor-package.yml
+++ b/.github/workflows/email-editor-package.yml
@@ -205,12 +205,6 @@ jobs:
         run: ./tools/vendor/composer.phar dump-autoload
         working-directory: mailpoet
 
-      - name: Run code phpstan 7.4
-        if: ${{ matrix.php-version == '7.4' }}
-        run: ../../../mailpoet/tools/vendor/composer.phar phpstan -- --php-version=70400
-        working-directory: packages/php/email-editor
-
-      - name: Run code phpstan 8.2
-        if: ${{ matrix.php-version == '8.2' }}
-        run: ../../../mailpoet/tools/vendor/composer.phar phpstan -- --php-version=80200
+      - name: Run code phpstan
+        run: ../../../mailpoet/tools/vendor/composer.phar phpstan -- --php-version=${{ matrix.php-version == '7.4' && '70400' || '80200' }}
         working-directory: packages/php/email-editor
--- a/.github/workflows/email-editor-package.yml
+++ b/.github/workflows/email-editor-package.yml
@@ -131,6 +131,27 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Cache Composer vendor dependencies for MailPoet
+        id: composer-mailpoet-cache
+        uses: actions/cache@v4
+        with:
+          path: mailpoet/vendor
+          key: ${{ runner.os }}-composer-mailpoet-${{ matrix.php-version }}-${{ hashFiles('mailpoet/composer.lock') }}-${{ hashFiles('mailpoet/composer.json') }}
+
+      - name: Cache Composer vendor-prefixed dependencies for MailPoet
+        id: vendor-prefixed-cache
+        uses: actions/cache@v4
+        with:
+          path: mailpoet/vendor-prefixed
+          key: ${{ runner.os }}-vendor-prefixed-${{ matrix.php-version }}-${{ hashFiles('mailpoet/prefixer/composer.lock') }}-${{ hashFiles('mailpoet/composer.json') }}
+
+      - name: Cache Composer dependencies for Email Editor
+        id: composer-email-editor-cache
+        uses: actions/cache@v4
+        with:
+          path: packages/php/email-editor/vendor
+          key: ${{ runner.os }}-composer-email-editor-${{ matrix.php-version }}-${{ hashFiles('packages/php/email-editor/composer.lock') }}-${{ hashFiles('mailpoet/composer.json') }}
+
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
@@ -143,9 +164,33 @@ jobs:
           touch .env
         working-directory: mailpoet
 
+      # Install MailPoet dependencies only if the cache was not hit
+      - name: Install mailpoet dependencies
+        if: steps.composer-mailpoet-cache.outputs.cache-hit != 'true'
+        run: ./tools/vendor/composer.phar install
+        working-directory: mailpoet
+
+      # Install Email Editor dependencies only if the cache was not hit
+      - name: Install email-editor dependencies
+        if: steps.composer-email-editor-cache.outputs.cache-hit != 'true'
+        run: ../../../mailpoet/tools/vendor/composer.phar install
+        working-directory: packages/php/email-editor
+
       - name: Install composer dependencies
         run: ../../tools/vendor/composer.phar install
         working-directory: mailpoet/tasks/phpstan
+
+      # Dump Email Editor autoload
+      # This is needed to refresh classmap autoload when the composer cache is hit
+      - name: Dump email-editor autoload
+        run: ../../../mailpoet/tools/vendor/composer.phar dump-autoload
+        working-directory: packages/php/email-editor
+
+      # Dump MailPoet autoload
+      # This is needed to refresh classmap autoload when the composer cache is hit
+      - name: Dump MailPoet autoload
+        run: ./tools/vendor/composer.phar dump-autoload
+        working-directory: mailpoet
 
       - name: Run code phpstan 7.4
         if: ${{ matrix.php-version == '7.4' }}
--- a/.github/workflows/email-editor-package.yml
+++ b/.github/workflows/email-editor-package.yml
@@ -119,3 +119,40 @@ jobs:
       - name: Run code style check
         run: ../../../mailpoet/tools/vendor/composer.phar code-style
         working-directory: packages/php/email-editor
+
+  phpstan-static-analysis:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php-version: ['7.4', '8.2']
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          extensions: gd
+
+      - name: Install tools
+        run: |
+          COMPOSER_DEV_MODE=1 php tools/install.php
+          touch .env
+        working-directory: mailpoet
+
+      - name: Install composer dependencies
+        run: ../../tools/vendor/composer.phar install
+        working-directory: mailpoet/tasks/phpstan
+
+      - name: Run code phpstan 7.4
+        if: ${{ matrix.php-version == '7.4' }}
+        run: ../../../mailpoet/tools/vendor/composer.phar phpstan -- --php-version=70400
+        working-directory: packages/php/email-editor
+
+      - name: Run code phpstan 8.2
+        if: ${{ matrix.php-version == '8.2' }}
+        run: ../../../mailpoet/tools/vendor/composer.phar phpstan -- --php-version=80200
+        working-directory: packages/php/email-editor
--- a/mailpoet/tasks/phpstan/email-editor-phpstan-custom.php
+++ b/mailpoet/tasks/phpstan/email-editor-phpstan-custom.php
@@ -1,0 +1,12 @@
+<?php declare(strict_types = 1);
+
+$config = [];
+$phpVersion = (int)getenv('ANALYSIS_PHP_VERSION') ?: PHP_VERSION_ID;
+$config['parameters']['phpVersion'] = $phpVersion;
+
+// Load PHP 8 specific config
+if ($phpVersion >= 80000) {
+  $config['includes'][] = 'email-editor-php-8-config.neon';
+}
+
+return $config;
--- a/mailpoet/tasks/phpstan/email-editor-phpstan-custom.php
+++ b/mailpoet/tasks/phpstan/email-editor-phpstan-custom.php
@@ -2,11 +2,6 @@
 
 $config = [];
 $phpVersion = (int)getenv('ANALYSIS_PHP_VERSION') ?: PHP_VERSION_ID;
-$config['parameters']['phpVersion'] = $phpVersion;
-
-// Load PHP 8.2 specific config
-if ($phpVersion >= 80200) {
-   $config['includes'][] = 'email-editor-php-8-config.neon';
-}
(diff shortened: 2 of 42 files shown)


.github/workflows/email-editor-package.yml
```diff
@@ -106,3 +106,40 @@ jobs:
...
+          touch .env
+        working-directory: mailpoet
+
+      - name: Install composer dependencies
+        run: ../../tools/vendor/composer.phar install
+        working-directory: mailpoet/tasks/phpstan
+
+      - name: Run code phpstan 7.4
+        if: ${{ matrix.php-version == '7.4' }}
+        run: ../../../mailpoet/tools/vendor/composer.phar phpstan -- --php-version=70400
```
lysyjan: Maybe we could avoid repetition here and use something like this:
```
      run: ../../../mailpoet/tools/vendor/composer.phar phpstan -- --php-version=${{ matrix.php-version | replace('.', '') }}00
```

mailpoet/tasks/phpstan/email-editor-phpstan.neon
```diff
@@ -0,0 +1,25 @@
+parameters:
+  level: 9
+  tmpDir: ../../temp/phpstan
+  bootstrapFiles:
+    - vendor/php-stubs/wordpress-stubs/wordpress-stubs.php
+  inferPrivatePropertyTypeFromConstructor: true
+  checkGenericClassInNonGenericObjectType: false
+  parallel:
+    processTimeout: 300.0
+    maximumNumberOfProcesses: 3 # Static analysis is running on Circle CI medium+ that has 3 CPUs
```
lysyjan: > Static analysis is running on Circle CI medium+ that has 3 CPUs

We don't run static analysis on CircleCI. Because this repository is `public` it should be possible by [the docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) to use 4 CPUs. But this could be good to test.

packages/php/email-editor/src/Engine/Renderer/class-renderer.php
```diff
@@ -87,8 +89,9 @@ public static function get_theme() {
 	 */
 	public function render( \WP_Post $post, string $subject, string $pre_header, string $language, $meta_robots = '' ): array { // phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
 		$template_id = 'mailpoet/mailpoet//' . ( get_page_template_slug( $post ) ? get_page_template_slug( $post ) : 'email-general' );
-		$template    = $this->templates->get_block_template( $template_id );
-		$theme       = $this->templates->get_block_template_theme( $template_id, $template->wp_id );
+		/** @var \WP_Block_Template $template */  // phpcs:ignore
```
lysyjan: It would be nice to add which rule is deactivated here or comment on why we need to deactivate the code sniffer here. There are multiple cases of this situation, so I won't comment on them.

packages/php/email-editor/src/Engine/Renderer/template-canvas.php
```diff
@@ -23,28 +23,28 @@
 ?><!DOCTYPE html>
 <html <?php language_attributes(); ?>>
 <head>
-	<title><?php echo esc_html( $subject ); ?></title>
+	<title><?php echo esc_html( $subject ); // @phpstan-ignore-line ?></title>
```
lysyjan, line 26: Because there is `@phpstan-ignore-line` for almost every line of PHP code, it could be better to add ignore for the file. What do you think?

packages/php/email-editor/src/Engine/Templates/class-utils.php
```diff
@@ -94,13 +94,25 @@ public function build_block_template_from_post( $post ) {
 		return $template;
 	}
 
+	// phpcs:disable
```
lysyjan: We could use `@phpstan-type` here to help us avoid disabling CodeSniffer. You can check how it works [here](https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases).

packages/php/email-editor/tasks/run-phpstan.php
```diff
@@ -0,0 +1,39 @@
+<?php // phpcs:ignoreFile
```
lysyjan: Do we need to ignore the coding standard in this file? It looks quite short, so we could follow WP Coding styles here.

mailpoet/tasks/phpstan/email-editor-phpstan.neon
```diff
@@ -15,6 +17,24 @@ parameters:
   ignoreErrors:
     - '#_before\(\) has no return type specified#' # skip integration test before errors
     - '#Container::get\(\) is not referenced in a parameter.#'
+    -
```
lysyjan, line 22: There are many new errors to ignore and a lot of using `@phpstan-ignore-line` in this commit. I have checked the Container class code, and if we change `@param string|class-string<T> $name` to  ` @param class-string<T> $name`, we can remove almost each of them. This change needs to update the test for this class, but I think it is worth it. Can you please take a closer look?

mailpoet/tasks/phpstan/email-editor-phpstan.neon
lysyjan, line 1: The new PHPStan config file is a part of the MailPoet plugin. Do you think it would be possible to move it to the email editor package? Or have you thought about a follow-up task? For example, similar to what we did with Codeception when we moved the dependencies to the top level for easier usage across our repository and simplified updating those dependencies. But maybe I am missing something.

lysyjan (changes requested):
I added some comments. I also noticed that the job is marked as successful on GitHub actions, but an error is visible under the job overview:

```
Ignored error pattern #Access to an undefined property MailPoet\\EmailEditor\\.*Test::\$tester.# in path /home/runner/work/mailpoet/mailpoet/packages/php/email-editor/tests/integration/* was not matched in reported errors.
```

So, we should address marking the job as successful when there were some errors. Please, can you take a look?

Nevertheless, you did a good job because it is not easy to configure this job from scratch.

.github/workflows/email-editor-package.yml
triple0t: This would not work. 
I tested this version with another version and got an error.
For example, here is the error for this
Screenshot 2024-11-26 at 8 32 56 AM
Screenshot 2024-11-26 at 8 36 20 AM

mailpoet/tasks/phpstan/email-editor-phpstan.neon
triple0t: It should be possible. I will give it a try

packages/php/email-editor/src/Engine/Renderer/class-renderer.php
triple0t: Okay. I will add an extra comment for that.

packages/php/email-editor/src/Engine/Renderer/template-canvas.php
triple0t: Unfortunately, PHPStan does not have an ignore file directive (`@phpstan-ignore-file`) that we can use.
The only way to ignore a file is to exclude it from the .neon config file. 
It shouldn't be necessary in this case. Let me know if you feel otherwise.

.github/workflows/email-editor-package.yml
lysyjan: I asked ChatGPT for this idea, so I apologize for the wrong idea. Because we have an issue with the ignore error related to PHP 8.2 maybe we can create two config files and use `matrix.php-version` as a part of the config name. It is possible to specify the used PHP version in [the configuration](https://phpstan.org/config-reference#phpversion). What do you think about this approach?

packages/php/email-editor/src/Engine/Templates/class-utils.php
triple0t: This will not work with a combination of PHPStan and CodeSniffer. Please check and test.

packages/php/email-editor/tasks/run-phpstan.php
triple0t: Noted. I will implement the update.

packages/php/email-editor/src/Engine/Templates/class-utils.php
lysyjan: I tested it yesterday and it worked. I used `@phpstan-type` in the class comment and then used the type in the annotation.
```
/**
 * Utils class.
 *
 * @phpstan-type TemplateObject object{
 *   slug: string,
 *   id: string,
 *   title: string,
 *   description: string,
 *   path: string,
 *   type: string,
 *   theme: string,
 *   source: string,
 *   post_types: array<string>
 * }
 */
 ```
 and
 `* @param TemplateObject $template_object Template object.`
 With this, were both commands without an error.

File: mailpoet/tasks/phpstan/email-editor-phpstan.neon (hunk 7 above)
Comment (triple0t) [reply mailpoet/tasks/phpstan/email-editor-phpstan.neon:22]: Noted. I will update this

File: mailpoet/tasks/phpstan/email-editor-phpstan.neon
Comment (triple0t) [reply mailpoet/tasks/phpstan/email-editor-phpstan.neon:1]: We can create that as a follow-up task. The config file will still need to be in the general phpstan directory because we import some shared libraries.

File: .github/workflows/email-editor-package.yml (hunk 1 above)
Comment (triple0t) [reply .github/workflows/email-editor-package.yml]: > Because we have an issue with the ignore error related to PHP 8.2 maybe we can create two config files and use matrix.php-version as a part of the config name. It is possible to specify the used PHP version in [the configuration](https://phpstan.org/config-reference#phpversion). What do you think about this approach?

I don't understand this comment. 

Are you suggesting we create two `email-editor-phpstan.neon` config files?

Comment (lysyjan) [reply .github/workflows/email-editor-package.yml]: > Are you suggesting we create two email-editor-phpstan.neon config files?

Yes, exactly. I would include the main config, and in the version-specific configs, I would specify the used PHP version.

What do you think about it?

Comment (triple0t) [reply .github/workflows/email-editor-package.yml]: Hmmm. That could work. Let me give it a try

Comment (triple0t) [reply .github/workflows/email-editor-package.yml]: @lysyjan Instead of creating two `email-editor-phpstan.neon` config files. I'm going to keep the current `email-editor-package.yml` file the same. I believe the way it works is easier to understand and doesn't add an extra overhead.

I will instead update the configuration for `email-editor-phpstan.neon` and allow it to load PHP 7 or PHP 8.2 specific config as it is required.

**Comment by triple0t** [comment]
All fixed. Thanks for the review, @lysyjan

File: packages/php/email-editor/tests/unit/Container_Test.php [hunk 8]
```diff
@@ -66,6 +70,6 @@ public function testExceptionForNonExistingService(): void {
 		$this->expectException( Exception::class );
 		$this->expectExceptionMessage( 'Service not found: non_existing_service' );
 
-		$container->get( 'non_existing_service' );
+		$container->get( 'non_existing_service' );  // @phpstan-ignore-line
```
lysyjan: You can use one of the classes created at the beginning of the test here. The check here is about throwing an error because the class was not set in the Container. When we use a random class or, for example stdClass, we can avoid using `@phpstan-ignore-line`, which would be nice.

packages/php/email-editor/tests/unit/Container_Test.php
```diff
@@ -10,28 +10,32 @@
...
 use PHPUnit\Framework\TestCase;
-use stdClass;
+
+class Simple_Service {} // phpcs:ignore
+
+class Singleton_Service {} // phpcs:ignore
 
 /**
  * Unit test for Container class.
+ * Ignoring Only one object structure is allowed in a file.
```
lysyjan, line 20: Should this comment be a part of the following line with phpcs:ignore?

mailpoet/tasks/phpstan/email-editor-phpstan-custom.php
```diff
@@ -0,0 +1,12 @@
+<?php declare(strict_types = 1);
+
+$config = [];
+$phpVersion = (int)getenv('ANALYSIS_PHP_VERSION') ?: PHP_VERSION_ID;
+$config['parameters']['phpVersion'] = $phpVersion;
+
+// Load PHP 8 specific config
+if ($phpVersion >= 80000) {
```
lysyjan: I guess this won't work on PHP 8.1 because the ignored error in the PHP 8 config is related only to PHP 8.2 and higher. It may not be a big deal, but we should be accurate with those conditions to avoid confusion among other developers because they can use this version.

packages/php/email-editor/src/Engine/Templates/class-utils.php
```diff
@@ -94,25 +94,25 @@ public function build_block_template_from_post( $post ) {
...
-	 *      title: string,
-	 *      description: string,
-	 *      path: string,
-	 *      type: string,
-	 *      theme: string,
-	 *      source: string,
-	 *      post_types: array<string>
-	 *  } $template_object Template object.
+	 * @template TemplateObject
+	 * @phpstan-type TemplateObject object{
```
lysyjan: I may have been inaccurate about this in my previous comment. But I would move `@phpstan-type` definition  to the class comment, like [here](https://github.com/mailpoet/mailpoet/blob/e3f7caea1425b9e02ca186223e42779bde36a6d8/mailpoet/lib/Automation/Engine/Storage/AutomationStorage.php#L12-L14). The readability in this place seems to me be a bit worse.

lysyjan (changes requested):
@triple0t As I mentioned at the end of the previous review. The GitHub Action job for the static analysis is not configured correctly. I tried to create my own branch and [test](https://github.com/mailpoet/mailpoet/actions/runs/12058626516) it. One of the static analyses returned an error, but the whole workflow was marked as successful.

I think we should address this problem to avoid false positive runs.

packages/php/email-editor/tests/unit/Container_Test.php
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/5943/commits/68d976b1fb919949ef1ee044c25b6660f610e5c1

packages/php/email-editor/tests/unit/Container_Test.php
triple0t: The code style will complain about the comment length, the code-style-fix will move it to the next line, and then phpstan will complain about the line all over again. Breaking the loop by moving the comment to the top. I commented on the other usage in https://github.com/mailpoet/mailpoet/pull/5943/commits/68d976b1fb919949ef1ee044c25b6660f610e5c1

packages/php/email-editor/src/Engine/Templates/class-utils.php
triple0t: Since this is a style and objective issue. I would leave it as it is.

mailpoet/tasks/phpstan/email-editor-phpstan-custom.php
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/5943/commits/68d976b1fb919949ef1ee044c25b6660f610e5c1

packages/php/email-editor/src/Engine/Templates/class-utils.php
lysyjan: @triple0t  I think my suggestion is how the @phpstan-type comment should be used.
The usage in the class comment used in [the official documentation](https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases) and also in other places in our codebase: [#](https://github.com/mailpoet/mailpoet/blob/e3f7caea1425b9e02ca186223e42779bde36a6d8/mailpoet/lib/Automation/Engine/Storage/AutomationStorage.php#L12-L14), [#](https://github.com/mailpoet/mailpoet/blob/0bbdf4b47fe8a4380ff3723ada4eddbe4f70e822/mailpoet/lib/Mailer/MailerLog.php#L8-L24), [#](https://github.com/mailpoet/mailpoet/blob/ba728c5612673f71c0200e609deff52bc722f149/mailpoet/lib/Migrator/Migrator.php).

triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/5943/commits/883c9b0a1f478bb323819a4fa098a22092a722e2

triple0t:
Thanks for the comment, @lysyjan.
All issues resolved.

lysyjan (approved):

PR #5948, opened 2024-11-22
Email editor - Document usage of private APIs from WP [MAILPOET-6319]

costasovo:
## Description

I searched for all unlock usages and moved them into a single file, `private-apis/index.ts` and removed the unlock export so that we can't use it in any other place. The goal is to have all private APIs we use in one place and with comments. I found a couple of cases where we could get rid of the unlock call and I removed one unnecessary private hook. 

## Code review notes

The very first commit is not related to the private APIs. To work with the newest version of Gutenberg, you need to set `template-locked` mode. I ran into that because I was testing with the new Gutenberg, so I fixed it immediately. I hope it is ok to include it here without an additional ticket.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6319]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6319]: https://mailpoet.atlassian.net/browse/MAILPOET-6319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/private-apis)

_The latest successful build from `email-editor/private-apis` will be used. If none is available, the link won't work._


packages/js/email-editor/src/private-apis/index.ts
```diff
@@ -0,0 +1,73 @@
...
+	// @ts-expect-error No types for privateApis.
+	privateApis as blockEditorPrivateApis,
+} from '@wordpress/block-editor';
+import { privateApis as componentsPrivateApis } from '@wordpress/components';
+import { store as coreStore } from '@wordpress/core-data';
+
+const { unlock } = __dangerousOptInToUnstableAPIsOnlyForCoreModules(
+	'I acknowledge private features are not for use in themes or plugins and doing so will break in the next version of WordPress.',
+	'@wordpress/edit-post' // The module name must be in the list of allowed, so for now I used the package name of the post editor
+);
```
triple0t, line 15: I noticed you switched from using the import [here](https://github.com/mailpoet/mailpoet/blob/e728e74d1166a5e572e90aee105a829d4ba0bc30/packages/js/email-editor/src/private-apis/lock-unlock.js). Instead you redefined the same variable. May I ask why that is so? Won't it be better to have a single unlock interface?

packages/js/email-editor/src/private-apis/index.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * WordPress dependencies
+ */
+import { __dangerousOptInToUnstableAPIsOnlyForCoreModules } from '@wordpress/private-apis';
+import {
+	// @ts-expect-error No types for privateApis.
+	privateApis as blockEditorPrivateApi,
+	// @ts-expect-error No types for privateApis.
+	privateApis as blockEditorPrivateApis,
+} from '@wordpress/block-editor';
```
triple0t: There is a single `privateApis` [export](https://github.com/WordPress/gutenberg/blob/9b0e2d96679ab6c82033946bead9521529e6b29a/packages/block-editor/src/index.js#L23). You can simplify this section.

triple0t (commented):
LGTM. Thank you for working on this.

I have minor remarks; kindly locate them below.

packages/js/email-editor/src/private-apis/index.ts
costasovo: Yes. This was intentional. I removed the other one in [a later commit](https://github.com/mailpoet/mailpoet/pull/5948/commits/e427d17cf3ba4df6ec79e862de07bee078ff5a5c). This unlock is not exported. This is for us to "prevent" adding new private API usage directly and "force" us to do all unlocking in this place. Does that make sense?

packages/js/email-editor/src/private-apis/index.ts
costasovo: Nice catch.  This is not intentional.

packages/js/email-editor/src/private-apis/index.ts
triple0t: In that case, can you delete this file https://github.com/mailpoet/mailpoet/blob/e427d17cf3ba4df6ec79e862de07bee078ff5a5c/packages/js/email-editor/src/private-apis/lock-unlock.js

costasovo: Aha. I see I have 2 now 🤦

costasovo: Thank you for noticing this. I missed that one.

packages/js/email-editor/src/private-apis/index.ts
triple0t: You need to update this too https://github.com/mailpoet/mailpoet/blob/1bf95f4b0a554e275006b24940e5a7e21a7cae31/packages/js/email-editor/src/private-apis/index.ts#L62

from `blockEditorPrivateApi`
to `blockEditorPrivateApis`

costasovo:
@triple0t I fixed the issues and it should be ready for second review.

triple0t (approved):
LGTM
Thank you

PR #5955, opened 2024-11-25
Add basic Personalization Tags support in the Gutenberg Email Editor [MAILPOET-6328]

lysyjan:
## Description

This PR adds the possibility of registering Personalization Tags implemented via HTML comments. Registered tags need to specify the name, tag, category, callback, and attributes.

## Code review notes

1. All new classes in the email editor package are covered with tests, but the implementation in the MailPoet plugin no. I want to focus on tests in the plugin in one of the follow-up tickets when we add more tags.
2. Replacement in attributes will be covered in one of the next PRs.
3. Insertion via UI will be added later.

## QA notes

- QA is not needed here or is optional by the reviewer.
- For testing, the Gutenberg plugin needs to be activated with the build from the latest trunk.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6328]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6328]: https://mailpoet.atlassian.net/browse/MAILPOET-6328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/tags-api)

_The latest successful build from `email-editor/tags-api` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/EmailEditor/Integrations/MailPoet/PersonalizationTagManager.php
+++ b/mailpoet/lib/EmailEditor/Integrations/MailPoet/PersonalizationTagManager.php
@@ -1,0 +1,42 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Integrations\MailPoet;
+
+use MailPoet\EmailEditor\Engine\PersonalizationTags\Personalization_Tags_Registry;
+use MailPoet\EmailEditor\Integrations\MailPoet\PersonalizationTags\Subscriber;
+
+class PersonalizationTagManager {
+  private Subscriber $subscriber;
+
+  public function __construct(
+    Subscriber $subscriber
+  ) {
+    $this->subscriber = $subscriber;
+  }
+
+  public function initialize() {
+    add_filter('mailpoet_email_editor_register_personalization_tags', function( Personalization_Tags_Registry $registry ): Personalization_Tags_Registry {
+      $registry->register(
+        __('First Name', 'mailpoet'),
+        'mailpoet/subscriber-firstname',
+        __('Subscriber', 'mailpoet'),
+        [$this->subscriber, 'getFirstName'],
+        ['default' => __('subscriber', 'mailpoet')],
+      );
+      $registry->register(
+        __('Last Name', 'mailpoet'),
+        'mailpoet/subscriber-lastname',
+        __('Subscriber', 'mailpoet'),
+        [$this->subscriber, 'getLastName'],
+        ['default' => __('subscriber', 'mailpoet')],
+      );
+      $registry->register(
+        __('Email', 'mailpoet'),
+        'mailpoet/subscriber-email',
+        __('Subscriber', 'mailpoet'),
+        [$this->subscriber, 'getEmail'],
+      );
+      return $registry;
+    });
+  }
+}
--- a/mailpoet/lib/EmailEditor/Integrations/MailPoet/PersonalizationTagManager.php
+++ b/mailpoet/lib/EmailEditor/Integrations/MailPoet/PersonalizationTagManager.php
@@ -2,6 +2,7 @@
 
 namespace MailPoet\EmailEditor\Integrations\MailPoet;
 
+use MailPoet\EmailEditor\Engine\PersonalizationTags\Personalization_Tag;
 use MailPoet\EmailEditor\Engine\PersonalizationTags\Personalization_Tags_Registry;
 use MailPoet\EmailEditor\Integrations\MailPoet\PersonalizationTags\Subscriber;
 
@@ -16,26 +17,26 @@ class PersonalizationTagManager {
 
   public function initialize() {
     add_filter('mailpoet_email_editor_register_personalization_tags', function( Personalization_Tags_Registry $registry ): Personalization_Tags_Registry {
-      $registry->register(
+      $registry->register(new Personalization_Tag(
         __('First Name', 'mailpoet'),
         'mailpoet/subscriber-firstname',
         __('Subscriber', 'mailpoet'),
         [$this->subscriber, 'getFirstName'],
         ['default' => __('subscriber', 'mailpoet')],
-      );
-      $registry->register(
+      ));
+      $registry->register(new Personalization_Tag(
         __('Last Name', 'mailpoet'),
         'mailpoet/subscriber-lastname',
         __('Subscriber', 'mailpoet'),
         [$this->subscriber, 'getLastName'],
         ['default' => __('subscriber', 'mailpoet')],
-      );
-      $registry->register(
+      ));
+      $registry->register(new Personalization_Tag(
         __('Email', 'mailpoet'),
         'mailpoet/subscriber-email',
         __('Subscriber', 'mailpoet'),
         [$this->subscriber, 'getEmail'],
-      );
+      ));
       return $registry;
     });
   }
--- a/packages/php/email-editor/src/Engine/PersonalizationTags/class-html-tag-processor.php
+++ b/packages/php/email-editor/src/Engine/PersonalizationTags/class-html-tag-processor.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * This file is part of the MailPoet plugin.
+ *
+ * @package MailPoet\EmailEditor
+ */
+
+declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Engine\PersonalizationTags;
+
+use WP_HTML_Tag_Processor;
+use WP_HTML_Text_Replacement;
+
+/**
+ * Class based on WP_HTML_Tag_Processor which is extended to replace
+ * tokens with their values in the email content.
+ */
+class HTML_Tag_Processor extends WP_HTML_Tag_Processor {
+	/**
+	 * List of deferred updates.
+	 *
+	 * @var WP_HTML_Text_Replacement[]
+	 */
+	private $deferred_updates = array();
+
+	/**
+	 * Replaces the token with the new content.
+	 *
+	 * @param string $new_content The new content to replace the token.
+	 */
+	public function replace_token( string $new_content ): void {
+		$this->set_bookmark( 'here' );
+		$here                     = $this->bookmarks['here'];
+		$this->deferred_updates[] = new WP_HTML_Text_Replacement(
+			$here->start,
+			$here->length,
+			$new_content
+		);
+	}
+
+	/**
+	 * Flushes the deferred updates to the lexical updates.
+	 */
+	public function flush_updates(): void {
+		foreach ( $this->deferred_updates as $key => $update ) {
+			$this->lexical_updates[] = $update;
+			unset( $this->deferred_updates[ $key ] );
+		}
+	}
+}
--- a/packages/php/email-editor/src/Engine/PersonalizationTags/class-html-tag-processor.php
+++ b/packages/php/email-editor/src/Engine/PersonalizationTags/class-html-tag-processor.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * This file is part of the MailPoet plugin.
+ * This file is part of the MailPoet Email Editor package.
  *
  * @package MailPoet\EmailEditor
  */
@@ -15,10 +15,13 @@ use WP_HTML_Text_Replacement;
 /**
  * Class based on WP_HTML_Tag_Processor which is extended to replace
  * tokens with their values in the email content.
+ *
+ * This class was inspired by a concept from the WordPress core,
+ * which could help us to avoid refactoring in the future.
  */
 class HTML_Tag_Processor extends WP_HTML_Tag_Processor {
 	/**
-	 * List of deferred updates.
+	 * List of deferred updates which will be replaced after calling flush_updates().
 	 *
 	 * @var WP_HTML_Text_Replacement[]
 	 */
--- a/packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
+++ b/packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
@@ -1,0 +1,75 @@
+<?php
+/**
+ * This file is part of the MailPoet plugin.
+ *
+ * @package MailPoet\EmailEditor
+ */
+
+declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Engine\PersonalizationTags;
+
+/**
+ * Registry for personalization tags.
+ */
+class Personalization_Tags_Registry {
+
+	/**
+	 * List of registered personalization tags.
+	 *
+	 * @var array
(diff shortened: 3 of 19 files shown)


packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @return void
+	 */
+	public function initialize(): void {
+		apply_filters( 'mailpoet_email_editor_register_personalization_tags', $this );
+	}
+
+	/**
+	 * Register a new personalization tag.
+	 *
+	 * @param string   $name        Unique identifier for the callback.
```
pavel-mailpoet: "Unique identifier for the callback"

Should this be the identifier of the tag? I think this needs a better description because I am not sure what is the difference between `$name`, `$tag`, and why do we need category here and what it does.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @param string   $tag         The tag to be used in the email content.
+	 * @param string   $category    The category of the personalization tag.
+	 * @param callable $callback    The callable function/method.
```
pavel-mailpoet: "The callable function/method." - this is technically correct, but as a reader, I have no idea what this is used for and when the callback is called.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @param array    $attributes  Additional data or settings for the callback (optional).
+	 * @return void
+	 */
+	public function register( string $name, string $tag, string $category, callable $callback, array $attributes = array() ): void {
+		if ( isset( $this->tags[ $tag ] ) ) {
+			return;
```
pavel-mailpoet, line 42: Is the silent return here the best approach? I haven't seen it used yet, but bugs like this are quite hard to debug. Should this throw an exception if it is not expected to call with an already registered tag. Since this is part of MailPoet would it be useful to at least log something to the error log?

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
```diff
@@ -0,0 +1,75 @@
...
+			'tag'        => $tag,
+			'name'       => $name,
+			'category'   => $category,
+			'callback'   => $callback,
+			'attributes' => $attributes,
+		);
+	}
+
+	/**
+	 * Retrieve a personalization tag by its tag.
```
pavel-mailpoet: The naming here is quite weird, and the comment doesn't help. Would an example help to understand it better?

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
```diff
@@ -0,0 +1,75 @@
...
+	 *
+	 * @param string $tag The tag of the personalization tag.
+	 * @return array|null The array data or null if not found.
```
pavel-mailpoet: Should we describe the array better? When you are outside of this class, having only an array as the description is difficult to work with and tooling doesn't help. Should we use a class instead of array? Or at least document the array [better](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes)?

packages/php/email-editor/src/Engine/class-email-editor.php
```diff
@@ -61,30 +62,40 @@ class Email_Editor {
...
-	 * @param Patterns             $patterns Patterns.
-	 * @param Settings_Controller  $settings_controller Settings controller.
-	 * @param Send_Preview_Email   $send_preview_email Preview email controller.
+	 * @param Email_Api_Controller          $email_api_controller Email API controller.
+	 * @param Templates                     $templates Templates.
+	 * @param Template_Preview              $template_preview Template preview.
+	 * @param Patterns                      $patterns Patterns.
+	 * @param Settings_Controller           $settings_controller Settings controller.
+	 * @param Send_Preview_Email            $send_preview_email Preview email controller.
+	 * @param Personalization_Tags_Registry $personalization_tags_controller Preview email controller.
```
pavel-mailpoet: I don't think the comment here is correct.

packages/php/email-editor/src/Engine/PersonalizationTags/class-html-tag-processor.php
```diff
@@ -0,0 +1,51 @@
...
+use WP_HTML_Tag_Processor;
+use WP_HTML_Text_Replacement;
+
+/**
+ * Class based on WP_HTML_Tag_Processor which is extended to replace
+ * tokens with their values in the email content.
+ */
+class HTML_Tag_Processor extends WP_HTML_Tag_Processor {
+	/**
+	 * List of deferred updates.
```
pavel-mailpoet: This comment is technically true. However, it lacks clarity. Would it be better to describe what those deferred updates are, why we need them, or how they work?

packages/php/email-editor/tests/integration/Engine/PersonalizationTags/HTML_Tag_Processor_Test.php
```diff
@@ -0,0 +1,102 @@
...
+		$html_content = '<div>Hello!</div>';
+
+		// Instantiate the HTML_Tag_Processor with the HTML content.
+		$processor = new HTML_Tag_Processor( $html_content );
+		$processor->next_token();
+		// Replace the token.
+		$processor->replace_token( 'John!' );
+
+		// Verify deferred updates.
+		$deferred_updates = $this->getPrivateProperty( $processor, 'deferred_updates' );
```
pavel-mailpoet: I am not sure if having this test testing the internals of the class is the best approach. Do you think it would be better to test the outside behaviour of the class? How does the class behave from the outside if it does what we need from it?

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
```diff
@@ -0,0 +1,75 @@
+<?php
+/**
+ * This file is part of the MailPoet plugin.
```
pavel-mailpoet: I am not sure I understand why is this class (and other classes from this Pull Request) part of the MailPoet plugin and not part of the Email Editor. I don't see anything MailPoet specific here.  We will need personalisation tags even in other integrations, won't we?

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalizer.php
```diff
@@ -0,0 +1,102 @@
...
+	 *
+	 * @param Personalization_Tags_Registry $tags_registry Personalization tags registry.
+	 */
+	public function __construct( Personalization_Tags_Registry $tags_registry ) {
+		$this->tags_registry = $tags_registry;
+		$this->context       = array();
+	}
+
+	/**
+	 * Set the context for the personalization.
```
pavel-mailpoet: Could you provide more details here on what the "context" is?

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalizer.php
```diff
@@ -0,0 +1,102 @@
+<?php
+/**
+ * This file is part of the MailPoet plugin.
+ *
+ * @package MailPoet\EmailEditor
+ */
+
+declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Engine\PersonalizationTags;
```
pavel-mailpoet: So this is the entry point to all this functionality, isn't it? Would it be better to have this class in this folder, and the "helper" classes like `HTML_Tag_Processor` and the Registry in a subfolder?

mailpoet/lib/EmailEditor/Integrations/MailPoet/PersonalizationTagManager.php
```diff
@@ -0,0 +1,42 @@
...
+
+  public function __construct(
+    Subscriber $subscriber
+  ) {
+    $this->subscriber = $subscriber;
+  }
+
+  public function initialize() {
+    add_filter('mailpoet_email_editor_register_personalization_tags', function( Personalization_Tags_Registry $registry ): Personalization_Tags_Registry {
+      $registry->register(
```
pavel-mailpoet: I am looking at the signature/usage of this method and I am wondering if it would be better to use a class as the argument here and pass it in. Instead of having a method with five arguments we would have a single argument. What do you think about it?

pavel-mailpoet (changes requested):
Thank you for the pull request. Overall, this looks great. I have some suggestions, and I am open to discussing them. I went commit by commit, so I saw some classes before their usage, which may have influenced some of my comments.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
lysyjan: You are right. The name should be displayed to users in the modal windows with tags overview or other UI parts. I will update the comment here.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
lysyjan: 👍 I will update the comment. This is a callback that returns a value of the tag.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalizer.php
lysyjan: Good idea. I had it in my mind, but for some reason, I didn't realize it.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
lysyjan: I considered using error_log, but it needs adding CodeSniffer ignore, which I wanted to avoid initially.
I decided to use the silent return in the first phase because if you call registration twice it could generate errors, but maybe it's correct 🤔

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
lysyjan: I was thinking about a name `token` instead of the tag. It sounds to me more descriptive. What do you think about it? I will try to improve the comments across all new files.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
pavel-mailpoet: I can see many approaches:

- silent return
- silently overwrite the original value
- throw an exception
- silent return with an `error_log`
- silent return with a MailPoet log

I am not saying one is the correct one, but if we want to go with one, it should probably be a conscious decision, and we should have a reason why.

packages/php/email-editor/tests/integration/Engine/PersonalizationTags/HTML_Tag_Processor_Test.php
lysyjan: That sounds good. I will update the test.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
pavel-mailpoet: Sounds good, but maybe also an example could help.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
lysyjan: You are right. I will update those comments. Or do you have another idea of what can be used here in the class header?

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalizer.php
lysyjan: Yes. I will add it.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
pavel-mailpoet: Well, the comment is fine with me, but I don't understand why the files are in the MailPoet plugin and not in the core editor. They seem generic, and we will need them (or something similar) when we port the editor to Woo Core or Fruita, won't we?

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
lysyjan: 👍 I will update the annotation.

mailpoet/lib/EmailEditor/Integrations/MailPoet/PersonalizationTagManager.php
lysyjan: I am not sure how you mean it. Can you please write a part of the code here for better understanding?

pavel-mailpoet: So the method `register` takes arguments "string $name, string $tag, string $category, callable $callback, array $attributes = array()". 
I am suggesting to create a class:

```
class PersonalizationTag {
	private string $name; 
	private string $tag; 
	private string $category; 
	private callable $callback; 
	private array $attributes = array();
  ...
}
```

and then use this class not just as an argument of the method `public function register( PersonalizationTag $tag )` but also inside the registry class instead of the array.

lysyjan:
@pavel-mailpoet Thanks for the great review. I hope I addressed all your suggestions.
I created a new Jira ticket related to the logs.

packages/php/email-editor/src/Engine/PersonalizationTags/class-personalization-tags-registry.php
```diff
@@ -23,6 +23,7 @@ class Personalization_Tags_Registry {
 
 	/**
 	 * Initialize the personalization tags registry.
+	 * This method should be called only once.
```
pavel-mailpoet, line 26: Not for this Pull request, but should we somehow ensure that the function is called only once? After it is called, set some property and then prevent calling it again.

pavel-mailpoet (approved):

PR #5957, opened 2024-11-26
Improve scheduled subscribers count check for PN and automation emails [MAILPOET-6346]

costasovo:
## Description

When the scheduler worker processes a post notification email or automation email, it enqueues subscribers by adding them to the scheduled_task_subscribers table. After it adds subscribers, there is a logic checking if there are any subscribers in the queue or not. The problem was that in case the subscribers were already enqueued (could be from a previous run if the multi insert timeouted), the method for adding subscribers would return 0 because all rows were already in the DB, and the logic for checking subscribers count would cancel or reschedule sending.

This PR fixes this by checking for the count of enqueued subscribers in the DB rather then looking at the number of newly inserted. 

## Code review notes

Scheduled task subscribers relation [is set EXTRA_LAZY](https://github.com/mailpoet/mailpoet/blob/7bacd887a9d0bc60280afa54173068d41afacf84/mailpoet/lib/Entities/ScheduledTaskEntity.php#L93-L97). 
That means we can call `$task->getSubscribers()->count();` and it will run an SQL query to get count instead of fetching them all. See [docs](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/tutorials/extra-lazy-associations.html).

## QA notes

Please test sending immediate and scheduled post notifications.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6346]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6346]: https://mailpoet.atlassian.net/browse/MAILPOET-6346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:fix-scheduled-count-check)

_The latest successful build from `fix-scheduled-count-check` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/API/JSON/v1/SendingQueue.php
+++ b/mailpoet/lib/API/JSON/v1/SendingQueue.php
@@ -166,7 +166,8 @@ class SendingQueue extends APIEndpoint {
         $segments = $newsletter->getSegmentIds();
 
         $this->scheduledTasksRepository->refresh($scheduledTask);
-        $subscribersCount = $this->subscribersFinder->addSubscribersToTaskFromSegments($scheduledTask, $segments, $newsletter->getFilterSegmentId());
+        $this->subscribersFinder->addSubscribersToTaskFromSegments($scheduledTask, $segments, $newsletter->getFilterSegmentId());
+        $subscribersCount = $scheduledTask->getSubscribers()->count();
 
         if (!$subscribersCount) {
           return $this->errorResponse([
--- a/mailpoet/lib/Cron/Workers/Scheduler.php
+++ b/mailpoet/lib/Cron/Workers/Scheduler.php
@@ -193,7 +193,8 @@ class Scheduler {
     }
 
     // ensure that subscribers are in segments
-    $subscribersCount = $this->subscribersFinder->addSubscribersToTaskFromSegments($task, $segments, $newsletter->getFilterSegmentId());
+    $this->subscribersFinder->addSubscribersToTaskFromSegments($task, $segments, $newsletter->getFilterSegmentId());
+    $subscribersCount = $task->getSubscribers()->count();
     if (empty($subscribersCount)) {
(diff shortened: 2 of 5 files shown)


websupporter:
Overall looks good to me and makes sense.

The other usage of `addSubscribersToTaskFromSegments()` where we rely on the returned count is in the API:
https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/API/JSON/v1/SendingQueue.php#L169-L171

From my understanding: This method gets called when the Send-Button is clicked and it is a normal newsletter. My first thought is: That is probably safe to rely on the count in that instance.

But I wonder, if we maybe should change the signature of `addSubscribersToTaskFromSegments()` so that it does not return the count anymore and maybe just a boolean. Otherwise I could see us reintroducing this problem. WDYT?

costasovo:
> This method gets called when the Send-Button is clicked and it is a normal newsletter. My first thought is: That is probably safe to rely on the count in that instance.

Yes. It is in a branch when the newsletter is sent immediately and we always create a new task for that so the issue can't happen.

> But I wonder, if we maybe should change the signature of addSubscribersToTaskFromSegments() so that it does not return the count anymore and maybe just a boolean. Otherwise I could see us reintroducing this problem. WDYT?

@websupporter Good idea! It seems handy that the method returns the count, but because of the error, void could be better.

costasovo:
> But I wonder, if we maybe should change the signature of addSubscribersToTaskFromSegments() so that it does not return the count anymore and maybe just a boolean. Otherwise I could see us reintroducing this problem. WDYT?

@websupporter I implemented the idea in https://github.com/mailpoet/mailpoet/pull/5957/commits/3d3cb4d47a8fee7176edab340ba022db2292759b

websupporter (approved):
lgtm! Thank you @costasovo 👏

Aschepikov (approved):
GJ!

PR #5964, opened 2024-11-29
Prevent incorrect newsletter sending status [MAILPOET-6241]

costasovo:
## Description

This PR fixes an issue: the Resend button on the Sending Status page switched the newsletter status to `sending` regardless of type.

Screenshot 2024-11-27 at 17 48 25

It also fixes the list refresh after the Resend button is clicked. The functionality has been broken for some time.

## Code review notes

I found two more potentially problematic places in https://github.com/mailpoet/mailpoet/commit/76baf9f4817ed8108f02c49ba96be9d1757d746e. I was not able to replicate the incorrect state for these two by any user action, but I still updated them to be safe.

### Replication

It is better to start with a clean plugin.

1) Create and automation that sends emails
2) Make it send a couple of emails
3) Go to DB admin and in the `wp_mailpoet_scheduled_task_subscribers` table set a couple of records as failed by adding 1 to `fail` column and a random error message to the `error` columns. 
4) Go to the Sending status page (Automation > Analytics > Click sent number in email step – see on the screenshot)

Screenshot 2024-11-29 at 9 33 17

5) Click the resend button
6) Observe 2 issues: I) The list doesn't refresh, II) Go to `wp_mailpoet_newsletters` and observe that the automation email has status `sending` (correct is `active`).

After installing and the plugin with fix. The incorrect state should be fixed via a migration. Note: to run migration you may need to deactivate and reactivate the plugin. Also the resend button shouldn't set the incorrect state any more. 

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6241]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6241]: https://mailpoet.atlassian.net/browse/MAILPOET-6241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:incorrect-newsletter-state)

_The latest successful build from `incorrect-newsletter-state` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/assets/js/src/listing/listing.jsx
+++ b/mailpoet/assets/js/src/listing/listing.jsx
@@ -23,8 +23,11 @@ class ListingComponent extends Component {
   }
 
   componentDidMount() {
-    const { params: propsParams, auto_refresh: autoRefresh = false } =
-      this.props;
+    const {
+      params: propsParams,
+      auto_refresh: autoRefresh = false,
+      refreshRef = null,
+    } = this.props;
     this.isComponentMounted = true;
     const params = propsParams || {};
     this.initWithParams(params);
@@ -33,6 +36,10 @@ class ListingComponent extends Component {
       jQuery(document).on('heartbeat-tick.mailpoet', () => {
         this.getItems();
       });
+    }
+
+    if (refreshRef) {
+      refreshRef.current = () => this.getItems();
     }
   }
 
@@ -751,6 +758,7 @@ ListingComponent.propTypes = {
   isItemDeletable: PropTypes.func,
   isItemToggleable: PropTypes.func,
   className: PropTypes.string,
+  refreshRef: PropTypes.shape({ current: PropTypes.func }),
 };
 
 ListingComponent.displayName = 'Listing';
--- a/mailpoet/assets/js/src/newsletters/sending-status.jsx
+++ b/mailpoet/assets/js/src/newsletters/sending-status.jsx
@@ -1,6 +1,6 @@
 import classnames from 'classnames';
 import { Link, useLocation, useParams } from 'react-router-dom';
-import { memo, useEffect, useState } from 'react';
+import { memo, useEffect, useState, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { __, _x } from '@wordpress/i18n';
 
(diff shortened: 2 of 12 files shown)


alex-mpoet (commented):
@costasovo I have one concern: are we really sure everyone whose past sending tasks are stuck because of this bug still wants the emails to be sent when they update the plugin? Maybe it's better to keep sending tasks paused in the migration so they could be resumed manually, or is this problematic, e.g. does not unclog the queue?

costasovo:
> I have one concern: are we really sure everyone whose past sending tasks are stuck because of this bug still wants the emails to be sent when they update the plugin? Maybe it's better to keep sending tasks paused in the migration so they could be resumed manually, or is this problematic, e.g. does not unclog the queue?

@alex-mpoet It would be ideal, but unfortunately, this is relevant mainly to automation and legacy automation emails (welcome and woo automatic), and for those emails, the paused state is not expected, and there is no UI for unpausing those. 
The options we have are:
- keep paused - they don't block the but are displayed as in progress forever
- unpause and let queue to process them
- mark as completed
- delete them as they never happened
- mark as failed - we display the resend button in the Sending Status screen for the failed tasks

Hmm perhaps we could also decide by the creation time. Something like unpause the task if it is 30 days old or newer and mark as failed (with an error message like "Sending timeout."). The failed ones would have the Resend button in the Sending Status screen. It would not be super convenient for UX because it doesn't support bulk action, but perhaps a safer approach. What do you think?

alex-mpoet:
@costasovo It's a great idea to have older tasks marked as failed! Regarding the time window for such tasks, I feel that 30 days may be too much. My gut feeling is that 7-14 days might be enough. But feel free to pick the value you prefer.

costasovo:
@alex-mpoet I updated the PR. In the end, I stayed with the 30 days - no special reason, just a different gut feeling :D

alex-mpoet:
Thank you for making the changes! Approving.

alex-mpoet (approved):

Aschepikov (approved):
GJ!

PR #5967, opened 2024-11-29
Retry email sending automation for unconfirmed subscribers [MAILPOET-6176]

oz:
## Description

When a subscriber isn’t already “Subscribed”, and some automation runs without any delay (to allow the subscriber to confirm their subscription), then any Send-Email actions will fail.

This PR fixes the issue by retrying the "Send Email Action" a couple of times if opt-in is required, and the subscriber has not yet confirmed their subscription.

## Code review notes

I discussed a bit of the approach with @JanJakes, and we found it was probably okay to re-schedule the "SendEmailAction" action a few times to wait for opt-in, and store state in the automation's run log. 💭 

> [!IMPORTANT]
> I'd like to get feedback on the current state of this PR, before spending a bit more time to improve the retry mechanism (see comments in code), or add more tests maybe. 😄 

## QA notes

Testing instructions:
1. Make sure signup confirmation is enabled in Settings.
2. Create a new automation with two steps, "Tag Added to Subscriber" as a trigger and "Send email" as an action. Activate it.
3. Add a new unconfirmed subscriber.
4. Add a tag to this subscriber on the subscriber edit page.
5. Wait until cron execution and observe that the automation run is in progress. Wait for some time to see that it is not advancing further.
6. Confirm the subscriber.
7. Check that the automation run is now completed.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6176](https://mailpoet.atlassian.net/browse/MAILPOET-6176)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6176]: https://mailpoet.atlassian.net/browse/MAILPOET-6176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:MAILPOET-6176_automations-subscribed-only)

_The latest successful build from `MAILPOET-6176_automations-subscribed-only` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
+++ b/mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
@@ -6,12 +6,14 @@ use MailPoet\AutomaticEmails\WooCommerce\Events\AbandonedCart;
 use MailPoet\Automation\Engine\Control\AutomationController;
 use MailPoet\Automation\Engine\Control\StepRunController;
 use MailPoet\Automation\Engine\Data\Automation;
+use MailPoet\Automation\Engine\Data\AutomationRunLog;
 use MailPoet\Automation\Engine\Data\Step;
 use MailPoet\Automation\Engine\Data\StepRunArgs;
 use MailPoet\Automation\Engine\Data\StepValidationArgs;
 use MailPoet\Automation\Engine\Exceptions\NotFoundException;
 use MailPoet\Automation\Engine\Integration\Action;
 use MailPoet\Automation\Engine\Integration\ValidationException;
+use MailPoet\Automation\Engine\Storage\AutomationRunLogStorage;
 use MailPoet\Automation\Integrations\MailPoet\Payloads\SegmentPayload;
 use MailPoet\Automation\Integrations\MailPoet\Payloads\SubscriberPayload;
 use MailPoet\Automation\Integrations\WooCommerce\Payloads\AbandonedCartPayload;
@@ -47,6 +49,19 @@ class SendEmailAction implements Action {
     4 * DAY_IN_SECONDS, // ~5 days
     25 * DAY_IN_SECONDS, // ~1 month
   ];
+
+  // Retry intervals for sending. These are used when the email address
+  // is not confirmed, and we need send non-transactional emails.
+  private const OPTIN_RETRY_INTERVALS = [
+    1 * MINUTE_IN_SECONDS, // ~1 minute
+    5 * MINUTE_IN_SECONDS, // ~5 minutes
+    20 * MINUTE_IN_SECONDS, // ~20 minutes
+    1 * HOUR_IN_SECONDS, // ~1 hour
+    12 * HOUR_IN_SECONDS, // ~12 hours
+    1 * DAY_IN_SECONDS, // ~1 day
+  ];
+  private const WAIT_OPTIN = 'wait_optin';
+  private const OPTIN_RETRIES = 'optin_retries';
 
   private const TRANSACTIONAL_TRIGGERS = [
     'woocommerce:order-status-changed',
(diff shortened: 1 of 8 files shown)


mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -171,40 +186,115 @@ public function run(StepRunArgs $args, StepRunController $controller): void {
...
+      }
+      $nextInterval = self::OPTIN_RETRY_INTERVALS[$runNumber - 1];
+    }
+
     $controller->scheduleProgress(time() + $nextInterval);
   }
 
+  private function isOptInRequired(NewsletterEntity $newsletter, SubscriberEntity $subscriber): bool {
+    $subscriberStatus = $subscriber->getStatus();
+    return ($newsletter->getType() !== NewsletterEntity::TYPE_AUTOMATION_TRANSACTIONAL && $subscriberStatus !== SubscriberEntity::STATUS_SUBSCRIBED);
```
websupporter: Maybe just taste, but I find this hard to read. So much combined negativity.

i think that would make it easier for me:
```
if transactional return false;
return !subscribed
```

oz: No problem with less negativity. 😄

FTR I merely extracted existing code here. 😇

websupporter:
Overall, I think that looks good. We talked yesterday about the retry count approach and it makes sense to me to store some kind of "retries for optin"-number.

mailpoet/lib/Automation/Integrations/MailPoet/Actions/SendEmailAction.php
```diff
@@ -171,40 +186,115 @@ public function run(StepRunArgs $args, StepRunController $controller): void {
...
+        $this->rerunLater($args, $controller, $newsletter, $subscriber);
+        return;
       }
 
+      $this->scheduleEmail($args, $newsletter, $subscriber);
     } else {
-      // run #N: check/sync sending status with the automation step
+      // Re-running for opt-in?
+      $state = $this->getRunLogData($args);
+      if (array_key_exists(self::WAIT_OPTIN, $state) && $state[self::WAIT_OPTIN] === 1) {
```
alex-mpoet, line 205: Not a strict requirement, but if `WAIT_OPTIN` is treated as a boolean, this could be simplified as `if (!empty($state[self::WAIT_OPTIN])) {`.

alex-mpoet (commented):
@oz I have an issue when trying to test this. After the `{"wait_optin":1}` flag gets saved to the DB, it is immediately wiped by `StepRunLogger->logProgress()` and other `StepRunLogger` methods, so it is absent during the next automation run. As a result, the code [here](https://github.com/mailpoet/mailpoet/pull/5967/files#diff-1e596733ccdea88d2ded67e24bca2d74c50867d0fe30c9137a81063fa64d7e3cR206) thinks the subscriber is confirmed and email scheduling fails like it did before this fix. Does it work for you?

oz:
> @oz I have an issue when trying to test this. After the `{"wait_optin":1}` flag gets saved to the DB, it is immediately wiped by `StepRunLogger->logProgress()` and other `StepRunLogger` methods, so it is absent during the next automation run. As a result, the code [here](https://github.com/mailpoet/mailpoet/pull/5967/files#diff-1e596733ccdea88d2ded67e24bca2d74c50867d0fe30c9137a81063fa64d7e3cR206) thinks the subscriber is confirmed and email scheduling fails like it did before this fix. Does it work for you?

Hi @alex-mpoet! Can you detail the steps to reproduce this?

alex-mpoet:
@oz

> Hi @alex-mpoet! Can you detail the steps to reproduce this?

I just tested the basic scenario described in the PR:
1. Make sure signup confirmation is enabled in Settings.
2. Create a new automation with two steps, "Tag Added to Subscriber" as a trigger and "Send email" as an action. Activate it.
3. Add a new unconfirmed subscriber.
4. Add a tag to this subscriber on the subscriber edit page.
5. Wait until cron execution and observe the failed automation run with the "'Email failed to schedule" error.

oz:
> @oz
> 
> > Hi @alex-mpoet! Can you detail the steps to reproduce this?
> 
> I just tested the basic scenario described in the PR:
> 
> 1. Make sure signup confirmation is enabled in Settings.
> 2. Create a new automation with two steps, "Tag Added to Subscriber" as a trigger and "Send email" as an action. Activate it.
> 3. Add a new unconfirmed subscriber.
> 4. Add a tag to this subscriber on the subscriber edit page.
> 5. Wait until cron execution and observe the failed automation run with the "'Email failed to schedule" error.

Ah thanks @alex-mpoet! 🙏🏻 

I found the issue. Indeed, it looks like the step-runner has its own instance of a logger, and will overwrite the one created in the action. I have a fix that involves passing the a step-run logger through (e.g. using a step-run controller), but I'm not very satisfied with it yet. I'll resubmit for review later.

oz:
@alex-mpoet can you take another look at this? To work-around the potential overwrites, I found that I had to share the instance of the "logger" from the class that handles step-runs. I'm not sure that this is the best approach overall, but it means that the data is no longer overwritten, and the action can log its own data too.

alex-mpoet (approved):
Thanks for fixing, the retrying mechanism works now and looks good!

Aschepikov (approved):
GJ!

PR #5972, opened 2024-12-02
Add start from scratch to the template selector - MAILPOET-6333

triple0t:
## Description

Add start from scratch to the template selector.

The button position will change depending on whether a template exists or not.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6333](https://mailpoet.atlassian.net/browse/MAILPOET-6333)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6333]: https://mailpoet.atlassian.net/browse/MAILPOET-6333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:add-start-from-scratch-to-the-template-selector)

_The latest successful build from `add-start-from-scratch-to-the-template-selector` will be used. If none is available, the link won't work._

### DIFF
--- a/packages/js/email-editor/src/components/template-select/select-modal.tsx
+++ b/packages/js/email-editor/src/components/template-select/select-modal.tsx
@@ -6,7 +6,11 @@ import {
 	Modal,
 	__experimentalHStack as HStack, // eslint-disable-line
 	Button,
+	Flex,
+	FlexItem,
 } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+
 import { Async } from './async';
 import { usePreviewTemplates } from '../../hooks';
 import { storeName, TemplatePreview } from '../../store';
@@ -113,6 +117,17 @@ export function SelectTemplateModal( { onSelectCallback } ) {
 							</div>
 						) ) }
 					</div>
+
+					<Flex justify="flex-end">
+						<FlexItem>
+							<Button
+								variant="tertiary"
+								onClick={ () => handleCloseWithoutSelection() }
+							>
+								{ __( 'Start from scratch', 'mailpoet' ) }
+							</Button>
+						</FlexItem>
+					</Flex>
 				</div>
 			</div>
 		</Modal>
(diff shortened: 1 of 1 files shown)


costasovo (commented):
@triple0t When testing I noticed that when there are no templates the button is present as primary and centered. But when I click it whem templates are loading it doesn't do anything.
What about hiding it or making it disabled during that period?

triple0t:

> What about hiding it or making it disabled during that period?

Quick question on this: How do you think we may know this period? We return an [empty array](https://github.com/mailpoet/mailpoet/blob/f9458c2b7f5b837bf48addf76c52ae123cb5318f/packages/js/email-editor/src/hooks/use-preview-templates.ts#L55-L57) when it's loading and when the template is not found.

costasovo:
> Quick question on this: How do you think we may know this period? We return an [empty array](https://github.com/mailpoet/mailpoet/blob/f9458c2b7f5b837bf48addf76c52ae123cb5318f/packages/js/email-editor/src/hooks/use-preview-templates.ts#L55-L57) when it's loading and when the template is not found.

I see. The information about the pending requests or actual requests because we need patterns and templates is not accessible as it is happening in a core store. But in the current state, the editor needs at least a blank template to be able to start, so for now, I think we could just wait for templates to be a non-empty array. If the template fetching fails, we can't start the editor. Would that work for you?

triple0t:
> But in the current state, the editor needs at least a blank template to be able to start, so for now, I think we could just wait for templates to be a non-empty array. If the template fetching fails, we can't start the editor. Would that work for you?

Yeah, that's fine with me, @costasovo.
Implemented in https://github.com/mailpoet/mailpoet/pull/5972/commits/2eaec4141d2df280867757320f28dfe4c30668da

Thanks.

costasovo (approved):
Thank you!

PR #5978, opened 2024-12-04
Email editor – polish UI for template edit, swap and restore [MAILPOET-6336]

costasovo:
## Description

This PR removes the temporary template experiment panel and adds the controls to the proper places.

https://github.com/user-attachments/assets/de5e71fa-c580-470c-b164-6a6136520e28

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6336]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6336]: https://mailpoet.atlassian.net/browse/MAILPOET-6336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/tempaltes-sidebar)

_The latest successful build from `email-editor/tempaltes-sidebar` will be used. If none is available, the link won't work._


packages/js/email-editor/src/components/sidebar/email-type-info.tsx
```diff
@@ -1,26 +1,125 @@
...
+								) }
+							</span>
+						</div>
+					</PanelRow>
+					{ template && (
+						<PanelRow>
+							<Flex justify={ 'start' }>
+								<FlexItem className="editor-post-panel__row-label">
+									Template
+								</FlexItem>
```
triple0t: Why not use the i18n function `__` here?

packages/js/email-editor/src/components/header/index.scss
```diff
@@ -39,3 +39,8 @@
     min-width: 280px;
   }
 }
+
+// We currently don't support command palette so we hide the shortcut
+.editor-document-bar__shortcut {
+  display: none;
+}
```
triple0t: I can still see the command palette.

Screenshot 2024-12-06 at 7 47 26 AM

triple0t (commented):
Well done. 
This PR looks good, and I'm excited about it.

I have minor remarks; please check them below.

Screenshot 2024-12-06 at 7 56 32 AM

Can we update this toast content? It's currently showing `site updated` instead of `template updated`.

Screenshot 2024-12-06 at 9 53 58 AM

Would it make sense to have a cancel button here?
It wasn't totally obvious that clicking outside the modal would exit it.


Screenshot 2024-12-06 at 9 58 56 AM

This button is very well hidden. Is there a reason why we placed it in such a non-descript place?


Screenshot 2024-12-06 at 9 59 05 AM

I feel this warning isn't too clear. The wording suggests it would clear my current customization when, in actual fact, it removes "ANY and ALL" customization done on the template.

Can we update the wording to something like:
`WARNING: This will clear ANY and ALL template customization. All updates made to the template will be lost. Do you want to proceed?`

costasovo:
@triple0t Thank you for the thorough review! Your remarks are great and will help to improve our app.

> Can we update this toast content? It's currently showing site updated instead of template updated.

Done in https://github.com/mailpoet/mailpoet/pull/5978/commits/9879ee4bc251c9eb7a6d4ca89e07c566a35664bc Note that this `Site updated` notice is in core used also for updating global styles. So for now, as styles are part of template data, we are fine, but in the future task where we introduce global styles, we will have to change it to something more suitable. Perhaps `Email design updated` could work. But that's just for the future.

> Would it make sense to have a cancel button here?
It wasn't totally obvious that clicking outside the modal would exit it.

The specification in the ticket explicitly mentions only a single button (point 3.f.I.2), but I agree that Cancel could be helpful. I just want to add that the native post editor doesn't show any dialog at all. I understood that the purpose of this dialog is to educate users about template mode a bit. I added the button in https://github.com/mailpoet/mailpoet/pull/5978/commits/80a825b8af6db1cb75412dd89d01bb00d7958670

> This button is very well hidden. Is there a reason why we placed it in such a non-descript place?

This is exactly the same place as in the site editor (see screenshot). I agree it is hidden, but at the same time, I think this action should be used rarely so I would keep it as it is for now.
Screenshot 2024-12-06 at 10 41 16

> I feel this warning isn't too clear. The wording suggests it would clear my current customization when, in actual fact, it removes "ANY and ALL" customization done on the template.

I think it is a good idea to make the message more informative. I used your message as was. I would like to ask @pavel-mailpoet here to give feedback on the change of the reset template warning to `This will clear ANY and ALL template customization. All updates made to the template will be lost. Do you want to proceed?` Is that ok?

packages/js/email-editor/src/components/header/index.scss
costasovo: Thank you! I fixed it by replacing the original commit. https://github.com/mailpoet/mailpoet/pull/5978/commits/223d8f6dc358baf88f884caa90ac7e6808755cc0

packages/js/email-editor/src/components/sidebar/email-type-info.tsx
costasovo: Nice catch! I added the translation function https://github.com/mailpoet/mailpoet/pull/5978/commits/739cb856322965693b1a7fb3709105a24a68efca

pavel-mailpoet:
> I think it is a good idea to make the message more informative. I used your message as was. I would like to ask @pavel-mailpoet here to give feedback on the change of the reset template warning to This will clear ANY and ALL template customization. All updates made to the template will be lost. Do you want to proceed? Is that ok?

Sounds good, but do you know what the core editor says?

costasovo:
> Sounds good, but do you know what the core editor says?

@pavel-mailpoet The core editor says: "Reset to default and clear all customizations?" I think it does no harm to be more informative in this case. But when we switch to the WP's native editor, we will get the original message. I have no strong opinion here.

costasovo:
@triple0t I believe I addressed/answered all your remarks and the PR should be ready for the second review.

triple0t (approved):
LGTM. 
Thanks for making the update.

I see this PR is linked to another PR. I'll assign it back to you. Please feel free to merge when ready.

PR #5975, opened 2024-12-03
Add categories to the select a template dialogue - MAILPOET-6331

triple0t:
## Description

This PR adds basic support for template categories within the "Select template modal"

## Code review notes

Note: we are currently hardcoding the template category. This process will be refined and updated in the [Templates: Recent templates behavior-MAILPOET-5949](https://mailpoet.atlassian.net/browse/MAILPOET-5949)

Note: The header style was updated in the child PR https://github.com/mailpoet/mailpoet/pull/5980

## QA notes

_N/A_

## Linked PRs

Related to https://github.com/mailpoet/mailpoet/pull/5980

## Linked tickets

[MAILPOET-6331](https://mailpoet.atlassian.net/browse/MAILPOET-6331)

## After-merge notes

_N/A_

## Tasks

- [ ] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [ ] I added sufficient test coverage
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:add-categories-to-the-select-a-template-dialogue)

_The latest successful build from `add-categories-to-the-select-a-template-dialogue` will be used. If none is available, the link won't work._

[MAILPOET-6331]: https://mailpoet.atlassian.net/browse/MAILPOET-6331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5980, opened 2024-12-05
Templates-recent templates behavior - MAILPOET-5949

triple0t:
## Description

Continuation of https://github.com/mailpoet/mailpoet/pull/5975

In this PR, we update the recent template behavior by pulling emails sent from the new email editor and using its post content as an email preset (template) for new emails.

Screenshot 2024-12-05 at 11 09 46 PM
Screenshot 2024-12-05 at 11 09 56 PM

## Code review notes

Please start review from https://github.com/mailpoet/mailpoet/pull/5980/commits/2bc05534c72a96ac6952d71bcd445eb9a4ea4e0b

## QA notes

For testing: You can add `draft` to this list https://github.com/mailpoet/mailpoet/blob/95a8b9bf0be8ea2bcabf5c810238609e1d765b6d/packages/js/email-editor/src/store/selectors.ts#L147
e.g.
```
status: 'publish,sent,draft',
```

## Linked PRs

Related to https://github.com/mailpoet/mailpoet/pull/5975

## Linked tickets

[MAILPOET-5949](https://mailpoet.atlassian.net/browse/MAILPOET-5949)

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:templates-recent-templates-behaviour)

_The latest successful build from `templates-recent-templates-behaviour` will be used. If none is available, the link won't work._

[MAILPOET-5949]: https://mailpoet.atlassian.net/browse/MAILPOET-5949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/js/email-editor/src/components/template-select/select-modal.tsx
+++ b/packages/js/email-editor/src/components/template-select/select-modal.tsx
@@ -47,8 +47,7 @@ export function SelectTemplateModal( {
 	const hasTemplates = templates?.length > 0;
 
 	const handleTemplateSelection = ( template: TemplatePreview ) => {
-		const templateIsPostContent =
-			template.template.type === 'mailpoet_email';
+		const templateIsPostContent = template.type === 'mailpoet_email';
 
 		const postContent = template.template as unknown as EmailEditorPostType;
 
--- a/packages/js/email-editor/src/components/template-select/select-modal.tsx
+++ b/packages/js/email-editor/src/components/template-select/select-modal.tsx
@@ -42,7 +42,7 @@ export function SelectTemplateModal( {
 	closeCallback = null,
 	previewContent = '',
 } ) {
-	const [ templates ] = usePreviewTemplates( previewContent );
+	let [ templates, emailPosts ] = usePreviewTemplates( previewContent );
 
 	const hasTemplates = templates?.length > 0;
 
@@ -82,6 +82,12 @@ export function SelectTemplateModal( {
 		},
 	];
 
+	let initialCategory = dummyTemplateCategories[ 0 ]; // Show the “Recent” category by default
+	if ( ! emailPosts || emailPosts?.length === 0 ) {
+		emailPosts = [];
+		initialCategory = dummyTemplateCategories[ 1 ]; // user does not recent category, show basic category
+	}
+
 	return (
 		<Modal
 			title={ __( 'Select a template', 'mailpoet' ) }
@@ -91,9 +97,9 @@ export function SelectTemplateModal( {
 			isFullScreen
 		>
 			<SelectTemplateBody
-				initialCategory={ dummyTemplateCategories[ 0 ] }
+				initialCategory={ initialCategory }
 				templateCategories={ dummyTemplateCategories }
-				templates={ templates }
+				templates={ [ ...templates, ...emailPosts ] }
 				handleTemplateSelection={ handleTemplateSelection }
 			/>
 
--- a/packages/js/email-editor/src/components/template-select/select-modal.tsx
+++ b/packages/js/email-editor/src/components/template-select/select-modal.tsx
@@ -32,13 +32,15 @@ function SelectTemplateBody( {
 	handleTemplateSelection,
 } ) {
 	const [ selectedCategory, setSelectedCategory ] = useState(
-		TemplateCategories[ 0 ].name // Show the “Recent” category by default
+		TemplateCategories[ 1 ].name // Show the “Basic” category by default
 	);
 
 	useEffect( () => {
-		if ( ! hasEmailPosts ) {
-			setSelectedCategory( TemplateCategories[ 1 ].name );
-		}
+		setTimeout( () => {
+			if ( hasEmailPosts ) {
+				setSelectedCategory( TemplateCategories[ 0 ].name );
+			}
+		}, 1000 ); // using setTimeout to ensure the template styles are available before block preview
 	}, [ hasEmailPosts ] );
 
 	return (
--- a/packages/js/email-editor/src/components/template-select/select-modal.tsx
+++ b/packages/js/email-editor/src/components/template-select/select-modal.tsx
@@ -59,9 +59,9 @@ export function SelectTemplateModal( {
 		}
 
 		void dispatch( storeName ).setTemplateToPost(
-			templateIsPostContent ? postContent?.template : template.slug,
+			templateIsPostContent ? postContent.template : template.slug,
 			templateIsPostContent
-				? {}
+				? template?.template?.mailpoet_email_theme || {}
 				: template.template.mailpoet_email_theme ?? {}
 		);
 		onSelectCallback();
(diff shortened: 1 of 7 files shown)


packages/js/email-editor/src/components/template-select/select-modal.tsx
```diff
@@ -1,34 +1,83 @@
...
+	storeName,
+	TemplateCategory,
+	TemplatePreview,
+} from '../../store';
+import { TemplateList } from './template-list';
+import { TemplateCategoriesListSidebar } from './template-categories-list-sidebar';
 
 const BLANK_TEMPLATE = 'email-general';
 
+const TemplateCategories: Array< { name: TemplateCategory; label: string } > = [
```
triple0t, line 18: We need to figure out what to do about this Template Category list. 
[WP_Block_Template](https://developer.wordpress.org/reference/classes/wp_block_template/) has an area property we can use to store the template category, but the field and value are not returned from the REST API. 

We would need to add another work-around [here](https://github.com/mailpoet/mailpoet/blob/b1cc53ec5bd0647e347a4ff281791b7e8148bad7/packages/php/email-editor/src/Engine/Templates/class-templates.php#L261-L287) by extending the `wp-template` endpoint. I'm not sure if there's value in doing that now or we should tackle it in this [ticket](https://mailpoet.atlassian.net/browse/MAILPOET-6356).

@costasovo, what do you think?

packages/js/email-editor/src/store/selectors.ts
```diff
@@ -140,6 +139,19 @@ export const getEditedEmailContent = createRegistrySelector(
 	}
 );
 
+export const getSentEmailEditorPosts = createRegistrySelector(
+	( select ) => () => {
+		const posts: EmailEditorPostType[] = select(
+			coreDataStore
+		).getEntityRecords( 'postType', 'mailpoet_email', {
+			per_page: 30, // show a maximum of 30 for now
```
costasovo: Did you try with 30 posts? I am concerned that 30 previews in the modal will cause performance issues, as every preview is basically a small block canvas. But we can still go with 30 and adjust later...

packages/js/email-editor/src/hooks/use-preview-templates.ts
```diff
@@ -40,6 +40,53 @@ function setPostContentInnerBlocks(
 	} );
 }
 
+const InternalCssThemeCache = {};
+
+/**
+ * We are reusing the template CSS and mailpoet theme by fetching the template from
+ * the list of email editor available templates.
+ * Note: This function may need an update when https://mailpoet.atlassian.net/browse/MAILPOET-6335 is merged
```
costasovo, line 54: Yeah. Great point! After MAILPOET-6335 there will be just global styles. Thank you for adding this workaround!

packages/js/email-editor/src/components/template-select/select-modal.tsx
costasovo: Thank you for pointing that out!

As for now we have only the two categories and we don't have problem figuring out what goes where so there is no rush.

I'm not sure the wp-template category and MailPoet selector categories are the same thing. In the selector user picks from template + initial content and I can imagine a template will be displayed in multiple categories because of a combination with a different pattern (or recent email content).
We will probably need some kind of registry and an extra endpoint for this in the future. I would wait with the implementation when we have a second category. The "recent emails" category is a specific automatic category.

costasovo (changes requested):
Great job! It works nicely. 

When testing, I ran into an issue with the recent email preview.
I sent an email created with the Simple Light template and then in the recent emails category the preview rendered only the content part, but not the template.
See screenshot. We could also place the "Start from scratch" button fixed to the bottom right. I think it would look cleaner.

Screenshot 2024-12-06 at 10 03 06

packages/js/email-editor/src/store/selectors.ts
triple0t: Yeah, I did. It worked well.

packages/js/email-editor/src/components/template-select/select-modal.tsx
triple0t: Noted. Thank you for the feedback. 
We will leave it as it is then.

triple0t:

> When testing, I ran into an issue with the recent email preview.
I sent an email created with the Simple Light template and then in the recent emails category the preview rendered only the content part, but not the template.

Fixed in https://github.com/mailpoet/mailpoet/pull/5980/commits/4588f92261101c3a89236eab8359478cb7f7cebf

> See screenshot. We could also place the "Start from scratch" button fixed to the bottom right. I think it would look cleaner.

Fixed in https://github.com/mailpoet/mailpoet/pull/5980/commits/ca4201f89b0e4f38248a41db386fdc6522f4ba8f

costasovo (commented):
@triple0t Thanks for the fix it works as expected!

I see there are conflicts with the trunk branch. It is most likely due to changes I made for email template swap functionality in https://github.com/mailpoet/mailpoet/commit/10b2c2b648af165898a01e8c550adea8347d9a49
This functionality ensures that when swapping a template all previews in the select modal use the content of the currently edited email so that user can see how the email will look like when they change the template.

I'm not sure if that behavior makes senses for recent emails category because it may end up showing just a bunch of same templates as many recent email will be based on the single template. I think we cold:
1) Disable recent emails category when swapping because all temlates should be listed anyway. We can detect it from previewContent being set 
2) Keep the recent email category with the currently edited email content. On select apply only template change
3) Keep the recent email category with the contents from sent emails. 
       3a) on change apply only template change
       3b) on change apply both template and content

I would personally go with the option 1) or 2). Any other ideas?

triple0t:
Oh, thanks for the ping, @costasovo 

Funny, I was focusing on fixing the merge conflict and wanted to message you about this upon completion. (You beat me to it)

> I'm not sure if that behavior makes senses for recent emails category because it may end up showing just a bunch of same templates as many recent email will be based on the single template. I think we cold:
> 
> 1. Disable recent emails category when swapping because all temlates should be listed anyway. We can detect it from previewContent being set
> 2. Keep the recent email category with the currently edited email content. On select apply only template change
> 3. Keep the recent email category with the contents from sent emails.
> 3a) on change apply only template change
> 3b) on change apply both template and content
> I would personally go with the option 1) or 2). Any other ideas?


Let's go with Option 2, as that is the current behavior.
If the need arises, we can change this later.

costasovo (dismissed):
@triple0t Thanks for fixing the conflict!

> Let's go with Option 2, as that is the current behavior.
If the need arises, we can change this later.

Makes sense! 👍

costasovo:
@triple0t I wanted to merge the PR, but I noticed a test started to fail after the latest change. Can you please take a look? I'll be able to look into it on Monday.

costasovo:
@triple0t Oh! And I realized I shouldn't merge it because it is on top of another PR. So, I'll leave merging to you.

triple0t:
Thank you, @costasovo 

Yeah, this PR is based on another.
I fixed the failing test in the parent PR and will rebase it after it is merged.

PR #5991, opened 2024-12-11
Add modal for inserting personalization tags to the new editor [MAILPOET-6354]

lysyjan:
## Description

This PR adds a new modal that can be displayed after clicking on new buttons.
- One button is a part of the rich text toolbar
- The next buttons are a part of the subject title and preheader title

The modal contains all registered personalization tags sorted by their categories.
Each tag can be inserted after clicking on the insert button and should be inserted at the last cursor position in the rich text.

## Code review notes

- Because the HTML comments implementation in Gutenberg does not append `[]`, we need to append them to our tags.
- I registered two new personalization tags for easier implementation of categories in the modal
- There is a known limitation now when two identical tags are in the same text part; when replacing, the editor changes the first of them. I have not found a better approach, and because it is an edge case, I would keep it as it is for now.
- The tags are inserted without attributes. I would like to add this functionality in a follow-up ticket.

## QA notes

1. Try inserting Personalization tags into subject, preheader, and text blocks
2. Try to replace a selected tag

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6354]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6354]: https://mailpoet.atlassian.net/browse/MAILPOET-6354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/tags-ui)

_The latest successful build from `email-editor/tags-ui` will be used. If none is available, the link won't work._


packages/js/email-editor/src/components/block-editor/editor.tsx
```diff
@@ -32,6 +32,12 @@ export function InnerEditor( {
 	initialEdits,
 	...props
 } ) {
+	const { loadPersonalizationTags } = useDispatch( storeName );
+
+	useEffect( () => {
+		void loadPersonalizationTags();
+	}, [ loadPersonalizationTags ] );
```
costasovo: I'm not sure I understand why we have this here. Is it an optimization to preload personalization tags, or is it the main mechanism for fetching them into the app? I don't see any other place where we fetch so I assume it is the latter. The wordpress/data has resolvers which are meant exactly for this see [the docs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#:~:text=normalized%20object.-,resolvers,-A%20resolver%20is)

We don't have any resolver in our store, so you will need to add the first one.  I was looking into docs and asked Chat GPT (`How to add a resolver to store made using @wordpress/data`), and it looks pretty straightforward.

mailpoet/lib/EmailEditor/Integrations/MailPoet/PersonalizationTagManager.php
```diff
@@ -37,6 +42,19 @@ public function initialize() {
...
+        'mailpoet/site-title',
+        __('Site', 'mailpoet'),
+        [$this->site, 'getTitle'],
+      ));
+      $registry->register(new Personalization_Tag(
+        __('Homepage URL', 'mailpoet'),
+        'mailpoet/site-homepage-url',
+        __('Site', 'mailpoet'),
+        [$this->site, 'getHomepageURL'],
+      ));
```
costasovo, line 57: Just an idea for the future... What about moving these two tags into the editor package? The logic behind this idea is similar to core blocks renderer being placed there. We could distribute the editor with a few basic or core tags.

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
```diff
@@ -9,6 +10,13 @@ const CategorySection = ( {
 	groupedTags: Record< string, PersonalizationTag[] >;
 	activeCategory: string | null;
 } ) => {
+	const { togglePersonalizationTagsModal } = useDispatch( storeName );
+
+	const { onInsert } = useSelect(
+		( select ) => select( storeName ).getPersonalizationTagsState(),
+		[]
+	);
```
costasovo: Can you please elaborate on why we need to add the callback into the state? It looks suspicious to me. I would expect we would use a custom action, perhaps a hook.

packages/js/email-editor/src/components/sidebar/details-panel.tsx
```diff
@@ -35,6 +35,220 @@ export function DetailsPanel() {
...
 	const { togglePersonalizationTagsModal, updateEmailMailPoetProperty } =
 		useDispatch( storeName );
+	const [ activeRichText, setActiveRichText ] = useState( null );
+	const [ selectionRange, setSelectionRange ] = useState( null );
+
+	const subjectRef = useRef( null );
+	const preheaderRef = useRef( null );
+
+	// Convert `RichText` DOM offset to stored value offset
+	const mapRichTextToValue = ( html ) => {
```
costasovo: I would suggest using `React.useCallback` for the long functions defined within a functional component. Without the useCallback the function is recreated on every re-render of the component.

packages/js/email-editor/src/components/personalization-tags/rich-text-utils.tsx
```diff
@@ -0,0 +1,191 @@
+import * as React from '@wordpress/element';
```
costasovo: It is just a detail. I think this file could be `rich-text-utils.ts` because there is no jsx syntax.

costasovo (commented):
@lysyjan Fantastic job! It is cool to play with this functionality in the editor. 

I did a time-boxed review and testing on my dev site. I didn't look in detail into the rich text related functions. I found a couple of tiny issues and have a few questions and notices. Let's handle these first and I'll do second round. 

### Issues from testing

**1) Link in modal is not clickable**
Screenshot 2024-12-11 at 17 21 10

**2) Tag in subject and preheader can break **
This is probably an edge case, but if it is an easy fix we should address it. When I placed a cursor next to the first part of the tag that was no two lines and inserted a new tag it broke and you can see some pieces of an HTML comment. Please see the video below.

**3) Modal opened for preheader inserted tag to subject**
This is also visible in the video. Notice that I opened the modal by clicking on the icon above the preheader but after selecting a tag the tag was inserted into the subject's rich text.


https://github.com/user-attachments/assets/cd51b201-ffbf-4aa5-9140-c34c503fb4c9

packages/js/email-editor/src/components/block-editor/editor.tsx
lysyjan: Good point with a resolver. I refactored this initialization into a proposed solution.

mailpoet/lib/EmailEditor/Integrations/MailPoet/PersonalizationTagManager.php
lysyjan: This sounds good to me. I think we can do it in a follow-up ticket where we can also add `date` tags.

lysyjan:
@costasovo Thank you for your testing and review.

> Link in modal is not clickable

I configured the link to the old shortcodes page and created a ticket to address the final adjustment.

> Tag in subject and preheader can break

I want to look closer at this issue in one of the follow-up tickets when I will add support for editing.

> Modal opened for preheader inserted tag to subject

I refactored the modal component, which solved the issue.

One of the acceptance tests is failing and seems to be irrelevant to this PR. I hope it will be fixed after merging another PR and rebasing.

packages/js/email-editor/src/blocks/core/rich-text.tsx
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Extend the rich text formats with a button for personalization tags.
+ */
+function extendRichTextFormats() {
+	registerFormatType( 'mailpoet-email-editor/shortcode', {
+		title: __( 'Personalization Tags', 'mailpoet' ),
+		className: 'mailpoet-email-editor-personalization-tags',
+		tagName: 'span',
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment -- The types does not match
+		// @ts-ignore
```
costasovo: Tiny detail, but please use `// @ts-expect-error` instead of `// @ts-ignore`. This is better because it will notify us to remove the comment in case we fix it or it is fixed in the core.

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
```diff
@@ -0,0 +1,57 @@
...
+									<div className="mailpoet-personalization-tags-modal__item-text">
+										<strong>{ item.name }</strong>
+										{ item.token }
+									</div>
+									<Button
+										variant="link"
+										onClick={ () => {
+											if ( onInsert ) {
+												onInsert( item.token );
+											}
```
costasovo, line 42: This could be simplified `onInsert && onInsert( item.token );` I think we use this syntax more often. But no need to update. Just FYI.

packages/js/email-editor/src/components/personalization-tags/personalization-tags-modal.tsx
```diff
@@ -0,0 +1,73 @@
...
+	const [ searchQuery, setSearchQuery ] = useState( '' );
+
+	const list = useSelect(
+		( select ) => select( storeName ).getPersonalizationTagsList(),
+		[]
+	);
+
+	if ( ! isOpened ) {
+		return null;
+	}
```
costasovo, line 21: Just a small performance optimization. If we put this logic outside of the modal component, we would save creating two states and fetching the list of tags from the store when the modal is not displayed, which we do now. I see that every component that uses this modal already controls the state of modal so we could just do 
```js
{ isModalOpen && <PersonalizationTagsModal onInsert={someFunction} closeCallback={closeCallbackFunction}}
```
With this the modal would be skipped entirely unless it is needed and we would select tag list from store only after user really opens it.

packages/js/email-editor/src/components/personalization-tags/rich-text-utils.ts
```diff
@@ -0,0 +1,191 @@
+import * as React from '@wordpress/element';
+
+/**
+ * Maps HTML indices to corresponding stored value indices in RichText content.
+ * This function skips over HTML tags, only mapping visible text content.
```
costasovo: I needed to read the whole function to understand what it does and why it is needed. I offer a slightly altered description that I believe would help me to understand easily. And I hope I understood it correctly 😅 

```
/**
 * Maps indices of characters in HTML representation of the value to corresponding characters of stored value in RichText content. The stored value doesn't contain tags.
 * This function skips over HTML tags, only mapping visible text content.
 *
 ```

File: packages/js/email-editor/src/components/personalization-tags/rich-text-with-button.tsx [hunk 10]
```diff
@@ -0,0 +1,138 @@
...
+			updateEmailMailPoetProperty( attributeName, updatedValue );
+
+			setSelectionRange( null );
+		},
+		[
+			attributeName,
+			mailpoetEmailData,
+			selectionRange,
+			updateEmailMailPoetProperty,
+		]
```
costasovo: I see the useCallback needs quite a lot of deps. Looking at it I think we could make it more "cachable" if we passed some values as parameters. I think we could pass: the current value and selection range, and it would improve the caching.
So in onInsert, we would call

```js
handleInsertPersonalizationTag( value, mailpoetEmailData[ attributeName ] ?? '', selectionRange );
```

What do you think?

costasovo (commented):
@lysyjan Thank you for addressing all the previous comments. 👍 

I did a second round of review and focused a bit more on details. I added a couple of comments with ideas for tiny and mostly performance improvements. 

After reading through the content to HTML mapping functions, I played a bit with adding tags to paragraphs and I found that when I try to place a tag right after a bold tag it crashes the block (see attached video). I'm able to replicate it reliably. We discussed that these issues will be addressed in follow-up PRs so no need to fix now. I just wanted to let you know so that you are aware also of this one.


https://github.com/user-attachments/assets/2911a588-9af3-4316-888a-7e71011d97a0

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
lysyjan: I tried it, but the linter in the email editor JS package does not allow it.

packages/js/email-editor/src/components/personalization-tags/personalization-tags-modal.tsx
lysyjan: I would do this as a part of another ticket because this also needs some spinner to improve UX. Do you agree?

lysyjan:
@costasovo I added two new commits with small adjustments.

costasovo (dismissed):

costasovo:
@Aschepikov There are two known issues we plan to address in a follow-up PR:

1) Tag in subject or preheader might break in case it is on multiple lines and the user places cursor in between (see video in https://github.com/mailpoet/mailpoet/pull/5991#pullrequestreview-2496173318)

2) Block may crash when placing a tag after bold and probably also other formatted text (see video in https://github.com/mailpoet/mailpoet/pull/5991#pullrequestreview-2502175600)

Aschepikov:
Hi @lysyjan looks like there is a failed test can you check it pls

lysyjan:
@Aschepikov The test should be fixed after merging and rebasing [the PR](https://github.com/mailpoet/mailpoet/pull/5993) created by Rosta. The PR is approved. Do you think I can merge it?

Aschepikov:
Hi @lysyjan can you rebase it now?

Aschepikov:
@lysyjan it is still failing

lysyjan:
@Aschepikov I will take a look and let you know.

costasovo:
@Aschepikov I fixed the problem with the test.

Aschepikov:
@lysyjan tags are not saving when I am choosing the recent emails with them is it expected?
image
image

lysyjan:
@Aschepikov I am not sure how recent emails should work. I will check it with the rest of the team, and we can address it in a follow-up ticket if needed. Would that work for you?

lysyjan:
@Aschepikov, I just checked how the old email editor works. The plugin does not copy the subject, and it copies only the content. So I think it is okay. cc @costasovo @pavel-mailpoet

Aschepikov:
@lysyjan  Sure it is ok. Also, I don't see the tags in the text block on the preview and sent email.
image

Aschepikov (approved):
GJ

ellatrix:
> Because the HTML comments implementation in Gutenberg does not append [], we need to append them to our tags.

Why do you need []? The whole point of using these HTML comments is so that it's forward compatible with the future Bits. If the pattern doesn't match `<!--/type attribute="key"-->`, then it's not forward compatible.

lysyjan:
There were several reasons for that.
- when the square brackets are missing, it can look like text with a color background only
- because we wanted to have a similar user experience as in the old editor, where our users have used this syntax for a couple of years. The idea was to have as similar as it is possible
- when we decide to migrate to Bits, we can prepare a simple DB migration, which migrates the old syntax to the new one
- this syntax is similar to WP shortcodes that are WordPress uses used to now

I tried to find out the final Bits syntax two months ago, but I was unsuccessful.
So adding `[]` inside the comment made sense to me

ellatrix:
> when the square brackets are missing, it can look like text with a color background only

You could use CSS?

PR #5993, opened 2024-12-12
Email editor – Global Styles [MAILPOET-6335]

costasovo:
## Description

This PR refactors how we store the email styles configurable via the style panels.

- the styles are no longer saved in the post meta, but they are saved in the `wp_posts` table as `wp_global_styles` CPT. The post is identified by `post_name` column value `wp-global-styles-mailpoet-email`
- there is only a single instance of the styles that works as global email styles and is used for all emails and templates
- the styles are saved independently from a template (see screenshot) and are no longer associated with a template

Screenshot 2024-12-12 at 16 20 10

## Code review notes

- I chose to create the global styles post lazily upfront instead of creating it from the editor. It is possible that later when we add custom variations, it will be more convenient to create the style posts from the editor, but for now, this was an easier variant to do.
- I didn't add any migration for existing emails. They will all switch to the default styles.

## QA notes

The email styles are now global, so they affect all emails. Please verify there is no regression in how the styles work and that they are applied in the editor and renderer, and if saved, they are applied to all emails.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6335]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6335]: https://mailpoet.atlassian.net/browse/MAILPOET-6335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/global-styles)

_The latest successful build from `email-editor/global-styles` will be used. If none is available, the link won't work._


lysyjan (approved):
It looks good to me. 👍
Great job with performance improvements.

Aschepikov (approved):
GJ!

PR #5996, opened 2024-12-16
Bundle WP RichText package as a part of the plugin [MAILPOET-6387]

lysyjan:
## Description

This PR temporarily bundles the WP RichText package with the plugin. This allows us to use Personalization Tags in the new email editor without an active Gutenberg plugin.

## Code review notes

- I started bundling RichText with the email editor JS file, but this solution does not work. I found out that the best possible approach is to replace the already registered RichText JS file with our own containing the required version.
- Because some CSS were missing (They are a part of the BlockEditor package), I added them to our styles.

## QA notes

Use the plugin on the page where the Gutenberg plugin is deactivated and check that HTML comments are correctly rendered in the email editor.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6387]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6387]: https://mailpoet.atlassian.net/browse/MAILPOET-6387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/rich-text-bundle)

_The latest successful build from `email-editor/rich-text-bundle` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/webpack.config.js
+++ b/mailpoet/webpack.config.js
@@ -563,24 +563,6 @@ const emailEditorRichText = Object.assign({}, wpScriptConfig, {
       },
     ],
   }),
-  plugins: [
-    ...wpScriptConfig.plugins,
-    new DependencyExtractionWebpackPlugin({
-      injectPolyfill: true,
-      requestToExternal: (request) => {
-        if (request.startsWith('@wordpress/')) {
-          return `wp.${request.replace('@wordpress/', '')}`;
-        }
-        return undefined;
-      },
-      requestToHandle: (request) => {
-        if (request.startsWith('@wordpress/')) {
-          return `wp-${request.replace('@wordpress/', '-')}`;
-        }
-        return undefined;
-      },
-    }),
-  ],
 });
 
 const emailEditorIntegration = Object.assign({}, wpScriptConfig, {
--- a/mailpoet/webpack.config.js
+++ b/mailpoet/webpack.config.js
@@ -530,6 +530,59 @@ const emailEditorCustom = Object.assign({}, wpScriptConfig, {
     : [...wpScriptConfig.plugins, new ForkTsCheckerWebpackPlugin()],
 });
 
+// Temporary solution to build rich-text package for email editor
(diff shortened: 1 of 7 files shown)


mailpoet/webpack.config.js
```diff
@@ -530,6 +530,59 @@ const emailEditorCustom = Object.assign({}, wpScriptConfig, {
...
+        return undefined;
+      },
+      requestToHandle: (request) => {
+        if (request.startsWith('@wordpress/')) {
+          return `wp-${request.replace('@wordpress/', '-')}`;
+        }
+        return undefined;
+      },
+    }),
+  ],
```
costasovo: I think we can remove these lines. If I understand correctly, they don't change anything in the default DependencyExtractionWebpackPlugin behavior. Locally it worked for me even without these lines.

costasovo (commented):
@lysyjan Great job! Awesome that we can replace just this one package!

I think that part of the config is not necessary. Please see my other comment. 

I needed to rebase this branch on top of email-editor/tags-ui. I assume we should give this to QA after the UI is merged. Correct?

mailpoet/webpack.config.js
lysyjan: I removed the whole plugins part, and it looks like it works.

lysyjan:
> I needed to rebase this branch on top of email-editor/tags-ui. I assume we should give this to QA after the UI is merged. Correct?

@costasovo I think it is possible to test it when the site contains a crated email with Personalization tags. We could skip QA if we both test it and it works. What do you think?

costasovo (approved):

costasovo:
> I think it is possible to test it when the site contains a crated email with Personalization tags. We could skip QA if we both test it and it works. What do you think?

@lysyjan Let's skip QA. It will be tested within the PR that adds the UI.

PR #5997, opened 2024-12-17
Soft deactivation of MailPoet [MAILPOET-6389]

websupporter:
## Description
If the current WP, Woo or PHP version is lower than needed, MailPoet deactivates itself. With this PR we will "soft-deactivate". This means, every code except for the mailpoet.php code will not be executed. Yet, as soon as the problem is resolved, MailPoet works as expected and you do not need to reactivate the plugin again. It also means that the notices we show like e.g.

image

will be permanently visible and inform the user about the underlying issue.

When testing, one problem I saw:
In that "soft-deactivated" mode we do not declare capability e.g. with HPOS. Therefore, WooCommerce will start to declare our plugin would not be compatible with HPOS. I believe, since we display our error message as well63, that is okay.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs
https://github.com/mailpoet/mailpoet-premium/pull/929
## Linked tickets

[MAILPOET-6389]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6389]: https://mailpoet.atlassian.net/browse/MAILPOET-6389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:MAILPOET-6389-soft-deactivation)

_The latest successful build from `MAILPOET-6389-soft-deactivation` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/mailpoet.php
+++ b/mailpoet/mailpoet.php
@@ -37,36 +37,6 @@ function mailpoet_deactivate_plugin() {
   }
 }
 
-// Check for minimum supported WP version
-if (version_compare(get_bloginfo('version'), MAILPOET_MINIMUM_REQUIRED_WP_VERSION, '<')) {
-  add_action('admin_notices', 'mailpoet_wp_version_notice');
-  // deactivate the plugin
-  add_action('admin_init', 'mailpoet_deactivate_plugin');
-  return;
-}
-
-// Check for minimum supported PHP version
-if (version_compare(phpversion(), '7.4.0', '<')) {
-  add_action('admin_notices', 'mailpoet_php_version_notice');
-  // deactivate the plugin
-  add_action('admin_init', 'mailpoet_deactivate_plugin');
-  return;
-}
-
-// Check for minimum supported WooCommerce version
-if (!function_exists('is_plugin_active')) {
-  require_once ABSPATH . 'wp-admin/includes/plugin.php';
-}
-if (is_plugin_active('woocommerce/woocommerce.php')) {
-  $woocommerceVersion = get_plugin_data(WP_PLUGIN_DIR . '/woocommerce/woocommerce.php', false, false)['Version'];
-  if (version_compare($woocommerceVersion, MAILPOET_MINIMUM_REQUIRED_WOOCOMMERCE_VERSION, '<')) {
-    add_action('admin_notices', 'mailpoet_woocommerce_version_notice');
-    // deactivate the plugin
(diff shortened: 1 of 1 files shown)


NeosinneR (dismissed):
LGTM

NeosinneR (approved):

mailpoet/mailpoet.php
```diff
@@ -132,60 +114,64 @@ function mailpoet_php_version_notice() {
...
-  $noticeP3 = str_replace('[/link]', '</a>', $noticeP3);
-
   $allowedTags = [
     'a' => [
       'href' => true,
       'target' => true,
     ],
   ];
   printf(
-    '<div class="error"><p><strong>%s</strong></p><p>%s</p><p>%s</p></div>',
```
websupporter, line 150: ouch. thanks. good catch!

mailpoet/mailpoet.php
```diff
@@ -97,8 +97,14 @@ function mailpoet_core_dependency_notice() {
 
 // Display PHP version error notice
 function mailpoet_php_version_notice() {
-  $noticeP1 = __('MailPoet requires PHP version 7.4 or newer (8.1 recommended). You are running version [version].', 'mailpoet');
-  $noticeP1 = str_replace('[version]', phpversion(), $noticeP1);
+  $noticeP1 = sprintf(
+    // translators: %1$s is the plugin name (MailPoet or MailPoet Premium), %2$s, %3$s, and %4$s are PHP version (e.g. "8.1.30")
+    __('%1$s requires PHP version %2$s or newer (%3$s recommended). You are running version %4$s.', 'mailpoet'),
```
websupporter, line 102: Not sure if the "MailPoet" "MailPoet Premium" replacement is of value, since this string is then just double in two different projects/text-domains (mailpoet and mailpoet-premium), but no objections. The other replacements are 👍

websupporter (commented):
lgtm

mailpoet/mailpoet.php
NeosinneR: Yeah, I originally wanted to use the `mailpoet` text domain also in premium, but I wasn't sure if that has any implications. But at least this way, Transifex can use the same translation for both strings, even if they are in different projects, so in practice, they should be translated automatically.

PR #6003, opened 2024-12-18
Add telemetry support for the email editor - MAILPOET-6365

triple0t:
## Description

We now collect basic telemetry information from the email editor and expose them to extenders via the `do_action` hook - `mailpoet_email_editor_events`. 

By default, the telemetry collection is **disabled**. Extenders would need to opt in by returning `true` to the filter hook - `mailpoet_email_editor_events_tracking_enabled`  

Here is an example of how it can be done:
https://github.com/mailpoet/mailpoet/blob/f2708f416d69470370ad5fdee58b90f49fbcfeea/mailpoet/assets/js/src/mailpoet-email-editor-integration/index.ts#L14-L24

## Code review notes

I tried to break down each commit to be related to specific (or group of related) components.
Most of the commit is basically on adding the tracking function.

Please review at your leisure.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6365](https://mailpoet.atlassian.net/browse/MAILPOET-6365)

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:add-user-tracking-to-mixpanel)

_The latest successful build from `add-user-tracking-to-mixpanel` will be used. If none is available, the link won't work._

[MAILPOET-6365]: https://mailpoet.atlassian.net/browse/MAILPOET-6365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/js/email-editor/src/components/header/send-button.tsx
```diff
@@ -41,6 +42,7 @@ export function SendButton() {
 		<Button
 			variant="primary"
 			onClick={ () => {
+				recordEvent( 'header_send_button_clicked' );
```
costasovo, line 45: This tracking is not 100% reliable. Mixpanel stacks events and sends them in batches and window.location.href will reload the page often before the event is sent. 
Mixpanel has a callback parameter and send immediately option for that ([documentation](https://github.com/mixpanel/mixpanel-js/blob/master/doc/readme.io/javascript-full-api-reference.md#mixpaneltrack)).
We use it for example in homepage when we open knowledge base links ([helper code](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/assets/js/src/homepage/tracking.ts#L4-L13)).
We will also need to add support for these types of events to our framework. I think it might be done in a follow-up PR.
This concern might also be valid when opening a link in a new tab because the browser stops JS execution in the inactive tab, and if the user doesn't revisit it, the event is not tracked.

packages/js/email-editor/src/events/event-pipeline.ts
```diff
@@ -0,0 +1,18 @@
+const EMAIL_STRING = 'email-editor-events';
+
+const dispatcher = new EventTarget();
```
costasovo, line 11: What is the benefit of using this dispatcher? Why can't we just call the `doAction( 'mailpoet-email-editor-events', eventData.detail );` directly in the `recordEvent`? I assume there is a reason for this I just struggle to see it.

costasovo (commented):
@triple0t, Thank you for adding such an extensive tracking coverage! Also, thank you for adding rendering optimizations!

I did a brief code walkthrough and locally tested some of the events with console.log. I'll check the code more deeply in the second review.

I've found an issue with tracking links. Please see the other comments.

Also, I am concerned that such extensive tracking might be costly. I see in our Mixpanel admin that we are over the quota every month. In December, we consumed 115% and we don't track granular actions like in the editor. The editor tracks two events just when it opens. I understand that having such detailed overview about the editor usage is helpful. I want to make sure we are aware of the potential high event volume. cc @pavel-mailpoet

pavel-mailpoet:
> Also, I am concerned that such extensive tracking might be costly. I see in our Mixpanel admin that we are over the quota every month. In December, we consumed 115% and we don't track granular actions like in the editor. The editor tracks two events just when it opens. I understand that having such detailed overview about the editor usage is helpful. I want to make sure we are aware of the potential high event volume.

Good comment, Rosta, @triple0t, could you please connect with Tautvidas and confirm what specifically he wants to track and if he is ok with the volume?

packages/js/email-editor/src/events/event-pipeline.ts
triple0t: For the V1, I wanted something easy to extend, highly versatile, and optimized performance-wise.
I also need to confirm if it's okay to continue using the library as it is (using `doAction`) or if there would be a performance gain if we just dispatch the event on the email editor DOM. 

Extenders would then listen for the events on the DOM (`mailpoet-email-editor`) element.

For now, I want to keep this flexibility.

packages/js/email-editor/src/components/header/send-button.tsx
triple0t: Thanks for noting this.
We will need to ensure whatever solution we implement works for both Mixpanel and Tracks. I will leave this here and update it in another PR.

triple0t:
Thanks for the in-depth review, @costasovo 

> I've found an issue with tracking links. Please see the other comments.

I replied the comment.

> Also, I am concerned that such extensive tracking might be costly. I see in our Mixpanel admin that we are over the quota every month. In December, we consumed 115% and we don't track granular actions like in the editor. The editor tracks two events just when it opens. I understand that having such detailed overview about the editor usage is helpful. I want to make sure we are aware of the potential high event volume. cc @pavel-mailpoet

I believe the primary use case here is for Tracks, where data volume would not be an issue. 
I agree with you; we need to figure out a way or introduce some kind of event log level that Extenders can opt-in to. 

For MailPoet Mixpanel, we can filter out events [here](https://github.com/mailpoet/mailpoet/blob/1ea261adbb6bfc88ad2832bab69e5c33c3647d12/mailpoet/assets/js/src/mailpoet-email-editor-integration/index.ts#L14-L17) before they are sent to Mixpanel.

I will confirm with Tautvidas and ensure we log only the data MailPoet requires on Mixpanel.

> Could you please connect with Tautvidas and confirm what specifically he wants to track and if he is ok with the volume?

Noted, I will do so. @pavel-mailpoet

packages/js/email-editor/src/events/event-pipeline.ts
costasovo: Ok. Thank you for the additional context! It makes sense to keep it.

packages/js/email-editor/src/components/header/send-button.tsx
costasovo: I see. We need an API that works for both. It makes sense to cover this in the follow-up PR, as we will know what works for Tracks.

triple0t:
All done. 
We are not sending [limited events](https://github.com/mailpoet/mailpoet/pull/6003/commits/3a600f45a351e837c43e40c1dd3f6ebf2f2d96e1) to Mixpanel.

packages/js/email-editor/src/components/block-editor/visual-editor/edit-template-blocks-notification.tsx
```diff
@@ -36,6 +37,10 @@ export default function EditTemplateBlocksNotification( { contentRef } ) {
 		};
 	}, [] );
 
+	recordEventOnce( 'edit_template_blocks_notification_opened', {
```
costasovo: I see this event is tracked on every load of the editor. I think we need to move this tracking call next to `setIsDialogOpen( true );` a couple of lines below.

packages/js/email-editor/src/components/styles-sidebar/panels/dimensions-panel.tsx
```diff
@@ -68,19 +79,26 @@ export function DimensionsPanel() {
...
 				className="tools-panel-item-spacing"
 			>
 				<SpacingSizesControl
 					label={ __( 'Block spacing', 'mailpoet' ) }
 					min={ 0 }
 					onChange={ ( value ) => {
 						updateStyleProp( [ 'spacing', 'blockGap' ], value.top );
+						recordEvent(
+							'styles_sidebar_screen_layout_dimensions_block_spacing_updated',
+							{ value }
```
costasovo: This event would use a debounce. When I use the range control, I see dozens of event-tracking calls within a second.

Screenshot 2025-01-08 at 16 55 28

packages/js/email-editor/src/components/styles-sidebar/panels/dimensions-panel.tsx
```diff
@@ -36,19 +40,26 @@ export function DimensionsPanel() {
...
 			>
 				<SpacingSizesControl
 					allowReset
 					values={ styles.spacing.padding }
 					onChange={ ( value ) => {
 						updateStyleProp( [ 'spacing', 'padding' ], value );
+						recordEvent(
+							'styles_sidebar_screen_layout_dimensions_padding_updated',
+							{ value }
+						);
```
costasovo: There is a range control and when moving the range control I see multiple events being tracked. I think we could use debounce or something to prevent multiple events triggered while user is adjusting the value and ideally track only the final value or one event to be able to see how many times user changed the setting.

packages/js/email-editor/src/components/styles-sidebar/panels/typography-element-panel.tsx
```diff
@@ -111,26 +112,76 @@ export function TypographyElementPanel( {
...
+		recordEvent(
+			'styles_sidebar_screen_typography_element_panel_set_font_size',
+			{
+				element,
+				headingLevel,
+				newValue,
+				selectedDefaultFontSize: newValue === defaultFontSize,
+			}
+		);
 	};
```
costasovo: These events could be logged multiple times when the user uses range control or up/down arrows on the keyboard. I think in this place we want to track that user adjusted font size and logging multiple events when user manipulate range control would skew our data. Am I correct?
I recorded a video how events are tracked.

https://github.com/user-attachments/assets/fc0ee4fc-eea7-4930-8eed-f76b001a3819

mailpoet/assets/js/src/newsletters/send.tsx
```diff
@@ -322,6 +322,13 @@ class NewsletterSendComponent extends Component<
           thumbnailPromise,
           validationError: validateNewsletter(response.data),
         });
+
+        if (response.data?.wp_post_id) {
+          MailPoet.trackEvent(
+            'New Email Editor > Newsletter created by BlockEmailEditor',
+          );
+        }
```
costasovo: I see that this event is tracked every time the send page is loaded. So if user reloads the page or return back to the editor and re-enters the page we track `'New Email Editor > Newsletter created by BlockEmailEditor'`. I'm not sure what to suggest here. Perhaps, could we track on user action when a user clicks "Send"?

costasovo (commented):
@triple0t Thank you for the update! I tested how the events  are tracked by logging them into console and added a couple of comments for cases which I found potentially problematic. In general it seems that we could use some kind of debounce to prevent tracking multiple events when user use controls like range control. Could we add something at the low level per event name?

mailpoet/assets/js/src/newsletters/send.tsx
triple0t: Hmm, the idea behind this section is to track when a newsletter created by the new editor gets to the Send page. 

I can update the wording, but this entails knowing that users find value in the editor. They might or might not be able to send a newsletter, but the fact that they got to this page is encouraging. 

Does this make sense, @costasovo?

costasovo: > Does this make sense, @costasovo?

Yes, it makes sense! With the current wording, seeing the event without the code context, I would assume it is tracked when a newsletter is created. So perhaps a different wording isn't a bad idea. Would `'New Email Editor > Send page accessed'` work?

packages/js/email-editor/src/components/block-editor/visual-editor/edit-template-blocks-notification.tsx
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/6003/commits/9e3898510e269127fc84053957c503a8e56abe37

packages/js/email-editor/src/components/styles-sidebar/panels/typography-element-panel.tsx
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/6003/commits/15bd07f23341807f82ad61f9246c18d82064fbec

packages/js/email-editor/src/components/styles-sidebar/panels/dimensions-panel.tsx
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/6003/commits/15bd07f23341807f82ad61f9246c18d82064fbec

mailpoet/assets/js/src/newsletters/send.tsx
triple0t: oppss, I saw your comment late.
Fixed in https://github.com/mailpoet/mailpoet/pull/6003/commits/8f41e1fa7d24b875bef2df32ab3ba1395cc85123

costasovo (commented):
@triple0t Thank you for the updates! I see that all the issues I mentioned were addressed. Great job!

I think I found one more thing we need to address. I missed that in the previous review.
When we track `template_select_modal_template_selected` event we send the variable to the Mixpanel. The potential problem is that Mixpanel has[ limits for event size and count of properties (included nested)](https://developer.mixpanel.com/reference/track-event#:~:text=Each%20event%20has%20the%20following,have%20fewer%20than%20255%20properties.) The template variable contains the full document and also parsed blocks so I'm afraid we can easily reach the property limit here (see screenshot). I think we need to send only a few selected properties.

Screenshot 2025-01-09 at 16 03 11

I checked the rest of the events we sent to Mixpanel and this was the only one with problematic data. I haven't checked the rest of the events that are not allowed to be sent to Mixpanel.

triple0t:
> I think I found one more thing we need to address. I missed that in the previous review.
When we track template_select_modal_template_selected event we send the variable to the Mixpanel. The potential problem is that Mixpanel has[ limits for event size and count of properties (included nested)](https://developer.mixpanel.com/reference/track-event#:~:text=Each%20event%20has%20the%20following,have%20fewer%20than%20255%20properties.) The template variable contains the full document and also parsed blocks so I'm afraid we can easily reach the property limit here (see screenshot). I think we need to send only a few selected properties.

Thanks for catching this, @costasovo. Yes, this was an error; we should not be sending the full object to Mixpanel.
Fixed in https://github.com/mailpoet/mailpoet/pull/6003/commits/fc9cb8a044e1a535b5280b64d7d532e9f25bed4b

costasovo (approved):
@triple0t Thanks for the fix! I think it is good to go. I'm not sure if we should send this to QA or skip QA and merge it. What do you think?

triple0t:
@costasovo Thanks. We can skip QA and merge

PR #6000, opened 2024-12-17
Email editor- register_block_template [MAILPOET-6356]

costasovo:
## Description

This PR is refactoring the templates to the new `register_block_template` function added in WP 6.7

## Code review notes

Please review commit by commit.
I had to update WordPress stubs for PHPStan because the new function was in the old stubs.
You may notice this change and a similar one with getPostTypes:
```php
- $this->wp->getTaxonomies([], 'object'),
+ $this->wp->getTaxonomies([], 'objects'),
```
It is safe because,[ internally, anything else than `name` or `names` is processed as `objects`.](https://github.com/WordPress/wordpress-develop/blob/e58bef6fdc10bbbbba02e58a34dbca022e956543/src/wp-includes/taxonomy.php#L285)

## QA notes

The templates should work the same as before.
You can test:

- creating a new email
- swapping template 
- editing template

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/930

## Linked tickets

[MAILPOET-6356]

## After-merge notes

_N/A_

## Tasks

- [x] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [x] I added sufficient test coverage
- [x] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-6356]: https://mailpoet.atlassian.net/browse/MAILPOET-6356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/register-templates)

_The latest successful build from `email-editor/register-templates` will be used. If none is available, the link won't work._

PR #6006, opened 2024-12-19
Email editor - Add basic template [MAILPOET-6334]

costasovo:
## Description

This PR replaces previous testing templates and email content patterns with "Newsletter" template and three patterns: 1 column, 2 column, and 3 column.

Screenshot 2024-12-19 at 16 39 06

## Code review notes

Please start review with https://github.com/mailpoet/mailpoet/commit/320572a3cbc089fddd778afe5987fe49ee5c56a6
The PR is on top of https://github.com/mailpoet/mailpoet/pull/6000 and must be merged after it.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6334]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6334]: https://mailpoet.atlassian.net/browse/MAILPOET-6334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/add-basic-template)

_The latest successful build from `email-editor/add-basic-template` will be used. If none is available, the link won't work._


packages/js/email-editor/src/components/template-select/select-modal.tsx
```diff
@@ -1,8 +1,15 @@
...
+ * WordPress dependencies
+ */
 import { useState, useEffect, memo } from '@wordpress/element';
 import { store as editorStore } from '@wordpress/editor';
 import { dispatch } from '@wordpress/data';
 import { Modal, Button, Flex, FlexItem } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
```
pavel-mailpoet, line 11: Do we really need those comments? It seems quite redundant to me.

packages/php/email-editor/src/Engine/Renderer/ContentRenderer/Preprocessors/class-spacing-preprocessor.php
```diff
@@ -39,13 +39,13 @@ private function add_block_gaps( array $parsed_blocks, string $gap = '', $parent
...
 			$block['email_attrs'] = $block['email_attrs'] ?? array();
+
 			/**
 			 * Do not add a gap to:
-			 * - the top level blocks - they are post-content, and header and footer wrappers and we don't want a gap between those
 			 * - first child
 			 * - parent block is a buttons block (where buttons are side by side).
 			 */
-			if ( $parent_block && 0 !== $key && $gap && 'core/buttons' !== $parent_block_name ) {
+			if ( 0 !== $key && $gap && 'core/buttons' !== $parent_block_name ) {
```
pavel-mailpoet, line 48: No action required, but I think we should really extract complicated conditions like this into a separate function.

pavel-mailpoet (dismissed):
Looks good to me, and works fine, approving.

packages/js/email-editor/src/components/template-select/select-modal.tsx
costasovo: I added those to a couple of components. Woo and Gutenberg do the same. AFAIK in Gutenberg it is enforced. IMO, it makes the imports visually more organized. But if you don't like them, I can remove them.

packages/php/email-editor/src/Engine/Renderer/ContentRenderer/Preprocessors/class-spacing-preprocessor.php
costasovo: The condition here should be simple. IMO the `'core/buttons' !== $parent_block_name ` should not be there and this code shouldn't depend on block names, because it is not extendible for new blocks. This is one of the reasons to look into better architecture for the renderer.

packages/js/email-editor/src/components/template-select/select-modal.tsx
pavel-mailpoet: Do not remove them now. 

I just don't understand the benefit. IMO the imports could be visually organised the same way without this comment and with only empty line. If this is not enforced by any tool I wonder if we will be able to maintain those sections.

costasovo: > If this is not enforced by any tool I wonder if we will be able to maintain those sections.

True. I'll keep it for now. I added a discussion point to our team meeting agenda. We can add the check or remove it...

Aschepikov:
Hi @costasovo I see there 5 templates not sure why 
image
When the page is loading there are just 3 templates, but after clicking on the templates again there are 5. Do you have an idea why?
image

costasovo:
> When the page is loading there are just 3 templates, but after clicking on the templates again there are 5. Do you have an idea why?

@Aschepikov, Can you please share what WP version and theme you use? Do you have Gutenberg plugin active?

Aschepikov:
@costasovo  wp version 6.7.1
Theme Storefront
Gutenberg plugin is active

costasovo:
@Aschepikov Which version of the Gutenberg plugin do you use? Can you try with inactive Gutenberg?

Aschepikov:
@costasovo Without Gutenberg I see the same, Version 19.9.0

costasovo:
@Aschepikov Thank you for providing the details. I was able to find and fix the issue. Could you please take another look?

Aschepikov (approved):
GJ!

PR #6018, opened 2024-12-31
Email editor - dependencies check [MAILPOET-6367]

costasovo:
## Description

This PR adds a mechanism for checking if a site has the dependencies needed to run the email editor. At this point, we are checking for a minimal WP (6.7).
When the editor can't run, we hide the button for creating an email in the new editor, and in case a user attempts to edit an email previously created in the new editor, they get a notice.

Screenshot 2025-01-09 at 18 27 14

## Code review notes

Please review commit by commit.

## QA notes

You can use [WP Downgrade](https://wordpress.org/plugins/wp-downgrade/) plugin to install older WP.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6367]

## After-merge notes

_N/A_

## Tasks

- [ ] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [ ] I added sufficient test coverage
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

[MAILPOET-6367]: https://mailpoet.atlassian.net/browse/MAILPOET-6367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/feature-check)

_The latest successful build from `email-editor/feature-check` will be used. If none is available, the link won't work._

PR #6029, opened 2025-01-08
Remove feature flag for the Gutenberg email editor - MAILPOET-6366

triple0t:
## Description

Remove feature flag for the Gutenberg email editor 

## Code review notes

_N/A_

## QA notes

We can skip QA.

## Linked PRs

Related to https://github.com/mailpoet/mailpoet/pull/6018

## Linked tickets

[MAILPOET-6366](https://mailpoet.atlassian.net/browse/MAILPOET-6366)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6366]: https://mailpoet.atlassian.net/browse/MAILPOET-6366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:remove-feature-flag)

_The latest successful build from `remove-feature-flag` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/assets/js/src/newsletters/types.tsx
+++ b/mailpoet/assets/js/src/newsletters/types.tsx
@@ -36,9 +36,7 @@ export function NewsletterTypes({
   const [isCreating, setIsCreating] = useState(false);
 
   const [isSelectEditorModalOpen, setIsSelectEditorModalOpen] = useState(false);
-  const isNewEmailEditorEnabled =
-    MailPoet.FeaturesController.isSupported('gutenberg_email_editor') &&
-    window.mailpoet_block_email_editor_enabled;
+  const isNewEmailEditorEnabled = window.mailpoet_block_email_editor_enabled;
 
   const setupNewsletter = (type): void => {
     if (type !== undefined) {
--- a/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
+++ b/mailpoet/assets/js/src/newsletters/campaign-stats/newsletter-stats-info.tsx
@@ -34,10 +34,7 @@ const redirectToNewsletterHome = () => {
 
 const getEditorLink = (newsletter: NewsletterType) => {
   let editorHref = `?page=mailpoet-newsletter-editor&id=${newsletter.id}`;
-  if (
-    MailPoet.FeaturesController.isSupported('gutenberg_email_editor') &&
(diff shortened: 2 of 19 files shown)


mailpoet/assets/js/src/newsletters/types.tsx
```diff
@@ -36,9 +36,7 @@ export function NewsletterTypes({
   const [isCreating, setIsCreating] = useState(false);
 
   const [isSelectEditorModalOpen, setIsSelectEditorModalOpen] = useState(false);
-  const isNewEmailEditorEnabled =
-    MailPoet.FeaturesController.isSupported('gutenberg_email_editor') &&
-    window.mailpoet_block_email_editor_enabled;
+  const isNewEmailEditorEnabled = window.mailpoet_block_email_editor_enabled;
```
lysyjan, line 39: I think this can be removed, and then we can also remove the window property `mailpoet_block_email_editor_enabled`. It is not critical. Or is there some reason to keep it?
From what I have checked, it only affects rendering a button, which should not be conditional anymore.

lysyjan (commented):
@triple0t Good job.
I have noticed only one tiny detail.

mailpoet/assets/js/src/newsletters/types.tsx
triple0t: We use the conditional to check if we can show the email editor button. The [variable](https://github.com/mailpoet/mailpoet/blob/1103e5fe49cf8f959139dced5c55e5409b2ac747/mailpoet/lib/AdminPages/Pages/Newsletters.php#L173) is set (or disabled) when the email editor [requirements aren't met](https://github.com/mailpoet/mailpoet/blob/1103e5fe49cf8f959139dced5c55e5409b2ac747/packages/php/email-editor/src/Engine/class-dependency-check.php#L23-L28).

lysyjan: Ah. This is something that was added or changed. I guess we didn't initially check the requirement, and it depended only on the feature flag. Thanks for the explanation 👍

lysyjan (approved):

PR #6030, opened 2025-01-08
Enable undo/redo keyboard shortcuts [MAILPOET-6362]

pavel-mailpoet:
## Description

_N/A_

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6362]

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:undo-shortcut)

_The latest successful build from `undo-shortcut` will be used. If none is available, the link won't work._

[MAILPOET-6362]: https://mailpoet.atlassian.net/browse/MAILPOET-6362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/js/email-editor/src/components/keybord-shortcuts/index.tsx
+++ b/packages/js/email-editor/src/components/keybord-shortcuts/index.tsx
@@ -1,5 +1,6 @@
 import { useSelect, useDispatch } from '@wordpress/data';
 import { useEffect } from '@wordpress/element';
+import { store as coreDataStore } from '@wordpress/core-data';
 import {
 	useShortcut,
 	store as keyboardShortcutsStore,
@@ -12,6 +13,7 @@ import { storeName } from '../../store';
 //    https://github.com/WordPress/gutenberg/blob/0ee78b1bbe9c6f3e6df99f3b967132fa12bef77d/packages/edit-site/src/components/keyboard-shortcuts/index.js
 
 export function KeyboardShortcuts(): null {
+	const { undo: undoAction, redo: redoAction } = useDispatch( coreDataStore );
 	const { isSidebarOpened, hasEdits, isSaving } = useSelect( ( select ) => ( {
 		isSidebarOpened: select( storeName ).isSidebarOpened(),
 		isSaving: select( storeName ).isSaving(),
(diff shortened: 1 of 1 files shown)


costasovo (commented):
@pavel-mailpoet We have [a file](https://github.com/mailpoet/mailpoet/blob/1c3ea9cd0a5fc8848a64d840e2fa16a6c7d8c1fe/packages/js/email-editor/src/components/keybord-shortcuts/index.tsx#L14-L13) where we place all keyboard shortcuts. The approach is similar to the Gutenberg. Could you move the shortcuts to that file? I think it is nice to have them in one place. Or do you prefer keeping the code in the Header component?

pavel-mailpoet:
> We have [a file](https://github.com/mailpoet/mailpoet/blob/1c3ea9cd0a5fc8848a64d840e2fa16a6c7d8c1fe/packages/js/email-editor/src/components/keybord-shortcuts/index.tsx#L14-L13) where we place all keyboard shortcuts. The approach is similar to the Gutenberg. Could you move the shortcuts to that file? I think it is nice to have them in one place. Or do you prefer keeping the code in the Header component?

I am aware of that, and I was thinking about where to put the code. I don't think there is a right or wrong approach to this. This is my reasoning why it makes sense to keep the code in the Header component:

- We use a similar approach in the [form editor](https://github.com/mailpoet/mailpoet/blob/aa74dfc0e42ba95b2fa74dd0297f9d6bd38f6ac4/mailpoet/assets/js/src/form-editor/components/history-undo.tsx#L21)
- A component should contain all the parts. When you look at a component, you should see its code, CSS, shortcuts and everything. You don't need to search for it everywhere. When you use the component, in this case, Header, you don't need to think about registering everything else, the component brings everything.

costasovo:
> I am aware of that and was thinking about where to put the code. I don't think there is a right or wrong approach to this. ?This is my reasoning why it makes sense to keep the code in the Header component:
> - We use a similar approach in the [form editor](https://github.com/mailpoet/mailpoet/blob/aa74dfc0e42ba95b2fa74dd0297f9d6bd38f6ac4/mailpoet/assets/js/src/form-editor/components/history-undo.tsx#L21)
> - A component should contain all the parts. When you look at a component, you should see its code, CSS, shortcuts and everything. You don't need to search for it everywhere. When you use the component, in this case, Header, you don't need to think about registering everything else, the component brings everything.

@pavel-mailpoet The reasoning about the component makes sense, but in this particular case, I would not apply it. I see the undo/redo functionality as decoupled from the component. IMO, it should be possible to remove/hide the component, and undo/redo should keep working via shortcodes. Or we may theoretically want to add multiple other buttons in various other components that could trigger undo/redo. Another argument why I would put it into the file is consistency with other shortcut definitions in this project and with the approach used in Gutenberg packages.

costasovo (approved):
I tested it locally and it works as expected

PR #6037, opened 2025-01-09
Remove default align center from group block [MAILPOET-6412]

lysyjan:
## Description

Because the group block has deactivated layout settings in the email editor, content alignment cannot be configured. The default value center could break blocks inside the group. 

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6412]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6412]: https://mailpoet.atlassian.net/browse/MAILPOET-6412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:list-align-fix)

_The latest successful build from `list-align-fix` will be used. If none is available, the link won't work._

### DIFF
--- a/packages/php/email-editor/src/Integrations/Core/Renderer/Blocks/class-group.php
+++ b/packages/php/email-editor/src/Integrations/Core/Renderer/Blocks/class-group.php
@@ -85,14 +85,13 @@ class Group extends Abstract_Block_Renderer {
 		)['declarations'];
 
 		$table_styles['background-size'] = empty( $table_styles['background-size'] ) ? 'cover' : $table_styles['background-size'];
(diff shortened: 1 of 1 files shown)


costasovo (approved):
@lysyjan Thank you for the fix! We can remove it if we don't support it. In case we reintroduce the alignment setting it should be easy to add back.

PR #6051, opened 2025-01-16
Fix misaligned opt-in checkbox on block checkout [MAILPOET-6436]

NeosinneR:
## Description

Fixes opt-in checkbox's label not vertically centered with the checkbox. 

Captura de pantalla de 2025-01-14 16-43-31

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6436]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6436]: https://mailpoet.atlassian.net/browse/MAILPOET-6436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:block-checkout-optin-alignment)

_The latest successful build from `block-checkout-optin-alignment` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/assets/js/src/global.d.ts
+++ b/mailpoet/assets/js/src/global.d.ts
@@ -39,13 +39,13 @@ interface JQuery {
 declare module '@woocommerce/blocks-checkout' {
   type CheckboxControlProps = {
     className?: string;
-    label?: string;
+    label?: string | React.ReactNode;
     id?: string;
-    instanceId?: string;
-    onChange?: (value: boolean) => void;
-    children?: React.ReactChildren | React.ReactElement;
+    onChange: (value: boolean) => void;
+    children?: React.ReactChildren;
     hasError?: boolean;
     checked?: boolean;
+    disabled?: string | boolean | undefined;
   };
   export const CheckboxControl: (props: CheckboxControlProps) => JSX.Element;
 }
--- a/mailpoet/assets/js/src/marketing-optin-block/edit.tsx
+++ b/mailpoet/assets/js/src/marketing-optin-block/edit.tsx
@@ -56,7 +56,11 @@ export function Edit({
     <div {...blockProps}>
       {optinEnabled ? (
         <div className="wc-block-checkout__marketing">
-          <CheckboxControl id="mailpoet-marketing-optin" checked={false} />
+          <CheckboxControl
+            id="mailpoet-marketing-optin"
+            checked={false}
+            onChange={() => {}}
+          />
           <RichText
             value={currentText}
             onChange={(value) => setAttributes({ text: value })}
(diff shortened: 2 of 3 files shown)


NeosinneR:
Just a ping, @restebanez, as this is waiting for your review :)

mailpoet/assets/js/src/marketing-optin-block/edit.tsx
```diff
@@ -56,7 +56,11 @@ export function Edit({
     <div {...blockProps}>
       {optinEnabled ? (
         <div className="wc-block-checkout__marketing">
-          <CheckboxControl id="mailpoet-marketing-optin" checked={false} />
+          <CheckboxControl
```
restebanez, line 59: This is not a code issue. I'm curious to know whether the checkbox is always empty at editing time?

mailpoet/assets/js/src/global.d.ts
```diff
@@ -39,13 +39,13 @@ interface JQuery {
...
+    label?: string | React.ReactNode;
     id?: string;
-    instanceId?: string;
-    onChange?: (value: boolean) => void;
-    children?: React.ReactChildren | React.ReactElement;
+    onChange: (value: boolean) => void;
+    children?: React.ReactChildren;
     hasError?: boolean;
     checked?: boolean;
+    disabled?: string | boolean | undefined;
```
restebanez, line 48: I'm kind of new to the code. Is  a good practice to use "undefined" as a type? Would there be an option to initialize the disabled property in the code instead?

restebanez (approved):
I have just added two questions that are more related to my lack of knowledge :)

I'm sorry about the long wait. I didn't notice the PR before.

mailpoet/assets/js/src/marketing-optin-block/edit.tsx
NeosinneR: Yes, it's always unchecked by default. The reason is that this checkbox can only subscribe you but can't unsubscribe you. If the customer is already subscribed, but won't check this checkbox, nothing will happen. This is to prevent unwanted unsubscriptions. 

But if you want to subscribe, you need to make an explicit action of checking the box.

mailpoet/assets/js/src/global.d.ts
NeosinneR: This is an imported component, so I only copied types from its source code at https://github.com/woocommerce/woocommerce/blob/644c13b828e30a0fc0c7894b1960f8f38094299c/plugins/woocommerce-blocks/packages/components/checkbox-control/index.tsx#L23 - I don't know the reasoning to use `undefined` but from maintenance point of view, it's easier to use the same types, rather than trying to remove `undefined` and handle it on our side.

Aschepikov:
@NeosinneR Hi there are some failed checks can you push the master pls

NeosinneR:
@Aschepikov rebased

Aschepikov (approved):
GJ!

PR #6053, opened 2025-01-16
Email editor - add @woocommerce/eslint-plugin [MAILPOET-6438]

costasovo:
## Description

This PR adds @woocommerce/eslint-plugin and fixes issues reported after adding the plugin.
The motivation was adding a check that would enforce comments and grouping of imports. The [@woocommerce/eslint-plugin](https://github.com/woocommerce/woocommerce/tree/trunk/packages/js/eslint-plugin) contains [a rule for checking dependency groups](https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/eslint-plugin/docs/rules/dependency-group.md) and is used by most of the packages in the Woo monorepo. As we plan to move the package to the monorepo, I think it makes sense to start using the whole package.

- When fixing the groups in the email editor codebase, I found that the rule for grouping reports an error when import contains a comment. We have a couple of comments in imports, mostly to suppress TS warnings. I worked around the issues by adding eslint ignore comments
- Apart from the import groups, I had to fix a couple of missing domains in translation functions and add comments to suppressed TS errors.
- I chose to disable @wordpress/no-unsafe-wp-apis check so that we don't need to use eslint ignore when we import an export with a name prefixed by __. This helps a bit with the problem with the import group check because we can remove some comments from imports. I think it is relatively safe to do because Gutenberg now uses private API instead of legacy unsafe exports, and most of these exports have remained unchanged for a long time.
- We can also disable `react/react-in-jsx-scope` because we use wordpress/scripts that [use the new JSX Transform](https://github.com/WordPress/gutenberg/pull/61692) and [it is recommended to disable it](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#eslint) for the new transform

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6438]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6438]: https://mailpoet.atlassian.net/browse/MAILPOET-6438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:email-editor/woo-eslint-plugin)

_The latest successful build from `email-editor/woo-eslint-plugin` will be used. If none is available, the link won't work._

### DIFF
--- a/packages/js/email-editor/.eslintrc.js
+++ b/packages/js/email-editor/.eslintrc.js
@@ -5,6 +5,7 @@ module.exports = {
 			files: [ '**/*.js', '**/*.ts', '**/*.jsx', '**/*.tsx' ],
 			rules: {
 				'react/react-in-jsx-scope': 'off',
+				'@wordpress/no-unsafe-wp-apis': 'off',
 				'@wordpress/i18n-text-domain': [
 					'error',
 					{
--- a/packages/js/email-editor/.eslintrc.js
+++ b/packages/js/email-editor/.eslintrc.js
@@ -1,0 +1,11 @@
+module.exports = {
+	extends: [ 'plugin:@woocommerce/eslint-plugin/recommended' ],
+	overrides: [
+		{
+			files: [ '**/*.js', '**/*.ts', '**/*.jsx', '**/*.tsx' ],
+			rules: {
+				'react/react-in-jsx-scope': 'off',
+			},
+		},
+	],
+};
--- a/packages/js/email-editor/.eslintrc.js
+++ b/packages/js/email-editor/.eslintrc.js
@@ -5,6 +5,12 @@ module.exports = {
 			files: [ '**/*.js', '**/*.ts', '**/*.jsx', '**/*.tsx' ],
 			rules: {
 				'react/react-in-jsx-scope': 'off',
+				'@wordpress/i18n-text-domain': [
+					'error',
+					{
+						allowedTextDomain: [ 'mailpoet' ],
+					},
+				],
 			},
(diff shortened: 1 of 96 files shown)


triple0t (commented):
I tested it and can confirm it works well.
Well done.

Quick question: Do you think it would be possible to support `pnpm run lint:js-fix` in fixing the eslint error message?

I can see the error with `pnpm run lint:js`, but I can't automatically fix it with `pnpm run lint:js-fix`

triple0t (approved):

costasovo:
> Quick question: Do you think it would be possible to support pnpm run lint:js-fix in fixing the eslint error message?
 > I can see the error with pnpm run lint:js, but I can't automatically fix it with pnpm run lint:js-fix

 @triple0t I'm not super familiar with the eslint rules, but I found that the auto-fix functionality must be implemented within a rule. I found in the docs that [the rule needs to provide a fix function](https://eslint.org/docs/latest/extend/custom-rules#applying-fixes). The [dependency-group rule doesn't provide the fix function](https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/eslint-plugin/rules/dependency-group.js#L149-L152), so it doesn't provide the automatic fix functionality. If we wanted it, we would need to add it to the rule.

PR #6055, opened 2025-01-16
Branch out of AbandonedCart jobs earlier

oz:
## Description

If the AbandonedCart automation is inactive or there are no subscribers matching when the cron-worker schedules this job, the worker can consider its job done, and return true earlier.

As I understand it, if `false` is returned, then the cron-worker will enqueue more similar jobs, clogging the queue.

See p1735156774223359-slack-C01H71ZLBGQ.

## Code review notes

_N/A_

## QA notes

Testing instructions:
1. Create an abandoned cart automation.
2. Put something in a cart to trigger the automation.
3. Disable the automation.
4. Wait until the scheduled task for the abandoned cart is triggered and see that it's marked as completed and not iterated over and over again.
5. You can also try to repeat the above steps, but delete the customer from the `mailpoet_subscribers` table instead of disabling the automation.

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6435]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6435]: https://mailpoet.atlassian.net/browse/MAILPOET-6435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:MAILPOET-6435_stuck-abandoned-cart)

_The latest successful build from `MAILPOET-6435_stuck-abandoned-cart` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/Cron/Workers/Automations/AbandonedCartWorker.php
+++ b/mailpoet/lib/Cron/Workers/Automations/AbandonedCartWorker.php
@@ -45,16 +45,16 @@ class AbandonedCartWorker extends SimpleWorker {
 
     $subscribers = $task->getSubscribers();
     if ($subscribers->count() !== 1) {
-      return false;
+      return true;
     }
     $subscriber = isset($subscribers[0]) ? $subscribers[0]->getSubscriber() : null;
     if (!$subscriber) {
-      return false;
+      return true;
     }
 
     $automation = $this->automationStorage->getAutomation((int)$automationId, (int)$automationVersion);
     if (!$automation || $automation->getStatus() !== Automation::STATUS_ACTIVE) {
-      return false;
+      return true;
     }
 
     $this->wp->doAction(
--- a/mailpoet/tests/integration/Automation/Integrations/WooCommerce/Triggers/AbandonedCartTriggerTest.php
+++ b/mailpoet/tests/integration/Automation/Integrations/WooCommerce/Triggers/AbandonedCartTriggerTest.php
@@ -8,6 +8,7 @@ use MailPoet\Automation\Engine\Data\NextStep;
 use MailPoet\Automation\Engine\Data\Step;
 use MailPoet\Automation\Engine\Data\Subject;
 use MailPoet\Automation\Engine\Storage\AutomationRunStorage;
+use MailPoet\Automation\Engine\Storage\AutomationStorage;
 use MailPoet\Automation\Integrations\WooCommerce\Subjects\AbandonedCartSubject;
 use MailPoet\Automation\Integrations\WooCommerce\Triggers\AbandonedCart\AbandonedCartTrigger;
 use MailPoet\Cron\Workers\Automations\AbandonedCartWorker;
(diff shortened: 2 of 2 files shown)


mailpoet/lib/Cron/Workers/Automations/AbandonedCartWorker.php
```diff
@@ -44,7 +44,10 @@ public function processTaskStrategy(ScheduledTaskEntity $task, $timer) {
     $lastActivityAt = $task->getCreatedAt();
 
     $subscribers = $task->getSubscribers();
-    if ($subscribers->count() !== 1) {
+    if ($subscribers->count() === 0) {
```
alex-mpoet: If a subscriber was deleted directly from the `subscribers` table in the DB (the case originally described in Slack), but the link table entry (`scheduled_task_subscribers`) remains, the count here will still be `1`, so none of the count-checking conditions will match, it will fail at the [subscriber retrieval step](https://github.com/mailpoet/mailpoet/pull/6055/files#diff-647e969e299019061c5a1bc914b5e61750cc04420bf386002b42c1bb00109c66R53) where the check still returns `false`.

oz: I've updated the code to also look at `scheduled_task_subscribers`. The return `true` when this table's records were deleted. This should fix the matching issue.

mailpoet/lib/Cron/Workers/Automations/AbandonedCartWorker.php
```diff
@@ -44,17 +44,20 @@ public function processTaskStrategy(ScheduledTaskEntity $task, $timer) {
...
+      return true;
+    }
+    if ($subscribers->count() > 1) {
```
alex-mpoet: @oz I'm not sure what's the purpose of this condition. Abandoned cart automations work with a single subscriber, so having more than one subscriber unless it's a bug is not normal. And this condition returns false, so the task will be iterated once again until extra subscribers are removed. Could you please elaborate?

oz: @alex-mpoet I'm keeping the spirit of the original code which branched out if subscriber count was `!== 1`. If you think this is now useless, I can remove it.

alex-mpoet: @oz Yes, the condition replicates the old one, but the return value is now different between the two. I would just keep the old `!== 1` and return `true` so it skips the task in both cases.

oz: Let's do that then.

alex-mpoet (approved):
@oz Looks good! Next time please consider adding tests (I have added them to this PR as an example) and QA instructions (I've also added an example). QA instructions can be skipped if the change is trivial to test or instructions are in the Jira ticket. If QA is not needed, please specify so explicitly.

Aschepikov (approved):

PR #6068, opened 2025-01-23
Make the email preview page independent of the MailPoet plugin - MAILPOET-6323

triple0t:
## Description

Make the email preview page independent of the MailPoet plugin

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6323](https://mailpoet.atlassian.net/browse/MAILPOET-6323)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6323]: https://mailpoet.atlassian.net/browse/MAILPOET-6323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:make-email-preview-page-independent-on-mailpoet-plugin)

_The latest successful build from `make-email-preview-page-independent-on-mailpoet-plugin` will be used. If none is available, the link won't work._


packages/php/email-editor/src/Engine/class-email-editor.php
```diff
@@ -282,16 +282,18 @@ public function load_email_preview_template( string $template ): string {
...
-			 * theme or child theme directories, so load it
-			 * from our plugin directory.
-			 */
-			return __DIR__ . '/Templates/single-post-template.php';
+		if ( ! in_array( $current_post_type, $email_post_types, true ) ) {
+			return $template;
 		}
 
-		return $template;
+		add_filter(
```
lysyjan, line 289: This new filter seems a bit hidden in this place. I would prefer having it in the `initialize` method like the others. What do you think?

lysyjan (commented):
It looks good. I only have a question about the new filter.

packages/php/email-editor/src/Engine/class-email-editor.php
triple0t: This is because we need to use the [$post](https://github.com/mailpoet/mailpoet/blob/c15c4e67514f000d6c0ba78489aef9a11f87cc6e/packages/php/email-editor/src/Engine/class-email-editor.php#L279) variable available in this [scope](https://github.com/mailpoet/mailpoet/blob/c15c4e67514f000d6c0ba78489aef9a11f87cc6e/packages/php/email-editor/src/Engine/class-email-editor.php#L279-L295). 

I discovered a weird bug when testing with some plugins. The post from the [apply_filter](https://github.com/mailpoet/mailpoet/blob/c15c4e67514f000d6c0ba78489aef9a11f87cc6e/packages/php/email-editor/src/Engine/Templates/single-email-post-template.php#L10) gets overwritten. 

Also, we sometimes experience weird inconsistencies with some platforms (e.g., Atomic), so it's safer to use the variable here.

lysyjan (approved):

PR #6070, opened 2025-01-24
Remove the use of MailPoet Vendor libraries [MAILPOET-6428]

pavel-mailpoet:
## Description

The email editor package used two libraries from `MailPoetVendor` namespace:
- Emogrifier\CssInliner
- Soundasleep\Html2Text

This pull request removes the dependency on MailPoet vendor.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6428]

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6428]: https://mailpoet.atlassian.net/browse/MAILPOET-6428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:remove-mailpoet-vendor)

_The latest successful build from `remove-mailpoet-vendor` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/build.sh
+++ b/mailpoet/build.sh
@@ -43,6 +43,9 @@ if [ -d 'vendor-prefixed' ]; then
 	mv vendor-prefixed vendor-prefixed-backup
 fi
 
+echo '[BUILD] Install email editor dependencies'
+cd ../packages/php/email-editor && ../../../mailpoet/tools/vendor/composer.phar install --no-dev --prefer-dist --optimize-autoloader --no-scripts && cd -
+
 # Production libraries.
 echo '[BUILD] Fetching production libraries'
 mkdir vendor-prefixed
--- a/mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
+++ b/mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
@@ -8,13 +8,13 @@ use MailPoetVendor\Pelago\Emogrifier\CssInliner;
 class MailPoetCssInliner implements Css_Inliner {
   private CssInliner $inliner;
 
-  public function from_html(string $unprocessed_html): self {// phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+  public function from_html(string $unprocessed_html): self {// phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps -- we need to match the interface
     $that = new self();
     $that->inliner = CssInliner::fromHtml($unprocessed_html);
     return $that;
   }
 
-  public function inline_css(string $css = ''): self {// phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+  public function inline_css(string $css = ''): self {// phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps -- we need to match the interface
     $this->inliner->inlineCss($css);
     return $this;
   }
--- a/mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
+++ b/mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
@@ -15,11 +15,17 @@ class MailPoetCssInliner implements Css_Inliner {
   }
 
   public function inline_css(string $css = ''): self {// phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps -- we need to match the interface
+    if (!isset($this->inliner)) {
+      throw new \LogicException('You must call from_html before calling inline_css');
+    }
     $this->inliner->inlineCss($css);
     return $this;
   }
 
   public function render(): string {
+    if (!isset($this->inliner)) {
+      throw new \LogicException('You must call from_html before calling inline_css');
+    }
     return $this->inliner->render();
   }
 }
--- a/mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
+++ b/mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
@@ -1,0 +1,25 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\EmailEditor\Integrations\MailPoet;
+
+use MailPoet\EmailEditor\Engine\Renderer\Css_Inliner;
+use MailPoetVendor\Pelago\Emogrifier\CssInliner;
+
+class MailPoetCssInliner implements Css_Inliner {
+  private CssInliner $inliner;
+
+  public function from_html(string $unprocessed_html): self {// phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    $that = new self();
+    $that->inliner = CssInliner::fromHtml($unprocessed_html);
+    return $that;
+  }
+
+  public function inline_css(string $css = ''): self {// phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    $this->inliner->inlineCss($css);
+    return $this;
+  }
+
+  public function render(): string {
+    return $this->inliner->render();
+  }
+}
--- a/packages/php/email-editor/composer.json
+++ b/packages/php/email-editor/composer.json
@@ -16,7 +16,8 @@
     ]
   },
   "require": {
-    "php": ">=7.4"
+    "php": ">=7.4",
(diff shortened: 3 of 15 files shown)


mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace MailPoet\EmailEditor\Integrations\MailPoet;
+
+use MailPoet\EmailEditor\Engine\Renderer\Css_Inliner;
+use MailPoetVendor\Pelago\Emogrifier\CssInliner;
+
+class MailPoetCssInliner implements Css_Inliner {
+  private CssInliner $inliner;
+
+  public function from_html(string $unprocessed_html): self {// phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
```
lysyjan: I can see in this implementation a possible danger when a method `inline_css` or `render` can be used before calling `from_html`. In that case, there is a possible calling method on `null`. In my opinion, we should avoid it. What do you think?

packages/php/email-editor/tests/integration/_bootstrap.php
```diff
@@ -49,6 +49,8 @@
 $console->writeln( 'Loading WP core... (' . $wp_load_file . ')' );
 require_once $wp_load_file;
 
+require_once __DIR__ . '/../../../../../mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php';
```
lysyjan, line 53: Tests are dependent on the MailPoet CSS Inliner implementation, which we should avoid. If we postpone it, we must challenge it later when we move this package into another repository.
I was thinking about it, and we can install Emogrifier as a dev dependency and use our simple implementation in tests. Do you think it could work?

packages/php/email-editor/composer.json
```diff
@@ -16,7 +16,8 @@
     ]
   },
   "require": {
-    "php": ">=7.4"
+    "php": ">=7.4",
+    "soundasleep/html2text": "^2.1"
```
lysyjan, line 20: You added a new dependency, but I am not sure if we install those dependencies during the build. What I have tested, after deleting installed dependencies from this PHP package, it was not installed during the build preparation. Please, can you check that?

lysyjan (commented):
I like your idea with the interface.
I have noticed a few things that I would like to discuss. Please check the comments.

mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
pavel-mailpoet: Good point, what do you suggest the action should be if the method `render` is called but the `$inliner` is null?

packages/php/email-editor/tests/integration/_bootstrap.php
pavel-mailpoet: My thinking was that we are testing the whole integration here. Yes, the tests are dependent on MailPoet. However, once we move the package, we will still need to implement this class for WooCommerce. If we forget, this line will remind us to write the new implementation and update it here.

mailpoet/lib/EmailEditor/Integrations/MailPoet/MailPoetCssInliner.php
lysyjan: I see two approaches here.
1. Throwing a logical exception when the property is not set yet
2. Using a constructor, which uses CssInliner::from_html(). But this could complicate using DI.

I would prefer throwing an exception. What do you think?

packages/php/email-editor/composer.json
pavel-mailpoet: I am sorry, but I am unsure what you are referring to. Could you please point me to the code that does the build and where I should orchestrate the deps installation?

packages/php/email-editor/tests/integration/_bootstrap.php
lysyjan: We agreed on a call to keep it as it is because we plan to refactor tests during the next project phase.

packages/php/email-editor/composer.json
lysyjan: I think you should focus on the command `./do install` and the file `build.sh`. Those are places where I would expect changes.

mailpoet/build.sh
```diff
@@ -52,6 +52,9 @@ COMPOSER_MIRROR_PATH_REPOS=1 ./tools/vendor/composer.phar install --no-dev --pre
 echo '[BUILD] Fetching prefixed production libraries'
 ./tools/vendor/composer.phar install --no-dev --prefer-dist --working-dir=./prefixer/
 
+echo '[BUILD] Install email editor dependencies'
+cd ../packages/php/email-editor && ../../../mailpoet/tools/vendor/composer.phar install --no-dev --prefer-dist --optimize-autoloader --no-scripts && cd -
```
lysyjan: The package dependencies are installed too late. I think they should be installed before the dependencies in line 50. I tried to check the build from CI, and the vendor directory for the `mailpoet` package was missing in the vendor.

lysyjan (dismissed):
LGTM

packages/php/email-editor/src/Engine/Renderer/class-renderer.php
```diff
@@ -12,9 +12,7 @@
 use MailPoet\EmailEditor\Engine\Templates\Templates;
 use MailPoet\EmailEditor\Engine\Theme_Controller;
 use MailPoetVendor\Html2Text\Html2Text;
```
triple0t: I think you missed this line.
I can see MailPoet Vendor use [here](https://github.com/mailpoet/mailpoet/blob/4122904d4beacc1f98db828b29b292bac3feab9b/packages/php/email-editor/src/Engine/Renderer/class-renderer.php#L14) for Html2Text

I also noticed this warning message in the logs

```txt
PHP Deprecated:  mb_convert_encoding(): Handling HTML entities via mbstring is deprecated; use htmlspecialchars, htmlentities, or mb_encode_numericentity/mb_decode_numericentity instead in packages/php/email-editor/vendor/soundasleep/html2text/src/Html2Text.php on line 48
```

triple0t (commented):
Sorry for jumping in, @pavel-mailpoet 
I noticed something while checking the PR and wanted to inform you.

Also, MailPoet seems to [use](https://github.com/mailpoet/mailpoet/blob/f20237ab1cfe76d90e3a331603dbe36a6a16aab8/mailpoet/prefixer/composer.json#L3-L6) its version of the library ([Repo here](https://github.com/mailpoet/html2text)). Do you know why?

pavel-mailpoet:
Thank you for pointing that out, I will take a look at your comment

> Also, MailPoet seems to [use](https://github.com/mailpoet/mailpoet/blob/f20237ab1cfe76d90e3a331603dbe36a6a16aab8/mailpoet/prefixer/composer.json#L3-L6) its version of the library ([Repo here](https://github.com/mailpoet/html2text)). Do you know why?

I saw that. I don't remember why we are using it. This is quite old, and the commits are also pretty old. The MailPoet fork is outdated and out of sync with the original repo, which contains some fixes. I want to avoid using the outdated fork so I used the original library. I couldn't find anything that would prevent us from using it. If we find during the testing in the future some issues we can address them, but right now, I am not aware.

packages/php/email-editor/src/Engine/Renderer/class-renderer.php
pavel-mailpoet: Thank you for pointing that out, I tried the PHPStorm git client and this change was somehow skipped

@triple0t For the error, I found a similar and they say it should be fixed: https://github.com/soundasleep/html2text/issues/102

I replaced the library here and now it shouldn't use the one from the MailPoet Vendor, can you please check if you still see the error?

triple0t: Thank you for working on this, @pavel-mailpoet 
I tested and I can't find the error any more.
But, I can still see MailPoet Vendor use [here](https://github.com/mailpoet/mailpoet/blob/c0f3657c1290c57d973843d0b20f0ab3b7898e35/packages/php/email-editor/src/Engine/Renderer/class-renderer.php#L14) for Html2Text

triple0t (approved):

PR #6072, opened 2025-01-24
MAILPOET-6447 GATracking escapes shortcodes

oz:
> [!IMPORTANT]
> Don't merge this. 😉

## Description

When Google Analytics tracking is active, and a user adds link to their own domain featuring shortcodes ([example on Zendesk](https://a8c.zendesk.com/agent/tickets/9286619)), our GATracking component can change these links by URL-encoding query parameters. This breaks links with shortcodes in newsletters.

For example, after adding tracking parameters, this URL `https://example.com.com/hi/?u=[subscriber:cf_10]` becomes `https://example.com.com/hi/?u=%5Bsubscriber%3Acf_10%5D&utm_source=mailpoet&utm_m...`

The `u=` value is encoded before the shortcode is expanded ; making the resulting link useless in lots of cases.

## Can we decode these params after adding tracking links?

This is what's done here. It's rather exploratory, but I wanted to verify whether decoding the URL *after* adding Google's tracking parameters would work. It does, kind of… but it introduces a different a bug by giving up on correct URL encoding. So, nope. 😏

It's easy to see that a subscriber could inject arbitrary content a break the URLs. For example, a with a first name like `M&M’s`, if we insert a link with a shortcode like `[subscriber:firstname]`, we could end up with a broken redirection URL like `http://site/?u=M&amp;M&#039;s&utm_source=mailpoet&...` 💣 

## Can we *not* encode these params?

From [a73e614](https://github.com/mailpoet/mailpoet/pull/6072/commits/a73e614dc0a81822452044aa0126ea136a1e9b57), we see that WP's `add_query_arg` encodes our URL, and I think it's fine as it prevents unencoded query params in the emails (which can easily result in invalid links as we've seen).

We rely on `add_query_arg` to add 3 query parameters to our URL, so it looks like maybe we could do without it, somehow... It's a battle tested utility function though, with lots of edge cases, so rewriting without it seems less safe.

## Can we expand shortcodes before adding tracking?

It seems like the most robust solution IMHO, but it requires a few changes to how we currently generate newsletters, and add our own tracking links.

### DIFF
--- a/mailpoet/lib/Statistics/GATracking.php
+++ b/mailpoet/lib/Statistics/GATracking.php
@@ -95,7 +95,7 @@ class GATracking {
 
       $processedLink = $this->wp->applyFilters(
         'mailpoet_ga_tracking_link',
-        $this->wp->addQueryArg($linkParams, $link),
+        urldecode($this->wp->addQueryArg($linkParams, $link)),
         $extractedLink['link'],
         $linkParams,
         $extractedLink['type']
--- a/mailpoet/lib/Statistics/GATracking.php
+++ b/mailpoet/lib/Statistics/GATracking.php
@@ -21,7 +21,7 @@ class GATracking {
   private $wp;
 
   /** @var TrackingConfig */
-  private $tackingConfig;
+  private $trackingConfig;
 
   public function __construct(
     NewsletterLinks $newsletterLinks,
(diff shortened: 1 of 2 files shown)


oz:
Let's close this while MailPoet decides whether to work on a proper fix. 🙅🏻

PR #6077, opened 2025-01-28
Refactor and remove the dependency on the mailpoet_email post-type - MAILPOET-6430

triple0t:
## Description

Refactor and remove the dependency on the mailpoet_email post-type

## Code review notes

Please try switching the [MAILPOET_EMAIL_POST_TYPE](https://github.com/mailpoet/mailpoet/blob/94115d0ef2a532bbbb7747e8a9cd6f8b71181cfa/mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailEditor.php#L10) from `mailpoet_email` to another string variable and confirm the editor works.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

[MAILPOET-6430](https://mailpoet.atlassian.net/browse/MAILPOET-6430)

## After-merge notes

_N/A_

## Tasks

[MAILPOET-6430]: https://mailpoet.atlassian.net/browse/MAILPOET-6430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:refactor-and-remove-the-dependency-on-the-mailpoet_email-post-type)

_The latest successful build from `refactor-and-remove-the-dependency-on-the-mailpoet_email-post-type` will be used. If none is available, the link won't work._

### DIFF
--- a/packages/js/email-editor/src/components/header/header.tsx
+++ b/packages/js/email-editor/src/components/header/header.tsx
@@ -27,7 +27,7 @@ import {
  * Internal dependencies
  */
 
-import { storeName } from '../../store';
+import { storeName, editorCurrentPostType } from '../../store';
 import { MoreMenu } from './more-menu';
 import { PreviewDropdown } from '../preview';
 import { SaveEmailButton } from './save-email-button';
@@ -94,8 +94,8 @@ export function Header() {
 	const { validateContent, isInvalid } = useContentValidation();
 
 	const { dirtyEntityRecords } = useEntitiesSavedStatesIsDirty();
-	const hasTemplatesEdits = dirtyEntityRecords.some(
-		( entity ) => entity.name === 'wp_template'
+	const hasNonEmailEdits = dirtyEntityRecords.some(
+		( entity ) => entity.name !== editorCurrentPostType
 	);
 
 	const preventDefault = ( event ) => {
@@ -238,7 +238,7 @@ export function Header() {
 			<div className="editor-header__settings edit-post-header__settings">
 				<SaveEmailButton />
 				<PreviewDropdown />
-				{ hasTemplatesEdits ? (
+				{ hasNonEmailEdits ? (
 					<SaveAllButton />
 				) : (
 					<SendButton
--- a/packages/js/email-editor/src/components/header/header.tsx
+++ b/packages/js/email-editor/src/components/header/header.tsx
@@ -94,8 +94,8 @@ export function Header() {
 	const { validateContent, isInvalid } = useContentValidation();
 
 	const { dirtyEntityRecords } = useEntitiesSavedStatesIsDirty();
-	const hasNonEmailEdits = dirtyEntityRecords.some(
-		( entity ) => entity.name !== 'mailpoet_email'
+	const hasTemplatesEdits = dirtyEntityRecords.some(
+		( entity ) => entity.name === 'wp_template'
 	);
 
 	const preventDefault = ( event ) => {
@@ -238,7 +238,7 @@ export function Header() {
 			<div className="editor-header__settings edit-post-header__settings">
 				<SaveEmailButton />
 				<PreviewDropdown />
-				{ hasNonEmailEdits ? (
+				{ hasTemplatesEdits ? (
 					<SaveAllButton />
 				) : (
 					<SendButton
--- a/packages/js/email-editor/src/components/template-select/select-modal.tsx
+++ b/packages/js/email-editor/src/components/template-select/select-modal.tsx
@@ -16,6 +16,7 @@ import {
 	storeName,
 	TemplateCategory,
 	TemplatePreview,
+	editorCurrentPostType,
 } from '../../store';
 import { TemplateList } from './template-list';
 import { TemplateCategoriesListSidebar } from './template-categories-list-sidebar';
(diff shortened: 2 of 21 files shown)


packages/js/email-editor/src/components/header/header.tsx
```diff
@@ -94,8 +94,8 @@ export function Header() {
 	const { validateContent, isInvalid } = useContentValidation();
 
 	const { dirtyEntityRecords } = useEntitiesSavedStatesIsDirty();
-	const hasNonEmailEdits = dirtyEntityRecords.some(
-		( entity ) => entity.name !== 'mailpoet_email'
+	const hasTemplatesEdits = dirtyEntityRecords.some(
+		( entity ) => entity.name === 'wp_template'
```
costasovo: We also need to check if we edited styles because they also require the Save button. Locally, when I edited the style, the editor ended up in a state that the send button was disabled, but the save button was not displayed.
Screenshot 2025-01-29 at 14 11 36
In the future we may also add template parts for header and footer so there might be more entities to save so I think it would be safer to stick with the `hasNonEmailEdits` logic if that's possible.
What do you think?

packages/js/email-editor/src/components/template-select/select-modal.tsx
```diff
@@ -59,7 +59,7 @@ function SelectTemplateBody( {
 				setSelectedCategory( TemplateCategories[ 0 ].name );
 			}
 		}, 1000 ); // using setTimeout to ensure the template styles are available before block preview
-	}, [ hasEmailPosts ] );
+	}, [ hasEmailPosts, hideRecentCategory ] );
```
costasovo, line 63: Nice catch!

packages/php/email-editor/src/Engine/class-email-editor.php
```diff
@@ -267,4 +268,28 @@ public function extend_email_theme_styles( WP_Theme_JSON $theme, WP_Post $post )
...
+		if ( ! $current_post_is_email_editor_type ) {
+			throw new \InvalidArgumentException( esc_html__( 'Invalid email post type', 'mailpoet' ) );
+		}
+
+		?>
+		<script type="text/javascript"> <?php // phpcs:ignore ?>
+			window.mailpoet_email_editor_current_post_type = <?php echo wp_json_encode( $email_editor_current_post_type ); ?>;
+		</script>
+		<?php
+	}
```
costasovo: Just a thought. We currently ask the integrator to provide a couple of [variables that are essential for the editor](https://github.com/mailpoet/mailpoet/blob/3d45ea92e2097c9211e735830f9ac667ae7d1393/mailpoet/lib/EmailEditor/Integrations/MailPoet/EditorPageRenderer.php#L151-L153). I feel that `mailpoet_email_editor_current_post_type` is very similar in that regard. It would be nicer if the editor package could prepare more itself. But, as we already have some variables there, I think that as for now we could add the current post type among them. Perhaps later, when we focus on simplifying bootstrapping the editor, we could move that somehow to the package. I'm suggesting that for the sake of having them all in one place.

packages/php/email-editor/src/Engine/class-email-editor.php
```diff
@@ -288,7 +288,8 @@ public function load_js_vars() {
 
 		?>
 		<script type="text/javascript"> <?php // phpcs:ignore ?>
-			window.mailpoet_email_editor_current_post_type = <?php echo wp_json_encode( $email_editor_current_post_type ); ?>;
+			window.mailpoet_email_editor_current_post_type = '<?php echo esc_js( $email_editor_current_post_type ); ?>';
+			window.mailpoet_email_editor_current_post_id = <?php echo esc_js( $post->ID ); ?>;
```
costasovo: I'm experiencing a weird issue related to this change. I already have a bunch of emails in the DB. When I create a new email in admin UI the editor loads an older email instead of the new one.
I noticed that the postId in the URL differs from the one we passed to the editor.
Screenshot 2025-01-29 at 14 39 58

Is the global $post variable reliable to determine the currently edited post? Perhaps we could move that to the integrator code as I suggested for the other variable where we know the id.

costasovo (commented):
@triple0t I'm happy to see these changes! 👍 I ran into a couple of issues I think we need to fix. Please see my other comments.

packages/js/email-editor/src/components/header/header.tsx
triple0t: Nice catch. Fixed in https://github.com/mailpoet/mailpoet/pull/6077/commits/9b01b840f54d0598968dfa4723739fbb7556b109

packages/php/email-editor/src/Engine/class-email-editor.php
triple0t: Hmm, Okay. Done in https://github.com/mailpoet/mailpoet/pull/6077/commits/12344da42842a10c6e08c007e0e1e414d8d4241d

packages/php/email-editor/src/Engine/class-email-editor.php
triple0t: Done here https://github.com/mailpoet/mailpoet/pull/6077/commits/12344da42842a10c6e08c007e0e1e414d8d4241d

triple0t:
Thank you for the in-depth review, @costasovo.
I've updated the PR.

costasovo (approved):
@triple0t Thanks for the refactor and fixes. Good job with noticing unnecessary code and the cleanup!
I tested with a different MAILPOET_EMAIL_POST_TYPE and all seem to work as expected! 

Please feel free to merge or send it to QA.

PR #6079, opened 2025-01-30
Moving fetching iframed assets into get_settings method

lysyjan:
## Description

Because we called the function for getting editor iframe assets in the nit hook, it could cause some properties to not be set yet. This PR moves to get assets into the `get_settings` method to avoid getting a property on null.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:iframe-assets-fix)

_The latest successful build from `iframe-assets-fix` will be used. If none is available, the link won't work._

### DIFF
--- a/packages/php/email-editor/src/Engine/class-settings-controller.php
+++ b/packages/php/email-editor/src/Engine/class-settings-controller.php
@@ -57,39 +57,13 @@ class Settings_Controller {
 	}
 
 	/**
-	 * Method to initialize the settings controller.
-	 *
-	 * @return void
-	 */
-	public function init(): void {
-		/*
-		 * We need to initialize these assets early because they are read from global variables $wp_styles and $wp_scripts
-		 * and in later WordPress page load pages they contain stuff we don't want (e.g. html for admin login popup)
-		 * in the post editor this is called directly in post.php.
-		 */
-		$this->iframe_assets = _wp_get_iframed_editor_assets();
-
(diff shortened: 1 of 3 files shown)


packages/php/email-editor/src/Engine/class-settings-controller.php
```diff
@@ -215,4 +189,36 @@ public function translate_slug_to_font_size( string $font_size ): string {
...
+		if ( ! empty( $this->iframe_assets ) ) {
+			return;
+		}
+
+		/*
+		 * We need to initialize these assets early because they are read from global variables $wp_styles and $wp_scripts
+		 * and in later WordPress page load pages they contain stuff we don't want (e.g. html for admin login popup)
+		 * in the post editor this is called directly in post.php.
+		 */
+		$this->iframe_assets = _wp_get_iframed_editor_assets();
```
costasovo: I think the comment is now a bit outdated. We no longer make sure it runs early. I think we can skip it, because it can be confusing as is.

costasovo (commented):
@lysyjan Good job! I have one suggestion about an outdated comment. Please see my other comment.

costasovo (approved):
Looks good! Thank you!

PR #6086, opened 2025-02-05
Fix Uncaught TypeError due to a type hint which affects some platforms

triple0t:
## Description

According to the [docs](https://developer.wordpress.org/reference/hooks/type_template/), this should be a string, but using NULL here as an input leads to a TypeError.

See p1738738617227929-slack-C01H71ZLBGQ for more information.

## Code review notes

_N/A_

## QA notes

We can skip QA.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:fix-error-notice)

_The latest successful build from `fix-error-notice` will be used. If none is available, the link won't work._

### DIFF
--- a/packages/php/email-editor/src/Engine/class-email-editor.php
+++ b/packages/php/email-editor/src/Engine/class-email-editor.php
@@ -279,7 +279,7 @@ class Email_Editor {
 	 * @param string $template post template.
 	 * @return string
 	 */
-	public function load_email_preview_template( string $template ): string {
+	public function load_email_preview_template( $template ): string {
 		$post = $this->get_current_post();
 
 		if ( ! $post instanceof \WP_Post ) {
--- a/packages/php/email-editor/src/Engine/class-email-editor.php
+++ b/packages/php/email-editor/src/Engine/class-email-editor.php
@@ -279,7 +279,7 @@ class Email_Editor {
 	 * @param string $template post template.
 	 * @return string
 	 */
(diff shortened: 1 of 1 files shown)


mhadidg (changes requested):
The return type is still `string`, but some paths (example below) just return `$template` as-is, which could be `null`—and that might throw a TypeError error:

```
if ( ! in_array( $current_post_type, $email_post_types, true ) ) {
	return $template;
}
```

I'd suggest either removing the return type definition or switching to the nullable version `?string` for both. Your call.

I'm wondering why it might be `null` in the first place? I don’t fully get the context of the code, but I'm just trying to figure out if there's an upstream issue causing this, or if it's normal to have `null` here.

triple0t:
Nice catch, @mhadidg. 
Fixed in https://github.com/mailpoet/mailpoet/pull/6086/commits/13526298a3e3d66a7b2e0d343244a4bb2acc7f9e

Thanks

> I also wonder why it might be null in the first place? I don’t fully get the context of the code, but I'm just trying to figure out if there's an upstream issue causing this, or if it's normal to have null here.

I don't know. The [docs](https://developer.wordpress.org/reference/hooks/type_template/) mentioned a `string` type.

mhadidg (approved):
LGTM :+1:

PR #6092, opened 2025-02-06
[WIP] Time based triggers experiment

websupporter:
## Description

Working on a solution for the time based triggers. This code is not tested at all. This PR atm is more for discussion and alignment.

## Code review notes

The general idea:
The new TimeBasedTriggerHandler registers a daily cron
In this cron, it searches for active automations that have a TimeBasedTrigger
For each of those a new action gets scheduled, which should execute immediately. It contains the trigger-key, the automation-id and an offset

When this fires, the automation gets pulled and now we ask the TimeBasedTrigger to create (or at least find, I think there is some architecture decision to make) the AutomationRuns.

It returns the number of runs created. If that is less than the max limit, we reschedule the same action but with a higher offset. This way, heavy automation run creations are executed in batches.

## QA notes

_N/A_

## Linked PRs

Implementation: https://github.com/mailpoet/mailpoet-premium/pull/935

## Linked tickets

[MAILPOET-6500](https://mailpoet.atlassian.net/browse/MAILPOET-6500)

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:time-based-triggers-experiment)

_The latest successful build from `time-based-triggers-experiment` will be used. If none is available, the link won't work._

[MAILPOET-6500]: https://mailpoet.atlassian.net/browse/MAILPOET-6500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mailpoet/lib/Automation/Engine/Control/ActionScheduler.php
+++ b/mailpoet/lib/Automation/Engine/Control/ActionScheduler.php
@@ -17,6 +17,11 @@ class ActionScheduler {
     return is_int($result) ? $result : 0;
   }
 
+  public function scheduleRecurring(int $timestamp, int $interval, string $hook, array $args = []): int {
+    $result = as_schedule_recurring_action($timestamp, $interval, $hook, $args, self::GROUP_ID);
+    return is_int($result) ? $result : 0;
+  }
+
   public function hasScheduledAction(string $hook, array $args = []): bool {
     return as_has_scheduled_action($hook, $args, self::GROUP_ID);
   }
--- a/mailpoet/lib/Automation/Engine/Control/TimeBasedTriggerHandler.php
+++ b/mailpoet/lib/Automation/Engine/Control/TimeBasedTriggerHandler.php
@@ -1,0 +1,135 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Automation\Engine\Control;
+
+use MailPoet\Automation\Engine\Data\Automation;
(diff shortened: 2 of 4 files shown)


JanJakes:
@websupporter What's the definition and use case of a time-based trigger? I'm just thinking... if it's something like `every monday at 8 AM` or an even more complex specification, then I guess a daily cron won't be enough to cover that. We could aim for a much smaller granularity.

A few thoughts that occurred to me when looking at the API. I like that it's managed by a trigger handler, which gives the ability for many different triggers to be time-based. Maybe it could somehow be moved to or injected into the existing `TriggerHandler`. A small detail...

But what if the time-based trigger interface exposed some kind of `getSchedule()` method where it would return any crontab record it desires, and the handler would make sure that it gets pinged more or less at the defined times? That's what I'd consider ideal from looking at the API from the "consumer" (= trigger implementor) side. "I give you a schedule, you make sure it runs." How to implement this is a question. Either a single recurring job that would run in much shorter intervals, or maybe leveraging Action Scheduler's `as_schedule_cron_action()`? Not sure what would it take to maintain a scheduled action per time-based trigger.

NeosinneR:
@websupporter I'm closing this but keeping the branch if/when we get back to time-based automations.

websupporter:
sounds good!

PR #6127, opened 2025-03-10
Add PHP 8.3 images and use it in automated tests [MAILPOET-6297]

lysyjan:
## Description

This PR adds dev configuration for PHP 8.3 and uses this version also in automated tests.

## Code review notes

I have some questions and would like to discuss them.
- Should I configure using PHP 8.3 by default for local development?
- Is there something that I overlooked and should be checked?

## QA notes

_N/A_

## Linked PRs

https://github.com/mailpoet/mailpoet-premium/pull/944
https://github.com/mailpoet/wordpress-images/pull/52

## Linked tickets

[MAILPOET-6297]

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:research-php-8-3)

_The latest successful build from `research-php-8-3` will be used. If none is available, the link won't work._

[MAILPOET-6297]: https://mailpoet.atlassian.net/browse/MAILPOET-6297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dev/php83/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+  rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+COPY dev/php.ini /usr/local/etc/php/conf.d/php_user.ini
+
+# msmtp config
+RUN printf "account default\nhost smtp\nport 1025" > /etc/msmtprc
+
+# xdebug build an config
+ENV XDEBUGINI_PATH=/usr/local/etc/php/conf.d/xdebug.ini
+RUN git clone -b "3.3.1" --depth 1 https://github.com/xdebug/xdebug.git /usr/src/php/ext/xdebug \
```
costasovo: There is xdebug [3.4.2 available](https://xdebug.org/updates#x_3_4_2). What about using the new version in the image? We could update also the PHP 8.2 image. The version 3.3.2 has a peformance fix for usage with Apache.

docker-compose.yml
```diff
@@ -60,31 +60,6 @@ services:
       - './packages:/var/www/html/wp-content/plugins/packages'
       - './templates:/var/www/templates'
 
-  test_wordpress:
```
costasovo, line 63: Nice! Happy to see this cleanup! We could probably also remove the [test database initialization in the DB service](https://github.com/mailpoet/mailpoet/blob/74e1616e15de23e2666ba1dcfb0c9e76e5e0ba1d/docker-compose.yml#L15). AFAIK it was used only for the test site.

costasovo (commented):
Great job! I added two comments with small suggestions for improvements we could consider.

costasovo:
@lysyjan, I forgot to answer the questions in the description.

> Should I configure using PHP 8.3 by default for local development?

I would switch. PHP 8.3 should be more performant and I guess it is better to use the newer version. 

> Is there something that I overlooked and should be checked?

Hmm. I see that the premium plugin also has some usages of PHP8.2 at least in Circle CI config. I'm not sure if that should be part of this task or not.

lysyjan:
@costasovo Thank you.
I had testing premium in my mind, but after moving to another task I forgot. Thank you. I will prepare another PR for the premium plugin.

lysyjan:
@costasovo I added new commits addressing your suggestions.

I also created two related PRs mentioned in the description.

costasovo (approved):
Great job! 

I checked the jobs in Circle CI and didn't notice any issues.
I built the WordPress image locally and ran a couple of random ./do commands and did basic check of admin MailPoet admin pages and didn't see any PHP Notices so I guess we are fine. 

I would merge the PR after the release.

PR #6169, opened 2025-04-22
Release 5.11.0

costasovo:
## Preview


NeosinneR (approved):
The code looks good, thanks for the skipped test fix. Approving now, not to block the release when CI passes.

PR #6178, opened 2025-04-25
Fix changelog and premium version release issues

NeosinneR:
## Description

This PR fixes two issues with the release.

### 1. When the version already exist in changelog, use it instead of using fallback

When the release is done, the script publishes the GitHub release and notifies on Slack. It uses the same Changelog controller as when generating it during release preparation. However, the Changelog controller was looking for the `x.x.x` template, and when it didn't find it (because it had already been replaced with the release version), it used a fallback changelog template.

For example, instead of

```
= 5.11.0 – 2025-04-22 =
* Added: Dynamic Products block in email editor. Useful in automations to render order products, cross-sells, abandoned cart content, or selected categories or tags;
* Updated: minimum required WordPress version to 6.7 and tested up to version to 6.8;
* Updated: minimum required WooCommerce version to 9.7.1 and tested up to version to 9.8.1.
```

it would output 

```
= 5.11.0 - 2025-04-23 =
* Improved: minor changes and fixes.
```

Now, it also accepts `$version`, and looks either for `x.x.x` or the specified version, in which case it uses the existing changelog.

### 2. Fixing the incorrect premium version when publishing the release 

Each plugin now uses its own GitHub client when fetching the latest release.

When only the free plugin client was used, after the free plugin was released, the release script for the premium plugin used the just-released version of the free plugin and bumped the minor version **again**, but that version didn't exist in the premium repository.

Now, each plugin checks its own releases, so even when 5.12.0 is released in the free, the latest premium release is still only 5.11.0, so it will correctly determine that it should release 5.12.0. This shouldn't affect determining whether a patch or minor version should be bumped, because the premium release always updates the minor version, and the free release checks for the presence of the `release` branch in the premium repository.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

STOMAIL-7398

## After-merge notes

_N/A_

## Tasks

### DIFF
--- a/mailpoet/tasks/release/ChangelogController.php
+++ b/mailpoet/tasks/release/ChangelogController.php
@@ -32,7 +32,7 @@ class ChangelogController {
       throw new \Exception('Version is required');
     }
     $changelog = $this->getChangelogFromReadme();
-    if (!$this->containsNewChangelog($changelog)) {
+    if (!$this->containsNewChangelogOrVersionExists($changelog, $version)) {
       $changelog = self::NEW_CHANGELOG_TEMPLATE . "\n" . self::FALLBACK_RECORD;
     }
     $changelog = $this->updateHeading($changelog, $version);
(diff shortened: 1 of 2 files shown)


costasovo (approved):
@NeosinneR Looks good! Thanks for the fixes.

I was wondering why we were using the free plugin client for the version, and I think it was to ensure that both premium and free versions have the same minor version, which is required for them to work together.
But AFAIK we never release only the free plugin minor version bump so we can safely assume that the latest free and premium minor versions always match.

NeosinneR:
> I was wondering why we were using the free plugin client for the version, and I think it was to ensure that both premium and free versions have the same minor version, which is required for them to work together.

@costasovo Exactly my thinking when implementing it. But it had this downside. And same as you, if we stick to updating the patch when releasing only the free plugin, the free and premium versions should always match.

PR #6183, opened 2025-04-29
Unsubscribe only on POST requests to List-Unsubscribe URL 

costasovo:
## Description

This PR addresses the issue that the URL we provide in the List-Unsubscribe header unsubscribes a recipient without confirmation when requested via both GET and POST HTTP methods. So when an antispam software accesses the URL, the recipient is unsubscribed.

According to the description by Certified Senders Alliance:

Screenshot 2025-04-29 at 9 13 02

The List-Unsubscribe header, when accessed via GET, should not immediately unsubscribe the recipient but display a page where they can unsubscribe (in our case, the confirmation page).

This PR fixes the issue described above by showing the confirmation page on GET and performing unsubscribe only on POST. 
I also refactored the confirmation page to use a form instead of a simple link for confirming unsubscribes. This is to further strengthen it against bots following links.

Closes STOMAIL-7396

## Code review notes

- I tried to explain decisions in commit messages.

## QA notes

Please test that unsubscribing from emails links works as expected.
- Check classic unsubscribe link in email body
- Check the link provided in the List-Unsubscribe email header. When opened in a browser, it should display a confirmation page, and when accessed via POST, it should immediately unsubscribe the subscriber.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

- [ ] I have added a changelog entry (pcNwfB-4Ov-p2#how-to-write-a-changelog)
- [ ] I followed [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers) for translations
- [ ] I added sufficient test coverage
- [ ] I embraced TypeScript by either creating new files in TypeScript or converting existing JavaScript files when making changes

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7396-subscribers-are-unsubscribed-on-get-request-to-url-provided)

_The latest successful build from `stomail-7396-subscribers-are-unsubscribed-on-get-request-to-url-provided` will be used. If none is available, the link won't work._

PR #6187, opened 2025-05-02
Add migration to re-subscribe subscribers unsubscribed by bots

costasovo:
## Description

This PR adds a migration that attempts to detect unsubscribes done by bots due to an issue fixed in https://github.com/mailpoet/mailpoet/pull/6183

In the migration, we consider subscribers who made three and more clicks and unsubscribed within 4 seconds before and 4 seconds after unsubscribe as unsubscribed by a bot.

From research of affected sites, it seems that the issue started in late March or early April. So we restrict the date to the start of March.

Closes STOMAIL-7399

## Code review notes

_N/A_

#### Testing

1. Install a plugin version without this migration e.g. 5.12.0
2. Create bot like unsubscribed subscribers:
 - Send an email
 - Click on multiple links (min 3)
 - Unsubscribe
 - Go to the DB `wp_mailpoet_statistics_unsubsribes` and check the `created_at` datetime of your unsubscribe
 - Update the clicks associated with the same subscriber in `wp_mailpoet_statistics_clicks` to have the `created_at` datetime set max. 4 seconds from the unsubscribe datetime
 3. Install the version with migration and make sure migration ran (e.g., deactivate and activate the plugin)
 4. Verify that the unsubscribed subscriber with multiple clicks at the same time is subscribed.
 5. Verify also that other unsubscribed subscribers (e.g, who clicked only one link) remained unsubscribed.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7399-add-migration-to-resubscribe-subscribers-unsubscribed-by-a)

_The latest successful build from `stomail-7399-add-migration-to-resubscribe-subscribers-unsubscribed-by-a` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/lib/Migrations/App/Migration_20250501_114655_App.php
+++ b/mailpoet/lib/Migrations/App/Migration_20250501_114655_App.php
@@ -1,0 +1,38 @@
+<?php declare(strict_types = 1);
+
+namespace MailPoet\Migrations\App;
+
+use MailPoet\Entities\StatisticsClickEntity;
+use MailPoet\Entities\StatisticsUnsubscribeEntity;
+use MailPoet\Entities\SubscriberEntity;
+use MailPoet\Migrator\AppMigration;
+
+class Migration_20250501_114655_App extends AppMigration {
+  public function run(): void {
+    $clicksStatsTable = $this->entityManager->getClassMetadata(StatisticsClickEntity::class)->getTableName();
+    $unsubscribeStatsTable = $this->entityManager->getClassMetadata(StatisticsUnsubscribeEntity::class)->getTableName();
+    $subscribersTable = $this->entityManager->getClassMetadata(SubscriberEntity::class)->getTableName();
+    $this->entityManager->getConnection()->executeQuery(
+      "UPDATE {$subscribersTable} SET status = :subscribedStatus WHERE id IN (
+        SELECT
+          mp_unsub.subscriber_id
+        FROM
+          {$unsubscribeStatsTable} AS mp_unsub
+        LEFT JOIN
+          {$clicksStatsTable} AS mp_click
+          ON mp_unsub.newsletter_id = mp_click.newsletter_id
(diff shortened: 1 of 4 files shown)


mailpoet/lib/Migrations/App/Migration_20250501_114655_App.php
```diff
@@ -6,32 +6,44 @@
...
+       SET status = :subscribedStatus
+       WHERE id IN (:subscriberIds)
+       AND status = :unsubscribedStatus",
       [
         'subscribedStatus' => SubscriberEntity::STATUS_SUBSCRIBED,
         'unsubscribedStatus' => SubscriberEntity::STATUS_UNSUBSCRIBED,
+        'subscriberIds' => $subscriberIds,
+      ],
+      [
+        'subscriberIds' => Connection::PARAM_INT_ARRAY,
```
lysyjan: I think we should use the `array_chunk` function here to avoid trouble with this migration. A high ID count can cause a too-long SQL query that ends with an error. This change also helps us to avoid performance issues.

lysyjan (commented):
I noticed one thing that could cause us problems.
It can be rare, but if it were to happen, it could be for a user with a significant number of subscribers.

mailpoet/lib/Migrations/App/Migration_20250501_114655_App.php
lysyjan: We use this approach already for importing subscribers etc. [Here](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/Subscribers/ImportExport/Import/Import.php#L421-L423) is an example.

lysyjan (approved):

Aschepikov (approved):
GJ!

PR #6189, opened 2025-05-05
Fix failing tests in May in years when February has 28 days

NeosinneR:
## Description

When February has 28 days in the year, starting on May 1, the [ReporterTest](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/tests/integration/Analytics/ReporterTest.php) tests will randomly fail, because subtracting 89 days will fall on the same day of the month, and if there is a 1 second difference, it will no longer technically belong in the "last 3 months" window.

For example, `May 1, 2025 13:00:00` - `89 days` = `February 1, 2025 13:00:00`. But if between creating the value, and asserting the test 1 second passes (or it might just be 1ms `13:00:00:999` vs. `13:00:01:000`), then a `last 3 months` since  `May 1, 2025 13:00:01` no longer includes `February 1, 2025 13:00:00` by only one second. 

The test was created in March 2024, when February had 29 days, which wasn't an issue. 

For the test purpose, there is no difference in subtracting `89 days` or `2 months`, so I just used 2 months.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

STOMAIL-7400

## After-merge notes

_N/A_

## Tasks

- [x] 🚫 I have added a changelog entry (pcNwfB-4Ov-p2#how-to-write-a-changelog)

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7400-failing-integration-reportertest)

_The latest successful build from `stomail-7400-failing-integration-reportertest` will be used. If none is available, the link won't work._


NeosinneR:
Merging as three consecutive CI builds passed.

PR #6216, opened 2025-06-06
Update tar-fs

costasovo:
## Description

Addresses https://github.com/mailpoet/mailpoet/security/dependabot/152

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:update-deps)

_The latest successful build from `update-deps` will be used. If none is available, the link won't work._


package.json
```diff
@@ -49,7 +49,9 @@
       "ws@>=6.0.0 <6.2.3": ">=6.2.3",
       "ws@>=8.0.0 <8.17.1": ">=8.17.1",
       "xmlhttprequest-ssl@<1.6.1": ">=1.6.1",
-      "xmlhttprequest-ssl@<1.6.2": ">=1.6.2"
+      "xmlhttprequest-ssl@<1.6.2": ">=1.6.2",
+      "tar-fs@<2.1.3": ">=2.1.3 <3.0.0",
+      "tar-fs@>=3.0.0 <3.0.8": ">=3.0.9"
```
pavel-mailpoet, line 54: @costasovo We've been adding dependencies to this `overrides` section. Do we also need to remove them eventually? Or do you know what the end plan is with this?

pavel-mailpoet (approved):

package.json
costasovo: @pavel-mailpoet Good question. I'm not aware we discussed that. The good thing is that they are all restricted by version, so they are not a problem for updates, but we should clean them up. We used to have tickets for updating dependencies regularly, and a task for cleaning up the overrides would IMO nicely fit into such a ticket.

PR #6264, opened 2025-07-21
Use email-editor packages installed via composer and npm

lysyjan:
## Description

This PR removes the email-editor packages from the repository and starts using packages installed via Composer and NPM.

## Code review notes

There is a list of major updates made in this PR:
- Adding email editor as a PHP dependency.
- Adding email editor as a JS dependency.
- Removing old packages from this repository.
- Deleting GH action with email editor package tests.
- Adding Jetpack Autoloader to avoid plugin conflicts with Woo.
- Moving Html2Text dependency into vendor to avoid conflicts with Woo.

## QA notes

Verify that the new email editor works correctly without and with the active WooCommerce plugin.

## Linked PRs

_N/A_

## Linked tickets

Closes STOMAIL-7471
Closes STOMAIL-7482

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:update-email-editor)

_The latest successful build from `update-email-editor` will be used. If none is available, the link won't work._


triple0t (commented):
Incredible. 
Well done, @lysyjan 👏🏾

Here is some quick feedback after reviewing the PR. 
I'm going to divide the comments by sections so we can decide which to tackle.

**MailPoet**

1. We need a `.gitkeep` file in the MailPoet JS and PHP packages [directory](https://github.com/mailpoet/mailpoet/tree/da3775a5ad75f9252f500ae12562b45552e393ea/packages)
2. We missed a few updates. [Here](https://github.com/mailpoet/mailpoet/blob/da3775a5ad75f9252f500ae12562b45552e393ea/mailpoet/tests/integration/EmailEditor/Integrations/MailPoet/EmailEditorTest.php#L24) and [here](https://github.com/mailpoet/mailpoet/blob/da3775a5ad75f9252f500ae12562b45552e393ea/mailpoet/assets/js/src/mailpoet-email-editor-integration/index.ts#L51).
3. UI issues with the alpha selector model  (Image)

| Before | After |
|--------|--------|
| Screenshot 2025-07-21 at 4 32 35 PM | Screenshot 2025-07-21 at 3 02 34 PM | 
4. Email Editor classes registered in MailPoet container: Do we need to do this? This is essentially replicating efforts, as we would need to keep both the WooCommerc package and the MailPoet container registration in sync. Can't we use `Email_Editor_Container::container()->get()`?
5. Unable to Switch / Swap newsletter template (Image)

| Before | After |
|--------|--------|
| Screenshot 2025-07-21 at 4 33 02 PM | Screenshot 2025-07-21 at 3 10 46 PM | 
6. Publish button showing in the editor on load. (Click on the create new newsetter, use the new editor, edit some content, notice the publish button. Save with "Ctrl + S" or  Click "Save draft", Reload the page, Edit content, and Notice the Publish button) (Image)
Screenshot 2025-07-21 at 3 09 43 PM Screenshot 2025-07-21 at 3 09 52 PM
7. Send button validation can be ignored. e.g., Edit the template and remove the Unsubscribe URL, Click on Save, Click on Send. Notice you can proceed to the send page without the Unsubscribe URL.
8. Used a Starter plan (Free plan) API key, but could not see the MailPoet logo
9. Paragraph block update logs in the browser console
Screenshot 2025-07-21 at 3 06 14 PM




**Email Editor General**
1. Global styles saved for all instances of the Email Editor (e.g set a global style like background color on MailPoet, notice the same background color on the WooCommerce emails)
2. Personalization tags showing from all registered editor instances (Not new).
3. WooCommerce Email content pattern showing on the MailPoet newsletter email preset selector modal. (Image)
Screenshot 2025-07-21 at 3 30 17 PM

**What works**  ✅ 
1. Email editor loads successfully 
2. Can edit content, add blocks, remove blocks, edit template  
3. Can preview emails and send email previews.   
4. Keyboard shortcuts works  
5. Can save global styles, edit global styles.  
6. Can rename emails and move emails to trash.  
7. Can preview sent emails  
8. Can use the email editor in both the MailPoet and WooCommerce plugins
9. No error in the browser console and PHP logs



@lysyjan,  Can we address some of the issues from the MailPoet part?

lysyjan:
Thank you for the nice summary @triple0t!

> 1. We need a .gitkeep file in the MailPoet JS and PHP packages [directory](https://github.com/mailpoet/mailpoet/tree/da3775a5ad75f9252f500ae12562b45552e393ea/packages)

Why do we need it? I think there is no reason to keep this directory. When we moved the email editor into a separate package, we created the php directory.

> 2. We missed a few updates. [Here](https://github.com/mailpoet/mailpoet/blob/da3775a5ad75f9252f500ae12562b45552e393ea/mailpoet/tests/integration/EmailEditor/Integrations/MailPoet/EmailEditorTest.php#L24) and [here](https://github.com/mailpoet/mailpoet/blob/da3775a5ad75f9252f500ae12562b45552e393ea/mailpoet/assets/js/src/mailpoet-email-editor-integration/index.ts#L51).

Fixed.

> 3. UI issues with the alpha selector model (Image)

I was not able to replicate this issue. Can you share your setup or steps to reproduce?

> 4. Email Editor classes registered in MailPoet container: Do we need to do this? This is essentially replicating efforts, as we would need to keep both the WooCommerc package and the MailPoet container registration in sync. Can't we use Email_Editor_Container::container()->get()?

There is [a follow-up ticket](https://linear.app/a8c/issue/STOMAIL-7482/use-the-email-editor-container-instead-of-the-di-container) for this step. Are you okay with that, or do you prefer to do it as a part of this PR?

> 5. Unable to Switch / Swap newsletter template (Image)

We made a change so that the swap action is displayed only when there is more than one template. I added a second template, and it works correctly.
Screenshot 2025-07-22 at 9 31 34

> 6. Publish button showing in the editor on load. (Click on the create new newsetter, use the new editor, edit some content, notice the publish button. Save with "Ctrl + S" or Click "Save draft", Reload the page, Edit content, and Notice the Publish button)

I can confirm this bug. It is a problem with the JS email-editor package. Do you think we need to address it before merging this PR?

> 7. Send button validation can be ignored. e.g., Edit the template and remove the Unsubscribe URL, Click on Save, Click on Send. Notice you can proceed to the send page without the Unsubscribe URL.

I will take a closer look at this issue. This has to be addressed.

> 8. Used a Starter plan (Free plan) API key, but could not see the MailPoet logo

This is a problem in [the current block implementation](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/EmailEditor/Integrations/MailPoet/Blocks/BlockTypes/PoweredByMailpoet.php#L22). We need to update the condition to a valid premium key. I will fix it.

> 9. Paragraph block update logs in the browser console

I cannot replicate this issue. Can you verify it with the inactive Woo plugin?

---

> 1. Global styles saved for all instances of the Email Editor (e.g set a global style like background color on MailPoet, notice the same background color on the WooCommerce emails)

This is something we need to address in the packages. I can create a ticket for that.

> 2. Personalization tags showing from all registered editor instances (Not new).

I created [a ticket](https://linear.app/a8c/issue/WOOPLUG-5118/add-the-possibility-to-set-supported-post-types-for-personalization) for this already. This is an edge case when Woo is active, and the block email editor feature is activated too.

> 3. WooCommerce Email content pattern showing on the MailPoet newsletter email preset selector modal. (Image)

I would say this is similar to the issues above, and we need to address it in the packages.

triple0t:
Thanks for the quick response, @lysyjan 

> Why do we need it? I think there is no reason to keep this directory. When we moved the email editor into a separate package, we created the php directory.

Even if the directory is empty, it's nice to keep the PHP Package folder to preserve the file/folder structure. 
For example, you can see it in the WooCommerce [monorepo](https://github.com/woocommerce/woocommerce/tree/c7a9f532b29fa8ed44dc256a6357442f2baa0e0f/packages/php) and the Jetpack repo.

> Fixed.

Thanks. I haven't found the commit addressing this. 


---


> I was not able to replicate this issue. Can you share your setup or steps to reproduce?

I tried with a fresh installation and it appears I only get the issue when the Gutenberg plugin is active.
We can ignore this for now.

> There is [a follow-up ticket](https://linear.app/a8c/issue/STOMAIL-7482/use-the-email-editor-container-instead-of-the-di-container) for this step. Are you okay with that, or do you prefer to do it as a part of this PR?

Nice. I'm okay with fixing this in another PR.

> We made a change so that the swap action is displayed only when there is more than one template. I added a second template, and it works correctly.

Nice.

> I can confirm this bug. It is a problem with the JS email-editor package. Do you think we need to address it before merging this PR?

It's confusing but it does not block this PR. 

> I will take a closer look at this issue. This has to be addressed.

Noted.

> This is a problem in [the current block implementation](https://github.com/mailpoet/mailpoet/blob/trunk/mailpoet/lib/EmailEditor/Integrations/MailPoet/Blocks/BlockTypes/PoweredByMailpoet.php#L22). We need to update the condition to a valid premium key. I will fix it.

Noted.

> I cannot replicate this issue. Can you verify it with the inactive Woo plugin?

I can't see the error on a fresh install. It looks like an issue with my local environment. We can ignore this.


> This is something we need to address in the packages. I can create a ticket for that.
> I created [a ticket](https://linear.app/a8c/issue/WOOPLUG-5118/add-the-possibility-to-set-supported-post-types-for-personalization) for this already. This is an edge case when Woo is active, and the block email editor feature is activated too.
> I would say this is similar to the issues above, and we need to address it in the packages.

Noted. Thank you.

triple0t (dismissed):

lysyjan:
@triple0t I hope I addressed all remaining issues. Because I tested the email editor package simplification, I added a commit using the container from the email editor package.

triple0t (commented):
Looks good. No errors.

Quick observations.

1. I can still see the WooContent starter template

Image

2. The editor canvas width is smaller

Before

Image

After

Image

3. Start from scratch button uses the WooTemplate

Image

triple0t (dismissed):
None of these issues are blocking, so I’m approving this PR.
Well done.

.circleci/wordpress/.htaccess
```diff
@@ -10,3 +10,19 @@ RewriteRule ^ - [L]
...
+RewriteEngine On
+RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+RewriteBase /
+RewriteRule ^index\.php$ - [L]
+RewriteCond %{REQUEST_FILENAME} !-f
+RewriteCond %{REQUEST_FILENAME} !-d
+RewriteRule . /index.php [L]
+</IfModule>
+
+# END WordPress
```
triple0t: This feels like a duplicate of the rewrite rule above. Do we need it?

mailpoet/assets/js/src/mailpoet-email-editor-integration/middleware/strip-post-status-on-save.ts
```diff
@@ -0,0 +1,148 @@
...
+  args: unknown[],
+  registry: Registry,
+  originalAction: ActionFunction,
+): Promise<unknown> {
+  try {
+    const [kind, name, recordOrId, options] = args;
+
+    // Validate kind and name
+    if (typeof kind !== 'string' || typeof name !== 'string') {
+      return await originalAction(...args);
```
triple0t, line 44: BTW, since `originalAction` is a promise, we can return it and skip the `await` part.
Is there a reason we are waiting for the result?

mailpoet/assets/js/src/mailpoet-email-editor-integration/middleware/strip-post-status-on-save.ts
```diff
@@ -0,0 +1,148 @@
...
+    // Proceed only for correct kind/name and when stripping is enabled
+    if (kind !== 'postType' || name !== postType) {
+      return await originalAction(...args);
+    }
+
+    // Ensure recordOrId is object with numeric id
+    if (
+      typeof recordOrId !== 'object' ||
+      recordOrId === null ||
+      typeof (recordOrId as Record<string, unknown>).id !== 'number'
```
triple0t:
```suggestion
      typeof (recordOrId as Record<string, unknown>)?.id !== 'number'
```

Also, if we are specifically checking for a number, we can typecast it to a numeric type for the check.
e.g.
```JavaScript
      typeof +recordOrId?.id !== 'number'
```

mailpoet/assets/js/src/mailpoet-email-editor-integration/middleware/strip-post-status-on-save.ts
```diff
@@ -0,0 +1,148 @@
...
+    const post = registry
+      .select(coreDataStore.name)
+      .getEntityRecord(
+        'postType',
+        postType,
+        typedRecordOrId.id as number,
+      ) as Record<string, unknown> | null;
+
+    // If post is missing or status is not defined, fallback to original action
+    if (!post || typeof post.status !== 'string') {
```
triple0t:
```suggestion
    if (!post || typeof post?.status !== 'string') {
```

mailpoet/prefixer/composer.json
triple0t, line 1: We can also remove this line: https://github.com/mailpoet/mailpoet/blob/7c7a322dff2389ae62f602971209cb222c73e4ed/mailpoet/prefixer/composer.json#L2-L7.

mailpoet/tasks/phpstan/email-editor-phpstan.neon
triple0t, line 1: I think we can totally remove this file as it's no longer useful.

triple0t (commented):
Tiny feedbacks

.circleci/wordpress/.htaccess
lysyjan: This is something that I didn't add. It was added by mistake. Probably generated by the IDE.

mailpoet/assets/js/src/mailpoet-email-editor-integration/middleware/strip-post-status-on-save.ts
lysyjan: When I didn't use `await` there was a lint error:
```
ESLint: Returning an awaited promise is required in this context. (@typescript-eslint/return-await)
```

triple0t: Ohh, this is because the method is [wrapped](https://github.com/mailpoet/mailpoet/blob/453e5bb1938cdacee08cc7205acbfc6eb5c2b9ff/mailpoet/assets/js/src/mailpoet-email-editor-integration/middleware/strip-post-status-on-save.ts#L39-L90) in a try-catch block. All good.

lysyjan:
@triple0t 
> 1. I can still see the WooContent starter template
> 3. Start from scratch button uses the WooTemplate

Those issues should be solved by using the Woo build from the trunk.

I added new commits addressing your suggestions 👍 Thanks

Aschepikov (approved):
GJ!

PR #6272, opened 2025-07-28
Exclude out-of-stock products in dynamic products block

costasovo:
## Description

Sending products that are out of stock in emails has a negative effect.

This PR adds an option to exclude out-of-stock products for the dynamic products block.

- When the checkbox is active, the dynamic product block includes only products that have `stock_status` `instock`, `lowstock`, or `onbackorder`. 
- Products that don't manage stock are always included.
- There is also a filter `mailpoet_products_exclude_out_of_stock_stock_status` that allows modifying the statuses.
- The PR also fixes an issue with the Dynamic product block that was not fetching products properly during sending.

Screenshot 2025-07-28 at 16 49 29

## Code review notes

_N/A_

## QA notes

1) Create a couple of products, including variable products
2) Setup the products to have managed stock
3) Verify that products that when the option "Exclude out-of-stock products" is checked, are out of stock and don't allow backorder, are excluded from the list.

## Linked PRs

_N/A_

## Linked tickets

STOMAIL-7484

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7484-hide-out-of-stock-products)

_The latest successful build from `stomail-7484-hide-out-of-stock-products` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **New Features**
  * Added an "Exclude out-of-stock products" checkbox to the dynamic products block settings in the newsletter editor.
  * Dynamic products block now supports excluding out-of-stock and unavailable products from display.

* **Bug Fixes**
  * Improved logic to prevent unintended product inclusion when the dynamic product type is not set to "selected".

* **Tests**
  * Added comprehensive tests to verify correct handling of out-of-stock and variable products in dynamic product listings.

### DIFF
--- a/mailpoet/assets/js/src/newsletter-editor/blocks/dynamic-products.ts
+++ b/mailpoet/assets/js/src/newsletter-editor/blocks/dynamic-products.ts
@@ -95,7 +95,7 @@ Module.DynamicProductsBlockModel = base.BlockModel.extend({
     base.BlockView.prototype.initialize.apply(this, args);
 
     this.on(
-      'change:amount change:contentType change:terms change:inclusionType change:displayType change:titleFormat change:featuredImagePosition change:titleAlignment change:titleIsLink change:imageFullWidth change:pricePosition change:readMoreType change:readMoreText change:sortBy change:showDivider change:dynamicProductsType change:titlePosition',
+      'change:amount change:contentType change:terms change:inclusionType change:displayType change:titleFormat change:featuredImagePosition change:titleAlignment change:titleIsLink change:imageFullWidth change:pricePosition change:readMoreType change:readMoreText change:sortBy change:showDivider change:dynamicProductsType change:titlePosition change:excludeOutOfStock',
       this._handleChanges,
       this,
     );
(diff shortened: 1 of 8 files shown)


NeosinneR (dismissed):
Thanks for this improvements, @costasovo. Looks and works as expected (`Polo` is out of stock). 

| Before | After | 
| ----  | ---- | 
| Screenshot 2025-07-29 at 10 17 18 | Screenshot 2025-07-29 at 10 16 45 | 

One thing I'd like to suggest is defaulting this option to `true`. I think we should set a reasonable defaults, even if they might not be backwards compatible when it's in the interest of most customers. WDYT?

costasovo:
@NeosinneR Thanks for the review!

> One thing I'd like to suggest is defaulting this option to true. I think we should set a reasonable defaults, even if they might not be backwards compatible when it's in the interest of most customers. WDYT?

Good point! I agree it is a reasonable default. Updated in https://github.com/mailpoet/mailpoet/pull/6272/commits/74e42f7b75377504ba837e70955e0f8ab00b2b1a

NeosinneR (approved):

Aschepikov (approved):
GJ!

PR #6275, opened 2025-07-30
Initial PHP 8.4 support

NeosinneR:
## Description

The first step in adding full support for PHP 8.4. We still can't switch local dev or tests to run on PHP 8.4, but some initial issues are already fixed. 

- Updated Tracy, Composer, and PHP-Scoper to the latest versions with PHP 8.4 support. 
- Added PHP 8.4 Dockerfile for local development.
- Fixed Carbon issue.
- Fixed deprecated implicitly marking params as nullable in vendor files.
    - I considered using https://cs.symfony.com/doc/rules/function_notation/nullable_type_declaration_for_default_null_value.html but ended up with a somewhat simple script instead of using a whole library. 
- Added `mailpoet/wordpress:8.4_20250729.1` and `mailpoet/wordpress:8.4-cli_20250729.1` Docker images in https://github.com/mailpoet/wordpress-images/pull/53

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

STOMAIL-7508

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7508-check-compatibility-with-php-84)

_The latest successful build from `stomail-7508-check-compatibility-with-php-84` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **New Features**
  * Added support for PHP 8.4 in the WordPress development environment, including a new Docker image with updated tools and PHP extensions.
  * Introduced an automated script to fix PHP 8.4 deprecation warnings in vendor code.
* **Bug Fixes**
  * Improved namespace prefixing for Doctrine dependencies to ensure compatibility.
  * Added attribute to suppress return type warnings in the Carbon library.
* **Chores**
  * Updated tool installation script to select Composer, PHP-Scoper, and Tracy versions based on the PHP runtime version.
  * Enhanced Composer scripts to run the new PHP 8.4 deprecation fixer automatically after install and update.

### DIFF
--- a/mailpoet/tools/install.php
+++ b/mailpoet/tools/install.php
@@ -1,14 +1,18 @@
 <?php // phpcs:ignore SlevomatCodingStandard.TypeHints.DeclareStrictTypes.DeclareStrictTypesMissing
 
+$composerVersion = '2.8.10';
+$phpScoperVersion = '0.18.17';
 $tracyVersion = '2.10.9'; // 2.10.10 causes SyntaxError in JS
 // The newer tracy version doesn't support PHP 7.X which we still support in tests and also in development environment.
 if (PHP_VERSION_ID < 80000) {
+  $composerVersion = '2.7.7';
+  $phpScoperVersion = '0.17.2';
   $tracyVersion = '2.9.4';
 }
 
 $tools = [
-  'https://github.com/composer/composer/releases/download/2.7.7/composer.phar' => 'composer.phar',
-  'https://github.com/humbug/php-scoper/releases/download/0.17.2/php-scoper.phar' => 'php-scoper.phar',
+  "https://github.com/composer/composer/releases/download/$composerVersion/composer.phar" => 'composer.phar',
+  "https://github.com/humbug/php-scoper/releases/download/$phpScoperVersion/php-scoper.phar" => 'php-scoper.phar',
   "https://github.com/nette/tracy/releases/download/v$tracyVersion/tracy.phar" => 'tracy.phar',
 ];
 // ensure installation in dev-mode only
--- a/mailpoet/tools/install.php
+++ b/mailpoet/tools/install.php
@@ -1,6 +1,6 @@
 <?php // phpcs:ignore SlevomatCodingStandard.TypeHints.DeclareStrictTypes.DeclareStrictTypesMissing
 
-$tracyVersion = '2.9.7';
+$tracyVersion = '2.10.9'; // 2.10.10 causes SyntaxError in JS
 // The newer tracy version doesn't support PHP 7.X which we still support in tests and also in development environment.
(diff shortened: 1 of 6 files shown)


mailpoet/tools/install.php
```diff
@@ -3,7 +3,7 @@
 $composerVersion = '2.8.10';
 $phpScoperVersion = '0.18.17';
 $tracyVersion = '2.10.10';
-// The newer tracy version doesn't support PHP 7.X which we still support in tests and also in development environment.
+// The newer versions drop support for PHP 7.4 which we still support in tests and also in development environment.
```
lysyjan: Does the comment match the PHP comparison? The comparison also excludes PHP 8.0 and 8.1.

lysyjan (commented):
When I tried to test PHP 8.4 locally, the Tracy debug bar was not visible when I installed dependencies on PHP 8.2 or higher. On PHP 8.1 it worked fine.

@NeosinneR Does the debug bar work for you with your environment?

mailpoet/tools/install.php
NeosinneR: I updated the comment in https://github.com/mailpoet/mailpoet/pull/6275/commits/0ed9ead3a1b144a60d20de3cf0b90d4ee56a7ca6 to mention also dropping 8.1. Technically, [Tracy 2.10.9 requires 8.0+](https://github.com/nette/tracy/blob/v2.10.9/composer.json#L18), [Composer 2.8.10 requires 7.2+](https://github.com/composer/composer/blob/2.8.10/composer.json#L25), but [PHP-Scoper 0.18.17 requires 8.2+](https://github.com/humbug/php-scoper/blob/0.18.17/composer.json#L20), so not to have two different conditions for 8.0+ and 8.2+, I consolidated all into just PHP 8.2+.

NeosinneR:
@lysyjan nice catch. As I was bumping Tracy versions to see if/when it breaks, I may have accidentally missed that the last bump to 2.10.10 actually broke it (there is syntax error in JS file, and it didn't seem like anything obvious to fix, so I didn't try to). I reverted it to 2.10.9 in https://github.com/mailpoet/mailpoet/pull/6275/commits/4dad009204e3f20377380e9d1c938d2e33b3be88 which works. 
Screenshot 2025-08-01 at 14 25 27

lysyjan (approved):
It looks good.
- Tracy works fine on PHP 8.4.
- No issues found

What I have noticed is:
- The command `./do doctrine:generate-cache` returns an error
- PHPStan does not work, and it needs to be updated to run on PHP 8.4
Screenshot 2025-08-06 at 13 41 16

PR #6307, opened 2025-08-27
Log error and provide default values for invalid segment statistics result

lysyjan:
## Description

This PR adds a fallback with logging when a segment statistics count query returns false.

## Code review notes

_N/A_

## QA notes

I guess we can skip QA here.

## Linked PRs

_N/A_

## Linked tickets

Closes STOMAIL-7537

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7537-imported-subscribers-not-showing-across-dashboard-pages)

_The latest successful build from `stomail-7537-imported-subscribers-not-showing-across-dashboard-pages` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for segment statistics: invalid or unexpected results are now logged and safely replaced with zeroed counts to prevent failures and ensure stable display.

* **Chores**
  * Added changelog entry for the fix.
  * Tightened static analysis baselines to remove previous suppressions and reflect stricter type checks.

### DIFF
--- a/mailpoet/changelog/2025-08-27-18-33-54-improved-error-handling-statistics-queries
+++ b/mailpoet/changelog/2025-08-27-18-33-54-improved-error-handling-statistics-queries
@@ -1,0 +1,5 @@
+# Type: Fixed
+
+# Description
+
+Improved error handling for invalid segment statistics results.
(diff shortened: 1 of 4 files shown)


NeosinneR (commented):
Should the same be also applied to the `getSubscribersGlobalStatusStatisticsCount()` method above?

Also, it doesn't really fix the `0 subscribers` issue when it should show a non-zero value. Have you considered throwing an error as this is really a non-desired state?

lysyjan:
> Should the same be also applied to the getSubscribersGlobalStatusStatisticsCount() method above?

@NeosinneR Good point, but the method is never used in the codebase. I similar behavior to all method in this repository because it's safer for all users.

> Also, it doesn't really fix the 0 subscribers issue when it should show a non-zero value. Have you considered throwing an error as this is really a non-desired state?

I think it is really an edge case, and because we have only one known staging site, I haven't considered it.

NeosinneR (approved):

PR #6334, opened 2025-09-18
Implement log level filtering for the email editor

yuliyan:
## Description

This PR implements a log level filtering for the email editor. This change aims at reducing the amount of log entries from the email editor by using the following strategy:

1. If the MailPoet log level is set to `Errors` only, log only Errors
1. If `WP_DEBUG` is enabled, MailPoet log level is set to `Everything`, do not limit the amount of logs
1. If `WP_DEBUG` is disabled, MailPoet log level is set to `Everything`, log the Info level and above, do not log the Debug level.

Closes [STOMAIL-7546: Implement Log Level Filtering for Email Editor Package](https://linear.app/a8c/issue/STOMAIL-7546/implement-log-level-filtering-for-email-editor-package).

## Code review notes

_N/A_

## QA notes

* Change the logging settings in MailPoet → Settings → Advanced → **Logging** and the email editor creates log entries based on the settings.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7546-implement-log-level-filtering-for-email-editor-package)

_The latest successful build from `stomail-7546-implement-log-level-filtering-for-email-editor-package` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

- **New Features**
  - Added log level filtering for the Email Editor so admins can control logging verbosity (errors-only to full debug); respects site debug mode to reduce noise or show more detail.

- **Documentation**
  - Added a changelog entry describing the Email Editor logging improvements.

### DIFF
--- a/mailpoet/lib/EmailEditor/Integrations/MailPoet/Logger.php
+++ b/mailpoet/lib/EmailEditor/Integrations/MailPoet/Logger.php
@@ -164,11 +164,13 @@ class Logger implements Email_Editor_Logger_Interface {
    * @return void
    */
   public function log(string $level, string $message, array $context = []): void {
-    if (!in_array($level, MonologLogger::getLevels()) && !is_numeric($level)) {
-      $level = MonologLogger::DEBUG;
+    try {
+      /** @phpstan-ignore-next-line toMonologLevel expects specific string or numeric values, but we handle invalid values gracefully with fallback */
+      $monologLevel = MonologLogger::toMonologLevel($level);
+    } catch (\Exception $e) {
+      $monologLevel = MonologLogger::DEBUG;
     }
-    /** @phpstan-ignore-next-line We check above that it's valid monolog level */
-    $monologLevel = MonologLogger::toMonologLevel($level);
+
     if ($this->shouldLogLevel($monologLevel)) {
       /** @phpstan-ignore-next-line PHPStan reports string in level as an error but it's okay */
       $this->mailpoetLogger->log($level, $message, $context);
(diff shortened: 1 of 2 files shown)


mailpoet/lib/EmailEditor/Integrations/MailPoet/Logger.php
```diff
@@ -116,7 +164,14 @@ public function debug(string $message, array $context = []): void {
    * @return void
    */
   public function log(string $level, string $message, array $context = []): void {
-    /** @phpstan-ignore-next-line PHPStan reports string in level as an error but it's okay */
-    $this->mailpoetLogger->log($level, $message, $context);
+    if (!in_array($level, MonologLogger::getLevels()) && !is_numeric($level)) {
+      $level = MonologLogger::DEBUG;
+    }
```
NeosinneR: `MonologLogger::getLevels())` is an associative array where keys are strings, but values are integers. 
```
array(8) {
  ["DEBUG"]     => int(100)
  ["INFO"       => int(200)
  ["NOTICE"     => int(250)
  ["WARNING"]   => int(300)
  ["ERROR"]     => int(400)
  ["CRITICAL"   => int(500)
  ["ALERT"]     => int(550)
  ["EMERGENCY"] => int(600)
}
```
And `in_array()` will take values and compare it against string `$level` which will always be false, practically resulting in `$level` being reset to `MonologLogger::DEBUG`.

This might need to be changed to use `array_keys`, but even then, there may be different cases so they should be normalized when comparing.

NeosinneR (changes requested):
Looks good when directly calling specific methods, but when using general `log()`, there is a bug, that would prevent it from properly logging depending on the setting.

mailpoet/lib/EmailEditor/Integrations/MailPoet/Logger.php
yuliyan: I reworked it to use try-catch with fallback approach in https://github.com/mailpoet/mailpoet/pull/6334/commits/0cc5580c15c66876b5d5bfc5c03e09e84dfa500e. `MonologLogger::toMonologLevel` accepts numeric, uppercase and lowercase level types and replicating the checks to cover the valid cases didn't seem justified.

NeosinneR (approved):
Thanks for the change, works as expected.

Aschepikov (approved):
GJ!

PR #6358, opened 2025-10-01
add front end form preprocess filter

PaulSchiretz:
## Description

If there is a mailpoet_form_widget_post_process filter there should also be a mailpoet_form_widget_pre_process where users have the chance to change the form title, styles etc... So i would say a filter where we pass the data array would be sufficient.

## Code review notes

Just added the filter.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Summary by CodeRabbit

* **New Features**
  * Enhanced form widget with a pre-processing step that allows adjustments before display, enabling greater customization and better compatibility with themes and plugins. Existing forms continue to work without changes.

* **Chores**
  * Hardened the form widget against direct file access to improve security and stability across environments.

### DIFF
--- a/mailpoet/lib/Form/Widget.php
+++ b/mailpoet/lib/Form/Widget.php
@@ -2,6 +2,9 @@
 
 namespace MailPoet\Form;
 
+if (!defined('ABSPATH')) exit;
+
+
 use MailPoet\API\JSON\API;
 use MailPoet\Config\Env;
 use MailPoet\Config\RendererFactory;
@@ -262,6 +265,7 @@ class Widget extends \WP_Widget {
       // render form
       $renderer = (new RendererFactory())->getRenderer();
       try {
+        $data = $this->wp->applyFilters('mailpoet_form_widget_pre_process', $data);
         $output = $renderer->render('form/front_end_form.html', $data);
         $output = WPFunctions::get()->doShortcode($output);
         $output = $this->wp->applyFilters('mailpoet_form_widget_post_process', $output);


mailpoet/lib/Form/Widget.php
```diff
@@ -2,6 +2,9 @@
 
 namespace MailPoet\Form;
 
+if (!defined('ABSPATH')) exit;
+
+
```
NeosinneR, line 7: This is added automatically in the build, and shouldn't be present in the files directly.

mailpoet/lib/Form/Widget.php
```diff
@@ -262,6 +265,7 @@ public function widget($args, $instance = null) {
       // render form
       $renderer = (new RendererFactory())->getRenderer();
       try {
+        $data = $this->wp->applyFilters('mailpoet_form_widget_pre_process', $data);
```
NeosinneR, line 268: Could you please add test, similarly to how [`mailpoet_form_widget_post_process` is tested](https://github.com/mailpoet/mailpoet/blob/4ea8a51d1dccfbdee74f634dc804c6b1390454b7/mailpoet/tests/integration/Form/WidgetTest.php#L38-L54)?

mailpoet/lib/Form/Widget.php
NeosinneR, line 245: The filter should be applied here, so it doesn't include sensitive data like nonce tokens.

NeosinneR (changes requested):
@PaulSchiretz thanks for creating the PR. I added a couple of comments that would need to be addressed before merging, but even before that, I wanted to ask, if you have an actual need for this filter. From the PR description, it sounds like you're adding it just because there is `*_post_*` filter, but that alone is not a good enough reason, as adding a filter means, we'd need to maintain it.

PaulSchiretz:
Hi @NeosinneR, your are right the pre filter might not be needed at all! sorry for the inconvenience! Cheers, Paul

PR #6365, opened 2025-10-08
feat: add WPML translation support for WooCommerce subscription opt-in label

mariusparaschiv:
## Description
Added WPML translation support for the WooCommerce checkout opt-in message to enable multilingual stores to translate the subscription checkbox label.

## QA notes
- Install and activate WPML
- Go to MailPoet → Settings → WooCommerce and set a custom opt-in message
- Go to WPML → String Translation
- Search for "mailpoet" or "woocommerce_checkout_optin_message"
- Translate the string to another language
- Switch site language and verify the translated message appears on WooCommerce checkout

## Linked ticket(s)
#6365

## Changelog entry
- Changed: WooCommerce checkout opt-in message is now translatable with WPML

## Translation best practices
✅ Translation ready

## Test coverage
✅ Existing tests cover the functionality

## Summary by CodeRabbit

* New Features
  * Added translation support for the WooCommerce checkout opt‑in message via WPML, enabling multilingual stores to localize the consent text.
  * Subscription field labels in WooCommerce are now translatable via WPML for consistent localization across checkout.
* Bug Fixes
  * Ensured the translated value is correctly displayed for subscription fields when translations are available.
* Chores
  * Automatically registers the checkout opt‑in message for translation when settings are saved, reducing manual setup for multilingual sites.

### DIFF
--- a/mailpoet/lib/API/JSON/v1/Settings.php
+++ b/mailpoet/lib/API/JSON/v1/Settings.php
@@ -139,6 +139,11 @@ class Settings extends APIEndpoint {
       $signupConfirmation = $this->settings->get('signup_confirmation.enabled');
       foreach ($settings as $name => $value) {
         $this->settings->set($name, $value);
+        
+        // Register WooCommerce opt-in message with WPML for translation
+        if ($name === 'woocommerce.optin_on_checkout.message' && function_exists('icl_register_string')) {
+          icl_register_string('mailpoet', 'woocommerce_checkout_optin_message', $value);
+        }
       }
 
       $this->onSettingsChange($oldSettings, $this->settings->getAll());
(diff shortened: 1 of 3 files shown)


NeosinneR (commented):
Hi @mariusparaschiv and thank you for the proposal. Unfortunately, I won't be able to accept this, with main reason being maintenance. WPML is a paid plugin, and MailPoet doesn't officially support it, nor do we have tests to be confident, that the proposed integration won't break at some point. And if it does, it'd be up to us to fix it.

For example, I tried installing WPML locally with your branch, and it resulted in fatal error caused by the `icl_register_string()`.

Image

These are the issue we'd need to be confident they can't occur in various setups (different PHP, MySQL, WordPress, or WooCommerce versions), and that'll likely only come with official WPML support, which would be heavily tested.  

If what you're trying to achieve can be done without relying explicitly on WPML (e.g. using WordPress' translation functions, or filters), I'd be happy to review again.

PR #6379, opened 2025-10-24
Add WooCommerce email context filters for newsletter rendering to include recipient email and type

lysyjan:
## Description

This PR updates the email-editor package, allowing the use of the Cart contents block in block emails.

Screenshot 2025-11-06 at 19 44 11

## Code review notes

_N/A_

## QA notes

Prerequisites: Install WooCommerce built from the trunk branch. 
You can use this zip [woocommerce.zip](https://github.com/user-attachments/files/23417667/woocommerce.zip)

1. Activate the email block editor for automations in MailPoet settings
2. Use the new block in Abandoned cart automation
3. Verify that email sent by automations contain the expected products

## Linked PRs

_N/A_

## Linked tickets

Closes STOMAIL-7500

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7500-cart-contents-in-email)

_The latest successful build from `stomail-7500-cart-contents-in-email` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Improved**
  * Mail editor now supports the Product collection block variant for Cart contents in email templates, enabling enhanced customization options for displaying shopping cart information.
  * Email rendering improved for automation-type newsletters, providing better context handling for automated communications.

### DIFF
--- a/mailpoet/lib/Newsletter/Renderer/Renderer.php
+++ b/mailpoet/lib/Newsletter/Renderer/Renderer.php
@@ -83,7 +83,24 @@ class Renderer {
     $wpPostEntity = $newsletter->getWpPost();
     $wpPost = $wpPostEntity ? $wpPostEntity->getWpPostInstance() : null;
     if ($wpPost instanceof \WP_Post) {
+      $filterCallback = function($context) use ($newsletter, $sendingQueue): array {
+        if ($sendingQueue) {
+          $task = $sendingQueue->getTask();
+          $subscribers = $task ? $task->getSubscribers() : null;
+          $subscriber = ($subscribers && $subscriber = $subscribers->first()) ? $subscriber->getSubscriber() : null;
+
+          if ($subscriber) {
+            $context['recipient_email'] = $subscriber->getEmail();
+          }
+        }
+        $context['email_type'] = $newsletter->getType();
+        return $context;
+      };
+      $this->wp->addFilter('woocommerce_email_editor_rendering_email_context', $filterCallback);
+
       $renderedNewsletter = $this->guntenbergRenderer->render($wpPost, $subject, $newsletter->getPreheader(), $language, $metaRobots);
+
+      $this->wp->removeFilter('woocommerce_email_editor_rendering_email_context', $filterCallback);
     } else {
       $body = (is_array($newsletter->getBody()))
         ? $newsletter->getBody()
--- a/mailpoet/lib/Newsletter/Renderer/Renderer.php
+++ b/mailpoet/lib/Newsletter/Renderer/Renderer.php
@@ -83,24 +83,38 @@ class Renderer {
     $wpPostEntity = $newsletter->getWpPost();
     $wpPost = $wpPostEntity ? $wpPostEntity->getWpPostInstance() : null;
     if ($wpPost instanceof \WP_Post) {
-      $filterCallback = function($context) use ($newsletter, $sendingQueue): array {
-        if ($sendingQueue) {
-          $task = $sendingQueue->getTask();
-          $subscribers = $task ? $task->getSubscribers() : null;
-          $subscriber = ($subscribers && $subscriber = $subscribers->first()) ? $subscriber->getSubscriber() : null;
+      // Only add the email context filter for automation emails.
+      // For bulk emails like newsletters, the first subscriber in the task isn't the unique recipient.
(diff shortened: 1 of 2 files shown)


costasovo (commented):
@lysyjan The code looks good. When I tried test it I received the email but there were no products.

Image
Image

lysyjan:
@costasovo I think the issue was caused by the same email editor version used in Woo `composer.lock` and MailPoet. I released the email editor version and use it in MailPoet.

lysyjan:
There is an issue with block templates. I prepared a fix in the email editor package and need to release a new version.

costasovo (approved):
@lysyjan If an incorrect version caused the issue, I think it is fine to merge. The mail with the block was correctly sent, so we can merge.

NeosinneR:
@lysyjan can this be merged once conflicts are resolved?

lysyjan:
@NeosinneR I need to release a new version of the email editor package and use it in this PR. It's at the top of my to-do list.

lysyjan:
@costasovo I updated the used email-editor package, and it should work now as expected.

mailpoet/lib/Newsletter/Renderer/Renderer.php
```diff
@@ -83,7 +83,24 @@ private function _render(NewsletterEntity $newsletter, ?SendingQueueEntity $send
...
+        if ($sendingQueue) {
+          $task = $sendingQueue->getTask();
+          $subscribers = $task ? $task->getSubscribers() : null;
+          $subscriber = ($subscribers && $subscriber = $subscribers->first()) ? $subscriber->getSubscriber() : null;
+
+          if ($subscriber) {
+            $context['recipient_email'] = $subscriber->getEmail();
+          }
+        }
+        $context['email_type'] = $newsletter->getType();
```
costasovo: I think this is a bit dangerous to do this way. In the renderer, we have no idea who the real recipient is, and we assume it is the first subscriber linked to the task.
This will work for automation emails, which have only one subscriber, but it will create incorrect context data for Newsletter emails, which have multiple subscribers.

costasovo (commented):
@lysyjan I shared one comment. I have a concern about creating the context inside the renderer.

mailpoet/lib/Newsletter/Renderer/Renderer.php
lysyjan: @costasovo Do you see here a better approach? Maybe we can set the recipient like this only when the newsletter type is `automation`. I can try to take a closer look, but at this moment, I have no idea.

costasovo: > Maybe we can set the recipient like this only when the newsletter type is automation.

Yeah, at least adding some guard like this and proper comment would be helpful.

> Do you see here a better approach?

I think logic like this could be better placed somewhere in a code related to email sending where we are perfectly aware of what and to whom we are sending. Maybe somewhere in sending queue processing?

lysyjan: @costasovo I slightly updated the code. Unfortunately, we cannot move this to sending queue processing because the context has to be set before block rendering.

costasovo (approved):
Thank you for adding the check for autmation emails!

mailpoet/lib/Newsletter/Renderer/Renderer.php
```diff
@@ -83,7 +83,38 @@ private function _render(NewsletterEntity $newsletter, ?SendingQueueEntity $send
...
     $wpPost = $wpPostEntity ? $wpPostEntity->getWpPostInstance() : null;
     if ($wpPost instanceof \WP_Post) {
+      // Only add the email context filter for automation emails.
+      // For bulk emails like newsletters, the first subscriber in the task isn't the unique recipient.
+      $automationTypes = [
+        NewsletterEntity::TYPE_AUTOMATION,
+        NewsletterEntity::TYPE_AUTOMATION_NOTIFICATION,
+        NewsletterEntity::TYPE_AUTOMATION_TRANSACTIONAL,
+      ];
+      $isAutomationType = in_array($newsletter->getType(), $automationTypes, true);
```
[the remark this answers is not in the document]
lysyjan: I think those old email types are not available for the new block email editor.

PR #6383, opened 2025-10-30
Update integration tests to work with WP 6.9

costasovo:
## Description

The PR updated integration tests to work with WP 6.9

There were two issues:

- empty passwords in wp_insert_user started throwing warnings  - fixed by switching to the helper for creating users
- updated PHPMailer, changed indexes in the array that holds addresses, causing tests to fail

Closes STOMAIL-7607

## Code review notes

There is a [test run where I switched integration tests to use WP 6.9](https://app.circleci.com/pipelines/github/mailpoet/mailpoet/21642/workflows/454bd025-ff3f-415f-b353-ce6e2e2cda41)

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:wp69-fixes)

_The latest successful build from `wp69-fixes` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Tests**
  * Streamlined test user creation by centralizing and extending the test helper to accept an optional username.
  * Replaced direct user creation in tests with the helper for consistent role assignment.
  * Simplified mailer reply-to assertions to validate the primary reply-to entry.
  * Removed some per-test manual user cleanup to rely on the centralized test harness.

### DIFF
--- a/mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
+++ b/mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
@@ -908,7 +908,7 @@ class SchedulerTest extends \MailPoetTest {
     return $newsletter;
   }
 
-  public function _createOrUpdateWPUser($role = null) {
+  public function _createOrUpdateWPUser(string $role) {
     $email = 'test@example.com';
     $username = 'phoenix_test_user';
     if (email_exists($email) === false) {
--- a/mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
+++ b/mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
@@ -912,13 +912,7 @@ class SchedulerTest extends \MailPoetTest {
     $email = 'test@example.com';
     $username = 'phoenix_test_user';
     if (email_exists($email) === false) {
-      wp_insert_user(
-        [
-          'user_login' => $username,
-          'user_email' => $email,
-          'user_pass' => '',
-        ]
-      );
+      $this->tester->createWordPressUser($email, $role, $username);
     }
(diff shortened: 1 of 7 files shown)


mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
```diff
@@ -912,13 +912,7 @@ public function _createOrUpdateWPUser($role = null) {
...
     $username = 'phoenix_test_user';
     if (email_exists($email) === false) {
-      wp_insert_user(
-        [
-          'user_login' => $username,
-          'user_email' => $email,
-          'user_pass' => '',
-        ]
-      );
+      $this->tester->createWordPressUser($email, $role, $username);
```
NeosinneR, line 915: `$role` can be `null` in this method. But `createWordPressUser` has `string $role` which is non-nullable, which might result in type error. So we should either ensure, that `$role` is always set here, or that it can be `null` in `createWordPressUser`

NeosinneR (commented):
Thanks for the fixes. I found one potential issue, in the comment below:

mailpoet/tests/integration/Cron/Workers/SchedulerTest.php
costasovo: Good catch. This looks like something the code rabbit could notice. I addressed it in https://github.com/mailpoet/mailpoet/pull/6383/commits/df3864277fc727c420e2d089a3329dc07fc8df4f

NeosinneR (approved):

PR #6388, opened 2025-11-03
Ensure the correct version of Woo Core is used

pavel-mailpoet:
## Description

Previously the build on alwasy CI used the latest version version of WooCommerce Core. This PR changes that and  ensures we use the version that is configured.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:print-plugins-versions)

_The latest successful build from `print-plugins-versions` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Chores**
  * Added conditional checks to skip core downloads when no specific WooCommerce version is specified, improving CI behavior.
  * Added clear, user-facing progress messages for plugin download steps (Memberships, Subscriptions, AutomateWoo, WooCommerce) to provide better build feedback.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -468,6 +468,7 @@ jobs:
                 name: Download WooCommerce Core
                 command: |
                   cd ../tests_env/docker
+                  echo "Downloading WooCommerce Core with tag: '${WOOCOMMERCE_VERSION}'"
                   docker compose run --rm -w /project --entrypoint "./do download:woo-commerce-zip ${WOOCOMMERCE_VERSION}" --no-deps -e WP_GITHUB_USERNAME=${WP_GITHUB_USERNAME} -e WP_GITHUB_WOOCOMMERCE_TOKEN=${WP_GITHUB_WOOCOMMERCE_TOKEN} codeception_acceptance
       - when:
           condition: << parameters.woo_subscriptions_version >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -467,9 +467,13 @@ jobs:
             - run:
                 name: Download WooCommerce Core
                 command: |
-                  cd ../tests_env/docker
-                  echo "Downloading WooCommerce Core with tag: '${WOOCOMMERCE_VERSION}'"
-                  docker compose run --rm -w /project --entrypoint "./do download:woo-commerce-zip ${WOOCOMMERCE_VERSION}" --no-deps -e WP_GITHUB_USERNAME=${WP_GITHUB_USERNAME} -e WP_GITHUB_WOOCOMMERCE_TOKEN=${WP_GITHUB_WOOCOMMERCE_TOKEN} codeception_acceptance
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -467,8 +467,13 @@ jobs:
             - run:
                 name: Download WooCommerce Core
                 command: |
-                  cd ../tests_env/docker
-                  docker compose run --rm -w /project --entrypoint "./do download:woo-commerce-zip ${WOOCOMMERCE_VERSION}" --no-deps -e WP_GITHUB_USERNAME=${WP_GITHUB_USERNAME} -e WP_GITHUB_WOOCOMMERCE_TOKEN=${WP_GITHUB_WOOCOMMERCE_TOKEN} codeception_acceptance
+                  if [ -z "${WOOCOMMERCE_VERSION}" ]; then
+                    echo "WooCommerce version is not set."
```
costasovo: "WooCommerce version is not set." sounds like an error, but it is a normal state. We are just not switching the version.

I would suggest using a message that would be more reassuring that this is not a problem.
Perhaps something like:
"The test run doesn't require a special WooCommerce version. Skipping download and continuing with the default version."

costasovo (commented):
Thanks for the fix and for adding the additional logs! I would suggest using a more detailed message. Please see my other comment.

.circleci/config.yml
```diff
@@ -752,8 +757,13 @@ jobs:
...
                 command: |
-                  cd ../tests_env/docker
-                  docker compose run --rm -w /project --entrypoint "./do download:woo-commerce-zip ${WOOCOMMERCE_VERSION}" --no-deps -e WP_GITHUB_USERNAME=${WP_GITHUB_USERNAME} -e WP_GITHUB_WOOCOMMERCE_TOKEN=${WP_GITHUB_WOOCOMMERCE_TOKEN} codeception_integration
+                  if [ -z "${WOOCOMMERCE_VERSION}" ]; then
+                    echo "WooCommerce version is not set."
+                  else
+                    cd ../tests_env/docker
+                    echo "Downloading WooCommerce Core with tag: '${WOOCOMMERCE_VERSION}'"
+                    docker compose run --rm -w /project --entrypoint "./do download:woo-commerce-zip ${WOOCOMMERCE_VERSION}" --no-deps -e WP_GITHUB_USERNAME=${WP_GITHUB_USERNAME} -e WP_GITHUB_WOOCOMMERCE_TOKEN=${WP_GITHUB_WOOCOMMERCE_TOKEN} codeception_acceptance
+                  fi
```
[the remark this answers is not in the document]
pavel-mailpoet: Thanks, fixed.

.circleci/config.yml
pavel-mailpoet: updated

costasovo (approved):
Looks good! Thank you!

PR #6398, opened 2025-11-10
A11Y - Improve accessing the new editor [STOMAIL-7619]

websupporter:
## Description
It is hard to access the new editor using only the keyboard.

1. The dropdown opening button has only an SVG, so screen readers would not communicate its purpose
2. The tab order was manipulated. A user would expect to enter the dropdown using tab, but instead the focus would jump into automations. 

This PR adds a `aria-label` to the button and focuses the dropdown on mount. Using git-blame I could not find a reason why we disabled this behavior. The documentation states:

> Specifying a false value disables the focus handling entirely (this should only be done when an appropriately accessible substitute behavior exists).
https://developer.wordpress.org/block-editor/reference-guides/components/dropdown/#focusonmount-firstelement-boolean

Additionally, I added some `aria-labels` for the "Create"-buttons, so it reads more precisely what is created.

Before:

https://github.com/user-attachments/assets/51a89cd7-c73e-40e2-8cae-0196fe398c5e

After:

https://github.com/user-attachments/assets/4ea5fe13-f927-41bf-8626-c62b3c3c40e1

## Code review notes

I tried also to get the info-toggle next to automations to be read out as the cursor stops at the heading, but this seems not to work. I used only text without div in the `<Tooltip>`, but this did not do the trick. Maybe we can revisit this the next time.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7619-tab-order-for-create-using-the-new-email-editor-is-wrong)

_The latest successful build from `stomail-7619-tab-order-for-create-using-the-new-email-editor-is-wrong` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Improvements**
  * Added descriptive ARIA labels to action buttons and dropdowns (create newsletter, choose editor version, create automation, create latest post notification, create re-engagement email) to improve screen reader support.
  * Adjusted dropdown focus behavior for a smoother keyboard navigation experience.

* **Documentation**
  * Added a changelog entry noting the accessibility improvements to the new newsletter editor.

### DIFF
--- a/mailpoet/assets/js/src/newsletters/types.tsx
+++ b/mailpoet/assets/js/src/newsletters/types.tsx
@@ -117,7 +117,6 @@ export function NewsletterTypes({
         {__('Create', 'mailpoet')}
       </Button>
       <Dropdown
-        focusOnMount={false}
         className="mailpoet-dropdown-button"
         contentClassName="mailpoet-dropdown-button-content"
         popoverProps={{ placement: 'bottom-end' }}
@@ -129,6 +128,7 @@ export function NewsletterTypes({
             isBusy={isCreating === 'standard'}
             disabled={isCreating !== null}
             aria-expanded={isOpen}
+            aria-label={__('Open dropdown menu', 'mailpoet')}
             data-automation-id="create_standard_email_dropdown"
           >
             <Icon icon={chevronDown} size={24} />
--- a/mailpoet/assets/js/src/newsletters/types.tsx
+++ b/mailpoet/assets/js/src/newsletters/types.tsx
@@ -129,7 +129,7 @@ export function NewsletterTypes({
             isBusy={isCreating === 'standard'}
             disabled={isCreating !== null}
             aria-expanded={isOpen}
-            aria-label={__('Open dropdown menu', 'mailpoet')}
+            aria-label={__('Choose editor version', 'mailpoet')}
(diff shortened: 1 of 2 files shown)


mailpoet/assets/js/src/newsletters/types.tsx
```diff
@@ -129,6 +128,7 @@ export function NewsletterTypes({
             isBusy={isCreating === 'standard'}
             disabled={isCreating !== null}
             aria-expanded={isOpen}
+            aria-label={__('Open dropdown menu', 'mailpoet')}
```
costasovo: As this dropdown is very specific for the new email editor, perhaps we could say: "Open dropdown with alternative options for creating emails" or something like this. Hmm, perhaps it would be too long. I'll leave the comment here for consideration.

costasovo (approved):
@websupporter Thank you for the improvements! I added one comment with a suggestion. Feel free to merge it

mailpoet/assets/js/src/newsletters/types.tsx
websupporter: "Choose editor version"? 477e7db71a

What do you think @costasovo

PR #6401, opened 2025-11-11
Fix schedule and recipients controls in email editor

yuliyan:
## Description

This PR resolves the following issues with the schedule and recipients fields in the email editor sidebar:

1. Set the `isScheduled` newsletter option accordingly when updating the `scheduledAt` from the email editor sidebar field.
2. Mark past dates in the scheduled date time picker as invalid.
3. Add a custom time option on the `/wp-admin/admin.php?page=mailpoet-newsletters#/send/<id>` screen to properly render a custom times set from the email editor.
4. Fix a bug with recipients lists being removed on subsequent email settings updates from within the email editor.

## Code review notes

_N/A_

## QA notes

* Navigate to MailPoet → Emails → Add New Email → Create using the new email editor (Alpha).
* Select a date to schedule the email and choose your recipients.
* Save the email, then go to Review & Schedule.
* Confirm that the recipients and scheduled date match your selections.
* Return to the email editor and update the recipients.
* Verify that the changes are reflected correctly on the Review & Schedule screen.

## Linked PRs

_N/A_

## Linked tickets

[STOMAIL-7622: Newsletter Send and Recipients controls in Email Editor sidebar not working correctly](https://linear.app/a8c/issue/STOMAIL-7622/newsletter-send-and-recipients-controls-in-email-editor-sidebar-not)

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7622-newsletter-send-and-recipients-controls-in-email-editor)

_The latest successful build from `stomail-7622-newsletter-send-and-recipients-controls-in-email-editor` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed schedule and recipients fields in the email editor sidebar.
  * Scheduling status now saves and displays correctly.
  * Prevents selecting past dates when scheduling (aligned to day boundary).

* **Improvements**
  * Time selector accepts and shows custom time values when not in presets.

* **Tests**
  * Added integration tests covering scheduling and segment management.

* **Documentation**
  * Added changelog entry describing the fixes.

### DIFF
--- a/mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/scheduled-row.tsx
+++ b/mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/scheduled-row.tsx
@@ -73,6 +73,9 @@ export function ScheduledRow() {
       .reverse()
       .join(''), // Reverse the string and test for "a" not followed by a slash.
   );
+  // Used for comparing today with DateTimePicker dates to determine validity.
+  // We set the hours to 0:00:00 to match the time format of DateTimePicker dates.
+  const today = new Date().setHours(0, 0, 0, 0);
 
   return (
     <PanelRow>
@@ -140,6 +143,7 @@ export function ScheduledRow() {
                     _x('dmy', 'date order', 'mailpoet')
                   }
                   is12Hour={is12HourTime}
+                  isInvalidDate={(date) => date.getTime() < today}
                 />
               </div>
             )}
--- a/mailpoet/assets/js/src/newsletters/send/time-select.jsx
+++ b/mailpoet/assets/js/src/newsletters/send/time-select.jsx
@@ -18,6 +18,20 @@ class TimeSelect extends Component {
         {timeOfDayItems[val]}
       </option>
     ));
+    // If the current value is not in the predefined timeOfDayItems list,
+    // create a custom option for it. This handles cases where the scheduled time
+    // is set from within the email editor, as the datetime picker allows for
+    // setting any time (not just the predefined options). This ensures the select
+    // renders the correct value and doesn't fall back to the first predefined time.
+    const predefinedTimeKeys = Object.keys(timeOfDayItems);
+    const isCustomTime = value && !predefinedTimeKeys.includes(value);
+    // To match the format of the predefined time options, we remove the seconds from the custom time value.
+    const lastColonIndex = value.lastIndexOf(':');
+    const customOptionLabel =
+      lastColonIndex > 0 ? value.slice(0, lastColonIndex) : value;
+    const customOption = isCustomTime ? (
+      <option value={value}>{customOptionLabel}</option>
+    ) : null;
 
     return (
       <Select
@@ -28,6 +42,7 @@ class TimeSelect extends Component {
         isMinWidth
         {...validation}
       >
+        {customOption}
         {options}
       </Select>
     );
--- a/mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
+++ b/mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
@@ -97,7 +97,7 @@ class EmailApiController {
       $newsletter->setDeletedAt($data['deleted_at']);
     }
 
-    if (isset($data['scheduled_at'])) {
+    if (array_key_exists('scheduled_at', $data)) {
       $this->updateScheduledAtOption($newsletter, $data['scheduled_at']);
     }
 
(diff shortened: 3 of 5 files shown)


mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/scheduled-row.tsx
```diff
@@ -73,6 +73,7 @@ export function ScheduledRow() {
       .reverse()
       .join(''), // Reverse the string and test for "a" not followed by a slash.
   );
+  const today = new Date().setHours(0, 0, 0, 0);
```
pavel-mailpoet, line 78: This is really a nitpick, but it really caught my eye. Do I understand correctly that you used this to get a current timestamp? Would `Date.now` work the same, but be more understandable for future developers? Even `setHours(0)` seems the same.

mailpoet/assets/js/src/newsletters/send/time-select.jsx
```diff
@@ -18,6 +18,10 @@ class TimeSelect extends Component {
         {timeOfDayItems[val]}
       </option>
     ));
+    const customOption =
+      value && Object.keys(timeOfDayItems).indexOf(value) === -1 ? (
+        <option value={value}>{value.slice(0, value.lastIndexOf(':'))}</option>
```
pavel-mailpoet: I find this code quite hard to read, especially the condition. I am not sure what value we need from this piece of code. I read the description:

> Add a custom time option on the /wp-admin/admin.php?page=mailpoet-newsletters#/send/<id> screen to properly render a custom times set from the email editor.

But I still am not sure what the value could be and how that works. Do we need a comment for this?

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
```diff
@@ -97,7 +97,7 @@ public function saveEmailData(array $data, \WP_Post $emailPost): void {
       $newsletter->setDeletedAt($data['deleted_at']);
     }
 
-    if (isset($data['scheduled_at'])) {
+    if (array_key_exists('scheduled_at', $data)) {
```
pavel-mailpoet, line 100: Can you explain this change? I am not sure I understand why it was needed.

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
```diff
@@ -142,17 +142,42 @@ private function updateScheduledAtOption($newsletter, $scheduledAtValue): void {
 
     // Set the value (null or the datetime string)
     $option->setValue($scheduledAtValue);
+
+    // Also update the isScheduled option
```
pavel-mailpoet, line 146: Do you think we need an integration test for the changes in this file?

pavel-mailpoet (commented):
Looks mostly good, I have a couple of questions.

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
yuliyan: Thanks for the reminder - I added tests in https://github.com/mailpoet/mailpoet/pull/6401/commits/581e861ca9387d4fa1c9cba2031ae44d915b4155 including tests for the recipients field.

mailpoet/lib/EmailEditor/Integrations/MailPoet/EmailApiController.php
yuliyan: We accept `null` as valid valud for `scheduled_at`, i.e used to remove the scheduled at. That said, using `isset( $data['scheduled_at'] )` where `$data['scheduled_at']` is `null` would prevent the update call.

mailpoet/assets/js/src/newsletters/send/time-select.jsx
yuliyan: I have split the code a bit and added some comments, which should make it a bit more clear.

mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/scheduled-row.tsx
yuliyan: The DateTimePicker dates use `00:00:00` for the time, and thus we need the current date at `00:00:00` for the date validity comparison.

As for using `.setHours(0)` - it will only set the hours, but the minutes, seconds, and milliseconds will remain the current ones. So we need all 4 parameters in this case.

I do agree this could be a little confusing, so I added some comments to clarify the case.

yuliyan:
@pavel-mailpoet, I have addressed your feedback. Could you take another look if everything looks good?

mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/scheduled-row.tsx
pavel-mailpoet: I understand now, thank you for the explanation.

pavel-mailpoet (approved):

Aschepikov (approved):
GJ!

PR #6414, opened 2025-11-21
Cleanup the Env class from db related constants

lysyjan:
## Description

  Clean up unused database-related parameters from the `Env` class after migrating to WordPress's `$wpdb` database abstraction layer. This reduces
  technical debt and simplifies the codebase by removing 12 parameters that were only needed for direct database connections.

  ### What changed:
  - **Removed 12 unused static properties** from `Env` class:
    - `$dbHost`, `$dbIsIpv6`, `$dbSocket`, `$dbPort`, `$dbName`
    - `$dbUsername`, `$dbPassword`, `$dbCharset`, `$dbCollation`
    - `$dbTimezoneOffset`, `$dbCharsetCollate`, `$wpDbPrefix`
  - **Kept only** `$dbPrefix` (WP prefix + plugin prefix)
  - **Simplified** `Env::init()` signature from 6 parameters to 2
  - **Replaced cached values** with direct `$wpdb` calls:
    - `Env::$dbCharsetCollate` → `$wpdb->get_charset_collate()`
    - `Env::$wpDbPrefix` → `$wpdb->prefix`
    - `DB_NAME` constant → `$wpdb->dbname`
  - **Updated tests** to use `$wpdb` properties instead of WordPress constants

  ### Impact:
  - 📊 **92

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7361-cleanup-env-class-from-db-related-constants)

_The latest successful build from `stomail-7361-cleanup-env-class-from-db-related-constants` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Refactor**
  * Database setup now uses WordPress's native DB settings; legacy per-parameter DB configuration was removed and related properties marked deprecated.

* **Chores**
  * Removed unused DB initialization parameters and simplified initialization flow.

* **Tests**
  * Updated integration tests to rely on WordPress runtime DB values and adjusted deprecated-property expectations.

* **Documentation**
  * Added changelog entry noting removal of unused DB-related parameters.

### DIFF
--- a/mailpoet/changelog/2025-11-21-10-28-36-removed-remove-unused-database-related-parameters-from-env.md
+++ b/mailpoet/changelog/2025-11-21-10-28-36-removed-remove-unused-database-related-parameters-from-env.md
@@ -1,0 +1,5 @@
+# Type: Removed
+
+# Description
+
+Remove unused database-related parameters from Env class after WPDB migration
--- a/mailpoet/lib/Config/Env.php
+++ b/mailpoet/lib/Config/Env.php
@@ -25,6 +25,54 @@ class Env {
   public static $pluginPrefix;
   /** @var string WP DB prefix + plugin prefix */
   public static $dbPrefix;
+  /**
+   * @deprecated Use global $wpdb->prefix instead
+   */
+  public static $wpDbPrefix = '';
+  /**
+   * @deprecated Database connection is handled by WordPress $wpdb
+   */
+  public static $dbHost = '';
+  /**
+   * @deprecated Database connection is handled by WordPress $wpdb
+   */
+  public static $dbIsIpv6 = '';
+  /**
+   * @deprecated Database connection is handled by WordPress $wpdb
+   */
+  public static $dbSocket = '';
+  /**
+   * @deprecated Database connection is handled by WordPress $wpdb
+   */
+  public static $dbPort = '';
+  /**
+   * @deprecated Use global $wpdb->dbname instead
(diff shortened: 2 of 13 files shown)


costasovo (commented):
@lysyjan Thanks for cleaning up the Env class. The plugin works as expected.

I have a concern that third-party plugins might be using these props. The thing is that accessing a non-existing public static property triggers a Fatal error.

My tip is that there is very little chance that someone is using those, but the risk is still here, and we learned that the WP ecosystem is full of surprises...

Should we address the concern? Maybe we could keep the empty definitions with the `@deprecated` annotation.

lysyjan:
Thanks for your suggestion @costasovo.

What do you think about using `@deprecated` for the whole class and stop using it? I think the usage of the last property can be removed and we could stop using the Env class.

costasovo:
> What do you think about using @deprecated for the whole class and stop using it? I think the usage of the last property can be removed and we could stop using the Env class.

@lysyjan Do you mean restoring the removed properties with empty values and marking the whole class deprecated? If so, then yes, I agree.

lysyjan:
@costasovo I set default values to the empty string and annotated the unused method and properties as deprecated.

costasovo (approved):
@lysyjan Thank you for the update and cleaning up all the usage! It is safer to keep the props for BC.

I did some basic smoke testing in the admin and didn't experienced any issues.

PR #6429, opened 2025-12-10
Fix race condition causing cron action to persist after deactivation

pavel-mailpoet:
## Description

Fix race condition where the `mailpoet/cron/daemon-trigger` Action Scheduler action can persist after plugin deactivation. 

**Problem:** When the MailPoet plugin is deactivated, a parallel request (e.g., opening the MailPoet emails page right before clicking deactivate) can re-initialise the daemon trigger action after unscheduleAllCronActions() has run. This keeps the action in a pending state and causes it to fail repeatedly because the plugin is no longer active. 

**Solution:** Introduce a deactivation flag 

## QA notes

Please verify the fix prevents orphaned cron actions

1. Activate the MailPoet plugin and set the cron method to "WordPress built-in cron (recommended)" in MailPoet > Settings > Advanced
2. Open the installed plugins page
3. Open the MailPoet Emails page in a new tab (from the sidebar menu) and immediately click deactivate MailPoet plugin
4. Go to Tools > Scheduled Actions and check for mailpoet/cron/daemon-trigger
5. Activate MailPoet plugin
6. Ensure cron method is set to "WordPress built-in cron (recommended)" in MailPoet > Settings > Advanced
7. Go to Tools > Scheduled Actions and verify mailpoet/cron/daemon-trigger is scheduled (status: pending)

Also do some smoke tests around scheduled newsletters and automations, that nothing is broken

## Linked PRs

_N/A_

## Linked tickets

STOMAIL-7283

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7283-action-scheduler-action-mailpoetcrondaemon-trigger-may)

_The latest successful build from `stomail-7283-action-scheduler-action-mailpoetcrondaemon-trigger-may` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition that could leave a scheduled daemon action active during plugin deactivation, improving reliability during deactivate/reactivate cycles.
* **New Features**
  * Added a deactivation flag mechanism to prevent scheduling while deactivation is in progress and a way to clear that flag.
* **Tests**
  * Added integration tests verifying the deactivation flag lifecycle and that triggers are not scheduled when deactivating.
* **Changelog**
  * Added an entry documenting the race‑condition fix.

### DIFF
--- a/mailpoet/lib/Cron/ActionScheduler/Actions/DaemonTrigger.php
+++ b/mailpoet/lib/Cron/ActionScheduler/Actions/DaemonTrigger.php
@@ -39,12 +39,13 @@ class DaemonTrigger {
   public function init() {
     $this->wp->addAction(self::NAME, [$this, 'process']);
 
-    // Don't schedule if plugin is being deactivated (prevents race condition)
-    if ($this->wp->getOption(DaemonActionSchedulerRunner::DEACTIVATION_FLAG_OPTION, false)) {
-      return;
-    }
-
     if (!$this->actionScheduler->hasScheduledAction(self::NAME)) {
+      // Don't schedule if plugin is being deactivated (prevents race condition with parallel requests)
+      // Note: We check the option directly instead of using DaemonActionSchedulerRunner::isDeactivating()
+      // to avoid a circular dependency (DaemonActionSchedulerRunner depends on DaemonTrigger)
+      if ($this->wp->getOption(DaemonActionSchedulerRunner::DEACTIVATION_FLAG_OPTION, false)) {
+        return;
+      }
       $this->actionScheduler->scheduleRecurringAction($this->wp->currentTime('timestamp', true), self::TRIGGER_RUN_INTERVAL, self::NAME);
     }
   }
--- a/mailpoet/lib/Cron/ActionScheduler/Actions/DaemonTrigger.php
+++ b/mailpoet/lib/Cron/ActionScheduler/Actions/DaemonTrigger.php
@@ -4,6 +4,7 @@ namespace MailPoet\Cron\ActionScheduler\Actions;
 
 use MailPoet\Cron\ActionScheduler\ActionScheduler;
 use MailPoet\Cron\ActionScheduler\RemoteExecutorHandler;
+use MailPoet\Cron\DaemonActionSchedulerRunner;
 use MailPoet\Cron\Triggers\WordPress;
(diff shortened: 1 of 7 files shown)


mailpoet/lib/Cron/ActionScheduler/Actions/DaemonTrigger.php
```diff
@@ -37,6 +38,12 @@ public function __construct(
 
   public function init() {
     $this->wp->addAction(self::NAME, [$this, 'process']);
+
+    // Don't schedule if plugin is being deactivated (prevents race condition)
+    if ($this->wp->getOption(DaemonActionSchedulerRunner::DEACTIVATION_FLAG_OPTION, false)) {
+      return;
+    }
```
costasovo: AFAIK this init method is called on every request and normally the option `DaemonActionSchedulerRunner::DEACTIVATION_FLAG_OPTION` will not be set. This means it will trigger an SQL query to the wp_options table on every request.
The query is probably super fast, but I think we could optimize that.

I think we can move this check into the following if block.
`if (!$this->actionScheduler->hasScheduledAction(self::NAME)) {`
then we would be fetching the option only when we actually need to add the action.

Also could we use `DaemonActionSchedulerRunner::isDeactivating` ?

costasovo (commented):
Thank you for looking into this issue! I have a suggestiong for an optimization. Please see my comment.

mailpoet/lib/Cron/ActionScheduler/Actions/DaemonTrigger.php
pavel-mailpoet: I've moved the check inside the `if (!$this->actionScheduler->hasScheduledAction(self::NAME))` block, so the option is only fetched when we're about to schedule a new action, not on every request. 

Regarding `DaemonActionSchedulerRunner::isDeactivating()` - unfortunately, we can't use it because it would create a circular dependency:
`DaemonActionSchedulerRunner` depends on `DaemonTrigger` (it's injected in the constructor)
If `DaemonTrigger` depended on `DaemonActionSchedulerRunner`, we'd have a circular dependency. I've added a comment in the code explaining this.

costasovo (approved):
Looks good! Thank you!

Aschepikov (approved):
GJ!

PR #6454, opened 2026-01-06
Send monthly stats through all the day, switch to WordPress timezone

websupporter:
## Description
It used to be that the monthly stats emails is send out at the first Monday of a month at 12:00. We switched to a bigger time window in #6452.

Now we decided to extend the time range in which we send the emails to the whole day. Additionally we switch to the WordPress timezone.

## Code review notes

_N/A_

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:mpi-171-randomize-mailpoet-monthly-stats-send-window)

_The latest successful build from `mpi-171-randomize-mailpoet-monthly-stats-send-window` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Improvements**
  * Monthly stats emails are now scheduled throughout the entire day using the site timezone, improving delivery distribution and randomness.
* **Tests**
  * Time-handling tests updated to reflect timezone-aware scheduling and the expanded full-day randomization window.

### DIFF
--- a/mailpoet/changelog/2026-01-06-10-41-44-improved-send-monthly-stats-emails-throughout-the-day-using.md
+++ b/mailpoet/changelog/2026-01-06-10-41-44-improved-send-monthly-stats-emails-throughout-the-day-using.md
@@ -1,0 +1,5 @@
+# Type: Improved
+
+# Description
+
+Send monthly stats emails throughout the day using site timezone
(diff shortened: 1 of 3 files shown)


websupporter:
> The WpDateTime dependency is instantiated directly in the constructor, while all other dependencies (MailerFactory, Renderer, SettingsController, etc.) are properly injected as constructor parameters. This inconsistency makes testing harder since you cannot easily provide a mock WpDateTime instance during tests.

Following precedent. Another way could be to make WpDateTime injectable and refactor the other classes. Wdyt?

costasovo (approved):
Thank you for the improvement!

I tested it locally and the following stats sending was scheduled correctly.

Image

PR #6458, opened 2026-01-12
Fetch total recipients count in the email editor from the API

yuliyan:
## Description

Fixes an issue in the email editor where **Total Recipients** shows all subscribers (including Unsubscribed/Unconfirmed) and can overcount due to duplicates.

Now uses the segments admin AJAX `subscriberCount` endpoint to return an accurate, de-duplicated count of **subscribed** recipients only.

#### Screenshot:
Qjr9MMhAOruKgywb

## Code review notes

_N/A_

## QA notes

* Create a new email using the email editor from MailPoet → Emails → Add new email → **Create using the new email editor (Alpha)**.
* Open the **Recipients** dropdown from the sidebar and select lists/segments with mixed subscriber statuses.
* Verify **Total recipients** matches actual send count.
* Save the email and navigate to the `Review & schedule` to verify the `Estimated recipients:` count is matching.

## Linked PRs

_N/A_

## Linked tickets

[STOMAIL-7685: In new block-based email editor, "total recipients" count shows all statuses](https://linear.app/a8c/issue/STOMAIL-7685/in-new-block-based-email-editor-total-recipients-count-shows-all)

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7685-in-new-block-based-email-editor-total-recipients-count-shows)

_The latest successful build from `stomail-7685-in-new-block-based-email-editor-total-recipients-count-shows` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed total recipients count in the email editor — count now updates dynamically for selected segments and displays a loading spinner while fetching.

* **Style**
  * Adjusted recipient loader alignment so the loading indicator displays correctly next to the total count.

### DIFF
--- a/mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/recipients-row.tsx
+++ b/mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/recipients-row.tsx
@@ -55,6 +55,8 @@ export function RecipientsRow() {
   const [recipientType, setRecipientType] = useState<
     'all_customers' | 'segment'
   >('segment');
+  const [recipientCount, setRecipientCount] = useState<number | null>(null);
+  const [isLoadingRecipientCount, setIsLoadingRecipientCount] = useState(false);
 
   // Fetch segments on mount.
   useEffect(() => {
@@ -125,6 +127,46 @@ export function RecipientsRow() {
     }
   }, [selectedSegmentIds, segments]);
 
+  // Fetch the subscriber count when selected segments change.
+  useEffect(() => {
+    if (recipientType !== 'segment' || selectedSegmentIds.length === 0) {
+      setRecipientCount(null);
+      setIsLoadingRecipientCount(false);
+      return undefined;
+    }
+
+    let mounted = true;
+    setIsLoadingRecipientCount(true);
+
+    void MailPoet.Ajax.post({
+      api_version: window.mailpoet_api_version,
+      endpoint: 'segments',
+      action: 'subscriberCount',
+      data: {
+        segmentIds: selectedSegmentIds,
+      },
+    })
+      .then((response) => {
+        if (mounted) {
+          setRecipientCount(response?.data?.count as number);
+        }
+      })
+      .catch(() => {
+        if (mounted) {
+          setRecipientCount(null);
+        }
+      })
+      .always(() => {
+        if (mounted) {
+          setIsLoadingRecipientCount(false);
+        }
+      });
+
+    return () => {
+      mounted = false;
+    };
+  }, [recipientType, selectedSegmentIds]);
+
   const updateEmailMailPoetProperty = (name: string, value: unknown) => {
     const postId = select(editorStore).getCurrentPostId();
     const currentPostType = 'mailpoet_email';
@@ -186,7 +228,7 @@ export function RecipientsRow() {
     (segment) => segment.type === 'woocommerce_users',
   );
   const allCustomersSegmentCount = parseInt(
-    allCustomersSegment?.subscribers_count?.all || '0',
+    allCustomersSegment?.subscribers_count?.subscribed || '0',
     10,
   );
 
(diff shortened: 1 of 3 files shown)


mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/recipients-row.tsx
```diff
@@ -319,7 +352,14 @@ export function RecipientsRow() {
                         />
                         <div className="mailpoet-status-panel__recipients-total-count">
                           {__('Total recipients: ', 'mailpoet')}{' '}
-                          {recipientCount.toLocaleString()}
+                          {isLoadingRecipientCount ? (
+                            <Spinner
+                              className="mailpoet-status-panel__recipients-loader"
+                              style={{ margin: '0' }}
```
eason9487: This `margin` could be moved to the [newly added style](https://github.com/mailpoet/mailpoet/blob/3f5e147ca29f688cb444467dd611d1f6d096af2e/mailpoet/assets/js/src/mailpoet-email-editor-integration/integration.scss#L53-L55) using higher specificity, for example: `&__recipients-loader.components-spinner`.

eason9487 (commented):
Thanks for the work! The total count of the selected segments is correct.

Before proceeding, I would like to check if the same approach used for selected segments should also be applied to "Send to all customers".

If I understand correctly, selecting "Send to all customers" will use the segment shown in the screenshot below.



In my WooCommerce Customers list, only one customer has the Subscribed status. However, using `subscribers_count?.all` here results in 2 being displayed.

https://github.com/mailpoet/mailpoet/blob/3f5e147ca29f688cb444467dd611d1f6d096af2e/mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/recipients-row.tsx#L211-L232

The newly added AJAX query will still be triggered as well when "Send to all customers" is selected, since `selectedSegmentIds` includes the WooCommerce Customers segment. However, the query result is not used—the component still relies on the `subscribers_count?.all` field instead.

This results in a redundant API call. Moreover, as shown in the screenshots below, the AJAX query result matches `subscribers_count.subscribed`, which would be the correct value to display.





---

Therefore, I would like to confirm:
1. Does the current "Send to all customers" option have a similar issue that needs to be addressed?
2. If the above issue is confirmed, at least two approaches seem viable. What do you think?
   1. Use the `subscribers_count.subscribed` field for "Send to all customers" and skip the AJAX query in this case
   2. Use the newly added AJAX query result for all scenarios, including "Send to all customers"

mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/recipients-row.tsx
yuliyan: Thanks for the suggestion - I've updated the PR. There was indeed a conflict with the component's styles requiring a higher specificity. However, I decided to use a nested selected of a compound selector with the component's `.components-spinner`, as it seemed like something that could possibly change upstream.

yuliyan:
@eason9487, thanks for catching the issues with the `all_customers` option. I have updated the label to use the subscribed count and updated the segment fetch to be limited only to when the `segment` option is selected.

mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/recipients-row.tsx
```diff
@@ -125,6 +127,45 @@ export function RecipientsRow() {
     }
   }, [selectedSegmentIds, segments]);
 
+  // Fetch the subscriber count when selected segments change.
+  useEffect(() => {
+    if (recipientType !== 'segment' || selectedSegmentIds.length === 0) {
+      setRecipientCount(null);
+      return undefined;
+    }
```
eason9487, line 136: Noticed an issue with the loading state when quickly removing all selected segments. When `selectedSegmentIds` becomes empty, this early return won't reset the loading state. If a previous AJAX request remains unfinished at this point, the UI gets stuck in an infinite loading state.

Steps to reproduce:
1. Select multiple segments
2. Quickly remove all selected segments using the backspace key

https://github.com/user-attachments/assets/017dfc01-4f3d-473b-bbeb-ef072ee52703

eason9487 (approved):
Thanks for addressing the feedback! The implementation now correctly handles the recipient count for both "Send to a segment" and "Send to all customers" options.

There's one minor issue remaining with the loading state, but it looks straightforward to fix. I'll go ahead and approve. Feel free to merge once that's addressed.

mailpoet/assets/js/src/mailpoet-email-editor-integration/sidebar/components/recipients-row.tsx
yuliyan: The issue was due to not resetting the loading state when `selectedSegmentIds.length === 0`. Fixed it and rebased branch to avoid generating noise with multiple commits.

PR #6467, opened 2026-01-21
Configure acceptance tests and developers environment to run on PHP 8.4

costasovo:
## Description

Updates the acceptance test environment to run on PHP 8.4 containers, along with necessary dependency updates and compatibility patches.

### Changes

- **Acceptance Tests**: Updated WordPress and Codeception containers to use PHP 8.4 images
- **Codeception Patches**: Added patch for PHP 8.4 deprecated `E_STRICT` constant in Codeception
- **Composer**: Updated to version 2.9.3 for improved PHP 8.4 compatibility
- **Tracy Debugger**: Added multi-version support (legacy for PHP 7.4-8.3, current for PHP 8.4+)
- **Dev Environment**: Updated default Docker configuration to PHP 8.4
- **CI/CD**: Added PHP version reporting step in CircleCI acceptance test pipeline

Related to STOMAIL-7697

### Test Plan

- Run acceptance tests and verify they pass on PHP 8.4
- Confirm Codeception patches apply successfully during `composer install`
- To test the developer environment, you need to rebuild the WordPress container. `docker compose build wordpress`

## Code review notes

I skipped the changelog because this is not a user-facing change. I think we can also skip QA for this PR.

## QA notes

_N/A_

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7697-compatibility-with-php-84)

_The latest successful build from `stomail-7697-compatibility-with-php-84` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Bug Fixes**
  * Added PHP 8.4 compatibility patches and error handling fixes for the testing environment to ensure proper test execution.

* **Chores**
  * Updated all development and testing infrastructure to PHP 8.4.
  * Updated Docker images to PHP 8.4-compatible versions.
  * Upgraded Composer to version 2.9.3 for improved stability.
  * Simplified PHP version compatibility handling across development tools.

### DIFF
--- a/mailpoet/tools/install.php
+++ b/mailpoet/tools/install.php
@@ -2,7 +2,7 @@
 
 
 // Tools versions for PHP 7.4+
-$composerVersion = '2.7.7';
+$composerVersion = '2.9.3';
 $phpScoperVersion = '0.17.2';
 $legacyTracyVersion = '2.9.4'; // Tracy 2.9.4 supports PHP 7.4
 $tracyVersion = '2.11.0'; // Tracy 2.11.0 supports PHP 8.4 and 8.5
--- a/mailpoet/tools/install.php
+++ b/mailpoet/tools/install.php
@@ -4,23 +4,13 @@
 // Tools versions for PHP 7.4+
 $composerVersion = '2.7.7';
 $phpScoperVersion = '0.17.2';
-$tracyVersion = '2.9.4';
-
-// Tools versions for PHP 8.0+
-if (PHP_VERSION_ID >= 80000) {
-  $tracyVersion = '2.9.7'; // Tracy 2.9.4 doesn't support PHP 7.4
-}
-
-// Tools versions for PHP 8.4+
-if (PHP_VERSION_ID >= 80400) {
-  $composerVersion = '2.8.10';
-  $phpScoperVersion = '0.18.17';
-  $tracyVersion = '2.10.9'; // 2.10.10 causes SyntaxError in JS
-}
+$legacyTracyVersion = '2.9.4'; // Tracy 2.9.4 supports PHP 7.4
+$tracyVersion = '2.11.0'; // Tracy 2.11.0 supports PHP 8.4 and 8.5
 
 $tools = [
   "https://github.com/composer/composer/releases/download/$composerVersion/composer.phar" => 'composer.phar',
   "https://github.com/humbug/php-scoper/releases/download/$phpScoperVersion/php-scoper.phar" => 'php-scoper.phar',
+  "https://github.com/nette/tracy/releases/download/v$legacyTracyVersion/tracy.phar" => 'tracy-legacy.phar',
   "https://github.com/nette/tracy/releases/download/v$tracyVersion/tracy.phar" => 'tracy.phar',
 ];
 // ensure installation in dev-mode only
--- a/tests_env/composer.json
+++ b/tests_env/composer.json
@@ -22,11 +22,13 @@
     "post-install-cmd": [
       "php ./tasks/fix-php82-codeception.php",
       "php ./tasks/fix-php83-codeception.php",
+      "php ./tasks/fix-php84-e-strict-codeception.php",
       "php ./tasks/FixPhp84Deprecations.php"
     ],
     "post-update-cmd": [
       "php ./tasks/fix-php82-codeception.php",
       "php ./tasks/fix-php83-codeception.php",
+      "php ./tasks/fix-php84-e-strict-codeception.php",
       "php ./tasks/FixPhp84Deprecations.php"
     ]
   }
--- a/tests_env/tasks/fix-php84-e-strict-codeception.php
+++ b/tests_env/tasks/fix-php84-e-strict-codeception.php
@@ -1,0 +1,85 @@
+<?php
+
+// throw exception if anything fails
+set_error_handler(function ($severity, $message, $file, $line) {
+  throw new ErrorException($message, 0, $severity, $file, $line);
+});
+
+// Fixes for PHP8.4 Compatibility - E_STRICT constant is deprecated
+
+// Replacement for ErrorHandler.php line 46
+$errorHandlerConstructorFind = <<<'CODE'
+    public function __construct()
+    {
+        $this->errorLevel = E_ALL & ~E_STRICT & ~E_DEPRECATED;
+    }
+CODE;
+
+$errorHandlerConstructorReplace = <<<'CODE'
+    public function __construct()
+    {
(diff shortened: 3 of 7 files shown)


tests_env/tasks/fix-php84-e-strict-codeception.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->errorLevel = E_ALL & ~E_STRICT & ~E_DEPRECATED;
+    }
+CODE;
+
+$errorHandlerConstructorReplace = <<<'CODE'
+    public function __construct()
+    {
+      $this->errorLevel = PHP_VERSION_ID >= 80400 ? (E_ALL & ~E_DEPRECATED) : (E_ALL & ~E_STRICT & ~E_DEPRECATED);
+    }
+CODE;
```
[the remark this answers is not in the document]
costasovo: It is in the vendor folder. So this small formatting problem can be ignored.

tests_env/composer.json
```diff
@@ -22,11 +22,13 @@
     "post-install-cmd": [
       "php ./tasks/fix-php82-codeception.php",
       "php ./tasks/fix-php83-codeception.php",
+      "php ./tasks/fix-php84-e-strict-codeception.php",
```
triple0t, line 25: Instead of patching it, what about updating the package to the latest version?
This error was fixed in [5.2.0](https://github.com/Codeception/Codeception/releases/tag/5.2.0), for example.

tests_env/tasks/fix-php84-e-strict-codeception.php
triple0t, line 1: This file might not be necessary if we update codeception to the most recent version https://packagist.org/packages/codeception/codeception

mailpoet/tools/install.php
```diff
@@ -2,25 +2,15 @@
...
-}
-
-// Tools versions for PHP 8.4+
-if (PHP_VERSION_ID >= 80400) {
-  $composerVersion = '2.8.10';
-  $phpScoperVersion = '0.18.17';
-  $tracyVersion = '2.10.9'; // 2.10.10 causes SyntaxError in JS
-}
+$legacyTracyVersion = '2.9.4'; // Tracy 2.9.4 supports PHP 7.4
+$tracyVersion = '2.11.0'; // Tracy 2.11.0 supports PHP 8.4 and 8.5
```
triple0t, line 8: Since you're making the same check in `mailpoet/mailpoet_initializer.php`, can you please explain why you implemented this change?

It has several disadvantages, from what I can see
1. You're forcing all users below PHP 8.4 to use an outdated version of Tracy (2.9.4) instead of the current approach, which provides an opportunity to use v2.9.7 for users with PHP 8
2. You're not updating the phpScoper version
3. Would using the latest version of Compoer (v2.9.3) cause conflict with the legacy version of  Tracy (v2.9.4)?

triple0t (commented):
Well done.
I have a few questions and suggestions. Please locate below.

tests_env/composer.json
costasovo: @triple0t I was looking into updating it, but the 5.2.0 requires PHP 8.1, and we still need to support 7.4. With 5.2.0 won't be able to run tests on PHP 7.4. I was also thinking about having two versions, but the patch seemed to be the less complex solution.

mailpoet/tools/install.php
costasovo: > Since you're making the same check in mailpoet/mailpoet_initializer.php, can you please explain why you implemented this change?

@triple0t  The idea is to download all versions of Tracy we use so that:
1) we don't have to reinstall dependencies every time we switch PHP version
2) On CI, the build job, which downloads tools, may run on a different PHP version than the test. 
With this change, I basically wanted to move this check to the `mailpoet/mailpoet_initializer.php` so that we are able to pick the correct version at runtime.

The composer was updated, because there were PHP Warnings on PHP 8.4.

> You're forcing all users below PHP 8.4 to use an outdated version of Tracy (2.9.4) instead of the current approach, which provides an opportunity to use v2.9.7 for users with PHP 8. 

I see I missed the PHP version between PHP 8.0 - PHP 8.4. I would fix that by downloading one more Tracy version. Would you agree?

> 2. You're not updating the PHPScoper version

This was not causing any errors, so I didn't need to update it. I would leave it for the update deps milestone.

> 3. Would using the latest version of Compoer (v2.9.3) cause conflict with the legacy version of Tracy (v2.9.4)?

I don't think this could cause a conflict. AFAIK, they are not used together.

tests_env/tasks/fix-php84-e-strict-codeception.php
costasovo: Same as the comment above. The 5.2.0 requires PHP 8.1, and we still need to support 7.4. With 5.2.0 won't be able to run tests on PHP 7.4.

tests_env/composer.json
triple0t: You're right. 5.2.0 does [require](https://github.com/Codeception/Codeception/blob/5.2.0/composer.json#L18) PHP 8.1. Plus, we can't bump the PHP version because we still need to run tests against PHP 7.4. This is a headache.

tests_env/tasks/fix-php84-e-strict-codeception.php
triple0t: Noted.

mailpoet/tools/install.php
triple0t: Noted. 

> I see I missed the PHP version between PHP 8.0 - PHP 8.4. I would fix that by downloading one more Tracy version. Would you agree?

That's fine with me.

tests_env/composer.json
costasovo: Yes. We need to drop 7.4 first. Ideally, together with 8.0

mailpoet/tools/install.php
costasovo: @triple0t I addressed this in https://github.com/mailpoet/mailpoet/pull/6467/commits/db4dfd87b30b633ade2e511e291ce974417a1acf.
Instead of adding a third version of Tracy, I updated the runtime check to use the newest version with PHP 8.2+ (the lowest it supports).
The legacy version 2.9.4 supports PHP 8.2 and lower, so for PHP 8.1, 8.0, and 7.4, it should be fine, and for newer PHP, we will have the latest version.

triple0t (approved):
Looks good to me.
Well done

PR #6487, opened 2026-01-30
Fix negative date errors when using large days values in segment filters

yuliyan:
## Description

When users set the `days` parameter to a very large number in dynamic segment filters, `Carbon::now()->subDays()` produces negative dates (e.g., year -1000). Depending on the database engine and the SQL mode, using negative dates in the queries can be rejected with errors.

This fix adds `getDateNDaysAgo()` and `getDateNDaysAgoImmutable()` helper methods to `FilterHelper` that clamp dates to a minimum of 1000-01-01 (MySQL's minimum valid DATE/DATETIME value).

Updated 12 filter files to use the new helper:
- WooCommerceNumberOfOrders
- WooCommerceUsedCouponCode
- WooCommerceTotalSpent
- WooCommerceSingleOrderValue
- WooCommerceNumberOfReviews
- WooCommerceAverageSpent
- WooCommerceUsedShippingMethod
- WooCommerceUsedPaymentMethod
- NumberOfClicks
- EmailsReceived
- EmailOpensAbsoluteCountAction
- DateFilterHelper

## Code review notes

To reproduce this error locally, I switched my local container to use `mysql:8.0` in `docker-compose.override.yml`:

```yml
services:
  db:
    image: mysql:8.0
    command: --sql_mode=STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION,ANSI,ONLY_FULL_GROUP_BY
    volumes:
      - mysql8-datavolume:/var/lib/mysql

volumes:
  mysql8-datavolume:
```

On creating a dynamic segment, e.g. `Repeat Buyers`, setting the `number of orders` to `in the last 999999` days, it is expected to get an error due to using an Invalid date in the query.

## QA notes

1. Navigate to MailPoet → Segments → Add new segment
2. Select any of the updated filters for the condition and confirm that after updating the number of days, the request to update the subscribers count is successful and updates the count as expected.
3. Set the `days` parameter to a large number, e.g. `999999`, and confirm the request doesn't throw any errors.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7773-date-error-after-updating-to-recent-version-of-the-plugin)

_The latest successful build from `stomail-7773-date-error-after-updating-to-recent-version-of-the-plugin` will be used. If none is available, the link won't work._

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed negative date errors when using very large day values in segment filters.

* **Tests**
  * Added tests to verify correct behavior and clamping for very large day values in date-based segment filters.

* **Refactor**
  * Centralized date calculation logic to make time-based filters more robust and consistent.

* **Changelog**
  * Added a release note entry documenting the fix.

### DIFF
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/DateFilterHelper.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/DateFilterHelper.php
@@ -16,6 +16,15 @@ class DateFilterHelper {
   const IN_THE_LAST = 'inTheLast';
   const NOT_IN_THE_LAST = 'notInTheLast';
 
+  /** @var FilterHelper */
+  private $filterHelper;
+
+  public function __construct(
+    FilterHelper $filterHelper
+  ) {
+    $this->filterHelper = $filterHelper;
+  }
+
   public function getValidOperators(): array {
     return array_merge(
       $this->getAbsoluteDateOperators(),
@@ -48,7 +57,7 @@ class DateFilterHelper {
         throw new InvalidFilterException('Invalid date value', InvalidFilterException::INVALID_DATE_VALUE);
       }
     } else if (in_array($operator, self::getRelativeDateOperators())) {
-      $carbon = CarbonImmutable::now()->subDays(intval($value) - 1);
+      $carbon = $this->filterHelper->getDateNDaysAgoImmutable(intval($value) - 1);
     } else {
       throw new InvalidFilterException('Incorrect value for operator', InvalidFilterException::MISSING_VALUE);
     }
--- a/mailpoet/lib/Segments/DynamicSegments/Filters/FilterHelper.php
+++ b/mailpoet/lib/Segments/DynamicSegments/Filters/FilterHelper.php
@@ -6,10 +6,18 @@ use MailPoet\Entities\DynamicSegmentFilterData;
 use MailPoet\Entities\SubscriberEntity;
 use MailPoet\Segments\DynamicSegments\Exceptions\InvalidFilterException;
 use MailPoet\Util\Security;
+use MailPoetVendor\Carbon\Carbon;
+use MailPoetVendor\Carbon\CarbonImmutable;
 use MailPoetVendor\Doctrine\DBAL\Query\QueryBuilder;
 use MailPoetVendor\Doctrine\ORM\EntityManager;
 
 class FilterHelper {
+  /**
+   * Minimum valid year for MySQL DATE/DATETIME fields.
+   * Used to clamp dates when subtracting large day values to prevent negative dates.
+   */
+  private const MIN_DATE_YEAR = 1000;
+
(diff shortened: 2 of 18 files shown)


mailpoet/lib/Segments/DynamicSegments/Filters/DateFilterHelper.php
```diff
@@ -48,7 +48,7 @@ public function getDateStringForOperator(string $operator, string $value): strin
         throw new InvalidFilterException('Invalid date value', InvalidFilterException::INVALID_DATE_VALUE);
       }
     } else if (in_array($operator, self::getRelativeDateOperators())) {
-      $carbon = CarbonImmutable::now()->subDays(intval($value) - 1);
+      $carbon = CarbonImmutable::now()->subDays(intval($value) - 1)->max(CarbonImmutable::createFromDate(1000, 1, 1));
```
lysyjan: I noticed that we could reuse here `FilterHelper::getDateNDaysAgoImmutable()`. What do you think?

mailpoet/lib/Segments/DynamicSegments/Filters/FilterHelper.php
```diff
@@ -81,4 +83,28 @@ public function validateDaysPeriodData(array $data): void {
...
+  /**
+   * Get a date by subtracting days from now, clamped to a minimum valid date.
+   * This prevents negative dates when users set very large day values,
+   * which can cause errors on some database engines like MySQL.
+   *
+   * @param int $days Number of days to subtract from current date
+   * @return Carbon The calculated date, clamped to minimum 1000-01-01
+   */
+  public function getDateNDaysAgo(int $days): Carbon {
+    return Carbon::now()->subDays($days)->max(Carbon::createFromDate(1000, 1, 1));
```
lysyjan: The minimum year 1000 appears in multiple places. Could be a class constant like `private const MIN_DATE_YEAR = 1000`

lysyjan (approved):
Good job 👍 I have just two tiny suggestions.

mailpoet/lib/Segments/DynamicSegments/Filters/DateFilterHelper.php
yuliyan: Ah, missed that - thanks for catching it! I've updated the commit.

mailpoet/lib/Segments/DynamicSegments/Filters/FilterHelper.php
yuliyan: Thanks for the suggestion, I've added the constant.

Aschepikov (approved):
GJ!

PR #6532, opened 2026-02-24
Fix release:prepare to skip prerelease builds

lysyjan:
## Description

The `release:prepare` command fails when a prerelease (e.g. "Trunk Snapshot Build") is the most recent GitHub release. The `getLastReleasedVersion()` method had three bugs:

1. **`per_page=1`** — Only fetched 1 release. If the newest is a prerelease with a non-semver tag, `$validReleases` becomes empty and throws an exception.
2. **No prerelease filtering** — The filter only checked version format but didn't exclude prereleases.
3. **Wrong array after filtering** — Used `reset($releases)` instead of `reset($validReleases)`, so even with successful filtering, the unfiltered entry would be returned.

## Code review notes

- `per_page` increased to 20 to look past recent prereleases
- Added `!$release['prerelease']` check to filter
- Fixed `reset()` to use `$validReleases`

## QA notes

Run `./do release:version:get:next` — it should return a valid `X.Y.Z` version and not fail.

## Linked PRs

_N/A_

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:fix/release-prepare-trunk-snapshot)

_The latest successful build from `fix/release-prepare-trunk-snapshot` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/tasks/release/GitHubController.php
+++ b/mailpoet/tasks/release/GitHubController.php
@@ -279,7 +279,7 @@ class GitHubController {
   public function getLastReleasedVersion(): ?string {
     $response = $this->httpClient->get('releases', [
       'query' => [
-        'per_page' => 1,
+        'per_page' => 20,
       ],
     ]);
     $releases = json_decode($response->getBody()->getContents(), true);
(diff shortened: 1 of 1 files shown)


pavel-mailpoet (approved):

pavel-mailpoet:
Thank you for fixing it, merging

PR #6538, opened 2026-02-25
Self-heal missing WP capabilities on admin requests

lysyjan:
## Description

MailPoet capabilities registered via `register_activation_hook` can be lost in environments that bypass activation hooks (e.g., plugins loaded via `include_once`) or due to race conditions. The existing `maybeRunActivator()` fallback only triggers on version changes (`db_version` mismatch) and won't detect or fix lost capabilities.

This adds a self-healing mechanism to `Capabilities::init()` that verifies all default permissions are present on the correct roles during admin requests, and restores any that are missing. The check is throttled by a 1-hour transient to avoid redundant work.

The existing `mailpoet_permission_*` filters in `AccessControl::getDefaultPermissions()` allow hosting environments to register additional custom roles, which are then automatically covered by the self-healing.

## Code review notes

- `ensureWPCapabilities()` is a new private method on `Capabilities`, called from `init()`
- Guards: `is_admin()` + transient short-circuit
- Only adds missing capabilities — does not override explicitly denied ones (`false`)
- Caches role objects to avoid redundant `getRole()` lookups (same pattern as `setupWPCapabilities()`)
- No changes to `Activator`, `setupWPCapabilities()`, or `removeWPCapabilities()`

## QA notes

1. Remove MailPoet capabilities:
   ```bash
   wp cap remove administrator mailpoet_access_plugin_admin mailpoet_manage_automations
   ```
2. Clear transient: `wp transient delete mailpoet_caps_verified`
3. Visit any wp-admin page
4. Verify caps restored: `wp cap list administrator | grep mailpoet`
5. Remove caps again — this time do NOT clear the transient
6. Visit wp-admin — caps should NOT be restored (transient still active, proving throttle works)

## Linked PRs

_N/A_

## Linked tickets

[STOMAIL-7847](https://linear.app/a8c/issue/STOMAIL-7847)

## After-merge notes

_N/A_

## Screenshots or screencast 

_N/A_

## Tasks

### DIFF
--- a/mailpoet/lib/Config/Capabilities.php
+++ b/mailpoet/lib/Config/Capabilities.php
@@ -7,6 +7,7 @@ use WP_Role;
 
 class Capabilities {
   const MEMBERS_CAP_GROUP_NAME = 'mailpoet';
+  const TRANSIENT_CAPS_VERIFIED = 'mailpoet_caps_verified';
 
   private $renderer = null;
   /** @var WPFunctions  */
@@ -30,6 +31,7 @@ class Capabilities {
 
   public function init() {
     $this->setupMembersCapabilities();
+    $this->ensureWPCapabilities();
   }
 
   public function setupWPCapabilities() {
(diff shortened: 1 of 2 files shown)


pavel-mailpoet (approved):
Works as described

pavel-mailpoet (approved):

PR #6544, opened 2026-03-02
Reduce unnecessary SQL queries on every page load

pavel-mailpoet:
## Description

A customer reported that MailPoet runs 10+ SQL queries against `mailpoet_settings` on every frontend page load, adding ~0.41s overhead. This PR addresses two root causes:

1. **`SettingsController::set()` always writes to DB** — now captures the previous value before mutation and skips the write when unchanged
2. **`Bridge::storeKeyAndState()` unconditionally writes key state and subscription type** — added change detection guards so writes only happen when values differ

**Steady-state impact (key/state unchanged):** eliminates up to 8-10 redundant queries per page load.

## Code review notes

- PHP `===` for arrays is order-sensitive — if the same data has different key ordering, the write proceeds anyway. This is a harmless false positive, never a missed write.
- `$previousState` is initialized to `null` at the top of `storeKeyAndState()` so it can be reused when the data-preservation branch already fetched it, avoiding a duplicate `settings->get()` call.

## QA notes

Load a frontend page with Query Monitor enabled and verify the number of `mailpoet_settings` queries is reduced compared to trunk. The `mta` and `premium` INSERT+SELECT pairs should no longer appear when their values haven't changed.

## Linked PRs

_N/A_

## Linked tickets

STOMAIL-7859

## After-merge notes

_N/A_

## Screenshots or screencast

_N/A_

## Tasks

**Issues Addressed: 1**

**Performance:** Unnecessary SQL queries on every page load
- MailPoet executed 8–10+ redundant queries against `mailpoet_settings` on every frontend page load (≈0.41s overhead).
- Fixed by adding change-detection to:
  1. `SettingsController::set()` — skips DB writes when the value is unchanged.
  2. `Bridge::storeKeyAndState()` — writes key state and subscription type only when values differ; avoids duplicate settings->get() calls by reusing $previousState.

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:reduce-settings-sql-queries)

_The latest successful build from `reduce-settings-sql-queries` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/changelog/2026-03-02-08-31-22-improved-reduce-unnecessary-database-queries-on-every-page-.md
+++ b/mailpoet/changelog/2026-03-02-08-31-22-improved-reduce-unnecessary-database-queries-on-every-page-.md
@@ -1,0 +1,5 @@
+# Type: Improved
+
+# Description
+
+Reduce unnecessary database queries on every page load
--- a/mailpoet/lib/Services/Bridge.php
+++ b/mailpoet/lib/Services/Bridge.php
@@ -304,6 +304,7 @@ class Bridge {
     }
 
     $previousKey = $this->settings->get($keySettingName);
+    $previousState = null;
     // If the key remain the same and the new state is not valid we want to preserve the data from the previous state.
     // The data contain information about state limits. We need those to display the correct information to users.
     if (empty($state['data']) && $previousKey === $key) {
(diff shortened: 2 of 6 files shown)


lysyjan (approved):

Aschepikov (approved):
Image
Insert select pairs are fixed, no repeated writes

PR #6559, opened 2026-03-10
Fix step name popover not closing on outside click

triple0t:
## Description

Fix the automation step name popover not dismissing when clicking outside of it (e.g. on the canvas background). The WordPress `Popover` component intentionally skips closing when `activeElement` is `document.body`, which happens when clicking non-focusable areas. Added a supplementary `mousedown` listener on the document to cover that gap.

## Code review notes

The WordPress Popover's `onDialogClose` handler has a guard: `if (activeElement === document.body) return;` — this prevents closing when clicking non-focusable areas. Since we can't modify core, the fix adds a document-level `mousedown` listener inside the dropdown content that explicitly calls `onClose()` when the click target is outside both the popover and the dropdown toggle.

## QA notes

1. Go to the Automation Editor and select a step that has a "Step name" field (e.g. "Send email")
2. Click the pencil icon next to the step name to open the rename popover
3. **Without typing anything**, click on the canvas background area
4. Verify the popover closes
5. Re-open the popover, type something, then click outside — verify it still closes
6. Re-open the popover and click inside the text input — verify it stays open
7. Re-open the popover and click the pencil icon again — verify it toggles closed

## Linked PRs

_N/A_

## Linked tickets

[STOMAIL-7884](https://linear.app/a8c/issue/STOMAIL-7884/cant-dismiss-automation-step-popover-by-clicking-outside)

## After-merge notes

_N/A_

## Screenshots or screencast 

Before:
https://github.com/user-attachments/assets/621a56db-84d0-4e7d-b4d6-1062434f3f86

After:
https://github.com/user-attachments/assets/5eb32af7-4929-4a86-9830-dceedcc0ff8b

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:stomail-7884-cant-dismiss-automation-step-popover-by-clicking-outside)

_The latest successful build from `stomail-7884-cant-dismiss-automation-step-popover-by-clicking-outside` will be used. If none is available, the link won't work._

**1 issue found:**

**Performance**: The `handleClickOutside` callback includes `dropdownRef` in its dependency array (line 48), but since `dropdownRef` is a stable ref object, this inclusion is unnecessary. More significantly, if `onClose` changes on each render (which is likely given it comes from the Dropdown component's renderContent prop), the callback is recreated frequently, causing the useEffect (line 54) to repeatedly remove and re-add the document's mousedown event listener. This unnecessary setup/teardown cycle on every render impacts performance. Consider removing `dropdownRef` from the dependency array since refs maintain stable identity, or refactoring to reduce callback recreation.

### DIFF
--- a/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
@@ -32,12 +32,14 @@ function StepNameContent({
 
   const handleClickOutside = useCallback(
     (event: MouseEvent) => {
+      if (event.button !== 0) return;
+
       const target = event.target as Node;
-      const popoverEl = contentRef.current?.closest('.components-popover');
+      const contentEl = contentRef.current;
       const dropdownEl = dropdownRef.current;
 
       if (
-        (!popoverEl || !popoverEl.contains(target)) &&
+        (!contentEl || !contentEl.contains(target)) &&
         (!dropdownEl || !dropdownEl.contains(target))
       ) {
         onClose();
--- a/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
+++ b/mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
@@ -1,4 +1,5 @@
 import { Dropdown, TextControl } from '@wordpress/components';
+import { useCallback, useEffect, useRef } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { edit, Icon } from '@wordpress/icons';
 import { PlainBodyTitle } from './plain-body-title';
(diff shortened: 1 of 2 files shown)


mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
```diff
@@ -9,44 +10,102 @@ type Props = {
...
+        (!dropdownEl || !dropdownEl.contains(target))
+      ) {
+        onClose();
+      }
+    },
+    [onClose, dropdownRef],
+  );
+
+  useEffect(() => {
+    document.addEventListener('mousedown', handleClickOutside);
```
lysyjan, line 52: `mousedown` fires for all mouse buttons, so right-clicking the canvas will also close the popover. Adding `if (event.button !== 0) return;` would limit this to left clicks only.

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
```diff
@@ -9,44 +10,102 @@ type Props = {
...
+  update,
+  onClose,
+  dropdownRef,
+}: ContentProps): JSX.Element {
+  const contentRef = useRef<HTMLDivElement>(null);
+
+  const handleClickOutside = useCallback(
+    (event: MouseEvent) => {
+      const target = event.target as Node;
+      const popoverEl = contentRef.current?.closest('.components-popover');
```
lysyjan: `.closest('.components-popover')` relies on a WordPress internal class name. If WP changes it, popoverEl becomes null and the popover starts closing on every click, including inside the text input. Consider checking against `contentRef.current` directly - it's simpler and doesn't depend on WP internals.

lysyjan (approved):
It works as expected. I noticed only some details.

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
triple0t: Fixed in https://github.com/mailpoet/mailpoet/pull/6559/changes/bd12fe770f307d37766a38e808ec1eebbdd0cfcb

mailpoet/assets/js/src/automation/editor/components/panel/step-name.tsx
triple0t: Thanks for the feedback. Implemented in https://github.com/mailpoet/mailpoet/pull/6559/changes/bd12fe770f307d37766a38e808ec1eebbdd0cfcb

PR #6560, opened 2026-03-11
Use gh CLI for WooCommerce test plugin downloads

pavel-mailpoet:
## Description

Replace the shared fine-grained GitHub PAT (`WP_GITHUB_WOOCOMMERCE_TOKEN`) with `gh` CLI for downloading private WooCommerce test plugins (Subscriptions, AutomateWoo, Memberships).

**Problem:** Developers either reused the shared CI token locally (bad practice) or needed personal fine-grained tokens requiring WooCommerce org admin approval.

**Solution:** New `GhDownloader` class with two strategies:
- **Local dev:** Uses `gh` CLI — developers authenticate via `gh auth login` (OAuth/SSO, no tokens)
- **CI containers:** Falls back to `curl` with `GH_TOKEN` env var (set from existing CI secret)

## Code review notes

The `GhDownloader` curl fallback preserves the same asset name matching logic from the old `GithubClient` for the release download path. The `gh release download --pattern` handles this automatically for the `gh` CLI path.

## QA notes

1. Run `cd mailpoet && php vendor/bin/robo download:woo-commerce-subscriptions-zip latest` — should download via `gh`
2. Run `GH_TOKEN=<token> php vendor/bin/robo download:automate-woo-zip latest` with `gh` unavailable — should use curl fallback
3. Without any auth, should print skip message and exit 0

## Linked PRs

_N/A_

## Linked tickets

[STOMAIL-7772](https://linear.app/a8c/issue/STOMAIL-7772/update-mailpoet-test-plugin-download-to-use-gh)

## After-merge notes

Publish the announcement: https://mailpoetdev.wordpress.com/wp-admin/post.php?post=18840&action=edit

## Screenshots or screencast

_N/A_

## Tasks

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:update/stomail-7772-use-gh-cli-for-test-plugin-downloads)

_The latest successful build from `update/stomail-7772-use-gh-cli-for-test-plugin-downloads` will be used. If none is available, the link won't work._


triple0t (approved):
Looks good to me.

Turns out the `*-info` file isn't useful again because it only stores the asset URL and does not include the version.

| Before | After |
| ------ | ----- |
|   Image     |   Image    |

## AI Review notes

### Bug: `curlDownload` leaves a corrupt file at the destination on failure

**Severity: Medium** | **Confirmed by test**

When the curl fallback path fails (e.g., 401 auth error), `fopen($destPath, 'w')` has already written to the destination file. The exception is thrown but the file remains — in testing, it was 112 bytes of GitHub's JSON error response masquerading as a `.zip`.

**Location:** `mailpoet/tasks/GhDownloader.php`, lines 153–181

```php
private function curlDownload(string $url, string $destPath, string $accept): void {
    $retries = 0;
    while (true) {
      $fp = fopen($destPath, 'w');
      // ... curl writes error response to file ...
      throw new \Exception("Failed to download $url (HTTP $httpCode)");
      // file left at $destPath containing error response, not a valid zip
    }
}
```

**Fix:** Add `@unlink($destPath)` before throwing the exception.

---

### Bug: `curlGetJson` swallows 401/403 and produces a misleading error

**Severity: Medium** | **Confirmed by test**

When the curl fallback hits a 401 or 403 (expired token, wrong org access), `curlGetJson` silently returns `null`. The caller then throws `"Release $tag not found for $repo"` — which is wrong. The release exists, the token just doesn't have access. This will send developers down the wrong debugging path.

**Location:** `mailpoet/tasks/GhDownloader.php`, lines 125–151

```php
private function curlGetJson(string $url): ?array {
    // ...
    if ($httpCode >= 500 && $retries++ < self::RETRIES_COUNT) {
      // only 5xx is retried
    }
    return null;  // 401/403 silently returns null, no diagnostic
}
```

**Fix:** Throw an exception for 4xx errors that includes the HTTP status code:

```php
if ($httpCode === 401 || $httpCode === 403) {
    throw new \Exception("Authentication failed for $url (HTTP $httpCode). Check your GH_TOKEN.");
}
```

---

### Suggestion: Use `escapeshellarg()` consistently in `downloadRawFile`

**Severity: Low** | Not exploitable with current callers, but fragile pattern

`$repo`, `$filePath`, and `$ref` are interpolated inside single quotes without `escapeshellarg()`, while `$destPath` on the same line is properly escaped. Current callers all pass hardcoded safe strings, so this isn't exploitable today, but it's inconsistent with `downloadReleaseAssetWithGh` which correctly escapes all arguments.

**Location:** `mailpoet/tasks/GhDownloader.php`, lines 55–56

```php
if ($this->hasGh) {
    $this->exec("gh api 'repos/$repo/contents/$filePath?ref=$ref' -H 'Accept: application/vnd.github.v3.raw' > " . escapeshellarg($destPath));
}
```

**Fix:** Use the same `sprintf` + `escapeshellarg()` pattern as `downloadReleaseAssetWithGh`.

PR #6570, opened 2026-03-23
Pass GH_TOKEN to Docker test containers automatically

lysyjan:
## Description

Docker test containers cannot download private WooCommerce plugins (Subscriptions, Memberships, AutomateWoo) because `gh` CLI is not available inside the container and `GH_TOKEN` is not forwarded.

This adds `ensureGhToken()` to `runTestsInContainer` which:
1. Checks if `GH_TOKEN` is already set (CI environment)
2. Falls back to extracting the token from `gh auth token` on the host
3. Forwards it to Docker via `-e GH_TOKEN` (without exposing the value in command output)

This matches the approach used in CI (`.circleci/config.yml`) and fixes local test runs that previously failed with "Skipping download — not authenticated".

## Code review notes

- The token is set via `putenv()` + `$_ENV` so Symfony Process (used by Robo) inherits it
- `-e GH_TOKEN` (without `=value`) tells Docker to forward the host env var, so the token never appears in the printed command
- If neither `GH_TOKEN` nor `gh` CLI is available, the flag is simply omitted (no change in behavior)

## QA notes

1. Reset the test Docker environment (this also deletes cached plugin zips): `./do reset:test-docker`
2. Run `./do test:integration --skip-deps` from `mailpoet/`
3. Verify WooCommerce Subscriptions, Memberships, and AutomateWoo are downloaded and installed in the Docker entrypoint
4. Verify the token is NOT visible in the printed command (only `-e GH_TOKEN` without the value)

## Linked PRs

- https://github.com/mailpoet/mailpoet-premium/pull/1018 — companion PR with the same change for the premium plugin

## Linked tickets

_N/A_

## After-merge notes

_N/A_

## Tasks

- [x] I have added a changelog entry — N/A, internal tooling change
- [x] I followed best practices for translations — N/A
- [ ] I added sufficient test coverage — build tooling, not unit-testable
- [x] I embraced TypeScript — N/A, PHP only

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:pass-gh-token-to-docker-tests)

_The latest successful build from `pass-gh-token-to-docker-tests` will be used. If none is available, the link won't work._

### DIFF
--- a/mailpoet/RoboFile.php
+++ b/mailpoet/RoboFile.php
@@ -1557,6 +1557,21 @@ class RoboFile extends \Robo\Tasks {
     return $exitCode;
   }
 
+  private function ensureTestPlugins(): void {
+    $pluginsDir = __DIR__ . '/tests/plugins';
+    $downloads = [
+      'woocommerce.zip' => 'download:woo-commerce-zip latest',
+      'woocommerce-subscriptions.zip' => 'download:woo-commerce-subscriptions-zip latest',
+      'woocommerce-memberships.zip' => 'download:woo-commerce-memberships-zip',
+      'automatewoo.zip' => 'download:automate-woo-zip latest',
+    ];
+    foreach ($downloads as $zip => $command) {
+      if (!file_exists("$pluginsDir/$zip")) {
+        $this->taskExec("./do $command")->dir(__DIR__)->run();
+      }
+    }
+  }
+
   private function createGhDownloader(): ?\MailPoetTasks\GhDownloader {
     require_once __DIR__ . '/tasks/GhDownloader.php';
     $downloader = new \MailPoetTasks\GhDownloader();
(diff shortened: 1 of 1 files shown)


pavel-mailpoet (changes requested):
After this pull request, the `GhDownloader` now has dead code. The code that downloads the packages using the `gh` command is never used. So we have two options:

1. Use the original idea and download the packages BEFORE running the tests outside of the Docker container, and remove the `ensureGhToken` 
2. Go with the improvement in this Pull Request and pass the TOKEN in the Docker container. But in this case, we will never use the `gh` command, so we should remove the requirements in the README and AGENTS files and remove the dead code in `GhDownloader`

lysyjan:
I would prefer downloading plugins outside Docker to be automated. I will update the PR.

pavel-mailpoet (approved):
Image

PR #6573, opened 2026-03-25
Fix: remove 'required' asterisk on email label, and remove space before others on Manage Subs page

dphyled:
## Description

Fix inconsistent copy on Manage Subscription page

Fixes two cosmetic issues on the subscriber-facing Manage Subscription page:

1. Email label asterisk — The email address was displayed with a * next to its label, suggesting it's a required input field. Since the email field is read-only on this page, the asterisk was misleading and has been removed.
2. Status label spacing — There was an extra space between the "Status" label text and its required-field asterisk (rendering as "Status " instead of "Status"). Removed the leading space in BlockRendererHelper::renderLabel() and renderLegend(), which affects all form labels consistently.

Unit tests updated to reflect the corrected output.

Closes STOMAIL-7748

## Tasks

- [x] I added sufficient test coverage (existing test updated)

## Preview

[Preview in WordPress Playground](https://account.mailpoet.com/playground/new/branch:fix/stomail-7748-inconsistent-manage-subscription-page-copy)

_The latest successful build from `fix/stomail-7748-inconsistent-manage-subscription-page-copy` will be used. If none is available, the link won't work._

## Issues Addressed

**Number of issues: 2**

**Categories:**
- Bug (2): Cosmetic visual inconsistencies on the Manage Subscription page
  1. Removes the required-field asterisk from the email label since the field is read-only
  2. Corrects inconsistent spacing before required-field asterisks by replacing regular spaces with non-breaking spaces (&nbsp;) for visual consistency across forms


mailpoet/lib/Form/Block/BlockRendererHelper.php
```diff
@@ -145,7 +145,7 @@ public function renderLabel(array $block, array $formSettings): string {
       $html .= static::getFieldLabel($block);
 
       if (static::getFieldIsRequired($block)) {
-        $html .= ' <span class="mailpoet_required" aria-hidden="true">*</span>';
```
lysyjan, line 148: I think we should keep the space here and below. It's better for readability, and, as I know, it's also best practice. This approach is also used in the editor and other pages.

lysyjan (commented):
Thank you for working on that 👍

I think we should remove only the asterisk for the email label and keep the space for the rest of the product.

For the next time, please use the pull-request template that is part of the repository.
Also, adding testing instructions is helpful :)

mailpoet/lib/Form/Block/BlockRendererHelper.php
dphyled: @lysyjan thank you for the PR review!! 🙇🏼‍♂️ 

Just want to double check: in [STOMAIL-7748](https://linear.app/a8c/issue/STOMAIL-7748), the two changes were 1) remove asterisk on Email label, and 2) "Remove the whitespace between the 'Status' label and asterisk".  Are we still doing #2, but leaving the space for required fields other places in the plugin? Or completely cancelling #2 and keeping the space everywhere?

dphyled: In Woo, looks like it includes a `&nbsp;` before the asterisk for required fields. Maybe should match that? (And keep the space between the 'Status' label and asterisk // cancel #2 from the Linear issue)
https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/wc-template-functions.php#L3201

lysyjan: > In Woo, looks like it includes a &nbsp; before the asterisk for required fields. Maybe should match that? (And keep the space between the 'Status' label and asterisk // cancel #2 from the Linear issue)

@dphyled That sounds good to me 👍 Can you update the PR?

lysyjan (approved):

